PR #713, opened 2023-09-08
Some cleanup work on ExpressionEditor

handeyeco:
## Summary:
Some Friday cleanup: removing prop types and most of underscore


packages/perseus-editor/src/widgets/expression-editor.tsx
```diff
@@ -9,22 +9,28 @@ import {
...
-    considered: PropTypes.oneOf(PerseusExpressionAnswerFormConsidered)
-        .isRequired,
-    value: PropTypes.string.isRequired,
-    form: PropTypes.bool.isRequired,
-    simplify: PropTypes.bool.isRequired,
-});
+type Props = {
+    widgetId?: any;
+    value?: string;
+} & Omit<PerseusExpressionWidgetOptions, "buttonsVisible"> &
```
jeremywiebe, line 25: Interesting that we don't use this `buttonsVisible` prop.  Looking around, my guess is this is historical and not actually used in our Perseus data... but rather a side effect that bubbled up from the `<MathInput>` that also has this prop.  

Not asking for change, just pondering.

packages/perseus-editor/src/widgets/expression-editor.tsx
```diff
@@ -282,7 +283,7 @@ class ExpressionEditor extends React.Component<any, any> {
         if (this.props.answerForms.length === 0) {
             issues.push("No answers specified");
         } else {
-            const hasCorrect = !!_(this.props.answerForms).find((form) => {
+            const hasCorrect = !!this.props.answerForms.find((form) => {
```
jeremywiebe, line 286: This could use `.some()` instead of `.find()` and get rid of the `!!` prefix trick to convert the numeric value to a boolean.

packages/perseus-editor/src/widgets/expression-editor.tsx
```diff
@@ -426,7 +422,7 @@ class ExpressionEditor extends React.Component<any, any> {
         e,
     ) => {
         const newProps: Record<string, any> = {};
-        newProps.functions = _.compact(e.target.value.split(/[ ,]+/));
+        newProps.functions = e.target.value.split(/[ ,]+/).filter(Boolean);
```
jeremywiebe, line 425: `.filter(Boolean)` is not correct in the new Typescript world (see https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/2201682693/TypeScript+for+Flow+Developers#Predicate-Functions-vs-Type-Predicates). 

```suggestion
        newProps.functions = e.target.value.split(/[ ,]+/).filter(isTruthy /* from wonder-stuff */);
```

jeremywiebe (approved):
Nice. Love these little improvements. Thanks!

One thing that should be fixed before landing is the use of `.filter(Boolean)`.

PR #714, opened 2023-09-08
ErrorBoundary additions to log widget type and component stack so these show up in sentry

pmcgill88:
## Summary:
Same as above! ~Guessing a change like this doesn't need a new changeset??~ Ok added a changeset.

Issue: LC-679

## Test plan:

### DIFF
--- a/.changeset/afraid-lizards-arrive.md
+++ b/.changeset/afraid-lizards-arrive.md
@@ -2,4 +2,4 @@
 "@khanacademy/perseus": minor
 ---
 
-Updated ErrorBoundary logging to include the componenetStack as well as the widget type as metadata.
+Updated ErrorBoundary logging to include the componentStack as well as the widget type as metadata.
--- a/.changeset/afraid-lizards-arrive.md
+++ b/.changeset/afraid-lizards-arrive.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": minor
+---
+
+Updated ErrorBoundary logging to include the componenetStack as well as the widget type as metadata.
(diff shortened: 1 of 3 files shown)


.changeset/afraid-lizards-arrive.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/perseus": minor
+---
+
+Updated ErrorBoundary logging to include the componenetStack as well as the widget type as metadata.
```
jeremywiebe, line 5: For trivial changes there is the option to run `yarn changeset --empty`, but I think it's safer to add a changeset for posterity. 

```suggestion
Updated ErrorBoundary logging to include the componentStack as well as the widget type as metadata.
```

packages/perseus/src/renderer.tsx
```diff
@@ -550,7 +550,10 @@ class Renderer extends React.Component<Props, State> {
             // filtered out in this.render(), so we shouldn't have to
             // worry about using this widget key and ref:
             return (
-                <ErrorBoundary key={"container:" + id}>
+                <ErrorBoundary
+                    key={"container:" + id}
+                    metadata={{widget_type: type}}
```
jeremywiebe, line 555: Looking a little more at this, I think it'd be useful to add the widget id also here (the `id` variable in scope).

```suggestion
                    metadata={{widget_type: type, widget_id: id}}
```

jeremywiebe (approved):
Love this little addition to our logging, I think it'll make the Sentry reports quite a bit more usable.

PR #715, opened 2023-09-10
Fix problems in image widget's caption

nixterrimus:
## Summary:
Fixes two problems:

- In content admin the caption overlapped with the image
- For articles the caption and image became misaligned.

Screen Shot 2023-09-10 at 10 29 15 AM
Screen Shot 2023-09-10 at 10 29 12 AM
Screen Shot 2023-09-10 at 10 29 08 AM
Screen Shot 2023-09-10 at 10 29 06 AM

Issue: https://khanacademy.atlassian.net/browse/LC-1214, https://khanacademy.atlassian.net/browse/LC-1215

## Test plan:
- Compare before/after
- **things should look better**

### DIFF
--- a/packages/perseus/src/widgets/__tests__/__snapshots__/image.test.ts.snap
+++ b/packages/perseus/src/widgets/__tests__/__snapshots__/image.test.ts.snap
@@ -75,6 +75,7 @@ exports[`image widget - isMobile %b should snapshot: first render 1`] = `
                   </span>
                   <figcaption
                     class="perseus-image-caption has-title"
+                    style="max-width: 420px;"
                   >
                     <div>
                       <div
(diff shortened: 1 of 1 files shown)


packages/perseus/src/styles/articles.less
```diff
@@ -121,7 +121,6 @@
         font-size: 14px;
         line-height: 19px;
         margin: 16px auto 42px;
-        max-width: 455px;
```
jeremywiebe, line 124: **thought**: I agree with removing this, but I always wonder where some of these seemingly arbitrary rules came from. :)

jeremywiebe (approved):
Nice. That's looking alot better! Thank-you!

PR #707, opened 2023-09-07
[Perseus Analytics] Add and implement 'perseus:expression-focused' event

jeremywiebe:
## Summary:

This PR implements a new analytics event: `perseus:expression-focused`. This event is fired whenever the `expression` widget input receives focus. 

Issue: LC-861

## Test plan:

`yarn test`
`yarn typecheck`

PR #719, opened 2023-09-11
Bump perseus-core (missed in previous release)

jeremywiebe:
## Summary:

In #707 I introduced a new analytics event (`perseus:expression-focused`) and implemented in Perseus. I missed marking `@khanacademy/perseus-core` as having a change in the changeset and none of our tooling warned, so the release went out without bumping that package, or the dependencies that used it. 

This PR is an attempt to just get core to version, which should version the packages that depend on it. 

Issue: LC-861

## Test plan:

`yarn test`

### DIFF
--- a/.changeset/poor-bikes-complain.md
+++ b/.changeset/poor-bikes-complain.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-core": minor
+---
+
+Added new analytics event "perseus:expression-focused"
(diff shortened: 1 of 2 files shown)


handeyeco (approved):
If it works, this looks good. Curious if the `changeset` tool requires code to change though.

jeremywiebe:
> If it works, this looks good. Curious if the `changeset` tool requires code to change though.

It might. I think I'm going to add a change that changes a comment in the affected file just to be sure!

PR #720, opened 2023-09-12
[math-input] Organize storybook stories

jeremywiebe:
## Summary:

This PR organizes the stories in the math-input package so that they're nicely grouped under a `Math Input` header in Storybook. It also moves two files so that they're slightly more obvious when you see them. 

See screenshots:

| Before | After |
|--------|-------|
| Before |  After |

Issue: "none"

## Test plan:


nedredmond (approved):
yes pls

PR #722, opened 2023-09-12
Added logic to convertDotToTimes for the v2 keypad.

SonicScrewdriver:
## Summary:
Ensured that we're getting the convertDotToTimes value from content's "times" so that we can display the intended multiplication symbol in the new v2 keypad. I also updated some variable names to maintain some consistency. 

Issue: LC-1092

## Test plan:
- Created new tests to ensure that the button changes as desired.


packages/math-input/src/components/keypad/keypad-mathquill.stories.tsx
```diff
@@ -78,7 +78,7 @@ export function V2KeypadWithMathquill() {
                             basicRelations
                             divisionKey
                             logarithms
-                            multiplicationDot
+                            convertDotToTimes
```
SonicScrewdriver, line 81: The three variable names sounded confusing as they were: `convertDotToTimes`, `multiplicationDot`, and `times`. multiplicationDot seems like it would technically be the inverse of the original variable logic (times), so I've renamed things in order to maintain readability.

packages/perseus/src/widgets/expression.tsx
```diff
@@ -610,11 +610,13 @@ export class Expression extends React.Component<Props, ExpressionState> {
  */
 const keypadConfigurationForProps = (
     widgetOptions: PerseusExpressionWidgetOptions,
-) => {
+): KeypadConfiguration => {
```
SonicScrewdriver, line 613: Added a type as we were missing one.

packages/perseus/src/widgets/expression.tsx
```diff
@@ -642,20 +644,20 @@ const keypadConfigurationForProps = (
 
     // TODO(charlie): Alert the keypad as to which of these symbols should be
     // treated as functions.
-    const extraVariables = Object.keys(uniqueExtraVariables);
+    const extraVariables: Key[] = Object.keys(uniqueExtraVariables) as Key[];
```
SonicScrewdriver, line 647: Necessary changes to support the new return type.

SonicScrewdriver:
NOTE : This will not be landed until given the all-clear, as we're currently in a code freeze.

packages/perseus/src/widgets/expression.tsx
handeyeco: This seems suspicious to me; if we need to coerce this at all, I think it needs to be `ReadonlyArray<Key>` rather than `Key[]`. I imagine the source of the issue is elsewhere though...

Oh this might be part of the problem:

``` JS
    if (!extraKeys.length) {
        // If there are no extra symbols available, we include Pi anyway, so
        // that the "extra symbols" button doesn't appear empty.
        extraKeys.push("PI");
    }
```

We won't be using the long-press "extra symbols" button in v2 keypad (though we still need it while v1 is alive). 🤔 Anyway, I think we could still figure out a way to make it `ReadonlyArray<Key>`, otherwise I wonder if this is a net loss: we made `keypadConfigurationForProps` more explicit, but had to add a bunch of `Key[]` which we typically try to avoid.

packages/math-input/src/index.ts
```diff
@@ -41,5 +41,8 @@ export type {default as Keys} from "./data/keys";
 export {default as KeyConfigs} from "./data/key-configs";
 export {type KeyType, KeypadType} from "./enums";
 
+// Keypad Configuration type
+export {type KeypadConfiguration} from "./types";
```
handeyeco, line 45: We already have `export type {KeypadAPI} from "./types";`, could we add `KeypadConfiguration` to that?

packages/math-input/src/full-math-input.stories.tsx
```diff
@@ -25,11 +25,17 @@ export const Basic = () => {
...
     // Whether to use Expression or Fraction keypad
     const [expression, setExpression] = React.useState<boolean>(false);
+    // Whether to use CDOT or TIMES
+    const [times, setTimes] = React.useState<boolean>(true);
     // Whether to use v1 or v2 keypad
     const [v2Keypad, setV2Keypad] = React.useState<boolean>(true);
     // Whether the keypad is open or not
     const [keypadOpen, setKeypadOpen] = React.useState<boolean>(false);
 
+    const input = React.useRef<any>(null);
```
handeyeco, line 35: Would something like this work?
```suggestion
    const input = React.useRef<typeof KeypadInput>(null);
```

handeyeco (approved):
My main asks:
- Confirm that existing code uses `convertDotToTimes`, `multiplicationDot`, and `times` interchangeably and that they're not subtly different.
- Explore options for using `ReadonlyArray<Key>` vs `Key[]`

PR #724, opened 2023-09-13
Fix check to _not_ run npm snapshot when on Version Packages PRs

jeremywiebe:
## Summary:

We've had an ongoing issue where the changeset action that produces a snapshot npm version for PRs also runs on the `Version Packages` PR, but fails. This is because our check for changesets finds none on those PRs. In these special `Version Packages` PRs, the changeset tool does all the work needed to get the affected packages ready for release and deletes all the changeset files. 

This PR fixes the issue. We (Juan, Marian, and I) [suspect](https://khanacademy.slack.com/archives/C4PE1QM5Y/p1694626375783619?thread_ts=1694619602.325429&cid=C4PE1QM5Y) that the `base_ref` is not the correct Github context variable to reference and instead are switching to use `head_ref` (which is already [in use](https://github.com/Khan/wonder-blocks/blob/fe03ac160334f19dd04e56fade97c4f2a0c80115/.github/workflows/chromatic-pr.yml#L10-L11) and seems to work properly in the Wonder Blocks repository)

Issue: LC-1091

## Test plan:

Do a release and make sure that Github doesn't try to run the npm snapshot step for the Version Packages PR.


.github/workflows/node-ci.yml
```diff
@@ -230,7 +230,7 @@ jobs:
         # on the main Changeset "Version Packages" PR
         if: |
             github.event.pull_request.draft == false
-            && github.ref_name != 'changeset-release/main'
+            ! startsWith(github.head_ref, 'changeset-release/')
```
jeremywiebe, line 233: See: https://github.com/Khan/wonder-blocks/blob/21a6c0e5feae8b0728e7e48a8906a08b9c55fccf/.github/workflows/chromatic-pr.yml#L11

.github/workflows/node-ci.yml
```diff
@@ -230,7 +230,7 @@ jobs:
         # on the main Changeset "Version Packages" PR
         if: |
             github.event.pull_request.draft == false
-            && github.ref_name != 'changeset-release/main'
+            !startsWith(github.head_ref, 'changeset-release/')
```
jandrade, line 233: Maybe adding `&&` would be still needed here?

nedredmond (approved):

.github/workflows/node-ci.yml
jeremywiebe: Doh! Yes, it would!!

jandrade (approved):
Makes sense to me! hoping it will work with the "Version Packages" PR 🤞

PR #727, opened 2023-09-15
Add alert for failure in publishing workflow

nedredmond:
Adds an alert for a failure in the publishing workflow.

Hard to test. I guess we just wait for it to fail?

For reference:
`failure()` (if any steps above fail): https://docs.github.com/en/actions/learn-github-actions/expressions#failure
`github.repository` (returns "Khan/perseus"): https://docs.github.com/en/actions/learn-github-actions/contexts#github-context


.github/workflows/release.yml
```diff
@@ -80,7 +94,7 @@ jobs:
                   SLACK_MSG_AUTHOR: ${{ github.event.pull_request.user.login }}
                   SLACK_USERNAME: GithubGoose
                   SLACK_ICON_EMOJI: ":goose:"
-                  SLACK_MESSAGE: "A new version of ${{ github.event.repository.name }} was published! 🎉 \nRelease notes → https://github.com/Khan/${{ github.event.repository.name }}/releases/"
+                  SLACK_MESSAGE: "A new version of ${{ github.event.repository.name }} was published! 🎉 \nRelease notes → https://github.com/${{ github.repository }}/releases/"
```
jeremywiebe, line 97: Oh, nice little improvement!

.github/workflows/release.yml
```diff
@@ -71,6 +71,20 @@ jobs:
...
+            - name: Send a Slack notification if a publish fails
+              if: failure()
+              uses: rtCamp/action-slack-notify@v2
+              env:
+                  SLACK_WEBHOOK: ${{ secrets.SLACK_EXERCISENEXT_WEBHOOK }}
+                  SLACK_CHANNEL: team-lc-perseus
+                  SLACK_MSG_AUTHOR: ${{ github.event.pull_request.user.login }}
+                  SLACK_USERNAME: GithubGoose
+                  SLACK_ICON_EMOJI: ":goose:"
+                  SLACK_MESSAGE: "Something went wrong with the release of ${{ github.event.repository.name }}. \nPlease check the logs → https://github.com/${{ github.repository }}/actions"
```
jeremywiebe, line 83: **idea**: Are we able to get a link directly to this action run?  Looks like it might be possible to build using the `github.run_id`.  

This PR is an improvement overall so if that's not possible, I'm fine landing this PR as is.

jeremywiebe (approved):
Thanks for this!  I have one non-blocking comment on how we might improve this just a touch more.

.github/workflows/release.yml
nedredmond: @jeremywiebe I will try! If we end up with a bad link we can fix it later

jeremywiebe (approved):

PR #728, opened 2023-09-18
Add hasBeenActivated to v2 keypad

handeyeco:
## Summary:
I did this in another PR but then removed it because I thought Third had already solved the issue.

Basically we just need to add the `visibility: hidden` to `dynamicStyle` so it's hidden from the start.

Issue: LC-1273

## Test plan:
- Emulate mobile
- Turn on v2 keypad
- Go to a page with MathInput
- The keypad should not flicker during the initial load

Before:

https://github.com/Khan/perseus/assets/16308368/e9a63104-ba88-4492-b09b-63e851888d57

After:

https://github.com/Khan/perseus/assets/16308368/d7b58289-8545-4e2f-b289-0a0bd7bdd928

### DIFF
--- a/packages/math-input/src/components/keypad/mobile-keypad.tsx
+++ b/packages/math-input/src/components/keypad/mobile-keypad.tsx
@@ -39,6 +39,7 @@ type Props = {
 type State = {
     active: boolean;
     containerWidth: number;
+    hasBeenActivated: boolean;
     keypadConfig?: KeypadConfiguration;
     keyHandler?: KeyHandler;
     cursor?: Cursor;
@@ -53,6 +54,7 @@ class MobileKeypad extends React.Component<Props, State> implements KeypadAPI {
     state: State = {
         active: false,
         containerWidth: 0,
+        hasBeenActivated: false,
     };
 
     componentDidMount() {
@@ -107,7 +109,10 @@ class MobileKeypad extends React.Component<Props, State> implements KeypadAPI {
     };
 
     activate: () => void = () => {
-        this.setState({active: true});
+        this.setState({
+            active: true,
+            hasBeenActivated: true,
+        });
     };
 
     dismiss: () => void = () => {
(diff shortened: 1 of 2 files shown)


SonicScrewdriver (approved):
Thank you greatly for this! Looks great to me, LGTM :)

packages/math-input/src/components/keypad/mobile-keypad.tsx
```diff
@@ -158,12 +163,18 @@ class MobileKeypad extends React.Component<Props, State> implements KeypadAPI {
...
         const {style} = this.props;
-        const {active, containerWidth, cursor, keypadConfig} = this.state;
+        const {active, hasBeenActivated, containerWidth, cursor, keypadConfig} =
+            this.state;
 
         const containerStyle = [
             // internal styles
             styles.keypadContainer,
             active && styles.activeKeypadContainer,
+            // If the keypad is yet to have ever been activated, we keep it invisible
```
jeremywiebe, line 173:
```suggestion
            // If the keypad is yet to have never been activated, we keep it invisible
```

packages/math-input/src/components/keypad/mobile-keypad.tsx
```diff
@@ -158,12 +163,18 @@ class MobileKeypad extends React.Component<Props, State> implements KeypadAPI {
...
+            // so as to avoid, e.g., the keypad flashing at the bottom of the page
+            // during the initial render.
+            // Done inline since stylesheets might not be loaded yet.
+            !hasBeenActivated && {visibility: "hidden"},
```
jeremywiebe, line 177: I don't understand how this fixes things.  The base `styles.keypadContainer` already has `visibility:hidden` on it. So if `active == false`, shouldn't that be the `visibility` style we have already?  

image

jeremywiebe (commented):
I have questions. :)

packages/math-input/src/components/keypad/mobile-keypad.tsx
handeyeco: I just copy/pasted and the wording is a little weird, but I think the original is more correct.

packages/math-input/src/components/keypad/mobile-keypad.tsx
handeyeco: Sorry, I pushed this to trigger a snapshot to test in Webapp but I didn't originally mark this as a WIP.

It needs to be added to `dynamicStyle` so that it happens before stylesheets are applied.

packages/math-input/src/components/keypad/mobile-keypad.tsx
jeremywiebe: Oooooh, it was my misreading!! 🙃

PR #729, opened 2023-09-18
Only check for changesets when there are changes to a package

nedredmond:
Changes glob pattern to only check for changesets when there is a change to a package. That way, we don't have to submit empty changesets for changes that would not increment the semver for any Perseus package.

LC-1280


jeremywiebe (approved):

.github/workflows/node-ci.yml
```diff
@@ -21,9 +21,9 @@ jobs:
             - uses: Khan/actions@check-for-changeset-v0
               if: github.actor != 'dependabot[bot]' && github.actor != 'dependabot-preview[bot]'
               with:
-                  exclude: .github/,.storybook/
                   exclude_extensions: .test.ts, .test.tsx, .stories.ts, .stories.tsx
-                  exclude_globs: "**/__tests__/*"
+                  # Inverted match for ./packages to exclude everything but packages
+                  exclude_globs: "**/__tests__/*, **/__stories__/*, **/dist/*, !(./packages/**/*.{ts,tsx,js,html,css,md})"
```
nedredmond, line 26: I have a feeling that this is supposed to be `!(packages...`
and SHOULD include `json`. Wdyt @jeremywiebe ?

jeremywiebe: Yes, I'd agree.  In cases where a package is changed, all dependencies are updated by `changeset` during the release.  But I think we'd want to force a changeset/release for _all_ package.json changes to be safe (ie. if we bump a dependency, we want to release that package).

PR #732, opened 2023-09-19
Bugfix: make FRAC buttons use the same FRAC

handeyeco:
## Summary:
v1 keypad was using `FRAC` and v2 keypad was using `FRAC_INCLUSIVE`.

Issue: LC-1267

## Test plan:
- Go to v1 keypad in "Expression" mode
- Type a number
- Hit the fraction button
- Note that the number does not become the numerator
- Go to v2 keypad in "Expression" mode
- Type a number
- Hit the fraction button
- Note that the number does not become the numerator (previously it did)

### DIFF
--- a/.changeset/fresh-countries-sleep.md
+++ b/.changeset/fresh-countries-sleep.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/math-input": patch
+---
+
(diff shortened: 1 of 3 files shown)


nedredmond (approved):
Nice find!

PR #733, opened 2023-09-20
FEI-5298: Replace Coveralls with Codecov

kevinbarabash:
## Summary:
We've found Coveralls to be too flaky to be useful.  It often hangs and sometimes it reports drops in coverages in types-only changes.

This PR switches back to Codecov but only uploads a single report for jest.  In the future we'll also generate coverage for cypress and merge the reports ourselves.  This PR is just to get a report of some sort showing up in the PRs.

Issue: FEI-5298

## Test plan:
- make a trivial change to a source code file, see that the report is uploaded to Codecov


packages/perseus/src/widgets/__tests__/expression.test.tsx
```diff
@@ -180,7 +180,7 @@ describe("Expression Widget", function () {
         });
     });
 
-    describe("invalid", function () {
+    describe.skip("invalid", function () {
```
jeremywiebe, line 183: Can we bring this back or is it flaky/broken in some way?

.prettierignore
```diff
@@ -30,6 +30,7 @@ LICENSE
 yarn.lock
 
 coverage/
+coverage-ts/
```
jeremywiebe, line 33: I don't see this path used anywhere in this PR. Does the codecov-action just write to here by default?

jeremywiebe (approved):
Thanks for flipping this back. Hopefully it's less hangy.

packages/perseus/src/widgets/__tests__/expression.test.tsx
kevinbarabash: I'll renable this before landing.

.prettierignore
kevinbarabash: It's possible it's something that's only on my machine because I was playing around with some other tooling.  I get rid of it and delete the dir locally.

.github/workflows/node-ci-main.yml
```diff
@@ -51,17 +51,14 @@ jobs:
...
-              # We don't bother uploading a record of coverage for dependabot changes
-              if: ${{ github.actor != 'dependabot[bot]' }}
-              uses: coverallsapp/github-action@v1
+              uses: codecov/codecov-action@v3
               with:
-                  github-token: ${{ secrets.GITHUB_TOKEN }}
-                  flag-name: jest
-                  parallel: true
+                  token: ${{ secrets.CODECOV_TOKEN }}
+                  files: ./coverage/coverage-final.json
```
kevinbarabash, line 57: @jeremywiebe should I set `continue-on-error: true` for this step?  I know some people had concerns about coverage drops show up as red `x`s on PRs.  Personally, I think we should be using required checks for those checks that we think are important enough to block merging.

jeremywiebe: 🤔  I don't think we should set `continue-on-error` to begin with. I'd rather tune the `target` value in our `codecov.yml` file if it's too sensitive.  

@handeyeco We can discuss and [disable this check](https://docs.codecov.com/docs/commit-status#informational) but still get the information even after Kevin lands this PR.

kevinbarabash: Things should fall in the green zone with the current settings once the cypress coverage results are merged in.

PR #735, opened 2023-09-21
Verify that there is a changeset for every changed package

nedredmond:
Switches out `check-for-changeset`, a github action that looks for the presence of a changeset file, for `changeset-per-package`, a new action that verifies that there is a changeset entry for each package with changes.

For details, see [the action repo](https://github.com/Khan/changeset-per-package).

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -16,14 +16,26 @@ jobs:
                 os: [ubuntu-latest]
                 node-version: [16.x]
         steps:
-            # We don't need changesets for infrastructure-type files (tests, stories,
-            # etc).
-            - uses: Khan/changeset-per-package@v0.0.0
-              if: github.actor != 'dependabot[bot]' && github.actor != 'dependabot-preview[bot]'
+            - name: Checkout repository
+              uses: actions/checkout@v4
               with:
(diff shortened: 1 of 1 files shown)


nedredmond:
Looks like we have an issue with using the action outside of its own repo. Probably an easy fix, but I'll close this for now since it's EOD.

nedredmond:
I'm just waffling back and forth on filters. I think I've got it now. Just tests and stories.

.github/workflows/node-ci.yml
```diff
@@ -9,21 +9,35 @@ on:
...
             matrix:
                 os: [ubuntu-latest]
                 node-version: [16.x]
         steps:
-            # We don't need changesets for infrastructure-type files (tests, stories,
-            # etc).
-            - uses: Khan/actions@check-for-changeset-v0
-              if: github.actor != 'dependabot[bot]' && github.actor != 'dependabot-preview[bot]'
+            - name: Checkout repository
+              uses: actions/checkout@v4
```
jeremywiebe, line 21: Oh, cool!  v4 now?  That must be a recent release.

jeremywiebe (approved):
Let's start using this! Thanks Ned!

PR #737, opened 2023-09-21
Add some unit tests for mobile MathInput

handeyeco:
## Summary:
Some tests for MathInput / Keypad integration

## Test plan:
- Run tests
- Note they pass


packages/math-input/src/components/__tests__/integration.test.tsx
```diff
@@ -0,0 +1,183 @@
...
+    );
+}
+
+describe("math input integration", () => {
+    it("renders", () => {
+        render(<ConnectedMathInput />);
+
+        expect(
+            screen.getByLabelText(
+                "Math input box Tap with one or two fingers to open keyboard",
```
SonicScrewdriver, line 85: I wonder if we should fix this label text as it seems messy. I wonder how it reads out in a screen reader. I know you didn't change this as part of your PR, just noticing that this looks strange. :)

packages/math-input/src/components/__tests__/integration.test.tsx
```diff
@@ -0,0 +1,183 @@
...
+        const input = screen.getByLabelText(
+            "Math input box Tap with one or two fingers to open keyboard",
+        );
+
+        fireEvent.touchStart(input);
+        userEvent.click(screen.getByRole("button", {name: "1"}));
+
+        // MathQuill is problematic,
+        // this is the only way I know how to test the "input"
+        const mathquillInput =
```
SonicScrewdriver, line 122: Makes sense! You might also be able to get a ref to the MathInput or check that setValue is called (since you're providing the method anyway) to get access to mathquill that way. Not necessary imo but might help clean things up.

packages/perseus/src/widgets/__tests__/expression-mobile.test.tsx
```diff
@@ -0,0 +1,350 @@
...
+        userEvent.click(screen.getByRole("tab", {name: "Numbers"}));
+        userEvent.click(screen.getByRole("button", {name: "9"}));
+
+        // MathQuill is problematic,
+        // this is the only way I know how to test the "input"
+        const mathquillInput =
+            // eslint-disable-next-line testing-library/no-node-access
+            document.getElementsByClassName("mq-root-block")[0] as HTMLElement;
+
+        expect(within(mathquillInput).getByText("x")).toBeVisible();
```
SonicScrewdriver, line 341: NIT: Hmm does this test that the equation was typed correctly / in the correct order? Would this pass if it said sin()9=x? Totally an ignorable nit as I'm trying to think of things to provide commentary on, but is there a way to verify the equation as written?

SonicScrewdriver (approved):
Looks great! I left a couple comments but they're all ignorable nits. Thank you for adding these tests!

packages/math-input/src/components/__tests__/integration.test.tsx
SonicScrewdriver: Mostly it seems like it needs some punctuation. "Math input box. Tap with one or two fingers to open keyboard."

packages/perseus/src/widgets/__tests__/expression-mobile.test.tsx
```diff
@@ -0,0 +1,350 @@
...
+            testDependencies,
+        );
+    });
+
+    afterEach(() => {
+        // The Renderer uses a timer to wait for widgets to complete rendering.
+        // If we don't spin the timers here, then the timer fires in the test
+        // _after_ and breaks it because we do setState() in the callback,
+        // and by that point the component has been unmounted.
+        jest.runOnlyPendingTimers();
```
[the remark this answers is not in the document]
jeremywiebe: :sadness:  Be nice if our component tore down the timer on unmount.

packages/perseus/src/widgets/__tests__/expression-mobile.test.tsx
jeremywiebe: I wonder if we could use the HTML element to [get a handle](http://docs.mathquill.com/en/latest/Api_Methods/#mqhtml_element) to the Mathquill API and then [ask it](http://docs.mathquill.com/en/latest/Api_Methods/#latex) for the `.latex()` ... and use that to verify what's in it.  This test works, but feels clunky and potentially difficult to maintain.

jeremywiebe (approved):
Nice, I like that we will now have integration tests for input+keypad! 

The one thing that I'd like to consider is if we can use the Mathquill API to check the input value instead of digging at the HTML directly.

packages/perseus/src/widgets/__tests__/expression-mobile.test.tsx
handeyeco: Updated

packages/math-input/src/components/__tests__/integration.test.tsx
handeyeco: On the more complex tests I used `latex()` from MathQuill, but I kept the simple tests as-is. I think it's useful to display the complexity of MathQuill to highlight some of the problematic bits: what's shown to the user is a bunch of characters wrapped by spans, so I wanted some tests to represent that.

packages/math-input/src/components/__tests__/integration.test.tsx
handeyeco: I think that's a good point; nervous about doing that here though since I think that would then need to go through the translation pipeline. 🤔

PR #738, opened 2023-09-22
Upload merged coverage report to Codecov

kevinbarabash:
## Summary:
While Codecov supports uploading multiple reports, the experience isn't great.  In particular, if only one report uploads Codecov will show a report with the coverage dropping because it only has coverage from half of the reports.  This PR avoids the issue by merging the reports ourselves.  We still run jest and cypress in parallel.  We use artifacts to pass the reports to another job which merges them and uploads them.

Issue: None

## Test plan:
- we should see an increase in the coverage


kevinbarabash:
@jeremywiebe there's an error with the merged report, see https://app.codecov.io/github/Khan/perseus/commit/e1187103a9f21805e9d449f3045601bf904a234b.  I'm not sure what I did wrong.  I was wondering if you had any insights based on your earlier work merging reports in webpp.

.codecov.yml
```diff
@@ -3,20 +3,12 @@ coverage:
...
     range: "70...100"
 
-    status:
-        project:
-            default: on
-        patch:
-            default: off
-        changes:
-            default: off
-
```
kevinbarabash, line 13: `default: on` is invalid.  Removing this appears to have made it so that as long as coverage is going up the check will be green.

.codecov.yml
```diff
@@ -3,20 +3,12 @@ coverage:
...
 comment:
     layout: "diff, reach, files, footer"
     behavior: default
-    require_changes: no
-    require_base: no
-    require_head: yes
```
kevinbarabash, line 19: This syntax was also incorrect.

jaredly: I think yes & no are [valid bools in yaml](https://yaml.org/type/bool.html), but I also prefer true/false 😅

jaredly (approved):

.github/workflows/node-ci.yml
```diff
@@ -140,11 +146,43 @@ jobs:
...
+                  name: jest-coverage
+                  # path to decompress the artifact into
+                  path: ./
+
+            - name: Download Cypress Coverage
+              uses: actions/download-artifact@v2
+              with:
+                  name: cypress-coverage
+                  # path to decompress the artifact into
+                  path: ./
```
jeremywiebe, line 179: Can you leave a comment explaining how we handle these coverage files? I'm confused why we're downloading artifacts (well, I have a guess but a comment would clarify)?  Do previous jobs upload the coverage artifacts to Github and then this process downloads them all and pushes them to Codecov?

.github/workflows/node-ci.yml
```diff
@@ -140,11 +146,43 @@ jobs:
...
+
             - name: Upload Coverage
               uses: codecov/codecov-action@v3
               with:
                   token: ${{ secrets.CODECOV_TOKEN }}
-                  files: ./coverage/coverage-final.json
+                  files: ./coverage-final.json,./out.json
```
jeremywiebe, line 185: This represents the downloaded artifacts, right?  My guess is one is the Jest coverage output and the other is the Cypress coverage output?

jeremywiebe (approved):
Looks good, thank-you. Would just ask that you comment the two things I've noted so that it's easier for future folks to connect the pieces.

.github/workflows/node-ci.yml
kevinbarabash: The only way to share files between jobs is using artifacts.  I'll leave a comment about this.

.github/workflows/node-ci.yml
kevinbarabash: That's correct.  I'll add a comment.

PR #743, opened 2023-09-26
Fixing a typo in mobile-keypad's unmount lifecycle. 

SonicScrewdriver:
## Summary:
Turns out the memory leak was a simple little typo on a react lifecycle. Fixing that allows us to properly call _ComponentWillUnmount_, which removes all of our event listeners and observers related to resizing the keypad.

## Video:
https://github.com/Khan/perseus/assets/12463099/ea18a0c6-d700-4137-b8ce-cafebded9683

## Test Plan:
- make tesc
- manual testing

### DIFF
--- a/.changeset/wild-suns-explain.md
+++ b/.changeset/wild-suns-explain.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/math-input": minor
+---
+
+Ensured that we're properly calling componentWillUnmount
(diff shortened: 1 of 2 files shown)


jaredly (approved):

jeremywiebe (approved):
That was fun. 🙃 

I think we should also consider cleaning up that `setTimeout` that could still be running after unmount. Although we didn't reproduce that during our pairing, I suspect this still could happen. Could you follow up with that in a separate PR?

PR #745, opened 2023-09-27
Provide each package's version on load

jeremywiebe:
## Summary:

This PR implements a way to see what version of each package is loaded on the page. 

We are implementing this because we recently had an issue in production where it was difficult to identify what Perseus code was running (an abort did not work). This should give us more visibility into what's going on in production.

When any Perseus package is used on a page, it will attach a value to the `__perseus_debug__` object on `window`.  To view the loaded Perseus packages and versions, we'll be able to simply evaluate the `__perseus_debug__` object. 

Issue: LC-1269

## Test plan:

I ran storybook locally and no logs were printed. 
I switched to webapp and ran `services/static/tools/dev_update_perseus.sh` to "vendor in" the dev versions of Perseus I'm working on. Next I ran `make serve-frontend` and navigated to an exercise page. Finally I opened the dev tools and evaluated `__perseus_debug__`.

image


package.json
```diff
@@ -124,10 +124,11 @@
...
   "scripts": {
     "gen:parsers": "yarn --cwd packages/kas gen:parsers",
-    "build": "yarn gen:parsers && rollup -c config/build/rollup.config.js",
+    "prebuild": "yarn gen:parsers",
+    "build": "rollup -c config/build/rollup.config.js",
     "build:types": "yarn tsc --build tsconfig-build.json",
-    "build:prodsizecheck": "yarn gen:parsers && rollup -c config/build/rollup.config.js --configEnvironment='production'",
-    "watch": "rollup -c config/build/rollup.config.js --watch",
+    "build:prodsizecheck": "yarn build --configEnvironment='production'",
+    "watch": "yarn build --watch",
```
jeremywiebe, line 131: This is just a bit of simplification and deduplication.  `prebuild` is automatically run by script runners any time `build` is requested. 

The other scripts can leverage the `build` script by adding on extra params as needed.

kevinbarabash (commented):
It feels a little weird to be logging a bunch of stuff in production all of the time.  I wonder if we could get the same information by setting a global.  Maybe we could do something like:
```
if (typeof ka_debug_info == "undefined") {
   ka_debug_info = {};
}
ka_debug_info.perseus = { version: "2.3.4", ... }
```
If developers need to access this info they could inspect `window.ka_debug_info` in the dev tools console.

nedredmond:
Non-blocking suggestion: console.debug instead of console.log

.github/workflows/node-ci.yml
```diff
@@ -31,7 +31,6 @@ jobs:
               id: match
               with:
                   changed-files: ${{ steps.changed.outputs.files }}
-                  files: packages/ # Only look for changes in packages
```
jeremywiebe, line 34: This is wrong. We _do_ want to look for changes in all workspace packages (`config/build` for example, should be checked).

packages/perseus-core/src/utils/add-library-version-To-perseus-debug.ts
```diff
@@ -0,0 +1,29 @@
+/**
```
kevinbarabash, line 1: **nit:** the naming of this file include a rogue uppercase `T`.

packages/perseus-core/src/utils/add-library-version-To-perseus-debug.ts
```diff
@@ -0,0 +1,29 @@
...
+    const s = `${prefix}${libraryVersion}`;
+
+    if (globalThis) {
+        globalThis.__perseus_debug__ = globalThis.__perseus_debug__ ?? {};
+
+        globalThis.__perseus_debug__[libraryName] = s;
+    } else {
+        // eslint-disable-next-line no-console
+        console.warn(`globalThis not found found (${s})`);
+    }
```
kevinbarabash, line 28: **question:** are there environments where this code runs where `globalThis` is not available?

**suggestion:** a more robust check for `globalThis` would be:
```
if (typeof globalThis !== "undefined") {
    ...
} else {
    ...
}
```

packages/pure-markdown/src/index.ts
```diff
@@ -4,6 +4,8 @@
  * Note that this file may be used in stand alone nodejs, thus
  * do not import anything from Perseus
  */
+export {libVersion} from "./version";
```
kevinbarabash, line 7: **suggestion (non-blocking):** `version` feels like it's more common than `libVersion`.

packages/perseus-core/src/utils/add-library-version-To-perseus-debug.ts
```diff
@@ -0,0 +1,29 @@
...
+    libraryName: string,
+    libraryVersion: string,
+) => {
+    // If the library version is the default value, then we don't want to
+    // prefix it with a "v" to indicate that it is a version number.
+    let prefix = "v";
+    if (libraryVersion === "__lib_version__") {
+        prefix = "";
+    }
+    const s = `${prefix}${libraryVersion}`;
```
kevinbarabash, line 19: **nit:** maybe use a more descriptive variable name here.

kevinbarabash (approved):
LGTM. 🎉 Thanks for switching to setting a global with all the versions instead of logging.  I left a few suggestions in the comments, but nothing blocking.  I could see us doing something similar in some of our other repos.

kevinbarabash:
What happens if multiple versions of a package are loaded accidentally?  We should maybe track all versions by making the entry in `__perseus_debug__` for that module an array of version and logging a warning that multiple versions of the library have been loaded.

packages/perseus-core/src/utils/add-library-version-To-perseus-debug.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    if (globalThis) {
+        globalThis.__perseus_debug__ = globalThis.__perseus_debug__ ?? {};
+
+        globalThis.__perseus_debug__[libraryName] = s;
```
jeresig, line 24: I think if we already have one in stored we should check to see if the version matches the old version - and if not we should turn this into an array. This will help you debug cases where we have multiple versions of the same library running side-by-side.

packages/perseus-core/src/utils/add-library-version-To-perseus-debug.ts
jeremywiebe: iOS 12.1 and lower [don't support](https://caniuse.com/?search=globalThis) `globalThis`.  I know that in our main webapp we recently dropped support for Safari/iOS 12. But I think I'm going to keep this in place just so we don't intentionally break things for iOS 12 users.

I'll fix the way I check though, to what you suggest.

packages/pure-markdown/src/index.ts
jeremywiebe: The only reason I like `libVersion` is that in Perseus, specifically, there are several "version" exports. With `@khanacademy/perseus` exporting `apiVersion`, `itemDataVersion`, and `itemVersion`, I thought `libVersion` was more clear than simply `version`.

image

packages/perseus-core/src/utils/add-library-version-To-perseus-debug.ts
jeremywiebe: I like these ideas. Thank-you. 

I've updated this function to do as suggested: first registration registers the key as the library name and the value as the version string. On subsequent calls for the same library, the entry is morphed into an array that includes a distinct list of versions registered. Added tests to validate all this.

PR #747, opened 2023-09-29
move StatefulKeypadContextProvider into Perseus

handeyeco:
## Summary:
<Add your summary here>

Issue: LC-1337

## Test plan:


packages/math-input/src/components/keypad-context.tsx
```diff
@@ -0,0 +1,52 @@
+/**
+ * KeypadContext provides a way to the Keypad and (Server)ItemRenderer to
+ * communicate.
+ *
+ * The KeypadContext.Provider wraps the ExerciseFooter while KeypadContext.Consumer
+ * wraps each (Server)ItemRenderer render site and the Keypad rendered in the
+ * ExerciseFooter.
```
jeremywiebe, line 7: **question**: Is this doc comment still accurate?  I feel like we should talk about the `<StatefulKeypadContextProvider />` at least, right?

packages/math-input/src/components/keypad-context.tsx
```diff
@@ -0,0 +1,52 @@
...
+        scrollableElement: null,
+    });
+
+type Props = React.PropsWithChildren<unknown>;
+
+export function StatefulKeypadContextProvider(props: Props) {
+    // used to communicate between the keypad and the Renderer
+    const [keypadElement, setKeypadElement] = useState<KeypadAPI | null>();
+    // this is a KeypadContextRendererInterface from Perseus
+    const [renderer, setRenderer] = useState<any>(null);
```
jeremywiebe, line 31: **idea**: Non-blocking. Maybe there's an opportunity to use an interface for the type here?

packages/math-input/src/types.ts
```diff
@@ -105,3 +106,16 @@ export interface KeypadAPI {
     setKeyHandler: (keyHandler: KeyHandler) => void;
     getDOMNode: () => ReturnType<typeof ReactDOM.findDOMNode>;
 }
+
+export type KeypadContext = {
```
jeremywiebe, line 110: **suggestion**: I'm mildly concerned about having a `keypadContext` and `KeypadContext`. 🤔  I also think that typically our contexts are named `TitleCase`.  So, what do you think about naming this type `KeypadContextType` and then keeping the naming of the original `KeypadContext` as it was.  This `KeypadContextType` looks like it's not needed outside of the `math-input` package so a slightly clumsy name internally feels easier to deal with.

jeremywiebe (approved):
This looks like it's already paying off in reduced duplication. Thanks!

kevinbarabash:
@handeyeco any idea of why coverage when up so much for this PR?

PR #748, opened 2023-09-29
Add keypadActive state to context

handeyeco:
## Summary:

1. [Webapp ~ Cleanup PR](https://github.com/Khan/webapp/pull/17025)
2. [Perseus ~ Bring StatefulKeypadContext into Perseus](https://github.com/Khan/perseus/pull/747)
3. **[> Perseus ~ Add "keypadActive" state](https://github.com/Khan/perseus/pull/748)**
4. [Webapp ~ Conditionally render Khanmigo button](https://github.com/Khan/webapp/pull/17063)

The main point of these PRs is to be able to tell Khanmigo when the keypad is open so it can stop putting a button over the keypad.

This PR is a good example of some of our problems in MathInput though: there wasn't really a single place for this state. v1 keypad stored this state in Redux, v2 keypad stored this in hooks, and the KeypadContext was just passing around refs to components that other components where calling methods on. I didn't scope creep and fix all of these problems, but I did try to start fixing one: we now have a StatefulKeypadContextProvider which provides `keypadActive` state.

So now the v2 MobileKeypad doesn't need to track this state, it just gets it from Context. At the same time, other things that want to know this state (like Khanmigo) can also consume the Context.

Issue: LC-1337

## Test plan:
- Nothing should really change
- Once all these PRs are landed:
  - Use or emulate a mobile device
  - Go to a page that uses a NumericInput or Expression
  - Open and close the keypad
  - Note things work the same

### DIFF
--- a/.changeset/tall-mangos-poke.md
+++ b/.changeset/tall-mangos-poke.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/math-input": major
+"@khanacademy/perseus": patch
+---
+
+Hoist keypad active state into keypad context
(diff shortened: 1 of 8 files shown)


kevinbarabash:
@handeyeco any thoughts on why this PR might be causing such a drop in coverage?

SonicScrewdriver (approved):
This all makes sense to me! Looks like codecov would like us to add some tests for provided-keypad and mobile-keypad, but I'm really not seeing a large difference in the logic. I'm happy to approve and I'll leave it to your judgement about whether these changes really require attention or not.

jeremywiebe (approved):

PR #749, opened 2023-09-29
Use fast-glob to find package.json's for aliases in storybook

jeremywiebe:
## Summary:

I was doing some work with exposing package versions in production and came across this code in our Storybook setup. I thought it was problematic so I re-worked it to use `fast-glob` and (arguably) simpler. 

Once I figured out this code wasn't the problem, I thought it still a worthwhile improvement so... here we are. 

Issue: "none"

## Test plan:

`yarn; yarn start`

Storybook should open in a browser window.


nedredmond (approved):
Looks good to me!

PR #750, opened 2023-09-29
Ensured that we're still validating against pi when it is provided in the answerForms.

SonicScrewdriver:
## Summary:
This PR fixes a bug where we were no longer validating against pi when `answer.strict=false` when pi was provided in the answerForms. It looks like this is a side effect that came about with a more recent bug fix that exposed a hard-to-find historical bug. Previously we were always validating against pi by default, but we changed this after having issue with several questions being incorrectly tagged as approximations of pi. 

I updated tests to make sure that we're still validating against pi when these conditions are met.

Issue: LC-1331

## Test plan:
- new test

### DIFF
--- a/packages/perseus/src/perseus-types.ts
+++ b/packages/perseus/src/perseus-types.ts
@@ -762,7 +762,8 @@ export type PerseusNumericInputAnswer = {
     // The forms available for this answer.  Options: "integer, ""decimal", "proper", "improper", "mixed", or "pi"
     // NOTE: perseus_data.go says this is required even though it isn't necessary.
     answerForms?: ReadonlyArray<MathFormat>;
-    // Whether validatorForms should include the defaultAnswerForms, or strictly match to the provided answerForms
+    // Whether we should check the answer strictly against the the configured answerForms (strict = true)
+    // or include the set of default answerForms (strict = false).
     strict: boolean;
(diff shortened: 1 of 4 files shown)


packages/perseus/src/perseus-types.ts
```diff
@@ -762,7 +762,7 @@ export type PerseusNumericInputAnswer = {
     // The forms available for this answer.  Options: "integer, ""decimal", "proper", "improper", "mixed", or "pi"
     // NOTE: perseus_data.go says this is required even though it isn't necessary.
     answerForms?: ReadonlyArray<MathFormat>;
-    // Whether the answerForms should be strictly matched
+    // Whether validatorForms should include the defaultAnswerForms, or strictly match to the provided answerForms
```
jeremywiebe, line 765: I'd like to keep the original terms here.  The current changes focus on the code, but these types are more like the schema for Perseus.  So it'd be great if they could be more "end user" focused to describe how these settings affect the question.  

Maybe something like this?

>  Whether we should check the answer strictly against the the configured answerForms (strict = true) or include a set of default answerForms (strict = false).

jeremywiebe (approved):
Great. I like those changes to make the logic a little more explicit even if it takes more space. Thanks for fixing this up!

PR #752, opened 2023-10-03
Fix SHA posted when updating npm snapshots

jeremywiebe:
## Summary:

Ned [noticed](https://khanacademy.slack.com/archives/C01AZ9H8TTQ/p1696342746848549) that the git SHA posted to our "npm Snapshot: Published" comment doesn't look right. It's not! When this process was originally created, we used the `GITHUB_SHA` value. Github's [docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) state: 

> Note that GITHUB_SHA for this event is the last merge commit of the pull request merge branch. If you want to get the commit ID for the last commit to the head branch of the pull request, use github.event.pull_request.head.sha instead.

So this PR switches to use `github.event.pull_request.head.sha` so that the commit pointed to is the latest commit on the PR, not the merge commit. 

Issue: "none"

## Test plan:

Merge and check the next PR's snapshot comment. It should be the last commit on that PR.

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -292,6 +292,11 @@ jobs:
 
             - name: Calculate short SHA for this commit
               id: short-sha
+              # Why not GITHUB_SHA here? Because that is the last merge-commit
+              # for the PR (ie. the ephemeral commit that Github creates for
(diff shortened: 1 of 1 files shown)


nedredmond (approved):
Thanks Jeremy!

PR #753, opened 2023-10-04
Make filter action conjunctive rather than disjunctive in changeset check

nedredmond:
Makes filters conjunctive rather than disjunctive.

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -31,8 +31,9 @@ jobs:
               id: match
               with:
                   changed-files: ${{ steps.changed.outputs.files }}
-                  files: packages/ # Only look for changes in packages
-                  globs: "!(**/__tests__/*), !(**/__stories__/*), !(**/dist/*)" # Ignore test files
+                  files: "packages/, config/build/" # Only look for changes in packages, build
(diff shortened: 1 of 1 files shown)


.github/workflows/node-ci.yml
```diff
@@ -31,8 +31,9 @@ jobs:
               id: match
               with:
                   changed-files: ${{ steps.changed.outputs.files }}
-                  files: packages/ # Only look for changes in packages
-                  globs: "!(**/__tests__/*), !(**/__stories__/*), !(**/dist/*)" # Ignore test files
+                  files: "packages/, config/build/" # Only look for changes in packages, build
```
jeremywiebe, line 34: I'm tempted to just not specify the packages we monitor.  The `changeset-per-package` will get the workspace info and then check that each change in a package has a changeset already, right?  So does it hurt letting all files that aren't ignored by the glob through and then letting changeset-per-package figure things out?  I'm probably being paranoid, but if we add another utility package somewhere else in the repo we'd have to remember to update this `files:` key. 🤷‍♂️ 

I'll leave it with you to decide.

jeremywiebe (approved):
Approved!  I've left a comment that you can decide on (I'm good either way!).

.github/workflows/node-ci.yml
nedredmond: Files has a default so if we don't provide something it'll be limited to `github/`

PR #755, opened 2023-10-04
Updating Article Renderer to be able to return the focusedElement so that we can scroll these elements into view above the keypad.

SonicScrewdriver:
## Summary:
It turns out we never set up our article renderer to return the focused element, which is required to be able to scroll the element above the keypad when the user focuses on it! 

This PR adds the missing logic to the article renderer, and also makes a couple of necessary modernizations as we were still using string refs. I've also created a new story for testing multiple inputs in a mobile article. Please note that the keypad does look a little strange in this view as we don't currently have the scrollable element logic in Perseus. 

(I suspect that eventually, we should probably move the scrollIntoView logic into Perseus.)

### DIFF
--- a/packages/perseus/src/__tests__/article-renderer.test.tsx
+++ b/packages/perseus/src/__tests__/article-renderer.test.tsx
@@ -1,0 +1,102 @@
+import {
+    StatefulKeypadContextProvider,
+    KeypadContext,
+} from "@khanacademy/math-input";
+import {RenderStateRoot} from "@khanacademy/wonder-blocks-core";
+import {render, screen} from "@testing-library/react";
+import * as React from "react";
+import "@testing-library/jest-dom"; // Imports custom matchers
+
+import {
+    testDependencies,
+    testDependenciesV2,
+} from "../../../../testing/test-dependencies";
+import {multiSectionArticleWithExpression} from "../__testdata__/article-renderer.testdata";
+import ArticleRenderer from "../article-renderer";
+import * as Dependencies from "../dependencies";
+import {ApiOptions} from "../perseus-api";
+import {registerWidget} from "../widgets";
+import ExpressionExport from "../widgets/expression";
+
+import MockWidgetExport from "./mock-widget";
+
+import type {APIOptions} from "../types";
+
+// This looks alot like `widgets/__tests__/renderQuestion.jsx', except we use
+// the ArticleRenderer instead of Renderer
+export const RenderArticle = (
+    apiOptions: APIOptions = Object.freeze({}),
+): {
+    container: HTMLElement;
+    renderer: ArticleRenderer;
+} => {
+    let renderer: ArticleRenderer | null = null;
+    const {container} = render(
+        <RenderStateRoot>
+            <StatefulKeypadContextProvider>
+                <KeypadContext.Consumer>
+                    {({keypadElement, setRenderer}) => (
+                        <ArticleRenderer
+                            ref={(node) => {
+                                renderer = node;
+                                setRenderer(node);
+                            }}
+                            json={multiSectionArticleWithExpression}
+                            dependencies={testDependenciesV2}
+                            apiOptions={{...apiOptions}}
+                            keypadElement={keypadElement}
+                        />
+                    )}
+                </KeypadContext.Consumer>
+                {/* The ItemRenderer _requires_ two divs: a work area and hints
+                area. Without both of these, it fails to render anything! */}
+                <div id="workarea" />
+                <div id="hintsarea" />
+            </StatefulKeypadContextProvider>
+        </RenderStateRoot>,
+    );
+    if (!renderer) {
+        throw new Error(`Failed to render!`);
+    }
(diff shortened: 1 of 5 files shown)


packages/perseus/src/article-renderer.tsx
```diff
@@ -167,13 +172,20 @@ class ArticleRenderer
 
         // TODO(alex): Add mobile api functions and pass them down here
         const sections = this._sections().map((section, i) => {
-            const refForSection = `section-${i}`;
+            const refForSection: any = i;
             return (
-                <DependenciesContext.Provider value={this.props.dependencies}>
+                <DependenciesContext.Provider
+                    key={i}
```
SonicScrewdriver, line 178: Do we really need to be creating multiple dependencies context providers? It doesn't seem specific to each of these sections at all. However, there could be something I'm missing.

For now, I added a key to stop react from throwing a noisy error about not having unique key props for our children in a list.

packages/perseus/src/article-renderer.tsx
```diff
@@ -87,21 +88,27 @@ class ArticleRenderer
...
         // input.
         let didFocusInput = false;
+        let focusedInput;
+
         if (this._currentFocus) {
             const [sectionRef, ...focusPath] = this._currentFocus;
-            // eslint-disable-next-line react/no-string-refs
-            // @ts-expect-error - TS2339 - Property 'getInputPaths' does not exist on type 'ReactInstance'.
-            const inputPaths = this.refs[sectionRef].getInputPaths();
+
```
SonicScrewdriver, line 95: String refs are old hat, and now we need a reference to the actual focused element so that we can return it to webapp with the passed in "onFocusChange" method.

packages/perseus/src/article-renderer.tsx
```diff
@@ -87,21 +88,27 @@ class ArticleRenderer
...
             });
+            focusedInput =
+                this.sectionRenderers[sectionRef].getDOMNodeForPath(focusPath);
         }
 
         if (this.props.apiOptions.onFocusChange != null) {
             this.props.apiOptions.onFocusChange(
                 this._currentFocus,
                 prevFocusPath,
                 didFocusInput ? keypadElement?.getDOMNode() : null,
```
SonicScrewdriver, line 110: Making sure we return the focusedInput now, if there is one.

handeyeco (commented):
Thanks for investigating this! Glad to see another area where we're removing string refs.

Are there tests we could write for this? For instance a test to make sure ArticleRenderer is calling `onFocusChange` with the correct focused element?

I think you're saying that the stories don't scroll into view. Would it be hard to make them actually scroll into view?

packages/perseus/src/article-renderer.tsx
jeremywiebe: No, I don't think we do.  We could move it to just wrap the final `<div />` that we render these sections into!

packages/perseus/src/article-renderer.tsx
```diff
@@ -167,13 +172,20 @@ class ArticleRenderer
...
         const sections = this._sections().map((section, i) => {
-            const refForSection = `section-${i}`;
+            const refForSection: any = i;
             return (
-                <DependenciesContext.Provider value={this.props.dependencies}>
+                <DependenciesContext.Provider
+                    key={i}
+                    value={this.props.dependencies}
+                >
                     <div key={i} className="clearfix">
```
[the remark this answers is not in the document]
SonicScrewdriver: I thiiiink we're okay for this one as we never remove or reorder sections in Articles. Sadly there's no data in the section that can be used, as it's either that single very very long json string for the Perseus content, or the question answers (that requires digging a little deep). That said, however, I'm totally open to implementing any alternative ideas!

Screenshot 2023-10-19 at 10 22 44 AM

packages/perseus/src/__tests__/article-renderer.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+            ...ApiOptions.defaults,
+            isMobile: false,
+            customKeypad: false,
+        });
+
+        // Assert
+        expect(screen.getByRole("textbox")).toBeInTheDocument();
+    });
+
+    it("should call the onFocusChanged callback when an input is focused", async () => {
```
handeyeco, line 108: Thank you!

packages/perseus/src/article-renderer.tsx
```diff
@@ -166,47 +173,53 @@ class ArticleRenderer
...
-                </DependenciesContext.Provider>
+                <div key={sectionIndex} className="clearfix">
+                    <Renderer
+                        {...section}
+                        ref={(elem) => {
+                            if (elem) {
+                                this.sectionRenderers[sectionIndex] = elem;
+                            }
+                        }}
+                        key={sectionIndex}
```
handeyeco, line 189: Do we need this key still?

handeyeco (approved):
Looks good to me!

PR #756, opened 2023-10-05
Prevent provided-keypad from re-rendering due to prop callabacks

jeremywiebe:
## Summary:

While doing some performance tuning in webapp I noticed that this component is re-rendering because we are creating the callback props inline. This means that they'll always be different. Moving them to be instance functions fixes this. 

Issue: LC-1346

## Test plan:

I was able to `yarn link` all the Perseus packages into webapp and then test. All re-renders of the `ProvidedKeypad` component stopped with these changes.

### DIFF
--- a/.changeset/fresh-peas-hang.md
+++ b/.changeset/fresh-peas-hang.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/math-input": patch
+---
+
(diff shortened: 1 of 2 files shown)


nedredmond (approved):
I like it!

PR #760, opened 2023-10-11
Remove "grid" role from keypad

nedredmond:
LC-1356

Removes "grid" role from keypad. Its presence without "row" and "gridcell" roles in child elements muddles the screen reader experience.


handeyeco (approved):
Looks good, thank you!

PR #762, opened 2023-10-12
Import addLibraryVersionToPerseusDebug from local file, not package

jeremywiebe:
## Summary:

While implementing our production version diagnostics for Perseus, I used some shell scripts to generate out the code for each package. In doing so, I implemented the `version.ts` for `perseus-core` incorrectly (as that's the package that hosts the helper function). 

All tests pass locally but when Perseus is integrated into a host application, it blows up with the error: `TypeError: perseusCore.addLibraryVersionToPerseusDebug is not a function` (@nedredmond).

Issue: "none"

## Test plan:

`yarn test` 
Let the snapshot build and then try it in webapp.


packages/perseus-core/src/__tests__/version.test.ts
```diff
@@ -0,0 +1,16 @@
...
+    it("adds the perseus-core version to globalThis", async () => {
+        await import("../version");
+
+        expect(globalThis).toHaveProperty("__perseus_debug__");
+        expect(globalThis.__perseus_debug__).toMatchInlineSnapshot(`
+            {
+              "@khanacademy/perseus-core": "__lib_version__",
+            }
+        `);
+    });
```
jeremywiebe, line 15: I'm not 100% sure if this is helpful.  If I revert the change in `version.ts` it still passes so this doesn't actually protect against regression here.  

Any thoughts?

benchristel: Modifying `globalThis` in a test makes me nervous. The fact that the test doesn't catch regressions makes me inclined to remove it.

I'd rather lean on a linter or something to check invariants like "a package should never import its own code from `node_modules`" (if that is in fact what we want to enforce... I'm not sure what the rule should be).

benchristel (approved):
Looks good! I'd lean towards removing the test.

packages/perseus-core/src/version.ts
```diff
@@ -2,7 +2,7 @@
 // version number during the release build.
 // In dev, you'll never see the version number.
 
-import {addLibraryVersionToPerseusDebug} from "@khanacademy/perseus-core";
+import {addLibraryVersionToPerseusDebug} from "./utils/add-library-version-to-perseus-debug";
```
nedredmond, line 5: Oh, I see the problem now!

nedredmond: Thanks for the quick fix!

nedredmond (approved):

packages/perseus-core/src/__tests__/version.test.ts
jeremywiebe: You've convinced me. I'll remove the test and land this.  I'll also do a small amount of research if this would be relatively easy to lint for (maybe with ESLint's dynamic rules).

PR #768, opened 2023-10-18
Transition mobile-keypad to use AphroditeCSSTransitionGroup (vendored from webapp)

jeremywiebe:
## Summary:

After setting up analytics for Keypad open/closed events and integrating it into webapp, we noticed that these events were over-firing by quite a bit. Turns out, the mobile keypad renders the full keypad and then animates displaying and hiding it (and toggles the `visibility` style) but never unmounts it. 

This PR changes things so that we import the `AphroditeCSSTransitionGroup` which manages animating in/our components as they are mounted/unmounted. This allows us to only render the Keypad when it's actually active. 

Issue: LC-1346

## Test plan:

- Integrate into webapp.
- Open a browser and enable dev tools and mobile emulation.
- Switch to the **Network** tab in dev tools.
- Navigate to an exercise that contains an `expression` widget.
- Note that there were NO CEDAR events fired for `MathInputVirtualKeypadOpened` or `MathInputVirtualKeypadClosed` during the rendering of the exercise.
- Tap into the expression and note that the `MathInputVirtualKeypadOpened` event fired.
- Close the keypad and note that the `MathInputVirtualKeypadClosed` event fired.


packages/math-input/src/components/keypad/mobile-keypad.tsx
```diff
@@ -39,21 +29,28 @@ type Props = {
 
 type State = {
     containerWidth: number;
-    hasBeenActivated: boolean;
     keypadConfig?: KeypadConfiguration;
     keyHandler?: KeyHandler;
     cursor?: Cursor;
 };
 
+/**
```
jeremywiebe, line 37: Moving this here allows this comment to be surfaced when you hover usages of this component (at least in VSCode).

jeremywiebe:
Tagged @kevinbarabash for the changes/import of `AphroditeCSSTransitionGroup` in case you have thoughts about using that approach.

packages/math-input/src/components/aphrodite-css-transition-group/util.ts
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    if (Array.isArray(list)) {
+        for (const item of list) {
+            result.push(...flatten(item));
+        }
+    } else {
+        // @ts-expect-error - TS2345 - Argument of type 'CSSProperties | NestedArray<CSSProperties | Falsy>' is not assignable to parameter of type 'CSSProperties'.
+        result.push(list);
+    }
```
kevinbarabash, line 22: I wonder why it thinks it's isn't a TypeScript error. 🤔

kevinbarabash (commented):
The addition of `AphroditeCSSTransitionGroup` seems reasonable to me.

packages/math-input/src/components/aphrodite-css-transition-group/util.ts
jeremywiebe: It doesn't show up locally!  I wonder if this is a TypeScript mismatch.  I'll investigate.

packages/math-input/src/components/keypad/__tests__/mobile-keypad.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+        // Act
+        rerender(
+            <MobileKeypad
+                onAnalyticsEvent={onAnalyticsEvent}
+                setKeypadActive={(keypadActive: boolean) => undefined}
+                keypadActive={false}
+            />,
+        );
+        unmount();
```
jeremywiebe, line 105: I am a bit unsure of this. This was the only way I could get the final analytics event to fire and "flush" before the end of the test.

somewhatabstract: Seems right to me. :)

somewhatabstract (approved):
Thanks for adding tests! This looks great. 💯

PR #775, opened 2023-10-20
Fix overfiring CEDAR event: ContentLibraryMathInputBoxEvaluated

nedredmond:
## Summary

Renderer sends a flag along with the rubric to let the Expression widget know when it's scoring.
We expected it only to be called on submit, but it was being called multiple times (once per `answerForm`) for every change event.

LC-1352

## Test

Verify that we only see this event in the network tab after `Check` button pressed.


jeremywiebe (commented):
Moving this up into the renderer and then using the if block on the `score` doesn't quite sit right with me. It sits apart from the expression and feels like one of those things that'll be easy to get out of sync.

You mentioned that the `simpleValidate` is called once per answer form, but from what I see the looping over answer forms is done within `validate` which is called by `simpleValidate` (https://github.com/Khan/perseus/blob/5bcf118c3190d92c2d5112b5f559e67cad5a906b/packages/perseus/src/widgets/expression.tsx#L197). 

I'm wondering if we can leave this in the `expression` widget but differentiate when we're actually scoring the user input vs something else (maybe a flag going into `scoreInput`? something else?). I'd be happy to pair on this with you.

.changeset/fuzzy-turtles-add.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Moved analytics callback for expression widget to Renderer to hook on to prevent overfiring
```
jeremywiebe, line 5: **todo**: Update this based on the latest round of changes.

packages/perseus/src/server-item-renderer.tsx
```diff
@@ -416,6 +416,7 @@ export class ServerItemRenderer
                     content={this.props.item.question.content}
                     widgets={this.props.item.question.widgets}
                     images={this.props.item.question.images}
+                    {...this.props.dependencies}
```
jeremywiebe, line 419: **thought**: We also have the ArticleRenderer. Might be good to ensure we're plumbing the dependencies through to the `Renderer` in there. It's not super common to place "answerable" widgets in an Article, but we do have the GradedGroup widget which enables checking answers in an Article and so it's probably good to do so we get accurate analytics from both articles and exercises.

jeremywiebe (approved):
Looks good Ned! Thanks for re-working the location of firing events.  I left two comments that you probably want to address before landing.

packages/perseus/src/server-item-renderer.tsx
nedredmond: Interesting-- that expansion of analytics not related to the current bug, so that should probably be a follow-up ticket for a clean git history.

.changeset/fuzzy-turtles-add.md
nedredmond: Ah, good eye! Thanks!

PR #776, opened 2023-10-23
Upgrade actions/checkout to v3

nedredmond:
LC-1268

Change all instances of `actions/checkout` to `v3`.

`v2` uses node 12, which is no longer supported.
`v3` uses node 16, which reached end of life last month, but it's the node version we currently use in Perseus.
`v4` is out but uses node 20, which we do not yet support.


kevinbarabash (approved):
LGTM 🚢

PR #778, opened 2023-10-23
Upgrade Wonder Blocks and Wonder Stuff libraries to latest

jeremywiebe:
## Summary:

Today our primary consumer of Perseus is Khan Academy's webapp (and the mobile app). Today we are upgrading new releases of Wonder Blocks and Wonder Stuff in webapp. 

Given that all of our WB and WS dependencies are peer/dev dependencies in Perseus, we essentially always use the host's version of those packages once we are integrated. 

So, this PR upgrades WB/WS packages to their latest releases to avoid us devving on aging/outdated WB/WS packages. 

**NOTE**: To get this all working, I had to upgrade Cypress and force use of Webpack 5 (by installing it explicitly in the root `package.json`). Webpack 5 was needed because Webpack 4 doesn't properly handle the `"exports"` key that the Phosphor icon's `package.json` uses). 

Issue: "none"

## Test plan:

Run tests (had to adjust Jest so that it could handle imports of our new Phosphor SVG icons)


config/test/svg.transform.js
```diff
@@ -0,0 +1,25 @@
+const path = require("path");
```
jeremywiebe, line 1: This is heavily inspired by web app's version...  ok, ok.... it's a blatant copy-paste. 😄

jeremywiebe:
No longer needed.

PR #780, opened 2023-10-24
Add analytics for widget rendering errors

jeremywiebe:
## Summary:

Today, we have Sentry error handling built into Perseus. But its not easy to track stats on how often a widget throws an error that causes a rendering failure by widget type. So this PR adds an analytics event for when a widget fails to render. 

This is implemented by adding an `onError` callback prop to the `ErrorBoundary` component. I've also moved the use of the `ErrorBoundary` for widget rendering _into_ the `WidgetContainer` component which wraps all widgets that we render. 

While I was in here, I migrated some related code away from string refs to React refs. 

Issue: LC-1383

## Test plan:

`yarn test` passes!

Integrate the snapshot PR from this PR into webapp and run tests and do some manual testing to make sure we haven't introduced a regression.


packages/perseus/src/__tests__/renderer-api.test.tsx
```diff
@@ -8,16 +8,19 @@ import TestUtils from "react-dom/test-utils";
 import _ from "underscore";
 import "jest-enzyme";
 
-import {ClassNames, Dependencies, Renderer} from "@khanacademy/perseus";
```
jeremywiebe, line 11: This was just wrong. We shouldn't be importing symbols from our own package like this.

packages/perseus/src/__tests__/renderer-api.test.tsx
```diff
@@ -165,11 +164,7 @@ describe("Perseus API", function () {
                 // version
                 expect(ClassNames.FOCUSED).toBe("perseus-focused");
 
-                const renderer = renderQuestionArea(inputNumber1Item, {
-                    interceptInputFocus: function (widgetId) {
```
jeremywiebe, line 169: I checked and this API option hasn't existed since we returned to this repo over a year ago. The test still passes with this removed, so 🤷‍♂️ .

packages/perseus/src/renderer.tsx
```diff
@@ -550,21 +550,25 @@ class Renderer extends React.Component<Props, State> {
...
-                    />
-                </ErrorBoundary>
+                    id={id}
+                    ref={(node) => {
+                        const containerId = "container:" + id;
+                        if (node != null) {
+                            this._widgetContainers.set(containerId, node);
+                        } else {
+                            this._widgetContainers.delete(containerId);
+                        }
```
jeremywiebe, line 562: Based on https://react.dev/learn/manipulating-the-dom-with-refs#how-to-manage-a-list-of-refs-using-a-ref-callback

packages/perseus/src/renderer.tsx
```diff
@@ -603,7 +607,6 @@ class Renderer extends React.Component<Props, State> {
 
         return {
             ...widgetProps,
-            ref: id,
```
jeremywiebe, line 606: This ref prop was never used, and it is a huge potential bug. Now that we use proper refs in the `WidgetContainer`, we can safely remove this.

packages/perseus/src/renderer.tsx
```diff
@@ -853,15 +856,11 @@ class Renderer extends React.Component<Props, State> {
         ];
     };
 
-    getWidgetInstance: (id: string) => Widget | null | undefined = (
-        id: string,
-    ): Widget | null | undefined => {
```
jeremywiebe, line 858: With arrow functions on classes, we only need the Typescript typing on the first part (no idea what it's called)... the arrow function definition after the `=` correctly infers the types if we properly type the member.

nedredmond: Yeah, you could do this either way, but since the function is so strongly defined in itself deleting the definition before "=" would work great as well!

packages/perseus/src/renderer.tsx
```diff
@@ -223,6 +223,8 @@ class Renderer extends React.Component<Props, State> {
     // The i18n linter.
     _translationLinter: TranslationLinter;
 
+    _widgetContainers: Map<string, WidgetContainer> = new Map();
```
nedredmond, line 226: Hmm... I am not 100% sure how this works in class components, but we should make sure we are not overwriting our refs on re-render. Just something to look out for!

packages/perseus/src/renderer.tsx
```diff
@@ -550,21 +550,25 @@ class Renderer extends React.Component<Props, State> {
...
-                        shouldHighlight={shouldHighlight}
-                        linterContext={PerseusLinter.pushContextStack(
-                            this.props.linterContext,
-                            "widget",
-                        )}
-                    />
-                </ErrorBoundary>
+                    id={id}
+                    ref={(node) => {
+                        const containerId = "container:" + id;
```
nedredmond, line 557: nitpick: Since you have to construct this ID in so many places, maybe you could make a simple function to ensure we don't get a typo somewhere?

`const containerId = (id: string) => "container:" + id;`

nedredmond (approved):
LGTM! Just a question and a tiny non-blocking nitpick.

PR #457, opened 2023-03-31
Remove almost all traces of Flow

kevinbarabash:
## Summary:
This PR makes the following changes:
- removes all flow related tooling except for flowgen
- updates almost all coments reference Flow to references TypeScript to be tool agnostic

Issue: None

## Test plan:
- yarn tsc

PR #781, opened 2023-10-25
Ensured we have access to transform.center when working with dilation in the Transform Widget 

SonicScrewdriver:
## Summary:
Team-LC has noticed recently that the skip rate on our transformer widget was quite a bit higher than it is normally. After further investigation, we discovered that the "homothety"/dilation tool is spitting out a large barrage of errors and is no longer functional. We suspect this is likely the main cause for the increase in reports. 

@jeanettehead and I looked into the history of the file and noticed that a key line of code was accidentally shifted to be a part of a parent comment as part of a [PR that removed all references to flow in Perseus](https://github.com/Khan/perseus/pull/457).

`center: self.dilationCircle.centerPoint.coord`

 It looks like a pesky little newline was accidentally (and likely automatically) removed, causing us to lose access to the required "center" property for our dilation tool. 

Since this widget is deprecated and only sees minimal use in our international content, the visibility of the issue was incredibly low. Thankfully the new Perseus Widget Analyzer by @pmcgill88 helped us catch this issue by providing greater insight into how our widgets are being actively used. 

Issue: LC-1384

## Test plan:
manual testing

### DIFF
--- a/packages/perseus/src/widgets/transformer.tsx
+++ b/packages/perseus/src/widgets/transformer.tsx
@@ -2521,6 +2521,7 @@ class Transformer extends React.Component<Props> {
                     // NOTE(kevinb): It's hard to convince that this.dilationCircle
                     // will still be defined here.                    center: self.dilationCircle.centerPoint.coord,
                     scale: newRadius / oldRadius,
+                    center: self.dilationCircle.centerPoint.coord,
                 });
             },
             circleNormalStyle: {
--- a/packages/perseus/src/widgets/transformer.tsx
+++ b/packages/perseus/src/widgets/transformer.tsx
@@ -2518,8 +2518,6 @@ class Transformer extends React.Component<Props> {
                 // @ts-expect-error - TS2345 - Argument of type '{ type: "dilation"; center: any; scale: number; }' is not assignable to parameter of type 'PerseusTransformerTransformation'.
                 self.doTransform({
                     type: "dilation",
-                    // NOTE(kevinb): It's hard to convince that this.dilationCircle
-                    // will still be defined here.
                     scale: newRadius / oldRadius,
                     center: self.dilationCircle.centerPoint.coord,
                 });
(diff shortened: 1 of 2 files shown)


packages/perseus/src/widgets/transformer.tsx
```diff
@@ -2521,6 +2521,7 @@ class Transformer extends React.Component<Props> {
                     // NOTE(kevinb): It's hard to convince that this.dilationCircle
                     // will still be defined here.                    center: self.dilationCircle.centerPoint.coord,
```
jeanettehead, line 2522: the code you need is actually here in this comment. the newline got accidentally deleted

packages/perseus/src/widgets/transformer.tsx
```diff
@@ -2518,9 +2518,8 @@ class Transformer extends React.Component<Props> {
                 // @ts-expect-error - TS2345 - Argument of type '{ type: "dilation"; center: any; scale: number; }' is not assignable to parameter of type 'PerseusTransformerTransformation'.
                 self.doTransform({
                     type: "dilation",
-                    // NOTE(kevinb): It's hard to convince that this.dilationCircle
```
SonicScrewdriver, line 2521: Realized we don't need this comment at all any more, as we've now removed all the flow overrides! No need to convince typescript. :)

packages/perseus/src/widgets/transformer.tsx
```diff
@@ -2518,9 +2518,8 @@ class Transformer extends React.Component<Props> {
...
-                    // will still be defined here.                    center: self.dilationCircle.centerPoint.coord,
                     scale: newRadius / oldRadius,
+                    center: self.dilationCircle.centerPoint.coord,
```
jeremywiebe, line 2522: This is a super small nit.   But could you move this to where it was above?

jeremywiebe (approved):
Nice catch!

PR #783, opened 2023-10-26
Add missed widgetId to 'perseus:widget-rendering-error' event

jeremywiebe:
## Summary:

Patrick and I discussed whether this event should have the `widgetId` on it. We agreed it should, but I forgot to add it back in. So here we go. 

I also realized that I missed a very important part of this implementation while adding missing tests. The part where we call the analytics. 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️   Added now!

Issue: LC-1383

## Test plan:

`yarn test`

### DIFF
--- a/packages/perseus-core/src/analytics.ts
+++ b/packages/perseus-core/src/analytics.ts
@@ -21,6 +21,13 @@ export type PerseusAnalyticsEvent =
           payload: null;
       }
     | {
+          type: "perseus:widget-rendering-error";
+          payload: {
+              widgetType: string;
+              widgetId: string;
+          };
+      }
+    | {
           type: "math-input:keypad-closed";
           payload: {
               virtualKeypadVersion: VirtualKeypadVersion;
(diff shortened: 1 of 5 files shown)


packages/perseus-core/src/analytics.ts
```diff
@@ -21,21 +21,22 @@ export type PerseusAnalyticsEvent =
           payload: null;
       }
     | {
-          type: "math-input:keypad-closed";
```
jeremywiebe, line 24: I moved this up to be together with the other Perseus events.

packages/perseus-core/src/analytics.ts
```diff
@@ -21,21 +21,22 @@ export type PerseusAnalyticsEvent =
...
+          type: "perseus:widget-rendering-error";
```
jeremywiebe, line 24: I moved this up to be together with the other Perseus events.

pmcgill88 (approved):
Thank you Jeremy!

PR #784, opened 2023-10-26
Fix horizontal stretch for graded group dots, + a11y tweaks

nedredmond:
LC-1291

Blocks some global list styles from affecting indicator dot `ul` container. This was causing a horizontal stretch in mobile, as well as other more subtle side effects.
Also, increases clickable area to 24 by 24 and adds hover/focus states. This required some refactoring of the current code. Now, instead of an `li` containing a `span`, it's an `li` containing a WonderBlocks `Clickable` containing a `span`... and when it's active, it contains another span! The interior span is to get around the background filling in the double border.

Style changes per this design:
https://www.figma.com/file/2lUPOSbOP8tbW7RLqbBFLh/Expression-Widget?type=design&node-id=5018-31744&mode=design&t=Ln3lgFaqAagSV97A-0

### DIFF
--- a/packages/perseus/src/styles/perseus-renderer.less
+++ b/packages/perseus/src/styles/perseus-renderer.less
@@ -133,7 +133,7 @@
         margin: -11px 0px 22px 0px; // first-level lists need padding
     }
 
-    .paragraph ul:not(.perseus-widget-radio) {
+    .paragraph ul:not(.perseus-widget-radio, .indicatorContainer) {
         padding-left: 35px;
         list-style-type: disc;
     }
--- a/.changeset/cool-days-sell.md
+++ b/.changeset/cool-days-sell.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": minor
+---
+
+Graded Group Set indicator dots now have a 24x24 clickable area as well as hover/focus states. Stretching issue on mobile resolved as well.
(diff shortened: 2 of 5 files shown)


packages/perseus/src/styles/perseus-renderer.less
```diff
@@ -133,7 +133,7 @@
         margin: -11px 0px 22px 0px; // first-level lists need padding
     }
 
-    .paragraph ul:not(.perseus-widget-radio) {
+    .paragraph ul:not(.perseus-widget-radio, .indicatorContainer) {
```
jeremywiebe, line 136: **thought**: CSS whackamole continues. 😞

jeremywiebe (approved):
Nice work figuring this out. Way more changes than I originally thought we'd need to fix "just a small stretching issue" (not a bad thing, just noting that sometimes the simple-looking fixes aren't so simple).  😂

nedredmond:
Thanks for the review!

FYI @jeremywiebe, not all of the changes are related to the bug. You and @SonicScrewdriver gave a 👍 to Sarah B's request for a11y improvement as part of the task, otherwise I would have split out that work.

https://khanacademy.atlassian.net/browse/LC-1291?focusedCommentId=280050

packages/perseus/src/styles/perseus-renderer.less
nedredmond: We really need to do something about that.

PR #786, opened 2023-10-26
Add Phosphor as a peer/dev dependency

handeyeco:
## Summary:
Here's my current understanding:
- WB added Phosphor as a peer dependency
- The WB team added Phosphor as a Webapp dependency
- Perseus uses some WB packages that use Phosphor
- Perseus did not have Phosphor as a dependency
    - This was not an issue when upgrading Webapp, because the WB team took care of it
    - When I tried to upgrade Perseus in Mobile, no Phosphor dependency

So this is part 1 of 2 in this solution: make Perseus use Phosphor as a dev/peer dependency

Part 2 of 2 will be: add Phosphor as a dependency in mobile


jeremywiebe (approved):

jandrade (approved):
Makes sense to me! 🚀

PR #788, opened 2023-10-31
Create randomized storybook stories for Radio, Categorizer, Dropdown, Explanation, Expression, and interactive graph

jeanettehead:
## Summary:
Created during hackathon 2023 - Basically manual fuzz testing - random configurations for widgets.

Issue: None

## Test plan:
None


packages/perseus/src/widgets/__testdata__/randomizers.ts
```diff
@@ -0,0 +1,48 @@
...
+    return Math.random() < threshold;
+};
+
+// generates an integer inclusive of the specified min and max
+export const randomInteger = (min: number, max: number): number => {
+    return Math.floor(Math.random() * (max + 1 - min)) + min;
+};
+
+// A helper to create an array of a certain length.
+export const arrayOfLength = (length: number): number[] => {
```
jeanettehead, line 38: is there a better way to do this? I want range from python - but didn't name this range because it's not exactly the same

jeanettehead:
So - check this out, use the [storybook build](https://650db21c3f5d1b2f13c02952-izkusraodp.chromatic.com/?path=/story/math-input-components-keypad-button--all-buttons). Do we want this? Is it helpful? 

I'm ok with this just being a silly hack if it's not useful to bring in.

jeremywiebe (approved):
Eventually we can hopefully the new comments you've added to `perseus-thypes.ts` into the actual type system. Right now it's not trivial because we have helper functions that return arrays instead of specifically-sized tuples. 


This is cool though!  Thanks.

PR #791, opened 2023-11-06
Moving ref'd element above the animation so that it doesn't return null

SonicScrewdriver:
## Summary:
Our keypadElement was only returning null to webapp after our new rendering improvements. This tiny PR should fix the issue so that we can use the keypadElement to scroll our inputs into view. 

## Test plan: 
- manual testing

### DIFF
--- a/.changeset/chilled-jeans-sing.md
+++ b/.changeset/chilled-jeans-sing.md
@@ -2,4 +2,4 @@
 "@khanacademy/math-input": patch
 ---
 
-Minor fix to ensure that the keypadElement is being provided to webapp.
+Minor fix to ensure that the keypadElement is always provided to mobile keypad consumers
--- a/.changeset/chilled-jeans-sing.md
+++ b/.changeset/chilled-jeans-sing.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/math-input": patch
+---
+
+Minor fix to ensure that the keypadElement is being provided to webapp.
--- a/packages/math-input/src/components/keypad/mobile-keypad.tsx
+++ b/packages/math-input/src/components/keypad/mobile-keypad.tsx
@@ -178,31 +178,28 @@ class MobileKeypad extends React.Component<Props, State> implements KeypadAPI {
         const convertDotToTimes = keypadConfig?.times;
 
         return (
-            <AphroditeCssTransitionGroup
-                transitionEnterTimeout={AnimationDurationInMS}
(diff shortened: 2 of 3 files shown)


.changeset/chilled-jeans-sing.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/math-input": patch
+---
+
+Minor fix to ensure that the keypadElement is being provided to webapp.
```
jeremywiebe, line 5: **thought**: For changelogs in Perseus, I try to write them without referencing the consuming app too much.  Can we adjust this slightly to be agnostic?  Something like "Minor fix to ensure that the keypadElement is always provided to mobile keypad consumers" ?

packages/math-input/src/components/keypad/mobile-keypad.tsx
```diff
@@ -178,31 +178,28 @@ class MobileKeypad extends React.Component<Props, State> implements KeypadAPI {
...
-                        transform: "translate3d(0, 100%, 0)",
-                    },
-                }}
-            >
-                {keypadActive ? (
-                    <View
-                        style={containerStyle}
-                        forwardRef={this._containerRef}
-                    >
+            <View style={containerStyle} forwardRef={this._containerRef}>
```
jeremywiebe, line 181: Just want to double-check that this was tested in webapp.  I feel like I tried this and ran into some layout issues.  I'm guessing this means that we'll always have a `<div />` that is the "visible size/position" of the keypad and the keypad will animate up/down within that div.

jeremywiebe (approved):
As long as my comment about animation/layout is covered, this looks good!

packages/math-input/src/components/keypad/mobile-keypad.tsx
SonicScrewdriver: Still testing as I needed the PR to test :) 

Sadly after testing -- this only fixes the null reference issue I was having in scrollElementIntoViewWithKeypad. It doesn't break anything else, but it still doesn't allow us to scroll our inputs into view. The `keypadElement.getBoundingRect().height` seems to be returning zero as there's no content available to give the necessary height for our calculations at the time of scroll.

SonicScrewdriver: This was solved in the scrollableElement PR! We're going to use a setTimeout in order to ensure we get the height AFTER the keypad has been enabled.

PR #794, opened 2023-11-08
Removing all instances of useV2Keypad

SonicScrewdriver:
## Summary:
<Add your summary here>

Issue: XXX-XXXX

## Test plan:


SonicScrewdriver:
Ooo I see I updated a couple files in the legacy keypad here that codecov is update there's no tests for. 

That makes me wonder ... should I remove the legacy keypad entirely as part of this PR? We're no longer using it at all. :)

PR #795, opened 2023-11-08
Add "do not track" query parameter to vimeo links in the video widget to ensure tracking/analytics cookies don't get set.

ioanacrant:
## Summary:
Issue: https://khanacademy.atlassian.net/browse/INFRA-9620

## Test plan:
- tested that parameter got added in video widget in storybook


jeremywiebe (commented):
Could you add a unit test or two for this change?  You should be able to use the tests in `video.test.ts` as a guide and then assert with something like: 

```js 
expect(document.getElementsByTagName("iframe")[0].src).toContain(
    "dnt=1",
);
```

packages/perseus/src/widgets/video.tsx
```diff
@@ -75,6 +76,16 @@ class Video extends React.Component<Props> {
...
             url = location;
+            if (IS_VIMEO.test(url)) {
+                // If this is a vimeo video, we need to add the query string
+                // parameter "dnt" so that analytics/tracking cookies aren't set.
+                // https://help.vimeo.com/hc/en-us/articles/12426260232977-Player-parameters-overview
+                if (url.indexOf("?") === -1) {
+                    url += "?dnt=1";
+                } else {
+                    url += "&dnt=1";
+                }
```
handeyeco, line 87: No action needed: I think this is fine to ship as-is, but there's also the [URLSearchParams API](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) now.

handeyeco (approved):
LGTM!

PR #797, opened 2023-11-09
Update the color of rationales

nixterrimus:
## Summary:
Updates the color of rationales to match wonderblocks

Issue: https://khanacademy.atlassian.net/browse/LC-1137

## Test plan:
- Verify it looks ok

### DIFF
--- a/packages/perseus/src/widgets/__tests__/__snapshots__/radio.test.ts.snap
+++ b/packages/perseus/src/widgets/__tests__/__snapshots__/radio.test.ts.snap
@@ -3638,7 +3638,7 @@ exports[`single-choice question reviewMode: true should snapshot the same: first
                       </button>
                     </div>
                     <div
-                      class="perseus-radio-rationale-content rationale_1w3tmyh"
+                      class="perseus-radio-rationale-content rationale_1ctj0s0"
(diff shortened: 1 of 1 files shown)


packages/perseus/src/widgets/radio/choice.tsx
```diff
@@ -346,6 +346,7 @@ const styles = StyleSheet.create({
         padding: intermediateCheckboxPadding,
         paddingTop: 0,
         marginLeft: 54,
+        color: Color.offBlack64,
```
jeremywiebe, line 349: There are two references to rationales in our `.less` styles: 

1. https://github.com/Khan/perseus/blob/cfff6ad16a3923418a3c75816af7c99689b94da4/packages/perseus/src/styles/articles.less#L237
2. https://github.com/Khan/perseus/blob/cfff6ad16a3923418a3c75816af7c99689b94da4/packages/perseus/src/styles/articles.less#L386

These are in Article styles. Just wanted to check if you verified this in Articles. The styles you've added here likely override those .less styles, but it might be useful to clean them up just so we don't introduce a regression at some point in the future if we move this color style for some reason.

jeremywiebe (approved):
Good stuff! Just one comment related to what looks like related `.less` styles for rationales.

PR #800, opened 2023-11-14
Conditionally switch decimal separator based on locale

handeyeco:
## Summary:
We were supposed to be switching between `,` and `.`, but we didn't see that through when updating how we're rendering icons for the v2 keypad. This updates that and adds some unit tests.

Screenshot 2023-11-14 at 10 21 29 AM
Screenshot 2023-11-14 at 10 20 55 AM

Issue: LC-1453

## Test plan:
- Use or simulate mobile web
- Go to an exercise with a MathInput
- Open the keypad by focusing the MathInput
- Note a `.` as the decimal separator (if you're in the US)
- Change your locale?
- Use or simulate mobile web
- Go to an exercise with a MathInput
- Open the keypad by focusing the MathInput
- Note a `,` as the decimal separator

### DIFF
--- a/.changeset/spotty-snails-serve.md
+++ b/.changeset/spotty-snails-serve.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/math-input": patch
+---
+
+Bugfix: conditionally switch between period/comma for decimal separator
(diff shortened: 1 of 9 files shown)


nedredmond (approved):
I guess the actual localization is handled in Webapp? I don't see here where we are conditionally switching based on locale.

handeyeco:
> I guess the actual localization is handled in Webapp? I don't see here where we are conditionally switching based on locale.

It's based off of `wonder-blocks-i18n`. The logic already existed from the v1 keypad days, we just needed to tweak some things to make it work with the v2 keypad.

```JS
// utils.ts
import {getDecimalSeparator} from "@khanacademy/wonder-blocks-i18n";
```

PR #790, opened 2023-11-03
Upgrade to latest Storybook (v7.5.3)

jeremywiebe:
## Summary:

What fun is a Hackathon if you don't upgrade some infra?? 

Steps I took: 
  * Upgrade all the babel stuff (fixes broken storybook issue that cause babel transpilation to fail)
  * Remove deprecated/unused useDraftEditor API option
  * Remove deprecated/unused inModal API option
  * npx storybook@latest upgrade
  * new-frameworks migration
  * storybook-binary migration
  * sb-scripts migration
  * autodocsTrue migration
  * wrap-require migration
  * Ensure css-loader config is correct (and safely replaced in main.js)

Screenshot 2023-11-03 at 12 09 53 PM

Issue: "hackathon"

## Test plan:

`yarn start` 😍

PR #802, opened 2023-11-15
[Storybook] Fix styles in editor stories

jeremywiebe:
## Summary:

After upgrading to Storybook v7 (#790), Nicole noticed that some of our styles in Editor stories was missing/broken. 

This PR fixes it with the following changes (not all required, but the move for `main.ts` => `main.ts` seemed really easy and very useful!):

  * 🛠️ Add explicit deps for less-loader, css-loader, and style-loader. Previously, we depended on these modules indirectly and didn't have explicit control over which version of each we had. Turns out we had some really old versions and upgrading them fixed some of the issues.
  * ✨ Migrate Storybook config to TS. 
  * ✨ Use Storybook config types and fix up framework ref
  * 🛠️ Make less-loader work as it used to (eager math evaluation)
  * 🛠️ Ensure perseus-editor.less is imported in Editor demo stories

image

## Dependency Details

I've included `yarn why` output for both `style-loader` and `css-loader` modules, which are part of this PR (for historical purposes). Both of these were "anchored" to a super-old version because of the `size-limit` dependency in the `simple-markdown` package. We don't use or need that package anymore so I've removed it. I think it's safe to upgrade them as we don't use them to build our production bundles (we use Rollup for that and not Webpack!).

<details>
<summary>
<code>style-loader</code>
</summary>

```sh
$ yarn why style-loader
yarn why v1.22.19
[1/4] 🤔  Why do we have the module "style-loader"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "style-loader@0.23.1"
info Reasons this module exists
   - "_project_#@khanacademy#simple-markdown#size-limit" depends on it
   - Hoisted from "_project_#@khanacademy#simple-markdown#size-limit#style-loader"
info Disk size without dependencies: "404KB"
info Disk size with unique dependencies: "756KB"
info Disk size with transitive dependencies: "3.39MB"
info Number of shared dependencies: 12
=> Found "@storybook/builder-webpack5#style-loader@3.3.3"
info This module exists because "_project_#@storybook#react-webpack5#@storybook#builder-webpack5" depends on it.
info Disk size without dependencies: "132KB"
info Disk size with unique dependencies: "132KB"
info Disk size with transitive dependencies: "132KB"
info Number of shared dependencies: 0
✨  Done in 0.48s.
```

</details>

<details>
<summary>
<code>css-loader</code>
</summary>

```sh
$ yarn why css-loader
yarn why v1.22.19
[1/4] 🤔  Why do we have the module "css-loader"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "css-loader@2.1.1"
info Reasons this module exists
   - "_project_#@khanacademy#simple-markdown#size-limit" depends on it
   - Hoisted from "_project_#@khanacademy#simple-markdown#size-limit#css-loader"
info Disk size without dependencies: "700KB"
info Disk size with unique dependencies: "3.31MB"
info Disk size with transitive dependencies: "7.15MB"
info Number of shared dependencies: 26
=> Found "@storybook/builder-webpack5#css-loader@6.8.1"
info This module exists because "_project_#@storybook#react-webpack5#@storybook#builder-webpack5" depends on it.
✨  Done in 0.51s.
```

</details>

Issue: LC-1459

## Test plan:

Run `yarn; yarn start`
Navigate to **Perseus** >> **Editor** >> **Demo** and compare to the Jira ticket's "Before" screenshot. It should match!


.storybook/main.ts
```diff
@@ -4,7 +4,11 @@ const path = require("path");
 const fs = require("fs");
 const glob = require("fast-glob");
 
-module.exports = {
+import type {StorybookConfig} from "@storybook/react-webpack5";
+
+const config: StorybookConfig = {
+    framework: "@storybook/react-webpack5",
```
jeremywiebe, line 10: This was at the bottom of the file, but to me, this feels like the most important thing so I've moved it to the top!

.storybook/main.ts
```diff
@@ -19,9 +23,9 @@ module.exports = {
     ],
 
     addons: [
-        getAbsolutePath("@storybook/addon-links"),
-        getAbsolutePath("@storybook/addon-essentials"),
-        getAbsolutePath("@storybook/addon-a11y"),
+        "@storybook/addon-links",
+        "@storybook/addon-essentials",
+        "@storybook/addon-a11y",
```
jeremywiebe, line 28: I created `getAbsolutePath()` in the original PR where we upgraded to Storybook v7, but I can't recall why and removing it works just fine!  😖

.storybook/main.ts
```diff
@@ -93,7 +97,10 @@ module.exports = {
                                 loader: "css-loader",
                                 options: {url: false},
                             },
-                            "less-loader",
+                            {
+                                loader: "less-loader",
+                                options: {lessOptions: {math: "always"}},
```
jeremywiebe, line 102: The newer version of Less that we now use does math slightly differently, so we need to set this setting to cause it to work like the older Less.

packages/perseus-editor/src/__stories__/editor.stories.tsx
```diff
@@ -11,6 +11,8 @@ import {registerAllWidgetsAndEditorsForTesting} from "../util/register-all-widge
 
 import type {PerseusRenderer} from "@khanacademy/perseus";
 
+import "../styles/perseus-editor.less";
```
jeremywiebe, line 14: This was a big part of the fix. I'm not 100% sure why it worked before, but we have to ensure we're actually importing the `.less` file that provides the style. I suspect it's a bundling issue where the new Webpack bundler (v5 compared to v4) skips the entry `index.ts` file which has the `.less` import. 🤷‍♂️

packages/simple-markdown/package.json
```diff
@@ -28,9 +28,7 @@
     "devDependencies": {
         "@types/react-dom": ">=16.0.0",
         "perseus-build-settings": "^0.2.1",
-        "size-limit": "^0.21.1",
-        "typescript": "^3.6.4",
-        "uglify-js": "^3.6.7"
```
jeremywiebe, line 33: `size-limit` and `uglify-js` are no longer used by `simple-markdown`!

nedredmond (approved):
Thanks Jeremy! Next time I'll ticket when I see it so that we could have handled it while you were out. Sorry for laying the burden on you!

PR #805, opened 2023-11-15
Reduce motion on label-image-widget dot

nixterrimus:
## Summary:
Respects a learners "reduce motion" setting.

https://github.com/Khan/perseus/assets/18454/0ace1667-a2c1-40e5-9427-9941dc5082cd

Issue: https://khanacademy.atlassian.net/browse/LC-1421

## Test plan:
- Load the the label image widget
- The dots should animate infinitely
- Turn on reduce motion
- Reload the page
- The dots should animation once

### DIFF
--- a/.changeset/wet-months-teach.md
+++ b/.changeset/wet-months-teach.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Reduce motion on dot animation when option is enabled


packages/perseus/src/widgets/label-image/marker.tsx
```diff
@@ -29,6 +29,11 @@ type State = {
     isFocused: boolean;
 };
 
+function shouldReduceMotion(): boolean {
+    const mediaQuery = window.matchMedia("(prefers-reduced-motion: reduce)");
+    return !mediaQuery || mediaQuery.matches;
```
nedredmond, line 34: Why default to reduced motion when no media query is found?

packages/perseus/src/widgets/label-image/marker.tsx
```diff
@@ -227,10 +235,20 @@ const styles = StyleSheet.create({
 
         animationDirection: "alternate",
         animationDuration: "0.8s",
-        animationIterationCount: "infinite",
         animationTimingFunction: "ease-in",
 
         transformOrigin: "50% 50%",
+
+        animationIterationCount: "0",
```
nedredmond, line 242: Is this line necessary when you'll always be providing one of the other two classes?

nedredmond (commented):
Thanks Nicole! Just please confirm whether the line that defaults to reduced motion when the setting is undefined is intentional, or am I misunderstanding something?

packages/perseus/src/widgets/label-image/marker.tsx
nedredmond: Thanks for clarifying. That makes sense!

packages/perseus/src/widgets/label-image/marker.tsx
nedredmond: I see. That does make sense to me, but I would like it if we had confirmation from Sarah B. I'll approve and humbly request you run this choice by our UX designer. Thanks Nicole!

nedredmond (approved):
Please run the unsupported default by Sarah B. Thanks again!

PR #808, opened 2023-11-17
Cleanup #problemarea CSS

SonicScrewdriver:
## Summary:
This PR fixes our #problemarea width and min height by removing the base setting. This CSS is not used anywhere, as the only two places the #problemarea exists (ItemEditor and in the mobile application) are both having to override these settings back to the baseline.  

Issue: LC-1475

## Test plan:
manual testing + confirming code use.

### DIFF
--- a/packages/perseus-editor/src/styles/perseus-editor.less
+++ b/packages/perseus-editor/src/styles/perseus-editor.less
@@ -279,15 +279,6 @@
 
     // Question area styles
     #problemarea {
-        /* Override the min-height property used for displaying the problem.
-        While editing, the size of the editor itself is a better constraint on
-        the size of the problemarea, and having a smaller problemarea is nicer
-        if it means it aligns better with the editor. */
-        min-height: 0;
-
-        // Force the hint width to take up the width of its containing viewport.
-        width: 100%;
-
         #workarea {
             margin: 0; // override ke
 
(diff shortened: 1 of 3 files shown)


SonicScrewdriver:
I have a few screenshots but didn't want to dirty up the main PR message. We seem to be overriding this style a lot across our many repos, and I suspect most of them are not needed at all anymore considering that the only actual implementation of this id is in our exercises in the mobile application, and the ItemEditor in Perseus. 

Usage of #problemarea in Perseus:
Screenshot 2023-11-17 at 1 47 45 PM

Usage of #problemarea in Webapp (can probably be cleaned up but that's out of scope of our mobile app work): 
Screenshot 2023-11-17 at 1 48 40 PM

Usage of #problemarea in our Mobile App:
Screenshot 2023-11-17 at 1 48 59 PM

SonicScrewdriver:
Hmm I distrust these Cypress failures as the interactive graph should have absolutely nothing to do with the #problemarea, which doesn't exist anywhere in Perseus outside of the ItemEditor, which is only imported into the Editor Page.

packages/perseus-editor/src/styles/perseus-editor.less
```diff
@@ -279,15 +279,6 @@
 
     // Question area styles
     #problemarea {
-        /* Override the min-height property used for displaying the problem.
```
jeremywiebe, line 282: I _think_ that once we're fully on `ServerItemRenderer` we won't have a `problemarea` anywhere anymore.  Looking around in this repo, I see it being rendered by the `item-editor.tsx` and nowhere else.

jeremywiebe (approved):

PR #811, opened 2023-11-21
Sync all Wonder-Blocks dependencies

jeremywiebe:
## Summary:

This PR updates all Wonder Blocks dependencies to be their latest release and ensures that each Perseus package depends on the same version of each Wonder Blocks package. Today all WB dependencies are `peerDependencies` which means that the host application that uses Perseus decides what version of Wonder Blocks that Perseus sees.

As part of these changes, I needed to do a few other changes to ensure it all works:

  * Bump all WB dependencies to latest and ensure they're in sync across all Perseus packages
  * Bring in svg transform for Jest to be able to run tests touching Phosphor icons (in WB)
  * Update snapshots now that we are on WB packages that use Phosphor icons
  * Fix HintRenderer stories (migrates to v7 Storybook format)

More details here: https://khanacademy.slack.com/archives/C0637C68TM0/p1700520252092829?thread_ts=1700513566.299299&cid=C0637C68TM0

I used a little Fish shell magic to do the actual bumping:

<details>
<summary>Fish shell scriptlet</summary>

```sh
# Update all wonder-blocks packages to the latest version.
for pkg in packages/*/package.json
    echo $pkg
    set --local wb_deps (jq -r '.devDependencies | keys[] | select(startswith("@khanacademy/wonder-"))' $pkg)
    set -a --local wb_deps (jq -r '.peerDependencies | keys[] | select(startswith("@khanacademy/wonder-"))' $pkg)

    if test (count $wb_deps) -gt 0
        pushd (dirname $pkg)
        yarn --quiet add --dev $wb_deps
        yarn --quiet add --peer $wb_deps
        popd
    end
end

```
</details>

Issue: LC-1489

## Test plan:

I ran tests: `yarn test`
I ran all Cypress tests: `yarn cypress:ci`
Checked most Storybook stories: `yarn start` and click around 

Note that there is one Storybook story that blows up, but it's related to the legacy keypad and its use of SVG icons in this repo (not Phosphor icons) and I didn't think it was worth going down the rabbit hole to fix them.


config/test/svg.transform.js
```diff
@@ -0,0 +1,25 @@
+const path = require("path");
```
jeremywiebe, line 1: This file imported directly from webapp.

packages/perseus/src/__stories__/hints-renderer.stories.tsx
```diff
@@ -3,56 +3,53 @@ import React from "react";
...
-type StoryArgs = {
-    hints: ReadonlyArray<any>;
-    hintsVisible: number;
-};
-
-type Story = {
-    title: string;
-    argTypes: any;
-};
+import type {Meta, StoryObj} from "@storybook/react";
```
jeremywiebe, line 6: This story was crashing trying to access `this.props.hints.length` in the `HintsRenderer`. I think it's because in Storybook v7 the origin `export default` from this file wasn't compatible.

packages/perseus/src/__tests__/__snapshots__/renderer.test.tsx.snap
```diff
@@ -367,18 +367,10 @@ exports[`renderer snapshots correct answer: correct answer 1`] = `
                 >
                   less than or equal to
                 </span>
-                <svg
+                <span
```
jeremywiebe, line 370: All these snapshot updates are because WB now uses Phosphor icons.

somewhatabstract (approved):
This looks right to me though I have only given it a cursory glance. Without a clear methodology/automation, or manually verifying version numbers, it's hard to know that it's "right".

Re: dev and peer deps
I know that in other repos, we tend to leave the package.json of each package to just have `peerDependencies` and then have the full list as dev dependencies in the root package.json.

Re: webapp tooling
The webapp tool is only built for one package.json. My plan, when I get it moved to `wonder-stuff-tools` is to have it be a little more versatile and understand our monorepos a little better.

PR #812, opened 2023-11-22
Remove enzyme and translate any remaining tests to RTL

jeremywiebe:
## Summary:

While working on our WB deps issues, I noticed we still use Enzyme for a few tests.  Two were `xit()` (ignored) in the math-input project so I deleted them as they were testing our legacy keypad components. The rest were easily migrated to our utility `renderQuestion()` function and Testing Library!

Issue: "none"

## Test plan:

`yarn test`


config/test/custom-matchers.ts
```diff
@@ -1,17 +1,9 @@
 // Ok here we are
 
-type Score =
```
jeremywiebe, line 3: This type didn't match our actual Score type.

config/test/custom-matchers.ts
```diff
@@ -32,7 +24,7 @@ expect.extend({
                 message: () => `Problem was not fully answered`,
             };
         }
-        if (score.earned !== 1) {
```
jeremywiebe, line 35: And incorrect score is one where the `earned` is not equal to `.total`.  Some questions have more than one "point" available.

packages/perseus/src/renderer.tsx
```diff
@@ -132,7 +132,7 @@ export type Widget = {
         path: FocusPath,
         newValue: string,
         // TODO(jeremy): I think this is actually a callback
-        focus: () => unknown,
+        focus?: () => unknown,
```
jeremywiebe, line 135: This callback/focus function is truly optional.  The API tests in this PR illustrate that by not passing it sometimes.

benchristel (approved):
LGTM! Thanks!

And thanks for the inline comments. Those were really helpful.

PR #813, opened 2023-11-23
Remove unnecessary Z-Index for #problemarea buttons/inputs

SonicScrewdriver:
## Summary:
On our mobile app, we're noticing that hints are rendering above our keypad due to this style that sets the z-index to 3. I don't think we need z-index here at all as the #problemarea is only used on Mobile and in the content editor - neither of which need the index. 

Rather than using .perseus-interactive, I'm now simply setting the position directly. The relative positioning is still important to ensure that the "+" symbol appears beside the "Next Hint" button. 

Issue: LC-1445

## Test plan:
manual

### DIFF
--- a/packages/perseus/src/styles/perseus-renderer.less
+++ b/packages/perseus/src/styles/perseus-renderer.less
@@ -96,13 +96,6 @@
 }
 
 .framework-perseus {
-    #problemarea {
-        input,
-        button {
-            .perseus-interactive;
-        }
-    }
-
     div.paragraph {
         font-family: "Lato", sans-serif;
         font-weight: 400;
--- a/packages/perseus/src/styles/perseus-renderer.less
+++ b/packages/perseus/src/styles/perseus-renderer.less
@@ -96,13 +96,6 @@
 }
 
 .framework-perseus {
-    #problemarea {
-        input,
-        button {
-            position: relative;
-        }
-    }
-
     div.paragraph {
         font-family: "Lato", sans-serif;
         font-weight: 400;
--- a/packages/perseus/src/styles/perseus-renderer.less
+++ b/packages/perseus/src/styles/perseus-renderer.less
@@ -96,6 +96,13 @@
 }
 
 .framework-perseus {
+    #problemarea {
+        input,
+        button {
+            position: relative;
+        }
+    }
+
     div.paragraph {
         font-family: "Lato", sans-serif;
         font-weight: 400;
--- a/.changeset/green-ligers-roll.md
+++ b/.changeset/green-ligers-roll.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Removing unused CSS from #problemarea.


packages/perseus/src/styles/perseus-renderer.less
```diff
@@ -99,7 +99,7 @@
     #problemarea {
         input,
         button {
-            .perseus-interactive;
+            position: relative;
```
jeremywiebe, line 102: Doing a grep through our codebase, the only place I see that we're rending an element with this class name is in the `item-editor.tsx` (which is only used for content editing). 

I'm wondering if this style is even needed or the removal of the `.perseus-interactive` bit is what's fixing it (rather than addition of `position:relative`). 

```sh 

$ rg problemarea
packages/perseus/src/styles/perseus-renderer.less
99:    #problemarea {

packages/perseus/src/styles/khan-exercise.css
131:#problemarea {
309:#problemarea a:link,
310:#problemarea input,
311:#problemarea label,  /* HACK(aria): for radios */
312:#problemarea select {
1062:#problemarea div#congruent-triangles {
1204:#problemarea .quarter-graph {
1210:#problemarea .estimation span {
1213:#problemarea .estimation span:first-child {
1235:#problemarea div.z-score-table > span {
1241:#problemarea div.z-score-table > span:first-child {
1247:#problemarea div.focus-information {
1253:#problemarea div.focus-information div {
1256:#problemarea div.focus-information-column-left {
1260:#problemarea div.focus-information-column-right {
1265:#problemarea .reading-tables > span {
1269:#problemarea .reading-tables > span:first-child {
1274:#problemarea .reading > span {
1280:#problemarea .reading.fake_header > span {
1284:#problemarea .reading > span:first-child {
1290:#problemarea .problem .graph-caption {

packages/perseus-editor/src/item-editor.tsx
121:                        <div id="problemarea">

packages/perseus-editor/src/styles/perseus-editor.less
281:    #problemarea {

$ _
```

SonicScrewdriver: I'm open to either option here! Although perhaps you're right and it's best to completely remove this CSS from Perseus, and to have the override/style in the mobile repo. I originally thought that both might need the position: relative, but we're rending hints quite a bit differently in the content editor.  Digging deeper, I'm not actually able to find any inputs in the content editor that is nested properly enough to get this style either.

I'll remove this style completely! 

NOTE: If we upgrade perseus in the mobile app before [Ben's PR is landed](https://github.com/Khan/mobile/pull/3457), hints will be broken on lc-release-candidate (see screenshot below). His PR manually sets the relative positioning that we need in the simple-button.css file he brought back. This isn't an actual issue, just wanted to make sure I keep track of all the moving parts. :) 

Some Information about why the relative positioning matters below:

We're also using #problemarea on our Mobile App, and removing the the relative positioning results in the following view, where the "+" becomes disconnected from the hint: 
Screenshot 2023-11-23 at 2 00 35 PM
(There are only two properties being set in .perseus-interactive: z-index and our relative positioning.)

jeremywiebe (approved):
This feels like a better fix especially for future maintenance. Thanks Third!

PR #815, opened 2023-11-24
[Docs] Architecture

jeremywiebe:
## Summary:

This PR starts to build out some documentation about how Perseus works. I've left some placeholder headers for topics I plan to write on in the future (so don't expect everything to be filled in in this PR :) )

Issue: "none"

## Test plan:

Read the docs


handeyeco:
Third showed this to me. This is cool, might be worth merging in even in it's incomplete state so it doesn't get lost.

PR #816, opened 2023-11-24
Wrap SvgImage in AssetContext's

jeremywiebe:
## Summary:

While trying to figure out why some Cypress tests weren't showing a background image on graphs, I discovered that we don't wrap all of our `<SvgImage />` components in an `AssetContext`. 

The `AssetContext` is used to communicate back up to the host application when things that happen asynchronously are completed. This allows the host to wait until all of the asynchronous assets are loaded/rendered (such as loading images or rendering Math). 

Wrapping these instances seems to fix the Cypress tests not showing background images. 

Issue: "none"

## Test plan:


packages/perseus/src/components/svg-image.tsx
```diff
@@ -180,6 +180,16 @@ type Props = {
     setAssetStatus: (assetKey: string, loaded: boolean) => void;
 };
 
+type DefaultProps = {
```
jeremywiebe, line 183: Just moved this up to be a named type (our convention in this repo) as opposed to an anonymous, inline type below.

.gitignore
```diff
@@ -8,4 +8,4 @@ coverage-ts
 storybook-static
 .nyc_output
 *.tsbuildinfo
-cypress/
+/cypress/
```
jeremywiebe, line 11: Without the leading `/`, we are ignoring any `cypress` anywhere in the repo... and our Cypress configuration lives in `config/cypress`. We don't want to ignore that dir!

config/cypress/cypress.config.js
```diff
@@ -32,11 +32,16 @@ const resourceTypes = /\.(png|svg|eot|woff|woff2|ttf|otf|svg(#Symbola)?)$/;
 module.exports = defineConfig({
     fixturesFolder: false,
     video: false,
-
+    // Prevent Cypress from scrolling to elements before clicking them.
+    scrollBehavior: false,
```
jeremywiebe, line 36: This prevents Cypress from auto-scrolling to the node we're clicking... which was messing up our coordinate calculations of where we wanted to drag to!

config/cypress/support.js
```diff
@@ -11,6 +11,20 @@ if (Cypress.env("CYPRESS_COVERAGE")) {
...
+// but I couldn't get it working in a short timebox. Leaving this breadcrumb
+// trail for future!
+// https://docs.cypress.io/guides/tooling/typescript-support#Clashing-Types-with-Jest
+// declare global {
+//     namespace Cypress {
+//         interface Chainable {
+//             dragTo(position: {x: number; y: number}): Chainable<any>;
+//         }
+//     }
+// }
```
benchristel, line 26: Should this have a TODO linked to a Jira issue? Otherwise, I suspect we'll probably forget to come back to this.

testing/render-question-with-cypress.tsx
```diff
@@ -31,12 +31,20 @@ const renderQuestion = (
         [assetKey: string]: boolean;
     } = {};
     const setAssetStatus = (assetKey: string, loaded: boolean) => {
+        console.log("setAssetStatus", assetKey, loaded);
```
benchristel, line 34: Should we remove these `console.log` calls before landing? (this one and line 39 below)

Seems like the changes to this file can just be reverted.

benchristel (approved):
I left a few minor comments inline, but once those are addressed, I think this will be good to go!

config/cypress/support.js
jeremywiebe: Good call.  LC-1495 now exists to cover migrating our Cypress configs to Typescript.

testing/render-question-with-cypress.tsx
jeremywiebe: Good catch. Thanks!

PR #817, opened 2023-11-25
[Spike] JSON Schema generation

jeremywiebe:
## Summary:

**SPIKE: DO NOT MERGE** 

While watching the Perseus Data Format meeting recording, I remembered an idea that we've batted around to generate a JSON Schema for our Perseus data format using the existing Typescript types. 

This PR introduces a package.json script `gen:schema` which generates a JSON schema for the `PerseusItem` type as a proof of concept. 

This wouldn't address all of our API needs, but it might help with external partners who consume the Perseus data format. 

Issue: 'none'

## Test plan:

### DIFF
--- a/.swcrc
+++ b/.swcrc
@@ -1,0 +1,15 @@
+{
+    "$schema": "https://swc.rs/schema.json",
+    "jsc": {
+        "parser": {
+            "syntax": "typescript",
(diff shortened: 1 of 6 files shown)


jeremywiebe:
Ben wrote a custom parsing system that does most of this now.

PR #818, opened 2023-11-27
In label image content editor add preferred direction

nixterrimus:
## Summary:
Adds a dropdown for preferred pop-over direction to the content editor. This value will get passed into the component for label image and label image will do its best to accomodate the preference with the available space.

Screenshot_2023-11-26_at_8_46_51_PM

Note: There are several follow on pieces that are not addressed in this PR
- Adding fields in webapp so that this option validates
- Making use of the preferred popover direction key on the learner-side

Issue: https://khanacademy.atlassian.net/browse/LC-1424

## Test plan:
- In storybook load the content editor for label image widget
- **There should be an option to choose preferred direction**

### DIFF
--- a/packages/perseus-editor/src/widgets/label-image/__stories__/behavior.stories.tsx
+++ b/packages/perseus-editor/src/widgets/label-image/__stories__/behavior.stories.tsx
@@ -27,5 +27,5 @@ export const Default = (args: StoryArgs): React.ReactElement => {
             setState({...state, ...newState});
         },
     } as const;
-    return <Behavior {...props} />;
+    return <Behavior preferredPopoverDirection="NONE" {...props} />;
 };


packages/perseus-editor/src/widgets/label-image/behavior.tsx
```diff
@@ -59,6 +70,27 @@ const Behavior = ({
...
+                <select
+                    value={preferredPopoverDirection}
+                    onChange={(e) => {
+                        onChange({
+                            preferredPopoverDirection: e.target
+                                .value as PreferredPopoverDirection,
+                        });
+                    }}
+                >
+                    <option value="NONE"> No Prference </option>
```
nedredmond, line 87: Typo in "Preference"

nedredmond (approved):
Thanks!

PR #820, opened 2023-11-28
Switch answer pill to WB Pill to fix WCAG violation

nedredmond:
## Summary
Replaces popover pill with implementation from WB Pill and `react-popper`, which is the same dep that powers WB's Tooltip and Popover; see https://github.com/Khan/wonder-blocks/blob/main/packages/wonder-blocks-tooltip/src/components/tooltip-popper.tsx

Fixes issue with context being forcibly changed when Hide Answers toggle is switched on or off

Issue: LC-1488

## Test
Using keyboard to activate/deactivate Hide Answers toggle doesn't force focus into answers/markers on activation


packages/perseus/package.json
```diff
@@ -64,9 +66,12 @@
         "prop-types": "^15.6.1",
         "react": "16.14.0",
         "react-dom": "16.14.0",
+        "react-popper": "^2.0.0",
```
nedredmond, line 69: This is what we use to power Tooltip and Popover, so not a new dependency for KA.
WB exposes a thin wrapper `TooltipPopper`, but there is a ticket to limit it to only Tooltip and Popover types, so it would break for us within a few versions.

packages/perseus/src/styles/widgets/label-image.less
```diff
@@ -5,6 +5,17 @@
             margin: 0;
         }
     }
+    [id^="answer-choice-"] {
+        // Reset text content.
+        div.paragraph {
+            all: revert;
```
nedredmond, line 11: https://developer.mozilla.org/en-US/docs/Web/CSS/all

packages/perseus/src/widgets/label-image.tsx
```diff
@@ -644,12 +645,21 @@ class LabelImage extends React.Component<LabelImageProps, LabelImageState> {
                 mediaQueries.xsOrSmaller.replace("@media ", ""),
             ).matches;
 
-            let side;
+            let side: "bottom" | "left" | "right" | "top";
+            let markerPosition;
```
nedredmond, line 649: Previously, we calculated the position of the marker relative to the popup but didn't write it to a variable. Storing now to use in dynamic CSS.

packages/perseus/src/widgets/label-image.tsx
```diff
@@ -658,21 +668,31 @@ class LabelImage extends React.Component<LabelImageProps, LabelImageState> {
...
             const answerChoicesActive = index === activeMarkerIndex;
 
-            const answerStyles: CSSProperties = {
+            const adjustPopoverDistance: CSSProperties = {
                 [side]: 15, // move the popover closer to the marker
             };
 
+            const adjustPillDistance: CSSProperties = {
+                [`margin${
+                    markerPosition.charAt(0).toUpperCase() +
```
nedredmond, line 682: I have to capitalize the first letter (i.e., `marginRight` to work as CSSProperties object.

PR #821, opened 2023-11-28
Remove console mocking in tests

jeremywiebe:
## Summary:

When we moved Perseus out of webapp we brought along a mechanism to mock `console.*()` calls. I think the original intent was to make it an explicit choice to have console logging in our production code. 

In reality, this has caused more headaches than it has saved us. Also, our build system can strip out console logging if we need that in the future. 

Issue: "none"

## Test plan:


nedredmond (approved):
🙏

PR #823, opened 2023-11-28
For wide images, label image popovers go above or below markers to lessen chances of covering markers

nedredmond:
## Summary
If an image is wider than it is tall, popovers and pills can only appear above or below markers.
Lessens changes of pills overlapping markers or each other.
Adds the number line story from Test Everything to verify.

Issue: LC-1504

Screenshot 2023-11-28 at 5 28 54 PM

https://github.com/Khan/perseus/assets/23404711/84093a82-d2eb-498c-8daf-553d1f164660

## Test strategy
Check out the story.

### DIFF
--- a/packages/perseus/src/widgets/label-image.tsx
+++ b/packages/perseus/src/widgets/label-image.tsx
@@ -645,11 +645,15 @@ class LabelImage extends React.Component<LabelImageProps, LabelImageState> {
                 mediaQueries.xsOrSmaller.replace("@media ", ""),
             ).matches;
 
+            // Determine whether the image is wider than it is tall.
+            const isWideImage =
+                this.props.imageWidth / 2 > this.props.imageHeight;
+
             let side: "bottom" | "left" | "right" | "top";
             let markerPosition;
             // Position popup closest to the center, preferring it renders
             // entirely within the image area.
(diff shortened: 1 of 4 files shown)


packages/perseus/src/widgets/label-image.tsx
```diff
@@ -645,11 +645,15 @@ class LabelImage extends React.Component<LabelImageProps, LabelImageState> {
                 mediaQueries.xsOrSmaller.replace("@media ", ""),
             ).matches;
 
+            // Determine whether the image is wider than it is tall.
+            const isWideImage =
+                this.props.imageWidth / 2 > this.props.imageHeight;
```
jeremywiebe, line 650: **question**: Is this a heuristic for deciding if this is essentially a number line question (more than twice as wide as tall)?

jeremywiebe (approved):

packages/perseus/src/widgets/label-image.tsx
nedredmond: Kind of! Also looks like our algorithm doesn't account for wide images.

This behavior isn't just to correct for number line images. Any wide images would put pills on the side towards the center (covering each other up) due to the algo only accounting for tall images.

PR #824, opened 2023-11-29
Pill color for correct & incorrect answers

nedredmond:
## Summary
Match marker color for correct & incorrect answers.
Also disables interaction for correct answers, matching marker behavior.

Isolates component and adds interactive story doc.

Issue: LC-1498

Screenshot 2023-11-29 at 10 28 25 AM

https://github.com/Khan/perseus/assets/23404711/64157cdc-ed97-4546-89e0-a5382ab3d2a5

## Testing Strategy:
Fire up storybook and play in the new story

### DIFF
--- a/packages/perseus/src/widgets/label-image/answer-pill.tsx
+++ b/packages/perseus/src/widgets/label-image/answer-pill.tsx
@@ -11,7 +11,7 @@ export const AnswerPill = (props: {
     id: string;
     selectedAnswers: readonly string[];
     showCorrectness?: "correct" | "incorrect";
-    markerRef: HTMLElement;
+    markerRef?: HTMLElement;
     side: "top" | "bottom" | "left" | "right";
     onClick: () => void;
     style?: CSSProperties;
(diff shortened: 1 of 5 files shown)


packages/perseus/src/widgets/label-image/answer-pill.tsx
```diff
@@ -0,0 +1,93 @@
...
+    );
+};
+
+const styles = StyleSheet.create({
+    bringToFront: {
+        boxShadow: `0 8px 8px ${Color.offBlack64}`,
+        zIndex: 1000,
+    },
+    correct: {
+        backgroundColor: "#00880b", // WB green darkened by 18%
```
jeremywiebe, line 88: **suggestion**: There is a WB function `fade()` that might encode this more clearly: https://github.com/Khan/wonder-blocks/blob/main/packages/wonder-blocks-color/src/util/utils.ts#L75

I think this colour could then be: 

```suggestion
        backgroundColor: fade(Color.green, 0.18),
```

jeremywiebe (approved):
Nice. Thanks for the test coverage and new stories!

packages/perseus/src/widgets/label-image/answer-pill.tsx
nedredmond: @jeremywiebe `fade` does the opposite-- it adds transparency to the color. Maybe I could use `mix` with black instead?

jeremywiebe: Oops, sorry for the rabbit hole. :)

PR #825, opened 2023-11-29
Remove the legacy keypad and replace it with MobileKeypad

benchristel:
## Summary:
Previously, the KeypadSwitch component let you toggle between the new
and old keypads. Now we always use the new keypad.

Issue: https://khanacademy.atlassian.net/browse/LC-739

Test plan:

`yarn test`


packages/math-input/src/full-mobile-input.stories.tsx
```diff
@@ -104,10 +110,17 @@ export function Wrapped() {
...
+                    keypadElement,
+                    setKeypadElement,
+                    keypadActive,
+                    setKeypadActive,
+                }) => (
                     <Basic
                         keypadElement={keypadElement}
                         setKeypadElement={setKeypadElement}
+                        keypadActive={keypadActive}
+                        setKeypadActive={setKeypadActive}
```
benchristel, line 123: To avoid breaking compatibility with existing callers, it might be better to wrap the MobileKeypad in a component that uses the `KeypadContext.Consumer` to get `keypadActive` and `setKeypadActive`. @handeyeco do you think that would be an improvement?

handeyeco: Yeah, that's definitely the way I'd like to go. See [this ticket](https://khanacademy.atlassian.net/browse/LC-1340):

> What’s even nicer is that this could likely all happen within Perseus - rather than passing things around in Webapp, we can pass things around in Perseus using Context; the Perseus widget passes a config up to the Perseus context which passes the config down to the Perseus keypad.

packages/perseus/src/mixins/provide-keypad.tsx
```diff
@@ -11,7 +11,7 @@
  * extend a `ProvideKeypad` component instead of using this mixin.
  */
 
-import {MobileKeypad} from "@khanacademy/math-input";
+import {KeypadContext, MobileKeypad} from "@khanacademy/math-input";
```
handeyeco, line 14: No action needed: just for reference, both [ProvideKeypad](https://khanacademy.atlassian.net/browse/LC-386) and [ItemRenderer](https://khanacademy.atlassian.net/browse/LC-408) are dead code now. It's fine if it's out of scope for this PR, but I also wouldn't spend much more time updating them.

handeyeco (approved):
Thank you!

PR #827, opened 2023-11-29
Revert expression-editor changes to answer form key handling

jeremywiebe:
## Summary:

In #711 we added some protective parsing code for AnswerForm `key`s. However, we discovered ([Slack](https://khanacademy.slack.com/archives/C3SLM30AW/p1701199431143119)) that there are instances of data where the key is entirely missing. 

Reverting the `throw` for now to unblock our content authors. I'll continue working on this in a separate PR/deploy to fix this more correctly as Ned was onto something!

Issue: LC-1502

## Test plan:

Run the tests. They should pass.

### DIFF
--- a/packages/perseus-editor/src/widgets/expression-editor.tsx
+++ b/packages/perseus-editor/src/widgets/expression-editor.tsx
@@ -37,11 +37,12 @@ type DefaultProps = {
 };
 
 const parseAnswerKey = ({key}: AnswerForm): number => {
-    const parsedKey = Number.parseInt(key ?? "");
-    if (Number.isNaN(parsedKey)) {
-        throw new Error(`Invalid answer key: ${key}`);
-    }
-    return parsedKey;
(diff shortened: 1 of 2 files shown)


packages/perseus-editor/src/widgets/expression-editor.tsx
```diff
@@ -38,11 +37,12 @@ type DefaultProps = {
 };
 
 const parseAnswerKey = ({key}: AnswerForm): number => {
-    const parsedKey = Number.parseInt(key ?? "");
```
nedredmond, line 41: I have no memory of this code and it doesn't look like my style. I rarely try/catch unless I am trying to _prevent_ errors from being thrown.

I'm not going to dig into git blame, but this code means nothing to me and I don't think we should be throwing here.

nedredmond (approved):
I was onto something? We don't need to make things more fragile. 

If this was really was me, I'm sorry!

PR #833, opened 2023-12-04
Remove ItemRenderer and ProvideKeypad

handeyeco:
## Summary:
Removing dead code

Issue: LC-1317


jeremywiebe (approved):
This is all exciting. One thought, I think we might want to do a Perseus release just before you land this so we can flush any small-ish changes before you land this (which is a major version bump)

PR #835, opened 2023-12-04
Minor editor bugfix for periodic table: uncheck key with table

nedredmond:
When table option unchecked, ensure key is unchecked.
This would only have made the setting "sticky" for the exercise and would have had no effect on the published, rendered result.

Issue: LC-1534

Testing strategy:
Unit tests added. In ZND or after deploy, try:
- Check table option
- Check key option
- Uncheck table option
- Verify that no "unsaved change" is registered
- Verify that checking table option again shows unchecked key

### DIFF
--- a/packages/perseus-editor/src/__tests__/items-extras-editors.test.tsx
+++ b/packages/perseus-editor/src/__tests__/items-extras-editors.test.tsx
@@ -1,0 +1,59 @@
+import "@testing-library/jest-dom";
+import {render, screen} from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import React from "react";
+
+import ItemExtrasEditor from "../item-extras-editor";
+
+describe("ItemExtrasEditor", () => {
+    it("should render correctly with default props", () => {
+        // Arrange
+        const onChangeMock = jest.fn();
+
+        // Act
+        render(<ItemExtrasEditor onChange={onChangeMock} />);
+
+        // Assert
+        expect(
+            screen.getByLabelText("Show periodic table:"),
+        ).toBeInTheDocument();
+        expect(screen.getByLabelText("Show periodic table:")).not.toBeChecked();
+    });
+
+    it("should call onChange with updated calculator value", () => {
+        // Arrange
+        const onChangeMock = jest.fn();
+        render(<ItemExtrasEditor calculator={false} onChange={onChangeMock} />);
+        const checkbox = screen.getByLabelText("Show calculator:");
+
+        // Act
+        userEvent.click(checkbox);
+
+        // Assert
+        expect(onChangeMock).toHaveBeenCalledWith({calculator: true});
+    });
+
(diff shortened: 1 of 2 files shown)


packages/perseus-editor/src/__tests__/items-extras-editors.test.tsx
```diff
@@ -0,0 +1,59 @@
+import "@testing-library/jest-dom";
+import {render, screen} from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import React from "react";
+
+import ItemExtrasEditor from "../item-extras-editor";
+
+describe("ItemExtrasEditor", () => {
+    it("should render correctly with default props", () => {
```
jeremywiebe, line 9: **thought**: I wonder if we should assert all of the "extras" here given the test name.  Otherwise I'd be inclined to name this `'should render period table option as unchecked with default props'`.

packages/perseus-editor/src/item-extras-editor.tsx
```diff
@@ -51,6 +51,13 @@ class ItemExtrasEditor extends React.Component<Props> {
                                     this.props.onChange({
                                         calculator: e.target.checked,
                                     });
+                                    if (!e.target.checked) {
+                                        // When we remove the periodic table,
+                                        // we should also remove sub-options.
+                                        this.props.onChange({
+                                            periodicTableWithKey: false,
+                                        });
+                                    }
```
jeremywiebe, line 60: **question**: Why are we doing this when toggling the calculator?

packages/perseus-editor/src/item-extras-editor.tsx
```diff
@@ -72,6 +79,13 @@ class ItemExtrasEditor extends React.Component<Props> {
                                     this.props.onChange({
                                         periodicTable: e.target.checked,
                                     });
+                                    if (!e.target.checked) {
+                                        // When we remove the periodic table,
+                                        // we should also remove sub-options.
+                                        this.props.onChange({
+                                            periodicTableWithKey: false,
+                                        });
+                                    }
```
jeremywiebe, line 88: **thought**: We could put these within the same `onChange` callback. That'd save one extra re-render. But maybe I shouldn't be thinking at that level and just let it re-render. :) You're call.

jeremywiebe (approved):
Looks good. One issue to verify before landing is the note about changing the `periodTableWithKey` option when toggling the calculator.

packages/perseus-editor/src/item-extras-editor.tsx
nedredmond: Oops-- did I add this twice??

packages/perseus-editor/src/item-extras-editor.tsx
nedredmond: Yeah, I can do that. I am just not used to class component patterns.

PR #837, opened 2023-12-04
Add tests for ExpressionEditor

handeyeco:
## Summary:
While working on [this PR](https://github.com/Khan/perseus/pull/838) I started to get nervous so I decided to add tests.

Ignore the branch name, there is no ticket for this.

PR #838, opened 2023-12-04
Remove TexButtons

handeyeco:
## Summary:
I don't think we need TexButtons anymore because:

1. We don't use them in webapp anymore so why show them to content creators
2. They're not accurate to what we do show
3. The input they use to submit answers shows the keypad

Screenshot 2023-12-04 at 2 59 31 PM
Screenshot 2023-12-04 at 2 59 49 PM

Ignore the branch name, there is no ticket for this.

## Test plan:
https://github.com/Khan/perseus/pull/837


handeyeco:
Codecov is dumb. This PR is just separate from the testing PR: https://github.com/Khan/perseus/pull/837

SonicScrewdriver (approved):
Looks great, thank you for cleaning the old tex buttons up!

PR #840, opened 2023-12-05
Remove Redux

handeyeco:
## Summary:
We don't need it anymore


jeremywiebe (approved):
I think there's a reference to `redux` in the root `package.json`.  I'm not 100% sure why, but I suspect that can go away now that math-input doesn't use it.

handeyeco:
@jeremywiebe I believe I removed every use of the word "redux" already

PR #842, opened 2023-12-05
Remove possible unused types

handeyeco:
## Summary:
Do we need these? I don't think MathInput uses Hubble, JSDiff, or Raphael?

There are three left, do we need those?
- Aphrodite (shouldn't Aphrodite provide these?)
- Assets (Phospher)
- Utility (TS helpers)

### DIFF
--- a/types/README.md
+++ b/types/README.md
@@ -1,0 +1,7 @@
+Warning:
+
+These type declarations are symlinked across each package.
+
+So even if one package doesn't use it, we shouldn't remove any of these type declarations unless **none** of the packages are using it.
+
+See: https://github.com/Khan/perseus/pull/842
--- a/types/hubble.d.ts
+++ b/types/hubble.d.ts
@@ -1,0 +1,1 @@
+declare module "hubble";
--- a/types/hubble.d.ts
+++ b/types/hubble.d.ts
@@ -1,1 +1,0 @@
-declare module "hubble";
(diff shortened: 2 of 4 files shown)


jeremywiebe (changes requested):
We still need this AFAIK.  `hubble` is used in `multi-renderer.tsx`. `jsdiff` is used in the perseus-editor package. and `raphael` is used by Graphie.

handeyeco:
@jeremywiebe Every package has a copy of these, this is just removing these from MathInput which, AFAIK, doesn't use those. I can keep them, but I feel like we probably don't need 11 `hubble.d.ts` files.

jeremywiebe:
> @jeremywiebe Every package has a copy of these, this is just removing these from MathInput which, AFAIK, doesn't use those. I can keep them, but I feel like we probably don't need 11 `hubble.d.ts` files.

Your changes are deleting the files in the root of the repo's `/type` folder.  This `type` folder is symlinked into each package as that is the only way for these Typescript type definitions to work properly with all the infra we have. :( So until no packages use them, these type defs have to stay, I think.

handeyeco:
@jeremywiebe Ah, thank you for clarifying. I'll update the PR to add a README so I don't try this again in two months.

jeremywiebe (approved):
Thank-you for adding this README!  I shoulda dun it when I set up the symlinks originally. :(

PR #846, opened 2023-12-05
Clean up common styles

handeyeco:
## Summary:
Breaks up `common-styles.ts` so much that it didn't really need to exist anymore.


packages/math-input/src/components/input/cursor-styles.ts
jeremywiebe, line 1: Can this just go directly into the `cursor-handle.tsx`? I can only see this being imported in that file.

jeremywiebe (approved):

packages/math-input/src/components/input/cursor-styles.ts
handeyeco: They're both used by `math-input.tsx` too

jeremywiebe: As I typed that I figured I was going to be wrong I just couldn't see that extra usage. 😭

PR #848, opened 2023-12-05
Add Analytics Events for Label Image Widget

nixterrimus:
Adds analytics events these events will be consumed by webapp and turned int CEDAR events.

Issue: https://khanacademy.atlassian.net/browse/LC-1500

## Test plan:
- Run the tests
- **They should pass**

### DIFF
--- a/packages/perseus/src/widgets/label-image.tsx
+++ b/packages/perseus/src/widgets/label-image.tsx
@@ -25,13 +25,13 @@ import {AnswerPill} from "./label-image/answer-pill";
 import {HideAnswersToggle} from "./label-image/hide-answers-toggle";
 import Marker from "./label-image/marker";
 
-import type {DependencyProps} from "../dependencies";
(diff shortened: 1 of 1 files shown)


nedredmond (approved):
I don't know a lot about this, so I'd like @jeremywiebe to look it over. AFAIK, g2g!

packages/perseus/src/dependencies.ts
```diff
@@ -22,6 +22,10 @@ export const getDependencies = (): PerseusDependencies => {
     );
 };
 
+export type DependencyProps = Partial<
```
jeremywiebe, line 25: I love that we're adding this.  Could I ask that you circle back on a separate PR and make sure other sites that used this `Partial<React.ContextType...` stuff be replaced with this new type?

jeremywiebe (commented):
@nixterrimus  We paired through most of this.  Just one follow-up request for a future PR.

PR #850, opened 2023-12-05
Make Perseus error reports easier to triage

jeremywiebe:
## Summary:

I was looking at our Sentry dashboard and noticed that _all_ Perseus rendering errors are grouped together under one event. This is largely due to the fact that the error boundary uses the same string for all errors it catches.

So this PR changes that error logging so that each different type of error caught by the boundary are logged as a different error (Sentry event). This will help us to triage/review errors in Sentry and not group all Perseus errors together in host applications.

Issue: "none"

## Test plan:

### DIFF
--- a/packages/perseus/src/error-boundary.tsx
+++ b/packages/perseus/src/error-boundary.tsx
@@ -35,7 +35,10 @@ class ErrorBoundary extends React.Component<Props, State> {
             {
                 cause: error,
                 loggedMetadata: {
-                    componentStack: info.componentStack,
+                    componentStack:
+                        "componentStack" in info
+                            ? info.componentStack
+                            : "componentStack not provided",
                     ...this.props.metadata,
                 },
             },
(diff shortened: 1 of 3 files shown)


nedredmond (approved):
Thanks Jeremy!

packages/perseus/src/error-boundary.tsx
```diff
@@ -22,13 +22,24 @@ class ErrorBoundary extends React.Component<Props, State> {
...
+            // reporting tool (currently Sentry). However, if we don't
+            // differentiate between the different errors that are thrown, they
+            // all end up being grouped as a single Sentry event, which is very
+            // unhelpful.
+            "Unhandled Perseus error: " + error.message,
+            Errors.Internal,
+            {
+                cause: error,
+                loggedMetadata: {
+                    componentStack: info.componentStack,
```
benchristel, line 38: If `info` is null or undefined, this will throw. Maybe it's better to do `info?.componentStack`?

jeremywiebe: The React [docs](https://react.dev/reference/react/Component#catching-rendering-errors-with-an-error-boundary) seem to imply that it will always be provided.  But I'm happy to make this uber fault-tolerant. The last thing we want to do is cause some sort of recursive error scenario. :)

jeremywiebe: Ok. I've made it more resilient if `info` is missing `componentStack`.  And added tests... because. :)

PR #852, opened 2023-12-05
Remove ItemRenderer again, but make ServerItemRenderer work with linting

handeyeco:
## Summary:
This reverts  #847 which reverted  #833. It also makes ServerItemRenderer work with `linterContext` so it can be used in the content editor.

Issue: LC-1546

## Test plan:
Once this lands to Webapp:
- Go to the content editor
- Go to some page where you can edit content
- Make a linting error (like writing `# hello world\n\n`)
- Note that ServerItemRenderer shows a linting error


packages/perseus/src/__tests__/server-item-renderer.test.tsx
```diff
@@ -529,4 +530,24 @@ describe("server item renderer", () => {
...
+                linterContext: {
+                    contentType: "exercise",
+                    highlightLint: true,
+                    paths: [],
+                    stack: [],
+                },
+            });
+
+            // Assert
+            expect(screen.getByRole("link")).toHaveAttribute(
```
handeyeco, line 547: I don't think there's a cleaner way to assert this. The link doesn't have text or alt text.

packages/perseus/src/server-item-renderer.tsx
```diff
@@ -462,7 +474,7 @@ const styles = StyleSheet.create({
 
 const ref = React.forwardRef<
     ServerItemRenderer,
-    Omit<React.ComponentProps<typeof ServerItemRenderer>, "onRendered">
+    Omit<PropsFor<typeof ServerItemRenderer>, "onRendered">
```
handeyeco, line 477: Thanks Jeremy!

packages/perseus/src/__tests__/server-item-renderer.test.tsx
jeremywiebe: Would this work? 

```js
expect(
    await screen.findByText(/Don't use level-1 headings:/),
).toBeVisible();
```

jeremywiebe (approved):
Nice!

packages/perseus/src/__tests__/server-item-renderer.test.tsx
handeyeco: Oh, for some reason the DOM around the text was greyed out and I didn't think to dig into it. Yep, there is text I can use. Changed 👍

PR #853, opened 2023-12-05
Adding a click event so that ChromeOS users using 3rd party browsers can interact with our math input.

SonicScrewdriver:
## Summary:
<Add your summary here>

Issue: LC-1522

## Test plan:

### DIFF
--- a/packages/math-input/src/components/input/math-input.tsx
+++ b/packages/math-input/src/components/input/math-input.tsx
@@ -626,6 +626,31 @@ class MathInput extends React.Component<Props, State> {
         }
     };
 
+    handleClick: (arg1: React.MouseEvent<HTMLDivElement>) => void = (e) => {
+        e.stopPropagation();
+
+        // Hide the cursor handle on touch start, if the handle itself isn't
+        // handling the touch event.
+        this._hideCursorHandle();
+
+        // Cache the container bounds, so as to avoid re-computing. If we don't
+        // have any content, then it's not necessary, since the cursor can't be
+        // moved anyway.
+        if (this.mathField.getContent() !== "") {
+            this._containerBounds = this._container.getBoundingClientRect();
+
+            // Make the cursor visible and set the handle-less cursor's
+            // location.
+            const {clientX, clientY} = e;
+            this._insertCursorAtClosestNode(clientX, clientY);
+        }
+
+        // Trigger a focus event, if we're not already focused.
+        if (!this.state.focused) {
+            this.focus();
+        }
+    };
+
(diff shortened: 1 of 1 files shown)


packages/math-input/src/components/input/math-input.tsx
```diff
@@ -626,6 +667,33 @@ class MathInput extends React.Component<Props, State> {
         }
     };
 
+    // We want to allow the user to be able to focus the input via click
+    // when using ChromeOS third-party browsers that use mobile user agents,
+    // but don't actually simulate touch events.
+    handleClick: (arg1: React.MouseEvent<HTMLDivElement>) => void = (e) => {
```
handeyeco, line 673: Is it possible to combine the typing with the declaration? Something like this?

```suggestion
    handleClick: (e: React.MouseEvent<HTMLDivElement>): void {
```

packages/math-input/src/components/input/math-input.tsx
```diff
@@ -626,6 +667,33 @@ class MathInput extends React.Component<Props, State> {
...
+        e.stopPropagation();
+
+        // Hide the cursor handle on touch start, if the handle itself isn't
+        // handling the touch event.
+        this._hideCursorHandle();
+
+        // Cache the container bounds, so as to avoid re-computing. If we don't
+        // have any content, then it's not necessary, since the cursor can't be
+        // moved anyway.
```
handeyeco, line 682: Do we still want to cache things? I know we talked about maybe not needing that.

handeyeco (approved):
Looks good! Feedback is all optional.

packages/math-input/src/components/input/math-input.tsx
```diff
@@ -196,9 +198,47 @@ class MathInput extends React.Component<Props, State> {
...
+                        // that the keypad may be anchored above the 'Check answer' bottom bar,
+                        // in which case we don't want to dismiss the keypad on check.
+                        if (
+                            (bounds.left <= x &&
+                                bounds.right >= x &&
+                                bounds.top <= y &&
+                                bounds.bottom >= y) ||
+                            bounds.bottom < y
+                        ) {
+                            touchDidStartInOrBelowKeypad = true;
```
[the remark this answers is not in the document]
SonicScrewdriver: Absolutely, I'll get that set up! I was a little worried it would make the logic fairly confusing to read, but I think a better solution is to make a "isWithinKeypadBounds" function that can return a boolean in order to clean up the if statement. That should make things far more readable and a little more dry :)

PR #859, opened 2023-12-06
Ensure that we're always getting the correct keypad bounds

SonicScrewdriver:
## Summary:
While investigating the blur issue from LC-1523, I discovered that we have some historical logic for caching our keypad bounds. None of that is necessary any longer. 

Fixing/removing this logic solves LC-1523 by ensuring that we are always getting the _current_ keypadBounds, which allows us to be able to click under the input again to blur it after the keypad has been closed/animated out of view.

It seems we've added a resize observer directly to the keypad, so we no longer need a bunch of event listeners in our MathInput either!

Issue: LC-1523

## Test plan:
- Manual testing of multiple device orientations and resizes

### DIFF
--- a/packages/math-input/src/components/input/math-input.tsx
+++ b/packages/math-input/src/components/input/math-input.tsx
@@ -361,26 +361,7 @@ class MathInput extends React.Component<Props, State> {
 
         this.mathField.focus();
         this.props?.onFocus();
-        this.setState({focused: true}, () => {
-            // NOTE(charlie): We use `setTimeout` to allow for a layout pass to
-            // occur. Otherwise, the keypad is measured incorrectly. Ideally,
-            // we'd use requestAnimationFrame here, but it's unsupported on
-            // Android Browser 4.3.
-            setTimeout(() => {
-                console.log("scroll");
-                if (this._isMounted) {
-                    // TODO(benkomalo): the keypad is animating at this point,
-                    // so we can't call _cacheKeypadBounds(), even though
-                    // it'd be nice to do so. It should probably be the case
-                    // that the higher level controller tells us when the
-                    // keypad is settled (then scrollIntoView wouldn't have
-                    // to make assumptions about that either).
-                    const maybeKeypadNode =
-                        this.props.keypadElement?.getDOMNode();
-                    scrollIntoView(this._container, maybeKeypadNode);
-                }
-            });
-        });
+        this.setState({focused: true});
     };
 
     /**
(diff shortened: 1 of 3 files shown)


packages/math-input/src/components/input/math-input.tsx
```diff
@@ -197,25 +196,6 @@ class MathInput extends React.Component<Props, State> {
...
-        // having the keypad notify of changes to us.
-        window.addEventListener("resize", this._clearKeypadBoundsCache);
-        window.addEventListener(
-            "orientationchange",
-            this._clearKeypadBoundsCache,
-        );
-    }
-
-    // eslint-disable-next-line react/no-unsafe
-    UNSAFE_componentWillReceiveProps(props: Props) {
```
handeyeco, line 215: Glad to see the `UNSAFE` go!

packages/math-input/src/components/input/math-input.tsx
```diff
@@ -263,13 +228,10 @@ class MathInput extends React.Component<Props, State> {
         this._root.style.fontSize = `${fontSizePt}pt`;
     };
 
-    /** Gets and cache they bounds of the keypadElement */
+    /** Returns the current bounds of the keypadElement */
     _getKeypadBounds: () => any = () => {
```
handeyeco, line 232: I wonder if this could be a good opportunity to type this method. Not sure if this right, but maybe:
```suggestion
    _getKeypadBounds(): DOMRect {
```

packages/math-input/src/components/input/math-input.tsx
```diff
@@ -263,13 +228,10 @@ class MathInput extends React.Component<Props, State> {
...
-        if (!this._keypadBounds) {
-            const node = this.props.keypadElement?.getDOMNode();
-            this._cacheKeypadBounds(node);
-        }
-        return this._keypadBounds;
+        const keypadNode = this.props.keypadElement?.getDOMNode() as Element;
```
handeyeco, line 233: I know this would be harder to type, but I think this would be safer.

```suggestion
        const keypadNode = this.props.keypadElement?.getDOMNode();
        
        if (!keypadNode) {
            return null
        }
```

handeyeco (approved):
Nice! Just some minor suggestions.

packages/math-input/src/components/input/math-input.tsx
SonicScrewdriver: This ended up working nicely with my new "isWithinKeypadBounds" function from the other PR

PR #861, opened 2023-12-07
Fix chevron getting hidden in label image widget

nixterrimus:
## Summary:
Fixes chevron getting hidden when a learner focuses with the keyboard on an open marker dot.

https://github.com/Khan/perseus/assets/18454/c45109f0-f38e-400d-8e84-c4f29b673dfa

Issue: https://khanacademy.atlassian.net/browse/LC-1544

## Test plan:
- In label image widget open a popover
- Use the keyboard to navigate to the marker dot
- **The Chevron should not be hidden**

### DIFF
--- a/.changeset/lemon-rabbits-vanish.md
+++ b/.changeset/lemon-rabbits-vanish.md
@@ -1,6 +1,5 @@
 ---
 "@khanacademy/perseus": patch
-"@khanacademy/perseus-core": patch
 ---
 
 Bug fix: dont hide chevron in label image on focus


.changeset/lemon-rabbits-vanish.md
```diff
@@ -0,0 +1,6 @@
+---
+"@khanacademy/perseus": patch
+"@khanacademy/perseus-core": patch
```
nedredmond, line 3: Should you be bumping core here?

PR #867, opened 2023-12-12
Add financial calculator checkboxes to Item Extras Editor

nedredmond:
## Summary:

- Add financial calculator checkboxes to Item Extras
- Add an indentation to dependent checkboxes
- Slight refactor to reduce redundant code
- Added item extras story

Issue: LC-1537

Testing strategy:
Check in new story that nothing has changed except new checkboxes added.
Clicking Financial Calculators checkbox should expand three checked options. Unchecking all options unchecks parent checkbox; unchecking parent unchecks options.

### DIFF
--- a/packages/perseus-editor/src/item-extras-editor.tsx
+++ b/packages/perseus-editor/src/item-extras-editor.tsx
@@ -1,29 +1,16 @@
-import {components} from "@khanacademy/perseus";
+import {components, ItemExtras} from "@khanacademy/perseus";
 import * as React from "react";
+
+import type {PerseusAnswerArea} from "@khanacademy/perseus";
 
 const {InfoTip} = components;
 
-const ItemExtras = [
-    "calculator",
-    "chi2Table",
-    "financialCalculator",
-    "financialCalculatorMonthlyPayment",
-    "financialCalculatorTotalAmount",
-    "financialCalculatorTimeToPayOff",
-    "periodicTable",
-    "periodicTableWithKey",
-    "tTable",
-    "zTable",
-] as const;
-
-type ItemExtrasProps = Record<typeof ItemExtras[number], boolean>;
-
-type Props = ItemExtrasProps & {
-    onChange: (props: Partial<ItemExtrasProps>) => void;
+type Props = PerseusAnswerArea & {
+    onChange: (props: Partial<PerseusAnswerArea>) => void;
 };
 
 class ItemExtrasEditor extends React.Component<Props> {
-    static defaultProps: ItemExtrasProps = {
+    static defaultProps: PerseusAnswerArea = {
         calculator: false,
         chi2Table: false,
         financialCalculator: false,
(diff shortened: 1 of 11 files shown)


packages/perseus-editor/src/item-extras-editor.tsx
```diff
@@ -20,159 +26,208 @@ class ItemExtrasEditor extends React.Component<Props> {
...
+            !this.props.financialCalculatorTotalAmount &&
+            !this.props.financialCalculatorTimeToPayOff
+        ) {
+            this.props.onChange({
+                financialCalculator: false,
+            });
+        }
+    }
+
     serialize: () => any = () => {
```
jeremywiebe, line 52: **request**: Can you type this `serialize` method result as the `PerseusAnswerArea` type from `@khanacademy/perseus`, please? That will mean you also need to add these new options to that type, but that's a good thing as that type should be the source of truth for what options are available in the answer area.

packages/perseus-editor/src/item-extras-editor.tsx
```diff
@@ -20,159 +26,208 @@ class ItemExtrasEditor extends React.Component<Props> {
...
+                        }}
+                    />
 
-                    <div>
-                        <label>
-                            Show z table (statistics):{" "}
-                            <input
-                                type="checkbox"
-                                checked={this.props.zTable}
+                    {this.props.financialCalculator && (
```
jeremywiebe, line 95: **question**: Are these three sub-types mutually exclusive?  If so, would it make sense for them to be a set of radio buttons instead?

jeremywiebe (approved):
This looks good.  One request before landing is to address the typing comment for the `serialize` method.

packages/perseus-editor/src/item-extras-editor.tsx
nedredmond: They are not mutually exclusive!

packages/perseus-editor/src/item-extras-editor.tsx
nedredmond: Can do! I wanted to type this but wasn't sure where it was used or how, didn't investigate. Always happy to improve types.

nedredmond: @jeremywiebe it doesn't seem advisable to import types from a different package. Each package should export its own types, no? This is reinforcing some very difficult coupling between the packages.

It doesn't look like `PerseusAnswerArea` is used anywhere.

nedredmond: OH I see now. I'll figure something out here.

packages/perseus-editor/src/item-extras-editor.tsx
jeremywiebe: Ah, then this makes sense.  Thanks.

packages/perseus-editor/src/item-extras-editor.tsx
jeremywiebe: With regards to exporting/importing types. In this case, the `perseus-editor` package is used to build objects that are consumed by Perseus itself.  So in that case I do think it makes sense for the editors to use types from Perseus to ensure we are building the Perseus data objects correctly. 

In the future, I'd love to see these editors run the data through a validation that's exported from Perseus to guarantee the objects are valid before saving them (ie. not _just_ a type check but a runtime validation).

PR #869, opened 2023-12-13
Remove unused field from PerseusAnswerArea

nedredmond:
One of the fields added in #867 is redundant and does not need to be exposed outside of Perseus. Switched it to component state.


Myranae (approved):
Looks good to me. Saw how you removed the `financialCalculator` prop and switched it to state inside the component.

PR #870, opened 2023-12-13
Add standin widget for deprecated widgets

nixterrimus:
## Summary:

This PR is groundwork for replacing the transformer widget.

- Adds a widget that is always correct
- Adds infrastructure and testing for replacing a widget

Screenshot 2023-12-26 at 11 30 18 AM

When it's time to replace the transformer widget we need to delete the code for that widget and use the commented out replacement function.

Issue: [LC-1582](https://khanacademy.atlassian.net/browse/LC-1582)

## Test plan:
- Visit storybook
- **You should be able to see the auto correct widget**

- Run the tests
- **Everything should pass**

[LC-1582]: https://khanacademy.atlassian.net/browse/LC-1582?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/perseus/src/widgets.ts
+++ b/packages/perseus/src/widgets.ts
@@ -62,9 +62,9 @@ export const replaceWidget = (name: string, replacementName: string) => {
 
     registerWidget(name, substituteWidget);
 
-    // The logger may not have been loaded at this point so we use the console
-    // eslint-disable-next-line no-console
-    console.log(`INFO: Replacing widget ${name} with ${replacementName}`);
+    if (Log) {
+        Log.log(`INFO: Replacing widget ${name} with ${replacementName}`);
+    }
 };
 
 export const registerEditors = (editorsToRegister: ReadonlyArray<Editor>) => {


packages/perseus/src/perseus-types.ts
```diff
@@ -187,6 +187,8 @@ export type SimpleMarkdownTesterWidget = Widget<'simple-markdown-tester', Perseu
 export type UnitInputWidget = Widget<'unit-input', PerseusUnitInputWidgetOptions>;
 // prettier-ignore
 export type VideoWidget = Widget<'video', PerseusVideoWidgetOptions>;
+//prettier-ignore
+export type AutoCorrectWidget = Widget<'auto-correct', any>;
```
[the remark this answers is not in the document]
handeyeco: I noticed in a comment later on that you mentioned some of these `any`s are because `auto-correct` can fill in for any other widget. If that's the case here, maybe it could use `PerseusWidgetOptions` which ORs together all of our existing widget options.

packages/perseus/src/widgets/auto-correct/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+                    layout="full-width"
+                />
+            </div>
+        );
+    }
+}
+
+export default {
+    name: "auto-correct",
+    displayName: "Auto Program",
```
handeyeco, line 55: Should this be "Auto Correct"?

handeyeco (commented):
Is there a plan to hook this up to CEDAR? I could see it being useful to see where learners are running into the `auto-correct` widget so that we can communicate with the content team where we need to update content.

packages/perseus/src/widgets/deprecated-standin/index.tsx
```diff
@@ -0,0 +1,57 @@
+import Banner from "@khanacademy/wonder-blocks-banner";
+import * as i18n from "@khanacademy/wonder-blocks-i18n";
+import React from "react";
+
+import type {PerseusScore, WidgetExports} from "../../types";
+
+// The props are type `any` on purpose so that this can receive props
+// from any deprecated widget
+type AutoCorrectProps = any;
+type Rubric = any;
```
handeyeco, line 10: In my heart I still want these to have types, but I also realize we're not actually using anything from them. Doesn't seem worth the effort to type these, but I wonder if it's worth circling back in a separate PR when @jeremywiebe is back or asking in #typescript if there's some TS magic that could work here.

handeyeco (approved):
Looks good 👍 Excited to see this in prod. I do think it'd be worth at least filing a ticket for adding CEDAR. I think it'd be beneficial to see where these are getting seen by learners.

packages/perseus-editor/src/util/register-all-widgets-and-editors-for-testing.test.ts
```diff
@@ -6,7 +6,9 @@ describe("Registering all widgets and editors", () => {
     it("should have an editor for every widget", () => {
         registerAllWidgetsAndEditorsForTesting();
 
-        const allWidgetsTypes = Widgets.getAllWidgetTypes();
+        const allWidgetsTypes = Widgets.getAllWidgetTypes()
+            // Auto correct is a special kind of widget, it DOES NOT need an editor
+            .filter((a) => a !== "deprecated-standin");
```
jeremywiebe, line 11: I'm reviewing after the fact, but it would be nice if this widget used a consistent name everywhere.  The name is `deprecated-standin` but we refer to it as "auto correct".  Also, auto-correct is somewhat of a taken term to mean the hidden feature in our consuming apps where a developer can "speed run" an exercise by auto-answering the question. 

FWIW: I like `deprecated standin` as the name. 

I'll leave it with you to decide if it's worth changing.

packages/perseus/src/widgets/deprecated-standin/__tests__/deprecated-standin.test.ts
```diff
@@ -0,0 +1,93 @@
...
+                static: false,
+            },
+            alignment: "default",
+        },
+    },
+} as const;
+
+describe("Deprecated Standin widget", () => {
+    beforeEach(() => {
+        jest.useRealTimers();
```
jeremywiebe, line 30: Just double-checking, is this necessary for these tests to pass?  I don't see any `async` tests so I'd be surprised if real timers are needed.

jeremywiebe (commented):
This all looks good.  One question I have is what does the editing experience look like? I know content authors can't use the `transformer` in new content, but would we want to update the editing side to make it clear that learners won't be able to interact with the widget in production and so it's useful to make changes to a `transformer` config in content.

PR #871, opened 2023-12-13
Ensure that we can reopen the keypad by clicking/tapping an already focused math-input.

SonicScrewdriver:
## Summary:
When users close the keypad using the "X" button, we retain focus on the input in order to vastly improve the user experience for screen readers. Unfortunately, this caused an issue where users would be unable to re-trigger the keypad until blurring the input. 

This PR adds some logic that allows us to open the keypad if our input is currently focused, and the keypadActive is currently set to false. 

Issue: LC-1524

## Test plan:
- Manually tested by bumping perseus in webapp to this PR, and then reviewing both mobile web using Chrome Dev Tools and iOS Simulator Safari.

## Videos:
I can't seem to upload them without them resulting in downloadable files, so instead I'll add them to the Jira ticket.

### DIFF
--- a/.changeset/blue-beds-fly.md
+++ b/.changeset/blue-beds-fly.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/math-input": minor
+---
+
+Ensured that tapping an already focused math input will reopen the keypad if it is closed.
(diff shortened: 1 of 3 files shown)


handeyeco (approved):
This looks good to me. Seems like something that should be relatively easy to add unit tests for; see `shows the keypad after input interaction` as precedent.

PR #873, opened 2023-12-14
Prevent paragraph class from being added to answer pills

nedredmond:
## Summary:
Rather than block paragraph styles from every avenue we come across, instead prevent them from being associated with the answer pill in the first place.

Adds a `inline` prop to `Renderer` that blocks the `paragraph` class from being added to answer pills.

### AFTER:
Screenshot 2023-12-14 at 12 26 57 PM
Screenshot 2023-12-14 at 12 59 09 PM

### BEFORE:
image (8)
image (9)

Issue: LC-1569

### Testing strategy:
Use NPM snapshot or link to check out editor and mobile view of label image widget exercises on webapp.

### DIFF
--- a/packages/perseus/src/question-paragraph.tsx
+++ b/packages/perseus/src/question-paragraph.tsx
@@ -17,7 +17,11 @@ class QuestionParagraph extends React.Component<Props> {
         // to attach some metadata to top-level QuestionParagraphs:
         return (
             <div
-                className={this.props.inline ? undefined : className}
+                className={
+                    this.props.inline
+                        ? this.props.className ?? undefined
+                        : className
+                }
                 data-perseus-component-index={this.props.translationIndex}
                 data-perseus-paragraph-index={this.props.paragraphIndex}
             >
--- a/packages/perseus/src/question-paragraph.tsx
+++ b/packages/perseus/src/question-paragraph.tsx
@@ -5,6 +5,7 @@ type Props = {
     translationIndex: number;
     paragraphIndex?: number;
     children?: React.ReactNode;
+    plainText?: boolean;
 };
 
 class QuestionParagraph extends React.Component<Props> {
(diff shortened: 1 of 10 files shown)


packages/perseus/src/widgets/passage-ref-target.tsx
```diff
@@ -50,8 +50,10 @@ class PassageRefTarget extends React.Component<any> {
         return (
             <Renderer
                 content={this.props.content}
-                // @ts-expect-error - TS2322 - Type '{ content: any; inline: boolean; apiOptions: any; linterContext: any; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Renderer> & Pick<Readonly<Props> & Readonly<{ children?: ReactNode; }>, "children" | ... 3 more ... | "legacyPerseusLint"> & InexactPartial<...> & InexactPartial<...>'.
-                inline={true}
+                // This was here, but the prop was not defined until now! (#873)
+                // Perhaps it once existed and was removed a long time ago.
+                // Commenting out to prevent unanticipated side effects
+                // inline={true}
```
nedredmond, line 56: This is weird! We had `inline` here already, but it didn't do anything.

packages/perseus/src/question-paragraph.tsx
```diff
@@ -16,7 +17,7 @@ class QuestionParagraph extends React.Component<Props> {
         // to attach some metadata to top-level QuestionParagraphs:
         return (
             <div
-                className={className}
+                className={this.props.inline ? undefined : className}
```
jeremywiebe, line 20: **issue**: We will lose the `className` provided in a prop with this approach.  I _think_ we need to split out that prop from the local `className` so we can prevent applying the `"paragraph"` class but still respect any provided. `props.className`

packages/perseus/src/widgets/passage-ref-target.tsx
jeremywiebe: Was added waaaaaaaay back: https://phabricator.khanacademy.org/D12113

I think it makes sense to comment this out for now, like you did.  Maybe add a reference to this Phabricator diff in the comment just for posterity?

jeremywiebe (approved):
Makes sense. I think this might allow us to simplify other CSS rules that dance around the `paragraph` class when we need to render inline (for future work).

packages/perseus/src/widgets/passage-ref-target.tsx
jeremywiebe: And then removed here: https://phabricator.khanacademy.org/D13454#inline-89200

PR #876, opened 2023-12-15
Fix `code` styling in content editor (and occasionally in production)

jeremywiebe:
## Summary:

@kmerrill27 noticed that in the content editor when previewing in Desktop mode, formatted code does not use a monospace font.

There are two reasons for this:

  1. There is a general `code` and `pre` rule in `khan-exercise.css` that should apply, but when Perseus is used in webapp, the webapp CSS reset is applied after, negating those styles.
  1. There is a desktop style for `code` and `pre` elements, but right now they are nested inside a `.perseus-mobile` rule, meaning they can't apply because we only apply that class in mobile contexts. :facepalm: (See [here](https://github.com/Khan/perseus/blob/9e522234f03bd9ae9fe5d4554827bb41012ee183/packages/perseus/src/styles/articles.less#L585)).

So, for now the simplest, least invasive fix is to hoist the common `code` and `pre` element styles scoped within `.framework-perseus`.  Other, existing styles for mobile and articles will still apply, but this will guarantee that we have a base style that has a higher specificity than any reset styles.

Issue: LC-1482

## Test plan:

  1. Navigate to https://www.khanacademy.org/devadmin/content/exercises/tracing-for-loops-with-range/xfb1f2616e5139982/x44d92fae4709f15d
  1. Tap the "Desktop" button in the right-hand preview
  1. Note that the radio widget options that include code are formatted width font


packages/perseus/src/server-item-renderer.tsx
```diff
@@ -2,12 +2,12 @@
...
- * renderer normally instead of ReactDOM.render()ing them into elements in the
- * DOM.
+ * The main item (aka Exercise) renderer in Perseus. This component renders its
+ * question renderer and hints renderer using standard React practices (an
+ * earlier ItemRenderer component used to use jQuery and ReactDOM.render()ing
+ * to render into elements it didn't own.
  *
- * This allows this component to be used in server-rendering of a perseus
- * exercise.
+ * This component is compatible with server-rendering of a perseus exercise.
```
jeremywiebe, line 10: This was done just to trigger an npm snapshot. For some reason that I couldn't figure out the Github Action to publish a snapshot was being skipped until I modified source code.  🤷‍♂️

nedredmond (approved):
While there will definitely be consequences to this... approved!

PR #877, opened 2023-12-15
## Summary:

jeremywiebe:
A round of Less file cleanup. There were a bunch of unused variables and files. I figured out what was unused by running `yarn build --watch` and then starting to delete variables. Any time there was a varaible that was still in use, there would be an error like this `[!] (plugin commonjs--resolver) Error: variable @gray95 is undefined`.

  * Get rid of variables.shared.less and move the single used variable to variables.less
  * Rename articles.less to styles.less as it has rules for both articles and exercises!
  * Remove completely unused variables.bootstrap.less
  * Remove outdated comment
  * Extract zoom styles to own file out of styles.less
  * Trim out more unused variables from variables.less

Issue: --none--

## Test plan:

`yarn build` is successful (our builds process Less files into CSS and so if there's variable references that no longer exist, it will fail)


packages/perseus/src/styles/styles.less
```diff
@@ -672,73 +671,4 @@
     }
 }
 
-/* Derived from the MIT-licensed zoom.js:
```
jeremywiebe, line 675: This chunk of styles was moved to `zoom.less`

nedredmond (approved):
All looks good to me! Thanks, @jeremywiebe !

PR #879, opened 2023-12-18
JIPT (Just In Place Translation) fix up dependency types

jeremywiebe:
## Summary:

Our JIPT configuration on the dependencies required the full JIPT configuration even when JIPT support was disabled (`useJIPT: false`). I've changed it so the types don't require the extra support functions when JIPT is not enabled. This makes the code clearer where we support this feature and should make dependency setup in host apps clearer as well.

Issue: "none"

## Test plan:

### DIFF
--- a/packages/perseus/src/types.ts
+++ b/packages/perseus/src/types.ts
@@ -466,24 +466,29 @@ type InitialRequestUrlInterface = {
 
 export type VideoKind = "YOUTUBE_ID" | "READABLE_ID";
 
-// An object for dependency injection, to allow different clients
-// to provide different methods for logging, translation, network
-// requests, etc.
-//
-// NOTE: You should avoid adding new dependencies here as this type was added
-// as a quick fix to get around the fact that some of the dependencies Perseus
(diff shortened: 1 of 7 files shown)


packages/perseus/src/types.ts
```diff
@@ -469,15 +476,14 @@ export type VideoKind = "YOUTUBE_ID" | "READABLE_ID";
 // could be used. Aim to shrink the footprint of PerseusDependencies and try to
 // use alternative methods where possible.
 export type PerseusDependencies = {
-    // JIPT
     JIPT: JIPT;
-    graphieMovablesJiptLabels: JiptLabelStore;
-    svgImageJiptLabels: JiptLabelStore;
-    rendererTranslationComponents: JiptTranslationComponents;
```
jeremywiebe, line 476: These only make sense when `useJIPT: true`, so I've moved these into the `JIPT` type union when `useJIPT: true`

jeremywiebe:
This is pretty stale. Closing ... can come back to it if we really want.

PR #881, opened 2023-12-19
Swap out LabelImage AnswerChoices WB Dropdown

nedredmond:
## Summary:
Swaps out custom dropdown for Wonderblocks dropdown.

This required a litany of changes-- some of the biggest:
 - Move AnswerPill into Marker (also gets rid of warning in console re: calling `findDomNode` in render function)
 - Moving/removing several `onClick` functions so as not to interfere with `opener`'s ClickableState magic-- as a result, analytics functions moved
 - Add logic to WonderBlocks Dropdown to pass to `opener` component when dropdown is open
 - Change how `innerIcon` in marker is rendered (for some reason, refused to be clickable when rendered conditionally)
 - [Also includes changes described in this ADR](https://khanacademy.atlassian.net/l/cp/FHith04E)

And there were some beneficial side-effects:
 - Many styles were consolidated
 - Handling of active/focused markers moved to WB Dropdown
 - Marker & AnswerPill get to share hover/focus states
 - Fixed minor UI issues with Marker being slightly off its mark
 

https://github.com/Khan/perseus/assets/23404711/baa7ebec-326b-468b-b7ce-2af218d1fd85

 Issue: LC-1425
 
 ## Testing strategy:
 Will be thoroughly tested to ensure if still operates the same or better as in prod.


packages/perseus/src/widgets/label-image/marker.tsx
```diff
@@ -102,90 +113,86 @@ export default class Marker extends React.Component<Props> {
...
                 ref={(node) => (this._icon = node)}
+                onClick={() => {
+                    this.props.analytics?.onAnalyticsEvent({
+                        type: "perseus:label-image:marker-interacted-with",
+                        payload: null,
+                    });
+                }}
             >
-                {innerIcon}
+                <Icon
```
nedredmond, line 131: For some unknown reason, this was not clickable when rendered conditionally. @SonicScrewdriver can attest!

Could be a weird Perseus Icon thing. Decided to hold off switching to WB.

packages/perseus/src/widgets/label-image/marker.tsx
```diff
@@ -102,90 +113,86 @@ export default class Marker extends React.Component<Props> {
...
+            selected,
+            showAnswer,
+            answerSide,
+            answerStyles,
+            hovered,
+            focused,
+        } = this.props;
 
         return (
-            <button
```
nedredmond, line 129: Moved the button logic out of this component. The `opener` function on the dropdown was acting funky and it was unclickable until I moved the clickable component.

packages/perseus/src/widgets/label-image/marker.tsx
```diff
@@ -102,90 +113,86 @@ export default class Marker extends React.Component<Props> {
...
-                className={css(
-                    styles.unstyledButton,
-                    styles.marker,
-                    isDisabled && styles.disabled,
+            <>
+                <View style={[styles.marker]}>
+                    {this.renderIcon({hovered, focused})}
+                </View>
+                {!!selected && showAnswer && (
+                    <AnswerPill
```
nedredmond, line 162: Moved this into here to take advantage of the ref (previously, had to do an illegal move and use `findDomNode` in the render function)

packages/perseus/src/widgets/label-image/marker.tsx
```diff
@@ -80,7 +91,7 @@ export default class Marker extends React.Component<Props> {
                     isSelected && styles.markerSelected,
                 ];
             }
-        } else if (focused) {
+        } else if (focused || hovered) {
```
nedredmond, line 94: Hover styles for initial view-- we enforce this everywhere else

packages/perseus/src/widgets/label-image.test.ts
```diff
@@ -319,544 +319,6 @@ describe("LabelImage", function () {
         });
     });
 
-    describe("navigateToMarkerIndex", function () {
```
jeremywiebe, line 322: Phew, were all these tests verifying things that the WB components handle now?

jeremywiebe (approved):
Whew! That's alot of changes.  😅 

But, the experience looks a bunch better and it's nice to remove more custom code in favour of WB components. Thanks for cleaning this up Ned!

packages/perseus/src/widgets/label-image.test.ts
nedredmond: This actually just doesn't work, but I meant to un-delete this since I left the non-working code in there (trying to control unnecessary scope creep where I can)

nedredmond: It doesn't work in prod either, not just in the PR code

packages/perseus/src/widgets/label-image.tsx
```diff
@@ -605,13 +599,14 @@ export class LabelImage extends React.Component<
...
                         // cannot change answer choices once question is answered
                         disabled={disabled}
-                        opener={({opened, text}) => (
+                        opener={({opened}) => (
                             <Clickable
                                 role="button"
                                 key={`marker-${marker.x}.${marker.y}`}
                                 aria-label={
-                                    marker.selected
-                                        ? (text as string)
```
nedredmond: This can't be exposed because it may include TeX and MD

packages/perseus/src/util/sre/tex-to-text.ts
```diff
@@ -0,0 +1,37 @@
...
+
+// Code adapted from https://github.com/Speech-Rule-Engine/sre-latex/blob/main/tex2mml/tex2mml.js
+
+//  Busproofs requires an output jax, which we aren't using
+const packages = AllPackages.filter((name) => name !== "bussproofs");
+
+//  Create the input jax
+const tex = new TeX({packages: packages});
+
+//  Create an HTML document using a LiteDocument and the input jax
```
benchristel, line 20: It might be worth clarifying here that liteAdaptor doesn't touch the browser's actual DOM.

benchristel (commented):
This is very cool! I like that we're able to reuse MathJax's capabilities for math-to-speech.

One thing that's concerning: the bundle size increased by a couple hundred kb, even though this PR removes more lines than it adds. That smells like MathJax is getting bundled somehow, even though it's only listed as a peer dependency. I think that's worth figuring out before we land this, since we definitely don't want to ship two copies of MathJax in webapp.

nedredmond:
@benchristel MathJax was not previously a part of this repository, so it's added as a peer dep and dev dep, and not actually included when Perseus is imported into webapp.

packages/perseus/src/widgets/label-image/answer-pill.tsx
```diff
@@ -8,16 +8,31 @@ import {Popper} from "react-popper";
 
 import Renderer from "../../renderer";
 
+const BringToFront: CSSProperties = {
```
jeremywiebe, line 11: **question**: Can this be moved to the `styles` object below?

jeremywiebe (commented):
I was looking through the recent changes and had one nitpicky question.

packages/perseus/src/widgets/label-image/answer-pill.tsx
nedredmond: The styles object doesn't take variables

nedredmond: Weirdly, it doesn't work when it's there.

PR #886, opened 2023-12-20
add tests for SorterEditor

handeyeco:
## Summary:
There wasn't a lot here to test that had accessible roles.


nedredmond (approved):
Thanks for fixing that typo too!

PR #891, opened 2023-12-20
add tests to DefinitionEditor

handeyeco:
## Summary:
Seems like things that use `Changeable` are a little problematic to test since they're not calling the `onChange` callback as expected.

Looking into it, it seems like there are several tickets to refactor it: https://khanacademy.atlassian.net/browse/LC-394


SonicScrewdriver (approved):
Looks good, thanks again! :)

PR #893, opened 2023-12-20
add tests for InputNumber

handeyeco:
## Summary:
I know it's depreciated, but it's still used quite often and it wasn't a lot of work to add.


packages/perseus-editor/src/widgets/input-number-editor.tsx
```diff
@@ -186,6 +186,7 @@ class InputNumberEditor extends React.Component<Props> {
                             type="text"
                             disabled={!this.props.inexact}
                             defaultValue={this.props.maxError}
+                            aria-label="Max error"
```
SonicScrewdriver, line 189: Thank you for adding more labels!!

SonicScrewdriver (approved):
Really appreciate you adding these! Sure it's deprecated, but this will help us catch any major regressions. :)

PR #895, opened 2023-12-21
ImageEditor improvements: Story, tests, refactor

jeremywiebe:
## Summary:

Adds a new Storybook story for the Image Editor.

Converts the editor component from legacy `createReactClass` to a slightly more modern `React.Component<>` class. 

image

Issue: --none--

## Test plan:

Check the component in storybook.
Review the component changes

### DIFF
--- a/.changeset/angry-ladybugs-drive.md
+++ b/.changeset/angry-ladybugs-drive.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": minor
+---
+
+Export utility Range and Size types from package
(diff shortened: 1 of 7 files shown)


Myranae (approved):
This looks good to me. Looks like you added a story, brought in some types, defined some types and defaults, and switched to class methods. It was interesting in the story to see that the dimensions only appeared after clicking into the `Image url` field. But once you click into the link field, the info populates. Might be good to have that as a note.
image

PR #896, opened 2023-12-21
Add confirmation before deleting a hint

jeremywiebe:
## Summary:

Adds a simple confirmation before proceeding with the hint is deleted. This will hopefully prevent alot of frustration by avoiding losing work.

Issue: LC-1591

## Test plan:

### DIFF
--- a/packages/perseus-editor/src/hint-editor.tsx
+++ b/packages/perseus-editor/src/hint-editor.tsx
@@ -346,6 +346,11 @@ class CombinedHintsEditor extends React.Component<CombinedHintsEditorProps> {
     };
 
     handleHintRemove: (i: number) => void = (i: number) => {
+        // eslint-disable-next-line no-alert
+        if (!confirm("Are you sure you want to delete this hint?")) {
+            return;
+        }
+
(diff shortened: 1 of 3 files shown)


packages/perseus-editor/src/hint-editor.tsx
```diff
@@ -346,6 +346,11 @@ class CombinedHintsEditor extends React.Component<CombinedHintsEditorProps> {
     };
 
     handleHintRemove: (i: number) => void = (i: number) => {
+        // eslint-disable-next-line no-alert
```
handeyeco, line 349: I imagine that this linting rule was made for user-facing code and it's okay here because it's internal users?

handeyeco (approved):
🌈

packages/perseus-editor/src/hint-editor.tsx
jeremywiebe: Correct. Ideally this would be a Wonder Blocks modal confirm thingy, but given this is only for internal users I think it's ok (we do this in webapp in a few places). IMHO, this is alot better than nothing for now.

PR #897, opened 2023-12-22
Convert a _bunch_ of string refs to React refs

jeremywiebe:
## Summary:

Perseus editor code still uses alot of string refs. I went through and converted the easy/simple ones to `React.createRef<>`s. 

Sorry: This is a large-ish PR, but I hope it's fairly straightforward to review.

  * Convert all string refs in Example widget and editor to React refs
  * Convert EditorPage itemEditor ref to a React ref
  * Convert EditorPage hintsEditor ref to a React ref
  * Remove unused manually-created <div> rendererMountNode
  * Convert ItemEditor's frame ref to a React ref
  * Convert ItemEditor's questionEditor ref to a React ref
  * Convert ItemEditor's itemExtrasEditor ref to a React ref
  * Convert all string refs in Editor to React refs
  * Convert IframeContentRenderer's container ref to a React ref
  * Convert all string refs in HintEditor to React refs
  * Convert all string refs in GradedGroupEditor to React refs
  * Convert all string refs in GroupEditor to React refs
  * Convert all string refs in InputNumberEditor to React refs
  * Convert all string refs in TableEditor to React refs
  * Convert all string refs in SimpleMarkdownTestEditor to React refs
  * Convert all string refs in SimpleMarkdownTester to React refs

Issue: --none--

## Test plan:

`yarn typecheck`
`yarn test`

I also did some exploring our Editor storybook stories.

### DIFF
--- a/packages/perseus-editor/src/components/__tests__/blur-input.test.tsx
+++ b/packages/perseus-editor/src/components/__tests__/blur-input.test.tsx
@@ -4,6 +4,8 @@ import * as React from "react";
 
 import BlurInput from "../blur-input";
 
+import "@testing-library/jest-dom"; // Imports custom matchers
+
 describe("BlurInput", () => {
     it("should render", () => {
         render(<BlurInput value="Hello world!" onChange={() => {}} />);
@@ -60,6 +62,6 @@ describe("BlurInput", () => {
         ref.current?.focus();
 
         // Assert
-        expect(document.activeElement).toBe(screen.getByRole("textbox"));
+        expect(screen.getByRole("textbox")).toHaveFocus();
     });
 });
--- a/packages/perseus-editor/src/components/__tests__/blur-input.test.tsx
+++ b/packages/perseus-editor/src/components/__tests__/blur-input.test.tsx
@@ -47,4 +47,19 @@ describe("BlurInput", () => {
         // Assert
         expect(onChange).not.toHaveBeenCalledTimes(1);
     });
+
+    it("should focus input through focus function", () => {
+        const ref = React.createRef<BlurInput>();
+
+        // Arrange
+        render(
+            <BlurInput ref={ref} value="Hello world!" onChange={() => {}} />,
+        );
+
+        // Act
+        ref.current?.focus();
+
+        // Assert
+        expect(document.activeElement).toBe(screen.getByRole("textbox"));
+    });
 });
--- a/packages/perseus-editor/src/components/__tests__/blur-input.test.tsx
+++ b/packages/perseus-editor/src/components/__tests__/blur-input.test.tsx
@@ -1,0 +1,50 @@
+import {render, screen} from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import * as React from "react";
+
+import BlurInput from "../blur-input";
+
+describe("BlurInput", () => {
+    it("should render", () => {
+        render(<BlurInput value="Hello world!" onChange={() => {}} />);
+    });
+
+    it("should call onChange prop on blur", () => {
+        // Arrange
+        const onChange = jest.fn();
+        render(<BlurInput value="Hello world!" onChange={onChange} />);
+
+        // Act
+        userEvent.type(screen.getByRole("textbox"), "Hello Khan Academy!");
+        userEvent.tab();
+
+        // Assert
+        expect(onChange).toHaveBeenCalledTimes(1);
+    });
+
+    it("should call onChange prop when typing into input", () => {
+        // Arrange
+        const onChange = jest.fn();
+        render(<BlurInput value="Hello world!" onChange={onChange} />);
+
+        // Act
+        userEvent.type(screen.getByRole("textbox"), "Hello Khan Academy!");
+        // NO TAB
+
+        // Assert
+        expect(onChange).not.toHaveBeenCalledTimes(1);
+    });
+
+    it("should call onChange prop when pasting into input", () => {
+        // Arrange
+        const onChange = jest.fn();
+        render(<BlurInput value="Hello world!" onChange={onChange} />);
+
+        // Act
+        userEvent.paste(screen.getByRole("textbox"), "Hello Khan Academy!");
+        // NO TAB
+
+        // Assert
+        expect(onChange).not.toHaveBeenCalledTimes(1);
+    });
+});
--- a/packages/perseus-editor/src/components/blur-input.tsx
+++ b/packages/perseus-editor/src/components/blur-input.tsx
@@ -27,6 +27,8 @@ type State = {
  */
 // eslint-disable-next-line react/no-unsafe
 class BlurInput extends React.Component<Props, State> {
+    input = React.createRef<HTMLInputElement>();
+
     constructor(props: Props) {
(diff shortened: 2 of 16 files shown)


packages/perseus-editor/src/components/blur-input.tsx
```diff
@@ -45,9 +47,14 @@ class BlurInput extends React.Component<Props, State> {
         this.props.onChange(e.target.value);
     };
 
+    focus() {
```
jeremywiebe, line 50: We call `.focus()` on refs for this component in some places. I suspect that was broken code previously and was likely broken by a refactoring from a bare `<input/>` element to this component.

packages/perseus-editor/src/editor-page.tsx
```diff
@@ -108,7 +109,6 @@ class EditorPage extends React.Component<Props, State> {
         // this.isMounted() but is still considered an anti-pattern.
         this._isMounted = true;
 
-        this.rendererMountNode = document.createElement("div");
```
jeremywiebe, line 111: This is never used after it is created and since it is never assigned an id or class, nor added to the DOM, I can't imagine how it's used!

packages/perseus-editor/src/editor.tsx
```diff
@@ -183,11 +184,11 @@ class Editor extends React.Component<Props, State> {
         // this.props.onChange during that, since it calls our parent's
         // setState
         this._sizeImages(this.props);
-        // eslint-disable-next-line react/no-string-refs
-        // @ts-expect-error - TS2769 - No overload matches this call.
-        $(ReactDOM.findDOMNode(this.refs.textarea))
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        $(this.textarea.current!)
```
jeremywiebe, line 188: jQuery doesn't like potential `undefined`, but this is exactly how the previous code was, and given this is in `componentDidMount`, we _should_ have a valid ref at this point.

packages/perseus-editor/src/editor.tsx
```diff
@@ -211,20 +210,11 @@ class Editor extends React.Component<Props, State> {
...
-             */
-            // @ts-expect-error - TS2339 - Property 'focus' does not exist on type 'Element | Text'.
+        if (this.lastUserValue != null && textarea) {
             textarea.focus();
-            // @ts-expect-error - TS2339 - Property 'value' does not exist on type 'Element | Text'.
             textarea.value = this.lastUserValue;
-            // @ts-expect-error - TS2339 - Property 'selectionStart' does not exist on type 'Element | Text'.
             textarea.selectionStart = 0;
-            // @ts-expect-error - TS2339 - Property 'select' does not exist on type 'Element | Text'.
-            textarea.select(0, prevProps.content.length);
```
jeremywiebe, line 227: `.select()` ([docs](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/select)) on the `<textarea>` element doesn't take any arguments, but `setSelectionRange` ([docs](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/setSelectionRange)) does!

packages/perseus-editor/src/editor.tsx
```diff
@@ -724,21 +710,21 @@ class Editor extends React.Component<Props, State> {
...
             this.props.content,
-            // @ts-expect-error - TS2339 - Property 'selectionStart' does not exist on type 'ReactInstance'. | TS2339 - Property 'selectionEnd' does not exist on type 'ReactInstance'.
             [textarea.selectionStart, textarea.selectionEnd],
             widgetType,
         );
-        // @ts-expect-error - TS2339 - Property 'focus' does not exist on type 'ReactInstance'.
         textarea.focus();
     };
 
-    addTemplate: (e: React.SyntheticEvent<HTMLTextAreaElement>) => void = (
```
jeremywiebe, line 740: This is used in an event handler for a `<select>` element, not `<textarea>`!

packages/perseus-editor/src/editor.tsx
```diff
@@ -820,32 +806,17 @@ class Editor extends React.Component<Props, State> {
     };
 
     focus: () => void = () => {
-        // eslint-disable-next-line react/no-string-refs
-        const textarea = ReactDOM.findDOMNode(this.refs.textarea);
+        const textarea = this.textarea.current;
         if (textarea) {
-            /**
-             * TODO(somewhatabstract, JIRA-XXXX):
-             * textarea should be refined with an instanceof check to
```
jeremywiebe, line 828: :ta-done:! ✅ 😄

packages/perseus-editor/src/components/__tests__/blur-input.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+        // Arrange
+        render(
+            <BlurInput ref={ref} value="Hello world!" onChange={() => {}} />,
+        );
+
+        // Act
+        ref.current?.focus();
+
+        // Assert
+        expect(document.activeElement).toBe(screen.getByRole("textbox"));
```
handeyeco, line 63: [Maybe?](https://github.com/testing-library/jest-dom#tohavefocus)

```suggestion
        expect(screen.getByRole("textbox")).toHaveFocus();
```

packages/perseus-editor/src/editor.tsx
handeyeco: TIL about the TS non-null assertion operator. Is this pretty common in TS or would it be worth commenting the code? I guess the ESLint exception is a clue, but maybe why we're adding the exception?

packages/perseus-editor/src/widgets/graded-group-editor.tsx
```diff
@@ -33,16 +33,17 @@ class GradedGroupEditor extends React.Component<Props> {
         hint: null,
     };
 
+    editor = React.createRef<Editor>();
+    hintEditor = React.createRef<Editor>();
```
handeyeco, line 37: They're both `Editor`? `hintEditor` doesn't need to be something like `HintEditor`? I imagine there's some shared interface or inheritance or [some other important sounding type word]?

handeyeco (approved):
I think this looks good. The logic is changed and I think in a perfect world we would have tests that covered the changed logic. I realize this work is somewhat timeboxed though and I'm glad to have these changes. Also for what logic changed, I imagine it's changed from "ref doesn't exist, code will explode now" to "ref doesn't exist, I'm going to exit early" which is almost certainly an improvement.

So I approve 👍 Thanks for doing this, it's a quality improvement.

packages/perseus-editor/src/editor.tsx
jeremywiebe: I don't think it's very common. I'm used to it as there is a similar operator in Swift that we sometimes used. 

I'll add a comment.

packages/perseus-editor/src/widgets/graded-group-editor.tsx
jeremywiebe: Yes, in this case both refs are an `Editor`. I think it's because a Hint in Perseus is very, very close to a regular question. 

image

PR #902, opened 2024-01-03
Update OnFocusChange logic so that it returns the keypadHeight rather than the entire DomNode for the keypad

SonicScrewdriver:
## Summary:
We have recently noticed a regression related to scrolling in the mobile application after fixing an issue related to our MathInput package not updating. The issue is that we're unable to properly calculate the keypadHeight using the new keypad mounting approach. 

This PR is a partial fix for that issue by simplifying the OnFocusChange method so that it returns the keypadHeight directly, rather than the DOMNode for the keypad. This will allow us to centralize the logic and keep things DRY across Mobile Web and our Mobile App. 

Issue: LC-1575

## Test plan: 
- manual testing
- updated unit tests accordingly

### DIFF
--- a/packages/perseus/src/widgets/__stories__/expression.stories.tsx
+++ b/packages/perseus/src/widgets/__stories__/expression.stories.tsx
@@ -1,4 +1,5 @@
 import {KeypadContext, KeypadType} from "@khanacademy/math-input";
+import {action} from "@storybook/addon-actions";
 import * as React from "react";
 
 import {ServerItemRendererWithDebugUI} from "../../../../../testing/server-item-renderer-with-debug-ui";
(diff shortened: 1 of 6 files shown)


packages/perseus/src/widgets/__stories__/expression.stories.tsx
```diff
@@ -46,6 +46,7 @@ const WrappedKeypadContext = ({
                             apiOptions={{
                                 isMobile: isMobile,
                                 customKeypad: customKeypad,
+                                onFocusChange: () => {},
```
handeyeco, line 49: Wonder if we could use a SB action here like we do in a few other stories.

```suggestion
                                onFocusChange: action("onFocusChange"),
```

handeyeco (approved):
Looks good, thank you.

PR #905, opened 2024-01-03
Type interaction-editor sub-components

handeyeco:
## Summary:
Prep work for LC-1541

Convert propTypes to TS types

Issue: LC-1541


benchristel (approved):
The new types caught a problem! There's a type error in `interaction-editor.stories.tsx` because the `apiOptions` prop doesn't exist. Other than that, this looks good.

PR #908, opened 2024-01-03
Change table zebra-stripe color to Wonder Blocks offWhite

benchristel:
## Summary:
This ensures that color-coded math expressions have accessible contrast
with the darker table rows.

Issue: https://khanacademy.atlassian.net/browse/LC-1577

Test plan:

Run storybook: `yarn storybook`.

Visit
`http://localhost:6006/?path=/story/perseus-renderers-item-renderer--label-image-item`

You should see a table of dog breeds. The background color of the
odd-numbered rows should be `#f7f8fa`.

Screen Shot 2024-01-03 at 3 44 35 PM

### DIFF
--- a/packages/perseus/src/styles/perseus-renderer.less
+++ b/packages/perseus/src/styles/perseus-renderer.less
@@ -314,7 +314,7 @@
         tr:nth-child(odd) {
             // Doesn't work in IE8 :(
             td {
-                background-color: #f7f8fa; // Wonder Blocks offWhite
+                background-color: @offWhite;
             }
         }
     }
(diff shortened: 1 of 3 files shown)


packages/perseus/src/styles/perseus-renderer.less
```diff
@@ -314,7 +314,7 @@
         tr:nth-child(odd) {
             // Doesn't work in IE8 :(
             td {
-                background-color: #ededed;
+                background-color: #f7f8fa; // Wonder Blocks offWhite
```
handeyeco, line 317: I think this is fine, but I think I'd prefer moving this color declaration to `variables.less`. This makes it reusable, easy to change, and is a little more self-documenting.

```suggestion
                background-color: @offWhite;
```

handeyeco (approved):

packages/perseus/src/styles/perseus-renderer.less
benchristel: I wish Wonder Blocks exported a .less file with these variables. This is a good alternative, though. Thanks!

PR #911, opened 2024-01-04
replace MathInput with MathquillInput

handeyeco:
## Summary:
Basically this just replaces the Perseus' MathInput with a minimalist MathquillInput.

Why tho?
- Perseus' MathInput was overkill for what we're doing in Perseus-Editor. The keypad was disabled, so it was basically just being used as a regular Mathquill input. Now that we have `createMathField` it's pretty easy for use to make a vanilla Mathquill input.
- While changing things for learners, we were breaking things for content creators. For instance, we were disabling the keypad...but not removing the button to toggle the keypad. So we just had a broken button in every input that no one noticed.
- Eventually we might want to merge Perseus' MathInput with MathInput's MathInput; having it used in more places would make that more complicated.
- Perseus' MathInput was only used for the Interaction widget which is depreciated. By severing ties, we can iterate more for learners without having to think about the content editing side of things.

Issue: [LC-1541](https://khanacademy.atlassian.net/browse/LC-1541)

https://github.com/Khan/perseus/assets/16308368/8f9d59e8-a160-4b4c-9a3a-59ff9f7863ca

[LC-1541]: https://khanacademy.atlassian.net/browse/LC-1541?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


benchristel (approved):
I think this is good. One minor point: as a newcomer to this code, I'm not totally clear on what the difference is between MathInput and MathquillInput, i.e. which one is which. The name difference seems arbitrary, since both inputs are ultimately using Mathquill, right? Or am I missing something?

PR #915, opened 2024-01-08
Remove Transformer

nixterrimus:
_We don't talk about transformer._

This PR removes the transformer widget and subs in the 'deprecated-standing' widget.

Issue: https://khanacademy.atlassian.net/browse/LC-1654

Testing:
- Run the tests
- **Everything should still pass**

### DIFF
--- a/packages/perseus/src/types.ts
+++ b/packages/perseus/src/types.ts
@@ -1,4 +1,3 @@
-import type {SerializedHighlightSet} from "./components/highlighting/types";
 import type {ILogger} from "./logging/log";
 import type {Item} from "./multi-items/item-types";
 import type {
@@ -92,9 +91,6 @@ export type ChangeHandler = (
         currentValue?: string;
         // perseus-all-package/widgets/dropdown.jsx
         selected?: number;
-        tools?: ReadonlyArray<any>;
-        transformations?: ReadonlyArray<any>;
-        highlights?: SerializedHighlightSet;
         // perseus-all-package/widgets/grapher.jsx
         plot?: any;
     },


packages/perseus/src/perseus-types.ts
```diff
@@ -456,7 +452,6 @@ export type PerseusGrapherWidgetOptions = {
               // Two points on the same slope in the tangent wave line.
               coords: [Coord, Coord];
           };
-    // TODO(jeremy): Unify with PerseusTransformerWidgetOptions's .graph
```
jeremywiebe, line 459: 🎉  Nothing better than deleting code to fulfill a TODO!

packages/perseus/src/types.ts
```diff
@@ -92,7 +92,6 @@ export type ChangeHandler = (
         currentValue?: string;
         // perseus-all-package/widgets/dropdown.jsx
         selected?: number;
-        // perseus-all-package/widgets/transformer.jsx
         tools?: ReadonlyArray<any>;
         transformations?: ReadonlyArray<any>;
         highlights?: SerializedHighlightSet;
```
jeremywiebe, line 97: **question**: The way the deleted comment worked is that it signified the following change handler props were from the `transformer` widget.  You might be able to delete these three args now that `transformer` is gone!

jeremywiebe (approved):
What a great day!   I had one thought as I worked through this. We'll need to look at the translation pipeline (I can pair with you on that) and see if there's any code there we need to update/delete.  I suspect that we can safely delete the `transformer` widget and leave the translation pipeline alone as we aren't backfilling to remove `transformer` references.  But it is probably worth doing a quick look to ensure we aren't missing anything.

packages/perseus/src/types.ts
jeremywiebe: It's no harm leaving it, but _if_ we did still use it, I suspect something somewhere would crash, yes.

PR #916, opened 2024-01-08
Introduce orderer story that uses images

jeremywiebe:
## Summary:

Khan Academy content for early math often asks learners to demonstrate understanding of numbers by adding items to a box. Example: "Add 6 flowers to the box" ([live link](https://www.khanacademy.org/math/early-math/cc-early-math-counting-topic/cc-early-math-counting/e/counting-out-1-20-objects)).  The learner then uses the `orderer` widget to drop 6 flower images into the grey box. 

image

This PR adds a story that illustrates this usage of the `orderer` widget.

Issue: "none"

## Test plan:

Navigate to the story and test it out.


benchristel (approved):
Looks great! Thanks!

PR #917, opened 2024-01-09
Update Perseus' Storybook to use MathJax instead of KaTeX

Myranae:
## Summary:
Storybook for Perseus was still using KaTeX for math expressions. This updates it to use MathJax.  

BEFORE:
image

NOW:
image

NOTE: There are still references to KaTeX in Perseus. More research is required to make sure MathJax is used everywhere in Perseus or at least to remove all references to KaTeX.

Issue: LC-1573

## Test plan:
- Confirm Perseus Storybook stories with math use MathJax instead of KaTex
- Confirm all tests pass


benchristel:
The code looks good! The GitHub actions are failing because they don't have access to download `@khanacademy/mathjax-renderer`, which is a private repo. We should figure out how e.g. webapp's GitHub actions get access to private repositories.

jeremywiebe (approved):

PR #920, opened 2024-01-10
Add SSH private key through a new shared workflow action

Myranae:
## Summary:
The MathJaxRenderer repo requires an SSH key to access because it is a private repo. This was preventing us from adding MathJax to Perseus. This change provides Perseus with the key needed to access a private repo. 
 
Issue: LC-1573

## Test plan:
- Confirm all checks pass
- Confirm checks on PR for adding MathJax to Perseus pass
- Confirm the update-browserslist-database Nightly Maintenance Task completes without error - may need to add these same changes to mobile for this though (currently showing the same access error due to lack of access to MathJax)


.gitignore
```diff
@@ -9,3 +9,4 @@ storybook-static
 .nyc_output
 *.tsbuildinfo
 /cypress/
+.idea
```
jeremywiebe, line 12: We're missing a trailing newline here. It's not super critical, but helps for future edits. 

Ref: https://gist.github.com/OleksiyRudenko/d51388345ea55767b7672307fe35adf3#

```suggestion
.idea

```

jeremywiebe (approved):
Let's go ahead and land this. If we need to fix anything, we can do that in a second PR. 

Thanks for doing this @Myranae

.gitignore
Myranae: Added!

PR #921, opened 2024-01-11
Make the Expression widget treat `sen` as equivalent to `sin`

benchristel:
## Summary:
`sen` is used instead of `sin` in Portuguese. This change allows the
Expression widget to accept both spellings. We simply convert any
instances of `sen` to `sin` when evaluating the learner's answer.

Issue: https://khanacademy.atlassian.net/browse/LC-290

Test plan:

- Start storybook: `yarn storybook`.
- Visit
  `http://localhost:6006/?path=/docs/perseus-widgets-expression--docs`.
  Enter `sen(x)` in one of the input fields. The word `sen` should be
  displayed in an upright font, indicating that it is an operator. `sin`,
  `cos`, `tan` etc. should also be displayed in an upright font.

Screen Shot 2024-01-11 at 2 26 02 PM

### DIFF
--- a/.changeset/new-colts-sip.md
+++ b/.changeset/new-colts-sip.md
@@ -1,0 +1,7 @@
+---
+"@khanacademy/math-input": minor
+"@khanacademy/perseus": minor
+---
+
+Make the Expression widget treat `sen` as equivalent to `sin`. The spelling
+`sen` is used in Portuguese.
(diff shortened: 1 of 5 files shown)


nedredmond (approved):
Looks good to me-- but where does the list of AutoOperatorNames come from? Do we need to anglicize more of those?

benchristel:
@nedredmond good question - I should add a comment explaining that.

The other names in the list are the defaults for MathQuill. The list is documented (sort of) here: https://docs.mathquill.com/en/latest/Config/#autooperatornames. That documentation links to the "Short Math Guide". The link is broken, but you can find a copy here: https://tug.ctan.org/info/short-math-guide/short-math-guide.pdf. The operators are in section 3.15.

PR #923, opened 2024-01-12
Update order of steps in action.yml

Myranae:
## Summary:
This reorders the steps so the SSH key is available before the npm install included in the shared-node-cache action.

Issue: LC-1573

## Test plan:
- Confirm checks pass
- After landing, confirm checks for PR #917 (updating MathJax in Perseus' Storybook) pass

### DIFF
--- a/.github/actions/shared-node-cache/action.yml
+++ b/.github/actions/shared-node-cache/action.yml
@@ -22,5 +22,3 @@ runs:
           uses: Khan/actions@shared-node-cache-v0
           with:
               node-version: ${{ inputs.node-version }}
-
-
--- a/.github/actions/shared-node-cache/action.yml
+++ b/.github/actions/shared-node-cache/action.yml
@@ -11,6 +11,9 @@ inputs:
         required: true
 
 # The steps this action runs.
+# The order of the two steps below needs to be maintained.
+# The ssh-agent step needs to be before the Node.js/Install step because our package.json uses
+# git+ssh. As this requires a valid SSH key, the ssh-agent action must be before the yarn install.
 runs:
     using: "composite"
     steps:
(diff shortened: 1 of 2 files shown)


.github/actions/shared-node-cache/action.yml
```diff
@@ -11,14 +11,17 @@ inputs:
         required: true
 
 # The steps this action runs.
+# The order of the two steps below needs to be maintained.
+# The ssh-agent step needs to be before the Node.js/Install step because our package.json uses
+# git+ssh. As this requires a valid SSH key, the ssh-agent action must be before the yarn install.
```
jeremywiebe, line 16: Yep. Thanks for adding this!

.github/workflows/node-ci.yml
```diff
@@ -239,7 +239,7 @@ jobs:
                   ssh-private-key: ${{ secrets.KHAN_ACTIONS_BOT_SSH_PRIVATE_KEY }}
             # Make sure our packages aren't growing unexpectedly
             # This must come last as it builds the old code last and so leaves the
-            # wrong code in place for the next job.
+            # wrong code in place for the next job; in other words, it leaves the repo on a base branch.
```
jeremywiebe, line 242: Nice.  I did like how in webapp they have a comment _after_ this step saying something like `# Don't add any steps after (see comment above this last step)` to avoid someone not seeing this comment you've just edited here.

jeremywiebe (approved):
Just asking to add a comment _after_ the `preact/compressed-size-action` step.  But other than that... good to go.  Thanks Tamara!

.github/workflows/node-ci.yml
Myranae: Added! :)

PR #925, opened 2024-01-17
Standin Widget messaging and content editor

nixterrimus:
## Summary:

- Updates the messaging and design of the widget for learners
- Makes the widget not visible in the content editor interface
- Adds a content editor so that when a deprecated widget is shown there is an editor that makes sense
- Lays the groundwork for swapping content editors in the future

Issue: https://khanacademy.atlassian.net/browse/LC-1636, https://khanacademy.atlassian.net/browse/LC-1637, https://khanacademy.atlassian.net/browse/LC-1645

## Test plan:
Tests + Storybook should be ok

### DIFF
--- a/packages/perseus/src/widgets/deprecated-standin/index.tsx
+++ b/packages/perseus/src/widgets/deprecated-standin/index.tsx
@@ -6,11 +6,11 @@ import type {PerseusScore, WidgetExports} from "../../types";
 
 // The props are type `any` on purpose so that this can receive props
 // from any deprecated widget
-type DeprecatedStandinProps = any;
+type Props = any;
 type Rubric = any;
 type UserInput = Empty;
 
-class DeprecatedStandin extends React.Component<DeprecatedStandinProps> {
+class DeprecatedStandin extends React.Component<Props> {
     static validate(userInput: UserInput, rubric: Rubric): PerseusScore {
         // Since this mean to replace an existing widget the learner
         // WILL earn points for this widget


packages/perseus-editor/src/widgets/deprecated-standin-editor.tsx
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    render(): React.ReactNode {
+        return (
+            <div>
+                <p>This widget has been deprecated and removed</p>
+                <p>
+                    Learners will see a message and they will not be graded on
+                    this part
+                </p>
```
jeremywiebe, line 23: **thought**: Would we want to provide further guidance here?  Perhaps "Please remove this widget (or replace it with a supported one) from this exercise and publish" ?

packages/perseus/src/init.ts
```diff
@@ -24,7 +24,7 @@ const init = function (options: PerseusOptions): Promise<undefined> {
     // This section replaces widgets
     // When it's time to replace the transformer widget
     // uncomment this line
-    // Widgets.replaceWidget("transformer", "always-correct");
+    // Widgets.replaceWidget("transformer", "deprecated-standin");
```
jeremywiebe, line 27: Thanks for unifying how we refer to this widget!

packages/perseus-editor/src/widgets/deprecated-standin-editor.tsx
```diff
@@ -0,0 +1,29 @@
...
+type Props = {
+    // Callback for when a widget prop is changed.
+    onChange: (options: any) => void;
+};
+
+class DeprecatedStandinEditor extends React.Component<Props> {
+    static widgetName = "deprecated-standin" as const;
+
+    serialize(): any {
+        return EditorJsonify.serialize.call(this);
```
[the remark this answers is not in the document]
jeremywiebe: I don't think you're missing context.  As it is now, the exercise can be re-saved and retain any props previously configured, you just can't modify them or see them. 

I think that actually makes sense. The author won't be able to update the `transformer` config, but at least this approach won't trigger a change to this widget configuration.

PR #926, opened 2024-01-18
Add generated labels to MathJax in Storybook

nedredmond:
Adds generated labels to MathJax in Storybook so that we can effectively test accessibility without involving webapp.

Issue: LC-1635

### DIFF
--- a/testing/test-mathjax.tsx
+++ b/testing/test-mathjax.tsx
@@ -23,11 +23,24 @@ type Props = {
 };
 
 export function TestMathjax({children: tex, onRender}: Props) {
-    const {domElement} = renderer.render(tex);
+    const ref = React.useRef<HTMLSpanElement>(null);
+    const {domElement, addLabelWhenPresentational} = React.useMemo(
+        () => renderer.render(tex),
+        [tex],
+    );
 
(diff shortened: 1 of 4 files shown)


testing/test-mathjax.tsx
```diff
@@ -23,11 +23,24 @@ type Props = {
...
     }, [tex, onRender]);
-    return <span dangerouslySetInnerHTML={{__html: domElement.outerHTML}} />;
+
+    React.useLayoutEffect(() => {
+        if (ref.current) {
+            addLabelWhenPresentational(ref.current);
+            ref.current.innerHTML = "";
+            ref.current.appendChild(domElement);
+        }
+    });
```
benchristel, line 43: Minor: Since `useLayoutEffect` callbacks run before `useEffect` callbacks, I wonder if this should be swapped with the `useEffect` block above, so the code executes in the same order that you read it.

benchristel (approved):
Looks good to me! I left a nonblocking comment inline - take it or leave it.

testing/test-mathjax.tsx
nedredmond: swapped!

PR #927, opened 2024-01-18
Delete unused code from tex.ts

benchristel:
## Summary:
There was a lot of code in this file related to our old KaTeX+MathJax2
renderer. Now that we have switched to MathJax 3, we can simplify things. I've
also deleted a couple functions that simply weren't used anywhere.

I changed the type of `processMath`'s `text` param to `string | number` because
the code that generates Graphie labels sometimes passes a number.

Issue: none

Test plan:

Review storybook stories that use TeX labels:

- Grapher widget
- Interactive Graph widget
- Number Line widget

The graph labels should display in the correct place.

### DIFF
--- a/packages/perseus/src/util/tex.ts
+++ b/packages/perseus/src/util/tex.ts
@@ -6,35 +6,12 @@ import {getDependencies} from "../dependencies";
 import KhanMath from "./math";
 import reactRender from "./react-render";
 
-declare const MathJax: any;
-
 function findChildOrAdd(elem: any, className: string) {
     const $child = $(elem).find("." + className);
     if ($child.length === 0) {
         return $("<span>").addClass(className).appendTo($(elem));
     }
     return $child;
-}
-
-function doCallback(
-    elem: HTMLElement,
-    callback: (() => unknown) | (() => void),
-) {
-    let tries = 0;
-    (function check() {
-        const height = elem.scrollHeight;
-        // Heuristic to guess if the font has kicked in
-        // so we have box metrics (magic number ick,
-        // but this seems to work mostly-consistently)
-        if (height > 18 || tries >= 10) {
-            callback();
-        } else {
-            tries++;
-            // TODO(jeff, CP-3128): Use Wonder Blocks Timing API
-            // eslint-disable-next-line no-restricted-syntax
-            setTimeout(check, 100);
-        }
-    })();
 }
 
 export default {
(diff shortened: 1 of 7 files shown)


packages/perseus/src/util/tex.ts
```diff
@@ -16,27 +14,6 @@ function findChildOrAdd(elem: any, className: string) {
...
-        // but this seems to work mostly-consistently)
-        if (height > 18 || tries >= 10) {
-            callback();
-        } else {
-            tries++;
-            // TODO(jeff, CP-3128): Use Wonder Blocks Timing API
-            // eslint-disable-next-line no-restricted-syntax
-            setTimeout(check, 100);
-        }
-    })();
```
benchristel, line 37: This hack was pretty horrible! We polled every 100ms waiting for the math fonts to load in so we could update the layout. We don't need to do this anymore, because MathJax 3 includes the font metrics in its JavaScript, so it can lay out the math accurately even before the fonts have loaded.

packages/perseus/src/util/tex.ts
```diff
@@ -64,29 +41,6 @@ export default {
...
-            const script: HTMLElement | undefined = $mathjaxHolder.find(
-                "script[type='math/tex']",
-            )[0];
-
-            // If text wasn't provided, we look in two places
-            if (text == null) {
-                if ($elem.attr("data-math-formula")) {
-                    // The old typeset formula
-                    // @ts-expect-error - TS2322 - Type 'string | undefined' is not assignable to type 'string'.
-                    text = $elem.attr("data-math-formula");
```
benchristel, line 83: hmm, I'm actually not confident we can get rid of this line. `text` is typed as non-null, so I didn't think this code was reachable, but I wonder if the types are telling the truth.

packages/perseus/src/util/tex.ts
```diff
@@ -102,45 +56,10 @@ export default {
...
-            $elem.attr("data-math-type", null);
-        }
-
-        return elem;
-    },
-
-    // Function to retrieve the formula of a typeset math node
-    retrieveMathFormula: function (elem: HTMLElement): any {
-        return $(elem).attr("data-math-formula");
-    },
```
benchristel, line 145: These functions were never called.

packages/perseus/src/util/tex.ts
jeremywiebe: Yeah. I'd spent some time working on this function a few years ago for mobile. Glad we can get rid of it today!!

jeremywiebe (approved):
This file is only used inside of Graphie today to render labels for points, etc.  I can't see any way that `text` can be null either, but beware that the types in this area could _easily_ lie. there's lots of `: any` in Graphie, etc.

PR #929, opened 2024-01-19
Remove dead code from graphie.ts

benchristel:
Issue: none

## Test plan:

Review the Grapher and Interactive Graph docs in Storybook


jeremywiebe (approved):
I guess if we ever need to plotPolar we'll just go back and grab this. :)  Thanks for starting the cleanup!

PR #930, opened 2024-01-19
Left-align image widget

nedredmond:
Removes centering style from image widget.

Issue: LC-1571


nedredmond:
@nixterrimus Great call. I was thinking the same-- or at least providing screenshots. Will do!

PR #933, opened 2024-01-23
Fixing some types in Perseus

handeyeco:
## Summary:
This is mostly just tweaking some types in Perseus and adding a type export for PerseusWidget so I can use it in Webapp.

## Test plan:
No major changes I don't think, tests and typechecks should continue to pass

### DIFF
--- a/.changeset/fast-jobs-type.md
+++ b/.changeset/fast-jobs-type.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/perseus": major
+"@khanacademy/perseus-editor": major
+---
+
+Update external and internal facing Perseus types


packages/perseus/src/perseus-types.ts
```diff
@@ -97,7 +97,7 @@ type Widget<Type extends string, Options> = {
     // NOTE: perseus_data.go says this is required even though it isn't necessary.
     alignment?: string;
     // Options specific to the type field of the widget.  See Perseus*WidgetOptions for more details
-    options: Options | null | undefined;
+    options: Options;
```
handeyeco, line 100: I couldn't find where a widget would be missing options?

jeremywiebe: I took a Quick Look at our English published content. Looks like there are two widget configs missing the `options` but they are also missing everything so I'd say they're invalid.  I'm in favour of making `options` non-optional. 👍

.changeset/fast-jobs-type.md
```diff
@@ -0,0 +1,6 @@
+---
+"@khanacademy/perseus": major
+"@khanacademy/perseus-editor": major
```
handeyeco, line 3: I'm open to debating the `major` here. From my perspective: we're changing external-facing types that have a fairly high chance of requiring us to make updates in Webapp. 🤷

jeremywiebe (approved):

PR #935, opened 2024-01-23
Define parameters of drawing functions explictly instead of using metaprogramming

benchristel:
Issue: LC-1662

## Test plan:

Review the Storybook docs for Interactive Graph, Number Line, and Grapher

### DIFF
--- a/packages/perseus/src/util/graphie.ts
+++ b/packages/perseus/src/util/graphie.ts
@@ -644,6 +644,24 @@ export class Graphie {
         );
     }
 
+    // (x, y) is coordinate of bottom left corner
+    rect(x, y, width, height, style?: Record<string, any>) {
+        return this.withStyle(style, () => {
+            // Raphael needs (x, y) to be coordinate of upper left corner
+            const corner = this.scalePoint([x, y + height]);
+            const dims = this.scaleVector([width, height]);
+            const elem = this.postprocessDrawingResult(
+                this.raphael.rect(...corner.concat(dims)),
+            );
+
+            if (this.isMobile) {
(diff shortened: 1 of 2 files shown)


packages/perseus/src/util/graphie.ts
```diff
@@ -580,93 +614,324 @@ export class Graphie {
...
+            const startVector = polar(radius, startAngle);
+            const endVector = polar(radius, endAngle);
+
+            const startPoint = this.scalePoint([
+                +(center[0] + startVector[0]).toFixed(6),
+                +(center[1] + startVector[1]).toFixed(6),
+            ]);
+            const endPoint = this.scalePoint([
+                +(center[0] + endVector[0]).toFixed(6),
+                +(center[1] + endVector[1]).toFixed(6),
```
benchristel, line 659: I should figure out a better way of rounding these, or a better way of phrasing the test so we don't have to round them.

packages/perseus/src/util/graphie.ts
```diff
@@ -690,21 +960,164 @@ export class Graphie {
...
+        });
+    }
 
     plot(
         fn: (x: number) => number,
         range: Interval,
-        style?: Record<string, any>,
-    ): RaphaelElement {}
+        arg3?: boolean | Record<string, any>,
+        arg4?: Record<string, any>,
```
benchristel, line 1103: We should have a method type for this similar to the one for Label. Otherwise you can pass two style objects.

jeremywiebe (approved):
Paired through the review.  LGTM.

PR #936, opened 2024-01-24
Copy over helper functions and tests for extracting data

Myranae:
## Summary:
This copies functions currently used in Webapp for extracting Perseus widget information into Perseus. The goal is to make all interactions with the internals of Perseus occur with code from its repo and external users then use an API to access information. The next step will be to refactor the usages in Webapp to use these helper functions and to remove this code from Webapp.

Issue: LC-1639

## Test plan:
- Use an NPM snapshot to test these changes with Webapp
- Confirm all PR checks pass
- Confirm all tests pass


handeyeco (approved):
My understand is that most of this is a straight copy/paste from Webapp without modifications per Jeremy's suggestion. If that understanding is correct then I think we're set to release this!

PR #938, opened 2024-01-25
add parsePerseusJSON

handeyeco:
## Summary:
I feel a little silly about this PR, but I think it'll be useful. Just exporting a wrapper around `JSON.parse` so we have a one-stop shop for Parsing Perseus JSON needs.

Issue: LC-1656

## Test plan:
- unit test

### DIFF
--- a/packages/perseus/src/util/parse-perseus-json.test.ts
+++ b/packages/perseus/src/util/parse-perseus-json.test.ts
@@ -1,0 +1,10 @@
+import parsePerseusJSON from "./parse-perseus-json";
+
+describe("parsePerseusJSON", () => {
+    it("should parse JSON", () => {
+        const result = parsePerseusJSON(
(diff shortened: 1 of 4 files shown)


packages/perseus/src/util/parse-perseus-json.test.ts
```diff
@@ -0,0 +1,10 @@
+import parsePerseusJSON from "./parse-perseus-json";
+
+describe("parsePerseusJSON", () => {
+    it("should parse JSON", () => {
+        const result = parsePerseusJSON(
+            `{ "question": { "content": "idk why I'm testing this" }}`,
+        );
+        expect(result.question.content).toBe("idk why I'm testing this");
```
jeremywiebe, line 8: Well, at a minimum we're validating that Typescript isn't complaining about types here. 🤷‍♂️

jeremywiebe (approved):

PR #940, opened 2024-01-25
Fix pills so they allow for multiline answers

nishasy:
## Summary

The latest update to Wonder Blocks Pill give it a set height. This makes it cut off multi-line answers in the answer pills in the label-image widget.

To fix this, I added `height: "auto"` to the pill style to reset its height and stop the clipping.

Issue: LC-1664

## Test plan:

- Comment out the added `height: "auto"`
- Go to label-image widget storybook
  - http://localhost:6006/?path=/story/perseus-widgets-label-image--label-image-numberline
- Click on a number line dot and select any of the options.
- Confirm that the answer pill is cut off.
- Add back the `height: "auto"`
- Go back to the number line and confirm that the answer pill is fully displayed now.

### Before

Screenshot 2024-01-25 at 1 56 17 PM

### After

Screenshot 2024-01-25 at 1 56 31 PM


packages/perseus/src/widgets/label-image/answer-pill.tsx
```diff
@@ -74,7 +74,7 @@ export const AnswerPill = (props: {
                     style={[
                         style,
                         popperStyle,
-                        styles.relative,
```
nishasy, line 77: I don't think `styles.relative` exists anymore?

nedredmond: Thanks for getting rid of that!

nedredmond:
Would you also please upgrade WonderBlocks as a part of this?

From root, you can run `yarn upgrade-interactive --latest` and choose all the `@khanacademy/wonder-blocks-` choices.

That way, we can confirm it looks right with the latest version of WB.

nishasy:
@nedredmond Thanks, that might explain why I don't see the issue in the Perseus storybook!

nedredmond (approved):
Looks great-- just changeset patch `math-input` to fix the ❌ and you're good to land!

PR #942, opened 2024-01-25
Creation of a callback that can allow us to determine the shuffle order of various widget choices. 

SonicScrewdriver:
## Summary:
A major pain point that we've been hitting lately for several initiatives is that there was no way to determine the shuffle order of our answers, as it happens during the exercise rendering phase. 

This PR adds a callback to apiOptions so that we can "getWidgetsStartProps", which will provide us with the current shuffled order of the answers. 

Issue: LC-1638

## Test plan:
- New test 
- Manual testing

### DIFF
--- a/packages/perseus/src/__tests__/renderer.test.tsx
+++ b/packages/perseus/src/__tests__/renderer.test.tsx
@@ -291,16 +291,16 @@ describe("renderer", () => {
             );
         });
 
-        it("should call the getAllWidgetsStartProps callback if provided in apiOptions", () => {
+        it("should call the onWidgetStartProps callback if provided in apiOptions", () => {
             // Arrange
-            const getAllWidgetsStartProps = jest.fn();
-            const apiOptions: APIOptions = {getAllWidgetsStartProps};
+            const onWidgetStartProps = jest.fn();
+            const apiOptions: APIOptions = {onWidgetStartProps};
 
             // Act
             renderQuestion(mockedRandomItem, apiOptions);
 
             // Assert
-            expect(getAllWidgetsStartProps).toHaveBeenCalledWith(
+            expect(onWidgetStartProps).toHaveBeenCalledWith(
                 mockedShuffledRadioProps,
             );
         });
--- a/packages/perseus/src/__tests__/renderer.test.tsx
+++ b/packages/perseus/src/__tests__/renderer.test.tsx
@@ -31,6 +31,7 @@ import type {
     PerseusImageWidgetOptions,
     PerseusInputNumberWidgetOptions,
 } from "../perseus-types";
+import type {APIOptions} from "../types";
 
 // NOTE(jeremy): We can't use an automatic mock for the translation linter,
 // because one of it's "instance" methods is created using `debounce` and Jest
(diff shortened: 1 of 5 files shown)


packages/perseus/src/__tests__/renderer.test.tsx
```diff
@@ -287,6 +289,20 @@ describe("renderer", () => {
                 /^less than or equal to$/,
             );
         });
+
+        it("should call the getAllWidgetsStartProps callback if provided in apiOptions", () => {
+            // Arrange
+            const getAllWidgetsStartProps = jest.fn();
+            const apiOptions: Record<string, any> = {getAllWidgetsStartProps};
```
handeyeco, line 296: Would this work?

```suggestion
            const apiOptions: APIOptions = {getAllWidgetsStartProps};
```

packages/perseus/src/renderer.tsx
```diff
@@ -487,13 +487,18 @@ class Renderer extends React.Component<Props, State> {
         },
         props: Props,
     ) => any = (allWidgetInfo, props) => {
+        const {apiOptions, problemNum} = props;
         // @ts-expect-error - TS2345 - Argument of type '{ [key: string]: PerseusWidget; }' is not assignable to parameter of type 'Partial<Record<string, CategorizerWidget>>'.
-        return mapObject(allWidgetInfo, (widgetInfo) => {
+        const widgetsStartProps = mapObject(allWidgetInfo, (widgetInfo) => {
```
handeyeco, line 492: I wonder if this is an opportunity to clean some old code. The `mapObject` helper seems confused about types so we have use expect error.
```suggestion
   type PerseusWidgetMap = {
            [key: string]: PerseusWidget;
        }
    _getAllWidgetsStartProps: (
        allWidgetInfo: PerseusWidgetMap,
        props: Props,
    ) => PerseusWidgetMap = (allWidgetInfo, props) => {
        const {apiOptions, problemNum} = props;
        const widgetsStartProps: PerseusWidgetMap = {};
        entries(allWidgetInfo).forEach([key, widgetInfo]) => {
            widgetsStartProps[key] = Widgets.getRendererPropsForWidgetInfo(
                widgetInfo,
                problemNum,
            );
        });
        apiOptions.getAllWidgetsStartProps?.(widgetsStartProps);
        return widgetsStartProps;
    };
```

packages/perseus/src/types.ts
```diff
@@ -228,6 +228,11 @@ export type APIOptions = Readonly<{
     // keystroke caused text typed in the text area to appear in it
     // only after a good few seconds.
     editorChangeDelay?: number;
+    // This is a callback function that returns all the Widget Start Props
+    // after they have been transformed by the widget's transform function.
+    // This is useful for when we need to know how a widget has shuffled its
+    // the available choices.
+    getAllWidgetsStartProps?: (widgets: any) => any;
```
handeyeco, line 235: I think it would be useful to have this `PerseusWidgetMap` type if we don't already.

```suggestion
    getAllWidgetsStartProps?: (widgets: PerseusWidgetMap) => PerseusWidgetMap;
```

handeyeco (commented):
This is looking good, I just wonder if there's an opportunity for use to add some types to this old code.

jeremywiebe (commented):
This is technically a good approach. But I'd like to question whether we've exhausted all ideas for shielding callers from knowing the internals of the Perseus JSON. Exposing a callback like this still encourages callers outside of Perseus to understand and "parse" these Perseus JSON objects. 

I understand I haven't been involved in all of the conversations, but I'd like to encourage us to move Perseus towards external callers treating Perseus data objects as opaque blobs instead of open data to be consumed/parsed/etc. Every case of external parsing further increases maintenance burdens and risk of hidden breakages if we need to change "internal" widget option structures.

handeyeco (approved):
🌈

PR #943, opened 2024-01-25
Add tests for Graphie drawing tools

benchristel:
These tests will cover the refactoring of the drawing tools (which will happen
in a future PR).

## Test plan:

`yarn test`

### DIFF
--- a/packages/perseus/src/util/graphie.test.ts
+++ b/packages/perseus/src/util/graphie.test.ts
@@ -17,6 +17,7 @@ function createMockRaphael() {
         setSize: jest.fn().mockName("raphael.setSize"),
         ellipse: jest.fn().mockName("raphael.ellipse"),
         rect: jest.fn().mockName("raphael.rect"),
+        path: jest.fn().mockName("raphael.path"),
     };
 }
 
@@ -235,9 +236,11 @@ describe("Graphie drawing tools", () => {
 
             graphie.ellipse([0, 0], [1, 1], {fill: "#123456"});
 
-            expect(mockRaphaelElement.attr).toHaveBeenCalledWith(expect.objectContaining({
-                fill: "#123456"
-            }));
+            expect(mockRaphaelElement.attr).toHaveBeenCalledWith(
+                expect.objectContaining({
+                    fill: "#123456",
+                }),
+            );
         });
 
         it("uses the default style, if none given", () => {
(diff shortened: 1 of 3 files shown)


packages/perseus/src/util/graphie.test.ts
```diff
@@ -0,0 +1,1251 @@
...
+        it("draws a path using the quadratic 'Q' SVG command", () => {
+            const graphie = createAndInitGraphie();
+            const mockRaphaelElement = createMockRaphaelElement();
+            graphie.raphael.path.mockReturnValue(mockRaphaelElement);
+
+            graphie.parabola(1, 2, 3);
+
+            expect(graphie.raphael.path).toHaveBeenCalledWith(
+                expect.stringContaining("Q"),
+            );
```
jeremywiebe, line 768: non-blocking: I wonder if it is it possible to assert a start or end point at least?  🤔

packages/perseus/src/util/graphie.ts
```diff
@@ -581,17 +587,135 @@ class Graphie {
         $.extend(this.currentStyle, processed);
     }
 
-    label(point: any, text: any, direction: any, latex?: any) {}
+    grid(
+        xr: Interval,
+        yr: Interval,
+        styleAttributes?: Record<string, any>,
+    ): RaphaelSet {
+        throw new Error("grid called on uninitialized Graphie");
```
jeremywiebe, line 595: *thought*: I am pretty sure we'll be able to get rid of these stubs once we get the `drawingTools` array fixed up, but I noticed we are handling these stubs in two different ways.  This stub throws whereas the `circle()` one just has a comment that its overwritten. Maybe always throw in the stub?

packages/perseus/src/util/graphie.ts
```diff
@@ -581,17 +587,135 @@ class Graphie {
...
+        scale: number,
+        padding: number,
+    ): {wrapper: HTMLDivElement; visibleShape: RaphaelElement} {
+        throw new Error("fixedEllipse called on uninitialized Graphie");
+    }
 
-    grid(xr: any, yr: any, styleAttributes: any) {}
+    arc(
+        center: Coord,
+        radius: Coord,
```
jeremywiebe, line 635: **nit**: In previous functions we called this `radii`.

jeremywiebe (approved):

PR #944, opened 2024-01-26
Remove unused parameters from plot and plotParametric

benchristel:
As I was moving methods into the Graphie class, I noticed that the shade and
fn2 params are never used. Removing them lets us simplify the code quite a bit.

## Test plan:

View the Storybook docs for Grapher, Interactive Graph, and Number Line.


benchristel:
@jeremywiebe it's going to be easiest to read this one commit at a time.

PR #946, opened 2024-01-29
Improve typechecking and fix @ts-expect-errors

benchristel:

Test plan:

Review the Interactive Graph docs in Storybook

### DIFF
--- a/packages/perseus/src/util/geometry.ts
+++ b/packages/perseus/src/util/geometry.ts
@@ -85,14 +85,6 @@ export function vector(a, b) {
     });
 }
 
-export function sumVectors(...points: Coord[]): Coord {
-    return points.reduce(addVectors, [0, 0]);
-}
-
-export function addVectors(a: Coord, b: Coord): Coord {
-    return [a[0] + b[0], a[1] + b[1]];
-}
-
 export function reverseVector(vector: Coord): Coord {
     return [-vector[0], -vector[1]];
 }
--- a/packages/perseus/src/util/geometry.ts
+++ b/packages/perseus/src/util/geometry.ts
@@ -83,6 +83,18 @@ export function vector(a, b) {
     });
 }
 
+export function sumVectors(...points: Coord[]): Coord {
+    return points.reduce(addVectors, [0, 0])
+}
+
+export function addVectors(a: Coord, b: Coord): Coord {
+    return [a[0] + b[0], a[1] + b[1]]
+}
+
+export function reverseVector(vector: Coord): Coord {
+    return [-vector[0], -vector[1]];
+}
+
 export function magnitude(v: ReadonlyArray<Coord>): number {
     return Math.sqrt(
         _.reduce(
--- a/packages/perseus/src/util/geometry.ts
+++ b/packages/perseus/src/util/geometry.ts
@@ -7,8 +7,9 @@ import _ from "underscore";
 
 import Util from "../util";
 
-import type {Coord, Line} from "../interactive2/types";
 import {sum} from "./math";
+
+import type {Coord, Line} from "../interactive2/types";
 
 const {eq, deepEq} = Util;
 
@@ -85,11 +86,11 @@ export function vector(a, b) {
 }
 
 export function sumVectors(...points: Coord[]): Coord {
-    return points.reduce(addVectors, [0, 0])
+    return points.reduce(addVectors, [0, 0]);
 }
 
 export function addVectors(a: Coord, b: Coord): Coord {
-    return [a[0] + b[0], a[1] + b[1]]
+    return [a[0] + b[0], a[1] + b[1]];
 }
 
 export function reverseVector(vector: Coord): Coord {
--- a/packages/perseus/src/util/geometry.ts
+++ b/packages/perseus/src/util/geometry.ts
@@ -8,6 +8,7 @@ import _ from "underscore";
 import Util from "../util";
 
 import type {Coord, Line} from "../interactive2/types";
+import {sum} from "./math";
 
 const {eq, deepEq} = Util;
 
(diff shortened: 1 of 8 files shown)


packages/perseus/src/util/geometry.ts
```diff
@@ -83,6 +85,31 @@ export function vector(a, b) {
...
+    return points.reduce(addVectors, [0, 0]);
+}
+
+export function addVectors(a: Coord, b: Coord): Coord {
+    return [a[0] + b[0], a[1] + b[1]];
+}
+
+export function reverseVector(vector: Coord): Coord {
+    return [-vector[0], -vector[1]];
+}
```
jeremywiebe, line 98: We have the `vector` module in `@khanacademy/kmath` that I think would be a better place for these more generic vector functions. In fact, we have `vector.add()` there already (https://github.com/Khan/perseus/blob/0775f5055acf53c9cca865f15dedd5ee5678ce8e/packages/kmath/src/vector.ts#L58). 

Could we move/replace these functions with using `vector`?

packages/perseus/src/util/graphie.ts
```diff
@@ -4,6 +4,7 @@ import {
     vector as kvector,
     number as knumber,
 } from "@khanacademy/kmath";
+import {entries} from "@khanacademy/wonder-stuff-core";
```
jeremywiebe, line 7: Now that we're using TS, I don't think we need to use this anymore. 

Webapp is just in the process of dropping the requirement to use `entries` and allow simply using `Object.entries()` directly. 

https://github.com/Khan/webapp/pull/19488

packages/perseus/src/util/graphie.ts
```diff
@@ -179,32 +221,33 @@ export class Graphie {
...
-                        ];
-                    }
-                } else if (typeof val === "number") {
-                    options[prop] = [
-                        [-val, val],
-                        [-val, val],
-                    ];
-                }
+                options[prop] = normalizeRange(options[prop])
+                // if (val.constructor === Array) {
```
jeremywiebe, line 225: Do these need to be deleted?

packages/perseus/src/util/graphie.ts
```diff
@@ -1005,13 +1041,6 @@ export class Graphie {
             };
             const clippedFn = (x) => clip(fn(x));
 
-            if (!this.currentStyle.strokeLinejoin) {
-                this.currentStyle.strokeLinejoin = "round";
-            }
-            if (!this.currentStyle.strokeLinecap) {
-                this.currentStyle.strokeLinecap = "round";
-            }
```
jeremywiebe, line 1013: Assuming this is unused code?

packages/perseus/src/util/graphie.ts
```diff
@@ -1386,8 +1410,7 @@ export class Graphie {
             if (typeof transformer === "function") {
                 $.extend(processed, transformer(value));
             } else {
-                const dasherized = key
-                    // @ts-expect-error - TS2339 - Property 'replace' does not exist on type 'string | number | symbol'.
+                const dasherized = String(key)
```
jeremywiebe, line 1413: I wonder if this would not be needed if line 1407 above used `attrs.forEach()` instead of `$.each(attrs`..

packages/perseus/src/util/graphie.ts
```diff
@@ -1411,6 +1434,57 @@ const labelDirections = {
...
+// - A pair of magnitudes for the x and y ranges, respectively. E.g. [5, 10]
+//   is equivalent to [[-5, 5], [-10, 10]].
+// - A single magnitude to be used for both x and y. E.g. 10 is equivalent to
+//   [[-10, 10], [-10, 10]]
+type RangeSpecifier = [Interval, Interval] | Coord | number;
+
+// Overload normalizeRange to specify that it only returns undefined if its
+// argument is undefined.
+export function normalizeRange(range: RangeSpecifier): [Interval, Interval];
+export function normalizeRange(range: RangeSpecifier | undefined): [Interval, Interval] | undefined;
```
jeremywiebe, line 1449: **question**: wouldn't this overload simply be this?

```suggestion
export function normalizeRange(range: undefined): undefined;
```

packages/perseus/src/util/graphie.ts
```diff
@@ -1411,6 +1434,57 @@ const labelDirections = {
     "above left": [-1.0, -1.0],
 } as const;
 
+// A range for a Graphie can be specified in one of three forms:
```
jeremywiebe, line 1437: I wonder if it's worth starting to pull out these types of things into their own files.  There are other concepts of `Range` that would be nice to start unifying:

```sh
rg "type Range"
packages/perseus/src/perseus-types.ts
6:export type Range = [number, number];

packages/perseus-editor/src/widgets/number-line-editor.tsx
11:type Range = [number, number];

packages/perseus-editor/src/widgets/interactive-graph-editor.tsx
46:type Range = [number, number]; // [min, max]

packages/perseus/src/util/geometry.ts
15:export type Range = [number, number];
```

I am good if you don't want to get distracted with this refactoring in this PR or just push it off as low priority too.

jeremywiebe (approved):

packages/perseus/src/util/graphie.ts
benchristel: It looks like `Object.entries` still doesn't have a very precise return type, so I'd prefer to keep using `entries`. I left a comment on the webapp PR.

packages/perseus/src/util/graphie.ts
benchristel: Yeah, nothing is setting `strokeLinejoin` or `strokeLinecap`.

packages/perseus/src/util/graphie.ts
benchristel: I thought about adding that overload, but it doesn't seem useful. If you know the input is always `undefined`, why call `normalizeRange`? 😂 

`RangeSpecifier | undefined` allows you to pass in a value that *might* be undefined, on the condition that you can handle getting an undefined value back. The other overload (without undefined) guarantees that if you don't pass undefined, you won't get undefined back.

packages/perseus/src/util/graphie.ts
benchristel: In general, duplicated types don't worry me much, because:

- they have no runtime cost
- TypeScript is structurally typed, so all these types are interchangeable
- it's nice to be able to see the definition of a commonly-used type without jumping around between files, so there is an ergonomic advantage to having types (re)defined close to where they're used.
- type checking should fail if the various definitions of a type get out of sync in a way that matters (I say *should* because our type coverage currently isn't great).

packages/perseus/src/util/graphie.ts
benchristel: Yes! Good catch :)

packages/perseus/src/util/graphie.ts
jeremywiebe: Ah, that makes sense. I hadn't thought about the types that _might_ be `undefined`, but might not.

PR #949, opened 2024-01-30
Use built-in JS functions instead of underscore/jQuery functions

benchristel:
## Summary:
The built-in APIs have better type support, and are idiomatic.

One underscore function that has no equivalent in vanilla JS is `_.zip`.
I've reimplemented zip so we can avoid importing underscore. This gets
us more robust typechecking around zip calls.

Test plan:

Review the Storybook docs for Grapher, Interactive Graph, and Number
Line.

### DIFF
--- a/packages/perseus/src/util/graphie.ts
+++ b/packages/perseus/src/util/graphie.ts
@@ -8,7 +8,6 @@ import {entries} from "@khanacademy/wonder-stuff-core";
 import $ from "jquery";
 // eslint-disable-next-line import/no-extraneous-dependencies
 import Raphael from "raphael";
-import _ from "underscore";
 
 // Minify Raphael ourselves because IE8 has a problem with the 1.5.2 minified
 // release
@@ -266,7 +265,7 @@ export class Graphie {
             ],
         ] as [Interval, Interval];
 
-        if (!_.isArray(unityLabels)) {
+        if (!Array.isArray(unityLabels)) {
             unityLabels = [unityLabels, unityLabels];
         }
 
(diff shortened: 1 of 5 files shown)


packages/perseus/src/util/graphie.ts
```diff
@@ -759,15 +758,17 @@ export class Graphie {
...
         center = center ? this.scalePoint(center) : null;
         createPath = createPath || this.svgPath;
 
         // Compute bounding box
-        const pathLeft = _.min(_.pluck(points, 0));
-        const pathRight = _.max(_.pluck(points, 0));
-        const pathTop = _.min(_.pluck(points, 1));
-        const pathBottom = _.max(_.pluck(points, 1));
+        const xs = points.map((p) => p[0]);
+        const ys = points.map((p) => p[1]);
```
jeremywiebe, line 767: Nice. This is clearer than the `pluck(, 0)` syntax!

packages/perseus/src/util/graphie.ts
```diff
@@ -950,16 +951,11 @@ export class Graphie {
...
             $span
-                .css(
-                    $.extend(
-                        {},
-                        {
-                            position: "absolute",
-                            padding: (pad != null ? pad : 7) + "px",
-                            color: "black",
-                        },
-                    ),
```
jeremywiebe, line 961: Weird...extending an empty object. That'd make sense if we wanted to force a copy but I don't understand why this would ever be needed.

packages/perseus/src/util/graphie.ts
```diff
@@ -1086,18 +1082,20 @@ export class Graphie {
...
-                (i === 0 ? "M" : "L") +
-                KhanMath.bound(scaled[0]) +
-                " " +
-                KhanMath.bound(scaled[1])
-            );
-        }).join("");
+        return points
+            .map((point, i) => {
+                if (point === true) {
+                    return "z";
```
jeremywiebe, line 1088: **observation**: It seems like you'd only ever want this as the last point (or can SVG paths close the shape more than once?).  Anyways... the code isn't changing in this PR. :)

jeremywiebe (approved):

packages/perseus/src/util/graphie.ts
benchristel: Yeah, the intent is to copy the object, but it's not necessary here, since there are no other references to the object we're copying. I suspect this is copy-pasta from some other code where we were updating `currentStyle` and wanted to do so without mutating the style object.

PR #950, opened 2024-01-31
Keypad: Move the FRAC button before the add/subtract/multiply/divide block

nishasy:
## Summary:
When the PERCENT button isn't present, the FRAC button
seems to be out of focus order when tabbing through the
Keypad. Also, even if it shows up in left-to-right order
when the PERCENT button is there, it still divides the
add/subtract/multiply/divide block, which is not intuitive
for screenreader users.

In this PR, I fixed this by moving the FRAC button just
before the PLUS button.

I also added a story for Keypad that includes all the
buttons excluding the navigation pad.

Issue: https://khanacademy.atlassian.net/browse/LC-1015

### Tab demos:

https://github.com/Khan/perseus/assets/13231763/0fbf1ea0-5264-4f20-80f4-c7974b62ac46

https://github.com/Khan/perseus/assets/13231763/dd59fd35-ce55-45e0-9951-d91d486e3bf3

https://github.com/Khan/perseus/assets/13231763/37410cb5-f3fd-46ce-b735-591b2f5b40fc

https://github.com/Khan/perseus/assets/13231763/fcde1305-a085-4219-9da5-f7252dd47d60

## Test plan:

Storybook
- Start storybook with `yarn start`
- Go to `/?path=/story/math-input-full-keypad--everything-minus-navigation-pad`
- Tab through all the keys within every keypad page
- The fraction key should receieve focus before the plus key on
  every page.

### DIFF
--- a/packages/math-input/src/components/keypad/shared-keys.tsx
+++ b/packages/math-input/src/components/keypad/shared-keys.tsx
@@ -37,6 +37,12 @@ export default function SharedKeys(props: Props) {
     return (
         <>
             <KeypadButton
+                keyConfig={Keys.FRAC}
+                onClickKey={onClickKey}
+                coord={fractionCoord}
+                secondary
+            />
+            <KeypadButton
                 keyConfig={Keys.PLUS}
                 onClickKey={onClickKey}
                 coord={[4, 0]}
@@ -46,12 +52,6 @@ export default function SharedKeys(props: Props) {
                 keyConfig={Keys.MINUS}
                 onClickKey={onClickKey}
                 coord={[5, 0]}
-                secondary
-            />
-            <KeypadButton
-                keyConfig={Keys.FRAC}
-                onClickKey={onClickKey}
-                coord={fractionCoord}
                 secondary
             />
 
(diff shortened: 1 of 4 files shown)


packages/math-input/src/components/keypad/shared-keys.tsx
```diff
@@ -37,21 +37,21 @@ export default function SharedKeys(props: Props) {
...
-                coord={[5, 0]}
+                coord={[4, 0]}
                 secondary
             />
             <KeypadButton
-                keyConfig={Keys.FRAC}
+                keyConfig={Keys.MINUS}
                 onClickKey={onClickKey}
-                coord={fractionCoord}
+                coord={[5, 0]}
```
jeremywiebe, line 54: So the order these show up in the React/HTML DOM is what defines the tab order, right?

jeremywiebe (approved):

packages/math-input/src/components/keypad/shared-keys.tsx
nishasy: Yup!

PR #951, opened 2024-01-31
rename helper to parsePerseusItem

handeyeco:
## Summary:
After digging more in the Webapp code, they're parsing more than just the PerseusItem.

Renaming this so it's more explicit about what it's for.


SonicScrewdriver (approved):
Makes sense :) Thank you!

PR #952, opened 2024-01-31
Label Image: screen reader reads answer alongside marker

nedredmond:
Moves marker label so that it's a sibling to the answer pill so that both are read. Also, ensures there is an explicit "button" label on Clickable so that we trigger the automated math label in MathJax Renderer.

https://github.com/Khan/perseus/assets/23404711/f63af37e-43e8-4b7f-b8d9-d37f737dcd16

Issue: LC-1701

Testing:
Open Safari with VoiceOver. Select an answer on Label Image Widget. Focus on marker and see that answer is read (and is accurate for math)

### DIFF
--- a/.changeset/plenty-tips-occur.md
+++ b/.changeset/plenty-tips-occur.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Screenreader reads Label Image Widget answer pills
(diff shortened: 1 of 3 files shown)


nedredmond:
@Myranae Were you using VoiceOver with Chrome? The supported combo is VoiceOver with Safari, where it does read math correctly.

nedredmond:
[@Myranae you can learn more about out supported SR/browser combos here](https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/1849524239)

PR #954, opened 2024-01-31
Fix @ts-expect-errors and `any` types in Graphie and nearby code

benchristel:
## Summary:

Test plan:

Review the Storybook docs

### DIFF
--- a/packages/kmath/src/vector.ts
+++ b/packages/kmath/src/vector.ts
@@ -26,6 +26,10 @@ export function zip<T>(...arrays: ReadonlyArray<T>[]): T[][] {
     return results;
 }
 
+export function map<T, U>(pair: [T, T], f: (a: T, i: number) => U): [U, U] {
+    return [f(pair[0], 0), f(pair[1], 1)];
+}
+
 /**
  * Checks if the given vector contains only numbers and, optionally, is of the
  * right dimension (length).
--- a/packages/perseus/src/components/graph.tsx
+++ b/packages/perseus/src/components/graph.tsx
@@ -1,5 +1,6 @@
 /* eslint-disable @khanacademy/ts-no-error-suppressions */
 /* eslint-disable react/no-unsafe */
+import {vector as kvector} from "@khanacademy/kmath";
 import $ from "jquery";
 import * as React from "react";
 import ReactDOM from "react-dom";
@@ -14,6 +15,7 @@ import SvgImage from "./svg-image";
 
 import type {Coord} from "../interactive2/types";
 import type {PerseusImageBackground} from "../perseus-types";
+import type {GridDimensions} from "../util";
 
 const defaultBackgroundImage = {
     url: null,
@@ -43,12 +45,12 @@ function numSteps(range: any, step: any) {
 }
 
 type Props = {
-    box: Readonly<[number, number]>;
+    box: [number, number];
     labels: ReadonlyArray<string>;
-    range: Readonly<[Coord, Coord]>;
-    step: Readonly<[number, number]>;
-    gridStep: Readonly<[number, number]>;
-    snapStep: Readonly<[number, number]>;
+    range: [Coord, Coord];
+    step: [number, number];
+    gridStep: [number, number];
+    snapStep: [number, number];
     markings: string;
     backgroundImage: PerseusImageBackground;
     showProtractor: boolean;
@@ -201,23 +203,26 @@ class Graph extends React.Component<Props> {
             Util.unescapeMathMode(label),
         );
         const range = this.props.range;
+        // @ts-expect-error - TS2345: Argument of type 'Element | Text | null' is not assignable to parameter of type 'HTMLElement'.
         const graphie = (this._graphie = GraphUtils.createGraphie(graphieDiv));
 
-        const gridConfig = this._getGridConfig();
+        const gridConfig: [GridDimensions, GridDimensions] =
+            this._getGridConfig();
+        // @ts-expect-error - TS2339: Property 'snap' does not exist on type 'Graphie'.
         graphie.snap = this.props.snapStep;
 
         if (this.props.markings === "graph") {
             graphie.graphInit({
                 range: range,
-                scale: _.pluck(gridConfig, "scale"),
+                scale: kvector.map(gridConfig, (g) => g.scale),
                 axisArrows: "<->",
                 labelFormat: function (s) {
                     return "\\small{" + s + "}";
                 },
                 gridStep: this.props.gridStep,
-                tickStep: _.pluck(gridConfig, "tickStep"),
+                tickStep: kvector.map(gridConfig, (g) => g.tickStep),
                 labelStep: 1,
-                unityLabels: _.pluck(gridConfig, "unityLabel"),
+                unityLabels: kvector.map(gridConfig, (g) => g.unityLabel),
                 isMobile: this.props.isMobile,
             });
             graphie.label(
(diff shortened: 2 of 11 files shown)


packages/kmath/src/vector.ts
```diff
@@ -26,6 +26,10 @@ export function zip<T>(...arrays: ReadonlyArray<T>[]): T[][] {
     return results;
 }
 
+export function map<T, U>(pair: [T, T], f: (a: T, i: number) => U): [U, U] {
```
jeremywiebe, line 29: **nit**: Most of the other `vector` functions are `ReadonlyArray<T>`.  This is specifically a 2-item tuple variant.  🤔   I guess if we want to make it [n-ary](https://xlinux.nist.gov/dads/HTML/naryfunc.html#:~:text=(definition),%2C%20binary%20function%2C%20trinary%20function.), we can! Never mind me. :)

packages/perseus/src/components/graph.tsx
```diff
@@ -43,12 +45,12 @@ function numSteps(range: any, step: any) {
...
+    box: [number, number];
     labels: ReadonlyArray<string>;
-    range: Readonly<[Coord, Coord]>;
-    step: Readonly<[number, number]>;
-    gridStep: Readonly<[number, number]>;
-    snapStep: Readonly<[number, number]>;
+    range: [Coord, Coord];
+    step: [number, number];
+    gridStep: [number, number];
+    snapStep: [number, number];
```
jeremywiebe, line 53: For posterity, Ben and I paired through this and decided that the use of `Readonly<>` caused more issues with types than we thought it helped. So we removed these and we can review if we want to use `Readonly<T>` in the future once we've improved types in this area further. It might be an easier thing to do once we have more type coverage.

packages/perseus/src/components/graphie.tsx
```diff
@@ -187,6 +189,7 @@ class Graphie extends React.Component<Props> {
             });
         }
 
+        // @ts-expect-error - TS2339: Property 'snap' does not exist on type 'Graphie'.
         graphie.snap = this.props.options.snapStep || [1, 1];
```
jeremywiebe, line 193: **thought**: It'd be nice if we passed this into `graphInit()` or `init()` instead of this direct assignment.  No change required, just musing on future improvements.

packages/perseus/src/util/graphie.test.ts
```diff
@@ -40,7 +40,7 @@ function createFakeRaphaelSet() {
 }
 
 function createAndInitGraphie(): Graphie {
-    const graphie = GraphUtils.createGraphie();
+    const graphie = GraphUtils.createGraphie(document.createElement("div"));
```
jeremywiebe, line 43: **thought**: Not sure if this is premature optimization, but it seems off that we don't add this div to the document.  Maybe the fact that it isn't needed is a statement that we don't assert anything about what Graphie does with this div yet?

packages/perseus/src/util/graphie.ts
```diff
@@ -1346,7 +1346,7 @@ export class Graphie {
         return this.drawingTransform().unscaleVector(point);
     };
 
-    private processAttributes(attrs: any) {
+    private processAttributes(attrs: Record<string, any> | undefined) {
```
jeremywiebe, line 1349: **question**: Should this type be the `StyleAttributes` type?

packages/perseus/src/util/graphie.ts
```diff
@@ -1538,22 +1538,12 @@ const GraphUtils: any = {
...
-    getDistance: function (point1, point2) {
+    getDistance: function (point1: Coord, point2: Coord): number {
         return kpoint.distanceToPoint(point1, point2);
     },
 
-    /**
-     * Round the given coordinates to a given snap value
-     * (e.g., nearest 0.2 increment)
-     */
-    snapCoord: function (coord, snap) {
```
jeremywiebe, line 1549: I'm really surprised this is unused. I wonder how we do coordinate snapping when dragging the movable points/lines around if this isn't used.  

No action required.

packages/perseus/src/widgets/measurer.tsx
```diff
@@ -121,17 +123,19 @@ const Measurer: any = createReactClass({
         const graphieDiv = ReactDOM.findDOMNode(this.refs.graphieDiv);
         // @ts-expect-error - TS2769 - No overload matches this call. | TS2339 - Property 'empty' does not exist on type 'JQueryStatic'.
         $(graphieDiv).empty();
+        // @ts-expect-error - Argument of type 'Element | Text | null' is not assignable to parameter of type 'HTMLElement'.
         const graphie = (this.graphie = GraphUtils.createGraphie(graphieDiv));
 
-        const scale = [40, 40];
-        const range = [
+        const scale: Coord = [40, 40];
+        const range: [Interval, Interval] = [
```
jeremywiebe, line 130: **idea**: I wonder if you can elide the types if you just add `as const` behind these literals?  I'm not sure which is better/more maintainable. 🤔

jeremywiebe (approved):
Nice. Although we're adding some ts-expect-errors, I think that's good as it means our `graphie` type is being "seen" in more places and we can begin improving it's type. 

Thanks for these improvements!

packages/kmath/src/vector.ts
benchristel: Yeah, the functions that take a `ReadonlyArray<T>` are a problem, because you can't assign their results to `Coord`. Since we are only working in 2 dimensions, I don't see any need to generalize this to more dimensions. If we do need to handle 3D someday, we can add an overload signature.

packages/perseus/src/components/graph.tsx
benchristel: Specifically, we couldn't figure out a way to type a readonly tuple that worked with type narrowing. The following code [does not typecheck](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABMOcAUAPAXIgSgUwEMATBAGwE8AeAbTBAFsAjfAJwBpF7m2BdAPkQAfLoxasAlDm7jEAbwBQiRDGCI0AQVatCFAHQwAzlp0VMEifKXLErfFBCskGGgAZeiANSIXARl7WAL7Wdg5OPgrBQA):

```ts
function foo(x: Readonly<[number, number]> | number): number {
  if (Array.isArray(x)) {
    return x[0] + x[1]
  }
  return x
}
```

packages/perseus/src/util/graphie.test.ts
benchristel: This is a cheap way of getting typechecking to pass. We apparently don't use this element in the code under test, since the tests were passing without it, but I figured that creating a dummy div is better than suppressing typechecking for this line.

packages/perseus/src/util/graphie.ts
benchristel: I haven't dug into it too deeply, but it looked to me like the Interactive code was defining and using its own snap methods.

packages/perseus/src/widgets/measurer.tsx
benchristel: The type names are useful as documentation. If you're just looking at a pair of numbers it's hard to know if `Coord` or `Interval` was intended.

packages/perseus/src/util/graphie.ts
benchristel: I'm not confident in giving it a type yet. The callers of `processAttributes` are passing `any` to it, so it's hard to tell what the actual type should be. I'll come back to this in a future PR.

PR #958, opened 2024-02-02
Test and simplify normalizeOptions

benchristel:
This is in preparation for making Movable an ES6 class. I attempted to do that,
but ran into problems because normalizeOptions was doing spooky magic. Hopefully
this will make the next refactoring easier.

## Test plan:

Review Storybook docs for Interactive Graph, Grapher, and Number Line

### DIFF
--- a/packages/perseus/src/interactive2/interactive-util.ts
+++ b/packages/perseus/src/interactive2/interactive-util.ts
@@ -120,7 +120,7 @@ const InteractiveUtil = {
      */
     normalizeOptions: function (options: Record<any, any>): Record<any, any> {
         const result = {...options};
-        _.each(FUNCTION_ARRAY_OPTIONS, function (eventName) {
+        FUNCTION_ARRAY_OPTIONS.forEach(function (eventName) {
             // Only propagate values which were set; not present values
             // shouldn't be added to options because we'd like them to
             // fall through to defaults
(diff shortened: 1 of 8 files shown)


packages/perseus/src/interactive2/interactive-util.ts
```diff
@@ -106,24 +118,16 @@ const InteractiveUtil = {
...
-        // function should really just modify this.state in place
-        // (and maybe be a function on MovableHelperMethods to get access
-        // to this.state), which would also be nicer because we could
-        // normalizeOptions once in this.modify
-        const result = _.clone(options);
-        _.each(arrayOptionNames, function (eventName) {
-            const funcOrArray = options[eventName];
+    normalizeOptions: function (options: Record<any, any>): Record<any, any> {
+        const result = {...options};
+        _.each(FUNCTION_ARRAY_OPTIONS, function (eventName) {
```
jeremywiebe, line 123: Do you want to switch this to `FUNCTION_ARRAY_OPTIONS.forEach(`... right away?

jeremywiebe, line 123: I was looking through this and was concerned that we've made the set of options the same for all movables now whereas before it looks like they each had their own set of option names they supported.  

With this change am I reading it right that we basically support the union of all option names, but the set of `options` passed in will be restricted/correct for the give Movable subclass that calls this function?

benchristel: `undefined` values on `options` are ignored, so it's fine if a particular movable doesn't have handlers for certain events. The runtime behavior is the same as before.

PR #960, opened 2024-02-02
Tests for Movable

jeremywiebe:
## Summary:

I had a few minutes yesterday so I started adding some tests around `Movable` as I know we'll likely be touching it for the `interactive-graph` work we're doing.

Issue: "none"

## Test plan:

`yarn test`

### DIFF
--- a/packages/perseus/src/interactive2/__tests__/movable.test.ts
+++ b/packages/perseus/src/interactive2/__tests__/movable.test.ts
@@ -69,7 +69,7 @@ describe("Movable", () => {
         expect(movable.cloneState()).not.toBe(movable.state);
     });
 
-    test("modify() resets movable to initial state", () => {
+    test("modify() resets movable to initial state when no params given", () => {
         const movable = new Movable(
             new Graphie(document.createElement("div")),
             {},
@@ -78,6 +78,10 @@ describe("Movable", () => {
         const state = movable.cloneState();
         movable.state.added = false;
 
+        // It seems that all usages of this function in our production code
+        // pass the original state + mutations to modify(). However, it _does_
+        // work to pass no arguments so we'll keep this Characterization test
+        // here for now.
         movable.modify();
 
         expect(movable.cloneState()).toEqual(state);
--- a/packages/perseus/src/interactive2/__tests__/movable.test.ts
+++ b/packages/perseus/src/interactive2/__tests__/movable.test.ts
@@ -164,12 +164,6 @@ describe("Movable", () => {
 
     it("grab() does not fire onClick when mouse up if not hovering", () => {
         const graphie = new Graphie(document.createElement("div"));
-        graphie.init({
-            range: [
-                [-10, 10],
-                [-10, 10],
-            ],
-        });
         const onClick = jest.fn();
         const movable = new Movable(graphie, {onClick});
 
--- a/packages/perseus/src/interactive2/__tests__/movable.test.ts
+++ b/packages/perseus/src/interactive2/__tests__/movable.test.ts
@@ -118,12 +118,13 @@ describe("Movable", () => {
                 [-10, 10],
                 [-10, 10],
             ],
+            scale: [10, 10],
         });
         const onMove = jest.fn();
         const movable = new Movable(graphie, {onMove});
 
         movable.grab([0, 0]);
-        fireEvent.mouseMove(document, {clientX: 480, clientY: 320});
+        fireEvent.mouseMove(document, {clientX: 120, clientY: 80});
 
         expect(onMove).toHaveBeenCalledWith([2, 2], [0, 0]);
     });
--- a/packages/perseus/src/interactive2/__tests__/movable.test.ts
+++ b/packages/perseus/src/interactive2/__tests__/movable.test.ts
@@ -161,16 +161,4 @@ describe("Movable", () => {
 
         expect(onClick).toHaveBeenCalledWith([0, 0], [0, 0]);
     });
-
-    it("grab() does not fire onClick when mouse up if not hovering", () => {
-        const graphie = new Graphie(document.createElement("div"));
-        const onClick = jest.fn();
-        const movable = new Movable(graphie, {onClick});
-
-        movable.grab([0, 0]);
-        movable.state.isHovering = false;
-        fireEvent.mouseUp(document);
-
-        expect(onClick).not.toHaveBeenCalled();
-    });
 });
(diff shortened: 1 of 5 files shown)


packages/perseus/src/interactive2/__tests__/movable.test.ts
```diff
@@ -0,0 +1,182 @@
...
+    });
+
+    it("grab() does not fire onClick when mouse up if not hovering", () => {
+        const graphie = new Graphie(document.createElement("div"));
+        graphie.init({
+            range: [
+                [-10, 10],
+                [-10, 10],
+            ],
+        });
```
benchristel, line 172: Does the graphie need to be initialized for this to work?

packages/perseus/src/interactive2/__tests__/movable.test.ts
```diff
@@ -0,0 +1,182 @@
...
+        const onClick = jest.fn();
+        const movable = new Movable(graphie, {onClick});
+
+        movable.grab([0, 0]);
+        movable.state.isHovering = false;
```
benchristel, line 177: It's a bit awkward that we're reaching into the state like this. How would isHovering actually be set?

packages/perseus/src/interactive2/__tests__/movable.test.ts
```diff
@@ -0,0 +1,182 @@
...
+            range: [
+                [-10, 10],
+                [-10, 10],
+            ],
+        });
+        const onMove = jest.fn();
+        const movable = new Movable(graphie, {onMove});
+
+        movable.grab([0, 0]);
+        fireEvent.mouseMove(document, {clientX: 480, clientY: 320});
```
benchristel, line 126: I assume these coordinates are derived from the scale of the Graphie somehow? It might be nice to set the scale explicitly (to decouple this test from whatever defaults Graphie sets) or otherwise document where these numbers come from.

packages/perseus/src/interactive2/__tests__/movable.test.ts
```diff
@@ -0,0 +1,182 @@
...
+    it("modify() retains original ID", () => {
+        const movable = new Movable(
+            new Graphie(document.createElement("div")),
+            {},
+        );
+        const id = movable.state.id;
+
+        movable.modify({added: true});
+
+        expect(movable.state.id).toStrictEqual(id);
```
benchristel, line 95: Is `toStrictEqual` needed here? `id` is just a string, so my understanding is that `toStrictEqual` is equivalent to `toBe` in this case.

packages/perseus/src/interactive2/__tests__/movable.test.ts
```diff
@@ -0,0 +1,182 @@
...
+            new Graphie(document.createElement("div")),
+            {},
+        );
+
+        const state = movable.cloneState();
+        movable.state.added = false;
+
+        movable.modify();
+
+        expect(movable.cloneState()).toStrictEqual(state);
```
benchristel, line 83: Is `toStrictEqual` needed here? Would `toEqual` be correct?

packages/perseus/src/interactive2/__tests__/movable.test.ts
```diff
@@ -0,0 +1,182 @@
...
+    it("cloneState() should copy state", () => {
+        const movable = new Movable(
+            new Graphie(document.createElement("div")),
+            {},
+        );
+
+        expect(movable.cloneState()).not.toBe(movable.state);
+    });
+
+    it("modify() resets movable to initial state", () => {
```
benchristel, line 72: I'm not sure this test really represents desired behavior. All the calls to `modify` I could find were passing the current state, with modifications as the argument. I think it's pretty strange that `modify` can be called without arguments, and it's probably not the intended use case.

Still, I think we might want to keep this test to document the legacy behavior. Can you add a comment noting that this is a [characterization test](https://en.wikipedia.org/wiki/Characterization_test)?

packages/perseus/src/interactive2/__tests__/movable.test.ts
```diff
@@ -0,0 +1,182 @@
...
+        expect(typeof movable.isHovering).toBe("function");
+        expect(typeof movable.cursor).toBe("function");
+        expect(typeof movable.added).toBe("function");
+        expect(typeof movable.isHovering).toBe("function");
+        expect(typeof movable.isMouseOver).toBe("function");
+        expect(typeof movable.isDragging).toBe("function");
+        expect(typeof movable.mouseTarget).toBe("function");
+    });
+
+    it("cloneState()", () => {
```
benchristel, line 36: Jest provides `test` as an alias for `it`. Using `test` might make some of these descriptions read a bit nicer.
```suggestion
    test("cloneState()", () => {
```

benchristel (approved):
I'm working on Movable as well, and also added some tests. Let's land this and then I'll rebase my changes.

packages/perseus/src/interactive2/__tests__/movable.test.ts
jeremywiebe: Good catch.  No, it passes without calling `init()`.

packages/perseus/src/interactive2/__tests__/movable.test.ts
jeremywiebe: In reviewing the code a bit more, it looks like `isHovering` might never be `false` in this case. I'm going to remove this test for now as I agree about digging into the internals being awkward.

packages/perseus/src/interactive2/__tests__/movable.test.ts
jeremywiebe: Agreed. Updated.

packages/perseus/src/interactive2/__tests__/movable.test.ts
jeremywiebe: Not sure how I decided on `.toStrictEqual`!  Fixed.

packages/perseus/src/interactive2/__tests__/movable.test.ts
jeremywiebe: :TIL: about the differences.  `toStrictEqual` verifies types as well, but I think `.toEqual()` is fine.

packages/perseus/src/interactive2/__tests__/movable.test.ts
jeremywiebe: True.  TBH, `it()` and `test()` dissolve into the background when I look at test files.  I'm open to changing this for others though.

PR #961, opened 2024-02-02
Fix Perseus so we register the transformer replacement

jeremywiebe:
## Summary:

When we pushed out the removal of the `transformer` widget, we missed uncommenting the replacement line. This fixes it and adds a test to the Perseus init() function to ensure we do that.

Issue: "none"

## Test plan:

`yarn test`

### DIFF
--- a/packages/perseus/src/types.ts
+++ b/packages/perseus/src/types.ts
@@ -411,9 +411,13 @@ export type WidgetExports<
 > = Readonly<{
     name: string;
     displayName: string;
+
+    // Widgets should provide _one_ of these two properties only!
     // TODO: figure out how to ensure that the component returned supports
     // the `validate` method.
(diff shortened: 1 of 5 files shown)


packages/perseus/src/types.ts
```diff
@@ -411,9 +411,13 @@ export type WidgetExports<
 > = Readonly<{
     name: string;
     displayName: string;
+
+    // Widgets should provide _one_ of these two properties only!
     // TODO: figure out how to ensure that the component returned supports
     // the `validate` method.
     getWidget?: () => T;
+    widget: T;
+
```
jeremywiebe, line 420: Unrelated, but it's just a code move to put these two, related fields together.

handeyeco (approved):
LGTM!

PR #966, opened 2024-02-06
Update "t" SVG for MathInput keypad

handeyeco:
## Summary:
We were using "s" instead of "t" for some reason

Issue: LC-1719


jeremywiebe (approved):
Nice

PR #970, opened 2024-02-07
Add Controller and Cursor types for MathQuill

nishasy:
## Summary:
Add types for MathQuill cursor and controller in
`mathquill-types.ts`.

We couldn't get these typed in mathquill.d.ts inside
the mathquill repo without it turning into a dumpster
fire of type errors. In Perseus, they're just using
type `any` since they're not defined in the mathquill API.

Since we don't want them to use type `any`, we can just
define them here in Perseus.

This PR includes
- Cursor and Controller types in mathquill-types.ts
- Updates to other files to match the new types

Issue: https://khanacademy.atlassian.net/browse/LC-1659

## Test plan:
`yarn jest`

Storybook
- Go to http://localhost:6006/?path=/docs/math-input-components-v2-keypad-with-mathquill--docs
- Play around with the buttons and confirm that the input looks correct
- Confirm there are no console errors

### DIFF
--- a/packages/math-input/src/components/input/mathquill-types.ts
+++ b/packages/math-input/src/components/input/mathquill-types.ts
@@ -22,13 +22,6 @@ export enum MathFieldActionType {
     MQ_END = 0,
 }
 
-/**
- * The MathQuill MathField Cursor
- * it's not part of the public API for MathQuill,
- * we reach into the internals to get it
- */
-export type MathFieldCursor = any;
-
 export type MathFieldUpdaterCallback = (
     mathField: MathFieldInterface,
     key: Key,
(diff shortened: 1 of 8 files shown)


packages/math-input/src/components/input/mathquill-types.ts
```diff
@@ -30,3 +33,60 @@ export type MathFieldUpdaterCallback = (
...
+
+/**
+ * The MathQuill API (see mathuill.d.ts) does not include types
+ * for cursor() and controller(), and adding these types there
+ * in the MathQuill repo causes unexpected conflicts with other types.
+ *
+ * We don't want to use the cursor and controller default type `any`
+ * so we declare the types here.
+ *
+ * Note: This is different from the MathFieldCursor defined above.
```
nedredmond, line 45: It should actually be replacing MathFieldCursor, an alias for "any" used to access Cursor. Can you delete that?

nedredmond (approved):
Thanks!!!

PR #971, opened 2024-02-07
Only include 'dist/' in NPM packages

benchristel:
## Summary:
Source files (`.ts`) were causing typechecking problems in webapp. We're
not sure exactly why they just started happening recently, but it seems
related to our improved typing around Graphie and Raphael.

Issue: none

## Test plan:

Upgrade Perseus packages to the new versions in webapp. There should be no type
errors when you run `yarn tsc -w` in services/static.

### DIFF
--- a/.changeset/short-otters-confess.md
+++ b/.changeset/short-otters-confess.md
@@ -1,0 +1,14 @@
+---
+"@khanacademy/kas": patch
+"@khanacademy/kmath": patch
+"@khanacademy/math-input": patch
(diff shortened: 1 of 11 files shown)


handeyeco:
For the record: https://khanacademy.slack.com/archives/CDHPTMSTB/p1707321428502869

handeyeco (approved):

jeremywiebe (approved):
Ditto what Matthew said. 😂

PR #975, opened 2024-02-08
Graphie: add tests for addMouseLayer()

jeremywiebe:
## Summary:

This PR introduces a set of tests that cover some of the `addMouseLayer()` function added to Graphie by `interactive.ts`.  I've also been able to add a bit more typing to `Graphie` which allows us to eliminate a few more `@ts-expect-error`s.

I could not get the `jquery.mobile.vmouse.js` events to fire in tests (`vmousedown`, `vmouseup`, etc), so the mouse handlers are not (yet) tested. 

Issue: LC-1662

## Test plan:

`yarn test`

### DIFF
--- a/packages/perseus/src/util/interactive.ts
+++ b/packages/perseus/src/util/interactive.ts
@@ -86,6 +86,7 @@ _.extend(GraphUtils.Graphie.prototype, {
         onMouseDown?: MouseHandler;
         onMouseOver?: MouseHandler;
         onMouseOut?: MouseHandler;
+        onMouseUp?: MouseHandler;
         allowScratchpad?: boolean;
         setDrawingAreaAvailable?: (available: boolean) => void;
     }) {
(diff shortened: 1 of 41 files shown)


packages/perseus/src/util/interactive.ts
```diff
@@ -99,11 +100,7 @@ _.extend(GraphUtils.Graphie.prototype, {
         );
 
         const mouselayerZIndex = 2;
-        graph.mouselayer = Raphael(
```
jeremywiebe, line 102: Fun fact, we create `graph.raphael` by passing in the `this.el`, so as far as I could tell it's always the same as `graph.raphael.canvas.parentNode`. This reduces how much mocking we need to do quite a bit. 

I also verified that Storybook worked for all of our graph widgets.

benchristel (approved):
Thanks for the thorough tests! It feels a bit like we're asserting on Graphie internals, but I'm assuming that these tests are temporary. Once we move the dynamically-defined methods over to the Graphie class, we can revisit/remove these tests?

nishasy (commented):
I don't think I'm qualified to be the final approver, but this looks good to me!

jeremywiebe:
> Thanks for the thorough tests! It feels a bit like we're asserting on Graphie internals, but I'm assuming that these tests are temporary. Once we move the dynamically-defined methods over to the Graphie class, we can revisit/remove these tests?

Yes. That makes sense to me. I was trying to get some sort of coverage that'd give us some confidence when we refactor.

PR #976, opened 2024-02-08
Ensure we do yarn install before Khan/changeset-per-package

jeremywiebe:
## Summary:

I noticed in a different [PR run](https://github.com/Khan/perseus/actions/runs/7823290861/job/21343971545?pr=975) that the `@changesets/cli` was being installed by `Khan/changeset-per-package`. This action checks if changesets is installed by trying to run it. This was always failing in the `node-ci` action because we didn't run `yarn install` initially! 

image

Issue: "none"

## Test plan:

Watch the next run and it shouldn't complain about "warning "@changesets/cli" is already in "devDependencies". Please remove existing entry first before adding it to "dependencies"."

### DIFF
--- a/.changeset/polite-guests-agree.md
+++ b/.changeset/polite-guests-agree.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Adds comment that I'll back out
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -23,6 +23,12 @@ jobs:
               with:
                   fetch-depth: 0
 
+            - name: Install & cache node_modules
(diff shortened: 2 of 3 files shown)


benchristel (commented):
Since none of the packages have changed in this PR, the changeset check exits early, so I can't verify (from the CI run on this PR) that this code is working.

I guess we can wait for the next PR. Alternatively, we could temporarily change one of the packages in this PR so we can watch the changeset job do its thing.

jeremywiebe:
@benchristel thanks for pointing out that this PR wasn't really tested yet.  After adding a code change and a changeset, I see my change still doesn't prevent the `yarn add` command I was trying to avoid.  

I'm going to abandon this PR for now.  I've messaged @nedredmond about this in case he has ideas.

PR #978, opened 2024-02-08
Stop keypad button clicks from removing focus from the input

nishasy:
## Summary:

Currently in the math input component, if a user has focus on the input field and then clicks
a keypad button, it shifts focus to the button before re-focusing on the input field. This
creates a jarring effect with the focus outline, and it causes the screen reader to read out the
full input value on every single click without reading the button key at all. I think it also sometimes
causes the screen reader to falsely start reading "popover closed" before getting interrupted.

To fix this, I'm adding `e.preventDefault()` on mouse down to the keypad buttons. This will
stop them from stealing focus from the input.

In this PR:
- Updated WB Clickable version in Perseus to support onMouseDown
- Added `e.preventDefault()` to the keypad buttons to stop them from stealing focus from the input

Issue: https://khanacademy.atlassian.net/browse/LC-1370

## Test plan:

Storybook
- Go to http://localhost:6006/iframe.html?id=perseus-components-math-input--default-with-basic-button-set&viewMode=story
- Open the math keypad
- Click on the keypad buttons with the cursor
- Confirm that focus didn't shift back and forth between the button and the input field
- Confirm that the button that was pressed is read out by the screen reader

### Before

Demo:
1. using keyboard to type
2. then using mouse to click keypad buttons

https://github.com/Khan/perseus/assets/13231763/c3313e5a-08fc-45dc-b942-2cd63f5a7b48

### After

Demo:
1. using keyboard to type
2. then using mouse to click keypad buttons

https://github.com/Khan/perseus/assets/13231763/4d918692-ebaa-4ca1-8ea0-231f17539f02


packages/perseus/src/components/math-input.tsx
```diff
@@ -75,6 +75,11 @@ type State = {
     keypadOpen: boolean;
     cursorContext: typeof CursorContext[keyof typeof CursorContext];
     openedWithEventType?: string;
+    // [count, label]
+    // count is just used to trigger a re-render when the label changes,
+    // even if the label is the same as the previous value.
+    // label is the last character typed.
+    lastChar: Array<number | string>;
```
benchristel, line 82: Why do we want to be able to explicitly trigger a re-render? If the intention is to get screenreaders to announce aria-live contents, I'm not sure this will always work. I think React will diff the new label string with its memory of the previous render, see that nothing changed, and skip updating the DOM.

If you're finding that this does actually work with a screenreader, I'd like to understand why it works.

If we decide to keep both the count and the label, could they both be top-level state properties? I.e.

```suggestion
    // count is just used to trigger a re-render
    count: number;
    // used in a role="alert" element to provide feedback as the user types
    lastCharacterTyped: string;
```

I feel like that's simpler, and more likely to be what future readers of this code will expect to see. (It's closer to what I expected, anyway).

nishasy: Ah, you're absolutely right. I was taking inspiration from this [screen reader alert component](https://github.com/Khan/webapp/blob/42a81b6872bc37c002af67107b903fbdd60bad07/services/static/javascript/python-coding/screen-reader-alert.tsx), but I totally missed the `key` prop. I'll update it now!

packages/perseus/src/components/math-input.tsx
```diff
@@ -236,10 +243,22 @@ class MathInput extends React.Component<Props, State> {
     // input is still focused
     blur: () => void = () => this.setState({focused: false});
 
-    handleKeypadPress: (key: Keys, e: any) => void = (key, e) => {
+    handleKeypadPress: (key: Keys, label: string, e: any) => void = (
```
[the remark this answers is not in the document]
nishasy: I believe I still need that so that the focus goes back to the input on the first keypad click. Otherwise it remains on the "open keypad" button.

nishasy:
Note: I don't think I can land this until Feb 22nd because of the level access manual testing period. I think landing this would block people from being able to deploy perseus.

nishasy:
I think I'm going to go back and simplify this PR so it doesn't include the key press read-aloud, since it was out of scope and I'm not even sure it's good/accessible anyway

nedredmond (approved):
Finally! Let's merge it!

PR #979, opened 2024-02-08
Graphie: Add tests for getMousePx and getMouseCoord

jeremywiebe:
## Summary:

This PR adds unit tests for Graphie's `getMousePx` and `getMouseCoord`. The pair of these translate an event from a mouse click to graph-relative coordinates in pixel coords (`getMousePx`) and graph coords (`getMouseCoord`).

Issue: LC-1662

## Test plan:

`yarn test`

### DIFF
--- a/packages/perseus/src/util/graphie.test.ts
+++ b/packages/perseus/src/util/graphie.test.ts
@@ -8,10 +8,6 @@ import GraphUtils, {normalizeRange} from "./graphie";
 
 import type {Graphie} from "./graphie";
 
-// Yay for side-effect imports!
-// eslint-disable-next-line import/no-unassigned-import
-import "./interactive";
-
 function createMockRaphaelElement(name = "mockRaphaelElement") {
     return {
         constructor: {prototype: Raphael.el},
--- a/packages/perseus/src/util/graphie.test.ts
+++ b/packages/perseus/src/util/graphie.test.ts
@@ -1,5 +1,6 @@
 // eslint-disable-next-line import/no-extraneous-dependencies
 import $ from "jquery";
+// eslint-disable-next-line import/no-extraneous-dependencies
 import Raphael from "raphael";
 
 import {testDependencies} from "../../../../testing/test-dependencies";
--- a/packages/perseus/src/util/graphie.test.ts
+++ b/packages/perseus/src/util/graphie.test.ts
@@ -1,4 +1,5 @@
 // eslint-disable-next-line import/no-extraneous-dependencies
+import $ from "jquery";
 import Raphael from "raphael";
 
 import {testDependencies} from "../../../../testing/test-dependencies";
@@ -1319,6 +1320,40 @@ describe("Graphie drawing tools", () => {
 
             expect(onSetDrawingAreaAvailable).toHaveBeenCalledWith(false);
         });
+    });
+
+    describe("getMousePx", () => {
+        it.each([
+            [{left: 0, top: 0}, {pageX: 10, pageY: 10}, [10, 10]],
+            [{top: 10, left: 20}, {pageX: 40, pageY: 40}, [20, 30]],
+        ])(
+            "should return the mouse position in pixel coordinates relative to graph",
+            (offset, mouseEvent, expectedMousePx) => {
+                const graphie = createAndInitGraphie();
+                jest.spyOn($.fn, "offset").mockReturnValue(offset);
+
+                const mousePx = graphie.getMousePx(mouseEvent);
+
+                expect(mousePx).toEqual(expectedMousePx);
+            },
+        );
+    });
+
+    describe("getMouseCoord", () => {
+        it.each([
+            [{left: 0, top: 0}, {pageX: 10, pageY: 10}, [2, 8]],
+            [{top: 10, left: 20}, {pageX: 40, pageY: 40}, [4, 4]],
+        ])(
+            "should return mosue position in graph coordinates relative to the graph",
+            (offset, mouseEvent, expectedMousePx) => {
+                const graphie = createAndInitGraphie();
+                jest.spyOn($.fn, "offset").mockReturnValue(offset);
+
+                const mousePx = graphie.getMouseCoord(mouseEvent);
+
+                expect(mousePx).toEqual(expectedMousePx);
+            },
+        );
     });
 });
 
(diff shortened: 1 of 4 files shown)


packages/perseus/src/util/graphie.ts
```diff
@@ -1445,6 +1440,16 @@ export class Graphie {
...
+
+    // This is a stub that's overridden in interactive.ts
+    getMouseCoord(event: Readonly<{pageX?: number; pageY?: number}>): Coord {
+        throw new Error("getMouseCoord is a stub, and is not implemented");
+    }
+
+    // This is a stub that's overridden in interactive.ts
+    getMousePx(event: Readonly<{pageX?: number; pageY?: number}>) {
+        throw new Error("getMousePx is a stub, and is not implemented");
+    }
```
jeremywiebe, line 1452: I just moved these down below the two functions that are related to `addMouseLayer` so they are all together.

packages/perseus/src/util/interactive.ts
```diff
@@ -196,14 +196,11 @@ _.extend(GraphUtils.Graphie.prototype, {
      * Get mouse coordinates in pixels
      */
     getMousePx: function (event: Readonly<{pageX?: number; pageY?: number}>) {
-        const graphie = this;
+        // @ts-expect-error - TS2532 - Object is possibly 'undefined'.
```
nedredmond, line 199: Which object is possible undefined, `el`? Can we add a check to narrow to a defined type?

packages/perseus/src/util/graphie.test.ts
```diff
@@ -1320,6 +1321,40 @@ describe("Graphie drawing tools", () => {
...
+        );
+    });
+
+    describe("getMouseCoord", () => {
+        it.each([
+            [{left: 0, top: 0}, {pageX: 10, pageY: 10}, [2, 8]],
+            [{top: 10, left: 20}, {pageX: 40, pageY: 40}, [4, 4]],
+        ])(
+            "should return mosue position in graph coordinates relative to the graph",
+            (offset, mouseEvent, expectedMousePx) => {
```
benchristel, line 1348: `expectedMousePx` should be named something like `expectedCoord`, no?

packages/perseus/src/util/graphie.test.ts
```diff
@@ -1320,6 +1321,40 @@ describe("Graphie drawing tools", () => {
             expect(onSetDrawingAreaAvailable).toHaveBeenCalledWith(false);
         });
     });
+
+    describe("getMousePx", () => {
+        it.each([
+            [{left: 0, top: 0}, {pageX: 10, pageY: 10}, [10, 10]],
+            [{top: 10, left: 20}, {pageX: 40, pageY: 40}, [20, 30]],
```
benchristel, line 1328: Is there a reason `left` and `top` are in the opposite order here? (same thing in the `it.each` on line 1343.)

packages/perseus/src/util/interactive.ts
```diff
@@ -196,14 +196,11 @@ _.extend(GraphUtils.Graphie.prototype, {
...
+        const mouseX = event.pageX - $(this.el).offset().left;
 
-        const mouseX =
-            // @ts-expect-error - TS2532 - Object is possibly 'undefined'.
-            event.pageX - $(graphie.raphael.canvas.parentNode).offset().left;
-        const mouseY =
-            // @ts-expect-error - TS2532 - Object is possibly 'undefined'.
-            event.pageY - $(graphie.raphael.canvas.parentNode).offset().top;
+        // @ts-expect-error - TS2532 - Object is possibly 'undefined'.
```
benchristel, line 202: I suspect these TS errors have changed. Are they still `Object is possibly 'undefined'`?

packages/perseus/src/util/graphie.test.ts
```diff
@@ -1320,6 +1321,40 @@ describe("Graphie drawing tools", () => {
...
+        ])(
+            "should return the mouse position in pixel coordinates relative to graph",
+            (offset, mouseEvent, expectedMousePx) => {
```
benchristel, line 1331: `offset` is the screen coordinates of the top-left corner of the graph, right? Could its name reflect that? Something like `graphPosition` might work.

packages/perseus/src/util/graphie.test.ts
```diff
@@ -1320,6 +1321,40 @@ describe("Graphie drawing tools", () => {
...
+
+                expect(mousePx).toEqual(expectedMousePx);
+            },
+        );
+    });
+
+    describe("getMouseCoord", () => {
+        it.each([
+            [{left: 0, top: 0}, {pageX: 10, pageY: 10}, [2, 8]],
+            [{top: 10, left: 20}, {pageX: 40, pageY: 40}, [4, 4]],
```
benchristel, line 1345: I'm having a hard time figuring out how the input values relate to the expected results. Could we explicitly set the scale and size of the graph, to make it a bit clearer?

packages/perseus/src/util/graphie.test.ts
```diff
@@ -1,4 +1,5 @@
 // eslint-disable-next-line import/no-extraneous-dependencies
+import $ from "jquery";
```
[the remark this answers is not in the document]
jeremywiebe: Yep.  It's used in our production code and I needed to mock it for tests. I think we could probably move to using `getBoundingClientRect`, but I wanted to avoid production code changes as much as possible in this PR.

packages/perseus/src/util/interactive.ts
jeremywiebe: Yes.  There's two errors that were being hidden. I've broken out the lines so that it's possible to see the two suppressions.  I'm not planing to resolve them though as this is the way the original code is (I just changed it from reaching into Raphael `graphie.raphael.canvas.parentNode` to `this.el`).

packages/perseus/src/util/graphie.test.ts
```diff
@@ -1320,6 +1322,50 @@ describe("Graphie drawing tools", () => {
...
+
+                expect(mousePx).toEqual(expectedPixelCoord);
+            },
+        );
+    });
+
+    describe("getMouseCoord", () => {
+        it.each([
+            [{left: 0, top: 0}, {pageX: 10, pageY: 10}, [2, 8]],
+            [{left: 20, top: 10}, {pageX: 30, pageY: 20}, [2, 8]],
```
benchristel, line 1346: I just realized... if one of these tests fails, is there any way to tell which one is failing? They have the same description, and the line numbers won't help either...

benchristel (approved):
Looks good! Thanks. I left one more question inline.

packages/perseus/src/util/graphie.test.ts
jeremywiebe: Um, yeah. Good point. Updated!

PR #980, opened 2024-02-08
Create a helper function for generating correctly typed Perseus test items

Myranae:
## Summary:
There are many places in webapp where custom types are created for Perseus items, frequently related to testing. This change adds a helper function that creates a generic Perseus test item that can have specific sections overridden with custom input. It will ensure tests are properly typed and help us know where updates need to be made if any changes happen to the types in Perseus.

Issue: LC-1658

## Test plan:
- Confirm all checks pass
- Test the new function in webapp with the npm snapshot
- In webapp, confirm related tests pass and related stories still work


packages/perseus/src/__tests__/test-utils.test.ts
```diff
@@ -0,0 +1,33 @@
...
+    basicObject,
+    expected1,
+    questionCustomInfo,
+    expected2,
+    answerAreaCustomInfo,
+    hintsCustomInfo,
+    expected3,
+} from "./test-items/test-utils-items";
+
+describe("generateTestPerseusItem", () => {
```
Myranae, line 15: Would love suggestions on ways to make these tests more useful or additional tests that could be created for this change. :)

packages/perseus/src/util/test-utils.ts
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Generate a Perseus item object for testing.
+ *
+ * In order to better type Perseus objects used in testing, this function
+ * uses a basic Perseus object and updates it with custom values as needed.
+ *
+ * @param {Partial<PerseusItem>} customFields
+ * @returns PerseusItem
+ */
```
Myranae, line 37: Is this useful here? I saw other functions had this information added, so wanted to include it. Seems handy to have a description.

packages/perseus/src/__tests__/test-items/test-utils-items.ts
```diff
@@ -0,0 +1,219 @@
...
+                        randomize: true,
+                    },
+                    type: "radio",
+                },
+            },
+        },
+    ],
+};
+
+export const expected3: PerseusItem = {
```
handeyeco, line 164: Could we name these to be a little clearer? Maybe like `expectedExtendingHintsCase` and `expectedExtendingQuestionCase` or something like that?

handeyeco (changes requested):
This looks good, very thorough tests.

Looking at `packages/perseus/src/util` I see that we have some tests next to the helpers. I think we should move these files a little bit:

```
packages/perseus/src/util
  - test-utils.ts
  - test-utils.test.ts
  - test-utils.testdata.ts
```

I can see why you put those there though, it seems like we have 3+ different ways of organizing things 😅 However I think tests are more discoverable when they're next to the files they test (esp when it comes to unit tests).

packages/perseus/src/__tests__/test-items/test-utils-items.ts
Myranae: 100% On it :) Thanks for the suggestion!

Myranae: Updated the names and re-requested review

Myranae: Didn't see your comment about structure until now. Just refactored the test files.

packages/perseus/src/util/test-utils.test.ts
```diff
@@ -0,0 +1,36 @@
+import {describe, it, expect} from "@jest/globals";
```
handeyeco, line 1: I don't think these need to be explicitly imported anymore and should be removed.

handeyeco (approved):
Thanks!

handeyeco:
Related to: https://github.com/Khan/webapp/pull/19775

PR #981, opened 2024-02-08
Add a Vite-based dev UI for interactive graphs

benchristel:
We're feeling the need for a faster, more reliable feedback loop than Storybook
can provide.

This commit introduces a Vite-based dev UI that (currently) renders a question
with an angle graph. We can add more graphs in future commits. I just wanted
to get a walking skeleton up for review.

Issue: https://khanacademy.atlassian.net/browse/LC-1741

## Test plan:

```
cd dev
yarn dev
open http://localhost:5173
```

You should see an interactive angle graph in the web browser.

Screen Shot 2024-02-08 at 2 05 49 PM


dev/index.jsx
```diff
@@ -0,0 +1,29 @@
+import * as React from "react"
+import {render} from "react-dom"
+import {Renderer} from "@khanacademy/perseus/src";
+import {
+    angleQuestion
+} from "@khanacademy/perseus/src/widgets/__testdata__/interactive-graph.testdata";
```
jeremywiebe, line 6: 🤔  I wonder if we should just use path-relative imports. 🤔  

I guess Vite properly handle these package references, which makes me wonder if we should just be importing like this `import {Renderer} from "@khanacademy/perseus"`.  🤷‍♂️   Or if this works, maybe this is where we start. :)

dev/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "publishConfig": {
+        "access": "public"
+    },
+    "repository": {
+        "type": "git",
+        "url": "https://github.com/Khan/perseus.git",
+        "directory": "dev"
+    },
+    "scripts": {
+        "dev": "vite"
```
jeremywiebe, line 16: I wonder if it'd be useful to put a `dev` script in the root `package.json` also so that we can `yarn dev` at the root.  it could just delegate to this `dev` script, but it'd remove the need to `cd dev` before `yarn dev`.

vendor/hubble/hubble.js
jeremywiebe, line 1: We intentionally didn't touch this file when we brought it in.  🤔 

But it seems like a pretty simple/small file, these refactors seem to mostly be removing the stuff whatever build system Hubble used.

No action necessary, just musing.

jeremywiebe (approved):
Nice.  This is a great start. Thanks.

dev/index.jsx
benchristel: I think relative imports are the way to go. I'd have a bit more confidence that we're actually getting the right code that way 😅 .

vendor/hubble/hubble.js
benchristel: Yeah, manually changing vendored dependencies is not great. But:

- Hubble is a small, simple library (and I've written very similar things myself) so I feel fine taking on responsibility for maintaining this code.
- Hubble is unmaintained (last commit was [10 years ago](https://github.com/joelburget/hubble), and there's not even an NPM package for it) so it's extremely unlikely that we'd ever want to merge in changes from upstream.

benchristel: If we really want to do this the "right" way, I suppose we could fork Hubble and publish our own package.

jeremywiebe: I think the "right" way is overkill.  This library has low usage in Perseus so let's just roll with what you've done.

PR #982, opened 2024-02-08
Inject generated and localized MathSpeak into MathQuill

nedredmond:
Swaps out MathQuill's hardcoded English MathSpeak for localized MathSpeak generated by MathJax/SRE

Issue: LC-1710

Testing:
Open Storybook and check generated labels. They should look the same!
If you can get Storybook into another locale supported by SRE, you'll see translated MathSpeak.
If you swap out the function in `createMathField` for something that returns "poot", all of the MathSpeak will be replaced with "poot".

### DIFF
--- a/.changeset/rich-glasses-type.md
+++ b/.changeset/rich-glasses-type.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/math-input": patch
+"@khanacademy/perseus": patch
+---
+
+Inject localized MathSpeak into MathQuill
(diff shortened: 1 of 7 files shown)


nishasy (approved):
Very cool! Are all the lint errors expected?

nedredmond:
@nishasy I accidentally accepted your MathQuill commit hash which rolled it back to a previous version-- whoops! Fixed.

PR #985, opened 2024-02-09
Bug fix: don't run methods on Editor widgets that haven't mounted

nedredmond:
The edit handler was immediately being called for some syntax corrections on values it was initialized with. This triggered Perseus' `onChange` callback, which was not yet prepared to handle them, causing the editor to crash.

Refactored the refs in widget-editor to not be string refs, then checked that current exists before calling methods

Issue: LC-1743

Testing:
[Follow @handeyeco's instructions for recreating the issue in a story from this comment (linked).](https://khanacademy.atlassian.net/browse/LC-1743?focusedCommentId=297416) You will notice that there is no crash, and the answer appears in the expression input as expected.

### DIFF
--- a/packages/perseus-editor/src/__stories__/editor-page.stories.tsx
+++ b/packages/perseus-editor/src/__stories__/editor-page.stories.tsx
@@ -26,9 +26,38 @@ export const Demo = (): React.ReactElement => {
     const [answerArea, setAnswerArea] = React.useState<
         PerseusAnswerArea | undefined | null
     >();
-    const [question, setQuestion] = React.useState<
-        PerseusRenderer | undefined
-    >();
+    const [question, setQuestion] = React.useState<PerseusRenderer | undefined>(
+        {
+            content: "[[☃ expression 1]]",
+            images: {},
+            widgets: {
+                "expression 1": {
+                    type: "expression",
+                    alignment: "default",
+                    static: false,
+                    graded: true,
+                    options: {
+                        answerForms: [
+                            {
+                                value: "++++",
+                                form: true,
+                                simplify: false,
+                                considered: "correct",
+                                key: "0",
+                            },
+                        ],
+                        buttonSets: ["basic", "prealgebra"],
+                        functions: ["f", "g", "h"],
+                        times: true,
+                    },
+                    version: {
+                        major: 1,
+                        minor: 0,
+                    },
+                },
+            },
+        },
+    );
     const [hints, setHints] = React.useState<ReadonlyArray<Hint> | undefined>();
 
     return (
--- a/packages/perseus-editor/src/__stories__/editor-page.stories.tsx
+++ b/packages/perseus-editor/src/__stories__/editor-page.stories.tsx
@@ -26,38 +26,9 @@ export const Demo = (): React.ReactElement => {
     const [answerArea, setAnswerArea] = React.useState<
         PerseusAnswerArea | undefined | null
     >();
-    const [question, setQuestion] = React.useState<PerseusRenderer | undefined>(
-        {
-            content: "[[☃ expression 1]]",
-            images: {},
-            widgets: {
-                "expression 1": {
-                    type: "expression",
-                    alignment: "default",
-                    static: false,
-                    graded: true,
-                    options: {
-                        answerForms: [
-                            {
-                                value: "++++",
-                                form: true,
-                                simplify: false,
-                                considered: "correct",
-                                key: "0",
-                            },
-                        ],
-                        buttonSets: ["basic", "prealgebra"],
-                        functions: ["f", "g", "h"],
-                        times: true,
-                    },
-                    version: {
-                        major: 1,
-                        minor: 0,
-                    },
-                },
-            },
-        },
-    );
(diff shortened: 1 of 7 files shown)


jeremywiebe:
This seems to introducing a new concept in the larger sense. As far as I can tell, our editors do not make changes to the Perseus content on load. Thinking about other types of editors, it always feels jarring for a document to be in a dirty state immediately after loading it. 

So my question is: do we have a choice in re-writing this tex on load? Can we be more graceful? 

The second concern I have is that we're driving this fix deep into the math input itself. I wonder if we could localize the fix closer to the editor as that's the part that seems sensitive to the way the new math input is doing things.

nedredmond:
It's a feature of Desmos' MathJax to correct the syntax.

nedredmond:
> The second concern I have is that we're driving this fix deep into the math input itself. I wonder if we could localize the fix closer to the editor as that's the part that seems sensitive to the way the new math input is doing things. 

The onChange callback is prop-drilled from Perseus to MathInput.

nedredmond:
> This seems to introducing a new concept in the larger sense. As far as I can tell, our editors do not make changes to the Perseus content on load. Thinking about other types of editors, it always feels jarring for a document to be in a dirty state immediately after loading it. 

I'm not sure if you mean this PR introduced this concept, but this is something that Desmos introduced to MathQuill, and does actually solve issues for us elsewhere (we were able to remove post-processing code).

jeremywiebe:
> > This seems to introducing a new concept in the larger sense. As far as I can tell, our editors do not make changes to the Perseus content on load. Thinking about other types of editors, it always feels jarring for a document to be in a dirty state immediately after loading it.
> 
> I'm not sure if you mean this PR introduced this concept, but this is something that Desmos introduced to MathQuill, and does actually solve issues for us elsewhere (we were able to remove post-processing code).

I think we're talking past each other. I understand that this is a Desmos change, but it's causing a change in how our Perseus widget editor works in that it now triggers a `onChange` callback on load. I'm not aware of other widgets mutating the edited props on load and that's the conceptual change.  

Further, I looked at the `widget-editor.tsx` where this error originates and it seems like the handler gets the updated props (including the fixes TeX).  So what if we just handle the ref being undefined and use whatever props we get.  Like, `this.refs.widget?.serialize() ?? {}` for line 82 of that file.

nedredmond:
I was thinking we would want to update bad data with correct syntax, but yeah, I could try that approach.

jeremywiebe:
> I was thinking we would want to update bad data with correct syntax, but yeah, I could try that approach.

I think the approach I showed _would_ include the updated data. The `newProps` parameter carries the updated value from what the MathInput fired up from what I can see.

packages/perseus-editor/src/components/widget-editor.tsx
```diff
@@ -49,12 +50,15 @@ class WidgetEditor extends React.Component<
     WidgetEditorProps,
     WidgetEditorState
 > {
+    widget: React.RefObject<Editor>;
```
jeremywiebe, line 53: **nit**: I think we can just create the ref here if you want. 
```suggestion
    widget = React.createRef<Editor>();
```

packages/perseus-editor/src/components/widget-editor.tsx
```diff
@@ -72,14 +76,15 @@ class WidgetEditor extends React.Component<
...
+        }
         const newWidgetInfo = Object.assign(
             {},
             this.state.widgetInfo,
         ) as PerseusWidget;
         newWidgetInfo.options = Object.assign(
-            // eslint-disable-next-line react/no-string-refs
-            // @ts-expect-error - TS2339 - Property 'serialize' does not exist on type 'ReactInstance'.
-            this.refs.widget.serialize(),
+            this.widget.current.serialize(),
```
jeremywiebe, line 87: Nice!  I like that we're switching to an actual ref. 

If we avoid calling `props.onChange` when the widget ref isn't set up.  Will that mean we don't propagate that initial TeX mixup up to the parent editors? Or is that the part that you said didnt' work quite as I thought? 

I'll approve and let you make the final call based on what worked for you.

jeremywiebe (approved):
Thanks Ned!

packages/perseus-editor/src/__stories__/editor-page.stories.tsx
```diff
@@ -26,9 +26,38 @@ export const Demo = (): React.ReactElement => {
...
+                        times: true,
+                    },
+                    version: {
+                        major: 1,
+                        minor: 0,
+                    },
+                },
+            },
+        },
+    );
```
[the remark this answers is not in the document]
nedredmond: I did!

packages/perseus-editor/src/components/widget-editor.tsx
nedredmond: We couldn't have done, because that onChange function relies on having the serialized widget. If I swap out for an empty object or something we'd REALLY muck up the data.

Yeah, we don't propagate the initial TeX mixup... which is what I thought you wanted

PR #989, opened 2024-02-09
Add python-program widget and editor

aag:
This commit adds a new widget: python-program. This widget renders
an iframe which loads an embedded Python program page. The purpose
of this widget is to add runnable Python programs to articles in
the Python course.

Issue: CL-1472

Test Plan:
- Temporarily change the iframe URL to include http://localhost:8090/.
- Run the webapp dev environment.
- Start storybook.
- Visit the Python Program story and verify that the program loads and that the
  buttons work.
image
- Visit the Python Program Editor story and verify that it renders the two input fields.
image

### DIFF
--- a/packages/perseus-editor/src/widgets/python-program-editor.tsx
+++ b/packages/perseus-editor/src/widgets/python-program-editor.tsx
@@ -1,40 +1,32 @@
-/* eslint-disable no-useless-escape */
 /**
  * This editor is for embedding Khan Academy Python programs.
  */
-
-import {
-    components,
-    Changeable,
-    EditorJsonify,
-} from "@khanacademy/perseus";
-import PropTypes from "prop-types";
+import {components, Changeable} from "@khanacademy/perseus";
 import * as React from "react";
-import _ from "underscore";
 
-import BlurInput from "../components/blur-input";
+const {NumberInput, TextInput} = components;
 
-const {NumberInput} = components;
+import type {PerseusPythonProgramWidgetOptions} from "@khanacademy/perseus";
 
-const DEFAULT_HEIGHT = 400;
+type Props = Changeable.ChangeableProps & {
+    programID: string;
+    height: number;
+};
 
-type Props = any;
+type DefaultProps = {
+    programID: Props["programID"];
+    height: Props["height"];
+};
 
 /**
  * This is the main editor for this widget, to specify all the options.
  */
 class PythonProgramEditor extends React.Component<Props> {
-    static propTypes = {
-        ...Changeable.propTypes,
-        programID: PropTypes.number,
-        height: PropTypes.number,
-    };
-
     static widgetName = "python-program" as const;
 
-    static defaultProps: any = {
-        programID: null,
-        height: DEFAULT_HEIGHT,
+    static defaultProps: DefaultProps = {
+        programID: "",
+        height: 400,
     };
 
     change: (...args: ReadonlyArray<unknown>) => any = (...args) => {
@@ -42,16 +34,18 @@ class PythonProgramEditor extends React.Component<Props> {
         return Changeable.change.apply(this, args);
     };
 
-    _handleHeightChange: (arg1: string) => void = (height) => {
-        this.change({height});
-    };
+    serialize(): PerseusPythonProgramWidgetOptions {
+        return {
+            programID: this.props.programID,
+            height: this.props.height,
+        };
+    }
 
-    _handleProgramIDChange: (arg1: string) => void = (programID) => {
-        this.change({programID});
-    };
-
-    serialize: () => any = () => {
-        return EditorJsonify.serialize.call(this);
+    getSaveWarnings: () => ReadonlyArray<string> = () => {
+        if (this.props.programID === "") {
+            return ["The program ID is missing."];
+        }
+        return [];
     };
 
     render(): React.ReactNode {
@@ -59,9 +53,8 @@ class PythonProgramEditor extends React.Component<Props> {
             <div>
                 <label>
                     User Program ID:{" "}
-                    <NumberInput
+                    <TextInput
                         value={this.props.programID}
-                        //onChange={this._handleProgramIDChange}
                         onChange={this.change("programID")}
                         placeholder="123"
                     />
@@ -71,7 +64,6 @@ class PythonProgramEditor extends React.Component<Props> {
                     Height:{" "}
                     <NumberInput
                         value={this.props.height}
-                        // onChange={this._handleHeightChange}
                         onChange={this.change("height")}
                         placeholder="400"
                     />
--- a/packages/perseus-editor/src/widgets/python-program-editor.tsx
+++ b/packages/perseus-editor/src/widgets/python-program-editor.tsx
@@ -1,0 +1,84 @@
+/* eslint-disable no-useless-escape */
+/**
+ * This editor is for embedding Khan Academy Python programs.
+ */
+
+import {
+    components,
+    Changeable,
+    EditorJsonify,
+} from "@khanacademy/perseus";
+import PropTypes from "prop-types";
+import * as React from "react";
+import _ from "underscore";
+
+import BlurInput from "../components/blur-input";
+
+const {NumberInput} = components;
+
+const DEFAULT_HEIGHT = 400;
+
+type Props = any;
+
+/**
+ * This is the main editor for this widget, to specify all the options.
+ */
+class PythonProgramEditor extends React.Component<Props> {
+    static propTypes = {
+        ...Changeable.propTypes,
+        programID: PropTypes.number,
+        height: PropTypes.number,
+    };
+
+    static widgetName = "python-program" as const;
+
+    static defaultProps: any = {
+        programID: null,
+        height: DEFAULT_HEIGHT,
+    };
+
+    change: (...args: ReadonlyArray<unknown>) => any = (...args) => {
+        // @ts-expect-error - TS2345 - Argument of type 'readonly unknown[]' is not assignable to parameter of type 'any[]'.
+        return Changeable.change.apply(this, args);
+    };
+
+    _handleHeightChange: (arg1: string) => void = (height) => {
+        this.change({height});
+    };
+
+    _handleProgramIDChange: (arg1: string) => void = (programID) => {
+        this.change({programID});
+    };
+
+    serialize: () => any = () => {
+        return EditorJsonify.serialize.call(this);
+    };
+
+    render(): React.ReactNode {
+        return (
+            <div>
+                <label>
+                    User Program ID:{" "}
+                    <NumberInput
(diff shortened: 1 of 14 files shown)


packages/perseus-editor/src/widgets/python-program-editor.tsx
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * This editor is for embedding Khan Academy Python programs.
+ */
+
+import {
+    components,
+    Changeable,
+    EditorJsonify,
+} from "@khanacademy/perseus";
+import PropTypes from "prop-types";
```
jeremywiebe, line 11: Modern Editors (well, all Perseus components), should not use `prop-types`. Can you convert this to use standard TypeScript `Props` instead, please?

packages/perseus-editor/src/widgets/python-program-editor.tsx
```diff
@@ -0,0 +1,84 @@
...
+import * as React from "react";
+import _ from "underscore";
```
jeremywiebe, line 13: Related to above, we're working to remove our dependance on `underscore`. Can we use modern, ES6 patterns instead?

packages/perseus-editor/src/widgets/python-program-editor.tsx
```diff
@@ -0,0 +1,84 @@
...
+        programID: PropTypes.number,
+        height: PropTypes.number,
+    };
+
+    static widgetName = "python-program" as const;
+
+    static defaultProps: any = {
+        programID: null,
+        height: DEFAULT_HEIGHT,
+    };
```
jeremywiebe, line 38: Our convention in Perseus is to define a `DefaultProps` and then use indexed access types to reference the Prop key. 

Eg. 

```suggestion
type DefaultProps = {
    programID: Props["programID"];
    height: Props["heigh"];
}

...

    static defaultProps: DefaultProps = {
        programID: null,
        height: DEFAULT_HEIGHT,
    };
```

packages/perseus-editor/src/widgets/python-program-editor.tsx
```diff
@@ -0,0 +1,84 @@
...
+        this.change({height});
+    };
+
+    _handleProgramIDChange: (arg1: string) => void = (programID) => {
+        this.change({programID});
+    };
+
+    serialize: () => any = () => {
+        return EditorJsonify.serialize.call(this);
+    };
```
jeremywiebe, line 55: We don't have a pattern for this encoded very clearly yet, but I'd love for this function to:

1) clearly define it's return type. It should be the type that is defined in `perseus-types.ts` as it's the type that is serialized as widget options to Datastore. 

2) Avoid using `EditorJsonify` and just manually, and clearly, build whatever the return value should be. We (I) have found that this dynamism just complicates things and obscures which properties we're serializing, which also contributes to properties being serialized that don't really need to be. 

```suggestion
    serialize(): PerseusPythonProgramWidgetOptions {
        return {
            programID: this.props.programID,
            height: this.props.height,
        };
    } 
```

packages/perseus-editor/src/widgets/python-program-editor.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+    render(): React.ReactNode {
+        return (
+            <div>
+                <label>
+                    User Program ID:{" "}
+                    <NumberInput
+                        value={this.props.programID}
+                        //onChange={this._handleProgramIDChange}
```
jeremywiebe, line 64: Can we clean up these commented out lines (and unused handler functions)?

packages/perseus/src/perseus-types.ts
```diff
@@ -1275,6 +1278,20 @@ export type PerseusCSProgramSetting = {
...
+    height: number | string;
+}
+
+// export type PerseusPythonIFrameWidgetOptions = {
+//     // A Python Program ID
+//     url: string;
+//     // The height of the widget
+//     height: number | string;
+// }
+
```
jeremywiebe, line 1294: Can we remove this type also? `PerseusPythonProgramWidgetOptions` looks good to me.

packages/perseus/src/widgets/__tests__/python-program.ts
```diff
@@ -0,0 +1,43 @@
...
+        } as const;
+
+        // Act
+        const {container} = renderQuestion(question1, apiOptions);
+
+        // Assert
+        expect(container).toMatchSnapshot("first mobile render");
+    });
+
+    // This widget doesn't have any direct behavior, it just renders an iframe
```
jeremywiebe, line 42: Thanks for leaving this here. One challenge we haven't tackled well in Perseus yet is dealing with Stories that use frames.

packages/perseus/src/widgets/python-program.tsx
```diff
@@ -0,0 +1,83 @@
+/**
+ * This widget is for embedding Khan Academy Python programs.
+ */
+
+import {StyleSheet, css} from "aphrodite";
+import PropTypes from "prop-types";
```
jeremywiebe, line 6: Same comment as the editor. Please define a `Typescript` Props type instead of using `prop-types`.

packages/perseus/src/widgets/python-program.tsx
```diff
@@ -0,0 +1,83 @@
...
+        programID: PropTypes.string,
+        height: PropTypes.number,
+    };
+
+    static defaultProps: any = {
+        height: 400,
+    };
+
+    render(): React.ReactNode {
+        if (!this.props.programID) {
```
jeremywiebe, line 36: Is there a reason why we'd ever allow a null/undefined `programID`? If not, then the editor should disallow it and we should type the `PerseusPythonProgramWidgetOptions` as having a required `programID` field.

packages/perseus/src/widgets/python-program.tsx
```diff
@@ -0,0 +1,83 @@
...
+function getUrlFromProgramID(programID: any) {
+    const {InitialRequestUrl} = getDependencies();
+
+    const path = `/python-program/${programID}/embedded`;
+
+    return toAbsoluteUrl(path);
+}
+
+/* This renders the program in an iframe. */
+class PythonProgram extends React.Component<any> {
```
jeremywiebe, line 25: This seems to be a really simple component. Is it possible to write it as a functional component?  We are working to write new code in modern React and want to avoid writing "legacy" code when we can.

packages/perseus/src/widgets/python-program.tsx
```diff
@@ -0,0 +1,83 @@
...
+            "allow-scripts",
+            "allow-top-navigation",
+        ].join(" ");
+
+        // We sandbox the iframe so that we allowlist only the functionality
+        //  that we need. This makes it a bit safer in case some content
+        //  creator "went wild".
+        // http://www.html5rocks.com/en/tutorials/security/sandboxed-iframes/
+        return (
+            <div className={css(styles.container)}>
```
jeremywiebe, line 59: Can this just be a WonderBlocks `View` instead? In which case we wouldn't need to use `css()` either as we can just use the style from line 42.

jeremywiebe (commented):
This is a great start Adam!  I know you're breaking new-ish ground here as we haven't introduced a new widget that needs an editor in a long time. 

I have several comments. Let me know when you've updated the PR and I'll happily review again. Thanks.

packages/perseus/src/widgets/python-program.tsx
aag: I'm not completely sure how to disallow an empty `programID` in the editor. I added a `getSaveWarnings` function, but I don't know how to test it.

packages/perseus/src/widgets/python-program.tsx
aag: I tried it, but the `renderQuestion()` function used in tests is not compatible with functional components. It appears all Widgets are currently class components. Given that, do you think it would be okay to leave it as a class component?

jeremywiebe: Ah! Yes.  We'll break that new ground when we have time.  I didn't realize we haven't migrated any actual widgets to functional React components.

packages/perseus/src/widgets/python-program.tsx
jeremywiebe: It looks like you resolved this. The `PerseusPythonProgramWidgetOptions` should mark the field(s) as non-optional and then in the widget you should be able to depend on the types optionality.

jeremywiebe: I also just figured out that a widget editor can implement a function `getSaveWarnings` which is basically a validation function for saving the widget. You can implement save validation there to ensure that there is a Program ID and valid Height in that function.

aag: I've added validation of both options to `getSaveWarnings` and added tests for the validation logic.

packages/perseus/src/widgets/python-program.tsx
```diff
@@ -0,0 +1,75 @@
...
+        margin: "auto",
+        width: "100%",
+    },
+});
+
+export default {
+    name: "python-program",
+    displayName: "Python Program",
+    widget: PythonProgram,
+} as WidgetExports<typeof PythonProgram>;
```
jeremywiebe, line 75: I just thought of this. We have no scoring implemented for this widget (ie. it can't be graded). So we should add `graded: false` to this structure to signal to Perseus that it shouldn't try (or expect) to grade instances of it. 

Ref: https://github.com/Khan/perseus/blob/main/packages/perseus/src/perseus-types.ts#L93

packages/perseus/src/widgets/python-program.tsx
```diff
@@ -0,0 +1,75 @@
...
+}
+
+type Props = {
+    programID: string;
+    height: number;
+};
+
+type DefaultProps = {
+    height: Props["height"];
+};
```
jeremywiebe, line 25: Thanks for converting these to TS types from prop-types!

jeremywiebe (approved):
This is looking great!! Thanks!

packages/perseus/src/widgets/python-program.tsx
aag: Are you sure this is the right place for that? This has the type `WidgetExports`, which [doesn't have a graded property](https://github.com/Khan/perseus/blob/9f2eafe4c4649fd1ee1a83611b53908eb9cf3468/packages/perseus/src/types.ts#L409). It seems like it needs to be a property of the `Widget`, but I don't understand how objects of that type are used.

jeremywiebe: Sorry Adam, I was digging through our types and got that wrong. 

Given that the initial target is to use this in Articles, let's leave this as is. If we do want to start using it as a helper in exercises, we'd have to do more work to provide scoring, but I think that's not required now that I've dug a bit further.

PR #991, opened 2024-02-12
Enable keyboard selection of movable points

benchristel:
## Summary:
This is a step toward making interactive graphs keyboard- and
screen-reader-accessible.

Issue: https://khanacademy.atlassian.net/browse/LC-1666

Test plan:

```
yarn dev
open http://localhost:5173
```

Press `tab` repeatedly. The movable points on the graphs should be
selected in turn, with a visible outline on the selected point.

### DIFF
--- a/packages/perseus/src/interactive2/movable-point-options.ts
+++ b/packages/perseus/src/interactive2/movable-point-options.ts
@@ -60,6 +60,7 @@ const draw = {
                 radii,
                 options,
             );
+            this.state.visibleShape.wrapper.setAttribute("tabindex", "0")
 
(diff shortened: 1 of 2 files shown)


benchristel:
@nedredmond this is a spike—I still need to figure out why exactly this works, if there's a better way to do it, how to test it, etc. But it's a first step toward a solution.

packages/perseus/src/interactive2/movable-point-options.ts
```diff
@@ -60,6 +60,7 @@ const draw = {
                 radii,
                 options,
             );
+            this.state.visibleShape.wrapper.setAttribute("tabindex", "0")
```
nedredmond, line 63: I think we should be using WB Clickable here for better control over states.

packages/perseus/src/util/interactive.ts
```diff
@@ -969,6 +969,7 @@ _.extend(GraphUtils.Graphie.prototype, {
                     radii,
                     options,
                 );
+                movablePoint.mouseTarget.wrapper.setAttribute("tabindex", "0")
```
nedredmond, line 972: We have some designs, don't we? Even if we get the browser focus defaults, this won't give us consistency across browsers. Let's look at our designs and at least give that a shot

benchristel:
Closing this, since we're moving ahead with a Mafs-based solution.

PR #992, opened 2024-02-13
Update our interactionCallback to return the current user input data for Khanmigo

SonicScrewdriver:
## Summary:
As part of our Khanmigo project, we want to be able to know what's currently been input into the widgets in our exercises. 

After some investigation, it was discovered that we already have a `interactionCallback` in our `apiOptions` that was being used to close a feedback popover in our numeric inputs. This PR updates that callback to now return the current user input data so that it can be utilized by Khanmigo to provide proactive hints and supportive content. 

Issue: LC-1754

## Test plan:
- New Stories to manually test with 
- Updated server renderer test

### DIFF
--- a/packages/perseus/src/__stories__/server-item-renderer.stories.tsx
+++ b/packages/perseus/src/__stories__/server-item-renderer.stories.tsx
@@ -6,6 +6,7 @@ import {
     itemWithInput,
     itemWithLintingError,
     labelImageItem,
+    itemWithMultipleInputs,
 } from "../__testdata__/server-item-renderer.testdata";
 import {ServerItemRenderer} from "../server-item-renderer";
 
@@ -21,6 +22,21 @@ export default {
 
 export const InputNumberItem = (args: StoryArgs): React.ReactElement => {
     return <ServerItemRendererWithDebugUI item={itemWithInput} />;
+};
+
+export const InputNumberWithInteractionCallback = (
+    args: StoryArgs,
+): React.ReactElement => {
+    return (
+        <ServerItemRendererWithDebugUI
+            item={itemWithMultipleInputs}
+            apiOptions={{
+                interactionCallback: (data) => {
+                    console.log(data);
+                },
+            }}
+        />
+    );
 };
 
 export const LabelImageItem = (args: StoryArgs): React.ReactElement => {
--- a/packages/perseus/src/__stories__/server-item-renderer.stories.tsx
+++ b/packages/perseus/src/__stories__/server-item-renderer.stories.tsx
@@ -7,7 +7,7 @@ import {
     itemWithLintingError,
     labelImageItem,
     itemWithMultipleInputNumbers,
-    itemWithMultipleWidgetTypes,
+    itemWithRadioAndExpressionWidgets,
 } from "../__testdata__/server-item-renderer.testdata";
 import {ServerItemRenderer} from "../server-item-renderer";
 
@@ -67,7 +67,7 @@ export const MultiWidgetWithInteractionCallback = (
 ): React.ReactElement => {
     return (
         <ServerItemRendererWithDebugUI
-            item={itemWithMultipleWidgetTypes}
+            item={itemWithRadioAndExpressionWidgets}
             apiOptions={{
                 interactionCallback: (data) => {
                     // We are logging the interaction callback data to the console
--- a/packages/perseus/src/__stories__/server-item-renderer.stories.tsx
+++ b/packages/perseus/src/__stories__/server-item-renderer.stories.tsx
@@ -6,7 +6,8 @@ import {
     itemWithInput,
     itemWithLintingError,
     labelImageItem,
-    itemWithMultipleInputs,
+    itemWithMultipleInputNumbers,
+    itemWithMultipleWidgetTypes,
 } from "../__testdata__/server-item-renderer.testdata";
 import {ServerItemRenderer} from "../server-item-renderer";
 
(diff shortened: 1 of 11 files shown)


packages/perseus/src/widgets/input-number.tsx
```diff
@@ -202,13 +202,6 @@ class InputNumber extends React.Component<Props> {
 
     _handleFocus: () => void = () => {
         this.props.onFocus([]);
```
SonicScrewdriver, line 204: The interaction callback is no longer necessary here as we're calling it in the server-item-renderer now.

packages/perseus/src/widgets/__stories__/expression.stories.tsx
```diff
@@ -48,6 +48,9 @@ const WrappedKeypadContext = ({
                                 isMobile: isMobile,
                                 customKeypad: customKeypad,
                                 onFocusChange: action("onFocusChange"),
+                                interactionCallback: (response) => {
```
SonicScrewdriver, line 51: I found having the console log incredibly helpful for testing the interactionCallback, but I am unsure if we want to present the story in this manner. 

I decided to leave it as is, but I'm open to feedback or alternative opinions on the matter!

packages/perseus/src/types.ts
```diff
@@ -147,7 +147,7 @@ export type APIOptions = Readonly<{
     readOnly?: boolean;
     answerableCallback?: (arg1: boolean) => unknown;
     getAnotherHint?: () => unknown;
-    interactionCallback?: () => unknown;
+    interactionCallback?: (widgetData: {[widgetId: string]: any}) => void;
```
SonicScrewdriver, line 150: I feel like this type will be improved as part of of LC-1756 when we implement Kevin's typing solution.

packages/perseus/src/server-item-renderer.tsx
```diff
@@ -149,6 +149,11 @@ export class ServerItemRenderer
     ) => {
         if (newFocus != null) {
             this._setCurrentFocus(newFocus);
+
```
SonicScrewdriver, line 152: We need to call the interactionCallback in both handleInteractWithWidget and onFocus, as we're currently using it to close a feedbackPopover in webapp. I believe there are may be some future use cases where we want to know if a user is currently clicking into widgets.

packages/perseus/src/__stories__/server-item-renderer.stories.tsx
```diff
@@ -42,3 +44,37 @@ export const WithLintingError = (args: StoryArgs): React.ReactElement => {
         />
     );
 };
```
SonicScrewdriver, line 46: I added two new stories to the Server Item Renderer as I felt the current examples were far too basic to test with:

1.  InputNumberWithInteractionCallback 
This story renders two input number widgets and console logs the response data. It's helpful for testing simple multi-widget states and scoring. 

2. MultiWidgetWithInteractionCallback
This story renders two different widgets (expression and radio) with the interaction callback, so that we can test multi-widget states, scoring, and the interactionCallback. 

I kept the console log in the interactionCallback as I felt like it provides decent value for testing, but I'd be happy to remove it if we feel it isn't appropriate to override eslint to console log in a Storybook. If it is felt that these new stories are unhelpful, I'd also be happy to simply remove them entirely.

SonicScrewdriver:
There seems to be an issue with our Jest tests across several PRs regarding non-unique keys. They're unrelated to this PR and seem like they should be looked into in a separate ticket.

packages/perseus/src/widgets/__tests__/input-number.test.ts
```diff
@@ -337,18 +337,4 @@ describe("focus state", () => {
         // Assert
         expect(gotFocus).toBeTrue();
     });
-
```
SonicScrewdriver, line 340: These are no longer required now that we're calling the interactionCallback in the server-item-renderer.

packages/perseus/src/__stories__/server-item-renderer.stories.tsx
handeyeco: > I kept the console log in the interactionCallback as I felt like it provides decent value for testing

I personally don't mind console logs, but I think you could do this with Storybook Actions and the data would show up in the SB UI.

packages/perseus/src/server-item-renderer.tsx
```diff
@@ -278,9 +283,10 @@ export class ServerItemRenderer
             questionHighlightedWidgets: withRemoved,
         });
 
-        if (this.props.apiOptions.interactionCallback) {
-            this.props.apiOptions.interactionCallback();
-        }
+        // Call the interactionCallback, if it exists, with the current user input data
+        this.props.apiOptions?.interactionCallback?.(
+            this.questionRenderer.getUserInputForWidgets(),
+        );
```
handeyeco, line 289: nit: wonder if it'd be worth making a shared method, something like:

`ServerItemRenderer._handleInteraction()`

packages/perseus/src/types.ts
handeyeco: That's next on the docket?

packages/perseus/src/widgets/__tests__/numeric-input.test.ts
```diff
@@ -485,20 +485,6 @@ describe("Numeric input widget", () => {
         );
     });
 
-    it("calls the interaction callback on focus", () => {
```
handeyeco, line 488: I think if we're going to remove this test, we should add a test for `numeric-input` in the new tests (which test `input-number` but not `numeric-input`).

handeyeco (approved):
This feels like we're adding some useful functionality while at the same time simplifying our code. My one concern that's not blocking, but I feel is worth mentioning:

My read of this change is that we're going to be calling the interaction callback a lot more now. Like it looks like `input-number` and `numeric-input` were calling the interaction callback on focus, but now we're calling it on any focus change. I don't know how that would affect things, but I'm worried about unintended consequences and I'm wondering how we could protect against those.

The code changes look good to me though and looking in Webapp at how we use `interactionCallback` I don't see how this will be a problem. Assuming stricter types happen in LC-1756 and that's next, I think we can ship this and start getting feedback from the Khanmigo team.

PR #995, opened 2024-02-14
Refactor Wrapped{Ellipse,Line,Path} to be ES6 classes

benchristel:
This enables better typechecking in the area of the code we want to change
next.

Issue: none

## Test plan:

```
yarn dev
open http://localhost:5173
```

You should see a bunch of working interactive graphs


packages/perseus/src/interactive2/wrapped-defaults.ts
```diff
@@ -10,13 +10,16 @@ import _ from "underscore";
...
+import type {Graphie} from "../util/graphie";
+
 /*
  * These functions, when called on the wrapped object, simply pass the
  * arguments to the underlying Raphael object.
  */
 const PASS_TO_RAPHAEL = ["attr", "animate"];
 
-const WrappedDefaults: any = _.extend(
+export const WrappedDefaultsOld: any = _.extend(
```
benchristel, line 22: We can delete this old ES5 code now!

packages/perseus/src/interactive2/wrapped-defaults.ts
```diff
@@ -84,4 +87,92 @@ const WrappedDefaults: any = _.extend(
...
+abstract class WrappedDefaults {
+    graphie: Graphie;
+    abstract wrapper: HTMLDivElement;
+    abstract visibleShape: {
+        /*
+         * These functions, when called on the wrapped object, simply pass the
+         * arguments to the underlying Raphael object.
+         */
+        attr: (...args: any[]) => void;
+        animate: (...args: any[]) => void;
```
benchristel, line 99: It's probably worth creating more specific types for `attr` and `animate`. Maybe we should also pull out an interface for the type of VisibleShape (or maybe one exists already?)

packages/perseus/src/interactive2/arrowhead.ts
```diff
@@ -0,0 +1,96 @@
...
+        const center: Coord = [0.75, 0];
+        let points: Coord[] = [
+            [-3, 4],
+            [-2.75, 2.5],
+            [0, 0.25],
+            center,
+            [0, -0.25],
+            [-2.75, -2.5],
+            [-3, -4],
+        ];
```
benchristel, line 27: TODO: inline this value into the `_.map()` so `points` can be a const?

(while we're at it, we can convert `_.map()` to `Array.map()`.)

packages/perseus/src/interactive2/arrowhead.ts
```diff
@@ -0,0 +1,96 @@
...
+                    points[i + 1][0] +
+                    " " +
+                    points[i + 1][1] +
+                    " " +
+                    points[i + 2][0] +
+                    " " +
+                    points[i + 2][1];
+            }
+            return path;
+        };
```
benchristel, line 57: TODO: move this function outside the class so the scopes are clearer. `points` shadows the variable from the outer scope.

packages/perseus/src/interactive2/get-clip-point.ts
```diff
@@ -0,0 +1,29 @@
+import {vector as kvector} from "@khanacademy/kmath";
+
+import type {Coord} from "@khanacademy/perseus";
+
+// Given `coord` and `angle`, find the point where a line extended
+// from `coord` in the direction of `angle` would be clipped by the
+// edge of the graphie canvas. Then draw an arrowhead at that point
+// pointing in the direction of `angle`.
+export const getClipPoint = function (graph, coord: Coord, angle: number) {
```
benchristel, line 9: Add a type annotation `graph: Graphie`?

packages/perseus/src/interactive2/arrowhead.ts
jeremywiebe: I did wonder if we could even go further and define the arrow head as a simple SVG element and then only use `transform()` to place it.  But that might be too big of a refactor to be worth it.  :)

packages/perseus/src/interactive2/get-clip-point.ts
```diff
@@ -0,0 +1,29 @@
+import {vector as kvector} from "@khanacademy/kmath";
+
+import type {Coord} from "@khanacademy/perseus";
+
+// Given `coord` and `angle`, find the point where a line extended
+// from `coord` in the direction of `angle` would be clipped by the
+// edge of the graphie canvas. Then draw an arrowhead at that point
+// pointing in the direction of `angle`.
```
jeremywiebe, line 8: The part about drawing an arrowhead is not really part of this function, is it?

packages/perseus/src/interactive2/wrapped-ellipse.ts
```diff
@@ -13,82 +14,90 @@ const DEFAULT_OPTIONS = {
...
-    center: Coord,
-    radii: [number, number],
-    options: any,
-) {
-    options = _.extend({}, DEFAULT_OPTIONS, options);
+class WrappedEllipse extends WrappedDefaults {
+    graphie: Graphie;
+    initialPoint: Coord;
+    wrapper: HTMLDivElement;
+    visibleShape: {
```
jeremywiebe, line 21: The base class already has `graphie`, `wrapper`, and `visibleShape`.  This feels risky/unnecessary for the subclasses to also define them.

jeremywiebe (commented):
Just a few comments from browsing this PR. I'll leave full review/approval for the keyboard team (Ned/Mark). :)

packages/perseus/src/interactive2/get-clip-point.ts
benchristel: Good catch!

packages/perseus/src/interactive2/get-clip-point.ts
benchristel: This ended up cascading into other type errors, so we decided not to do it.

PR #996, opened 2024-02-14
Parse markdown list when rendering in JIPT context

mpolyak:
## Summary:
It was noticed that we're not properly previewing lists in the Translation Editor.
Leading to confusion for our translators as to how correctly translate the markdown.

article_list_before

The preview is even more confusing when lists are nested.

exercise_nested_lists_before

After investigating the code in both webapp and perseus repos, I found this commit from 2016: https://github.com/Khan/webapp/commit/0708615029644f6eb7d29a3d4c361344b281d577

Reverting it appears to resolve the preview issue when rendering in JIPT context.

article_list_after

exercise_nested_lists_after

The original commit summary reads:

> When translating perseus articles, we split up the text of the article
into paragraphs by splitting at double-newlines.

Yes indeed, we do this in two places:

- In [webapp](https://github.com/Khan/webapp/blob/c7fd0d7893c05e3a04d96b1154ac586e6e43d25d/services/static/javascript/manticore-package/manticore-utils.ts#L514) we split the `crwdns...:0...crwdne...:0` strings that are used in the initial preview rendering.
- Then in [perseus](https://github.com/Khan/perseus/blob/ede8afab152a0d9c285ac7512ffefa926732c2e1/packages/perseus/src/renderer.tsx#L987) we split again, this time the translated strings that are replaced in the DOM by Crowdin's [jipt.js](https://github.com/Khan/webapp/blob/c7fd0d7893c05e3a04d96b1154ac586e6e43d25d/services/static/third_party/jipt/jipt.js) integration.

So far so good. The summary continues:

> However, lists can also have double newlines in them, which can cause problems. This change just disables
the parsing of lists while doing jipt stuff.

This is where I have been racking my brain to find an example of such a case
in our translation database.

Since this "disabling" via `isJipt` state passed into `PerseusMarkdown.parse[Inline]`
only applies when rendering in JIPT context, I searched the database for any Crowdin
`crwdns...:0...crwdne...:0` strings that contain double-newlines - I was not able to find any.

Under normal rendering Perseus handles markdown lists as-is just fine, this
implies that the original issue must have applied to just these Crowdin strings,
breaking the `jipt.js` integration in some way, preventing it from replacing with translations in the DOM?

In this "fix" I chose to revert this "disabling" as I cannot find any _current_
Crowdin strings that might be affected. I'm inclined to deploy this and wait for
reports from our translators of preview regressions. I'm guessing that if this
breaks something obvious, it would be noticed right away.

In such a case a more involved fix might be to "enable" the list parsing rule
_after_ `replaceJiptContent` has been called to replace the initial-render Crowdin
string. This might require adding an additional state, as `isJipt` controls another
[crowdinId](https://github.com/Khan/perseus/blob/ede8afab152a0d9c285ac7512ffefa926732c2e1/packages/pure-markdown/src/index.ts#L148) rule in the markdown parser.

Issue: https://khanacademy.atlassian.net/browse/CP-7983

Test Plan:
 - Pull this change into webapp
 - Open Article Translator and verify strings with markdown list formatted correctly
  - `/devadmin/translations/edit/es/Article/xa0ed5200/preview`
 - Open Exercise Translator and verify strings with markdown list formatted correctly
  - `/devadmin/translations/edit/es/Exercise/xc13c1f966be9a7a6/by-order`
 - Open WYWIWYG and verify strings with markdown list are formatted correctly
  - `/science/biology/classical-genetics/variations-on-mendelian-genetics/a/multiple-alleles-incomplete-dominance-and-codominance?lang=en-pt`

### DIFF
--- a/.changeset/blue-actors-pull.md
+++ b/.changeset/blue-actors-pull.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/perseus": minor
+"@khanacademy/pure-markdown": minor
+---
+
+Revert disabling markdown parser list rule in JIPT context
--- a/packages/perseus/src/__tests__/perseus-markdown.test.ts
+++ b/packages/perseus/src/__tests__/perseus-markdown.test.ts
@@ -453,69 +453,6 @@ describe("perseus markdown", () => {
 
         it.each([
             {
-                content: "1. test\n\n" + "2. boo\n\n",
-                expected: [
-                    {
-                        type: "paragraph",
-                        content: [
-                            {
-                                type: "text",
-                                content: "1",
-                            },
-                            {
-                                type: "text",
-                                content: ". test",
-                            },
-                        ],
-                    },
-                    {
-                        type: "paragraph",
-                        content: [
-                            {
-                                type: "text",
-                                content: "2",
-                            },
-                            {
-                                type: "text",
-                                content: ". boo",
-                            },
-                        ],
-                    },
-                ],
-            },
-            {
-                content: "* test\n\n" + "* boo\n\n",
-                expected: [
-                    {
-                        type: "paragraph",
-                        content: [
-                            {
-                                type: "text",
-                                content: "* test",
-                            },
-                        ],
-                    },
-                    {
-                        type: "paragraph",
-                        content: [
-                            {
-                                type: "text",
-                                content: "* boo",
-                            },
-                        ],
-                    },
-                ],
-            },
-        ])("should ignore lists in jipt mode", ({content, expected}) => {
-            // Arrange, Act
-            const parsed = parse(content, {isJipt: true});
-
-            // Assert
-            expect(parsed).toEqual(expected);
-        });
-
-        it.each([
-            {
                 content: "$",
                 expected: [
                     {
--- a/packages/pure-markdown/src/__tests__/index.test.ts
+++ b/packages/pure-markdown/src/__tests__/index.test.ts
@@ -441,69 +441,6 @@ describe("pure markdown", () => {
 
         it.each([
             {
-                content: "1. test\n\n" + "2. boo\n\n",
-                expected: [
-                    {
-                        type: "paragraph",
-                        content: [
-                            {
-                                type: "text",
-                                content: "1",
-                            },
-                            {
-                                type: "text",
-                                content: ". test",
-                            },
-                        ],
-                    },
-                    {
-                        type: "paragraph",
-                        content: [
-                            {
-                                type: "text",
-                                content: "2",
-                            },
-                            {
-                                type: "text",
-                                content: ". boo",
-                            },
-                        ],
-                    },
-                ],
-            },
-            {
-                content: "* test\n\n" + "* boo\n\n",
-                expected: [
-                    {
-                        type: "paragraph",
-                        content: [
-                            {
-                                type: "text",
-                                content: "* test",
-                            },
-                        ],
-                    },
-                    {
-                        type: "paragraph",
-                        content: [
-                            {
-                                type: "text",
-                                content: "* boo",
-                            },
-                        ],
-                    },
-                ],
-            },
-        ])("should ignore lists in jipt mode", ({content, expected}) => {
-            // Arrange, Act
-            const parsed = parse(content, {isJipt: true});
-
-            // Assert
-            expect(parsed).toEqual(expected);
-        });
-
-        it.each([
-            {
                 content: "$",
                 expected: [
                     {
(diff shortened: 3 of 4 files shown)


jeremywiebe (approved):
Thanks for the detailed writeup!  I think the testing approach makes sense.

mpolyak:
Thank you @jeremywiebe for helping me build a webapp ZND: https://github.com/Khan/webapp/pull/19894

As soon as I opened the test case Article (`/devadmin/translations/edit/es/Article/xa0ed5200/preview`), I stumbled on a regression!

I think I'm beginning to understand what the 2016 fix was for. The "lists can also have double newlines in them, which can cause problems." is talking about something like this example case:
```
* Allele pairs may have a variety of dominance relationships (that is, one allele of the pair may not completely “hide” the other in the heterozygote).

* There are often many different alleles of a gene in a population.
```

And this is how this snipper looks in the raw `Article.perseusContent`:
```
...\n\n* Allele pairs may have a variety of dominance relationships (that is, one allele of the pair may not completely “hide” the other in the heterozygote).\n\n* There are often many different alleles of a gene in a population.\n\n...
```

There is indeed "double-newlines" in the middle of it. I was confused why I couldn't find this in our translation database, and this reason is that we _extract_ these two list items as _two separate_ strings in Crowdin:

https://crowdin.com/editor/khanacademy/30972/enus-es?view=comfortable&filter=basic&value=0#3469164
https://crowdin.com/editor/khanacademy/30972/enus-es?view=comfortable&filter=basic&value=0#3638873

This is how it looks in our database. What we upload to Crowdin:
```
crowdin-go [PRODUCTION] => select msgid from crowdinstrings where msgkey='7e91342b5ad085138e372afc8e173893c086f5d2';
                                                                          msgid
---------------------------------------------------------------------------------------------------------------------------------------------------------
 * Allele pairs may have a variety of dominance relationships (that is, one allele of the pair may not completely “hide” the other in the heterozygote).
(1 row)

crowdin-go [PRODUCTION] => select msgid from crowdinstrings where msgkey='9db25e2fe4499e4b865c796a14b24f0daa5b956e';
                                msgid
---------------------------------------------------------------------
 * There are often many different alleles of a gene in a population.
(1 row)
```

What we download from Crowdin to drive the `jipt.js` integration:
```
crowdin-go [PRODUCTION] => select * from translation where locale='en-PT' and msgkey='7e91342b5ad085138e372afc8e173893c086f5d2';
 locale |                  msgkey                  | deleted | approved | updatedseqno |      updatedtimestamp       |             msgstr             | msgstrplural | linterror | isautotranslated
--------+------------------------------------------+---------+----------+--------------+-----------------------------+--------------------------------+--------------+-----------+------------------
 en-PT  | 7e91342b5ad085138e372afc8e173893c086f5d2 | f       | t        |            0 | 2019-12-04 23:30:43.8277+00 | crwdns3469164:0crwdne3469164:0 | {}           |           | f
(1 row)

crowdin-go [PRODUCTION] => select * from translation where locale='en-PT' and msgkey='9db25e2fe4499e4b865c796a14b24f0daa5b956e';
 locale |                  msgkey                  | deleted | approved | updatedseqno |      updatedtimestamp       |             msgstr             | msgstrplural | linterror | isautotranslated
--------+------------------------------------------+---------+----------+--------------+-----------------------------+--------------------------------+--------------+-----------+------------------
 en-PT  | 9db25e2fe4499e4b865c796a14b24f0daa5b956e | f       | t        |            0 | 2019-12-04 23:30:43.8277+00 | crwdns3638873:0crwdne3638873:0 | {}           |           | f
(1 row)
```

And here's a recording of what happens if the 2016 change is reverted. The result appears that we're misattributing which translation goes with which string:

cp-7983-article-issue

Since these list items are uploaded to Crowdin as distinct strings, my initial thinking that this is strictly a Translation Editor issue in how we split the strings _prior_ to rendering, I'm guessing in [getArticleShowableJiptContent](https://github.com/Khan/webapp/blob/c7fd0d7893c05e3a04d96b1154ac586e6e43d25d/services/static/javascript/manticore-package/manticore-utils.ts#L514).

We render the two list items as a single paragraph to the DOM, which confuses `jipt.js`.

cp-7983-article-issue

Instead, we should be splitting them into two paragraphs, to match how these list items exist as two separate strings on Crowdin, something like:

cp-7983-article-issue-split

I'm going to continue investigating this in the Translation Editor to see if I can properly split these for rendering.

mpolyak:
> Or try the alternative proposed in the PR summary of passing additional state into PerseusMarkdown.parse[Inline] on the re-render after replaceJiptContent, to "enable" the markdown parser list rule.
>
> The latter, if it works, would be the more contained minimal change, so I'll try that first.

N/M that doesn't work for the same reason - section is rendered as a single paragraph.

PR #874, opened 2023-12-15
Update workflows to use node 20

somewhatabstract:
## Summary:
This updates the GitHub Actions workflows to use node 20 instead of node 16.

Issue: FEI-5405

## Test plan:
1. `nvm use 20` to setup node 20
2. `yarn` to check upgrade path
3. `rm -rf node_modules`
4. `yarn` to check "from scratch" path
5. `yarn typecheck`
6. `yarn test`
7. `yarn lint`
8. `yarn build`
8. Put up this PR and see that the GitHub Actions all work.

PR #998, opened 2024-02-14
Update .nvmrc to Node 18

nedredmond:
Node 18 seems to be the minimum version to run Perseus now. This change only applies to devs using node version managers (nvm, n, fnm, asdf, etc), and automatically switches the version to 18+ when `cd`ing into the directory.

Given that we are updating to 20 everywhere, I don't think this will be controversial.


jeremywiebe (commented):
Can we just jump to v20? FE Infra is actively moving everything to v20 so we might as well jump to that instead.

https://github.com/Khan/perseus/pull/874

nedredmond:
@jeremywiebe done!

jeremywiebe (approved):

PR #1000, opened 2024-02-15
[Bug Fixes] Numerous issues with the Number-line widget

SonicScrewdriver:
## Summary:
This PR fixes two issues related to the number line widget (MOB-5422 & MOB-5416). It also adds a storybook story to help test number-lines, as we didn't have anything showing the widget with "isTickCtrl" set to "True". 

MOB-5422 
This is solved by programmatically focusing on the input element during click and tap events so that the onKeyUp handler can be fired correctly for keyboard events. 

MOB-5416 
This is solved by fixing a callback in math input, where we were returning "false" rather than a proper callback. 

Issue: MOB-5422 & MOB-5416

## Test plan:
- manual

### DIFF
--- a/packages/math-input/src/components/input/math-input.tsx
+++ b/packages/math-input/src/components/input/math-input.tsx
@@ -330,7 +330,7 @@ class MathInput extends React.Component<Props, State> {
         }
     };
 
-    blur: (e?: React.SyntheticEvent) => void = (e) => {
+    blur: () => void = () => {
         this.mathField.blur();
 
         this.setState((prevState) => {
--- a/packages/math-input/src/components/input/math-input.tsx
+++ b/packages/math-input/src/components/input/math-input.tsx
@@ -46,7 +46,6 @@ type HandleState = {
 type State = {
     focused: boolean;
     handle: HandleState;
-    showFocusStyle: boolean;
 };
 
 // eslint-disable-next-line react/no-unsafe
@@ -77,7 +76,6 @@ class MathInput extends React.Component<Props, State> {
 
     state: State = {
         focused: false,
-        showFocusStyle: false,
         handle: {
             animateIntoPosition: false,
             visible: false,
@@ -363,7 +361,7 @@ class MathInput extends React.Component<Props, State> {
         this.mathField.focus();
         this.props?.onFocus();
 
-        this.setState({focused: true, showFocusStyle: true}, () => {
+        this.setState({focused: true}, () => {
             // NOTE(charlie): We use `setTimeout` to allow for a layout pass to
             // occur. Otherwise, the keypad is measured incorrectly. Ideally,
             // we'd use requestAnimationFrame here, but it's unsupported on
@@ -919,13 +917,13 @@ class MathInput extends React.Component<Props, State> {
     };
 
     render(): React.ReactNode {
-        const {showFocusStyle, handle} = this.state;
+        const {focused, handle} = this.state;
         const {style} = this.props;
 
         const innerStyle = {
             ...inlineStyles.innerContainer,
             borderWidth: this.getBorderWidthPx(),
-            ...(showFocusStyle
+            ...(focused
                 ? {
                       borderColor: Color.blue,
                   }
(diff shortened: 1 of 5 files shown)


packages/math-input/src/components/input/math-input.tsx
```diff
@@ -857,7 +872,7 @@ class MathInput extends React.Component<Props, State> {
             const value = this.mathField.getContent();
             if (this.props.value !== value) {
                 this.mathField.setContent(this.props.value);
```
SonicScrewdriver, line 874: This is the fix for MOB-5416.

packages/math-input/src/components/input/math-input.tsx
```diff
@@ -658,6 +666,13 @@ class MathInput extends React.Component<Props, State> {
...
             this.focus();
         }
+
+        // If the user clicked on the input using a mouse, we want to
+        // set the focus to the inputRef so that the keyUp event will
+        // be triggered when the user types on the keyboard.
+        // This is necessary to support Chromebooks as they use mobile user
+        // agents, do not simulate touch events, and have physical keyboards.
+        this.inputRef?.focus();
     };
```
SonicScrewdriver, line 676: This is the fix for MOB-5422

packages/math-input/src/components/input/math-input.tsx
```diff
@@ -76,6 +77,7 @@ class MathInput extends React.Component<Props, State> {
 
     state: State = {
         focused: false,
```
SonicScrewdriver, line 79: I ended up deciding that we needed to separate the state regarding the focus style and the actual focus of the input, in order to support tabbing to the math input and receiving focusStyling WITHOUT closing the keypad. 

Separating the logic allows us to (1) separately handle the focus styling, (2) show the custom keypad on input focus, (3) keep the custom keypad open if the user tabs into the keypad, and (4) maintain the blurOnClickOutside functionality. 

I'm unsure if I should update the focused state property to something like "inputActive". Totally open to suggestions on this!

packages/perseus/src/widgets/__stories__/number-line.stories.tsx
```diff
@@ -12,3 +18,42 @@ type StoryArgs = Record<any, any>;
 export const Question1 = (args: StoryArgs): React.ReactElement => {
     return <RendererWithDebugUI question={question1} />;
 };
+
```
SonicScrewdriver, line 21: I wanted to add a story that shows this particular widget form as there's currently no where else to test the number-line with the tick controller!

packages/math-input/src/components/input/math-input.tsx
```diff
@@ -623,6 +624,13 @@ class MathInput extends React.Component<Props, State> {
         if (!this.state.focused) {
             this.focus();
         }
+
+        // If the user clicked on the input using a mouse or tap gesture,
+        // we want to set the focus to the inputRef so that the keyUp
+        // event will be triggered when the user types on the keyboard.
+        // This is necessary to support Chromebooks as they use mobile user
+        // agents, do not simulate touch events, and have physical keyboards.
+        this.inputRef?.focus();
```
jeremywiebe, line 633: For both of these fixes, immediately above them we do the check on `this.state.focused` and then call `this.focus()`. 

I'm wondering if these `inputRef?.focus()` calls might belong in there?  Of if the state.focused is false in some cases? 

I trust you've tested this, but just had the question in case you knew. 

I'll approve it so you aren't blocked and leave it to you if you stick with this or integrate into the `this.focus()` function.

jeremywiebe (approved):
This looks reasonable.  One question that I'll leave with you.

packages/math-input/src/components/input/math-input.tsx
SonicScrewdriver: Yes! Unfortunately state.focused might be false in some cases, and we want to focus the input ref every time the user clicks or taps on the input. 

I have an upcoming PR that should help clean up things slightly, but I had to revert those commits as they were causing a flumoxing test failure that we haven't been able to nail down yet. I've also been chatting with design about different approaches to focus handling (perhaps a button like desktop), or @nishasy mentioned the possibility of having Chromebooks be recognized as "desktop" machines which would let us cut out a lot of the edge-cases we're having to handle specifically for Chromebooks. 

I'll land this PR as is for now, but I definitely want to clean up the focus function in the future.

PR #1001, opened 2024-02-15
Set Expression widget's `accessible` field to true.

nishasy:
## Summary:

Now that expression widget is screenreader accessible, we want to mark it as such.
This will make it so that the exercise editor will not auto-check and auto-disable the
"requires screen or mouse" checkbox when the expression widget is used.

This will change the value of `hasViolatingWidgets` in [item-controls.tsx](https://github.com/Khan/webapp/blob/6d54f60f3a3bbb0758e4d499f9eb123ff27d6ef5/services/static/javascript/perseus-one-package/item-controls.tsx#L670-L682) where the
checkbox is rendered.

NOTE: This will just make it so that the checkbox won't auto-disable. It will _not_
update the existing values to reflect Expression now being accessible. A backfill
will still be necessary for that with our current architecture.

Issue: none

## Test plan:
N/A?

### Checkbox in question

Screenshot 2024-02-14 at 5 29 03 PM

### DIFF
--- a/packages/perseus/src/widgets/expression.tsx
+++ b/packages/perseus/src/widgets/expression.tsx
@@ -687,6 +687,7 @@ ExpressionWithDependencies.getOneCorrectAnswerFromRubric =
 export default {
     name: "expression",
     displayName: "Expression / Equation",
+    accessible: true,
     defaultAlignment: "inline-block",
     widget: ExpressionWithDependencies,
     transform: (widgetOptions: PerseusExpressionWidgetOptions): RenderProps => {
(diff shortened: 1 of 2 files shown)


nedredmond:
This is great for new exercises, but I do not believe this will fix the issue retroactively without the backfill.

nedredmond (approved):

nishasy:
@nedredmond Yes, that's correct! I'll update the description to clarify this.

packages/perseus/src/widgets/expression.tsx
```diff
@@ -687,6 +687,7 @@ ExpressionWithDependencies.getOneCorrectAnswerFromRubric =
 export default {
     name: "expression",
     displayName: "Expression / Equation",
+    accessible: true,
```
jeremywiebe, line 690: 😮   Today I learned!!!

PR #1003, opened 2024-02-15
Allow movable points to be selected with the tab key

benchristel:
Issue: https://khanacademy.atlassian.net/browse/LC-1668

## Test plan:

- Open the interactive graph dev UI (`yarn dev`)
- Press Tab repeatedly
- You should see each movable point get selected in turn. The selected point
  should have a blue outline.

### DIFF
--- a/packages/perseus/src/interactive2/wrapped-ellipse.tsx
+++ b/packages/perseus/src/interactive2/wrapped-ellipse.tsx
@@ -1,3 +1,4 @@
+import * as React from "react";
 import {vector as kvector} from "@khanacademy/kmath";
 import _ from "underscore";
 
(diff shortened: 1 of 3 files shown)


nedredmond (approved):
Looks like snapshots need updating

packages/perseus/src/interactive2/wrapped-ellipse.tsx
```diff
@@ -53,6 +58,21 @@ class WrappedEllipse extends WrappedDrawing {
             // movable wrapper so that when moved the browser does not scroll page
             this.wrapper.style.touchAction = "none";
 
+            // Option 2: instead of addToMouseLayerWrapper(this.wrapper)
+            // we could do addToMouseLayerWrapper(reactRoot)
+            reactRender(
```
jeremywiebe, line 63: Can we use a React portal instead?  `reactRender` is an abstraction over `ReactDOM.render()` that I think we should remove eventually as React Context's don't flow through it whereas they do with Portals.

packages/perseus/src/util/graphie.ts
```diff
@@ -736,10 +740,12 @@ export class Graphie {
                 left: left + "px",
                 top: top + "px",
             });
-            // wrapper.setAttribute("data-graphie-type", "ellipse");
+
+            const raphaelContainer = document.createElement("div");
```
jeremywiebe, line 744: **question**: What's the reason for needing to nest a div here?

PR #1004, opened 2024-02-16
Replace hardcoded Aria strings with translatable ones

nedredmond:
Uses newly-added API on MathQuill to replace hardcoded strings intended for `aria` alerts with translatable strings. Creates a file to host said strings.

> [!NOTE]
> This will not be landed until after the Level Access review completes on Feb 22.

Issue: LC- 1712

Testing: Try replacing the strings with silly things, start Storybook, and [test any consumer of MathQuill](http://localhost:6006/?path=/story/math-input-components-v2-keypad-with-mathquill--v-2-keypad-with-mathquill) either with the accessibility tree open in dev tools or screen reader on. You should see your strings replace things as you navigate your cursor through or select an expression.


packages/math-input/src/components/input/strings.ts
```diff
@@ -0,0 +1,24 @@
+import * as i18n from "@khanacademy/wonder-blocks-i18n";
+
+import type {MathQuillAriaStatisStringsMap} from "./mathquill-types";
+
+export const mathQuillStrings = {
+    mathInputBox: i18n._("Math input box"),
+    ariaStaticStringsMap: {
+        before: (obj) => i18n._("before %(obj)s", {obj}),
```
jeremywiebe, line 8: Just to be sure, could you run `yarn extract-strings` in the root of the Perseus repo on this PR and then double-check that `packages/math-input/dist/strings.js` has these strings in it?

jeremywiebe (approved):

PR #1005, opened 2024-02-16
Update ArticleRenderer to split sections into paragraphs in JIPT context

mpolyak:
## Summary:
This is a continuation to https://github.com/Khan/perseus/pull/996, where "disabling" of the `list` rule was removed
from the markdown parser in order to correctly render lists in JIPT context.

However, that change broke Article preview rendering in the Translation Editor,
due to a mismatch in paragraph indexes. Causing the `jipt.js` integration to not
be able to attribute rendered strings to translations on Crowdin. This is fixed in: https://github.com/Khan/webapp/pull/19917

With these two changes in place the Translation Editor is now rendering markdown
lists correctly without any other regression.

Unfortunately this is not case for articles containing markdown lists rendered in
the WYSIWYG. These are regressed due to the same paragraph index mismatch,
as unlike the Translation Editor, the JIPT WYSIWYG renders articles using `ArticleRenderer`
(same as normal page rendering.)

cp-7983-broken-article-wysiwyg

To fix this, we apply the same change of splitting article sections into paragraphs
when rendering in JIPT context. This enables the `jipt.js` integration to attribute
the strings in the DOM to translations on Crowdin.

cp-7983-article-wysiwyg

Issue: https://khanacademy.atlassian.net/browse/CP-7983

Test Plan:
 - Open any article in any KA locale site and verify rendering is unaffected
 - Open any article in WYSIWYG and verify there are no regressions to rendering
 - Open article with a markdown list in WYSIWYG and verify lists are rendered correctly
   - `/science/biology/classical-genetics/variations-on-mendelian-genetics/a/multiple-alleles-incomplete-dominance-and-codominance?lang=en-pt`

### DIFF
--- a/packages/perseus/src/article-renderer.tsx
+++ b/packages/perseus/src/article-renderer.tsx
@@ -7,7 +7,8 @@ import * as PerseusLinter from "@khanacademy/perseus-linter";
 import classNames from "classnames";
 import * as React from "react";
 
-import {DependenciesContext} from "./dependencies";
+import {DependenciesContext, getDependencies} from "./dependencies";
+import JiptParagraphs from "./jipt-paragraphs";
 import {ClassNames as ApiClassNames, ApiOptions} from "./perseus-api";
 import Renderer from "./renderer";
 import Util from "./util";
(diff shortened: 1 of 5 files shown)


packages/perseus/src/article-renderer.tsx
```diff
@@ -162,9 +163,32 @@ class ArticleRenderer
...
+                    (paragraph) => {
+                        paragraphs.push({
+                            ...section,
+                            content: paragraph,
+                        });
+                    },
+                );
+            }
+
+            return paragraphs;
```
jeremywiebe, line 188: One thought I had. This logic is now in 2 (or is it 3) places between webapp and Perseus.  Would it make sense to add a function onto the `JIPT` dependency object that does this work?  Then at least we'd just call that function from this location and the function could live in webapp closer to the other parsing/splitting code from Khan/webapp#19917. Just a thought.

jeremywiebe (approved):
This makes sense. I'd be happy to help if you want to try the route of extracting this logic to be a dependency we push into Perseus.

packages/perseus/src/article-renderer.tsx
mpolyak: Good idea, this can definitely be captured in a higher level utility function.

mpolyak:
>  I'd be happy to help if you want to try the route of extracting this logic to be a dependency we push into Perseus.

I think it's worth doing this as a follow up set of changes, once these 2 perseus PRs and 1 webapp PR go out without issue.

PR #1006, opened 2024-02-16
Allow learners to use any locale's trig operators in the Expression widget

benchristel:
Previously, learners had to use the English versions of trig operators like
`tan` and `csc`, no matter what language they were learning in. This was
awkward because different languages use different names for these operators,
e.g. `tg` and `cosec`.

This commit updates the expression widget so learners can use the operators for
their locale.

Issue: https://khanacademy.atlassian.net/browse/LC-1632

## Test plan:

```
GIVEN I am looking at an expression widget (e.g. on Test Everything)
WHEN I type any of the trig operators listed in the spreadsheet linked from the Jira issue
THEN the operator is displayed in an upright font

GIVEN I have typed the correct answer to a trig problem using operators appropriate to my locale
WHEN I submit the answer
THEN it is accepted as correct
```


jeremywiebe (approved):
I was starting to think if needed to add locale awareness to Perseus, but I think this approach is probably better (at least for this problem). 

I think it's probably useful for learners to be able to use a trig operator in their own language, or in English. 

👍

PR #1009, opened 2024-02-20
Update graph-settings.tsx to React class component

nishasy:
## Summary:

- Update GraphSettings to be a React class instead of old `createReactClass` const.
- Add types

Tests coming in follow-up PR.

## Test plan:
`yarn jest`
`yarn typecheck`

### DIFF
--- a/packages/perseus-editor/src/components/graph-settings.tsx
+++ b/packages/perseus-editor/src/components/graph-settings.tsx
@@ -60,10 +60,21 @@ type State = {
 class GraphSettings extends React.Component<Props, State> {
     _isMounted = false;
 
+    static stateFromProps(props) {
+        return {
+            labelsTextbox: props.labels,
+            gridStepTextbox: props.gridStep,
+            snapStepTextbox: props.snapStep,
+            stepTextbox: props.step,
+            rangeTextbox: props.range,
+            backgroundImage: _.clone(props.backgroundImage),
+        };
+    }
+
     constructor(props: Props) {
         super(props);
 
-        this.state = this.stateFromProps(props);
+        this.state = GraphSettings.stateFromProps(props);
     }
 
     static defaultProps = {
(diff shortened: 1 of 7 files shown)


packages/perseus-editor/src/components/graph-settings.tsx
```diff
@@ -101,13 +111,13 @@ const GraphSettings = createReactClass({
...
+    }
 
-    componentWillUnmount: function () {
+    componentWillUnmount() {
         this._isMounted = false;
-    },
+    }
 
-    stateFromProps: function (props) {
+    stateFromProps(props) {
```
jeremywiebe, line 120: Can you mark this as `static`?  It will mean a few changes to call sites, but it'll ensure we don't "cheat" in this function ever and start accessing instance fields.

packages/perseus-editor/src/widgets/interaction-editor/interaction-editor.tsx
```diff
@@ -222,11 +222,16 @@ class InteractionEditor extends React.Component<Props, State> {
...
-                        step={this.props.graph.tickStep}
-                        gridStep={this.props.graph.gridStep}
+                        range={
+                            this.props.graph.range as [
+                                [number, number],
+                                [number, number],
+                            ]
+                        }
+                        step={this.props.graph.tickStep as [number, number]}
+                        gridStep={this.props.graph.gridStep as [number, number]}
```
jeremywiebe, line 234: **suggestion**: In the next (or near future) PR you could change this components props so that they use tuples instead of `ReadonlyArray<>`. That'd eliminate the need to do the casting you're doing here... and I think it's more accurate.

jeremywiebe (approved):
Awesome! Thanks.

packages/perseus-editor/src/widgets/interaction-editor/interaction-editor.tsx
nishasy: I think the props are already tuples. The problem is that we're doing a `slice()` on them [just above this](https://github.com/Khan/perseus/blob/3110c79bb524c78dd7ac0c737bb749f79b64f109/packages/perseus-editor/src/components/graph-settings.tsx#L303-L307), which is causing it to become `number[]` types. And then I get a type error because that means it can be shorter than the tuple expects.

nishasy: I can move the `as [number, number]` to those lines instead?

jeremywiebe: Actually, all those `.slice()` calls look like they're just used to copy the array.  If you switch to spreading the array, Typescript seems to retain that their a tuple and you can eliminate the `as` casting entirely. 

https://www.typescriptlang.org/play?#code/C4TwDgpgBAggRgYygXigbwFBW1ATgQwDsBzCALigG1CBXAWzglwBopaGmBdAbgwF9eGBAHtCAZ2BQAHhXhJUmHHiKkKlAIwAGVls78MQ0RKghZiFOiw4CJclQB0jqfZuk9fIA

Note: you are welcome to land this PR as is and do these changes in a new PR.  Whatever works for you. :)

PR #1014, opened 2024-02-21
Fix Spacebar Usage on Explanation Widget

mark-fitzgerald:
## Summary:
The explanation widget should respond to keyboard interactions such that both the Enter/Return key and the spacebar activate it. This bug calls out that the spacebar is instead scrolling the page (see "Before" video), which is the default browser behavior. Changing the widget container from `<Link>` to `<Clickable>` improves semantics and properly handles keystrokes (it is now a `<button>` element instead of an `<a>` element).

Issue: LC-212

## Test plan:
1. Open the Perseus Storybook (locally: `http://localhost:6006/?path=/story/perseus-widgets-explanation--question-1`)
1. Using the keyboard, tab until the text "[Explanation]" is in focus.
    * To avoid numerous tab keystrokes, you can click the "explanation" word prior to the widget, then press the tab key to advance focus to the widget.
1. With the "[Explanation]" widget in focus, press the spacebar.
    * The extra text associated with the widget should now be showing.
    * The focus indicator should still be on the "[Explanation]" widget.
1. Press the spacebar again.
    * The extra text associated with the widget should now be hidden.
1. The same steps using the Enter key instead of the spacebar should yield the same results.

### Before - Spacebar scrolls the page

https://github.com/Khan/perseus/assets/13896410/5c63dc9a-dff2-4420-9560-a89ffe854680

### After - Spacebar now activates the widget

https://github.com/Khan/perseus/assets/13896410/d20a345a-a177-49e9-828d-788149827a7f

### DIFF
--- a/packages/perseus/src/widgets/__tests__/explanation.test.ts
+++ b/packages/perseus/src/widgets/__tests__/explanation.test.ts
@@ -120,7 +120,7 @@ describe("Explanation", function () {
         renderQuestion(question1);
 
         // Act
-        const expandLink = screen.getByRole("button", {expanded: false});
+        const expandLink = screen.getByTestId("expand-button");
         userEvent.click(expandLink);
 
         // Assert
@@ -133,9 +133,9 @@ describe("Explanation", function () {
         renderQuestion(question1);
 
         // Act
-        const expandLink = screen.getByRole("button", {expanded: false});
+        const expandLink = screen.getByTestId("expand-button");
         userEvent.click(expandLink); // expand and then
-        const collapseLink = screen.getByRole("button", {expanded: true});
+        const collapseLink = screen.getByTestId("expand-button");
         userEvent.click(collapseLink); // collapse
 
         // Assert
--- a/packages/perseus/src/widgets/__tests__/explanation.test.ts
+++ b/packages/perseus/src/widgets/__tests__/explanation.test.ts
@@ -115,28 +115,77 @@ describe("Explanation", function () {
         expect(container).toMatchSnapshot("expanded");
     });
 
-    it("can be expanded", function () {
+    it("can be expanded and collapsed with a mouse click", function () {
         // Arrange
         renderQuestion(question1);
 
-        // Act
-        const expandLink = screen.getByTestId("expand-button");
-        userEvent.click(expandLink);
+        // Act - expand with a click
+        const expandButton = screen.getByRole("button", {
+            name: "[Explanation]",
+        });
+        userEvent.click(expandButton);
 
         // Assert
         // get asserts if it doesn't find a single matching element
         expect(screen.getByText("This is an explanation")).toBeVisible();
+
+        // Act - collapse with a click
+        const collapseButton = screen.getByRole("button", {
+            name: "[Hide explanation!]",
+        });
+        userEvent.click(collapseButton); // collapse
+
+        // Assert
+        expect(screen.queryByText("This is an explanation")).toBeNull();
     });
 
-    it("can be collapsed", function () {
+    it("can be expanded and collapsed with the keyboard - Enter key", function () {
         // Arrange
         renderQuestion(question1);
 
-        // Act
-        const expandLink = screen.getByTestId("expand-button");
-        userEvent.click(expandLink); // expand and then
-        const collapseLink = screen.getByTestId("expand-button");
-        userEvent.click(collapseLink); // collapse
+        // Act - expand with a click
+        const expandButton = screen.getByRole("button", {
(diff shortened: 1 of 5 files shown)


packages/perseus/src/widgets/__tests__/explanation.test.ts
```diff
@@ -133,9 +133,9 @@ describe("Explanation", function () {
         renderQuestion(question1);
 
         // Act
-        const expandLink = screen.getByRole("button", {expanded: false});
+        const expandLink = screen.getByTestId("expand-button");
         userEvent.click(expandLink); // expand and then
-        const collapseLink = screen.getByRole("button", {expanded: true});
+        const collapseLink = screen.getByTestId("expand-button");
```
benchristel, line 138: Is there a way we could avoid using a test ID here, e.g. by selecting the element by text instead? My understanding is that if a feature is hard to test without a test ID, it's likely inaccessible (users can't select elements by test ID, after all). I don't think there's an accessibility problem here—just want to promote exemplary testing practices.

packages/perseus/src/widgets/explanation.tsx
```diff
@@ -110,53 +111,15 @@ class Explanation extends React.Component<Props, State> {
...
-                            styles.linkContainer,
-                            styles.exerciseLink,
-                        )}
-                    >
-                        <Link
-                            className={css(styles.explanationLink)}
-                            href={href}
-                            onClick={onClick}
-                            role="button"
-                            aria-expanded={this.state.expanded}
```
benchristel, line 153: Do we not need `aria-expanded` anymore?

packages/perseus/src/widgets/explanation.tsx
```diff
@@ -110,53 +111,15 @@ class Explanation extends React.Component<Props, State> {
...
-                // anchor tag, see: `perseus-api.jsx`.
-                linkContainer = (
-                    <div
-                        className={css(
-                            styles.linkContainer,
-                            styles.exerciseLink,
-                        )}
-                    >
-                        <Link
-                            className={css(styles.explanationLink)}
```
benchristel, line 149: What did these styles do? (`linkContainer`, `exerciseLink`, `articleLink`, and `explanationLink`)

benchristel (commented):
The test plan and video look amazing! I left some questions inline.

packages/perseus/src/widgets/__tests__/explanation.test.ts
mark-fitzgerald: Good catch. I re-implemented the use of `getByRole()` with an updated content reference.

packages/perseus/src/widgets/explanation.tsx
mark-fitzgerald: I'll be doing another PR that completes the partial accessibility in this widget. Even though this attribute doesn't aid in accessibility without other related attributes, I've restored it in the `<Clickable>` element so that it's not forgotten and can be leveraged in the forthcoming PR.

packages/perseus/src/widgets/explanation.tsx
mark-fitzgerald: `explanationLink` provides color and italics.
`exerciseLink` and `articleLink` provide font size and line height.
I don't know what value `linkContainer` provides. I think originally it was there to convert the containing `<div>` tag to an inline element. Now that the container is replaced with a `<button>`, there is no need for that styling. I will remove it (testing shows that doing so does not impact usage).

packages/perseus/src/widgets/explanation.tsx
```diff
@@ -110,53 +111,18 @@ class Explanation extends React.Component<Props, State> {
...
-                        >
-                            {`[${linkAnchor}]`}
-                        </Link>
-                    </div>
-                );
-            }
+            const viewStyling = [styles.linkContainer, styles.explanationLink];
+            viewStyling.push(
+                isArticle ? styles.articleLink : styles.exerciseLink,
+            );
```
benchristel, line 117: Could this style be defined inline?
```suggestion
            const viewStyling = [
                styles.linkContainer,
                styles.explanationLink,
                isArticle ? styles.articleLink : styles.exerciseLink,
            ];
```
In general, our style is to avoid mutating arrays and POJOs where possible (though the mutation here is pretty harmless). TypeScript's type checking is less robust in the presence of mutability, because TS doesn't have a concept of [variance](https://flow.org/en/docs/lang/variance/).

benchristel (approved):
Looks great, thanks! I left one comment inline on a minor stylistic thing, but I don't think this will need another round of review.

PR #1018, opened 2024-02-23
Add question data for the graph flipbook

benchristel:
...and a script to generate it

Issue: none

## Test plan:

```
yarn dev
open 'http://localhost:5173/#flipbook'

# Copy a bunch of files to your clipboard
cat data/questions/0{0,1,2,3}/*/* | pbcopy
```

Paste the copied data into the flipbook input.

You should then be able to flip through the questions with the "next" and
"previous" buttons.


benchristel:
I have mixed feelings about committing all this data to the repo, but I feel like it's easier than making everyone download the content snapshot and extract the questions themselves.

I believe this dataset is comprehensive (we won't need to test any questions that aren't in this dataset). Questions in English and Hindi are included. Hindi is one of the hardest languages for MathJax to render, so if it works we are good.

jeremywiebe (approved):

PR #1019, opened 2024-02-23
Use arrow functions in GraphSettings

nishasy:
## Summary:
I didn't update all the old functions to arrow functions
in GraphSettings when updating it to be a React class,
which is now causes `this.setState()` to fail because
`this` is not bound.

The easiest way to make sure the functions are bound
to `this` is to make them arrow functions.

(This issue was found while attempting to add tests.
That's not done yet, but it should be the following PR.)

In this PR:
- Update GraphSettings old functions to arrow functions
  so they bind `this`.

## Test plan:
- Go to http://localhost:6006/?path=/docs/perseus-editor-editorpage--docs
- Click the "Add a widget" dropdown
- Choose interactive graph
- Open the "interactive-graph 1" section
- Open the web console
- Make a change to every single text field, dropdown, button, or other input.
- Check the console to make sure there are no errors trying to call
  `this.setState()`

### DIFF
--- a/.changeset/olive-cows-pump.md
+++ b/.changeset/olive-cows-pump.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": patch
+---
+
+Use arrow functions in GraphSettings to bind `this` properly
--- a/packages/perseus-editor/src/components/graph-settings.tsx
+++ b/packages/perseus-editor/src/components/graph-settings.tsx
@@ -128,21 +128,21 @@ class GraphSettings extends React.Component<Props, State> {
         this._isMounted = false;
     }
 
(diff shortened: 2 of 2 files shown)


jeremywiebe (changes requested):
I know we discussed this, but as I was looking at this I had the thought that a safer approach would be to just revert the PR that converted this component to a class component.  That would mean we essentially don't ship any changes to GraphSettings in the upcoming release.  Then you can re-apply the class PR and the rest of your changes in a cohesive release.

nishasy:
Reverted the breaking changes instead.

Going to add the arrow functions in the PR re-updating graph settings later.

PR #1022, opened 2024-02-23
Add stories for GraphSettings and InteractiveGraphEditor

nishasy:
## Summary:
Right now, there aren't really stories for GraphSettings
or InteractiveGraphEditor.

Adding them here so that it's easier to visually test
them later.

Note: I added prop comments in interactive-graph-editor.tsx
so that storybook automatically uses those props. I didn't do
that for graph-settings.tsx yet here, because that will be
converted to a React class soon. It will be easier to add the
prop comments after that conversion.

## Test plan:
Graph Settings
http://localhost:6006/?path=/docs/perseus-editor-components-graph-settings--docs

InteractiveGraphEditor
http://localhost:6006/?path=/docs/perseus-editor-widgets-interactive-graph-editor--docs

### Graph Settings story + control panel

Screenshot 2024-02-26 at 8 25 52 PM

Screenshot 2024-02-26 at 8 26 06 PM

### Interactive Graph Editor story + control panel

Screenshot 2024-02-26 at 8 26 24 PM

Screenshot 2024-02-26 at 8 26 32 PM

Screenshot 2024-02-26 at 8 26 37 PM

### DIFF
--- a/packages/perseus-editor/src/widgets/interactive-graph-editor.tsx
+++ b/packages/perseus-editor/src/widgets/interactive-graph-editor.tsx
@@ -80,24 +80,24 @@ type Props = {
     /**
      * The range of the graph in the x and y directions.
      */
-    range: [Range, Range]; // x, y
+    range: [x: Range, y: Range];
     /**
      * How far apart the tick marks on the axes are in the x and y
      * directions.
      */
-    step: [number, number]; // x, y
+    step: [x: number, y: number];
     /**
      * How far apart the grid lines are in the x and y directions.
      */
-    gridStep: [number, number]; // x, y
+    gridStep: [x: number, y: number];
     /**
      * How far apart the snap-to points are in the x and y directions.
      */
-    snapStep: [number, number]; // x, y
+    snapStep: [x: number, y: number];
     /**
      * The size of the graph in pixels.
      */
-    box: [number, number]; // x, y
+    box: [x: number, y: number];
 
     /**
      * An error message to display in the graph area, or true if the
@@ -135,6 +135,9 @@ type Props = {
      * The number of ticks to display on the ruler.
      */
     rulerTicks: number;
+    /**
+     * The data for checking if an answer is correct.
+     */
     correct: any; // TODO(jeremy)
 
     /**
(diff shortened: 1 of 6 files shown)


packages/perseus-editor/src/widgets/interactive-graph-editor.tsx
```diff
@@ -73,23 +73,73 @@ const FieldLabel = (props: {children: string}) => {
 type Props = {
     apiOptions: APIOptionsWithDefaults;
 
+    /**
+     * The labels for the x and y axes.
+     */
     labels: ReadonlyArray<string>;
```
jeremywiebe, line 79: Thanks for adding these in!

packages/perseus-editor/src/widgets/interactive-graph-editor.tsx
```diff
@@ -73,23 +73,73 @@ const FieldLabel = (props: {children: string}) => {
...
+    /**
+     * The range of the graph in the x and y directions.
+     */
     range: [Range, Range]; // x, y
```
jeremywiebe, line 83: One thing we can do now in Typescript is add names to the tuple items. I don't think we can use them when referring to the tuple items, but I think it might help with documentation. 

Not requiring this in this PR, but you can if you like the idea. 

https://blog.logrocket.com/exploring-use-cases-typescript-tuples/#what-are-named-tuples

```suggestion
    range: [x: Range, y: Range]; // x, y
```

packages/perseus-editor/src/widgets/interactive-graph-editor.tsx
```diff
@@ -73,23 +73,73 @@ const FieldLabel = (props: {children: string}) => {
...
     showTooltips: boolean;
+    /**
+     * The label to display on the ruler, if any.
+     */
     rulerLabel: string;
+    /**
+     * The number of ticks to display on the ruler.
+     */
     rulerTicks: number;
     correct: any; // TODO(jeremy)
```
jeremywiebe, line 138: This represents the data for checking if the answer is correct. We definitely need to improve types here. No action required.

jeremywiebe (approved):
Love it. I have one request.  Could you add a screenshot of the two new stories to this PR just for posterity?

PR #1023, opened 2024-02-24
Mafs graph foundation with segment

nedredmond:
Summary:
Beginnings of Mafs migration. Includes shim to swap in Mafs implementation, a directory with Mafs files and components, and an untested method of updating the `graph` state for verification.

Needs more testing. Looking forward to the flipbook.

Screenshot 2024-02-23 at 7 29 58 PM

### DIFF
--- a/packages/perseus/src/types.ts
+++ b/packages/perseus/src/types.ts
@@ -228,6 +228,8 @@ export type APIOptions = Readonly<{
     // This is useful for when we need to know how a widget has shuffled its
     // the available choices.
     onWidgetStartProps?: (widgets: PerseusWidgetMap) => PerseusWidgetMap;
+    // Feature flags that can be passed from consuming application
+    flags?: Record<string, boolean | Record<string, boolean>>;
 }>;
 
 type TeXProps = {
--- a/packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
+++ b/packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
@@ -1,4 +1,5 @@
 import {describe, beforeEach, it} from "@jest/globals";
+import {waitFor} from "@testing-library/react";
 
 import {clone} from "../../../../../testing/object-utils";
 import {testDependencies} from "../../../../../testing/test-dependencies";
@@ -16,7 +17,6 @@ import type {Coord} from "../../interactive2/types";
 import type {PerseusRenderer} from "../../perseus-types";
 import type Renderer from "../../renderer";
 import type {APIOptions} from "../../types";
-import {waitFor} from "@testing-library/dom";
 
 const updateWidgetState = (renderer: Renderer, widgetId: string, update) => {
     const state = clone(renderer.getSerializedState());
@@ -110,7 +110,7 @@ describe("interactive-graph widget", function () {
     );
 });
 
-describe.only("segment graph", () => {
+describe("segment graph", () => {
     const apiOptions = {flags: {mafs: {segment: true}}};
 
     it("should render", () => {
--- a/packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
+++ b/packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
@@ -1,5 +1,6 @@
 import {describe, beforeEach, it} from "@jest/globals";
 import {waitFor} from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
 
 import {clone} from "../../../../../testing/object-utils";
 import {testDependencies} from "../../../../../testing/test-dependencies";
@@ -118,19 +119,53 @@ describe("segment graph", () => {
         renderQuestion(segmentQuestion, apiOptions);
     });
 
-    it("rejects incorrect answer", async () => {
+    it("should reject when has not been interacteracted with", () => {
+        // Arrange
         const {renderer} = renderQuestion(segmentQuestion, apiOptions);
 
+        // Act
+        // no action
+
+        // Assert
+        expect(renderer).toHaveInvalidInput();
+    });
+
+    it("rejects incorrect answer", async () => {
+        // Arrange
+        const {renderer, container} = renderQuestion(
+            segmentQuestion,
+            apiOptions,
+        );
+
+        // eslint-disable-next-line testing-library/no-container, testing-library/no-node-access
+        const movablePoints = container.querySelectorAll(
+            "circle.mafs-movable-point-hitbox",
+        );
+
+        // Act
+        await userEvent.type(movablePoints[0], "{arrowup}{arrowdown}");
+
+        // Assert
         await waitFor(() => {
             expect(renderer).toHaveBeenAnsweredIncorrectly();
         });
     });
 
     it("accepts correct answer", async () => {
-        const {renderer} = renderQuestion(
+        const {renderer, container} = renderQuestion(
             segmentQuestionDefaultCorrect,
             apiOptions,
         );
+
+        // eslint-disable-next-line testing-library/no-container, testing-library/no-node-access
+        const movablePoints = container.querySelectorAll(
+            "circle.mafs-movable-point-hitbox",
+        );
+
+        // Act
+        await userEvent.type(movablePoints[0], "{arrowup}{arrowdown}");
+
+        // Assert
         await waitFor(() => {
             expect(renderer).toHaveBeenAnsweredCorrectly();
         });
--- a/packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
+++ b/packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
@@ -4,7 +4,10 @@ import {clone} from "../../../../../testing/object-utils";
 import {testDependencies} from "../../../../../testing/test-dependencies";
 import * as Dependencies from "../../dependencies";
 import {ApiOptions} from "../../perseus-api";
-import {questionsAndAnswers} from "../__testdata__/interactive-graph.testdata";
+import {
+    questionsAndAnswers,
+    segmentQuestion,
+} from "../__testdata__/interactive-graph.testdata";
 
 import {renderQuestion} from "./renderQuestion";
 
@@ -103,4 +106,11 @@ describe("interactive-graph widget", function () {
             });
         },
     );
+});
+
+describe("segment graph", () => {
+    it("should render", () => {
+        // Mafs isn't very RTL friendly...
+        renderQuestion(segmentQuestion, blankOptions);
+    });
 });
--- a/packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
+++ b/packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
@@ -1,6 +1,6 @@
 import {describe, beforeEach, it} from "@jest/globals";
 import {waitFor} from "@testing-library/react";
-import userEvent from "@testing-library/user-event";
+import {userEvent as userEventLib} from "@testing-library/user-event";
 
 import {clone} from "../../../../../testing/object-utils";
 import {testDependencies} from "../../../../../testing/test-dependencies";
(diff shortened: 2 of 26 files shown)


packages/perseus/src/widgets/interactive-graphs/grid.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+import type {PerseusImageBackground} from "../../perseus-types";
+import type {SizeClass} from "../../util/sizing-utils";
+
+const renderLineLabel = (n: number, [min, max]: [number, number]) =>
+    n !== -1 && n !== min && n !== max;
+
+export const Grid = (props: {
+    step: number[];
+    gridStep: number[];
```
jeremywiebe, line 12: Can these be `[number, number]` instead (to make sure there's only ever 2 numbers)?

packages/perseus/src/widgets/interactive-graphs/grid.tsx
```diff
@@ -0,0 +1,29 @@
...
+    range: [[number, number], [number, number]];
+    backgroundImage?: PerseusImageBackground;
+    containerSizeClass: SizeClass;
+}) => (
+    <Coordinates.Cartesian
+        xAxis={{
+            lines: props.step[0],
+            subdivisions: props.step[0] / props.gridStep[0],
+            labels: (n) => (renderLineLabel(n, props.range[0]) ? n : ""),
```
jeremywiebe, line 21: I _think_ in Graphie the axis labels are rendered with KaTeX, but I'd be curious to see if that's really ever needed (like do we only ever render numbers or can there be more complex expressions in the axis labels).  I suspect the answer is "numbers only".

packages/perseus/src/widgets/interactive-graphs/legacy-grid.tsx
```diff
@@ -0,0 +1,49 @@
...
+import AssetContext from "../../asset-context";
+import {SvgImage} from "../../components";
+import {interactiveSizes} from "../../styles/constants";
+import {getInteractiveBoxFromSizeClass} from "../../util/sizing-utils";
+
+import type {PerseusImageBackground} from "../../perseus-types";
+import type {SizeClass} from "../../util/sizing-utils";
+
+/**
+ * If a graphie-to-png URL is provided in `backgroundImage`, will return the rendered graphie background.
```
jeremywiebe, line 13: **terminology**: [graphie-to-png](https://khanacademy.atlassian.net/wiki/spaces/LC/pages/2535227849/Graph+Terminology#Graphie-to-png) is the name for the website to build Graphies.  I think in this case we can just call it a "graphie URL".

https://khanacademy.atlassian.net/wiki/spaces/LC/pages/2535227849/Graph+Terminology

packages/perseus/src/widgets/interactive-graphs/legacy-grid.tsx
```diff
@@ -0,0 +1,49 @@
...
+    const {url, width, height} = backgroundImage ?? {};
+    if (url && typeof url === "string") {
+        const box = getInteractiveBoxFromSizeClass(containerSizeClass);
+        const scale = box[0] / interactiveSizes.defaultBoxSize;
+        return (
+            <View
+                style={{
+                    position: "absolute",
+                    bottom: 0,
+                    left: 0,
```
jeremywiebe, line 29: **question**: Is the bottom, left anchoring due to the fact that we put the `y` label at the top of the graph and the `x` to the right?

packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
```diff
@@ -104,3 +107,10 @@ describe("interactive-graph widget", function () {
         },
     );
 });
+
+describe("segment graph", () => {
+    it("should render", () => {
+        // Mafs isn't very RTL friendly...
+        renderQuestion(segmentQuestion, blankOptions);
```
jeremywiebe, line 114: The `renderQuestion` helper returns the `renderer` itself. So you should be able to call `guessAndScore()` on that to check if the new Mafs grid implementation supports grading correctly. 

We can work on this tomorrow while we pair.

jeremywiebe (commented):
I've done a once-over. To me, it all looks like a great start!  I suspect we'll find more places in the `interactive-graph.tsx` that we need to add a seam.  Hopefully we can get the Cypress tests going to help with that.

packages/perseus/src/widgets/interactive-graphs/legacy-grid.tsx
nedredmond: Yes!

packages/perseus/src/widgets/interactive-graph.tsx
```diff
@@ -1672,22 +1678,32 @@ class InteractiveGraph extends React.Component<Props, State> {
...
-    };
+    getUserInput: () => PerseusGraphType = () =>
+        this.mafsRef.current ??
+        InteractiveGraph.getUserInputFromProps(this.props);
 
-    simpleValidate: (rubric: Rubric) => PerseusScore = (rubric) => {
-        return InteractiveGraph.validate(this.getUserInput(), rubric, this);
-    };
+    simpleValidate: (rubric: Rubric) => PerseusScore = (rubric) =>
+        InteractiveGraph.validate(this.getUserInput(), rubric, this);
```
benchristel, line 1686: I don't think this way of scoring the graph is going to work. `InteractiveGraph.validate` looks pretty entangled with our existing graph code, so I'd prefer to rethink how we do scoring as part of the Mafs rewrite.

Could we return a dummy value here for now?
```suggestion
        ({type: "invalid"}); // TODO: implement scoring
```

packages/perseus/src/widgets/interactive-graphs/graphs/segment.tsx
```diff
@@ -0,0 +1,105 @@
...
+    React.useEffect(() => {
+        // update state when points move
+        const segment = [p1.point, p2.point];
+        props.onGraphChange((current) => ({
+            ...current,
+            coords: current.coords
+                ? current.coords.map((c, i) => (i === props.i ? segment : c))
+                : [segment],
+        }));
+    });
```
benchristel, line 96: Do we want `p1` and `p2` to be listed as deps here? If their identities aren't stable, maybe we can do this (re-run the useEffect callback when the points change)?
```suggestion
    }, [...p1.point, ...p2.point]);
```

packages/perseus/src/widgets/interactive-graphs/graphs/segment.tsx
```diff
@@ -0,0 +1,105 @@
...
+export const SegmentGraph = (props: SegmentProps) => {
+    const segments = getSegmentCoords(props);
+
+    React.useEffect(() => {
+        // set initial state
+        props.onGraphChange((current) => ({
+            ...current,
+            coords: segments,
+        }));
+    });
```
benchristel, line 57: Should `segments` be a useEffect dep?

packages/perseus/src/widgets/interactive-graphs/utils.ts
```diff
@@ -0,0 +1,86 @@
...
+    coordsList: ReadonlyArray<Coord>,
+    ranges: PerseusInteractiveGraphWidgetOptions["range"],
+): Array<Coord> =>
+    coordsList.map<Coord>(
+        (coords) =>
+            coords.map((coord, i) => {
+                const extent = ranges[i][1] - ranges[i][0];
+                return (coord + ranges[i][1]) / extent;
+            }) as Coord,
+    );
```
benchristel, line 75: I think these `normalize*` functions should have unit tests.

benchristel:
This is very exciting! I left some comments inline.

packages/perseus/src/widgets/interactive-graph.tsx
nedredmond: Actually, `validate` is one of the only pure functions in the class. Grading works well, now that I've fixed the mafsRef.

packages/perseus/src/widgets/interactive-graphs/graphs/segment.tsx
nedredmond: I'm reworking this area now that I see there's a requirement that a graph should not be graded before it's been interacted with.

packages/perseus/src/types.ts
```diff
@@ -228,6 +228,8 @@ export type APIOptions = Readonly<{
     // This is useful for when we need to know how a widget has shuffled its
     // the available choices.
     onWidgetStartProps?: (widgets: PerseusWidgetMap) => PerseusWidgetMap;
+    // Feature flags that can be passed from consuming application
+    flags?: Record<string, boolean | Record<string, boolean>>;
```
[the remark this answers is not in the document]
benchristel: I believe the below would also work, but I don't have super strong feelings one way or the other:

```ts
this.props.apiOptions?.flags?.[`mafs-${this.props.graph.type}-graph`] &&
mafsGraphTypes.includes(this.props.graph.type)
```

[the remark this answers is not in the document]
mark-fitzgerald: I like the use of optional chaining to simplify the check.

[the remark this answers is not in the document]
jeremywiebe: **thought**: Can we be more explicit with what the flags can contain here?  I think at "compile time" we should know what flags Perseus supports.  That'd help integration too as we'd have invalid types in the host app if they try to provide flag keys that we don't support. 

This will also help in the Perseus code so that we don't try to read flags that will never be populated. 

So something like this: 

```js 
 // Feature flags that can be passed from consuming application
     flags?: {
         mafs?: {
             segment?: true // or boolean
         }
     }
```

packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
```diff
@@ -104,3 +111,70 @@ describe("interactive-graph widget", function () {
...
+            apiOptions,
+        );
+
+        // eslint-disable-next-line testing-library/no-container, testing-library/no-node-access
+        const movablePoints = container.querySelectorAll(
+            "circle.mafs-movable-point-hitbox",
+        );
+
+        // Act
+        await userEvent.type(movablePoints[0], "{arrowup}{arrowdown}");
```
jeremywiebe, line 153: Really really glad we can drive these with keyboard events. It's a double-win in that we are testing keyboard support right out of the gate!  Nicely done!

packages/perseus/src/widgets/interactive-graph.tsx
```diff
@@ -174,6 +177,8 @@ class InteractiveGraph extends React.Component<Props, State> {
         },
     };
 
+    mafsRef = React.createRef<Widget>();
```
jeremywiebe, line 180: `<Widget>` 🔥

jeremywiebe (approved):
Awesome. I have a thought about the types for `flags`, but that's non-blocking.

packages/perseus/src/types.ts
jeremywiebe, line 232: I don't know where my comment went. 🙃 

I wonder if we should type this more strictly. Having explicit keys will make integration into the host applications easier as it'll be obvious which flags Perseus supports. It will also help Perseus because we'll only access keys/flags that we know the host knows about. 

So maybe something like this?

```suggestion
    flags?: {
        mafs?: {
            segment: boolean 
        }
    }
```

nedredmond: You commented on a resolved conversation, but I saw it!

packages/perseus/src/widgets/interactive-graphs/grid.tsx
```diff
@@ -0,0 +1,33 @@
...
+import * as React from "react";
+
+import type {SizeClass} from "../../util/sizing-utils";
+
+interface GridProps {
+    step: [number, number];
+    gridStep: [number, number];
+    range: [[number, number], [number, number]];
+    containerSizeClass: SizeClass;
+    markings: "graph" | "grid" | "none";
```
mark-fitzgerald, line 11: I've seen this `markings` type in a couple places in `perseus-types.ts`. Should we have a single type for `markings` that we can then reference in each of these places (including here)?

mark-fitzgerald (approved):

packages/perseus/src/types.ts
nedredmond: I was really hoping to avoid temporary types that would require cleanup. Index signatures of string to boolean are the usual pattern for feature flags-- it's how it's implemented everywhere else.

PR #1028, opened 2024-02-26
Fix Perseus Widget Option Types

SonicScrewdriver:
## Summary:
The goal of this PR (and LC-1756) is to support Khanmigo by ensuring that there are specific types for our WidgetOptions, so that we can infer the types with little effort. 

1. First, this required type fixes/specifications for several individual widgets as they had their WidgetOptions type set to "any", which was causing our union type to be "any". 
2. After the WidgetOptions were improved, I've added @kevinbarabash's proposed new approach to setting the Widgets type. This new solution seems to work significantly better, and our types are now being inferred correctly. 
3. Finally, I went around and did some code cleanup. I removed any previous cases where I could find where we were casting the types as a work around, and ensured that we're using the new PerseusWidgets type wherever possible. 

Issue: LC-1756

## Test plan:
- ensure all previous tests pass


packages/perseus/src/__tests__/renderer.test.tsx
```diff
@@ -192,12 +188,10 @@ describe("renderer", () => {
                 ...question1,
                 widgets: {
                     ...question1.widgets,
```
SonicScrewdriver, line 190: We're specifically breaking our type formatting here, so we still require the override for this test. 

However, it does make me wonder if there could be some issues with our content with the new, more specific types.

packages/perseus/src/perseus-types.ts
```diff
@@ -14,6 +14,88 @@ type StyleParams = {
 // TODO(FEI-5054): Figure out how to get global .d.ts files working with monorepos
 type Empty = Record<never, never>;
 
+export type WidgetsType = {
```
SonicScrewdriver, line 17: I am fully open to hearing any alternative naming suggestions here. I feel like our Perseus type names are incredibly confusing currently.

packages/perseus/src/widgets.ts
```diff
@@ -302,7 +302,7 @@ export const getRendererPropsForWidgetInfo = (
         // It shouldn't matter what we return here, but for consistency
         // we return the untransformed options, as if the widget did
         // not have a transform defined.
```
SonicScrewdriver, line 304: We should be returning the whole item here "as-is" if we're unable to recognize it. Ideally we should never hit this case.

packages/perseus/src/widgets/__testdata__/plotter.testdata.ts
```diff
@@ -19,6 +16,10 @@ export const question1: PerseusRenderer = {
                     "$4^{\\text{th}} \\text{}$",
                     "$5^{\\text{th}} \\text{}$",
                 ],
```
SonicScrewdriver, line 18: Dummy data added to complete type

packages/perseus/src/widgets/__tests__/explanation.test.ts
```diff
@@ -181,8 +179,7 @@ describe("Explanation", function () {
         it("should always return 0 points", () => {
             const result = ExplanationWidgetExports.widget.validate(
                 {},
```
SonicScrewdriver, line 181: We don't need to cast these types anymore -- they're properly inferred!

packages/perseus/src/__tests__/mock-asset-loading-widget.tsx
```diff
@@ -8,15 +8,15 @@ import type {WidgetExports} from "../types";
 
 export const mockedAssetItem: PerseusItem = {
     question: {
```
SonicScrewdriver, line 10: This one was a little tricky as it was faking a widget that doesn't exist. I've changed the widget to use the ExampleWidget, as it is one of the simplest widgets, and the test passes without issues.

packages/perseus/src/__tests__/server-item-renderer.test.tsx
```diff
@@ -272,7 +272,7 @@ describe("server item renderer", () => {
         // everything is loaded.
 
         // Arrange
```
SonicScrewdriver, line 274: This is related to the mocked-asset changes above

packages/perseus/src/perseus-types.ts
```diff
@@ -14,6 +13,124 @@ type StyleParams = {
 // TODO(FEI-5054): Figure out how to get global .d.ts files working with monorepos
 type Empty = Record<never, never>;
 
+export type Widgets = {
```
SonicScrewdriver: Thank you to @kevinbarabash for this solution!

packages/perseus/src/perseus-types.ts
SonicScrewdriver: I ended up going with "PerseusWidgets" but I'm still not the happiest with the naming clarity.

packages/perseus/src/widgets/__tests__/__snapshots__/plotter.test.tsx.snap
```diff
@@ -222,603 +222,21 @@ exports[`plotter widget should snapshot basic question: initial render 1`] = `
...
-            <span
-              class="graphie-label rotate"
-              style="position: absolute; padding: 7px; color: black; left: 14.999999999999995px; top: 175px; margin-left: 0px; margin-top: 0px; fill: none; stroke: #71B307; font-weight: bold;"
-            >
-              Number of absent students
-            </span>
-          </div>
+            <title>
+              Rendering Error!
+            </title>
```
jeremywiebe, line 233: This is probably something we should fix before checking in.

packages/perseus/src/widgets/lights-puzzle.tsx
```diff
@@ -7,8 +7,36 @@ import _ from "underscore";
 import * as Changeable from "../mixins/changeable";
 import WidgetJsonifyDeprecated from "../mixins/widget-jsonify-deprecated";
 
+import type {PerseusLightsPuzzleWidgetOptions} from "../perseus-types";
 import type {WidgetExports} from "../types";
 
+// Types
```
jeremywiebe, line 13: :sad-panda: I've been trying to delete this widget for about 2 years. lol

jeremywiebe (commented):
I took a swing through and have a few comments. I'll leave final +1 to someone from the API project.

packages/perseus/src/perseus-types.ts
```diff
@@ -14,6 +14,88 @@ type StyleParams = {
 // TODO(FEI-5054): Figure out how to get global .d.ts files working with monorepos
 type Empty = Record<never, never>;
 
+export type PerseusWidgets = {
```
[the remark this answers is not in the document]
handeyeco: I think we could also follow up with a helper that does roughly the same thing. Each widget should have a `type` field, so I wonder if we could do something like...

``` JS
function getWidget(perseusItem: PerseusItem, widgetId: PerseusWidgetId): Something {
  const widgetData = perseusJson.widgets[widgetId]

  if (!widgetData?.type) {
    throw new Error("Widget data not found")
  }

  switch(widgetData.type) {
    case 'categorizer':
      return widgetData as CategorizerWidget
    default: 
      throw new Error("Widget type not found")
  }
}
```

That would further help isolate consumers from thinking about `radio 1` and whatnot; they could just provide the key without thinking about what's inside of it and receive a properly typed object.

I still think the type in the PR is useful, even if just internally. It's really technically not `[key: string]: whatever` and this type has already pointed out some places where we he incorrectly formatted data.

packages/perseus/src/widgets/lights-puzzle.tsx
```diff
@@ -105,17 +133,12 @@ class Tile extends React.Component<any> {
     }
 
     _flip = () => {
-        this.props.onChange(!this.props.value);
+        this.props.onChange({value: !this.props.value});
```
handeyeco, line 136: This is a logic change, are we sure this is going to work as expected?

packages/perseus/src/widgets/lights-puzzle.tsx
```diff
@@ -130,8 +153,7 @@ class TileGrid extends React.Component<any> {
                                             size={this.props.size}
                                             onChange={_.partial(
                                                 this.props.onChange,
-                                                y,
-                                                x,
+                                                {value: {y, x}},
```
handeyeco, line 156: Same here

packages/perseus/src/widgets/reaction-diagram.tsx
```diff
@@ -1,20 +1,17 @@
...
 import {Molecule} from "./molecule";
 
+import type {
+    PerseusReactionDiagramSeparators,
+    PerseusReactionDiagramWidgetOptions,
+} from "../perseus-types";
 import type {WidgetExports} from "../types";
 
-class Separator extends React.Component<any> {
+class Separator extends React.Component<PerseusReactionDiagramSeparators> {
```
handeyeco, line 12: Some here: I'm not sure we really want to be using types from `perseus-types` as PropTypes, but I imagine I'm just wrong or overly cautious.

packages/perseus/src/widgets/unit.tsx
```diff
@@ -61,7 +62,7 @@ type ValidationResult =
 type Props = {
     apiOptions: APIOptions;
     onChange: ChangeHandler;
-    value: string;
+    value: PerseusUnitInputWidgetOptions["value"];
```
handeyeco, line 65: Same here

packages/perseus/src/widgets/reaction-diagram.tsx
```diff
@@ -118,13 +116,13 @@ class ReactionDiagramWidget extends React.Component<any> {
...
                             <Molecule
-                                id={id}
+                                widgetId={id}
                                 rotationAngle={this.props.rotationAngle[i]}
                                 smiles={s}
                             />
                             {i === this.props.smiles.length - 1 ? null : (
                                 <Separator
-                                    data={this.props.separators[i]}
+                                    data={this.props.separators[i].data}
```
handeyeco, line 125: Another place where logic changes I think. Are we sure it's safe? I realize these are deprecated widgets, but I think they're still in use. Although you might be fixing the bugs that made them problematic for all I know 😂 Just want to make sure we know we're not breaking prod.

handeyeco (commented):
This is a big PR and I ran out of time. Maybe we can go over it 1/1 just to answer lingering questions?

packages/perseus/src/perseus-types.ts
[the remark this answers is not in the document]
SonicScrewdriver: Love it! I'll change that as that's far more descriptive.

packages/perseus-editor/src/widgets/simple-markdown-tester.tsx
```diff
@@ -1,21 +1,21 @@
...
 
+import type {PerseusSimpleMarkdownTesterWidgetOptions} from "@khanacademy/perseus";
+
 const mdParse = PerseusMarkdown.parse;
 const mdOutput = PerseusMarkdown.basicOutput;
 
-type Props = any;
-
+type Props = Changeable.ChangeableProps & {
+    value: PerseusSimpleMarkdownTesterWidgetOptions["value"];
```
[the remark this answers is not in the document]
SonicScrewdriver: I paired with Jeremy on this component a little earlier and I _believe_ that this is the desired approach! We ended up following the pattern that we were seeing in the other widgets.

packages/perseus/src/perseus-types.ts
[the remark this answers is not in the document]
jeremywiebe: That might work! :)  I think we'd have to make the function a generic to get back a useful type, but that would mean the caller would be providing the type it expects and that doesn't feel like a full solution. I do like the direction you're thinking  though! :)

handeyeco (approved):
Third and I talked through this. I have concerns about some of the logic checks in lights-puzzle and the reaction thing; Third is going to manually test that these components still work and might add a couple of unit tests.

packages/perseus/src/perseus-types.ts
```diff
@@ -1342,25 +1445,49 @@ export type PerseusInputNumberWidgetOptions = {
...
+
+export type PerseusSimpleMarkdownTesterWidgetOptions = {
+    value: string;
+};
+export type PerseusUnitInputWidgetOptions = {
+    value: string;
+};
 
 export type PerseusWidgetOptions =
+    | PerseusExpressionWidgetOptions
```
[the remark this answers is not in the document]
SonicScrewdriver: No it's not! I'll put it back where it belongs. :)

packages/perseus/src/widgets/lights-puzzle.tsx
SonicScrewdriver: Got the approval to delete the widget now! :D

packages/perseus/src/widgets/lights-puzzle.tsx
SonicScrewdriver: Got approval to delete this widget!

PR #1029, opened 2024-02-27
Upgrade @testing-library/user-event to v14

jeremywiebe:
## Summary:

This PR upgrades the user-event library to v14. This contains a similar set of changes to what is contained in Khan/webapp#20109. It prepares the way for Khan Academy to move to React v18. 

In short:
  * All calls to the `userEvent` library are now treated as async (prefixed with `await`)
  * Most usages of the library now need to be setup before we can use it. This is because the new library does not know intrinsically how to advance Jest's fake timers. As a result, we have to tell it how because most of our Perseus tests use fake timers. This is a bit verbose but overall pretty unintrusive based on the pattern I figured out (open to feedback and improvements!!).
  * 
  
Issue: LC-1753

## Test plan:

`yarn test` :thumbsup:
`yarn lint` :thumbsup:


config/test/test-setup.ts
```diff
@@ -9,6 +9,8 @@ import {StyleSheetTestUtils} from "aphrodite";
 import jestSerializerHtml from "jest-serializer-html";
 import {addSerializer} from "jest-specific-snapshot";
 
+import "@testing-library/jest-dom";
```
jeremywiebe, line 12: We had sprinkled this in most test files. This makes starting a new test file just that little bit easier!

package.json
```diff
@@ -50,10 +50,9 @@
     "@storybook/react-vite": "^7.6.17",
     "@swc-node/register": "^1.6.6",
     "@swc/core": "^1.3.68",
-    "@testing-library/dom": "^8.11.0",
```
jeremywiebe, line 53: `jest-dom` already depends on this library and we never use it directly. So this avoids version mismatches (which I did hit while working on this PR) when we bump testing library versions.

packages/math-input/src/components/__tests__/integration.test.tsx
```diff
@@ -1,12 +1,11 @@
-import "@testing-library/jest-dom";
 import {
     screen,
     render,
     fireEvent,
     within,
     waitFor,
 } from "@testing-library/react";
-import userEvent from "@testing-library/user-event";
+import {userEvent as userEventLib} from "@testing-library/user-event";
```
jeremywiebe, line 8: By aliasing the library import, we can avoid renaming _all_ usages of the `userEvent` throughout these tests. See the `beforeEach()` below.

packages/math-input/src/components/__tests__/integration.test.tsx
```diff
@@ -91,7 +97,7 @@ describe("math input integration", () => {
         ).toBeInTheDocument();
     });
 
-    it("doesn't show the keypad initially", () => {
+    it("doesn't show the keypad initially", async () => {
```
jeremywiebe, line 100: I used some of the automation John figured out for migrating webapp and so there might be a few tests in this PR that are migrated to be async that don't actually call any async functions. In practice, this didn't seem to bother. I'm happy to roll back these if the reviewer has strong feelings. :)

packages/perseus/src/__tests__/renderer.test.tsx
```diff
@@ -260,13 +263,13 @@ describe("renderer", () => {
             });
 
             // Assert
-            expect(sawDropdown2).toBeTrue();
+            expect(sawDropdown2).toBe(true);
```
jeremywiebe, line 266: I started getting type errors on `.toBeTrue()` and `.toBeFalse()`. They appear to be from Jasmine and I'm unclear how this ever worked. 😦

packages/perseus/src/__tests__/renderer.test.tsx
```diff
@@ -1439,9 +1442,10 @@ describe("renderer", () => {
                 },
             });
 
-            screen
-                .getAllByRole("textbox")
-                .forEach((el, i) => userEvent.type(el, i.toString()));
+            const textboxes = screen.getAllByRole("textbox");
+            for (let i = 0; i < textboxes.length; i++) {
+                await userEvent.type(textboxes[i], i.toString());
+            }
```
jeremywiebe, line 1448: I just could not get `.forEach()` to work with these async calls.

packages/perseus/src/components/__tests__/zoomable.test.tsx
```diff
@@ -23,11 +22,29 @@ const mockSize = (
     }
 };
 
+// The zoomable does some measuring after the initial render to determine what
```
jeremywiebe, line 25: As I dug into these tests for the Zoomable, I discovered that they weren't really validating things correctly. I've fixed them up so that I _think_ they are now correct _and_ compatible with user-event@14

packages/perseus/src/widgets/__testdata__/numeric-input.testdata.ts
```diff
@@ -258,8 +258,8 @@ export const withCoefficient: PerseusRenderer = {
...
 
-export const question1AndAnswer: [PerseusRenderer, string, string] = [
-    question1,
-    "1252",
-    "1/2",
-];
+export const question1AndAnswer: [
+    question: PerseusRenderer,
+    correct: string,
+    incorrect: string,
```
jeremywiebe, line 264: Just added names to the tuple items. Makes it a bit clearer.

jeresig (approved):
The general pattern you've used here for user-event makes sense to me. An alternative for the beforeEach would be to change the `render()` call to have a custom function that renders and returns a new userEvent object correctly-bound to the jest fake timers - but that's possibly more work? I'm defer to others on the review of the specifics of the tests. Thanks for the user-event upgrade!

PR #1031, opened 2024-02-27
Allow higher versions of react

ErBlack:
This change is intended to enable the package to be compatible with higher versions of React beyond the currently specified version 16.14.0.

It appears that the package is compatible with higher versions than currently allowed in the `package.json` file.


jeremywiebe:
Hi there. Thank-you so much for submitting this pull request. I will be closing this PR per our [Contributing Guide](https://github.com/Khan/perseus?tab=readme-ov-file#contributing). This is a good idea, however, and I will be updating the packages in this repo to accept a broader range of React/React DOM dependencies. Note that the project is not compatible with React 17 or 18 quite yet!

PR #1033, opened 2024-02-27
Update interactive-graph-settings.tsx to React class component

nishasy:
## Summary:
This PR is pretty much the same as https://github.com/Khan/perseus/pull/1009.

The changes from that PR got rolled back, and now interactive-graph-settings.tsx
has been split out as a copy of graph-settings.tsx.

Instead of updating graph-settings.tsx, we're keeping that as is as it will
only be used by deprecated components. interactive-graph-settings.tsx will
continue to be updated along with InteractiveGraph.

To make these updates easier, we're making InteractiveGraphSettings a
React class component instead of it using the old `createReactClass` const.

- Update InteractiveGraphSettings to be a React class
- Add types
- Use arrow functions so `this` binds correctly

Additional cleanup coming in a following PR.

## Test plan:
`yarn jest packages/perseus-editor/src/components/__tests__/interactive-graph-settings.test.tsx`
`yarn typecheck`

### DIFF
--- a/.changeset/nervous-news-cheat.md
+++ b/.changeset/nervous-news-cheat.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/perseus": patch
+"@khanacademy/perseus-editor": patch
+---
+
+Udpate InteractiveGraphSettings to React class. Update `markings` to be "graph" | "grid" | "none" instead of string.
--- a/.changeset/nervous-news-cheat.md
+++ b/.changeset/nervous-news-cheat.md
@@ -3,4 +3,4 @@
 "@khanacademy/perseus-editor": patch
 ---
 
-Udpate InteractiveGraphSettings to React class. Update `markings` to be "graph" | "grid" | "none" instead of string.
+Update InteractiveGraphSettings to React class. Update `markings` to be "graph" | "grid" | "none" instead of string.
--- a/packages/perseus-editor/src/components/interactive-graph-settings.tsx
+++ b/packages/perseus-editor/src/components/interactive-graph-settings.tsx
@@ -1,7 +1,4 @@
 /* eslint-disable react/forbid-prop-types, react/no-unsafe */
-/**
- * Used in the editor for the InteractiveGraph widget.
- */
 import {
     components,
     interactiveSizes,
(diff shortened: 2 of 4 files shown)


.changeset/nervous-news-cheat.md
```diff
@@ -0,0 +1,6 @@
+---
+"@khanacademy/perseus": patch
+"@khanacademy/perseus-editor": patch
+---
+
+Udpate InteractiveGraphSettings to React class. Update `markings` to be "graph" | "grid" | "none" instead of string.
```
jeremywiebe, line 6:
```suggestion
Update InteractiveGraphSettings to React class. Update `markings` to be "graph" | "grid" | "none" instead of string.
```

packages/perseus-editor/src/components/interactive-graph-settings.tsx
```diff
@@ -1,7 +1,4 @@
 /* eslint-disable react/forbid-prop-types, react/no-unsafe */
-/**
- * Used in the editor for the InteractiveGraph widget.
- */
```
jeremywiebe, line 4: I thought this comment was quite useful.

jeremywiebe (approved):

packages/perseus-editor/src/components/interactive-graph-settings.tsx
nishasy: Oops! I think I accidentally removed it in a copy/paste. I'll add it back.

packages/perseus-editor/src/widgets/interactive-graph-editor.tsx
```diff
@@ -114,7 +114,7 @@ type Props = {
      * - grid: shows only the grid lines
      * - none: shows no markings
      */
-    markings: string; // TODO(jeremy)
+    markings: "graph" | "grid" | "none";
```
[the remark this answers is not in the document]
jeremywiebe: We should probably have a pattern where we document them in `perseus-types.ts` and then reference those prop types using indexed accessors `markings: InteractiveGraphWidgetOptions["markings"]`.  The only place that doesn't work is when the widget defines a `transform` function whereby the widget options are not the same as the React component's Props.  But I think that's rare.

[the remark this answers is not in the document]
nishasy: @nedredmond For context, I added the prop defs here so that they would show up in storybook. Maybe we could switch around some stuff and add them to `types` later!

PR #1035, opened 2024-02-28
[🔥AUDIT🔥] Fix Storybook deployed on GH Pages

jeremywiebe:
## Summary:

We just switched to using Vite with Storybook. Vite outputs files that have an underscore prefix and Github has some native Jekyll support that ignores these types of files. 

I found a Github issue where the poster found [the issue](https://github.com/storybookjs/storybook/issues/21711#issuecomment-1496464011).  That linked to a Github article explaining the behaviour. 

So this PR adds a `poststorybook-build` script to the `package.json` file which emits the `.nojekyll` file into the `storybook-static` folder before publishing. 

Issue: "none"

## Test plan:

I have no way to test this without landing and releasing it, so that's how I'll test

### DIFF
--- a/package.json
+++ b/package.json
@@ -147,6 +147,7 @@
     "storybook": "storybook dev -p 6006",
     "start": "storybook dev -p 6006",
     "build-storybook": "storybook build",
+    "post:build-storybook": "touch ./storybook-static/.nojekyll",
     "cypress": "cypress open --component",
     "cypress:ci": "cross-env BABEL_COVERAGE=1 cypress run --component --env CYPRESS_COVERAGE=1",
     "format": "prettier --write .",
--- a/package.json
+++ b/package.json
@@ -147,7 +147,7 @@
     "storybook": "storybook dev -p 6006",
     "start": "storybook dev -p 6006",
     "build-storybook": "storybook build",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -147,6 +147,7 @@
     "storybook": "storybook dev -p 6006",
     "start": "storybook dev -p 6006",
     "build-storybook": "storybook build",
+    "postbuild-storybook": "touch ./storybook-static/.nojekyll",
```
jeremywiebe, line 150: See: https://docs.npmjs.com/cli/v10/using-npm/scripts#pre--post-scripts

package.json
```diff
@@ -147,6 +147,7 @@
     "storybook": "storybook dev -p 6006",
     "start": "storybook dev -p 6006",
     "build-storybook": "storybook build",
+    "postbuild-storybook": "install -d ./storybook-static/.nojekyll",
```
jeremywiebe, line 150: We use `install -d` here because there are cases where this target might not exist (like when we publish to Chromatic where the `build-storybook` command is passed a custom output dir). In this case, we don't worry about it, but don't want the command to fail because the dir is missing.

PR #1039, opened 2024-02-28
Clarify what versions of React the project supports

jeremywiebe:
## Summary:

We received a [PR](https://github.com/Khan/perseus/pull/1031) that opened up the versions of React for simple-markdown a bit. This was a good idea so I've created a PR to clarify what versions of React we support.

When we complete the React v18 migration, we'll need to update these, but this puts the ranges in place so we're prepared.

Issue: "none"

## Test plan:

### DIFF
--- a/.changeset/plenty-forks-sing.md
+++ b/.changeset/plenty-forks-sing.md
@@ -1,0 +1,8 @@
+---
+"@khanacademy/math-input": patch
+"@khanacademy/perseus": patch
+"@khanacademy/perseus-editor": patch
+"@khanacademy/simple-markdown": patch
+---
+
+Update package.json to make supported React version ranges clear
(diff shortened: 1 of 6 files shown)


Myranae (approved):
Looks good! How do you know what the minimum version for these packages is though? Do these need to be tested?

jeremywiebe:
> Looks good! How do you know what the minimum version for these packages is though? Do these need to be tested?

The min version Perseus supports is mostly defined by what APIs we use.  In reality, our minimum should be at what webapp and mobile are using. FEI is working on moving everything to React 18. I have Khan/perseus#1064 open that moves Perseus to React 18, so I'm inclined to abort this PR as its changes would be very short-lived.

jeremywiebe:
Closing as #1064 will obviate the need for this PR once it lands (shortly!)

PR #1041, opened 2024-02-28
Fix fin calc editor bug

nedredmond:
Summary:

The financial calculator appeared to be unchecked if the component reloaded. This was caused by the `financialCalculatorOptionsExpanded` defaulting to "off" despite props. Instead, we will set it to "true" if any of the Financial Calculator Options are set to true.

Also fixes minor comment typos.

Issue: LC-1797

Testing:
Unit test that checks if expected items are present and checked passes-- did not pass before these changes.
To test in the editor experience, check Financial Calculator, click the Preview tab, then return. Financial Calculator appears to be unchecked, even though the checked state is true in the JSON data.

### DIFF
--- a/packages/perseus-editor/src/item-extras-editor.tsx
+++ b/packages/perseus-editor/src/item-extras-editor.tsx
@@ -26,18 +26,24 @@ class ItemExtrasEditor extends React.Component<Props, State> {
         zTable: false,
     };
 
+    finCalcOptions = [
+        this.props.financialCalculatorMonthlyPayment,
+        this.props.financialCalculatorTotalAmount,
+        this.props.financialCalculatorTimeToPayOff,
+    ];
+
     state = {
-        financialCalculatorOptionsExpanded: false,
+        financialCalculatorOptionsExpanded: this.finCalcOptions.some(
+            (opt) => opt === true,
+        ),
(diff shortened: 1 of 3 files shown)


packages/perseus-editor/src/item-extras-editor.tsx
```diff
@@ -26,18 +26,24 @@ class ItemExtrasEditor extends React.Component<Props, State> {
         zTable: false,
     };
 
+    finCalcOptions = [
+        this.props.financialCalculatorMonthlyPayment,
+        this.props.financialCalculatorTotalAmount,
+        this.props.financialCalculatorTimeToPayOff,
+    ];
```
jeremywiebe, line 33: This makes me a bit nervous. We're accessing props when instantiating a class field.  I know in the past we've had "race"-type errors where `this` is not properly set up during field initialization.  See: https://github.com/microsoft/TypeScript/issues/52331 (I'm not sure it's still an issue in our version of Typescript, but I think it's worth avoiding this pattern). 

Could we move this, and the `state` initialization into a constructor?

jeremywiebe (commented):
This all looks good, except for my comments about field initialization touching `this`.

PR #1048, opened 2024-03-01
And Controlled stories for InteractiveGraphSettings and InteractiveGraphEditor

nishasy:
## Summary

Add a controlled story to storybook for InteractiveGraphSettings and InteractiveGraphEditor so that their behaviors are easy to test. (As opposed to having to manually change everything in the control panel)

Instead of adding a `useState` for every prop, I used `useReducer`, which did the trick.

### Test Plan
Interactive Graph Editor controlled story
http://localhost:6006/?path=/story/perseus-editor-widgets-interactive-graph-editor--controlled

Interactive Graph Settings
http://localhost:6006/?path=/story/perseus-editor-components-interactive-graph-settings--controlled

### Interactive Graph Settings demo

https://github.com/Khan/perseus/assets/13231763/77ee738d-e1c6-4f6c-b705-f7e2b8c373b1

### Interactive Graph Editor demo

Note: The demo has some weird things like the graph disappearing or duplicating. This is not a bug introduced here, it's already like this in prod.

https://github.com/Khan/perseus/assets/13231763/6676b8b0-560e-4636-ba5e-e604ab0582c3


packages/perseus-editor/src/components/__stories__/graph-settings.argtypes.ts
```diff
@@ -50,7 +50,6 @@ export default {
             type: "select",
         },
         table: {
-            defaultValue: {summary: "graph"},
```
nishasy, line 53: Removed this everywhere because I learned that `defaultValue` under `table` is no longer the favored approach. Instead, set args directly in the stories.

jeremywiebe (approved):
Very neat!  Thanks for improving the Storybook experience for these editors.

PR #1049, opened 2024-03-01
Use Wonder Blocks in InteractiveGraphEditor

nishasy:
## Summary:

In preparation for adding a wonder blocks dropdown for adding a locked figure, 
I'm making some other components in InteractiveGraphEditor use wonder blocks
as well for the sake of consistency an accessibility.

- Use wonder-blocks-dropdown SingleSelect instead of `<select>`
- Update PropCheckBox so it uses WB Checkbox
- Use WB PhosphorIcon instead of a custom image that was indiscernible anyway 
- Use WB Typography's LabelSmall everywhere
- Update labels so they work programmatically (clicking on them also highlights/clicks on the element that the label applies to)
- Move `correct answer` down by the graph while I'm here. Requested by content authors.

Note: I don't think WB dropdown has any equivalent for `<optgroup>` yet. In the
meantime, I'm just using a horizontal rule to separate the metric and imperial
units in the "Ruler label" dropdown.

## Test plan:
`yarn jest`

Storybook
- Go to the EditorPage story http://localhost:6006/?path=/docs/perseus-editor-editorpage--docs
- Click the "Add a widget" dropdown
- Select "Interactive graph"
- Open the new "interactive-graph 1" dropdown
- Check that the checkboxes are wonder blocks checkboxes
- Confirm that the "Type of graph" dropdown is a wonder blocks dropdown
- Click the "Show ruler" checkbox
- Confirm that the "Ruler label" and "Ruler ticks" dropdowns are wonder blocks dropdowns

### Screenshot

Before is on the left. After is on the right.

Screenshot 2024-03-04 at 6 05 36 PM


packages/perseus-editor/src/components/labeled-row.tsx
nishasy, line 1: Moved over from interactive-graph-editor.tsx (it was called Row before) into its own dedicated file here. Noticed everything inside Row included FieldLabel (which wasn't even setting labels programmatically before) so I combined them here.

packages/perseus-editor/src/components/__tests__/interactive-graph-settings.test.tsx
```diff
@@ -227,7 +231,12 @@ describe("InteractiveGraphSettings", () => {
         render(<InteractiveGraphSettings onChange={onChange} />);
 
         // Act
-        const input = screen.getByRole("textbox", {name: "x Range"});
+        // Note: The textbox's `name` attribute is "x Range 10" because it's
+        // encased in a <label> element, which applies everything surrounding
+        // the first element to the first element's label. This means it
+        // includes the "10" from the second textbox. Same for the other
+        // RangeInput tests below.
+        const input = screen.getByRole("textbox", {name: "x Range 10"});
```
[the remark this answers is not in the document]
nishasy: After some more testing weirdness, I really do think we should switch to unique IDs at some point. I might do it in a following PR? Or maybe I'll wait for a larger refactor in case we can get to a point with functional components so I can use the hook for it.

packages/perseus-editor/src/components/graph-points-count-selector.tsx
```diff
@@ -1,9 +1,19 @@
+import {OptionItem, SingleSelect} from "@khanacademy/wonder-blocks-dropdown";
+import {StyleSheet} from "aphrodite";
 import * as React from "react";
 
 import {UNLIMITED, parsePointCount} from "../util/points";
 
 import type {PointValue} from "../util/points";
 
+const NUMERIC_OPTIONS = [...Array(7).keys()].map((n) => (
```
jeremywiebe, line 9: 🤷‍♂️  This feels off to generate the React components and assign them to a const here.  I haven't seen this done before so I'd defer to you if it's a pattern you've used before.

packages/perseus-editor/src/components/graph-type-selector.tsx
```diff
@@ -7,25 +9,34 @@ type GraphTypeSelectorProps = {
...
-        <select
-            value={props.graphType}
-            onChange={(e) => {
-                const type = e.target.value;
-                props.onChange(type);
+        <SingleSelect
+            selectedValue={props.graphType}
+            onChange={(newValue) => {
+                props.onChange(newValue);
             }}
```
jeremywiebe, line 16: I'm ok either way, but we can shorten this up if we want:
```suggestion
            onChange={props.onChange}
```

packages/perseus-editor/src/components/interactive-graph-settings.tsx
```diff
@@ -8,9 +8,15 @@ import {
...
     Util,
 } from "@khanacademy/perseus";
+import {View} from "@khanacademy/wonder-blocks-core";
+import {OptionItem, SingleSelect} from "@khanacademy/wonder-blocks-dropdown";
+import {color, spacing} from "@khanacademy/wonder-blocks-tokens";
+import {css, StyleSheet} from "aphrodite";
 import * as React from "react";
 import _ from "underscore";
 
+import LabeledRow from "./labeled-row";
```
jeremywiebe, line 18: Thanks for starting to extract things into components!

packages/perseus-editor/src/components/interactive-graph-settings.tsx
```diff
@@ -545,112 +559,139 @@ class InteractiveGraphSettings extends React.Component<Props, State> {
...
+                    label="Background image URL:"
+                    style={styles.resetSpaceTop}
+                >
+                    <input
+                        type="text"
+                        className={css(styles.backgroundUrlInput)}
+                        ref={this.bgUrlRef}
+                        value={this.state.backgroundImage.url || ""}
+                        onChange={(e) => {
+                            const image = _.clone(this.props.backgroundImage);
```
jeremywiebe, line 572: I think you swapped this `_.clone()` in other places for a simple object spread.

packages/perseus-editor/src/components/interactive-graph-settings.tsx
```diff
@@ -545,112 +559,139 @@ class InteractiveGraphSettings extends React.Component<Props, State> {
...
+                                <SingleSelect
+                                    id="ruler-ticks-select"
+                                    selectedValue={`${this.props.rulerTicks}`}
+                                    onChange={(newValue) => {
+                                        this.change({rulerTicks: newValue});
+                                    }}
+                                    placeholder="10"
+                                    style={styles.singleSelectShort}
+                                >
+                                    {RULER_TICKS.map((value) => (
```
jeremywiebe, line 649: If this is the only usage of `RULER_TICKS` I'd be inclined not to create a const as it adds some indirection to understand what it is.

packages/perseus-editor/src/widgets/__tests__/interactive-graph-editor.test.tsx
```diff
@@ -0,0 +1,445 @@
+import {ApiOptions, Dependencies} from "@khanacademy/perseus";
+import {render, screen} from "@testing-library/react";
+import {userEvent as userEventLib} from "@testing-library/user-event";
+import * as React from "react";
+
+import {testDependencies} from "../../../../../testing/test-dependencies";
+import InteractiveGraphEditor from "../interactive-graph-editor";
+
+const baseProps = {
```
jeremywiebe, line 9: Would it be useful to type this as `Partial<PropsFor<typeof InteractiveGraphEditor>>`?  Or is it safe just because we pass it to the component in this file anyways so type inference should keep things correct?

packages/perseus-editor/src/widgets/__tests__/interactive-graph-editor.test.tsx
```diff
@@ -0,0 +1,445 @@
...
+        const onChangeMock = jest.fn();
+
+        render(
+            <InteractiveGraphEditor {...baseProps} onChange={onChangeMock} />,
+        );
+
+        // Act
+        const dropdown = screen.getByRole("button", {name: "Type of Graph:"});
+        await userEvent.click(dropdown);
+        await userEvent.click(screen.getByRole("option", {name: "Polygon"}));
```
jeremywiebe, line 51: **thought**: If we're going to be growing our use of WB (especially the dropdown), I wonder if it'd be worth writing our own `userEvent` extension to select a WB dropdown option. 🤔

packages/perseus-editor/src/widgets/interactive-graph-editor.tsx
```diff
@@ -541,47 +516,68 @@ class InteractiveGraphEditor extends React.Component<Props> {
...
+                                Congruency requires only that the angle measures
+                                are the same. An exact match implies congruency,
+                                but also requires that the angles have the same
+                                orientation and that the vertices are in the
+                                same position.
+                            </p>
+                        </InfoTip>
+                    </LabeledRow>
                 )}
+                <LabeledRow label="Correct answer:">
```
jeremywiebe, line 547: Glad it was so easy to move this.

jeremywiebe (approved):
Cool.  One suggestion I'd have is that this might be a good place to wrap things up and deploy.  I think we want to also give the content authors a heads-up that this visual change is coming.

packages/perseus-editor/src/components/__tests__/interactive-graph-settings.test.tsx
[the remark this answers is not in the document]
nishasy: This has to do with the fact that RangeInput is actually two text fields wrapped under one label. Only the first text field receives the label, and it treats everything around it as a label (both the actual text and the second text field).

The way to get around this is to give the desired text field an ID and set the label's `htmlFor` to that ID, but this is a problem if there are two interactive graph editors because you can't have multiple IDs. The second label will point to the first instance of the text field. So what we'd need to do is use a unique ID provider. Which I _can_ do here with the unique ID HOC. Honestly I might just do that...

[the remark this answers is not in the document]
nishasy: I just really don't like using HOCs when hooks are so much better.

README.md
```diff
@@ -37,6 +37,27 @@ The components and widgets of Perseus are developed using [Storybook](https://gi
...
+2. Run the following command to update the dev dependencies and the peer dependencies.
+```
+yarn add --dev [dependency name]
+yarn add --peer [dependency name]
+
+// Example
+yarn add --dev @khanacademy/wonder-blocks-button
+yarn add --peer @khanacademy/wonder-blocks-button
+```
+
```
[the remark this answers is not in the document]
jeremywiebe: The original reason was exactly asWe've debated this in the past. I think our entire team is pretty much on the same page that our packages should use regular `dependencies` for WB/WS dependencies.  

In fact, we've created 3 tickets and a document for this. 😬 

LEMS-911, LEMS-1490, LEMS-1549, and https://khanacademy.atlassian.net/wiki/spaces/LC/pages/2473951425/Should+Perseus+dependencies+be+peer+dependencies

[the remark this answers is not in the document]
jeremywiebe: Follow-on to this. I think we could clarify that we use `peerDependencies` for dependencies that webapp is already using (which I think is all of WB, WS, and React).  But, we can probably get the same effect more safely if we just use `dependencies` and mark them as external in our bundling process.

packages/perseus-editor/src/components/__tests__/interactive-graph-settings.test.tsx
```diff
@@ -61,8 +61,10 @@ describe("InteractiveGraphSettings", () => {
         );
 
         // Act
-        const select = screen.getByRole("combobox", {name: "Ruler label:"});
-        await await userEvent.selectOptions(select, ["cm"]);
+        const select = screen.getByRole("button", {name: "Ruler label:"});
+        select.click();
```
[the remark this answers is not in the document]
nishasy: I removed the `userEvent.selectOptions` because WB dropdown isn't a combobox. 

I used `select.click()` instead of `userEvent.click()` because I was under the incorrect assumption that it was more straightforward to use click. But your question prompted me to read the userEvent docs, and now I think it makes more sense to change it. I'll bring back userEvent!

packages/perseus-editor/src/widgets/__tests__/interactive-graph-editor.test.tsx
nishasy: I think it's safe, but I'll add types anyway for clarity.

README.md
```diff
@@ -37,6 +37,32 @@ The components and widgets of Perseus are developed using [Storybook](https://gi
 
 We use [changesets](https://github.com/changesets/changesets) to help manage our versioning/releases. Before pushing a new PR, add a changeset by running `yarn changeset`. Commit and submit that with the PR.
 
+### Updating Dependencies
+
+If you want to use another library in Perseus, you will need up update the dependencies.
+Use `peerDependencies` for dependencies that webapp is already using, such as
```
jeremywiebe, line 43:
```suggestion
Use `peerDependencies` and `devDependencies` for dependencies that webapp is already using, such as
```

jeremywiebe (approved):

PR #1055, opened 2024-03-05
Fixing InteractionCallback calling too frequently 

SonicScrewdriver:
## Summary:
This PR fixes an issue where our hints or helpful popups were being immediately closed, in a glitchy manner. 

Previously I had a poor understanding of when we were intending to call the interactionCallback. Looking at several widgets (numeric-input, input-number), they had been calling the interactionCallback both on answer update AND on focus. After reviewing all the widgets, I suspect that the logic I saw in numeric-input/input-number was actually a bug and not how we want to be using the interactionCallback! 

As a result, I have removed the interactionCallback from the onFocus method in our server-item-renderer. While we will no longer be dismissing the popover **on focus** for numeric-input and input-number, I now believe that this historical functionality was unintended. 

Issue: LEMS-1804

## Test plan:
- Manually tested PR in Webapp, and can confirm that it fixes the issue.

### DIFF
--- a/.changeset/empty-impalas-watch.md
+++ b/.changeset/empty-impalas-watch.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": minor
+---
+
+Fixing the frequency that we're calling interactionCallback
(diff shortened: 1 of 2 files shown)


handeyeco (approved):
Cool, let's try it out!

jeremywiebe (approved):

PR #1058, opened 2024-03-06
Move enum logic from Webapp to Perseus

handeyeco:
## Summary:
In [this PR](https://github.com/Khan/webapp/pull/20014) I added some logic in Webapp knowing I would need to move it to Perseus; now that I'm trying to add the CEDAR event to Mobile, the time has come.

---

### I didn't get any ideas on how to implement the below and I need to move on, so I'm going to move forward with the original plan.

I'm wondering if this could be done in a cleaner way. In each widget we have an export; for example this is Radio's:

``` TS
export default {
    name: "radio", // <- type: string
    displayName: "Radio / Multiple choice",
    accessible: true,
    widget: Radio,
    transform: transform,
    staticTransform: transform,
    version: {major: 1, minor: 0},
    propUpgrades: propUpgrades,
    isLintable: true,
} as WidgetExports<typeof Radio>;
```

We also do something like this in `perseus-types.ts`:

``` TS
export type RadioWidget = Widget<'radio', PerseusRadioWidgetOptions>;

type Widget<Type extends string, Options> = {
    // The "type" of widget which will define what the Options field looks like
    type: Type;

    // ...etc...
};
```

Then:

``` TS
export type PerseusWidgetsMap = {
    // ...etc...
} & {
    [key in `radio ${number}`]: RadioWidget;
} & {
    // ...etc...
```

It feels like we're having to maintain a lot of places where there are lists of widgets; now this would add even more. Is there a way to derive this type information from the widgets themselves? For instance:

``` TS
// in each widget file
export default {
    name: "radio",
    enum: "RADIO",
    // ...etc...
} as WidgetExports<typeof Radio>;

// in a shared file
const perseusWidgets = [
    RadioWidget,
    LightPuzzleWidget,
    // ...etc...
]

// Could theses be derived types?
const PerseusWidgetNames = perseusWidgets.map(w => w.name) as const;
const PerseusWidgetEnums = perseusWidgets.map(w => w.enum) as const;
```

That way the widgets themselves are responsible for providing their own name and enum values? And we have one master list of widgets that is used to derive other lists?

Issue: MOB-5441


packages/perseus/src/util/widget-enum-utils.ts
```diff
@@ -0,0 +1,153 @@
...
+const widgetNameToEnum: Record<WidgetName, WidgetEnum> = {
+    categorizer: "CATEGORIZER",
+    "cs-program": "CS_PROGRAM",
+    definition: "DEFINITION",
+    "deprecated-standin": "DEPRECATED_STANDIN",
+    dropdown: "DROPDOWN",
+    explanation: "EXPLANATION",
+    expression: "EXPRESSION",
+    "graded-group-set": "GRADED_GROUP",
+    "graded-group": "GRADED_GROUP_SET",
```
nedredmond, line 97: Are these reversed?

handeyeco: Oh that's weird. That's copy/pasted from prod webapp which means it's live on the site. 🤔 Not that I think we care all that much about graded-group statistics (compared to more interactive widgets).

packages/perseus/src/util/widget-enum-utils.ts
```diff
@@ -0,0 +1,153 @@
...
+export function convertWidgetNameToEnum(name: WidgetName): WidgetEnum {
+    const widgetEnum = widgetNameToEnum[name];
+
+    if (!widgetEnum) {
+        throw new Error(`Unknown widget name: ${name}`);
+    }
+
+    return widgetEnum;
+}
+
```
SonicScrewdriver, line 138: Oh lovely this can replace that function in the coach report utils! :)

SonicScrewdriver (approved):
Wonderful, thank you!

PR #1060, opened 2024-03-06
Accessibility and Usability Enhancements for Explanation Widget

mark-fitzgerald:
## Summary:
The Explanation widget should be consistent regardless of its content type, as well as across device platforms. Therefore, the `<Clickable>` element is now changed to a Wonder Blocks `<Button>`. Additionally, minor custom formatting that was tied to the content type (i.e. article) or the platform (i.e. mobile) has been removed.

The widget should also follow accessibility best practices. The attribute `aria-controls` has been added to the button to indicate what will be affected when it is toggled. Although [this attribute is not currently being used](https://github.com/w3c/aria/issues/995) by assistive technologies, it is [highly recommended by the ARIA working group](https://github.com/w3c/aria/issues/995#issuecomment-600643723) that this attribute be applied in cases such as this. Therefore, in addition to the attribute being added to the button, the content that is shown/hidden by the button is now rendered in the DOM at all times in order to be referenced by the `aria-controls` attribute.

Interactions with the widget should respond in an intuitive manner. When activation of an element causes something else to happen on the page, an [animation can ease the cognitive aspects](https://www.nngroup.com/articles/animation-usability/) of following that change. Therefore, CSS transitions have been applied to properties of the content wrapper to change the show/hide action to be a reveal/conceal action (see the Before and After videos for a comparison). In order to honor the preferences of the user, the widget checks the `prefers-reduced-motion` media preference before applying the transitions.

Other interesting changes:
* Refactored the widget to take advantage of the common visuals across use cases.
   * Removed unneeded logic branches.
   * Removed unneeded styling.
   * Removed unneeded unit tests.
   * Moved styling logic outside of the JSX block.
* Added a new Storybook story (with new test data) to show a normal use case.
* Fixed unit tests for Graded Group widget which improperly referenced the Explanation widget.
   * Changed the test data to use the GG widget's internal Explain/Hide button.
   * Updated the snapshots.
* Added UI regression tests to guard against CSS mishaps that alter keyboard interactions.

Issues: LP-5838, LEMS-199, LEMS-259, LP-11557 (possibly), LEMS-277, LC-212, MOB-5434 (possibly)

## Test plan:
1. Open the Perseus Storybook (locally: http://localhost:6006/?path=/story/perseus-widgets-explanation--question-1).
   * Notice the change in appearance of the "Explanation" button.
1. Activate the "Explanation" button (either by clicking it or by tabbing to it and pressing Enter/Spacebar keys).
   * Notice that the widget content is revealed via an animation.
   * Activating the button (now showing "Hide") again causes the content to be concealed, but slightly faster than when it is revealed.
1. Change your accessibility settings to prefer reduced motion.
   * In MacOS, open System Settings => Accessibility => Display => Reduce motion
1. Activate the "Explanation" button again.
   * Notice that the widget content is shown without any animation.
   * Don't forget to reset your accessibility settings.

### Before - Current UX

https://github.com/Khan/perseus/assets/13896410/dcb1fcdc-c4cd-4e01-b342-e4a7f39ba583

### After - New UX

https://github.com/Khan/perseus/assets/13896410/a8622fba-c61e-4107-9a87-80fe1f6829bb


packages/perseus/src/widgets/__testdata__/graded-group.testdata.ts
```diff
@@ -34,40 +34,31 @@ export const question1: PerseusRenderer = {
...
-                                            "Some chemosynthetic bacteria introduce energy and fixed carbon into communities where photosynthesis is not possible (e.g., deep-sea vents).",
-                                            "Some bacteria live symbiotically inside of host organisms and provide the host with nutrients.",
-                                        ],
-                                        categories: ["True", "False"],
-                                        values: [0, 1, 0, 0],
-                                        randomizeItems: false,
-                                    },
-                                    version: {major: 0, minor: 0},
-                                },
+                },
```
jeremywiebe, line 37: Thanks for fixing up this test data!

packages/perseus/src/widgets/__tests__/explanation.cypress.ts
```diff
@@ -0,0 +1,75 @@
+import renderQuestion from "../../../../../testing/render-question-with-cypress";
```
jeremywiebe, line 1: **question**: What was the reason for needing Cypress tests here?  FWIW: we prefer @testing-library (RTL) tests over Cypress except in cases where we can't test a component with RTL.  Cypress tests are much slower and clumsier to write/work with so that's why.

packages/perseus/src/widgets/explanation.tsx
```diff
@@ -71,85 +71,75 @@ class Explanation extends React.Component<Props, State> {
...
+            this.state.expanded
+                ? styles.contentExpanded
+                : styles.contentCollapsed,
+            allowTransition &&
+                (this.state.expanded
+                    ? styles.transitionExpanded
+                    : styles.transitionCollapsed),
+        );
+
+        const contentTestId =
```
jeremywiebe, line 103: **thought**: We typically haven't prevented test IDs in prod code. In my mind, it's not a big deal if we have test IDs in prod markup. It also removes the need for the test where we alter the `NODE_ENV` variable, which will be problematic when running tests in parallel.

packages/perseus/src/widgets/explanation.tsx
```diff
@@ -71,85 +71,75 @@ class Explanation extends React.Component<Props, State> {
...
+                            endIcon={caretIcon}
+                            kind="tertiary"
+                            onClick={this._onClick}
+                            size="large"
+                            style={buttonStyleOverrides}
+                        >
+                            {promptText}
+                        </Button>
+
+                        <div
```
jeremywiebe, line 125: Today, we mostly prefer using a Wonder Blocks [`View`](https://khan.github.io/wonder-blocks/?path=/docs/core-view--docs) (which is basically a `<div />`).  This then means you don't need to use the `Aphrodite` `css()` function and can pass the styles directly to the `<View style={} />` prop.

jeremywiebe: Also, I realize that the original content used `<div />`'s, so if this change is too dramatic, I'm ok with you skipping the change to using a `<View />`.

packages/perseus/src/widgets/explanation.tsx
```diff
@@ -71,85 +71,75 @@ class Explanation extends React.Component<Props, State> {
...
-        }
-
-        const expandedStyle = isMobile
-            ? styles.contentExpandedMobile
-            : styles.contentExpanded;
+        const caretIcon = this.state.expanded ? caretUp : caretDown;
+
+        const allowTransition = window.matchMedia(
+            "(prefers-reduced-motion: no-preference)",
+        ).matches;
```
jeremywiebe, line 82: :til: !

packages/perseus/src/widgets/__tests__/explanation.test.ts
```diff
@@ -41,167 +85,184 @@ describe("Explanation", function () {
...
+
+        // Act - call the widget's "change" function
+        widget.change("foo", "bar", callbackMock);
+
+        // Assert
+        expect(changeMock.mock.contexts[0]).toEqual(widget);
+        expect(changeMock).toHaveBeenCalledWith("foo", "bar", callbackMock);
+
+        // Clean-up
+        changeMock.mockRestore();
```
jeremywiebe, line 265: Our Jest tests automatically do this, so this is unnecessary. 

```suggestion
```

packages/perseus/src/widgets/__tests__/explanation.test.ts
```diff
@@ -41,167 +85,184 @@ describe("Explanation", function () {
...
 
-        // Assert
-        expect(screen.queryByText("This is an explanation")).toBeNull();
+        // Assert - transition when concealing
+        expect(screen.getByTestId("content-container").className).not.toContain(
+            "transitionCollapsed",
+        );
+
+        // Clean-up
+        matchMediaMock.mockRestore();
```
jeremywiebe, line 243: Jest should be cleaning up all mocks at the end of each test, so this isn't needed.

```suggestion
```

packages/perseus/src/widgets/__tests__/explanation.test.ts
```diff
@@ -41,167 +85,184 @@ describe("Explanation", function () {
...
 
-        // Assert
-        expect(screen.queryByText("This is an explanation")).toBeNull();
+        // Assert - transition when concealing
+        expect(screen.getByTestId("content-container").className).toContain(
+            "transitionCollapsed",
+        );
+
+        // Clean-up
+        matchMediaMock.mockRestore();
```
jeremywiebe, line 207: Jest should be cleaning up all mocks at the end of each test, so this isn't needed.

```suggestion
```

packages/perseus/src/widgets/__tests__/explanation.test.ts
```diff
@@ -41,167 +85,184 @@ describe("Explanation", function () {
...
+        //     classes and aria that are applied.
         expect(container).toMatchSnapshot("expanded");
     });
 
-    it("should snapshot for mobile", async () => {
-        // Arrange and Act
-        const {container} = renderQuestion(question1, {
-            isMobile: true,
-        });
+    it("does NOT have a test ID when not in a test environment", async () => {
```
jeremywiebe, line 93: **thought**: I mentioned this later on also, but I think it's overkill to remove test ids for prod. To date, I'm not aware of any Khan Academy code that suppresses test IDs in prod, so I think we'd want to talk about it as a team before we start doing that.

jeremywiebe (commented):
I've reviewed, but want to leave this open just so others have a chance to see/review this PR. There's some good stuff in here. This is awesome work for your first major contribution to Perseus!!! Thank-you!

packages/perseus/src/widgets/explanation.tsx
benchristel: `View`s differ from divs mainly by having `display: flex; flex-direction: column;` by default. This div uses `display: inline-grid`, so the View defaults get overridden. I think it's fine to use a div here.

packages/perseus/src/widgets/__tests__/explanation.cypress.ts
```diff
@@ -0,0 +1,75 @@
...
+        cy.waitUntil(() =>
+            cy
+                .focused()
+                .siblings()
+                .first()
+                .should("have.attr", "aria-hidden", "true")
+                .should("have.css", "visibility", "hidden"),
+        );
+        cy.realPress("Tab");
+        cy.focused().should("have.text", postContentLinkText);
```
benchristel, line 73: It kind of feels like this test is testing too much. We prefer to have each test focus on a single behavior, so that when it fails you can easily tell what's broken.

Since the collapse behavior is covered by unit tests, I don't think we need to test it here. Is there a need to test the backwards tab order? It's the browser default, and we'd have to go out of our way to break it. Testing the forward tab order should be sufficient IMO.

packages/perseus/src/widgets/__tests__/explanation.cypress.ts
mark-fitzgerald: RTL is unable to test the styling that is applied to a given element. These tests are intended to verify that the styling that affects tab order do not regress. I will add comments to give the reasoning.

packages/perseus/src/widgets/explanation.tsx
mark-fitzgerald: Removed.

packages/perseus/src/widgets/explanation.tsx
mark-fitzgerald: I'm cool with switching to `<View>` since it helps with consistency in using WB.

packages/perseus/src/widgets/__tests__/explanation.test.ts
mark-fitzgerald: Removed.

packages/perseus/src/widgets/__tests__/explanation.cypress.ts
mark-fitzgerald: I broke out the "verify collapsing still works" portion of the 2nd test to it's own test to reduce the number of behaviors being evaluated in that test.

Call me paranoid, but I've seen situations where event handling works fine for "forward" tabbing, but messes things up when "reverse" tabbing. We don't have event handling happening here, but if we add it in the future, I think it would be good to confirm that we didn't regress. Also, since [LEMS-199](https://khanacademy.atlassian.net/browse/LEMS-199) references moving backwards and having problems (though that is a screen reader action, not a keyboard tab), it was fresh in my mind as something to check.

packages/perseus/src/widgets/__tests__/explanation.cypress.ts
```diff
@@ -4,26 +4,27 @@ import * as Dependencies from "../../dependencies";
 import * as Perseus from "../../index";
 import {ipsumExample} from "../__testdata__/explanation.testdata";
 
+//  NOTE: The regression tests in this file use Cypress because they are intended to validate styling that is applied.
+//        Since React Testing Library isn't applying the CSS to the elements,
+//          we can't use Jest to verify that some keyboard interactions work properly.
```
jeremywiebe, line 9: Nice. Thanks for adding this.

jeremywiebe (approved):

PR #1062, opened 2024-03-07
Debounce Perseus Math Input Updates

nedredmond:
## Summary

Accounts for a bug in the editor in consuming webapp code, which uses the `facebook/flux` library and cannot handle immediate updates.

The new, accessible MathQuill update comes with a feature that reformats/disambiguates certain TeX and immediately updates itself. For example, `x^2` immediately updates to `x^{2}`.

### Issue
LEMS-1805

### Testing
Open any exercise with multiple "malformed" TeX in the editor. There are examples in [the ticket](https://khanacademy.atlassian.net/browse/LEMS-1805). The editor should load without `flux` throwing errors and halting the render.

### DIFF
--- a/packages/perseus/src/components/math-input.tsx
+++ b/packages/perseus/src/components/math-input.tsx
@@ -157,7 +157,7 @@ class MathInput extends React.Component<Props, State> {
                 (baseConfig) => ({
                     ...baseConfig,
                     handlers: {
-                        edit: (mathField: MathFieldInterface) => {
+                        edit: debounce((mathField: MathFieldInterface) => {
                             // This handler is guaranteed to be called on change, but
                             // unlike React it sometimes generates false positives.
                             // One of these is on initialization (with an empty string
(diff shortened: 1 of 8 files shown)


packages/perseus/src/components/math-input.tsx
```diff
@@ -399,6 +399,21 @@ const MathInputIcon = ({hovered, focused, active}) => {
     );
 };
 
+const debounce = <T extends unknown[]>(
```
handeyeco, line 402: I imagine we're defining our own debounce function because we're trying to reduce our usage of underscore and because this will provide better types.

I'd advocate moving this out of `math-input` and into `perseus/src/util.ts` or one of the files in `perseus/src/util`, but I'm not going to lose sleep if we keep it here for now.

handeyeco (approved):
Thanks for figuring this out so fast! LGTM

packages/perseus/src/components/math-input.tsx
nedredmond: It didn't even occur to me to try to use underscore, to be honest. That implementation is probably fine! Switching now.

nedredmond: jk, underscore's debounce causes a million test failures.

PR #1064, opened 2024-03-12
React 18

jeremywiebe:
## Summary:

~I've got the changes ready, but we need to wait on WB and webapp to be ready before I land this PR and do a release.~ **We're now ready!**

This PR moves this repo to use React 18. There were a few changes that needed to be made to be fully compatible with React 18.

  * Migrate a few tests from the old `ReactDOM.TestUtils` renderer to @testing-library.
  * Update some snapshots as the snapshot renderer seems to have slightly different whitespace settings.
  * Fixes to the `ErrorBoundary` tests as React's `componentDidCatch` provides a different stack trace format (which is irrelevat to our code).
  * Migrate direct uses of `ReactDOM.render()` to `ReactDOMClient.createRoot()` (we hope to move these to use Portals in the future, but I couldn't get them working).
  * Move all `devDependencies` in the root `package.json` to be `dependencies` (`import/no-extraneous-dependencies` was complaining about `@types/react` being in `devDependencies`). Since the root package.json is just a host for the workspace and doesn't build any production code, the distinction between `dependencies` and `devDependencies` is irrelevant.
  * Force Storybook to use `react-docgen` instead of the default `react-docgen-typescript` (`react-docgen` is the default in Storybook 8 so we can revert this once we upgrade).  See [this](https://github.com/storybookjs/storybook/issues/24665#issuecomment-1792084305) comment that pointed to the fix.

Issue: LEMS-1802

## Test plan:

`yarn test` ✅ 
`yarn tsc` ✅
`yarn storybook` ✅ - I sampled many of the stories. I also checked the stories that were affected by the move from `ReactDOM.render()` to `ReactDOMClient.createRoot()`


benchristel:
@jeremywiebe is this PR still needed?

jeremywiebe:
> @jeremywiebe is this PR still needed?

@benchristel Yes, but it's waiting on FE Infra work to move WB and Webapp to React 18 before I feel comfortable landing it. Ultimately, webapp (or any host app) defines what version of React Perseus uses so I think we could get into trouble if we move to React 18 and accidentally use some API/feature that is only available in React 18+.

benchristel:
@jeremywiebe React 18 is deployed in webapp, so I think this PR is now unblocked!

package.json
```diff
@@ -152,8 +148,5 @@
         "cypress:ci": "cross-env BABEL_COVERAGE=1 cypress run --component --env CYPRESS_COVERAGE=1",
         "format": "prettier --write .",
         "typecheck": "tsc"
-    },
-    "dependencies": {
-        "tiny-invariant": "^1.3.1"
```
jeremywiebe, line 157: Now that we only have `dependencies`, this has moved up into the main body of dependencies above.

handeyeco:
Just out of curiosity, do you have a sense for what would have caused a jump to the `math-input` size?

Screenshot 2024-07-17 at 2 10 54 PM

PR #1066, opened 2024-03-12
Dev Tooling: Better mobile support

jeremywiebe:
## Summary:

This PR improves our mobile support for our dev tooling (both Storybook and the custom Dev UI):

  * Adds the `perseus-mobile` class to the parent div of the renderers. Some of our Perseus styling depends on an ancestor node having the `perseus-mobile` class in order to render for mobile.
  * Adds a mobile switch to the `RendererWithDebugUI` component so we can manually flip to mobile rendering in Storybook (already exists in the Dev UI)

Finally, I've changed the`SideBySide` component slightly so that the right panel will move below the left panel on narrower screens. This will make inspecting a story with mobile emulation on much cleaner (the right panel, usually the Perseus JSON, would overlap the left panel previously).

### Dev UI (with mobile)

image

### Storybook (with mobile) 

image

### Inspecting a Story with mobile emulation enabled

image

Issue: "none"

## Test plan:

### DIFF
--- a/dev/gallery.tsx
+++ b/dev/gallery.tsx
@@ -107,7 +107,10 @@ export function Gallery() {
                 </nav>
             </header>
             <main className={css(styles.main)}>
-                <View style={styles.cards}>
+                <View
+                    style={styles.cards}
+                    className={isMobile ? "perseus-mobile" : ""}
+                >
                     {questions.map((question, i) => (
                         <QuestionRenderer
                             key={i}
--- a/testing/renderer-with-debug-ui.tsx
+++ b/testing/renderer-with-debug-ui.tsx
@@ -1,8 +1,10 @@
 /* eslint-disable @khanacademy/ts-no-error-suppressions */
 import Button from "@khanacademy/wonder-blocks-button";
-import {View} from "@khanacademy/wonder-blocks-core";
+import {PhosphorIcon} from "@khanacademy/wonder-blocks-icon";
 import {Strut} from "@khanacademy/wonder-blocks-layout";
+import Switch from "@khanacademy/wonder-blocks-switch";
 import {HeadingSmall} from "@khanacademy/wonder-blocks-typography";
+import deviceMobile from "@phosphor-icons/core/regular/device-mobile.svg";
 import * as React from "react";
 import ReactJson from "react-json-view";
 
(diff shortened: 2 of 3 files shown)


dev/gallery.tsx
```diff
@@ -107,7 +107,10 @@ export function Gallery() {
                 </nav>
             </header>
             <main className={css(styles.main)}>
-                <View style={styles.cards}>
+                <View
+                    style={styles.cards}
+                    className={isMobile ? "perseus-mobile" : ""}
```
jeremywiebe, line 112: This is the magic to flip Perseus into mobile rendering for some widgets (used in our `.less` files mostly).

testing/renderer-with-debug-ui.tsx
```diff
@@ -28,22 +31,42 @@ export const RendererWithDebugUI = ({
...
+            leftTitle={
+                <View
+                    style={{
+                        flexDirection: "row",
+                        alignItems: "center",
+                        width: "100%",
+                    }}
+                >
+                    Widget
+                    <View style={{marginLeft: "auto"}}>
```
jeremywiebe, line 47: This was the only way I could figure out pushing the switch to the right side.  `alignSelf: 'end'` didn't work. 😞

benchristel: `marginLeft` is a fine solution. Another thing I often do in flexbox layouts is add a blank element between the left-aligned and right-aligned things, and set it to `flexGrow: 1` so it expands to fill the space.

testing/side-by-side.tsx
```diff
@@ -39,15 +40,18 @@ const SideBySide = ({
...
     sideBySide: {
         display: "flex",
+        flexWrap: "wrap",
         flexDirection: "row",
     },
     leftPanel: {
         paddingRight: "30px",
-        flexGrow: 1,
+        flexBasis: "400px",
+        margin: Spacing.medium_16,
```
benchristel, line 49: Do we need both padding and margin on this element?

testing/side-by-side.tsx
```diff
@@ -39,15 +40,18 @@ const SideBySide = ({
...
     },
     rightPanel: {
         flexGrow: 1,
         maxWidth: "50%",
+        flexBasis: "400px",
```
benchristel, line 54: I wonder if this `400px` magic number should be extracted to a constant. I suspect there are other places where we reference 400px, too.

benchristel (approved):
LGTM! Thanks!

I left a couple comments inline.

testing/side-by-side.tsx
jeremywiebe: It wasn't working as I intended so I've tweaked the styles to be simpler.

testing/side-by-side.tsx
jeremywiebe: I found a few places that had a 400 constant already.  I've refactored to use one of them.  I think 400 is a graph constant mostly, but for our stories I think it mostly makes sense for now.

PR #1067, opened 2024-03-12
View a locked point with mafs

nishasy:
## Summary:
The ability to view a locked points was added as part of
[LEMS-1702](https://khanacademy.atlassian.net/browse/LEMS-1702), but now that we've
switched to mafs, we need to implement viewing a point again with mafs.

- Add a layer for locked figures to the mafs graph
- Createa a new component for said locked feature
- Add an interactive graph story to confirm that locked points can be viewed

Issue: https://khanacademy.atlassian.net/browse/LEMS-1702

## Test plan:
- Open http://localhost:6006/?path=/story/perseus-widgets-interactive-graph--segment-with-mafs-and-locked-points
- Confirm that red points are there at (-7, -7) and (2, -5)
- Drag the points of the line segment over the locked points and confirm that
  the blue points sit on top of the red points. (Check the interactive layer
  is on top of the locked layer.)

[LEMS-1702]: https://khanacademy.atlassian.net/browse/LEMS-1702?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### Story

Screenshot 2024-03-12 at 3 17 13 PM

### DIFF
--- a/packages/perseus/src/perseus-types.ts
+++ b/packages/perseus/src/perseus-types.ts
@@ -645,7 +645,7 @@ export type PerseusInteractiveGraphWidgetOptions = {
     correct: PerseusGraphType;
     // Shapes (points, chords, etc) displayed on the graph that cannot
     // be moved by the user.
-    lockedFigures?: ReadonlyArray<LockedFigure>;
+    lockedFigures?: Array<LockedFigure>;
 };
 
 export type LockedFigure = LockedPoint;
--- a/packages/perseus/src/perseus-types.ts
+++ b/packages/perseus/src/perseus-types.ts
@@ -645,7 +645,7 @@ export type PerseusInteractiveGraphWidgetOptions = {
     correct: PerseusGraphType;
     // Shapes (points, chords, etc) displayed on the graph that cannot
     // be moved by the user.
-    lockedFigures?: Array<LockedFigure>;
+    lockedFigures?: ReadonlyArray<LockedFigure>;
 };
 
 export type LockedFigure = LockedPoint;
--- a/packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
+++ b/packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
@@ -193,7 +193,7 @@ describe("locked layer", () => {
         // eslint-disable-next-line testing-library/no-container, testing-library/no-node-access
         const points = container.querySelectorAll(
             // Filter out the interactive points' circles
-            "circle:not([class*='mafs-movable-point'])",
+            "circle:not([class*='movable-point'])",
         );
 
         // Act
@@ -212,7 +212,7 @@ describe("locked layer", () => {
         // Act
         // eslint-disable-next-line testing-library/no-container, testing-library/no-node-access
         const points = container.querySelectorAll(
-            "circle:not([class*='mafs-movable-point'])",
+            "circle:not([class*='movable-point'])",
         );
 
         // Assert
--- a/packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
+++ b/packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
@@ -193,7 +193,7 @@ describe("locked layer", () => {
         // eslint-disable-next-line testing-library/no-container, testing-library/no-node-access
         const points = container.querySelectorAll(
             // Filter out the interactive points' circles
-            "circle:not(.mafs-movable-point):not(.mafs-movable-point-hitbox):not(.mafs-movable-point-focus):not(.mafs-movable-point-ring):not(.mafs-movable-point-point)",
+            "circle:not([class*='mafs-movable-point'])",
         );
 
         // Act
@@ -212,7 +212,7 @@ describe("locked layer", () => {
         // Act
         // eslint-disable-next-line testing-library/no-container, testing-library/no-node-access
         const points = container.querySelectorAll(
-            "circle:not(.mafs-movable-point):not(.mafs-movable-point-hitbox):not(.mafs-movable-point-focus):not(.mafs-movable-point-ring):not(.mafs-movable-point-point)",
+            "circle:not([class*='mafs-movable-point'])",
         );
 
         // Assert
--- a/packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
+++ b/packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
@@ -1,4 +1,5 @@
 import {describe, beforeEach, it} from "@jest/globals";
+import {color} from "@khanacademy/wonder-blocks-tokens";
 import {waitFor} from "@testing-library/react";
 import {userEvent as userEventLib} from "@testing-library/user-event";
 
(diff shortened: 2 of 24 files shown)


packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -94,7 +95,19 @@ export const MafsGraph = React.forwardRef<
...
                     height={height}
                 >
+                    {/* Background layer */}
                     {!legacyGrid && <Grid {...props} />}
+
+                    {/* Locked layer */}
+                    {props.lockedFigures && (
+                        <MafsLockedLayer
+                            // Make a copy of lockedFigures since the prop
+                            // type is readonly.
```
benchristel, line 105: Is `lockedFigures` still readonly? The type is `Array<LockedFigure>`, so I think you should be able to use the array without copying it.

packages/perseus/src/widgets/interactive-graphs/mafs-locked-layer.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+                /**
+                 * Devlopment-time future-proofing: This `never` should
+                 * fail during type-checking if we add a new locked
+                 * shape type and forget to handle it in any other
+                 * switch case here.
+                 */
+                // eslint-disable-next-line @typescript-eslint/no-unused-vars
+                const exhaustiveCheck: never = figure.type;
+                throw new Error(`Unknown locked shape type: ${figure.type}`);
```
benchristel, line 35: You can use `UnreachableCaseError` from `wonder-stuff-core` to do the exhaustiveness check in one line:

```ts
throw new UnreachableCaseError(figure.type);
```

benchristel (commented):
The code looks good! The graph in the screenshot doesn't look quite right though—I wonder if you need to import the Mafs CSS? Something like this:

```ts
import "mafs/core.css";
import "./mafs-styles.css";
```

I also left a couple suggestions inline.

packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
```diff
@@ -178,3 +180,43 @@ describe("segment graph", () => {
...
+        const {container} = renderQuestion(
+            segmentWithLockedPointsQuestion,
+            apiOptions,
+        );
+
+        // eslint-disable-next-line testing-library/no-container, testing-library/no-node-access
+        const points = container.querySelectorAll(
+            // Filter out the interactive points' circles
+            "circle:not(.mafs-movable-point):not(.mafs-movable-point-hitbox):not(.mafs-movable-point-focus):not(.mafs-movable-point-ring):not(.mafs-movable-point-point)",
+        );
```
nishasy, line 197: This is absolutely not ideal. Would love to pair with someone if they know a better to way to find these circles.

mark-fitzgerald: Would this work: `"circle:not([class*='mafs-movable-point'])"`

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
nishasy: The problem is that `lockedFigures` is readonly in perseus-types.ts. I can. change it to not be readonly though

packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
nishasy: @mark-fitzgerald YES!

packages/perseus/src/perseus-types.ts
```diff
@@ -645,7 +645,7 @@ export type PerseusInteractiveGraphWidgetOptions = {
     correct: PerseusGraphType;
     // Shapes (points, chords, etc) displayed on the graph that cannot
     // be moved by the user.
-    lockedFigures?: ReadonlyArray<LockedFigure>;
+    lockedFigures?: Array<LockedFigure>;
```
jeremywiebe, line 648: Ideally, the types here should remain read-only. These are the public API of Perseus and represent the data passed into Perseus. As such, I think we should treat them as immutable and copy them first if we want to mutate them.

Why did we need to make it mutable?

packages/perseus/src/widgets/interactive-graphs/mafs-locked-layer.tsx
```diff
@@ -0,0 +1,41 @@
+import {UnreachableCaseError} from "@khanacademy/wonder-stuff-core";
+import {Point} from "mafs";
+import * as React from "react";
+
+import type {LockedFigure} from "../../perseus-types";
+
+type Props = {
+    lockedFigures: Array<LockedFigure>;
```
jeremywiebe, line 8: In most places in Perseus we type array props as `ReadonlyArray<>`. If you update that here, I think that'll remove the need to make adjustments to `perseus-types.ts`.

packages/perseus/src/widgets/interactive-graphs/mafs-locked-layer.tsx
```diff
@@ -0,0 +1,41 @@
+import {UnreachableCaseError} from "@khanacademy/wonder-stuff-core";
```
jeremywiebe, line 1: :til: !

jeremywiebe (approved):
This is really cool...especially how little code you had to write!

packages/perseus/src/perseus-types.ts
nedredmond: +1!

packages/perseus/src/widgets/interactive-graphs/mafs-locked-layer.tsx
```diff
@@ -0,0 +1,41 @@
...
+type Props = {
+    lockedFigures: Array<LockedFigure>;
+};
+
+const MafsLockedLayer = (props: Props) => {
+    const {lockedFigures} = props;
+    return (
+        <>
+            {lockedFigures.map((figure, index) => {
+                if (figure.type === "point") {
```
nedredmond, line 16: non-blocking: You mention switch cases in the comment below, and the custom error references "case". Plus, we use switch cases elsewhere in the Mafs code for similar implementations. Why not make this a switch case?

packages/perseus/src/widgets/interactive-graphs/mafs-locked-layer.tsx
nedredmond, line 1: super-tiny-nitpick: Other files that render Mafs internals (like every graph type) do not include Mafs in the name. I think this may be a little redundant, but again, tiny nitpick, not blocking, not a hill to die on, not even a hill at all.

nedredmond (approved):
+1 to @jeremywiebe's comment about keeping props readonly. Otherwise, looks great!

packages/perseus/src/perseus-types.ts
nishasy: Ah, I changed it so it's more straightforward here:

https://github.com/Khan/perseus/pull/1067#discussion_r1522306486

But I hear you on the public API needing to be immutable. I'll go back to the first approach.

packages/perseus/src/widgets/interactive-graphs/mafs-locked-layer.tsx
nishasy: Ah! That would do it.

PR #1068, opened 2024-03-13
Style movable points and segments on Mafs graphs

benchristel:
## Summary:
See this slack thread for designs and discussion:
https://khanacademy.slack.com/archives/C067UM1QAR4/p1709847469823399

Issue: LEMS-1672

Test plan:

```
yarn dev
open http://localhost:5173
```

Check the "segment" box in the dropdown at the top of the screen to
switch to the Mafs version of the segment graph.

You should be able to move the points and line segment with the mouse
and keyboard. The hover and focus states of the points should be clearly distinct.
The line segment should be highlighted blue when focused.

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/movable-point.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/movable-point.tsx
@@ -1,0 +1,52 @@
+import Color from "@khanacademy/wonder-blocks-color";
+import {vec, useMovable, useTransformContext} from "mafs";
+import * as React from "react";
+import {useRef} from "react";
+
+type Props = {
+    point: vec.Vector2;
+    onMove: (newPoint: vec.Vector2) => unknown;
+};
+
+export const MovablePoint = (props: Props) => {
+    const hitboxRef = useRef<SVGCircleElement>(null);
+    const {point, onMove} = props;
+
+    useMovable({
+        gestureTarget: hitboxRef,
+        point,
+        onMove,
+        constrain: (p) => p,
+    });
+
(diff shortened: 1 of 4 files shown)


packages/perseus/src/widgets/interactive-graphs/movable-point.tsx
```diff
@@ -0,0 +1,52 @@
...
+                    "--movable-point-color": Color.blue,
+                } as any
+            }
+        >
+            <circle className="movable-point-hitbox" r={30} cx={x} cy={y} />
+            <circle className="movable-point-halo" r={11} cx={x} cy={y} />
+            <circle className="movable-point-ring" r={8} cx={x} cy={y} />
+            <circle
+                className="movable-point-center"
+                r={6}
```
nishasy, line 45: Would it make sense to break out these scalar numbers into a `constants` file? I'm not entirely sure since it doesn't look like there are a lot of reused ones (between this file and the CSS file), but I think naming them could make it easier to figure out why the sizes are what they are in the future.

nishasy (approved):
Very cool!

packages/perseus/src/widgets/interactive-graphs/movable-point.tsx
```diff
@@ -38,11 +38,10 @@ export const MovablePoint = (props: Props) => {
             }
         >
             <circle className="movable-point-hitbox" r={30} cx={x} cy={y} />
```
nishasy, line 40: **Optional suggestion**: Maybe also move the hitbox radius to the styles for consistency?

PR #1070, opened 2024-03-13
Enable inlay hints for VSCode

kevinbarabash:
## Summary:
Inlay hints can be useful at see the types for things without having to hover over everything.  They can also be useful in understand what args are being passed as what params especially when the args are literals.

This PR enables all of the inlay hints in our .vscode/settings.json.  In order to see the inlay hints developers must press and hold Ctrl-Opt at the same time otherwise they're not visible.

If developers want to customize this behaviour they can do so in their User settings.json for VSCode.

Issue: None

## Test plan:
- open a .ts file in VSCode
- press Ctrl-Opt
- see all of the inlay hints

Screen Shot 2024-03-12 at 10 23 07 AM


benchristel (approved):
Thanks!

PR #1071, opened 2024-03-13
Add "Check answer" buttons to the interactive graph flipbook

benchristel:
Issue: LEMS-1814

## Test plan:

```
yarn dev
open http://localhost:5173/#flipbook
```

Follow the onscreen instructions to get graph data into the flipbook.

Clicking the "Check answer" button should pop up an alert telling you
if you've answered the question correctly.

### DIFF
--- a/dev/flipbook.tsx
+++ b/dev/flipbook.tsx
@@ -123,7 +123,7 @@ function GradableRenderer(props: QuestionRendererProps) {
             case "invalid":
                 return "You didn't answer the question.";
             case "points":
-                return score.earned > 0 ? "Correct!" : "Incorrect.";
+                return score.earned >= score.total ? "Correct!" : "Incorrect.";
         }
     }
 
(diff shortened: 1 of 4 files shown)


dev/flipbook.tsx
```diff
@@ -83,30 +89,63 @@ type QuestionRendererProps = {
...
+function GradableRenderer(props: QuestionRendererProps) {
+    const {question, apiOptions} = props;
+    const rendererRef = useRef<Renderer>(null);
+
+    function describeScore(score: PerseusScore): string {
+        switch (score.type) {
+            case "invalid":
+                return "You didn't answer the question.";
+            case "points":
+                return score.earned > 0 ? "Correct!" : "Incorrect.";
```
jeremywiebe, line 126: This is not quite right. For a multi-widget question, the `total` field on the score may be more than 1 and so to check for correctness, we should be using `score.earned >= score.total` (I'm not sure how it could be MORE, but that's how https://github.com/Khan/perseus/blob/881da724d36561b2a200af0258ab81523d4d236a/packages/perseus/src/util.ts#L273 does correctness checking).

jeremywiebe (approved):
Nice!

PR #1073, opened 2024-03-13
Make Mafs line segment color offBlack64

benchristel:
## Summary:
We changed it (accidentally) to offBlack16 in
https://github.com/Khan/perseus/pull/1067. This commit fixes it.

Issue: none

Test plan:

View the Mafs segment graph in Storybook (http://localhost:6006/?path=/story/perseus-widgets-interactive-graph--segment-with-mafs-and-locked-points).
Verify that the segment is a darker gray than the gridlines.
Screen Shot 2024-03-13 at 6 11 27 PM

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/mafs-styles.css
+++ b/packages/perseus/src/widgets/interactive-graphs/mafs-styles.css
@@ -23,7 +23,7 @@
 }
 
 .MafsView .movable-segment {
-    --mafs-segment-stroke-color: var(--mafs-line-color);
+    --mafs-segment-stroke-color: rgba(33, 36, 44, 0.64); /* offBlack64 */;
     --mafs-segment-stroke-weight: 2px;
 }
 
--- a/packages/perseus/src/widgets/interactive-graphs/mafs-styles.css
+++ b/packages/perseus/src/widgets/interactive-graphs/mafs-styles.css
@@ -6,6 +6,8 @@
     --mafs-bg: transparent;
     --mafs-fg: rgb(33, 36, 44); /* WB color.offBlack */
 
+    --mafs-line-color: rgba(33, 36, 44, 0.16); /* WB color.offBlack16*/
+
     --mafs-blue: #1865f2; /* WB color.blue */
     --mafs-red: #d92916; /* WB color.red */
     --mafs-green: #00a60e; /* WB color.green */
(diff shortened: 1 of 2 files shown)


benchristel:
I wonder if we should have visual snapshot tests to catch regressions like this 🤔

jeremywiebe (approved):

nedredmond:
The screenshot is a Graphie background, not a Mafs grid

nedredmond:
@benchristel there's some confusion here-- the line segment color should use "fg", or foreground, not "lime-color", which is for gridlines

packages/perseus/src/widgets/interactive-graphs/mafs-styles.css
```diff
@@ -6,8 +6,6 @@
     --mafs-bg: transparent;
     --mafs-fg: rgb(33, 36, 44); /* WB color.offBlack */
 
-    --mafs-line-color: rgba(33, 36, 44, 0.16); /* WB color.offBlack16*/
```
nedredmond, line 9: @nishasy and I looked at this-- this variable is used for gridlines and needs to be offBlack16 or offBlack24

packages/perseus/src/widgets/interactive-graphs/mafs-styles.css
```diff
@@ -23,7 +21,7 @@
 }
 
 .MafsView .movable-segment {
-    --mafs-segment-stroke-color: var(--mafs-line-color);
```
nedredmond, line 26: This was the wrong variable for segment stroke, should have been fg (or you can make it offBlack64 like you did)

nedredmond (commented):
Please verify that you are not causing a regression in the Mafs grid rendering

nedredmond:
I am getting the impression you mistakenly thought it was a Perseus-side custom variable. Unfortunately, Mafs doesn't have great documentation on the variables available. See its use here:

https://github.com/stevenpetryk/mafs/blob/main/src/display/Coordinates/Cartesian.tsx

Please restore it. Again, it's used to color gridlines.

benchristel:
@nedredmond Thanks, and good catch. I had a brain fart and thought `--mafs-line-color` was unused because searching for it showed no usages.

It's fixed now.

packages/perseus/src/widgets/interactive-graphs/mafs-styles.css
nishasy: @nedredmond I ended up changing it back to offBlack16, because I looked and offBlack24 is not there in WB. 24% was only added for certain colors, not black. My mistake!

nedredmond:
> @nedredmond Thanks, and good catch. I had a brain fart and thought `--mafs-line-color` was unused because searching for it showed no usages.
> 
> It's fixed now.

Great! Sorry for repeating myself a bunch i reviewed on the GitHub app and couldn't see my comments

PR #1074, opened 2024-03-14
Create UI for adding a locked point to interactive graph editor (mafs only)

nishasy:
## Summary:
UI for adding a locked point to interactive graph editor! This will only be for mafs graphs as mafs graphs have a slightly different look and API (and also it was just not working very well with graphie anyway).

This UI includes:
- A dropdown that allows adding elements (just "Point" for now) added to the interactive graph editor.
- Once added, settings for points
  - The ability to delete the setting in question
  - The x and y coordinates of the point
- The removal of the locked point investigation from graph.tsx and related files.

[The ability to change the color of a point](https://khanacademy.atlassian.net/browse/LEMS-1751) is coming in a future PR.

Issue: LEMS-1813

## Test plan:
`yarn jest packages/perseus-editor/src/components/__tests__/util.test.ts`

Storybook
- Go to http://localhost:6006/?path=/story/perseus-editor-widgets-interactive-graph-editor--with-locked-points
- Scroll down to the "Add an element" dropdown
- Click "Add an element"
- Click "Point"
- Confirm that a black point appears at (0, 0) on the graph
- Change the coordinates and confirm that the point moves
- Add another point
- Use the trash button to delete a point
- Confirm that the correct point is deleted

## Demo:

https://github.com/Khan/perseus/assets/13231763/9eca725b-651a-4312-b639-bda26163ab8e


packages/perseus-editor/src/widgets/__tests__/interactive-graph-editor.test.tsx
```diff
@@ -43,6 +61,9 @@ describe("InteractiveGraphEditor", () => {
 
         render(
             <InteractiveGraphEditor {...baseProps} onChange={onChangeMock} />,
+            {
+                wrapper: RenderStateRoot,
+            },
```
nishasy, line 66: Needed to add `RenderStateRoot` because of the unique ID provider.

packages/perseus-editor/src/components/labeled-row.tsx
nishasy, line 1: The click target was taking up the whole row when it wasn't supposed to. Figured I'd fix it really quickly while I was here.

packages/perseus-editor/src/components/__tests__/util.test.ts
```diff
@@ -0,0 +1,42 @@
+import {getValidNumberFromString, getDefaultFigureForFigureType} from "../util";
+
+describe("getValidNumberFromString", () => {
```
jeremywiebe, line 3: Thanks for adding tests for this!

packages/perseus-editor/src/components/locked-figure-select.tsx
```diff
@@ -0,0 +1,49 @@
+/**
+ * Dropdown for selecting a locked figure to add to an interactive graph.
+ * Locked figures are elements (points, segmeents, etc.) that are not
```
jeremywiebe, line 3:
```suggestion
 * Locked figures are elements (points, segments, etc.) that are not
```

packages/perseus-editor/src/components/locked-figure-select.tsx
```diff
@@ -0,0 +1,42 @@
...
+    onChange: (value: string) => void;
+};
+
+const LockedFigureSelect = (props: Props) => {
+    const {id, onChange} = props;
+
+    return (
+        <View style={styles.container}>
+            <ActionMenu menuText="Add element" style={styles.addElementSelect}>
+                {[
```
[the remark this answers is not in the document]
jeremywiebe: :til: I would have used a fragment here.

packages/perseus-editor/src/components/locked-point-settings.tsx
```diff
@@ -0,0 +1,131 @@
...
+                <IconButton
+                    icon={trashIcon}
+                    aria-label={`Delete locked point at ${coordState[0]}, ${coordState[1]}`}
+                    onClick={onRemove}
+                />
+            </View>
+
+            {/* Coordinates */}
+            <View>
+                <View style={styles.row}>
```
jeremywiebe, line 73: **thought**: From a design perspective, I wonder if we should put the x and y coord inputs on the same row.  Vertical space is so precious in our editors and so I think we should try to avoid scrolling if at all possible.  

What do you think about something like this: 

```
*Point* 
Coordinate [x ] [y ]
```

Where the inputs would have a watermark/placeholder for which axis they're for?  I think authors using this editor would understand "x" always comes first so even with data in the fields they'd likely understand it's an x, y coordinate pair.

packages/perseus-editor/src/components/locked-figures-section.tsx
```diff
@@ -0,0 +1,87 @@
...
+        onChange(newProps);
+    }
+
+    return (
+        <View style={styles.container}>
+            {figures?.map((figure, index) => (
+                <LockedFigureSettings
+                    key={`${uniqueId}-locked-${figure}-${index}`}
+                    {...figure}
+                    onRemove={() => removeLockedFigure(index)}
```
jeremywiebe, line 69: I have a possibly-unfounded concern that using indexes for these callbacks could be problematic.  I know for react, you aren't supposed to use array indexes for keys.  Does that apply here?

packages/perseus-editor/src/widgets/__tests__/interactive-graph-editor.test.tsx
jeremywiebe: :til: about `render()`'s `wrapper`!  Is it worth a `render` helper in this test file?

packages/perseus-editor/src/widgets/interactive-graph-editor.tsx
```diff
@@ -135,6 +137,12 @@ type Props = {
      * editor page.
      */
     correct: any; // TODO(jeremy)
+    /**
+     * The locked figures to display in the graph area.
+     * Locked figures are graph elements (points, lines, line segmeents,
+     * etc.) that are locked in place and not interactive.
+     */
+    lockedFigures?: Array<LockedFigure>;
```
jeremywiebe, line 145:
```suggestion
    lockedFigures?: ReadonlyArray<LockedFigure>;
```

jeremywiebe (commented):
Coooool.   I left some comments, nothing that _has_ to be done. :)

PR #1077, opened 2024-03-14
Implement adjusting colors for locked points

nishasy:
## Summary:
Now that the interactive graph editor can add locked
points, we also need to make it so that they can update
the color of those points.

In this PR, I added a dropdown that allows for changing
the color. I set the default color to blue. I also
made is so that the actual color value isn't saved
as part of the locked figure styles, just the name.
This then gets converted to the wonder blocks color
later in the mafs graph. I did it this way to avoid
having to convert back and forth between the color name
and value in the color UI.

Issue: https://khanacademy.atlassian.net/browse/LEMS-1751

## Test plan:

Storybook
- Go to http://localhost:6006/?path=/story/perseus-widgets-interactive-graph--segment-with-mafs-and-locked-points
- Confirm that the locked points are blue by default
- Go to http://localhost:6006/?path=/story/perseus-editor-widgets-interactive-graph-editor--with-locked-points
- There should be two locked points: one blue and one gold
- Confirm that using the color dropdown changes the color for the appropriate point
- Add a new point
- Confirm that it's blue by default
- Change its color and confirm all the colors work
- Change the coordinates of the points and make sure that nothing breaks
- Delete points and make sure nothing breaks
- Add points again and make sure nothing breaks 
- Click on the "Color" label and confirm that it opens the color dropdown.
- ^ Repeat for each dropdown and make sure it opens the correct one (IDs are unique and work as expected)

## Demo:

Screenshot 2024-03-14 at 3 19 08 PM

https://github.com/Khan/perseus/assets/13231763/6c8306b1-8364-4096-bfb7-4994db3a33e7


packages/perseus-editor/src/components/locked-point-settings.tsx
```diff
@@ -18,13 +19,24 @@ import {getValidNumberFromString} from "./util";
...
 
+const possibleColors = [
+    "purple",
+    "blue",
+    "teal",
+    "green",
+    "gold",
+    "red",
+    "pink",
+];
```
nishasy, line 30: These are not the exact colors from the [designs](https://www.figma.com/file/JroPokeCLBv2VRIw3GPvY5/LC---Interactive-graph-widget?type=design&node-id=739-6745&mode=design&t=9dixLbZO1PneX7r0-0), but I'm not sure where those are? I did my best to match them with Wonder Blocks colors.

nedredmond (approved):
I took a few looks over to see if I could find anything objectionable so you wouldn't call me out for green-stamping 😛, but this change legitimately LGTM!

packages/perseus-editor/src/components/locked-point-settings.tsx
jeremywiebe: I think there's a similar list defined in the `mathjax-renderer` repo.  You might want to work with Ben and/or Tamara to see if that list should be defined somewhere centrally?

packages/perseus/src/widgets/interactive-graphs/graph-locked-layer.tsx
```diff
@@ -1,13 +1,27 @@
...
+/**
+ * Takes a style object with color names and converts them to actual
+ * Wonder Blocks color values. This way, we can use colors and their
+ * names in the editor UI without having to convert them back and forth.
+ */
+function convertStyle(style?: StyleParams): StyleParams {
+    return {
+        ...style,
+        stroke: color[style?.stroke || "blue"],
+        fill: color[style?.fill || "blue"],
```
jeremywiebe, line 21: I wonder if we should have a `DefaultColor` variable somewhere?  🤔   (Or `DefaultPointColor` as we'll likely have other locked features that need a default color).

PR #1078, opened 2024-03-15
Fix MathJax dependency

Myranae:
## Summary:
The mathjax-renderer dependency is needed for Storybook test rendering. As such, it should be added to the root package.json instead of the Perseus package package.json. This commit makes that change.

Issue: LEMS-1820

## Test plan:
- Confirm all checks pass
- Confirm with Learning Equity that they are now able to upgrade


nedredmond (approved):
Thanks @Myranae!

PR #1080, opened 2024-03-15
Migrate Linear and Linear System to Mafs

nedredmond:
## Summary
Adds **Linear** and **Linear System** graph types behind a feature flag.

Also refactors parts of the Segment graph to be reusable, since they share logic. Refactored some types to reduce redundancy / duplication / maintenance burden.

https://github.com/Khan/perseus/assets/23404711/716d5938-35c6-4384-9b1c-8cd0fbe2d1e2

### Issues
LEMS-1774 and LEMS-1815

### Testing
Fire up `yarn dev` and activate the flags.


SonicScrewdriver (approved):
Paired with Ned to review this! It looks lovely and has a bunch of great improvements! :D

packages/perseus/src/widgets/interactive-graphs/types.ts
```diff
@@ -12,3 +16,21 @@ export type MafsGraphProps<T extends InteractiveGraphState> = {
     graphState: T;
     dispatch: (action: InteractiveGraphAction) => unknown;
 };
+
+export type InteractiveGraphState = SegmentGraphState | LinearGraphState;
+
+export interface InteractiveGraphStateCommon {
+    hasBeenInteractedWith: boolean;
+    // range = [[xMin, xMax], [yMin, yMax]] in Cartesian units
+    range: [Interval, Interval];
```
jeremywiebe, line 25: Do you think Typescript's ability to label tuple members would help here a bit? I know you can't use these labels to reference the values, but I do find it helps a bit as it removes some of the need for a comment like we have above. 

```suggestion
    range: [x: Interval, y: Interval];
```

packages/perseus/src/widgets/interactive-graphs/mafs-styles.css
```diff
@@ -1,11 +1,12 @@
 /**
- * Overrides of mafs styles-- we will want to move this into JS land
- * to take advantage of WB tokens, but I cannot get that to work at the moment
+ * Overrides of mafs theme-- we will want to move this into JS land
```
jeremywiebe, line 2: Would a good "half step" improvement be to make this a `.less` file and then at least be able to have Less variables? 

I think we already have some of our WB colours defined in [`variables.shared.less`](https://github.com/Khan/perseus/blob/c431c4b8147ae0630df2d2b19b0f5a5b5f04d4bf/packages/perseus/src/styles/variables.shared.less#L79)

packages/perseus/src/widgets/interactive-graphs/interactive-graph-state.ts
```diff
@@ -48,7 +49,7 @@ const getDefaultSegments = (props: {
     graph: PerseusGraphTypeSegment;
     step: vec.Vector2;
     range: [Interval, Interval];
-}): Segment[] => {
+}): CollinearTuple[] => {
```
jeremywiebe, line 52: :til: the term "collinear" :)  "lying in the same straight line"

packages/perseus/src/widgets/interactive-graphs/interactive-graph-reducer.ts
```diff
@@ -117,24 +127,27 @@ function clamp(value: number, min: number, max: number) {
...
-    newValue: A[number],
-): A {
-    const copy = [...array] as any;
+function setAtIndex<T, A extends readonly T[]>(args: {
+    array?: A;
+    index: number;
+    newValue: A[number];
+}): A {
+    const {array, index, newValue} = args;
+    const copy: T[] = [...(array || [])];
```
jeremywiebe, line 155: **suggestion**: Instead of dealign with the default array value on line 150, would this be a bit easier to see?

```suggestion
function setAtIndex<T>({
    array = [], 
    index, 
    newValue
}: {
    array?: ReadonlyArray<T>;
    index: number;
    newValue: T;
}): ReadonlyArray<T> {
    const copy: T[] = [...array];
    copy[index] = newValue;
    return copy;
}
```

packages/perseus/src/widgets/interactive-graphs/graphs/types.ts
```diff
@@ -0,0 +1,10 @@
+import type {CollinearTuple} from "../../../perseus-types";
+import type {vec} from "mafs";
+
+export interface InteractiveLineProps {
+    collinearPair: Readonly<CollinearTuple>;
```
jeremywiebe, line 5: Can this be a tuple like other places you updated things?

```suggestion
    collinearPair: [CollinearTuple, CollinearTuple];
```

packages/perseus/src/widgets/interactive-graphs/graphs/linear.tsx
```diff
@@ -0,0 +1,85 @@
...
+    const {dispatch} = props;
+    const {coords: lines, snapStep, range, type} = props.graphState;
+
+    const colors = ["var(--movable-line-stroke-color)", "var(--mafs-violet)"];
+
+    return (
+        <>
+            {lines?.map((line, i) => (
+                <LineView
+                    key={i}
```
jeremywiebe, line 24: **question**: 🤔  I wonder if there's any "stronger" line identifier than its index?  or if we should give each line some type of stable ID?

jeremywiebe (commented):
Looks like I was too slow to submit. :) If there's anything here you like, we can do it in a future PR.

packages/perseus/src/perseus-types.ts
```diff
@@ -2,9 +2,11 @@
 // TODO(FEI-4011): Use types generated by https://github.com/jaredly/generate-perseus-flowtypes
 
 import type {Coord} from "./interactive2/types";
+import type {Interval, vec} from "mafs";
```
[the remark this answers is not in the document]
nedredmond: Without it, we'd be duplicating the types we use elsewhere. Tough call.

packages/perseus/src/widgets/interactive-graphs/types.ts
nedredmond: TIL!

packages/perseus/src/widgets/interactive-graphs/interactive-graph-reducer.ts
nedredmond: @jeremywiebe I like adding the default array, that would be cleaner!

packages/perseus/src/widgets/interactive-graphs/graphs/types.ts
nedredmond: Yes, it should be! This was changed very early in the process. Good catch!

packages/perseus/src/widgets/interactive-graphs/graphs/linear.tsx
nedredmond: I don't think there is at the moment. However, the index for these lines will not change for the user. Definitely something we will need to consider for more dynamic graphs like the Unlimited Points.

packages/perseus/src/widgets/interactive-graphs/graphs/types.ts
nedredmond: Oh-- this isn't a ReadonlyArray, it's just Readonly!

PR #1081, opened 2024-03-15
Handle the case where matchMedia is not defined

benchristel:
## Summary:
This is relevant for e.g. SSR in webapp. Previously, we were seeing an
error during SSR when there was an Explanation widget on the page.

Issue: none

Test plan:

- Install a dev build of Perseus in local webapp
- `make serve-ssr`
- As a pre-phantom user (no cookies) with JS disabled in your browser,
  visit
  `/internal-courses/test-everything-qe-sandbox/xa527aaa0ed54bca0:test-everything-1/xa527aaa0ed54bca0:te-explanation/a/explanation-article`
- You should not see an error. The article should render.

### DIFF
--- a/.changeset/tricky-dolphins-double.md
+++ b/.changeset/tricky-dolphins-double.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Fix a bug in the Explanation widget that was causing SSR to error out.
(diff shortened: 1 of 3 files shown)


mark-fitzgerald (approved):

jeremywiebe (approved):
This makes sense. Thanks for fixing.  I do wonder if we'll end up re-breaking SSR in the future with some other API usage... and how we might catch it earlier. 🤔

PR #1084, opened 2024-03-15
Migrate `Ray` graph type to Mafs

nedredmond:
## Summary
Adds `Ray` graph type behind a feature flag.

Also refactors tests to iterate over questions and reduce repetition.

https://github.com/Khan/perseus/assets/23404711/52c0a1d8-05f1-4fcc-a081-9eb917cf0796

### Issues
LEMS-1823

### Testing
Fire up `yarn dev` and activate the flag.

### DIFF
--- a/packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
+++ b/packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
@@ -19,7 +19,7 @@ import {
 import {renderQuestion} from "./renderQuestion";
 
 import type {Coord} from "../../interactive2/types";
-import type {PerseusGraphType, PerseusRenderer} from "../../perseus-types";
+import type {PerseusRenderer} from "../../perseus-types";
 import type Renderer from "../../renderer";
 import type {APIOptions} from "../../types";
 import type {UserEvent} from "@testing-library/user-event";
--- a/packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
+++ b/packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
@@ -9,13 +9,11 @@ import * as Dependencies from "../../dependencies";
 import {ApiOptions} from "../../perseus-api";
 import {
     questionsAndAnswers,
-    segmentQuestion,
     segmentWithLockedPointsQuestion,
     segmentQuestionDefaultCorrect,
-    linearQuestion,
     linearQuestionWithDefaultCorrect,
     linearSystemQuestionWithDefaultCorrect,
-    linearSystemQuestion,
+    rayQuestionWithDefaultCorrect,
 } from "../__testdata__/interactive-graph.testdata";
 
 import {renderQuestion} from "./renderQuestion";
@@ -131,6 +129,7 @@ describe("mafs graphs", () => {
         "segment",
         "linear",
         "linear-system",
+        "ray",
     ];
 
     const graphTypeFlags = graphsTypesToEnable.reduce((acc, type) => {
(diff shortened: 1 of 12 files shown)


packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
```diff
@@ -242,125 +198,42 @@ describe("mafs graphs", () => {
         });
     });
 
-    describe("linear", () => {
```
nedredmond, line 245: I moved this below the graph-specific tests

packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
```diff
@@ -242,125 +198,42 @@ describe("mafs graphs", () => {
...
-
-            // Act
-            // no action
-
-            // Assert
-            expect(renderer).toHaveInvalidInput();
-        });
-
-        it("rejects incorrect answer", async () => {
+    describe("locked layer", () => {
```
nedredmond, line 201: I just swapped the order of the tests

packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
```diff
@@ -142,56 +141,19 @@ describe("mafs graphs", () => {
...
-        });
-    });
-
-    describe("segment", () => {
+    describe.each([
+        segmentQuestionDefaultCorrect,
+        linearQuestionWithDefaultCorrect,
+        linearSystemQuestionWithDefaultCorrect,
+        rayQuestionWithDefaultCorrect,
+    ])("segment", (question) => {
```
nedredmond, line 149: Since all of the sets of tests are basically identical, we can just loop through them this way

packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
```diff
@@ -142,56 +141,19 @@ describe("mafs graphs", () => {
...
-            );
-
-            // Assert
-            expect(points[0]).toHaveStyle({fill: color.red, stroke: color.red});
-            expect(points[1]).toHaveStyle({fill: color.red, stroke: color.red});
-        });
-    });
-
-    describe("segment", () => {
+    describe.each([
```
jeremywiebe, line 144: +++

packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
```diff
@@ -142,56 +141,19 @@ describe("mafs graphs", () => {
...
+        segmentQuestionDefaultCorrect,
+        linearQuestionWithDefaultCorrect,
+        linearSystemQuestionWithDefaultCorrect,
+        rayQuestionWithDefaultCorrect,
+    ])("graph type", (question) => {
```
jeremywiebe, line 149: **idea**: this isn't a massive issue, but if any of these fail, the test name won't reveal which graph type failed.  

What do you think about this?
```suggestion
        ["segment", segmentQuestionDefaultCorrect],
        ["linear", linearQuestionWithDefaultCorrect],
        ["linear-system", linearSystemQuestionWithDefaultCorrect],
        ["ray", rayQuestionWithDefaultCorrect],
    ])("graph type: %s", (type, question) => {
```

This adds a bit of maintenance overhead but then we know which graph type failed.

jeremywiebe (approved):

packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
nedredmond: I was thinking about doing this as well, just stopped because I thought I was being a bit extra. I'll get back to it Monday

PR #1086, opened 2024-03-18
Create helpers to replace snowman regex uses on the front-end

Myranae:
## Summary:
There are several locations in webapp where regexes are used to extract widget placeholder helpers or related information from Perseus JSON. These helpers should allow us to remove those regexes, providing us with more flexible when changing structures in Perseus. It also allows for easier testing when related changes are made.

Issue: LEMS-1800

## Test plan:
- Confirm all checks pass
- Use npm snapshot to confirm these helpers can correctly replace regex uses in webapp


packages/perseus/src/util/snowman-utils.ts
```diff
@@ -0,0 +1,149 @@
+export enum WidgetType {
```
SonicScrewdriver, line 1: Interesting! I think [Matthew is right](https://github.com/Khan/perseus/pull/1058), we may want to discuss how the heck to manage all these widget lists as they are starting to get [voluminous](https://github.com/Khan/perseus/pull/1058/files). 

They all seem unique, but I wonder if there's something we can do to simplify the amount of work that we would need to do if we add a new widget, or rename an old one. Sync tags could work, but I'm not super stoked on the idea of them as I always found them a little frustrating to deal with.

packages/perseus/src/util/snowman-utils.ts
```diff
@@ -0,0 +1,149 @@
...
+    PythonProgram = "python-program",
+    Radio = "radio",
+    Sorter = "sorter",
+    Table = "table",
+    Unit = "unit",
+    Video = "video",
+    VideoTranscriptLink = "video-transcript-link",
+}
+
+const QUESTION_WIDGETS = [
```
SonicScrewdriver, line 41: Is it possible to add a comment to explain what QUESTION WIDGETs are? I think it could help our future selves understand what these are. :) 

Iirc I THINK we determined that these are all widgets that actually require user input -- but not all of them.

packages/perseus/src/util/snowman-utils.ts
```diff
@@ -0,0 +1,149 @@
...
+    "interactive-number-line",
+    "categorizer",
+    "plotter",
+    "orderer",
+    "protractor",
+    "matcher",
+    "sorter",
+];
+
+// Regex for widget placeholders in a string.
```
SonicScrewdriver, line 55: Thank you for these comments! They really help explain what we're trying to grep.

packages/perseus/src/util/snowman-utils.ts
```diff
@@ -0,0 +1,149 @@
...
+    const widgets: Array<string> = [];
+
+    let match = widgetRegex.exec(content);
+
+    while (match !== null) {
+        widgets.push(match[1]);
+        match = widgetRegex.exec(content);
+    }
+
+    if (widgets.length === 0) {
```
SonicScrewdriver, line 95: Interesting! I think this makes sense, but it made me wonder if we have any really strange exercises somewhere _without_ widgets. Hopefully not!

packages/perseus/src/util/snowman-utils.ts
```diff
@@ -0,0 +1,149 @@
...
+        if (!widgetTypes.includes(match[1])) {
+            widgetTypes.push(match[1]);
+        }
+        match = widgetTypeRegex.exec(content);
+    }
+    return widgetTypes;
+}
+
+/**
+ * Check if a specific widget is a question widget type using its ID.
```
SonicScrewdriver, line 126: A brief explanation of what the question widget type means could also be here! Not completely necessary however, so long as one of the two places explains what a "question widget type" is, or why we're checking for it.

packages/perseus/src/util/snowman-utils.ts
Myranae: I'll add the comment "// These widgets all require user-input, though this is not an exhaustive list" or something like that. How does that sound?

packages/perseus/src/util/snowman-utils.test.ts
```diff
@@ -0,0 +1,93 @@
...
+        const content =
+            "> [[☃ passage 3]]\n\n=====\n\n###Questions for Passage 1\n\n[[☃ group 1]]  \n[[☃ group 12]]  \n\n$$\n";
+        const actual = getAllWidgetIds(content);
+        const expected = ["passage 3", "group 1", "group 12"];
+        expect(actual).toEqual(expected);
+    });
+});
+
+describe("getAllWidgetIds", () => {
+    it("returns all widget names present in a string", () => {
```
SonicScrewdriver, line 30: nit: should this be widget ids for consistency with the test names? (non blocking)

packages/perseus/src/util/snowman-utils.ts
```diff
@@ -0,0 +1,149 @@
...
+    let match = widgetRegex.exec(content);
+
+    while (match !== null) {
+        widgets.push(match[1]);
+        match = widgetRegex.exec(content);
+    }
+
+    if (widgets.length === 0) {
+        throw new Error("Exercises should contain widgets");
+    }
```
handeyeco, line 97: This might be fine, but I think it might be equally fine to return an empty array or null.

SonicScrewdriver (approved):
Looks wonderful, thank you! I left a couple small comments about adding comments, but nothing blocking! :)

packages/perseus/src/util/snowman-utils.ts
Myranae: Yea, I didn't want to keep this as I'm not sure it is necessary to throw an error, but the error was something the repatriating file had. I originally thought there was a type error if I took this out, but it doesn't seem to happen anymore. Do you think I should check to see if the code in the repatriating file can be changed to not need this error?

packages/perseus/src/util/snowman-utils.ts
SonicScrewdriver: This was to help out with some of the cheat utils, correct? I wonder if something like: "These are all the currently widgets that require user-input and are supported for automatic scoring." would work? 

Hmm I wonder if we should document somewhere all these places where we're only supporting SOME widgets.🤔 It might be a good callout for potential places of improvement in the future.

packages/perseus/src/util/snowman-utils.ts
```diff
@@ -0,0 +1,149 @@
...
+
+/**
+ * Extract all widget types from a Perseus JSON content string.
+ * This does not include the instance number and prevents duplicates.
+ * example output: ['radio', 'categorizer']
+ *
+ * @param {string} content
+ * @returns {Array<string>}
+ */
+export function getAllWidgetTypes(content: string): Array<string> {
```
handeyeco, line 110: I _think_ if we change WidgetType from an enum to a const, we might be able to do `Array<WidgetType>` which would be a little more specific.

packages/perseus/src/util/snowman-utils.ts
SonicScrewdriver: Naw, I think this is fine as is. :) It just piqued my interest as I was reading through. 

The amount of errors are doubtful to be high, and it's likely useful to know if we start seeing a lot of errors for exercises with no widgets at all.

packages/perseus/src/util/snowman-utils.ts
```diff
@@ -0,0 +1,149 @@
...
+    let match = widgetTypeRegex.exec(content);
+
+    while (match !== null) {
+        // Might need to take this check out and just list them all
+        if (!widgetTypes.includes(match[1])) {
+            widgetTypes.push(match[1]);
+        }
+        match = widgetTypeRegex.exec(content);
+    }
+    return widgetTypes;
```
handeyeco, line 122: In this helper we return an empty array instead of throwing an error. I think we need to be consistent, my gut is either an empty array or null.

packages/perseus/src/util/snowman-utils.ts
```diff
@@ -0,0 +1,149 @@
...
+}
+
+/**
+ * Check if a specific widget is a question widget type using its ID.
+ * The widget ID includes the widget type and the instance number.
+ *
+ * @param {string} widgetId
+ * @returns {boolean}
+ */
+const isQuestionWidgetType = (widgetId: string): boolean => {
```
handeyeco, line 132: nit: you're using both `const funName = () => {}` and `function funName() {}`. It's not an issue to use one or the other, but it's nice to be generally consistent. My personal style is to use `function` for named functions and arrow functions for anonymous functions.

packages/perseus/src/util/snowman-utils.ts
```diff
@@ -0,0 +1,149 @@
...
+    return QUESTION_WIDGETS.includes(widgetId.split(" ")[0]);
```
handeyeco, line 133: Ultra tiny thought, but I wonder if it's worth making this more universal. For instance:

``` JS
function isQuestionWidgetType(widget: string): boolean {
  const split = widget.split(" ")
  return QUESTION_WIDGETS.includes(widget) || QUESTION_WIDGETS.includes(split[0])
}

console.log(isQuestionWidgetType("radio")) // true
console.log(isQuestionWidgetType("radio 1")) // true
```

No action needed, just a thought.

packages/perseus/src/util/snowman-utils.ts
Myranae: Could add a TODO above the places that require it and leave a comment about how only some widgets are listed and it will need to be updated in the future? 🤔

packages/perseus/src/util/snowman-utils.test.ts
Myranae: Ooo, thank you so much! :D

packages/perseus/src/util/snowman-utils.test.ts
```diff
@@ -0,0 +1,93 @@
...
+describe("getQuestionWidgetIds", () => {
+    it("returns all question widgets in a string", () => {
+        // Arrange
+        const questionContent = `${addWidget(
+            WidgetType.NumericInput,
+            1,
+        )} foo ${addWidget(WidgetType.Radio, 1)}`;
+
+        // Act
+        const questionWidget = getQuestionWidgetIds(questionContent);
```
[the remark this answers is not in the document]
Myranae: The question one looks through a specific list of widgets. We are assuming that the list is widgets that require user-input for automatic sorting, from what Third and I saw when we were looking into the code.

packages/perseus/src/util/snowman-utils.ts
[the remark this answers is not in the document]
Myranae: 100%

packages/perseus/src/util/snowman-utils.ts
```diff
@@ -0,0 +1,149 @@
...
+    "sorter",
+];
+
+// Regex for widget placeholders in a string.
+// Widget IDs are identified using capture groups. Ex. 'radio 1'
+const widgetRegex = /\[\[☃ ([^\]]+)\]\]/g;
+
+// Regex for widget placeholders in a string. Ex. 'radio'
+// Widget types are identified using capture groups.
+const widgetTypeRegex = /\[\[☃ ([a-z]+) +\d+\]\]/g;
```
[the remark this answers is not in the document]
Myranae: You were right! Added the tests and fixed the regex

packages/perseus/src/util/snowman-utils.ts
Myranae: Was able to do the specific type with the enum :)

packages/perseus/src/util/snowman-utils.ts
Myranae: Will return empty array instead

packages/perseus/src/util/snowman-utils.ts
Myranae: Fixed!

packages/perseus/src/util/snowman-utils.ts
Myranae: Love it. Added it.

packages/perseus/src/util/snowman-utils.ts
```diff
@@ -0,0 +1,157 @@
...
+ * @param {WidgetId} widgetId
+ * @returns {boolean}
+ */
+function isQuestionWidgetType(widgetId: WidgetId): boolean {
+    const widgetIdString = widgetId as string;
+    return (
+        QUESTION_WIDGETS.includes(widgetIdString.split(" ")[0] as WidgetType) ||
+        QUESTION_WIDGETS.includes(widgetId as WidgetType)
+    );
+}
```
handeyeco, line 142: Take a look at this code as a way to maybe prevent the need for all these `as` casting things. It's still not very correct though...I'm sure there's a better way.

``` TS
const RADIO = "radio"
const PASSAGE = "passage"

const WIDGET_TYPES = ["radio", "passage"] as const;
type WidgetType = (typeof WIDGET_TYPES)[number];

const QUESTION_WIDGETS = ["radio"] as const;
type QuestionWidgetType = (typeof QUESTION_WIDGETS)[number];

type WidgetId = `${WidgetType} ${number}`;

const radioId: WidgetId = "radio 4";
const passageId: WidgetId = "passage 2";
// @ts-expect-error
const wrongId: WidgetId = "radio"; // catches when we're missing a number
// @ts-expect-error
const wrongId2: WidgetId = "rabio 4"; // catches when we misspell radio

const radioType: WidgetType = "radio";
const passageType: WidgetType = "passage";
// @ts-expect-error
const wrongType: WidgetType = "radio 4" // catches when we add a number
// @ts-expect-error
const wrongType2: WidgetId = "rabio"; // catches when we misspell radio

const radioQuestion: QuestionWidgetType = "radio";
// @ts-expect-error
const passageQuestion: QuestionWidgetType = "passage"; // catches when we think passage is a question type

function isQuestionWidgetType(widgetId: WidgetType | WidgetId): boolean {
    return (
        // Thanks to typing the params right, we can feel safe that this will work
        // either it's already a widget type ("radio".split(" ") is ["radio"])
        // or it's a valid id for a widget ("radio 1". split(" ") is ["radio", "1"])
        // so this plus the fact that it's incredibly short-lived makes it a pretty safe "any"
        QUESTION_WIDGETS.includes((widgetId.split(" ")[0] as any))
    );
}

console.log(isQuestionWidgetType(radioId));
console.log(isQuestionWidgetType(passageId));
console.log(isQuestionWidgetType(radioType));
console.log(isQuestionWidgetType(passageType));

// @ts-expect-error
console.log(isQuestionWidgetType("hello world")); // catches arbitrary strings
```

[Playground](https://www.typescriptlang.org/play?#code/MYewdgzgLgBASgQQCIEkDyMC8MBEAnAQwBMBLEHAKFElgAUEBlBhAcQFEtcAHAiCAgOYBTSlXDQYAdRRJ2AFQD6cgJq02DTgG18xMjgA03Xv2E4AujF4xq0ANwUoATy5CpJIsKhznr7AAonFxAAMykZeSVVdQBKTTAAVwBbACMhPDN7MRoYAEUAVXU5dAA5BWlZNjkNbG1CUnILKxsoe0DXHPihaDIwSXdPbxdOAJ8Q3IKGIrRS8vkGWISUtIyKBx83DyEoFCJOAAMAEgBvPs2vHwBfGGPF1LwLvczmmDqyHYAuDc8dzh16mAALDh7M8eHxBEIPl8tj9sDgwSZXAAmYEUAD0aJgAAEoBAALRCAAeLmAUAJeDwIDwWQkAHdKWABFDTt9dnDXuRbDAMdYCFBgAALLowWlCsAioQAcjwrkSJD4JEZlhgtzS6MxOPxRJJZLSlOpz3p4CZRCRnxZMLZuEIyTIgOB3MxwD5guFoqE4tpsvlEBcABs-S9dCBVs8OYMhOb+lsI78OajQcYIRGo2dY3CERDUTzNQTiUJSeT9TTYEbGSnoechuzg-bHbz+UKICKxRLLERdgQVUk7ursbi8zqi1SSyKGQII2bK7DrQRbZz687G27W16YHK+P7AxzQ+JYByOl0oD1PofuuALVXfLP6tmNQPtQXdRSR4nwcIz8fwKfOufetGr1+TNTC5Hkl1dZt3U9VwoAFRUAGsYGA1x5WVABHX8v3FNpVmCeIwFJHoYHlT8ekvCM-FpADmQA2MAB9pyIaJPmSEAQD9IQCHFI4KBgPiXi2eI8HFPxeP48SeTkAUuPg5soBAGBAkVARFKFJCCEIRJmzwEgBAFKBDDXZ1xWCIQhEDfhTNUvlVNQqiAxFKl4LE8S+J5IQSFgtJiKgSVmwIP0ZWIRxlSos5FPWPw-j0AA6X0-U8qLcGiYjm1qYNzGiFzXJ5KkfL85UADcAvcYjdmCPKuzCzwYCijkYAARhwGKYHixKcGS1KYHS29DBwJqzCy1z+J5CAFNg1CuD9eI5LU4ICFJazYE8grFWAGVSGSP0QogAUqTJBLCqEXZEgIeDhU85UuBlKAnFaggrJwLjHEoYa+PyQoSjKcJKgYGK1umoguj8SjqKIOKpva5LNAABkafywEcaIhv46J7AuXdIHYoQYr9EABD8EjMLI2ifD8DkdmRkFxGx3H8cJiBSIvUmXD8ZDKbRmlabxgmiaPEm0zJ8MfCprmOLp3nGeJ5nBdZ5CI1Fvtc0fQs9VfGnxZ5hmmf-WWhCioUAwU2kqT9IgcCpxcXSbSw8FtKBCDwHaHeUiAgA)

packages/perseus/src/util/snowman-utils.ts
jeremywiebe: I keep thinking that there must be a way in TypeScript to use a utility type to extract the widget types from the `all-widgets.ts` export (https://github.com/Khan/perseus/blob/9917dceaf750b18122880d9096bddc3b859b9dfa/packages/perseus/src/all-widgets.ts#L5-L6) 

Maybe a question for #typescript ?

packages/perseus/src/util/snowman-utils.ts
```diff
@@ -0,0 +1,157 @@
...
+];
+
+/**
+ * Regex for widget placeholders in a string.
+ *
+ * First capture group is the widget ID (ex. 'radio 1')
+ * Second capture group is the widget type (ex. "radio)
+ * exec return will look like: ['[[☃ radio 1]]', 'radio 1', 'radio']
+ */
+export const widgetRegex = /\[\[☃ (([a-z-]+) \d+)\]\]/g;
```
handeyeco, line 65: I think this might be a bug. Since this is declared globally and adding the `g` makes it stateful, I think it's going to cause erratic behavior:

``` JS
const regexWithG = /\[(\w+) (\d+)\]/g
console.log(regexWithG.exec("[hello 42] [world 5]"))
// ["[hello 42]", "hello", "42"]

console.log(regexWithG.exec("[something 42] [else 5]"))
// ["[else 5]", "else", "5"]

console.log(regexWithG.exec("[new 42] [thing 5]"))
// null
```

handeyeco (changes requested):
My only blocking asks are:
- We make sure having a global, stateful regex isn't a bug
- We use the exported regex in the test, rather than redeclaring it

packages/perseus/src/util/snowman-utils.ts
Myranae: Thanks again for pointing this out, Third, and for chiming in, Jeremy! I created a ticket for us to come back to this and to consider extracting the types from `all-widgets.ts`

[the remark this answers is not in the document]
Myranae: Created a ticket for this: https://khanacademy.atlassian.net/browse/LEMS-1844

packages/perseus/src/util/snowman-utils.ts
```diff
@@ -0,0 +1,157 @@
...
+
+    while (match !== null) {
+        widgets.push(match[1] as WidgetId);
+        match = widgetRegex.exec(content);
+    }
+
+    return widgets;
+}
+
+function isWidgetType(value: string): value is WidgetType {
```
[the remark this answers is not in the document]
Myranae: Yea, I found this in one of the examples for checking if a string is in an enum!  Here's some relevant documentation for reference: https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

packages/perseus/src/util/snowman-utils.ts
```diff
@@ -0,0 +1,157 @@
...
+    return Object.values(WidgetType).some((widgetType) => value === widgetType);
```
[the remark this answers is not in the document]
Myranae: Includes in that example gives an error though. I would need to do `.includes(value as WidgetType)` for it to work. I was thinking avoiding casting would be better, but totally okay to change it to includes. I think reading it with includes makes more sense, but the functionality seems to work both ways. Here's the slack chat where I brought this up: https://khanacademy.slack.com/archives/C028T2HDX3P/p1710799481158499

packages/perseus/src/util/snowman-utils.test.ts
```diff
@@ -0,0 +1,120 @@
...
+    WidgetType,
+    getQuestionWidgetIds,
+    getAllWidgetTypes,
+} from "./snowman-utils";
+
+describe("widgetRegex", () => {
+    it("locates match and capture groups as expected for non-hyphenated widget types", () => {
+        // Arrange
+        const exampleContent = "Here is some content [[☃ group 1]]";
+        const widgetRegex = /\[\[☃ (([a-z-]+) \d+)\]\]/g;
```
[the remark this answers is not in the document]
Myranae: For sure. I initially did it this way because the regex is stateful, and that persists. I've written out two potential ways to fix this though. It looks like most instances in the code of using this regex use it in a while loop, which resets the lastIndex property of the regex to 0 once the search finishes. It can also be reset manually by setting .lastIndex to 0. I'll go over these during our group pair time today.

packages/perseus/src/util/snowman-utils.ts
Myranae: For sure. The webapp instances where I took the regex from all use a while loop to specifically loop through all matches of the regex, so they want it to be stateful. Not sure how to fix it so that the lastIndex property resets after each use other than to have each user make sure they reset it. Is there something else we could do here? We could make it not global and then refactor wherever it uses a for loop to somehow take that into account?

packages/perseus/src/util/snowman-utils.ts
Myranae: Thanks for writing out this example! Is this what it could partially look like if I used the values found in the enum utils file? Those aren't set as consts though. Would I need to create `const WIDGET_TYPES = ["radio", "passage"] as const;`with all the widgets listed and `const QUESTION_WIDGETS = ["radio"] as const;` with all the question type widgets listed? Also, if I am still declaring the type in this version, isn't it the same? Or is just having as few as possible still better?

packages/perseus/src/util/snowman-utils.ts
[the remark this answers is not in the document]
handeyeco: You might try `values` from `wonder-stuff-core`. I think the reason they exist is to help with typing.

packages/perseus/src/util/snowman-utils.ts
Myranae: Going to use the generator function for this! Thanks!

packages/perseus/src/util/snowman-utils.ts
Myranae: Will move forward with the current version and will try to remove the type declarations at a later time.

packages/perseus/src/util/snowman-utils.test.ts
```diff
@@ -0,0 +1,122 @@
...
+        const firstCaptureGroup = match ? match[1] : null;
+        const secondCaptureGroup = match ? match[2] : null;
+
+        //Assert
+        expect(widgetPlaceholder).toEqual("[[☃ group 1]]");
+        expect(firstCaptureGroup).toEqual("group 1");
+        expect(secondCaptureGroup).toEqual("group");
+    });
+
+    it("locates match and capture groups as expected for hyphenated widget types", () => {
```
handeyeco, line 28: I appreciate how focused these tests are

handeyeco (approved):
LGTM! Very clean PR, appreciate the work you put into this.

PR #1088, opened 2024-03-18
Avoid rendering legacy interactive graph when we really want mafs

benchristel:
Mark and I ran into an issue when testing these graphs and toggling between
different feature flag states.  When switching from the legacy interactive
graph to the Mafs version, the InteractiveGraph component wasn't remounted, so
the state of the legacy graph was preserved. This caused the line segment to
mysteriously disappear from a segment graph when the feature flag was switched
off!

To prevent similar gotchas in the future, this PR isolates the legacy graph code
in its own component, which is only rendered if the relevant Mafs feature flag
is turned off. This is a more accurate reflection of our intent.

Issue: none

## Test plan:

- View the interactive graph gallery (`yarn dev`)
- Toggle the "Segment" feature flag on.
- Move the line segment on the first graph.
- Toggle the "Segment" feature flag off.
- There should still be a line segment on the graph.

### DIFF
--- a/packages/perseus/src/widgets/interactive-graph.tsx
+++ b/packages/perseus/src/widgets/interactive-graph.tsx
@@ -245,9 +245,9 @@ class LegacyInteractiveGraph extends React.Component<Props, State> {
         }
     }
 
-    private _getShouldShowInstructions: (arg1?: Props | null | undefined) => boolean = (
-        props,
-    ) => {
+    _getShouldShowInstructions: (
+        arg1?: Props | null | undefined,
+    ) => boolean = (props) => {
         props = props || this.props;
         return (
             this.isClickToAddPoints(props) &&
@@ -258,12 +258,12 @@ class LegacyInteractiveGraph extends React.Component<Props, State> {
 
     deprecatedProps: any = deprecatedProps;
 
-    private setGraphie: (arg1: any) => void = (newGraphie) => {
+    setGraphie: (arg1: any) => void = (newGraphie) => {
         this.graphie = newGraphie;
         this.setupGraphie();
     };
 
-    private handleAddPointsMouseDown: (arg1: Coord) => void = (coord) => {
+    handleAddPointsMouseDown: (arg1: Coord) => void = (coord) => {
         // This function should only be called when this.isClickToAddPoints()
         // is true
         if (!this.isClickToAddPoints()) {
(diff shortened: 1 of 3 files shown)


packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -37,8 +38,12 @@ const renderGraph = (props: {
     }
 };
 
+export interface HasUserInput {
+    getUserInput(): PerseusGraphType
+}
+
 export const MafsGraph = React.forwardRef<
-    Partial<Widget>,
```
nedredmond, line 41: The intent of using the Widget type, which was @jeremywiebe's request, was that we would eventually implement the other methods (focus, blur, etc) on this component. This change makes that more difficult in the future.

packages/perseus/src/widgets/interactive-graph.tsx
```diff
@@ -934,7 +933,7 @@ class InteractiveGraph extends React.Component<Props, State> {
         return point;
     };
 
-    removePoint: (arg1: Coord) => number | null | undefined = (point) => {
```
nedredmond, line 937: As I'm sure you're learning now, all of these functions were used by the editor to add/remove features from the graphs. What you're trying to do would be a much bigger refactor.

nedredmond (changes requested):
This fundamentally breaks the editor. I'm not sure the juice is worth the squeeze here-- it's a pretty minor DX issue. We can just refresh the page to get the legacy graph back.

packages/perseus/src/widgets/interactive-graph.tsx
benchristel: I tested the editor in storybook and didn't see any obvious problems. It seems like making these methods private shouldn't break anything, because although the `InteractiveGraphEditor` component retains a ref to the `InteractiveGraph`, it only calls the `getUserInput` method on it (https://github.com/Khan/perseus/blob/main/packages/perseus-editor/src/widgets/interactive-graph-editor.tsx#L624-L627).

Where are these `add`/`update`/`remove` methods called, outside the InteractiveGraph class itself?

nedredmond: Oh, looks like I misread these errors
Screenshot 2024-03-18 at 4 24 39 PM

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
benchristel: I've changed this back to use `Partial<Widget>` I think we lose some type safety by doing that (the Widget type uses `any` for some fields) but it's not a big deal.

benchristel:
@nedredmond what are your current thoughts on this? I've addressed your comments. Do you still think it's too risky, or can we land this?

PR #1092, opened 2024-03-19
Ensure that our mobile math input can be visibly focused via tab

SonicScrewdriver:
## Summary:
This PR updates our Math Input to ensure that we're properly displaying the focus styling on the input when it is focused via a tab event. This was done by adding a state for `showInputFocusStyle` that is separate from the original `focused` state. 

This was required, as we have several cases where we _don't_ want to remove focus styling from the input, such as when the user moves the focus to the keypad by tabbing or clicking a button. The `focused` state will now indicate whether the input is the actively focused DOM element, and `showInputFocusStyle` will indicate whether we want the input to _appear_ focused. This allows for a smoother and more natural interaction between our Math Input and our Keypad. 

Issue: MOB-5431

## Video example of new focus handling:
https://github.com/Khan/perseus/assets/12463099/cc6ab02b-5c9b-46da-9bc3-d36580869b50

## Test plan:
- New test 
- Automatic tests
- Manual Testing

### DIFF
--- a/packages/math-input/src/components/input/math-input.tsx
+++ b/packages/math-input/src/components/input/math-input.tsx
@@ -626,9 +626,15 @@ class MathInput extends React.Component<Props, State> {
             this.focus();
         }
 
+<<<<<<< HEAD
         // If the user clicked on the input using a mouse or tap gesture,
         // we want to set the focus to the inputRef so that the keyUp
         // event will be triggered when the user types on the keyboard.
+=======
+        // If the user clicked on the input using a mouse, we want to
+        // set the focus to the inputRef so that the keyUp event will
+        // be triggered when the user types on the keyboard.
+>>>>>>> 92f3bd38 (Fix for the issue where keyboard events were not firing when the user uses a mouse to click into the input.)
         // This is necessary to support Chromebooks as they use mobile user
         // agents, do not simulate touch events, and have physical keyboards.
         this.inputRef?.focus();
@@ -668,9 +674,15 @@ class MathInput extends React.Component<Props, State> {
             this.focus();
         }
 
+<<<<<<< HEAD
         // If the user clicked on the input using a mouse or tap gesture,
         // we want to set the focus to the inputRef so that the keyUp
         // event will be triggered when the user types on the keyboard.
+=======
+        // If the user clicked on the input using a mouse, we want to
+        // set the focus to the inputRef so that the keyUp event will
+        // be triggered when the user types on the keyboard.
+>>>>>>> 92f3bd38 (Fix for the issue where keyboard events were not firing when the user uses a mouse to click into the input.)
         // This is necessary to support Chromebooks as they use mobile user
         // agents, do not simulate touch events, and have physical keyboards.
         this.inputRef?.focus();
--- a/packages/math-input/src/components/input/math-input.tsx
+++ b/packages/math-input/src/components/input/math-input.tsx
@@ -626,15 +626,9 @@ class MathInput extends React.Component<Props, State> {
             this.focus();
         }
 
-<<<<<<< HEAD
         // If the user clicked on the input using a mouse or tap gesture,
         // we want to set the focus to the inputRef so that the keyUp
         // event will be triggered when the user types on the keyboard.
-=======
-        // If the user clicked on the input using a mouse, we want to
-        // set the focus to the inputRef so that the keyUp event will
-        // be triggered when the user types on the keyboard.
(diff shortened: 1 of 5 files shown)


packages/math-input/src/components/input/math-input.tsx
```diff
@@ -328,11 +336,21 @@ class MathInput extends React.Component<Props, State> {
...
+                handle: {
+                    visible: false,
+                },
+                focused: prevState.focused,
+            };
+        });
     };
 
-    focus: () => void = () => {
+    focus: (setKeypadActive: (keypadActive: boolean) => void) => void = (
```
SonicScrewdriver, line 351: Now whenever we focus the input, we ensure the keypad is active.

packages/math-input/src/components/input/math-input.tsx
```diff
@@ -328,11 +336,21 @@ class MathInput extends React.Component<Props, State> {
 
     blur: () => void = () => {
         this.mathField.blur();
-        this.props.onBlur && this.props.onBlur();
-        this.setState({focused: false, handle: {visible: false}});
+
```
SonicScrewdriver, line 339: We don't want to be calling the props.onBlur method here as that automatically closes the keypad. This results in glitchy behaviour as clicking on a keypad button or tabbing away from the math input would result in the keypad closing.

packages/math-input/src/components/input/math-input.tsx
```diff
@@ -973,6 +993,10 @@ class MathInput extends React.Component<Props, State> {
                             ref={(node) => {
                                 this.inputRef = node;
                             }}
+                            onFocus={() => {
```
SonicScrewdriver, line 996: Updated this to ensure that we're making proper use of the focus and blur options, so that we can support adding focus styling when a user tabs to focus a math input. As a result, some minor changes were required to the blur and focus methods.

packages/math-input/src/components/input/math-input.tsx
```diff
@@ -183,6 +188,8 @@ class MathInput extends React.Component<Props, State> {
             // multi-touch.
             if (this.state.focused && this.didTouchOutside && !this.didScroll) {
                 this.blur();
+                this.mathField.blur();
+                this.props.onBlur && this.props.onBlur();
```
handeyeco, line 192: Just a minified version

```suggestion
               this.props.onBlur?.();
```

packages/math-input/src/components/input/math-input.tsx
```diff
@@ -328,11 +336,21 @@ class MathInput extends React.Component<Props, State> {
...
+
+        this.setState((prevState) => {
+            return {
+                showInputFocusStyle: false,
+                handle: {
+                    visible: false,
+                },
+                focused: prevState.focused,
+            };
+        });
```
handeyeco, line 348: I don't think the callback is needed. If you don't pass `focused` React should just let it be.

```suggestion
        this.setState({
                showInputFocusStyle: false,
                handle: {
                    visible: false,
                },
        });
```

packages/math-input/src/components/input/math-input.tsx
handeyeco, line 351: It looks like we use `(keypadActive: boolean) => void` in a few places and this particular type of nested functions is hard to read. I think it's worth making a defined type for this that can be used throughout this file.

handeyeco (approved):
Seems good to me!

This work really highlights a pain point in this code (a pain point we inherited, not as a result of your work): I don't think we can nest the `input/math-input` component within another React component because of the non-React way we stick methods onto components and access them via refs. If this were written in the React style, we could have done something like:

``` JSX
export default function WrappedMathInput() {
  return (
    <KeypadContext.Consumer>
      {({keypadActive, setKeypadActive}) => (
        <MathInput
          keypadActive={keypadActive}
          setKeypadActive={setKeypadActive} />
      )}
    </KeypadContext.Consumer>
  )
}
```

Then we'd have access to this information throughout the MathInput component as props. Instead we're bending over backwards trying to get this stuff because we have to wrap the JSX in the render method. 😭

Again, not a reflection of your work, it just makes me sad.

packages/math-input/src/components/input/math-input.tsx
SonicScrewdriver: Ooo good callout! This is all coming from the KeypadContext Provider, so I've made sure to update all the locations to use either `KeypadContextType['setKeypadActive']` or `KeypadContextType['keypadActive']`

PR #1096, opened 2024-03-19
Upgrade all eslint-related libraries to latest

jeremywiebe:
## Summary:

This PR upgrades our ESLint tooling (and related dev libraries) to their latest version. In doing so I've also removed eslint rules that are on, by default.

I did need to run 'eslint --fix' afterwards which applied a few formatting changes (due to the Prettier upgrade) but they all seemed pretty minor.

Note: the reduction in coverage is the result of indentation due to Prettier v3 formatting changes. In reality coverage hasn't changed. 

Issue: LEMS-1847

## Test plan:

`yarn` (make sure you install first)
`yarn lint`


.eslintrc.js
```diff
@@ -320,16 +320,13 @@ module.exports = {
...
          * testing-library
          */
+        "testing-library/await-async-queries": "error",
         "testing-library/prefer-user-event": "error",
-        "testing-library/no-wait-for-empty-callback": "error",
-        "testing-library/no-wait-for-multiple-assertions": "error",
-        "testing-library/no-wait-for-side-effects": "error",
-        "testing-library/no-wait-for-snapshot": "error",
+        // TODO(jeremy): It might be nice to enable
+        // "testing-library/prefer-explicit-assert": "error",
```
nedredmond, line 326: I'm in favor of this!

.eslintrc.js
```diff
@@ -320,16 +320,13 @@ module.exports = {
         /**
          * testing-library
          */
+        "testing-library/await-async-queries": "error",
         "testing-library/prefer-user-event": "error",
-        "testing-library/no-wait-for-empty-callback": "error",
-        "testing-library/no-wait-for-multiple-assertions": "error",
-        "testing-library/no-wait-for-side-effects": "error",
-        "testing-library/no-wait-for-snapshot": "error",
```
nedredmond, line 327: Can you explain all of these removals?

nedredmond (approved):
Thanks @jeremywiebe !

Just occurred to me last thing to update would be Prettier to make sure it doesn't yell at us for using the new TS features.

.eslintrc.js
jeremywiebe: These are all enabled, by default, in the base `plugin:testing-library/react` ruleset: https://github.com/testing-library/eslint-plugin-testing-library?tab=readme-ov-file#supported-rules

packages/perseus-editor/src/widgets/interaction-editor/interaction-editor.tsx
```diff
@@ -132,20 +132,20 @@ class InteractionEditor extends React.Component<Props, State> {
...
+                          ? _.clone(MovableLineEditor.defaultProps)
+                          : elementType === "function"
+                            ? _.clone(FunctionEditor.defaultProps)
+                            : elementType === "parametric"
+                              ? _.clone(ParametricEditor.defaultProps)
+                              : elementType === "label"
+                                ? _.clone(LabelEditor.defaultProps)
+                                : elementType === "rectangle"
+                                  ? _.clone(RectangleEditor.defaultProps)
+                                  : {},
```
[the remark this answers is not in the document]
jeremywiebe: 😱 I hadn't noticed this beast before.

jeremywiebe:
> Thanks @jeremywiebe !
> 
> Just occurred to me last thing to update would be Prettier to make sure it doesn't yell at us for using the new TS features.

Prettier did get a bump in this PR also :) because the updaed `eslint-plugin-prettier` needed it: https://github.com/Khan/perseus/pull/1096/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R104

.eslintrc.js
jeremywiebe: I'll put up a separate PR and see how disruptive it is, and if we like it.

PR #1097, opened 2024-03-19
Add query params to dev ui

nedredmond:
https://github.com/Khan/perseus/assets/23404711/43efa352-2e46-4780-a640-bd95ce79803d

## Summary
Add search feature and query params to dev ui to prevent needing to reset state and scroll to graph on each fast module reload.

### Issue
N/A

### DIFF
--- a/dev/gallery.tsx
+++ b/dev/gallery.tsx
@@ -2,12 +2,13 @@
 import {useUniqueIdWithMock, View} from "@khanacademy/wonder-blocks-core";
 import {OptionItem, MultiSelect} from "@khanacademy/wonder-blocks-dropdown";
 import {Strut} from "@khanacademy/wonder-blocks-layout";
+import SearchField from "@khanacademy/wonder-blocks-search-field";
 import Spacing from "@khanacademy/wonder-blocks-spacing";
 import Switch from "@khanacademy/wonder-blocks-switch";
 import {color} from "@khanacademy/wonder-blocks-tokens";
 import {css, StyleSheet} from "aphrodite";
 import * as React from "react";
-import {useState} from "react";
+import {useEffect, useMemo, useState} from "react";
 
 import {Renderer} from "../packages/perseus/src";
 import * as grapher from "../packages/perseus/src/widgets/__testdata__/grapher.testdata";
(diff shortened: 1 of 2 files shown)


dev/gallery.tsx
```diff
@@ -76,33 +77,78 @@ const styles = StyleSheet.create({
 
 export function Gallery() {
     const ids = useUniqueIdWithMock();
+    const params = useMemo(
+        () => new URLSearchParams(window.location.search),
+        [],
```
jeremywiebe, line 82: **question**: Does `window.location.search` need to be a dependency?  Otherwise won't the memo only be calculated once ever?

dev/gallery.tsx
```diff
@@ -113,16 +159,49 @@ export function Gallery() {
...
-                                flags: {mafs: mafsFlagsObject},
-                            }}
-                        />
-                    ))}
+                    {questions
+                        .filter((question) => {
+                            if (!search) {
+                                return true;
+                            }
+                            const widgetKey = Object.keys(question.widgets)[0];
```
jeremywiebe, line 167: **thought**: We have some data in our example corpus where there is more than one widget in the `widgets` dict. We can fix this by doing `Object.keys(questions.widgets).some(key => { ... })` and nest this `switch` statement inside that. Otherwise we're only searching on the first (possibly non-interactive-widget) widget in each question.

dev/gallery.tsx
```diff
@@ -113,16 +159,49 @@ export function Gallery() {
...
+                            const widget = question.widgets[widgetKey];
+                            switch (widget.type) {
```
jeremywiebe, line 169: **idea**: Wonder if it'd be worth extracting this matcher function to it's own top-level function to shorten up this component a bit. Right now all the extra indentation is starting to make it difficult to read.

jeremywiebe (approved):
Great idea! Thanks!

dev/gallery.tsx
nedredmond: No, otherwise it'd cause the component to reload every time the URL was updated in the `useEffect`. We only want it to be calculated once.

PR #1099, opened 2024-03-20
Testing Library: enable prefer-explicit-assert rule and fix violations

jeremywiebe:
## Summary:

After updating ESLint & Friends (#1096) to latest, we got a new testing-library/eslint-plugin-testing-library rule named: `testing-library/prefer-explicit-assert`. This rule requires that there is an explicit `expect()` to assert presenve of a DOM node instead of just using the implicit `getBy()`. 

🔉 I'm on the fence if this is a net benefit. **Please weigh in**.

Issue: LEMS-1847

## Test plan:

Review the changes in this PR. Which looks better? Before or after?

### DIFF
--- a/.changeset/three-jars-cheer.md
+++ b/.changeset/three-jars-cheer.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Test updates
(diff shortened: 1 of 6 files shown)


nedredmond (approved):
I'm a big fan of this. Without an explicit assert, there is the potential for a false positive.

PR #1100, opened 2024-03-20
Get the interactive graph flipbook ready for QE testing

benchristel:
QE has a lot on their plate right now, so we need to use their time
efficiently.

This PR makes the Interactive Graph flipbook available via Storybook,
so QE testers will be able to access it at https://khan.github.io/perseus.

I've also made some quality-of-life improvements to the flipbook to make
testing easier. The main one being that you can now see the index of the
question you're viewing, and jump to any question by index.

Issue: none

## Test plan:

- `yarn storybook`
- `open https://localhost:6006`
- Search for "Flipbook"
- You should be able to paste JSON from the WIP
  [test plan](https://khanacademy.atlassian.net/wiki/spaces/LC/pages/2738520299/Test+plan+for+segment+graph)
  into the box and flip through the questions.

### DIFF
--- a/dev/flipbook-model.test.ts
+++ b/dev/flipbook-model.test.ts
@@ -6,6 +6,7 @@ import {
     setQuestions,
     selectQuestions,
     removeCurrentQuestion,
+    jumpToQuestion,
 } from "./flipbook-model";
 
 import type {FlipbookModel} from "./flipbook-model";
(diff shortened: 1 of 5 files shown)


dev/flipbook-model.test.ts
```diff
@@ -75,7 +79,45 @@ describe("flipbookModelReducer", () => {
...
+        const state: FlipbookModel = {questions: "foo", requestedIndex: 0};
+        const action = jumpToQuestion("blah");
+
+        expect(flipbookModelReducer(state, action)).toEqual({
+            questions: "foo",
+            requestedIndex: 0,
+        });
+    });
+
+    it("does nothing given a negative value", () => {});
```
jeremywiebe, line 95: 😄 I'm not sure how much value this test has.

jeremywiebe (approved):
Nice. This'll be handy to have published somewhere.

PR #1104, opened 2024-03-20
Migrate ViewportResizer and DeviceFramer to functional components

jeremywiebe:
## Summary:

This PR migrates two small editor components to be functionnal components and adds stories for them. I started into these components because I thought they might be useful in our main ServerItemRenderer stories so that we can easily change the viewport to be what our content editors use. 

  * Migrate ViewportResizer to functional component and add stories for it
  * Migrate DeviceFramer to functional component and add stories for it

Issue: "none"

## Test plan:

### DIFF
--- a/packages/perseus-editor/src/components/device-framer.tsx
+++ b/packages/perseus-editor/src/components/device-framer.tsx
@@ -2,8 +2,7 @@
  * A component that displays its contents inside a device frame.
  */
 
-import {constants} from "@khanacademy/perseus";
-import PropTypes from "prop-types";
+import {DeviceType, constants} from "@khanacademy/perseus";
 import * as React from "react";
 
 const SCREEN_SIZES = {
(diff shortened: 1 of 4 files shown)


packages/perseus-editor/src/components/device-framer.tsx
```diff
@@ -24,67 +25,72 @@ const SCREEN_SIZES = {
...
             height: SCREEN_SIZES[deviceType].height,
             border: "solid 1px #CCC",
             margin: 8,
             zoom: scale,
-        } as const;
+        }),
+        [deviceType, scale],
+    );
 
+    if (nochrome) {
```
jeremywiebe, line 62: The rendering part of this component required some re-organization to migrate it because of the conditional rendering.

packages/perseus-editor/src/components/device-framer.tsx
```diff
@@ -24,67 +25,72 @@ const SCREEN_SIZES = {
...
+        // Render content inside a variable height iframe.  Used on the
+        // "edit" table of the content editor. In this mode, PerseusFrame
+        // will draw the border and reserve space on the right for
+        // lint indicators.
         return (
-            <div key="screen" className="screen" style={screenStyle}>
-                {this.props.children}
+            <div>
```
benchristel, line 68: Do we need this extra wrapper div? I see that it was in the original code, but I wonder what it's for.

benchristel (approved):
LGTM! Thanks!

packages/perseus-editor/src/components/device-framer.tsx
jeremywiebe: No. Looking at git history it looks like this component used to actually draw a frame that looked like a real device.

PR #1105, opened 2024-03-20
Migrate to WB Tokens package

jeremywiebe:
## Summary:

The Wonder Blocks team has moved all color and spacing tokens together into a single WB package. This reduces the number of packages we'll need to reference. 

This PR migrates all usages of the old packages to use `@khanacademy/wonder-blocks-tokens`. There was only one non-trivial change which was changing a `fade(Color.red, 0.8)` call to use a provided color that the new Tokens package provides. 

Issue: "none"

## Test plan:


nedredmond (approved):
Oh I thought we'd already done this, someone assigned it to themselves sprints ago. Thanks Jeremy!

jandrade:
🎉 Thanks @jeremywiebe! Now I'm going to get rid of the now fully deprecated `Color` and `Spacing` packages in the WB repo.

PR #1107, opened 2024-03-20
Spike: we can render TeX labels with MathJax CHTML

benchristel:
Issue: https://khanacademy.atlassian.net/browse/LEMS-1806

## Test plan:
TODO

Co-authored-by: Nicole W. <nicole@khanacademy.org>

### DIFF
--- a/dev/gallery.tsx
+++ b/dev/gallery.tsx
@@ -19,6 +19,7 @@ import type {APIOptions, PerseusRenderer} from "../packages/perseus/src";
 import "../packages/perseus/src/styles/perseus-renderer.less";
 
 const questions = [
+    interactiveGraph.linearQuestionWithLockedPoints,
     interactiveGraph.segmentQuestion,
     interactiveGraph.pointQuestion,
     interactiveGraph.angleQuestion,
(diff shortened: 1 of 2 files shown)


nedredmond (changes requested):
Two things:
- This will render on top of everything in the Mafs graph, which may make using interactive elements difficult. That's one reason using svg between the graph and interactive elements may be desirable. Have you verified that this doesn't interfere with manipulating the interactive elements?
- If we want to use this method, please put it into its own component.

benchristel:
> This will render on top of everything in the Mafs graph, which may make using interactive elements difficult.

I set `pointerEvents: none` on the TeX layer, so you can click through the labels to the interactive elements. Although, now that I'm looking at the legacy graphs, it looks like the labels are rendered underneath the interactive elements.

This is just a proof of concept for future reference, not meant to be merged. I'll close it!

nedredmond:
For future ref, here's the spike to render MathJax as TeX between layers in Mafs
https://github.com/Khan/perseus/blob/c5346a958789b7754d4de514e7509291b3c5d36a/packages/perseus/src/widgets/interactive-graphs/label.tsx

PR #1108, opened 2024-03-21
Graph dev ui: add navigation back to gallery from flipbook

nedredmond:
Makes header its own component, adds it to flipbook, and adds link back to gallery.

To keep item placement consistent and also to make header more flexible, items spread to take up header space and wrap when container becomes too narrow.

Row order is reversed so that the lone item in the flipbook nav is in the same place as in gallery (top right).

https://github.com/Khan/perseus/assets/23404711/3541b0d8-e64e-42a0-a197-da6e9483d5dc

### DIFF
--- a/dev/flipbook.tsx
+++ b/dev/flipbook.tsx
@@ -22,6 +22,7 @@ import {
     selectNumQuestions,
     jumpToQuestion,
 } from "./flipbook-model";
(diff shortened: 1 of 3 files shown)


benchristel (approved):
Thanks for adding this! LGTM.

PR #1109, opened 2024-03-21
Fix test following eslint changes

nedredmond:
One test failed because of an `eslint` error due to new (and welcome!) rules.

An element was asserted as existing by only a query, which violates the rule. Also, the test didn't explicitly assert that the element was focused per the test title. This ensures that the expected element has focus.


Myranae (approved):
Thanks for fixing the test!

PR #1111, opened 2024-03-21
Ensure default segments are positioned correctly on Mafs graphs

benchristel:
## Summary:
Previously, there was a discrepancy between our `getDefaultSegments`
function and the equivalent code in the legacy InteractiveGraph.
`getDefaultSegments` was not "normalizing" the points correctly (see the
test added in this commit for an explanation of why we do this).

This commit ports the legacy behavior over.

Issue: https://khanacademy.atlassian.net/browse/LEMS-1841

Test plan:

See the linked issue for repro steps. The bug should not repro!

### DIFF
--- a/.changeset/twenty-ads-type.md
+++ b/.changeset/twenty-ads-type.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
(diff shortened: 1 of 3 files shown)


nedredmond (approved):
Thanks Ben!

PR #1112, opened 2024-03-21
[🔥AUDIT🔥] Add linear Mafs flag to flipbook

benchristel:
🖍 _This is an audit!_ 🖍

This lets you see the Mafs rendering of linear graphs side-by-side with the
legacy rendering in the flipbook UI.

Issue: none

## Test plan:

```
yarn dev
open http://localhost:5173/#flipbook
grep -rl '"type":"linear"' data/questions/ | xargs cat | pbcopy
```

Paste the data on your clipboard into the flipbook. The graphs on the right
should be rendered by Mafs.

### DIFF
--- a/dev/flipbook.tsx
+++ b/dev/flipbook.tsx
@@ -135,7 +135,15 @@ function SideBySideQuestionRenderer({
                     question={question}
                     apiOptions={{
                         ...apiOptions,
-                        flags: {mafs: {segment: true, linear: true}},
+                        flags: {
+                            mafs: {
+                                segment: true,
+                                linear: true,
+                                "linear-system": true,
+                                polygon: true,
+                                ray: true,
+                            },
+                        },
                     }}
                 />
             </View>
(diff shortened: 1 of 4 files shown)


dev/flipbook.tsx
```diff
@@ -133,7 +133,10 @@ function SideBySideQuestionRenderer({
                 />
                 <GradableRenderer
                     question={question}
-                    apiOptions={{...apiOptions, flags: {mafs: {segment: true}}}}
+                    apiOptions={{
+                        ...apiOptions,
+                        flags: {mafs: {segment: true, linear: true}},
```
nedredmond, line 138: Oh-- I didn't realize we needed to add this here as well! May as well add `linear-system`, `polygon`, and `ray` too, right?
```suggestion
                        flags: {mafs: {segment: true, linear: true, linear-system: true, ray: true, polygon: true}},
```

nedredmond: Prettier will have something to say about the whitespace

nedredmond (commented):
Could you just add all of the currently implemented graphs while you're at it?

nedredmond (approved):
One thing I'd suggest is moving the new file to a `.testdata.ts`, but it's not super important. Thanks!

PR #1116, opened 2024-03-22
Add optional angle label to polygon

nedredmond:
## Summary
Adds optional angel label to Polygon. Isolates to its own component so that it can be reused for the Angle graph.

Points are always facing inside the graph.
Labels adjust their position when too close to the movable points.
Special arc for 90-degree angles.
Like original, omits "≈" when angle evaluates to integer.

https://github.com/Khan/perseus/assets/23404711/331dfb48-cbf0-4af6-8578-6fdb71e41859

### Issue
lems-1854

## Testing
Run `yarn dev`, flip mafs polygon flag, search "poly"

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/components/angle.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/components/angle.tsx
@@ -131,19 +131,54 @@ const shouldDrawArcInside = (
     range: [Interval, Interval],
     polygonPoints: readonly vec.Vector2[],
 ) => {
-    const throughLine = getRangeIntersectionVertex(vertex, midPoint, range);
+    const rangeIntersectionPoint = getRangeIntersectionVertex(
+        vertex,
+        midPoint,
+        range,
+    );
 
     const lines = getLines(polygonPoints);
-    let blackLineIntersections = 0;
+    let lineIntersections = 0;
 
     lines.forEach(
         (line) =>
-            linesIntersect([vertex, throughLine], line) &&
-            blackLineIntersections++,
+            linesIntersect([vertex, rangeIntersectionPoint], line) &&
+            lineIntersections++,
     );
 
+    // The intersection check will sometimes return false if it intersects with
+    // another vertex, so in the case we get 0 intersections, we check for points.
+    if (lineIntersections === 0) {
+        // find point in array
+        const midpointIndex = polygonPoints.findIndex(
+            ([x, y]) => x === midPoint[0] && y === midPoint[1],
+        );
+        // get array sans point and adjacent points
+        const [prev, next] = [
+            (midpointIndex - 1 + polygonPoints.length) % polygonPoints.length,
+            (midpointIndex + 1) % polygonPoints.length,
+        ];
+        const [[lineAx, lineAy], [lineBx, lineBy]] = [
+            midPoint,
+            rangeIntersectionPoint,
+        ];
+        for (let i = 0; i < polygonPoints.length; i++) {
+            if ([prev, midpointIndex, next].includes(i)) {
+                continue;
+            }
+            const [x, y] = polygonPoints[i];
+            const intersectsPoint =
+                (y - lineAy) * (lineBx - lineAx) ===
+                (lineBy - lineAy) * (x - lineAx);
+            if (intersectsPoint) {
+                return false;
+            }
+        }
+        return true;
+    }
+
     // If the number of intersections is even, the angle is inside the polygon
-    return isEven(blackLineIntersections);
+    return isEven(lineIntersections);
 };
 
 const getLines = (points: readonly vec.Vector2[]): CollinearTuple[] =>


packages/perseus/src/widgets/interactive-graphs/graphs/components/angle.tsx
```diff
@@ -0,0 +1,173 @@
...
+    const a = vec.dist(centerPoint, endPoints[0]);
+    const b = vec.dist(centerPoint, endPoints[1]);
+    const c = vec.dist(endPoints[0], endPoints[1]);
+
+    // Law of cosines
+    const angle = Math.acos((a ** 2 + b ** 2 - c ** 2) / (2 * a * b));
+
+    const startAngle = Math.atan2(startY - centerY, startX - centerX);
+    const endAngle = Math.atan2(endY - centerY, endX - centerX);
+    // const normalizedAngle = angle > 2 * Math.PI ? angle - 2 * Math.PI : angle;
```
benchristel, line 43:
```suggestion
```

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle.tsx
```diff
@@ -0,0 +1,173 @@
...
+    const x1 = centerX + radius * Math.cos(startAngle);
+    const y1 = centerY + radius * Math.sin(startAngle);
+    const x2 = centerX + radius * Math.cos(endAngle);
+    const y2 = centerY + radius * Math.sin(endAngle);
```
benchristel, line 47: I think there's a simpler way to calculate these `x1,y1,x2,y2` constants that doesn't need trigonometry:

```ts
const y1 = centerY + (startY - centerY) / a * radius;
const x2 = centerX + (endX - centerX) / b * radius;
const x1 = centerX + (startX - centerX) / a * radius;
const y2 = centerY + (endY - centerY) / b * radius;
```

With this approach, you don't need to calculate `startAngle` and `endAngle`.

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle.tsx
```diff
@@ -0,0 +1,173 @@
...
+    range: [Interval, Interval];
+    color?: string;
+}
+
+export const Angle = ({
+    centerPoint,
+    endPoints,
+    range,
+    polygonPoints,
+    color = "var(--movable-line-stroke-color)",
```
benchristel, line 24: Color isn't used.

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+    return (
+        <>
+            <defs>
+                <filter
+                    id="background"
+                    x="-5%"
+                    width="110%"
+                    y="0%"
+                    height="100%"
```
benchristel, line 83: What do these dimensions on the filter do? I tried deleting them and nothing seemed to change...

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle.tsx
```diff
@@ -0,0 +1,173 @@
...
+                }
+            >
+                {angleLabel}°
+            </Text>
+        </>
+    );
+};
+
+const isRightAngle = (angle) => {
+    if (angle > Math.PI) {
```
benchristel, line 124: When would `angle` ever be greater than pi? Could we just return `false` here, or omit this conditional?

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle.tsx
```diff
@@ -0,0 +1,173 @@
...
+        angle = Math.abs(angle - Math.PI);
+    }
+    return Math.abs(angle - Math.PI / 2) < 0.01;
```
benchristel, line 127: I think the tolerance on this should be smaller than `0.01`. 0.01 radians is ~0.5 degrees.
Screen Shot 2024-03-22 at 9 33 38 AM

packages/perseus/src/widgets/interactive-graphs/graphs/use-transform.ts
```diff
@@ -0,0 +1,10 @@
+import {useTransformContext, vec} from "mafs";
+
+export const useTransform = (...points: vec.Vector2[]) => {
+    const {viewTransform, userTransform} = useTransformContext();
+    const transformToPx = vec.matrixMult(viewTransform, userTransform);
+    const transformedPoints = points.map((point) =>
+        vec.transform(point, transformToPx),
+    );
+    return transformedPoints;
```
benchristel, line 9: Nice!

packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx
```diff
@@ -50,6 +51,24 @@ export const PolygonGraph = (props: Props) => {
                         : "var(--movable-line-stroke-weight)",
                 }}
             />
+            {showAngles &&
+                points.map((point, i) => {
+                    const pt1 = points.at(i - 1);
+                    const pt2 = points[(i + 1) % points.length];
+                    if (!pt1 || !pt2) {
+                        return null;
```
benchristel, line 59: This return should be unreachable, right? I wonder if we should throw an exception here.

More likely than `pt1` or `pt2` being null is that they might be the same point, which I think will happen if there is only one point in the `points` array. In that case, I think we won't be able to compute an angle, so we probably should return null in that case.

benchristel (commented):
This is pretty cool! I learned some math today.

It would be nice to have unit tests for all the math. I also left some suggestions inline.

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle.tsx
nedredmond: It adds 5% padding on the left and right of the filter.

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle.tsx
nedredmond: Radian math is wild, Ben. I learned a lot yesterday.

packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx
nedredmond: This could come in to play in "unlimited" mode.

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle.tsx
nedredmond: Removed!

benchristel (approved):
I think this is good to go if you've read through my suggestions and decided whether or not to implement them.

Although the bug about angle measurements we're discussing in Slack right now sounds like it might be a blocker. I'll leave fixing that up to you. Let me know if you want another review.

PR #1117, opened 2024-03-22
WIP: move some helpers over

handeyeco:
## Summary:
move test helpers

Issue: LEMS-1845

## Test plan:

### DIFF
--- a/packages/perseus/src/widget-type-utils.ts
+++ b/packages/perseus/src/widget-type-utils.ts
@@ -1,0 +1,81 @@
+import type {
+    PerseusItem,
+    PerseusWidget,
+    PerseusWidgetsMap,
+} from "./perseus-types";
+
+export function getWidgetTypeByWidgetKey(
+    widgetKey: string,
+    widgetMap: PerseusWidgetsMap,
+): string | null {
+    const widget = widgetMap[widgetKey];
+    return widget?.type ?? null;
+}
+
+export function contentHasWidgetType(
+    type: string,
+    content: string,
+    widgetMap: PerseusWidgetsMap,
+): boolean {
+    return getWidgetIdsFromContentByType(type, content, widgetMap).length > 0;
+}
+
+export function getWidgetIdsFromContent(
+    content: string,
+): ReadonlyArray<string> {
+    const output: Array<string> = [];
+    const regex = /\[\[☃ ([A-Za-z0-9- ]+)\]\]/g;
+
+    let match = regex.exec(content);
+    while (match) {
+        output.push(match[1]);
+        match = regex.exec(content);
+    }
+
+    return output;
+}
+
+export function getWidgetIdsFromContentByType(
+    type: string,
+    content: string,
+    widgetMap: PerseusWidgetsMap,
+): ReadonlyArray<string> {
+    const rv: Array<string> = [];
+    const widgetIdsInContent = getWidgetIdsFromContent(content);
+    widgetIdsInContent.forEach((widgetId) => {
+        const widget = widgetMap[widgetId];
+        if (widget?.type === type) {
+            rv.push(widgetId);
+        }
+    });
+    return rv;
+}
+
+export function getWidgetsMapFromItemData(
+    itemData: PerseusItem,
+): PerseusWidgetsMap | null {
+    return itemData?.question?.widgets ?? null;
+}
+
+export function getWidgetFromWidgetMap(
+    widgetId: string,
+    widgetMap: PerseusWidgetsMap,
+): PerseusWidget | null {
+    return widgetMap[widgetId] ?? null;
+}
+
+export function getWidgetsFromWidgetMap(
+    widgetIds: ReadonlyArray<string>,
+    widgetMap: PerseusWidgetsMap,
+): PerseusWidgetsMap {
+    const widgets: PerseusWidgetsMap = {};
+
+    widgetIds.forEach((widgetId) => {
+        const widget = getWidgetFromWidgetMap(widgetId, widgetMap);
+        if (widget) {
+            widgets[widgetId] = widget;
+        }
+    });
+
+    return widgets;
+}
--- a/packages/perseus/src/index.ts
+++ b/packages/perseus/src/index.ts
@@ -94,6 +94,11 @@ export {parsePerseusItem} from "./util/parse-perseus-json";
 export {
     generateTestPerseusItem,
     genericPerseusItemData,
+    generateTestRadioWidget,
+    generateTestInteractiveGraphWidget,
+    generateTestCategorizerWidget,
+    generateTestExpressionWidget,
+    generateTestNumericInputWidget,
 } from "./util/test-utils";
 export {
     convertWidgetNameToEnum,
--- a/packages/perseus/src/util/test-utils.ts
+++ b/packages/perseus/src/util/test-utils.ts
@@ -1,4 +1,11 @@
-import type {PerseusItem} from "../perseus-types";
+import type {
+    CategorizerWidget,
+    ExpressionWidget,
+    InteractiveGraphWidget,
+    NumericInputWidget,
+    PerseusItem,
+    RadioWidget,
+} from "../perseus-types";
 
 export const genericPerseusItemData: PerseusItem = {
     question: {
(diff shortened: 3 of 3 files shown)


packages/perseus/src/util/extract-perseus-data.ts
```diff
@@ -554,24 +562,28 @@ const SUPPORTED_WIDGETS = [
...
 ];
 
-export const getWidgetTypeFromWidgetKey = (widgetKey: string): string => {
-    return widgetKey.split(" ")[0];
-};
-
 /* Verify if the perseus item has supported widgets for automatic scoring */
-export const isWrongAnswerSupported = (widgetKeys: Array<string>): boolean => {
+export const isWrongAnswerSupported = (
+    widgetKeys: Array<string>,
```
Myranae, line 567: Question, nb: Were we trying to move away from using "keys" in relation to the widget name/ID? That's probably something to worry about another time, but just noticed "key" is used a lot here

packages/perseus/src/widget-type-utils.ts
```diff
@@ -0,0 +1,140 @@
...
+ * @param {string} content the string to parse
+ * @param {PerseusWidgetsMap} widgetMap widget ID to widget map
+ * @returns {ReadonlyArray<string>} the widget type (ie "radio")
+ */
+export function getWidgetIdsFromContentByType(
+    type: string,
+    content: string,
+    widgetMap: PerseusWidgetsMap,
+): ReadonlyArray<string> {
+    const rv: Array<string> = [];
```
Myranae, line 81: Question: What does "rv" mean? Would it be better to spell out what it refers to? I think we usually try to stay away from abbreviations from what I remember. Totally fine either way; just wanted to point it out :)

packages/perseus/src/widget-type-utils.test.ts
```diff
@@ -0,0 +1,271 @@
...
+            const content = `[[☃ ${widgetId}]]`;
+            const widgetMap = {
+                [widgetId]: generateTestRadioWidget(),
+            };
+
+            // Act
+            const result = contentHasWidgetType(type, content, widgetMap);
+
+            // Assert
+            expect(result).toBeTrue();
```
Myranae, line 59: Praise: So cool to see this! Love finding the type from the widget. Feels much better.

packages/perseus/src/widget-type-utils.test.ts
```diff
@@ -0,0 +1,271 @@
...
+        });
+
+        it("returns false when found", () => {
```
Myranae, line 62: Comment, nb: The names for the tests are a little confusing. I like how short they are, but it might be better to describe what the test is doing a bit more.

packages/perseus/src/widget-type-utils.test.ts
```diff
@@ -0,0 +1,271 @@
...
+
+            // Assert
+            expect(result.length).toBe(2);
+            expect(result[0]).toBe(widgetId1);
+            expect(result[1]).toBe(widgetId2);
+        });
+
+        it("returns an empty array when nothing is found", () => {
+            // Assemble
+            const content = `hello world`;
```
Myranae, line 110: Suggest
```suggestion
            const content = "hello world";
```
Suggestion, nb:  I don't think it matters, but should this use double quotes when there is no interpolation?

packages/perseus/src/widget-type-utils.ts
```diff
@@ -0,0 +1,140 @@
...
+    return widget?.type ?? null;
+}
+
+/**
+ * Does the content have a specific type of widget?
+ *
+ * @param {string} type the type of the widget in question (ie "radio")
+ * @param {string} content the string to search through
+ * @param {PerseusWidgetsMap} widgetMap widget ID to widget map
+ * @returns {boolean} if the content includes the widget type
```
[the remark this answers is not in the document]
handeyeco: Screenshot 2024-03-26 at 9 11 43 AM

This is the format of JSDoc, the computers know to display it correctly

packages/perseus/src/widget-type-utils.ts
handeyeco: It's a common shorthand for "return value"

packages/perseus/src/util/extract-perseus-data.ts
```diff
@@ -599,7 +611,8 @@ export const getCorrectAnswerForWidgetKey = (
     itemData: PerseusItem,
 ): string | undefined => {
     const rubric = itemData.question.widgets[widgetKey].options;
-    const widgetType = getWidgetTypeFromWidgetKey(widgetKey);
+    const widgetMap = getWidgetsMapFromItemData(itemData) as PerseusWidgetsMap;
+    const widgetType = getWidgetTypeByWidgetId(widgetKey, widgetMap) as string;
```
Myranae: Question, nb: Is the type declaration here because the function could also be null? The function on line 614 already returns something of type PerseusWidgetsMap though, right?

packages/perseus/src/widget-type-utils.ts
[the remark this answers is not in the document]
Myranae: Awesome! I guess it would be more for whoever is reading the comment in place then, but glad the pop-up formats it correctly :D

packages/perseus/src/widget-type-utils.test.ts
```diff
@@ -0,0 +1,271 @@
...
+            // Assert
+            expect(widgetType).toBeNull();
+        });
+    });
+
+    describe("contentHasWidgetType", () => {
+        it("returns true when found", () => {
+            // Assemble
+            const type = "radio";
+            const widgetId = "dont-look-for-type-in-key";
```
Myranae: Suggestion, nb: Oh, just noticed that all these IDs refer to themselves as keys. Should it use "id"?
```suggestion
            const widgetId = "dont-look-for-type-in-id";
```

packages/perseus/src/widget-type-utils.test.ts
handeyeco: I tend to think of a test name as being a combination of the describe + test, in this case: "contentHasWidgetType returns false when found" which covers:

- The name of the function: "contentHasWidgetType"
- What we're searching: "content"
- What we're looking for: "widget type"
- The assertion aka the expected return value: "false" (also "Has" implies a boolean return type)

IMO: I think that + focused functions + focused tests should be enough here unless you think there's something specific that we're missing.

Myranae:
Looks like a lot of checks are failing. Seems to be related to this error: `[!] Error: 'isWidgetKeyInContent' is not exported by packages/perseus/src/util/extract-perseus-data.ts, imported by packages/perseus/src/index.ts`

packages/perseus/src/util/extract-perseus-data.ts
handeyeco: Line 614: getWidgetsMapFromItemData used to return `PerseusWidgetsMap | null` and now returns `PerseusWidgetsMap`. I forgot to remove the `as`.

Line 615: while unlikely, it isn't beyond the realm of possibility that we have `[[radio 1]]` and no `"radio 1"` key in `widgets`. I thought it would be best to make sure consumers handle that possibility. However there's already code that exists that _don't_ handle that possibility. Rather than refactor all their logic (which would be scope creep), I just cast for now.

packages/perseus/src/widget-type-utils.test.ts
Myranae: I think it just sounds confusing to say that the function returns false when it finds the widget type. It returns false when the widget type is _not_ found, right?

Myranae (approved):
Looks like all checks passed! Thanks for replying to my comments~ Love all the tests and the comments and the small helper functions. Looks good :)

packages/perseus/src/widget-type-utils.ts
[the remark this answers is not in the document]
handeyeco: Screenshot 2024-03-26 at 10 05 53 AM

For people reading the comments in the code, editors tend to apply syntax highlighting correctly. It's just GH that doesn't seem to get it.

packages/perseus/src/widget-type-utils.test.ts
handeyeco: I see what you mean. I thought you were talking more generally about the names. This one in particular is wrong, so I fixed it. Thanks for pointing that out.

packages/perseus/src/widget-type-utils.ts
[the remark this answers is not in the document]
Myranae: Thank you for pointing that out!

nicolecomputer (approved):
This is a great step forward, thank you!

PR #1119, opened 2024-03-22
Polygon graph: how right angle whether `showAngles` is true or not

nedredmond:
## Summary
Show right angles any time a right angle is created, whether or not `showAngles` is true, which is the current behavior of `interactiveGraph`

Screenshot 2024-03-22 at 4 42 50 PM

Screenshot 2024-03-22 at 3 47 50 PM

### Issue
LEMS-1857

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/components/angle.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/components/angle.tsx
@@ -13,7 +13,7 @@ interface Props {
     polygonPoints: readonly vec.Vector2[];
     active: boolean;
     range: [Interval, Interval];
-    color?: string;
+    showAngles: boolean;
 }
 
 export const Angle = ({
@@ -21,7 +21,7 @@ export const Angle = ({
     endPoints,
     range,
     polygonPoints,
-    color = "var(--movable-line-stroke-color)",
+    showAngles,
 }: Props) => {
     const [centerX, centerY] = centerPoint;
     const areClockwise = clockwise([centerPoint, ...endPoints]);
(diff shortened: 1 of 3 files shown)


packages/perseus/src/widgets/interactive-graphs/graphs/components/angle.tsx
```diff
@@ -123,6 +134,20 @@ export const Angle = ({
     );
 };
 
+const RightAngleArc = ({start: [x1, y1], vertex: [x2, y2], end: [x3, y3]}) => (
```
benchristel, line 137: Suggestion: should this be called `RightAngleSquare`, since it's not really an arc?

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle.tsx
```diff
@@ -123,6 +134,20 @@ export const Angle = ({
...
+    <g
+        style={{
+            transform: `var(--mafs-view-transform) var(--mafs-user-transform)`,
+        }}
+    >
```
benchristel, line 142: Suggestion: the group with the transform is duplicated here and on line 105 above. Maybe the `g` element could be moved outside the ternary that chooses the square or arc—something like:

```ts
<g
  style={{
    transform: `var(--mafs-view-transform) var(--mafs-user-transform)`,
  }}
>
  {!isInside && isRightAngle(angle) ? (
    <RightAngleArc
      // ...
    />
  ) : (
    <path d={arc} strokeWidth={0.02} fill="none" />
  )}
</g>
```

Actually, since only the `d` property of the path is different, maybe it could even be:

```ts
<g
  style={{
    transform: `var(--mafs-view-transform) var(--mafs-user-transform)`,
  }}
>
  <path
    strokeWidth={0.02}
    fill="none"
    d={
      !isInside && isRightAngle(angle)
        ? square
        : arc
    }
  />
</g>
```

benchristel (approved):
LGTM! I left a couple suggestions inline, which you can take or leave.

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle.tsx
nedredmond: I'm using the right angle square elsewhere-- see the early return in the code above

PR #1120, opened 2024-03-22
Suppress console logging in tests

benchristel:
Some of our tests were logging errors and React warnings to the console, which
made test failures harder to spot. This PR makes the test logs a bit cleaner
(they're still not perfect, though).

This PR will probably be easiest to read one commit at a time. It does the following
things:

- Fixes React warnings related to missing keys
- Fixes warnings about Jest fake timers not being enabled
- Suppresses console output in error-boundary.test.tsx
- Removes a noisy test for the Plotter widget - the test could never fail
- Removes the fake analytics logger from tests; it is now only used in Storybook

Issue: none

## Test plan:

`yarn test`


packages/perseus/src/widgets/__tests__/plotter.test.tsx
```diff
@@ -14,14 +14,6 @@ describe("plotter widget", () => {
...
     });
 
-    it("should snapshot basic question", () => {
-        // Arrange and Act
-        const {container} = renderQuestion(question1);
-
-        // Assert
-        expect(container).toMatchSnapshot("initial render");
-    });
-
```
benchristel, line 24: This test didn't fail even when I threw an error in the plotter's render method. Tests that can't fail are useless, so I opted to delete this one.

packages/perseus/src/widgets/__tests__/__snapshots__/plotter.test.tsx.snap
```diff
@@ -1,245 +0,0 @@
...
-          class="perseus-widget-container widget-nohighlight widget-block"
-        >
-          <svg
-            height="16"
-            role="img"
-            viewBox="0 0 16 16"
-            width="16"
-          >
-            <title>
-              Rendering Error!
```
benchristel, line 232: I don't think this was supposed to be an error...

PR #1122, opened 2024-03-25
Fix reflex angle bug

nedredmond:
## Summary
Fixes bug where reflex angles (greater than 180) flip between inside-outside.
The algorithm used to check if an angle is internal will sometimes give a false negative when directly "facing" another angle (i.e., a line drawn from the center of each angle would hit opposing angle directly).
The algorithm counts the number of intersecting lines from a ray drawn from the middle of the angle to the edge of the graph. If the count is odd, it's an inside angle.

Instead of drawing from the angle through the polygon, we can draw a line away from the polygon, and flip the logic. This makes the situation of angles directly facing each other a distant edge case that is difficult to recreate.

### Issue
LEMS-1858

## Test
Run `yarn dev` and create a reflex angle with any polygon graph with more than 3 sides. The angle arc and measurement should remain inside the polygon and not flip back-and-forth as its angle changes.


benchristel (changes requested):
This feels pretty complex, and as you point out, there's still a (rare) edge case where it doesn't quite work.

I wonder if an approach like the one described [here](https://gamedev.stackexchange.com/questions/160344/algorithms-for-calculating-vertex-normals-in-2d-polygon) would cover all cases and be simpler to implement. The "vertex normals" of a polygon are vectors at each vertex that point toward the outside of the polygon. I think you could use those vectors to reliably determine which side of a vertex is the "inside" and draw the angle there.

image

I'd like to lean on an existing, battle-tested approach like that, if we can. Requesting changes because I think we should at least investigate that approach before committing further to this one.

nedredmond:
@benchristel , I wouldn't say what you have there is simpler than what I've done:
https://stackoverflow.com/questions/217578/how-can-i-determine-whether-a-2d-point-is-within-a-polygon

nedredmond:
And funny enough, I used vertex normals as the rays to count intersections.

benchristel (approved):
Talked this over with @nedredmond and I think we can land this as-is because it does address the bug. In the future we may want to revisit this and use the vertex normals approach but that would basically be a rewrite of the angle-calculation code.

PR #1124, opened 2024-03-25
Enable concurrency on most Github action workflows

jeremywiebe:
## Summary:

This PR enables [concurrency](https://docs.github.com/en/actions/using-jobs/using-concurrency) for the `node-ci-main.yml` and `node-ci.yml` Github Action workflow files. 

For `node-ci-main.yml` concurrency is enabled but `cancel-in-progress` is `false`. This is so that all workflow jobs run for a landed PR before the next PR's changes are processed. This ensures that we get updated coverage metrics for every PR that is landed. 

For `node-ci.yml` concurrency is enabled _and_ `cancel-in-progress` is `true`. This should save us some Github Action processing time and $$$. 

Issue: LEMS-1761

## Test plan:

There's not much to test other than watching how things go on subsequent PRs after this is landed. 

I checked that there are no steps that'd be affected by being interrupted in `node-ci.yml` and turning off concurrency should just make things work better in `node-ci-main.yml` as we'll avoid race conditions that could happen due to two PRs landing very close together.

### DIFF
--- a/.github/workflows/node-ci-main.yml
+++ b/.github/workflows/node-ci-main.yml
@@ -12,7 +12,7 @@ on:
 # workflow.
 # See https://docs.github.com/en/actions/using-jobs/using-concurrency
 concurrency:
-    group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+    group: ${{ github.workflow }}
     cancel-in-progress: false
 
 # Our jobs run like this to minimize wasting resource cycles:
--- a/.github/workflows/node-ci-main.yml
+++ b/.github/workflows/node-ci-main.yml
@@ -7,6 +7,14 @@ on:
             - changeset-release/*
             - feature/*
 
+# When a PR is landed to main, we want to ensure that all of these jobs run and
+# that they all complete before the next PR's changes are processed by this
+# workflow.
+# See https://docs.github.com/en/actions/using-jobs/using-concurrency
+concurrency:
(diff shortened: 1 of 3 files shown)


.github/workflows/node-ci-main.yml
```diff
@@ -7,6 +7,14 @@ on:
             - changeset-release/*
             - feature/*
 
+# When a PR is landed to main, we want to ensure that all of these jobs run and
+# that they all complete before the next PR's changes are processed by this
+# workflow.
+# See https://docs.github.com/en/actions/using-jobs/using-concurrency
+concurrency:
+    group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
jeremywiebe, line 15: I _think_ we could just use `gitHub.ref` for the second part of the group. I don't think we ever have `main` as the PR branch.

benchristel: Should the group just be something like `node-ci-main`? All CI runs for `main` could be in the same group, right?

Not that it should really matter, since cancel-in-progress is false.

jeremywiebe: Actually, I think having the group just be `node-ci-main` is a good idea.  Although we have `cancel-in-progress: false`, we are saying that we are concurrent on the workflow (`node-ci-main`) and the PR number or Ref.  

So if multiple PRs land close together on main, they'd have different PR #s and/or refs which isn't what we want.  Thanks for calling this out!

PR #1128, opened 2024-03-27
Use new widget props callback when available for radio widget in `injectWidgets`

nedredmond:
## Summary
Allows Khanmigo to know the order of radio widget answers when helping learners with exercises. Also, adds the option letter to the prompt context.
If there are no widget props available, there is no change in behavior.
Also adds tests!

### Issue
TUT-342

## Testing
There are some new unit tests, but this will also be tested using the NPM preview in webapp.


.changeset/big-weeks-fly.md
sxkosone, line 1: What an interesting name! Are these auto-generated?

packages/perseus/src/util/extract-perseus-data.test.ts
```diff
@@ -0,0 +1,96 @@
...
+            const content = "[[☃ Radio 1]]";
+            const widgets = {
+                "Radio 1": {
+                    type: "radio",
+                    options: {
+                        choices: [
+                            {content: "option 1"},
+                            {content: "option 2"},
+                            {content: "option 3"},
+                        ] satisfies PerseusRadioChoice[],
```
sxkosone, line 17: TIL about `satisfies`!

sxkosone (commented):
Nice tests!

SonicScrewdriver (approved):
We worked through this together! :D

.changeset/big-weeks-fly.md
nedredmond: Yeah, `changesets` auto-generates names. They can get pretty out there!

packages/perseus/src/util/extract-perseus-data.test.ts
nedredmond: I love it!

packages/perseus/src/util/extract-perseus-data.ts
```diff
@@ -233,6 +232,28 @@ function getAnswersFromWidgets(
...
+ * @param {number} index
+ * @returns {string}
+ * @example
+ * toOptionLetter(0) // 'A'
+ * toOptionLetter(1) // 'B'
+ * toOptionLetter(25) // 'Z'
+ * // Once the index goes past 25, it will start returning special characters
+ * toOptionLetter(26) // '['
+ */
+const toOptionLetter = (index: number): string =>
```
jeremywiebe, line 254: I'm late to this, but the function that the radio widget uses is already available for use: https://github.com/Khan/perseus/blob/53fd3768652b702df4a79b97a0ab4766b11748fe/packages/perseus/src/widgets/radio/util.ts#L8

packages/perseus/src/util/extract-perseus-data.ts
```diff
@@ -269,14 +291,25 @@ function injectWidgets(
         switch (widget.type) {
             case "radio":
                 // Replace radio with the radio options
-                // '[[☃ Radio 1]]' -> 'option 1\noption 2\noption 3'
+                // '[[☃ Radio 1]]' -> 'choice 1\nchoice 2\nchoice 3'
+                // or if the current widget state is available with proper order,
+                // '[[☃ Radio 1]]' ->
+                //   'Option A: choice 1\nOption B: choice 2\nOption C: choice 3'
                 const radio = widget;
+                const radioProps = widgetProps?.[widgetID] as
```
jeremywiebe, line 299: I realize that this has already been reviewed and landed, but I'd like to socialize a different approach. 

The radio widget, like all others, supports a `transform` function. This function is the one that does all the heavy lifting to morph the radio widget options into the set that the radio's React component receives and renders. 

So, it would probably mean for more stable less brittle code if we used that transform on the given widget options and then took the `.choices` in the result to derive the order. The only extra bit you'd need is the `problemNum` as that is the randomization seed used. 

https://github.com/Khan/perseus/blob/20a23a3976a70378e86c8871e47d5497ddc9d51e/packages/perseus/src/widgets/radio.ts#L98

Maybe in the future we can refactor to do this.

jeremywiebe (commented):
Sorry, I was late to review. But I think they're worth looking at if you have time. /cc @SonicScrewdriver @handeyeco

packages/perseus/src/util/extract-perseus-data.ts
nedredmond: I see this one translates the letter. Since the rest of the prompt is not translated, I'd be worried about the inconsistency causing some confusion in the completion.

packages/perseus/src/util/extract-perseus-data.ts
```diff
@@ -261,14 +287,26 @@ function injectWidgets(
         switch (widget.type) {
             case "radio":
                 // Replace radio with the radio options
-                // '[[☃ Radio 1]]' -> 'option 1\noption 2\noption 3'
+                // '[[☃ Radio 1]]' -> 'choice 1\nchoice 2\nchoice 3'
+                // or if the serialized state is available,
+                // '[[☃ Radio 1]]' ->
+                //   'Option A: choice 1\nOption B: choice 2\nOption C: choice 3'
                 const radio = widget;
+                const radioSerializedState = serializedState?.question?.[
```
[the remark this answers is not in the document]
nedredmond: @SonicScrewdriver walked me through the preferred method from the perseus data team-- a new callback she added in the last few weeks.

packages/perseus/src/util/extract-perseus-data.ts
jeremywiebe: i18n, that's a good point! I didn't notice this one wasn't translated. This differs in that it can extend past `Z`, but I am 99.9% sure that a radio widget with > 26 choices is problematic in many other ways! :)

PR #1129, opened 2024-03-27
Upgrade @changesets/ tooling to latest

jeremywiebe:
## Summary:

I have been chasing an issue where bumping to a snapshot version sometimes yields and inconsistent set of updates (some snapshotted packages don't reference the snapshot version of a dependency). 

For example: In one case I bumped to a snapshot version where `perseus`, `perseus-editor` and `math-input` were all modified. As a result, `perseus-editor` should be pointing to the snapshot version of `math-input`. But in the case I encountered, it was pointing to the unchanged, production version.  

I am bumping changesets to the latest v2.27.1 (see https://github.com/changesets/changesets/releases?q=cli&expanded=true). I think v2.23.1 fixed the issue (see [release notes](https://github.com/changesets/changesets/releases/tag/%40changesets%2Fcli%402.23.1)), specifically: 

> https://github.com/changesets/changesets/pull/857 [7febb59](https://github.com/changesets/changesets/commit/7febb599167234ae071b5d223b80cbc8a9375709) Thanks [@dotansimha](https://github.com/dotansimha)! - Fixed an issue with dependency ranges still using pre-existing range modifiers instead of fixed package versions when performing a snapshot release. This ensures that installs of snapshot versions are always reproducible.

Issue: "none"

## Test plan:

1. Make some changes to perseus, perseus-editor, and math-input
2. Stage all the changes: `git add .`
3. Run the command to snapshot version the packages: `yarn changeset version --snapshot PR0000`
4. `git diff` - verify that each package that references a modified dependency is using the snapshot version


jeremywiebe:
This wasn't actually needed to fix the issue I was chasing! Closing.

PR #1131, opened 2024-03-27
Adjust Explanation Widget to Handle Lots of Text in its Button

mark-fitzgerald:
## Summary:
Some implementations of the Explanation widget have a lot of text that is placed in the button that toggles the content visibility. The current version of the widget does not account for the available width in those cases, and this causes the button to extend beyond its parent bounds which results in horizontal scrolling. There are options in the WonderBlocks `<Button>` component to handle these situations gracefully by word-wrapping the button content to fit the available width (see **Before** and **After** screenshots for comparison).

Issue: LEMS-1838

## Test plan:
1. Open Perseus Storybook (locally: http://localhost:6006/?path=/story/perseus-widgets-explanation--wide-button)
   * Note that the disclosure button has a large amount of text in it.
   * The text could be wrapping to multiple lines - this is the desired outcome.
   * Shrinking the viewport width should cause the disclosure button to shrink at some point, along with the surrounding paragraphs. (You may have to reduce the width by a significant amount to see this result.)

### Before:

Wide Button - Before

### After:

Wide Button - After


packages/perseus/src/widgets/__testdata__/explanation.testdata.ts
```diff
@@ -125,3 +125,46 @@ export const ipsumExample: PerseusRenderer = {
...
+                Fluctuations in energy readings from it, Captain. 
+                All transporters off. 
+                A strange set-up, but I'd say the graviton generator is depolarized. 
+                The dark colourings of the scrapes are the leavings of natural rubber, 
+                    a type of non-conductive sole used by researchers experimenting with electricity. 
+                The molecules must have been partly de-phased by the anyon beam.
+                \n[[\u2603 explanation 1]]\n\nSensors indicate no shuttle or other ships in this sector.
+                According to coordinates, we have travelled 7,000 light years and are located near [the system J-25](#).
+                Tractor beam released, sir. Force field maintaining our hull integrity.
+                Damage report? Sections 27, 28 and 29 on decks four, five and six destroyed.
```
jeanettehead, line 140: excellent test data

packages/perseus/src/widgets/explanation.tsx
```diff
@@ -91,9 +91,17 @@ class Explanation extends React.Component<Props, State> {
...
             marginLeft: "-2px",
-            padding: "0 2px",
+            marginRight: "2px",
+            paddingLeft: "2px",
+        };
+
+        const labelStyle = {
+            fontSize: "18px",
+            lineHeight: "20px",
+            marginRight: "-6px",
```
benchristel, line 103: What does this negative margin do? It feels a bit strange that we are adding `marginRight: 2px` to the button but then removing margin from the label. Might be worth a comment explaining why we do this.

benchristel (approved):
LGTM! I left one question inline.

packages/perseus/src/widgets/explanation.tsx
mark-fitzgerald: Comments added.

packages/perseus/src/widgets/__testdata__/explanation.testdata.ts
mark-fitzgerald: Thank you! I think everyone should use this kind of [filler content](https://vlad-saling.github.io/star-trek-ipsum/) for their examples. (I might be biased, though.)

PR #1132, opened 2024-03-27
Make `coords` required on interactive graph state types

benchristel:
## Summary:
This lets us remove some null checks.

The new types are more specific about exactly which fields are relevant
to the state. Previously, some of the state types included fields from
the graph definitions that were not needed. I'm hopeful that the new
types will be more straightforward for new contributors to the project
to understand.

Issue: none

Test plan:

`yarn tsc && yarn test`


mark-fitzgerald (approved):
Much cleaner. Thank you!

PR #1135, opened 2024-03-28
Flipbook: persist questions input in local storage while working

handeyeco:
## Summary:
This is more a POC to check interest, but I noticed that when I was making changes in Perseus that Flipbook wasn't keeping the questions data I had given it. That meant I had to:

1. Make a change
2. Wait for Flipbook to reload
3. Copy/paste data
4. Check my work
5. Repeat

By stashing the data in local storage, the data persists after a reload.

### DIFF
--- a/dev/flipbook.tsx
+++ b/dev/flipbook.tsx
@@ -23,6 +23,7 @@ import {
     selectNumQuestions,
     jumpToQuestion,
     selectCurrentQuestionAsJSON,
+    loadQuestionsFromStorage,
 } from "./flipbook-model";
 import {Header} from "./header";
 
@@ -59,13 +60,14 @@ export function Flipbook() {
     const noTextEntered = questionsState === "";
 
     useEffect(() => {
-        const localStorageQuestions = localStorage.getItem(LS_QUESTIONS_KEY);
-        if (noTextEntered && localStorageQuestions) {
-            dispatch(setQuestions(localStorageQuestions));
-        }
+        const localStorageQuestions =
+            localStorage.getItem(LS_QUESTIONS_KEY) || "";
+        dispatch(loadQuestionsFromStorage(localStorageQuestions));
+    }, []);
 
+    useEffect(() => {
         localStorage.setItem(LS_QUESTIONS_KEY, questionsState);
-    }, [noTextEntered, questionsState]);
+    }, [questionsState]);
 
     function handleDiscardQuestion() {
         dispatch(removeCurrentQuestion);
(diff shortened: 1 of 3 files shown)


dev/flipbook.tsx
```diff
@@ -53,7 +55,22 @@ export function Flipbook() {
...
 
-    const noTextEntered = state.questions.trim() === "";
+    const questionsState = state.questions.trim();
+    const noTextEntered = questionsState === "";
+
+    useEffect(() => {
+        const localStorageQuestions = localStorage.getItem(LS_QUESTIONS_KEY);
+        if (noTextEntered && localStorageQuestions) {
+            dispatch(setQuestions(localStorageQuestions));
+        }
```
benchristel, line 65: I wonder if we could simplify this `useEffect` callback, and get this logic under test, by unconditionally dispatching an event like `loadQuestionsFromStorage` and letting the reducer handle it in different ways based on the state of the question input. Something like this:

```ts
useEffect(() => {
  const storedQuestions = localStorage.getItem(LS_QUESTIONS_KEY);
  dispatch(loadQuestionsFromStorage(storedQuestions));
}, [])

useEffect(() => {
  localStorage.setItem(LS_QUESTIONS_KEY, questionsState);
}, [questionsState])
```

dev/flipbook.tsx
```diff
@@ -53,7 +55,22 @@ export function Flipbook() {
...
+
+        localStorage.setItem(LS_QUESTIONS_KEY, questionsState);
+    }, [noTextEntered, questionsState]);
+
+    function handleDiscardQuestion() {
+        dispatch(removeCurrentQuestion);
+        localStorage.removeItem(LS_QUESTIONS_KEY);
```
benchristel, line 72: Is this `removeItem` needed? Discarding a question is supposed to remove just that one question from the list, so I don't think we should clear out the storage.

benchristel (commented):
Thank you for doing this! I left a couple suggestions inline. The one about moving logic to the reducer you can take or leave. I'm pretty sure we shouldn't clear the store when a question is discarded, though.

dev/flipbook.tsx
handeyeco: Open to alternatives, but it works like this:
- Dispatch to remove the one question (React state updates are async IIRC, so really we're queuing an update here)
- Remove the data from local storage
- State update causes `questionsState` to change
- The change to `questionsState` causes `useEffect` to rerun
- The effect stores the updated `questionsState` in local storage

benchristel: Makes sense, but, since the updated `questionsState` will ultimately be stored (overwriting whatever state is already there), why do we need to clear it first?

benchristel (approved):
You've convinced me that my comments should be non-blocking, so I'm approving now. Thanks again for doing this!

dev/flipbook.tsx
handeyeco: Updated

dev/flipbook.tsx
handeyeco: I have my reasons!

Just kidding, I didn't need to write that code. Fixed.

PR #1136, opened 2024-03-28
Storybook: Prevent Vite from inlining SVGs

jeremywiebe:
## Summary:

Mark noticed that the new @phosphor icons were not showing up in Storybook on Github Pages (`khan.github.com/pereseus`). 

I tracked it down to a feature introduced in Vite 5 (vitejs/vite#15534), but it seems like it has bugs ([#15378](https://github.com/vitejs/vite/issues/15378) and [#15444](https://github.com/vitejs/vite/issues/15444)). 

I don't see how we could "quote" the URL as it's handed of to Wonder Blocks which seems to be handling the import correctly. 

So instead we'll just force Vite to never inline SVGs using Vite's `assetsInlineLimit` (https://vitejs.dev/config/build-options#build-assetsinlinelimit) option. 

Issue: LEMS-1887

## Test plan:

I haven't tested on GH Pages yet, but I did the following and verified that there are now svg files in the build output folder:

```sh
yarn build-storybook
ls storybook-static/assets
```

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -4,10 +4,6 @@ import {mergeConfig} from "vite";
 
 import type {StorybookConfig} from "@storybook/react-vite";
 
-function getAbsolutePath(value: string): any {
-    return dirname(require.resolve(join(value, "package.json")));
-}
-
 const config: StorybookConfig = {
     framework: "@storybook/react-vite",
 
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -4,6 +4,10 @@ import {mergeConfig} from "vite";
 
 import type {StorybookConfig} from "@storybook/react-vite";
 
+function getAbsolutePath(value: string): any {
+    return dirname(require.resolve(join(value, "package.json")));
+}
+
 const config: StorybookConfig = {
     framework: "@storybook/react-vite",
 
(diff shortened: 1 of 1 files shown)


.storybook/main.ts
```diff
@@ -55,6 +70,9 @@ const config: StorybookConfig = {
         });
     },
 
+    core: {
+        disableTelemetry: true,
+    },
```
benchristel, line 75: Is this change related to the others?

.storybook/main.ts
```diff
@@ -4,6 +4,10 @@ import {mergeConfig} from "vite";
 
 import type {StorybookConfig} from "@storybook/react-vite";
 
+function getAbsolutePath(value: string): any {
+    return dirname(require.resolve(join(value, "package.json")));
+}
```
benchristel, line 9: I think this `getAbsolutePath` function is unused.

benchristel (approved):
I left a couple notes inline on changes that might not be needed, but otherwise this looks good. Thanks!

.storybook/main.ts
jeremywiebe: It's not.  I had been comparing to the WB config and it had that in there. I'll remove it and put it in a separate PR.

.storybook/main.ts
jeremywiebe: 🤦‍♂️ You're right. While trying to figure out the issue I pulled in this function as it was also in the WB config (they're on Storybook 8 though). I'll clean this up. Thanks for catching that.

PR #1137, opened 2024-03-29
Add arrows to graph grid

handeyeco:
## Summary:
Add arrows to axes 

Screenshot 2024-03-29 at 3 54 39 PM

Issue: [LEMS-1879](https://khanacademy.atlassian.net/browse/LEMS-1879)

- Wrapped our Mafs stuff in a context provider for graph state
- Created a hook to help get graph state: `useGraphState`
- Used the graph state to get range information
- Used the range information to place arrows

## Test plan:

[LEMS-1879]: https://khanacademy.atlassian.net/browse/LEMS-1879?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/grid-arrows.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/grid-arrows.tsx
@@ -1,0 +1,50 @@
+import {useTransformContext, vec} from "mafs";
+
+import useGraphState from "./reducer/use-graph-state";
+
+type ArrowProps = {
+    x: number;
+    y: number;
+    rotate: number;
+};
+
+function Arrow(props: ArrowProps) {
+    const {userTransform, viewTransform} = useTransformContext();
+
+    const point: vec.Vector2 = [props.x, props.y];
+    const userTransformedPoint = vec.transform(point, userTransform);
+    const viewTransformedPoint = vec.transform(
+        userTransformedPoint,
+        viewTransform,
+    );
+
+    return (
+        <g
+            transform={`translate(${viewTransformedPoint[0]} ${viewTransformedPoint[1]}) rotate(${props.rotate}) scale(1.25)`}
+        >
+            <g transform="translate(-1)">
+                <path
+                    d="M-3 4 C-2.75 2.5 0 0.25 0.75 0C0 -0.25 -2.75 -2.5 -3 -4"
+                    fill="none"
+                />
+            </g>
+        </g>
+    );
+}
+
+export default function GridArrows() {
+    const {state} = useGraphState();
+
+    const range = state.range;
+    const [xMin, xMax] = range[0];
+    const [yMin, yMax] = range[1];
+
+    return (
+        <>
+            <Arrow x={xMax} y={0} rotate={0} />
+            <Arrow x={0} y={yMin} rotate={90} />
+            <Arrow x={xMin} y={0} rotate={180} />
+            <Arrow x={0} y={yMax} rotate={270} />
+        </>
+    );
+}
--- a/packages/perseus/src/widgets/interactive-graphs/grid-arrows.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/grid-arrows.tsx
@@ -1,4 +1,5 @@
 import {useTransformContext, vec} from "mafs";
+import React from "react";
 
 import useGraphState from "./reducer/use-graph-state";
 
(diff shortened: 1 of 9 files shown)


packages/perseus/src/widgets/interactive-graphs/grid-arrows.tsx
```diff
@@ -0,0 +1,51 @@
...
+        viewTransform,
+    );
+
+    return (
+        <g
+            transform={`translate(${viewTransformedPoint[0]} ${viewTransformedPoint[1]}) rotate(${props.rotate}) scale(1.25)`}
+        >
+            <g transform="translate(-1)">
+                <path
+                    d="M-3 4 C-2.75 2.5 0 0.25 0.75 0C0 -0.25 -2.75 -2.5 -3 -4"
```
handeyeco, line 28: This path is from `util/graphie.ts`.

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -61,54 +62,63 @@ export const MafsGraph = React.forwardRef<
     }));
 
     return (
-        <View
-            style={{
-                width,
-                height,
-                position: "relative",
+        <GraphStateContext.Provider
```
handeyeco, line 65: I don't know why GH is being so silly about this file. I think the only thing I changes was wrapping existing things in the Provider.

packages/perseus/src/widgets/interactive-graphs/reducer/use-graph-state.ts
```diff
@@ -0,0 +1,18 @@
...
+
+import type {InteractiveGraphAction} from "./interactive-graph-action";
+import type {InteractiveGraphState} from "../types";
+
+export const GraphStateContext = createContext<{
+    state: InteractiveGraphState;
+    dispatch: React.Dispatch<InteractiveGraphAction>;
+}>({
+    state: {} as any,
+    dispatch: () => {},
```
handeyeco, line 11: TS makes me provide defaults, but this will immediately get overwritten when the provider component provides values for these.

packages/perseus/src/widgets/interactive-graphs/grid.tsx
```diff
@@ -26,8 +28,11 @@ const axisOptions = (
...
-        <Coordinates.Cartesian
-            xAxis={axisOptions(props, 0)}
-            yAxis={axisOptions(props, 1)}
-        />
+        <>
+            <Coordinates.Cartesian
+                xAxis={axisOptions(props, 0)}
+                yAxis={axisOptions(props, 1)}
+            />
+            <GridArrows />
```
benchristel, line 36: I'm not sure `props.markings === "none"` is the right condition to check for this. There are three possible settings for `markings`: "graph", "grid", and "none", I believe. I think we only render the axes when `markings` is "graph".

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
benchristel: You can get a cleaner diff if you check the "hide whitespace" box:

Screen Shot 2024-04-01 at 11 14 30 AM

benchristel (commented):
This looks great! I left one question inline about whether we want to show the axis arrows when `markings` is `grid`.

packages/perseus/src/widgets/interactive-graphs/grid-arrows.tsx
```diff
@@ -0,0 +1,51 @@
...
+                    fill="none"
+                />
+            </g>
+        </g>
+    );
+}
+
+export default function GridArrows() {
```
benchristel, line 36: Minor: I wonder if it would make more sense to call this `AxisArrows`, since the arrows go on the axes, not the grid per se.

packages/perseus/src/widgets/interactive-graphs/reducer/use-graph-state.ts
```diff
@@ -0,0 +1,18 @@
...
+});
+
+export default function useGraphState() {
+    const {state, dispatch} = React.useContext(GraphStateContext);
+
+    return {state, dispatch};
```
benchristel, line 17: Why do we destructure `state` and `dispatch` and then bundle them up into a new object to return them? Could we do this instead?

```suggestion
export default function useGraphState() {
    return React.useContext(GraphStateContext);
```

benchristel (approved):
LGTM! Thanks!

PR #1138, opened 2024-03-29
Refactor the Mafs interactive graph state to accommodate circle graphs

benchristel:
This doesn't actually implement the circle graph; that will be done
in a future PR.

This PR adds a library, `tiny-invariant`, which helps us narrow TypeScript types
in tests. We already depend on this library indirectly via react-router and Mafs,
so we don't need an ADR to use it.

Issue: https://khanacademy.atlassian.net/browse/LEMS-1828

## Test plan:

`yarn test`

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
@@ -40,6 +40,8 @@ const renderGraph = (props: {
             return <PolygonGraph graphState={state} dispatch={dispatch} />;
         case "point":
             return <PointGraph graphState={state} dispatch={dispatch} />;
+        case "circle":
+            throw new Error("the circle graph type is not yet implemented");
         default:
             return new UnreachableCaseError(type);
     }
--- a/packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
@@ -7,7 +7,11 @@ import {
     MOVE_ALL,
     MOVE_CONTROL_POINT,
     MOVE_LINE,
-    MOVE_POINT, MoveAll, MoveControlPoint, MoveLine, MovePoint,
+    MOVE_POINT,
+    MoveAll,
+    MoveControlPoint,
+    MoveLine,
+    MovePoint,
 } from "./interactive-graph-action";
 
 import type {CollinearTuple} from "../../../perseus-types";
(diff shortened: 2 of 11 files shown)


packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -40,6 +40,8 @@ const renderGraph = (props: {
             return <PolygonGraph graphState={state} dispatch={dispatch} />;
         case "point":
             return <PointGraph graphState={state} dispatch={dispatch} />;
+        case "circle":
+            throw new Error("the circle graph type is not yet implemented");
```
benchristel, line 44: I added the circle type as a kind of test case to make sure the refactored code will be flexible in the ways we need it to be. But since circle graphs are not actually implemented yet, I should probably take this out (and remove the `circle` graph state type) before landing this PR.

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -8,60 +8,53 @@ import {
...
-                          }),
-                      });
+    switch (state.type) {
+        case "segment":
+        case "linear":
+        case "linear-system":
+        case "ray": {
+            if (action.itemIndex == null) {
+                throw new Error(
+                    "MoveControlPoint.itemIndex cannot be null when moving a point on a line",
```
benchristel, line 38: The need for this exception is a sign that the types still aren't quite right. @mark-fitzgerald maybe we can pair on addressing this.

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -164,6 +204,27 @@ export function interactiveGraphReducer<
...
+        default:
+            return state;
+    }
+}
+
+// Generic type makes returned state match input state
+export function interactiveGraphReducer(
+    state: InteractiveGraphState,
+    action: InteractiveGraphAction,
+): InteractiveGraphState {
```
benchristel, line 217: Since this `interactiveGraphReducer` is the main export of this file, IMO it should be at the top of the file where people can easily find it.

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -81,12 +74,43 @@ export function interactiveGraphReducer<
...
+                    point: bound({
+                        snapStep,
+                        range,
+                        point: action.destination,
+                    }),
+                }),
+            });
+            return {
+                ...state,
+                hasBeenInteractedWith: true,
```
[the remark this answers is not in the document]
benchristel: It looks like some actions are not setting `hasBeenInteractedWith` in some early-return scenarios, so I'm opting not to do this refactoring.

mark-fitzgerald (approved):

PR #1139, opened 2024-04-01
Mafs callback on change (interactive graph editor change bug)

handeyeco:
## Summary:
I don't know if this is the preferred way to do things, but it works

https://github.com/Khan/perseus/assets/16308368/ade4f43a-798a-4b75-bd63-23b05d43b944

Issue: LEMS-1882

## Test plan:
- In Perseus SB, go to: Interactive Graph Editor / With Locked point
- Move a point
- Note the "Correct answer" field updates as you modify the graph

### DIFF
--- a/packages/perseus/src/types.ts
+++ b/packages/perseus/src/types.ts
@@ -7,6 +7,7 @@ import type {
     PerseusWidgetsMap,
 } from "./perseus-types";
 import type {SizeClass} from "./util/sizing-utils";
+import type {InteractiveGraphState} from "./widgets/interactive-graphs/types";
 import type {KeypadAPI} from "@khanacademy/math-input";
 import type {AnalyticsEventHandlerFn} from "@khanacademy/perseus-core";
 import type {LinterContextProps} from "@khanacademy/perseus-linter";
@@ -94,6 +95,8 @@ export type ChangeHandler = (
         selected?: number;
         // perseus-all-package/widgets/grapher.jsx
         plot?: any;
+        // Interactive Graph callback (see legacy: interactive-graph.tsx)
+        graph?: InteractiveGraphState;
     },
     callback?: () => unknown | null | undefined,
     silent?: boolean,
--- a/packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
@@ -22,28 +22,6 @@ function getBaseMafsGraphProps(): MafsWrapperProps {
         containerSizeClass: "small",
         onChange: () => {},
     };
-
-    // const widgetBaseProps: InteractiveGraphProps = {
-    //     ...mafsGraphBaseProps,
-    //     alignment: null,
-    //     apiOptions: {
-    //         ...ApiOptions.defaults,
-    //     },
-    //     containerSizeClass: "small",
-    //     findWidgets: () => [],
-    //     isLastUsedWidget: false,
-    //     onBlur: () => {},
-    //     onChange: () => {},
-    //     onFocus: () => {},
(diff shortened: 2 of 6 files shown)


packages/perseus/src/types.ts
```diff
@@ -94,6 +95,8 @@ export type ChangeHandler = (
         selected?: number;
         // perseus-all-package/widgets/grapher.jsx
         plot?: any;
+        // Interactive Graph callback (see legacy: interactive-graph.tsx)
+        graph?: InteractiveGraphState;
```
jeremywiebe, line 99: Eventually, it would be really  nice if each widget that had an `onChange` callback would type it using a generic. For now, this is fine.

jeremywiebe (approved):
Thanks for writing a test for this!

packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+        reviewModeRubric: {
+            ...mafsGraphBaseProps,
+        },
+        static: true,
+        trackInteraction: () => {},
+        widgetId: "mafs-graph",
+        linterContext: {} as any,
+    };
+
+    return widgetBaseProps;
```
benchristel, line 54: I had no idea `MafsGraph` required all these props! I think very few of them are actually used.

I think I would prefer it if, instead of having to create these do-nothing props in tests, we trimmed down the props type of MafsGraph to include only what it needs. Currently, the props type is derived from the InteractiveGraph props:

```ts
React.PropsWithChildren<InteractiveGraphProps> & {box: [number, number]}
```

packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+    beforeEach(() => {
+        userEvent = userEventLib.setup({
+            advanceTimers: jest.advanceTimersByTime,
+        });
+    });
+
+    it("renders", () => {
+        render(<MafsGraph {...getBaseMafsGraphProps()} box={[400, 400]} />);
+
+        expect(screen.getByTestId("mafs-graph__wrapper")).toBeInTheDocument();
```
benchristel, line 68: We don't use this test ID anywhere else, so it feels like this test isn't adding much value. If this test fails, what will that tell us? Likely just that the test ID has changed or gone missing.

I'd rather have this first test verify the assumptions made by the second test. E.g. it could check for the presence of a `circle.movable-point-hitbox`.

packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
handeyeco: I'm a little confused with this one. `widgetBaseProps` is based on `WidgetProps` which is used by all widgets. `mafsGraphBaseProps` is based on `PerseusInteractiveGraphWidgetOptions` which is used by the legacy Interaction widget. I'd be concerned about modifying either of those due to impact on non-Mafs widgets.

I could nest `MafsGraph` - the parent could take `PerseusInteractiveGraphWidgetOptions` and pick out the pieces that the child `MafsGraph` needs. That way the parent could be easily swapped with the original Interactive Graph widget and the child could focus only on what it requires to function?

I'm tempted to suggest that be a follow-up though.

benchristel: I think we can change the types and tests without having to change any production code. Here's a commit showing what I mean. You should be able to cherry-pick it into your branch, if you agree it's an improvement: https://github.com/Khan/perseus/commit/cfee1eaa3e4a9812a92696d0cff2e52220fa5dcc

PR #1141, opened 2024-04-01
Update store for `range` and `snapStep` when props change

handeyeco:
## Summary:
Update reducer state for `range` and `snapStep` when props change (for editing experience).

https://github.com/Khan/perseus/assets/16308368/419f4933-d888-4160-9bfa-3b0a72f1eb0d

Issue: [LEMS-1824](https://khanacademy.atlassian.net/browse/LEMS-1824)

## Test plan:
- Go to the "With Locked Points" story
- Move the points around to see the step they move by
- Change `snapStep` to 5, 5
- Note the snap now works on 5, 5

[LEMS-1824]: https://khanacademy.atlassian.net/browse/LEMS-1824?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


benchristel (commented):
The code looks good. But I wonder: should we also re-snap and bound existing points when the snap step or range changes? Otherwise, we could end up with points that are not snapped or are outside the graph.

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -197,6 +206,40 @@ function doMovePoint(
...
+function doChangeSnapStep(
+    state: InteractiveGraphState,
+    action: ChangeSnapStep,
+): InteractiveGraphState {
+    if (
+        // Deep equality check since these are arrays
+        _.isEqual(state.snapStep, action.snapStep)
+    ) {
+        return state;
+    }
```
benchristel, line 218: I like this — it's an optimization to prevent unnecessary re-renders, right? It makes me wonder if we should do an `isEqual` check on every action, perhaps right before returning from the root reducer function:

```ts
if (_.isEqual(newState, oldState)) {
  return oldState;
}

return newState;
```

packages/perseus/src/perseus-types.ts
```diff
@@ -483,6 +483,9 @@ export type PerseusGradedGroupSetWidgetOptions = {
     gradedGroups: ReadonlyArray<PerseusGradedGroupWidgetOptions>;
 };
 
+// 2D range: xMin, xMax, yMin, yMax
+export type GraphRange = [[number, number], [number, number]];
```
benchristel, line 487: I recently learned that tuple members can have names in TypeScript! So you can do this:

```ts
export type GraphRange = [
  x: [min: number, max: number],
  y: [min: number, max: number],
];
```

This seems a bit more self-documenting.

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -73,15 +78,25 @@ export const MafsGraph = React.forwardRef<
...
 
+    const {snapStep, range} = props;
+
+    useEffect(() => {
+        dispatch(changeSnapStep(snapStep));
+    }, [snapStep]);
+
+    useEffect(() => {
+        dispatch(changeRange(range));
+    }, [range]);
```
benchristel, line 98: I wonder if, to prevent unnecessary re-renders, the individual numbers in `snapStep` and `range` could be listed as deps here:

```ts
    useEffect(() => {
        dispatch(changeSnapStep(snapStep));
    }, snapStep);

    useEffect(() => {
        dispatch(changeRange(range));
    }, range.flat());
```

We could also maybe `useLayoutEffect` so the state update happens before React renders anything to the DOM.

handeyeco:
> But I wonder: should we also re-snap and bound existing points when the snap step or range changes? Otherwise, we could end up with points that are not snapped or are outside the graph.

@benchristel idk, maybe I'm missing something, but this feels like a relatively big ask. I think I would advocate for breaking it out into a separate ticket to be prioritized separately. Personally I think it'd be useful, but not sure the added complexity would be worth the benefit. I think I would file a ticket and wait for someone to ask for it to be addressed.

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
handeyeco: I did what you asked but in a more verbose way. The linter didn't seem happy with your suggestion and I think being a little more verbose makes the intent a little more clear. It's a good suggestion though, solid improvement.

benchristel (approved):
LGTM!

PR #1142, opened 2024-04-02
Split grid vs tick step

handeyeco:
## Summary:
Separate out Grid Step from Tick Step in Mafs charts

https://github.com/Khan/perseus/assets/16308368/51bdb324-a763-4859-9d68-906a394b7f32

Issue: [LEMS-1883](https://khanacademy.atlassian.net/browse/LEMS-1883)

## Test plan:
- Go to "With Locked Points"
- Change grid step and note it has minimal effect on tick step
- Change tick step and note it doesn't affect grid step

[LEMS-1883]: https://khanacademy.atlassian.net/browse/LEMS-1883?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/grid.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/grid.tsx
@@ -46,7 +46,7 @@ export const lineLabelText = (
  * @param {GridProps} props
  * @param {number} axisIndex which axis we're getting options for
  */
-export const axisOptions = (
+const axisOptions = (
     props: Omit<GridProps, "containerSizeClass">,
     axisIndex: number,
 ) => {
--- a/packages/perseus/src/widgets/interactive-graphs/grid.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/grid.tsx
@@ -7,7 +7,7 @@ import type {SizeClass} from "../../util/sizing-utils";
 import type {vec} from "mafs";
 
 interface GridProps {
-    step: vec.Vector2;
+    tickStep: vec.Vector2;
     gridStep: vec.Vector2;
     range: [[number, number], [number, number]];
     containerSizeClass: SizeClass;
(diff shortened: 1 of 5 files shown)


nishasy:
Looking at the [storybook preview from the checks](https://650db21c3f5d1b2f13c02952-gloegfmzdi.chromatic.com/?path=/story/perseus-editor-widgets-interactive-graph-editor--with-locked-points): I think there might be a bug, or at least a mismatch with the non-mafs behavior. I think it won't let me change the tick step to a value smaller than the grid step. Can you repro?

benchristel (approved):
LGTM! I noticed what I think is a bug in the logic for deciding whether we render a label, but since this PR doesn't introduce the bug I think we could fix it later if you prefer. Let me know if you decide not to fix it, and I'll create a ticket.

packages/perseus/src/widgets/interactive-graphs/grid.tsx
```diff
@@ -4,30 +4,63 @@ import * as React from "react";
 import AxisArrows from "./axis-arrows";
 
 import type {SizeClass} from "../../util/sizing-utils";
+import type {vec} from "mafs";
 
 interface GridProps {
-    step: [number, number];
-    gridStep: [number, number];
+    step: vec.Vector2;
```
nicolecomputer, line 10: Should this maybe be called tickStep? (not a blocker for this PR)

handeyeco:
I think this PR is still progress, so I'm going to land it. Filed some tickets though:
- https://khanacademy.atlassian.net/browse/LEMS-1890 per @nishasy's comment
- https://khanacademy.atlassian.net/browse/LEMS-1891 per @benchristel's comment

PR #1143, opened 2024-04-02
Graph tick marks

nicolecomputer:
## Summary:

@nicolecomputer + @SonicScrewdriver co-authored this.

Adds graph tick marks.

image

Issue: https://khanacademy.atlassian.net/browse/LEMS-1877

## Test plan:
- Check out the graphs

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/axis-ticks.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/axis-ticks.tsx
@@ -12,19 +12,16 @@ const tickStyle: React.CSSProperties = {
 };
 
 const YGridTick = ({y}: {y: number}) => {
-    const startingPoint: vec.Vector2 = [0, y];
-    let [startPtPx] = useTransform(startingPoint);
-
-    const endPtPx = [startPtPx[0] + tickSize / 2, startPtPx[1]];
-    startPtPx = [startPtPx[0] - tickSize / 2, startPtPx[1]];
+    const pointOnAxis: vec.Vector2 = [0, y];
+    const [[xPosition, yPosition]] = useTransform(pointOnAxis);
 
     return (
-        <g className="y-axis-ticks" tabIndex={0} style={{} as any}>
+        <g className="y-axis-ticks">
             <line
-                x1={startPtPx[0]}
-                y1={startPtPx[1]}
-                x2={endPtPx[0]}
-                y2={endPtPx[1]}
+                x1={xPosition - tickSize / 2}
+                y1={yPosition}
+                x2={xPosition + tickSize / 2}
+                y2={yPosition}
                 style={tickStyle}
             />
         </g>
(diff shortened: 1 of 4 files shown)


packages/perseus/src/widgets/interactive-graphs/axis-ticks.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+const XGridTick = ({x}: {x: number}) => {
+    const startingPoint: vec.Vector2 = [x, 0];
+    let [startPtPx] = useTransform(startingPoint);
+
+    const endPtPx = [startPtPx[0], startPtPx[1] + tickSize / 2];
+    startPtPx = [startPtPx[0], startPtPx[1] - tickSize / 2];
+
+    return (
+        <g className="x-axis-ticks" tabIndex={0} style={{} as any}>
```
benchristel, line 42: Why pass an empty `style` object here? (and in `YGridTick` above?)

And do we really want these groups to be tabbable?

packages/perseus/src/widgets/interactive-graphs/axis-ticks.tsx
```diff
@@ -0,0 +1,95 @@
...
+                x2={endPtPx[0]}
+                y2={endPtPx[1]}
+                style={tickStyle}
+            />
+        </g>
+    );
+};
+
+const XGridTick = ({x}: {x: number}) => {
+    const startingPoint: vec.Vector2 = [x, 0];
```
benchristel, line 35: This `startingPoint` is really the middle of the tick mark, right? I wonder if something like this, without the reassignment, would be clearer:

```ts
const pointOnAxis: vec.Vector2 = [x, 0];
const [[xPx, yPx]] = useTransform(pointOnAxis);

return (
  <g className="x-axis-ticks">
    <line
      x1={xPx}
      y1={yPx - tickSize / 2}
      x2={xPx}
      y2={yPx + tickSize / 2}
      style={tickStyle}
    />
  </g>
);
```

packages/perseus/src/widgets/interactive-graphs/axis-ticks.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+const YGridTick = ({y}: {y: number}) => {
+    const startingPoint: vec.Vector2 = [0, y];
+    let [startPtPx] = useTransform(startingPoint);
+
+    const endPtPx = [startPtPx[0] + tickSize / 2, startPtPx[1]];
+    startPtPx = [startPtPx[0] - tickSize / 2, startPtPx[1]];
+
+    return (
+        <g className="y-axis-ticks" tabIndex={0} style={{} as any}>
```
mark-fitzgerald, line 22: What does the `tabIndex` do for us on this element? I'm wondering what the use case is for it.

benchristel (commented):
Very cool! I left ideas for some simplifications inline — mostly pure refactoring, but one that will affect behavior (removing `tabIndex`). You can take or leave the refactorings, but I think we should decide what to do about `tabIndex`.

packages/perseus/src/widgets/interactive-graphs/axis-ticks.tsx
SonicScrewdriver: I believe this snuck in here as we copied the "g" element from elsewhere! I'll remove them :D

packages/perseus/src/widgets/interactive-graphs/axis-ticks.tsx
SonicScrewdriver: I believe this snuck in here as we copied the "g" element from elsewhere! Currently getting rid of it :)

benchristel (approved):
Looks great! Thanks!

PR #1145, opened 2024-04-02
Ensure graphs in the dev gallery have stable React keys

benchristel:
## Summary:
This fixes a bug. Repro steps:

1. Run `yarn dev` and visit `http://localhost:5173`
2. Ensure all Mafs flags are turned OFF and the searchbar is clear.
3. Paste the text `poly` into the searchbar.

Expected: All graphs render without errors.

Actual: One graph crashes to the Renderer arrow boundary (red circle
with exclamation).

The bug happened because a segment graph component got reused for a
polygon graph. The component gets into a weird state when this happens
and crashes. Ensuring each graph has a stable key fixes the bug because
now each component renders the same graph for its whole lifetime.

Issue: none

Test plan:

Follow the repro steps above and confirm that the bug does not repro.

### DIFF
--- a/dev/gallery.tsx
+++ b/dev/gallery.tsx
@@ -20,7 +20,7 @@ import type {APIOptions, PerseusRenderer} from "../packages/perseus/src";
 
 import "../packages/perseus/src/styles/perseus-renderer.less";
 
-const questions = [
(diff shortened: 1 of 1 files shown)


handeyeco (approved):
LGTM!

PR #1146, opened 2024-04-02
Fix Safari clipping the focus circle on the movable point

jeremywiebe:
## Summary:

We noticed that the focus ring on movable points was being clipped on Safari 17.2.1. This wasn't happening on Safari 16 (older) nor Safari Technology Preview (newer).

We were able to re-work how the focus ring was drawn and arrive at something that looks exactly the same but works on all browsers. 

Design spec: https://www.figma.com/file/JroPokeCLBv2VRIw3GPvY5/LC---Interactive-graph-widget?type=design&node-id=855-9644&mode=design&t=f6QWv1t1RXJtI6ky-4

Issue: LEMS-1889

## Test plan:

`yarn start`
Navigate to: http://localhost:6006/?path=/story/perseus-widgets-interactive-graph--segment-with-mafs-and-locked-points
Press `{tab}` on the keyboard until one of the movable points is focused.
Note that the focus ring (outside of the light-blue "halo") is a full circle and not clipped.

## Screenshot (fixed)

image

Co-authored-by: Ben Christel <benchristel@khanacademy.org>
Co-authored-by: Sarah Third <sarahthird@khanacademy.org>

### DIFF
--- a/.changeset/forty-tomatoes-lick.md
+++ b/.changeset/forty-tomatoes-lick.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Fix focus ring on movable points in interactive-graph widget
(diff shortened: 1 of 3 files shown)


benchristel (approved):
LGTM!

PR #1149, opened 2024-04-03
Helper: find questions in `data` that match a predicate

handeyeco:
## Summary:
I thought maybe this would be helpful. I needed a way to find questions where grid step was greater than our tick step.

The idea is this can be modified as needed to help other people find the data they need.

I know `jq` exists, but I'd rather stick to my JS roots.

### DIFF
--- a/data/find-questions.js
+++ b/data/find-questions.js
@@ -1,0 +1,81 @@
+/**
+ * To use this:
+ * - Modify `predicateCallback` to look for questions that
+ *   match your requirements
+ * - run: node find-questions.js
+ */
+
+const Path = require("path");
+const fs = require("fs");
+
+// ==========================
+// MODIFY THIS WHEN SEARCHING
+// ==========================
+
+/**
+ * predicateCallback used to check if the given question matches
+ * the rules you're looking for
+ *
+ * @param {PerseusRenderer} q the question to check
+ * @returns {boolean} whether the question is a match
+ */
+function predicateCallback(q) {
+    // Look through each widget in the question
+    for (let widget of Object.values(q.widgets)) {
+        // Skip everything that's not an interactive-graph
+        if (widget.type !== "interactive-graph") continue;
+
+        // Make sure the interactive-graph has the data we're comparing
+        if (!widget?.options?.gridStep || !widget?.options.step) continue;
+
+        // Do that actual check
+        const [xGridStep, yGridStep] = widget.options.gridStep;
+        const [xTickStep, yTickStep] = widget.options.step;
+        if (xGridStep > xTickStep || yGridStep > yTickStep) {
+            return true;
+        }
+    }
+
+    return false;
+}
+
+// ================================
+// DON'T MODIFY THIS WHEN SEARCHING
+// ================================
+
+// Find all `.json` files in a directory (recursive)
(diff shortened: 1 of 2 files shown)


data/find-questions.js
```diff
@@ -0,0 +1,88 @@
...
+        const data = fs.readFileSync(fileName, "utf8");
+        const json = JSON.parse(data);
+        if (predicateCallback(json)) {
+            output.push(data);
+        }
+    }
+}
+
+function main() {
+    findJsonFiles("./questions");
```
benchristel, line 79: Why not have `findJsonFiles` return the files it found? I think that's clearer and less mistake-prone than using a global.

data/find-questions.js
```diff
@@ -0,0 +1,88 @@
...
+    checkFiles();
```
benchristel, line 80: Same thing here, `checkFiles` could return the `output`.

benchristel (approved):
This is a great idea. I left a couple small suggestions inline which I would prefer to be addressed, but I won't insist on it. I don't think this needs another round of review in any case!

data/find-questions.js
```diff
@@ -0,0 +1,88 @@
+/**
+ * To use this:
+ * - Modify `predicateCallback` to look for questions that
+ *   match your requirements
+ * - run: node find-questions.js
```
jeremywiebe, line 5: One idea.  If you want to use TypeScript and ES module imports in this file, you could:
  * Rename it to `.ts`
  * Add `#!/usr/bin/env -S node -r @swc-node/register` as the first line in this file
  * Run `chmod +x dev/find-questions.js`

Then we'd be able to use types here and run it by simply doing `./dev/find-questions.js`.  You can see an example of this in `utils/extract-strings.ts`

Totally optional though!

jeremywiebe (approved):
Love the idea. Had one thought about using TS, but it's not required.

data/find-questions.js
handeyeco: It was just laziness. I copy/pasted some recursive code from SO and I didn't want to think about it. Fixed.

data/find-questions.js
handeyeco: For `findJsonFiles` I copy/pasted and `checkFiles` I just matched the other function's style. Fixed.

data/find-questions.js
```diff
@@ -0,0 +1,88 @@
...
+        // Do that actual check
+        const [xGridStep, yGridStep] = widget.options.gridStep;
+        const [xTickStep, yTickStep] = widget.options.step;
+        if (xGridStep > xTickStep || yGridStep > yTickStep) {
+            return true;
+        }
+    }
+
+    return false;
+}
```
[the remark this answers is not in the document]
handeyeco: I like this idea! It got me scope creeping though...what if we add flags too so that people don't have to modify the file at all if the predicate they want exists! I think I'll land it and we can iterate.

data/find-questions.js
handeyeco: Done, thanks

benchristel: TIL: `#!/usr/bin/env -S node -r @swc-node/register`

That's a neat trick!

PR #1151, opened 2024-04-03
math-input: upgrade mathquill to v1.0.1

jeremywiebe:
## Summary:

We've had a very long-standing issue where the Mathquill package referenced its fonts using an absolute path (eg. `/fonts/Symbola.woff2`). This has been difficult (let's be real, unsolved) for our build systems to deal with for most of the time we've been using Mathquill. 

Today we fix it! Mathquill v1.0.1 now references its fonts using relative paths, which are then trivially resolved by our build tools (be it Webpack, Rollup, or Vite).

Issue: LEMS-375

## Test plan:

Run `yarn start`
Navigate to http://localhost:6006/?path=/story/math-input-full-mobile-mathinput--wrapped
Open the browser dev tools and switch to the Network tab. Filter on "Symbola" and note that at least one of the "Symbola" fonts were successfully loaded.

|   | Before | After |
| - | ------ | ----- |
| UI |  image | image |
| Actual font | image | image |
| Network Tab | before | after |

### DIFF
--- a/.changeset/cold-mugs-pay.md
+++ b/.changeset/cold-mugs-pay.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/math-input": patch
+---
+
+Upgrade mathquill dependency to v1.0.1 to fix Symbola font references
--- a/packages/math-input/package.json
+++ b/packages/math-input/package.json
@@ -24,7 +24,7 @@
     "scripts": {},
     "dependencies": {
         "@khanacademy/perseus-core": "1.4.2",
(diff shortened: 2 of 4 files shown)


nedredmond (approved):
LGTM, except for the `v` missing from the semver tag, but that norm doesn't really matter at all

jeremywiebe:
Abandoning!  See [LEMS-375](https://khanacademy.atlassian.net/browse/LEMS-375) for details!

PR #1152, opened 2024-04-03
Increase the size of MovablePoint hitboxes to 48px

benchristel:
## Summary:
This is a holdover from the legacy interactive graph. 44px is WCAG AAA,
but there's no reason not to go up to 48px since that's the existing
behavior.

Issue: https://khanacademy.atlassian.net/browse/LEMS-1895

Test plan:

View a segment graph in the dev UI. Inspect one of the point hitboxes.
It should be 48 by 48 pixels.

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.tsx
@@ -3,7 +3,6 @@ import {useMovable} from "mafs";
 import * as React from "react";
 import {useRef} from "react";
 
-import {TARGET_SIZE} from "../../utils";
 import {useTransform} from "../use-transform";
 
 import type {vec} from "mafs";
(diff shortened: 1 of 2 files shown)


packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.tsx
```diff
@@ -14,6 +13,10 @@ type Props = {
     color?: string;
 };
 
+// The hitbox size of 48px by 48px is preserved from the legacy interactive
+// graph.
```
jeremywiebe, line 17: Not only legacy. I remember Sarah B mentioning this is also the Android minimum tap target size. 

https://support.google.com/accessibility/android/answer/7101858

But the fun begins, the current [Material Design](https://m3.material.io/foundations/accessible-design/accessibility-basics#20bcdd28-186d-4884-b592-3fb069e10a73) from Google mentions 44 (which matches [WCAG](https://www.w3.org/WAI/WCAG21/Understanding/target-size.html#:~:text=If%20the%20target%20is%20the,44%20by%2044%20CSS%20pixels.&text=A%20footnote%20or%20an%20icon,from%20the%20minimum%20target%20size.)).

🤷‍♂️

jeremywiebe (approved):

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.tsx
jeremywiebe: Actually, I read wrong.  The Material Design spec makes a distinction between "Touch and pointer" (48 dp) and strictly "pointer" targets (44 dp).

Ignore me. :)

PR #1153, opened 2024-04-04
Remove wonder-blocks-i18n from math-input.

jeresig:
## Summary:
Remove usage of wonder-blocks-i18n from math-input. Is replaced by using a context in which the strings and locale are specified. This will need to be used in conjunction with other changes to the `perseus` and `perseus-editor` packages. We shouldn't release this change until we roll out all of these changes to all 3 of these packages at the same time.

Proposal: https://khanacademy.atlassian.net/wiki/spaces/FRONTEND/pages/2775777284/LinguiJS+Perseus+String+Handling+Proposal

Issue: FEI-5550

## Test plan:
All tests, eslint, and TS should pass.


packages/math-input/src/utils.ts
```diff
@@ -46,9 +65,10 @@ const TIMES_ONLY = ["fr", "tr", "pt-pt"];
  * @param {boolean} convertDotToTimes - the setting set by content creators
  * @returns {boolean} - true to convert to × (TIMES), false to use · (CDOT)
  */
-export function convertDotToTimesByLocale(convertDotToTimes: boolean): boolean {
-    const locale = getLocale();
-
+export function convertDotToTimesByLocale(
+    convertDotToTimes: boolean,
+    locale: string,
+): boolean {
```
jeremywiebe, line 71: I have a loose preference for `locale` showing up first because it feels most important for any function that takes a `locale` (and gives us consistency for where it shows up in other params).

testing/render-keypad-with-cypress.tsx
```diff
@@ -3,9 +3,12 @@ import * as React from "react";
 
 import Keypad from "../packages/math-input/src/components/keypad";
 
+import {strings} from "./mock-strings";
+
 const renderSingleKeypad = () =>
     mount(
         <Keypad
+            strings={strings}
```
jeremywiebe, line 11: Do we need a `locale` here?

jeremywiebe (commented):
Overall I'm good with this approach. I did comment about possibly using a React context to provide the `locale` across the libraries, but that brings along its own complications. What do you think?

packages/math-input/src/utils.ts
jeresig: Good call - I can switch this!

packages/math-input/src/index.ts
```diff
@@ -31,6 +31,16 @@ export {MobileKeypad} from "./components/keypad";
 // Unwrapped v2 keypad for desktop
 export {default as DesktopKeypad} from "./components/keypad";
 
+// For strings used in MathQuill and the keypad
+export {strings} from "./strings";
```
jeresig, line 35: @jeremywiebe I'm not happy about this, I would much prefer that we build and export this separately, so that it's not part of the bundle. For example I could import it like so:

```
import strings from "@khanacademy/math-input/strings";
```

And not from the main export itself. Do you have a sense of how to best configure something like that? If not, I can experiment with making that happen in rollup.

jeresig: Just to follow up on this, I got this working! There are now multiple exports.

.changeset/five-readers-boil.md
```diff
@@ -0,0 +1,7 @@
+---
```
SonicScrewdriver, line 1: Do we want two changesets?

packages/math-input/src/components/input/mathquill-instance.ts
```diff
@@ -102,17 +101,31 @@ const createBaseConfig = (): MathFieldConfig => ({
...
 export function createMathField(
     container: HTMLDivElement | HTMLSpanElement,
+    strings: MathInputStrings,
     configCallback?: (baseConfig: MathFieldConfig) => MathFieldConfig,
 ): MathFieldInterface {
     const baseConfig = createBaseConfig();
     const config = configCallback ? configCallback(baseConfig) : baseConfig;
 
     const mathField = mathQuillInstance
         .MathField(container, config)
```
SonicScrewdriver, line 111: This seems much cleaner than the old strings file!

packages/math-input/src/types.ts
```diff
@@ -68,3 +68,69 @@ export type KeypadContextType = {
...
+    "nothing above": string;
+    labelValue: ({label, value}: {label: string; value: string}) => string;
+    plus: string;
+    minus: string;
+    negative: string;
+    multiply: string;
+    divide: string;
+    decimal: string;
+    percent: string;
+    multiply2: string;
```
SonicScrewdriver, line 99: I feel it might be a little strange having a string called multiply2. I wonder if it makes sense to name them strings.cdot and strings.times, and just have the resulting string be "Multiply" for both? It might help make it a little easier to understand in key-config. 

This is a total nit though, and not something I consider an issue.

packages/math-input/src/utils.ts
```diff
@@ -1,20 +1,39 @@
...
-export const decimalSeparator: string =
-    getDecimalSeparator() === ","
-        ? DecimalSeparator.COMMA
-        : DecimalSeparator.PERIOD;
+/**
+ *  Get the character used for separating decimals.
+ */
+export const getDecimalSeparator = (locale: string): string => {
+    let separator: string = DecimalSeparator.PERIOD;
+
```
SonicScrewdriver, line 11: Oh interesting, I didn't know about this issue! Thanks for adding a fix.

.changeset/five-readers-boil.md
jeresig: Oh whoops - I don't think we do. I'll remove one!

packages/math-input/src/types.ts
jeresig: Ok, yeah - I can change that!

config/build/rollup.config.js
```diff
@@ -276,28 +282,61 @@ const getPackageInfo = (commandLineArgs, pkgName) => {
...
-            format: "esm",
-            platform: "browser",
-            file: pkgJson.module,
-            // We care about the file size of this one.
-            plugins: [filesize()],
-        });
+
+    if (pkgJson.exports) {
+        for (const exportConfig of Object.values(pkgJson.exports)) {
+            if (exportConfig.require && formats.has("cjs")) {
```
jeremywiebe, line 288: Unrelated to this PR, but we are _extremely_ close to not needing `cjs` builds anymore (in fact, I think we can drop them any time we have bandwidth). :)

packages/math-input/package.json
```diff
@@ -18,6 +18,18 @@
...
+            "require": "./dist/index.js",
+            "source": "./src/index.ts"
+        },
+        "./strings": {
+            "import": "./dist/es/strings.js",
+            "require": "./dist/strings.js",
+            "source": "./src/strings.ts"
+        }
+    },
     "files": [
```
jeremywiebe, line 33: Really neat!  Thanks for figuring this out (especially with the Rollup build changes!)

jeremywiebe (approved):
Really happy where this ended up.  Thanks John!

PR #1154, opened 2024-04-04
Movable point bug: use snap in constrain callback

handeyeco:
## Summary:
Use `snap` in `constrain` callback so Mafs can properly position points when navigating with a keyboard.

Issue: [LEMS-1885](https://khanacademy.atlassian.net/browse/LEMS-1885)

## Test plan:
- In Flipbook
- Paste the below content
- Click a movable point
- Use your arrow keys to try an move it
- Note they move

```
{"content":"This one was always working [[☃ interactive-graph 1]]","images":{},"widgets":{"input-number 1":{"alignment":"default","graded":true,"options":{"answerType":"number","inexact":false,"maxError":0.1,"simplify":"required","size":"normal","value":6},"static":false,"type":"input-number","version":{"major":0,"minor":0}},"interactive-graph 1":{"alignment":"default","graded":true,"options":{"backgroundImage":{"bottom":0,"height":0,"left":0,"scale":1,"url":"","width":0},"correct":{"coords":[[3,2],[9,2],[3,6],[9,6]],"numPoints":4,"type":"point"},"graph":{"numPoints":4,"type":"point"},"gridStep":[1,1],"labels":["x","y"],"markings":"graph","range":[[-10,10],[-10,10]],"rulerLabel":"","rulerTicks":10,"showProtractor":false,"showRuler":false,"snapStep":[0.5,0.5],"step":[1,1]},"static":false,"type":"interactive-graph","version":{"major":0,"minor":0}}}}

{"content":"This one previously did not work [[☃ interactive-graph 1]]","images":{},"widgets":{"input-number 1":{"alignment":"default","graded":true,"options":{"answerType":"number","inexact":false,"maxError":0.1,"simplify":"required","size":"normal","value":6},"static":false,"type":"input-number","version":{"major":0,"minor":0}},"interactive-graph 1":{"alignment":"default","graded":true,"options":{"backgroundImage":{"bottom":0,"height":0,"left":0,"scale":1,"url":"","width":0},"correct":{"coords":[[3,2],[9,2],[3,6],[9,6]],"numPoints":4,"type":"point"},"graph":{"numPoints":4,"type":"point"},"gridStep":[1,1],"labels":["x","y"],"markings":"graph","range":[[-10,10],[-10,10]],"rulerLabel":"","rulerTicks":10,"showProtractor":false,"showRuler":false,"snapStep":[1,1],"step":[1,1]},"static":false,"type":"interactive-graph","version":{"major":0,"minor":0}}}}

{"content":"This one previously did not work [[☃ interactive-graph 1]]","images":{},"widgets":{"input-number 1":{"alignment":"default","graded":true,"options":{"answerType":"number","inexact":false,"maxError":0.1,"simplify":"required","size":"normal","value":6},"static":false,"type":"input-number","version":{"major":0,"minor":0}},"interactive-graph 1":{"alignment":"default","graded":true,"options":{"backgroundImage":{"bottom":0,"height":0,"left":0,"scale":1,"url":"","width":0},"correct":{"coords":[[3,2],[9,2],[3,6],[9,6]],"numPoints":4,"type":"point"},"graph":{"numPoints":4,"type":"point"},"gridStep":[1,1],"labels":["x","y"],"markings":"graph","range":[[-1,10],[-1,10]],"rulerLabel":"","rulerTicks":10,"showProtractor":false,"showRuler":false,"snapStep":[0.5,0.5],"step":[1,1]},"static":false,"type":"interactive-graph","version":{"major":0,"minor":0}}}}
```

[LEMS-1885]: https://khanacademy.atlassian.net/browse/LEMS-1885?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
@@ -3,6 +3,8 @@ import {UnreachableCaseError} from "@khanacademy/wonder-stuff-core";
 import {vec} from "mafs";
 import _ from "underscore";
 
+import {snap} from "../utils";
+
 import {
     type InteractiveGraphAction,
     MOVE_ALL,
@@ -61,14 +63,14 @@ function doMoveControlPoint(
                     setAtIndex({
                         array: tuple,
                         index: action.pointIndex,
-                        newValue: snap({
+                        newValue: snap(
                             snapStep,
-                            point: bound({
+                            bound({
                                 snapStep,
                                 range,
                                 point: action.destination,
                             }),
-                        }),
+                        ),
                     }),
             });
 
@@ -117,14 +119,8 @@ function doMoveLine(
             });
 
             const newLine: PairOfPoints = [
-                snap({
-                    snapStep,
-                    point: vec.add(currentLine[0], change),
-                }),
-                snap({
-                    snapStep,
-                    point: vec.add(currentLine[1], change),
-                }),
+                snap(snapStep, vec.add(currentLine[0], change)),
+                snap(snapStep, vec.add(currentLine[1], change)),
             ];
 
             const newCoords = setAtIndex({
@@ -160,7 +156,7 @@ function doMoveAll(
             });
 
             const newCoords = state.coords.map((point: vec.Vector2) =>
-                snap({snapStep, point: vec.add(point, change)}),
+                snap(snapStep, vec.add(point, change)),
             );
 
             return {
@@ -188,14 +184,14 @@ function doMovePoint(
                 coords: setAtIndex({
                     array: state.coords,
                     index: action.index,
-                    newValue: snap({
-                        snapStep: state.snapStep,
-                        point: bound({
+                    newValue: snap(
+                        state.snapStep,
+                        bound({
                             snapStep: state.snapStep,
                             range: state.range,
                             point: action.destination,
                         }),
-                    }),
+                    ),
                 }),
             };
         }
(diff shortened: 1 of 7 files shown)


packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -117,14 +119,8 @@ function doMoveLine(
...
-                snap({
-                    snapStep,
-                    point: vec.add(currentLine[0], change),
-                }),
-                snap({
-                    snapStep,
-                    point: vec.add(currentLine[1], change),
-                }),
+                snap(snapStep, vec.add(currentLine[0], change)),
+                snap(snapStep, vec.add(currentLine[1], change)),
```
benchristel, line 123: Suggestion: I'm not sure I want to lose the named arguments on `snap`. Since the point and snap step params have the same structure (a pair of numbers) it's too easy to mix them up.

Maybe a compromise like this would work?

```ts
snap(point, {snapStep: [0.5, 0.5]})
```

This makes it pretty clear which argument is which, and it has the additional advantage that we can pass the whole graph state as the second argument:

```ts
snap(point, state)
```

benchristel (approved):
Very cool! Thanks for digging into this.

I left one suggestion inline about the parameters of `snap`. TL;DR I think the named parameters had some advantages, but a hybrid solution might give us the best of both worlds.

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
handeyeco: I'll revert the change. I just really dislike the "all arguments as an arg object" pattern; I think it's needless complexity and we're adding that complexity to solve a problem that IDEs fix:

Screenshot 2024-04-04 at 2 02 27 PM

I also think this makes the TS signature more useful.

Screenshot 2024-04-04 at 2 05 53 PM

Screenshot 2024-04-04 at 2 04 55 PM

But I figured I'd get push back.

PR #1155, opened 2024-04-04
Fix bug when a segment endpoint is dragged after the question is graded

benchristel:
## Summary:
Repro steps for the bug:

- View a segment graph question (e.g. the JSON below)

```
{"content":"The graph of $y=h(x)$ is a line segment joining the points $(1,9)$ and $(3,2)$. \n\n**Drag the endpoints of the segment below to graph $y=h^{-1}(x)$.\n **\n\n[[☃ interactive-graph 1]]","images":{},"widgets":{"interactive-graph 1":{"alignment":"default","graded":true,"options":{"backgroundImage":{"bottom":0,"height":425,"left":0,"scale":1,"url":"web+graphie://ka-perseus-graphie.s3.amazonaws.com/a82b4098da55c75b68d2c1fd9f348923b8935ccb","width":425},"correct":{"coords":[[[9,1],[2,3]]],"type":"segment"},"graph":{"type":"segment"},"gridStep":[1,1],"labels":["x","y"],"markings":"none","range":[[-10,10],[-10,10]],"rulerLabel":"","rulerTicks":10,"showProtractor":false,"showRuler":false,"snapStep":[0.5,0.5],"step":[1,1]},"type":"interactive-graph","version":{"major":0,"minor":0}}}}
```

- Drag the leftmost point of the segment to the right, so the
  left-to-right order of the points is reversed.
- Grade the question (it doesn't matter if your answer is correct or
  incorrect)
- Drag one of the points with the mouse

Expected: the point you drag moves.

Actual: the *other point* moves.

The bug happened because the grading code mutated arrays from the graph
state. This commit fixes that by cloning the arrays before mutating
them.

Issue: https://khanacademy.atlassian.net/browse/LEMS-1900

Test plan:

Visit the interactive graph flipbook (`yarn dev`) and follow the repro
steps. The bug should not repro.

### DIFF
--- a/packages/perseus/src/util.ts
+++ b/packages/perseus/src/util.ts
@@ -919,6 +919,23 @@ const unescapeMathMode: (label: string) => string = (label) =>
 
 const random: RNG = seededRNG(new Date().getTime() & 0xffffffff);
 
+// TODO(benchristel): in the future, we may want to make deepClone work for
+// Record<string, Cloneable> as well. Currently, it only does arrays.
+type Cloneable =
+    | null
+    | undefined
+    | boolean
+    | string
+    | number
+    | Cloneable[]
+    | readonly Cloneable[];
+function deepClone<T extends Cloneable>(obj: T): T {
+    if (Array.isArray(obj)) {
+        return obj.map(deepClone) as T;
+    }
+    return obj;
+}
+
 const Util = {
     inputPathsEqual,
     nestedMap,
@@ -965,6 +982,7 @@ const Util = {
     textarea,
     unescapeMathMode,
     random,
+    deepClone,
 } as const;
 
 export default Util;
--- a/packages/perseus/src/widgets/interactive-graph.test.tsx
+++ b/packages/perseus/src/widgets/interactive-graph.test.tsx
@@ -171,6 +171,8 @@ describe("InteractiveGraph.validate on a segment question", () => {
 
         InteractiveGraph.widget.validate(guess, rubric, null);
 
+        // Narrow the type of `rubric.correct` to segment graph; otherwise TS
+        // thinks it might not have a `coords` property.
         invariant(rubric.correct.type === "segment");
         expect(rubric.correct.coords).toEqual([
             [
--- a/packages/perseus/src/widgets/interactive-graph.test.tsx
+++ b/packages/perseus/src/widgets/interactive-graph.test.tsx
@@ -1,0 +1,182 @@
+import invariant from "tiny-invariant";
+
+import InteractiveGraph, {type Rubric} from "./interactive-graph";
+
+import type {PerseusGraphType} from "@khanacademy/perseus";
+
+function createRubric(graph: PerseusGraphType): Rubric {
+    return {graph, correct: graph};
+}
+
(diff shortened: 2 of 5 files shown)


packages/perseus/src/util.ts
```diff
@@ -919,6 +919,23 @@ const unescapeMathMode: (label: string) => string = (label) =>
...
+// Record<string, Cloneable> as well. Currently, it only does arrays.
+type Cloneable =
+    | null
+    | undefined
+    | boolean
+    | string
+    | number
+    | Cloneable[]
+    | readonly Cloneable[];
+function deepClone<T extends Cloneable>(obj: T): T {
```
handeyeco, line 932: NAN: This makes me think we need to switch from Underscore to Lodash - Lodash has a `cloneDeep` helper, though I'm not sure how well it works with TS.

packages/perseus/src/widgets/interactive-graph.test.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+        expect(guess.coords).toEqual([
+            [
+                [1, 1],
+                [0, 0],
+            ],
+        ]);
+    });
+
+    it("does not modify the `rubric` data", () => {
```
handeyeco, line 152: I don't really understand this test.
1. Are the coords supposed to be inverted like in the test above?
2. Why would type not be "segment"?

handeyeco (approved):

packages/perseus/src/widgets/interactive-graph.test.tsx
benchristel: I should probably add a comment to clarify why this test is needed. But it fails if I replace the `deepClone` of `rubric.correct.coords` with a shallow-copy `slice`:

```ts
let guess = Util.deepClone(userInput.coords);
let correct = rubric.correct?.coords.slice();
```

The elements of `correct` later get sorted in-place. This test ensures that that sorting does not disturb the original rubric data.

Re: the `invariant(rubric.correct.type === "segment");` - that's needed for TS to understand that `rubric.correct` is always going to have `coords`. The type of `rubric.correct` is `PerseusGraphType`, so we have to narrow the type to just segment graphs.

packages/perseus/src/util.ts
benchristel: Eventually we will be able to use the browser-native `structuredClone`.

packages/perseus/src/util.ts
```diff
@@ -919,6 +919,23 @@ const unescapeMathMode: (label: string) => string = (label) =>
...
+    if (Array.isArray(obj)) {
+        return obj.map(deepClone) as T;
+    }
+    return obj;
```
jeremywiebe, line 936: This feels a bit misleading. We aren't doing any cloning here, just returning the original value (which I guess is correct given the types that Cloneable currently covers).  

We do already have a `clone` (which is actually a deep clone) in https://github.com/Khan/perseus/blob/53fd3768652b702df4a79b97a0ab4766b11748fe/testing/object-utils.ts#L3 

Could we rationalize these two implementations?

PR #1157, opened 2024-04-05
New Interactive Graph Axis Tick Labels (rendered with the MathJax font)

SonicScrewdriver:
## Summary:
This PR adds a new approach to rendering our interactive graph axis tick labels. In order to ensure that we can support cases where the [tickSteps are less than the gridSteps](https://khanacademy.atlassian.net/browse/LEMS-1890), we're now manually rendering these labels. 

The labels have also been updated to render using the `MathJax_Main-Regular` (or `MJXTEX`) font. 

NOTE: There's still a bug regarding overlapping labels on the initial negative axis points, but that will be tackled in [LEMS-1891](https://khanacademy.atlassian.net/browse/LEMS-1891). 

Issue: LEMS-1878

## Test plan:
- manual testing in flipbook using all edge cases 
- manual testing in webapp to ensure the fonts are still correctly identified 

## Screen shots:
Screenshot 2024-04-10 at 4 18 44 PM

Screenshot 2024-04-10 at 4 19 34 PM

Screenshot 2024-04-10 at 4 20 05 PM

[LEMS-1891]: https://khanacademy.atlassian.net/browse/LEMS-1891?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/axis-ticks.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/axis-ticks.tsx
@@ -97,7 +97,6 @@ export function generateTickLocations(
 type Props = {
     tickStep: [number, number];
     range: [[number, number], [number, number]];
-    graphSize: vec.Vector2;
 };
 
 export const AxisTicks = (props: Props) => {
--- a/packages/perseus/src/widgets/interactive-graphs/axis-ticks.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/axis-ticks.tsx
@@ -71,6 +71,8 @@ export const AxisTicks = (props: Props) => {
     const [yMin, yMax] = range[1];
 
     const yTickStep = props.tickStep[1];
+
+    const yTickStep = props.tickStep[1];
     const xTickStep = props.tickStep[0];
 
     const yGridTicks = generateTickLocations(yTickStep, yMin, yMax);
--- a/packages/perseus/src/widgets/interactive-graphs/axis-ticks.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/axis-ticks.tsx
@@ -1,6 +1,4 @@
 import * as React from "react";
-import "@khanacademy/mathjax-renderer/src/css/mathjax.css";
-import "@khanacademy/mathjax-renderer/src/css/safari-hacks.css";
 
 import {useTransform} from "./graphs/use-transform";
 
--- a/packages/perseus/src/widgets/interactive-graphs/axis-ticks.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/axis-ticks.tsx
@@ -17,8 +17,6 @@ const YGridTick = ({y}: {y: number}) => {
     const pointOnAxis: vec.Vector2 = [0, y];
     const [[xPosition, yPosition]] = useTransform(pointOnAxis);
 
-    const labelString = y < 0 ? y.toString() : " " + y.toString();
-
     return (
         <g className="y-axis-ticks">
             <line
@@ -40,7 +38,7 @@ const YGridTick = ({y}: {y: number}) => {
                     x={xPosition - 10}
                     y={yPosition + 5}
                 >
-                    {labelString}
+                    {y.toString()}
                 </text>
             )}
         </g>
@@ -51,7 +49,6 @@ const XGridTick = ({x}: {x: number}) => {
     const pointOnAxis: vec.Vector2 = [x, 0];
     const [[xPosition, yPosition]] = useTransform(pointOnAxis);
 
-    const labelString = x < 0 ? x.toString() : x.toString();
     return (
         <g className="x-axis-ticks">
             <line
@@ -73,7 +70,7 @@ const XGridTick = ({x}: {x: number}) => {
                     x={xPosition}
                     y={yPosition + 25}
                 >
-                    {labelString}
+                    {x.toString()}
                 </text>
             )}
         </g>
--- a/packages/perseus/src/widgets/interactive-graphs/axis-ticks.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/axis-ticks.tsx
@@ -1,7 +1,9 @@
 import * as React from "react";
 
 import {useTransform} from "./graphs/use-transform";
+import useGraphState from "./reducer/use-graph-state";
 
+import type {GridProps} from "./grid";
 import type {vec} from "mafs";
 
 const tickSize = 10;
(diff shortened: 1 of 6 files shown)


packages/perseus/src/widgets/interactive-graphs/grid.tsx
```diff
@@ -8,12 +8,18 @@ import type {GraphRange} from "../../perseus-types";
 import type {SizeClass} from "../../util/sizing-utils";
 import type {vec} from "mafs";
 
+/*
```
SonicScrewdriver, line 11: I added this comment for my own sanity, and wondered if I should keep it. Happy to remove it however!

packages/perseus/src/widgets/interactive-graphs/axis-ticks.tsx
```diff
@@ -24,6 +24,21 @@ const YGridTick = ({y}: {y: number}) => {
...
             />
+            {
+                // TODO (LEMS-1891): Negative one is a special case as the labels can
+                // overlap with the axis line. We should handle this case more gracefully.
+            }
+            {y !== -1 && (
+                <text
+                    height={20}
+                    width={50}
+                    textAnchor="end"
```
SonicScrewdriver, line 35: We want "end" here as we want to right-align the y axis labels.

packages/perseus/src/widgets/interactive-graphs/axis-ticks.tsx
```diff
@@ -41,6 +56,21 @@ const XGridTick = ({x}: {x: number}) => {
...
             />
+            {
+                // TODO (LEMS-1891): Negative one is a special case as the labels can
+                // overlap with the axis line. We should handle this case more gracefully.
+            }
+            {x !== -1 && (
+                <text
+                    height={20}
+                    width={50}
+                    textAnchor="middle"
```
SonicScrewdriver, line 67: We want middle here, as we want to ensure the x axis tick labels are centered under their appropriate 'tick'.

packages/perseus/src/widgets/interactive-graphs/grid.tsx
```diff
@@ -8,12 +8,18 @@ import type {GraphRange} from "../../perseus-types";
...
+gridStep: Where the lines on the grid will render.
+tickStep: Where the little black axis lines & labels (ticks) should render. (e.g. 1, 2, 3, 4, 5)
+snapStep: Where the graph points will lock to when they are dragged.
+*/
```
SonicScrewdriver, line 15: I added these comments for my own sanity, and wondered if I should keep them. Happy to delete this if it's not useful however.

jeremywiebe: I love that these exist, please don't remove them entirely. 

But, could these go on the `GridProps` type itself?  Or some other type where these keys are defined that we could document?

packages/perseus/src/widgets/interactive-graphs/mafs-styles.css
```diff
@@ -124,3 +124,8 @@
 .MafsView .movable-point:focus-visible .movable-point-focus-outline {
     visibility: visible;
 }
+
+.MafsView .axis-ticks text {
+    font-size: 14px;
+    font-family: "MJXTEX";
```
benchristel, line 130: Where is this font family defined? Will it always be available?

I kind of thought we'd have to add a font-face definition like:

```css
@font-face {
  font-family: MJXTEX;
  src: url("https://cdn.kastatic.org/fonts/mathjax/MathJax_Main-Regular.woff") format("woff");
}
```

But maybe that definition already exists elsewhere in Perseus?

If we don't have an explicit font-face definition, then I think the correct font will only be available if MathJax has rendered an expression since the last full-page refresh.

benchristel (commented):
It looks SO GOOD!

I left one question inline about how the font face gets loaded.

packages/perseus/src/widgets/interactive-graphs/mafs-styles.css
benchristel: If we end up defining the font-face in Perseus CSS, we should probably call it something other than `MJXTEX` to avoid any possible conflict with the font-face that MathJax defines. They _should_ always be the same font, but you never know...

SonicScrewdriver: We found that it's actually already well defined in MathJax itself, which I believe gets pulled into index.css when we compile the dist. 

Screenshot 2024-04-11 at 2 51 24 PM

I made sure to test in webapp, and it loads in properly even if I go directly to the segment graph test page without loading any ancillary MathJax. :)

SonicScrewdriver: I believe @jeremywiebe thought this would be bundled properly after we did some digging! 

It also seems to load properly in flipbook, even when I make sure that there's no other Mathjax on the page at all.
 
Screenshot 2024-04-11 at 3 04 31 PM

I was unable to find any examples of content in webapp that don't _also_ end up using ancillary MathJax as they usually have to indicate points or some kind of value to set up the question. Even the segment testing page has some minor bits of MathJax floating around.🤔

benchristel: It does indeed seem like MathJax sneaks in in a lot of places!

However, I'm still wary of relying on MathJax being the math renderer that is used in production. We allow clients of Perseus to inject any TeX component, and I suspect that third-party users of Perseus (e.g. [Learning Equity](https://khanacademy.slack.com/archives/CJDRXTGQ7/p1712852640716959)) are still using KaTeX.

I tried out switching to a custom `@font-face` and it works. The fonts are only loaded once if we use the same CDN URL that the MathJax renderer is configured to use:

```css
@font-face {
    font-family: Mafs-MJXTEX;
    src: url("https://cdn.kastatic.org/fonts/mathjax/MathJax_Main-Regular.woff") format("woff");
}

.MafsView .axis-ticks text {
    font-size: 14px;
    font-family: "Mafs-MJXTEX";
}
```

SonicScrewdriver: Lovely, that seems safer! I've implemented your suggestion, thank you greatly. :)

packages/perseus/src/widgets/interactive-graphs/axis-ticks.tsx
```diff
@@ -63,10 +97,12 @@ export function generateTickLocations(
 type Props = {
     tickStep: [number, number];
     range: [[number, number], [number, number]];
+    graphSize: vec.Vector2;
```
benchristel, line 100: `graphSize` doesn't seem to be used. I think we can remove it?

benchristel (approved):
LGTM! Just one small thing—I think the `graphSize` prop is unused and can be removed.

PR #1158, opened 2024-04-05
Update workflows to use Node 20-ready actions

somewhatabstract:
## Summary:
This updates the workflows to use action versions that have been updated to use Node 20. This should mean no more Node 16 deprecation warnings.

Issue: FEI-5484

## Test plan:
Put this up and check the workflows run without Node 16 deprecation warnings.

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -30,11 +30,11 @@ jobs:
                   fetch-depth: 0
 
             - name: Get changed files
-              uses: Khan/actions@get-changed-files-v1
+              uses: Khan/actions@get-changed-files-v2
               id: changed
 
             - name: Filter out files that don't need a changeset
-              uses: Khan/actions@filter-files-v0
+              uses: Khan/actions@filter-files-v1
               id: match
               with:
                   changed-files: ${{ steps.changed.outputs.files }}
@@ -70,7 +70,7 @@ jobs:
                   ssh-private-key: ${{ secrets.KHAN_ACTIONS_BOT_SSH_PRIVATE_KEY }}
 
             - name: Get All Changed Files
-              uses: Khan/actions@get-changed-files-v1
+              uses: Khan/actions@get-changed-files-v2
               id: changed
 
             - name: Check formatting
@@ -79,14 +79,14 @@ jobs:
 
             - id: js-files
               name: Find .js(x)/.ts(x) changed files
-              uses: Khan/actions@filter-files-v0
+              uses: Khan/actions@filter-files-v1
               with:
                   changed-files: ${{ steps.changed.outputs.files }}
                   extensions: ".js,.jsx,.ts,.tsx"
                   files: "yarn.lock"
 
             - id: eslint-reset
-              uses: Khan/actions@filter-files-v0
+              uses: Khan/actions@filter-files-v1
               name: Files that would trigger a full eslint run
               with:
                   changed-files: ${{ steps.changed.outputs.files }}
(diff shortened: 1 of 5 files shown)


jeresig (approved):

.github/workflows/storybook.yml
```diff
@@ -39,7 +39,7 @@ jobs:
                   ssh-private-key: ${{ secrets.KHAN_ACTIONS_BOT_SSH_PRIVATE_KEY }}
 
             - name: Publish to Chromatic
-              uses: chromaui/action@latest
+              uses: chromaui/action@v11
```
somewhatabstract, line 42: **note:** Best to pin to a known version to avoid unexpected breaking changes that a major update might introduce.

.github/workflows/node-ci.yml
```diff
@@ -334,7 +334,7 @@ jobs:
             # Note: these two actions are locked to the latest version that were
             # published when I created this yml file (just for security).
             - name: Find existing comment
-              uses: peter-evans/find-comment@034abe94d3191f9c89d870519735beae326f2bdb
+              uses: peter-evans/find-comment@3eae4d37986fb5a8592848f6a574fdf654e61f9e
```
jeremywiebe, line 337: Originally I used a commit hash, but every time I see these hashes I think I should change this to just use the [version numbers](https://github.com/peter-evans/find-comment/tags) instead.  Would you be ok with incorporating this change in this PR? 

```suggestion
              uses: peter-evans/find-comment@v3
```

.github/workflows/storybook.yml
jeremywiebe: TBH, for a service like this, I'd rather us be using `@latest` and deal with issues if they come up.  Otherwise we tend to live on aging versions. I view this similar to how we live on evergreen browsers these days. Not a blocker for this PR, but I'd be in favour of keeping `@latest`.

.github/workflows/node-ci.yml
somewhatabstract: A blog a year or so ago recommend using commit SHAs for actions that are less trusted to avoid malicious releases getting into workflows. That's why it has been preferred for this specific set of actions. Some have suggested that ALL actions should use SHAs this way to avoid accidentally invoking malicious releases. We have walked a line somewhere between where trusted orgs we tend to use the tags approach, but for some third-party providers, like this one, we use SHAs.

somewhatabstract: It's actually recommended practice from GitHub.


> Pin actions to a full length commit SHA
> 
> Pinning an action to a full length commit SHA is currently the only way to use an action as an immutable release. Pinning to a particular SHA helps mitigate the risk of a bad actor adding a backdoor to the action's repository, as they would need to generate a SHA-1 collision for a valid Git object payload. When selecting a SHA, you should verify it is from the action's repository and not a repository fork. 

https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

somewhatabstract: And:

> Pin actions to a tag only if you trust the creator
> 
> Although pinning to a commit SHA is the most secure option, specifying a tag is more convenient and is widely used. If you’d like to specify a tag, then be sure that you trust the action's creators. The ‘Verified creator’ badge on GitHub Marketplace is a useful signal, as it indicates that the action was written by a team whose identity has been verified by GitHub. Note that there is risk to this approach even if you trust the author, because a tag can be moved or deleted if a bad actor gains access to the repository storing the action.

With all that in mind, I think keeping this particular set of actions as SHAs is likely the right thing to do.

.github/workflows/storybook.yml
somewhatabstract: The problem with `@latest` is it opts us in to protentially breaking changes. We should be in control of that.

.github/workflows/node-ci.yml
jeremywiebe: Nice.  Thanks for the documentation links!  That makes sense from a security point of view.

PR #1160, opened 2024-04-08
Adjust Button Styling to Match Intended UI from Content

mark-fitzgerald:
## Summary:
Adjust button styling to be left-aligned, and to use the line spacing of the surrounding text.

Issue: none

## Before:
Screenshot 2024-04-08 at 1 09 13 PM

## After:
Screenshot 2024-04-08 at 1 08 12 PM

### DIFF
--- a/packages/perseus/src/widgets/explanation.tsx
+++ b/packages/perseus/src/widgets/explanation.tsx
@@ -92,6 +92,7 @@ class Explanation extends React.Component<Props, State> {
         // While the button is not normally included in a block of text, it needs to be able to accommodate such a case.
         const buttonStyleOverrides = {
             height: "auto",
+            lineHeight: "inherit",
             marginLeft: "-2px", // transfer space on the left side to the padding
             marginRight: "2px", // adds space between the right border and any text that follows
             paddingLeft: "2px", // adds space between the left border and the button text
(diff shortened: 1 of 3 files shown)


nishasy (approved):
Nice!

packages/perseus/src/widgets/explanation.tsx
```diff
@@ -92,14 +92,16 @@ class Explanation extends React.Component<Props, State> {
         // While the button is not normally included in a block of text, it needs to be able to accommodate such a case.
         const buttonStyleOverrides = {
             height: "auto",
+            lineHeight: "inherit",
```
jeremywiebe, line 95: Not directly related to this change, but these styles (`buttonStyleOverrides` and `labelStyle`) could (should?) go into the `const styles` object later on if there's no dynamic style calculation in them.

jeremywiebe (approved):
One comment (non-blocking)

packages/perseus/src/widgets/explanation.tsx
mark-fitzgerald: (face-palm)
I over-read the WB documentation and implied something that wasn't there. Still getting used to this Stylesheet (CSS-in-JS) implementation.

PR #1162, opened 2024-04-09
Update focus style for line segments to match designs

benchristel:
Issue: https://khanacademy.atlassian.net/browse/LEMS-1892

## Test plan:

- `yarn dev`
- `open http://localhost:5173`
- Enable Mafs for segment graphs
- tab to the line segment
- the line should be outlined in blue

Screen Shot 2024-04-09 at 11 31 01 AM

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-line.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-line.tsx
@@ -73,6 +73,18 @@ export const MovableLine = (props: Props) => {
                 <SVGLine
                     start={startPtPx}
                     end={endPtPx}
+                    className="movable-line-focus-outline"
+                    style={{}}
+                />
+                <SVGLine
+                    start={startPtPx}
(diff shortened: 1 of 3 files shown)


packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-line.tsx
```diff
@@ -70,6 +70,18 @@ export const MovableLine = (props: Props) => {
...
+                    start={startPtPx}
+                    end={endPtPx}
+                    className="movable-line-focus-outline"
+                    style={{}}
+                />
+                <SVGLine
+                    start={startPtPx}
+                    end={endPtPx}
+                    className="movable-line-focus-outline-gap"
+                    style={{}}
```
jeremywiebe, line 83: Given that `style` is made optional in this PR, is there a reason we need/want to explicitly pass an empty `style` here?

jeremywiebe (approved):
Looks good to me.

PR #1164, opened 2024-04-09
Restore Padding Around Mafs Graph (Match Legacy Position)

mark-fitzgerald:
## Summary:
Add styling that provides the legacy padding at the top and the right of the graph.

Issue: LEMS-1901

## Test plan:

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
@@ -120,6 +120,8 @@ export const MafsGraph = React.forwardRef<
                     width,
                     height,
                     position: "relative",
+                    padding: "25px 25px 0 0",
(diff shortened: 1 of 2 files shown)


benchristel (approved):
Looks great!

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -120,6 +120,8 @@ export const MafsGraph = React.forwardRef<
                     width,
                     height,
                     position: "relative",
+                    padding: "25px 25px 0 0",
+                    boxSizing: "content-box",
```
benchristel, line 124: It's a bit unfortunate that we have to use `content-box` here, but since we're trying to preserve legacy behavior, it's probably the most straightforward and least error-prone approach.

PR #1166, opened 2024-04-10
Split MafsGraph into two pieces to make it easier to test

handeyeco:
## Summary:
This is pretty simple PR conceptually, but I think it's going to be really helpful for actual DOM unit tests (vs reducer/logic unit test). In a perfect world, the reducer would be the single source of all logic and the UI would be the visual representation of the output of that logic (state). However Mafs _also_ has logic in it; while I could test what the reducer would do given a certain action, I was having trouble testing the result of a user interaction with Mafs.

So this PR splits MafsGraph into two pieces:
- StatefulMafsGraph: which holds the `useReducer`
- MafsGraph: which basically holds everything else

The benefits:
- I can pass MafsGraph an initial state, so bootstrapping a test is a breeze
- I can mock and spy on a `dispatch` function to assert that I'm getting the actions I want

I added two regression tests for bugs I was having trouble testing before.

I got Ben's blessing to explore this idea before I started.

Part of: [LEMS-1907](https://khanacademy.atlassian.net/browse/LEMS-1907)

## Test plan:
- Tests pass

[LEMS-1907]: https://khanacademy.atlassian.net/browse/LEMS-1907?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-line.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-line.tsx
@@ -58,6 +58,7 @@ export const MovableLine = (props: Props) => {
                 ref={line}
                 tabIndex={0}
                 className="movable-line"
+                data-test-id="movable-line"
                 style={{cursor: dragging ? "grabbing" : "grab"}}
             >
                 {/**
@@ -76,6 +77,7 @@ export const MovableLine = (props: Props) => {
                         strokeWidth: "var(--movable-line-stroke-weight)",
                     }}
                     dragging={dragging}
+                    data-test-id="movable-line__line"
                 />
             </g>
 
@@ -94,9 +96,10 @@ function SVGLine(props: {
     style: SVGProps<SVGLineElement>["style"];
     dragging?: boolean;
 }) {
-    const {start, end, style, dragging} = props;
+    const {start, end, style, dragging, ...rest} = props;
     return (
         <line
+            {...rest}
             x1={start[0]}
             y1={start[1]}
             x2={end[0]}
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-line.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-line.tsx
@@ -2,8 +2,7 @@ import {vec, useMovable, Vector} from "mafs";
 import {useRef} from "react";
 import * as React from "react";
 
-import useGraphState from "../../reducer/use-graph-state";
-import {TARGET_SIZE, snap} from "../../utils";
+import {TARGET_SIZE} from "../../utils";
 import {useTransform} from "../use-transform";
 import {getRayIntersectionCoords} from "../utils";
 
@@ -26,7 +25,6 @@ type Props = {
 };
 
 export const MovableLine = (props: Props) => {
-    const {state} = useGraphState();
     const {start, end, onMove, extend, stroke = defaultStroke} = props;
     const midpoint = vec.midpoint(start, end);
 
@@ -51,7 +49,9 @@ export const MovableLine = (props: Props) => {
         onMove: (newPoint) => {
             onMove(vec.sub(newPoint, midpoint));
         },
-        constrain: (p) => snap(state.snapStep, p),
+        constrain: (p) => {
+            return p;
+        },
     });
 
     return (
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-line.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-line.tsx
@@ -77,7 +77,7 @@ export const MovableLine = (props: Props) => {
                         strokeWidth: "var(--movable-line-stroke-weight)",
                     }}
                     dragging={dragging}
-                    data-test-id="movable-line__line"
+                    testId="movable-line__line"
                 />
             </g>
 
@@ -95,17 +95,18 @@ function SVGLine(props: {
     end: vec.Vector2;
     style: SVGProps<SVGLineElement>["style"];
     dragging?: boolean;
+    testId?: string;
 }) {
-    const {start, end, style, dragging, ...rest} = props;
+    const {start, end, style, dragging, testId} = props;
     return (
         <line
-            {...rest}
             x1={start[0]}
             y1={start[1]}
             x2={end[0]}
             y2={end[1]}
             style={style}
             className={dragging ? "movable-dragging" : undefined}
+            data-test-id={testId}
         />
     );
 }
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-line.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-line.tsx
@@ -49,9 +49,7 @@ export const MovableLine = (props: Props) => {
         onMove: (newPoint) => {
             onMove(vec.sub(newPoint, midpoint));
         },
-        constrain: (p) => {
-            return p;
-        },
+        constrain: (p) => p,
     });
 
     return (
--- a/packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
@@ -9,6 +9,7 @@ import {interactiveGraphReducer} from "./reducer/interactive-graph-reducer";
 import type {Props as MafsGraphProps} from "./mafs-graph";
 import type {InteractiveGraphState} from "./types";
 import type {UserEvent} from "@testing-library/user-event";
+import type {vec} from "mafs";
 
 function getBaseMafsGraphProps(): MafsGraphProps {
     return {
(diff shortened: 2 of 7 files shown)


packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-line.tsx
```diff
@@ -94,9 +96,10 @@ function SVGLine(props: {
     style: SVGProps<SVGLineElement>["style"];
     dragging?: boolean;
 }) {
-    const {start, end, style, dragging} = props;
+    const {start, end, style, dragging, ...rest} = props;
```
handeyeco, line 99: Just to pass `data-test-id`, I could also just add `data-test-id` to props if we think that would be better than using `...rest`

packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
```diff
@@ -63,3 +68,192 @@ describe("MafsGraph", () => {
         expect(mockChangeHandler).toHaveBeenCalled();
     });
 });
+
+function map(
```
handeyeco, line 72: This is really common functionality. In [P5](https://p5js.org/reference/#/p5/map) it's called `map`; I think Math speak it's called a linear transform. I left this here basically to ask if anyone knows if we have this already (maybe in underscore/lodash/jquery with a different name?)

packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
```diff
@@ -63,3 +68,192 @@ describe("MafsGraph", () => {
...
+
+        expect(line).toBeInTheDocument();
+        expect(line.getAttribute("x1")).toBe("0");
+        expect(line.getAttribute("y1")).toBe("0");
+
+        // Map graph coordinates to SVG coordinates
+        const halfXBox = baseMafsGraphProps.box[0] / 2;
+        const halfYBox = baseMafsGraphProps.box[1] / 2;
+        const expectedX2 = map(-7, -10, 10, halfXBox * -1, halfXBox);
+        const expectedY2 = map(0.5, -10, 10, halfYBox * -1, halfYBox) * -1;
```
handeyeco, line 131: You'll see this a lot in the tests. Basically I'm converting graph space (-10 to 10) to pixel space (-200 to 200 when width/height is 400).

I think it's equally valuable to be testing logic numbers (graph space, the product of the reducer) and real numbers (pixel space, the product of actual SVG rendering).

packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
```diff
@@ -63,3 +68,192 @@ describe("MafsGraph", () => {
...
+            />,
+        );
+
+        const group = screen.getByTestId("movable-point");
+        group.focus();
+        await userEvent.keyboard("[ArrowRight]");
+        const action = movePoint(0, [4, 2]);
+        expect(mockDispatch).toHaveBeenCalledWith(action);
+
+        const updatedState = interactiveGraphReducer(state, action);
```
handeyeco, line 171: Since the dispatch is mocked, I'm manually recalculating state and rerendering.

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -91,7 +106,7 @@ export const MafsGraph = React.forwardRef<
     const [xSnap, ySnap] = props.snapStep;
     useEffect(() => {
         dispatch(changeSnapStep([xSnap, ySnap]));
-    }, [xSnap, ySnap]);
+    }, [dispatch, xSnap, ySnap]);
```
handeyeco, line 109: `dispatch` is required by the linter, but IIRC `dispatch` is guaranteed to not change.

packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
handeyeco: Example: `map(2, 0, 10, 0, 100)` scales what would be 2 in a 0-10 range into 20 in a 0-100 range.

mark-fitzgerald: I wonder if we should call the function something like `linearTransform` (or something similar) so that we don't imply a meaning that isn't there due to the JavaScript `Array.prototype.map` function.

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-line.tsx
jeremywiebe: Although TypeScript will still properly type-check this, I tend to favour explicit props so that it's easier to read and see what exactly is being passed down. But I'm not adamant about this. :)

packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
benchristel: Mafs exports utilities for creating and applying transformation matrices, which I think could also be used to implement `graphToPixel`. E.g.

```ts
const graphToPixelTransform = vec.matrixBuilder().translate(-rangeMin, 0)
  .scale(fullPixelSpace / (rangeMax - rangeMin)).translate(halfPixelSpace);

vec.transform(point, graphToPixelTransform);
```

I don't think that's the right logic, but something like that structure could work.

packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
```diff
@@ -63,3 +69,198 @@ describe("MafsGraph", () => {
...
+    targetMin: number,
+    targetMax: number,
+): number {
+    return (
+        ((num - startMin) / (startMax - startMin)) * (targetMax - targetMin) +
+        targetMin
+    );
+}
+
+function graphToPixel(
```
jeremywiebe, line 86: What do you think of basically copy-pasting the Mafs transform logic into this function (it doesn't look like it's exposed in Mafs anywhere). 

https://github.com/stevenpetryk/mafs/blob/adc9122f102dc18128644d752d07390a27b92a64/src/display/Point.tsx#L21

jeremywiebe (commented):
I really like this idea.  I think Ben is reviewing this right now too so I'll leave it to him to do the final approval.

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-line.tsx
handeyeco: Changed

packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
```diff
@@ -63,3 +69,198 @@ describe("MafsGraph", () => {
...
+
+        // Map graph coordinates to SVG coordinates
+        const expectedX1 = graphToPixel(0);
+        const expectedY1 = graphToPixel(0) * -1;
+        const expectedX2 = graphToPixel(-7);
+        const expectedY2 = graphToPixel(0.5) * -1;
+        expect(line.getAttribute("x1")).toBe(expectedX1 + "");
+        expect(line.getAttribute("y1")).toBe(expectedY1 + "");
+        expect(line.getAttribute("x2")).toBe(expectedX2 + "");
+        expect(line.getAttribute("y2")).toBe(expectedY2 + "");
```
benchristel, line 150: Idea: could we configure the graph so one unit = one pixel? Then we wouldn't have to transform the `expected` coordinates.

benchristel (approved):
This looks good, but I think I'm missing something: how did you fix the bug?

handeyeco:
@benchristel This branch is forked from the bug fix since I wanted to keep the changes focused: https://github.com/Khan/perseus/pull/1163

packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
handeyeco: @benchristel Done

packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
handeyeco: I ended up using Ben's [recommendation](https://github.com/Khan/perseus/pull/1166#discussion_r1559726655) which will use Mafs transform helpers.

packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
handeyeco: I personally like that this really approximates what a user would see.

packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
handeyeco: I think you're right, but I think I'm going to keep this as-is if for no other reason than to show how nuanced we can get.

packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
handeyeco: I answered this in another comment, but I ended up using helpers from Mafs for this.

PR #1167, opened 2024-04-10
Possible fix for Chrome repositioning bug

handeyeco:
## Summary:
This works, but I don't know why and I don't know what repercussions there will be. I wasn't able to repro in FF, but I was in Chrome/Safari (this fix works for both). It makes me wonder if this is a browser bug.

It's overriding [this CSS](https://github.com/stevenpetryk/mafs/blob/adc9122f102dc18128644d752d07390a27b92a64/core.css#L4) from Mafs' first commit.

Before
https://github.com/Khan/perseus/assets/16308368/fa2c76aa-c3ee-4672-92ef-e784857fe276

After
https://github.com/Khan/perseus/assets/16308368/145584fe-cc32-44bb-9be2-83c7b99cf792

Issue: [LEMS-1906](https://khanacademy.atlassian.net/browse/LEMS-1906)

## Alternative

An alternative that seems to work: `.MafsView > svg { display: block; }`. Given the amount of spacing at the bottom, it looks suspiciously like this is `inline` spacing accidentally surfacing.

I'm actually leaning towards this solution. Curious what others think.

[LEMS-1906]: https://khanacademy.atlassian.net/browse/LEMS-1906?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/mafs-styles.css
+++ b/packages/perseus/src/widgets/interactive-graphs/mafs-styles.css
@@ -29,6 +29,7 @@
 }
 
 .MafsView > svg {
+    /* Chrome/Safari bugfix for LEMS-1906 */
     display: block;
 }
 
--- a/packages/perseus/src/widgets/interactive-graphs/mafs-styles.css
+++ b/packages/perseus/src/widgets/interactive-graphs/mafs-styles.css
@@ -26,6 +26,7 @@
     --movable-line-stroke-color: var(--mafs-blue);
     --movable-line-stroke-weight: 2px;
     --movable-line-stroke-weight-active: 4px;
+    overflow: initial;
 }
 
 .MafsView .movable-line:hover,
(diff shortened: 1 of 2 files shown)


packages/perseus/src/widgets/interactive-graphs/mafs-styles.css
```diff
@@ -26,6 +26,7 @@
     --movable-line-stroke-color: var(--mafs-blue);
     --movable-line-stroke-weight: 2px;
     --movable-line-stroke-weight-active: 4px;
+    overflow: initial;
```
benchristel, line 29: Yet another possible solution: `position: absolute` on the `svg` child of `.MafsView` fixes it for me (in Edge). I think that only works, though, because the parent of `.MafsView` happens to have `position: relative`. So to make this not context-dependent, I think we'd want `.MafsView` to have `position: relative`.

handeyeco:
I decided to go with option 2: make the SVG element a block element. It feels like the least invasive: we don't modify `.MafsView` and we're already acting like the graph would be a block element anyway (so why wouldn't the SVG be one?)

benchristel (approved):
Looks good. My one suggestion would be to put a comment in the CSS explaining why we're setting `display: block`.

PR #1169, opened 2024-04-15
Add "hairLines" for movable point in mobile

handeyeco:
## Summary:
Add hairlines when dragging a movable point (mobile only).

https://github.com/Khan/perseus/assets/16308368/a71eb93a-387b-464d-96ae-b47cd4ef79e2

Issue: [LEMS-1913](https://khanacademy.atlassian.net/browse/LEMS-1913)

## Test plan:
- Use a phone or simulate mobile
- Go to an exercise that has an [interactive chart](https://khanacademy.org/internal-courses/test-everything/test-everything-1/te-interactive-graph/e/test-everything-segment-graph-exercise)
- Drag a point and note there are cross hairs

[LEMS-1913]: https://khanacademy.atlassian.net/browse/LEMS-1913?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/perseus/src/widgets/interactive-graph.tsx
```diff
@@ -1826,6 +1826,7 @@ class InteractiveGraph extends React.Component<Props, State> {
                     gridStep={gridStep}
                     snapStep={snapStep}
                     box={box}
+                    isMobile={this.props.apiOptions.isMobile}
```
handeyeco, line 1829: I kind of want to pass all of APIOptions and maybe have a context that lets components consume information as needed from APIOptions, but I was worried that would lead to scope creep.

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.tsx
```diff
@@ -33,37 +33,67 @@ export const StyledMovablePoint = (props: Props) => {
...
-        </g>
+        <>
+            {showHairlines && (
+                <g>
+                    <line
+                        x1={verticalStartX}
+                        y1={y}
+                        x2={verticalEndX}
+                        y2={y}
+                        stroke={WBColor.green}
```
handeyeco, line 54: This green looks worse than the original green, but this is the green WB provides and I think we're trying to stick to those colors.

benchristel (approved):
LGTM. In response to [your question in Slack](https://khanacademy.slack.com/archives/C067UM1QAR4/p1713211118877839) — I think there's a way to keep `showHairlines` out of the state without too much duplication.

```tsx
const renderGraph = (props: {
    state: InteractiveGraphState;
}) => {
    const {state, dispatch} = props;
    const {type} = state;
    const commonProps = {
        dispatch,
        showHairlines: false,
    }
    switch (type) {
        case "segment":
            return <SegmentGraph graphState={state} {...commonProps} />;
        case "linear":
        case "linear-system":
            return <LinearGraph graphState={state} {...commonProps} />;
        case "ray":
            return <RayGraph graphState={state} {...commonProps} />;
        case "polygon":
            return <PolygonGraph graphState={state} {...commonProps} />;
        case "point":
            return <PointGraph graphState={state} {...commonProps} />;
        case "circle":
            throw new Error("the circle graph type is not yet implemented");
        default:
            return new UnreachableCaseError(type);
    }
};
```

PR #1173, opened 2024-04-16
Change APIOptions' 'flags' field to be strongly typed

jeremywiebe:
## Summary:

We recently introduced feature flags into Perseus via `APIOptions`. These flags were originally loosely typed with the `Record<>` type, but this is problematic for a few reasons:

  1. The host application does not know what flags are supported (or that the structure is nested)
  2. The host has no way of knowing when a Perseus update drops support for a flag 
  3. Inside Pereus, there's no way of knowing what flags might be passed in, so typos can easily cause a flag not to be used. 

So, this PR changes the flags field to be strongly and explicitly typed. Adding flags should be relatively trivial (if it's a Mafs flag, just add the flag name to the `MafsFlags` array, otherwise add it as you would to any Typescript type)

Issue: "none"

## Test plan:

`yarn test`
`yarn tsc`

### DIFF
--- a/packages/perseus/src/types.ts
+++ b/packages/perseus/src/types.ts
@@ -127,6 +127,19 @@ type TrackInteractionArgs = {
 } & Partial<TrackingGradedGroupExtraArguments> &
     Partial<TrackingSequenceExtraArguments>;
 
+export const MafsFlags = [
+    /** Enables the `segment` interactive-graph type.  */
+    "segment",
+    /** Enables the `linear` interactive-graph type.  */
+    "linear",
+    /** Enables the `linear-system` interactive-graph type.  */
+    "linear-system",
+    /** Enables the `ray` interactive-graph type.  */
+    "ray",
+    /** Enables the `polygon` interactive-graph type.  */
+    "polygon",
+] as const;
+
 /**
  * APIOptions provides different ways to customize the behaviour of Perseus.
  *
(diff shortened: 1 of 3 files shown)


packages/perseus/src/types.ts
```diff
@@ -127,6 +127,19 @@ type TrackInteractionArgs = {
 } & Partial<TrackingGradedGroupExtraArguments> &
     Partial<TrackingSequenceExtraArguments>;
 
+export const MafsFlags = [
+    /** Enables the `segment` interactive-graph type.  */
```
jeremywiebe, line 131: I'm on the fence if these are actually helpful. 🤔

benchristel (approved):
So, my understanding is that Ned originally structured the `flags` object as a flexible `Record` so that we could remove flags without a breaking change to the Perseus API. With this new type setup, we'll technically (if we're following semver) have to do a major release when we remove the mafs flags, because clients of Perseus will get type errors if they upgrade and don't remove usages of those flags from their code.

Personally, I think the type safety we gain in the meantime is worth the small inconvenience, so I've approved the PR. Wanted to make sure you're aware of the tradeoff, though.

jeremywiebe:
> So, my understanding is that Ned originally structured the `flags` object as a flexible `Record` so that we could remove flags without a breaking change to the Perseus API. With this new type setup, we'll technically (if we're following semver) have to do a major release when we remove the mafs flags, because clients of Perseus will get type errors if they upgrade and don't remove usages of those flags from their code.
> 
> Personally, I think the type safety we gain in the meantime is worth the small inconvenience, so I've approved the PR. Wanted to make sure you're aware of the tradeoff, though.

Thanks for bringing the point about backwards compatibility up.  I did think a bit about that and think that I'd rather be forced to update the host application when I upgrade Perseus than keep passing flags, unsure if they are correct or unused.

PR #1175, opened 2024-04-16
Ensure axis ticks are only displayed if the markings are set to graph

SonicScrewdriver:
## Summary:
Issue: LEMS-1922 LEMS-1922

This is a simple PR that fixes a bug where our axis ticks are appearing even when our markings type is set to "grid" or "none".

## Test plan:


benchristel (commented):
I feel like this might be worth a test. My opinion on automated UI testing is basically 🤷 most of the time, _but_ when we find bugs I like to have regression tests to make sure the bug doesn't come back.

benchristel (approved):
LGTM!

PR #1176, opened 2024-04-16
Style the arrowheads on Mafs linear graphs

benchristel:
## Summary:
The arrowheads now match the legacy graph.

Issue: https://khanacademy.atlassian.net/browse/LEMS-1893

Test plan:

- `yarn dev`
- Visit: http://localhost:5173/?flags=linear%2Clinear-system&search=linear

The graphed lines should have arrowheads to indicate that the line
extends beyond the graph bounds. The arrowheads should stick to the
line as you move it and point in the correct direction.
Screen Shot 2024-04-16 at 3 53 58 PM

### DIFF
--- a/packages/perseus/src/util/svg.ts
+++ b/packages/perseus/src/util/svg.ts
@@ -1,0 +1,63 @@
+export function pathBuilder(): PathBuilder {
+    return new PathBuilder()
+}
+
+export class PathBuilder {
+    private path: Command[] = []
+    private scaleFactor: number = 1
+
+    build(): string {
+        return this.path
+            .map(command => command.scaleBy(this.scaleFactor))
+            .join("")
+    }
+
+    // add a move (M) command to the path
+    move(x: number, y: number): PathBuilder {
+        this.path.push(new MoveCommand(x, y))
+        return this;
+    }
+
+    // add a curve (C) command to the path
+    curve(control1X: number, control1Y: number, control2X: number, control2Y: number, endX: number, endY: number): PathBuilder {
+        this.path.push(new CurveCommand(control1X, control1Y, control2X, control2Y, endX, endY));
+        return this;
+    }
+
+    scale(factor: number): PathBuilder {
+        this.scaleFactor *= factor;
+        return this;
+    }
+}
+
+abstract class Command {
+    abstract type: string
+    coords: number[] = []
+
+    toString() {
+        return `${this.type}${this.coords.join(" ")}`
+    }
+
+    scaleBy(factor: number) {
+        this.coords = this.coords.map(coord => coord * factor)
+        return this;
+    }
+}
+
+class MoveCommand extends Command {
+    type = "M"
+    constructor(x, y) {
+        super();
+        this.coords = [x, y]
+    }
+}
+
+class CurveCommand extends Command {
+    type = "C"
+    constructor(control1X: number, control1Y: number, control2X: number, control2Y: number, endX: number, endY: number) {
+        super();
+        this.coords = [
+            control1X, control1Y, control2X, control2Y, endX, endY
+        ];
+    }
+}
--- a/packages/perseus/src/util/svg.ts
+++ b/packages/perseus/src/util/svg.ts
@@ -1,26 +1,42 @@
 export function pathBuilder(): PathBuilder {
-    return new PathBuilder()
+    return new PathBuilder();
 }
 
 export class PathBuilder {
-    private path: Command[] = []
-    private scaleFactor: number = 1
+    private path: Command[] = [];
+    private scaleFactor: number = 1;
 
     build(): string {
         return this.path
-            .map(command => command.scaleBy(this.scaleFactor))
-            .join("")
+            .map((command) => command.scaleBy(this.scaleFactor))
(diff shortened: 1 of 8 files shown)


packages/perseus/src/widgets/interactive-graphs/graphs/components/arrowhead.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+// We use the pathBuilder here to scale up the SVG path coordinates used
+// elsewhere for arrowheads. We scale in the path itself instead of using a CSS
+// transform because a transform would also scale up the stroke weight, and we
+// don't want that.
+const arrowPath = pathBuilder()
+    .move(-3, 4)
+    .curve(-2.75, 2.5, 0, 0.25, 0.75, 0)
+    .curve(0, -0.25, -2.75, -2.5, -3, -4)
+    .scale(1.4)
```
benchristel, line 19: Note that this also changes the scaling of the arrowheads on the axis graph, since the `Arrowhead` component is used for the axis arrows as well as the linear graph lines. Previously, the scale factor was `1.25`. The new scale matches the legacy graph backgrounds better AFAICT.

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-line.tsx
```diff
@@ -88,7 +89,7 @@ export const MovableLine = (props: Props) => {
                         stroke,
                         strokeWidth: "var(--movable-line-stroke-weight)",
                     }}
-                    dragging={dragging}
+                    className={dragging ? "movable-dragging" : ""}
```
benchristel, line 92: The `dragging` prop on `SVGLine` was only used in this one place, so I opted to remove it and set the className directly. It seems like `SVGLine` shouldn't know about dragging states anyway, just based on its name. Dragging, and the CSS classes related to it, are application-level concepts, not SVG concepts.

packages/perseus/src/util/svg.ts
```diff
@@ -0,0 +1,63 @@
+export function pathBuilder(): PathBuilder {
+    return new PathBuilder()
+}
+
+export class PathBuilder {
```
jeremywiebe, line 5: And so we start rebuilding Raphãel.  (I kid, I kid 😂 )

packages/perseus/src/widgets/interactive-graphs/graphs/components/arrowhead.tsx
```diff
@@ -0,0 +1,47 @@
...
+type Props = {
+    x: number;
+    y: number;
+    rotate: number; // degrees counterclockwise from the positive x-axis
+};
+
+// We use the pathBuilder here to scale up the SVG path coordinates used
+// elsewhere for arrowheads. We scale in the path itself instead of using a CSS
+// transform because a transform would also scale up the stroke weight, and we
+// don't want that.
```
jeremywiebe, line 14: Thanks for explaining why we scale here vs. CSS transform.

packages/perseus/src/widgets/interactive-graphs/graphs/components/arrowhead.tsx
```diff
@@ -0,0 +1,47 @@
+import * as React from "react";
+import {useTransformContext, vec} from "mafs";
+import {pathBuilder} from "../../../../util/svg";
+
+type Props = {
+    x: number;
+    y: number;
+    rotate: number; // degrees counterclockwise from the positive x-axis
```
jeremywiebe, line 8: Would it be clearer to label this "angle" ? 🤔

packages/perseus/src/widgets/interactive-graphs/graphs/components/arrowhead.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export function Arrowhead(props: Props) {
+    const {userTransform, viewTransform} = useTransformContext();
+
+    const point: vec.Vector2 = [props.x, props.y];
+    const userTransformedPoint = vec.transform(point, userTransform);
+    const viewTransformedPoint = vec.transform(
+        userTransformedPoint,
+        viewTransform,
+    );
```
jeremywiebe, line 30: Could we just use our `useTransform()` hook here instead?

jeremywiebe (approved):
Nice.  I like that you've started extracting the individual components into the `./graphs/components` dir.

packages/perseus/src/widgets/interactive-graphs/graphs/components/arrowhead.tsx
benchristel: Yes! This was existing code, so I didn't change it, but I can now.

benchristel:
There is a regression from the legacy graph: arrowheads get cut off at the top and bottom of the graph. I'll fix that.
Screen Shot 2024-04-17 at 8 31 49 AM

benchristel:
> There is a regression from the legacy graph: arrowheads get cut off at the top and bottom of the graph. I'll fix that.

After looking into this, it seems like it's going to be a surprisingly large amount of work, so maybe I'll do a follow-up PR.

PR #1177, opened 2024-04-16
Update locked points in preparation for locked lines

nishasy:
## Summary:
Update to locked points in preparation for locked lines:
- Use Wonder Blocks Accordion's AccordionSection to make the points info collapsible. This will take up less vertical space when necessary.
- Add `toggleable` and `onToggle` props to LockedPointSettings. This should be allow the future LockedLineSettings component to make the points show/hide.
- Add a story to demonstrate said toggleable functionality
- Refactoring to make LockedFiguresSection more universal, not point-specific
- Add more info in the header so the point is easily identifiable (coords, color, filled)

Issue: none

## Test plan:
`yarn jest packages/perseus-editor/src/components/__tests__/locked-point-settings.test.tsx`

Storybook
http://localhost:6006/?path=/story/perseus-editor-widgets-interactive-graph-editor--with-locked-points

## Screenshots
Screenshot 2024-04-16 at 6 28 30 PM

Screenshot 2024-04-16 at 6 28 36 PM

Screenshot 2024-04-16 at 6 28 43 PM

Screenshot 2024-04-16 at 6 52 13 PM

Screenshot 2024-04-16 at 6 52 18 PM

### DIFF
--- a/packages/perseus-editor/src/components/__stories__/locked-point-settings.stories.tsx
+++ b/packages/perseus-editor/src/components/__stories__/locked-point-settings.stories.tsx
@@ -1,0 +1,76 @@
+import * as React from "react";
+
+import LockedPointSettings from "../locked-point-settings";
+import {getDefaultFigureForType} from "../util";
+
+import type {Meta, StoryObj} from "@storybook/react";
+
+export default {
+    title: "Perseus Editor/Components/Locked Point Settings",
+    component: LockedPointSettings,
+} as Meta<typeof LockedPointSettings>;
+
+export const Default = (args): React.ReactElement => {
+    return <LockedPointSettings {...args} />;
+};
+
+type StoryComponentType = StoryObj<typeof LockedPointSettings>;
+
+// Set the default values in the control panel.
+Default.args = {
+    ...getDefaultFigureForType("point"),
+    onChangeProps: () => {},
+};
+
+export const Controlled: StoryComponentType = {
+    render: function Render() {
+        const [props, setProps] = React.useState(
+            getDefaultFigureForType("point"),
+        );
+
+        const handlePropsUpdate = (newProps) => {
+            setProps({
+                ...props,
+                ...newProps,
+            });
+        };
+
+        return (
+            <LockedPointSettings {...props} onChangeProps={handlePropsUpdate} />
+        );
+    },
+};
+
+/**
+ * In some cases, the locked point may be shown or hidden from the graph,
+ * such as when the point is used to define a locked line.
+ *
+ * Use `onToggle` and `toggled` to update the visibility of the locked point.
+ * When `onToggled` is passed in, a switch will be rendered to allow
+ * toggling the point, which will then show/hide the relevant properties
+ * for the locked point.
+ */
+export const Toggleable: StoryComponentType = {
+    render: function Render() {
+        const [toggled, setToggled] = React.useState(true);
+        const [props, setProps] = React.useState(
+            getDefaultFigureForType("point"),
+        );
+
+        const handlePropsUpdate = (newProps) => {
+            setProps({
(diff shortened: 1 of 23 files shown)


packages/perseus-editor/src/components/__tests__/locked-point-settings.test.tsx
jeremywiebe, line 1: **praise**: I always enjoy going through the tests you write. They're well-written, clear, and clean!  Thanks for that!

packages/perseus-editor/src/components/__stories__/locked-point-settings.stories.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * In some cases, the locked point may be shown or hidden from the graph,
+ * such as when the point is used to define a locked line.
+ *
+ * Use `onToggle` and `toggled` to update the visibility of the locked point.
+ * When `onToggled` is passed in, a switch will be rendered to allow
+ * toggling the point, which will then show/hide the relevant properties
+ * for the locked point.
+ */
```
jeremywiebe, line 52: **thought**: I wonder if this would be better as rendered text in the story? Then engineers viewing this story would see this text in Storybook. What do you think?

jeremywiebe:
Tagged @mark-fitzgerald just for visibility (I think you and Nisha may have already talked about this PR). :)

packages/perseus/src/index.ts
```diff
@@ -151,7 +151,7 @@ export type {ParsedValue} from "./util";
 export type {
     LockedFigure,
     LockedFigureType,
-    LockedPoint,
+    LockedPointType,
```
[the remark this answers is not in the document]
nishasy: I added that suffix because in the refactor, I created a `LockedPoint` component for the actual locked point on the mafs graph. Typescript didn't like them having the same name.

packages/perseus-editor/src/components/locked-point-settings.tsx
```diff
@@ -68,105 +78,194 @@ const LockedPointSettings = (props: Props) => {
...
-                        onBlur={handleBlur}
-                        style={styles.textField}
-                    />
-                </View>
+        <View
+            // More specificity so that we can override the default
+            // universal styles from the .less file.
+            className="locked-figure-accordion"
+        >
+            <AccordionSection
```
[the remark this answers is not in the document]
nishasy: I do agree that it would be awesome to break this down. I think I might favor a different approach myself - I think I'll break out the color circle and the color select into different components, and that should make this significantly easier to follow. And then they can be reused for locked lines!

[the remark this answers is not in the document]
nishasy: I'll give that a try and if you still feel like it should be split up by accordion, header, and panel, I'll go ahead and do that

packages/perseus-editor/src/components/color-circle.tsx
```diff
@@ -0,0 +1,41 @@
...
+    color: LockedFigureColor;
+    filled?: boolean;
+};
+
+const ColorCircle = (props: Props) => {
+    const {color, filled = true} = props;
+
+    return (
+        <View
+            aria-label={`Color: ${color}, ${filled ? "filled" : "open"}`}
```
mark-fitzgerald, line 18: What does having `aria-label` here do for us? Isn't this circle more of a decorative indicator next to the text of the option?

nishasy: Hmm, it's used both for both the color dropdown and for the overall summary section. That's a good point though, that would cause it to read twice in the dropdown. I guess I could add a `decorative` prop that causes the label to be added or not?

PR #1178, opened 2024-04-17
Add flag for M1 locked features

jeremywiebe:
## Summary:

This PR adds the feature flag `interactive-graph-locked-features-m1` which gates access to locked features being built in Milestone 1 (M1) of the interactive-graph rebuild. 

This flag will gate access to adding/editing locked points and locked lines. 

Issue: "none"

## Test plan:

### DIFF
--- a/packages/perseus/src/types.ts
+++ b/packages/perseus/src/types.ts
@@ -127,7 +127,7 @@ type TrackInteractionArgs = {
 } & Partial<TrackingGradedGroupExtraArguments> &
     Partial<TrackingSequenceExtraArguments>;
 
-export const MafsFlags = [
+export const MafsGraphTypeFlags = [
     /** Enables the `segment` interactive-graph type.  */
     "segment",
     /** Enables the `linear` interactive-graph type.  */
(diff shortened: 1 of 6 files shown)


packages/perseus/src/types.ts
```diff
@@ -127,7 +127,7 @@ type TrackInteractionArgs = {
 } & Partial<TrackingGradedGroupExtraArguments> &
     Partial<TrackingSequenceExtraArguments>;
 
-export const MafsFlags = [
+export const MafsGraphTypeFlags = [
```
jeremywiebe, line 130: This needed to split because we use this array in the dev Gallery to populate the graph type dropdown. I think it's nice to have it split out anyway.

nishasy (approved):
🎉

packages/perseus/src/renderer.tsx
```diff
@@ -161,34 +161,32 @@ export type Widget = {
...
+            [key: string]: any;
+        }) => unknown;
+        // If linterContext.highlightLint is true, then content will be passed
+        // to the linter and any warnings will be highlighted in the rendered
+        // output.
+        linterContext: LinterContextProps;
+        legacyPerseusLint?: ReadonlyArray<string>;
+        // Skip adding paragraph class
+        inline?: boolean;
+    };
```
benchristel, line 189: Why did this type change? Looks like we lost some properties (images, content, and widgets).

jeremywiebe: I reverted this.  I had rebased on top of `main` and incorrectly done a conflict resolution. Thanks for catching this!

PR #1179, opened 2024-04-17
Change data-test-id to data-testid to match React Testing Library's default

jeremywiebe:
## Summary:

Per a recent conversation in [Slack](https://khanacademy.slack.com/archives/CDHPTMSTB/p1711990121678439) about why `data-testid` isn't working in a Testing Library test, we re-discovered that Wonder Block's choice to use `data-test-id` as the HTML test id attribute doesn't match Testing Library's default (`data-testid`). This predates RTL.

This PR upgrades WB dependencies that used/specified test ids and aligns Perseus Wonder Blocks test id name with RTL's so that we can remove some customization and make it easier for engineers to do the right thing (ie. follow the RTL [docs](https://testing-library.com/docs/queries/bytestid/)).

Issue: FEI-5551

### Related PRs
  * https://github.com/Khan/wonder-blocks/pull/2212
  * https://github.com/Khan/webapp/pull/21335

## Test plan:

`yarn test` :checkmark:
`yarn start` - navigate around (`View`'s documentation correctly mentions its `testId` prop renders `data-testid`.


dev/flipbook.tsx
```diff
@@ -1,9 +1,8 @@
 /* eslint monorepo/no-internal-import: "off", monorepo/no-relative-import: "off", import/no-relative-packages: "off" */
 import Button from "@khanacademy/wonder-blocks-button";
-import Color from "@khanacademy/wonder-blocks-color";
 import {View} from "@khanacademy/wonder-blocks-core";
 import {Strut} from "@khanacademy/wonder-blocks-layout";
-import {spacing} from "@khanacademy/wonder-blocks-tokens";
+import {color, spacing} from "@khanacademy/wonder-blocks-tokens";
 import * as React from "react";
```
jeremywiebe, line 6: I think I missed this when landing the move to tokens, or this resulted from an incorrect merge conflict resolution.  Anyways, easy fix as we just move to tokens.

packages/perseus/package.json
```diff
@@ -32,28 +32,27 @@
...
         "mafs": "https://github.com/Khan/mafs/releases/download/v0.18.1%2Bbuild-cd8b2467893342ba0b02100073fe21f93f8b1125/mafs-0.18.1.tgz"
     },
     "devDependencies": {
-        "@khanacademy/wonder-blocks-button": "^6.2.8",
-        "@khanacademy/wonder-blocks-clickable": "^4.1.0",
-        "@khanacademy/wonder-blocks-core": "^6.3.1",
-        "@khanacademy/wonder-blocks-data": "^13.0.6",
-        "@khanacademy/wonder-blocks-dropdown": "^5.1.6",
-        "@khanacademy/wonder-blocks-form": "^4.4.3",
+        "@khanacademy/wonder-blocks-banner": "^3.0.41",
```
jandrade, line 38: **praise:** yay! thanks for upgrading to the latest Banner version (I think the previous one was still using `wb-color` internally).

packages/perseus/src/widgets/interactive-graphs/graphs/linear.tsx
```diff
@@ -35,8 +35,6 @@ export const LinearGraph = (props: LinearGraphProps) => {
                             moveControlPoint(endpointIndex, destination, i),
                         )
                     }
-                    // "linear" or "linear-system" + index
-                    data-testid={type + i}
```
jandrade, line 39: **question:** Why is this not needed anymore? Is the change intended?

packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx
```diff
@@ -113,7 +113,6 @@ export const PolygonGraph = (props: Props) => {
                     onMove={(destination: vec.Vector2) =>
                         dispatch(movePoint(i, destination))
                     }
-                    data-testid={type + i}
```
jandrade, line 116: Same question here

yarn.lock
```diff
@@ -2441,221 +2441,119 @@
...
   dependencies:
     "@babel/runtime" "^7.18.6"
-    "@khanacademy/wonder-blocks-core" "^6.3.1"
+    "@khanacademy/wonder-blocks-clickable" "^4.2.0"
+    "@khanacademy/wonder-blocks-core" "^6.4.0"
+    "@khanacademy/wonder-blocks-layout" "^2.0.31"
     "@khanacademy/wonder-blocks-tokens" "^1.2.0"
+    "@khanacademy/wonder-blocks-typography" "^2.1.11"
 
-"@khanacademy/wonder-blocks-color@^3.0.0":
```
jandrade, line 2576: **praise:** Yay! no more color 🧹

jandrade (approved):
Looks great to me! Added a couple of questions (that are related to the Linting issues). Besides that, all of this makes sense, thanks!! 🚢

packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx
jeremywiebe: This wasn't used in any tests so I thought it was better to remove them for now. If future tests need them we can bring back the test ids.

packages/perseus/src/widgets/interactive-graphs/graphs/linear.tsx
jeremywiebe: This wasn't used in any tests so I thought it was better to remove them for now. If future tests need them we can bring back the test ids.

PR #1182, opened 2024-04-17
View a locked line

nishasy:
## Summary:
Show lines on on the interactive graph widget when they're
passed in via the `lockedFigures` prop.

Issue: https://khanacademy.atlassian.net/browse/LEMS-1904

## Test plan:

Storybook
- Go to http://localhost:6006/?path=/story/perseus-widgets-interactive-graph--segment-with-mafs-and-locked-lines
- Confirm that a green segment and a purple line show up with a close left point and an open right point

Screenshot 2024-04-17 at 10 07 03 AM

NOTE: The dashed line segments are weird, but this will be fixed in a future PR

Screenshot 2024-04-17 at 10 36 15 AM

Screenshot 2024-04-17 at 10 36 19 AM


packages/perseus-editor/src/components/locked-figures-section.tsx
```diff
@@ -62,14 +62,23 @@ const LockedFiguresSection = (props: Props) => {
...
-                    onChangeProps={(newProps) => changeProps(index, newProps)}
-                    onRemove={() => removeLockedFigure(index)}
-                />
-            ))}
+            {figures?.map((figure, index) => {
+                // TODO(LEMS-1740): Remove this check when
+                // locked lines are added to LockedFigureSettings.
+                if (figure.type === "line") {
+                    return;
+                }
```
nishasy, line 70: There's a codecov warning, but this section is going to be added + tested in the "Add a locked line" PR.

packages/perseus/src/widgets/interactive-graphs/locked-line.tsx
```diff
@@ -0,0 +1,66 @@
...
+                />
+            )}
+            {showEndPoint && (
+                <Point
+                    x={point2.coord[0]}
+                    y={point2.coord[1]}
+                    svgCircleProps={{
+                        style: {
+                            fill: point2.filled
+                                ? wbColor[point2.color]
```
nishasy, line 54: I don't think there's a way for me to test the visual style of this circle.

packages/perseus-editor/src/components/__stories__/locked-point-settings.stories.tsx
```diff
@@ -54,7 +55,7 @@ export const Toggleable: StoryComponentType = {
     render: function Render() {
         const [toggled, setToggled] = React.useState(true);
         const [props, setProps] = React.useState(
-            getDefaultFigureForType("point"),
+            getDefaultFigureForType("point") as LockedPointType,
         );
```
jeremywiebe, line 59: Can we use `satisfies` here instead of `as`?

packages/perseus-editor/src/components/locked-figures-section.tsx
```diff
@@ -53,7 +53,7 @@ const LockedFiguresSection = (props: Props) => {
                 {
                     ...lockedFigures[index],
                     ...newProps,
-                },
+                } as LockedFigure,
```
jeremywiebe, line 56:
```suggestion
                } satisfies LockedFigure,
```

packages/perseus-editor/src/components/util.ts
```diff
@@ -58,6 +59,23 @@ export function getDefaultFigureForType(type: LockedFigureType): LockedFigure {
...
+                        ...getDefaultFigureForType("point"),
+                        coord: [2, 2],
+                    },
+                ],
+                color: "blue",
+                lineStyle: "solid",
+                showArrows: false,
+                showStartPoint: false,
+                showEndPoint: false,
+            } as LockedLineType;
```
jeremywiebe, line 78: I don't think we need these `as ...` statements.  We're returning a type union and because each of these default objects has a `type: "..."` in them, Typescript should properly validate that they conform to the correct member of the union.

packages/perseus/src/perseus-types.ts
```diff
@@ -664,6 +664,17 @@ export type LockedPointType = {
     filled: boolean;
 };
 
+export type LockedLineType = {
+    type: "line";
+    kind: "line" | "ray" | "segment";
+    points: [startPoint: LockedPointType, endPoint: LockedPointType];
+    color: string;
+    lineStyle: "solid" | "dashed";
```
jeremywiebe, line 672: **nit**: since this is in a locked line, maybe this can just be `style`?  Although, using `lineStyle` will allow us to be consistent when we do circle and polygon.  Ignore me. :)

packages/perseus/src/perseus-types.ts
```diff
@@ -664,6 +664,17 @@ export type LockedPointType = {
     filled: boolean;
 };
 
+export type LockedLineType = {
+    type: "line";
+    kind: "line" | "ray" | "segment";
+    points: [startPoint: LockedPointType, endPoint: LockedPointType];
+    color: string;
```
jeremywiebe, line 671: Can we make this the type union of available colors instead?

packages/perseus/src/widgets/interactive-graphs/locked-line.tsx
jeremywiebe: I was thinking you might be able to add a test that uses a snapshot... but I'm not sure how busy that'd be. 🤔   Someone else might have an idea.

.changeset/kind-parrots-serve.md
```diff
@@ -0,0 +1,6 @@
+---
+"@khanacademy/perseus": patch
+"@khanacademy/perseus-editor": patch
```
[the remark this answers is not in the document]
nishasy: I was thinking that too, but I then I wasn't sure if we'd want all patches until the major release with the flag switching on. I'll change it.

packages/perseus/src/perseus-types.ts
nishasy: I think the reason I did `lineStyle` here is that it's not a style type, it's just two possible values. I didn't want it to be confused with svg style props.

packages/perseus/src/perseus-types.ts
nishasy: Doing this in the LockedPoint PR so it can apply to both 👍🏽

packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
```diff
@@ -314,4 +315,112 @@ describe("locked layer", () => {
...
+        });
+        expect(points[1]).toHaveStyle({
+            fill: color.white,
+            stroke: color.purple,
+        });
+        expect(points[2]).toHaveStyle({
+            fill: color.white,
+            stroke: color.green,
+        });
+        expect(points[3]).toHaveStyle({fill: color.green, stroke: color.green});
```
mark-fitzgerald, line 386: NIT: Should this have the same line breaks as the other `expect` statements before it?

mark-fitzgerald (approved):
Looks good!

packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
nishasy: Oh yeah, I think that fix is in a follow-up PR (the MathJax colors one), so I'll leave it like this here for now

PR #1183, opened 2024-04-17
New logic to handle bug regarding overlapping axis tick labels in Interactive Graph.

SonicScrewdriver:
## Summary:
Sometimes our negative axis labels will overlap, which causes visual issues. 

The previous approach for this issue in graphie consisted of hardcoded logic that hid ALL instances of any axis labels for `-1`. Unfortunately, this logic only works properly if the tickStep is specially set to `1`.

This PR adds new logic that ensures that we **_only_** show the first negative axis tick label IF the gridStep is larger than the tick step. This should account for cases where there's clearly enough room to render the label, and alternative tickSteps such as 0.5. 

e.g. If our gridStep = 1 and our tickStep = 2, it means that there are 2 grid lines for every 1 tick. This allows enough room for both these tick labels to render. 

Issue: LEMS-1891

## Test plan:
- manual testing
- new tests 

## Screenshots:
Screenshot 2024-04-17 at 11 09 40 AM
Screenshot 2024-04-17 at 11 09 25 AM
Screenshot 2024-04-17 at 11 07 35 AM

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/axis-ticks.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/axis-ticks.tsx
@@ -11,7 +11,24 @@ const tickStyle: React.CSSProperties = {
     strokeWidth: 1,
 };
 
-const YGridTick = ({y}: {y: number}) => {
+const showTickLabel = (
+    gridStep: number,
+    tickStep: number,
+    label: number,
+): boolean => {
+    const showLabel = gridStep > tickStep ? true : label !== -tickStep;
+    return showLabel;
+};
+
+const YGridTick = ({
+    y,
+    gridStep,
+    tickStep,
+}: {
+    y: number;
+    gridStep: number;
+    tickStep: number;
+}) => {
     const pointOnAxis: vec.Vector2 = [0, y];
     const [[xPosition, yPosition]] = useTransform(pointOnAxis);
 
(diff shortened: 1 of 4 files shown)


packages/perseus/src/widgets/interactive-graphs/axis-ticks.tsx
```diff
@@ -28,7 +49,7 @@ const YGridTick = ({y}: {y: number}) => {
                 // TODO (LEMS-1891): Negative one is a special case as the labels can
```
SonicScrewdriver, line 49: Whoops need to remove this TODO now

benchristel:
I don't really understand why the grid step should affect whether we show the tick numbers. Below is an example where reducing the grid step from the default caused the `-1` labels to show up and collide.

What if instead, we skipped rendering the first negative number on both axes, whatever that number is? Is that feasible? Would it have undesirable effects?

Screen Shot 2024-04-17 at 12 46 17 PM

benchristel:
Okay, I see now that this preserves behavior from the legacy code. I got confused by this sentence in the PR description:

> The pre-existing logic for this this issue on the legacy interactive graph consisted of hardcoded logic that hid ALL instances of any -1 axis labels.

which made me think that the hardcoded logic for `-1` labels came from the legacy (graphie) graph.

benchristel (approved):
LGTM, thanks! I left a suggestion inline for making the tests more self-explanatory, but you can take it or leave it.

I also think it would be good to have a comment in the code explaining that this is legacy logic that we're porting over from Graphie.

PR #1184, opened 2024-04-17
Add `showSolutions`, implement for radio widget (others coming in following PRs)

nedredmond:
Summary
This is a completely backwards-compatible solution that provides an functional-React-friendly API alternative to `showRationalesForCurrentlySelectedChoices` while providing a way to reveal all rationales.

PRs will follow for `label-image`, `group`, and `graded-group`.

Naming it `showSolutions` instead of `showRationales` because for `label-image`, the method displays the correctness of each answer without rationales.

https://github.com/Khan/perseus/assets/23404711/67137a1c-1e3d-4632-9d24-524937c2b80c

Issue: TUT-425

Testing:
Unit tests added and storybook updated to reflect the changes!

### DIFF
--- a/packages/perseus/src/renderer.tsx
+++ b/packages/perseus/src/renderer.tsx
@@ -36,6 +36,7 @@ import type {
     PerseusWidget,
     PerseusWidgetOptions,
     PerseusWidgetsMap,
+    ShowRationales,
 } from "./perseus-types";
 import type {
     APIOptions,
@@ -157,6 +158,9 @@ export type Widget = {
             message?: string | null | undefined,
         ) => unknown | null | undefined,
     ) => PerseusScore;
+    /**
+     * @deprecated Use `showRationales` prop instead.
+     */
     showRationalesForCurrentlySelectedChoices?: (options?: any) => void;
     examples?: () => ReadonlyArray<string>;
 };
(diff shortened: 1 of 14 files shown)


packages/perseus/src/renderer.tsx
```diff
@@ -210,21 +222,25 @@ type Context = LinterContextProps & {
...
+        | "onInteractWithWidget"
+        | "onRender"
+        | "onSerializedStateUpdated"
+        | "questionCompleted"
+        | "showSolutions"
+        | "reviewMode"
+        | "serializedState"
+        | "widgets"
+    >
+>;
```
nedredmond, line 243: I just think this is easier to read and prettier.

packages/perseus/src/widgets/__stories__/radio.stories.tsx
```diff
@@ -9,78 +9,78 @@ import {
...
+>;
 
 export default {
     title: "Perseus/Widgets/Radio",
     args: {
         static: false,
         crossOutEnabled: false,
         reviewMode: false,
+        showSolutions: "none",
+        question: questionWithPassage,
```
nedredmond, line 33: By adding `question` to args, I accidentally made an interactive and more grokkable version of the Perseus JSON panel!

Screenshot 2024-04-17 at 1 45 08 PM

testing/renderer-with-debug-ui.tsx
```diff
@@ -15,18 +15,22 @@ import {registerAllWidgetsForTesting} from "../packages/perseus/src/util/registe
...
     question: PerseusRenderer;
-    apiOptions?: APIOptions;
-    reviewMode?: boolean;
-};
+} & Partial<
+    Omit<
+        ComponentProps<typeof Renderer>,
+        "content" | "images" | "widgets" | "problemNum"
+    >
+>;
```
nedredmond, line 27: Allows other props to be passed into the Renderer if the story has the need

packages/perseus/src/widgets/__stories__/radio.stories.tsx
```diff
@@ -9,78 +9,78 @@ import {
...
-        <RendererWithDebugUI
-            question={applyStoryArgs(questionWithPassage, args)}
-            apiOptions={buildApiOptions(args)}
-            reviewMode={args.reviewMode}
-        />
-    );
+export const SingleSelect = {
+    args: {
+        question: questionWithPassage,
+    },
```
jeremywiebe, line 73: Oh, I do like this!!

packages/perseus/src/widgets/radio/radio.tsx
```diff
@@ -334,6 +350,7 @@ class Radio extends React.Component<Props> {
      * Turn on rationale display for the currently selected choices. Note that
      * this leaves rationales on for choices that are already showing
      * rationales.
+     * @deprecated Internal only. Use `showSolutions` prop instead.
```
jeremywiebe, line 353: Thanks for deprecating this.  I hadn't ever looked too closely at this function, but I really don't like it. Calling an imperative method which takes it's prop value and mutates it and then tells its parent to re-render it.

jeremywiebe (approved):
This looks good. I'm so not a fan of the imperative functions for managing this now that I've looked at them, but that's not your problem. 

Thanks for doing some Storybook tweaking here.

packages/perseus/src/widgets/radio/radio.tsx
nedredmond: After I make similar changes to use the prop elsewhere we can deprecate it on the renderer as well

PR #1185, opened 2024-04-17
Clarify the role of our graph context

benchristel:
## Summary:
We use a React context to convey contextual information about an
interactive graph down the component tree. I think when this context was
introduced, we envisioned it being the one-stop shop for all information
about a graph, but it turns out that doesn't work well. The different
graph types have different state shapes, so getting state from the
context in the individual graph components would force us to add a bunch
of redundant conditionals to narrow the types.

As a result of this limitation, the context was only used in practice
for information that doesn't vary between the graph types.

This PR renames and restructures the context to clarify its role.

Issue: none

Test plan:

`yarn test`

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.tsx
@@ -20,7 +20,7 @@ type Props = {
 const hitboxSizePx = 48;
 
 export const StyledMovablePoint = (props: Props) => {
-    const {state} = useGraphModel();
+    const {range, snapStep, markings} = useGraphModel();
     const hitboxRef = useRef<SVGCircleElement>(null);
     const {point, onMove, color = WBColor.blue} = props;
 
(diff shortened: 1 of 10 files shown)


packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.tsx
```diff
@@ -20,29 +20,29 @@ type Props = {
...
+        constrain: (p) => snap(snapStep, p),
     });
     const pointClasses = `movable-point ${dragging ? "movable-point--dragging" : ""}`;
 
     const [[x, y]] = useTransform(point);
 
-    const [xMin, xMax] = state.range[0];
-    const [yMin, yMax] = state.range[1];
+    const [xMin, xMax] = range[0];
+    const [yMin, yMax] = range[1];
```
jeremywiebe, line 38: I see you restructured this in one line in the first file like this:

```ts
   range: [[xMin, xMax], [yMin, yMax]]
```

That feels nice and compact and avoids spreading (no pun intended) out this destructuring of the graph config across multiple lines.

packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
```diff
@@ -163,14 +163,14 @@ describe("MafsGraph", () => {
             coords: [[2, 2]],
         };
 
-        const baseMafsGraphProps = getBaseMafsGraphProps();
+        const props = {
+            ...getBaseMafsGraphProps(),
+            range: state.range,
+            snapStep: state.snapStep,
```
jeremywiebe, line 169: Should `state` then be renamed to `config`?

jeremywiebe (approved):
This makes a ton of sense. Thanks.

PR #1187, opened 2024-04-18
Don't extend graphed lines all the way to the graph bounds

benchristel:
## Summary:
Our SVG drawings get clipped off at the bounds of the Mafs SVG
container. This was causing the arrowheads at the end of lines to get
clipped in certain circumstances, e.g. when a line intersects the edge
of the graph at a small acute angle.

The solution implemented in this commit is to draw the arrowheads a few
pixels inside the graph bounds. Note that the legacy interactive graph
takes a similar approach to solving this problem.

Issue: LEMS-1893

Test plan:

- `yarn dev`
- View the Mafs linear graphs at `http://localhost:5173`. Drag the lines
  around. The arrowheads should remain visible.
Screen Shot 2024-04-18 at 10 10 50 AM

### DIFF
--- a/.changeset/afraid-moons-live.md
+++ b/.changeset/afraid-moons-live.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Fix a visual bug where the arrowheads of graphed lines were sometimes clipped off at the edge of the graph.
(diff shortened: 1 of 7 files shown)


handeyeco (approved):
Nice!

PR #1188, opened 2024-04-18
Expose showSolutions prop on ServerItemRenderer

nedredmond:
Adds `showSolutions` to `ServerItemRenderer`
Quick follow-up to #1184

### DIFF
--- a/packages/perseus/src/server-item-renderer.tsx
+++ b/packages/perseus/src/server-item-renderer.tsx
@@ -21,7 +21,7 @@ import {ApiOptions} from "./perseus-api";
 import Renderer from "./renderer";
 import Util from "./util";
 
-import type {PerseusItem} from "./perseus-types";
+import type {PerseusItem, ShowSolutions} from "./perseus-types";
 import type {APIOptions, FocusPath, PerseusDependenciesV2} from "./types";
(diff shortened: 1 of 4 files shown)


packages/perseus/src/server-item-renderer.tsx
```diff
@@ -51,11 +52,9 @@ type HOCProps = {
 
 type Props = OwnProps & HOCProps;
 
-type DefaultProps = {
-    apiOptions: Props["apiOptions"];
-    onRendered: Props["onRendered"];
-    linterContext: Props["linterContext"];
-};
+type DefaultProps = Required<
+    Pick<Props, "apiOptions" | "onRendered" | "linterContext">
```
jeremywiebe, line 56: Not asking for changes, but I actually find the original easier to read as opposed to the use of `Pick<>`. (I do like wrapping this all in `Required<>`)

jeremywiebe (approved):

packages/perseus/src/server-item-renderer.tsx
nedredmond: In this case I agree! It gets really busy when there are a lot of DefaultProps, but with just three I did feel a little silly. I was just trying to be consistent with the other files I touched.

PR #1190, opened 2024-04-18
Reboot Mafs state when there's a major graph change

handeyeco:
## Summary:
I wrote a lot of context around this in the ticket.

The **tl;dr**:
- Sometimes in the InteractiveGraphEditor there's a change so monumental that it doesn't make sense to surgically update our existing graph state
- When that happens, just do a hard reboot of graph state

~~I'm not 100% on this change. It feels a little extreme. So I thought I'd make a PR and get feedback before making tests for it.~~

Ben suggested a better option which I adopted.

Demo video (with sound):
https://github.com/Khan/perseus/assets/16308368/811b3c9c-34e7-4f6f-b5cc-9892c402a6ec

Issue: [LEMS-1924](https://khanacademy.atlassian.net/browse/LEMS-1924)

## Test plan:
- Go to Storybook
  - `?path=/story/perseus-editor-widgets-interactive-graph-editor--with-mafs`
- Pick a "segment" chart
- Change the number of segments
- Note the number of segments changes in the preview

[LEMS-1924]: https://khanacademy.atlassian.net/browse/LEMS-1924?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/perseus-editor/src/widgets/interactive-graph-editor.tsx
+++ b/packages/perseus-editor/src/widgets/interactive-graph-editor.tsx
@@ -441,7 +441,7 @@ class InteractiveGraphEditor extends React.Component<Props> {
                                     correct: {
                                         type: "segment",
                                         numSegments: sides,
-                                        coords: null,
+                                        // coords: null,
                                     },
                                 });
                             }}
--- a/packages/perseus-editor/src/widgets/interactive-graph-editor.tsx
+++ b/packages/perseus-editor/src/widgets/interactive-graph-editor.tsx
@@ -213,6 +213,8 @@ class InteractiveGraphEditor extends React.Component<Props> {
 
         const sizeClass = containerSizeClass.SMALL;
         if (this.props.valid === true) {
+            const correct = this.props.correct;
+
             // TODO(aria): send these down all at once
             const graphProps = {
                 ref: "graph",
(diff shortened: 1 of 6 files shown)


benchristel (approved):
I like this approach! Since this is just for the editor, I wonder if it would make sense to instead have the editor force-remount the graph component (by changing its key) when there is a major change.

The key could be something like:

```tsx
key={`${graph.type}:${graph.coords.length}`}
```

I don't have a sense of which approach we should prefer, though. Either one is good with me.

packages/perseus-editor/src/widgets/interactive-graph-editor.tsx
```diff
@@ -244,13 +246,18 @@ class InteractiveGraphEditor extends React.Component<Props> {
                     this.props.onChange({correct: correct});
                 },
             } as const;
+
+            // This is used to force a remount of the graph component
+            // when there's a significant change
+            const graphKey = `${correct.type}:${correct.numSegments || 0}`;
```
jeremywiebe, line 252: This is specific to the `segment` graph, right? Will we need to adjust as we do other graph types?

PR #1192, opened 2024-04-18
Add WB Tooltip to StyledMovablePoint

mark-fitzgerald:
## Summary:

A tooltip for a movable point can be requested at the content creator's discretion. When this option is toggled "on" in the editor, a tooltip with the point's coordinates should appear whenever the user interacts with the point (either via mouse dragging or via keyboard movement).

Issue: LEMS-1860

## Test plan:

1. Launch the developer testing app locally (`yarn dev`).
1. Navigate to `localhost:5173` to view the various graphs.
1. Toggle one or more graphs to use Mafs (via the "Mafs Flags" dropdown menu).
1. Toggle the "Show Tooltips" switch.
1. Interact with a point via mouse to see the tooltip.
   * Move your mouse pointer over one of the points in a Mafs graph and notice that a tooltip appears.
   * Drag one of the points around on the graph and notice that the tooltip follows the point and updates the coordinate readout.
1. Interact with a point via keyboard to see the tooltip.
   * Tab until one of the points has focus and notice that a tooltip appears.
   * Move the point around by using the arrows on your keyboard and notice that the tooltip follows the point and updates the coordinate readout.

### New Tooltip Functionality

https://github.com/Khan/perseus/assets/13896410/493930e0-dbd6-4ef3-bb1c-4249929322b6


dev/gallery.tsx
```diff
@@ -189,9 +217,10 @@ export function Gallery() {
                                 ? graphTypeContainsText(question, search)
                                 : true,
                         )
+                        .map(insertShowTooltips)
                         .map(([question, i]) => (
                             <QuestionRenderer
-                                key={i}
+                                key={`${i}${showTooltips ? "-with-tooltips" : ""}`}
```
mark-fitzgerald, line 223: Adding text to the key to force a re-render when toggling the "Show Tooltips" switch.

dev/gallery.tsx
```diff
@@ -95,6 +95,9 @@ export function Gallery() {
     );
 
     const [isMobile, setIsMobile] = useState(params.get("mobile") === "true");
+    const [showTooltips, setShowTooltips] = useState(
+        params.get("tooltips") === "true",
```
nishasy, line 99: Can this be shortened to `useState(params.get("tooltips"))`? or `!!useState(params.get("tooltips"))`?

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.tsx
```diff
@@ -20,10 +21,16 @@ type Props = {
 const hitboxSizePx = 48;
 
 export const StyledMovablePoint = (props: Props) => {
-    const {range, snapStep, markings} = useGraphConfig();
+    const {range, snapStep, markings, showTooltips} = useGraphConfig();
     const hitboxRef = useRef<SVGCircleElement>(null);
     const {point, onMove, color = WBColor.blue} = props;
 
+    // WB Tooltip requires a color name for the background color.
+    // Since the color in props is a hex value, a reverse lookup is needed.
```
nishasy, line 29: 😭

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.test.tsx
nishasy, line 1: I love these tests! 🎉

nishasy (commented):
Very exciting! Could we consider giving the tooltip a fixed width if it's possible? To stop the size change from being so jarring while moving the point with the half steps adding more digits?

dev/gallery.tsx
nishasy: I didn't see that "true" is a string. You can ignore my suggestion!

.changeset/dry-lemons-doubt.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/perseus": minor
+---
+
+Display tooltip for movable point when option is indicated
```
jeremywiebe, line 5: Our changelogs aren't critically important, but I think it'd be good to clarify here that we're talking about the `interactive-graph` widget.

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.tsx
```diff
@@ -20,10 +21,16 @@ type Props = {
 const hitboxSizePx = 48;
 
 export const StyledMovablePoint = (props: Props) => {
-    const {range, snapStep, markings} = useGraphConfig();
+    const {range, snapStep, markings, showTooltips} = useGraphConfig();
     const hitboxRef = useRef<SVGCircleElement>(null);
     const {point, onMove, color = WBColor.blue} = props;
```
jeremywiebe, line 26: Note that as far as I know, there are no plans to allow content authors to adjust the color of any of the movables. That is something that is strictly reserved for Locked Features. 

So if it simplifies things, I think we can remove the ability to change the color of this StyledMovablePoint.

jeremywiebe (approved):
Overall I think this looks really good. I think the ability to color the movable point predates this PR so you can leave that for now, but I do know that currently Graphie doesn't allow you to change the color of the movables and so I think we want to do the same (ie. default to WB Blue and not allow customization).

nishasy (approved):
You can ignore my suggestion about the width, since the current implementation matches the previous tooltip behavior. (Also just adding when we talked about online for the paper trail: it doesn't look that much better with a fixed/min width anyway, as the numbers are still kind of jumping back and forth.)

mark-fitzgerald:
> Very exciting! Could we consider giving the tooltip a fixed width if it's possible? To stop the size change from being so jarring while moving the point with the half steps adding more digits?

After looking into setting a minimum width for the tooltip bubble, the jarring aspect that occurs when moving is still there due to the snap-to-grid nature of the movement. Keeping the width consistent didn't produce the "smoothness" that we were hoping for. Leaving it without for now.

PR #1193, opened 2024-04-19
Add axis labels to Mafs graphs

Myranae:
## Summary:

Legacy graphs have labels on the graph axes. This adds that feature to Mafs graphs as well.

Examples - Left is legacy, right is Mafs: 
(From the flipbook)

image

image

From localhost storybook:
image

Issue: LEMS-1819

## Test plan:
- Confirm labels show up for both axes
- Confirm all other currently available features are working
- Confirm all checks pass


packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
```diff
@@ -26,12 +29,14 @@ function getBaseMafsGraphProps(): MafsGraphProps {
...
         containerSizeClass: "small",
         onChange: () => {},
+        labels: [],
     };
 }
 
 describe("StatefulMafsGraph", () => {
     let userEvent: UserEvent;
     beforeEach(() => {
+        setDependencies(testDependencies);
```
Myranae, line 39: Was getting an error saying dependencies were not set for Perseus here, so followed the instructions in the error to get this set up. Looked up each part to set up the imports.

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -92,6 +93,7 @@ type MafsGraphProps = Props & {
 export const MafsGraph = (props: MafsGraphProps) => {
     const {state, dispatch} = props;
     const [width, height] = props.box;
+    const {labels} = useGraphConfig();
```
Myranae, line 96: I originally had labels in the props, but since labels can be found in the context, I switched to this. Since the jsx returned by this component is what contains the reference to the context though, can I do this?

Myranae: Nope. Looks like that's a no XD

nishasy:
Very cool! I was just doing some quick manual testing in the [interactive graph editor](https://650db21c3f5d1b2f13c02952-rnwpoaywce.chromatic.com/?path=/story/perseus-editor-editorpage--demo), and I think the y label might be clipping down?

| Before | After |
|-------|-------|
| Screenshot 2024-04-22 at 10 30 14 AM | Screenshot 2024-04-22 at 10 30 21 AM |

packages/perseus/src/widgets/interactive-graphs/grid.tsx
```diff
@@ -60,13 +62,31 @@ const axisOptions = (
...
+                    width={width}
+                    height={height}
+                />
+            </clipPath>
+            <g clipPath={"url(#myClip)"}>
+                <Coordinates.Cartesian
+                    xAxis={axisOptions(props, 0)}
+                    yAxis={axisOptions(props, 1)}
+                />
+            </g>
```
benchristel, line 89: Why do we need the `clipPath` here?

Also, I'm worried that `url(#myClip)` won't work if there are multiple graphs with different dimensions onscreen, since the `myClip` ID will not be unique. We might need to `useUniqueId` from `@khanacademy/wonder-blocks-core` to get a unique ID.

packages/perseus/src/widgets/interactive-graphs/axis-labels.tsx
```diff
@@ -0,0 +1,43 @@
...
+    return (
+        <>
+            <span
+                id="x-axis-label"
+                style={{position: "absolute", left: x1, top: y1}}
+            >
+                <TeX>{xAxisLabelText}</TeX>
+            </span>
+            <span
+                id="y-axis-label"
```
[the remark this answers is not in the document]
Myranae: Moved everything to inline styles and out of the css file.

packages/perseus/src/widgets/interactive-graphs/axis-labels.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+export default function AxisLabels(props) {
+    const {range, labels, width, height} = useGraphConfig();
+
+    const yAxisLabelLocation: vec.Vector2 = [0, range[1][1]];
+    const xAxisLabelLocation: vec.Vector2 = [range[0][1], 0];
+
+    const [xAxisLabelText, yAxisLabelText] = labels;
+    const graphInfo = {
+        range: range,
```
jeremywiebe, line 17:
```suggestion
        range,
```

packages/perseus/src/widgets/interactive-graphs/graphs/use-transform.ts
```diff
@@ -1,10 +1,36 @@
...
+const matrixBuilder = vec.matrixBuilder;
+
+export type GraphDimensions = {
+    range: [Interval, Interval];
+    width: number; // pixels
+    height: number; // pixels
+};
+
+export function vectorToPixel(
+    points: vec.Vector2[],
```
jeremywiebe, line 14: Related to Ben's comment below.  Maybe we name this `vectors` to enforce the point that we're dealing with vectors (distances) and not x,y locations on the graph?

packages/perseus/src/widgets/interactive-graphs/graphs/use-transform.ts
```diff
@@ -1,10 +1,36 @@
...
+    return points.map((p) => vec.transform(p, transformToPx));
+}
+
+export function pointToPixel(point: vec.Vector2, graphState: GraphDimensions) {
+    const [[xMin], [, yMax]] = graphState.range;
+    const [a, b] = [-xMin, -yMax];
+    return vectorToPixel([point], graphState, [a, b]);
+}
+
+export const useTransformVectorToPixel = (...points: vec.Vector2[]) => {
```
jeremywiebe, line 33: same comment about usage of `points` as a parameter name here. 
```suggestion
export const useTransformVectorToPixel = (...vectors: vec.Vector2[]) => {
```

packages/perseus/src/widgets/interactive-graphs/graphs/components/arrowhead.tsx
```diff
@@ -21,7 +22,13 @@ const arrowPath = pathBuilder()
...
 
 export function Arrowhead(props: Props) {
-    const [point] = useTransform([props.x, props.y]);
+    const {range, width, height} = useGraphConfig();
+    const arrowContext: GraphDimensions = {
+        range: range,
+        width: width,
+        height: height,
+    };
+    const [point] = vectorToPixel([props.tip], arrowContext);
```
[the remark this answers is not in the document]
Myranae: Updated it to use `useTransformVectorToPixel`. Arrowheads aren't showing up right now in general though, so can't confirm it's working quite yet.

packages/perseus/src/widgets/interactive-graphs/axis-labels.tsx
Myranae: Thank you!

nishasy:
Cool! I like that it's not clipping down anymore, but I think it might be too high now? Are you seeing the same thing [locally](http://localhost:6006/?path=/story/perseus-editor-widgets-interactive-graph-editor--with-mafs)?

image

packages/perseus/src/widgets/__tests__/__snapshots__/interactive-graph.test.ts.snap
```diff
@@ -3250,6 +3250,24 @@ exports[`snapshots should render correctly with locked lines 1`] = `
...
+              </span>
+              <span
+                style="position: absolute; left: 200px; top: -24px; font-size: 14px; transform: translate(-50%, -7px);"
+              >
+                <span
+                  class="mock-TeX"
+                >
+                  y
+                </span>
+              </span>
```
Myranae, line 3270: It looks like this addition and the clippath content below are the only actual changes. The other content is still the same, just shifted down to align incorrectly. If there is a way to prevent it from doing that and to only highlight the actual changes, would love to know.

packages/perseus/src/widgets/interactive-graphs/axis-labels.tsx
```diff
@@ -0,0 +1,58 @@
...
+        width,
+        height,
+    };
+    const [x1, y1] = vec.add(
+        pointToPixel(xAxisLabelLocation, graphInfo)[0],
+        [0, 0],
+    );
+    const [x2, y2] = vec.add(
+        pointToPixel(yAxisLabelLocation, graphInfo)[0],
+        [0, -24],
```
mark-fitzgerald, line 27: It might be helpful to leave a comment here that explains what the -24 accomplishes.

packages/perseus/src/widgets/interactive-graphs/axis-labels.tsx
```diff
@@ -0,0 +1,58 @@
...
+    return (
+        <>
+            <span
+                style={{
+                    position: "absolute",
+                    left: x1,
+                    top: y1,
+                    fontSize: "14px",
+                    transform: "translate(7px, -50%)",
+                }}
```
mark-fitzgerald, line 41: I'm thinking that we don't need to calculate a position to place these labels. We should be able to do this entirely in CSS.

For the "x" label, the CSS would look like:
```
.x-axis-label {
    position: absolute;
    left: 100%;
    top: 50%;
    font-size: 14px;
    transform: translate(7px, -50%); /* assumes we want a fixed 7px spacing */
}
```
and the "y" label would look like:
```
.y-axis-label {
    position: absolute;
    left: 50%;
    font-size: 14px;
    transform: translate(-50%, calc(-100% - 7px));
}
```

If we decide to change the font size in the future, this will auto-adapt to the new size of the element. Another benefit is that we don't have to execute the two calls to `pointToPixel()`.

Myranae: Hmm, I think Ben mentioned not having both inline styles and CSS, so I moved everything to be inline. I suppose I could move everything to be on the css file instead. But doesn't this method require the axes to be like a plus sign? Sometimes the axes will be more of an "L" shape on the page, which I don't think would work with what you have here. I could be wrong though.

mark-fitzgerald: I just realized that the CSS that I provided only works for centered labels. If we need to place the labels towards one edge of the graph, this won't work. Meaning that the call to `pointToPixel()` would still need to be made. While my preference is still to place the styling in CSS classes, and then adjust the position (if needed) by setting a CSS variable (via the `style` attribute), I think the main premise for my original recommendation has lost much of its allure. So, the current method is fine.

packages/perseus/src/widgets/interactive-graphs/grid.tsx
Myranae: Ooo, good point! I'll update the ID.

The clip path element defines the shape to be used as the clipping shape. Here the clippath surrounds the rect element, which defines the shape of the clipping. Everything outside of the clip shape gets hidden and only content inside the clip shape is shown. So the clipPath element is the definition and the clipPath attribute is there to say, "clip all children in this element using the shape defined in the id listed in the url call. Does that make sense?

packages/perseus/src/widgets/interactive-graphs/axis-labels.tsx
Myranae: Thank you for pointing this out though! Your recommendations are getting me thinking more about using CSS and trying to use a CSS file more. Personally, I prefer seeing everything in a CSS file, but I also prefer not using classes unless there are multiple instances of the class. I think that's not the preference generally though, so I defaulted to using the inline styling. Definitely appreciate your CSS reminders though :)

packages/perseus/src/widgets/interactive-graphs/graphs/use-transform.ts
Myranae: Did some additional refactoring in this file to get plural/singular syntax into alignment. Thanks for pointing this out in so many places!

packages/perseus/src/widgets/interactive-graphs/graphs/use-transform.ts
```diff
@@ -1,10 +1,39 @@
...
+}
+
+export function pointToPixel(
+    points: vec.Vector2[],
+    graphState: GraphDimensions,
+) {
+    const [[xMin], [, yMax]] = graphState.range;
+    const [a, b] = [-xMin, -yMax];
+    return vectorToPixel(points, graphState, [a, b]);
+}
```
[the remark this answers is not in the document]
Myranae: Added some comments to both vectorsToPixels and pointToPixel. I used the wording you provided; thank you so much! I don't think I actually understood this difference, so I appreciate you clarifying. Please let me know if you think useTransformVectorsToPixels also needs a comment. I'm assuming the other two needed comments because they were harder to differentiate though.

packages/perseus/src/widgets/interactive-graphs/grid.tsx
```diff
@@ -60,13 +62,31 @@ const axisOptions = (
 };
 
 export const Grid = (props: GridProps) => {
+    const {width, height} = useGraphConfig();
+    const xRange = useTransformVectorToPixel(props.range[0]);
+    const yRange = useTransformVectorToPixel(props.range[1]);
```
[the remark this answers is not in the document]
Myranae: Hmm, when I look at the definition for GraphRange, it looks like a 2D array ` [x: [min: number, max: number], y: [min: number, max: number]`. So then grabbing the first index grabs the x range and grabbing the second index grabs the y range. Is that not correct? I think you can find the GraphRange definition in perseus-types.ts on line 483.

packages/perseus/src/widgets/interactive-graphs/axis-labels.tsx
Myranae: Thank you for pointing this out! Added a comment to explain. Originally this was a +1 on a coordinate point, so that might have been more clear. Please let me know if this explanation here is not clear though

packages/perseus/src/widgets/interactive-graphs/grid.tsx
Myranae: Not sure if this should be "WithMock" or "WithoutMock." Just went with "Without" but please let me know if I need to use "With." Thank you again for pointing this out :)

benchristel: > So the clipPath element is the definition and the clipPath attribute is there to say, "clip all children in this element using the shape defined in the id listed in the url call. Does that make sense?

I thought the Mafs SVG element already clipped children placed outside the bounds of the SVG, but maybe I'm missing something. Do you have an example of a problem that happens if the `<clipPath>` element is removed?

Oh, I see, you've set `overflow: visible` in the CSS to override Mafs' default clipping behavior. So we need this clipPath to restore it for the elements that should get clipped.

Could we instead render the axis labels outside the `Mafs` element to prevent them from being clipped? If we did that, I think we wouldn't need `<clipPath>` or `overflow: visible`.

mark-fitzgerald: We could, but I suspect that would complicate things. Wouldn't we need to maintain two separate layers? I suppose we could investigate rendering items outside of the `Mafs` element and see if that simplifies things.

One of the benefits of keeping things inside the `Mafs` element is that we have reference to the boundary of the graph for styling/positioning purposes. With the `<clipPath>` defined, we can apply that path to any group within the graph. Since we currently just need that clip path for everything _except_ for axis labels and marking labels, perhaps we can have a top-level `<g>` for "clipped" items, and another for "un-clipped". Then if we have need for something else that is outside of the clipping, we can simply add it to the "un-clipped" group.

Myranae: Actually, looks like I was able to remove the clippath! Ben and I switched to doing the labels outside of the mafs-graph, so we didn't need the clippath anymore. I'm not sure how this will affect labels though, so maybe adding clipping on a more global level needs to be addressed at some point. But for this work, looks like it is not needed.

packages/perseus/src/widgets/__tests__/__snapshots__/interactive-graph.test.ts.snap
Myranae: Changed the code so the snapshot doesn't have extra changes due to alignment

packages/perseus/src/widgets/__tests__/__snapshots__/interactive-graph.test.ts.snap
```diff
@@ -3250,6 +3250,24 @@ exports[`snapshots should render correctly with locked lines 1`] = `
...
+                  class="mock-TeX"
+                >
+                  x
+                </span>
+              </span>
+              <span
+                style="position: absolute; left: 200px; top: -24px; font-size: 14px; transform: translate(-50%, -7px);"
+              >
+                <span
+                  class="mock-TeX"
```
Myranae, line 3266: Is it a problem that the class is "mock-TeX" here?

Myranae:
> Cool! I like that it's not clipping down anymore, but I think it might be too high now? Are you seeing the same thing [locally](http://localhost:6006/?path=/story/perseus-editor-widgets-interactive-graph-editor--with-mafs)?
> 
> image

@nishasy 
I can make it so that it doesn't look too high in the editor:
image

But then it looks too low in the flipbook:
image

packages/perseus/src/widgets/__tests__/__snapshots__/interactive-graph.test.ts.snap
benchristel: No, that's fine. The TeX component being used here is a lightweight version just for testing. Webapp will inject its own TeX component.

packages/perseus/src/widgets/interactive-graphs/axis-labels.tsx
```diff
@@ -0,0 +1,61 @@
...
+    const [xAxisLabelText, yAxisLabelText] = labels;
+    const graphInfo = {
+        range,
+        width,
+        height,
+    };
+    const [x1, y1] = vec.add(
+        pointToPixel(xAxisLabelLocation, graphInfo),
+        [0, 0],
+    );
```
benchristel, line 24: Not a big deal, but... this `vec.add()` call has no effect, since it's adding `[0, 0]`.

benchristel (approved):
Looks great! 🎉

mark-fitzgerald (approved):
Looks good.

nishasy (approved):
🎉 🎉 🎉 🎉 🎉 
🚀 🚀 🚀 🚀 🚀

PR #1194, opened 2024-04-19
Centralize where we control feature flags in perseus-editor for Storybook

jeremywiebe:
## Summary:

This PR enables the new `interactive-graph-locked-features-m1` feature flag for all `interactive-graph` editor stories (so that you can actually see the locked features as we build them). 

I also centralized our feature flags for `Perseus-editor` stories so it'll be just a bit easier as we start doing more graph types and M2 locked features. 

Issue: "none"

## Test plan:

`yarn start`
Navigate to the [With Mafs](http://localhost:6006/?path=/story/perseus-editor-widgets-interactive-graph-editor--with-mafs) sory
Note that the locked figures editing tools appear below the graph

It should also be available in the [With Locked Points](http://localhost:6006/?path=/story/perseus-editor-widgets-interactive-graph-editor--with-locked-points) story.

### DIFF
--- a/.changeset/twelve-llamas-tap.md
+++ b/.changeset/twelve-llamas-tap.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": patch
+---
+
+Enable M1 locked features in Perseus Editor storybook stories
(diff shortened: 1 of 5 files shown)


nishasy (approved):
LGTM, but don't forget to remove unused imports before landing 🙂

PR #1196, opened 2024-04-22
Rename ColorCircle to ColorSwatch

nishasy:
## Summary:
Rename ColorCircle to ColorSwatch so that the term "circle"
isn't confusd with the circle graph or other circle terms.

Issue: none

## Test plan:
`yarn jest`


benchristel (approved):
LGTM!

I don't have all the context here so feel free to ignore this suggestion, but... as long as we're renaming this concept, would `ColorSwatch` be an appropriate name?

nishasy:
@benchristel I like ColorSwatch better! I'll go ahead and change it.

PR #1198, opened 2024-04-22
Add Storybook stories for Mafs visual regression testing

benchristel:
In a future PR, we'll configure Storybook to run snapshot tests on these
stories, which should help us catch visual regressions in Mafs graphs.

Issue: https://khanacademy.atlassian.net/browse/LEMS-1926

## Test plan:

- `yarn storybook`
- `open http://localhost:6006/?path=/story/perseus-widgets-interactive-graph--mafs-with-custom-axis-labels`

### DIFF
--- a/.changeset/chilly-shoes-sniff.md
+++ b/.changeset/chilly-shoes-sniff.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Internal: Add Storybook stories for visual regression testing of Mafs interactive graphs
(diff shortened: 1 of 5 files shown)


jeremywiebe (approved):
This all looks great.  I like the builder pattern in use here.

One thought I had (which might be premature) is to have these stories that we plan to use for visual regression testing organized into their own file or demarcated some way.  Not sure that's necessary, but it'd help us remember that we want to keep them around even if they appear less useful for direct development.

PR #1200, opened 2024-04-22
Radio Editor - Use a percent based width calculation to avoid horizontal scrolling when entering radio widget options.

jeanettehead:
Issue: LEMS-1954

## Test plan:
* Open a radio widget in the editor
* Observe that when there is lots of text in an option, there is no horizontal scroll
* Observe that when there is very little text in an option, it still occupies the width of its container


mark-fitzgerald (approved):
Gotta love simple.

PR #1201, opened 2024-04-22
Default radio to 4 options in the editor

jeanettehead:
## Summary:
From a content team member:
"Hi all! Just a question I have been wondering about: Why does the multiple choice widget default to two options? Don't all Khan Academy MC questions have at least four options? Is there a setting I can change somewhere to make it default to four?"

Issue: LEMS-1808

## Test plan:
* Open the content editor
* Add a radio widget
* Observe that there are four empty options instead of 2


mark-fitzgerald (approved):
Easy-peasy.

PR #1202, opened 2024-04-22
Update snapshots

nishasy:
## Summary:
Looks like some snapshots on main are failing.
Updating them here.

Issue: none

## Test plan:
`yarn jest`


.changeset/four-llamas-double.md
nishasy, line 1: The empty changeset should have worked since the only update was to a test/snap, but it was complaining so 🤷🏽‍♀️

jeremywiebe (approved):

PR #1205, opened 2024-04-23
Configure Chromatic to run visual regression tests for graphs

benchristel:
## Summary:
Context: Chromatic.com is a web service that does visual regression
testing of user interfaces, by taking screenshots of Storybook stories
and diffing them across commits. This commit sets up Perseus to use
our existing Chromatic account for visual regression testing of our
interactive graph widget.

Previously, we were only running visual snapshot tests on Chromatic when
a storybook file changed. This doesn't really help us catch regressions.

We can't run all our stories through Chromatic on every commit because
we'd hit our limit of 35,000 snapshots per month. So this commit tells
Chromatic to only take snapshots of interactive graph stories designed
for regression testing.

Issue: LEMS-1926

Test plan:

Watch CI run on this PR. I'd expect to see a link to Chromatic in the
"checks" section of the PR on GitHub, titled "UI Tests". Clicking the
link should take you to a Chromatic.com page for the build, which should
say that 17 "tests" ran.


nishasy (commented):
Exciting stuff! Glad we can limit the usage so we don't go over the limit.

It looks good to me, but I'll let Jeremy or someone else do the final approval since I'm not very well-versed in this stuff.

.github/workflows/storybook.yml
```diff
@@ -44,6 +43,4 @@ jobs:
                   token: ${{ secrets.GITHUB_TOKEN }}
                   projectToken: ${{ secrets.CHROMATIC_APP_CODE }}
                   autoAcceptChanges: "main"
-                  # Generate snapshots for only changed stories
-                  # See: https://www.chromatic.com/docs/turbosnap
-                  onlyChanged: true
+                  onlyStoryFiles: "**/*regression.stories.tsx"
```
mark-fitzgerald, line 46: Love it!

mark-fitzgerald (approved):
Looks good!

.github/workflows/storybook.yml
```diff
@@ -43,4 +43,4 @@ jobs:
                   token: ${{ secrets.GITHUB_TOKEN }}
                   projectToken: ${{ secrets.CHROMATIC_APP_CODE }}
                   autoAcceptChanges: "main"
-                  onlyStoryFiles: "**/*regression.stories.tsx"
```
jeremywiebe: 😞  https://github.com/chromaui/chromatic-cli/issues/721

benchristel: Yeah, I can't seem to get this to work. The `chromaui` GitHub action logs something like "stories will be limited to these files" indicating that it understands the option, but then all the stories get uploaded anyway. And then no tests run.

Somehow, WB is getting snapshot testing to work. And setting it up for the mathjax-renderer was pretty painless. I wonder if maybe the fact that we're running Storybook through Vite is getting in our way?

jeremywiebe (approved):

.storybook/preview.tsx
```diff
@@ -58,8 +58,6 @@ const preview: Preview = {
                 value,
             })),
         },
-        // Disables Chromatic's snapshotting on a global level
-        chromatic: {disableSnapshot: true},
```
jeremywiebe, line 62: Urg, I didn't even think about just searching for "chromatic" in the repo. 😞

jeremywiebe (approved):

PR #1206, opened 2024-04-23
[🔥AUDIT🔥] Fix `yarn dev`

jeremywiebe:
🖍 _This is an audit!_ 🖍

## Summary:

As part of the move away from `@khanacademy/wonder-blocks-i18n` we added some Storybook config to check if we're in a dev/storybook env and just use hard-coded (mock) strings. 

This config was applied in `.storybook/main.ts` but not to our custom dev server (in `/dev`). 

This PR moves the Vite config into `dev/vite.config.ts` so it works for both (we "inherit" the dev vite config in Storybook - maybe that's confusing and we should move it)

Issue: "none"

## Test plan:

`yarn dev` - dev UI works

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -41,11 +41,6 @@ const config: StorybookConfig = {
     viteFinal: async (config, {configType}) => {
         return mergeConfig(config, {
             ...viteConfig,
-            define: {
-                // This is used to determine if we are running in a
-                // Storybook environment.
-                "process.env.STORYBOOK": JSON.stringify(true),
-            },
             build: {
                 // Vite 5 has a bug with how it builds `url(data: )` urls when
                 // it inlines SVGs. Given this is mostly used for static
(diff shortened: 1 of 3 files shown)


.storybook/main.ts
```diff
@@ -41,11 +41,6 @@ const config: StorybookConfig = {
     viteFinal: async (config, {configType}) => {
         return mergeConfig(config, {
             ...viteConfig,
```
jeremywiebe, line 43: This `viteConfig` comes from `dev/vite.config.ts`.  Maybe it should move to the repo root. 🤔

benchristel (approved):

dev/vite.config.ts
```diff
@@ -15,6 +15,11 @@ glob.sync(resolve(__dirname, "../packages/*/package.json")).forEach(
 );
 
 export default defineConfig({
+    define: {
```
jeresig, line 18: Ahh - thank you!

PR #1208, opened 2024-04-24
Implement locked rays in Interactive Graph

nishasy:
## Summary:
Locked rays were not implemented along with lines. But here they are!

I used the movable ray for inspiration for the implementation. This is
using the "in-house" Vector component rather than than Mafs Vector.

Note that the ray arrow cannot be hidden/shown. This will likely be
done as part of LEMS-1929.

Issue: https://khanacademy.atlassian.net/browse/LEMS-1928

## Test plan:

Relevant storybook stories
http://localhost:6006/?path=/story/perseus-editor-widgets-interactive-graph-editor--with-locked-lines
http://localhost:6006/?path=/story/perseus-widgets-interactive-graph--segment-with-mafs-and-locked-lines
http://localhost:6006/?path=/story/perseus-widgets-interactive-graph--all-locked-rays
http://localhost:6006/?path=/docs/perseus-editor-editorpage--docs

## Storybook screenshots

| Interactive Graph | Interactive Graph Editor | Editor Page |
|---|---|---|
| Screenshot 2024-04-23 at 6 19 46 PM | Screenshot 2024-04-23 at 6 23 22 PM | Screenshot 2024-04-23 at 6 23 50 PM |


packages/perseus/src/widgets/__tests__/__snapshots__/interactive-graph.test.ts.snap
```diff
@@ -4178,6 +4178,42 @@ exports[`snapshots should render correctly with locked lines 1`] = `
...
                     />
                   </g>
+                  <g
+                    class="locked-line"
+                  >
+                    <g
+                      style="stroke: #B25071; stroke-width: 2;"
+                    >
+                      <line
+                        class="undefined"
```
mark-fitzgerald, line 4188: Is `undefined` expected?

packages/perseus/src/widgets/interactive-graphs/graphs/components/svg-line.tsx
```diff
@@ -20,7 +20,7 @@ export function SVGLine(props: Props) {
             x2={end[0]}
             y2={end[1]}
             style={style}
-            className={`${className}`}
```
nishasy, line 23: This was making svg `<line>` have `class="undefined"`. Now it should have no class name if one isn't passed in.

packages/perseus/src/widgets/__tests__/__snapshots__/interactive-graph.test.ts.snap
nishasy: Good catch! Fixed.

packages/perseus/src/widgets/interactive-graphs/graph-locked-layer.tsx
```diff
@@ -4,14 +4,16 @@ import * as React from "react";
 import LockedLine from "./locked-line";
 import LockedPoint from "./locked-point";
 
-import type {LockedFigure} from "../../perseus-types";
+import type {LockedFigure, Range} from "../../perseus-types";
 
 type Props = {
     lockedFigures: ReadonlyArray<LockedFigure>;
+    range: [Range, Range];
```
benchristel, line 11: minor: this type is a bit confusing. "The range is a pair of `Range`s". I might prefer to use the `Interval` type from Mafs, and label the elements of the pair:

```ts
import type {Interval} from "mafs";
// ...
type Props = {
  // ...
  range: [x: Interval, y: Interval];
}
```

packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
```diff
@@ -360,11 +360,16 @@ describe("locked layer", () => {
         // Act
         // eslint-disable-next-line testing-library/no-container, testing-library/no-node-access
         const lines = container.querySelectorAll(".locked-line line");
+        // eslint-disable-next-line testing-library/no-container, testing-library/no-node-access
+        const ray = container.querySelector(".locked-line g");
```
benchristel, line 364: This selector seems brittle - could we use a different class for locked rays? E.g. `.locked-ray`?

packages/perseus/src/widgets/interactive-graphs/graph-locked-layer.tsx
```diff
@@ -22,6 +24,7 @@ const GraphLockedLayer = (props: Props) => {
                         break;
                     case "line":
                         Figure = LockedLine;
+                        otherProps = {range: props.range};
```
benchristel, line 27: Could we move the declaration of `otherProps` inside the `map` callback, next to Figure?

```ts
{lockedFigures.map((figure, index) => {
    let Figure;
    let otherProps;
```

The way it is now, the `otherProps` will persist between calls to the `map` callback. This means that if the first element in `lockedFigures` is a line, and the second is a point, the `range` prop will get mistakenly added to the point. Probably not a huge deal, but could create headaches down the line.

I'm also kind of surprised TypeScript accepts this, and I wonder if it's just throwing up its hands and implicitly typing `otherProps` as `any` or something.

An alternative would be to render the figure inside the `switch`:

```tsx
{lockedFigures.map((figure, index) => {
    switch (figure.type) {
        case "point":
            return <LockedPoint key={`point-${index}`} {...figure} />;
            break;
        case "line":
            return <LockedLine key={`line-${index}`} range={props.range} {...figure} />;
            break;
        default:
            /**
             * Devlopment-time future-proofing: This should
             * fail during type-checking if we add a new locked
             * shape type and forget to handle it in any other
             * switch case here.
             */
            throw new UnreachableCaseError(figure);
    }
})}
```

jeremywiebe (commented):
This is so exciting to see these locked features coming through!  

image

@benchristel and @mark-fitzgerald did you have any reservations that you saw were blocking approving this PR once your comments were resolved?

mark-fitzgerald (approved):
Looks good. I like Ben's points, especially the one about the `otherProps` in `graph-locked-layer.tsx`.

packages/perseus/src/widgets/interactive-graphs/graph-locked-layer.tsx
nishasy: oh yeah rendering directly inside the `swtich` makes way more sense. The current implementation is left over from how it was done pre-refactor where it was super long awkward to put in a swtich. Thanks!

packages/perseus/src/widgets/__testdata__/interactive-graph.testdata.ts
```diff
@@ -1639,4 +1662,118 @@ export const segmentWithAllLockedLineVariations: PerseusRenderer = {
...
+                    {
+                        type: "line",
+                        kind: "ray",
+                        points: [
+                            {
+                                type: "point",
+                                coord: [-7, -5],
+                                color: "green",
+                                filled: true,
+                            },
```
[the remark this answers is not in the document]
nishasy: That's a good idea. I'm going to look into it in a new PR.

packages/perseus/src/widgets/__stories__/interactive-graph.stories.tsx
```diff
@@ -121,6 +122,19 @@ export const AllLockedLines = (args: StoryArgs): React.ReactElement => (
...
 );
 
+export const AllLockedRays = (args: StoryArgs): React.ReactElement => (
+    <RendererWithDebugUI
+        apiOptions={{
+            flags: {
+                mafs: {
+                    segment: true,
+                },
+            },
```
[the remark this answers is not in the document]
nishasy: (talked offline)
Since this is in a different package, I'll hold off on doing that. If I keep adding to this file, I'll move the flags so they can be used in perseus though.

packages/perseus/src/widgets/interactive-graphs/locked-line.tsx
```diff
@@ -4,16 +4,54 @@ import * as React from "react";
...
+            <Vector
+                tail={point1.coord}
+                tip={endExtend}
+                color={lockedFigureColors[color]}
+                style={{
+                    strokeDasharray:
+                        lineStyle === "dashed"
+                            ? // TODO(lems-1930): Uncomment this line when the
+                              // dashed style is updated in Mafs.
+                              // ? "var(--mafs-line-stroke-dash-style)"
```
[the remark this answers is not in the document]
nishasy: I don't think our in-house Vector (build offs of SVGLine) has solid/dashed style option? I'm not sure it would work the same way?

benchristel (approved):
LGTM!

PR #1210, opened 2024-04-24
Limit colorset for locked figures

nishasy:
## Summary:
Limit locked figures colorset as per this slack conversation.

Issue: https://khanacademy.atlassian.net/browse/LEMS-1931

## Test plan:

Storybook pages
http://localhost:6006/?path=/docs/perseus-editor-widgets-interactive-graph-editor--docs
http://localhost:6006/?path=/docs/perseus-widgets-interactive-graph--docs

Screenshot 2024-04-24 at 10 38 53 AM


packages/perseus-editor/src/components/__stories__/color-select.stories.tsx
```diff
@@ -1,8 +1,9 @@
+import {type LockedFigureColor} from "@khanacademy/perseus";
```
handeyeco, line 1:
```suggestion
import {LockedFigureColor} from "@khanacademy/perseus";
```
Does it work like this? Do we still need `type`?

.changeset/gorgeous-phones-float.md
```diff
@@ -0,0 +1,6 @@
+---
+"@khanacademy/perseus": minor
+"@khanacademy/perseus-editor": minor
+---
+
+Limit colorest for locked figures in Interactive Graph
```
jeremywiebe, line 6: ??

```suggestion
Limit color set for locked figures in Interactive Graph
```

packages/perseus-editor/src/components/__stories__/color-select.stories.tsx
jeremywiebe, line 1: Can we move this back down to line 5 and revert it to 

```suggestion 
import type {LockedFigureColor} from "@khanacademy/perseus";
```

packages/perseus/src/perseus-types.ts
```diff
@@ -658,30 +658,20 @@ export type PerseusInteractiveGraphWidgetOptions = {
 // TODO: If/when these colors are added to Wonder Blocks, we should remove
 // them from here and use Wonder Blocks everywhere else instead.
```
jeremywiebe, line 659: Is this comment still something we plan/need to do?

jeremywiebe (approved):
That makes sense to me.

packages/perseus-editor/src/components/locked-line-settings.tsx
```diff
@@ -36,7 +36,7 @@ const LockedLineSettings = (props: Props) => {
     const {
         kind,
         points,
-        color: lineColor = "blue",
+        color: lineColor,
```
handeyeco, line 39: Not really related to your change, but where is `lineColor` coming from? I'm like super confused. VS Code understands what it is, but I don't see where it's declared.

EDIT: nm I'm dumb. You're renaming `color` to `lineColor`. I guess that makes sense, but since it's so close to how TS assigns a type I wonder if it'd be better to be more verbose in how you're renaming destructured properties.

packages/perseus-editor/src/components/util.ts
```diff
@@ -52,6 +52,8 @@ export function getValidNumberFromString(value: string) {
     return isNaN(parsed) ? 0 : parsed;
 }
 
+const DEFAULT_COLOR = "green";
```
handeyeco, line 55: It's the default color for everything?

handeyeco (approved):
Seems good to me!

packages/perseus-editor/src/components/__stories__/color-select.stories.tsx
nishasy: Yeah, it looks like I do need `type`

packages/perseus-editor/src/components/locked-line-settings.tsx
nishasy: What would you suggest?

Maybe something like
```
// Renaming because `color` can be too ambiguous.
const lineColor = props.color;
```
?

packages/perseus-editor/src/components/util.ts
nishasy: Hmm... good question. I'll double check.

nishasy:
I'm going to land this for now, and if the default color or anything else needs to change, I'll do it in a new PR.

PR #1212, opened 2024-04-24
Bugfix: changing graph markings resetting number of segments

handeyeco:
## Summary:
When `hasBeenInteractedWith` was false and `getGradableGraph` was called, `getGradableGraph` would strip everything but `type` off of `initialGraph`. When this happened in the editor, it meant that certain properties were getting reset to their defaults.

https://github.com/Khan/perseus/assets/16308368/6bbba3de-7dae-4e01-bed6-c586747a3b15

Issue: [LEMS-1935](https://khanacademy.atlassian.net/browse/LEMS-1935)

## Test plan:
- Go to the `Interactive Graph Edior > With Mafs` story
- Change type to "Line Segment(s)"
- Change "Number of segments" to 4
- Change "Marking" from "Graph" to "Grid"
- Note "Number of segments" stays on 4 (before it would reset to 1)

[LEMS-1935]: https://khanacademy.atlassian.net/browse/LEMS-1935?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/math-input/src/components/keypad/__tests__/keypad-v2-mathquill.test.tsx
+++ b/packages/math-input/src/components/keypad/__tests__/keypad-v2-mathquill.test.tsx
@@ -228,7 +228,7 @@ describe("Keypad v2 with MathQuill", () => {
         );
     });
 
-    it.skip("deletes from the input using the backspace button", async () => {
+    it("deletes from the input using the backspace button", async () => {
         // Arrange
         const mockMathInputCallback = jest.fn();
         render(
@@ -275,7 +275,7 @@ describe("Keypad v2 with MathQuill", () => {
         }
 
         expect(mockMathInputCallback).toHaveBeenLastCalledWith("");
-    });
+    }, 10000);
 
     // Keypad event tests
     it("fires the keypad open event on open", async () => {
--- a/packages/math-input/src/components/keypad/__tests__/keypad-v2-mathquill.test.tsx
+++ b/packages/math-input/src/components/keypad/__tests__/keypad-v2-mathquill.test.tsx
@@ -228,7 +228,7 @@ describe("Keypad v2 with MathQuill", () => {
         );
     });
 
-    it("deletes from the input using the backspace button", async () => {
+    it.skip("deletes from the input using the backspace button", async () => {
         // Arrange
         const mockMathInputCallback = jest.fn();
         render(
--- a/packages/perseus-editor/src/components/widget-editor.tsx
+++ b/packages/perseus-editor/src/components/widget-editor.tsx
@@ -11,7 +11,7 @@ import _ from "underscore";
 import SectionControlButton from "./section-control-button";
 
 import type Editor from "../editor";
-import type {Alignment, PerseusWidget} from "@khanacademy/perseus";
+import type {APIOptions, Alignment, PerseusWidget} from "@khanacademy/perseus";
 
 const {InlineIcon} = components;
 const {iconChevronDown, iconChevronRight, iconTrash} = icons;
@@ -25,7 +25,7 @@ type WidgetEditorProps = {
         silent?: boolean,
     ) => unknown;
     onRemove: () => unknown;
-    apiOptions: any;
+    apiOptions: APIOptions;
 } & Omit<PerseusWidget, "key">;
 
 type WidgetEditorState = {
--- a/packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
+++ b/packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
@@ -154,6 +154,17 @@ describe("mafs graphs", () => {
                 renderQuestion(question, apiOptions);
             });
 
+            it("should reject when has not been interacted with", () => {
+                // Arrange
+                const {renderer} = renderQuestion(question, apiOptions);
+
+                // Act
+                // no action
+
+                // Assert
+                expect(renderer).toHaveInvalidInput();
+            });
+
             it("rejects incorrect answer", async () => {
                 // Arrange
                 const {renderer, container} = renderQuestion(
--- a/packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
+++ b/packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
@@ -154,17 +154,6 @@ describe("mafs graphs", () => {
                 renderQuestion(question, apiOptions);
             });
 
-            it("should reject when has not been interacted with", () => {
-                // Arrange
-                const {renderer} = renderQuestion(question, apiOptions);
-
-                // Act
-                // no action
-
-                // Assert
-                expect(renderer).toHaveInvalidInput();
-            });
-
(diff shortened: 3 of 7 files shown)


packages/math-input/src/components/keypad/__tests__/keypad-v2-mathquill.test.tsx
```diff
@@ -228,7 +228,7 @@ describe("Keypad v2 with MathQuill", () => {
         );
     });
 
-    it("deletes from the input using the backspace button", async () => {
+    it.skip("deletes from the input using the backspace button", async () => {
```
handeyeco, line 231: This just seems broken. It keeps timing out.

packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
```diff
@@ -154,17 +154,6 @@ describe("mafs graphs", () => {
                 renderQuestion(question, apiOptions);
             });
 
-            it("should reject when has not been interacted with", () => {
```
handeyeco, line 157: Warning to reviewers: this test broke with my change and I'm worried I might be introducing a regression.

However when I run this test with Mafs turned off, the test also fails. My change makes the external Mafs API more consistent with the legacy API.

At its core, the root of the problem is that the legacy API passes back more than just the type of the initial graph; it also includes things like `numSegments`. The fact that we aren't passing that back up means we're overwriting our settings with an object that only has a `type` property.

So I _think_ my changes are correct and that this test is incorrect. I'd appreciate thoughts though.

packages/math-input/src/components/keypad/__tests__/keypad-v2-mathquill.test.tsx
```diff
@@ -275,7 +275,7 @@ describe("Keypad v2 with MathQuill", () => {
         }
 
         expect(mockMathInputCallback).toHaveBeenLastCalledWith("");
-    });
+    }, 10000);
```
handeyeco, line 278: It intermittently fails on me with a timeout error, so I'm trying this.

packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
benchristel: I think we should keep the test, because it covers important behavior. Without this test, I don't think there will be anything to ensure that we mark unanswered questions as unanswered rather than incorrect. Maybe there's a bug in `toHaveInvalidInput` that we should fix?

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-state.ts
```diff
@@ -161,7 +161,7 @@ export function getGradableGraph(
     initialGraph: PerseusGraphType,
 ): PerseusGraphType {
     if (!state.hasBeenInteractedWith) {
-        return {type: initialGraph.type};
+        return {...initialGraph};
```
benchristel, line 164: I think this change is good. I manually tested it in the dev UI flipbook, and question grading still seems to work. Here is the grading logic that this is coupled to, from [interactive-graph.tsx](https://github.com/Khan/perseus/blob/322a2287bb6c4e3341e948d811769225e3d0949e/packages/perseus/src/widgets/interactive-graph.tsx)

```ts
const hasValue = Boolean(
    // @ts-expect-error - TS2339 - Property 'coords' does not exist on type 'PerseusGraphType'.
    userInput.coords ||
        // @ts-expect-error - TS2339 - Property 'center' does not exist on type 'PerseusGraphType'. | TS2339 - Property 'radius' does not exist on type 'PerseusGraphType'.
        (userInput.center && userInput.radius),
);

// The input wasn't correct, so check if it's a blank input or if it's
// actually just wrong
if (!hasValue || _.isEqual(userInput, rubric.graph)) {
    // We're where we started.
    return {
        type: "invalid",
        message: null,
    };
}
return {
    type: "points",
    earned: 0,
    total: 1,
    message: null,
};
```

Previously, this code was taking the `!hasValue` path to a score of `invalid`. Now, it's taking the `_.isEqual(userInput, rubric.graph)` path.

packages/perseus-editor/src/components/widget-editor.tsx
```diff
@@ -25,7 +25,7 @@ type WidgetEditorProps = {
         silent?: boolean,
     ) => unknown;
     onRemove: () => unknown;
-    apiOptions: any;
+    apiOptions: APIOptions;
```
benchristel, line 28: Thanks for fixing this!

benchristel (commented):
I think the code change is okay, but I don't really understand why it fixes the bug. Why is the editor calling `getUserInput` at all?

I'd also like to repair and keep the test that was failing, if we can.

packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
handeyeco: So only the `point` graph is failing and it's failing for both Mafs and legacy. Console logging things shows that they exhibit the same behavior when getting user input.

I now think there was an existing issue with `point` that was getting hidden by the issue I fixed in this PR. I'll keep investigating.

handeyeco: Okay @benchristel! I did a lot more research (documented in that bug ticket), here's the conclusion I same to:
- At least since before 2024 (before this work kicked off), the legacy interactive graph grades a point graph as correct if:
  - The initial state is correct
  - Even if the user hasn't interacted with the chart
- These tests are testing that circumstance: the test starts in a default state and we try to validate it before interaction
- The tests were passing because of the bug I fixed

I don't think this is the time to fix this issue because:
- It's a really unlikely edge case (why would we start the exercise with the correct answer already set?)
- Changing this part of the logic will also change the legacy behavior, possibly introducing regressions
- After we retire the legacy experience, we could refactor to include `hasBeenInteractedWith` (which is only available to Mafs right now)

If we don't want to remove these tests, I propose we change them so that this particular test is testing data that _doesn't_ begin in the correct state.

I'm going to update this PR and I've thrown time on the calendar for us to talk.

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-state.test.ts
```diff
@@ -193,3 +197,31 @@ describe("initializeGraphState for point graphs", () => {
         },
     );
 });
+
+describe("getGradableGraph", () => {
+    /**
+     * Originally `getGradableGraph` was returning a PerseusGraphType with just a
+     * `type` property, stripping everything else off of `initialGraph`.
+     * This caused the editor to keep resetting certain properties (ie `numSegments`).
+     */
```
benchristel, line 206: Thanks for adding this regression test and the description of why it's needed!

benchristel (approved):
LGTM!

PR #1215, opened 2024-04-24
Change method binding for NumericInputEditor.

jeresig:
## Summary:
Jeremy, you were right! Turns out there are some places that call the methods directly without using .bind(). I'm not sure I can find them all so I just changed all the methods to be bound. I hope that works!

Issue: FEI-5550

## Test plan:
I've confirmed that this page renders correctly after this change:
http://localhost:8090/devadmin/content/exercises/lateral-and-total-surface-area-of-cylinders/x48fc188364f60f35/x18cc8eaa537fc877

### DIFF
--- a/.changeset/dull-lamps-sneeze.md
+++ b/.changeset/dull-lamps-sneeze.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": patch
+---
+
+Fix up a binding issue with NumericInputEditor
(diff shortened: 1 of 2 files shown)


jeresig:
Hey folks - I'm going to land this so that we can get this fix out for the content creators, some of the tools are broken right now!

jeremywiebe (commented):
Perseus still has too many sharp edges. I think making them all arrow functions is a safe approach for now. I'd love to reduce how much we use refs as we continue working on Perseus, but it's a project that is hard to justify. :)

PR #1221, opened 2024-04-25
[Dev UI] Display correctness in WB Banner

jeremywiebe:
## Summary:

I changed the "Check Answer" button result to be displayed in a WB banner so that it doesn't require a second click to dismiss it. Now it just disappears after a short timeout. 

image

Issue: "none"

## Test plan:

`yarn dev`
Paste some questions into the Flipbook 
Click "Check answer"

### DIFF
--- a/.changeset/clean-comics-stare.md
+++ b/.changeset/clean-comics-stare.md
@@ -1,5 +1,0 @@
----
-"@khanacademy/perseus-dev-ui": patch
----
-
-✨ [Dev UI] Display Flipbook correctness in a WB Banner
--- a/.changeset/clean-comics-stare.md
+++ b/.changeset/clean-comics-stare.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-dev-ui": patch
+---
+
+✨ [Dev UI] Display Flipbook correctness in a WB Banner
--- a/dev/flipbook.tsx
+++ b/dev/flipbook.tsx
@@ -208,7 +208,7 @@ function SideBySideQuestionRenderer({
 function GradableRenderer(props: QuestionRendererProps) {
     const {question, apiOptions} = props;
     const rendererRef = useRef<Renderer>(null);
-    const [score, setScore] = useState<PerseusScore | undefined>();
+    const [score, setScore] = useState<PerseusScore>();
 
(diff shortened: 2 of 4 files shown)


dev/flipbook.tsx
```diff
@@ -204,6 +208,9 @@ function SideBySideQuestionRenderer({
 function GradableRenderer(props: QuestionRendererProps) {
     const {question, apiOptions} = props;
     const rendererRef = useRef<Renderer>(null);
+    const [score, setScore] = useState<PerseusScore | undefined>();
```
handeyeco, line 211:
```suggestion
    const [score, setScore] = useState<PerseusScore>();
```
Does the same thing with less code.

.changeset/clean-comics-stare.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/perseus-dev-ui": patch
```
handeyeco, line 2: Do we need a changeset for `\dev` stuff? Or is that a holdover from when it was in `\packages`?

handeyeco (approved):

dev/flipbook.tsx
jeremywiebe: Weird. I thought I tried that and it complained that I wasn't providing a value. It works as you say. Thanks!

.changeset/clean-comics-stare.md
jeremywiebe: I think our tooling will yell if I don't make a changeset for this.  We could disable it, but 🤷

handeyeco: Usually I wouldn't care about extra changesets, but `@khanacademy/perseus-dev-ui` isn't a think right? Like there's no `packages/perseus-dev-ui`. Seems like it'd confuse new contributors.

jeremywiebe: Going to try to mark the `package.json` as `private: true` https://docs.npmjs.com/cli/v10/configuring-npm/package-json#private  

I think that'll stop the tooling from asking for a changeset in this "package".

jeremywiebe: Ok. That worked.  We shouldn't be publishing this package anyways so I think that's the right call.

PR #1223, opened 2024-04-25
Upgrade to Mafs 0.18.5 (includes consistent dash styles for lines)

jeremywiebe:
## Summary:

Nisha noticed that some Mafs line types used different dash styles. We fixed this upstream in Mafs. Upgrading to Mafs v0.18.5 includes this fix.

| Line | Segment |
| --- | --- | 
| image |  image |

Issue: LEMS-1930

## Test plan:

`yarn start`
Navigate to [All Locked Line Segments](http://localhost:6006/?path=/story/perseus-widgets-interactive-graph--all-locked-line-segments)
Navigate to [All Locked Lines](http://localhost:6006/?path=/story/perseus-widgets-interactive-graph--all-locked-lines)

Note that the dash styles are the same.

### DIFF
--- a/.changeset/many-cougars-drop.md
+++ b/.changeset/many-cougars-drop.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Upgrade Mafs to v0.18.5 - includes fix to make dashed lines use a consistent dash style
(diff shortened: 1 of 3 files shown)


nishasy (approved):
Yes!!!!

PR #1225, opened 2024-04-25
Fix Storybook sorting

jeremywiebe:
## Summary:

It was noted that the Storybook navigation panel was not properly nesting stories under the "Perseus" node. This PR fixes this by removing spaces from the "Perseus Editor" stories. It appears that Storybook splits names on spaces and so I suspect that's why the Perseus and Perseus Editor stories were intermingled.

Issue: "none"

https://github.com/Khan/perseus/assets/77138/2e41965c-de81-451d-958d-8b5946201dd8

## Test plan:

### DIFF
--- a/.changeset/dirty-snails-run.md
+++ b/.changeset/dirty-snails-run.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": patch
+---
+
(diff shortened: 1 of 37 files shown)


benchristel (approved):
Thanks! LGTM.

PR #1229, opened 2024-04-26
Show Arrowheads on Locked Lines

mark-fitzgerald:
## Summary:

Issue: LEMS-1929

## Test plan:


packages/perseus/src/widgets/interactive-graphs/graphs/components/vector.tsx
```diff
@@ -15,17 +15,11 @@ type Props = {
...
 export function Vector(props: Props) {
     const {tail, tip, color, style} = props;
-
     const [tailPx, tipPx] = useTransformVectorsToPixels(tail, tip);
-    const direction = vec.sub(tip, tail);
+
     return (
         <g style={{stroke: color, strokeWidth: 2}}>
             <SVGLine start={tailPx} end={tipPx} style={style} />
-            <Arrowhead tip={tip} angle={angleDegrees(direction)} />
```
nishasy, line 24: Does removing this here not remove the arrows in `movable-line.tsx`?

benchristel: I think it will remove those arrowheads. We should have visual regression tests for this!

I have created a ticket for myself to add those tests.

packages/perseus/src/widgets/interactive-graphs/graphs/utils.ts
```diff
@@ -1,5 +1,35 @@
...
+}
+
+/**
+ * Given two points, generate the tip and rotation angle that an arrowhead should use when drawn in the graph.
+ * @param referencePoint - The point of the associated line used for reference to help determine arrowhead location.
+ * @param throughPoint - A point that the associated line passes through. Must be different from referencePoint.
+ * @param box - The box with which to intersect the line, in the form [[xMin, xMax], [yMin, yMax]]
+ */
+export const getArrowheadValues = (
+    referencePoint: vec.Vector2,
```
benchristel, line 17: Possibly relevant: the point where a ray starts is called the "initial point": https://en.wikipedia.org/wiki/Line_(geometry)#Ray

packages/perseus/src/widgets/interactive-graphs/graphs/utils.ts
```diff
@@ -1,5 +1,35 @@
+import {vec} from "mafs";
+
 import type {CollinearTuple} from "../../../perseus-types";
-import type {Interval, vec} from "mafs";
+import type {Interval} from "mafs";
+
+export function calculateAngleInDegrees([x, y]: vec.Vector2) {
```
benchristel, line 6: There's another function that does this, `angleDegrees` in `vector.tsx`. Maybe the definitions could be deduplicated and moved to a common location?

It's too bad that the existing math functions (in `@khanacademy/kmath`) are such a poor fit for our use case that we can't just call them. I think the main problem is that they generalize to 3D (and maybe higher dimensions?) when AFAIK nothing actually needs 3D.

packages/perseus/src/widgets/interactive-graphs/graphs/utils.ts
```diff
@@ -1,5 +1,35 @@
...
+): {
+    tip: vec.Vector2;
+    angle: number;
+} => {
+    const extendedPoint = getIntersectionOfRayWithBox(
+        throughPoint,
+        referencePoint,
+        box,
+    );
+    const direction = vec.sub(extendedPoint, referencePoint);
```
benchristel, line 29: Do we need the `extendedPoint` to calculate the direction? Or is it equally correct to use `throughPoint`?

I'd prefer using `throughPoint` if we can, because that makes it clear that the direction does not depend on the extendedPoint calculation.

packages/perseus/src/widgets/interactive-graphs/locked-line.tsx
```diff
@@ -15,52 +16,72 @@ type Props = LockedLineType & {
...
     const [point1, point2] = points;
+    // NOTE: Segments should not show both a point and an arrow.
+    //       Arrows supersede points.
+    const showStartPoint =
+        props.showStartPoint && (kind !== "segment" || !showArrows);
+    const showEndPoint =
+        props.showEndPoint && (kind !== "segment" || !showArrows);
+
+    let arrowHeadValues = getArrowheadValues(point1.coord, point2.coord, range);
+    let arrowTip = kind === "segment" ? point2.coord : arrowHeadValues.tip;
```
benchristel, line 29: The fact that we need this ternary to ignore `arrowHeadValues.tip` in the case of a segment graph makes me think that `getArrowheadValues` is the wrong abstraction. If it were inlined, this code could become:

```ts
const direction = vec.sub(throughPoint, referencePoint);
const angle = calculateAngleInDegrees(direction);
const tip = kind === "segment"
    ? throughPoint
    : getIntersectionOfRayWithBox(throughPoint, referencePoint, range);
```

packages/perseus/src/widgets/interactive-graphs/locked-line.tsx
```diff
@@ -15,52 +16,72 @@ type Props = LockedLineType & {
...
-                        lineStyle === "dashed"
-                            ? // TODO(lems-1930): Uncomment this line when the
-                              // dashed style is updated in Mafs.
-                              // ? "var(--mafs-line-stroke-dash-style)"
-                              "4, 3"
-                            : undefined,
-                }}
-            />
+            <>
+                {showArrows && startArrowHead}
```
benchristel, line 43: `startArrowhead` is only set if `showArrows` is true, so it feels like one of the checks for `showArrows` could be removed.

packages/perseus/src/widgets/interactive-graphs/graphs/components/vector.tsx
mark-fitzgerald: Good catch. It does indeed. Fixing...

packages/perseus/src/widgets/interactive-graphs/locked-line.tsx
```diff
@@ -15,52 +16,72 @@ type Props = LockedLineType & {
...
-        kind,
-        points,
-        showStartPoint,
-        showEndPoint,
-        range,
-    } = props;
+    const {color, lineStyle, kind, points, showArrows, range} = props;
     const [point1, point2] = points;
+    // NOTE: Segments should not show both a point and an arrow.
+    //       Arrows supersede points.
```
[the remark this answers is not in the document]
mark-fitzgerald: The editor has the option for line segments, and my understanding was that this is desired, but that is now under review. I will discuss during the on-site.

packages/perseus/src/widgets/interactive-graphs/locked-line.tsx
mark-fitzgerald: Yep. That was my intent, and then I just spaced it. Good catch!

packages/perseus/src/widgets/interactive-graphs/locked-line.tsx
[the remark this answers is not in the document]
mark-fitzgerald: This issue has been reviewed, and it has been determined that only lines and rays should have arrows.

packages/perseus/src/widgets/interactive-graphs/locked-line.tsx
mark-fitzgerald: Updated.

packages/perseus/src/widgets/__testdata__/interactive-graph.testdata.ts
```diff
@@ -1159,6 +1159,120 @@ export const segmentWithLockedLineQuestion: PerseusRenderer = {
     },
 };
 
+export const segmentWithLockedLineAndArrowheadsQuestion: PerseusRenderer = {
```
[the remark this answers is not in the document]
nishasy: I'd be down to add that functionality for locked figures in a future PR. I can make an M2 ticket for it.

packages/perseus/src/widgets/interactive-graphs/graphs/components/arrowhead.test.tsx
```diff
@@ -0,0 +1,22 @@
...
+        expect(svgArrow?.getAttribute("style")).toContain("stroke: green");
+    });
+
+    it("Inherits the color of the parent element when no color is specified", () => {
+        const {container} = render(<Arrowhead tip={[0, 0]} angle={0} />);
+        // eslint-disable-next-line testing-library/no-container,testing-library/no-node-access
+        const svgArrow = container.querySelector("path");
+        expect(svgArrow?.getAttribute("style")).toContain("stroke: inherit");
+    });
+});
```
nishasy, line 22: Do you think it would make sense to add a test for the SVG's `transform` property given a certain `angle` here?

nishasy (approved):
Very exciting!

packages/perseus/src/widgets/interactive-graphs/graphs/utils.ts
mark-fitzgerald: Refactoring makes this issue moot.

packages/perseus/src/widgets/interactive-graphs/graphs/components/arrowhead.test.tsx
mark-fitzgerald: Philosophically, it shouldn't matter how the arrow is rotated. So if we decide in the future to do that differently, we shouldn't have to update our tests. I think that if we want to verify that the arrow is rotated properly, then we should look into setting up visual regression tests (i.e. screen snapshots).

benchristel (approved):

PR #1230, opened 2024-04-26
Get `@khanacademy/mathjax-renderer` from NPM

benchristel:
## Summary:
Previously, we were installing mathjax-renderer directly from a private
GitHub repository. This caused third-party clients of Perseus to be unable
to use Perseus, since they don't have access to that repository.

This commit fixes that by installing `@khanacademy/mathjax-renderer`
from NPM.

Issue: LEMS-1818

Test plan:

- Install the test build of Perseus in webapp.
- Update the version of `@khanacademy/mathjax-renderer` used in webapp to 1.7.2.
- Verify that TeX rendering works.
- Verify that `yarn.lock` lists no GitHub URLs for `@khanacademy/mathjax-renderer`.

### DIFF
--- a/.changeset/strange-crabs-bake.md
+++ b/.changeset/strange-crabs-bake.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/math-input": minor
+---
+
+Get `@khanacademy/mathjax-renderer` from NPM, to allow third-party clients of Perseus to install it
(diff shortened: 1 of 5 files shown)


benchristel:
@jeremywiebe I've verified that MathJax can render in webapp (both clientside and serverside) with this version of Perseus.

jeremywiebe (approved):

PR #1235, opened 2024-04-29
Update mathjax-renderer to v2.

jeresig:
## Summary:
This adds in a required locale param to createMathField, which is the biggest change, and technically a breaking change, so I'm going to do a major version bump on math-input.

Issue: FEI-5571

## Test plan:
Tests all pass.


benchristel (approved):
Thanks! LGTM.

PR #1237, opened 2024-04-29
Circle Graph foundation

handeyeco:
## Summary:
Roughing out the Mafs Circle Graph.

Learner Experience

https://github.com/Khan/perseus/assets/16308368/f9624567-8b44-4549-98a1-0bf4e413cab7

Editor Experience

https://github.com/Khan/perseus/assets/16308368/3cd23ff1-c917-4fa8-b662-8da4909c1890

Issue: [LEMS-1828](https://khanacademy.atlassian.net/browse/LEMS-1828)

## Test plan:
- TODO

[LEMS-1828]: https://khanacademy.atlassian.net/browse/LEMS-1828?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/perseus/src/widgets/__tests__/__snapshots__/interactive-graph.test.ts.snap
+++ b/packages/perseus/src/widgets/__tests__/__snapshots__/interactive-graph.test.ts.snap
@@ -3302,7 +3302,7 @@ exports[`snapshots should render correctly with locked lines 1`] = `
                   >
                     <pattern
                       height="20"
-                      id="cartesian-9"
+                      id="cartesian-31"
                       patternUnits="userSpaceOnUse"
                       width="20"
                       x="0"
@@ -3310,7 +3310,7 @@ exports[`snapshots should render correctly with locked lines 1`] = `
                     >
                       <pattern
                         height="20"
-                        id="cartesian-9-subdivision"
+                        id="cartesian-31-subdivision"
                         patternUnits="userSpaceOnUse"
                         width="20"
                       >
@@ -3319,7 +3319,7 @@ exports[`snapshots should render correctly with locked lines 1`] = `
                         />
                       </pattern>
                       <rect
-                        fill="url(#cartesian-9-subdivision)"
+                        fill="url(#cartesian-31-subdivision)"
                         height="20"
                         width="20"
                       />
@@ -3353,7 +3353,7 @@ exports[`snapshots should render correctly with locked lines 1`] = `
                       </g>
                     </pattern>
                     <rect
-                      fill="url(#cartesian-9)"
+                      fill="url(#cartesian-31)"
                       height="640"
                       width="640"
                       x="-320"
@@ -4493,7 +4493,7 @@ exports[`snapshots should render correctly with locked points 1`] = `
                   >
                     <pattern
                       height="20"
-                      id="cartesian-8"
+                      id="cartesian-30"
                       patternUnits="userSpaceOnUse"
                       width="20"
                       x="0"
@@ -4501,7 +4501,7 @@ exports[`snapshots should render correctly with locked points 1`] = `
                     >
                       <pattern
                         height="20"
-                        id="cartesian-8-subdivision"
+                        id="cartesian-30-subdivision"
                         patternUnits="userSpaceOnUse"
                         width="20"
                       >
@@ -4510,7 +4510,7 @@ exports[`snapshots should render correctly with locked points 1`] = `
                         />
                       </pattern>
                       <rect
-                        fill="url(#cartesian-8-subdivision)"
+                        fill="url(#cartesian-30-subdivision)"
                         height="20"
                         width="20"
                       />
@@ -4544,7 +4544,7 @@ exports[`snapshots should render correctly with locked points 1`] = `
                       </g>
                     </pattern>
                     <rect
-                      fill="url(#cartesian-8)"
+                      fill="url(#cartesian-30)"
                       height="640"
                       width="640"
                       x="-320"
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/utils.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/utils.ts
@@ -46,8 +46,25 @@ export const getIntersectionOfRayWithBox = (
     }
 };
 
-export const getLines = (points: readonly vec.Vector2[]): CollinearTuple[] =>
-    points.map((point, i) => {
+export const getLines = (points: readonly vec.Vector2[]): CollinearTuple[] => {
+    return points.map((point, i) => {
         const next = points[(i + 1) % points.length];
         return [point, next];
     });
+};
+
+/**
+ * determine radius of a circle given the center point
+ * and a point on the circle's circumference
+ *
+ * @param center - center point of the circle
+ * @param edgePoint - point on the circumference of the circle
+ * @returns the radius of the circle
+ */
+export function getRadius(center: vec.Vector2, edgePoint: vec.Vector2): number {
+    const [centerX, centerY] = center;
+    const [edgeX, edgeY] = edgePoint;
+    return Math.sqrt(
+        Math.pow(edgeX - centerX, 2) + Math.pow(edgeY - centerY, 2),
+    );
+}
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/utils.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/utils.ts
@@ -52,19 +52,3 @@ export const getLines = (points: readonly vec.Vector2[]): CollinearTuple[] => {
         return [point, next];
     });
 };
-
-/**
- * determine radius of a circle given the center point
- * and a point on the circle's circumference
- *
- * @param center - center point of the circle
- * @param edgePoint - point on the circumference of the circle
- * @returns the radius of the circle
- */
-export function getRadius(center: vec.Vector2, edgePoint: vec.Vector2): number {
-    const [centerX, centerY] = center;
-    const [edgeX, edgeY] = edgePoint;
-    return Math.sqrt(
-        Math.pow(edgeX - centerX, 2) + Math.pow(edgeY - centerY, 2),
-    );
-}
--- a/packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
@@ -15,6 +15,7 @@ import {
 } from "./graphs";
 import {SvgDefs} from "./graphs/components/text-label";
 import {PointGraph} from "./graphs/point";
+import {getRadius} from "./graphs/utils";
 import {Grid} from "./grid";
 import {LegacyGrid} from "./legacy-grid";
 import {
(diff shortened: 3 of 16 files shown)


packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -202,6 +210,66 @@ function doMovePoint(
...
+}
+
+function doMoveRadiusPoint(
+    state: InteractiveGraphState,
+    action: MoveRadiusPoint,
+): InteractiveGraphState {
+    switch (state.type) {
+        case "circle": {
+            const nextRadiusPoint: vec.Vector2 = [
+                // The +0 is to convert -0 to +0
```
benchristel, line 244: ugh, floating point.

no action needed :P

benchristel (commented):
This looks like a good start! Was there something in particular you wanted feedback on?

I noticed that it's possible to drag either control point outside the graph bounds. We should have constraints that keep both points visible.

A minor UX thing: it might be nice if the cursor changed to a left-right arrow when you hover over the radius point. Otherwise, it's slightly surprising that that point can only move right and left.

packages/perseus/src/widgets/interactive-graphs/types.ts
```diff
@@ -71,6 +71,7 @@ export interface CircleGraphState extends InteractiveGraphStateCommon {
     type: "circle";
     center: Coord;
     radius: number;
+    radiusPoint: Coord;
```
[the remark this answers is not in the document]
handeyeco: Originally I was just using `radiusPoint`. I added back `radius` when I looked at grading, but I think I could remove it again.

handeyeco:
https://github.com/Khan/perseus/assets/16308368/93c768c6-1aee-434c-a2f9-fb9ab02bd7cc

Made the constraints keep the center dot in the chart and did what I could to keep the radius handle in view.

packages/perseus/src/widgets/interactive-graphs/types.ts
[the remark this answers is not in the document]
handeyeco: Okay, I remember what it was @benchristel: I wanted to store `radiusPoint` because I figured `radius` could just be computed from that and it'd give me some flexibility around positioning the handle.

However the way things worked was that `MafsGraph` was just blindingly passing its state up to `InteractiveGraph`. `InteractiveGraph` merges the object from the `onChange` callback into its state and without `radius` the editor doesn't work as expected.

This made me wonder: do we _have_ to be bound to the structure of state in the legacy interactive graph? So I updated my code to include `mafStateToInteractiveGraph`, the idea being that we could store whatever we want, however we want in the new code as long as we transform it to be compatible with the legacy code. I expect this could be controversial, but I also imagine it could be useful as we implement the weirder charts and try to add a11y improvements.

Obviously I could also just add `radius` back to the store.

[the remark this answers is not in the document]
benchristel: FWIW, I like the approach of keeping only the state we need in our `useReducer` store, and transforming it back to the legacy format only when we have to.

Currently, someone who's only looking at the useReducer state type has to wonder "why did they design it this way" and it's unsatisfying if the answer is "an accident of history."

benchristel:
> Screen.Recording.2024-04-29.at.4.14.21.PM.mov 
> Made the constraints keep the center dot in the chart and did what I could to keep the radius handle in view.

Cool! One thought, though: when the radius handle swaps over to the left side, the keyboard controls get reversed: left arrow now _expands_ the circle, whereas normally it contracts it. This is likely to be confusing for screenreader and keyboard users.

Note that with the current setup, I think this is the _only_ way to get the point to go over to the left side if you're only moving things with the keyboard.

Let's start with the assumption that the entire circle must remain visible within the graph bounds, and if we need to revisit that decision, we will.

packages/perseus/src/widgets/interactive-graphs/graphs/utils.ts
```diff
@@ -46,8 +46,25 @@ export const getIntersectionOfRayWithBox = (
...
+ * @param edgePoint - point on the circumference of the circle
+ * @returns the radius of the circle
+ */
+export function getRadius(center: vec.Vector2, edgePoint: vec.Vector2): number {
+    const [centerX, centerY] = center;
+    const [edgeX, edgeY] = edgePoint;
+    return Math.sqrt(
+        Math.pow(edgeX - centerX, 2) + Math.pow(edgeY - centerY, 2),
+    );
+}
```
benchristel, line 70: Minor: I think this `getRadius` function could live with the definition of the circle graph state type, and take the entire graph state as its argument:

```ts
getRadius(circleGraph);
```

...and then `getRadius` could live in `interactive-graph-state.ts`.

Our way of representing a circle, as a center point and a point on the edge, is pretty specific to the graph state format. Passing the entire graph state to `getRadius` would better encapsulate the structure of the state, and possibly let us change it more easily in the future.

(in general, the practice of passing entire state objects around opens up other possibilities, too, like caching the results of getter functions — see [`flipbook-model.ts`](https://github.com/Khan/perseus/blob/db041577dda780e2a871ad9e63c295a1d3fbe185/dev/flipbook-model.ts#L123-L152) for examples. I don't think caching is relevant here, though)

handeyeco:
@nishasy I added you specifically as a reviewer because I'm confused about why I needed to change the snapshots and I was hoping you might know.

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-state.ts
```diff
@@ -335,3 +350,17 @@ const getPolygonCoords = ({
...
+ *
+ * @param graph - the graph object containing the circle
+ * @returns the radius of the circle
+ */
+export function getRadius(graph: CircleGraphState): number {
+    const [centerX, centerY] = graph.center;
+    const [edgeX, edgeY] = graph.radiusPoint;
+    return Math.sqrt(
+        Math.pow(edgeX - centerX, 2) + Math.pow(edgeY - centerY, 2),
+    );
```
benchristel, line 365: The Pythagorean theorem feels like overkill here, but I guess it's less likely to break in the future.

benchristel (approved):
I left a few suggestions inline, but no blockers. LGTM!

packages/perseus/src/widgets/__tests__/__snapshots__/interactive-graph.test.ts.snap
```diff
@@ -3302,15 +3302,15 @@ exports[`snapshots should render correctly with locked lines 1`] = `
                   >
                     <pattern
                       height="20"
-                      id="cartesian-9"
+                      id="cartesian-31"
```
nishasy, line 3305: I'm not 100% certain, but I think this snapshot update has to do with [how Mafs does the ID](https://github.com/stevenpetryk/mafs/blob/168edc7816dce6a5402b83d62f5b58ba74c6a9b9/src/display/Coordinates/Cartesian.tsx#L29) for the Cartesian plane. [It uses an incrementer as a hack for unique IDs](https://github.com/stevenpetryk/mafs/blob/168edc7816dce6a5402b83d62f5b58ba74c6a9b9/src/display/Coordinates/Cartesian.tsx#L7). I really don't know why there are 22 more cartesian planes here than there were before though??

I also don't understand why the snapshot even previously only had `cartesian-8` and `cartesian-9`. Not sure where 1-7 went 🤔 

I wonder if @jeremywiebe would know?

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -202,6 +210,84 @@ function doMovePoint(
...
+    action: MoveCenter,
+): InteractiveGraphState {
+    switch (state.type) {
+        case "circle": {
+            // Constrain the center of the circle to the chart range
+            const [xMin, xMax] = state.range[0];
+            const [yMin, yMax] = state.range[1];
+            const constrainedCenter: vec.Vector2 = [
+                Math.min(Math.max(xMin, action.destination[0]), xMax),
+                Math.min(Math.max(yMin, action.destination[1]), yMax),
```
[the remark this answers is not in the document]
benchristel: Oh, actually, `bound` might do what you need, and then you don't even have to destructure the range. One problem with `bound` might be that it keeps the bounded point one snap step away from the graph's edge, so this *might* cause problems for existing content that lets you go right up to the top and right edges. I haven't checked.

packages/perseus/src/widgets/__tests__/__snapshots__/interactive-graph.test.ts.snap
jeremywiebe: I suspect that tests that involve the `<Cartesian />` component "burn" ids even if they don't involve capturing a snapshot. As long as the tests (including snapshot tests) run in the same order (which it seems they do) we should be ok with re-numbering these ids. 

It'd be nice to have something like [`useUniqueId()`](https://khan.github.io/wonder-blocks/?path=/docs/packages-core-useuniqueid--docs) from WB, but I don't think it's reasonable to introduce that into Mafs.

handeyeco: I think we need to find a fix for this, I keep having to regenerate snapshots.

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -45,6 +52,10 @@ export type Props = {
     labels: InteractiveGraphProps["labels"];
 };
 
+type MafsChange = {
+    graph: InteractiveGraphState;
+};
```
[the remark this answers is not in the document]
handeyeco: [Filed a ticket to follow up on this](https://khanacademy.atlassian.net/browse/LEMS-1974)

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -202,6 +210,84 @@ function doMovePoint(
...
+            // Try to position the radius handle in a visible spot
+            // if it otherwise would be off the chart
+            // ex: if the handle is on the right and we move the center
+            // to the rightmost position, move the handle to the left
+            const [radX] = newRadiusPoint;
+            if (radX < xMin || radX > xMax) {
+                const xJumpDist = (radX - constrainedCenter[0]) * 2;
+                const possibleNewX = newRadiusPoint[0] - xJumpDist;
+                if (possibleNewX >= xMin && possibleNewX <= xMax) {
+                    newRadiusPoint[0] = possibleNewX;
```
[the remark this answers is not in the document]
handeyeco: I added a spread operator to at least make sure it's not the same array object:

``` JS
            const newRadiusPoint: vec.Vector2 = [
                ...vec.add(
                    state.radiusPoint,
                    vec.sub(constrainedCenter, state.center),
                ),
            ];
```

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-state.ts
handeyeco: I personally like it for being future-proof.

handeyeco:
Everything's passing except some Changeset weirdness. I think it's safe to move ahead.

PR #1238, opened 2024-04-29
Update line segment tab order

benchristel:
During playtesting, Sarah B pointed out that the tab order for segment graphs
was not very intuitive. Tabbing to a segment first selects the whole segment.
Tabbing further selects each endpoint.

This PR updates the tab order so the first endpoint is selected first, then the
segment, then the second endpoint.

Issue: LEMS-1936

## Test plan:

View a Mafs segment graph in the dev ui (`yarn dev`). Tabbing between UI
elements should feel intuitive.

### DIFF
--- a/.changeset/tricky-goats-beg.md
+++ b/.changeset/tricky-goats-beg.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Fixed tab order for movable segments to be more intuitive. Previously, the tab order was (whole segment, point 1, point2). Now it's (point 1, whole segment, point 2).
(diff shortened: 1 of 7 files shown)


handeyeco:
Do we need to think about what happens when a learner flips control points?

https://github.com/Khan/perseus/assets/16308368/8223d647-8e61-4006-87d3-249018c005ba

handeyeco (approved):
Nice! This seems much more intuitive.

I think this would be really easy for someone in the future to change, causing a regression - especially without commenting on the importance of the render order or without a test (which I think would be straight-forward to write).

benchristel:
I am going to try a different approach, since the current approach has issues:

- There's a visual glitch when the line segment is selected: the line renders in front of the first point.
- The first workaround I tried to fix the visuals works on Edge, but breaks horribly on Firefox.

I'll open a new PR for my second attempt.

PR #1239, opened 2024-04-29
Prevent mobile styling from overriding graph axis label size

benchristel:
There was a bug in the exercise editor where the axis labels on Mafs interactive
graphs were too big.  The cause was the small-screen-specific styling changed
in this commit.

Issue: LEMS-1964

## Test plan:

- View the visual regression test added in this commit.
- The axis labels should not overlap with the graph grid.


handeyeco (approved):
Makes sense

PR #1240, opened 2024-05-02
Fix line segment tab order

benchristel:
During playtesting, Sarah B pointed out that the tab order for segment graphs
was not very intuitive. Tabbing to a segment first selects the whole segment.
Tabbing further selects each endpoint.

This PR updates the tab order so the first endpoint is selected first, then the
segment, then the second endpoint.

Issue: LEMS-1936

## Test plan:

View a Mafs segment graph in the dev ui (`yarn dev`). Tabbing between UI
elements should feel intuitive.


packages/perseus/src/widgets/__testdata__/interactive-graph.testdata.ts
```diff
@@ -769,7 +769,6 @@ export const polygonQuestionDefaultCorrect: PerseusRenderer = {
                         [2.5, 4],
                         [1.5, 2],
                     ],
-                    match: "congruent",
```
benchristel, line 772: The removal of this line is kind of a long story.

- I changed the RTL tests for interative graph to use `tab` to select the first movable element
- This meant that the test for polygon was selecting the entire polygon instead of the first point
- The polygon question was graded based on congruence
- That meant that moving the entire polygon, without changing its shape, resulted in a correct answer
- That caused the test for answering incorrectly to fail.

To test that incorrect answers are graded correctly, we make the grading more strict. When `match` is not present, the points must be exactly where the content author put them for the question to be marked correct.

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point-view.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+import useGraphConfig from "../../reducer/use-graph-config";
+import {useTransformVectorsToPixels} from "../use-transform";
+
+import type {vec} from "mafs";
+import type {ForwardedRef} from "react";
+
+type Props = {
+    point: vec.Vector2;
+    onMove?: (newPoint: vec.Vector2) => unknown; // FIXME: remove onMove
```
benchristel, line 14: I need to fix the FIXMEs in this file before this lands.

packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx
```diff
@@ -20,21 +20,17 @@ export const PolygonGraph = (props: Props) => {
...
 
-    const pointsSum = points.reduce(
-        (acc, point) => vec.add(acc, point),
-        [0, 0],
-    );
-    const midpoint =
-        pointsSum && vec.scale(pointsSum, 1 / (points.length ?? 1));
-
     const ref = React.useRef<SVGPolygonElement>(null);
+    const dragReferencePoint = points[0];
```
benchristel, line 28: This is a sneaky bugfix.

The bug was that sometimes, selecting an entire polygon and pressing an arrow key would not move the polygon. Our tests for the polygon graph happened to catch it (recall that since we're now using `tab` to select the first movable element in our tests, we're now selecting the entire polygon in the tests instead of selecting the first point)

The bug happened because the `midpoint` of the polygon was, in general, not on a snap step. This meant that trying to move the polygon sometimes caused it to snap back to its original location, effectively canceling the move.

Since the points of a polygon are always on a snap step, using the first point as the reference point fixes this.

nishasy:
@benchristel Do you have a video of the new tabbing experience?

benchristel:
@nishasy I've attached a video to the PR description.

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point-view.tsx
```diff
@@ -0,0 +1,130 @@
+import {color as WBColor} from "@khanacademy/wonder-blocks-tokens";
```
nishasy, line 1: For posterity, could you add a small docblock at the top of this file explaning what MovablePointView is and where/how it's used?

packages/perseus/src/widgets/interactive-graphs/graphs/ray.tsx
```diff
@@ -12,35 +11,22 @@ type Props = MafsGraphProps<RayGraphState>;
...
+        <Segment
+            // a ray only has one line
+            points={lines[0]}
+            onMoveLine={handleMoveLine}
+            onMovePoint={handleMovePoint}
+            extend={{
+                start: false,
+                end: true,
+            }}
+        />
```
nishasy, line 30: This is beautiful ✨

packages/perseus/src/widgets/interactive-graphs/graphs/components/segment.tsx
```diff
@@ -1,44 +1,136 @@
...
+    /* Extends the line to the edge of the graph with an arrow */
+    extend?:
+        | undefined
+        | {
+              start: boolean;
+              end: boolean;
+          };
+};
+
+export const MovableLine = (props: MovableLineProps) => {
```
nishasy, line 122: I don't think `MovableLine` needs to be exported anymore?

packages/perseus/src/widgets/interactive-graphs/graphs/components/segment.tsx
```diff
@@ -1,44 +1,136 @@
...
+        extend,
+    } = props;
+
+    const {visiblePoint: visiblePoint1, focusableHandle: focusableHandle1} =
+        useControlPoint(start, color, (p) => props.onMovePoint(0, p));
+    const {visiblePoint: visiblePoint2, focusableHandle: focusableHandle2} =
+        useControlPoint(end, color, (p) => props.onMovePoint(1, p));
+
+    return (
+        <>
```
nishasy, line 42: Beautiful! For posterity, could you add a comment here explaining that the focusHandles/visiblePoints are done this way because of the weirdness with SVGs and z-index/tabindex?

nishasy (approved):

PR #1241, opened 2024-05-06
Out of Bounds Axis Tick Labels and Tick Lines 

SonicScrewdriver:
## Summary:
There are many cases where we are required to render the axis tick labels outside of the graph bounds. This PR adds the necessary logic to ensure that we are handling all of the following edge cases:

1. Out-of-bounds X and Y labels 
2. Logic for handling graphs locked purely in the positive or negative quadrants for either X or Y 
3. Custom styling for when the y-axis labels are left-aligned (positioned on the right side of the graph) AND have a mix of positive and negative numbers.
4. Cases where an axis tick label is directly on the edge of the graph (and thus unable to fully render the label)
5. Cases where we have half or partial gridSteps visible on the graph 
6. Cases where the axis tick label widths need to be larger than the graph width 
7. Cases where we need to conditionally render the 0 label 
8. Cases where the y-axis is off-center

This PR also adds the following logic:
1. New logic for handling negative symbols to improve legibility and consistency with legacy graphs
2. New logic for adding half-tick lines on the edge of the graph to help with legibility/cognitive load.

Note: ~~I've had to hard-code some strange padding numbers here to fit with our custom implementation. I am happy to take any suggestions on better approaches to this issue!~~

Major props to Mark for helping me get rid of ALL the magic numbers! 

Issue: LEMS-1938

## Test plan:
- make tesc / new tests 
- manual testing 

## Video:
https://github.com/Khan/perseus/assets/12463099/6aa3fb32-42b8-41ce-a809-b81cebb2d8fc


packages/perseus/src/widgets/interactive-graphs/types.ts
```diff
@@ -74,3 +74,9 @@ export interface CircleGraphState extends InteractiveGraphStateCommon {
 }
 
 export type PairOfPoints = [Coord, Coord];
+
```
SonicScrewdriver, line 77: I moved this type here as I noticed it was being used across several different files

packages/perseus/src/widgets/interactive-graphs/graphs/components/axis-tick-labels.tsx
```diff
@@ -0,0 +1,236 @@
...
+    );
+};
+
+const XGridLabel = ({label, gridStep, tickStep, graphInfo}: GridLabel) => {
+    const pointOnAxis: vec.Vector2 = [label, 0];
+    const pixelPoint = pointToPixel(pointOnAxis, graphInfo);
+
+    // If the label are outside of the graphs viewbox/range we want to make sure the
+    // labels are still visible by clamping them to the appropriate edge of the graph
+    const topPadding = graphInfo.range[1][1] < 0 ? -16 : 7;
```
SonicScrewdriver, line 114: I think I should break these padding values out to named variables for clarity. I'm still not a fan of these hard-coded values but it makes a big difference for positioning depending on whether the axis is at the top or bottom of the graph.

packages/perseus/src/widgets/interactive-graphs/axis-ticks.tsx
```diff
@@ -11,87 +13,78 @@ const tickStyle: React.CSSProperties = {
...
+    }
+    if (graphInfo.range[1][1] < 0) {
+        // If the graph is on the negative side of the x-axis, lock the ticks to the right side of the graph
+        yPointOnAxis = graphInfo.range[1][1];
+    }
+
+    const pointOnAxis: vec.Vector2 = [x, yPointOnAxis];
     const [[xPosition, yPosition]] = useTransformVectorsToPixels(pointOnAxis);
 
+    // If the tick is on the edge of the graph's range, don't render it
```
SonicScrewdriver, line 72: This one is probably worth a new test, but it's an extreme edge case. I'll get a test made up for it.

Myranae (approved):
Wow! This is a very complex PR! (at least for me XD) It works other than some of the ticks being out of alignment (mentioned to you previously) and I noticed the tick labels overlap when the numbers get large and are in the negatives. Not sure there is much that can be done for that though. Left some non-blocking comments. Thanks so much for all your work on this!!
Negatives overlap
image

Ticks out of alignment
image

packages/perseus/src/widgets/interactive-graphs/mafs-styles.css
```diff
@@ -154,7 +154,50 @@
...
     font-size: 14px;
     font-family: "Mafs-MJXTEX";
+    line-height: 1.5em;
+}
+
+.y-axis-tick-labels {
+    width: 35px;
+    display: flex;
+    flex-flow: column;
+    transform: translateX(calc(-100% - 0.5em));
```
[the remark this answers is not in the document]
SonicScrewdriver: Oooo you're completely right, I think the pixel values were legacy values from early on in testing. I've updated them to use em, which seems much cleaner!

packages/perseus/src/widgets/interactive-graphs/mafs-styles.css
```diff
@@ -154,7 +154,50 @@
...
+
+    position: absolute;
+    text-align: right;
```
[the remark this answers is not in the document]
SonicScrewdriver: This was here before! :) We want to right align the numbers by default for the y-axis.

PR #1246, opened 2024-05-06
Changed locked figures default color to gray

nishasy:
## Summary:
SarahB suggested that the default color for locked figures
should be gray so that they don't look more interactive than
they need to.

Updating that here.

https://khanacademy.slack.com/archives/C067UM1QAR4/p1714408357638599?thread_ts=1714070893.730669&cid=C067UM1QAR4

Issue: none

## Test plan:
`yarn jest`

Storybook
- Add a new line and confirm it's gray, and its points are gray
- Add a new point and confirm it's gray

image

### DIFF
--- a/.changeset/small-donuts-applaud.md
+++ b/.changeset/small-donuts-applaud.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": patch
+---
+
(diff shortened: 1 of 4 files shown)


jeremywiebe (approved):
Nice. That was easy. 😄

PR #1247, opened 2024-05-07
Only show ruler/protractor warning when either is enabled

jeremywiebe:
## Summary:

This PR updates the banner warning content authors against using the ruler or protractor so that its only shown if one (or both) are enabled. This will cut down on visual clutter when neither are selected (which is most common).

Issue: LEMS-1963

## Test plan:

`yarn start`
Navigate to the EditorPage story
Add an `interactive-graph` widget
Enable the ruler - note that the banner appears
Disable the ruler and enable the protractor - note that the banner appears
Enable both ruler and protractor - note that the banner appears

### DIFF
--- a/.changeset/curvy-falcons-crash.md
+++ b/.changeset/curvy-falcons-crash.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": patch
+---
+
+Only show a11y banner for ruler and protractor when either or both is enabled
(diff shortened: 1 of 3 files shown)


nishasy (approved):
LGTM! I have a personal preference for the banner to be under the "show ruler" and "show protractor" checkboxes so that the checkbox doesn't jump after being clicked, but that's super not important.

Not 100% sure why codecov says that some of the lines aren't covered - maybe you need to add a test if `showProtractor` is true and `showRuler` is false?

jeremywiebe:
> LGTM! I have a personal preference for the banner to be under the "show ruler" and "show protractor" checkboxes so that the checkbox doesn't jump after being clicked, but that's super not important.
> 
> Not 100% sure why codecov says that some of the lines aren't covered - maybe you need to add a test if `showProtractor` is true and `showRuler` is false?

I don't have super-strong opinions. I've moved it below the checkboxes (but above the ruler config settings).  I think that works ok. Ideally this is not something most content authors see very often.

image

PR #1249, opened 2024-05-08
Suppress spurious errors logged by tests

benchristel:
## Summary:
Our Jest tests were logging that `<g>` is not a recognized HTML element.
That's because it's not HTML; it's only valid in SVGs.

The noisy errors made it harder to spot actual test failures in the
output. This commit wraps the rendered elements in an `<svg>` tag to
prevent the errors.

Issue: none

Test plan:

`yarn test`


jeremywiebe (approved):
Thanks for fixing this and making the logs less noisy!

PR #1250, opened 2024-05-08
Rename `Segment` to `MovableLine`

benchristel:
## Summary:
My last PR (#1240) left a rename half-finished. This PR finishes it.

Issue: none

Test plan:

`yarn test`


packages/perseus/src/widgets/interactive-graphs/graphs/components/segment.test.ts
benchristel, line 1: These tests were duplicated in `movable-line.test.tsx`!

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-line.tsx
```diff
@@ -130,7 +130,8 @@ type MovableLineProps = {
           };
 };
 
-export const MovableLine = (props: MovableLineProps) => {
+// Exported for testing
+export const Line = (props: LineProps) => {
```
benchristel, line 134: I don't love the way this code is split up - it's hard to find a name for this internal component that makes sense and clarifies how its purpose is different from that of `MovableLine`. But this `Line` component has tests pinning it down, so refactoring across the two components is costly.

jeremywiebe (approved):
That rename matches the component behaviour much better. Thanks!

PR #1255, opened 2024-05-09
Ensure axis lines and arrowheads are 2px thick.

mark-fitzgerald:
## Summary:
The default thickness for lines in Mafs is 1px. Our designs indicate 2px. These styling updates ensure that axis lines and arrowheads have the correct thickness.

Issue: LEMS-1929

## Test plan:
1. Launch storybook for Perseus (`yarn start`)
1. View the [Interactive Graph editor with locked lines story](http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph-editor--with-locked-lines)
1. Inspect the graph parts (hard to see visually, so inspecting is best for verifying)
   * Axis lines should have a `stroke-width` of 2px.
   * Arrowheads on lines and rays should also have a `stroke-width` of 2px.

### Before
Line Thickness - Before

### After
Line Thickness - After


packages/perseus/src/widgets/interactive-graphs/mafs-styles.css
```diff
@@ -154,7 +154,17 @@
...
+
 .MafsView .axis-ticks text {
     font-size: 14px;
     font-family: "Mafs-MJXTEX";
 }
+
+/* This is a gross method of referencing the axis lines within the graph,
+        but there are no other identifying traits that we can use to target these lines
+        since they are controlled by Mafs.
+ */
```
jeremywiebe, line 167: Indeed.  So much so that I'd be inclined to see if we can upstream the line thickness as an option.  I'd like to avoid hacks like this while we're in active development.  Steven is very open to us proposing changes.

mark-fitzgerald: I'll draw up a PR for Mafs.

mark-fitzgerald: Mafs has been updated (v0.18.6), and the latest commit to Perseus includes that update.

nishasy (approved):
Nice! Much cleaner!

PR #1256, opened 2024-05-09
Add a "drag handle" to the center of graphed circles

benchristel:
This brings the circle graph more in line with Sarah B's [Figma
designs](https://www.figma.com/file/JroPokeCLBv2VRIw3GPvY5/LC---Interactive-graph-widget?type=design&node-id=1150-51899&mode=design&t=rLzMLVWek25t36gx-0)

Issue: LEMS-1976, LEMS-1977

## Test plan:

Play around with the circle graphs in the `yarn dev` UI.

Video:

https://github.com/Khan/perseus/assets/693920/ebcb0e39-1418-480e-8169-4b0b043d2e42

### DIFF
--- a/packages/perseus/src/widgets/__stories__/interactive-graph-regression.stories.tsx
+++ b/packages/perseus/src/widgets/__stories__/interactive-graph-regression.stories.tsx
@@ -157,10 +157,16 @@ export const MafsWithMultipleSegments = (
     />
 );
 
-export const MafsCircleGraphWithNonsquareRange = (args: StoryArgs): React.ReactElement => (
+export const MafsCircleGraphWithNonsquareRange = (
+    args: StoryArgs,
+): React.ReactElement => (
     <div className="framework-perseus perseus-mobile">
         <MafsQuestionRenderer
-            question={interactiveGraphQuestionBuilder().withCircle().withYRange(-5, 5).withXRange(-10, 10).build()}
+            question={interactiveGraphQuestionBuilder()
+                .withCircle()
+                .withXRange(-10, 10)
+                .withYRange(-5, 5)
+                .build()}
         />
     </div>
 );
--- a/packages/perseus/src/widgets/__stories__/interactive-graph-regression.stories.tsx
+++ b/packages/perseus/src/widgets/__stories__/interactive-graph-regression.stories.tsx
@@ -157,6 +157,14 @@ export const MafsWithMultipleSegments = (
     />
 );
 
+export const MafsCircleGraphWithNonsquareRange = (args: StoryArgs): React.ReactElement => (
+    <div className="framework-perseus perseus-mobile">
+        <MafsQuestionRenderer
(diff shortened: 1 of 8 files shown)


packages/perseus/src/widgets/interactive-graphs/graphs/circle.tsx
```diff
@@ -38,3 +38,91 @@ export function CircleGraph(props: CircleGraphProps) {
...
+                fillOpacity={0}
+                color={color.blue}
+            />
+            <DragHandle center={center} />
+        </g>
+    );
+}
+
+const dragHandleDimensions: vec.Vector2 = [24, 14];
+const dragHandlePointPositions = crossProduct([-4.4, 0, 4.4], [-2.1, 2.1]);
```
mark-fitzgerald, line 88: Just curious - where did these numbers come from?

benchristel: I fiddled with them until the drag handle looked close to the design mock and reasonably legible.

packages/perseus/src/widgets/interactive-graphs/graphs/circle.tsx
```diff
@@ -17,17 +24,10 @@ export function CircleGraph(props: CircleGraphProps) {
...
                 radius={getRadius(graphState)}
-                fillOpacity={0}
-                color={color.blue}
-            />
-            <StyledMovablePoint
-                point={center}
-                onMove={(newCenter) => {
-                    dispatch(moveCenter(newCenter));
-                }}
+                onMove={(newCenter) => dispatch(moveCenter(newCenter))}
```
Myranae, line 30: Question: Are we going to add the fill color on hover? I think you mentioned it wasn't added yet, but will that be a separate PR?

packages/perseus/src/widgets/__stories__/interactive-graph-regression.stories.tsx
```diff
@@ -157,6 +157,20 @@ export const MafsWithMultipleSegments = (
     />
 );
 
+export const MafsCircleGraphWithNonsquareRange = (
```
Myranae, line 160: Question: Was this just a specific case that we want to keep an eye on? The graph definitely looks unusual. Is something like this in production?

Myranae (approved):
Looks good to me. Might want to update the video in the summary since we decided to make the drag handle blue. Love the new pixel functions too. Added a couple non-blocking comments as well.

packages/perseus/src/widgets/__stories__/interactive-graph-regression.stories.tsx
benchristel: I don't think we have any examples like this in production, but since the old graph supports this case, I wanted to make sure we don't regress.

packages/perseus/src/widgets/interactive-graphs/graphs/circle.tsx
benchristel: That's coming in https://khanacademy.atlassian.net/browse/LEMS-1978

PR #1257, opened 2024-05-09
Introduce EditorWithLayout

jeremywiebe:
## Summary:

This PR introduces a new Perseus Editor component that allows the host to control layout of individual editor pieces. It does this by using a callback function for the `children` prop. This is a similar approach to how the `MultiRenderer` works today. 

The associated Story for the `EditrorWithLayout` component provides an example of placing each piece of the editor wherever is needed. 

## What's next?

This PR can be thought of as a proof-of-concept for how this could be built. There are components that are nested inside other, high-level components (such as inside the `Editor`) that would need to be raised up to the `EditorWithLayout` so that they can be exposed individually.

The preview component `IframeContentRenderer` is also likely to be problematic as it uses an iframe that is typically pointed at production webapp. The LEMS team has ticket https://khanacademy.atlassian.net/browse/LEMS-1809 to fix this, but it is not on our near-term roadmap. We could adjust that if necessary. 

Issue: LEMS-1972

## Test plan:

`yarn`
`yarn start`
Navigate to http://localhost:6006/?path=/story/perseuseditor-editorwithlayout--controlled and test things out. 

Open `packages/perseus-editor/src/__stories__/editor-with-layout.stories.tsx` and move things around and note that the layout can be adjusted without making any changes to the `EditorWithLayout` component.

### DIFF
--- a/packages/perseus-editor/src/__stories__/editor-with-layout.stories.tsx
+++ b/packages/perseus-editor/src/__stories__/editor-with-layout.stories.tsx
@@ -1,5 +1,12 @@
-import EditorWithLayout from "../editor-with-layout";
+import {View} from "@khanacademy/wonder-blocks-core";
+import {color, spacing} from "@khanacademy/wonder-blocks-tokens";
+import {HeadingSmall} from "@khanacademy/wonder-blocks-typography";
+import {action} from "@storybook/addon-actions";
+import * as React from "react";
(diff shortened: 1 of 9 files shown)


jeremywiebe:
Closing as this is likely not the direction we'll go to make the editor more usable.

PR #1258, opened 2024-05-10
Remove unused wonder-blocks-i18n

jeremywiebe:
## Summary:

John migrated Perseus to use the new Khan Academy localization system where translated strings are passed in. This removed the need/use for wonder-blocks-i18n.  This PR removes a dangling dependency that we no longer need. 

Issue: "none"

## Test plan:

`yarn`
`yarn test`
`yarn tsc`


jeresig (approved):
huh - I definitely removed this, but I guess it got lost in a merge somewhere! Thanks for the update :)

PR #1259, opened 2024-05-10
Script to sync peer dependencies from host project (webapp)

jeremywiebe:
## Summary:

This PR introduces a script that can take a `package.json` (from webapp) and update all `peerDependencies` and `devDependencies` in package versions from that `package.json`.

Issue: "none"

## Test plan:

`./utils/sync-peer-dependencies.ts ~/khan/webapp/package.json`


package.json
```diff
@@ -1,160 +1,160 @@
...
-    "build-storybook": "storybook build",
-    "cypress": "cypress open --component",
-    "cypress:ci": "cross-env BABEL_COVERAGE=1 cypress run --component --env CYPRESS_COVERAGE=1",
-    "format": "prettier --write .",
-    "typecheck": "tsc"
-  },
-  "dependencies": {
-    "tiny-invariant": "^1.3.1"
-  }
+    "name": "perseus",
```
jeremywiebe, line 2: This file was using 2-spaces indentation, but our whole repo is set to 4-spaces.... so turn of "Showing whitespace" to view this PR).

image

jeremywiebe:
Tagging @jandrade in case you have thoughts.

utils/sync-peer-dependencies.ts
```diff
@@ -0,0 +1,109 @@
...
+function filterUnusableTargetVersions(
+    targetVersions: Record<string, string>,
+    packagesInThisRepo: ReadonlyArray<string>,
+): Record<string, string> {
+    return Object.fromEntries(
+        Object.entries(targetVersions).filter(
+            ([pkgName, pkgVersion]) =>
+                !RestrictedPackageVersions.some((r) => r.test(pkgVersion)) &&
+                !RestrictedPackageNames.some((name) => name === pkgName) &&
+                !packagesInThisRepo.some((name) => name === pkgName),
```
Myranae, line 48: Questions: So the last line filters out packages in the target versions object that are not in Perseus, right? The middle line filters out packages that are restricted, while the first line filters out package versions we can't use. Does the first line need to do a regex search because the package versions sometimes have characters like `^`?

Myranae:
Really cool! I love how this works. Looks like you have some merge conflicts, need a changeset, and the lockfile needs to be updated though. I'm curious how these were staying synced before, but I'm glad there's a script for it now!

jeremywiebe:
> Really cool! I love how this works. Looks like you have some merge conflicts, 
Fixed! 

> need a changeset, 
Added!

> and the lockfile needs to be updated though. 
Yikes, fixed!

> I'm curious how these were staying synced before, but I'm glad there's a script for it now!
It was all done manually and ad-hoc. Going forward, I'm hoping we can simply run this script and run tests.

utils/sync-peer-dependencies.ts
jeremywiebe: Yes. I added more comments here, but the two `RestrictedXyz` constants have comments on them explaining what they are for.

Basically, some package versions we don't/can't use, such as `link://...` versions in webapp which are vendored-in packages in the webapp tree. We use a similar package in Perseus but can't copy in those packages verbatim. In practice, those dependencies basically don't change so we're safe there. 

Next, some packages we don't want to sync because they are things like tool (Typescript). We maintain our own version in Perseus and upgrade when we need/want to. The packages we build have types that are compatible with webapp. 

The last filter removes any package that is in this repo, `changeset` manages these cross-repo package dependencies for us so we don't want to "pull in" the versions from webapp.

jandrade:
@jeremywiebe thanks for tagging me!

Did you find any issues with Perseus + webapp not being in sync (in terms of WB deps)? I do wonder if this could be done with a different take: Perseus geting all the latest versions of WB packages from npm instead? I'm not fully sure about the context behind this decision, so any extra clarification will be super helpful.

jeremywiebe:
> @jeremywiebe thanks for tagging me!
> 
> Did you find any issues with Perseus + webapp not being in sync (in terms of WB deps)? 

Well, until now, our versions of WB have not usually been updated when WB is released and so we've sometimes discovered version differences when we go to deploy a new Perseus release. 

I wonder if a different take would be to use Dependabot to simply open PRs any time those deps change.

mark-fitzgerald (approved):
Simple. I like it!

PR #1260, opened 2024-05-13
Rename start and end points to point1 and point2 in Interactive Graph locked lines

nishasy:
## Summary:
During playtesting, it was pointed out that "start point" and "end point" are
confusing names for the defining points of a line.

Instead, we're changing them to "point1" and "point2" (which is also in line
with how Mafs names them).

Issue: https://khanacademy.atlassian.net/browse/LEMS-1989

## Test plan:
`yarn jest`

Storybook
- Open the line accordion
- Confirm that the point accordions say "point 1" and "point 2" in the summaries.

image


jeremywiebe (approved):

nishasy:
Note: The fact that the UI tests say "328 tests unchanged here" means the locked figure accordions are never being snapshot tested in any story. I suspect this is because they're in the collapsed state on initial load, which is where the snapshot is taken.

I need to make a story where they're all expanded. I think this should be covered in the upcoming expand all / collapse all task though.

PR #1261, opened 2024-05-13
Fix bug where arrow keys did not always move segment

Myranae:
## Summary:
On some graphs, after focusing on the line segment, using the arrow keys would not move the segment. After fixing the line component to consider the snap step and adjusting dragging movement to be based on start point instead of midpoint, the line segments now move with all directional keys.

Issue: LEMS-1970

## Test plan:
- On this branch, open localhost with `yarn dev`
- Add in the following content in the flipbook and confirm all segments move in all 4 directions with all 4 arrow keys
```
{"content":"The transformation $\\text E\\,\\vec{x}$ is defined by the following matrix.\n\n$\\text E = \\left[\\begin{array}{rr}0 & 5 \\\\ -5 & 5\\end{array}\\right]$\n\n**Plot the result of $\\text E\\,\\vec{v}$ for $\\vec{v}=\\left[\\begin{array}{r}5 \\\\ 2\\end{array}\\right]$.** \n\n[[☃ interactive-graph 1]]\n\n","images":{},"widgets":{"interactive-graph 1":{"alignment":"default","graded":true,"options":{"backgroundImage":{"height":425,"url":"web+graphie://ka-perseus-graphie.s3.amazonaws.com/e230590dbd3ad30f570b592c71475f899f110097","width":425},"correct":{"coords":[[[0,0],[10,-15]]],"type":"segment"},"graph":{"type":"segment"},"gridStep":[5,5],"labels":["x","y"],"markings":"none","range":[[-25,25],[-25,25]],"rulerLabel":"","rulerTicks":10,"showProtractor":false,"showRuler":false,"showTooltips":false,"snapStep":[5,5],"step":[5,5]},"static":false,"type":"interactive-graph","version":{"major":0,"minor":0}}}}
{"content":"The graph of $y=h(x)$ is a line segment joining the points $(1,9)$ and $(3,2)$. \n\n**Drag the endpoints of the segment below to graph $y=h^{-1}(x)$.\n **\n\n[[☃ interactive-graph 1]]","images":{},"widgets":{"interactive-graph 1":{"alignment":"default","graded":true,"options":{"backgroundImage":{"bottom":0,"height":425,"left":0,"scale":1,"url":"web+graphie://ka-perseus-graphie.s3.amazonaws.com/a82b4098da55c75b68d2c1fd9f348923b8935ccb","width":425},"correct":{"coords":[[[9,1],[2,3]]],"type":"segment"},"graph":{"type":"segment"},"gridStep":[1,1],"labels":["x","y"],"markings":"none","range":[[-100,100],[-100,100]],"rulerLabel":"","rulerTicks":10,"showProtractor":false,"showRuler":false,"snapStep":[10,10],"step":[10,10]},"type":"interactive-graph","version":{"major":0,"minor":0}}}}
{"content":"Line segment $\\overline{OG}$ is rotated $180^\\circ$ about the point $(-2,4)$.  \n\n**Draw the image of this rotation using the interactive graph.**\n\n*The direction of a rotation by a positive angle is counter-clockwise.* \n\n[[☃ interactive-graph 1]]\n\n","images":{},"widgets":{"interactive-graph 1":{"graded":true,"options":{"backgroundImage":{"bottom":0,"height":425,"left":0,"scale":1,"url":"web+graphie://ka-perseus-graphie.s3.amazonaws.com/567155d84afcfa01baf0a44afd994fce3df17b5c","width":425},"correct":{"coords":[[[-7,7],[2,5]]],"type":"segment"},"graph":{"type":"segment"},"gridStep":[1,1],"labels":["x","y"],"markings":"none","range":[[-150,150],[-150,150]],"rulerLabel":"","rulerTicks":10,"showProtractor":false,"showRuler":false,"snapStep":[20,20],"step":[10,10]},"type":"interactive-graph","version":{"major":0,"minor":0}}}}
```
- Confirm all tests pass
- Confirm added tests are good regression tests
- After deploying, confirm graphs 2, 4, 5, and  7 now move on the following page: https://www.khanacademy.org/devadmin/content/articles/segment-graph/xa85475650c4b3749


packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
```diff
@@ -339,4 +332,178 @@ describe("MafsGraph", () => {
         expect(line.getAttribute("x2")).toBe(expectedX2 + "");
         expect(line.getAttribute("y2")).toBe(-expectedY2 + "");
     });
+
+    it("MovableLine moves up based on up keystroke ", async () => {
```
Myranae, line 336: Added tests for movement in each direction. I know this can also be written as a single test that runs through an array of options and plugs those values in, but since it was only 4 I didn't bother. Also I decided to use different coordinates for the line each time, which wouldn't work with the single test approach. Totally not necessary to use different points too though.

benchristel: I'm not sure if these tests actually protect us against regressions. In any case, I think the developer experience of running them could be improved. When I tried bringing back the old logic in `movable-line.tsx`, I got a weird failure:

```
Error: expect(jest.fn()).toHaveBeenCalledWith(...expected)

Expected: {"delta": [0, 0.5], "itemIndex": 0, "type": "move-line"}
Received
       1
          Object {
        -   "delta": Array [
        -     0,
        -     0.5,
        +   "snapStep": Array [
        +     1,
        +     1,
            ],
        -   "itemIndex": 0,
        -   "type": "move-line",
        +   "type": "change-snap-step",
          },
       2
          Object {
        -   "delta": Array [
        -     0,
        -     0.5,
        +   "range": Array [
        +     Array [
        +       -10,
        +       10,
        +     ],
        +     Array [
        +       -10,
        +       10,
              ],
        -   "itemIndex": 0,
        -   "type": "move-line",
        +   ],
        +   "type": "change-range",
          },
       3
          Object {
            "delta": Array [
              0,
        -     0.5,
        +     0.4,
            ],
            "itemIndex": 0,
            "type": "move-line",
          },

Number of calls: 4
```

It looks like there are some other actions getting dispatched that we don't care about. I think the part we actually care about is this:

```
        -     0.5,
        +     0.4,
```

but even that doesn't tell the developer running the test "the line isn't going to move because Mafs isn't snapping the point correctly".

Looking at the failure, I also noticed that a `change-snap-step` action is getting dispatched which sets the snap step to `[1, 1]`. This will probably interfere with the test logic. I think that action is getting dispatched because `baseMafsGraphProps` has a snapStep of `[1, 1]`.

Long story short, I think that rather than mock `dispatch`, I'd prefer to simulate the `useReducer` store like this:

```ts
// at the top level:

function createFakeStore<S, A>(reducer: (state: S, action: A) => S, state: S) {
    return {
        dispatch(action: A) {
            state = reducer(state, action);
        },

        getState() {
            return state;
        }
    }
}

// ...and then, in each test:

  const {dispatch, getState} = createFakeStore(interactiveGraphReducer, initialState)

  const baseMafsGraphProps = getBaseMafsGraphProps();

  render(
      <MafsGraph
          state={getState()}
          dispatch={dispatch}
          {...baseMafsGraphProps}
          snapStep={[0.5, 0.5]}
      />,
  );

  const group = screen.getByTestId("movable-line");
  group.focus();
  await userEvent.keyboard("[ArrowDown]");

  expect(getState().coords).toEqual(expectedCoords);
```

If we do that, it's easy to see from the test failure where the line ended up (whether it moved or not). We don't have to convert the line coordinates to pixel coordinates or re-render the graph.

I'd be happy to pair on reworking the tests if you want. Maybe tomorrow?

Myranae: Awesome! Thanks for the updated test suggestion! Yea, I'll play around with this a bit and see what I can manage. Definitely pairing if I can't get it working. I just took over a test that was there (and no longer needed), so I definitely appreciate the improvement :)

Myranae: Fixed!

mark-fitzgerald (approved):
Looks good!

nishasy (approved):

packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
```diff
@@ -101,6 +114,7 @@ describe("MafsGraph", () => {
         userEvent = userEventLib.setup({
             advanceTimers: jest.advanceTimersByTime,
         });
+        setDependencies(testDependencies);
```
jeremywiebe, line 117: I know we have two ways of doing this in Perseus, but I'd recommend doing the following instead of calling `setDependencies()` here. 

Top of file: 

```js
```

```suggestion
        jest.spyOn(Dependencies, "getDependencies").mockReturnValue(
            testDependencies,
        );
```

This has one major advantage: jest will automatically undo the mock at the end of the test. Calling `setDependencies` sets the global state of Perseus and means we really should be "undoing" this set at the end of the test. If we use a `spyOn` it'll be done automatically.

jeremywiebe (approved):

PR #1263, opened 2024-05-13
Fix a bug where line-ending arrows would flip around or vanish

benchristel:
The bug happened when `-0` was passed as a coordinate (yes, `-0` is a distinct
value in floating point, because the sign bit is stored separately from the
mantissa).

This PR patches the bug by handling horizontal and vertical lines as a special
case, and then refactors to a design where the special-casing is no longer
necessary. It's probably easiest to read the PR one commit at a time.

Issue: LEMS-1955

## Test plan:

- `yarn dev`
- Play around with the linear graphs at `localhost:5173`

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-line.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-line.tsx
@@ -148,10 +148,10 @@ const Line = (props: LineProps) => {
     if (extend) {
         const trimmedRange = trimRange(range, graphDimensionsInPixels);
         startExtend = extend.start
-            ? getIntersectionOfRayWithBox(start, end, trimmedRange)
+            ? getIntersectionOfRayWithBox(end, start, trimmedRange)
             : undefined;
         endExtend = extend.end
-            ? getIntersectionOfRayWithBox(end, start, trimmedRange)
+            ? getIntersectionOfRayWithBox(start, end, trimmedRange)
             : undefined;
     }
 
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/utils.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/utils.ts
@@ -24,11 +24,12 @@ export const getIntersectionOfRayWithBox = (
     const yDiff = bY - aY;
     const xDiff = bX - aX;
     const slope = yDiff / xDiff;
+    const inverseSlope = 1 / slope
 
-    const yAtXMin = slope * (xMin - aX) + aY;
-    const yAtXMax = slope * (xMax - aX) + aY;
-    const xAtYMin = (yMin - aY) / slope + aX;
-    const xAtYMax = (yMax - aY) / slope + aX;
+    const yAtXMin = aY + (xMin - aX) * slope;
+    const yAtXMax = aY + (xMax - aX) * slope;
+    const xAtYMin = aX + (yMin - aY) * inverseSlope;
+    const xAtYMax = aX + (yMax - aY) * inverseSlope;
 
     // clock analogy to describe quadrants
     switch (true) {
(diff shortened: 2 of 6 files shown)


packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-line.tsx
```diff
@@ -148,10 +148,10 @@ const Line = (props: LineProps) => {
     if (extend) {
         const trimmedRange = trimRange(range, graphDimensionsInPixels);
         startExtend = extend.start
-            ? getIntersectionOfRayWithBox(start, end, trimmedRange)
+            ? getIntersectionOfRayWithBox(end, start, trimmedRange)
```
jeremywiebe, line 151: Not directly related to this PR, but Nisha recently changed the locked line component to simply use `point1` and `point2` for lines. This feels like it might be applicable for the `MovableLine` component(s) as well as for line and ray they don't really have a start/end but rather two points that the line/ray intersect.  🤔  

For a future PR if you agree.

packages/perseus/src/widgets/interactive-graphs/graphs/utils.ts
```diff
@@ -18,35 +18,31 @@ export const getIntersectionOfRayWithBox = (
...
+        case isBetween(yAtXExtreme, yMin, yMax):
+            return [xExtreme, yAtXExtreme];
+
+        // does the ray intersect the top or bottom edge?
+        case isBetween(xAtYExtreme, xMin, xMax):
+            return [xAtYExtreme, yExtreme];
+
         default:
-            return [xMax, yAtXMax];
+            return [0, 0];
```
jeremywiebe, line 45: Is this basically stating that we'll never hit this `default`?  If so, could you comment on that as I might be confused for a bit if I arrived at this line of code without having read this PR or surrounding code.

packages/perseus/src/widgets/interactive-graphs/graphs/utils.ts
```diff
@@ -18,35 +18,31 @@ export const getIntersectionOfRayWithBox = (
...
-        case yDiff <= 0 && xDiff > 0:
-            // xAtYMin evaluates to -Infinity here, so we use absolute value
-            return Math.abs(xAtYMin) > xMax ? [xMax, yAtXMax] : [xAtYMin, yMin];
-        // 9 o'clock to 11:59
-        case yDiff >= 0 && xDiff < 0:
-            return xAtYMax < xMin ? [xMin, yAtXMin] : [xAtYMax, yMax];
-        // 6 o'clock to 8:59
-        case yDiff < 0 && xDiff <= 0:
-            return xAtYMin < xMin ? [xMin, yAtXMin] : [xAtYMin, yMin];
+        // does the ray intersect the left or right edge?
```
jeremywiebe, line 36: I'm not sure if it's obvious to others, but this is: 

```suggestion
        // does the ray intersect the left or right edge within the y graph range?
```

right?

jeremywiebe (approved):
This feels much more clear now and needs less complex calculations (the clock analogy). Thanks!

packages/perseus/src/widgets/interactive-graphs/graphs/utils.ts
```diff
@@ -8,8 +8,8 @@ export function calculateAngleInDegrees([x, y]: vec.Vector2) {
 /**
  * Given a ray and a rectangular box, find the point where the ray intersects
  * the edge of the box. Assumes the `initialPoint` is inside the box.
- * @param initialPoint - The starting point of the ray.
- * @param throughPoint - A point that the ray passes through. Must be different from initialPoint.
+ * @param initialPoint - A point that the ray passes through. Must be different from initialPoint.
+ * @param throughPoint - The starting point of the ray.
```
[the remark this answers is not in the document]
benchristel: Good catch!

PR #1264, opened 2024-05-14
Show images on their own line when included in a radio widget.

mark-fitzgerald:
## Summary:

When images are added to a radio widget, they are often intended to be on their own line. Creators indicate this by entering two new-line (`\n\n`) characters. This gets rendered as a separate `div.paragraph` in the HTML. This code change assigns CSS to the elements that match this rendered pattern. Content that has only a single new-line character around the image are rendered with the same `div.paragraph`, and therefore do not receive this styling.

Issue: LEMS-1917

## Test plan:
1. Launch Storybook locally (`yarn start`)
1. Navigate to "[Perseus > Widgets > Radio > Select With Images](http://localhost:6006/?path=/story/perseus-widgets-radio--select-with-images)"
1. Note the options that have images on their own line. _(There should be only 1 that is not on its own line.)_

### Before
Images in Radio Widget - Before

### After
Images in Radio Widget - After


mark-fitzgerald:
> I'm curious about a case where there is text before AND after an image. Or maybe one with multiple images? Could you add those as more radio options in the story?

Added - one with text before and after the image, and another with just 2 images.

packages/perseus/src/styles/widgets/radio.less
```diff
@@ -51,6 +51,14 @@
             border: 1px solid #ccc;
         }
     }
+
+    .perseus-radio-option {
+        /* Radio options with images that are set apart should be on their own line */
+        .paragraph + .paragraph:has(img:only-child), /* Image that is preceded by content */
+        .paragraph:has(img:only-child):has(+ .paragraph) /* Image that precedes other content */ {
+            display: block;
+        }
```
benchristel, line 60: I don't think we can use `:has` because it is [not supported on Safari 14](https://caniuse.com/?search=%3Ahas). Our policy is to keep support for any browser version with > 1% market share.

Our list of supported browsers is maintained here: https://docs.google.com/spreadsheets/d/1rHNwOvDu7B_RRU4EClUrqA3epsp93z7twQzJ5T2MptE/edit#gid=99235600

benchristel (changes requested):
The output looks great! Requesting changes because I don't think we can use `:has` yet. Maybe talk to the frontend infra team about re-running the browser numbers? It's been a few months since we last generated them.

packages/perseus/src/styles/widgets/radio.less
mark-fitzgerald: Good callout. I wasn't aware of the 1% threshold. Also, technically `:has()` only works on 15.4 and above, so lower versions of v15 won't work either. Regarding that 1% threshold, is it intended to be 1% for the browser (8,497,328), or 1% of all sessions (49,877,646)?

This leads to a new question: Are we OK with a "progressive enhancement" approach to something like this? Since the widget is still function without the styling in place, I'm inclined to follow this approach. I'm open to other ways of implementing this fix, but I'm not sure how to do it without adding logic to the markdown renderer.

benchristel: > Regarding that 1% threshold, is it intended to be 1% for the browser (8,497,328), or 1% of all sessions (49,877,646)

Good question - I don't know! I've just been trusting the percentages in that spreadsheet to be the ones we care about.

> Are we OK with a "progressive enhancement" approach to something like this?

Personally, my ideal would be to have code that works the same in all browsers we support. But I think I'd be okay with this feature working differently or not working on older Safaris, as long as the UX doesn't regress from the way it is now. Eventually those versions will age out and then this code _will_ work the same everywhere.

benchristel: I guess one question is: do we even have access to old versions of Safari we can test this on? I'm running 15.6 — too new. AFAIK it's not easy to downgrade.

benchristel (approved):
Okay, you've convinced me — since this is purely additive CSS, it's unlikely to cause a regression in Safari < 15.4, where `:has` isn't supported. I think this is good to land.

PR #1266, opened 2024-05-14
Remove range and snapStep from MafsGraph props

benchristel:
The `range` and `snapStep` are available on the `state`. When pairing, Tamara
and I saw some surprising test failures that were caused by this duplication.
Our props and state got out of sync, and since some parts of the graph code
were getting snapStep from props, and other parts were getting it from state,
the resulting behavior was inconsistent.

This commit removes the duplication, ensuring that `range` and `snapStep` are
only passed to each component once.

Issue: none

## Test plan:

- `yarn storybook`
- Visit http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph-editor--with-mafs
- Change the range of the graph and observe that the line segment is
  constrained to the new range when you drag it around.

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
@@ -7,15 +7,11 @@ import invariant from "tiny-invariant";
 import {testDependencies} from "../../../../../testing/test-dependencies";
 import {setDependencies} from "../../dependencies";
 
-import {
-    MafsGraph,
-    StatefulMafsGraph,
-    StatefulMafsGraphProps,
-} from "./mafs-graph";
+import {MafsGraph, StatefulMafsGraph} from "./mafs-graph";
 import {movePoint} from "./reducer/interactive-graph-action";
 import {interactiveGraphReducer} from "./reducer/interactive-graph-reducer";
 
-import type {MafsGraphProps} from "./mafs-graph";
+import type {MafsGraphProps, StatefulMafsGraphProps} from "./mafs-graph";
 import type {InteractiveGraphState} from "./types";
 import type {GraphRange} from "../../perseus-types";
 import type {UserEvent} from "@testing-library/user-event";
--- a/packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
@@ -23,7 +23,6 @@ function getBaseMafsGraphProps(): MafsGraphProps {
         gridStep: [1, 1],
         markings: "graph",
         containerSizeClass: "small",
-        onChange: () => {},
         showTooltips: false,
         labels: ["x", "y"],
         dispatch: () => {},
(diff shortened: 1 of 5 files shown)


packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
```diff
@@ -21,17 +25,41 @@ function getBaseMafsGraphProps(): MafsGraphProps {
         box: [400, 400],
         step: [1, 1],
         gridStep: [1, 1],
-        snapStep: [1, 1],
         markings: "graph",
+        containerSizeClass: "small",
+        showTooltips: false,
+        labels: ["x", "y"],
+        dispatch: () => {},
+        state: {
```
Myranae, line 29: Question: Was state added here so that the custom state objects in the tests can overwrite this information?

packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
```diff
@@ -21,17 +25,41 @@ function getBaseMafsGraphProps(): MafsGraphProps {
...
+            snapStep: [1, 1],
+            range: [
+                [-10, 10],
+                [-10, 10],
+            ],
+        },
+    };
+}
+
+function getBaseStatefulMafsGraphProps(): StatefulMafsGraphProps {
```
Myranae, line 43: Questions: Is this separate function needed just to more closely match the types required by certain tests? And this stateful one is the same as the old base one was, right?

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-action.ts
```diff
@@ -98,10 +97,10 @@ export function moveRadiusPoint(destination: vec.Vector2): MoveRadiusPoint {
 export const CHANGE_SNAP_STEP = "change-snap-step";
 export interface ChangeSnapStep {
     type: typeof CHANGE_SNAP_STEP;
-    snapStep: MafsGraphProps["snapStep"];
+    snapStep: [x: number, y: number];
```
Myranae, line 100: Question: Why were these changed to be explicit?

Myranae (approved):
Looks good! Left some non-blocking comments. Thanks for fixing this! Does look like you still have some PR checks to fix though.

packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
benchristel: Yep!

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-action.ts
benchristel: My change to the prop type names broke the import of `MafsGraphProps`, and I took the opportunity to inline the types rather than fix the import.

My reasoning is: if you want to dispatch an action, you need to know what values to pass as the arguments. A type like `MafsGraphProps["snapStep"]` doesn't tell you what value you can pass — you have to take the extra step of looking up what `MafsGraphProps["snapStep"]` actually is.

The inlined types are more self-documenting. The downside, arguably, is that the types are now duplicated. But this duplication is the harmless kind. Normally, the problem with duplication is that things can get out of sync. But there's no danger that types will get out of sync because we have a tool (the typescript compiler) that ensures they stay consistent.

PR #1267, opened 2024-05-14
Remove markings from interactive graph state

benchristel:
It seems to be unused. `markings` is still available via the MafsGraph props.
It doesn't need to be part of the state because nothing you can do to the graph
can change it, and it's not needed to compute state updates.

Issue: none

## Test plan:

`yarn typecheck`

### DIFF
--- a/.changeset/angry-islands-cough.md
+++ b/.changeset/angry-islands-cough.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/perseus": patch
+"@khanacademy/perseus-editor": patch
+---
+
+Internal: remove the `markings` property from the interactive graph state
--- a/packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
@@ -30,7 +30,6 @@ function getBaseMafsGraphProps(): MafsGraphProps {
             type: "segment",
             hasBeenInteractedWith: false,
             coords: [],
-            markings: "none",
             snapStep: [1, 1],
             range: [
                 [-10, 10],
@@ -145,7 +144,6 @@ describe("MafsGraph", () => {
         const mockDispatch = jest.fn();
         const state: InteractiveGraphState = {
             type: "segment",
-            markings: "graph",
             hasBeenInteractedWith: true,
             range: [
                 [-10, 10],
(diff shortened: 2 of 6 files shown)


packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -36,7 +36,7 @@ import type {Widget} from "../../renderer";
 import "mafs/core.css";
 import "./mafs-styles.css";
 
-export type Props = {
+export interface StatefulMafsGraphProps {
```
jeremywiebe, line 39: I noticed that in Mafs code it uses `interface` for Props types. In Perseus we've typically used `type`.  Is there a benefit/improvement from using `interface`? I don't have strong opinions other than a desire to be consistent. :)

jeremywiebe (approved):
I think this makes sense. It looks like the logic changes were mostly pulling up the `useEffect()` hooks into the StatefulMafsGraph, right?

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
benchristel: Sorry, this change came from [another PR](https://github.com/Khan/perseus/pull/1266) that this one was stacked on top of. https://github.com/Khan/perseus/pull/1266

The difference between `type` and `interface` is something I need to study more. The major difference I can think of is that only `interface`s can be `extend`ed. So you can do:

```ts
interface DerivedType extends Supertype {
  someProperty: string
}
```

whereas with `type`, your only option is

```
type DerivedType = Supertype & {someProperty: string}
```

which has surprising behavior when Supertype has a `someProperty` that's not a string. It looks like this may have been fixed in the latest version of TS to be less surprising (someProperty is correctly typed as the intersection of the property types from `Supertype` and `{someProperty: string}`) but IMO the `extends` behavior (subtype overrides properties of supertype) is more intuitive and useful.

To be super hand-wavy: OO subtyping and algebraic type intersection have different semantics and different use cases.

There is also a difference in editor tooling: JetBrains displays these little icons in the gutter to tell you which member of an interface a value is implementing.

Screen Shot 2024-05-15 at 9 40 35 AM

In this specific case, I don't think there's a practical difference between type and interface. I'll probably switch this back to `type` (on the other branch).

PR #1268, opened 2024-05-14
Collapse/Expand all button for locked figure settings

nishasy:
## Summary:
Add expand/collapse all button so that all the locked figures
settings accordion sections can be easily expanded or collapsed
all at once.

I chose to make it "collapse all" if only some are expaneded, since
it's easier to collapse and then expand than it is the expand and
then collapse because of the scroll distance. 

Issue: https://khanacademy.atlassian.net/browse/LEMS-1992

## Test plan:
`yarn jest`

Stories
- http://localhost:6006/?path=/docs/perseuseditor-components-locked-point-settings--docs
- http://localhost:6006/?path=/docs/perseuseditor-components-locked-line-settings--docs

## Demo:

https://github.com/Khan/perseus/assets/13231763/cde0248c-8132-4083-a007-7fd40d9205f6


packages/perseus-editor/src/components/__stories__/locked-point-settings.stories.tsx
```diff
@@ -43,18 +43,25 @@ export const Controlled: StoryComponentType = {
...
+};
+
 /**
  * In some cases, the locked point may be shown or hidden from the graph,
  * such as when the point is used to define a locked line.
  *
- * Use `onToggle` and `toggled` to update the visibility of the locked point.
- * When `onToggled` is passed in, a switch will be rendered to allow
+ * Use `onExtrasToggle` and `extrasToggled` to update the visibility of the locked point.
+ * When `onextrasToggled` is passed in, a switch will be rendered to allow
```
mark-fitzgerald, line 58: Typo: Should this be "When `onExtrasToggled` is passed in..."?

mark-fitzgerald:
Not a big deal, but I noticed that if all the accordions are collapsed, and then one is opened and then deleted, the button stays with "Collapse All", even though the remaining one(s) are already collapsed.

https://github.com/Khan/perseus/assets/13896410/1acc531a-2e8f-400f-a61a-406fa1b3f545

mark-fitzgerald (approved):
I left some minor, non-blocking comments. Otherwise, looks good!

packages/perseus-editor/src/components/__stories__/locked-point-settings.stories.tsx
nishasy: Yes! But this is also going away in my next PR, so I'm going to leave it as is for now

PR #1270, opened 2024-05-15
Update WB toolbar depenency in dev

nishasy:
## Summary:
This check was failing on everyone's branches, so fixing it here
https://github.com/Khan/perseus/actions/runs/9088536946/job/24978356027?pr=1268

Issue: none

## Test plan:
`yarn typecheck`
`yarn run lint`

### DIFF
--- a/dev/package.json
+++ b/dev/package.json
@@ -34,6 +34,10 @@
         "@phosphor-icons/core": "^2.0.2"
     },
     "devDependencies": {
+        "@khanacademy/wonder-blocks-toolbar": "^3.0.31",
         "vite": "^5.1.0"
+    },
+    "peerDependencies": {
+        "@khanacademy/wonder-blocks-toolbar": "^3.0.31"
     }
 }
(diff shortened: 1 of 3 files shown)


benchristel (commented):
I think the toolbar is only used by the dev UI, right? If that's true, I'd prefer to add the dependency to `dev/package.json` if we can. Otherwise, clients of Perseus will have to install `wonder-blocks-toolbar` when nothing in Perseus itself actually needs it.

benchristel (approved):
LGTM! 🎉

nishasy:
🤦🏽‍♀️ This PR isn't doing what I thought it's doing. The toolbar dependency was already there, and switching from 3.0.30 to 3.0.31 is trivial and won't actually fix the linting/typecheck issues we've been seeing.

I'm going to abandon this for now, and we can update this dependency when there's actually a need for it. I don't think I can land it anyway because of the type error that we're still trying to figure out.

nishasy:
Cache is fixed here, so I guess I'll land it anyway

PR #1271, opened 2024-05-15
Correct how arrowhead rotation is calculated

mark-fitzgerald:
## Summary:
Arrowhead rotation (aka "direction") was calculated using graph coordinates. While this works properly most of the time, there are situations where it doesn't. Specifically, when grid units are displayed more rectangularly than as a regular square. This code change resolves that issue by using pixel coordinates, instead.

Issue: LEMS-1996

## Test plan:
1. Launch the Dev environment (`yarn dev`)
1. Go to the flipbook (http://localhost:5173/#flipbook)
1. Paste in the following graph definition:
```
{"content":"[[☃ interactive-graph 1]]\n\n\n","images":{},"widgets":{"input-number 1":{"alignment":"default","graded":true,"options":{"answerType":"number","inexact":false,"maxError":0.1,"simplify":"required","size":"small","value":"2.6666666666666665"},"type":"input-number","version":{"major":0,"minor":0}},"interactive-graph 1":{"alignment":"default","graded":true,"options":{"backgroundImage":{"bottom":0,"height":0,"left":0,"scale":1,"url":null,"width":0},"correct":{"coords":[[0,0],[3,8]],"type":"linear"},"graph":{"type":"linear"},"gridStep":[1,1],"labels":["t","d"],"markings":"graph","range":[[-10,1],[-10,10]],"rulerLabel":"","rulerTicks":10,"showProtractor":false,"showRuler":false,"snapStep":[0.5,0.5],"step":[1,1]},"type":"interactive-graph","version":{"major":0,"minor":0}}}}
```
4. Move one of the line points around and note that the arrowheads appear to align properly with the line.

### Before
Arrowhead Rotation - Before

### After
Arrowhead Rotation - After


packages/perseus/src/widgets/__tests__/__snapshots__/interactive-graph.test.ts.snap
```diff
@@ -4185,7 +4185,7 @@ exports[`snapshots should render correctly with locked lines 1`] = `
                     />
                     <g
                       class="interactive-graph-arrowhead"
-                      transform="translate(-200 153.33333333333331) rotate(-192.52880770915152)"
+                      transform="translate(-200 153.33333333333331) rotate(167.47119229084848)"
```
mark-fitzgerald, line 4188: This is actually the same angle: -192.52880770915152 + 360 = 167.47119229084848

packages/perseus/src/widgets/interactive-graphs/graphs/components/__snapshots__/movable-line.test.tsx.snap
```diff
@@ -330,7 +330,7 @@ exports[`Rendering Does render extensions of line when option is enabled 1`] = `
         />
         <g
           class="interactive-graph-arrowhead"
-          transform="translate(0 0) rotate(-45)"
+          transform="translate(0 0) rotate(0)"
```
mark-fitzgerald, line 333: The line associated with these arrowheads are drawn with weird point references ([0,0], [0,0]) ([see preceding `<line>` element](https://github.com/Khan/perseus/blob/f67858a4c7bdd76116aa79eebbad298227f0a86b/packages/perseus/src/widgets/interactive-graphs/graphs/components/__snapshots__/movable-line.test.tsx.snap#L325-L330)). This line has no slope, so the rotation should be zero.

benchristel (approved):
LGTM!

An idea: it might be good to add a visual regression test for this bug, in `interactive-graph-regression.stories.tsx`. To do that, you'd probably need to add a new `InteractiveFigureConfig` subtype for linear graphs in `interactive-graph-question-builder.ts`. LMK if you'd like to pair on that.

PR #1275, opened 2024-05-15
Stub getDependencies instead of calling setDependencies in tests

benchristel:
We had several tests that were calling setDependencies, which has the
possibility of contaminating tests that run afterward, since the dependencies
are not reset after every test.

This PR changes the tests to always stub getDependencies instead of calling
setDependencies. Jest automatically removes the stubs after every test.

Issue: https://khanacademy.atlassian.net/browse/LEMS-1997

## Test plan:

`yarn test`

### DIFF
--- a/packages/perseus/src/__tests__/widgets.test.ts
+++ b/packages/perseus/src/__tests__/widgets.test.ts
@@ -1,5 +1,5 @@
 import {testDependencies} from "../../../../testing/test-dependencies";
-import {setDependencies} from "../dependencies";
+import * as dependencies from "../dependencies";
 import {registerAllWidgetsForTesting} from "../util/register-all-widgets-for-testing";
 import * as Widgets from "../widgets";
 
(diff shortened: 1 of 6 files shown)


packages/perseus/src/__tests__/widgets.test.ts
```diff
@@ -1,5 +1,5 @@
 import {testDependencies} from "../../../../testing/test-dependencies";
-import {setDependencies} from "../dependencies";
+import * as dependencies from "../dependencies";
```
jeremywiebe, line 2: **nit**: I think everywhere that we use `import *` we name the import using a capital. 🤷‍♂️  

```suggestion
import * as Dependencies from "../dependencies";
```

packages/perseus/src/logging/log.test.ts
```diff
@@ -16,10 +19,13 @@ describe("Perseus logging", () => {
         expect(logSpy).toHaveBeenCalledWith("test message", {a: 1, b: "two"});
     });
 
-    it("should proxy error() calls to the logger provided through setDependencies", () => {
+    it("should proxy error() calls to the logger obtained from getDependencies", () => {
```
jeremywiebe, line 22: Thanks!  That's what we were _actually_ testing. :)

PR #1277, opened 2024-05-15
Show error if locked line has length 0

nishasy:
## Summary:
It doesn't make sense for a locked line to have length 0, so
I'm adding an error message if the user were to set the two 
points to the same value.

Implementation notes/questions:
- I really tried to use WB TextField `validate` for this, but it
  was extremely convoluted to get the four text fieds to stay in
  sync. Or even to keep two of them in sync. If it has the error
  in one, but you update it in the other point, it would just
  not update the original error. I sort of used a manual state
  for this instead and replicated the style. It's much easier
  and cleaner this way.
- This implementation does not stop the graph from receiving
  a line with the same two defining points. Is that something
  worth doing? I'd have to change where the error gets set,
  but it's doable.

Issue: https://khanacademy.atlassian.net/browse/LEMS-1991

## Test plan:
`yarn jest packages/perseus-editor/src/components/__tests__/locked-line-settings.test.tsx`

Storybook
http://localhost:6006/?path=/story/perseuseditor-components-locked-line-settings--controlled

## Demo

https://github.com/Khan/perseus/assets/13231763/c4ddf49a-ee2f-4b7f-a51c-105c4fae8800


jeremywiebe (commented):
I don't have any concerns with the code, but showing 2 errors feels off to me. It also causes quite a bit of junk as they are shown/hidden.  

Also, the editor system has a way to generate errors to block saving via the `getSaveWarnings()` function.  Any widget editor can implement that and return an array of strings representing the issues.  Webapp will then prevent saving if the top-level function returns any issues. I'd recommend doing that here so that we do more than simply render an error, we actually block saving.

nishasy:
@jeremywiebe 

> showing 2 errors for the same issue feels off to me. It also causes quite a bit of junk as they are shown/hidden. I wonder if we could put the error above the two points

That makes sense to me. I'll move the error. Do you think it still makes sense to highlight those four text fields in red so it's apparent which inputs are related to the error?

jeremywiebe:
> @jeremywiebe
> 
> > showing 2 errors for the same issue feels off to me. It also causes quite a bit of junk as they are shown/hidden. I wonder if we could put the error above the two points
> 
> That makes sense to me. I'll move the error. Do you think it still makes sense to highlight those four text fields in red so it's apparent which inputs are related to the error?
> 
> image

Nice.  I think that's much better.  It still highlights the fields that contribute to the error state, but it's not so "loud".  :)

packages/perseus-editor/src/components/locked-line-settings.tsx
```diff
@@ -26,6 +26,8 @@ import type {
     LockedPointType,
 } from "@khanacademy/perseus";
 
+const lengthZeroStr = "The line cannot have length 0.";
```
benchristel, line 29: This string doesn't need to be added to `PerseusStrings` for translation, does it? Do we have a policy of not doing i18n for the editor?

benchristel (approved):
LGTM! Just one question inline about i18n.

packages/perseus-editor/src/components/locked-line-settings.tsx
nishasy: I did a quick search, and it doesn't look like any of the editor strings are translated anywhere else.

jeremywiebe: Confirmed. We don't translate the content editor.

PR #1278, opened 2024-05-16
Alphabetically order locked line kind dropdown options

nishasy:
## Summary:
This is just a nit. Putting Ray before Segment makes it so that
1. The options are in alpahbetical order
2. The options are in length order

Issue: none

## Test plan:
Storybook
http://localhost:6006/?path=/story/perseuseditor-editorpage--mafs-with-locked-figures

Screenshot 2024-05-16 at 10 29 42 AM

### DIFF
--- a/.changeset/rare-oranges-count.md
+++ b/.changeset/rare-oranges-count.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": patch
+---
+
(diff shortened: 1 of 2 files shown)


benchristel (approved):
LGTM!

jeremywiebe (approved):

PR #1280, opened 2024-05-16
Bugfix: first added locked figure settings accordion should be expanded on add

nishasy:
## Summary:
There was an issue where the first added locked figure settings would be collapsed, but all the settings are supposed to be expanded when added. This was happening because the initial expanded states array had one element in it when it was supposed to be empty. 

Issue: https://khanacademy.atlassian.net/browse/LEMS-1992

## Before

https://github.com/Khan/perseus/assets/13231763/4a1423d7-6f14-46dc-8431-f98ac8c64efe

## After

https://github.com/Khan/perseus/assets/13231763/2ab6a799-4216-4b29-9e3c-f3351b94dc32

## Test plan:
Storybook
- Add an interactive graph widget
- Use a segment graph
- Click "Add locked element"
- Select any element
- The settings accordion should be expanded


packages/perseus-editor/src/components/locked-figures-section.tsx
```diff
@@ -25,9 +25,12 @@ type Props = {
 const LockedFiguresSection = (props: Props) => {
     // Keep track of all figures' accordions' expanded states for the
     // expand/collapse all button. Set the whole array to false initially.
-    const [expandedStates, setExpandedStates] = React.useState(
-        Array(props.figures?.length).fill(false),
-    );
+    const collapsedStateArray =
+        props.figures?.length && props.figures.length > 0
```
Myranae, line 29: Question: Wouldn't the first part of this check be sufficient? Undefined would be falsy, so if `figures` is undefined, this equals the empty array, but also if it is zero it is falsy. So if there is no `figures` defined or if `length` is zero, it's falsy. So the only way `props.figures?.length` can be truthy is if it exists and is greater than 0.

Myranae (approved):
Looks good! I was able to follow the testing steps and saw the accordion starts expanded when locked figures are added. Added one question, but I don't think it's a blocker. Thanks! :)

packages/perseus-editor/src/components/locked-figures-section.tsx
nishasy: That's what I thought! My guess is that it saw the length is not null, and so it evaluated to `true`? which also = 1?

Would love if someone with better knowledge of JavaScript had some insight into this. (Maybe @benchristel?)

jeremywiebe: > Wouldn't the first part of this check be sufficient?

This is subjective, but I'm not a fan of using bare numbers for truthy checks (`if props.figures?.length`). I find them tricky to reason about. 

@Myranae is right about `0` being falsey though.

image


I _think_ I'd re-arrange this slightly and avoid needing to know the details of what is and isn't falsy/truthy in JS. 

```js
    const collapsedStateArray =
        Array((props.figures ?? []).length)
            .fill(false);
```

Myranae: @jeremywiebe So if it's an empty array that you try to fill with `false`, it will just stay an empty array, right?

jeremywiebe: Correct.  `array.fill` with one argument replaces all instances in the array with the provided argument.  

image

nishasy: ah, that's much more elegant. thank you!

PR #1282, opened 2024-05-16
Fix low color contrast issues across several widgets

Myranae:
## Summary:
Worked on as part of GAAD. 
The issue here was specific to the low contrast found for the dot and line in number-line. This fix changes the KhanColors object in Perseus which is used across a number of widgets. The changes here specifically affect grapher, number-line, interaction, and interactive-graph (based on the updated snapshots).

[Discussion about what colors to use. Includes input from a designer](https://khanacademy.slack.com/archives/C01AZ9H8TTQ/p1715881111389849?thread_ts=1715880696.430059&cid=C01AZ9H8TTQ)

Before:
image
image
image

After:
image
image
image

Here's also some manual testing of how it looks in webapp (test everything course). Checked the other three types of graphs also affected by this as well.

Before:
image

After:
image

Issue: LEMS-252

## Test plan:
- Check all the widgets that were affected in Storybook and confirm the colors have at least 3:1 contrast with surroundings
- Also confirm the designs are not negatively affected
- Confirm all checks pass


packages/perseus/src/util/colors.ts
```diff
@@ -81,13 +83,9 @@ const KhanColors = {
...
     KA_GREEN: "#71B307",
-    // Don't actually use _BACKGROUND! Make things transparent instead. The
-    // background color used in exercises is subject to change at the whim
-    // of any redesigns.
-    _BACKGROUND: "#FDFDFD", // TODO(eater): Get rid of this altogether.
-    INTERACTING: "#71B307", // KA_GREEN
-    INTERACTIVE: "#71B307", // KA_GREEN
-    DYNAMIC: "#6495ED", // BLUE
+    INTERACTING: color.green,
+    INTERACTIVE: color.green,
```
Myranae, line 87: Could also just combine these into one since they are both the same color. Thoughts?

packages/perseus/src/util/colors.ts
```diff
@@ -81,13 +83,9 @@ const KhanColors = {
     GRAY_I: "#21242C",
     KA_BLUE: "#314453",
     KA_GREEN: "#71B307",
-    // Don't actually use _BACKGROUND! Make things transparent instead. The
-    // background color used in exercises is subject to change at the whim
-    // of any redesigns.
-    _BACKGROUND: "#FDFDFD", // TODO(eater): Get rid of this altogether.
```
Myranae, line 87: Got rid of this part since both comments seem to think it is not needed. Replaced where they are used with just `"transparent"`.

catandthemachines:
Do you have screen shots of the UX changes with the new colors? :)

catandthemachines:
Were you able to run an accessibility scan and confirm the color doesn't have any more issues?

catandthemachines:
Not sure how it works here just yet, but were you able to confirm or check with a designer on these new colors?

Myranae:
> Do you have screen shots of the UX changes with the new colors? :)

Ooo, thank you! Will get these added :D

Myranae:
> Were you able to run an accessibility scan and confirm the color doesn't have any more issues?

@catandthemachines Hmm, I'm not sure about an accessibility scan. How would I do that and what other issues are possible? I did triple check the ratio was better. I noticed the ratio between the lines and the dots might not be ideal, but not sure if that is fixable since the dot color has to have good contrast with the background (and so does the line).

Myranae:
> Not sure how it works here just yet, but were you able to confirm or check with a designer on these new colors?

@catandthemachines I added a link to the PR summary that leads to the conversation I had with the team and a designer about what colors to use. Planning to make a ZND and check with our QE person on the change as well. (Before merging but after things slow down a bit).

catandthemachines:
> > Were you able to run an accessibility scan and confirm the color doesn't have any more issues?
> 
> @catandthemachines Hmm, I'm not sure about an accessibility scan. How would I do that and what other issues are possible? I did triple check the ratio was better. I noticed the ratio between the lines and the dots might not be ideal, but not sure if that is fixable since the dot color has to have good contrast with the background (and so does the line).

There are a lot of great accessibility scan tools for Chrome browsers (I use Accessibility Insights when I want to confirm is my site has any issues or not https://chromewebstore.google.com/detail/accessibility-insights-fo/pbjjkligggfmakdaogkfomddhfmpjeni ). For the question on the dot vs. the line, that might be a great question for Level Access to give us suggestions. I can't remember off the top of my head but I believe distinguishing content in a visual diagram has different color contrast requirements than text vs. background.

catandthemachines (approved):

packages/perseus/src/util/colors.ts
Myranae: Hmm, there are 45 instances of INTERACTIVE and 26 instances of INTERACTING all across 13 files. I would probably make `INTERACTING` just be `INTERACTIVE`, so it would only add 5 files with changes. 

I don't think it would have wider consequences since both are using the same color, though I don't know if the change in meaning would be troublesome. 

Perhaps this should be part of a wider change of updating `KhanColors` at some point and saved for another time since it has so many instances. I think I assumed it would be a smaller change. 😅

catandthemachines: Yeah in that case I would hold off and just submit a backlog task to revisit in the future 😉

packages/perseus/src/widgets/number-line.tsx
```diff
@@ -472,25 +472,25 @@ class NumberLine extends React.Component<Props, State> {
         // it can't be interacted with.
         let fill;
         if (isOpen) {
-            fill = KhanColors._BACKGROUND;
+            fill = "transparent";
         } else if (props.static) {
-            fill = KhanColors.DYNAMIC;
+            fill = KhanColors.BLUE;
```
nishasy, line 477: How would you feel about replacing this (and other uses below) with WB `color.blue` and not importing `KhanColors` at all where possible? I think that would cut out the intermediate step and make it clear exactly where the color is coming from.

nishasy (approved):
Left a question/suggestion about using WB color instead of KhanColors in the files directly, but otherwise LGTM!

packages/perseus/src/widgets/number-line.tsx
Myranae: I think I may have used some of my understanding of backend coding here to make this decision, but I think it might not be the same for frontend. Let me see what you think. I was thinking having one place where we import WB colors would be more efficient than importing it in all the places the colors are used, specifically because WB is an external library. So we're only doing the import here and not all over. Maybe the import still ends up happening in all the places anyway as each time it goes to to the KhanColors file and it ends up importing from WB at that point anyway. 🤔

PR #1285, opened 2024-05-16
Clarify the code for initializing linear graphs

benchristel:
## Summary:
Previously, the code for initializing movable lines looked like it would
use `graph.coords` if that was provided, and fall back to default values
if not. That's not what it did, because the `??` operator has higher
precedence than the ternary operator. If any coords were passed, we'd
treat the graph as a linear-system graph.

I suspect the reason we didn't notice this bug is that `coords` is never
provided for `linear` or `ray` graphs. Linear and ray graphs always have
one line, and linear-system graphs always have two lines.

Issue: none

Test plan:

- `yarn dev`
- Go to `http://localhost:5173#flipbook`
- View the examples obtained by running each of:
  ```
  ag -l '"type":"linear"' data/questions/ | xargs cat | grep -v 'grapher 1' | pbcopy
  ag -l '"type":"ray"' data/questions/ | xargs cat | pbcopy
  ag -l '"type":"linear-system"' data/questions/ | xargs cat | pbcopy
  ```
  (run these one at a time; each one overwrites your clipboard)
- Verify that none of the Mafs graphs differ from the legacy version.

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-state.test.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-state.test.ts
@@ -117,13 +117,27 @@ describe("initializeGraphState for linear-system graphs", () => {
             ...baseGraphData,
             graph: {
                 type: "linear-system",
-                coords: [[[1, 2], [3, 4]]],
+                coords: [
+                    [
+                        [1, 2],
+                        [3, 4],
+                    ],
+                ],
             },
         });
 
         invariant(state.type === "linear-system");
-        expect(state.coords).toEqual([[[-5, 5], [5, 5]], [[-5, -5], [5, -5]]])
-    })
+        expect(state.coords).toEqual([
+            [
+                [-5, 5],
+                [5, 5],
+            ],
+            [
+                [-5, -5],
+                [5, -5],
+            ],
+        ]);
+    });
 });
 
 describe("initializeGraphState for polygon graphs", () => {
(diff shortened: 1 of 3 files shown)


packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-state.ts
```diff
@@ -295,13 +295,20 @@ const getLineCoords = ({
...
-}: getLineCoordsArg): PairOfPoints[] =>
-    // Return two lines for a linear system, one for a ray or linear
-    graph.coords ?? graph.type === "linear-system"
-        ? defaultLinearCoords.map((collinear) =>
-              normalizePoints(range, step, collinear),
-          )
-        : [normalizePoints(range, step, defaultLinearCoords[0])];
+}: getLineCoordsArg): PairOfPoints[] => {
+    //  Return two lines for a linear system, one for a ray or linear
+    switch (graph.type) {
```
jeremywiebe, line 300: Nice. It makes it much clearer depending strictly on the `graph.type`.

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-state.test.ts
```diff
@@ -109,6 +109,21 @@ describe("initializeGraphState for line graphs", () => {
             ],
         ]);
     });
+
+    it("ignores any provided coords", () => {
+        // This is a characterization test. I don't know if it's desired
+        // behavior, but it's the existing behavior.
```
jeremywiebe, line 115: 😢  I'm pretty sure this is a byproduct of the way that our editors work. They serialize the props passed into the graph component and don't always strip out unnecessary data. Thus we have `.graph.coords` that is populated, when we really only care about `.correct.coords` (the rubric). 

We also automatically derive the starting location of the movable lines and so that's why the coords in the data are ignored.

jeremywiebe (approved):

PR #1289, opened 2024-05-16
Replace interactive graph snapshot tests with Storybook stories

benchristel:
## Summary:
There is a problem with using Jest snapshot tests to test Mafs: Mafs
uses globally autoincrementing IDs for some of its elements, so when
new tests are added to a suite, the IDs shift and all later snapshot
tests fail.

To work around this, I've removed the problematic Jest snapshot tests
and replaced them with Storybook stories.

These stories will be run as visual snapshot tests on Chromatic,
providing similar regression protection with hopefully less test
flakiness.

Issue: LEMS-1975

Test plan:

Watch CI run and accept the baselines for the new stories on Chromatic.


Myranae (approved):
Looks good~ I see the regression tests in the UI Tests area of the PR checks. So after this, the goal is for everyone to not make any more Jest snapshot tests but instead use these Storybook regression tests, right? Also, looks like some checks are failing, but they don't seem to be related to this PR specifically.
image
image

benchristel:
> the goal is for everyone to not make any more Jest snapshot tests but to instead use these Storybook regression tests, right?

You can write Jest snapshot tests for individual components. It's just the main `MafsGraph` component that's the problem — `MafsGraph` renders the axis grid, which uses autoincrementing IDs.

jeremywiebe:
I would love (or would have loved) to find a different solution for this. This means folks now have to be aware that some tests can use Jest snapshot tests and some Storybook visual snapshot tests.  

I know Jest supports custom [Property Matchers](https://jestjs.io/docs/snapshot-testing#property-matchers) for snapshots, but I haven't seen how to configure this globally.

PR #1290, opened 2024-05-17
Upgrade to @khanacademy/wonder-blocks-timing@5 everywhere

benchristel:
## Summary:
Previously, we were using version 4 in the dev UI, which caused
TypeScript errors on CI. TypeScript was finding the types for v5 when it
should have been using the types for v4.

Issue: LEMS-2002

Test plan:

View an interactive graph in the dev UI flipbook and use the "check
answer" button. It should pop up a banner that goes away after a couple
seconds.

### DIFF
--- a/dev/flipbook.tsx
+++ b/dev/flipbook.tsx
@@ -5,7 +5,7 @@ import {View} from "@khanacademy/wonder-blocks-core";
 import IconButton from "@khanacademy/wonder-blocks-icon-button";
 import {Strut} from "@khanacademy/wonder-blocks-layout";
 import Link from "@khanacademy/wonder-blocks-link";
-import {useTimeout} from "@khanacademy/wonder-blocks-timing";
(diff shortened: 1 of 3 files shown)


dev/flipbook.tsx
```diff
@@ -267,7 +267,9 @@ function GradableRenderer(props: QuestionRendererProps) {
     const rendererRef = useRef<Renderer>(null);
     const [score, setScore] = useState<PerseusScore>();
 
-    useTimeout(() => setScore(undefined), 2500, !!score);
+    const clearScoreTimeout = useTimeout(() => setScore(undefined), 2500, {
+        schedulePolicy: SchedulePolicy.OnDemand,
```
jeremywiebe, line 271: 🤔  I noticed this too where I used `useTimeout`... it would have been nice for it to start automatically.

jeremywiebe (approved):

PR #1293, opened 2024-05-17
Fill interactive circles with blue on hover

benchristel:
## Summary:
Designs: https://www.figma.com/design/JroPokeCLBv2VRIw3GPvY5/LC---Interactive-graph-widget?node-id=1150-51899&t=kJh1CRKGniWKyer1-0

Issue: LEMS-1978

Test plan:

Visit http://localhost:5173 and turn on Mafs for the circle graph.

The circle should light up blue when hovered.

Note that there is currently an issue where axis numbers intercept the
hover gesture, so the circle will not be blue if your cursor is over an
axis number. This will be fixed in LEMS-2003.

Screen Shot 2024-05-17 at 1 10 41 PM

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/mafs-styles.css
+++ b/packages/perseus/src/widgets/interactive-graphs/mafs-styles.css
@@ -152,6 +152,17 @@
     cursor: grab;
 }
 
+.MafsView .movable-circle .circle {
+    stroke: var(--mafs-blue);
+    stroke-width: 2px;
+    fill: transparent;
+}
+
+.MafsView .movable-circle:hover .circle {
+    fill: var(--mafs-blue);
+    fill-opacity: 0.16;
+}
+
 .MafsView .movable-circle.movable-circle--dragging {
     cursor: grabbing;
     outline: none;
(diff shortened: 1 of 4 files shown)


packages/perseus/src/widgets/interactive-graphs/mafs-styles.css
```diff
@@ -152,6 +152,17 @@
     cursor: grab;
 }
 
+.MafsView .movable-circle .circle {
+    stroke: var(--mafs-blue);
```
Myranae, line 156: Question: Was the color here decided in the design document or is that just what we're using for mafs graphs? Not WB color tokens?

Myranae (approved):
Looks great! Just as you described. Hover changes the color to blue except when over the numbers. Thanks for setting this up! Left a non-blocking question.

packages/perseus/src/widgets/interactive-graphs/mafs-styles.css
benchristel: We've been using this `--mafs-blue` variable in the CSS for Mafs components. It's the same as WB blue. We can't use WB tokens directly in CSS because (AFAIK?) WB only exports them as JavaScript.

Myranae: Thanks! Good to know it's the same as WB blue :)

PR #1295, opened 2024-05-20
Fix a literal corner case in intersecting a ray with a box

benchristel:
## Summary:
The bug: the arrowhead of a drawn ray would always be at the origin when
floating-point imprecision caused the computed intersection with the
graph bounds to be just outside the actual bounds in both dimensions.
This happened when the ray was pointed directly at one of the corners.

Issue: LEMS-2004

Test plan:

Follow the repro steps in the linked issue. The bug should not repro.

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/utils.test.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/utils.test.ts
@@ -18,7 +18,7 @@ describe("getIntersectionOfRayWithBox", () => {
         expect(intersection).toEqual([0, 0]);
     });
 
-    test("given a horizontal ray passing through the origin to the left", () => {
+    test("a horizontal ray passing through the origin to the left", () => {
         const box: [Interval, Interval] = [
             [-7, 7],
             [-11, 11],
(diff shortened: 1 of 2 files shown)


packages/perseus/src/widgets/interactive-graphs/graphs/utils.test.ts
```diff
@@ -3,22 +3,7 @@ import {getIntersectionOfRayWithBox} from "./utils";
...
-        const throughPoint: vec.Vector2 = [3, 5];
-        const intersection = getIntersectionOfRayWithBox(
-            initialPoint,
-            throughPoint,
-            box,
-        );
-        expect(intersection).toEqual([0, 0]);
-    });
-
-    test("given a horizontal ray passing through the origin to the left", () => {
```
benchristel, line 21: This removed test was testing an invalid case. I'm okay leaving that behavior undefined.

packages/perseus/src/widgets/interactive-graphs/graphs/utils.ts
```diff
@@ -42,10 +42,7 @@ export const getIntersectionOfRayWithBox = (
             return [xAtYExtreme, yExtreme];
 
         default:
-            // This default case is only reachable if the input is invalid
-            // (initialPoint is outside the graph bounds, or initialPoint and
-            // throughPoint are the same).
-            return [0, 0];
+            return [xExtreme, yExtreme];
```
benchristel, line 45: I don't know why codecov says this line isn't covered. It definitely is covered (by the test I added). If I remove it, the test fails.

jeremywiebe, line 45: Just so I understand, the test is verifying when floating point is imprecise and the `default` here is hit because of the imprecise floating point calculations?

jeremywiebe (approved):

PR #1296, opened 2024-05-21
Try to force Node v20

jeremywiebe:
## Summary:

We recently moved to Node v20. In doing so, we added an `engines` key to our root `package.json`. This file is only used for local dev and controlling tooling versions and has nothing to do with the packages we publish. 

An example of a failed run looks like this: 

image

My working theory is that this job is getting the default Node version for the `ubuntu-latest` image (full list of installed software is [here](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#language-and-runtime)).  

In this PR I've added an explicit step that uses the `actions/setup-node` action and forces it to Node v20.  This seems to fix the situation (and align with my working theory of why it was broken). 

Issue: n/a

## Test plan:

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -30,7 +30,7 @@ jobs:
                   fetch-depth: 0
 
             - name: Force Node version
-              uses: actions/setup-node
+              uses: actions/setup-node@v4
               with:
                   node-version: ${{ matrix.node-version }}
                   cache: yarn
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -29,6 +29,12 @@ jobs:
               with:
                   fetch-depth: 0
 
+            - name: Force Node version
+              uses: actions/setup-node
+              with:
+                  node-version: ${{ matrix.node-version }}
+                  cache: yarn
+
             - name: Get changed files
               uses: Khan/actions@get-changed-files-v2
               id: changed


kevinbarabash (approved):
LGTM 🚢

kevinbarabash:
I missed Jeff's comment in Slack before I approved this so I've re-requested a review from myself:
> I would recommend not changing engines settings like that yet if you can help it. I’m working on that task right now and to avoid mishaps, I need to get some safety checks in webapp first.
Our E2E tests run on Node 17 for the time being and any dependencies that our E2E test infrastructure uses must work with Node 17 so engine rules will break that and stop them running at all.
I am working on putting in some safety checks to avoid this.

.github/workflows/node-ci.yml
```diff
@@ -29,6 +29,12 @@ jobs:
               with:
                   fetch-depth: 0
 
+            - name: Force Node version
+              uses: actions/setup-node@v4
+              with:
+                  node-version: ${{ matrix.node-version }}
+                  cache: yarn
```
somewhatabstract, line 36: **praise:** Looks good to me. I suspect the ubuntu images have a default version of Node installed. Being explicit with `setup-node` is smart.

somewhatabstract (approved):
LGTM

.github/workflows/node-ci.yml
jeremywiebe: You're right.  Once we move to ubuntu 24 (which is in Beta) it'll default to Node 20. But I suspect we still want to define the Node version explicitly. 

https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#language-and-runtime

PR #1298, opened 2024-05-22
Use getSaveWarnings to stop 0 length locked line from saving

nishasy:
## Summary:
Right now, setting the locked line's points to be the same
will show an error, but it won't stop the exercise from
being able to be saved.

By using `getSaveWarnings` here, we can do that.

This PR includes:
- Storybook story that shows the save warnings on the side
- Util function for checking locked points + tests
- Util function being used in `getSaveWarnings()` as well as inside locked line settings

Issue: https://khanacademy.atlassian.net/browse/LEMS-2008

## Test plan:
Storybook 
- http://localhost:6006/?path=/story/perseuseditor-editorpage--with-save-warnings
- Set the points to be the same in the locked line settings
- Make sure that a save warning shows up in the save warning box

Screenshot 2024-05-21 at 5 14 05 PM

### DIFF
--- a/packages/perseus-editor/src/components/util.ts
+++ b/packages/perseus-editor/src/components/util.ts
@@ -80,3 +80,13 @@ export function getDefaultFigureForType(type: LockedFigureType): LockedFigure {
             throw new UnreachableCaseError(type);
     }
 }
+
+export function lockedPointsEqual(
+    point1: LockedPointType,
+    point2: LockedPointType,
+) {
+    return (
+        point1.coord[0] === point2.coord[0] &&
+        point1.coord[1] === point2.coord[1]
+    );
+}
--- a/packages/perseus-editor/src/components/util.ts
+++ b/packages/perseus-editor/src/components/util.ts
@@ -80,13 +80,3 @@ export function getDefaultFigureForType(type: LockedFigureType): LockedFigure {
             throw new UnreachableCaseError(type);
     }
 }
-
-export function lockedPointsEqual(
-    point1: LockedPointType,
-    point2: LockedPointType,
-) {
-    return (
-        point1.coord[0] === point2.coord[0] &&
-        point1.coord[1] === point2.coord[1]
-    );
-}
--- a/packages/perseus-editor/src/widgets/interactive-graph-editor.tsx
+++ b/packages/perseus-editor/src/widgets/interactive-graph-editor.tsx
@@ -22,6 +22,7 @@ import InteractiveGraphSettings from "../components/interactive-graph-settings";
 import LabeledRow from "../components/labeled-row";
 import LockedFiguresSection from "../components/locked-figures-section";
 import SegmentCountSelector from "../components/segment-count-selector";
+import {lockedPointsEqual} from "../components/util";
 import {parsePointCount} from "../util/points";
 
 import type {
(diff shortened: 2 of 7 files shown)


packages/perseus-editor/src/components/util.ts
```diff
@@ -80,3 +80,13 @@ export function getDefaultFigureForType(type: LockedFigureType): LockedFigure {
...
 }
+
+export function lockedPointsEqual(
+    point1: LockedPointType,
+    point2: LockedPointType,
+) {
+    return (
+        point1.coord[0] === point2.coord[0] &&
+        point1.coord[1] === point2.coord[1]
+    );
```
nishasy, line 91: Why is it saying line 90 isn't covered by tests?

packages/perseus-editor/src/widgets/interactive-graph-editor.tsx
```diff
@@ -661,6 +662,24 @@ class InteractiveGraphEditor extends React.Component<Props> {
...
+        // A locked line on the graph cannot have length 0.
+        if (this.props.lockedFigures) {
+            _.each(this.props.lockedFigures, (figure) => {
+                if (
+                    figure.type === "line" &&
+                    lockedPointsEqual(figure.points[0], figure.points[1])
+                ) {
+                    issues.push("The line cannot have length 0.");
+                }
+            });
```
jeremywiebe, line 678: We try to avoid using underscore if there's a native equivalent. 

```suggestion
            this.props.lockedFigures
                .filter((f) => f.type === "line")
                .forEach((figure) => {
                    if (
                        lockedPointsEqual(figure.points[0], figure.points[1])
                    ) {
                        issues.push("The line cannot have length 0.");
                    }
                });
```

jeremywiebe (approved):
This looks great, the only thing I'd suggest changing is that I don't think you need the new function you added (`lockedPointsEqual`).

packages/perseus-editor/src/components/util.ts
```diff
@@ -80,3 +80,13 @@ export function getDefaultFigureForType(type: LockedFigureType): LockedFigure {
             throw new UnreachableCaseError(type);
     }
 }
+
+export function lockedPointsEqual(
```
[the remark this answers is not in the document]
nishasy: Ah, I created this so I could pass in a whole LockedPoint without having to keep doing  `point1.coord[0]` etc. I can still remove it if you'd prefer.

packages/perseus-editor/src/components/util.ts
jeremywiebe: Strange. I'm not sure. Codecov gets its coverage info from the test runs we upload, so somewhere that data must be wrong. :/ When I run `yarn coverage` locally and then `open coverage/lcov-report/index.html`, I can see that the line is covered.

packages/perseus-editor/src/components/util.ts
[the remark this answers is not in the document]
jeremywiebe: 🤔  To be pedantic, we're really checking only that the points are at the same place, not that the points are entirely the same (color, filledness).  So if you used one of the suggestions you could just call them with `kvector.equal(point1.coord, point2.coord)`.  

But this isn't a deal-breaker. Whatever you decide is totally fine. Go with your gut. :)

packages/perseus-editor/src/widgets/interactive-graph-editor.tsx
nishasy: I have NO IDEA how I ended up using `_.each` 🤦🏽
I'm going to use a for of loop since that's what I would have actually done had I written this properly myself in the first place.

PR #1299, opened 2024-05-22
Minor CSS change to ensure that interactive elements are rendered above our labels

SonicScrewdriver:
## Summary:
<Add your summary here>

Issue: LEMS-2009

## Test plan:

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/mafs-styles.css
+++ b/packages/perseus/src/widgets/interactive-graphs/mafs-styles.css
@@ -211,6 +211,8 @@
      * drag interactions on the graph. */
     user-select: none;
     pointer-events: none;
+    /* Prevent labels from covering up interaction points. */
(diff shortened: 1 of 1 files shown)


benchristel (approved):
LGTM. One thought: I notice the gridlines now intersect the axis label numbers. Did we previously have some whitespace around the numbers?

SonicScrewdriver:
> LGTM. One thought: I notice the gridlines now intersect the axis label numbers. Did we previously have some whitespace around the numbers?

They always intersected, but I definitely counted it as a bug that I'm trying to solve in another PR! 

Screenshot 2024-05-22 at 11 08 54 AM

When the labels were svg text, it was easy to add a white stroke. I'm currently just finishing up a test with text-shadow to achieve the same thing on the new label approach + upcoming darker grid lines, and I'm hoping to have a PR up by EOD.

PR #1301, opened 2024-05-22
Remove unused APIOptions and types

jeremywiebe:
## Summary:

This PR removes some code/APIOptions that are completely unused today. To verify, I checked the webapp and mobile source code. I also checked and these options are not used in Perseus so even if they were set by calling applications, they do nothing. 

It's time to go!

  * Remove Empty type as its now centrally defined in `utility.d.ts` which is symlinked into each package src
  * Remove unused and deprecated `inModal` APIOption
  * Remove unused and deprecated `useDraftEditor` APIOption

Issue: "none"

## Test plan:

### DIFF
--- a/.changeset/brave-lions-melt.md
+++ b/.changeset/brave-lions-melt.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": major
+---
+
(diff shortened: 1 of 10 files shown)


benchristel (approved):
Yay! Thanks for cleaning these up.

PR #1304, opened 2024-05-22
New Perseus Preview component

jeremywiebe:
## Summary:

Today, the Perseus editor preview component uses a complex `<iframe />` based system that depends on a live website providing the iframe source. This makes it challenging to mimic the preview system in Storybook and has its own set of issues in production. 

Added to that, the mechanism to communicate between the Perseus Editor and the preview frame is not at all easy to understand and is split between this repo (Khan/perseus) and our main webapp repo. This makes maintaining the preview system tricky and error prone (basically, we don't touch it!) 

This PR introduces a new preview component that simply renders the `ServerItemRenderer` using the current PerseusItem. This eliminates all of the cross-frame communication that was needed and seems to work just as good as the iframe system. 

A few notes before we commit to this:

  * We need preview components that support hints and articles (either we make the preview component more flexible or render different preview components where they're needed - I think I'd lean towards the latter as I don't think there's a need to have One Master Preview Component To Rule Them All™️ )
  * The `expression` widget when simulating mobile can display the math keypad, but it doesn't automatically dismiss when the field is blurred. In testing, Nisha and I found that this is actually semi-broken in production anyways so this wouldn't be a regression. 
  * The old iframe preview system uses a helper tool from the [HammerJS library](hammerjs.github.io/touch-emulator/), which is vendored in, to simulate touches when previewing for Phone and Tablet. This isn't possible with a non-iframe preview system because that library "mutates" the global env, so we'd have to reload the page to switch between preview devices. I think a way forward might be to change the math keypad to support Pointer Events instead of being so touch-centric. 

Issue: LEMS-1809

## Test plan:

`yarn start`
Navigate to the `Editor Page` story and add any widgets.  The preview should update in near realtime as you edit the content and widget configs.

### DIFF
--- a/packages/perseus-editor/src/components/device-framer.tsx
+++ b/packages/perseus-editor/src/components/device-framer.tsx
@@ -60,7 +60,7 @@ const DeviceFramer = ({
     );
 
     if (nochrome) {
-        // Render content inside a variable height iframe.  Used on the
+        // Render content inside a variable height div.  Used on the
         // "edit" table of the content editor. In this mode, PerseusFrame
         // will draw the border and reserve space on the right for
         // lint indicators.
--- a/packages/perseus-editor/src/content-renderer.tsx
+++ b/packages/perseus-editor/src/content-renderer.tsx
@@ -1,0 +1,50 @@
+import {
+    KeypadContext,
+    MobileKeypad,
+    StatefulKeypadContextProvider,
+} from "@khanacademy/math-input";
+import {Renderer} from "@khanacademy/perseus";
+import {View} from "@khanacademy/wonder-blocks-core";
+import * as React from "react";
+
+// eslint-disable-next-line import/no-relative-packages
+import {mockStrings} from "../../perseus/src/strings";
+
+import type {APIOptions, PerseusRenderer} from "@khanacademy/perseus";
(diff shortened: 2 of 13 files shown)


mpolyak:
Very cool, thank you for working on this!

In this approach of rendering `ServerItemRenderer` directly, would we still be able to highlight content lint warnings/errors in the editor?

jeremywiebe:
> Very cool, thank you for working on this!
> 
> In this approach of rendering `ServerItemRenderer` directly, would we still be able to highlight content lint warnings/errors in the editor?

Yes, linting, good call!  I've added that to the PR description as something we'd need to figure out!

benchristel:
@jeremywiebe I'm going to close this because it's more than a year old and has tons of merge conflicts. Feel free to reopen if you want to keep it around.

jeremywiebe:
@benchristel good idea.  In fact, based on the [proposal document](https://khanacademy.atlassian.net/wiki/spaces/~505982898/blog/2025/07/16/4146200767/Enabling+flexible+Perseus+editing#Separate-Out-Editor-and-Preview-Components) for handling better layouts in editors, we likely won't go this route anyways.

PR #1305, opened 2024-05-22
Add the `quadratic` question type for the new Mafs-based Interactive Graph

SonicScrewdriver:
## Summary:
This PR adds the quadratic question type to the new Mafs-based Interactive Graph widget. It should function identically to the original implementation within Interactive Graph (with 3 control points). 

**Note:** [There is a bug](https://khanacademy.atlassian.net/browse/LEMS-2014) related to the "getValidDestination" method that is incompatible with the innate keyboard support of Mafs. If a user tries to use the keyboard to move the control point to an invalid location, the control point will simply refuse to move further in that direction. As keyboard support is part of an upcoming stage of work on Interactive Graph, I have opted to leave this bug for now. I believe that we will need to find some way to differentiate between a mouse event and a keyboard event to solve this issue.

### User View: 
https://github.com/Khan/perseus/assets/12463099/bab8ef09-6bfb-4e96-89c4-ef8f3542fd32

### Content Creator View:
https://github.com/Khan/perseus/assets/12463099/ef74ccf7-e0a7-440f-a1cc-42eb77665045

Issue: LEMS-1831

## Test plan:
make tesc / yarn jest

### DIFF
--- a/packages/perseus/src/perseus-types.ts
+++ b/packages/perseus/src/perseus-types.ts
@@ -770,7 +770,7 @@ export type PerseusGraphTypePolygon = {
 export type PerseusGraphTypeQuadratic = {
     type: "quadratic";
     // expects a list of 3 coords
-    readonly coords: [Coord, Coord, Coord];
+    coords?: [Coord, Coord, Coord];
 } & PerseusGraphTypeCommon;
 
 export type PerseusGraphTypeSegment = {
--- a/packages/perseus/src/perseus-types.ts
+++ b/packages/perseus/src/perseus-types.ts
@@ -770,7 +770,7 @@ export type PerseusGraphTypePolygon = {
 export type PerseusGraphTypeQuadratic = {
     type: "quadratic";
     // expects a list of 3 coords
-    coords?: ReadonlyArray<Coord>;
+    readonly coords: [Coord, Coord, Coord];
 } & PerseusGraphTypeCommon;
 
 export type PerseusGraphTypeSegment = {
--- a/packages/perseus/src/widgets/interactive-graph.tsx
+++ b/packages/perseus/src/widgets/interactive-graph.tsx
@@ -34,6 +34,7 @@ import {getInteractiveBoxFromSizeClass} from "../util/sizing-utils";
 
 import {StatefulMafsGraph} from "./interactive-graphs";
 
+import type {QuadraticGraphState} from "./interactive-graphs/types";
 import type {Coord} from "../interactive2/types";
 import type {
     PerseusGraphType,
@@ -2216,7 +2217,7 @@ class InteractiveGraph extends React.Component<Props, State> {
         return InteractiveGraph.getQuadraticCoefficients(coords);
     }
 
-    static defaultQuadraticCoords(props: Props): ReadonlyArray<Coord> {
+    static defaultQuadraticCoords(props: Props): QuadraticGraphState["coords"] {
         const coords = [
             [0.25, 0.75],
             [0.5, 0.25],
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/quadratic.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/quadratic.tsx
@@ -33,12 +33,6 @@ export function QuadraticGraph(props: QuadraticGraphProps) {
     // Calculate the y value based on the current x value
     const y = (x) => (a * x + b) * x + c;
 
-    // Set the xy values for the Parametric plot
-    const xy = (x: number) => [x, y(x)] as vec.Vector2;
-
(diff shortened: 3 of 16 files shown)


jeremywiebe (commented):
Sorry, I know you said this was just to self-review... I got looking through it and had two tiny comments. :)

packages/perseus/src/perseus-types.ts
```diff
@@ -770,7 +770,7 @@ export type PerseusGraphTypePolygon = {
 export type PerseusGraphTypeQuadratic = {
     type: "quadratic";
     // expects a list of 3 coords
```
SonicScrewdriver, line 772: Updated as requested!

packages/perseus/src/widgets/interactive-graph.tsx
```diff
@@ -2216,7 +2217,7 @@ class InteractiveGraph extends React.Component<Props, State> {
         return InteractiveGraph.getQuadraticCoefficients(coords);
     }
 
-    static defaultQuadraticCoords(props: Props): ReadonlyArray<Coord> {
+    static defaultQuadraticCoords(props: Props): QuadraticGraphState["coords"] {
```
SonicScrewdriver, line 2220: This was just a type update to address PR feedback

packages/perseus/src/perseus-types.ts
```diff
@@ -770,7 +770,7 @@ export type PerseusGraphTypePolygon = {
...
-    coords?: ReadonlyArray<Coord>;
+    coords?: [Coord, Coord, Coord];
 } & PerseusGraphTypeCommon;
```
SonicScrewdriver, line 774: Updated type as requested! :)

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -184,8 +187,26 @@ function doMovePoint(
...
+                        bound({
+                            snapStep: state.snapStep,
+                            range: state.range,
+                            point: action.destination,
+                        }),
+                    ),
+                }),
+            };
+        }
+        case "quadratic": {
```
SonicScrewdriver, line 209: These had to be separated out due to the new type conflict as we're no longer using Coord[]

packages/perseus/src/widgets/interactive-graphs/graphs/quadratic.tsx
```diff
@@ -0,0 +1,120 @@
...
+        if (validDestination === false) {
+            return;
+        }
+
+        dispatch(movePoint(elementId, destination));
+    };
+
+    return (
+        <>
+            <Plot.Parametric xy={xy} t={t} color={color.blue} />
```
benchristel, line 58: Is `Plot.Parametric` needed here? Mafs also has a `Plot.OfX` component that just lets you pass a function to compute y from x. https://mafs.dev/guides/display/plots

packages/perseus/src/widgets/interactive-graphs/graphs/quadratic.tsx
```diff
@@ -0,0 +1,120 @@
...
+        (p3[0] * p3[0] * (p1[1] - p2[1]) +
+            p2[0] * p2[0] * (p3[1] - p1[1]) +
+            p1[0] * p1[0] * (p2[1] - p3[1])) /
+        denom;
+    const c =
+        (p2[0] * p3[0] * (p2[0] - p3[0]) * p1[1] +
+            p3[0] * p1[0] * (p3[0] - p1[0]) * p2[1] +
+            p1[0] * p2[0] * (p1[0] - p2[0]) * p3[1]) /
+        denom;
+    return [a, b, c];
```
benchristel, line 119: This seems extremely magical 😂 . Is this a widely known algorithm? It might be nice to include a comment linking to an explanation.

One thing that isn't immediately clear to me (okay, at all clear to me) from looking at this is whether the order of the provided `coords` matters. I believe it doesn't, but it would be cool to have a test demonstrating that.

benchristel (approved):
I left a few suggestions inline, which you can take or leave. LGTM!

packages/perseus/src/widgets/interactive-graphs/graphs/quadratic.tsx
SonicScrewdriver: (We've discussed this elsewhere, but I figured it's good to keep all the data on the ticket in case we ever search back.) 

I added a comment and a test for this to confirm that the order does not matter! 

I found the source for the original logic, which was created 11 years ago (!) : https://phabricator.khanacademy.org/D2413

When I have some spare time, I'll try to fully break down the math so that I can provide a clearer description of what is going on. (Mostly as a self-curiosity)

PR #1306, opened 2024-05-22
[Interactive Graph Editor] Move the 'Add locked figure' and 'Collapse/Expand all' buttons to the top

nishasy:
## Summary:
<Add your summary here>

Issue: XXX-XXXX

## Test plan:
todo


jeremywiebe:
I _think_ we settled that this would not be landed, right?  /cc @mark-fitzgerald

nishasy:
Turns out this might not be the preferred ordering of things. We might come back to this later, but we'll just leave it as for now.

PR #1307, opened 2024-05-23
Upgrade to Storybook 8 (latest)

jeremywiebe:
## Summary:

Storybook has been prompting and prodding that there's a new version out (8). So today I ran the upgrade script and tested it. There was one minor manual fix but otherwise it seems to "just work". 

  1. Ran: `npx storybook@latest upgrade`
  2. Removed `vite-plugin-turbosnap` plugin, this functionality is now built-in to Storybook and is accessed by passing --stats-json

Issue: "none"

## Test plan:

`yarn start`
Navigate around 

`yarn build-storybook`
`cd storybook-static && python3 -m http.server` (serves the static Storybook build using the Python3 http module)


jeremywiebe:
:( So it's not as easy as the steps I gave.  `yarn lint` now breaks with the error `TypeError: prettier.resolveConfig is not a function`.  I'll close this and come back ... sometime... maybe. :)

PR #1309, opened 2024-05-24
[Interactive Graph] View locked circle

nishasy:
## Summary:
View a locked circle!

This PR is kind of long, but that was necessary due to cascading type updates.
It's also a lot of tests.

Issue: https://khanacademy.atlassian.net/browse/LEMS-1940

## Test plan:
`yarn jest`

Storybook
http://localhost:6006/?path=/story/perseus-widgets-interactive-graph--locked-circle

Screenshot 2024-05-23 at 5 05 51 PM

### DIFF
--- a/.changeset/mighty-mayflies-whisper.md
+++ b/.changeset/mighty-mayflies-whisper.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": minor
+---
+
+[Interactive Graph] Render locked circles on interactive graphs
--- a/packages/perseus-editor/src/components/locked-figures-section.tsx
+++ b/packages/perseus-editor/src/components/locked-figures-section.tsx
@@ -18,6 +18,9 @@ import type {Props as InteractiveGraphEditorProps} from "../widgets/interactive-
 import type {LockedFigure, LockedFigureType} from "@khanacademy/perseus";
 
 type Props = {
+    // Whether to show the M2 features in the locked figure settings.
+    // TODO(LEMS-2016): Remove this prop once the M2 flag is fully rolled out.
+    showM2Features?: boolean;
     figures?: Array<LockedFigure>;
     onChange: (props: Partial<InteractiveGraphEditorProps>) => void;
 };
@@ -102,6 +105,7 @@ const LockedFiguresSection = (props: Props) => {
 
                 return (
                     <LockedFigureSettings
+                        showM2Features={props.showM2Features}
                         expanded={expandedStates[index]}
                         onToggle={(newValue) => {
                             const newExpanded = [...expandedStates];
--- a/packages/perseus-editor/src/components/locked-figures-section.tsx
+++ b/packages/perseus-editor/src/components/locked-figures-section.tsx
@@ -92,12 +92,13 @@ const LockedFiguresSection = (props: Props) => {
 
     return (
         <View>
-            {figures?
-                // TODO(LEMS-1941): Implement circle locked figure settings.
-                // Remove this block once circle locked figure settings are
-                // implemented.
-                .filter(f => f.type !== "circle")
-                .map((figure, index) => {
+            {figures?.map((figure, index) => {
+                if (figure.type === "circle") {
+                    // TODO(LEMS-1941): Implement circle locked figure settings.
+                    // Remove this block once circle locked figure settings are
+                    // implemented.
+                    return;
+                }
 
                 return (
                     <LockedFigureSettings
--- a/packages/perseus-editor/src/components/locked-figures-section.tsx
+++ b/packages/perseus-editor/src/components/locked-figures-section.tsx
@@ -92,13 +92,12 @@ const LockedFiguresSection = (props: Props) => {
 
     return (
         <View>
-            {figures?.map((figure, index) => {
-                if (figure.type === "circle") {
-                    // TODO(LEMS-1941): Implement circle locked figure settings.
-                    // Remove this block once circle locked figure settings are
-                    // implemented.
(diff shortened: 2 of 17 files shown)


packages/perseus-editor/src/components/locked-figures-section.tsx
```diff
@@ -92,20 +92,31 @@ const LockedFiguresSection = (props: Props) => {
...
-                    onRemove={() => removeLockedFigure(index)}
-                />
-            ))}
+            {figures?.map((figure, index) => {
+                if (figure.type === "circle") {
+                    // TODO(LEMS-1941): Implement circle locked figure settings.
+                    // Remove this block once circle locked figure settings are
+                    // implemented.
+                    return;
+                }
```
jeremywiebe, line 101: **nit of all nits**: You could use `.filter()` here to remove any circles and then it would yield a less disruptive change here (to revert, you just remove the `.filter()`).  

I say this in full realization that this code will be reverted in the very near future.  :)   But, thinking more "functionally" by chaining array functions might be useful for other things.  

```suggestion
            {figures?
                // TODO(LEMS-1941): Implement circle locked figure settings.
                // Remove this block once circle locked figure settings are
                // implemented.
                .filter(f => f.type !== "circle")
                .map((figure, index) => {
```

.changeset/mighty-mayflies-whisper.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/perseus": minor
+---
+
+[Interactive Graph] Render locked circles on interactive graphs
```
jeremywiebe, line 5: Thanks for the clear change log entry!

packages/perseus/src/perseus-types.ts
```diff
@@ -690,6 +690,15 @@ export type LockedLineType = {
     showPoint2: boolean;
 };
 
+export type LockedCircleType = {
+    type: "circle";
+    center: Coord;
+    radius: number;
+    color: LockedFigureColor;
+    fillOpacity: number;
```
jeremywiebe, line 698: 🤔  I don't know if there's a requirement to give the authors the ability to control _how_ translucent the fill is.  What do you think about just having `boolean` for this?  Like `opaqueFill: boolean` or `translucent: boolean`?

jeremywiebe: 🤔  Thinking a bit further.  Maybe it's two flags:

```js
  filled: boolean,
  translucentFill: boolean,
```

What do you think?

packages/perseus/src/widgets/__stories__/interactive-graph.stories.tsx
```diff
@@ -135,6 +136,19 @@ export const AllLockedRays = (args: StoryArgs): React.ReactElement => (
     />
 );
 
+export const LockedCircle = (args: StoryArgs): React.ReactElement => (
+    <RendererWithDebugUI
+        apiOptions={{
+            flags: {
+                mafs: {
+                    segment: true,
+                },
```
jeremywiebe, line 145: Let's use the `flagsForApiOptions` helper here.  That way as we flip more flags to true overall, we just update them in one place. 

https://github.com/Khan/perseus/blob/main/packages/perseus-editor/src/__stories__/flags-for-api-options.ts

packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
```diff
@@ -527,4 +528,34 @@ describe("locked layer", () => {
...
         });
     });
+
+    test("should render locked circles", async () => {
+        // Arrange
+        const {container} = renderQuestion(segmentWithLockedCircles, {
+            flags: {
+                mafs: {
+                    segment: true,
+                },
```
jeremywiebe, line 538: Same comment as earlier about using flagsForApiOptions.

I had another thought though.  Should these new locked figures be behind an M2 feature flag (like we [had for M1](https://github.com/Khan/perseus/blob/main/packages/perseus-editor/src/__stories__/flags-for-api-options.ts#L7))?

Maybe work that out with Mark.  I could see us shipping each locked figure as they're ready and tested, especially given we talked to Sarah P about possibly not needing full QE passes on locked figures.

packages/perseus/src/widgets/interactive-graphs/locked-circle.tsx
```diff
@@ -0,0 +1,10 @@
+import {Circle} from "mafs";
+import * as React from "react";
+
+import {lockedFigureColors, type LockedCircleType} from "../../perseus-types";
+
+const LockedCircle = (props: LockedCircleType) => {
+    return <Circle {...props} color={lockedFigureColors[props.color]} />;
```
jeremywiebe, line 7: Well that's easy! :)

jeremywiebe (approved):
Looking great!!  The one thing I think that needs to be resolved is whether we allow control over opacity.

packages/perseus/src/perseus-types.ts
jeremywiebe: Yep!  That's great!!

packages/perseus/src/widgets/__stories__/interactive-graph.stories.tsx
nishasy: I'll make sure to do that in the editor, but I don't think that necessary for viewing locked features. Also, I think `flags-for-api-toptions.ts` is in the `perseus-editor` package, whereas this stories file is in the `perseus` package.

mark-fitzgerald (approved):

PR #1310, opened 2024-05-24
Upgrade Mafs to 0.18.7

benchristel:
## Summary:
This version fixes the issue we had with Mafs 0.18.6, which was that the
node version was pinned too strictly. It was causing webapp CI to fail
because an earlier node version was used there.

Mafs 0.18.7 declares support for node >= 20, which should work for us.

Issue: none

Test plan:

- Release Perseus
- Deploy the new version of Perseus into webapp
- webapp CI should pass

### DIFF
--- a/.changeset/wild-kids-care.md
+++ b/.changeset/wild-kids-care.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": minor
+---
+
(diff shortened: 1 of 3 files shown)


mark-fitzgerald (approved):
Yahoo!

jeremywiebe (approved):

PR #1312, opened 2024-05-24
Remove :before pseudo-element to fix axis tick labels in Safari 15.6

benchristel:
## Summary:

There was a bug where the axis tick numbers did not render in older versions of Safari
(Ben repro'd the issue on Safari 15.6). Third tracked it down to the `:before` pseudo-element that we were using to display the negative sign without affecting the alignment of the numbers below the tick marks. By removing the `:before`, we sacrifice the nice alignment,
but fix the Safari bug. In the future, we can consider other ways of achieving the alignment,
e.g. absolute positioning.

## Test plan:

- Review Storybook on Safari 15.6

### DIFF
--- a/.changeset/rich-spoons-deliver.md
+++ b/.changeset/rich-spoons-deliver.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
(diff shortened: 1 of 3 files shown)


SonicScrewdriver (approved):
We paired on this solution together! :)

PR #1315, opened 2024-05-27
Creation of new Mafs-based Sinusoid Graph

SonicScrewdriver:
## Summary:
<Add your summary here>

Issue: LEMS-1832

## Test plan:

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/sinusoid.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/sinusoid.tsx
@@ -1,0 +1,104 @@
+import {color} from "@khanacademy/wonder-blocks-tokens";
+import {Plot} from "mafs";
+import * as React from "react";
+
+import {movePoint} from "../reducer/interactive-graph-action";
+
+import {StyledMovablePoint} from "./components/movable-point";
+
+import type {Coord} from "../../../interactive2/types";
+import type {SinusoidGraphState, MafsGraphProps} from "../types";
+import type {vec} from "mafs";
+
+type SinusoidGraphProps = MafsGraphProps<SinusoidGraphState>;
+
+export type SineCoefficient = [
+    number, // amplitude
+    number, // angularFrequency
+    number, // phase
+    number, // verticalOffset
+];
+
+export function SinusoidGraph(props: SinusoidGraphProps) {
+    const {dispatch, graphState} = props;
+
+    const {coords} = graphState;
+
+    // Destructure the coefficients for calculating the quadratic equation
+    const [a, b, c, d] = getSinusoidCoefficients(coords);
+
+    // Move a point to a new destination
+    const handleOnMove = (destination: vec.Vector2, elementId: number) => {
+        // If the destination is invalid, we do not want to move the point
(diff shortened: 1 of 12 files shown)


packages/perseus/src/widgets/interactive-graphs/graphs/sinusoid.tsx
```diff
@@ -0,0 +1,107 @@
...
+import type {vec} from "mafs";
+
+type SinusoidGraphProps = MafsGraphProps<SinusoidGraphState>;
+
+export type SineCoefficient = [
+    number, // amplitude
+    number, // angularFrequency
+    number, // phase
+    number, // verticalOffset
+];
```
benchristel, line 20: A 4-tuple is a lot for me to hold in my head. Would it be easier if this were an object with named fields?

```ts
export type SineCoefficient = {
  amplitude: number,
  angularFrequency: number,
  phase: number,
  verticalOffset: number,
}
```

`computeSine` could take this whole object as a param and destructure it internally.

packages/perseus/src/widgets/interactive-graphs/graphs/sinusoid.tsx
```diff
@@ -0,0 +1,107 @@
...
+    const [a, b, c, d] = getSinusoidCoefficients(coords);
+
+    // Move a point to a new destination
+    const handleOnMove = (destination: vec.Vector2, elementId: number) => {
+        // If the destination is invalid, we do not want to move the point
+        const validDestination = isValidDestination(
+            destination,
+            elementId,
+            coords,
+        );
```
benchristel, line 40: Question: do we need to validate the destination here, instead of in the reducer? If we did the validation in the reducer, it would be easier to write tests that assert that invalid moves really get rejected.

packages/perseus/src/widgets/interactive-graphs/types.ts
```diff
@@ -78,6 +79,11 @@ export interface QuadraticGraphState extends InteractiveGraphStateCommon {
     coords: [Coord, Coord, Coord];
 }
 
+export interface SinusoidGraphState extends InteractiveGraphStateCommon {
+    type: "sinusoid";
+    coords: vec.Vector2[];
```
benchristel, line 84: Instead of an array, should `coords` be a pair of points `[vec.Vector2, vec.Vector2]`?

benchristel (approved):
I left some suggestions inline, which you can take or leave but I don't see anything that will need another review, so I'm approving now.

mark-fitzgerald (approved):
Great job! Looks fantastic!!

I like Ben's comments, too.

packages/perseus/src/widgets/interactive-graphs/graphs/sinusoid.tsx
SonicScrewdriver: Ah yes, this logic makes far more sense in the reducer -- Thank you! I've moved it over as requested, and added a test to the reducer suite. :) 

I also took a look at the quadratic validation and I think that can probably move to the reducer as well, so long as we don't mind importing the getQuadraticCoefficients function. :) I'll make a micro ticket to follow up on that.

PR #1316, opened 2024-05-28
[Accessibility Bug] Fixing alt and aria-hidden issues with image widgets.

catandthemachines:
Fixing some issues around the accessibility of the image widget as it relates to alt text and aria-hidden usage. The fix is as follows:

- If the alt text and the caption are the same, set the alt text to "". This will ensure the image is still accessibility compliant, while avoiding reading out duplicate information. This is similar to a pattern that WCAG recommends for decorative images, https://www.w3.org/WAI/tutorials/images/decorative/.
- Remove hidden screen reader text blocks. This is unnecessary HTML content as we should make use of the alt text in images instead of re-creating it.
- Removing tabindex=0 and aria-hidden attribute on svg images. Non-interactive content should not be tabbable and like wise we shouldn't be aria-hiding images. If the images add no additional information for users we should simply set alt="".

### DIFF
--- a/packages/perseus/src/widgets/image.tsx
+++ b/packages/perseus/src/widgets/image.tsx
@@ -110,9 +110,7 @@ class ImageWidget extends React.Component<Props> {
                     {({setAssetStatus}) => (
                         <SvgImage
                             src={url}
-                            alt={
-                                alt
-                            }
+                            alt={alt}
                             width={backgroundImage.width}
                             height={backgroundImage.height}
                             preloader={apiOptions.imagePreloader}
--- a/packages/perseus/src/widgets/image.tsx
+++ b/packages/perseus/src/widgets/image.tsx
@@ -92,7 +92,7 @@ class ImageWidget extends React.Component<Props> {
 
     render(): React.ReactNode {
         let image;
-        let alt = this.props.caption === this.props.alt ? "" : this.props.alt;
+        const alt = this.props.caption === this.props.alt ? "" : this.props.alt;
         const {apiOptions} = this.props;
 
         const backgroundImage = this.props.backgroundImage;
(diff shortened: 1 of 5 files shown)


packages/perseus/src/widgets/image.tsx
```diff
@@ -97,37 +97,20 @@ class ImageWidget extends React.Component<Props> {
 
         const backgroundImage = this.props.backgroundImage;
 
+        if (this.props.alt && this.props.caption && this.props.caption == this.props.alt) {
```
mark-fitzgerald, line 100: Do we need to test for the presence of `alt` and `caption` since their [types are NonNullable](https://github.com/Khan/perseus/blob/0de2bd66e017d6fefc32c5570f83d8c16ce9906a/packages/perseus/src/widgets/image.tsx#L46-L47) (and they have [default values set](https://github.com/Khan/perseus/blob/0de2bd66e017d6fefc32c5570f83d8c16ce9906a/packages/perseus/src/widgets/image.tsx#L74-L75))?

packages/perseus/src/widgets/image.tsx
```diff
@@ -97,37 +97,20 @@ class ImageWidget extends React.Component<Props> {
 
         const backgroundImage = this.props.backgroundImage;
 
+        if (this.props.caption === this.props.alt) {
+            alt = "";
+        } else {
+            alt = this.props.alt;
+        }
```
mark-fitzgerald, line 104: Nit: I'm thinking that we can now write this as a ternary:
`const alt = this.props.caption === this.props.alt ? "" : this.props.alt`

mark-fitzgerald (approved):
Simple - I love it!

PR #1318, opened 2024-05-28
Make the cursor for the circle edge point a left-right arrow

benchristel:
## Summary:
You can only drag this particular point to the left and right. The new
cursor makes that apparent.

Issue: LEMS-1979

Test plan:

Paste this data into the flipbook:

```
{"content":"**Graph the circle $(x+2)^2+(y+1)^2=49$.**\n\n[[☃ interactive-graph 1]]","images":{},"widgets":{"interactive-graph 1":{"alignment":"default","graded":true,"options":{"backgroundImage":{"url":null},"correct":{"center":[-2,-1],"radius":7,"type":"circle"},"graph":{"type":"circle"},"gridStep":[1,1],"labels":["x","y"],"markings":"graph","range":[[-10,10],[-10,10]],"rulerLabel":"","rulerTicks":10,"showProtractor":false,"showRuler":false,"snapStep":[1,1],"step":[1,1]},"static":false,"type":"interactive-graph","version":{"major":0,"minor":0}}}}
```

The cursor should change to a left-right arrow when you hover over the
movable point that changes the size of the circle.

https://github.com/Khan/perseus/assets/693920/036bf0ed-aae0-40f3-8e4b-3599a3f4b9ca

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/components/css-cursor.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/components/css-cursor.ts
@@ -1,3 +1,3 @@
 // TODO(benchristel): add more possibilities for `cursor` if we need them
 // in the future.
-export type CSSCursor = "move" | "ew-resize"
+export type CSSCursor = "move" | "ew-resize";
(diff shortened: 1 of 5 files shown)


packages/perseus/src/widgets/interactive-graphs/graphs/components/css-cursor.ts
SonicScrewdriver, line 1: It might be good to add a comment that describes that this file / type is for, just to help reduce cognitive load for future devs.

SonicScrewdriver (approved):
Looks great! Thank you for setting this up :) I had one small comment about adding a code-comment, but it's nothing blocking.

PR #1320, opened 2024-05-29
Prevent the sides of a Mafs polygon from crossing

benchristel:
## Summary:
The legacy interactive graph constrained vertex movement on polygon
graphs to prevent sides from crossing. As of this commit, we now do the
same for Mafs interactive graphs.

Issue: LEMS-1896

Test plan:

Run:

```
grep -rl '"type":"polygon"' data/questions/ | xargs cat | pbcopy
```

Paste the data into `http://localhost:5173#flipbook`.

Play around with the polygons and verify that you can't make their lines
cross.

### DIFF
--- a/.changeset/eleven-ligers-accept.md
+++ b/.changeset/eleven-ligers-accept.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": minor
+---
+
+Prevent the sides of polygons on Mafs interactive graphs from crossing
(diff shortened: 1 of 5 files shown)


Myranae (approved):
Looks good! For the polygon graph types where the shape is already defined, I was not able to get it to cross itself. When it was an unlimited points/sides polygon, I was able to get the legacy version to cross itself (image below), but the mafs version wasn't available since I don't think that's set up yet. Might want to keep in mind that the legacy version in this situation does allow crossing sides and determine if that is on purpose.
image

Also, when switching to the branch and opening the flipbook with an example that defined coordinates with lines crossing, it started off with lines crossing (image below), but once the related points are moved, it doesn't cross over itself again for both legacy and mafs.
image

PR #1321, opened 2024-05-30
View a Locked Vector in Interactive Graph

mark-fitzgerald:
## Summary:
A vector is similar to a ray, but has a magnitude (limited length) and is not infinite. Like a Locked Ray, there is a need to have a Locked Vector.

Issue: LEMS-1949

## Test plan:
1. Launch Storybook (`yarn start`)
1. Navigate to Perseus => Widgets => Interactive Graph => [Locked Vector](http://localhost:6006/?path=/story/perseus-widgets-interactive-graph--locked-vector)
1. Note the two vectors on the graph (ray-like lines with a defined length)

## Example:
Screenshot 2024-05-30 at 1 40 11 PM


benchristel (approved):
Looks good! I notice that `point1` and `point2` are used in several places to refer to the endpoints of a vector. I wonder if these variables could be named something like `tail` and `tip` instead. Up to you, though!

packages/perseus/src/widgets/interactive-graphs/locked-vector.tsx
```diff
@@ -0,0 +1,24 @@
...
+    const {color, points} = props;
+    const [point1, point2] = points;
+
+    return (
+        <g className="locked-vector">
+            <Vector
+                tail={point1}
+                tip={point2}
+                color={lockedFigureColors[color]}
+            />
```
nishasy, line 19: Oh nice! I didn't realize this was going to be so easy to add.

nishasy (approved):
LGTM! Solid work! 🥳 ⬅️ ➡️

mark-fitzgerald:
> Looks good! I notice that `point1` and `point2` are used in several places to refer to the endpoints of a vector. I wonder if these variables could be named something like `tail` and `tip` instead. Up to you, though!

Good call. I changed those to be `tail` and `tip`.

PR #1322, opened 2024-05-30
Update Wonder Blocks Form version in perseus-editor package

nishasy:
## Summary:
WB Form has a new update: TextField now has `min`, `max`, and `step` props!

Updating dependencies here so they can be used in the interactive graph editor
locked features UI.

## Test plan:
- Add a random WB `<TextField>` somewhere with trivial required props.
- Give it a `type="number"`
- Use `min`, `max`, and/or `step` props
- Confirm it works!
- Change it to `type="text"`
- Confirm that typescript complains
- Delete random TextField


SonicScrewdriver (approved):
LGTM :) Thanks for the update!

PR #1323, opened 2024-05-30
[Interactive Graph Editor] Restrict the range on locked figures (arrow buttons only)

nishasy:
## Summary:
Now that `min` and `max` are available on WB TextField, let's use them
to restrict the coordinates to the graph range for locked figures. 

Here, I'm passing range down from the interactive graph editor to the
locked figure settings so that they can use the range to set `min` and
`max` on their coordinate inputs.

Note! This will stop the arrows on the input's spinbutton from going below
the min or above the max, but it will not stop a user from typing a number
out side the given range. 

^^ I'm going to add validation for typed numbers outside the range in
a follow-up PR.

## Test plan:
`yarn jest`

Manual testing
- Go to Storybook: http://localhost:6006/?path=/story/perseuseditor-editorpage--mafs-with-locked-figures
- Confirm that using the arrow buttons does not let the point or line go
  beyond the graph range.
- To go faster, hold down the arrow button instead of clicking one at
  a time.

### Point

https://github.com/Khan/perseus/assets/13231763/634fe3bd-e09b-4cc2-b46e-f95a9fa3f1a1

### Line

https://github.com/Khan/perseus/assets/13231763/5cc63ace-1b2e-4176-b5c8-d42421e83bf7


benchristel (approved):
Looks good. Does the range prop type have to be optional in all these places? It seems like the graph range should always exist.

nishasy:
@benchristel I was thinking about that... I made it optional for the reusable components so they're more flexible, but I can make it required for `LockedPointSettings`, `LockedLineSettings`, and everything further up the tree.

PR #1324, opened 2024-05-31
Use coords from props to initialize Mafs graph state

benchristel:
This fixes an bug in the exercise editor. Previously, if you loaded an
existing exercise with an interactive graph and opened the interactive graph
widget editor, the graph displayed in the widget editor would *not* reflect the
correct answer.  This was inconsistent/misleading, because content creators
specify the correct answer for graphing questions by moving the elements on the
graph to the correct position.

The way this was implemented for the legacy interactive graph widget was pretty
straightforward: when initializing graph state, we used any coordinates for
interactive elements that were provided in the props. This commit makes Mafs
graphs do the same.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2021

## Test plan:

- Deploy the dev build of Perseus into a webapp ZND.
- Visit
  `https://www.khanacademy.org/devadmin/content/exercises/interactive-graph-exercise/x1454c1ae30522aff/x692a1df2f20f9a4a`
- Expand the interactive graph widget editor
- You should see a visualization of the correct answer for the graph (compare
  to the equations shown in the correct answer field).

### DIFF
--- a/.changeset/pink-weeks-cheer.md
+++ b/.changeset/pink-weeks-cheer.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Fix a bug where, after a page refresh, the interactive graph widget editor displayed the default interactive elements instead of the correct answer to the graph.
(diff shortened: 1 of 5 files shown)


mark-fitzgerald (approved):
Consistency - I like it!

PR #1330, opened 2024-06-06
Implement the "angles" snapTo type for Polygon

Myranae:
## Summary:
Polygon previously was only able to snap to grid lines. Functionality was needed to allow it to snap to whole angles regardless of the grid or snapStep. This implements that feature.
 
Issue: LEMS-1855

## Test plan:
- Confirm all tests pass
- Confirm test coverage for new code is good
- Run `yarn dev` in the command line
- Navigate to localhost:5173
- Turn on mafs flags for all graphs
- Confirm the new gallery test for snapTo angles works correctly
- Confirm the other types of graphs work correctly
- Go to the flipbook and enter the following content for polygon
```
{"content":"[[☃ interactive-graph 1]]\n","images":{"web+graphie://ka-perseus-graphie.s3.amazonaws.com/4fbb5026607ec9ee083b365aa6bcda133adca567":{"width":440,"height":425}},"widgets":{"interactive-graph 1":{"type":"interactive-graph","alignment":"default","static":false,"graded":true,"options":{"step":[2,2],"backgroundImage":{"url":null},"markings":"graph","labels":["x","y"],"showProtractor":false,"showRuler":false,"showTooltips":false,"rulerLabel":"","rulerTicks":10,"range":[[-10,10],[-10,10]],"gridStep":[1,1],"snapStep":[1,1],"graph":{"type":"polygon","numSides":4,"showAngles":true,"snapTo":"grid"},"correct":{"type":"polygon","numSides":4,"coords":[[4,3],[4,7],[-1,7],[-1,3]],"snapTo":"grid"}},"version":{"major":0,"minor":0}}}}
```
- Move the points in "snapTo" grid mode and confirm the points snap to the grid
- Move the polygon and confirm it snaps to the grid
- Change the "snapTo" property of the graph object in the content to `angles`
- Move the points and confirm the angles are whole numbers and do not snap to the grid
- Move the entire polygon and confirm it does not snap to the grid and the angles stay whole numbers


packages/perseus/src/widgets/interactive-graph.tsx
```diff
@@ -965,11 +965,7 @@ class LegacyInteractiveGraph extends React.Component<Props, State> {
             throw makeInvalidTypeError("createPointForPolygonType", "polygon");
         }
 
-        // TODO(alex): check against "grid" instead, use constants
-        const snapToGrid = !_.contains(
-            ["angles", "sides"],
-            this.props.graph.snapTo,
-        );
+        const snapToGrid = this.props.graph.snapTo === "grid";
```
Myranae, line 968: Completing the TODO from Alex, though I tried to match how other parts of the code does this.

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -359,24 +398,48 @@ function doChangeRange(
...
+    delta: vec.Vector2,
+    constraintOpts: Omit<ConstraintArgs, "point" | "snapStep">,
+): vec.Vector2 => {
+    const maxMoves = coords.map((point: vec.Vector2) =>
+        maxNoSnapMove({...constraintOpts, point}),
+    );
+    const minMoves = coords.map((point: vec.Vector2) =>
+        minNoSnapMove({...constraintOpts, point}),
+    );
+    const [dx, dy] = getDeltaVertex(maxMoves, minMoves, delta);
```
benchristel, line 442: The near-duplication in all of these functions (`get(NoSnap)Change`, `(noSnap)bound`, `max(NoSnap)move`) makes me nervous. I think we could instead pass `snapStep: [0, 0]` to  the existing `getChange` function. This works because the snap step passed to `getChange` is only used to control how close a point can be to the edge of the graph — the actual snapping is done separately. Maybe the `snapStep` argument to `getChange` and its callees could be renamed to `minDistanceFromEdge` or `inset` or something like that to clarify how it's used.

Myranae: Totally no issue reverting that refactor. I just thought it was confusing, but it's not really worth it if it seems risky.

I'll work on updating the functions you mentioned. I didn't like the duplication either, but wasn't sure how to handle the snapStep. I'll try what you mentioned. Thanks! :)

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.tsx
```diff
@@ -16,18 +16,20 @@ type Props = {
...
-
+    const {point, onMove, cursor, color = WBColor.blue, snapTo} = props;
+    const snapToValue = snapTo ?? "grid";
     const {dragging} = useMovable({
         gestureTarget: elementRef,
         point,
         onMove,
-        constrain: (p) => snap(snapStep, p),
+        constrain: (p) =>
+            ["angles", "sides"].includes(snapToValue) ? p : snap(snapStep, p),
```
Myranae, line 32: The tests I added were meant to test this and the similar constrain update in the polygon file. They still pass when this is returned to the original constrain. Would love suggestions on how to test this.

benchristel (approved):
LGTM! Thanks for the updates.

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.tsx
benchristel: I think that in order to test this logic, you need a test where:

- angle snapping is on
- you try to move a point to a certain (x, y)
- angle snapping causes it to end up at a point different from (x, y)

So, e.g. in this test:

```ts
const updated = interactiveGraphReducer(
    state,
    movePoint(0, [3.130341262927298, -2.932437813845241]),
);

invariant(updated.type === "polygon");
expect(updated.coords[0]).toEqual([
    3.130341262927298, -2.932437813845241,
]);
```

The coords passed to `movePoint` need to be different from the coords in the expectation.

This is the kind of test where I'd run it once to find out what the resulting point coords are, and then just paste those into the expectation. The correctness of the resulting test will be very non-obvious, but at least it will guard against regressions.

Myranae: I can change it to be that direction for sure. Unfortunately, when I change the constrain on `StyledMovablePoint` back to just `snap(snapStep,p)`, the polygon points start snapping to the grid, and the test still passes.
image

benchristel: oh, my bad... I misread the code. The reducer tests don't invoke this code at all, since it's in the point component, not the reducer. To test the `constrain`, you'd have to write a React Testing Library test.

I actually think you *can* revert the constrain to `snap(snapStep, p)` and everything will work fine. The reducer will make sure the points are snapped correctly.

This `constrain` is only needed for keyboard support (it makes sure that pressing an arrow key always moves the point by a single snap step). Since we've decided not to pursue keyboard support for angle snapping, I think we can leave it as-is.

Myranae: When I revert the constrain, it starts snapping to the grid, but the tests don't catch it. Here's a video showing that, though it does use the flipbook. I'll start working on a RTL test for this tomorrow unless you still don't think it's needed. :)

https://github.com/Khan/perseus/assets/60857422/93db6f30-c7ce-491a-a250-973571c12a17

Myranae: Ah, if we can't use keyboard to move a point with angle snapping, then I don't think I can test using RTL, right? I'll just move forward with this and we can change things or add tests later if we need to :)

PR #1331, opened 2024-06-06
[Interactive-Graph] Merging isValidDestination into interactive-graph-reducer

catandthemachines:
The logic in isValidDestination is quite redundant and would be better served in the interactive-graph-reducer file along with other graphing movement logic.

In addition to this I added quadratic tests to the reducer test to ensure test coverage.

Task: https://khanacademy.atlassian.net/browse/LEMS-2023

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/quadratic.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/quadratic.tsx
@@ -10,8 +10,8 @@ import type {QuadraticGraphState, MafsGraphProps} from "../types";
 import type {vec} from "mafs";
 
 type QuadraticGraphProps = MafsGraphProps<QuadraticGraphState>;
-type QuadraticCoords = QuadraticGraphState["coords"];
 type QuadraticCoefficient = [number, number, number];
+export type QuadraticCoords = QuadraticGraphState["coords"];
 
 export function QuadraticGraph(props: QuadraticGraphProps) {
     const {dispatch, graphState} = props;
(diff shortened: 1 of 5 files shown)


packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.test.ts
```diff
@@ -385,6 +400,33 @@ describe("movePoint on a polygon graph", () => {
...
+        const state: InteractiveGraphState = {
+            ...baseQuadraticGraphState,
+            coords: [
+                [-5, 5],
+                [0, -5],
+                [5, 5],
+            ],
+        };
+
+        const updated = interactiveGraphReducer(state, movePoint(0, [0, 0]));
```
benchristel, line 423: It might be worth adding a comment here explaining why this move is invalid.

packages/perseus/src/widgets/interactive-graphs/graphs/quadratic.tsx
```diff
@@ -34,16 +34,6 @@ export function QuadraticGraph(props: QuadraticGraphProps) {
...
-            destination,
-            elementId,
-            coords,
-        );
-        if (validDestination === false) {
-            return;
-        }
-
         dispatch(movePoint(elementId, destination));
     };
```
benchristel, line 38: Since `handleOnMove` is just calling `dispatch` now, I wonder if its definition could be inlined into the JSX. Of course, we might have to undo that later if we want to optimize things with `useCallback`. Not a big deal either way.

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -227,6 +229,16 @@ function doMovePoint(
...
+            // Set up the new coords and check if the quadratic coefficients are valid
+            const newCoords: QuadraticCoords = [...state.coords];
+            newCoords[action.index] = action.destination;
+            const QuadraticCoefficients = getQuadraticCoefficients(newCoords);
+
+            // If the new destination results in an invalid quadratic equation, we don't want to move the point
+            if (QuadraticCoefficients === undefined) {
+                return state;
+            }
+
```
benchristel, line 241: I don't know why codecov is complaining about this. I think your tests cover this line. You can probably ignore the warning.

catandthemachines:
> LGTM! Thanks for simplifying this code. I left a couple suggestions inline, which you can take or leave.

Thanks Ben! And 100% agree, this makes it so much nicer :)

PR #1333, opened 2024-06-11
[Interactive Graph Editor] Remove m1 flag, hook up m2 flag for locked figures

nishasy:
## Summary:
Now that locked figures m1 has been rolled out, we can remove the flag from
the code. After that, we can also remove the flag from Growthbook.

While I'm here, I also realized the m2 flag wasn't hooked up anywhere, so
I did that as well.

Issue: https://khanacademy.atlassian.net/browse/LEMS-1999

## Test plan:
Storybook
- Confirm that locked ellipses and vectors don't show up in the UI
  - check the settings are not there
  - check that the "add locked figure" dropdown options are not there
- Confirm that locked ellipses and vectors show in the UI
  - check that the settings are there
  - check that the "add locked figure" dropdown options are there

| Current | M2 |
| --- | --- |
| Screenshot 2024-06-11 at 12 43 32 PM | Screenshot 2024-06-11 at 12 43 24 PM |

### DIFF
--- a/.changeset/weak-flowers-heal.md
+++ b/.changeset/weak-flowers-heal.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/perseus": minor
+"@khanacademy/perseus-editor": minor
+---
+
+[Interactive Graph Editor] Remove m1 flag from the code, and put locked vector and locked ellipse UI behind the m2 flag.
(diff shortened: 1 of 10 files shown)


benchristel (approved):
LGTM!

PR #1335, opened 2024-06-12
[Interactive Graph Editor] Remove unnecessary uses of useUniqueId

nishasy:
## Summary:
There are number of places in the code using `useUniqueIdWithMock` for
programmatic labels, but that is not necessary. You can nest the labeled
element within the `<label>` tag instead.

Changes in this PR:
- Update Wonder Blocks dependency so we can remove id="" from WB TextField
- Disable chromatic on a couple of stories that don't need it
- Remove `useUniqueIdWithMock`

There are a couple of places within the Interactive Graph Editor UI
where `uniqueIdWithMock` is still necessary. Specifically, within
`locked-figures-section.tsx`, as this unique ID is used to make
sure that the individual locked figure sections don't get mixed up
between multiple interacticve graph editors on the same page.

Note: I tried to remove `RenderStateRoot` from all the relevant tests
now that `useUniqueIdWithMock` is no longer there in most files.
But it couldn't be removed. I suspect this is because a number of
Wonder Blocks components that use `useUniqueIdWithMock` anyway.

Issue: none

## Test plan:
`yarn jest`

Storybook
- http://localhost:6006/?path=/story/perseuseditor-editorpage--mafs-with-locked-figures-m-2-flag
- Play around with every single input for every single locked figure
- Confirm they all work as expected.
- Try making two of the same kidn of figure
- Confirm that the label for the second one does not focus on an input
  from the first.

### DIFF
--- a/packages/perseus-editor/src/__stories__/editor-page.stories.tsx
+++ b/packages/perseus-editor/src/__stories__/editor-page.stories.tsx
@@ -128,6 +128,14 @@ export const MafsWithLockedFiguresCurrent = (): React.ReactElement => {
     );
 };
 
+MafsWithLockedFiguresCurrent.parameters = {
+    chromatic: {
+        // Disabling because this isn't visually testing anything on the
+        // initial load of the editor page.
+        disable: true,
+    },
+};
+
 export const MafsWithLockedFiguresM2Flag = (): React.ReactElement => {
     const [previewDevice, setPreviewDevice] =
         React.useState<DeviceType>("phone");
@@ -174,6 +182,14 @@ export const MafsWithLockedFiguresM2Flag = (): React.ReactElement => {
             }}
         />
     );
+};
+
+MafsWithLockedFiguresM2Flag.parameters = {
+    chromatic: {
+        // Disabling because this isn't visually testing anything on the
+        // initial load of the editor page.
+        disable: true,
+    },
 };
 
 export const WithSaveWarnings = (): React.ReactElement => {
--- a/packages/perseus-editor/src/components/color-select.tsx
+++ b/packages/perseus-editor/src/components/color-select.tsx
@@ -1,6 +1,7 @@
 import {lockedFigureColors} from "@khanacademy/perseus";
-import {View, useUniqueIdWithMock} from "@khanacademy/wonder-blocks-core";
+import {View} from "@khanacademy/wonder-blocks-core";
 import {OptionItem, SingleSelect} from "@khanacademy/wonder-blocks-dropdown";
+import {Strut} from "@khanacademy/wonder-blocks-layout";
 import {spacing} from "@khanacademy/wonder-blocks-tokens";
 import {LabelMedium} from "@khanacademy/wonder-blocks-typography";
 import {StyleSheet} from "aphrodite";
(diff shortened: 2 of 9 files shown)


packages/perseus-editor/src/components/color-select.tsx
```diff
@@ -22,34 +23,36 @@ type Props = {
...
-                {possibleColors.map((colorName) => (
-                    <OptionItem
-                        key={colorName}
-                        value={colorName}
-                        label={colorName}
-                        leftAccessory={
-                            <ColorSwatch color={colorName} decorative={true} />
-                        }
+            <LabelMedium tag="label">
+                <View style={styles.row}>
```
nishasy, line 29: I tried to make this work without this extra `View` here, but adding `style={styles.row}` directly on the `LabelMedium` didn't do anything... open to other suggestions if anyone knows how I can avoid this extra `View`.

packages/perseus-editor/src/__stories__/editor-page.stories.tsx
```diff
@@ -128,6 +128,14 @@ export const MafsWithLockedFiguresCurrent = (): React.ReactElement => {
     );
 };
 
+MafsWithLockedFiguresCurrent.parameters = {
+    chromatic: {
+        // Disabling because this isn't visually testing anything on the
+        // initial load of the editor page.
+        disable: true,
```
benchristel, line 135: What does this do? Note that we currently only run visual snapshot tests for files named `*regression.stories.tsx` ([configured here](https://github.com/Khan/perseus/blob/18be26f35f735e6ec985d65d48d9859e5227fa39/.github/workflows/storybook.yml#L47)). This file won't be tested on Chromatic with our current setup.

packages/perseus-editor/src/components/color-select.tsx
benchristel: Looks like `styles.row` sets the flex direction, but does not set `display: flex` on the element. `View`s are `display: flex` by default, but I guess `LabelMedium`s aren't.

I'm guessing if you add `display: flex` to `styles.row`, it will work.

benchristel (approved):
LGTM. I left a couple comments/questions inline.

packages/perseus-editor/src/__stories__/editor-page.stories.tsx
nishasy: 🤔  I disabled this story because I noticed it show up in the chromatic "UI tests" check in the last Version Packages PR. I'm pretty sure that means it's still being run. Is it possible that there's some place where it's not being restricted properly?

packages/perseus-editor/src/components/color-select.tsx
nishasy: You're right! I never would have guessed that a Wonder Blocks element doesn't have `display: "flex"`.

packages/perseus-editor/src/__stories__/editor-page.stories.tsx
benchristel: I haven't figured out why, but the number on the UI Tests check is not the actual number of tests that run. E.g. on this PR it says "350 tests unchanged"

Screen Shot 2024-06-12 at 3 14 01 PM

but if you click through to Chromatic, it shows that 350 stories were _uploaded_, and only 48 visual tests were run.

Screen Shot 2024-06-12 at 3 14 23 PM

I'm pretty sure the numbers on Chromatic.com are correct, because if we were really running 350 snapshot tests on every PR, we'd run out of monthly quota for our current (free) Chromatic plan.

nishasy: (Discussed offline.) Definitely something weird going on 🤔 I'll keep an eye out and point it out next time it happens. In the meantime, it doesn't hurt to keep this story disabled on Chromatic, so I'll leave this as is.

Thanks Ben!

PR #1337, opened 2024-06-12
[Interactive Graph] Fixing blue dot in Firefox Line graph

catandthemachines:
After a thorough investigation the blue dot that was appearing in Firefox was a result of an empty container having focus and Firefox attempting to put a focus-visible outline. This empty container is necessary for the visual tabbing structure of the graph. So to resolve this we need to set the outline to none for this empty focus handler.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2034

Before:
image

After:
image


packages/perseus/src/widgets/interactive-graphs/mafs-styles.css
```diff
@@ -145,6 +145,11 @@
     );
 }
 
+.MafsView
+    .movable-point__focusable-handle:focus-visible {
+    outline: none;
```
nishasy, line 150: Very nice! Can you add a comment explaining that this is here because empty elements still have a focus outline in Firefox?

nishasy (approved):
LGTM! 🎉 Just left one small comment

benchristel (approved):
LGTM. Thanks for the fix!

PR #1338, opened 2024-06-12
Implement the "sides" snapTo type for Polygon

Myranae:
## Summary:

Polygon previously was only able to snap to grid lines and angles. Functionality was needed to allow it to snap to whole side lengths regardless of the grid, snapStep, or angles. This implements that feature.

Known Strange Behavior: 
In the image below, the gallery shows it is possible for the lines to end up on top of each other when snapping to lines (shown on the right). When snapping to grid (on the left), this doesn't happen. 
image

This behavior is not a regression as it also happens on legacy graphs.
image

Issue: LEMS-1902

## Test plan:
- Confirm all tests pass
- Confirm test coverage for new code is good
- Run `yarn dev` in the command line
- Navigate to localhost:5173
- Turn on mafs flags for all graphs
- Confirm the new gallery test for snapTo sides works correctly
- Confirm the other types of graphs work correctly
- Go to the flipbook and enter the following content for polygon:
```
{"content":"[[☃ interactive-graph 1]]","images":{},"widgets":{"interactive-graph 1":{"alignment":"default","graded":true,"options":{"backgroundImage":{"url":""},"correct":{"coords":[[-2,1],[-2,6],[-5,6],[-5,1]],"numSides":4,"snapTo":"sides","type":"polygon"},"graph":{"numSides":7, "showAngles":true, "showSides": true, "snapTo":"grid","type":"polygon"},"gridStep":[1,1],"labels":["x","y"],"markings":"graph","range":[[-10,10],[-10,10]],"rulerLabel":"","rulerTicks":10,"showProtractor":false,"showRuler":false,"showTooltips":false,"snapStep":[2,2],"step":[2,2]},"static":false,"type":"interactive-graph","version":{"major":0,"minor":0}}}}
```
- Move the points in "snapTo" grid mode and confirm the points snap to the grid
- Move the polygon and confirm it snaps to the grid
- Change the "snapTo" property of the graph object in the content to `sides`
- Move the points and confirm the sides are whole numbers and do not snap to the grid or angles
- Move the entire polygon and confirm it does not snap to the grid or angles and the sides stay whole numbers
- Confirm no errors appear in the console


packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -200,10 +201,8 @@ function doMoveAll(
         case "polygon": {
             let newCoords: vec.Vector2[];
             switch (state.snapTo) {
-                // TODO(LEMS-1902): Implement sides snapping for polygons
                 case "grid":
-                case undefined:
-                case "sides": {
+                case undefined: {
```
SonicScrewdriver, line 205: Would `case undefined` be better handled by a default for the switch case?

Myranae: Hmm, I'm not sure what the best way to handle it would be. It should use the exact same bit of code as grid, so that's why I put it there, but maybe it would be better as a default since something other than undefined might be passed and we want it to always use the grid options for that. Thanks! I'll change it :)

Myranae: Should I just not have a case for grid then and have anything other than angles and sides go to default? Or would that be confusing?

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle.tsx
```diff
@@ -82,7 +84,9 @@ export const Angle = ({
         angleInDegrees = 360 - angleInDegrees;
     }
 
-    const angleLabelNumber = parseFloat(angleInDegrees.toFixed(1));
+    const angleLabelNumber = parseFloat(
+        angleInDegrees.toFixed(snapTo === "angles" ? 0 : 1),
```
SonicScrewdriver, line 88: Non-blocking nit: It might be good to add a comment here explaining that we only want to show whole angles in angle mode. :)

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -199,35 +200,24 @@ function doMoveAll(
     switch (state.type) {
         case "polygon": {
             let newCoords: vec.Vector2[];
```
SonicScrewdriver, line 202: Love it, this seems much cleaner!

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -545,22 +550,90 @@ function boundAndSnapToAngle(
...
+
+    // Solve for angle by using the law of cosines
+    const innerAngle = lawOfCosines(sides[0], sides[2], sides[1]);
+
+    const outerAngle = GraphUtils.findAngle(
+        coordsCopy[rel(1)],
+        coordsCopy[rel(-1)],
+    );
+
+    const onLeft =
```
SonicScrewdriver, line 629: Non-blocking nit: I'd love it if there could be a comment here explaining what the onLeft code is doing / how it affects the offset!

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -545,22 +550,90 @@ function boundAndSnapToAngle(
 
     const knownSide = magnitude(
         // @ts-expect-error - TS2345 - Argument of type 'number[]' is not assignable to parameter of type 'readonly Coord[]'.
-        vector(coords[rel(-1)], coords[rel(1)]),
+        vector(coordsCopy[rel(-1)], coordsCopy[rel(1)]),
     );
 
     const onLeft =
```
SonicScrewdriver, line 556: Non-blocking nit: I'd ALSO love it if there could be a comment here explaining what the onLeft code is doing / how it affects the offset! 

Since this is shared with both components, I wonder if it could be a little function so the comment doesn't need to be repeated. I'll leave this to your discretion, however, as perhaps this is a small component that simply isn't worth separating out.

SonicScrewdriver (approved):
This looks great from what I'm seeing! This logic gets pretty complex, so I made some requests for more commenting, but I don't consider them blockers!

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
Myranae: I would love to pull out some functions that are used in both, but I think that requires a bit more refactoring. And I tried it a bit, but we'd have to pass in a lot of information that makes it a bit harder to parse. The comment should be added now though :)

PR #1340, opened 2024-06-12
Removing purple color for second LinearSystem line.

catandthemachines:
## Summary:
Changing the second line in a Linear System from the color purple, to blue.

Issue: LEMS-2058

Before:
image

After:
image

## Test plan:
To manually test, go to storybook, select Interactive Graph Editor and toggle to Linear System.
You will see that instead of the second line being purple, it will be blue now.

### DIFF
--- a/.changeset/strong-boats-sip.md
+++ b/.changeset/strong-boats-sip.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Updating for all Linear Systems to have the same color line.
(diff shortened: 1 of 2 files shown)


Myranae (approved):
Looks good! 
image

What was the context for changing the purple? Purple is the best color, so I'm a little sad, but I'm pretty sure it's because the blue color is how we indicate interactivity :)

catandthemachines:
> Looks good! image
> 
> What was the context for changing the purple? Purple is the best color, so I'm a little sad, but I'm pretty sure it's because the blue color is how we indicate interactivity :)

I think it was a design choice to be consistent with the interactive blue, with this purple being a little confusing.

PR #1341, opened 2024-06-12
Changing circle graph default radius.

catandthemachines:
## Summary:
Updating default circle radius to 2.

Issue: LEMS-2063

Before:
image

After:
image

## Test plan:
Updated unit tests to reflect new setting.
To manually test, go to storybook, select Interactive Graph Editor and toggle to Circle.
You will see that instead of the radius being 1, that it will be set to 2 by default.


Myranae (approved):
Default circle shows radius of 2. Looks great! Thanks for the update :)

PR #1342, opened 2024-06-12
[Interactive Graph Editor] Show error if a coordinate pair value is out of range.

nishasy:
## Summary:
Adding `min` and `max` to the number text fields works for the arrow buttons,
but it doesn't stop someone from typing in a number out of range.

Add error messages so people know if a locked figure is out of range.

I'm going to add this to `getSaveWarnings()` in a future PR.

## Test plan:
Storybook
- https://khan.github.io/perseus/?path=/story/perseuseditor-editorpage--mafs-with-locked-figures-m-2-flag
- For all the locked figure types, type in the coordinate so they're out of range
- Confirm that the respective error messages show up
- Change the range of the graph
- Confirm that the error messages stay or don't stay accordingly

## Screenshots:

| x only | y only | x and y|
| --- | --- | --- |
| Screenshot 2024-06-12 at 4 14 42 PM | Screenshot 2024-06-12 at 4 14 54 PM | Screenshot 2024-06-12 at 4 14 47 PM |

### Demo: errors updating after changing the graph range

https://github.com/Khan/perseus/assets/13231763/36c91725-06b9-4d76-aa74-818e07f459d6

### DIFF
--- a/packages/perseus-editor/src/components/coordinate-pair-input.tsx
+++ b/packages/perseus-editor/src/components/coordinate-pair-input.tsx
@@ -47,8 +47,8 @@ const CoordinatePairInput = (props: Props) => {
         const xOutOfRange = coord[0] < range[0][0] || coord[0] > range[0][1];
         const yOutOfRange = coord[1] < range[1][0] || coord[1] > range[1][1];
 
-        const xRangeError = `${labels[0]} out of range ${range[0][0]} to ${range[0][1]}`;
-        const yRangeError = `${labels[1]} out of range ${range[1][0]} to ${range[1][1]}`;
+        const xRangeError = `${labels[0]} out of range (${range[0][0]} to ${range[0][1]})`;
+        const yRangeError = `${labels[1]} out of range (${range[1][0]} to ${range[1][1]})`;
 
         setXRangeError(xOutOfRange ? xRangeError : null);
         setYRangeError(yOutOfRange ? yRangeError : null);
--- a/packages/perseus-editor/src/components/coordinate-pair-input.tsx
+++ b/packages/perseus-editor/src/components/coordinate-pair-input.tsx
@@ -55,15 +55,16 @@ const CoordinatePairInput = (props: Props) => {
                     <TextField
                         type="number"
                         value={coordState[0]}
-                        min={range ? range[0][0] : undefined}
-                        max={range ? range[0][1] : undefined}
-                        onChange={(newValue) => handleCoordChange(newValue, 0)}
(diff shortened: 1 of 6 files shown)


packages/perseus-editor/src/components/locked-line-settings.tsx
```diff
@@ -183,6 +183,7 @@ const LockedLineSettings = (props: Props) => {
 
 const styles = StyleSheet.create({
     row: {
+        display: "flex",
```
nishasy, line 186: This got missed in my last PR, oops.

packages/perseus-editor/src/components/coordinate-pair-input.tsx
```diff
@@ -47,38 +74,56 @@ const CoordinatePairInput = (props: Props) => {
...
                     <Strut size={spacing.xxSmall_6} />
                     <TextField
                         type="number"
                         value={coordState[0]}
                         min={range ? range[0][0] : undefined}
                         max={range ? range[0][1] : undefined}
-                        onChange={(newValue) => handleCoordChange(newValue, 0)}
+                        onChange={(newValue) =>
+                            handleCoordChange(newValue, 0)
+                        }
```
nishasy, line 102: I tried using WB TextField's `validate` and `onValidate` props here, but I couldn't get it to update when I wanted it to.

packages/perseus-editor/src/components/coordinate-pair-input.tsx
```diff
@@ -17,7 +17,18 @@ type Props = {
...
+    const {
+        coord,
+        labels = ["x coord", "y coord"],
+        error,
+        range,
+        onChange,
+    } = props;
+
+    const [xRangeError, setXRangeError] = React.useState<string | null>(null);
+    const [yRangeError, setYRangeError] = React.useState<string | null>(null);
```
nishasy, line 29: I did separate `xRangeError` and `yRangeError` instead of `rangeError: [string, string]`, because using an array was causing the component to rerender infinitely from the `useEffect`. I could have kept the array by doing a deep equality check, but it seemed excessive for an array of 2 elements.

nishasy (commented):
Note: I tried adding unit tests in `interactive-graph-editor-locked-figures.test.tsx` to confirm that typing a value outside the range in the input would cause the error to show. But I couldn't get the tests to work 🤔 Not sure if it has something to do with userEvent?

nishasy:
We had a discussion and decided not to do this. We can trust the content authors.

PR #1343, opened 2024-06-13
[Interactive Graph Editor]: Show save warnings for invalid locked points, lines, ellipses, and vectors

nishasy:
## Summary:
NOTE: WE ARE NOT DOING THIS. I'm just pushing this code to a PR so I won't
lose whatever work I've already done on it.

## Test plan:
N/A


nishasy:
After discussing it, we decided we don't want to block the authors with this.

PR #1344, opened 2024-06-13
Revert "[Interactive Graph Editor] Restrict the range on locked figures (arrow buttons only) (#1323)"

nishasy:
## Summary:
We decided we no longer want to restrict the range on locked figures.
We trust the content authors to use the locked figures how they see fit.

## Test plan:
`yarn jest`

Storybook:
- Confirm that the arrow buttons no longer restrict the range


Myranae (approved):
Looks good! No longer restricts the range :)

PR #1345, opened 2024-06-13
Reset interactive graph state when the graph type or number of segments changes

benchristel:
## Summary:
There was a bug in the exercise editor where the preview did not reflect
changes to the graph type or number of segments. This happened because
changes to the `graph.type` and `graph.numSegments` props were ignored.
We now reinitialize the graph state when those props change.

Issue: LEMS-2041

Test plan:

Deploy a dev build of Perseus into the a webapp ZND. Visit:

https://www.khanacademy.org/devadmin/content/exercises/linear-graph-exercise/x3239cabc044aa28b/x325a92c13bc37711

Change the graph type of one of the interactive graphs to "segment".
The preview on the right-hand side of the editor should update.

Change the number of segments on the graph. The preview should update.

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
@@ -25,6 +25,7 @@ import {initializeGraphState} from "./reducer/initialize-graph-state";
 import {
     changeRange,
     changeSnapStep,
+    reinitialize,
     type InteractiveGraphAction,
 } from "./reducer/interactive-graph-action";
 import {interactiveGraphReducer} from "./reducer/interactive-graph-reducer";
@@ -33,6 +34,7 @@ import {GraphConfigContext} from "./reducer/use-graph-config";
 
 import type {InteractiveGraphState, InteractiveGraphProps} from "./types";
 import type {Widget} from "../../renderer";
+import type {PerseusGraphType} from "../../perseus-types";
 import type {vec} from "mafs";
 
 import "mafs/core.css";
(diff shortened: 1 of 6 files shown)


packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -149,6 +151,19 @@ export const StatefulMafsGraph = React.forwardRef<
...
     }, [dispatch, xMinRange, xMaxRange, yMinRange, yMaxRange]);
 
+    const numSegments = graph.type === "segment" ? graph.numSegments : null;
+    const originalPropsRef = useRef(props);
+    const latestPropsRef = useLatestRef(props);
+    useEffect(() => {
+        // This conditional prevents the state from being "reinitialized" right
+        // after the first render. This is an optimization, but also prevents
+        // a bug where the graph would be marked "incorrect" during grading
+        // even if the user never interacted with it.
```
benchristel, line 161: I feel like there's probably a simpler way to achieve this that doesn't require keeping a ref to two `props` objects. Maybe I should extract a `useEffectExceptOnMount` hook — unless we already have one?

nishasy: Looking at the [exports from wonder-blocks-core](https://github.com/Khan/wonder-blocks/blob/main/packages/wonder-blocks-core/src/index.ts), I'm not sure there's something that fits your use case. Maybe `useIsMounted` is the closest one? But I don't think that's the same thing... 🤔

nishasy (approved):
LGTM. Testing this must have been annoying 🥲

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
Myranae: Question: What's the downside to having the two refs? If you can extract a hook that makes things simpler, that always sounds better to me :) Unless it is harder to follow of course.

PR #1348, opened 2024-06-13
[Interactive Graph Editor] Update the locked ellipse settings so they only take degrees for the angle input

nishasy:
## Summary:
We had it so that locked ellipse settings allow for either degrees or radians
for the angel input. However, I realized that we're not saving the data for
what unit is being used. This means that if an angle were saved as "30 degrees"
and then the page were reloaded, it would then show up as "0.523599 radians".

For the sake of simplicty and consistency, I'm changing this so that it only
takes degrees, and I'm changing it to a number input since it no longer needs
to evaluate expressions with pi (that required it being a text input).

Issue: https://khanacademy.atlassian.net/browse/LEMS-1941

## Test plan:
`yarn jest`

Storybook
- http://localhost:6006/?path=/story/perseuseditor-editorpage--mafs-with-locked-figures-m-2-flag
- Confirm that the ellipse settings angle input shows degrees only
- Confirm that it is rotated correctly

Screenshot 2024-06-13 at 2 53 47 PM


packages/perseus-editor/src/components/locked-line-settings.tsx
```diff
@@ -191,6 +191,7 @@ const LockedLineSettings = (props: Props) => {
 
 const styles = StyleSheet.create({
     row: {
+        display: "flex",
```
nishasy, line 194: Oops, forgot this in my last PR

benchristel (approved):
LGTM!

packages/perseus-editor/src/components/__tests__/angle-input.test.tsx
```diff
@@ -15,91 +15,52 @@ describe("AngleInput", () => {
         });
     });
 
-    test("calls onChange with new angle (radians)", async () => {
+    test("displays angle in degrees", () => {
         // Arrange
-        const onChangeProps = jest.fn();
-        render(<AngleInput angle={0} onChange={onChangeProps} />, {
+        render(<AngleInput angle={Math.PI / 4} onChange={() => {}} />, {
```
Myranae, line 20: Question: What is the `Math.PI / 4` here?

packages/perseus-editor/src/components/angle-input.tsx
```diff
@@ -1,19 +1,16 @@
...
 import * as React from "react";
 
-const {InfoTip} = components;
-
 const degreeToRadian = (degrees: number) => {
     return (degrees / 180) * Math.PI;
 };
+const radianToDegree = (radians: number) => {
+    return (radians / Math.PI) * 180;
+};
```
mark-fitzgerald, line 13: Would these two functions be helpful to have in a Utils file?

packages/perseus-editor/src/components/__tests__/angle-input.test.tsx
nishasy: That is 45 degrees in radians! I think I just wanted to make it test at an angle.

PR #1353, opened 2024-06-14
[Interactive Graph] View a locked polygon

nishasy:
## Summary:
View a locked polygon!

- Add the types necessary to view a locked polygon
- Add the "locked polygon" component that displays on an interactive graph
- Add a storybook story to confirm that locked polygons render on the graph

Small note: I decided to switch the order of ellipse and vector, because I
thought it would make more sense for vector to be next to line and for
ellipse to be next to polygon.

Note that markings (congruent/similar/parallel/right angle) are part of
a separate tasking.

Issue: https://khanacademy.atlassian.net/browse/LEMS-1942

## Test plan:
`yarn jest`

Storybook
- http://localhost:6006/?path=/story/perseus-widgets-interactive-graph--locked-polygon

Screenshot 2024-06-14 at 4 29 24 PM


benchristel (approved):
Cool! LGTM!

PR #1354, opened 2024-06-17
Make graph specific properties persist and display on preview graph

Myranae:
## Summary:
Previously, graph specific properties, like number of sides and snapTo type for polygons, did not appear in the preview graph and did not persist or actually affect the preview graph. This makes it so the options selected affect the preview graph.

Issue: LEMS-2093

## Test plan:
- Pull up http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph-editor--with-mafs
- Play around with the different types of graphs and confirm the options affect the preview graph
- Confirm moving the shape or lines around does not reset the options
- Confirm all tests pass

### DIFF
--- a/.changeset/witty-moons-develop.md
+++ b/.changeset/witty-moons-develop.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": minor
+---
+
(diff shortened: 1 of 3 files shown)


nishasy (approved):
Interesting!

PR #1356, opened 2024-06-17
Interactive Graph Bug Fix: Ensure that we're bounding and snapping BEFORE checking if the new destination results in a valid graph.

SonicScrewdriver:
## Summary:
This is a fix for an issue in our Sinusoid graphs, where moving the points to the edge of the graph would cause the entire browser to crash. I believe that this was due the fact that we were bounding and snapping our points to the graph settings AFTER determining if the newDestination is a valid location, which means that points placed at the edge of the graph could end up in an invalid location. This made it possible to get the graph into a bit of an infinite state. 

By moving the boundAndSnap logic prior to checking whether it is a valid location, we can make sure that this loop is impossible. Repeating this order of operations for the Quadratic Graph actually manages to solve a separate issue that was considered a "Won't Do" earlier! :) 

Issue: LEMS-2064 & LEMS-2066

## Test plan:
- Manual testing in webapp (local dev) to confirm bug is reproducible 
- Manual testing with PR snapshot in webapp to confirm bug is solved.

### DIFF
--- a/.changeset/smooth-worms-wave.md
+++ b/.changeset/smooth-worms-wave.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+A fix for performance issues related to Sinusoid and Quadratic graphs
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/sinusoid.test.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/sinusoid.test.ts
@@ -18,16 +18,29 @@ describe("SinusoidGraph", () => {
         expect(getSinusoidCoefficients(coords)).toEqual(expected);
     });
 
-    it("should accurately calculate the sine wave", () => {
+    it("should accurately calculate the sine wave for a given x-coordinate", () => {
         const coords: SinusoidGraphState["coords"] = [
             [0, 0],
             [2, 2],
         ];
-        const coefficients = getSinusoidCoefficients(coords);
 
+        // Ensure that the coefficients are defined
+        const coefficients = getSinusoidCoefficients(coords);
+        expect(coefficients).toBeDefined();
+
+        // Grab a point where the sine wave should be 0
         const pointToTest = coords[0][0] + 4;
 
         // The sine wave should be roughly 0 at this point when accounting for floating point errors
-        expect(Math.round(computeSine(pointToTest, coefficients))).toEqual(0);
+        // We already know that the coefficients are defined from the previous test
+        expect(Math.round(computeSine(pointToTest, coefficients!))).toEqual(0);
+    });
+
+    it("should return undefined when the coefficients are invalid", () => {
+        const coords: SinusoidGraphState["coords"] = [
+            [0, 0],
+            [0, 0],
+        ];
+        expect(getSinusoidCoefficients(coords)).toBe(undefined);
     });
 });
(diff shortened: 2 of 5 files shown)


nishasy (approved):
Neat! Good job tracking this down too 🕵🏽

benchristel (commented):
Can you add a test for the changed behavior? Also, do you know where the infinite loop was occurring? Did it happen because Mafs tried to plot an infinite number of points? I'm not seeing any potential for infinite loops in the old reducer logic itself.

If the infinite loop is in Mafs, then we might want to submit a bugfix upstream. I could take that on.

SonicScrewdriver:
> Can you add a test for the changed behavior? Also, do you know where the infinite loop was occurring? Did it happen because Mafs tried to plot an infinite number of points? I'm not seeing any potential for infinite loops in the old reducer logic itself.
> 
> If the infinite loop is in Mafs, then we might want to submit a bugfix upstream. I could take that on.

@benchristel Ooo I absolutely will! I didn't add one initially as I wasn't sure what it would really show 🤔, but I can definitely see two tests that I could add now:

- Add a test to the reducer to ensure that a point cannot be placed at an invalid location, even if it's at the very edge of the graph. 
- After updating the logic in the sinusoid component (see notes below), make sure that we fall back to rendering the last "valid" coordinates for our graph when we are provided coordinates that result in an invalid graph that would require calculating by infinity. 

# Key Investigation notes:

- I was never able to get any kind of error state to appear as the whole browser would freeze and crash (blocking further discovery), so I'm not confident in the full identification of the core issue. 
- I have been utterly unable to recreate the issue in Perseus myself - it seems to be something connected with webapp. I tried testing in Flipbook, a new (local) Storybook story for testing Sinusoid+Mafs, and our Storybook Interactive Graph Editor w/ Mafs story.
- It's possible it could be related to how we're always getting the user's current answer for Khanmigo, which would be invalid at this point, but I am unable to verify yet. 

# What is the issue?

The issue, as far as I can tell, was as follows:

- A sinusoid graph shouldn't be allowed to hit a position that results in an invalid graph that requires dividing by infinity
- We have bounding logic to ensure we keep our points bound to the range of the graph 
- We were binding points to the graph AFTER doing the invalid location check, which meant the graph could hit this invalid state when users tried to place a point past the edge of the graph _**if**_ the other control point is also on the graph (x-axis) edge. For example, if the graph range is [-10, 10], with one point already at [10, 2] and the user tries to drag the second point to [12, 1] - then we would determine this as a valid location based on the current logic. THEN we'd run our bound logic, which would bounce the point back to [10,1], and result in an invalid graph state. 
- We have an equation for calculating coefficients in the Sinusoid component that does not have the safety check / fallback that Quadratic does when the coefficients hit "infinity". 
- Our computeSine currently logic runs whether the coefficients are valid or not, and the result is provided _directly_ into the mafs graph. 

# Next Steps for this ticket:

Given this .. I feel like the issue _probably_ isn't with mafs so much as something that was happening between webapp and Perseus when the graph was allowed to hit the invalid state. I feel like I should be able to recreate this issue in Perseus if it was an issue with Mafs specifically. 

With the logic added prior to this comment**_, it should never be possible to hit this invalid state any more_**. That being said, I'll still work on adding the same fallback checks to Sinusoid as we are using in Quadratic just to keep things safe. I'll also add the two tests I mentioned above! 

# In terms of completing the investigation: 

I'll make a new DRAFT PR on a different branch with some strategic console logs/logic changes to test my theory that it's the fact that the sinusoid graph is allowed to become invalid at all that results in the browser crashing. I'll do these tests tomorrow morning and report back here if I'm able to find the exact root cause.

If I'm still not able to find anything concrete about this exact root cause after these tests, I may simply leave a note on LEMS-2056 about investigating this issue further with a summary of the current discovery. LEMS-2056 is an investigation ticket made to specifically look into performance issues with our Graphs, with a particular focus on Quadratics and Sinusoids. 

Hopefully that clearly explains my discoveries so far! I'm happy to jump on a call tomorrow if you want to discuss further or if an alternative approach is desired. :)

PR #1363, opened 2024-06-20
Simplify InputWithExamples

handeyeco:
## Summary:
While working on [LEMS-2081](https://khanacademy.atlassian.net/browse/LEMS-2081) I started to get the hunch that InputWithExamples didn't use MathInput or MathOutput anymore.

The only uses I could find of InputWithExamples was InputNumber and NumericInput; both had hardcoded values for `type="text"`.

I checked in `webapp` and `mobile` too. Think we're safe to clean this up if tests are passing.

[LEMS-2081]: https://khanacademy.atlassian.net/browse/LEMS-2081?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/perseus/src/components/input-with-examples.tsx
```diff
@@ -19,7 +17,6 @@ import type {StyleType} from "@khanacademy/wonder-blocks-core";
 const {captureScratchpadTouchStart} = Util;
 
 type Props = {
-    type: "math" | "text" | "tex";
```
benchristel, line 22: The removal of this prop will cause type errors in webapp when Perseus is updated. We'll need to bump the major version for this.

Alternatively, we could just mark this prop as deprecated, e.g.

```ts
/**
 * @deprecated type has no effect and will be removed in a future release of Perseus
 */
```

 and release this change under a minor version bump. Then we could remove the prop in a future major release. I'd rather just bump the major version, though.

benchristel (approved):
LGTM! I think we should bump the major version instead of the minor, though.

nishasy (approved):
This is so weird! I also confirmed that `InputWithExamples` is not being used anywhere with a different type than `"text"`, so I agree with you that this is the right way to go 👍🏽 

It's odd to me that `NumericInput` doesn't use the `type="math"` version of this. And also that it uses `InputWithExamples` _at all_ if it's not utilizing that?

packages/perseus/src/components/input-with-examples.tsx
handeyeco: I debated major vs minor (and even considered patch); I decided on minor since it's a change to a _component_ and not a _widget_. The component is not exported, so I don't believe the external API of Perseus is changing - we're just changing how widgets consume an internal component.

I already started the upgrade process in [Webapp](https://github.com/Khan/webapp/pull/22886) to see if it breaks anything; if it throws errors I'll bump to major.

benchristel: Ah cool! Yes, you are totally right about it not being exported. I would actually lean toward a patch release then, since I don't think this changes any observable behavior? I.e. we're not adding a feature, just deleting dead code.

handeyeco:
@nishasy My guess is it was code predating the `math-input` package (though I could totally be wrong). IIRC those widgets use that input in mobile now and content that needs complex math on desktop uses the Expression widget.

PR #1360, opened 2024-06-18
[Interactive Graph Editor] Add the ability to reorder locked figure settings

nishasy:
## Summary:
Added some icon buttons so that content authors can reorder the locked figure
settings. This can be used to make sure that one figure can render on top
of another figure as desired. (The lower they are in the settings, the more
in the front they are on the graph.)

Note: I added more tests to the `interactive-graph-editor-locked-figures.test.tsx` file.
I think it's getting _too_ long, so I'm considering breaking it up into a different file per
`describe` block in the future in its own PR.

Issue: https://khanacademy.atlassian.net/browse/LEMS-1951

## Test plan:
`yarn jest packages/perseus-editor/src/widgets/__tests__/interactive-graph-editor-locked-figures.test.tsx`

Play around with the buttons in the storybook file, as demonstrated in the videos below.

http://localhost:6006/?path=/story/perseuseditor-editorpage--mafs-with-locked-figures-m-2-flag

### Reordering in editor

https://github.com/Khan/perseus/assets/13231763/6bc92a9f-aa88-442d-bc47-7e10b3901354

### Updating order on graph

https://github.com/Khan/perseus/assets/13231763/05f9ce2a-28fc-4e30-9a12-77f8651a64f5

PR #1365, opened 2024-06-20
[Interactive Graph Editor] Put the action buttons under the m2 flag

nishasy:
## Summary:
The action buttons that were added in https://github.com/Khan/perseus/pull/1360
need to be behind the m2 flag. Adding the check here.

Issue: https://khanacademy.atlassian.net/browse/LEMS-1951

## Test plan:
Storybook
- It should not have the action buttons in the locked figure settings
- Expand all the locked figure settings.
- They should all have the action buttons.

### Current
Screenshot 2024-06-20 at 2 02 23 PM

### M2 flag
Screenshot 2024-06-20 at 2 02 39 PM

Screenshot 2024-06-20 at 2 02 46 PM

### DIFF
--- a/.changeset/funny-sheep-retire.md
+++ b/.changeset/funny-sheep-retire.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": patch
+---
+
+[Interactive Graph Editor] Put the action buttons behind the m2 flag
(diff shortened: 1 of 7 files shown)


mark-fitzgerald (approved):
It's too bad that we don't have a React Context way of accessing these flags so that we wouldn't need to pass it as a prop in so many places.

nishasy:
@mark-fitzgerald I agree! I think it would make sense for API options to be a context.

PR #1366, opened 2024-06-20
[Interactive Graph Editor] Add locked features M2b flag

nishasy:
## Summary:
Adding the m2b flag and its prop to all the relevant components.

This way, it should be easy to use a simple conditional when we add
the upcoming locked figures (function graphs, labels, maybe polygon markings).

Issue: none

## Test plan:
- Confirm that only point and line show up in the settings
- Point, line, vector, ellipse, and polygon should show up
- Go to http://localhost:6006/?path=/story/perseuseditor-editorpage--mafs-with-locked-figures-m-2-b-flag
- It should be identical to the m2 one for now.

### DIFF
--- a/.changeset/sour-coats-agree.md
+++ b/.changeset/sour-coats-agree.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/perseus": patch
+"@khanacademy/perseus-editor": patch
+---
+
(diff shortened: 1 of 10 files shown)


mark-fitzgerald (approved):
2b or not 2b? That is a boolean. _(sorry, Shakespeare)_

PR #1367, opened 2024-06-20
Create our own wrapper around Mafs' `useMovable`

benchristel:
## Summary:
We need the `useMovable` hook to do a couple things that Mafs doesn't
yet enable:

- let us handle keyboard events explicitly (needed for angle graphs, and
  maybe polygons if we decide to continue supporting side-length
  snapping). Currently, Mafs uses heuristics and search to convert
  keyboard input into movement vectors based on the provided `constrain`
  function, but this only works for grid-based snapping.

- get the movement vector in *pixel coordinates*. This is needed to
  implement the protractor rotation handle. The protractor is a
  fixed-size image that doesn't scale with the graph, so we really want
  to be working in pixels - otherwise, we have to do a bunch of
  confusing coordinate transformations.

We'll likely want to iterate on these features a bit before contributing
them back to Mafs. Therefore, this commit creates a wrapper around
`useMovable` where we can add our new functionality.

I have named the wrapper `useDraggable` instead of `useMovable`
so people don't mistake it for the Mafs hook.

Issue: none

Test plan:

- Run `yarn dev` and view the graphs at `http://localhost:5173/`.
- Turn on the Mafs flags for all graph types.
- Verify that you can move all the interactive graph elements with
  mouse and keyboard.

### DIFF
--- a/.changeset/shaggy-moles-promise.md
+++ b/.changeset/shaggy-moles-promise.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Internal: wrap the Mafs `useMovable` hook, creating a seam where we can add new functionality.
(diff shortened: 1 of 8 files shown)


SonicScrewdriver (approved):
Looks pretty straight forward to me -- thank you so much for setting this up! Only thing I can think of would be to maybe add a comment to explain the use-draggable.ts file, but I don't think that's a blocker at all.

PR #1368, opened 2024-06-20
Organize locked figures into own folder

mark-fitzgerald:
The number of files dedicated to locked figures has reached a point where we should group them for better organization.


nishasy (approved):
Thanks for doing this!

PR #1369, opened 2024-06-21
Interactive ServerItemRenderer story

handeyeco:
## Summary:
Super hacky, but a bad tool is better than no tool (maybe?)

https://github.com/Khan/perseus/assets/16308368/97f97f0a-7c72-45d5-8e2a-85256f39c24d

### DIFF
--- a/packages/perseus/src/__stories__/server-item-renderer.stories.tsx
+++ b/packages/perseus/src/__stories__/server-item-renderer.stories.tsx
@@ -147,7 +147,7 @@ export const Interactive = () => {
                         <button onClick={() => setDebug(!debug)}>
                             {debug
                                 ? "Use ServerItemRenderer"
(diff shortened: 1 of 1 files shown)


benchristel (approved):
> Super hacky, but a bad tool is better than no tool (maybe?)

[Worse is better](https://cs.stanford.edu/people/eroberts/courses/cs181/projects/2010-11/WorseIsBetter/index.php/Worse-is-better.html)! 😂 

My one question is whether this should be part of the `yarn dev` UI rather than Storybook, but we can always move it later (or reference the same component in both places)

handeyeco:
@benchristel I think long-term it would be better to add this to `yarn dev`. I have some questions about `yarn dev`'s access to Perseus CSS though. I'm just going to cut this so we can start using it and we can iterate.

PR #1370, opened 2024-06-21
Copy the implementation of `useMovable` into `useDraggable`

benchristel:
## Summary:

This copies Mafs' implementation of the `useMovable` hook into our own `useDraggable`
hook with no substantive changes.  In a future PR, I'll add the new features I need.

Context: We need the useMovable hook to do a couple things that Mafs doesn't
yet enable:

- let us handle keyboard events explicitly (needed for angle graphs, and
maybe polygons if we decide to continue supporting side-length
snapping). Currently, Mafs uses heuristics and search to convert
keyboard input into movement vectors based on the provided `constrain`
function, but this only works for grid-based snapping.

- get the movement vector in pixel coordinates. This is needed to
implement the protractor rotation handle. The protractor is a
fixed-size image that doesn't scale with the graph, so we really want
to be working in pixels - otherwise, we have to do a bunch of
confusing coordinate transformations.

We'll likely want to iterate on these features a bit before contributing
them back to Mafs. Therefore, this commit creates a wrapper around
useMovable where we can add our new functionality.

Issue: none

Test plan:

- Run `yarn dev` and view the graphs at `http://localhost:5173/`.
- Turn on the Mafs flags for all graph types.
- Verify that you can move all the interactive graph elements with
  mouse and keyboard.

### DIFF
--- a/packages/perseus/package.json
+++ b/packages/perseus/package.json
@@ -46,6 +46,7 @@
         "@khanacademy/perseus-linter": "^0.4.0",
         "@khanacademy/pure-markdown": "^0.3.5",
         "@khanacademy/simple-markdown": "^0.12.0",
+        "@use-gesture/react": "^10.2.27",
         "mafs": "0.18.7"
     },
     "devDependencies": {
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/use-draggable.test.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/use-draggable.test.tsx
@@ -12,16 +12,16 @@ function TestDraggable(props: {
     point: vec.Vector2;
     constrain: (point: vec.Vector2) => vec.Vector2;
 }) {
-    const gestureTarget = useRef<HTMLParagraphElement>(null);
+    const gestureTarget = useRef<HTMLButtonElement>(null);
     const {dragging} = useDraggable({
         ...props,
         gestureTarget,
         onMove: () => {},
     });
     return (
-        <p role="button" ref={gestureTarget}>
+        <button ref={gestureTarget}>
             dragging: {String(dragging)}
-        </p>
+        </button>
     );
 }
 
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/use-draggable.test.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/use-draggable.test.tsx
@@ -1,0 +1,78 @@
+import {render, screen} from "@testing-library/react";
+import {userEvent as userEventLib} from "@testing-library/user-event";
+import {Mafs} from "mafs";
+import * as React from "react";
+import {useRef} from "react";
+
+import {useDraggable} from "./use-draggable";
+
+import type {vec} from "mafs";
+
+function TestDraggable(props: {
+    point: vec.Vector2;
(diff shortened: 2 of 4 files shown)


packages/perseus/package.json
benchristel, line 1: `@use-gesture/react` is a dependency of Mafs that we are now importing directly. I verified that adding it here doesn't cause multiple copies/versions of `@use-gesture/react` to get installed.

packages/perseus/src/widgets/interactive-graphs/graphs/use-draggable.test.tsx
benchristel, line 1: These tests are unfortunately incomplete. I couldn't figure out how to get `@use-gesture/react` to register a drag gesture, either with `userEvent` or `fireEvent`.

packages/perseus/src/widgets/interactive-graphs/graphs/use-draggable.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+    const gestureTarget = useRef<HTMLParagraphElement>(null);
+    const {dragging} = useDraggable({
+        ...props,
+        gestureTarget,
+        onMove: () => {},
+    });
+    return (
+        <p role="button" ref={gestureTarget}>
+            dragging: {String(dragging)}
+        </p>
```
Myranae, line 24: Question: Is it okay to set the role of a `p` element to "button" like this? I know this is only a test, so maybe it's not super important.

packages/perseus/src/widgets/interactive-graphs/graphs/use-draggable.test.tsx
Myranae: Suggestion: Is this [Moving a Pointer](https://testing-library.com/docs/user-event/pointer#moving-a-pointer) section of the Testing Library docs useful? It seems like it is showing how to user userEvent.pointer to drag, though I could be reading it wrong.

Myranae (approved):
Looks good! Was able to test that all the Mafs graphs (or everything I could tab to) moved with keyboard and also with mouse. I was surprised to see the points on a polygon with angle snapping moving with keyboard, but the points on the polygon with sides snapping still don't move with the keyboard. Added some non-blocking comments.

packages/perseus/src/widgets/interactive-graphs/graphs/use-draggable.test.tsx
benchristel: Yeah, I tried what those docs suggest. `@use-gesture/react` doesn't see it as a drag, though.

catandthemachines (approved):
Approved! Though I am curious why it is necessary to move the logic in Mafs into a useDraggable function for our internal use, instead of utilizing useMovable.

packages/perseus/src/widgets/interactive-graphs/graphs/use-draggable.test.tsx
benchristel: Update: I looked at the tests for `@use-gesture/react` itself and found that they're using `fireEvent`, so I came up with a testing solution based on that.

benchristel:
@catandthemachines I've updated the PR description explaining the reasoning behind this change. Copied from here: https://github.com/Khan/perseus/pull/1367

PR #1372, opened 2024-06-21
[Storybook: Editor Page] Add a working preview for the editor page in storybook

nishasy:
## Summary:
The preview iframe doesn't work in storybook, because in links to a local URL
to webapp when it's in prod. It has never worked in storybook.

We need to be able to see what the initial state of the questions on the editor
page would look like without making any changes to the actual preview (yet).

I've added a preview here that shows the question and the hints in the editor
page. This works for all widgets. It's sticky, so it follows as you scroll
down a long editor. You can also close it if it's taking too much space.

This is a prerequisite to viewing an interactive graph with initial coordinates.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2051

## Test plan:
http://localhost:6006/?path=/story/perseuseditor-editorpage--demo

### Interactive graph widget

Screenshot 2024-06-21 at 2 58 45 PM

### Expression widget

Screenshot 2024-06-21 at 2 58 14 PM

### Demo video

https://github.com/Khan/perseus/assets/13231763/c39db9c4-380d-4e3f-a556-5542198046c8

### DIFF
--- a/packages/perseus-editor/src/__stories__/editor-page-util.tsx
+++ b/packages/perseus-editor/src/__stories__/editor-page-util.tsx
@@ -1,0 +1,168 @@
+import {
+    Renderer,
+    type APIOptions,
+    type DeviceType,
+    type Hint,
+    type PerseusAnswerArea,
+    type PerseusRenderer,
+} from "@khanacademy/perseus";
+import Button from "@khanacademy/wonder-blocks-button";
+import {View} from "@khanacademy/wonder-blocks-core";
+import IconButton from "@khanacademy/wonder-blocks-icon-button";
+import {Strut} from "@khanacademy/wonder-blocks-layout";
+import {color, spacing} from "@khanacademy/wonder-blocks-tokens";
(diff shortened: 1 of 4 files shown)


packages/perseus-editor/src/__stories__/editor-page-util.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+type Props = {
+    apiOptions?: APIOptions;
+    question?: PerseusRenderer;
+    hints?: ReadonlyArray<Hint>;
+};
+
+const onChangeAction = action("onChange");
+
+function EditorPageUtil(props: Props) {
```
benchristel, line 34: The name `EditorPageUtil` isn't particularly intuitive / self-explanatory to me. Maybe this component could be named something like `EditorPageWithStorybookPreview`?

benchristel (approved):
Thank you for adding this and cleaning up that duplicated story code!

One comment about naming inline; otherwise, LGTM!

PR #1373, opened 2024-06-22
[Interactive Graph] Update the builder with all currently migrated graph types

nishasy:
## Summary:

I updated the builder utility with all currently migrated graph types. This
meant adding linear, linear system, ray, quadratic, and sinusoid.

I also added the ability to give them all starting coords. This is necessary
to prove that the starting coords can be set at all.

Now that it's clear they can be set, we can move onto adding the ability
to edit these coordinates in the editor UI in the next task.

NOTE: The task also includes start coords for the Points graph type, but
that doesn't seem to have the mafs flag on in storybook yet, so I didn't
include that here.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2051

## Test plan:
Storybook
- [segment](http://localhost:6006/?path=/story/perseuseditor-editorpage--interactive-graph-segment-with-starting-coords)
- [multiple segments](http://localhost:6006/?path=/story/perseuseditor-editorpage--interactive-graph-segments-with-starting-coords)
- [linear](http://localhost:6006/?path=/story/perseuseditor-editorpage--interactive-graph-linear-with-starting-coords)
- [linear system](http://localhost:6006/?path=/story/perseuseditor-editorpage--interactive-graph-linear-system-with-starting-coords)
- [ray](http://localhost:6006/?path=/story/perseuseditor-editorpage--interactive-graph-ray-with-starting-coords)
- [circle](http://localhost:6006/?path=/story/perseuseditor-editorpage--interactive-graph-circle-with-starting-coords)
- [quadratic](http://localhost:6006/?path=/story/perseuseditor-editorpage--interactive-graph-quadratic-with-starting-coords)
- [sinusoid](http://localhost:6006/?path=/story/perseuseditor-editorpage--interactive-graph-sinusoid-with-starting-coords)

## Storybook previews showing custom start coords

| segment | segments | linear |
| --- | --- | --- |
| Screenshot 2024-06-21 at 5 08 25 PM | Screenshot 2024-06-21 at 5 08 31 PM | Screenshot 2024-06-21 at 5 08 36 PM |
| linear system | ray | circle |
| Screenshot 2024-06-21 at 5 08 40 PM | Screenshot 2024-06-21 at 5 08 44 PM | Screenshot 2024-06-21 at 5 08 49 PM |
| quadratic | sinusoid | |
| Screenshot 2024-06-21 at 5 08 53 PM | Screenshot 2024-06-21 at 5 08 58 PM | |

### DIFF
--- a/.changeset/short-badgers-wonder.md
+++ b/.changeset/short-badgers-wonder.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/perseus": patch
+"@khanacademy/perseus-editor": patch
+---
+
+[Interactive Graph] Update the interactive graph builder with all currently migrated graph types
--- a/packages/perseus-editor/src/__stories__/editor-page.stories.tsx
+++ b/packages/perseus-editor/src/__stories__/editor-page.stories.tsx
@@ -7,7 +7,17 @@ import {StyleSheet} from "aphrodite";
 import * as React from "react";
 
 import {EditorPage} from "..";
-import {segmentWithLockedFigures} from "../../../perseus/src/widgets/__testdata__/interactive-graph.testdata";
+import {
+    circleWithStartingCoordsQuestion,
+    linearSystemWithStartingCoordsQuestion,
+    linearWithStartingCoordsQuestion,
+    quadraticWithStartingCoordsQuestion,
+    rayWithStartingCoordsQuestion,
+    segmentWithLockedFigures,
+    segmentWithStartingCoordsQuestion,
+    segmentsWithStartingCoordsQuestion,
+    sinusoidWithStartingCoordsQuestion,
+} from "../../../perseus/src/widgets/__testdata__/interactive-graph.testdata";
 import {registerAllWidgetsAndEditorsForTesting} from "../util/register-all-widgets-and-editors-for-testing";
 
 import EditorPageWithStorybookPreview from "./editor-page-with-storybook-preview";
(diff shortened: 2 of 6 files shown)


benchristel (approved):
Looks great, thank you! My one question is whether we want to let callers of the builder pass both `startCoords` and `numSegments` to the segment graph. In the common case, the information in `numSegments` is redundant and could be derived from `startCoords`. I could _maybe_ see an argument that we'd want to let `numSegments` be different from `startCoords` in order to test what happens when the data is inconsistent. That's such an edge case, though, that I don't think it's worth complicating the builder for the sake of the `O(1)` tests we might write that take advantage of it.

I'll leave the decision up to you, though.

nishasy:
@benchristel I think it's still nice to have `numSegments` because it's accurate to the experience when `coords` isn't used. It would be nice to use the builder to create the segment graph with the default coords for a number of segments rather than having to find out and type all the coords out.

benchristel:
> It would be nice to use the builder to create the segment graph with the default coords for a number of segments rather than having to find out and type all the coords out.

I agree. I guess the way I'd prefer to accomplish that is by having a separate method to set non-default coords, so you could do either:

```ts
const question = interactiveGraphBuilder()
  .withNumSegments(3)
  .build();
```

or

```ts
const question = interactiveGraphBuilder()
  .withSegments([
    // ...
  ])
  .build();
```

PR #1375, opened 2024-06-25
[Interactive Graph Editor] Fix the editor preview for segments, points, and polygons

nishasy:
## Summary:
There was an issue where some of the custom settings for segments, points, and polygon
graphs were not updating the right-side storybook preview for interactive graphs
in the EditorPage stories.

The issue is that the `onChange` functions in `interactive-graph-editor.tsx` were
updating the `correct` field, but they were not updating the `graph` field.

Updating the `graph` field as well in this PR.

Issue: none

## Test plan:
Storybook
- Go to http://localhost:6006/?path=/story/perseuseditor-editorpage--interactive-graph-segment-with-starting-coords
- Change the number of segments
- Confirm that the number changes in the right panel, and it matches the preview in the left panel
- Go to http://localhost:6006/?path=/story/perseuseditor-editorpage--interactive-graph-polygon
- Change the number of sides 
- Check the "show angle measures" checkbox
- Check the "show side measure" checkbox
- Confirm that the right side panel updates accordingly and matches the left side panel

### DIFF
--- a/packages/perseus-editor/src/widgets/interactive-graph-editor.tsx
+++ b/packages/perseus-editor/src/widgets/interactive-graph-editor.tsx
@@ -433,7 +433,6 @@ class InteractiveGraphEditor extends React.Component<Props> {
                                             },
                                             graph: {
                                                 ...this.props.graph,
-                                                type: "polygon",
                                                 showAngles:
                                                     !this.props.graph
                                                         .showAngles,
@@ -467,8 +466,7 @@ class InteractiveGraphEditor extends React.Component<Props> {
                                             graph: {
                                                 ...this.props.graph,
                                                 showSides:
-                                                    !this.props.correct
-                                                        .showSides,
+                                                    !this.props.graph.showSides,
                                             },
                                         });
                                     }
--- a/packages/perseus-editor/src/widgets/interactive-graph-editor.tsx
+++ b/packages/perseus-editor/src/widgets/interactive-graph-editor.tsx
@@ -317,6 +317,10 @@ class InteractiveGraphEditor extends React.Component<Props> {
                                         type: "point",
                                         numPoints: points,
                                     },
+                                    graph: {
+                                        type: "point",
+                                        numPoints: points,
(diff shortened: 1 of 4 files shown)


packages/perseus-editor/src/widgets/interactive-graph-editor.tsx
```diff
@@ -439,13 +456,22 @@ class InteractiveGraphEditor extends React.Component<Props> {
...
+                                            correct: {
+                                                ...this.props.correct,
+                                                showSides:
+                                                    !this.props.correct
+                                                        .showSides,
+                                            },
+                                            graph: {
+                                                ...this.props.graph,
+                                                showSides:
+                                                    !this.props.correct
```
Myranae, line 470: Question: I might have asked this when we were pairing, but, should this be from `graph` instead of `correct`?

Myranae (approved):
Thanks so much for fixing this! Reviewing this made me wonder why we use the `correct` object for all the checks in the `interactive-graph-editor.tsx` file. I think I'll bring this up to discuss during a sync, but I'm glad to see your new checks working off of `graph`!

packages/perseus-editor/src/widgets/interactive-graph-editor.tsx
nishasy: Yes! Graph and correct are largely the same, but we don't want `coords` to come from `correct`. That would result in the right side preview changing to the correct answer whenever the checkbox is checked. We want `coords` (and other props) to be the same as whenever `graph` already was.

nishasy: Yes! I think it doesn't matter that much since they should be the same, but it should be `graph` as a principle. And graph also has better type safety 👍🏽 I'll change that.

nishasy: I did it right for angles but not sides 🤦🏽‍♀️

PR #1376, opened 2024-06-25
Add the protractor to Mafs graphs

benchristel:
This PR ports the protractor to Mafs interactive graphs, but does not add
any new behavior (e.g. keyboard or screenreader accessibility).

Issue: LEMS-1862

## Test plan:

Paste the following into the `yarn dev` flipbook:

```json
{"content":"The graph of $y=h(x)$ is a line segment joining the points $(1,9)$ and $(3,2)$. \n\n**Drag the endpoints of the segment below to graph $y=h^{-1}(x)$.\n **\n\n[[☃ interactive-graph 1]]","images":{},"widgets":{"interactive-graph 1":{"alignment":"default","graded":true,"options":{"backgroundImage":{"bottom":0,"height":425,"left":0,"scale":1,"url":"web+graphie://ka-perseus-graphie.s3.amazonaws.com/a82b4098da55c75b68d2c1fd9f348923b8935ccb","width":425},"correct":{"coords":[[[9,1],[2,3]]],"type":"segment"},"graph":{"type":"segment"},"gridStep":[1,1],"labels":["x","y"],"markings":"none","range":[[-10,10],[-10,10]],"rulerLabel":"","rulerTicks":10,"showProtractor":false,"showRuler":false,"snapStep":[0.5,0.5],"step":[1,1],"showProtractor":true},"type":"interactive-graph","version":{"major":0,"minor":0}}}}
```

Verify:

- You can move and rotate the protractor
- You cannot lose the protractor outside the graph bounds
- Nothing has regressed from the legacy graph

Note that the above graph is a segment graph. It would be more realistic
to test the protractor with an angle graph, but those aren't ported
to Mafs yet.

https://github.com/Khan/perseus/assets/693920/e995d643-cee1-4421-89fd-deebe8548f05


nishasy (approved):
Yay! Admittedly, I don't think I scrutinized this PR too much since we want to get rid of the protractor soon anyway, but the overall implementation makes sense to me.

Nit: I think the old arrow still allowed the numbers to be seen, whereas the new one doesn't? Not sure if that's a priority though.

| Old | New |
| --- | --- |
| Screenshot 2024-06-25 at 5 54 31 PM | Screenshot 2024-06-25 at 5 54 43 PM |

packages/perseus/src/widgets/interactive-graphs/utils.ts
```diff
@@ -53,3 +53,33 @@ export function snap(snapStep: vec.Vector2, point: vec.Vector2): vec.Vector2 {
         Math.round(requestedY / snapY) * snapY,
     ];
 }
+
+// Returns the closest point to the given `point` that is within the graph
+// bounds given in `state`.
+export function bound({
```
Myranae, line 59: Praise: I like having these in the utils file! Nice change :D

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -475,7 +475,7 @@ const getDeltaVertex = (
 const getChange = (
     coords: readonly vec.Vector2[],
     delta: vec.Vector2,
-    constraintOpts: Omit<ConstraintArgs, "point">,
+    constraintOpts: {snapStep: vec.Vector2; range: [Interval, Interval]},
```
Myranae, line 478: Question: Do we lose anything by disconnecting `constraintOpts` from `ConstraintArgs`? I think I like it being more clearly stated like this, but curious if there was a reason to have it there other than to not repeat the types.

packages/perseus/src/util/svg.ts
```diff
@@ -30,7 +55,7 @@ export class PathBuilder {
     ): PathBuilder {
         this.path.push({
             action: "C",
-            coords: [control1X, control1Y, control2X, control2Y, endX, endY],
+            args: [control1X, control1Y, control2X, control2Y, endX, endY],
```
Myranae, line 58: Question: Is the change to `args` to better match the code you are bringing in or does that just better reflect this information?

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
benchristel: We don't lose any type safety by doing this. It is more verbose to write the type out explicitly, but I agree that it's clearer.

benchristel:
> Nit: I think the old arrow still allowed the numbers to be seen, whereas the new one doesn't? Not sure if that's a priority though.

Ooh, good catch. I don't think it actually matters, though, because none of the graphs that use the protractor have axes or a grid (or any background images).

packages/perseus/src/util/svg.ts
benchristel: Previously, this was called `coords` because all the values were in fact coordinates. But the `A` (arc) command for SVG paths has some "flag" arguments too, which determine the direction of the arc. I changed the name to reflect that.

PR #1380, opened 2024-06-27
[Interactive Graph] Update Circle graph type to use  instead of

nishasy:
## Summary:
I was getting a lot of annoying type errors from the Circle graph type
being a little inconsistent with the other graph types. Specifically,
all the other graph types have `coords` as a field, whereas circle uses
`center` to do essentially the same thing.

To avoid type issues for upcoming work, I'm updating the circle graph type
to use `coords` instead of `center`.

Issue: none

## Test plan:
`yarn jest`
`yarn typecheck`
`yarn run lint`

storybook
- https://khan.github.io/perseus/?path=/story/perseuseditor-editorpage--interactive-graph-circle-with-starting-coords
- Confirm that moving the left preview circle updates the "Correct answer" box correctly.

### DIFF
--- a/packages/perseus/src/widgets/interactive-graph.tsx
+++ b/packages/perseus/src/widgets/interactive-graph.tsx
@@ -103,7 +103,6 @@ const deprecatedProps = {
 const _getShouldShowInstructions: (arg1: Props) => boolean = (props) => {
     return (
         _isClickToAddPoints(props) &&
-        // @ts-expect-error - TS2339 - Property 'coords' does not exist on type 'PerseusGraphType'. | TS2339 - Property 'coords' does not exist on type 'PerseusGraphType'.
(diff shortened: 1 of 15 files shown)


packages/perseus/src/widgets/interactive-graph.tsx
```diff
@@ -1909,8 +1907,8 @@ class InteractiveGraph extends React.Component<Props, State> {
         graph: PerseusGraphType,
         props: Props,
     ): ReadonlyArray<Coord> {
+        // @ts-expect-error - TS2339 - Property 'coords' does not exist on type 'PerseusGraphType'.
```
nishasy, line 1910: I have no idea why these had to be moved.

nishasy:
Never mind, this PR is breaking the experience.

PR #1381, opened 2024-06-27
Add a feature flag for Mafs point graphs with fixed numPoints

benchristel:
## Summary:
The point graph has two separate user experiences:

- fixed number of points, where you can drag points around, but not add
  or delete them
- unlimited points, where you can add and delete points

We have built the first of these experiences in Mafs, but not the
second. We want to release Mafs graphs with fixed-numPoints before
implementing unlimited points.

This commit adds logic to make that possible with feature flags. It also
moves the existing (similar) logic for polygon graphs so all decisions
about which graph to render are made in one, testable place.

Issue: LEMS-2126

Test plan:

Deploy a webapp ZND. Visit
`/internal-courses/test-everything/test-everything-1/te-interactive-graph/a/survey-of-graph-types`
and verify that only the graphs that should use Mafs are doing so.

### DIFF
--- a/.changeset/tricky-rice-nail.md
+++ b/.changeset/tricky-rice-nail.md
@@ -2,4 +2,4 @@
 "@khanacademy/perseus": minor
 ---
 
-Add a "point-fixed" flag to the Mafs feature flags, to control whether point graphs with fixed numbers of points should use Mafs.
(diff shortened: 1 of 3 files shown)


nishasy (approved):
LGTM

PR #1383, opened 2024-06-28
View Locked Function in the Interactive Graph

mark-fitzgerald:
## Summary:
Existing locked figures include lines that are straight. The need exists to also have lines that follow a function/equation. Plotting arbitrary functions as a locked figure will provide content creators with a powerful drawing tool for interactive graphs.

Issue: LEMS-1944

## Test plan:
1. Navigate to Perseus => Widgets => Interactive Graph => [Locked Functions](http://localhost:6006/?path=/docs/perseus-widgets-interactive-graph-locked-functions--docs)
1. Note that there are multiple stories available to demonstrate various functions and options

## Examples:
**Quadratic**
Quadratic

**Sine**
Sine

**Logarithmic**
Logarithmic


packages/perseus/src/perseus-types.ts
```diff
@@ -731,6 +736,19 @@ export type LockedPolygonType = {
     strokeStyle: "solid" | "dashed";
 };
 
+export type LockedFunctionType = {
+    type: "function";
+    color: LockedFigureColor;
+    strokeStyle: "solid" | "dashed";
+    equation: string;
+    equationParsed?: {
```
nishasy, line 744: Could you add comments here about the usage for / difference between `equation` and `equationParsed`?

packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
```diff
@@ -257,7 +260,7 @@ describe("a mafs graph", () => {
             expect(points).toHaveLength(2);
         });
 
-        test("should render locked points with styles", async () => {
+        it("should render locked points with styles", async () => {
```
nishasy, line 263: 😢 I don't think this needed to be changed, since `it` is an alias of `test`. No need to change it back now though.

packages/perseus/src/widgets/interactive-graphs/interactive-graph-question-builder.ts
```diff
@@ -252,6 +258,20 @@ class InteractiveGraphQuestionBuilder {
...
+            color: "grayH",
+            strokeStyle: "solid",
+            directionalAxis: "x",
+            ...options,
+        };
+
+        this.addLockedFigure(lockedFunction);
+        return this;
+    }
+
```
nishasy, line 274: Don't forget to add a unit test for this in `interactive-graph-question-builder.test.ts`!

nishasy (approved):
THIS IS SO EXCITING! A few small comments, but nothing blocking.

packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
mark-fitzgerald: Agreed that it isn't necessary. I saw that we had a mix of uses (`it` vs `test`) and thought that I would clean things up for consistency.

packages/perseus/src/perseus-types.ts
mark-fitzgerald: Added.

packages/perseus/src/widgets/interactive-graphs/interactive-graph-question-builder.ts
mark-fitzgerald: Good catch. Thanks.

Added.

packages/perseus/src/perseus-types.ts
```diff
@@ -731,6 +736,23 @@ export type LockedPolygonType = {
     strokeStyle: "solid" | "dashed";
 };
 
+export type LockedFunctionType = {
+    type: "function";
+    color: LockedFigureColor;
+    strokeStyle: "solid" | "dashed";
+    equation: string; // This is the user-defined equation (as it was typed)
+    equationParsed?: {
```
jeremywiebe, line 744: @mark-fitzgerald  The types in this file represent the "at rest" data representations for Perseus Data. As such, I try to avoid having keys that would never show up in Perseus JSON/data. 

As I read this PR, this key would be created the first time this locked figure was rendered and then "cached" here for future renders, right?  

My concern is folks thinking they can put data into this key in Perseus JSON and beginning to misuse the intent of this key (which I read as a memoization/cache of the parsed equation).

packages/perseus/src/widgets/interactive-graphs/locked-figures/locked-function.tsx
```diff
@@ -0,0 +1,30 @@
...
+import type {LockedFunctionType} from "../../../perseus-types";
+
+const LockedFunction = (props: LockedFunctionType) => {
+    const {color, strokeStyle, directionalAxis, domain} = props;
+    const plotProps = {
+        color: lockedFigureColors[color],
+        style: strokeStyle,
+        domain,
+    };
+    const equation = props.equationParsed || KAS.parse(props.equation).expr;
```
jeremywiebe, line 16: When is `equationParsed` populated?

packages/perseus/src/perseus-types.ts
mark-fitzgerald: Ah, thanks for that clarification. I wasn't aware of this. I will refactor this accordingly when I implement the add/edit capability.

packages/perseus/src/widgets/interactive-graphs/locked-figures/locked-function.tsx
mark-fitzgerald: I was initially thinking that it would be populated when the content editor types the equation. I'm not sure how that would persist, though, so I'm rethinking the timing/management of this.

PR #1384, opened 2024-06-28
[Expression Remediation] Create arrow key navigation for TabBar component.

catandthemachines:
## Summary:
Updating TabBar and TabbarItem component to support arrow key navigation.

Included in this PR
- The creation of arrow key navigation for Tabbar and specifically the TabItems used for changing the presenting panel.
- Implementation of functionality for tab bar items, specifically the focus property and ability to change the role type.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2130

## Test plan:
Storybook
- http://localhost:6006/?path=/story/math-input-full-keypad--everything-minus-navigation-pad
- Confirm that tabbing goes through only the selected tabitem, and to select another tab a user must use their left and right arrow keys.


catandthemachines:
Still need to add unit tests and test out a few cases but here is a rough POC of how we can add the the arrow key navigation.

packages/math-input/src/components/tabbar/item.tsx
```diff
@@ -75,70 +76,146 @@ function imageTintColor(
...
-                        focused,
-                        pressed,
-                    );
+type TabItemProps = {
+    onClick: () => void;
+    itemState: ItemState;
+    itemType: KeypadPageType;
+};
+
+export function ArrowKeyTabbarItem(
```
handeyeco, line 94: The render functions for `ArrowKeyTabbarItem` and `TabbarItem` are almost 99% identical. Regularly we run into issues in Perseus when copy/pasting code because the two copies grow and diverge and after awhile the reasons for copying the code gets lost to time.

I think this might be an opportunity to consolidate these two components and make one, flexible component or to use composition to separate the focus logic from the component logic.

packages/math-input/src/components/tabbar/item.tsx
```diff
@@ -75,70 +76,146 @@ function imageTintColor(
     return color.offBlack64;
 }
 export type ItemState = "active" | "inactive" | "disabled";
-type Props = {
+type ArrowKeyTabItemProps = {
     onClick: () => void;
     itemState: ItemState;
     itemType: KeypadPageType;
+    focus: boolean;
+    setFocus: React.Dispatch<React.SetStateAction<number>>;
```
handeyeco, line 84: I wonder if we could simplify this to just `(index: number) => void` and wrap the setter when providing the prop.

Here:

```suggestion
    setFocus: (index: number) => void;
```

In `tabbar.tsx`:

``` JSX
                       setFocus={(index: number) => setFocus(index)}
```

packages/math-input/src/components/tabbar/item.tsx
```diff
@@ -75,70 +76,146 @@ function imageTintColor(
...
+    itemState: ItemState;
+    itemType: KeypadPageType;
+};
+
+export function ArrowKeyTabbarItem(
+    props: ArrowKeyTabItemProps,
+): React.ReactElement {
+    const {onClick, itemType, itemState, focus, setFocus, index} = props;
+
+    const ref = useRef(null);
```
handeyeco, line 99: 1. Tiny nit: I think `ref` could use a clearer name.
2. I think if this were typed you could avoid suppressing a TS error below; something like `const ref = useRef<Clickable>()`

packages/math-input/src/components/tabbar/item.tsx
```diff
@@ -75,70 +76,146 @@ function imageTintColor(
...
 
-                    return (
+    useEffect(() => {
+        if (focus) {
+            // Move element into view when it is focused
+            // @ts-expect-error - TS2339 - Property 'current' does not exist on type 'never'.
+            ref.current?.focus();
```
handeyeco, line 105: This is the TS error I was talking about. IIRC `@ts-expect-error` suppresses **all** TS errors. This was recently an issue that I ran into: someone did a find/replace and a very obvious bug made it to prod because someone had suppressed a totally unrelated error.

packages/math-input/src/components/tabbar/item.tsx
```diff
@@ -7,6 +7,7 @@ import * as React from "react";
 import IconAsset from "./icons";
 
 import type {KeypadPageType} from "../../types";
```
handeyeco, line 9: All of these GitHub Actions errors makes me think your IDE might be messed up; I think a lot of these could be auto-fixed for you by your IDE. LMK if you want help troubleshooting.

packages/math-input/src/components/tabbar/tabbar.tsx
```diff
@@ -31,17 +32,25 @@ type Props = {
...
-                {items.map((item) => (
-                    <TabbarItem
+                {items.map((item, index) => (
+                    <ArrowKeyTabbarItem
                         key={`tabbar-item-${item}`}
                         itemState={
                             item === selectedItem ? "active" : "inactive"
                         }
                         itemType={item}
+                        index={index}
```
handeyeco, line 47: I'm sure you'll notice the GH errors, but I don't think you need this prop.

packages/math-input/src/components/tabbar/tabbar.tsx
```diff
@@ -62,4 +71,35 @@ function Tabbar(props: Props): React.ReactElement {
     );
 }
 
+// Custom focus hook to handle arrow key navigation for the TabBar for each ArrowKeyTabItem.
+function useArrowKeyFocus(size) {
```
handeyeco, line 75: Open to pushback on this, but:
- If this logic is only used by this component, could it just be part of the component? Or is it just cleaner to split it out?
- If this logic could be used by other components, could we split it out into its own file?

packages/math-input/src/components/tabbar/tabbar.tsx
```diff
@@ -62,4 +71,35 @@ function Tabbar(props: Props): React.ReactElement {
...
+        },
+        [size, currentFocus, setCurrentFocus],
+    );
+
+    useEffect(() => {
+        document.addEventListener("keydown", handleKeyDown, false);
+        return () => {
+            document.removeEventListener("keydown", handleKeyDown, false);
+        };
+    }, [handleKeyDown]);
```
handeyeco, line 100: After reading [this article](https://kentcdodds.com/blog/usememo-and-usecallback), I've been a skeptic of `useCallback` without a demonstrable need. Is there a reason this doesn't work?

```suggestion
    function handleKeyDown(e) {
        if (e.keyCode === 39) {
            // Right arrow
            e.preventDefault();
            setCurrentFocus(
                currentFocus === size - 1 ? size - 1 : currentFocus + 1,
            );
        } else if (e.keyCode === 37) {
            // Left arrow
            e.preventDefault();
            setCurrentFocus(currentFocus === 0 ? 0 : currentFocus - 1);
        }
    };

    useEffect(() => {
      document.addEventListener("keydown", handleKeyDown, false);
      return () => {
          document.removeEventListener("keydown", handleKeyDown, false);
      });
```

handeyeco (commented):
Looking good! I noticed this is still a draft, so hopefully you don't mind the early feedback. Curious about the test plan, but I think RTL can handle focus tests?

packages/math-input/src/components/tabbar/item.tsx
catandthemachines: I was getting some weird errors when attempting to do just this. But let me try again 😉

packages/math-input/src/components/tabbar/tabbar.tsx
catandthemachines: Ideally in the future we should work with WonderBlocks to use their future TabPanel which will have this functionality baked in. I see this as a temporary stop-gap to ensure users can navigate via arrow key until we can align on a longer term implementation.

But I can totally move this into a utility function as well 😊

.changeset/polite-snails-know.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/math-input": minor
+---
+
+Updating tabbar experience to use arrow-key navigation to access the other tabitems.
```
mark-fitzgerald, line 5: This comment is typically echoed in release notes, etc. So, including the widget name somewhere in the description might be helpful.

packages/math-input/src/components/keypad/keypad.tsx
```diff
@@ -151,7 +151,6 @@ export default function Keypad(props: Props) {
                 <View style={styles.keypadInnerContainer}>
                     <View
                         style={[styles.keypadGrid, gridStyle]}
-                        tabIndex={0}
                         aria-label="Keypad"
```
mark-fitzgerald, line 154: With the `tabIndex` removed, is the `aria-label` still needed?

catandthemachines:
> Looking good! I noticed this is still a draft, so hopefully you don't mind the early feedback. Curious about the test plan, but I think RTL can handle focus tests?

I added a new test file to test the focus functionality for item.tsx. I tried to add some tests in tabbar.test.tsx, but I couldn't get the mocking of the arrow-key behavior to work. Let me know if I should take another stab at it.

packages/math-input/src/components/tabbar/tabbar.tsx
catandthemachines: Totally agree with you, it was a section of code I pulled from some examples online. I haven't used useCallback very often and was looking to remove/refactor anyway as it seemed gratuitous. Deleted! :)

packages/math-input/src/components/keypad/keypad.tsx
catandthemachines: Let me check and see how it sounds before removing the aria-label.

packages/math-input/src/components/tabbar/tabbar.tsx
```diff
@@ -32,16 +33,41 @@ type Props = {
...
+
+    const [focus, setFocus] = useState(
+        selectedIndex === -1 ? 0 : selectedIndex,
+    );
+
+    // Custom function to handle arrow key navigation for the TabBar for each TabItem.
+    const useArrowKeyFocus = (e) => {
+        if (e.keyCode === 39) {
+            // Right arrow
+            setFocus(focus === items.length - 1 ? items.length - 1 : focus + 1);
```
mark-fitzgerald, line 48: Just curious: Do we not want the focus to cycle to the other side of the tabbar when it reaches one of the ends? I guess I was expecting it to act like a menu where after reaching the end, the next arrow would put focus back on the first item. Maybe that's just me.

mark-fitzgerald: I guess maybe for a tablist, this is the correct behavior. Just a curiousity.

catandthemachines: For me it felt like it would be more confusing if it cycled back, but I'm good with either implementation. Let me ask Caitlyn real quick.

mark-fitzgerald: So, it looks like cycling back to the other end is a proper (though optional) behavior: [ARIA Tabs Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/)

I also noticed [ARIA Toolbar Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/toolbar/), and wondered if that might be more appropriate here (for `role` definition). Same expectations for arrow key navigation. I'm not saying that this needs changing, just pointing out another option, in case it helps.

packages/math-input/src/components/keypad/keypad.tsx
mark-fitzgerald: Just curious - how did the screen reader test go with/without the `aria-label`?

mark-fitzgerald (approved):
Feels cleaner in here - thanks!

PR #1385, opened 2024-07-01
[Interactive Graph Editor] Remove the use of graphKey for remounting

nishasy:
## Summary:
It turns out that the use of `graphKey` for remounting in
`interactive-graph-editor.tsx` is redundant, since this is also being
done by the `useEffect` in `mafs-graph.tsx`.

`graphKey` was also causing a warning to show up in the logs, possibly
from a remounting loop.

Here, `graphKey` is being removed altogether to stop the remounting issue.

[More info on Slack](https://khanacademy.slack.com/archives/C067UM1QAR4/p1719854190583499)

Issue: none

## Test plan:
Storybook
- Add an interactive graph
- Open the browser console
- Move the graph in the "correct" preview (quickly! Don't hold for time before dragging)
- Confirm the warning doesn't appear
- Change properties of the graph (e.g. number of segments)
- Confirm that the preview changes correctly
- Repeat for every graph type

### DIFF
--- a/.changeset/heavy-buckets-think.md
+++ b/.changeset/heavy-buckets-think.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": patch
+---
+
+[Interactive Graph Editor] Remove the use of graphKey for remounting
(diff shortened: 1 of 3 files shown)


benchristel (approved):
LGTM! Thanks for debugging this issue and fixing it :)

PR #1388, opened 2024-07-01
[Interactive Graph: Circle] Add a key prop to the circle drag handle

nishasy:
## Summary:
There's always a console error saying that the circle graph drag handle
is missing the `key` prop.

Adding a `key` here.

Issue: none

## Test plan:
Open the console and make sure the error isn't showing up anymore.

Make sure the logs don't show the console error anymore.

Screenshot 2024-07-01 at 3 10 31 PM

### DIFF
--- a/.changeset/sour-cougars-draw.md
+++ b/.changeset/sour-cougars-draw.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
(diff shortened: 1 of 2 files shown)


benchristel (approved):
Oh, this is because the six dots in the drag handle are generated from coordinates. Thanks for the fix!

catandthemachines (approved):

PR #1390, opened 2024-07-03
Move math functions to a `math/` folder

benchristel:
Note to reviewers: it's going to be easiest to review this PR one commit
at a time!

This is the first part of a multi-part refactor. I am submitting this
intermediate PR to get feedback and hopefully avoid running into too many
merge conflicts.

In future PRs, I plan to:

- move more functions into `math/`
- reuse the existing functions in more places
- re-export functions from `kmath` in `math/index.ts`, possibly
  wrapping them with more suitable type signatures
- fix TODOs added in this PR.

Adding a `math/` folder gives us a clear place to put new mathy functions,
and will help us locate and reuse the functions that already exist.

Issue: LEMS-2135

## Test plan:

`yarn test`

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/math/box.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/math/box.ts
@@ -1,6 +1,6 @@
 import {clamp} from "./clamp";
 import {X, Y} from "./coordinates";
-import {MAX, MIN} from "./interval";
+import {trim} from "./interval";
 
 import type {Interval, vec} from "mafs";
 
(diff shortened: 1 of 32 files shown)


packages/perseus/src/widgets/interactive-graphs/math/box.ts
```diff
@@ -0,0 +1,41 @@
...
+// `amount`. If a component of `amount` is negative, the box is expanded in
+// that dimension instead.
+export function inset(amount: vec.Vector2, box: Box): Box {
+    return ensureValid([
+        [box[X][MIN] + amount[X], box[X][MAX] - amount[X]],
+        [box[Y][MIN] + amount[Y], box[Y][MAX] - amount[Y]],
+    ]);
+}
+
+export function ensureValid(box: Box): Box {
```
mark-fitzgerald, line 25: I like the comments for the other functions. Perhaps we could have one here, too? Mostly because I don't know what setting the min/max values to the average does.

benchristel: I think `ensureValid` is the wrong abstraction. The behavior of setting the values to the average is specific to `inset`. The purpose of setting min and max to the average is that the center of the box should not be changed by `inset`.

benchristel: I've reworked this code. Let me know what you think!

mark-fitzgerald: That makes a LOT more sense! Thank you!!

mark-fitzgerald (approved):
The singular-function files (i.e. `clamp.ts` and `coordiantes.ts`) feel a bit light to me, but I understand the intent. I think that the organization and comments and tests help greatly to make this more readable.

PR #1392, opened 2024-07-03
Create Angle Mafs Graph 

SonicScrewdriver:
## Summary:
This PR adds the Angle graph type to the new Mafs-based Interactive Graph widget. This should work identically to the legacy Angle graph, albeit with some fixed functionality regarding reflexive angles. 

The graph supports: 
- Snapping to whole angles 
- Snapping to custom step sizes (snapDegrees)
- Allowing for reflexive or non-reflexive angles 
- Offsetting the angle by a number of degrees
- Showing or hiding the angle

There are some pre-existing bugs that I've found, but have elected to either solve at a later date, or to simply ignore: 
- The original graph coordinate order is backwards, imo, and not intuitive. As a result, the tab order is also unintuitive. Unfortunately, I can't change this without breaking the scoring for the legacy graphs. (Going to give this one last test before making a ticket). I'd recommend looking into fixing this after removing the legacy graph so that we can update the scoring logic that is currently shared. 
- The original graph ensures that the vertex can never be less than 30px away from the end points. Unfortunately our reducer does not have a way to access any of the pixel dimensions or hooks for recreating this logic. For now, I have set it up so that the vertex cannot get any closer than 2 graph units to the end points, which should work for all the current graphs in our content. It is possible that a new reducer action could be created for moving JUST angle graphs -- but, after checking the content, it seems like this would be unnecessary. 
- Graphs that don't use a "balanced" range (ie [5, 10]) results in the angle indicators warping. Given that all angle graphs have ranges of [-10,10], this doesn't appear to be an issue worth spending additional time on. I am unable to think of a use case or reason why an angle graph would ever require an unbalanced ranged - in fact, most turn off the gridlines entirely. 

Final Note:
- I'd love to get design input regarding the angle indicator. I feel like the text should probably be blue as it's essentially an "interactive" element, in a way. 

Issue: LEMS-1827

## Test plan:
- new tests 
- manual testing of all edge cases in flipbook using a set of hand-built graphs, and a couple live-questions. Feel free to message me for the json

## Screenshots:
Screenshot 2024-07-04 at 7 55 58 AM

## Video:
https://github.com/Khan/perseus/assets/12463099/92795848-81dd-4350-89a0-6efa8550327d


packages/perseus/src/widgets/interactive-graphs/mafs-styles.css
```diff
@@ -257,3 +257,10 @@
     text-align: center;
     width: var(--x-axis-label-width, 20px);
 }
+
```
SonicScrewdriver, line 260: Having the angle be black really looked wrong to me. I feel like the angle text should also be blue, but I'll double check with design on that one.

packages/perseus/src/widgets/interactive-graphs/reducer/initialize-graph-state.ts
```diff
@@ -344,3 +352,48 @@ function getCircleCoords(graph: PerseusGraphTypeCircle): {
...
+    const offset = ((angleOffsetDeg || 0) * Math.PI) / 180;
+
+    let defaultCoords: [Coord, Coord] = [
+        [0.85, 0.5],
+        [0.5, 0.5],
+    ];
+
+    defaultCoords = normalizePoints(range, step, defaultCoords, true);
+
+    // @ts-expect-error - TS2345 - Argument of type 'number[]' is not assignable to parameter of type 'readonly Coord[]'. | TS2556 - A spread argument must either have a tuple type or be passed to a rest parameter.
```
SonicScrewdriver, line 382: I wasn't sure about whether I should bother with this expect error that got carried over from the old code. It would require creating new properly typed functions for both magnitude and vector, and seemed like a rabbit hole that wasn't worth diving into yet.

packages/perseus/src/widgets/__testdata__/interactive-graph-random.testdata.ts
```diff
@@ -37,7 +37,9 @@ const randomGraphTypeAngle = (): PerseusGraphTypeAngle => {
         snapDegrees: randomInteger(1, 90),
         match: "congruent",
         // must be 3 coords for an angle graph
```
SonicScrewdriver, line 39: Just create a literal array and call randomCoord 3 times

packages/perseus/src/widgets/__testdata__/interactive-graph.testdata.ts
```diff
@@ -63,6 +63,64 @@ export const angleQuestion: PerseusRenderer = {
...
+                    height: 400,
+                    left: 0,
+                    scale: 1,
+                    url: "https://ka-perseus-graphie.s3.amazonaws.com/807ea77cf7031c1b9a45694083f05b5e09b01946.png",
+                    width: 400,
+                },
+                correct: {
+                    allowReflexAngles: false,
+                    angleOffsetDeg: 0,
+                    coords: [
```
SonicScrewdriver, line 86: Could make this a right angle?

packages/perseus/src/widgets/interactive-graphs/graphs/angle.tsx
```diff
@@ -0,0 +1,113 @@
...
+import {Vector} from "./components/vector";
+import {useTransformVectorsToPixels} from "./use-transform";
+import {getIntersectionOfRayWithBox} from "./utils";
+
+import type {CollinearTuple} from "../../../perseus-types";
+import type {AngleGraphState, MafsGraphProps} from "../types";
+import type {vec, Interval} from "mafs";
+
+type AngleGraphProps = MafsGraphProps<AngleGraphState>;
+type AngleProps = {
```
SonicScrewdriver, line 19: Should export this type from the indicators file

packages/perseus/src/widgets/interactive-graphs/graphs/angle.tsx
```diff
@@ -0,0 +1,113 @@
...
+        coords,
+        showAngles,
+        range,
+        allowReflexAngles,
+        angleOffsetDeg,
+        snapDegrees,
+    } = graphState;
+
+    // Break the coords into the two end points and the center point
+    const endPoints = [coords[0], coords[2]] as [vec.Vector2, vec.Vector2];
```
SonicScrewdriver, line 43: satisfies

packages/perseus/src/widgets/interactive-graphs/graphs/angle.tsx
```diff
@@ -0,0 +1,113 @@
...
+    };
+
+    // Render the lines, angle, and then movable points
+    return (
+        <>
+            {svgLines}
+            <Angle {...angleParams} />
+            {coords.map((point, i) => (
+                <StyledMovablePoint
+                    key={"point-" + i}
```
SonicScrewdriver, line 103: Could be anglePoint

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.tsx
```diff
@@ -0,0 +1,428 @@
...
+    const clockwiseCoords = shouldReverseCoords ? coords : coords.reverse();
+
+    // Calculate the angles between the two points
+    const startAngle = findAngle(clockwiseCoords[0], vertex);
+    const endAngle = findAngle(clockwiseCoords[1], vertex);
+    const angle = (startAngle + 360 - endAngle) % 360;
+    const halfAngle = (endAngle + angle / 2) % 360;
+    const angleLabel = parseFloat(angle.toFixed(0)); // Only want to show whole angles
+
+    // Check if the angle is reflexive and if we should allow it
```
SonicScrewdriver, line 170: no longer checking if we should allow it

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.tsx
```diff
@@ -0,0 +1,428 @@
...
+                    y="0%"
+                    height="100%"
+                >
+                    <feFlood floodColor="#FFF" floodOpacity="0.5" />
+                    <feComposite operator="over" in="SourceGraphic" />
+                </filter>
+            </defs>
+
+            {!isReflexive && isRightAngle(angle) ? (
+                <RightAngleSquare
```
SonicScrewdriver, line 237: Update CSS so new arc styles also target right angle square

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.tsx
```diff
@@ -0,0 +1,428 @@
...
+                    start={[x1, y1]}
+                    vertex={[x2, y2]}
+                    end={[x3, y3]}
+                />
+            ) : (
+                <MafsCssTransformWrapper>
+                    <path
+                        d={arc}
+                        strokeWidth="2000"
```
SonicScrewdriver, line 246: Fix stroke width. This was a test that discovered the stroke width is doing nothing.

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.tsx
```diff
@@ -0,0 +1,428 @@
...
+            </defs>
+
+            {!isReflexive && isRightAngle(angle) ? (
+                <RightAngleSquare
+                    start={[x1, y1]}
+                    vertex={[x2, y2]}
+                    end={[x3, y3]}
+                />
+            ) : (
+                <MafsCssTransformWrapper>
```
SonicScrewdriver, line 243: Add an Arc component?

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.tsx
```diff
@@ -0,0 +1,428 @@
...
+                    width="110%"
+                    y="0%"
+                    height="100%"
+                >
+                    <feFlood floodColor="#FFF" floodOpacity="0.5" />
+                    <feComposite operator="over" in="SourceGraphic" />
+                </filter>
+            </defs>
+
+            {!isReflexive && isRightAngle(angle) ? (
```
SonicScrewdriver, line 236: Don't need isReflexive here

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.tsx
```diff
@@ -0,0 +1,428 @@
...
+        const y = point1[1] - point2[1];
+        if (!x && !y) {
+            return 0;
+        }
+        return (180 + (Math.atan2(-y, -x) * 180) / Math.PI + 360) % 360;
+    }
+    return findAngle(point1, vertex) - findAngle(point2, vertex);
+};
+
+function scaledPolarRad(radius: number, radians: number): vec.Vector2 {
```
SonicScrewdriver, line 351: This is how the old text was being positioned, we have redone the math from scratch

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.tsx
```diff
@@ -0,0 +1,428 @@
...
+        y: sum.y / sumMagnitude,
+    };
+
+    // Calculate the initial distance of the bisector direction from the origin
+    const initialDistance = Math.sqrt(
+        bisectorDirection.x ** 2 + bisectorDirection.y ** 2,
+    );
+
+    // Determine the minimum radius to ensure the point is at least 2 units away in any direction
+    const requiredDistance = 3;
```
SonicScrewdriver, line 413: pass in radius just for future sanity rather than hard coding

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.tsx
```diff
@@ -0,0 +1,428 @@
...
+        radius = 1; // Use the unit vector as is
+    }
+
+    // Scale the bisector direction by the radius
+    const bisectorPoint = [
+        bisectorDirection.x * radius,
+        bisectorDirection.y * radius,
+    ] as vec.Vector2;
+
+    return vec.add(bisectorPoint, vertex);
```
SonicScrewdriver, line 427: Comment to explain the addition

packages/perseus/src/widgets/interactive-graphs/reducer/initialize-graph-state.ts
```diff
@@ -344,3 +352,48 @@ function getCircleCoords(graph: PerseusGraphTypeCircle): {
...
+    const radius = magnitude(vector(...defaultCoords));
+
+    const coords: [Coord, Coord, Coord] = [...defaultCoords, [0, 0]];
```
SonicScrewdriver, line 385: Make anote that 0,0 is just a placeholder to appease typescript

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -507,9 +534,181 @@ function boundAndSnapToGrid(
     return snap(snapStep, bound({snapStep, range, point}));
 }
 
-function boundAndSnapToAngle(
+function tooClose(
```
SonicScrewdriver, line 537: Put this below the vertex function

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -507,9 +534,181 @@ function boundAndSnapToGrid(
...
+        snapStep,
+        range,
+        point: snap(snapStep, destination),
+    });
+    const delta = kvector.add(newVertex, reverseVector(startingVertex));
+
+    let valid = true;
+    const newPoints: Record<string, any> = {};
+
+    _.each([0, 2], function (i) {
```
SonicScrewdriver, line 575: No more underscores!

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -507,9 +534,181 @@ function boundAndSnapToGrid(
...
+        const oldPoint = coordsCopy[i];
+        let newPoint = kvector.add(oldPoint, delta);
+
+        let angle = GraphUtils.findAngle(newVertex, newPoint);
+        angle *= Math.PI / 180;
+
+        newPoint = constrainToBoundsOnAngle(newPoint, angle, range, snapStep);
```
SonicScrewdriver, line 582: This works but could I just use bound and snap for the vertex since it's using the snap step

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -507,9 +534,181 @@ function boundAndSnapToGrid(
...
+        range: angleRange,
+        point: destinationPoint,
+    });
+    coordsCopy[index] = boundPoint;
+
+    // Get the vertex of the angle
+    const vertex = coords[1];
+
+    // Gets the angle between the coords and the vertex
+    let angle = GraphUtils.findAngle(coordsCopy[index], vertex);
```
SonicScrewdriver, line 693: Decide where the true source of findAngle should live. Graph utils probably

SonicScrewdriver: JK use angle-indicators or move it to a mafs-utils file

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -507,9 +534,181 @@ function boundAndSnapToGrid(
...
+
+    // Snap the angle to the nearest multiple of snapDegrees (if provided)
+    angle = Math.round((angle - offsetDegrees) / snap) * snap + offsetDegrees;
+    const distance = GraphUtils.getDistance(coordsCopy[index], vertex);
```
SonicScrewdriver, line 697: Check if there's a better getDistance OR move the logic to the same place as findAngle

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -507,9 +534,181 @@ function boundAndSnapToGrid(
...
+    const snappedValue = kvector.add(vertex, polar(distance, angle));
+
```
SonicScrewdriver, line 699: See if vec.add would work here instead of kvector.

Also move polar to same place as findAngle/getDistance.

jeremywiebe (approved):
Reviewed with Third during a pairing session! LGTM.

packages/perseus/src/widgets/interactive-graph.tsx
```diff
@@ -2134,20 +2134,18 @@ class InteractiveGraph extends React.Component<Props, State> {
         coords = InteractiveGraph.pointsFromNormalized(props, [
             [0.85, 0.5],
             [0.5, 0.5],
-        ]);
+        ]) as [Coord, Coord, Coord];
```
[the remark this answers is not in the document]
SonicScrewdriver: Oooo I think no, actually, as the return from pointsFromNormalized is a readonly array. 

This was a necessary type update to support the older graphs so that we can better types for the new graph. This code will be stripped out when we remove the legacy graph, so I wonder if it's better to just leave it as "as" for now.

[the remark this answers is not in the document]
jeremywiebe: If it's destined to be removed eventually, I think leaving it with `as` is fine for now.

PR #1393, opened 2024-07-03
Move tests next to implementation, and fix filename

benchristel:
## Summary:
The tests for axis ticks looked lonely hanging out in `__tests__`, so I
moved them next to the associated production code. I also renamed one
file from `tests.ts` to `test.ts` - the name was preventing Jest from
running the tests in that file.

Issue: none

## Test plan:

`yarn test`


mark-fitzgerald (approved):
Loving the spring/summer cleaning!

PR #1394, opened 2024-07-03
Move functions from protractor to `math/`

benchristel:
These math functions weren't tested, and had some lurking bugs which I've
now fixed.

Issue: LEMS-2135

## Test plan:

`yarn test`


jeremywiebe (approved):
LGTM. 👍

PR #1395, opened 2024-07-08
Create section in Storybook for Graph Editor

nicolecomputer:
## Summary:

Moves storybook stories about the graph editor into their own section.

Pasted_Image_7_8_24__12_02 PM-2

Issue: https://khanacademy.atlassian.net/browse/LEMS-2138 

## Test plan:
- Visit storybook
- **You should still be able to find all of the stories that were there previously**


packages/perseus-editor/src/__stories__/interactive-graph-editor.stories.tsx
```diff
@@ -0,0 +1,284 @@
...
+    DeviceType,
+    Hint,
+    PerseusAnswerArea,
+    PerseusRenderer,
+} from "@khanacademy/perseus";
+
+registerAllWidgetsAndEditorsForTesting(); // SIDE_EFFECTY!!!! :cry:
+
+export default {
+    title: "PerseusEditor/Interactive Graph",
```
jeremywiebe, line 37: Our other widget-level editors are placed in `PerseusEditor/Widgets/*`.  Eg ["Categorizer Editor"](https://github.com/Khan/perseus/blob/be7de3a5edb9c08b6976be508c56dd4253e3c86c/packages/perseus-editor/src/widgets/__stories__/categorizer-editor.stories.tsx#L14). 


What do you think of: 
```suggestion
    title: "PerseusEditor/Widgets/Interactive Graph",
```

jeremywiebe (approved):
Looks good but I suggest a minor change in where this new section lives.

packages/perseus-editor/src/__stories__/interactive-graph-editor.stories.tsx
nishasy: I was thinking something similar! There's already [Interactive Graph Editor](https://650db21c3f5d1b2f13c02952-vxdazuuqkt.chromatic.com/?path=/docs/perseuseditor-widgets-interactive-graph-editor--docs) in that section, though.

I think my personal preference would be to put it under `PerseusEditor/EditorPage/InteractiveGraph`, since the main thing it's rendering is still the `EditorPage` component, just specifically examples of interactive graphs. That should add an "Interactive Graph" folder under the `docs` and `demo` pages, which should then contain all the interactive graph stories inside.

PR #1397, opened 2024-07-09
Split StatefulMafsGraph and MafsGraph into separate files

benchristel:
Each of these components used different helper functions and types defined
alongside them.  I've moved each component into a separate file along with its
associated functions and types. This makes the files more cohesive, so it's
easier to see which groups of entities go together.

Issue: none

## Test plan:
`yarn test`


handeyeco (approved):
My understanding is this is just splitting up one file into two. If that's the case and there are no changes to logic, I think this makes a lot of sense.

PR #1398, opened 2024-07-09
Move getLines to polygon.tsx

benchristel:
## Summary:
It's only used by the polygon graph, and doesn't seem like a good candidate for
reuse (with a name like getLines, how would people find it?) so I moved it
closer to its usage site.

Issue: none

## Test plan:

`yarn test`

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx
@@ -11,7 +11,7 @@ import {TextLabel} from "./components/text-label";
 import {useDraggable} from "./use-draggable";
 
 import type {MafsGraphProps, PolygonGraphState} from "../types";
-import {CollinearTuple} from "@khanacademy/perseus";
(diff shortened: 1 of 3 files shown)


Myranae (approved):
I had a failing test for a second there after running `yarn test`, but I ran a `yarn install` and everything is passing now, so looks good to me!

packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx
```diff
@@ -9,9 +9,9 @@ import {PolygonAngle} from "./components/angle-indicators";
 import {StyledMovablePoint} from "./components/movable-point";
 import {TextLabel} from "./components/text-label";
 import {useDraggable} from "./use-draggable";
-import {getLines} from "./utils";
 
 import type {MafsGraphProps, PolygonGraphState} from "../types";
+import {CollinearTuple} from "@khanacademy/perseus";
```
Myranae, line 14: Looks like you might want to add `type` to this

PR #1399, opened 2024-07-09
Move `segmentsIntersect` to the math folder

benchristel:
and add tests. There are some interesting edge cases!

Issue: LEMS-2135

## Test plan:
`yarn test`

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/math/geometry.test.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/math/geometry.test.ts
@@ -16,7 +16,7 @@ describe("segmentsIntersect", () => {
     });
 
     it("returns false when segments are the same", () => {
-        // TODO(benchristel): is this desired behavior?
+        // intersecting segments must have a SINGLE intersection point.
         const segment1: Segment = [
             [0, 0],
             [1, 1],
@@ -29,7 +29,7 @@ describe("segmentsIntersect", () => {
     });
 
     it("returns false when an endpoint touches the other segment (lambda = 1)", () => {
-        // TODO(benchristel): is this desired behavior?
+        // We treat segments as open (exclusive of their endpoints).
         const segment1: Segment = [
             [0, 0],
             [1, 1],
@@ -42,7 +42,7 @@ describe("segmentsIntersect", () => {
     });
 
     it("returns false when an endpoint touches the other segment (lambda = 0)", () => {
-        // TODO(benchristel): is this desired behavior?
+        // We treat segments as open (exclusive of their endpoints).
         const segment1: Segment = [
             [0, 0],
             [1, 1],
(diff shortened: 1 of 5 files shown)


packages/perseus/src/widgets/interactive-graphs/math/geometry.test.ts
```diff
@@ -0,0 +1,178 @@
...
+        ];
+        const segment2: Segment = [
+            [1, 1],
+            [1, 1],
+        ];
+        expect(segmentsIntersect(segment1, segment2)).toBe(false);
+    });
+
+    it("returns false when segments are the same", () => {
+        // TODO(benchristel): is this desired behavior?
```
jeremywiebe, line 19: I _think_ it is mathematically.  [This](https://www.cuemath.com/geometry/intersection-of-two-lines/#) page says that intersecting lines always meet at a single point (two identical lines would meet at all points on both lines).

It also states that "The intersecting lines can cross each other at any angle. This angle formed is always greater than 0° and less than 180°." Again, identical lines would form an angle of 0°.

That page is basically the first hit for "do identical lines intersect" so it isn't authoritative... but seems to make logical sense. :)

packages/perseus/src/widgets/interactive-graphs/math/geometry.test.ts
```diff
@@ -0,0 +1,178 @@
...
+            [1, 1],
+        ];
+        const segment2: Segment = [
+            [0, 0],
+            [1, 1],
+        ];
+        expect(segmentsIntersect(segment1, segment2)).toBe(false);
+    });
+
+    it("returns false when an endpoint touches the other segment (lambda = 1)", () => {
```
jeremywiebe, line 31: This behaviour smells of a rounding issue. :) It feels like a segment ending somewhere on the other segment should intersect (given the "rules" from my earlier comment.  Probably something to verify with the Math content folks.

jeremywiebe (approved):
Makes sense!

packages/perseus/src/widgets/interactive-graphs/math/geometry.test.ts
benchristel: Changing the check from `lambda < 1` to `lambda <= 1` makes the function return `true` in this case. Rounding could also be an issue but my tests don't check for that.

[According to Wikipedia](https://en.wikipedia.org/wiki/Line_segment), line segments can be closed (including both endpoints), open (excluding both endpoints) or half-open. So any combination of `<=` or `<` checks might be mathematically valid depending on what type of segment we're considering. I think this is all worth documenting.

packages/perseus/src/widgets/interactive-graphs/math/geometry.ts
```diff
@@ -2,11 +2,14 @@ import type {vec} from "mafs";
 
 export type Segment = [vec.Vector2, vec.Vector2];
 
-// https://stackoverflow.com/a/24392281/7347484
+// Determines whether two line segments have exactly one intersection point.
+// The segments are treated as "open" - that is, not inclusive of their
+// endpoints.
```
jeremywiebe, line 7: Thank-you for adding this!

PR #1400, opened 2024-07-10
Upgrade testing-library to React 18-compatible version

jeremywiebe:
## Summary:

This PR is the part of a chain of PRs upgrading Perseus to React 18.

  1. This PR --> #1400
  2. #1401
  3. #1402
  4. #1403

I thought I had the React 18 (#1064) upgrade all ready (and it was even approved), but when I returned to that PR, I noticed linter and test failures. This lead me to discover that I need to upgrade testing-library. 

This then breaks a bunch of tests (which I'll fix in future PRs). 

Issue: LEMS-1802

## Test plan:

This PR will introduce many test failures. That's ok, they'll be fixed in follow-up PRs so that hopefully each step is relatively easy to review.


somewhatabstract:
When upgrading webapp, we couldn't get the newer RTL to properly work with React 16 and had to update React first.

jeresig (approved):
Ah great!

PR #1402, opened 2024-07-10
React 18: Use RTL functions instead of direct HTML DOM APIs

jeremywiebe:
## Summary:

This PR is the part of a chain of PRs upgrading Perseus to React 18.

  1. #1400
  2. #1401
  3. This PR --> #1402
  4. #1403

This PR fixes a few instances where there exists an RTL API for tests instead of making direct DOM object calls. RTL recommends this approach and this fixes a few more tests.

Issue: LEMS-1802

## Test plan:

`yarn test` more tests pass

### Before

```sh
$ yarn test 
... 
Test Suites: 4 failed, 1 skipped, 184 passed, 188 of 189 total
Tests:       9 failed, 14 skipped, 2564 passed, 2587 total
Snapshots:   4 failed, 1 file obsolete, 155 passed, 159 total
Time:        24.056 s, estimated 25 s
Ran all test suites.
error Command failed with exit code 1.
```

### After 

```sh
$ yarn test 
...
Test Suites: 3 failed, 1 skipped, 185 passed, 188 of 189 total
Tests:       6 failed, 14 skipped, 2567 passed, 2587 total
Snapshots:   4 failed, 1 file obsolete, 155 passed, 159 total
Time:        20.517 s, estimated 24 s
Ran all test suites.
error Command failed with exit code 1.
```

Close! :)

PR #1401, opened 2024-07-10
React 18: Wrap non-RTL actions in act()

jeremywiebe:
## Summary:

This PR is the part of a chain of PRs upgrading Perseus to React 18.

  1. #1400
  2. This PR --> #1401
  3. #1402
  4. #1403

It appears that React 18 and RTL are more sensitive to not having mutating calls wrapped in `act()`. The changes in this PR are all simple changes that wrap function calls that cause DOM changes in `act()` calls. There are a few special cases that are failing yet, but this PR cleans up the straightforward ones. 

Issue: LEMS-1802

## Test plan:

`yarn test` should have many fewer test failures than PR #1400

Upcoming PRs will address the remaining test failures.

### Before 

```sh
$ yarn test 
...
Test Suites: 20 failed, 1 skipped, 168 passed, 188 of 189 total
Tests:       104 failed, 14 skipped, 2469 passed, 2587 total
Snapshots:   22 failed, 1 file obsolete, 137 passed, 159 total
Time:        25.267 s, estimated 26 s
Ran all test suites.
error Command failed with exit code 1.
```

### After

```sh
$ yarn test 
... 
Test Suites: 4 failed, 1 skipped, 184 passed, 188 of 189 total
Tests:       9 failed, 14 skipped, 2564 passed, 2587 total
Snapshots:   4 failed, 1 file obsolete, 155 passed, 159 total
Time:        24.056 s, estimated 25 s
Ran all test suites.
error Command failed with exit code 1.
```

PR #1403, opened 2024-07-11
React 18: Custom test fixes for React 18/RTL update

jeremywiebe:
## Summary:

This PR is the part of a chain of PRs upgrading Perseus to React 18.

  1. #1400
  2. #1401
  3. #1402
  4. This PR --> #1403

A few tests needed larger changes than simply wrapping the actions in `act()` (see #1402). This PR covers the remaining tests that were failing for various reasons. In some cases the tests were largely re-written (sometimes for better test clarity while re-working them to pass). 

In the case of the Zoomable tests, we were waiting for the text to be updated in the element, but the styling changes had not yet settled as the updates caused multiple `setState()` calls to be made which all needed to be called and then resolved by a subsequent render (thus the new `waitForStyle()` helper).

In the Graded Group test update, I'm not sure how it worked as written. In the test that was fixed, we were testing mobile and in that case the "Check" button reads "Try again" on the second and subsequent attempts. 

In the Group tests, I needed to wrap some actions in `act()` as usual, but I also took the opportunity to change the element finding logic away from the `getAll...()` functions to depend more on the aria labels. This makes these tests read more clearly (instead of accessing item at index `[0]` we get the textbox with the string `"value rounded to the nearest ten"`).

Lastly, there was one outdated snapshot due to a test file being renamed.

Issue: LEMS-1802

## Test plan:

`yarn test` all tests now pass!

### Before 

```sh
$ yarn test 
...
Test Suites: 3 failed, 1 skipped, 185 passed, 188 of 189 total
Tests:       6 failed, 14 skipped, 2567 passed, 2587 total
Snapshots:   4 failed, 1 file obsolete, 155 passed, 159 total
Time:        22.175 s, estimated 27 s
Ran all test suites.
```

### After 

```sh
$ yarn test 
...
Test Suites: 1 skipped, 188 passed, 188 of 189 total
Tests:       14 skipped, 2573 passed, 2587 total
Snapshots:   159 passed, 159 total
Time:        26.765 s
Ran all test suites.
```


packages/perseus/src/components/__tests__/zoomable.test.tsx
```diff
@@ -214,8 +214,10 @@ describe("Zoomable", () => {
         // Arrange
         // Simulate window resize event
         const resizeWindowTo = (width: number, height: number) => {
-            const resizeEvent = document.createEvent("Event");
-            resizeEvent.initEvent("resize", true, true);
```
jeremywiebe, line 218: ESLint was complaining that `initEvent` is deprecated. Given this is test code and the test still passes, I considered it a safe update.

packages/perseus/src/widgets/__tests__/__snapshots__/interactive-graph.test.ts.snap
jeremywiebe, line 1: The file under test was renamed at some point earlier to `.tsx` and so this snapshot is unused today.

packages/perseus-editor/src/__tests__/editor.test.tsx
```diff
@@ -123,7 +123,6 @@ describe("Editor", () => {
         await userEvent.clear(captionInput);
         await userEvent.type(captionInput, "A picture of kittens");
         await userEvent.tab(); // blurring the input triggers onChange to be called
-        jest.runOnlyPendingTimers();
```
jeremywiebe, line 126: Tests still pass without this, so I'm removing it.

jeresig (approved):
Glad you were able to find some good fixes here, seems tricky!

PR #1404, opened 2024-07-11
Support visible/aria label for the Expression widget

handeyeco:
## Summary:

Adds two new options for the Expression widget via the ExpressionEditor:

- visibleLabel
- ariaLabel

Caitlyn wanted these to be separate so that they could read differently. Her example is if the label should be "Number in cm": a screen reader might read that as "Number in C M" whereas we want it to read "Number in centimeters." So this supports that case.

Issue: LEMS-2081

## Test plan:

Looking at the labels:
- Go to a page that has an Expression with label ([like this page](https://prod-znd-240724-23739-eb7fb49.khanacademy.org/internal-courses/test-everything/test-everything-1/expression/a/expression-widget-for-level-access))
- Note there is a visible label
- Note that when you click the visible label, it focuses the MathQuill input
- Note that the MathQuill input (a textarea) has an aria label

Editing the labels:
- Go to an editing page that has an Expression widget ([like this page](https://prod-znd-240724-23739-eb7fb49.khanacademy.org/devadmin/content/articles/expression-widget-for-level-access/x24be113497bc5d30))
- Note in the dropdown for the Expression options there are two new fields for visible and aria labels
- Edit the labels, publish, wait for release, and note the labels are saved


packages/perseus/src/widgets/expression.tsx
```diff
@@ -537,32 +554,44 @@ export class Expression extends React.Component<Props, ExpressionState> {
...
-                                    this._handleFocus();
-                                }
-                            },
-                        );
-                    }}
-                    onBlur={this._handleBlur}
-                />
+                <View style={{padding: "15px 4px 0"}}>
+                    {!!this.props.visibleLabel && (
+                        <label
```
handeyeco, line 562: Mobile visible label

packages/perseus/src/widgets/expression.tsx
```diff
@@ -574,55 +603,73 @@ export class Expression extends React.Component<Props, ExpressionState> {
...
-                onFocus={() =>
-                    this.setState({
-                        showErrorTooltip: false,
-                    })
-                }
-            >
-                {/**
+            <View style={{margin: "5px 5px 0"}}>
+                {!!this.props.visibleLabel && (
+                    <label
```
handeyeco, line 608: Desktop visible label

anakaren-rojas:
noob question: when do the new labels get saved and pushed to the backend?

packages/perseus/src/widgets/expression.tsx
```diff
@@ -356,6 +362,17 @@ export class Expression extends React.Component<Props, ExpressionState> {
         // TODO(scottgrant): This is a hack to remove the deprecated call to
         // this.isMounted() but is still considered an anti-pattern.
         this._isMounted = true;
+
+        // HACK: imperatively add an ID onto the Mathquill input
+        // (which in mobile is a span; desktop a textarea)
+        // in order to associate a visual label with it
+        if (this.refs.input) {
+            const isMobile = this.props.apiOptions.customKeypad;
+            const container = ReactDOM.findDOMNode(this.refs.input);
```
catandthemachines, line 371: Nit: This code is fine for right now, but just want to highlight that if it is at all possible to avoid using `findDOMNode` function as it's deprecated in future versions of React. This isn't a problem now as we're just moving to version 18. But wanted to call it out as a food for thought 😄

packages/perseus/src/widgets/expression.tsx
```diff
@@ -537,32 +554,44 @@ export class Expression extends React.Component<Props, ExpressionState> {
         );
     };
 
-    render():
-        | React.ReactNode
-        | React.ReactElement<React.ComponentProps<"div">> {
```
catandthemachines, line 542: Ooo, I'm curious why did we need to remove the typing of the render function. If possible I would love to bring it back.

packages/perseus/src/widgets/expression.tsx
```diff
@@ -537,32 +554,44 @@ export class Expression extends React.Component<Props, ExpressionState> {
...
-                            () => {
-                                if (this._isMounted) {
-                                    this._handleFocus();
-                                }
-                            },
-                        );
-                    }}
-                    onBlur={this._handleBlur}
-                />
+                <View style={{padding: "15px 4px 0"}}>
```
catandthemachines, line 560: nit: not sure if this is an agreed upon style guidance, I would highly recommend placing style objects as const at the top of the file and injecting it here. I find it makes it easier to find all the style values instead of searching through the React code.

packages/perseus/src/widgets/expression.tsx
catandthemachines, line 562: Is there a WonderBlocks component they would recommend for creating labels? I could be wrong, but I think you can use any text component block as a label for an input, it just needs to have the htmlFor or the ariaLabelledby properties linking them together.

packages/perseus/src/widgets/expression.tsx
catandthemachines, line 608: Same here about using a WonderBlocks component for the label.

packages/perseus/src/widgets/expression.tsx
```diff
@@ -574,55 +603,73 @@ export class Expression extends React.Component<Props, ExpressionState> {
...
+                            className={ApiClassNames.INTERACTIVE}
+                            value={this.props.value}
+                            onChange={this.changeAndTrack}
+                            convertDotToTimes={this.props.times}
+                            buttonSets={this.props.buttonSets}
+                            onFocus={this._handleFocus}
+                            onBlur={this._handleBlur}
+                            hasError={this.state.showErrorStyle}
+                            ariaLabel={
+                                this.props.ariaLabel || "Expression input"
```
catandthemachines, line 659: Question for curiosity, can we globalize this string and inject it in here?

catandthemachines (commented):
It's getting so close! The biggest questions I have is whether we can use a WonderBlocks components for the labels so the styling is consistent with our design system.

packages/math-input/src/components/input/math-wrapper.ts
```diff
@@ -38,20 +38,28 @@ class MathWrapper {
     mobileKeyTranslator: Record<Key, MathFieldUpdaterCallback>;
 
     constructor(
-        element,
```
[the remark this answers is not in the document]
handeyeco: I renamed `element` to be `mathFieldMount` so it's clearer what it's for.

packages/perseus-editor/src/widgets/expression-editor.tsx
```diff
@@ -175,6 +175,37 @@ class ExpressionEditor extends React.Component<Props> {
...
+                <div className="perseus-widget-row">
+                    <label>
+                        Aria label:{" "}
+                        <TextInput
+                            value={this.props.ariaLabel}
+                            onChange={this.change("ariaLabel")}
+                        />
+                    </label>
+                    <InfoTip>
+                        <p>Label text that's read by screenreaders.</p>
```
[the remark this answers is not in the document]
handeyeco: I'd like to avoid that until we know it's stable and socialize expectations around aria-labels to content creators.

packages/perseus/src/widgets/expression.tsx
handeyeco: Screenshot 2024-07-26 at 8 41 25 AM

`render` gets typed as a result of Expression being typed as `React.Component` so explicit typing is redundant: `class Expression extends React.Component<Props, ExpressionState>`.

But I doubt I would have removed that without having hit an error, I just don't remember what it is.

packages/perseus-editor/src/widgets/expression-editor.tsx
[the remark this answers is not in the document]
handeyeco: Ticket to follow up: https://khanacademy.atlassian.net/browse/LEMS-2212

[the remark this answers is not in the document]
catandthemachines: ++ I agree with that. Can we in the interum update the InfoTip to say:

> Label text that's read by screen readers. Highly recommend adding a label here to ensure your exercise is accessible. For more information on writting accessible labels, please see [this article](https://www.w3.org/WAI/tips/designing/#ensure-that-form-elements-include-clearly-associated-labels) or reach out to the [accessibility guild](https://khanacademy.slack.com/archives/C0A4SN2NQ)

packages/perseus-editor/src/__stories__/article-editor.stories.tsx
catandthemachines, line 1: Ooo interesting, I'm guessing this was missing in the storybook, or is this related to expressions?

packages/perseus/src/widgets/expression.tsx
```diff
@@ -574,59 +602,81 @@ export class Expression extends React.Component<Props, ExpressionState> {
...
+                            }
+                        />
+                    </Tooltip>
+                </div>
+            </View>
         );
     }
 }
 
+const styles = StyleSheet.create({
```
catandthemachines, line 671: Did Caitlyn get a chance to take a look at these padding values for mobile and desktop?

catandthemachines (approved):

packages/perseus-editor/src/__stories__/article-editor.stories.tsx
handeyeco: It was missing. I added it when debugging why Articles weren't saving changes.

packages/perseus-editor/src/widgets/expression-editor.tsx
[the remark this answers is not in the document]
handeyeco: I updated the text and added the WCAG link. I didn't add the Slack link since this is a public project and only internal KA folks can access that channel.

PR #1405, opened 2024-07-12
Fix console errors and warnings printed in tests

benchristel:
Our tests printed a lot of noisy warnings to the logs, which made it harder to
spot real test failures. This PR fixes up the code so the warnings are no
longer logged.

Each commit in the PR fixes a different warning. It's probably going to be
easiest to review it one commit at a time.

Issue: LEMS-2141

## Test plan:

Run `yarn test`, and verify that no warnings are printed.

### DIFF
--- a/packages/perseus-editor/src/widgets/numeric-input-editor.tsx
+++ b/packages/perseus-editor/src/widgets/numeric-input-editor.tsx
@@ -292,9 +292,14 @@ class NumericInputEditor extends React.Component<Props, State> {
         const addAnswerButton = (
             <div>
                 <a
-                    href="javascript:void(0)"
+                    href="#"
                     className="simple-button orange"
-                    onClick={() => this.addAnswer()}
+                    onClick={(e) => {
+                        // preventDefault ensures that href="#"
+                        // doesn't scroll to the top of the page
+                        e.preventDefault();
+                        this.addAnswer();
+                    }}
                     onKeyDown={(e) => this.onSpace(e, this.addAnswer)}
                 >
                     <span>Add new answer</span>
@@ -425,9 +430,14 @@ class NumericInputEditor extends React.Component<Props, State> {
                             ) : null}
                             <div className="value-divider" />
                             <a
-                                href="javascript:void(0)"
+                                href="#"
                                 className={"answer-status " + answer.status}
-                                onClick={() => this.onStatusChange(i)}
+                                onClick={(e) => {
+                                    // preventDefault ensures that href="#"
+                                    // doesn't scroll to the top of the page
+                                    e.preventDefault();
+                                    this.onStatusChange(i);
+                                }}
                                 onKeyDown={(e) =>
                                     this.onSpace(e, this.onStatusChange)
                                 }
@@ -435,10 +445,15 @@ class NumericInputEditor extends React.Component<Props, State> {
                                 {answer.status}
                             </a>
                             <a
-                                href="javascript:void(0)"
+                                href="#"
                                 className="answer-trash"
                                 aria-label="Delete answer"
-                                onClick={() => this.onTrashAnswer(i)}
+                                onClick={(e) => {
+                                    // preventDefault ensures that href="#"
+                                    // doesn't scroll to the top of the page
+                                    e.preventDefault();
+                                    this.onTrashAnswer(i);
+                                }}
                                 onKeyDown={(e) =>
                                     this.onSpace(e, this.onTrashAnswer)
                                 }
@@ -446,10 +461,15 @@ class NumericInputEditor extends React.Component<Props, State> {
                                 <InlineIcon {...iconTrash} />
                             </a>
                             <a
-                                href="javascript:void(0)"
+                                href="#"
                                 className="options-toggle"
                                 aria-label="Toggle options"
-                                onClick={() => this.onToggleOptions(i)}
+                                onClick={(e) => {
+                                    // preventDefault ensures that href="#"
+                                    // doesn't scroll to the top of the page
+                                    e.preventDefault();
+                                    this.onToggleOptions(i);
+                                }}
                                 onKeyDown={(e) =>
                                     this.onSpace(e, this.onToggleOptions)
                                 }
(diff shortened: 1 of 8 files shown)


packages/perseus-editor/src/widgets/numeric-input-editor.tsx
```diff
@@ -425,31 +430,46 @@ class NumericInputEditor extends React.Component<Props, State> {
...
+                                href="#"
                                 className="options-toggle"
                                 aria-label="Toggle options"
-                                onClick={() => this.onToggleOptions(i)}
+                                onClick={(e) => {
+                                    // preventDefault ensures that href="#"
+                                    // doesn't scroll to the top of the page
+                                    e.preventDefault();
+                                    this.onToggleOptions(i);
+                                }}
```
benchristel, line 472: We should really be using a button element here, but I decided to just make the minimal change to get rid of the warning.

packages/perseus/src/widgets/__tests__/radio.test.ts
```diff
@@ -429,9 +426,6 @@ describe("single-choice question", () => {
 
             // Act
             await userEvent.keyboard(" ");
-            await act(async () => {
-                await jest.runAllTimers();
-            });
```
benchristel, line 434: @nicolecomputer I don't _think_ these `act` calls are necessary, because `userEvent` will advance the Jest timers for us. Do you agree? (asking because you added the act calls, but no worries if you have no memory of this)

yarn.lock
```diff
@@ -6034,25 +6034,10 @@ caniuse-api@^3.0.0:
...
-  integrity sha512-cPiMKZgqgkg5LY3/ntGeLFUpi6tzddBNS58A4tnTgQw1zON7u2sZMU7SzOeVH4tj20++9ggL+V6FDOFMTaFFYA==
-
-caniuse-lite@^1.0.30001587:
-  version "1.0.30001587"
-  resolved "https://registry.yarnpkg.com/caniuse-lite/-/caniuse-lite-1.0.30001587.tgz#a0bce920155fa56a1885a69c74e1163fc34b4881"
-  integrity sha512-HMFNotUmLXn71BQxg8cijvqxnIAofforZOwGsxyXJ0qugTdspUF4sPSJ2vhgprHCB996tIDzEq1ubumPDV8ULA==
+caniuse-lite@^1.0.0, caniuse-lite@^1.0.30001332, caniuse-lite@^1.0.30001541, caniuse-lite@^1.0.30001587:
+  version "1.0.30001641"
+  resolved "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001641.tgz"
+  integrity sha512-Phv5thgl67bHYo1TtMY/MurjkHhV4EDaCosezRXgZ8jzA/Ub+wjxAvbGvjoFENStinwi5kCyOYV3mi5tOGykwA==
```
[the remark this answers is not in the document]
jeremywiebe: Yes! It's fine to update manually.  I've created a task to implement auto-updating so that these warnings don't return in a few days/weeks.

https://khanacademy.atlassian.net/browse/LEMS-2177

packages/perseus/src/widgets/__tests__/radio.test.ts
```diff
@@ -828,13 +828,15 @@ describe("multi-choice question", () => {
...
+        async (...choices) => {
             // Arrange
             const {renderer} = renderQuestion(question, apiOptions);
 
             // Act
             const option = screen.getAllByRole("checkbox");
-            choices.forEach(async (i) => await userEvent.click(option[i]));
+            for (const i of choices) {
+                await userEvent.click(option[i])
+            }
```
jeremywiebe, line 839: More and more I'm becoming a fan of using this type of for loop over `.forEach()`. To me, `.forEach()` isn't a higher-order function like other, more common ones (`.map()`, `.filter()`, etc).

jeremywiebe (approved):
Thanks for taking the time to investigate and fix these!

**idea**: Webapp has a Jest setup that fails a test if there are console.error or console.warn calls during the test. We used to have that but removed it... I wonder if we want to bring that back to help keep our test output clean. 🤔

packages/perseus/src/widgets/__tests__/radio.test.ts
benchristel: Yeah, the `forEach` callback needs to have a side effect in order to do anything useful. So there's not much advantage over a for loop. The one case where I might use it is at the end of a functional pipeline, like:

```js
things.map(...)
  .filter(...)
  .map(...)
  .forEach(console.log)
```

benchristel:
> idea: Webapp has a Jest setup that fails a test if there are console.error or console.warn calls during the test. We used to have that but removed it... I wonder if we want to bring that back to help keep our test output clean. 

@jeremywiebe I like that idea, but there is one obstacle. There is actually still one error being printed:

```
    console.error
      Warning: React does not recognize the `uniqueId` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `uniqueid` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
          in div (created by Popper)
```

We can't easily fix this, because it's coming from a library (`react-popper`). I believe we're on the latest version of `react-popper`, so I think we'd have to fork it and fix the issue ourselves.

jeremywiebe:
> > idea: Webapp has a Jest setup that fails a test if there are console.error or console.warn calls during the test. We used to have that but removed it... I wonder if we want to bring that back to help keep our test output clean.
> 
> @jeremywiebe I like that idea, but there is one obstacle. There is actually still one error being printed:
> 
> ```
>     console.error
>       Warning: React does not recognize the `uniqueId` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `uniqueid` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
>           in div (created by Popper)
> ```
> 
> We can't easily fix this, because it's coming from a library (`react-popper`). I believe we're on the latest version of `react-popper`, so I think we'd have to fork it and fix the issue ourselves.

Ah! I saw that yesterday too! I thought I traced it to the WB popover and not react-popper specifically. Does [this](https://github.com/Khan/wonder-blocks/blob/main/packages/wonder-blocks-popover/src/components/popover.tsx#L253) look like the culprit?  If so, might be worth raising with the WB team.

PR #1406, opened 2024-07-12
remove profanity rule

handeyeco:
## Summary:
See: https://khanacademy.slack.com/archives/C01AZ9H8TTQ/p1720805367737349

Why:
- My understanding is that these linters are rules for content creators
- The linter only checked for 7 English words and only looking for the whole word
- While not a big concern, it _is_ code that has to run and be maintained
- Also it's code in our public codebase, so it feels weird to have

So...
- We're running it against people we trust
- It's not very effective at what it's trying to do
- And there are other reasons to remove it

So I just thought I'd remove it.

### DIFF
--- a/.changeset/sour-carrots-flash.md
+++ b/.changeset/sour-carrots-flash.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-linter": major
+---
+
+No longer lint for profanity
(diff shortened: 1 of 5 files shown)


catandthemachines (approved):
Easy, simple and not needed anymore. Sounds like a great change/update to Perseus <3

PR #1409, opened 2024-07-12
[Interactive Graph: Test Utils] Update the interactive graph question builder with all options + update testdata to use builder

nishasy:
## Summary:
I wanted to update all our interactive graph testdata so that it's easy to
read through, understand what it's testing, and reuse in the future.

Changes included:
- Updated the builder so that all the graph types include
  `coord` and `startCoord` in the `options` param, as well
  as any other field that tests should be able to specify
  for that respective graph type.
- Updated the builder so that it uses the correct default
  `correct` coords, allowing it to be used for the
  `-DefaultCorrect` tests without passing in options.
- Updated the test data so that it uses the builder
- Updated the test data so that the question (`content` field)
  actually makes sense for the graph. (for example, questions
  that asked the user to change the triangle, but the graph is
  actually a square-- or maybe not even a polygon)
- Updated unit tests to reflect the above updates

Issue: https://khanacademy.atlassian.net/browse/LEMS-2137

## Test plan:
`yarn jest packages/perseus/src/widgets/interactive-graphs/interactive-graph-question-builder.test.ts`

`yarn run cypress` --> run every perseus related e2e test

`yarn dev`
- Check that the questions' wording makes sense for their respective graphs

## yarn dev screenshots

Changes to notice:
- I used locked figures for some instead of background images, so they don't have the A or B labels anymore
- I removed the question description for some where it didn't feel relevant, but I can add them back if they were helpful
- The linear one no longer has any figures in it

| Before | After |
| --- | --- |
| Screenshot 2024-07-12 at 4 00 55 PM | Screenshot 2024-07-12 at 4 01 25 PM |
| Screenshot 2024-07-12 at 4 01 01 PM | Screenshot 2024-07-12 at 4 01 29 PM |
| Screenshot 2024-07-12 at 4 01 05 PM | Screenshot 2024-07-12 at 4 01 33 PM |
| Screenshot 2024-07-12 at 4 01 09 PM | Screenshot 2024-07-12 at 4 01 37 PM |


packages/perseus/src/widgets/__stories__/interactive-graph.stories.tsx
```diff
@@ -6,7 +6,6 @@ import {
     angleQuestion,
     circleQuestion,
     linearQuestion,
-    linearQuestionWithLockedPoints,
```
nicolecomputer, line 9: This didn't do anything and wasn't a test we needed. It was removed.

packages/perseus/src/widgets/__tests__/__snapshots__/interactive-graph.test.ts.snap
```diff
@@ -352,26 +338,26 @@ exports[`interactive-graph widget question Shoud render predictably: after inter
                 </span>
                 <span
                   class="graphie-label"
-                  data-math-formula="5.0"
-                  style="position: absolute; padding: 7px; color: rgb(100, 149, 237); left: 269.14285714285717px; top: 126.85714285714285px; fill: none; stroke: #6495ED;"
```
nicolecomputer, line 356: This probably changed from moving from unlimited points to a limited number of points

packages/perseus/src/widgets/__tests__/__snapshots__/interactive-graph.test.ts.snap
```diff
@@ -84,7 +84,7 @@ exports[`interactive-graph widget question Shoud render predictably: after inter
                     style="-webkit-tap-highlight-color: rgba(0, 0, 0, 0);"
                   />
                   <path
-                    d="M371.42857142857144,228.57142857142858L142.85714285714286,57.142857142857146L142.85714285714286,228.57142857142858Z"
```
nicolecomputer, line 87: Not a ship blocker, just curious: I wonder why this one is different?

packages/perseus/src/widgets/__tests__/interactive-graph.test.ts
```diff
@@ -422,12 +422,12 @@ describe("locked layer", () => {
 
         // Assert
         expect(points[0]).toHaveStyle({
-            fill: lockedFigureColors.green,
-            stroke: lockedFigureColors.green,
+            fill: lockedFigureColors.grayH,
```
nicolecomputer, line 425: Gray is the actual default for locked figures, that's why these changed

packages/perseus/src/widgets/interactive-graphs/interactive-graph-question-builder.ts
```diff
@@ -111,85 +149,149 @@ class InteractiveGraphQuestionBuilder {
...
+    withSegments(options?: {
+        numSegments?: number;
+        startCoords?: CollinearTuple[];
+        coords?: CollinearTuple[];
+    }): InteractiveGraphQuestionBuilder {
+        this.interactiveFigureConfig = new SegmentGraphConfig(options);
         return this;
     }
 
-    withNumSegments(numSegments: number): InteractiveGraphQuestionBuilder {
```
nicolecomputer, line 124: This got removed to increase clarity

packages/perseus/src/widgets/interactive-graphs/interactive-graph-question-builder.ts
```diff
@@ -111,85 +149,149 @@ class InteractiveGraphQuestionBuilder {
...
+        allowReflexAngles?: boolean;
+        angleOffsetDeg?: number;
+        snapDegrees?: number;
+        match?: "congruent";
+    }): InteractiveGraphQuestionBuilder {
+        this.interactiveFigureConfig = new AngleGraphConfig(options);
         return this;
     }
 
     // TODO(benchristel): if we want other attributes of locked points to be
```
nicolecomputer, line 249: I think we solved this TODO, can you mark it done

packages/perseus/src/widgets/interactive-graphs/interactive-graph-question-builder.ts
```diff
@@ -272,8 +374,20 @@ class InteractiveGraphQuestionBuilder {
...
+        y: number,
+        options?: {
+            color?: LockedFigureColor;
+            filled?: boolean;
+        },
+    ): LockedPointType {
+        return {
+            type: "point",
+            coord: [x, y],
+            color: options?.color ?? "grayH",
```
nicolecomputer, line 388: Nice! I see you going from green to gray!

packages/perseus/src/widgets/interactive-graphs/interactive-graph-question-builder.ts
```diff
@@ -288,21 +402,33 @@ interface InteractiveFigureConfig {
...
 
-    constructor(numSegments: number, startCoords?: CollinearTuple[]) {
-        this.numSegments = numSegments;
-        this.startCoords = startCoords;
+    constructor(options?: {
+        numSegments?: number;
+        startCoords?: CollinearTuple[];
+        coords?: CollinearTuple[];
+    }) {
+        this.numSegments =
```
nicolecomputer, line 413: Not a ship blocker: Should this throw an error if these are not the same length?

packages/perseus/src/widgets/interactive-graphs/interactive-graph-question-builder.ts
```diff
@@ -315,20 +441,25 @@ class SegmentGraphConfig implements InteractiveFigureConfig {
...
 
-    constructor(startCoords?: CollinearTuple) {
-        this.startCoords = startCoords;
+    constructor(options?: {
+        coords?: CollinearTuple;
+        startCoords?: CollinearTuple;
+    }) {
+        this.startCoords = options?.startCoords;
+        this.correctCoords = options?.coords ?? [
+            [-5, 5],
```
nicolecomputer, line 454: Note: These are the default correct coords (before we were using an arbitrary set of coords)

packages/perseus/src/widgets/interactive-graphs/interactive-graph-question-builder.ts
```diff
@@ -454,29 +617,147 @@ class SinusoidGraphConfig implements InteractiveFigureConfig {
...
+    constructor(options?: {
+        coords?: [Coord, Coord, Coord];
+        startCoords?: [Coord, Coord, Coord];
+        showAngles?: boolean;
+        allowReflexAngles?: boolean;
+        angleOffsetDeg?: number;
+        snapDegrees?: number;
+        match?: "congruent";
+    }) {
+        // Default correct answer is 20 degree angle at (0, 0)
```
nicolecomputer, line 726: Thank you for adding this note, this helps make it understandable!

packages/perseus/src/widgets/__testdata__/interactive-graph.testdata.ts
```diff
@@ -6,1715 +6,451 @@ import type {LockedFunctionOptions} from "../interactive-graphs/interactive-grap
...
-            },
-            alignment: "default",
-        },
-    },
-};
+export const polygonQuestionDefaultCorrect: PerseusRenderer =
+    interactiveGraphQuestionBuilder()
+        .withPolygon("grid", {
+            startCoords: [
+                // TODO: figure out why this test is off by 0.5.
```
nicolecomputer, line 231: Please make a ticket for this to figure out in the future. This isn't a ship-blocker

nicolecomputer (approved):
A couple of minor nits and things to clean up but this looks great! Thank you for taking this on!

packages/perseus/src/widgets/__tests__/__snapshots__/interactive-graph.test.ts.snap
nishasy: I figured out the difference by pasting these into an SVG path editor tool. Before, it was showing a right triangle (the answer), and now it's showing a default starting triangle.

This particular snapshot was looking at `polygonQuestion`, referenced in `questionsAndAnswers` in the testdata file.

This new triangle is actually the expected triangle. I'm not sure why it was showing a different triangle before, as the `graph` field in the testdata wasn't specifying that it should be the same as the `correct` coords. 🤔 The fact that it was wrong is weird, but at least it's fixed now?

| Before | After |
| --- | --- |
| image | image |

packages/perseus/src/widgets/__testdata__/interactive-graph.testdata.ts
nishasy: https://khanacademy.atlassian.net/browse/LEMS-2186

packages/perseus/src/widgets/interactive-graphs/interactive-graph-question-builder.ts
nishasy: I keep going back and forth on this. I almost want to say no, because if we were to add it here, we would also have to add errors for any difference in length between coords and startCoords for all the other graph types that have a coord array.

I kind of want to just trust the devs, but maybe @benchristel has stronger opinions on this.

benchristel (commented):
Thanks for doing this. It's SO MUCH BETTER. 💖 🎉 

How do we know this refactoring doesn't change the test data? Or are we pretty sure that it doesn't matter if there are changes to the data, as long as no tests fail?

I've had bad experiences in the past where changes to test inputs caused tests to become tautological or test the wrong thing, so I'm looking for some assurance that we're protected from that.

packages/perseus/src/widgets/interactive-graphs/interactive-graph-question-builder.ts
benchristel: I'm inclined not to add any validations until we mess up a test and we're like "gosh I wish the builder had validations". The risk incurred by doing nothing seems low.

packages/perseus/src/widgets/interactive-graphs/interactive-graph-question-builder.ts
```diff
@@ -111,85 +149,147 @@ class InteractiveGraphQuestionBuilder {
...
 
-    withSegments(
-        startCoords: CollinearTuple[],
-    ): InteractiveGraphQuestionBuilder {
-        this.interactiveFigureConfig = new SegmentGraphConfig(
-            startCoords.length,
-            startCoords,
-        );
+    withSegments(options?: {
+        numSegments?: number;
```
benchristel, line 153: It might be worth having a comment here to say that if `numSegments` is not passed, it gets defaulted appropriately.

Alternatively, I might be inclined to split `withSegments` into multiple methods for different use cases:

```
withDefaultSegments(numSegments)

withSegments({startCoords, coords})
```

packages/perseus/src/widgets/interactive-graphs/interactive-graph-question-builder.ts
```diff
@@ -111,85 +149,147 @@ class InteractiveGraphQuestionBuilder {
...
+        },
     ): InteractiveGraphQuestionBuilder {
-        this.interactiveFigureConfig = new SinusoidGraphConfig(startCoords);
+        this.interactiveFigureConfig = new PolygonGraphConfig(snapTo, options);
         return this;
     }
 
-    withPolygon(
-        snapTo: "grid" | "angles" | "sides",
+    withPoint(
```
benchristel, line 225: Should this method be named `withPoints`? Since the first argument is `numPoints`, and `withPoints(3)` makes more sense than `withPoint(3)`.

packages/perseus/src/widgets/interactive-graphs/interactive-graph-question-builder.ts
nishasy: (talked about offline) I think for now, I'd prefer to leave it as one method so it's consistent with the other graph type methods. Even for the `withSegments` example, there's a potential mismatch between the length of `startCoords` and the length of `coords`. Ideally, we'd be able to trust the devs. What I really think would happen is that if they needed to test the length, it would fail.

If it becomes an issue, we can change it later. For now, I found this way to be the most intuitive to me while I was trying to use this method in the test data.

nishasy:
> How do we know this refactoring doesn't change the test data? Or are we pretty sure that it doesn't matter if there are changes to the data, as long as no tests fail?

(talked about offline) I did do a full repo search for the uses of every single question in the testdata file. I made sure that the part that was being tested for each question stayed in tact. There were definitely some changes, but to my knowledge, they were not important in the first place. Hopefully that's good enough?

PR #1410, opened 2024-07-12
Clarify which actions go with which graph type

benchristel:
I could never remember which actions are supposed to be used with which
graph type, so I thought I'd fix that. Now you don't have to guess: you
can type `actions.<graph type>` and your IDE will autocomplete the
available actions for you.

Issue: none

## Test plan:

`yarn test`


mark-fitzgerald (approved):
I love this! Thanks for improving the DX of the Interactive Graph!!!

PR #1411, opened 2024-07-15
add PerseusError to perseus-core

handeyeco:
## Summary:
When working on adding READMEs, I noticed `perseus-error` wasn't doing much and that there was a duplicate `perseus-error.ts` file in `perseus`.

`perseus-error` was also just one tiny file and we have `perseus-core` now.

This PR moves `perseus-error` into `perseus-core`, removes the `perseus-error` package, and remove `perseus-error.ts`.

Issue: LEMS-2178

## Test plan:
- Everything should work the same


jeremywiebe (approved):
👍  Thanks for cleaning this up!

PR #1413, opened 2024-07-15
Update browserslist and caniuse-lite packages weekly

jeremywiebe:
## Summary:

This PR introduces a new Github Action that will run weekly (on Monday's at 7:30 am UTC - which maps to somewhere in the night in North America).

We will still need to manually approve and land these PRs to merge the changes (#security). Also, if we do not land the PR before this action runs again, the PR will simply be updated.

Issue: "none"

## Test plan:

We can run this action manually to test. Upon running it, a new PR with the title 'chore: update browserslist' should appear.

### DIFF
--- a/.github/workflows/maintenance.yml
+++ b/.github/workflows/maintenance.yml
@@ -42,19 +42,19 @@ jobs:
                   commit-message: "chore: update browserslist"
                   title: Update browserslist
                   body: |
-                     # Summary
+                      # Summary
 
-                     Updates the `browserslist` and `caniuse-lite` npm packages
+                      Updates the `browserslist` and `caniuse-lite` npm packages
 
-                     ## Reviewing notes:
+                      ## Reviewing notes:
 
-                     There should only be changes to the `yarn.lock` file in
-                     this PR. Check that there is only 1 `caniuse-lite` package
-                     reference in the `yarn.lock` file (the intent of this
-                     update is to ensure that `caniuse-lite` is on the latest
-                     version and that there aren't multiple, conflicting
-                     versions that different tools might see).
+                      There should only be changes to the `yarn.lock` file in
+                      this PR. Check that there is only 1 `caniuse-lite` package
+                      reference in the `yarn.lock` file (the intent of this
+                      update is to ensure that `caniuse-lite` is on the latest
+                      version and that there aren't multiple, conflicting
+                      versions that different tools might see).
 
-                     If everything looks fine, please approve this PR and then
-                     land it (either with the Big Green Merge Button ™️ or by
-                     using `git land <this pr #>` in a terminal.
+                      If everything looks fine, please approve this PR and then
+                      land it (either with the Big Green Merge Button ™️ or by
+                      using `git land <this pr #>` in a terminal.
--- a/.github/workflows/maintenance.yml
+++ b/.github/workflows/maintenance.yml
@@ -5,7 +5,7 @@ on:
     workflow_dispatch:
     schedule:
         # https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#schedule
-        #        ┌────────── minute (0 - 59)
+        #        ┌─────────── minute (0 - 59)
         #        │ ┌───────── hour (0 - 23)
         #        │ │ ┌─────── day of the month (1 - 31)
         #        │ │ │ ┌───── month (1 - 12 or JAN-DEC)
--- a/.github/workflows/maintenance.yml
+++ b/.github/workflows/maintenance.yml
@@ -1,0 +1,32 @@
+name: Weekly maintenance/update tasks
+on:
+    workflow_dispatch:
+    schedule:
+        - cron: "30 1 * * 1" # Run on Mondays @ 7:30am UTC (3:30am EDT, 12:30 am PDT)
+jobs:
+    update-browserslist:
+        name: Update the browserslist database (used by caniuse-lite)
+        runs-on: ${{ matrix.os }}
+        strategy:
+            matrix:
+                os: [ubuntu-latest]
+                node-version: [20.x]
+        steps:
+            - uses: actions/checkout@v4
+            - name: Install & cache node_modules
+              uses: ./.github/actions/shared-node-cache
+              with:
+                  node-version: ${{ matrix.node-version }}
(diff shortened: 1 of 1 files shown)


.github/workflows/maintenance.yml
```diff
@@ -0,0 +1,32 @@
+name: Weekly maintenance/update tasks
+on:
+    workflow_dispatch:
+    schedule:
+        - cron: "30 1 * * 1" # Run on Mondays @ 7:30am UTC (3:30am EDT, 12:30 am PDT)
```
jeremywiebe, line 5: This field is compatible with POSIX `cron` syntax. More details here: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule

somewhatabstract: **suggestion:** For easier workflow maintenance, I usually paste this useful info into the workflow as a comment...like this:
https://github.com/Khan/webapp/blob/906784c2673fa8fcd7e1a6aced02601692d646d7/.github/workflows/static-service-nightly-e2e.yml#L7-L15

.github/workflows/maintenance.yml
```diff
@@ -0,0 +1,60 @@
...
+    workflow_dispatch:
+    schedule:
+        # https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#schedule
+        #       ┌─────────── minute (0 - 59)
+        #       │  ┌───────── hour (0 - 23)
+        #       │  │ ┌─────── day of the month (1 - 31)
+        #       │  │ │ ┌───── month (1 - 12 or JAN-DEC)
+        #       │  │ │ │ ┌─── day of the week (0 - 6 or SUN-SAT)
+        #       v  v v v v
+        - cron: 30 1 * * 1
```
benchristel, line 14: Thank you for documenting this! :)

How did you choose 7:30 AM? (i.e. why not 7 AM?)

benchristel (approved):
Very cool. LGTM!

.github/workflows/maintenance.yml
jeremywiebe: Props to @somewhatabstract for the prompt and source of the docs. I really like the format!

I probably overthought this by a bunch.  I picked a time that would be at night in North America (considering East Coast and West Coast time zones).  Then I picked a time not at the top of the hour because tasks are often scheduled at the top of the hour and this would avoid adding to a potential thundering herd. In the end, it's probably much overthought! :)

jeremywiebe: And oh no!  I think I've got it at 1:30 UTC right now (contradicting the doc string)!  I'll fix.

.github/workflows/maintenance.yml
```diff
@@ -0,0 +1,60 @@
...
+    schedule:
+        # https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#schedule
+        #        ┌─────────── minute (0 - 59)
+        #        │ ┌───────── hour (0 - 23)
+        #        │ │ ┌─────── day of the month (1 - 31)
+        #        │ │ │ ┌───── month (1 - 12 or JAN-DEC)
+        #        │ │ │ │ ┌─── day of the week (0 - 6 or SUN-SAT)
+        #        v v v v v
+        - cron: 30 7 * * 1
+        #       Run on Mondays @ 7:30am UTC (3:30am EDT, 12:30 am PDT)
```
somewhatabstract, line 15: **praise:** Thanks for adding the comments here. I am wondering if it's useful to also note that all times are in UTC. It's tripped me up before and while the docs mention it, it might be useful to add in these comments too. Not super important (something to carry over to other places too), but thought I'd mention the idea.

.github/workflows/maintenance.yml
```diff
@@ -0,0 +1,60 @@
...
+                  ssh-private-key: ${{ secrets.KHAN_ACTIONS_BOT_SSH_PRIVATE_KEY }}
+
+            - name: Update browserslist DB
+              run: npx update-browserslist@latest
+
+            - name: Create Pull Request
+              uses: peter-evans/create-pull-request@v6
+              with:
+                  token: ${{ secrets.GITHUB_TOKEN }}
+                  team-reviewers: Perseus
```
somewhatabstract, line 41: **question:** Should these sorts of PRs cast a slightly wider net of reviewers? It feels  like Frontend Infra Web would be a good thing to add so there's more people that can review and land the PR. Maybe Security too?

.github/workflows/maintenance.yml
```diff
@@ -0,0 +1,60 @@
...
+              uses: ./.github/actions/shared-node-cache
+              with:
+                  node-version: ${{ matrix.node-version }}
+                  ssh-private-key: ${{ secrets.KHAN_ACTIONS_BOT_SSH_PRIVATE_KEY }}
+
+            - name: Update browserslist DB
+              run: npx update-browserslist@latest
+
+            - name: Create Pull Request
+              uses: peter-evans/create-pull-request@v6
```
somewhatabstract, line 38: suggestion: Would using `gh` be better than relying on a third-party action?

Also, I think enabling auto-merge for this PR would be a good idea. That way, when all actions are run AND the PR is reviewed, it will just land itself. What do you think?

somewhatabstract (approved):
This looks good to me. I have some suggestions inline.

Do we need to do anything with Changesets to make sure the changesets action is happy with this generated PR?

.github/workflows/maintenance.yml
jeremywiebe: I've noted that the time is in UTC below, but I'll make it explicit.

.github/workflows/maintenance.yml
jeremywiebe: I'm hesitant to overload #Security with PRs that are routine like this. If FEI is happy to see these, I'll add them. I think reviewers can bring in the Security folks if something feels off when reviewing.

.github/workflows/maintenance.yml
jeremywiebe: I'm going to stick with this action because it has some really useful behaviour that would take some amount of new code to reproduce: https://github.com/peter-evans/create-pull-request?tab=readme-ov-file#action-behaviour

Also, auto-merge is not enabled on the Perseus repo and I'm a bit hesitant to open that can of worms. It's unclear how that would affect our OLC tooling and other workflows/tools we have in place.

PR #1417, opened 2024-07-15
Remove jest-extended package

jeremywiebe:
## Summary:

When we moved Perseus out of webapp and back into this repo, we kept a dependency on `jest-extended`. Today, Perseus' use of this package is minimal and there are obvious Jest-native matchers that we can use instead. 

So, today we drop the `jest-extended` dependency.

Issue: "none"

## Test plan:

`yarn test` ✅

### DIFF
--- a/config/test/test.config.js
+++ b/config/test/test.config.js
@@ -61,7 +61,6 @@ module.exports = {
         "<rootDir>/**/*.test.tsx",
     ],
     setupFilesAfterEnv: [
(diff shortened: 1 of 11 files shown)


mark-fitzgerald (approved):
Removing dependencies (even for dev/testing) tickles my toes. Thank you!

catandthemachines (approved):
Looks great! Thanks for cleaning this up!

PR #1418, opened 2024-07-16
React 18: Final Fixes

jeremywiebe:
## Summary:

This PR represents what should be the final set of fixes for React 18. It includes the following changes that quiet any remaining console errors as well as test failures:

  * caching React "roots" created by `ReactDOMClient.createRoot()` in `tex.ts`. React warns when you call `ReactDOMClient.createRoot()` multiple times on the same DOM element. So this PR caches elements that have been processed in a `WeakMap`. The `WeakMap` will avoid a "memory leak" by "over retention" of DOM nodes.
  * Fixes a "uniqueId" warning coming out of Wonder Blocks Popover. It turns out that even though the WB `Popover` constrains it's `content` property to only be `PopoverContents` or `PopoverContentsCore`, Typescript happily allows any JSX to be assigned to it. So I've fixes `packages/math-input/src/components/keypad/__tests__/keypad-v2-mathquill.test.tsx` to wrap it's `contents` in a `PopoverContentsCore`. See [this](https://khanacademy.slack.com/archives/C4PE1QM5Y/p1721077735704319) Slack thread.
  * Introduction of `waitForDeferredRenders` which uses testing library's `waitFor()` to wait for the next React render. We have a pattern in our old `reactRender()` that causes testing library to complain that a DOM update was not wrapped in `act()`. This function silences this warning by essentially wrapping the effects of `reactRender()` inside of an `act()` by way of `waitFor()`.
  * Other actions that caused render/updates wrapped in `waitFor()` also.

Issue: LEMS-1802

## Test plan:

`yarn test` - they should all pass and there should be no console errors logged

### DIFF
--- a/packages/math-input/src/components/keypad/__tests__/keypad-v2-mathquill.test.tsx
+++ b/packages/math-input/src/components/keypad/__tests__/keypad-v2-mathquill.test.tsx
@@ -1,4 +1,4 @@
-import {Popover} from "@khanacademy/wonder-blocks-popover";
+import {Popover, PopoverContentCore} from "@khanacademy/wonder-blocks-popover";
 import {color} from "@khanacademy/wonder-blocks-tokens";
 import {render, screen} from "@testing-library/react";
 import {userEvent as userEventLib} from "@testing-library/user-event";
@@ -68,7 +68,7 @@ function V2KeypadWithMathquill(props: Props) {
         <div style={{maxWidth: "400px", margin: "2em"}}>
             <Popover
                 content={
-                    <div>
+                    <PopoverContentCore>
                         <Keypad
                             extraKeys={["a", "b", "c"]}
                             onClickKey={handleClickKey}
@@ -86,7 +86,7 @@ function V2KeypadWithMathquill(props: Props) {
                             }
                             showDismiss
                         />
-                    </div>
+                    </PopoverContentCore>
                 }
                 dismissEnabled
                 opened={keypadOpen}
--- a/packages/perseus-editor/src/widgets/__tests__/interactive-graph-editor.test.tsx
+++ b/packages/perseus-editor/src/widgets/__tests__/interactive-graph-editor.test.tsx
@@ -5,7 +5,7 @@ import {userEvent as userEventLib} from "@testing-library/user-event";
 import * as React from "react";
 
 import {testDependencies} from "../../../../../testing/test-dependencies";
-import {waitForDeferredRenders} from "../../../../../testing/wait";
+import {waitForInitialGraphieRender} from "../../../../../testing/wait";
 import {flags} from "../../__stories__/flags-for-api-options";
 import {getDefaultFigureForType} from "../../components/util";
 import InteractiveGraphEditor from "../interactive-graph-editor";
@@ -596,7 +596,7 @@ describe("InteractiveGraphEditor", () => {
                 wrapper: RenderStateRoot,
             },
         );
-        await waitForDeferredRenders();
+        await waitForInitialGraphieRender();
 
         // Assert
         expect(ref.current?.getSaveWarnings()).toEqual([
@@ -628,7 +628,7 @@ describe("InteractiveGraphEditor", () => {
                 wrapper: RenderStateRoot,
             },
         );
-        await waitForDeferredRenders();
+        await waitForInitialGraphieRender();
 
         // Assert
         expect(ref.current?.getSaveWarnings()).toEqual([]);
--- a/packages/perseus-editor/src/widgets/__tests__/interactive-graph-editor.test.tsx
+++ b/packages/perseus-editor/src/widgets/__tests__/interactive-graph-editor.test.tsx
@@ -1,10 +1,11 @@
 import {ApiOptions, Dependencies} from "@khanacademy/perseus";
 import {RenderStateRoot} from "@khanacademy/wonder-blocks-core";
-import {render, screen} from "@testing-library/react";
+import {render, screen, waitFor} from "@testing-library/react";
 import {userEvent as userEventLib} from "@testing-library/user-event";
 import * as React from "react";
 
 import {testDependencies} from "../../../../../testing/test-dependencies";
+import {waitForDeferredRenders} from "../../../../../testing/wait";
 import {flags} from "../../__stories__/flags-for-api-options";
 import {getDefaultFigureForType} from "../../components/util";
 import InteractiveGraphEditor from "../interactive-graph-editor";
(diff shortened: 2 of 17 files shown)


packages/math-input/src/components/keypad/__tests__/keypad-v2-mathquill.test.tsx
```diff
@@ -68,7 +68,7 @@ function V2KeypadWithMathquill(props: Props) {
         <div style={{maxWidth: "400px", margin: "2em"}}>
             <Popover
                 content={
-                    <div>
+                    <PopoverContentCore>
```
jeremywiebe, line 71: The Wonder Blocks `Popover` tries to restrict the `content` prop to either a `PopoverContent` or `PopoverContentCore` element, but Typescript doesn't properly recognize that arbitrary JSX isn't valid.

packages/perseus-editor/src/widgets/__tests__/interactive-graph-editor.test.tsx
```diff
@@ -130,12 +131,13 @@ describe("InteractiveGraphEditor", () => {
             },
         );
 
-        const defaultType = screen.queryByText("Linear function");
-        const otherType = screen.getByText("Polygon");
-
         // Assert
-        expect(defaultType).not.toBeInTheDocument();
-        expect(otherType).toBeInTheDocument();
+        await waitFor(() =>
```
jeremywiebe, line 135: This will become a common theme throughout this PR. There are many places where our Perseus React components queue future renders through `setTimeout()` or `setState()` calls (or through a call to `reactRender()` helper). 

These renders do not happen unless we use `waitFor()` or the async `findBy...()` query methods. 

Without these, the tests either fail because the DOM has not been fully updated or testing-library spews warnings: `Warning: An update to Root inside a test was not wrapped in act(...).`

packages/perseus/src/__tests__/renderer-api.test.tsx
```diff
@@ -215,10 +215,12 @@ describe("Perseus API", function () {
         });
     });
 
-    describe("widget placeholders", function () {
-        it("should replace widgets with [WIDGET]", function () {
+    describe("placeholders", function () {
+        it("should replace widgets with widgetPlaceholder", function () {
             const widgetPlaceholder = (
-                <span data-testid="widget-container">[WIDGET]</span>
+                <span data-testid="widget-container" key="1">
```
jeremywiebe, line 221: The lack of `key` on these elements was causing a React warning because widgets are rendered as a list of children (even when there's only 1).

packages/perseus/src/util/react-render.tsx
```diff
@@ -2,6 +2,23 @@ import {RenderStateRoot} from "@khanacademy/wonder-blocks-core";
...
 import * as ReactDOMClient from "react-dom/client";
 
+// We keep a cache of elements that have already had createRoot() called on
+// them. If we don't do this, React 18 warns every subsequent call:
+//   > You are calling ReactDOMClient.createRoot() on a container that has
+//   > already been passed to createRoot() before.
+//
+// Note that this cache uses a WeakMap<> so that the cache isn't causing a
+// memory leak (by retaining elements after they're released everywhere else).
+const reactRootCache = new WeakMap<HTMLElement, ReactDOMClient.Root>();
```
jeremywiebe, line 12: ACHTUNG! I'd appreciate some eyes on these changes just to make sure I've not done something bad. :)

benchristel: One question: what is the behavior in the current code (with React 16) if `render` is called multiple times with a stateful component? Is the state preserved across renders? If it isn't, then I think caching the React roots will change the behavior (the state of rendered components will now be preserved).

benchristel (commented):
I left a few questions inline. I think there might be some changes needed to `waitForDeferredRenders` (though maybe I'm just misunderstanding what it does).

packages/perseus/src/util/react-render.tsx
jeremywiebe: The deprecated `render()` [documentation](https://react.dev/reference/react-dom/render#caveats) says that multiple calls with the same host DOM node will maintain state (at least that's how I read it):

> If you call render on the same domNode more than once, React will update the DOM as necessary to reflect the latest JSX you passed. React will decide which parts of the DOM can be reused and which need to be recreated by [“matching it up”](https://react.dev/learn/preserving-and-resetting-state) with the previously rendered tree. Calling render on the same domNode again is similar to calling the [set function](https://react.dev/reference/react/useState#setstate) on the root component: React avoids unnecessary DOM updates.

testing/wait.ts
```diff
@@ -91,3 +93,11 @@ const FRAME_DURATION = 17;
  */
 export const waitForAnimationFrame: () => Promise<void> = () =>
     wait({delay: FRAME_DURATION});
+
+/**
+ *
+ * @returns a Promise<void> that resolves after the
```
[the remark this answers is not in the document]
jeremywiebe: Come back next week for the conclusion of this riveting PR!

Thanks for calling out the issue here. I had found this solution on a Github issue comment and it seemed to work. But based on your comment and finding Kent Dodds "common mistakes" [entry](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#passing-an-empty-callback-to-waitfor) I've re-worked this a bit and renamed it. 

Now it specifically waits for Graphie to do a render pass. This causes all related tests to pass and no `act()` wrapper warnings. 

I think this is cleaner and hopefully easier to follow/more accurate.

testing/wait.ts
```diff
@@ -95,9 +95,14 @@ export const waitForAnimationFrame: () => Promise<void> = () =>
     wait({delay: FRAME_DURATION});
 
 /**
+ * Waits for the initial Graphie render.
  *
- * @returns a Promise<void> that resolves after the
+ * @returns a Promise<void> that resolves when the Graphie graph has renderred.
+ * The behaviour is undefined if there are multiple Graphies on the page (which
+ * is unlikely in tests).
```
benchristel, line 102: Thanks for updating this! Much clearer now :)

benchristel (approved):

packages/perseus/src/util/react-render.tsx
```diff
@@ -2,6 +2,23 @@ import {RenderStateRoot} from "@khanacademy/wonder-blocks-core";
 import * as React from "react";
 import * as ReactDOMClient from "react-dom/client";
 
+// We keep a cache of elements that have already had createRoot() called on
+// them. If we don't do this, React 18 warns every subsequent call:
+//   > You are calling ReactDOMClient.createRoot() on a container that has
+//   > already been passed to createRoot() before.
```
jeresig, line 8: Interesting... I'm surprised we haven't seen this in webapp. cc: @somewhatabstract

jeremywiebe: It looks like webapp needed to do this also: https://github.com/Khan/webapp/blob/5133d22218437f9da3a7ef11bc89850b61378ac3/services/static/javascript/rendering/deprecated-react-render.tsx#L19

PR #1420, opened 2024-07-16
Simplify and rename an interactive graph reducer action

benchristel:
## Summary:
The name `moveControlPoint` was not an accurate description of what this action
was doing.  I've renamed it to `movePointInFigure` which hopefully makes it
clear what it's doing and why it takes 3 arguments.  The linear and ray graph
types were using this action, passing a `0` as the `figureIndex` argument
because those graph types only have one figure. I've wrapped the action
constructor for these graph types so they no longer have to pass the dummy
argument.

Issue: none

## Test plan:

`yarn test`

### DIFF
--- a/.changeset/tricky-ravens-kneel.md
+++ b/.changeset/tricky-ravens-kneel.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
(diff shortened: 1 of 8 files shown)


jeremywiebe (approved):
That's a nice little improvement in readability and clarity!  Thank-you.

PR #1421, opened 2024-07-17
Type configured userEvent field for easier testing DX

jeremywiebe:
## Summary:

When I upgraded Perseus to `user-event` v14, tests that used fake timers needed to initialize an instance of the `user-event` library (because we needed to tell it how to advance the fake timers in Jest). 

The configured variable was untyped and meant that we don't get any intellisense/code completion when using it, so this PR properly types the `let userEvent` declarations!

### Before 

image

### After 

image

Issue: "none"

## Test plan:

Open one of the affected test files and type `userEvent.` in one of the tests. Note that your editor (if it supports this) should list the functions available on the `userEvent` instance!

### DIFF
--- a/packages/math-input/src/components/__tests__/integration.test.tsx
+++ b/packages/math-input/src/components/__tests__/integration.test.tsx
@@ -80,7 +80,7 @@ function ConnectedMathInput({keypadConfiguration = defaultConfiguration}) {
 }
 
 describe("math input integration", () => {
-    let userEvent;
+    let userEvent: ReturnType<typeof userEventLib.setup>;
     beforeEach(() => {
         userEvent = userEventLib.setup({
             advanceTimers: jest.advanceTimersByTime,
--- a/packages/math-input/src/components/__tests__/integration.test.tsx
+++ b/packages/math-input/src/components/__tests__/integration.test.tsx
@@ -15,6 +15,7 @@ import {MobileKeypad} from "../keypad";
 import {KeypadContext, StatefulKeypadContextProvider} from "../keypad-context";
 
 import type {KeypadConfiguration} from "../../types";
+import type {UserEvent} from "@testing-library/user-event";
 
 const MQ = MathQuill.getInterface(3);
 
@@ -80,7 +81,7 @@ function ConnectedMathInput({keypadConfiguration = defaultConfiguration}) {
 }
 
 describe("math input integration", () => {
-    let userEvent: ReturnType<typeof userEventLib.setup>;
+    let userEvent: UserEvent;
     beforeEach(() => {
         userEvent = userEventLib.setup({
             advanceTimers: jest.advanceTimersByTime,
--- a/packages/perseus-editor/src/__tests__/editor.test.tsx
+++ b/packages/perseus-editor/src/__tests__/editor.test.tsx
@@ -45,7 +45,7 @@ describe("Editor", () => {
         Widgets.registerEditors([ImageEditor]);
     });
 
-    let userEvent;
+    let userEvent: ReturnType<typeof userEventLib.setup>;
     beforeEach(() => {
         userEvent = userEventLib.setup({
             advanceTimers: jest.advanceTimersByTime,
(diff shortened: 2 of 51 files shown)


packages/math-input/src/components/__tests__/integration.test.tsx
```diff
@@ -80,7 +80,7 @@ function ConnectedMathInput({keypadConfiguration = defaultConfiguration}) {
 }
 
 describe("math input integration", () => {
-    let userEvent;
+    let userEvent: ReturnType<typeof userEventLib.setup>;
```
Myranae, line 83: question, nb: So this means whatever the return type is for the `userEventLib.setup` function is the type of this userEvent variable, right?

packages/perseus-editor/src/__tests__/editor.test.tsx
```diff
@@ -80,8 +80,9 @@ describe("Editor", () => {
         render(<Harnessed onChange={onChangeMock} />);
 
         // Act
-        screen.getByRole("button", {name: "Remove image widget"}),
-            await userEvent.click();
+        await userEvent.click(
```
handeyeco, line 83: 👍

packages/perseus-editor/src/__tests__/editor.test.tsx
```diff
@@ -80,8 +80,9 @@ describe("Editor", () => {
...
+            screen.getByRole("button", {name: "Remove image widget"}),
+        );
```
Myranae, line 85: question, nb: Did the syntax that was previously here not work? Or is this change mostly for readability?

packages/math-input/src/components/__tests__/integration.test.tsx
handeyeco, line 83: I'm wondering if it'd be useful to alias this so it's clearer.

```
type UserEvent = ReturnType<typeof userEventLib.setup>;
```
Then
```
let userEvent: UserEvent;
```

Not a blocker, I'll defer to you.

handeyeco (approved):
Nice

Myranae (approved):
Was able to go into an affected test and see options after typing `userEvent.` into it! Yay code completion~ Added a couple questions just for learning purposes, but this looks pretty straightforward to me :) I'm curious though, why didn't it get a type after it was redeclared?

packages/perseus-editor/src/__tests__/editor.test.tsx
jeremywiebe: It was incorrect. It looks like deep in the library there's some defaulting to use the previous target as the target when `click()` is called, but the API types definitely don't allow for a missing element. I think it was sheer luck that this worked.

packages/math-input/src/components/__tests__/integration.test.tsx
jeremywiebe: That's exactly correct.

jeremywiebe:
> I'm curious though, why didn't it get a type after it was redeclared?

I just didn't think to! :(

Myranae:
@jeremywiebe Oh, no, I meant, why didn't it automatically happen after this line: `userEvent = userEventLib.setup({
            advanceTimers: jest.advanceTimersByTime,
        });`? Shouldn't this show that now `userEvent` should be the type returned by `userEventLib.setup`?  I thought with TypeScript it would then infer the type of `userEvent` moving forward, but we needed to declare the type manually.

jeremywiebe:
> @jeremywiebe Oh, no, I meant, why didn't it automatically happen after this line: `userEvent = userEventLib.setup({ advanceTimers: jest.advanceTimersByTime, });`? Shouldn't this show that now `userEvent` should be the type returned by `userEventLib.setup`? I thought with TypeScript it would then infer the type of `userEvent` moving forward, but we needed to declare the type manually.

Ah, that! When we declare `let userEvent;` TypeScript has no way of knowing what type it will be so it's typed as `any`. TypeScript doesn't know Jest and so it doesn't understand when `beforeEach()` will be called or if will be called at all. In a normal function TS would be able to do flow analysis and narrow a type based on `if` and `switch case` checks.

packages/math-input/src/components/__tests__/integration.test.tsx
jeremywiebe: Nice.  The library even provides an exported type for this: `UserEvent`!

PR #1423, opened 2024-07-18
Resolve LEMS-2186 - no longer need 'off by 0.5' startCoords

jeremywiebe:
## Summary:

At the end of the React 18 work I noticed two failing tests in the Interactive Graph suite. I found that its the two tests that had to be customized to pass related to `startCoords`. 

If I remove this customization, all tests pass. My hunch is that with React 16 and the older `@testing-library` package, there was some async DOM updates/events that didn't flush all the way leaving the graph in a non-final state after keypresses, thus failing the tests. 

Issue: LEMS-2186

## Test plan:

`yarn test` ✅✨🎉😍

### DIFF
--- a/types/global.d.ts
+++ b/types/global.d.ts
@@ -1,2 +1,0 @@
-// eslint-disable-next-line import/no-unassigned-import
-import "jest-extended"; // we love the side effects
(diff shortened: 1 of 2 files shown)


types/global.d.ts
jeremywiebe, line 1: I had introduced this earlier in my React 18 work (sorry, it was unrelated), but I've since removed `jest-extended` from Perseus entirely. This was left behind after merging.

benchristel (approved):
LGTM! Thanks!

FWIW, I find it a little strange that these test data fixtures are named `DefaultCorrect` but there's nothing in the builder call that sets up the correct answer to be the same as the initial coordinates. The point of the builder is that every test can be explicit about what it needs and let the stuff it doesn't care about default. Here it seems like we're relying on the builder's defaults to set things up the way we need, which isn't necessarily readable or safe.

nishasy (approved):
Idk how this fixed the off by 0.5 issue, but that is very exciting!

PR #1424, opened 2024-07-18
[Expression Widget] Disabling portal functionality on Popover implementation

catandthemachines:
## Summary:
This update includes:
- Updating @khanacademy/wonderblocks-popover package version
- Disabiling portaling popover functionality, ensuring keyboard navigation works for expression widget.
- Removing timeout logic for portal-hack.

Issue: LEMS-2087

## Test plan:

- Navigate to /?path=/story/perseus-widgets-expression--desktop-kitchen-sink
- Open expression widget
- Notice that you can't hit tab key to navigate all tabs in expression widget. Tab will get you to the first element, then use left and right arrow-keys to navigate to other tab keys. Hit tab again to go to tab content.

### DIFF
--- a/packages/math-input/package.json
+++ b/packages/math-input/package.json
@@ -44,7 +44,7 @@
         "@khanacademy/mathjax-renderer": "^2.0.0",
         "@khanacademy/wonder-blocks-clickable": "4.2.1",
         "@khanacademy/wonder-blocks-core": "6.4.0",
-        "@khanacademy/wonder-blocks-popover": "3.2.2",
+        "@khanacademy/wonder-blocks-popover": "3.2.9",
         "@khanacademy/wonder-blocks-timing": "5.0.0",
         "@khanacademy/wonder-blocks-tokens": "1.3.0",
         "@khanacademy/wonder-stuff-core": "1.5.2",
--- a/packages/math-input/package.json
+++ b/packages/math-input/package.json
@@ -64,7 +64,7 @@
         "@khanacademy/mathjax-renderer": "^2.0.0",
         "@khanacademy/wonder-blocks-clickable": "4.2.1",
         "@khanacademy/wonder-blocks-core": "6.4.0",
-        "@khanacademy/wonder-blocks-popover": "3.2.2",
+        "@khanacademy/wonder-blocks-popover": "3.2.9",
         "@khanacademy/wonder-blocks-timing": "5.0.0",
         "@khanacademy/wonder-blocks-tokens": "1.3.0",
         "@khanacademy/wonder-stuff-core": "1.5.2",
--- a/packages/math-input/src/components/keypad/__tests__/__snapshots__/keypad.test.tsx.snap
+++ b/packages/math-input/src/components/keypad/__tests__/__snapshots__/keypad.test.tsx.snap
@@ -19,13 +19,13 @@ exports[`keypad should snapshot expanded: first render 1`] = `
             aria-disabled="false"
             aria-label="Numbers"
             aria-selected="true"
-            class="button_vr44p2-o_O-reset_152ygtm-o_O-link_13xlah4-o_O-clickable_1ncqa8p"
+            class="button_vr44p2-o_O-reset_152ygtm-o_O-link_13xlah4-o_O-focused_en8zhl-o_O-clickable_1ncqa8p"
             role="tab"
             tabindex="0"
             type="button"
           >
             <div
-              class="default_xu2jcg-o_O-base_1pupywz"
+              class="default_xu2jcg-o_O-base_1pupywz-o_O-focused_2hwz1v"
             >
               <div
                 class="default_xu2jcg-o_O-innerBox_qdbgl3"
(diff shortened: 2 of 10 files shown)


packages/math-input/package.json
```diff
@@ -80,4 +80,4 @@
         "react-transition-group": "^4.4.1"
     },
     "keywords": []
-}
+}
```
jeremywiebe, line 83: Would you mind adding the trailing newline back in?  Some tools (especially CLI ones) misbehave if they're missing.

packages/math-input/src/components/tabbar/item.tsx
```diff
@@ -89,26 +89,12 @@ function TabbarItem(props: TabItemProps): React.ReactElement {
...
-            timeout = setTimeout(() => {
-                if (tabRef?.current) {
-                    // Move element into view when it is focused
-                    tabRef?.current.focus();
-                }
-            }, 0);
+            if (tabRef?.current) {
+                // Move element into view when it is focused
+                tabRef?.current.focus();
+            }
```
jeremywiebe, line 96: **suggestion**: We can make this a bit more concise by using optional chaining. 

```suggestion
            // Move element into view when it is focused
            tabRef?.current?.focus();
```

packages/perseus/package.json
```diff
@@ -118,4 +118,4 @@
         "underscore": "^1.4.4"
     },
     "keywords": []
-}
+}
```
jeremywiebe, line 121: **nitpick**: newline

packages/math-input/package.json
```diff
@@ -44,7 +44,7 @@
         "@khanacademy/mathjax-renderer": "^2.0.0",
         "@khanacademy/wonder-blocks-clickable": "4.2.1",
         "@khanacademy/wonder-blocks-core": "6.4.0",
-        "@khanacademy/wonder-blocks-popover": "3.2.2",
+        "@khanacademy/wonder-blocks-popover": "3.2.9",
```
jeremywiebe, line 47: **todo**: our WB dependencies in Perseus are a bit annoying to manage.  We need to keep the `devDependencies` in sync with the `peerDependencies` later in the same file.

The `devDependencies` are what our build/test/storybook tooling look at, but the `peerDependencies` help to notify consuming apps (webapp) what our base level of the dependency is.

packages/perseus/package.json
```diff
@@ -61,7 +61,7 @@
         "@khanacademy/wonder-blocks-layout": "2.0.32",
         "@khanacademy/wonder-blocks-link": "6.1.1",
         "@khanacademy/wonder-blocks-pill": "2.2.1",
-        "@khanacademy/wonder-blocks-popover": "3.2.2",
+        "@khanacademy/wonder-blocks-popover": "3.2.9",
```
jeremywiebe, line 64: **todo**: our WB dependencies in Perseus are a bit annoying to manage.  We need to keep the `devDependencies` in sync with the `peerDependencies` later in the same file.

The `devDependencies` are what our build/test/storybook tooling look at, but the `peerDependencies` help to notify consuming apps (webapp) what our base level of the dependency is.

packages/math-input/src/components/keypad/__tests__/__snapshots__/keypad.test.tsx.snap
```diff
@@ -19,13 +19,13 @@ exports[`keypad should snapshot expanded: first render 1`] = `
             aria-disabled="false"
             aria-label="Numbers"
             aria-selected="true"
-            class="button_vr44p2-o_O-reset_152ygtm-o_O-link_13xlah4-o_O-clickable_1ncqa8p"
+            class="button_vr44p2-o_O-reset_152ygtm-o_O-link_13xlah4-o_O-focused_en8zhl-o_O-clickable_1ncqa8p"
             role="tab"
             tabindex="0"
             type="button"
           >
```
jeremywiebe, line 26: **question**: Is the fact that these now have `focused` in the class name relevant to this PR? I realize it's a snapshot, but this seems like a behaviour change.

jeremywiebe (approved):
This seems reasonable to me. 👍  Just note the `package.json` comments as I think they need to be addressed before landing.

packages/perseus/src/components/math-input.tsx
```diff
@@ -66,6 +66,20 @@ type Props = {
...
+     * the trigger location.
+     *
+     * Set to "true" by default.
+     *
+     * CAUTION: Turning off portal could cause some clipping issues
+     * especially around legacy code with usage of z-indexing,
+     * Use caution when turning this functionality off and ensure
+     * your content does not get clipped or hidden.
+     */
+    portal?: boolean;
```
handeyeco, line 82: I could totally be wrong, but I think at this point the only component that uses `perseus/src/components/math-input` is Expression. So a different approach would be to remove this prop and hard-code it to `true` on L348.

handeyeco (approved):
Looks good! 🌈 Do we need tests for this?

PR #1425, opened 2024-07-18
React 18: fix bundle size bloat

jeremywiebe:
## Summary:

@handeyeco [noticed](https://github.com/Khan/perseus/pull/1064#issuecomment-2234053372) that the bundle size for `math-input` grew considerably in the React 18 base branch (#1064). 

image

It turns out that `react` and `react-dom` were no longer listed as `peerDependencies` in its `package.json`. I found the same regression in `@khanacademy/perseus`'s `package.json` file. 

We use a Rollup plugin ([`rollup-plugin-auto-external`](https://github.com/stevenbenisek/rollup-plugin-auto-external)) to automatically mark peer dependencies as external to avoid bundling them into our packages. When `react` and `react-dom` were excluded from `peerDependencies` they got bundled in and that was why the bundle size exploded. 

Issue: LEMS-1802

## Test plan:

I ran `yarn build` locally and checked the generated `index.js` file in each `dist` folder. They no longer contain the React code.

I will also re-check the bundle size comment in the PR once it updates.


handeyeco (approved):
Thanks!

PR #1428, opened 2024-07-18
Remove 'katex' from @khanacademy/perseus

jeremywiebe:
## Summary:

KaTeX was retired from use in Perseus in 2023. Today, I noticed that we still reference the `katex` package in the perseus `package.json`, but we never import it anywhere in the code. 

So I'm removing that dependency. I've also updated some code comments and storybook references to not mention KaTeX anymore. I've left some references to CSS classes and the KaTeX_Main font ref. I'm unsure if they might be important somewhere (especially if host applications may still set it outside of Perseus).

So this doesn't fully remove all references to katex, but it moves us another step closer. 

Issue: "none"

## Test plan:

`yarn tsc`
`yarn lint`

I ran storybook and opened the `ZoomableTex` stories and played with them. 

Note(jeremy): I've restored the code that swaps `\begin{align}` to `\begin{aligned}` and noted LEMS-1608 based on comments from Ben (thanks!).

~I also went to the `EditorPage` story and pasted in the following Tex into the **Question** field (which uses the macro that we used to have to replace for KaTeX, but MathJax supports properly: switching `\begin{align}` to `\begin{aligned}`).~


packages/perseus/src/components/__stories__/graph.stories.tsx
```diff
@@ -1,7 +1,6 @@
 import * as React from "react";
 
 import Graph from "../graph";
-// TODO(scottgrant): Katex is unavailable here. Fix!
```
jeremywiebe, line 4: I couldn't even make sense of what was to be done here so I'm deleting this comment.

packages/perseus/src/renderer.tsx
```diff
@@ -1289,12 +1289,6 @@ class Renderer extends React.Component<Props, State> {
             );
         }
         if (node.type === "math") {
-            // Replace uses of \begin{align}...\end{align} which KaTeX doesn't
-            // support (yet) with \begin{aligned}...\end{aligned} which renders
-            // the same is supported by KaTeX.  It does the same for align*.
-            // TODO(kevinb) update content to use aligned instead of align.
-            const tex = node.content.replace(/\{align[*]?\}/g, "{aligned}");
```
jeremywiebe, line 1296: MathJax supports the `\begin{align}` macro just fine (tested in Storybook).

packages/perseus/src/types.ts
```diff
@@ -318,7 +318,6 @@ export type APIOptions = Readonly<{
 
 type TeXProps = {
     children: string;
-    katexOptions?: any;
```
jeremywiebe, line 321: I checked and this prop/string is not used in webapp anywhere!

packages/perseus/src/util/tex-preprocess.ts
jeremywiebe, line 1: Git(hub) doesn't seem to recognize that this is a rename.

packages/perseus/src/renderer.tsx
benchristel: Removing this code might break the exercise editor, which still uses KaTeX behind the scenes to validate that the input TeX _can_ render with KaTeX. This gets complicated to reason about, because the reason it does that is to ensure that the TeX content creators type is compatible with old versions of the mobile app, which use KaTeX (and old versions of Perseus which have this replacement logic, so `\begin{align}` will work fine there).

I _think_ we could move this logic to the exercise editor and it would be fine. But I'd lean toward keeping this code the same until we can remove KaTeX everywhere. If you agree, can you add a TODO comment here referencing [LEMS-1608](https://khanacademy.atlassian.net/browse/LEMS-1608)?

packages/perseus/src/util/tex.ts
```diff
@@ -15,8 +15,8 @@ function findChildOrAdd(elem: any, className: string) {
 }
 
 export default {
-    // Process a node and add math inside of it. This attempts to use KaTeX to
-    // format the math, and if that fails it falls back to MathJax.
+    // Process a node and add math inside of it. This uses MathJax
```
benchristel, line 18:
```suggestion
    // Process a node and add math inside of it. This uses MathJax to
```

benchristel (commented):
I think removing the `\begin{align}` replacement will break the exercise editor. I left a note inline. So we either need to keep the replacement code, or move it to the exercise editor. Otherwise, this looks good!

packages/perseus/src/renderer.tsx
benchristel: FWIW, I looked up our new [mobile deprecation policy](https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/2883780614/ADR+732+Formalize+the+graphql+deprecation+policy+for+fields+used+by+the+mobile+app) and I believe we will be able to completely remove KaTeX starting September 18th!

jeremywiebe: I'd like to move forward and land this, so I'll do as Ben suggests and restore this code with a new TODO.  Thanks for thinking of that Ben!

benchristel (approved):
LGTM!

PR #1431, opened 2024-07-22
Add hintMode to widget and Renderer props

benchristel:
## Summary:
<Add your summary here>

Issue: XXX-XXXX

## Test plan:

### DIFF
--- a/packages/perseus/src/mixins/widget-prop-denylist.ts
+++ b/packages/perseus/src/mixins/widget-prop-denylist.ts
@@ -23,6 +23,7 @@ const denylist = [
     "apiOptions",
     "questionCompleted",
     "findWidgets",
+    "hintMode",
     // added by src/editor.jsx, for widgets removing themselves
     // this is soooo not the right place for this, but alas.
     "onRemove",
--- a/packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.tsx
@@ -31,6 +31,7 @@ export type StatefulMafsGraphProps = {
     showTooltips: Required<InteractiveGraphProps["showTooltips"]>;
     showProtractor: boolean;
     labels: InteractiveGraphProps["labels"];
+    hintMode: boolean;
 };
 
 // Rather than be tightly bound to how data was structured in
--- a/.changeset/pretty-points-drop.md
+++ b/.changeset/pretty-points-drop.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/perseus": minor
+"@khanacademy/perseus-editor": minor
+---
+
+Add a `hintMode` prop to `Renderer` and widgets.
--- a/packages/perseus-editor/src/__stories__/editor-page-with-storybook-preview.tsx
+++ b/packages/perseus-editor/src/__stories__/editor-page-with-storybook-preview.tsx
@@ -123,6 +123,7 @@ function EditorPageWithStorybookPreview(props: Props) {
                             <Renderer
                                 strings={mockStrings}
                                 apiOptions={apiOptions}
+                                hintMode={true}
                                 {...hint}
                             />
                         </View>
--- a/packages/perseus/src/__stories__/hints-renderer.stories.tsx
+++ b/packages/perseus/src/__stories__/hints-renderer.stories.tsx
@@ -2,9 +2,9 @@ import {View} from "@khanacademy/wonder-blocks-core";
 import React from "react";
 
 import HintsRenderer from "../hints-renderer";
+import {interactiveGraphQuestionBuilder} from "../widgets/interactive-graphs/interactive-graph-question-builder";
 
 import type {Meta, StoryObj} from "@storybook/react";
-import {interactiveGraphQuestionBuilder} from "../widgets/interactive-graphs/interactive-graph-question-builder";
 
 const meta: Meta<typeof HintsRenderer> = {
     title: "Perseus/Renderers/Hints Renderer",
--- a/packages/perseus/src/__stories__/hints-renderer.stories.tsx
+++ b/packages/perseus/src/__stories__/hints-renderer.stories.tsx
@@ -4,6 +4,9 @@ import React from "react";
 import HintsRenderer from "../hints-renderer";
 
 import type {Meta, StoryObj} from "@storybook/react";
+import {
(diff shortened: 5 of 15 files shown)


packages/perseus/src/mixins/widget-prop-denylist.ts
```diff
@@ -23,6 +23,7 @@ const denylist = [
     "apiOptions",
     "questionCompleted",
     "findWidgets",
+    "hintMode",
```
benchristel, line 26: Stuff like this makes me nervous that adding new props to widgets is not an entirely trivial process :/

Myranae: Question: How did you know this needed to get added here? Is that what made you nervous? That there are steps involved that we may not be fully aware of or that these steps even exist? I'm guessing the latter as I write this out, but could be both :)

benchristel: I searched the codebase for one of the other props in this list, with the goal of finding places where widget props are created. That led me to this file.

Tangent:

Looking more closely at how this list is used, I see: https://github.com/Khan/perseus/blob/7bf32b73395a8bc14f9198ce183de0e7706bee86/packages/perseus/src/widgets/passage-ref.tsx#L79-L83

This function is supposed to return a `UserInput`, so perhaps what we ought to be doing here is just creating an object that has the fields in `UserInput`. Then we could remove the denylist.

```ts
getUserInput: () => UserInput = () => ({
  static: this.props.static,
  reviewModeRubric: this.props.reviewModeRubric,
  linterContext: this.props.linterContext,
  isLastUsedWidget: this.props.isLastUsedWidget,
  alignment: this.props.alignment,
});
```

However, a quick search makes it seem like the return value of `getUserInput` is not actually used (it's passed to `PassageRef.validate()`, but that function doesn't do anything with it and just returns a hardcoded value). So I think maybe we can just delete all this code.

packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.tsx
```diff
@@ -31,6 +31,7 @@ export type StatefulMafsGraphProps = {
     showTooltips: Required<InteractiveGraphProps["showTooltips"]>;
     showProtractor: boolean;
     labels: InteractiveGraphProps["labels"];
+    hintMode: boolean;
```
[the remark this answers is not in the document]
benchristel: `InteractiveGraphProps["hintMode"]` is equivalent to `boolean`, as far as the TypeScript compiler is concerned. But the meaning of `boolean` is immediately obvious to anyone who knows TypeScript, while to find out what `InteractiveGraphProps["hintMode"]` means, you have to go digging through two other files. You also have to know that `hintMode` is common to all widgets and not a special interactive-graph prop; otherwise, you're unlikely to look in the right place. I dislike having to dig through layers of type declarations to figure out what the actual type of something is, so I prefer to write types like `boolean` instead of `InteractiveGraphProps["hintMode"]`.

packages/perseus/src/mixins/widget-prop-denylist.ts
benchristel: Okay, `getUserInput` _is_ used, by the Renderer (which calls it on each widget).

Myranae: 😂 Got excited about potentially just deleting it all. It looks like the denylist is used in other places as well, but `removeDenylistProps` only seems to be used with the `getUserInput` here. Is the `getUserInput` used by Renderer this same method from PassageRef? In any case, probably still good to add `hintMode` here if it should be excluded when widgets serialize themselves. Thanks for looking into this!

packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.tsx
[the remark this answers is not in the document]
Myranae: I always assumed it was done that way so that if the type of one changed it updated the type of the other. I think there are probably other ways of making sure to update it in all the right places though, such as tests and such, right?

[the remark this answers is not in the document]
benchristel: The type system / `tsc` itself will ensure that the types are kept consistent. That's the main benefit of a type system - keeping different things consistent 😄

packages/perseus/src/mixins/widget-prop-denylist.ts
jeremywiebe: `getUserInput` is used as part of the scoring system. Each widget provides a `simpleValidate()` function that the Renderer calls. This function scores the widget in its current state. 

`simpleValidate` typically calls the `getUserInput` instance function on itself. This function returns the input(s) that the user has provided (the selected radio option(s), the graph coordinate that the user has moved the point to, etc).  The widget then calls a `validate()` static method on itself and passes the user input and the "rubric" (the widget options that define the correct answer) and returns the score (correct/incorrect). 

The `removeDenylistProps` in this flow is a cheat and I'd love to get rid of it. Basically, it returns all widget props as the user input, except those that show up in the `denylist`.  

Instead `getUserInput` on each widget should just return an object that represents the minimum necessary set of values (from props) that are needed by `simpleValidate`.

packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.tsx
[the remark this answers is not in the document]
jeremywiebe: `InteractiveGraphProps` and all of the types defined in `perseus-types.ts` should not include props that the `Renderer` provides programmatically. The types in `perseus-types.ts` could be thought of as the data schema for Perseus (the serialized, at rest data format). 

So Ben has it the right way here. 

In the widget definitions, we've used the format `InteractiveGraphProps["some_prop"]` as a signal that the prop comes from the serialized widget options as opposed to something that is derived at runtime. All of these types/props could use a big cleanup/standardization, but it didn't get done during our Perseus Stabilization project and hasn't become a high enough priority yet. 

I'd love to see a "gold standard" widget definition that included a convention for defining all the necessary types, file layout, test file location, etc. :) #someday.

jeremywiebe (approved):

Myranae (approved):
Looks good to me, and I was able to get my changes working after pulling in your branch, so it works!

PR #1432, opened 2024-07-22
Ensure the new Mafs-based Angle Graphs score correctly 

SonicScrewdriver:
## Summary:
The new Mafs-based Angle graphs were being marked as incorrect if the user moves the bottom point clockwise, which was not caught earlier as there was a bug in the implementation of the legacy Angle graphs that allowed users to create reflexive angles when they were not supposed to be able to. 

The new Mafs-based Angle graph solves this issue, but it uncovered a bug in our scoring that fails to handle this particular state. 

This ticket adds the logic necessary to reverse the coordinates (for scoring purposes) when they're clockwise and reflexive graphs are not allowed. This will be the approach used until we are able to remove/deprecate the legacy Angle graphs. 

When the legacy Angle graph is removed, it's highly recommended that this logic be moved directly to the scoring function in `interactive-graph.tsx`, so that there's no future confusion about how the scoring is being calculated.  Upgrading this graph was very difficult due to the previous logic being split across many different files, so this will save us a lot of headaches down the road. The ticket to house that work can be found [here](https://khanacademy.atlassian.net/browse/LEMS-2190). 

Issue: LEMS-2165

## Test plan:
- Manual testing using Flipbook
- Creation of new regression tests to test all 3 states to ensure that there's no future regressions during the transition (These should be moved along with the logic during LEMS-2190) 

## Video: 
https://github.com/user-attachments/assets/de76c0be-085d-4090-9ba8-598de27de0b4

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-state.test.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-state.test.ts
@@ -31,8 +31,9 @@ describe("getGradableGraph", () => {
         expect(result).toEqual(initialGraph);
     });
 
-    // (LEMS-2190): This test is to ensure that the new Mafs graph is reversing coordinates when the angle graph is reflexive
-    // in a clockwise direction in order to temporarily maintain the same angle scoring with the legacy graph.
+    // (LEMS-2190): This test is to ensure that the new Mafs graph is reversing coordinates when the angle graph
+    // is reflexive in a clockwise direction in order to temporarily maintain the same angle scoring with the
+    // legacy graph. This logic (& the tests) should be moved to the scoring function after removing the legacy graph.
     it("returns reversed coordinates if the angle graph is reflexive when not allowed", () => {
         const state: InteractiveGraphState = {
             type: "angle",
--- a/packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-state.test.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-state.test.ts
@@ -1,3 +1,5 @@
+import invariant from "tiny-invariant";
+
 import {getGradableGraph} from "./interactive-graph-state";
 
 import type {InteractiveGraphState} from "../types";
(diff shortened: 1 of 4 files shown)


packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-state.ts
```diff
@@ -77,9 +79,24 @@ export function getGradableGraph(
...
+        // the same angle scoring with the legacy graph (which had a bug).
+        // (LEMS-2190): When we remove the legacy graph, move this logic to the scoring function.
+        const areClockwise = clockwise([
+            state.coords[0],
+            state.coords[2],
+            state.coords[1],
+        ]);
+        const shouldReverseCoords = areClockwise && !state.allowReflexAngles;
+        const coords: [Coord, Coord, Coord] = shouldReverseCoords
+            ? (state.coords.slice().reverse() as [Coord, Coord, Coord])
```
jeremywiebe, line 94: I _think_ you can get away without the `as` if you use: 

```suggestion
            ? ([...state.coords].reverse()
```

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-state.ts
```diff
@@ -77,9 +79,24 @@ export function getGradableGraph(
...
+            : state.coords;
+
         return {
             ...initialGraph,
-            coords: state.coords,
+            coords: coords,
```
jeremywiebe, line 99: **nit**

```suggestion
            coords,
```

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-state.ts
```diff
@@ -77,9 +79,24 @@ export function getGradableGraph(
     }
 
     if (state.type === "angle" && initialGraph.type === "angle") {
+        // We're going to reverse the coords for scoring if the angle is clockwise and
+        // we don't allow reflex angles. This is because the angle is largely defined
+        // by the order of the points in the coords array, and we want to maintain
+        // the same angle scoring with the legacy graph (which had a bug).
+        // (LEMS-2190): When we remove the legacy graph, move this logic to the scoring function.
```
jeremywiebe, line 86: Thanks for leaving such a great, explanatory comment.  I was thinking about removing the old code and how we'd know about LEMS-2190 at that time.  What do you think of also adding a reference to LEMS-2190 at the top of the legacy interactive-graph file?  Or near some code that we know we'll be touching when we delete it.

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-state.test.ts
```diff
@@ -28,4 +30,94 @@ describe("getGradableGraph", () => {
 
         expect(result).toEqual(initialGraph);
     });
+
+    // (LEMS-2190): This test is to ensure that the new Mafs graph is reversing coordinates when the angle graph
+    // is reflexive in a clockwise direction in order to temporarily maintain the same angle scoring with the
+    // legacy graph. This logic (& the tests) should be moved to the scoring function after removing the legacy graph.
+    it("returns reversed coordinates if the angle graph is reflexive when not allowed", () => {
+        const state: InteractiveGraphState = {
```
jeremywiebe, line 38: Thinking about the builder work that Ben/Nisha have done... in these three tests, the only bits that change are the `coords` and `allowReflexAngles` right? 

I wonder if it's worth building a InteractiveGraphState like this: 

```ts
const defaultState: InteractiveGraphState = {
    type: "angle",
    coords: [
        [-5, 0],
        [0, 0],
        [5, 5],
    ],
    hasBeenInteractedWith: true,
    range: [
        [-10, 10],
        [-10, 10],
    ],
    snapStep: [1, 1],
    showAngles: true,
    allowReflexAngles: false,
};
```

And then in the tests:

```ts
    const state = {
        ...defaultState,
        coords: [[], [], []],
        allowReflexAngles: true, // or false
    }
```

That would distill the data to only the bits that area essential to the test. 🤔

jeremywiebe (approved):

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-state.ts
SonicScrewdriver: Sadly no luck here :( 

Reverse returns a generic array (Coord[]), and we're looking specifically for [Coord, Coord, Coord], so typescript becomes unhappy.

jeremywiebe: Ah right!  It's `reverse()` that's part of the problem!

PR #1434, opened 2024-07-22
Ensure endpoints cannot be placed too close to the vertex on Angle Graphs

SonicScrewdriver:
## Summary:
In a play-test bug report, it was noted that sometimes the angle graph gets stuck. Upon further investigation, it was determined that this was due to the "tooClose" logic only being considered when we are moving the vertex, and not the end points. 

This PR rectifies that omission and ensures that points can never be closer than 2 graph units to the vertex. The resulting behaviour mimics the legacy graph. Some new tests were also added to help catch any future regressions of this behaviour. 

Issue: LEMS-2153

## Test plan:
- Manual Testing in Flipbook 
- New regression tests

### DIFF
--- a/.changeset/soft-mangos-teach.md
+++ b/.changeset/soft-mangos-teach.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Bug fix to ensure that Angle graph endpoints cannot get too close to the vertex.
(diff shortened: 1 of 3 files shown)


jeremywiebe (approved):
That makes sense.  LGTM.

benchristel:
@SonicScrewdriver I think this issue is fixed by https://github.com/Khan/perseus/pull/1472. Can this PR be closed now?

benchristel:
I think this behavior is implemented another way now, so I'm going to close this PR. Feel free to reopen if it's still needed!

PR #1435, opened 2024-07-23
Remove StatefulArticleEditor

handeyeco:
## Summary:
I couldn't find that it was being used at all

[Search](https://github.com/search?q=org%3AKhan%20StatefulArticleEditor&type=code)


SonicScrewdriver (approved):
Sweet! Thanks for getting rid of some archaic legacy code that's not being used.

PR #1437, opened 2024-07-23
Update Polygon to fill only on hover

Myranae:
## Summary:
Previously, Polygon graphs had a blue fill by default. This changes them so they are transparent by default and only fill with Mafs blue on hover. 

### Before
https://github.com/user-attachments/assets/c72d26fd-8939-41a0-9b2f-b133655fe1c4

### After
https://github.com/user-attachments/assets/7d4733d4-9594-438d-be58-c9e301dc63d2

Issue: LEMS-2099

## Test plan:
- Confirm all checks and tests pass
- Run yarn dev. Open http://localhost:5173/. Turn on the flag for Polygon. 
- Confirm: 
--- Polygon graphs fill on hover. 
--- The fill stays when the graph is being dragged and goes away after the mouse leaves the shape.


packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx
```diff
@@ -103,11 +104,15 @@ export const PolygonGraph = (props: Props) => {
...
                         cursor: dragging ? "grabbing" : "grab",
+                        fill: hovered ? "var(--mafs-blue)" : "transparent",
                     },
                     onFocus: () => setFocused(true),
                     onBlur: () => setFocused(false),
                     onMouseEnter: () => setHovered(true),
-                    onMouseLeave: () => setHovered(false),
+                    onMouseLeave: () => {
+                        setHovered(false);
+                        setFocused(false);
```
benchristel, line 114: Why do we `setFocused(false)` here? If the user has focused the polygon with the keyboard, we don't want mousing over it and then away from it to make it appear unfocused.

benchristel (commented):
A thought: since we are managing the focus/hover styling with React state instead of CSS, it would be possible to pull out a stateless component that takes `active` as a prop. That would allow us to add visual snapshot tests for the active and inactive state. I don't think that needs to block this PR, but it might be a nice addition.

packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx
Myranae: When testing, if I move the polygon, the lines stay weighted even after moving the shape. This isn't how other Mafs graphs act, so I was trying to get it to act the same. First I made it so the weighting only happens when dragging and on hover and not on focus, but that wasn't quite right. Looks like this isn't quite right either 🤔 Not sure how to fix that, but I can try to brainstorm another fix tomorrow too. Or just make a new ticket.

benchristel: Oh, that's because we're managing the focus styling ourselves, based on `onFocus`. The issue is that an element can be _focused_ but not have _visible focus_, which is the state where the browser applies focus styling to it by default. The behavior you're seeing happens because whenever you click an element, it gets _focus_, and `onFocus` is called. However, an element generally doesn't get _visible focus_ unless you select it with the keyboard. I say "generally" because I seem to recall that browsers have a lot of complicated logic around exactly when to show focus. [Docs.](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible)

I think in order to get this to work the way we want, we would either need to:

- manage the hover/focus visuals entirely in CSS, using `:hover` and `:focus-visible` pseudo-classes.
- detect `:focus-visible` by holding a ref to the polygon element and, when `onFocus` is called, do something like this:
  ```ts
  if (polygonRef.current.matches(":focus-visible")) {
    setFocused(true);
  }
  ```
  Ideally, there would be a separate event callback we could register like `onFocusVisible`, but I don't think that exists.

catandthemachines (commented):
I'm curious, what does the keyboard focus look like with this addition? I see a lot of fill behavior on hover and it seems like we should also be applying it based on focus-visible. <3

packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx
```diff
@@ -51,9 +51,7 @@ export const PolygonGraph = (props: Props) => {
                 points={[...points]}
                 color="var(--movable-line-stroke-color)"
                 svgPolygonProps={{
-                    strokeWidth: active
-                        ? "var(--movable-line-stroke-weight-active)"
-                        : "var(--movable-line-stroke-weight)",
```
Myranae, line 56: After discussions with our designer, we realized that other similar graphs do not have their lines become weighted when active. The line related graphs (linear, linear system, etc) become weighted because you can move the line by dragging it. Polygon and other graphs are dragged by clicking anywhere in the area filled with blue. When focusing any graph, the lines do not become weighted, just outlined.

packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx
```diff
@@ -51,9 +51,7 @@ export const PolygonGraph = (props: Props) => {
...
+                    style: {fill: "transparent"},
```
Myranae, line 54: Setting the color on line 52 starts the graph out with a colored background. Removing this line turns the graph gray though. If we leave line 52 and just change the fill here, it works as needed.

packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx
Myranae: After chatting with Caitlyn, we realized that polygon should not actually have its lines weighted in any way, because that is the same behavior as angle, sinusoid, and circle (which are more similar to this graph than the linear ones). The area that is interactive for the user is the center, upon which the user can grab and drag. So the lines themselves are not what are interactive. Updated the code to reflect this.

Myranae:
> I'm curious, what does the keyboard focus look like with this addition? I see a lot of fill behavior on hover and it seems like we should also be applying it based on focus-visible. <3

Thanks for pointing this out! I don't think any of the graphs fill on keyboard focus, but an outline goes around the entire shape indicating it has keyboard focus. I think it might be a larger discussion if we want to also add fill for this. Maybe something to consider when we get to the keyboard part of this project?

benchristel:
I agree with @catandthemachines that we need some kind of focus styling now that the line thickness doesn't change on focus. Currently there is no visual indicator that anything is focused if you focus the whole polygon.

Myranae:
Ooo, yes, it needs to have the same outline as circle. Will add that!

Myranae:
@catandthemachines @benchristel I'm moving the line work into another PR. Want to get the fill on hover work completed while I explore how to get the outline working. Removed the code related to the outline, so should just be for the hover now.

benchristel (approved):
LGTM!

PR #1439, opened 2024-07-23
Removing tests: trying to get to the root problem

jeremywiebe:
## Summary:
<Add your summary here>

Issue: XXX-XXXX

## Test plan:


jeremywiebe:
We figured it out in #1445

PR #1440, opened 2024-07-24
Fix flaky tests (following React 18/RTL upgrade)

jeremywiebe:
## Summary:

Matthew noticed that these tests fail sometimes in CI. They also fail sporadically (especially, it seems, on older/slower laptops). 

Extending the timeout for the final `waitFor()` calls seems to fix them and give the graph enough time to process the actions that are dispatched by the keypresses. 

Issue: LEMS-1802 

## Test plan:

All tests pass. 
All tests pass in CI. 
All tests pass in CI again.

### DIFF
--- a/packages/perseus/src/widgets/__tests__/interactive-graph.test.tsx
+++ b/packages/perseus/src/widgets/__tests__/interactive-graph.test.tsx
@@ -224,7 +224,9 @@ describe("a mafs graph", () => {
                 renderQuestion(question, apiOptions);
             });
 
-            it("rejects incorrect answer", async () => {
+            // TODO(Jeremy): This test is currently skipped because it is
+            // failing sporadically after the React 18 upgrade.
+            xit("rejects incorrect answer", async () => {
                 // Arrange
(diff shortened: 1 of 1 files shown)


benchristel (commented):
I don't love that we're skipping these tests, because they're now the only tests we have for the interactive behavior of InteractiveGraph.

Could we increase the [`waitFor` timeout](https://testing-library.com/docs/dom-testing-library/api-async/#waitfor) instead? I suspect the tests are "flaky" just because the code is slow.

jeremywiebe:
> I don't love that we're skipping these tests, because they're now the only tests we have for the interactive behavior of InteractiveGraph.
> 
> Could we increase the [`waitFor` timeout](https://testing-library.com/docs/dom-testing-library/api-async/#waitfor) instead? I suspect the tests are "flaky" just because the code is slow. According to the docs, the default timeout is 1 second.

That's a good point.  I'll try that.

benchristel (approved):
🤞 LGTM! Thanks.

PR #1441, opened 2024-07-24
Turn off interactivity when Interactive Graph in hint mode

Myranae:
## Summary:
Hint mode allows content creators to copy graphs directly from content, paste them into a hint, and have it automatically detect that the graph should be rendered as a hint. This work removes interactivity from the graph once it has been rendered as a hint.

(Need to confirm that adding a new graph into the hint lets you modify the graph.)

Issue: LEMS-2054

## Test plan:
- Run `yarn storybook` after checking out the branch associated with this PR
- Confirm the Hints Renderer graphs are not interactive (one story showing all graphs and also one story for each type): http://localhost:6006/?path=/story/perseus-renderers-hints-renderer--with-all-interactive-graphs

For further testing:
- Go to the Editor Page Demo: http://localhost:6006/?path=/story/perseuseditor-editorpage--demo
- Add an interactive graph widget to the Question box (try linear for now since it has start coords)
- Copy and paste that into the Hint input box
- Confirm with the preview that the hint graph is not interactive with the mouse or the keyboard and that the start coords change how it looks
- In the same way, confirm the other graph types are not interactive with either the mouse or the keyboard


packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -125,7 +126,7 @@ export const MafsGraph = (props: MafsGraphProps) => {
                                 range={state.range}
                             />
                         )}
-                        /* Protractor */
+                        {/* Protractor */}
```
Myranae, line 129: Just noticed this was missing the brackets, so added them in since this seems to be a comment.

benchristel (changes requested):
The code looks great! But the graph elements can still be moved with the keyboard when hint mode is on. I think we should fix that before landing this change.

One possible approach might be:

- add a `hintMode` field to the [`GraphConfig` context](https://github.com/Khan/perseus/blob/7a448e77c18d9c8437c24e6567f1e4fa03efc6b9/packages/perseus/src/widgets/interactive-graphs/reducer/use-graph-config.ts#L5) and set it in [MafsGraph](https://github.com/Khan/perseus/blob/7a448e77c18d9c8437c24e6567f1e4fa03efc6b9/packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx#L58-L67). This will allow the individual components to change their behavior when in hintMode without any prop drilling.
- set `tabIndex` to `-1` on all graph elements where it's currently set to `0`, when hint mode is on, to prevent tabbing to those elements.

A possible complication: components defined by Mafs might have tab indexes, and those components don't know about hint mode. Let's cross that bridge if/when we get to it, though.

Myranae:
@benchristel I was able to use your suggestion to make the hint mode graphs non-interactive via keyboard too. I set up some more stories to help test this as well.

benchristel (approved):
Looks great! Thanks for the updates.

PR #1444, opened 2024-07-24
Throttle events in polygon

nicolecomputer:
## Summary:

Throttles the events in polygon to increase performance and decrease CPU load. This is especially important for Safari. This change lowers CPU usage from 87% to 30% on a M3 macbook. The increase in performance will be more on phones and iPads where this will also decrease energy usage.

40 FPS was chosen as a mark that feels response and good to use but doesn't put too much stress on the machine.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2185

## Test plan:

To test before:
- Load the dev environment on Safari
- Open the Timelines pane and start recording
- The CPU usage should be high (it'll vary based on your machine)

To test after:
- Repeat test above with this change
- **The CPU usage should be lower**


packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx
```diff
@@ -42,6 +42,7 @@ export const PolygonGraph = (props: Props) => {
     });
 
     const active = hovered || focused || dragging;
+    const [lastMoveTime, setLastMoveTime] = React.useState<number>(0);
```
benchristel, line 45: Could we use a ref instead of state here? That would prevent unnecessary re-renders. I believe a ref is appropriate because we will never need to update DOM elements when `lastMoveTime` changes.

packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx
```diff
@@ -116,9 +117,16 @@ export const PolygonGraph = (props: Props) => {
                     key={"point-" + i}
                     snapTo={snapTo}
                     point={point}
-                    onMove={(destination: vec.Vector2) =>
-                        dispatch(actions.polygon.movePoint(i, destination))
-                    }
+                    onMove={(destination: vec.Vector2) => {
+                        const now = Date.now();
+                        const targetFPS = 40;
```
benchristel, line 122: What's the rationale for capping FPS at 40? For displays that refresh at 60Hz (or browsers that repaint at 60Hz?), this will result in every third frame getting dropped. I guess my question is, why not 30 fps?

nicolecomputer: Loosely held opinion: 40 fps feels about right. I've noticed in games it's where things start to feel smooth (which is totally subjective). Some data, though: If I cap at 40fps I get around 26% CPU usage. If I cap at 60 I get around 40%.

packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx
nicolecomputer: Yepppppp. Your absolutely right.

benchristel (approved):
🎉 LGTM!

PR #1445, opened 2024-07-24
Fix use-draggable hook to ignore 'keyup' events

jeremywiebe:
## Summary:

While @nicolecomputer and I were investigating why the `interactive-graph.test.tsx` tests _sometimes_ failed in CI and other slower environments, we found that the `useDraggable()` hook was handling both `keydown` as well as `keyup` events.

Through some testing we discovered also that the browser sends only a `keydown` event for a single keypress. When the key is held down, we saw a series of `keydown` events and then a single `keyup` event when the key was released. 

To that end, we've adjusted `useDraggable` to ignore all `keyup` events.

Issue: LEMS-1820, LEMS-1934

## Test plan:

`yarn test` all tests pass
In storybook, you can notice that there's one fewer moves when selecting a point and then holding down the arrow key to move it. Previously, there was a final move because of the `keyup` processing. Now that's gone and the effect "feels" more correct.

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/use-draggable.test.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/use-draggable.test.tsx
@@ -216,7 +216,7 @@ describe("useDraggable", () => {
         expect(screen.getByRole("button")).toHaveFocus();
 
         // Act:
-        await userEvent.keyboard("{arrowright>1}{arrowup>1}");
+        await userEvent.keyboard("{arrowright}{arrowup}");
 
         // Assert: the element moved one step to the right and then one step up
         expect(onMoveSpy.mock.calls).toEqual([[[1, 0]], [[0, 1]]]);
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/use-draggable.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/use-draggable.ts
@@ -61,6 +61,17 @@ export function useDraggable(args: Params): DragState {
             const isKeyboard = type.includes("key");
             if (isKeyboard) {
                 event?.preventDefault();
+
+                // When a key is held down, we see multiple "keydown" events,
+                // followed by a single "keyup" event.
+                // For a single keypress, we only see a "keydown" event, no
+                // "keyup".
+                // We never want to process the keyup event as an intent to
+                // move so we bail on further processing here.
+                if (type === "keyup") {
+                    return;
+                }
+
                 const {
                     direction: yDownDirection,
                     altKey,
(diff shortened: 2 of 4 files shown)


packages/perseus/src/widgets/interactive-graphs/graphs/use-draggable.test.tsx
```diff
@@ -216,7 +216,7 @@ describe("useDraggable", () => {
         expect(screen.getByRole("button")).toHaveFocus();
 
         // Act:
-        await userEvent.keyboard("{arrowright>1}{arrowup>1}");
```
jeremywiebe, line 219: The `>1` construct within curly braces `{}` causes the `userEvent` library to not lift the key. So this test is pressing the right arrow, holding it down, and pressing the up arrow. It ends with both keys down. 

See: https://testing-library.com/docs/user-event/keyboard/

packages/perseus/src/widgets/interactive-graphs/graphs/use-draggable.ts
```diff
@@ -61,6 +61,17 @@ export function useDraggable(args: Params): DragState {
...
+
+                // When a key is held down, we see multiple "keydown" events,
+                // followed by a single "keyup" event.
+                // For a single keypress, we only see a "keydown" event, no
+                // "keyup".
+                // We never want to process the keyup event as an intent to
+                // move so we bail on further processing here.
+                if (type === "keyup") {
+                    return;
+                }
```
benchristel, line 73: I bet this `keyup` weirdness was the cause of [this mysterious bug](https://khanacademy.atlassian.net/browse/LEMS-1934). I believe @nishasy has the hardware to test whether the bug is fixed or not 🙏 

The same bug probably exists in upstream Mafs, so maybe we should submit the fix there.

benchristel (approved):
Wow, this is a great find and fix. LGTM! 🎉 👏

packages/perseus/src/widgets/interactive-graphs/graphs/use-draggable.ts
jeremywiebe: I tested with an iPhone 13 and a Bluetooth Magic Keyboard. It seems to be fixed. I've asked SarahP to retest. If it works, I'll mark that bug as fixed! :) 

I'll also open a PR against Mafs as you suggest.

benchristel: @jeremywiebe Does the bug occur in prod/main with your keyboard setup?

jeremywiebe: Yes. Here's video where I'm pressing the arrow key once every few seconds. You can see the double-movement each time. 

https://github.com/user-attachments/assets/c5e433b5-0f3d-41fa-98ed-8ef251985380

PR #1446, opened 2024-07-24
Rename StartCoordSettings to StartCoordsSettings

nishasy:
## Summary:
Nit: It should be StartCoordsSettings, not StartCoordSettings, because
the field is called startCoords.

Issue: none

## Test plan:
`yarn jest`


jeremywiebe (approved):
🤓 I very much appreciate the attention to detail!

Approved

PR #1448, opened 2024-07-25
[Interactive Graph Editor] Refactor and clean up start coords UI implementation

nishasy:
## Summary:
The StartCoordsSettings file was getting unruly and hard to follow. I'm refactoring
it here to make it easier to add more types, and to make it generally more
maintainable.

- Broke out the main UI input section for each graph type into its own file
- Removed the unnecessary switch statement (only need to pass in `undefined` to
  reset to default coords)
- Updated the button to "Use default start coordinates" instead of "Use default
  start coords" as per Sarah Third's recommendation
- Updated the labels from "x/y coords" to just "x/y" to save space and look
  cleaner. (I did get [permission from the content authors](https://khanacademy.slack.com/archives/C04EA1QAKN0/p1721860247138389) for this.)
- Removed all the checks for the graph type. The feature is behind a feature
  flag, so no one should run into the start coords for an unimplemented
  graph type in prod or anything
- Updated tests accordingly

Issue: https://khanacademy.atlassian.net/browse/LEMS-2052

## Test plan:
`yarn jest packages/perseus-editor/src/components/__tests__/start-coords-settings.test.tsx`

Storybook

## Screenshots

| Before | After |
| --- | --- |
| Screenshot 2024-07-24 at 5 28 48 PM | Screenshot 2024-07-24 at 5 28 55 PM |
| Screenshot 2024-07-24 at 5 29 14 PM | Screenshot 2024-07-24 at 5 29 19 PM |

### DIFF
--- a/packages/perseus-editor/src/widgets/__tests__/interactive-graph-editor-locked-figures.test.tsx
+++ b/packages/perseus-editor/src/widgets/__tests__/interactive-graph-editor-locked-figures.test.tsx
@@ -54,6 +54,22 @@ describe("InteractiveGraphEditor locked figures", () => {
         );
     });
 
+    const assertAndCloseStartCoordsSection = async () => {
+        // Confirm the "Start coordinates" section is open.
+        const resetCoordsButton = screen.getByRole("button", {
+            name: "Use default start coordinates",
+        });
+        expect(resetCoordsButton).toBeInTheDocument();
+
+        // Close the "Start coordinates" section first to remove
+        // other inputs from the DOM.
+        const startCoordinatesHeading = screen.getByRole("button", {
+            name: "Start coordinates",
+        });
+
+        await userEvent.click(startCoordinatesHeading);
+    };
+
     // Basic functionality
     describe.each`
         figureType   | figureName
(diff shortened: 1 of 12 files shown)


packages/perseus-editor/src/widgets/__tests__/interactive-graph-editor-locked-figures.test.tsx
```diff
@@ -1202,6 +1202,13 @@ describe("InteractiveGraphEditor locked figures", () => {
                 lockedFigures: [defaultPolygon],
             });
 
+            // Close the "Start coordinates" section first to remove
+            // other accordions from the DOM.
+            const startCoordinatesHeading = screen.getByRole("button", {
+                name: "Start coordinates",
+            });
+            await userEvent.click(startCoordinatesHeading);
```
benchristel, line 1210: I just noticed that the code doesn't quite match the comment here: we're clicking on the heading, which I assume might either expand or collapse the section depending on its current state.

A few ideas for fixing this:

- Could we assert that the section is expanded before clicking the heading?
- Could we assert that the section is collapsed after clicking the heading?
- Could we avoid the problem by testing a smaller component (i.e. only rendering the locked figures section)?

benchristel (approved):
Looks good! I noted one non-blocking concern inline.

packages/perseus-editor/src/widgets/__tests__/interactive-graph-editor-locked-figures.test.tsx
nishasy: Ah, I did it that way because it's always open by default. I can add an assertion to make it more clear though.

I think some smaller component testing is there in `locked-figures-section.test.tsx`. I remember doing it this way specifically to test it within the larger context of the editor page, because I wanted to test the editor's `onChange` function.

PR #1449, opened 2024-07-25
KAS: try to type some things

handeyeco:
## Summary:
Functionality shouldn't change, just trying to modernize some things.


packages/kas/package.json
```diff
@@ -22,7 +22,7 @@
         "dist"
     ],
     "scripts": {
-        "gen:parsers": "node src/parser-generator.js",
+        "gen:parsers": "node src/parser-generator.ts",
```
jeremywiebe, line 25: Does this work (passing a TypeScript file to node directly)?

Previously we had to use SWC to run TS scripts as command line apps: https://github.com/Khan/perseus/blob/main/utils/pre-publish-check-ci.ts#L1

packages/kas/src/compare.ts
```diff
@@ -1,12 +1,15 @@
-/* eslint-disable */
-import _ from "underscore";
```
jeremywiebe, line 2: Yay!

packages/kas/src/compare.ts
```diff
@@ -16,18 +19,17 @@ export const compare = function (expr1, expr2, options) {
      */
 
     if (options !== undefined) {
-        // eslint-disable-next-line no-undef
-        options = _.extend(defaults, options);
+        options = {...defaults, ...options};
     } else {
         options = defaults;
     }
```
jeremywiebe, line 25: FWIW, with spread syntax, we can just spread the `options` (even if they're `undefined`).

image

mark-fitzgerald (approved):
Thanks for the clean-up!

packages/kas/package.json
handeyeco: It seems to work when I run it 🤷

PR #1453, opened 2024-07-25
[Interactive Graph Editor] Implement the UI for adding start coords for circle graphs

nishasy:
## Summary:
Add the UI to specify start coords for Circle graph type.

- Add the circle graph type to `start-coor-settings.tsx`
- Create a `start-coords-circle.tsx` file with the main UI
- Update the circle graph type so that `startCoords` includes both
  the center and the radius.
- Update tests

Issue: https://khanacademy.atlassian.net/browse/LEMS-2206

## Test plan:
`yarn jest`

Storybook
- Go to http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-circle-with-starting-coords
- Confirm that the preview shows the circle at the already set coords + radius
- Confirm that the UI reflects these coords + radius
- Press "Use default start coordinates"
- Confirm that the circle goes to [0, 0] with a radius of 2.

- Go to http://localhost:6006/?path=/story/perseuseditor-editorpage--demo
- Create an interactive graph widget
- Change the graph to circle type
- Confirm that the start coords UI shows the center at [0, 0] with radius 2

Screenshot 2024-07-25 at 4 29 34 PM


packages/perseus/src/perseus-types.ts
```diff
@@ -789,7 +789,10 @@ export type PerseusGraphTypeCircle = {
     center?: Coord;
     radius?: number;
     // The initial coordinates the graph renders with.
-    startCoords?: Coord;
+    startCoords?: {
+        center: Coord;
+        radius: number;
+    };
```
nishasy, line 795: ADMITTEDLY, this is a bit janky. I tried to split them out into separate fields too, or use `radiusPoint` instead of radius, but those implementations got really tricky.

If anyone has any opinions or suggestions, I'd love to hear them!

benchristel: What's janky about it? It looks fine to me.

benchristel (approved):
I left one comment inline about maybe avoiding the dependency on `mafs`, but it's your call. LGTM!

packages/perseus/src/perseus-types.ts
nishasy: Ah, I just didn't like that all the other `startCoord` types are arrays, but this one is an object. It also means the type is not as clean/streamlined when using it in other places - although I suppose I can make it its own type and import it if that were really an issue.

PR #1455, opened 2024-07-26
Update Polygon graphs to have weighted lines only after focus from keyboard

Myranae:
## Summary:
Previously, Polygon had its shape weighted upon hover and after clicking in the shape. This is not how circle is currently working. To bring them into alignment, this changes Polygon to become weighted only when the shape is tabbed to by the keyboard or if the shape is moved via keyboard after the user clicks on it. 

### Before
https://github.com/user-attachments/assets/16f3aef8-3312-4e0c-8815-542ad6ea7c98

### After
https://github.com/user-attachments/assets/b6058ae6-b31e-4942-ae24-9ae2853bfcc3

Issue: LEMS-2200

## Test plan:
- Checkout the associated branch
- Run `yarn dev` 
- On the gallery page, turn on the flag for polygon and circle
- Confirm Polygon does not have weighted lines on hover or when dragging. After clicking the shape, confirm the shape becomes weighted when you move it with the arrow keys. Confirm the lines return to normal after clicking outside of the shape.
- Confirm this functionality matches Circle graphs


benchristel (commented):
I see that it works, but I don't fully understand _how_ it works. It seems like this relies on the order in which events fire? (`onMouseDown`, `onFocus`, `onMouseUp`)?

I prototyped an alternative implementation that I think covers all the same use cases: https://github.com/Khan/perseus/pull/1458. It uses `:focus-visible`, so it will use the the browser's logic for deciding when focus should be shown. What do you think about this approach?

Myranae:
Thanks, @benchristel ! I'm not so comfortable with refs here, but that is exactly what I needed! I don't think I would have known about jsdom not supporting `:focus-visible` even if I had used a ref, so thanks for sharing that try/catch block in the function you shared. Going to try to add in a test before finalizing this, unless you don't think that's necessary.

Myranae:
@benchristel I was able to use your idea, but I had to add some more logic around user actions to get it to do everything we need it to do. Would you mind taking another look and letting me know what you think? It seems like the main issue might be getting the line weighting / shape to re-render when certain changes occur (like blur or tabbing). I used state to do that, but would love another suggestion more in-line with your alternative approach.

benchristel (approved):
Looks great! I think I like this better than my original approach.

PR #1456, opened 2024-07-29
Remove 'Show ruler' option from interactive graph editor

benchristel:
The ruler is not used in published content. We haven't implemented it for
the new Mafs version of the InteractiveGraph widget, so there's no point
having a checkbox for it in the editor.

Issue: LEMS-2096

## Test plan:

View an interactive graph exercise in the exercise editor:
`/devadmin/content/exercises/linear-graph-exercise/x3239cabc044aa28b/x325a92c13bc37711`

There should be no option to show the ruler in the graph settings.

Screen Shot 2024-07-29 at 2 35 35 PM


packages/perseus/src/widgets/interactive-graphs/interactive-graph-question-builder.ts
```diff
@@ -64,10 +64,7 @@ class InteractiveGraphQuestionBuilder {
                         labels: this.labels,
                         markings: this.markings,
                         range: [this.xRange, this.yRange],
-                        rulerLabel: "",
```
nicolecomputer, line 67: This builder pattern is lovely because we don't have to go find out every place that we have test data and remove these. 💪

nicolecomputer (approved):
This looks great 🚢

mark-fitzgerald (approved):
Code cleansing is so refreshing!

PR #1457, opened 2024-07-29
Add static mode to interactive graph

nicolecomputer:
## Summary:
Adds static mode and a helper to the builder function to mark a graph as static.

New button in content editor:
Screenshot 2024-07-29 at 2 41 59 PM

Grading:
Screenshot 2024-07-29 at 2 31 29 PM
Screenshot 2024-07-29 at 2 31 05 PM

Issue: LC-2019

## Test plan:
- Visit storybook
- There should be two new stories

- **In "Static Graph" you should not be able to interact with the graph**
- **In "Static Graph With Another Widget" you should be graded only on the radio widget**

- Go to the editor demo
- **You should be able to toggle static"**

### DIFF
--- a/.changeset/gentle-dragons-hide.md
+++ b/.changeset/gentle-dragons-hide.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": minor
+---
+
+Add static mode to interactive graph
--- a/packages/perseus/src/widgets/__stories__/interactive-graph.stories.tsx
+++ b/packages/perseus/src/widgets/__stories__/interactive-graph.stories.tsx
@@ -20,6 +20,8 @@ import {
     segmentWithLockedEllipses,
     segmentWithLockedVectors,
     segmentWithLockedPolygons,
+    staticGraph,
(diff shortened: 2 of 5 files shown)


nicolecomputer:
@jeremywiebe Is using identity for the transform ok here?

jeremywiebe (commented):
I feel like there's a piece missing yet.  I don't see any changes to the interactive graph widget itself to make the widget static. 

Reading the Jira ticket, I think it's talking about two unique concepts: static mode and a mode with no interactive-elements. 

Traditionally, in Perseus, `static: true` just locks the widget so that the learner can't interact with it. This is used when the learner has completed the question but has not yet moved on to the next one. They can see the answer but cannot change anything. 

The other mode mentioned is where we want to display an `interactive-graph` but have no interactive elements on it (ie. it has no `type`). 

I think that `static` mode is necessary for the second feature, but it does not cover it entirely (we'll need a `type: "none"` or something for the `interactive-graph`.

nicolecomputer:
This was superseded by other changes.

PR #1458, opened 2024-07-29
Use :focus-visible to show focus indictor on polygon

benchristel:
I felt like suggesting these changes inline on https://github.com/Khan/perseus/pull/1455 would be too cumbersome, so I'm doing it in a PR instead.


packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.tsx
```diff
@@ -16,7 +16,6 @@ interface PolygonAngleProps {
     centerPoint: vec.Vector2;
     endPoints: [vec.Vector2, vec.Vector2];
     polygonLines: readonly CollinearTuple[];
-    active: boolean;
```
benchristel, line 19: This prop wasn't used. Removing it so we can remove the logic to compute its value in `polygon.tsx`.

benchristel:
Tamara incorporated these changes into https://github.com/Khan/perseus/pull/1455 manually, so there's no need for this PR to stay open. Closing.

PR #1459, opened 2024-07-29
Guard against matchMedia access during initial render

somewhatabstract:
## Summary:
This PR makes sure that we don't call `matchMedia` during the initial render. This is because `matchMedia` is not available during the initial render, and calling it will throw an error.

Issue: FEI-5743

## Test plan:
1. Put up this PR and await the pre-release packages
2. Update webapp with the pre-release packages
3. Put up a webapp PR and check that the ZND is working on URLs that were previously erroring

### DIFF
--- a/packages/perseus/src/widgets/explanation.tsx
+++ b/packages/perseus/src/widgets/explanation.tsx
@@ -66,6 +66,15 @@ class Explanation extends React.Component<Props, State> {
     state: State = {
         expanded: false,
     };
+    _mounted: boolean = false;
+
+    componentDidMount() {
+        this._mounted = true;
+    }
+
+    componentWillUnmount() {
+        this._mounted = false;
+    }
 
     change: (arg1: any, arg2: any, arg3: any) => any = (...args) => {
         return Changeable.change.apply(this, args);
@@ -85,9 +94,9 @@ class Explanation extends React.Component<Props, State> {
 
         const caretIcon = this.state.expanded ? caretUp : caretDown;
 
-        const allowTransition = mediaQueryIsMatched(
-            "(prefers-reduced-motion: no-preference)",
-        );
+        const allowTransition =
+            this._mounted &&
+            mediaQueryIsMatched("(prefers-reduced-motion: no-preference)");
 
         // Special styling is needed to fit the button in a block of text without throwing off the line spacing.
         // While the button is not normally included in a block of text, it needs to be able to accommodate such a case.
--- a/packages/perseus/src/widgets/label-image.tsx
+++ b/packages/perseus/src/widgets/label-image.tsx
@@ -109,6 +109,7 @@ export class LabelImage extends React.Component<
 
     // The rendered markers on the question image for labeling.
     _markers: Array<Marker | null | undefined>;
+    _mounted: boolean = false;
 
     static gradeMarker(marker: InteractiveMarkerType): InteractiveMarkerScore {
         const score = {
@@ -375,6 +376,14 @@ export class LabelImage extends React.Component<
         };
     }
 
+    componentDidMount() {
+        this._mounted = true;
+    }
+
+    componentWillUnmount() {
+        this._mounted = false;
+    }
+
     simpleValidate(rubric: LabelImageProps): PerseusScore {
         return LabelImage.validate(this.getUserInput(), rubric);
     }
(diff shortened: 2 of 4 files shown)


packages/perseus/src/widgets/explanation.tsx
```diff
@@ -85,9 +94,9 @@ class Explanation extends React.Component<Props, State> {
 
         const caretIcon = this.state.expanded ? caretUp : caretDown;
 
-        const allowTransition = mediaQueryIsMatched(
-            "(prefers-reduced-motion: no-preference)",
-        );
+        const allowTransition =
+            this._mounted &&
+            mediaQueryIsMatched("(prefers-reduced-motion: no-preference)");
```
somewhatabstract, line 99: **note:** We don't want animation on by default on initial render in case the user has turned it off (long term, it would be useful if this also could be overridden by webapp profile setting too).

packages/perseus/src/widgets/label-image/marker.tsx
```diff
@@ -106,7 +119,7 @@ export default class Marker extends React.Component<Props> {
         } else if (showPulsate) {
             iconStyles = [
                 styles.markerPulsateBase,
-                shouldReduceMotion()
+                this._mounted && shouldReduceMotion()
```
somewhatabstract, line 122: **note:** We don't want animation on by default on initial render in case the user has turned it off (long term, it would be useful if this also could be overridden by webapp profile setting too).

packages/perseus/src/widgets/label-image.tsx
```diff
@@ -503,17 +512,17 @@ export class LabelImage extends React.Component<
...
+        const isNarrowPage =
+            this._mounted &&
+            window.matchMedia(mediaQueries.xsOrSmaller.replace("@media ", ""))
+                .matches;
 
-            // Determine whether the image is wider than it is tall.
-            const isWideImage =
-                this.props.imageWidth / 2 > this.props.imageHeight;
+        // Determine whether the image is wider than it is tall.
+        const isWideImage = this.props.imageWidth / 2 > this.props.imageHeight;
```
somewhatabstract, line 522: **note:** Moved out of the loop; doesn't need recalculating multiple times.

packages/perseus/src/widgets/label-image.tsx
```diff
@@ -503,17 +512,17 @@ export class LabelImage extends React.Component<
...
-        return markers.map((marker, index): React.ReactElement => {
-            // Determine whether page is rendered in a narrow browser window.
-            const isNarrowPage = window.matchMedia(
-                mediaQueries.xsOrSmaller.replace("@media ", ""),
-            ).matches;
+        // Determine whether page is rendered in a narrow browser window.
+        const isNarrowPage =
+            this._mounted &&
+            window.matchMedia(mediaQueries.xsOrSmaller.replace("@media ", ""))
+                .matches;
```
somewhatabstract, line 519: **note:** Moved out of the loop since it won't change for each item, and modified to return `false` on initial render so that it will SSR and hydrate correctly.

jeresig (approved):
These changes all seem reasonable to me, thank you for figuring this out, Jeff! I'll defer to the Perseus team for final approval here, though.

packages/perseus/src/widgets/label-image.tsx
jeremywiebe: Thanks for fixing this up!

packages/perseus/src/widgets/explanation.tsx
jeremywiebe: I'm confused how this fixes things.  The `mediaQueryIsMatched` function guards against `window.matchMedia` not being a function. Is that helper written wrong or is the fix more about only running on a mounted component?

packages/perseus/src/widgets/label-image.tsx
```diff
@@ -503,17 +512,17 @@ export class LabelImage extends React.Component<
...
-        // Render all markers for widget.
-        return markers.map((marker, index): React.ReactElement => {
-            // Determine whether page is rendered in a narrow browser window.
-            const isNarrowPage = window.matchMedia(
-                mediaQueries.xsOrSmaller.replace("@media ", ""),
-            ).matches;
+        // Determine whether page is rendered in a narrow browser window.
+        const isNarrowPage =
+            this._mounted &&
+            window.matchMedia(mediaQueries.xsOrSmaller.replace("@media ", ""))
```
jeremywiebe, line 518: This usage _doesn't_ use a helper to guard `window.matchMedia` being undefined. I wonder if we should move all these checks to call a helper that guards usage properly. Would that alleviate us needing the `_mounted` flag in each component?

jeremywiebe (approved):
The fixes mostly make sense. I left one question and I think it would make sense to move the `mediaQueryIsMatched` helper to a new file in `packages/perseus/src/util`. Not a deal-breaker, but if you decide not to please let me know and I'll create a ticket to do this after the fact.

Thanks for chasing this down Jeff.

packages/perseus/src/widgets/explanation.tsx
somewhatabstract: This is a subtle thing, but we don't just want to guard against it being accessed; the initial render on both server and client must be identical, so we don't want to use `matchMedia` at all on first render as the client could give a different rendered result than the server.

packages/perseus/src/widgets/label-image.tsx
somewhatabstract: Guarding against it being undefined is really a red herring; I think perhaps that check should be removed where we have it. I don't think we support any browsers that won't have it at this point, and we shouldn't be determining by that measure if we call it or not on initial render as we need that to match on both client (where it exists) and server (where it doesn't exist or might give a different response than on the client).

PR #1461, opened 2024-07-30
Reverting portal disable functionality, causing clipping behavior.

catandthemachines:
## Summary:
Adding back portal functionality for Expression Widget to ensure it does not get clipped.

This implementation was originally added in this PR: https://github.com/Khan/perseus/pull/1424

Issue: LEMS-2224

## Test plan:
Go to https://khan.github.io/perseus/?path=/story/perseus-widgets-expression--desktop-kitchen-sink
Confirm that you have to tab through all the tab pannels to get the the main keypad, instead of going directly to the keypad.

### DIFF
--- a/packages/math-input/src/components/tabbar/item.tsx
+++ b/packages/math-input/src/components/tabbar/item.tsx
@@ -89,10 +89,26 @@ function TabbarItem(props: TabItemProps): React.ReactElement {
     const tabRef = useRef<{focus: () => void}>(null);
 
     useEffect(() => {
+        let timeout;
         if (role === "tab" && focus) {
-            // Move element into view when it is focused
-            tabRef?.current?.focus();
+            /**
+             * When tabs are within a WonderBlocks Popover component, the
+             * manner in which the component is rendered and moved causes
+             * focus to snap to the bottom of the page on first focus.
+             *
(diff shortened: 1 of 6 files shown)


packages/math-input/src/components/tabbar/item.tsx
```diff
@@ -89,10 +89,26 @@ function TabbarItem(props: TabItemProps): React.ReactElement {
...
+            /**
+             * When tabs are within a WonderBlocks Popover component, the
+             * manner in which the component is rendered and moved causes
+             * focus to snap to the bottom of the page on first focus.
+             *
+             * This timeout moves around that by delaying the focus enough
+             * to wait for the WonderBlock Popover to move to the correct
+             * location and scroll the user to the correct location.
+             * */
+            timeout = setTimeout(() => {
```
handeyeco, line 103: This is part of the revert?

handeyeco (approved):
Thanks Cat!

packages/math-input/src/components/tabbar/item.tsx
catandthemachines: Yes, I removed it because without the portal it was no longer necessary. Now that we're doing portaling again we need the timeout behavior.

PR #1463, opened 2024-07-30
[Interactive Graph Editor - Stories] Fix broken default and controlled stories

nishasy:
## Summary:
The stories are broken because of a missing null check.

Adding that here.

Issue: none

## Test plan:
`yarn start`
Check that the following stories work
- http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph-editor--default
- http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph-editor--controlled


jeremywiebe (approved):
Thanks for fixing! :)

PR #1465, opened 2024-07-30
[Hint Mode: Start Coords] Add separate flags for graph types

nishasy:
## Summary:
We want to be able to release the start coords UI even if
only a subset of graph types are ready. To accomodate this,
I'm adding individual flags for each graph type.

Issue: none

## Test plan:
Storybook
- http://localhost:6006/?path=/docs/perseuseditor-widgets-interactive-graph--docs
- Confirm that the UI shows up for the currently landed graphs - segment, ray, linear, linear-system, circle
- Confirm that the UI does not show up for the graphs that are not done yet - sinusoid, quadratic, angle, polygon, point

### DIFF
--- a/packages/perseus-editor/src/widgets/interactive-graph-editor.tsx
+++ b/packages/perseus-editor/src/widgets/interactive-graph-editor.tsx
@@ -62,6 +62,15 @@ const POLYGON_SIDES = _.map(_.range(3, 13), function (value) {
     );
 });
 
+// TODO(LEMS-2228): Remove flags once this is fully released
+const startCoordsUiPhase1Types = [
+    "linear",
+    "linear-system",
+    "ray",
+    "segment",
+    "circle",
+];
+
 type Range = [min: number, max: number];
 
 export type Props = {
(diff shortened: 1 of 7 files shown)


packages/perseus-editor/src/widgets/interactive-graph-editor.tsx
```diff
@@ -473,7 +473,7 @@ class InteractiveGraphEditor extends React.Component<Props> {
                     </LabeledRow>
                 )}
                 {this.props.graph?.type &&
-                    this.props.apiOptions?.flags?.mafs?.["start-coords-ui"] && (
+                    (this.props.apiOptions?.flags?.mafs?.["start-coords-ui"][this.props.graph.type]) && (
```
benchristel, line 476: I think this needs to be

```suggestion
                    (this.props.apiOptions?.flags?.mafs?.["start-coords-ui"]?.[this.props.graph.type]) && (
```

or it will throw an error when `mafs["start-coords-ui"]` is undefined.

benchristel (approved):
This looks good to me. I kind of wonder if it will be hard to manage all these different flags, e.g. to know which ones are turned on in Growthbook, turn them on for everyone who needs to playtest, etc. I might prefer to have milestone-based flags like we do for locked features. But I'm okay with it either way.

benchristel (approved):
LGTM!

PR #1468, opened 2024-07-31
[Hint Mode: Start Coords] Add start coords UI for sinusoid graphs

nishasy:
## Summary:
Add the UI to specify start coords for Sinusoid graph type.

- Add the sinusoid graph type to start-coord-settings.tsx
- Create a start-coords-sinusoid.tsx file with the main UI
- Add a utility for getting the sinusoid equation from coordinates (less complex
  than the static method on InteractiveGraph that does the same)
- Add start coords UI phase 2 flag

Issue: https://khanacademy.atlassian.net/browse/LEMS-2207

## Test plan:
`yarn jest`

Storybook
- http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-sinusoid-with-starting-coords

PR #1469, opened 2024-07-31
[Hint Mode: Start Coords] Add start coords UI for quadratic graphs

nishasy:
## Summary:
Add the UI to specify start coords for Quadratic graph type.

- Add the quadratic graph type to start-coord-settings.tsx
- Create a start-coords-quadratic.tsx file with the main UI
- Add a utility for getting the quadratic equation from coordinates (less complex
  than the static method on InteractiveGraph that does the same)
- Add quadratic graph to the phase 2 flag

Issue: https://khanacademy.atlassian.net/browse/LEMS-2210

## Test plan:
`yarn jest`

Storybook
- http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-quadratic-with-starting-coords

### DIFF
--- a/packages/perseus-editor/src/components/__tests__/util.test.ts
+++ b/packages/perseus-editor/src/components/__tests__/util.test.ts
@@ -4,6 +4,7 @@ import {
     radianToDegree,
     getDefaultGraphStartCoords,
     getSinusoidEquation,
+    getQuadraticEquation,
 } from "../util";
 
 import type {PerseusGraphType, Range} from "@khanacademy/perseus";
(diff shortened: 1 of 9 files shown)


packages/perseus-editor/src/components/__tests__/util.test.ts
```diff
@@ -276,3 +277,34 @@ describe("getSinusoidEquation", () => {
...
+    test.each`
+        point1     | point2      | point3
+        ${[-5, 5]} | ${[-5, -5]} | ${[5, 5]}
+        ${[-5, 5]} | ${[0, -5]}  | ${[-5, 5]}
+        ${[-5, 5]} | ${[0, 5]}   | ${[0, 5]}
+    `("should return division by zero error", ({point1, point2, point3}) => {
+        // Act
+        const equation = getQuadraticEquation([point1, point2, point3]);
+
+        expect(equation).toBe("Division by zero error");
```
benchristel, line 308: This is an interesting case! Thanks for testing it.

I am curious: how does the graph preview render in this case? What happens to the graph when the start coords aren't valid?

packages/perseus-editor/src/components/start-coords-quadratic.tsx
```diff
@@ -0,0 +1,95 @@
...
+                    onChange={(value) =>
+                        onChange([startCoords[0], startCoords[1], value])
+                    }
+                />
+            </View>
+        </>
+    );
+};
+
+const styles = StyleSheet.create({
```
mark-fitzgerald, line 73: NIT: Grrr. Same styling as all the [other "start-coords-*" implementations](https://github.com/Khan/perseus/pull/1468/files#diff-c2eba6387e6940c42713ad16223d5c2689f99010e62db4d6e34b2f1dc822c692R58-R78).
No need to change anything in this PR.

mark-fitzgerald (approved):
LGTM

packages/perseus-editor/src/components/start-coords-quadratic.tsx
nishasy: Ah, I did admittedly copy/paste the entire sinusoid file and build off of it. I should probably break out `Tile` into a separate component. Maybe in a different PR.

packages/perseus-editor/src/components/util.ts
```diff
@@ -215,3 +222,34 @@ export const getSinusoidEquation = (startCoords: [Coord, Coord]) => {
...
+    const b =
+        (p3[0] * p3[0] * (p1[1] - p2[1]) +
+            p2[0] * p2[0] * (p3[1] - p1[1]) +
+            p1[0] * p1[0] * (p2[1] - p3[1])) /
+        denom;
+    const c =
+        (p2[0] * p3[0] * (p2[0] - p3[0]) * p1[1] +
+            p3[0] * p1[0] * (p3[0] - p1[0]) * p2[1] +
+            p1[0] * p2[0] * (p1[0] - p2[0]) * p3[1]) /
+        denom;
```
benchristel, line 250: This is copied from `InteractiveGraph.getQuadraticCoefficients`?

I wish we had a central place where this logic could live, and both `perseus` and `perseus-editor` could import it from there. I think `kmath` is probably the best candidate that place. Would it be too hard to move `getQuadraticCoefficients` to `kmath`?

If you don't want to move it in this PR, I can move it once this is landed.

benchristel (approved):
LGTM!

packages/perseus-editor/src/components/util.ts
nishasy: Yes, that would be awesome! Sorry, I didn't see your comment before landing.

packages/perseus-editor/src/components/__tests__/util.test.ts
nishasy: It just shows all the points where they would be, but the quadratic graph doesn't go through the points. The equation box says "Division by zero error"

Screenshot 2024-07-31 at 3 43 19 PM
Screenshot 2024-07-31 at 3 43 33 PM

nishasy: I'm not too worried about it because I think this is within the "trust the content authors" territory

PR #1470, opened 2024-07-31
[Interactive Graph Locked Figures] Remove m2 flag from the code

nishasy:
## Summary:
Now that locked figures m2 has been released for a while with no issues coming up,
we can remove the flag conditionals from the code.

After this, I'll remove the flag from webapp and growthbook as well.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2016

## Test plan:
`yarn jest`
`yarn typecheck`

Scour the Perseus repo for instances of `m2Features`, `features-m2`, and other
variations. Confirm they no longer exist.

Look for m2b and make sure m2 is no longer next to it in each location.

### DIFF
--- a/.changeset/gentle-melons-jam.md
+++ b/.changeset/gentle-melons-jam.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/perseus": patch
+"@khanacademy/perseus-editor": patch
+---
+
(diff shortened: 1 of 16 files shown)


mark-fitzgerald (approved):
Flag removal = progress
Woo-hoo!

PR #1472, opened 2024-08-01
Add keyboard controls to angle graphs

benchristel:
This PR makes it so angle graph questions can be solved with only a keyboard -
no mouse required!

There are several tricky constraints to this problem:

- Every distinct state of the graph needs to be reachable using only the
  keyboard.
- In order for the movement to feel natural, pressing an arrow key should
  always move the point in the corresponding direction: up, down, left, or right.
- However, the points need to snap, not to a grid, but to angle measures
  (in degrees).
- The side points of the angle should not be allowed to get too close to the
  vertex - if they overlap the vertex there's no way to drag them away again.
- The points shouldn't be able to move outside the graph bounds
- It should never be possible for an edge point to get "stuck" or blocked in its
  movement by the edge of the graph, or by the angle vertex.

I believe this PR solves for all these constraints. To make it work, I had to do
a few slightly weird things:

- I constrained the movement of the vertex so no part of the angle-measure arc
  can go outside the graph bounds. This ensures that all the control points of
  the angle can always be placed onscreen, and prevents the side points from
  getting stuck on the edges of the graph.
- I reworked the interface of `useDraggable`, renaming `constrain` to
  `constrainKeyboardMovement`. Movement constraints for dragging are now
  implemented entirely in the reducer. (They were almost all in the reducer
  before; circles were the only graph type that still depended on the constrain
  function for dragging.)
- I also introduced a new type of `useDraggable` constraint, where the destination
  points for the up, left, right, and down arrows are listed explicitly. I think
  this ultimately makes the code clearer. It's also more efficient (since
  useDraggable does not have to do a bunch of repeated trig calculations to
  search for valid destination points).

Issue: https://khanacademy.atlassian.net/browse/LEMS-2134

## Test plan:

View `/?path=/story/perseus-widgets-interactive-graph--angle-with-mafs` in
Storybook. Try to break the graph.

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/angle.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/angle.tsx
@@ -1,19 +1,22 @@
+import {vec} from "mafs";
 import * as React from "react";
 
+import {calculateAngleInDegrees, polar} from "../math";
+import {findIntersectionOfRays} from "../math/geometry";
 import {actions} from "../reducer/interactive-graph-action";
 import useGraphConfig from "../reducer/use-graph-config";
 
 import {Angle} from "./components/angle-indicators";
 import {trimRange} from "./components/movable-line";
-import {StyledMovablePoint} from "./components/movable-point";
+import {MovablePoint} from "./components/movable-point";
 import {SVGLine} from "./components/svg-line";
 import {Vector} from "./components/vector";
 import {useTransformVectorsToPixels} from "./use-transform";
 import {getIntersectionOfRayWithBox} from "./utils";
 
 import type {CollinearTuple} from "../../../perseus-types";
+import type {Segment} from "../math/geometry";
 import type {AngleGraphState, MafsGraphProps} from "../types";
-import type {vec} from "mafs";
 
 type AngleGraphProps = MafsGraphProps<AngleGraphState>;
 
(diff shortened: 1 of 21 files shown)


packages/perseus/src/widgets/interactive-graphs/graphs/use-draggable.test.tsx
```diff
@@ -127,21 +130,55 @@ describe("useDraggable", () => {
...
-        // Assert: the draggable element was moved to (1, -1) due to the
-        // constrain function. If you see (1.2, -1.3) instead, that means the
-        // constraint is not being applied.
-        expect(onMoveSpy).toHaveBeenCalledWith([1, -1]);
+        // Assert: the draggable element was moved one step to the right
+        expect(onMoveSpy.mock.calls).toEqual([[[1, 0]]]);
+    });
+
+    it("constrains the destination point using a constrainKeyboardMovement object", async () => {
+        // Arrange: a 200x200px graph with a 20-unit range in each dimension.
```
SonicScrewdriver, line 148: I really appreciate these comment explanations about what the graph is setting up!

packages/perseus/src/widgets/interactive-graphs/graphs/angle.tsx
```diff
@@ -89,16 +85,102 @@ export function AngleGraph(props: AngleGraphProps) {
...
-                    key={"point-" + i}
-                    snapTo={"angles"}
-                    point={point}
-                    onMove={(destination: vec.Vector2) =>
-                        dispatch(actions.angle.movePoint(i, destination))
-                    }
-                />
-            ))}
+            {/* vertex */}
+            <MovablePoint
```
SonicScrewdriver, line 89: Also appreciate these callouts to help explain which points they are, now that we're not mapping them.

SonicScrewdriver (approved):
Very well commented and incredibly appreciated! Thank you for setting all of this up Ben, I think it's much cleaner!

PR #1474, opened 2024-08-01
ImageLoader: fix types and remove @ts-expect-error instances

jeremywiebe:
## Summary:

While chasing down a question from LearningEquality, I noticed some `@ts-expect-error` usages that made me wonder. After removing them I found it was quite easy to make the types correct.

This is also a common problem in our codebase following the port from Flow to TypeScript. We often set props to have the type `SomeType | null | undefined`. However, often we don't need to support it being `null`. In this case, removing the assignment of a default value (`null`) fixed the main issue.

Issue: "none"

## Test plan:

`yarn typecheck`
`yarn lint`

### DIFF
--- a/packages/perseus/src/components/image-loader.tsx
+++ b/packages/perseus/src/components/image-loader.tsx
@@ -28,7 +28,7 @@ export type ImageProps = {
     alt: string;
     title?: string;
     ["aria-hidden"]?: boolean;
-    tabIndex?: string;
+    tabIndex?: number;
     onClick?: (e: React.SyntheticEvent) => void;
     style?: Dimensions;
 };
@@ -41,7 +41,7 @@ type Props = {
     // When the DOM updates to replace the preloader with the image, or
     // vice-versa, we trigger this callback.
     onUpdate: (status: (typeof Status)[keyof typeof Status]) => void;
-    preloader: () => React.ReactElement | null | undefined;
+    preloader: (() => React.ReactNode) | null | undefined;
     src: string;
 };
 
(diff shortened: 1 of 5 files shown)


packages/perseus/src/components/image-loader.tsx
```diff
@@ -28,7 +28,7 @@ export type ImageProps = {
     alt: string;
     title?: string;
     ["aria-hidden"]?: boolean;
-    tabIndex?: string;
+    tabIndex?: number;
```
jeremywiebe, line 31: This type error was exposed once I removed the `@ts-expect-error`s.

packages/perseus/src/components/image-loader.tsx
```diff
@@ -41,7 +41,7 @@ type Props = {
     // When the DOM updates to replace the preloader with the image, or
     // vice-versa, we trigger this callback.
     onUpdate: (status: (typeof Status)[keyof typeof Status]) => void;
-    preloader: () => React.ReactElement | null | undefined;
+    preloader: (() => React.ReactNode) | null | undefined;
```
jeremywiebe, line 44: The main fix here is wrapping the function that returns a React element, otherwise it's a function that returns `React.ReactElement | null | undefined`. I changed the return type to `React.ReactNode` as that is what a `render` function returns.

packages/perseus/src/server-item-renderer.tsx
```diff
@@ -72,7 +72,6 @@ type SerializedState = {
     hints: any;
 };
 
-/* eslint-disable-next-line react/no-unsafe */
```
jeremywiebe, line 75: This was not actually suppressing anything so I'm removing it.

mark-fitzgerald (approved):
Thanks for cleaning these up.

PR #1475, opened 2024-08-01
fix(LEMS-2132): add dialog title and description to expression widget

anakaren-rojas:
## Summary:
- Adds dialog title and description for expression widget 
- Bumps wonder blocks popover versions 

Note:
- Screen reader will not read both the `aria-label` and the `aria-describedby` attributes. User has the option to navigate into the expression widget navigation instructions. 

Issue: LEMS-2132

## Test plan:
1. Navigate to [Math Input stories](https://650db21c3f5d1b2f13c02952-njskcgaclq.chromatic.com/?path=/docs/perseus-components-math-input--docs)  
2. Using Screen Reader - open the Math Input Expression Widget 

### Expected behavior:
Screen reader reads "mathematics keyboard"

## Screenshots: 
https://github.com/user-attachments/assets/7693b8d1-f34d-4b09-944a-c145c323ab5a

### With unique id
Screenshot 2024-08-05 at 3 18 21 PM


yarn.lock
```diff
@@ -2765,17 +2765,17 @@
     "@khanacademy/wonder-blocks-tokens" "^1.3.0"
     "@khanacademy/wonder-blocks-typography" "^2.1.11"
 
-"@khanacademy/wonder-blocks-popover@3.2.9":
-  version "3.2.9"
-  resolved "https://registry.yarnpkg.com/@khanacademy/wonder-blocks-popover/-/wonder-blocks-popover-3.2.9.tgz#f82212edd117832b7eab6066d8b6381fc982126e"
-  integrity sha512-WjstA4acn/gOfo1CedA90oq5hWZnkUl4Lj2hpMCHYSTflfEhz0mjcFWzWXPAta+rtk6pAEnAKzqWT1oJ5q+Hvg==
+"@khanacademy/wonder-blocks-popover@3.2.11":
```
anakaren-rojas, line 2768: Do we usually push the `.lock` files?

handeyeco: Yep

packages/perseus/src/components/math-input.tsx
```diff
@@ -330,25 +332,38 @@ class InnerMathInput extends React.Component<InnerProps, State> {
...
-                                    convertDotToTimes={
-                                        this.props.convertDotToTimes
-                                    }
-                                    {...(this.props.keypadButtonSets ??
-                                        mapButtonSets(this.props?.buttonSets))}
-                                />
-                            </PopoverContentCore>
+                            <>
+                                <HeadingMedium
+                                    id="popover-content"
```
handeyeco, line 340: Right now it's possible to have multiple keypads open at the same time. Should we auto-generate an ID here to make sure everything connects correctly?

handeyeco (approved):
Thanks! Just a concern about IDs being unique on a page.

packages/perseus/src/components/math-input.tsx
anakaren-rojas: Good call out! Updated to use the first 8 chars of a UUID - the new ids are `popover-content-${shortened-uuid}`

PR #1477, opened 2024-08-01
Migrate ItemExtrasEditor to Wonder-Blocks

jeremywiebe:
## Summary:

While working in the Perseus Editor I took a few minutes to convert the item extras editor to use WB checkboxes. This results in much nicer layout (checkboxes all on the left, better typography).

| Before | After | 
|---|---|
| Screenshot 2024-08-01 at 4 09 02 PM | Screenshot 2024-08-01 at 4 08 52 PM | 

Issue: LEMS-1809

## Test plan:

Run storybook and look at the Editor Page or Item Extras Editor stories. 

I also verified that it still serializes the right data out and that nested checkboxes work.

### DIFF
--- a/packages/perseus-editor/src/__tests__/item-extras-editor.test.tsx
+++ b/packages/perseus-editor/src/__tests__/item-extras-editor.test.tsx
@@ -19,19 +19,19 @@ describe("ItemExtrasEditor", () => {
         render(<ItemExtrasEditor onChange={() => {}} />);
 
         // Assert
-        expect(screen.getByLabelText("Show calculator:")).not.toBeChecked();
-        expect(screen.getByLabelText("Show periodic table:")).not.toBeChecked();
+        expect(screen.getByLabelText("Show calculator")).not.toBeChecked();
+        expect(screen.getByLabelText("Show periodic table")).not.toBeChecked();
         expect(
-            screen.queryByText("Include key/legend with periodic table:"),
+            screen.queryByText("Include key/legend with periodic table"),
         ).not.toBeInTheDocument();
         expect(
-            screen.getByLabelText("Show z table (statistics):"),
+            screen.getByLabelText("Show z table (statistics)"),
         ).not.toBeChecked();
         expect(
-            screen.getByLabelText("Show t table (statistics):"),
+            screen.getByLabelText("Show t table (statistics)"),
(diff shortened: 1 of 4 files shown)


packages/perseus-editor/src/item-extras-editor.tsx
```diff
@@ -43,75 +47,69 @@ class ItemExtrasEditor extends React.Component<Props> {
             <div className="perseus-answer-editor">
                 <div className="perseus-answer-options">
                     <ItemExtraCheckbox
-                        label="Show calculator:"
+                        label="Show calculator"
                         infoTip="Use the calculator when completing difficult calculations is NOT the intent of the question. DON’T use the calculator when testing the student’s ability to complete different types of computations."
                         checked={this.props.calculator}
                         onChange={(e) => {
```
nishasy, line 53: nit (probably out of scope here): I'd prefer if this (and other instances below) were named something clearer, like `newValue` or `newCheckedValue` or `newCheckedState` or something instead of `e`

nishasy (approved):
This is so awesome! Looks good to me. Left one small nonblocking comment.

nishasy:
Also! Don't forget to update the tests for the labels that no longer have the ":" in them

packages/perseus-editor/src/item-extras-editor.tsx
jeremywiebe: I agree. I've updated it to match what the WB checkbox uses: `newCheckedState`

packages/perseus-editor/src/__tests__/item-extras-editor.test.tsx
```diff
@@ -58,15 +58,15 @@ describe("ItemExtrasEditor", () => {
...
+        const checkbox = screen.getByLabelText("Show periodic table");
 
         // Act
         await userEvent.click(checkbox);
 
         // Assert
         // visible when periodicTable is checked
         expect(
-            screen.getByText("Include key/legend with periodic table:"),
+            screen.getByText("Include key/legend with periodic table"),
```
nishasy, line 69: Nit - optional: I think it's supposed to be better practice to use `getByRole` if possible? So I'd probably do `screen.getByRole("checkbox", {name: "Include key/legend with periodic table"})`

PR #1478, opened 2024-08-02
[Interactive Graph Editor] Add white as a fill option for locked ellipses and polygons

nishasy:
## Summary:
We got a request from content authors to add a "white" fill type for locked ellipses
and polygons. Adding that here.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2119

## Test plan:
Storybook
- http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--mafs-with-locked-figures-m-2-b-flag
- Go through all the fill options for locked ellipses and locked polygons
- Confirm that the swatch in the summary also updates corrently

### DIFF
--- a/.changeset/gentle-roses-pump.md
+++ b/.changeset/gentle-roses-pump.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/perseus": patch
+"@khanacademy/perseus-editor": patch
+---
+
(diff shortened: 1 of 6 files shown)


nishasy:
Added tests for the rendering of the actual polygons/ellipses on the graph. I didn't test the swatch color in the editor summary because it seemed like overkill.

PR #1479, opened 2024-08-02
Remove PropCheckBox

handeyeco:
## Summary:
While looking at WB-ifying ExpressionEditor I noticed this weird component called PropCheckBox and thought we should just switch to WB Checkbox because:

- We're trying to offload work to WB when possible
- PropCheckBox was using ancient `createReactClass`
- It wasn't properly typed
- IMO what it was doing was a little too cute

Issue: Part of [LEMS-2222](https://khanacademy.atlassian.net/browse/LEMS-2222)

## Test plan:
Nothing should change, tests should all pass

[LEMS-2222]: https://khanacademy.atlassian.net/browse/LEMS-2222?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/perseus-editor/src/components/interactive-graph-settings.tsx
```diff
@@ -547,10 +548,12 @@ class InteractiveGraphSettings extends React.Component<Props, State> {
...
                             <div className="perseus-widget-left-col">
-                                <PropCheckBox
+                                <Checkbox
                                     label="Show tooltips"
-                                    showTooltips={this.props.showTooltips}
-                                    onChange={this.change}
+                                    checked={this.props.showTooltips}
+                                    onChange={(value) => {
+                                        this.change({showTooltips: value});
+                                    }}
```
Myranae, line 556: Suggestion: It seems like this might not have had coverage before, but it might be easy to cover with something like the following added to the `graph-settings.test.tsx` file. Looks like there are a number of places marked as missing coverage. Probably not necessary to fix all the coverage, but here's a possible test here at least:

```
test("calls onChange when tooltips label is changed", async () => {
        // Arrange
        const onChange = jest.fn();
        render(
            <GraphSettings
                editableSettings={["graph"]}
                showTooltips={true}
                onChange={onChange}
            />,
        );

        // Act
        const checkbox = screen.getByRole("checkbox", {
            name: "Show tooltips",
        });
        await userEvent.click(checkbox);

        // Assert
        expect(onChange).toHaveBeenCalledWith(
            expect.objectContaining({showTooltips: false}),
            undefined,
        );
    });
```

Myranae (approved):
Looks good. Love moving anything over to WB instead of custom. Was able to manually check radio, interactive graph, number input, number line, and expression in storybook. The checkboxes seem to work as intended. Tried sorter and categorizer, but the stories don't seem to work. Number line tooltips don't work on prod or this branch, so that's fine XD Let me know if you'd like me to also manually check the rest, but seems good as they all follow a similar pattern.

PR #1480, opened 2024-08-02
WB-ify ExpressionEditor

handeyeco:
## Summary:
Move ExpressionEditor to using WB.

Issue: Part of [LEMS-2222](https://khanacademy.atlassian.net/browse/LEMS-2222)

## Test plan:
Behavior shouldn't change, just move components over to WB.

[LEMS-2222]: https://khanacademy.atlassian.net/browse/LEMS-2222?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/perseus-editor/src/widgets/expression-editor.tsx
```diff
@@ -491,58 +493,62 @@ class AnswerOption extends React.Component<
...
-            removeButton = (
-                <button
-                    type="button"
-                    className="simple-button orange"
+        const removeButton = this.state.deleteFocused ? (
+            <>
+                <Button
+                    size="small"
                     onClick={this.handleImSure}
-                    onBlur={this.handleDeleteBlur}
```
handeyeco, line 510: WB buttons don't have `onBlur`, so I used a two button system

packages/perseus/src/components/text-input.tsx
```diff
@@ -63,7 +63,6 @@ class TextInput extends React.Component<Props> {
         const formattedValue = value === null ? "" : value.toString();
 
         return (
-            // $FlowIgnore
```
handeyeco, line 66: We don't use Flow anymore

packages/perseus-editor/src/components/sortable.tsx
```diff
@@ -1,11 +1,17 @@
...
 import PropTypes from "prop-types";
 import * as React from "react";
 import ReactDOM from "react-dom";
 
 const PT = PropTypes;
 
-// Takes an array of components to sort
+/**
+ * Takes an array of components to sort.
+ * As of this writing, there are two sortable components
```
anakaren-rojas, line 11: nit: `As of 08/05/24` vs `As of this writing` might be helpful for future us

packages/perseus-editor/src/widgets/expression-editor.tsx
```diff
@@ -491,58 +493,62 @@ class AnswerOption extends React.Component<
...
-                <div className="answer-handle" />
+            <div className={css(styles.answerOption)}>
+                <div className={css(styles.answerHandle)} />
 
-                <div className="answer-body">
-                    <div className="answer-considered">
-                        <div
+                <div className={css(styles.answerBody)}>
+                    <div>
+                        <button
```
jeremywiebe, line 539: Does it make sense to migrate this to be a WB button right away (or at least a Clickable)?

jeremywiebe (approved):
Nice to see more wb-ifying going on!  Thanks!

packages/perseus-editor/src/widgets/expression-editor.tsx
handeyeco: I tried to move it to Clickable (WB button I think would be too difficult to customize) and I ran into the most bizarre bug I've ever seen and now I'm tired and I don't want to fight it anymore so it's staying a `button`.

https://khanacademy.slack.com/archives/CEA6W0F6F/p1722982631694429

PR #1481, opened 2024-08-02
[Interactive Graph Editor] Stop page scrolling on number text field focused scroll

nishasy:
## Summary:
Consider an `<input>` element with `type="number"`. The default behavior in React (except Firefox)
if it's focused and wheel over it is that it does BOTH (a) change the number and (b) scroll
the entire page.

There is a specific request from content authors to make it so that only _one_ of those
bevahiors happens.

In this PR:
- Make a new component called `ScrolllessNumberTextField` that uses `stopPropagation`
  to stop the page scroll. The number changes on scroll (except in Firefox).
- Stories to test this
- Updates to all the `type="number"` text fields in the perseus editor to use
  the new ScrolllessNumberTextField component.

Notes:
- Even with this custom implementation, Firefox inputs scroll the page and not the number.
  This is in line with native (non-React) HTML input behavior. As this still only executes
  _one_ of the actions, this is still fine.
- It doesn't seem to be possible to test the input change on `wheel` event.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2121

## Test plan:
`yarn jest packages/perseus-editor/src/components/__tests__/scrollless-number-text-field.test.tsx`

Storybook
- Check all the inputs for all the start coords settings and the locked figures settings

### DIFF
--- a/packages/perseus-editor/src/components/scrollless-number-text-field.tsx
+++ b/packages/perseus-editor/src/components/scrollless-number-text-field.tsx
@@ -1,0 +1,45 @@
+import {TextField} from "@khanacademy/wonder-blocks-form";
+import * as React from "react";
+
+import type {PropsFor} from "@khanacademy/wonder-blocks-core";
+
+/**
+ * This is a custom text field of type="number" for use in Perseus Editors.
+ *
+ * This makes it so that the text field's input number updates on scroll
+ * without scrolling the page.
+ *
+ * NOTE 1: Native HTML number inputs do not update the number value on scroll,
+ * they only scroll the page. For some reason, inputs in React do NOT work
+ * this way. By default, scrolling on a focused number input in React causes
+ * BOTH the input value to change AND the page to scroll. The behavior in
+ * this component is an improvement on the React behavior, but it's the
+ * opposite of the native HTML behavior.
+ *
+ * NOTE 2: Firefox seems to have a custom override for this. Even with this
+ * stopPropogation, Firefox matches the native HTML behavior.
+ */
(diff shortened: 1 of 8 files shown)


packages/perseus-editor/src/components/scrollless-number-text-field.tsx
```diff
@@ -0,0 +1,45 @@
...
+import type {PropsFor} from "@khanacademy/wonder-blocks-core";
+
+/**
+ * This is a custom text field of type="number" for use in Perseus Editors.
+ *
+ * This makes it so that the text field's input number updates on scroll
+ * without scrolling the page.
+ *
+ * NOTE 1: Native HTML number inputs do not update the number value on scroll,
+ * they only scroll the page. For some reason, inputs in React do NOT work
```
benchristel, line 13: The reason this happens is explained here: https://stackoverflow.com/a/68266494

Might be worth linking to that answer so future maintainers know what's up.

benchristel (approved):
LGTM. I feel like this is really a bug in React, but since we probably won't get it fixed there, we might consider it a bug in Wonder Blocks. I don't think we'd ever want a scroll gesture to both change the input and scroll the page.

nishasy:
@benchristel I did explore updating this in Wonder Blocks, but I don't think it's feasible to change this everywhere across Khan Academy. (a) Because this behavior is the exact opposite of expected native behavior, and (b) because it's inconsistent across browsers. I think this is fine for the content authors as it was specifically requested, but it seems like it would warrant a much larger discussion if we wanted to address this elsewhere. 😞

PR #1482, opened 2024-08-02
[Interactive Graph Editor: Storybook] Add a story for Point graph type start coords

nishasy:
## Summary:
There currently isn't a good way to manually test the start coords for Point graph
in storybook, so I'm adding a story for that here.

- Added story for point graph start coords in the interactive graph editor stories.
- Shortened the names of the stories so they show up neater in the files structure.
- Added `numPoints` to the stateful mafs graph useEffect so it appropriately updates
  the preview when the number of points is changed.
- Fixed the builder for point graphs so the default coords value is undefined and it
  calculates the correct coords itself. It was defaulting to [0, 0] before, which
  really didn't work when there was a different number of points.
- Added `points` graph to the flag utility so it uses the mafs version in storybook.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2209

## Test plan:
Storybook
- Change the number of points and confirm that it looks correct in both previews

## Screenshot

Screenshot 2024-08-02 at 3 57 23 PM


packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.tsx
```diff
@@ -96,6 +96,7 @@ export const StatefulMafsGraph = React.forwardRef<
     }, [dispatch, xMinRange, xMaxRange, yMinRange, yMaxRange]);
 
     const numSegments = graph.type === "segment" ? graph.numSegments : null;
+    const numPoints = graph.type === "point" ? graph.numPoints : null;
```
Myranae, line 99: Comment: I wonder if there is a good way for us to leave a note/reminder about needing to add new information like this here and to the useEffect below. 🤔

packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.tsx
```diff
@@ -96,6 +96,7 @@ export const StatefulMafsGraph = React.forwardRef<
     }, [dispatch, xMinRange, xMaxRange, yMinRange, yMaxRange]);
```
Myranae, line 97: Suggestion: Might be helpful to have some comments here explaining what this chunk is for. Apologies, should have added that when I worked on this section too!

packages/perseus/src/widgets/interactive-graphs/interactive-graph-question-builder.ts
```diff
@@ -695,7 +695,7 @@ class PointsGraphConfig implements InteractiveFigureConfig {
         },
     ) {
         this.numPoints = numPoints;
-        this.coords = options?.coords ?? [[0, 0]];
+        this.coords = options?.coords;
```
Myranae, line 698: Question: I don't think this affects this PR, but you mentioned in the summary that you changed this so it "calculates the correct coords itself." How does it do that?

Myranae (approved):
Looks good to me! Added some non-blocking questions/comments :)

packages/perseus/src/widgets/interactive-graphs/interactive-graph-question-builder.ts
nishasy: If I don't pass in any coords, then the graph automatically initializes with the default starting coords in `initialize-graph-state.ts` when it renders. If I hard-code it here, it doesn't initialize to the right one based on the number of points. I could also call the initializer function here, but that would be redundant.

PR #1484, opened 2024-08-05
clean up info-tip

handeyeco:
## Summary:
While working on another issue, I noticed that InfoTip could use some touchups


Myranae (approved):
Looks good! Thanks for the cleanup!

PR #1486, opened 2024-08-05
[Hint Mode: Start Coords] Add start coords UI for point graphs (limited only)

nishasy:
## Summary:
Add the UI to specify start coords for Point graph type.

- Add the point graph type to start-coord-settings.tsx
- Create a start-coords-point.tsx file with the main UI
- Add the start coords UI point flag
- Move the logic for determining whether the UI should be
  shown based on the feature flags out into the util file.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2209

## Test plan:
`yarn jest`

Storybook

image

### DIFF
--- a/packages/perseus-editor/src/components/__tests__/util.test.ts
+++ b/packages/perseus-editor/src/components/__tests__/util.test.ts
@@ -256,6 +256,76 @@ describe("getDefaultGraphStartCoords", () => {
 
         expect(defaultCoords).toEqual({center: [0, 0], radius: 2});
     });
+
+    test("should get default start coords for a sinusoid graph", () => {
+        // Arrange
+        const graph: PerseusGraphType = {type: "sinusoid"};
+        const range = [
+            [-10, 10],
+            [-10, 10],
+        ] satisfies [Range, Range];
+        const step = [1, 1] satisfies [number, number];
+
+        // Act
+        const defaultCoords = getDefaultGraphStartCoords(graph, range, step);
+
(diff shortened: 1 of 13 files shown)


packages/perseus-editor/src/components/__tests__/util.test.ts
```diff
@@ -256,6 +256,76 @@ describe("getDefaultGraphStartCoords", () => {
 
         expect(defaultCoords).toEqual({center: [0, 0], radius: 2});
     });
+
+    test("should get default start coords for a sinusoid graph", () => {
```
nishasy, line 260: oops, forgot to add tests for the sinusoid and quadratic start coords before

packages/perseus-editor/src/components/util.ts
```diff
@@ -253,3 +260,37 @@ export const getQuadraticEquation = (startCoords: [Coord, Coord, Coord]) => {
...
+    if (
+        startCoordsPoint &&
+        graph.type === "point" &&
+        graph.numPoints !== "unlimited"
+    ) {
+        return true;
+    }
+
+    return false;
+};
```
nishasy, line 296: Moved the flag logic out into the util file because it was getting super cluttered inside the interactive-graph-editor.tsx.

benchristel: 👍 It's nice to have a place for this logic!

benchristel (approved):
Looks great! 🚢

PR #1487, opened 2024-08-05
Add readOnly mode to interactive graphs to prevent keyboard interactions when question marked correct

Myranae:
## Summary:
Widgets are placed into a `readOnly` mode when they are part of an exercise that was answered correctly. This is to prevent a state where the question is marked correct, but the user has modified the graph to an incorrect state. Mouse interactions are prevented by adding an invisible div over the widget, but this does not prevent keyboard interactions. This change adds a `readOnly` property to graphs that reduces the tabIndex to -1 in instances where `readOnly` mode is active (when the question was answered correctly).

Issue: LEMS-2175

## Test plan:
- Confirm all tests pass
- Checkout the branch associated with this PR. Run `yarn storybook`
- Confirm the graph in the readOnly mode story is non-interactive: http://localhost:6006/?path=/story/perseus-widgets-interactive-graph--polygon-with-mafs-read-only
- Compare the interactivity with the corresponding story not in readOnly mode: http://localhost:6006/?path=/story/perseus-widgets-interactive-graph--polygon-with-mafs
- Use the PR npm snapshot to test this in webapp with a ZND to confirm when a graph related exercise is answered correctly that the graph cannot be interacted with either via mouse or keyboard (don't see a snapshot, so maybe this isn't possible)


packages/perseus/src/types.ts
```diff
@@ -189,6 +189,10 @@ export type APIOptions = Readonly<{
     ) => unknown;
     GroupMetadataEditor?: React.ComponentType<StubTagEditorType>;
     showAlignmentOptions?: boolean;
+    /**
+     * A boolean that indicates whether the associated problem has been
+     * answered correctly and should no longer be interactive.
+     */
```
benchristel, line 195: Thanks for documenting this!

packages/perseus/src/widgets/interactive-graphs/graphs/circle.tsx
```diff
@@ -46,7 +46,7 @@ function MovableCircle(props: {
     onMove: (newCenter: vec.Vector2) => unknown;
 }) {
     const {center, radius, onMove} = props;
-    const {snapStep, hintMode} = useGraphConfig();
+    const {snapStep, hintMode, readOnly} = useGraphConfig();
```
benchristel, line 49: Since both `hintMode` and `readOnly` both come from the graph config and are always used together, what do you think about consolidating them? Creating the graph config would look like this:

```tsx
<GraphConfigContext.Provider
  value={{
    // ...
    disableKeyboardInteraction: hintMode || readOnly,
  }}
>
```

And then individual components would just look at `disableKeyboardInteraction`. That way, the `||` logic wouldn't have to be duplicated in every component.

benchristel (approved):
This looks good! I'd prefer it if the `hintMode || readOnly` logic didn't have to be duplicated in multiple components. I left a comment inline describing one way to achieve that. But I don't think that will need another round of review.

packages/perseus/src/widgets/__stories__/interactive-graph.stories.tsx
```diff
@@ -71,6 +82,15 @@ export const PolygonWithMafs = (args: StoryArgs): React.ReactElement => (
...
 );
 
+export const PolygonWithMafsReadOnly = (
+    args: StoryArgs,
+): React.ReactElement => (
+    <RendererWithDebugUI
+        {...polygonReadOnlyOptions}
+        question={polygonQuestion}
+    />
+);
```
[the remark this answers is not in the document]
Myranae: Love it! I tried to refactor the `{...mafsOptions}` part, but wasn't quite sure how to do it. I think I might have been missing the double brackets. 🤦‍♀️ Thank you for pointing this out! Updated :)

packages/perseus/src/widgets/interactive-graphs/graphs/circle.tsx
benchristel: I don't recall exactly where the mouse interactions get disabled. If the code that disables them is reading from `useGraphConfig()` to decide what to do, then `disableInteraction` sounds good to me.

Myranae: I'll double check :)

Myranae: Looks like mouse interactions are disabled just below where we set the config. Doesn't look like it's reading from the config, so will stick with `disableKeyboardInteractions`. 
image

Myranae: Updated! 😁

jeremywiebe (approved):
This all looks great. I wonder if it would be worth one test (maybe per graph type) that ensures when `readOnly: true` that nothing can be interacted with.  

One idea could be to run the "answer graph correctly" tests with `readOnly: true` and ensure they fail because nothing can be moved with the keyboard?

PR #1489, opened 2024-08-06
Hints: fix hints type on PerseusItem to be correct

jeremywiebe:
## Summary:

Our `hints` type on `ItemRenderer` used the `PerseusRenderer` type. That's mostly correct, except that a Hint can also have the `replace` key. 

Issue: "none"

## Test plan:

`yarn typecheck` ✅

### DIFF
--- a/packages/perseus-editor/src/hint-editor.tsx
+++ b/packages/perseus-editor/src/hint-editor.tsx
@@ -318,14 +318,13 @@ class CombinedHintsEditor extends React.Component<CombinedHintsEditorProps> {
         silent: boolean,
     ) => {
         // TODO(joel) - lens
-        const hints = _(this.props.hints).clone();
+        const hints = [...this.props.hints];
         hints[i] = _.extend(
             {},
             this.serializeHint(i, {keepDeletedWidgets: true}),
             newProps,
         );
 
-        // @ts-expect-error - TS2740 - Type 'Hint' is missing the following properties from type 'readonly Hint[]': length, concat, join, slice, and 18 more.
         this.props.onChange({hints: hints}, cb, silent);
     };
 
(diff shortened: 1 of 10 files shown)


packages/perseus-editor/src/hint-editor.tsx
```diff
@@ -318,14 +318,13 @@ class CombinedHintsEditor extends React.Component<CombinedHintsEditorProps> {
         silent: boolean,
     ) => {
         // TODO(joel) - lens
-        const hints = _(this.props.hints).clone();
```
jeremywiebe, line 321: Moving away from `_` gives us better typing.

packages/perseus-editor/src/hint-editor.tsx
```diff
@@ -360,10 +354,9 @@ class CombinedHintsEditor extends React.Component<CombinedHintsEditorProps> {
     };
 
     addHint: () => void = () => {
-        const hints = _(this.props.hints)
-            .clone()
-            // @ts-expect-error - TS2339 - Property 'concat' does not exist on type 'Hint'.
-            .concat([{content: ""}]);
+        const hints = [...this.props.hints].concat([
+            {content: "", images: {}, widgets: {}},
+        ]);
```
benchristel, line 359: `.concat` returns a copy, so you don't need the array-spread syntax here:

```suggestion
        const hints = this.props.hints.concat([
            {content: "", images: {}, widgets: {}},
        ]);
```

packages/perseus/src/perseus-types.ts
```diff
@@ -127,6 +127,15 @@ export type PerseusRenderer = {
...
 };
 
+export type Hint = PerseusRenderer & {
+    /**
+     * When `true`, causes the previous hint to be replaced with this hint when
+     * displayed. When `false`, the previous hint remains visible when this one
+     * is displayed.
+     */
+    replace?: boolean;
+};
```
benchristel, line 137: The `PerseusRenderer` type already has a `replace` property — should that be removed now?

benchristel (approved):
Looks good overall. I left a couple suggestions inline. Thanks for fixing this!

packages/perseus/src/perseus-types.ts
jeremywiebe: Good catch! Yes, we can remove it now!

PR #1490, opened 2024-08-06
Export compiled CSS in package.json

jeremywiebe:
## Summary:

Recently we changed some of our packages to use the modern syntax for declaring packge exports: namely the `exports` key. When a package defines this key, it no longer allows arbitrary relative exports (ie. `import X from '@khanacademy/perseus/dist/some-file.js';`). 

This: 

image

Becomes: 

image

As a result, our bundled CSS was no longer exported. In webapp today, we need to import the bundled CSS using a fully-relative path (ie. `import '../../node_modules/@khanacademy/perseus/dist/index.css''`). This is clunky and the import paths are not portable. 

So, this PR fixes it by exporting all CSS files from our bundles.

Issue: "none"

## Test plan:

Update webapp with the snapshot npm package and then check if we can change our CSS imports to `import '@khanacademy/perseus/index.css';`

### DIFF
--- a/packages/math-input/package.json
+++ b/packages/math-input/package.json
@@ -30,7 +30,8 @@
             "require": "./dist/strings.js",
             "types": "./dist/strings.d.ts",
             "source": "./src/strings.ts"
-        }
+        },
+        "./*.css": "./dist/*.css"
     },
     "files": [
         "dist"
--- a/packages/math-input/package.json
+++ b/packages/math-input/package.json
@@ -31,7 +31,7 @@
             "types": "./dist/strings.d.ts",
             "source": "./src/strings.ts"
         },
-        "./index.css": "./dist/index.css"
+        "./styles.css": "./dist/index.css"
     },
     "files": [
         "dist"
--- a/packages/math-input/package.json
+++ b/packages/math-input/package.json
@@ -31,7 +31,7 @@
             "types": "./dist/strings.d.ts",
             "source": "./src/strings.ts"
         },
-        "./*.css": "./dist/*.css"
+        "./index.css": "./dist/index.css"
     },
     "files": [
         "dist"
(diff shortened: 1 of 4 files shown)


packages/math-input/package.json
```diff
@@ -30,7 +30,8 @@
             "require": "./dist/strings.js",
             "types": "./dist/strings.d.ts",
             "source": "./src/strings.ts"
-        }
+        },
+        "./index.css": "./dist/index.css"
```
jeresig, line 34: Is index.css a good name? `perseus.css` may be better?

jeresig (approved):
Nice! One name nit inline.

packages/math-input/package.json
jeremywiebe: `index.css` is a [default filename](https://github.com/Anidetrix/rollup-plugin-styles?tab=readme-ov-file#css-extraction) (although it's not documented explicitly there) provided by the `rollup-plugin-styles` plugin we use. 

https://github.com/Khan/perseus/blob/3d7e3b6854cf9f3a945e86988c6bbcb2b6c9e2cb/config/build/rollup.config.js#L216

I can change it, but it would take a bit more effort as we use a single rollup build file to generate Rollup build configs for each package. I'll see how difficult that is (basically, I think we'd have the `perseus` package have `perseus.css`, `perseus-editor` have `perseus-editor.css`.  More customization than that I don't think is worth it. :)

jeresig: To clarify, leave the file as `./dist/index.css`, but export it as `./perseus.css` - having an export named "index" doesn't really make sense? (or export it as `./styles.css` - something that provides a context clude)

jeremywiebe: I think I like `./styles.css`. That gives a consistent naming across all Perseus packages, but signals what it is.

PR #1491, opened 2024-08-06
[Hint Mode: Start Coords] Correct flag logic for polygon graph start coords UI

nishasy:
## Summary:
I noticed that the "Reset to default coordinates" button still shows up for
polygon graphs with unlimited sides, but it's not supposed to. I figured out
that this is because I accidentally used `numPoints` instead of `numSides` in
the logic.

- Fix the flags logic
- Add tests

Issue: https://khanacademy.atlassian.net/browse/LEMS-2072

## Test plan:

Storybook
- Go to http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-polygon
- Set number of sides to "unlimited sides"
- Confirm that the start coords UI no longer shows up

### DIFF
--- a/.changeset/clean-glasses-watch.md
+++ b/.changeset/clean-glasses-watch.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": patch
+---
+
+[Hint Mode: Start Coords] Correct flag logic for polygon graph start coords UI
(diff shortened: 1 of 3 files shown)


benchristel (approved):
Good catch! LGTM.

PR #1492, opened 2024-08-06
[Hint Mode: Start Coords] Build the foundation for adding start coords UI for angle graphs

nishasy:
## Summary:
This PR includes the foundation/prerequisites for adding the start coords UI for angle graphs.

- Add the polygon graph type to start-coords-settings.tsx
- Create a WORK IN PROGRESS start-coords-angle.tsx file. This will be properly filled in
  in the following PR, whose main purpose will be adding the body for the UI.
- Add the start coords UI angle flag
- Add a storybook story for angle graphs (and update polygon story to start with start coords)
- Update the angle initializer to account for start coords

Issue: https://khanacademy.atlassian.net/browse/LEMS-2073

## Test plan:
`yarn jest`

Storybook
- http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-angle

image

### DIFF
--- a/.changeset/eleven-paws-serve.md
+++ b/.changeset/eleven-paws-serve.md
@@ -1,4 +1,5 @@
 ---
+"@khanacademy/perseus": minor
 "@khanacademy/perseus-editor": minor
 ---
 
--- a/.changeset/friendly-lemons-battle.md
+++ b/.changeset/friendly-lemons-battle.md
@@ -2,4 +2,4 @@
 "@khanacademy/perseus-editor": minor
 ---
 
-[Hint Mode: Start Coords] Add start coords UI for angle graphs
+[Hint Mode: Start Coords] Build the foundation for adding start coords UI for angle graphs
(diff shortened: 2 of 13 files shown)


nishasy:
Note: Because of my polygon story update, I realized that the "correct" preview's interactive graph doesn't show the right number of sides. This also led me to discover that the segment graph's "correct" preview with start coords doesn't show the right number of segments.

Fixing this in a separate PR.

Myranae (approved):
Looks good to me :) Checked out storybook and the reset to default coords button even does what I expected it to do. I love how you broke up the work like this. I need to get better at this as well :D

nishasy:
@Myranae Yeah, trying to figure out how to incorporate the angle in degrees is what inspired me to break this out into two PRs instead! I suspect that it may get somewhat complicated. It's temporarily just displaying the coordinates because that's what ultimately gets stored so that the graph knows how to render the angle.

PR #1494, opened 2024-08-06
Frameless Preview - 1: ContentRenderer

jeremywiebe:
## Summary:

This PR introduces a new component `ContentRenderer` (in the spirit of the original `IframeContentRenderer`). It previews Perseus content but no longer uses an `<iframe />` for preview. 

The IFrame preview system has several shortcomings that this series of PRs will address:

  1. The preview system is slow to load (it loads the entire webapp App Shell) including user authorization status.
  1. The preview system does not work in Storybook - we could likely get something working but it would involve copying some code from webapp over and then exposing a new http endpoint to serve up the preview page
  1. The preview system is complicated in that it uses `postMessage` to communicate between the host editor and the preview page. This code is split between Perseus and webapp and the contract is entirely implicit (there's no strong Typescript types or even any sort of code contract that defines what is passed between the host and iframe). 

image

And an example showing lint errors being displayed:

image

Issue: LEMS-1809

## Test plan:

`yarn start`
Navigate to the `ContentRenderer` story

### DIFF
--- a/dev/vite.config.ts
+++ b/dev/vite.config.ts
@@ -10,7 +10,41 @@ const packageAliases = {};
 glob.sync(resolve(__dirname, "../packages/*/package.json")).forEach(
     (packageJsonPath) => {
         const pkg = JSON.parse(fs.readFileSync(packageJsonPath, "utf-8"));
-        packageAliases[pkg.name] = join(dirname(packageJsonPath), pkg.source);
+
+        // "exports" is the more modern way to declare package exports. Some,
+        // but not all, Perseus packages delcare "exports".
+        if ("exports" in pkg) {
+            // Not all packages export strings, but for those that do we need
+            // to set up an alias so Vite properly resolves them.
+            // Eg `import {strings, mockStrings} from "@khanacademy/perseus/strings";`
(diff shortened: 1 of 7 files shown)


dev/vite.config.ts
```diff
@@ -10,7 +10,41 @@ const packageAliases = {};
 glob.sync(resolve(__dirname, "../packages/*/package.json")).forEach(
     (packageJsonPath) => {
         const pkg = JSON.parse(fs.readFileSync(packageJsonPath, "utf-8"));
-        packageAliases[pkg.name] = join(dirname(packageJsonPath), pkg.source);
+
```
jeremywiebe, line 13: This is a bit :( 

We need to build a set of aliases for Vite that mimic what we export from each package so that our cross-repo package dependencies work. When we go to production these are package dependencies like anything else, but within this monorepo we need to alias the package imports to the source files within this repo so that we don't have to rebuild to see changes. 

The `package.json` in each `packages/*` folder is still the source of truth, we're just using that data to build the aliases here.

jeremywiebe:
Closing - the challenge of matching styles is a dealbreaker (`@media` queries cannot be mocked/controlled without the iframe).

PR #1496, opened 2024-08-07
add lato to SB

handeyeco:
## Summary:

We use Lato at KA, but we weren't importing Lato into Storybook. I mostly just copied what they were doing in WB.


jeremywiebe (approved):
Cool.  I thought of the wonder-blocks while reviewing this and wondered what they do...pretty much exactly the same. But they don't package these fonts nor the `lato.css` file, as far as I can tell. So this feels like the right approach. 👍 

* https://github.com/Khan/wonder-blocks/blob/29ad5b48db78075861eb42103dc930b254a6a550/static/lato.css
* https://github.com/Khan/wonder-blocks/blob/29ad5b48db78075861eb42103dc930b254a6a550/.storybook/preview-head.html.

PR #1497, opened 2024-08-07
make the trig buttons real text so they can be translated

handeyeco:
## Summary:
It looks basically the same, but now it can be translated.

Previously "sin", "cos", and "tan" were SVGs.

Screenshot 2024-08-07 at 2 32 50 PM

Issue: LEMS-1621

## Test plan:


benchristel (approved):
Cool! I wonder why we were using SVGs for this...

LGTM.

PR #1502, opened 2024-08-08
Export keypad-context separately to avoid bundle bloat

handeyeco:
## Summary:
<Add your summary here>

Issue: LEMS-2246

## Test plan:


jeresig:
Great! So a question I have is that what's being exported here will be a different file than the one that's referenced internally inside Math Input. Is that a problem? It's possible that the actual context object may be different between this exported one and the "internal" one?

packages/keypad-context/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "source": "src/index.ts",
+    "files": [
+        "dist"
+    ],
+    "scripts": {
+        "test": "bash -c 'yarn --silent --cwd \"../..\" test ${@:0} $($([[ ${@: -1} = -* ]] || [[ ${@: -1} = bash ]]) && echo $PWD)'"
+    },
+    "dependencies": {},
+    "devDependencies": {},
+    "peerDependencies": {},
```
jeremywiebe, line 29: We should be listing 'react' here as a peer dep.

packages/keypad-context/src/keypad-context.tsx
```diff
@@ -32,7 +32,7 @@ export function StatefulKeypadContextProvider(props: Props) {
     // whether or not to display the keypad
     const [keypadActive, setKeypadActive] = useState<boolean>(false);
     // used to communicate between the keypad and the Renderer
-    const [keypadElement, setKeypadElement] = useState<KeypadAPI | null>();
+    const [keypadElement, setKeypadElement] = useState<any>();
```
jeremywiebe, line 35: 😬  `any` ?

packages/keypad-context/types
```diff
@@ -0,0 +1 @@
+../../types/
```
jeremywiebe, line 1: Github doesn't show this well, but wanted to confirm this is a symlink.

packages/math-input/src/components/__tests__/integration.test.tsx
```diff
@@ -1,3 +1,9 @@
+import {
+    KeypadContext,
+    StatefulKeypadContextProvider,
+    // @ts-expect-error it can't find type declaration
+    // eslint-disable-next-line monorepo/no-internal-import
```
jeremywiebe: Is this error still happening?

jeremywiebe (approved):
Let's hope this works! :)

packages/keypad-context/src/keypad-context.tsx
handeyeco: Yeah. I tried to move types from `@khanacademy/math-input` and things quickly spiraled. I think short-term it's good to have this separate package, but long-term we need to find a better solution.

PR #1503, opened 2024-08-08
Add "static" toggle to Interactive Graph widget

benchristel:
## Summary:
Some of our widgets (e.g. Radio) have a "static" mode which can be toggled on
in the exercise editor. When a widget is in static mode, the "answer" to the
widget is displayed, the widget is not graded, and the learner cannot alter
the widget's state.

Content creators need to be able to set interactive graphs to static mode so
that graphs can be added to hints, to demonstrate the solution steps for an
exercise. Previously, we implemented a special "hint mode" for interactive
graphs to accommodate this use case. Hint mode was based on a couple faulty
assumptions:

- We did not clearly understand what static mode was or when it was turned on.
  We confused static mode with the "read only" state of a widget, which occurs
  when a learner has successfully answered the question. Similar to static mode,
  the read-only state makes the widget non-interactable. However, while static
  mode is configurable by content creators, read-only state is not; it is set
  or unset dynamically, based on the learner's progress through an exercise.
- We assumed that graphs in hints would always be non-interactable, so we
  turned on hint mode automatically in hints (and turned it off in question
  stems) rather than giving content creators a manual toggle. This assumption
  turned out to be incorrect; we now know of at least one question where a hint
  contains a graph that the learner is supposed to be able to manipulate. See:
  https://khanacademy.slack.com/archives/C067UM1QAR4/p1723045477082029

Hint mode (which this feature replaces) was never released to content creators,
so we are okay to drop support for it.

This PR removes hint mode and adds a "static" toggle to the Interactive
Graph widget editor. It builds on work that @nicolecomputer did in
https://github.com/Khan/perseus/pull/1457. Her feature branch is merged into
this one.

Issue: LEMS-2251, LEMS-2053, LEMS-2054

## Test plan:

- Deploy this to a ZND
- Create an exercise in the Test Everything course
- Create a static interactive graph
- Verify that the preview is non-interactable (with mouse and keyboard) when
  the "static" toggle is on, and interactable when "static" is off.
- Save and publish the exercise.
- Go through the exercise as a learner and verify that you cannot manipulate
  the static graphs.
- Verify that the presence of a static graph does not affect how a question is
  graded.

### DIFF
--- a/packages/perseus/src/__tests__/renderer.test.tsx
+++ b/packages/perseus/src/__tests__/renderer.test.tsx
@@ -1352,30 +1352,6 @@ describe("renderer", () => {
             expect(el).not.toBeNull();
         });
 
-        it("should force the widget to be non-static if it has a problem number", () => {
-            // Arrange/Act
-            const {renderer} = renderQuestion(
-                {
-                    ...question1,
-                    widgets: {
-                        ...question1.widgets,
-                        "dropdown 1": {
-                            ...question1.widgets["dropdown 1"],
-                            static: true,
-                        },
-                    },
-                },
-                {},
-                {problemNum: 1},
-            );
-
-            // Assert
-            const [dropdownWidget] = renderer.findWidgets("dropdown 1");
-
-            // Act
-            expect(dropdownWidget.props.static).toBe(false);
-        });
-
         it("should ask each widget to show rationales", () => {
             // Arrange
             const {renderer} = renderQuestion(mockedItem);
--- a/packages/perseus/src/widgets/__testdata__/interactive-graph.testdata.ts
+++ b/packages/perseus/src/widgets/__testdata__/interactive-graph.testdata.ts
@@ -809,25 +809,8 @@ export const quadraticQuestion: PerseusRenderer =
 export const quadraticQuestionWithDefaultCorrect: PerseusRenderer =
     interactiveGraphQuestionBuilder().withQuadratic().build();
 
-export const staticGraphQuestion: PerseusRenderer = interactiveGraphQuestionBuilder()
-    .addLockedPointAt(-7, -7)
-    .addLockedLine([-7, -5], [2, -3])
-    .addLockedVector([0, 0], [8, 2], "purple")
-    .addLockedEllipse([0, 5], [4, 2], {angle: Math.PI / 4, color: "blue"})
-    .addLockedPolygon(
-        [
-            [-9, 4],
-            [-6, 4],
-            [-6, 1],
-            [-9, 1],
-        ],
-        {color: "pink"},
-    )
-    .withStaticMode(true)
-    .build();
-
-export const staticGraphQuestionWithAnotherWidget: () => PerseusRenderer = () => {
-    const result = interactiveGraphQuestionBuilder()
+export const staticGraphQuestion: PerseusRenderer =
+    interactiveGraphQuestionBuilder()
         .addLockedPointAt(-7, -7)
         .addLockedLine([-7, -5], [2, -3])
         .addLockedVector([0, 0], [8, 2], "purple")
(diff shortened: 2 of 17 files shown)


Myranae:
Haven't finished properly reviewing, but I believe some of this code is code I also changed in my [bug fix PR](https://github.com/Khan/perseus/pull/1487). It's been approved, but I wanted to manually test before landing. Hoping to do that today, just ran into some local dev server / ZND issues. I'll totally update my PR if this one moves forward first though :)

benchristel:
@Myranae I'll wait for you to land your PR. I need to test this on a ZND before landing it anyway.

Myranae:
@benchristel It's landed :)

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -45,11 +45,11 @@ export type MafsGraphProps = {
     labels: InteractiveGraphProps["labels"];
     state: InteractiveGraphState;
     dispatch: React.Dispatch<InteractiveGraphAction>;
-    hintMode: boolean;
+    static: boolean | null | undefined;
```
jeremywiebe, line 48: I'm not sure if types from other uses of `static` force this, but can we make this `static: boolean | undefined`?  I think there are many places that use this `X | null | undefined` pattern, but in reality, we rarely distinguish between `null` and `undefined`. 

This is a nitpicky thing, but it's another place where the type is broader/more complicated for no reason.

packages/perseus/src/widgets/interactive-graphs/reducer/use-graph-config.ts
```diff
@@ -15,7 +15,11 @@ export type GraphConfig = {
     width: number; // pixels
     height: number; // pixels
     labels: readonly string[];
-    hintMode?: boolean;
+    // The name isStatic (instead of static) avoids a collision with the
+    // `static` keyword in ES6. `static` is not a valid variable name, so
+    // naming this `isStatic` allows the object to be destructured without
+    // renaming the resulting variables.
+    isStatic: boolean;
```
jeremywiebe, line 22: Guess we'll want to avoid having a Perseus concept named "const" then, eh? 😂

jeremywiebe (approved):
Cool.  I was initially surprised by the size of the PR (number of files touched) but most of that is removing `hintMode`.  I'm pretty happy we can use `static` mode and not introduce another new concept (assuming it all works out). Thanks Ben!

Myranae:
Doing some manual testing on this, I noticed the hints-renderer stories related to interactive graph hint mode don't work anymore because hint mode isn't a thing. We may want to delete all the interactive graph stories from the Hints Renderer stories since you now have static graph stories and graphs shouldn't automatically be turning non-interactive when copied to the hints section. [This story](http://localhost:6006/?path=/story/perseus-renderers-hints-renderer--with-all-interactive-graphs) and all the interactive graph ones under it I mean.

benchristel:
@jeremywiebe @Myranae I realized I missed a crucial part of this feature, which is displaying the correct answer instead of the start coords when the graph is in static mode. So I'll have to ask for another review.

I'd really like Nicole's eyes on this when she's well again, because I suspect she's already done the research to figure out how displaying the correct answer is supposed to work. I've got a solution, but I'm not sure it follows established patterns.

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
benchristel: The type `boolean | null | undefined` comes from [WidgetProps](https://github.com/Khan/perseus/blob/81db6a703677a3447da04ca05ee1707b7f97a851/packages/perseus/src/types.ts#L582), so I had to use the same type here.

I don't really feel comfortable changing WidgetProps, since it's so widely used (and it's one of our few defenses against writing widgets that can't handle production data).

packages/perseus/src/__tests__/renderer.test.tsx
```diff
@@ -1352,30 +1352,6 @@ describe("renderer", () => {
             expect(el).not.toBeNull();
         });
 
-        it("should force the widget to be non-static if it has a problem number", () => {
```
jeremywiebe, line 1355: Really glad we don't need to further commit to this hack. 😅

packages/perseus/src/widgets/__testdata__/interactive-graph.testdata.ts
```diff
@@ -808,3 +808,91 @@ export const quadraticQuestion: PerseusRenderer =
...
+                    [-6, 1],
+                    [-9, 1],
+                ],
+                {color: "pink"},
+            )
+            .withStaticMode(true)
+            .build();
+        result["widgets"] = {
+            ...result["widgets"],
+            "radio 1": {
```
jeremywiebe, line 853: Hrm... seems like we now we need a `radioQuestionBuilder()` 😃

packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.tsx
```diff
@@ -128,5 +129,15 @@ export const StatefulMafsGraph = React.forwardRef<
         graph.startCoords,
     ]);
 
+    if (props.static) {
```
jeremywiebe, line 132: This becomes obvious fairly quickly as I read the code, but could you leave a comment here that if we're in `static` mode we display the correct answer? (unless you don't feel that's necessary)

jeremywiebe (approved):
This is great. LGTM!

PR #1505, opened 2024-08-08
[Interactive Graph: test utils] Do not set the coords field by default in the builder

nishasy:
## Summary:
The builder should not be setting the `coords` inside the graph's `correct` field by default.
The correct answer `coords` get set after the content authors interact with the graph's
"correct answer" preview.

When the coords are undefined, the graph renders with the default starting coords. If we were
to pass in the coords explicitly in the builder, it would (a) not be an accurate representation
of the initial state of a graph, and (b) we would have to make sure we use the initialization
logic to confirm that we're setting coords based on the number of segments, sides, or points
for segment graphs, polygon graphs, and point graphs respectively.

Updating the coords to be undefined by default inside the builder also fixes an issue where
the "correct answer" preview was showing the wrong graph values because the values were
set to the wrong default values in the builder.

Issue: none

## Test plan:
`yarn jest packages/perseus/src/widgets/interactive-graphs/interactive-graph-question-builder.test.ts`

Storybook
- Segments
  - http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-segments
- Polygons
  - http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-polygon

## Screenshots

| Before | After |
| --- | --- |
| Screenshot 2024-08-08 at 4 26 16 PM | Screenshot 2024-08-08 at 4 26 22 PM |
| Screenshot 2024-08-08 at 4 26 47 PM | Screenshot 2024-08-08 at 4 26 53 PM |

### DIFF
--- a/.changeset/wise-lamps-tell.md
+++ b/.changeset/wise-lamps-tell.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+[Interactive Graph: test utils] Do not set the coords field by default in the builder
--- a/packages/perseus/src/widgets/interactive-graphs/interactive-graph-question-builder.test.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/interactive-graph-question-builder.test.ts
@@ -145,7 +145,6 @@ describe("InteractiveGraphQuestionBuilder", () => {
                 correct: {
                     type: "segment",
                     numSegments: 1,
-                    coords: [expect.anything()],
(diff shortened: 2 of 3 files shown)


benchristel (approved):
Thanks! I just ran into a situation yesterday where `correct.coords` was undefined, and it surprised me. This clarifies things. LGTM.

PR #1506, opened 2024-08-09
Remove .babelrc.js files in package dirs that are no longer needed (Jest is bugfixed)

jeremywiebe:
## Summary:

When we extracted Perseus to its own repo, we had to put a small `.babelrc.js` file in some package folders so that Jest would work (https://github.com/facebook/jest/issues/11741). We are now on a version of Jest that is fixed and so no longer need these files (we do have a root babel config that covers everything). 

Issue: "none"

## Test plan:

`yarn test` ✅

### DIFF
--- a/packages/kmath/.babelrc.js
+++ b/packages/kmath/.babelrc.js
@@ -1,8 +1,0 @@
-/**
- * HACK(somewhatabstract): Due to https://github.com/facebook/jest/issues/11741,
- * we need to have this file, or updating inline snapshots can fail rather
- * cryptically.
- *
- * We should remove this when jest is fixed.
- */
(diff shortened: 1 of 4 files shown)


jeremywiebe:
Shoot. I tested further and discovered that the flow that was broken in Jest (editing a test in a `.tsx` file which uses an inline snapshot cannot be automatically updated by Jest) is indeed fixed, but breaks again because [Jest doesn't support `prettier` v3](https://jestjs.io/docs/configuration/#prettierpath-string) (see also the [issue](https://github.com/jestjs/jest/issues/14305)). Prettier is used to format inline snapshots. 

Closing this PR for now.

PR #1511, opened 2024-08-12
Lint against static widgets in question stems

benchristel:
We have logic in the Perseus renderer that prevents widgets from being static
in question stems. This logic was put in place to prevent content creators from
accidentally creating exercises that couldn't be answered (because all their
widgets were static). However the implementation is pretty hacky (it checks
problemNum to know if the renderer is in a question stem), and doesn't work in
the exercise editor preview, or in Perseus storybook. The resulting behavior is
very surprising: `static` widgets in question stems are non-interactive in the
exercise editor but are interactive in the learner experience.

Because this is so confusing, we want to remove the special-case logic and let
`static` always do what it says on the tin. But we still want to ensure that
unanswerable exercises aren't published. To that end, this commit adds a linter
that will do that.

Issue: LEMS-2251

## Test plan:

Deploy a ZND and edit an exercise.

- Add a static widget to a question stem; you should get a linter warning.
- Add a static widget to a hint; you should not get a linter warning.
- Add a non-static widget to a question; you should not get a linter warning.
- Add a static widget to an article; you should not get a linter warning.

Note that only Radio widgets can be made static, currently.

Screen Shot 2024-08-13 at 2 40 06 PM

### DIFF
--- a/packages/perseus-linter/src/__tests__/rules.test.ts
+++ b/packages/perseus-linter/src/__tests__/rules.test.ts
@@ -540,9 +540,7 @@ describe("Individual lint rules tests", () => {
             stack: ["hint"],
             widgets: {
                 "radio 1": {
-                    options: {
-                        static: true,
-                    },
+                    static: true,
                 },
             },
         })
@@ -556,9 +554,7 @@ describe("Individual lint rules tests", () => {
             stack: [],
             widgets: {
                 "radio 1": {
-                    options: {
-                        static: true,
-                    },
+                    static: true,
                 },
             },
         })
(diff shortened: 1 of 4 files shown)


packages/perseus-linter/src/__tests__/rules.test.ts
```diff
@@ -532,4 +533,79 @@ describe("Individual lint rules tests", () => {
...
+                "radio 1": {
+                    options: {
+                        static: true,
+                    },
+                },
+            },
+        })
+
+        expect(problems).toBe(null)
+    });
```
benchristel, line 551: I originally wrote these tests using the `expectPass` etc. helper functions defined in this file. Then I realized that Jest wasn't running all the tests (or maybe Goland just wasn't displaying all the results?) because some of them had duplicate titles. In an ideal world, I think this file would be table tests.

jeremywiebe (approved):
This looks good. Do we need to register this run anywhere to make it active though?

PR #1513, opened 2024-08-13
[SPIKE] Perseus Preview - styling overrides

jeremywiebe:
This PR illustrates an approach to the iframe-less preview using CSS variables and `var()` with defaults to provide overrides. 

At this point I don't feel it's the right approach as it "pollutes" the CSS quite a bit and is finicky to keep correct (ie. it's easy to introduce a style in a `@media` query and forget to provide an override. Finally, given that CSS/Styling in Perseus is tricky to test, this feels like a non-maintainable solution. 

I'm putting it up here so that I can build a ZND for some folks to look at, but am leaning towards different solutions.


jeremywiebe:
Closing. We definitely aren't pursuing this approach. See: https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/3159195681/Perseus+Preview+System

PR #1515, opened 2024-08-14
[Start Coords] Implement start coords UI for angle graphs

nishasy:
## Summary:
Implement the body of the start coords UI for angle graphs (replace the "WIP" placeholder).

The implementation for `getAngleEquation()` util function is taken from [interactive-graph.tsx](https://github.com/Khan/perseus/blob/f083200340545f41275f0696dbfc967f45028b0c/packages/perseus/src/widgets/interactive-graph.tsx#L2354)
and rewritten to work here without having to import the InteractiveGraphWidget component from
perseus. Using this custom implementation, we can pass in just the coordinates to get our equation
instead of having to pass in all the graph props in the original function, which leads to a whole 
slew of TypeScript errors.

There's also the consideration of adding an "angle" field to allow content authors to edit the angle
directly, but this adds a lot of complexity. If this feature is requested in the future, we can add it then.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2073

## Test plan:
`yarn jest packages/perseus-editor/src/components/__tests__/start-coords-settings.test.tsx`
`yarn jest packages/perseus-editor/src/components/__tests__/util.test.ts`

Storybook
http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-angle

Screenshot 2024-08-13 at 5 12 05 PM

### DIFF
--- a/.changeset/lovely-crabs-talk.md
+++ b/.changeset/lovely-crabs-talk.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": minor
+---
+
+[Start Coords] Implement start coords UI for angle graphs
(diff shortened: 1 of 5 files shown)


benchristel (approved):
LGTM!

PR #1517, opened 2024-08-14
Replace the 'Set as static' button with a switch

benchristel:
## Summary:
We discussed this [in slack][1], and Charlie thought it was a good idea.

[1]: https://khanacademy.slack.com/archives/C067UM1QAR4/p1723576041577949?thread_ts=1723487579.845949&cid=C067UM1QAR4

Issue: none

## Test plan:

Run `yarn storybook`

View: http://localhost:6006/?path=/docs/perseuseditor-widgets-interactive-graph--docs

Verify that you can toggle the "Static" switch. The graph in the preview should
be interactive when the switch is turned off, and should be non-interactive
(and display the answer) when the switch is on.
Screenshot of Perseus Storybook

### DIFF
--- a/.changeset/stupid-steaks-flow.md
+++ b/.changeset/stupid-steaks-flow.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": minor
+---
+
+Replace the "(un)set as static" button in the widget editor with a toggle switch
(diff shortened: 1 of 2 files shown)


jeremywiebe (approved):
Nice. That feels like a small change that'll make things so much clearer!

PR #1521, opened 2024-08-14
Perseus Editor: Add ContentPreview component

jeremywiebe:
## Summary:

This PR introduces a new editor component: `ContentPreview`. It is a thin wrapper around `ServerItemRenderer` which provides an easy way for editors to preview a given Perseus Item. 

It should be noted that due to current Perseus styling, the preview will _not_ match production 100%. This is because Perseus styles are adjusted based `@media` queries on device width.

image

Issue: LEMS-1809

## Test plan:

View the prevew in Storybook. `yarn start` ==> http://localhost:6006/?path=/docs/perseuseditor-content-preview--docs

### DIFF
--- a/dev/vite.config.ts
+++ b/dev/vite.config.ts
@@ -12,7 +12,7 @@ glob.sync(resolve(__dirname, "../packages/*/package.json")).forEach(
         const pkg = JSON.parse(fs.readFileSync(packageJsonPath, "utf-8"));
 
         // "exports" is the more modern way to declare package exports. Some,
-        // but not all, Perseus packages delcare "exports".
+        // but not all, Perseus packages declare "exports".
         if ("exports" in pkg) {
             // Not all packages export strings, but for those that do we need
             // to set up an alias so Vite properly resolves them.
--- a/dev/vite.config.ts
+++ b/dev/vite.config.ts
@@ -10,7 +10,41 @@ const packageAliases = {};
 glob.sync(resolve(__dirname, "../packages/*/package.json")).forEach(
     (packageJsonPath) => {
         const pkg = JSON.parse(fs.readFileSync(packageJsonPath, "utf-8"));
-        packageAliases[pkg.name] = join(dirname(packageJsonPath), pkg.source);
+
+        // "exports" is the more modern way to declare package exports. Some,
+        // but not all, Perseus packages delcare "exports".
+        if ("exports" in pkg) {
+            // Not all packages export strings, but for those that do we need
+            // to set up an alias so Vite properly resolves them.
+            // Eg `import {strings, mockStrings} from "@khanacademy/perseus/strings";`
+            // And MOST IMPORTANTLY, this alias _must_ precede the main
+            // import, otherwise Vite will just use the main export and tack
+            // `/strings` onto the end, resulting in a path like this:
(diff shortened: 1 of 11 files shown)


dev/vite.config.ts
```diff
@@ -10,7 +10,41 @@ const packageAliases = {};
...
+            // import, otherwise Vite will just use the main export and tack
+            // `/strings` onto the end, resulting in a path like this:
+            // `packages/perseus/src/index.ts/strings`
+            const stringsSource = pkg.exports["./strings"]?.source;
+            if (stringsSource != null) {
+                packageAliases[`${pkg.name}/strings`] = join(
+                    dirname(packageJsonPath),
+                    stringsSource,
+                );
+            }
```
jeremywiebe, line 30: This is necessary so we can import strings from other packages in this repo in Storybook correctly.  We basically make an alias for Vite that mimics the `exports` key in the package's `package.json` file. 

This fixes the following import in `perseus-editor` to work as you'd expect based on package [exports](https://github.com/Khan/perseus/blob/main/packages/perseus/package.json#L28):

```
import {mockStrings} from "@khanacademy/perseus/strings";
```

packages/perseus/src/components/lint.tsx
```diff
@@ -1,10 +1,11 @@
 import {StyleSheet, css} from "aphrodite";
 import * as React from "react";
-import ReactDOM from "react-dom";
+import Tooltip from "@khanacademy/wonder-blocks-tooltip";
```
jeremywiebe, line 3: I needed to migrate to the WB tooltip as the old one was getting clipped by the preview element(s).  Moving to WB also simplifies this component quite a bit.

packages/perseus/src/components/lint.tsx
```diff
@@ -64,43 +61,12 @@ type State = {
  * that has a right margin (like anything blockquoted) the circle will appear
  * to the left of where it belongs.  And if there is more
  **/
-class Lint extends React.Component<Props, State> {
+class Lint extends React.Component<Props> {
```
mpolyak, line 64: I love that we can improve/simplify the tooltip code for both the existing iframe and the new preview component.

dev/vite.config.ts
```diff
@@ -10,7 +10,41 @@ const packageAliases = {};
 glob.sync(resolve(__dirname, "../packages/*/package.json")).forEach(
     (packageJsonPath) => {
         const pkg = JSON.parse(fs.readFileSync(packageJsonPath, "utf-8"));
-        packageAliases[pkg.name] = join(dirname(packageJsonPath), pkg.source);
+
+        // "exports" is the more modern way to declare package exports. Some,
+        // but not all, Perseus packages delcare "exports".
```
mpolyak, line 15:
```suggestion
        // but not all, Perseus packages declare "exports".
```

mpolyak (approved):
I know that this is not what was originally hoped for, but this is still something that we can leverage in the editors to offer huge improvements in the future.

Thank you very much Jeremy.

jeremywiebe:
Given that this is mostly for CP's upcoming Exercise Editor v2 work, I'm going to land this. We can improve/fix as needed as CP starts using it.

PR #1523, opened 2024-08-14
[Interactive Graph] Stop the Mafs graphs from being user selectable

nishasy:
## Summary:
Currently, graphs can be selected and copied when the user drag-selects them.
We don't want this to be the case. It causes a bunch of graph information to
be awkwardly copied, such as the background image or all the axis tick labels.

Here, I'm adding `user-select: none` to the Mafs graph to stop this behavior.

NOTE: This works perfectly in Safari and Firefox, but there is a known bug
in Chrome that makes it so that the drag-selected element (which does
successfully look non-selected) still gets copied to the clipboard.
As this is a browser-specific issue, I did not try to fix it here. I think
there might be a super roundabout way to to this, like blocking mouse events
on specific parts of the graph, but I'm not sure that's worth it for a
behavior like this that's out of our control.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2151

## Test plan:
Storybook
- For tick labels
  - http://localhost:6006/?path=/story/perseus-widgets-interactive-graph--segment-with-mafs-and-locked-points
- For background image
  - http://localhost:6006/?path=/story/perseus-widgets-interactive-graph--angle-with-mafs

- Try to copy the graph (start dragging from any text above it to make it easier)
- Paste into a notes app. Confirm that the graph info doesn't paste. (Except in Chrome)

### Before

https://github.com/user-attachments/assets/0baced92-e0b5-4df3-9198-2e8ae6439370

### After

https://github.com/user-attachments/assets/92c55f1e-2e91-4c43-b97f-9852317c0576

### DIFF
--- a/.changeset/unlucky-carrots-sparkle.md
+++ b/.changeset/unlucky-carrots-sparkle.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+[Interactive Graph] Stop the Mafs graphs from being user selectable
--- a/packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
@@ -80,6 +80,7 @@ export const MafsGraph = (props: MafsGraphProps) => {
                     marginLeft: "20px",
                     marginBottom: "20px",
                     pointerEvents: props.static ? "none" : "auto",
+                    userSelect: "none",
                 }}
             >
                 <LegacyGrid


benchristel (approved):
This looks good to me. But I'm a little confused about why it's not working in Chrome. I just tried setting `user-select: none` on a random element in a relatively simple HTML file (for the record, it was an `<li>` in the [Perseus changelog](https://github.com/Khan/perseus/blob/main/packages/perseus/CHANGELOG.md)) and got the expected behavior in Chrome 127: the text of the element can't be copied, but the text before and after it can be. So it seems like `user-select: none` works in Chrome in at least some cases. I'd like to understand exactly when it doesn't work a bit better. Maybe it has something to do with SVGs?

Still, this change is an improvement, so let's ship it!

nishasy:
@benchristel I tested it manually on the [MDN user-select docs](https://developer.mozilla.org/en-US/docs/Web/CSS/user-select#result).

Under the section result section, where it says
> You should be able to select this text.
Hey, you can't select this text!
Clicking once will select all of this text.

I was able to select the middle line when copying from Chrome, but not with Safari or Firefox.

I wonder if there's some sort of special case with `<li>`? Or maybe it's just inconsistent...

benchristel:
@nishasy interesting... I just tried the example on that page in Chrome 127 and Edge 126, and it works correctly in both browsers. What version of Chrome do you have?

nishasy:
@benchristel Huh...

My Chrome version says `Version 126.0.6478.183 (Official Build) (arm64)`. Maybe I need to update?

If it works correctly on all the browsers, that would be amazing.

Updating now. I'll let you know the result.

nishasy:
@benchristel Okay now I have `Version 127.0.6533.120 (Official Build) (arm64)`

It still looks buggy for me.

https://github.com/user-attachments/assets/0fdcac31-b38e-44e5-a079-7d2bfacef300

nishasy:
I'm going to land this, but it would be nice to understand why it's inconsistent for different people at some point.

PR #1507, opened 2024-08-09
Add tests for mixed number in KAS

handeyeco:
## Summary:

This is a failed attempt at [LEMS-2198](https://khanacademy.atlassian.net/browse/LEMS-2198), but I want to keep the tests to help the next person who tries.

[LEMS-2198]: https://khanacademy.atlassian.net/browse/LEMS-2198?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1524, opened 2024-08-15
Fix broken KAS expect

handeyeco:
## Summary:
While working on [this PR](https://github.com/Khan/perseus/pull/1507), I noticed tests were passing when they shouldn't have been.

That PR might be a dead end, so I wanted to submit a fix in a separate PR.

### DIFF
--- a/packages/kas/src/__tests__/evaluating.test.ts
+++ b/packages/kas/src/__tests__/evaluating.test.ts
@@ -20,8 +20,9 @@ expect.extend({
 
         if (actual !== expected) {
             return {
-                pass: actual !== expected,
-                message: () => `${input} evaluates as ${expected}`,
+                pass: false,
+                message: () =>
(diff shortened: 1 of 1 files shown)


mark-fitzgerald (approved):
Doh! Great catch!

packages/kas/src/__tests__/evaluating.test.ts
```diff
@@ -44,6 +45,13 @@ declare global {
 }
 
 describe("evaluating", () => {
+    // Due to a bug in `toEvaluateAs`, all tests were passing
+    // whether or not they should have been.
+    // This is to make sure we don't regress again.
+    test.failing("should be possible to fail", () => {
```
jeremywiebe, line 51: Cool!  TIL: https://jestjs.io/docs/api#testfailingname-fn-timeout

PR #1525, opened 2024-08-15
PerseusItem: split and clarify top-level types

jeremywiebe:
## Summary:

In Perseus, the type `PerseusItem` represents the top-level data structure that the two main Perseus question renderers accept: `ServerItemRenderer` and `MultiRenderer`. 

This PR formalizes the fact that, although we had a single type for both, they are really two completely separate types. For now, I've retained the concept of the `PerseusItem` type, but in reality, most (all?) places should be using the `StandardItem` for `ServerItemRenderer` and `MultiItem` for `MultiRenderer`.

Issue: "none"

## Test plan:

`yarn tsc`

### DIFF
--- a/packages/perseus/src/perseus-types.ts
+++ b/packages/perseus/src/perseus-types.ts
@@ -84,15 +84,13 @@ export type PerseusWidgetsMap = {
 };
 
 /**
- * A PerseusItem is the top-level Perseus data type.
- */
-export type PerseusItem = StandardItem | MultiItem;
-
-/**
- * A "StandardItem" is a classic Perseus item. It is rendered by the
+ * A "PerseusItem" is a classic Perseus item. It is rendered by the
  * `ServerItemRenderer` and the layout is pre-set.
+ *
+ * To render more complex Perseus items, see the `Item` type in the multi item
+ * area.
  */
-export type StandardItem = {
+export type PerseusItem = {
     // The details of the question being asked to the user.
     question: PerseusRenderer;
     // A collection of hints to be offered to the user that support answering the question.
--- a/packages/perseus/src/perseus-types.ts
+++ b/packages/perseus/src/perseus-types.ts
@@ -83,19 +83,36 @@ export type PerseusWidgetsMap = {
     [key in `video ${number}`]: VideoWidget;
 };
 
-export type PerseusItem = {
+/**
+ * A PerseusItem is the top-level Perseus data type.
+ */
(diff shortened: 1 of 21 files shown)


packages/perseus/src/perseus-types.ts
```diff
@@ -83,21 +83,38 @@ export type PerseusWidgetsMap = {
...
 
+/**
+ * A "MultiItem" is an advanced Perseus item. It is rendered by the
+ * `MultiRenderer` and you can control the layout of individual parts of the
+ * item.
+ */
+export type MultiItem = {
+    // Multi-item should only show up in Test Prep content and it is a variant of a PerseusItem
+    _multi: any;
+};
```
nishasy, line 116: **No op, just a question for myself:**

Since I don't see `MultiItem` being used anywhere in this PR, I assume the main place this is used would be in `webapp` as an imported type, right?

nishasy: Would it ever be possible to fully add types for this?

nishasy (approved):
🎉 You may still want to get thoughts from someone who knows Perseus better than I do, but LGTM.

packages/perseus/src/util/fix-passage-refs.ts
```diff
@@ -101,8 +103,8 @@ const fixRendererPassageRefs = (options: any) => {
     return traverse(options, null, fixRadioWidget, fixWholeOptions);
 };
 
-const FixPassageRefs = (itemData: any): any => {
-    if (itemData._multi) {
+const fixPassageRefs = (itemData: PerseusItem): any => {
+    if ("_multi" in itemData) {
```
handeyeco, line 107: If `_multi` could be in `itemData`, should the type be `PerseusItem | MultiItem`?

handeyeco (approved):
Glad to see all this `_multi` removal! Some thoughts:

- Are we sure we still need MultiItems? Didn't we sunset SAT and whatnot?
- I think this would be a major version bump because it would require the consumer to update their code when upgrading versions if they have `_multi: null` on their PerseusItem objects.

packages/perseus/src/perseus-types.ts
jeremywiebe: Yes. But these types in this area are complicated and somewhat "stress test" the TypeScript engine.  We found that bringing that typing all the way up to here causes the TypeScript service to crash sometimes. I never figured out how to resolve that.

PR #1526, opened 2024-08-15
[Interactive Graph Locked Figures] Add controls to move a whole locked polygon

nishasy:
## Summary:
Right now, if someone wants to move an entire locked polygon, they’d have to
manually change every single point’s coordinate pair.

I added some arrow buttons to accomplish this.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2115

## Test plan:
`yarn jest packages/perseus-editor/src/components/__tests__/locked-polygon-settings.test.tsx`

Storybook
- http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--mafs-with-locked-figures-current
- Play around with the buttons and confirm that the polygon moves accordingly

## Screenshots & video:

NOTE: The arrows might look like they're too far to the right, but it doesn't look like this
in prod where the width of the settings is shortened to 306px by webapp styles.

Screenshot 2024-08-14 at 4 14 34 PM

https://github.com/user-attachments/assets/87b6a085-24a2-4085-90d4-6c0e11b97de3

### DIFF
--- a/packages/perseus-editor/src/components/__tests__/locked-polygon-settings.test.tsx
+++ b/packages/perseus-editor/src/components/__tests__/locked-polygon-settings.test.tsx
@@ -164,4 +164,140 @@ describe("LockedPolygonSettings", () => {
         // Assert
         expect(onToggle).toHaveBeenCalled();
     });
+
+    test("calls onChange when the whole polygon is moved up", async () => {
+        // Arrange
+        const onChangeSpy = jest.fn();
+        render(
+            <LockedPolygonSettings
+                {...defaultProps}
+                onChangeProps={onChangeSpy}
+                points={[
+                    [1, 1],
+                    [2, 1],
+                    [2, 2],
+                    [1, 2],
+                ]}
+            />,
+            {
+                wrapper: RenderStateRoot,
+            },
+        );
+
+        // Act
+        const moveUpButton = screen.getByLabelText("Move polygon up");
+        await userEvent.click(moveUpButton);
+
+        // Assert
+        expect(onChangeSpy).toHaveBeenCalledWith({
+            points: [
+                [1, 2],
+                [2, 2],
+                [2, 3],
+                [1, 3],
+            ],
+        });
+    });
+
+    test("calls onChange when the whole polygon is moved down", async () => {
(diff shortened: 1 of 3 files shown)


packages/perseus-editor/src/components/graph-locked-figures/locked-polygon-settings.tsx
```diff
@@ -54,6 +58,31 @@ const LockedPolygonSettings = (props: Props) => {
         onChangeProps({color: newValue});
     }
 
+    function handlePolygonMove(movement: "up" | "down" | "left" | "right") {
```
aemandine, line 61: [Non-blocking, newbie question]
This is a cool way to specify what string values the function should accept! Is this strategy preferred over an enum? (Maybe for the sake of simplicity, since it's only used once?) Thanks:)

aemandine (approved):
Hi Nisha! This is my first code review, so I would perhaps wait for someone else's  approval as well. That said, I paired with Third for this, and we think it looks great! I left two comments, one of which is just a question because I'm still learning TypeScript style. Thanks!

packages/perseus-editor/src/components/__tests__/locked-polygon-settings.test.tsx
```diff
@@ -164,4 +164,140 @@ describe("LockedPolygonSettings", () => {
...
+                    [1, 2],
+                ]}
+            />,
+            {
+                wrapper: RenderStateRoot,
+            },
+        );
+
+        // Act
+        const moveUpButton = screen.getByLabelText("Move polygon down");
```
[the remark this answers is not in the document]
nishasy: Oops! Yes 😅 I see I forgot that below as well.

packages/perseus-editor/src/components/graph-locked-figures/locked-polygon-settings.tsx
nishasy: As far as I can tell, enums and union types like this pretty much serve the same function. I do it like this since it's more consistent with the rest of the codebase. I'd be curious to hear what other people on the Perseus team think though.

PR #1529, opened 2024-08-15
Add unlimited points to graph

nicolecomputer:
## Summary:

Allows adding an unlimited number of points to a graph.

https://github.com/user-attachments/assets/9645c943-460f-4295-8696-12d2fcd50c6e

Issue: LEMS-1816

## Test plan:
- Run the tests


packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -499,6 +503,31 @@ function doChangeRange(
...
+): InteractiveGraphState {
+    if (state.type !== "point") {
+        return state;
+    }
+    const {snapStep} = state;
+    const snappedPoint = snap(snapStep, action.location);
+
+    // Check if there's already a point in that spot
+    for (const point of state.coords) {
+        if (point[0] === snappedPoint[0] && point[1] === snappedPoint[1]) {
```
benchristel, line 518: We have constants for the X and Y components of a vector, which might make this more self-explanatory:
```suggestion
        if (point[X] === snappedPoint[X] && point[Y] === snappedPoint[Y]) {
```

packages/perseus/src/widgets/interactive-graphs/types.ts
```diff
@@ -52,6 +52,7 @@ export interface LinearSystemGraphState extends InteractiveGraphStateCommon {
 export interface PointGraphState extends InteractiveGraphStateCommon {
     type: "point";
     coords: Coord[];
+    numPoints?: number | "unlimited";
```
benchristel, line 55: Since the number of points is given by `coords.length` for graphs with fixed numbers of points, could this property just be a boolean? E.g. `unlimitedPoints: boolean`?

In any case, I think this property shouldn't be optional, since it's always set in `initializeGraphState`.

packages/perseus/src/widgets/interactive-graphs/graphs/point.tsx
```diff
@@ -1,17 +1,61 @@
...
+                    x={left}
+                    y={top}
+                    onClick={(event) => {
+                        const elementRect =
+                            event.currentTarget.getBoundingClientRect();
+
+                        const x = event.clientX - elementRect.x;
+                        const y = event.clientY - elementRect.y;
+
+                        const graphCoordiantes = pixelsToVectors(
```
benchristel, line 49:
```suggestion
                        const graphCoordinates = pixelsToVectors(
```

packages/perseus/src/widgets/interactive-graphs/graphs/point.tsx
```diff
@@ -1,17 +1,61 @@
...
+    const width = maxX - minX;
+    const height = maxY - minY;
+    const [[widthPx, heightPx]] = useTransformDimensionsToPixels([
+        width,
+        height,
+    ]);
+    const [[left, top]] = useTransformVectorsToPixels([minX, maxY]);
     return (
         <>
+            {numPoints === "unlimited" && (
```
benchristel, line 33: It might be worth having a comment somewhere around here to make it clear what this mysterious `<rect>` is for :)

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -70,79 +72,106 @@ export const MafsGraph = (props: MafsGraphProps) => {
...
-                    backgroundImage={props.backgroundImage}
-                />
                 <View
+                    className="mafs-graph"
                     style={{
-                        position: "absolute",
-                        bottom: 0,
-                        left: 0,
+                        height,
+                        width,
```
benchristel, line 83: Do we need both this `width` and the one on the wrapper `View`, above? It seems like maybe we could get away with just one...

packages/perseus/src/widgets/interactive-graphs/graphs/point.tsx
benchristel, line 1: Most of the code in this file is now only used in the unlimited points case. That makes me think that maybe the unlimited point graph should be a separate component. What do you think?

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -70,79 +72,106 @@ export const MafsGraph = (props: MafsGraphProps) => {
...
+                            {renderGraph({
+                                state,
+                                dispatch,
+                            })}
+                        </Mafs>
+                    </View>
                 </View>
+                {props.state.type === "point" &&
+                    props.state.numPoints === "unlimited" &&
+                    renderPointGraphControls({state, dispatch})}
```
benchristel, line 151: This feels a bit like points-specific logic is leaking out into this component. Currently, `renderGraph` is the thing that knows what components to render for each graph type. Maybe `renderGraph` should return two things: the graph itself, and the controls.

benchristel (commented):
It's really cool to see this coming together! The behavior seems solid; I left some suggestions for structural improvements inline.

packages/perseus/src/widgets/interactive-graphs/graphs/point.tsx
nicolecomputer: I broke them out into two components. I think it's easier to read now

packages/perseus/src/widgets/interactive-graphs/types.ts
nicolecomputer: I think I'd prefer to keep numPoints. It matches the state we receive and I don't think there are going to be any more places where it would get tricky.

benchristel (approved):
Thanks for the updates! LGTM.

PR #1530, opened 2024-08-19
Make a SharedRendererProps type

handeyeco:
## Summary:
I don't know how useful this is, but this is my motivation: https://github.com/Khan/webapp/pull/24684

## Test plan:
Nothing should change but types


jeremywiebe (approved):
Thanks. I think eventually we would refactor that shared type away, but today it makes sense to have it for the preview system.

PR #1532, opened 2024-08-20
Add PhET simulation content editor

aemandine:
## Summary:
Create the content editor for PhET simulations, allowing users to set:
* The URL of the simulation 
* A description of the simulation (used to communicate the widget's purpose to screen readers and Khanmigo) 

I have documented the Perseus widget creation process [here](https://khanacademy.atlassian.net/wiki/spaces/LC/pages/3075276818/Creating+a+New+Widget+Type).

Issue: [LEMS-2168](https://khanacademy.atlassian.net/browse/LEMS-2168)

Editor page storybook page, displaying PhET widget, which contains an iframe displaying the Projectile Data Lab PhET simulation. The URL of the widget is the URL for Projectile Data Lab, and the description is 'Projectile Data Lab.'

Editor page storybook page, displaying PhET widget, which contains an iframe saying 'Sorry, this simulation cannot load.' The URL of the widget is the Google homepage URL, and the description is 'Google.'

## Test plan:

- Verify that the widget content editor shows up appropriately in Storybook's EditorPage (`/docs/perseuseditor-editorpage--docs`)
  - Set URL to a valid PhET simulation (like https://phet.colorado.edu/sims/html/projectile-data-lab/latest/projectile-data-lab_all.html) and verify that the simulation appears correctly
  - Change URL to a non-PhET origin and verify that the error shows up
  - Change URL to a non-existent resource from the PhET origin and verify that the error shows up
  - Click the fullscreen button and verify that the iframe goes into fullscreen mode
- `yarn jest packages/perseus-editor/src/widgets/__tests__/phet-sim-editor.test.tsx`

[LEMS-2168]: https://khanacademy.atlassian.net/browse/LEMS-2168?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/perseus-editor/src/widgets/phet-sim-editor.tsx
```diff
@@ -1,45 +1,55 @@
...
 import * as React from "react";
 
-import BlurInput from "../components/blur-input";
+import type {PerseusPhetSimWidgetOptions} from "@khanacademy/perseus/src/perseus-types";
 
-type PhetSimEditorProps = any;
+type Options = {
+    url: PerseusPhetSimWidgetOptions["url"];
+    description: PerseusPhetSimWidgetOptions["description"];
+};
```
aemandine, line 11: Two questions about this!

1. Is `Options` an okay name for this instead of `DefaultProps`? I want to use it as the return type of `this.serialize()` function which makes `DefaultProps` look like a bit of a misnomer.
2. Is this an okay way of setting the type for default props and serialize? If we just import `PerseusPhetSimWidgetOptions`, we get `static` as well, so this was our best strategy.

jeremywiebe: 1. The value returned by `serialize()` should really be a [`PerseusPhetSimWidgetOptions `](https://github.com/Khan/perseus/blob/phet-widget/packages/perseus/src/perseus-types.ts#L1544).  This data will be what's passed back to the `phet-sim`'s `transform` function and ultimately [provided](https://github.com/Khan/perseus/blob/phet-widget/packages/perseus/src/widgets/phet-sim.tsx#L20...L21) to the `PhetSim` React component.

I think it's fair to have a separate `DefaultProps` type which represents the default props for this editor widget. This could differ from the widget options and it is conceptually a very different type.

2. Yes, this is the conventional way we've defined default props (by using indexed properties into `Props`). As I said above, I would keep the `PerseusPhetSimWidgetOptions` type separate from `DefaultProps`. 

Is there a reason that `static` was added to `PerseusPhetSimWidgetOptions`? I don't think it's needed.

packages/perseus-editor/src/widgets/phet-sim-editor.tsx
```diff
@@ -1,45 +1,55 @@
...
+                        this.props.onChange({description})
+                    }
+                />
             </div>
         );
     }
 
-    serialize: () => any = () => {
-        return EditorJsonify.serialize.call(this);
+    serialize: () => Options = () => {
```
jeremywiebe, line 48: I realize you're likely following another editor's pattern, but I don't think this needs to be an arrow function. That's only needed if the class method is passed around and we want to ensure `this` stays bound.

```suggestion
    serialize(): PerseusPhetSimWidgetOptions {
```

packages/perseus/src/widgets/phet-sim.tsx
```diff
@@ -215,6 +215,6 @@ export default {
     displayName: "PhET Simulation",
     widget: PhetSim,
     // Let's not expose it to all content creators yet
-    hidden: true,
+    hidden: false,
```
jeremywiebe, line 218: Wooooohooo! 🎉

jeremywiebe (commented):
This looks great!  Before we deploy this we have to make sure that the publish pipeline is updated though. 😄 

I'll leave this unapproved just so Third and/or Tamara have a chance to look at it.

packages/perseus-editor/src/widgets/phet-sim-editor.tsx
aemandine: Huh! I guess I could leave `static` out. When I was first making the PerseusPhetSimWidgetOptions, I was operating off the iframe, and iframe and some other widgets had this at the end of their widget options (which did not see a use for, so I thought it was for something I did not understand):
```
// Always false
static: boolean;
```
But I'm now seeing that some of the other widgets don't have a `static`, so I guess it's not a hard and fast rule! I'll go ahead and remove `static` and change the `serialize()` return type to `PerseusPhetSimWidgetOptions`. Thank you so much!

P.S. Do you know why `static` is on some widgets but not others?

packages/perseus-editor/src/widgets/phet-sim-editor.tsx
aemandine: Oh, noted! Thanks. I'd been wondering what the difference was, and was just pattern-matching in the meantime. That makes sense! Thank you!

packages/perseus/src/widgets/phet-sim.tsx
aemandine: At risk of raining on everyone's parade, we still have a few minor design/a11y tickets to push through before totally releasing! But because we have the feature branch, so the code won't hit prod until everything's done, I think okay to un-hide! And release coming soon 👀

packages/perseus-editor/src/widgets/phet-sim-editor.tsx
jeremywiebe: The individual widget options should never have `static` on them. It is already [defined](https://github.com/Khan/perseus/blob/f083200340545f41275f0696dbfc967f45028b0c/packages/perseus/src/types.ts#L582) in the base `WidgetProps<T>` type which is what all widgets [receive](https://github.com/Khan/perseus/blob/f083200340545f41275f0696dbfc967f45028b0c/packages/perseus/src/renderer.tsx#L658) from the renderer.

packages/perseus/src/widgets/phet-sim.tsx
aemandine: Okay, re-hiding so we can double check it works on TestEverything before we fully launch!

SonicScrewdriver (approved):
We've been peer reviewing this during our Mentorship pairing sessions! Great job, let's get this moving! :)

PR #1535, opened 2024-08-21
wip: Add tests and rerender when showTooltips changes to update editor preview

Myranae:
## Summary:

--Unable to get the tooltip working from this. Eventually we will be updating the tool used for tooltip, which should fix this issue. Closing PR for now.--

Tooltips currently do not show up in the right hand side content editor preview window. This fixes that by inducing a rerender when showTooltips changes using useEffect.

(Waiting for npm snapshot to do manual testing and confirm issue fixed)

Before (hovering over point)
image

After
(in progress)

Issue: LEMS-2276

## Test plan:
- Confirm all tests pass
- Manually test with a webapp ZND

### DIFF
--- a/.changeset/witty-spies-learn.md
+++ b/.changeset/witty-spies-learn.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/perseus": patch
+"@khanacademy/perseus-editor": patch
+---
+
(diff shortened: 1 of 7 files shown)


Myranae:
Unable to get the tooltip working from this. Eventually we will be updating the tool used for tooltip, which should fix this issue. Closing PR for now.

PR #1538, opened 2024-08-21
Update Mathquill Mathfield with Portuguese trig functions

handeyeco:
## Summary:
So basically:
- These two PRs ([1](https://github.com/Khan/perseus/pull/921) and [2](https://github.com/Khan/webapp/pull/19111)) helped add support for Portuguese trig functions when typed in the input.
- This [PR](https://github.com/Khan/perseus/pull/1497) added support for the text of trig buttons to be translated, however the input would still update with "sin" if the "sen" button was pressed.
- The PR being reviewed updates the input correctly; so if the "sen" button is pressed, the input is updated with "sen"

How this works:
- When a button is pressed, the keypad yells into the void that "SIN" was pressed
- The keypad translator has a callback for updating the input when "SIN" is pressed
  - BEFORE: it would just updated the input with `sin()`
  - NOW: it looks for the translated version of "sin", checks it against known safe translations, and either updates it with a known safe translation or the default "sin"

Issue: LEMS-1621

## Test plan:
Once the strings are translated:
- Go to a Portuguese page with an expression widget
- Go to a question that requires the "sin" button
- Open the keypad
- Note the button is "sen" instead of "sin"
- Click the "sen" button
- Note the input is updated with "sen()"
- Fill out the correct answer and submit
- Note the validator accepts "sen" instead of "sin"

### DIFF
--- a/.changeset/orange-rules-knock.md
+++ b/.changeset/orange-rules-knock.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/math-input": patch
+"@khanacademy/perseus": patch
+---
+
+Use Portuguese sen and tg when updating Mathquill from the keypad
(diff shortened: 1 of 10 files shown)


benchristel (approved):
Cool! LGTM.

PR #1540, opened 2024-08-21
[Graphie] Adjust Scaling of Text to Render Properly in Mobile

mark-fitzgerald:
## Summary:
When a graphie image is displayed in a scaled-down environment (i.e. mobile viewport), SVG elements were scaled properly, but text was not. The issue was due to scaling not being applied to the text. This change adds in the calculations to properly scale and position text that exists in the image.

Issue: LEMS-2022

## Test plan:

1. Obtain a Graphie image.
   * From Graphie editor:
      * Open a Graphie editor for an existing image.
      * Click on the "Convert to image" button.
      * Wait for the image web address to be generated.
   * Or, just use this example image: <br/> web+graphie://ka-perseus-graphie.s3.amazonaws.com/2c7f2619eec00ab1992b4a491a0359542012b286.
2. Open an [editor page](https://www.khanacademy.org/devadmin/content/articles/test-article-for-the-test-course/xf02c39bb55c6da4a) (Storybook editor pages aren't working properly with Graphie images at the time of this PR).
3. Add an image widget.
4. Paste the image URL into the available entry field.
5. Compare the image preview using the available Desktop and Mobile toggles.
6. The text elements (i.e. point labels) should be scaled with the image (lines, etc.) and placed correctly.

### Before - bad text
Bad Text

### After - good text
Good Text


nishasy (approved):
I wouldn't say I have the expertise in this area to find any fault with your logic, so I'm just putting my full trust in all your testing and hard work. Thanks for hopping on this extremely tricky issue!

packages/perseus/src/components/svg-image.tsx
```diff
@@ -485,12 +485,22 @@ class SvgImage extends React.Component<Props, State> {
                 // without coordinates. They don't seem to have any content, so
                 // it seems fine to just ignore them (rather than error), but
                 // we should figure out why this is happening.
+
+                // 'fontSize' - When a default scale (1) is used,
+                //     setting the font size to 100% is redundant.
+                // Also, setting the font size to 100% can counteract other scale-based styling.
+                // Therefore, when the scale is not applicable,
+                //     we set the value to 'null' so that it won't be set to anything.
+                const fontSize =
```
jeremywiebe, line 494: I think we'd be slightly more clear calling this variable `style` as that's what the last parameter is (extra styles).  If we call this `fontSize` here, I could see folks being confused how to integrate other custom styles. 

Also, I realize this is ancient code that we probably won't touch much before deleting it (oh what a glorious day that would be)... but I think updating the variable name could still avoid future confusion.

packages/perseus/src/components/svg-image.tsx
```diff
@@ -485,12 +485,22 @@ class SvgImage extends React.Component<Props, State> {
                 // without coordinates. They don't seem to have any content, so
                 // it seems fine to just ignore them (rather than error), but
                 // we should figure out why this is happening.
+
+                // 'fontSize' - When a default scale (1) is used,
+                //     setting the font size to 100% is redundant.
+                // Also, setting the font size to 100% can counteract other scale-based styling.
+                // Therefore, when the scale is not applicable,
+                //     we set the value to 'null' so that it won't be set to anything.
```
jeremywiebe, line 493: Thanks for this very clear comment!

packages/perseus/src/util/graphie.ts
```diff
@@ -1675,10 +1675,34 @@ const setLabelMargins = function (span: HTMLElement, size: Coord): void {
             marginTop: -height / 2 - y * scale,
         });
     } else {
+        const $container = $span.closest(".svg-image");
+        // Ensuring that the line-height of the text doesn't throw off placement of the text element.
+        // Inherited line-height values can really mess up placement.
```
jeremywiebe, line 1680: How does setting it to `normal` help? (for a CSS newb)

packages/perseus/src/util/graphie.ts
```diff
@@ -1675,10 +1675,34 @@ const setLabelMargins = function (span: HTMLElement, size: Coord): void {
...
+        $container.css("line-height", "normal");
+
+        // The 'max-width' value is the same as the expected (100% scale) width of the graphie.
+        // We are using jQuery to get this information since we don't have a way to pass it to this function.
+        const expectedWidth = $container.css("max-width") ?? "0px";
+        // We can calculate the true scale of the graphie by taking actual width and dividing by the expected width.
+        // NOTE: Using 'slice' to remove the "px" unit from the end of the expected width value.
```
jeremywiebe, line 1687: Would this value ever be in a unit other than `px`?

packages/perseus/src/util/graphie.ts
```diff
@@ -1675,10 +1675,34 @@ const setLabelMargins = function (span: HTMLElement, size: Coord): void {
...
+        // "multipliers" basically move the position of the text by using 1, 0, -1
+        // 'margin-left' and 'margin-top' are used to position the text.
+        // Margin and font size need to be scaled accordingly.
         const multipliers = labelDirections[direction || "center"];
         $span.css({
-            marginLeft: Math.round(width * multipliers[0]),
-            marginTop: Math.round(height * multipliers[1]),
+            marginLeft: Math.round(width * multipliers[0] * scale) / 100,
+            marginTop: Math.round(height * multipliers[1] * scale) / 100,
+            fontSize: `${Math.round(scale * 100) / 100}%`,
```
jeremywiebe, line 1704: Do we need to elide `fontSize` here also (like in `svg-image.tsx`) for a scale of 1?

jeremywiebe (approved):
Amazing.  Hard to believe such a small amount of code for so much investigation and testing, but that's how it goes sometimes. I have some questions that might be best answered with some more comments. Thanks for fixing this very long-standing issue!!!

packages/perseus/src/util/graphie.ts
mark-fitzgerald: Most default `line-height` settings are 1.2. However, that value isn't appropriate for all fonts. So, setting this property to `"normal"` sets it to the preferred value for the specified `font-family`.

packages/perseus/src/util/graphie.ts
mark-fitzgerald: Great question! I can't say for sure (I would need to dive into the code some more to find it). I think what is happening with `max-width` is it is taking the measurement from the image and slapping a "px" on it - that's my guess.

jeremywiebe: I wonder if, instead of digging through, you could use an approach like this to only trim `px` if it exists: `expectedWidth.replace(/px$/, '');`?

Also, I suspect this is a nitpicky concern and likely is fine the way you have it. :)

packages/perseus/src/util/graphie.ts
mark-fitzgerald: We don't need to here, though I think that's a good idea to do it anyway.

The reason I did it in `svg-image.tsx` is that in some cases, it was obstructing a prior `font-size` setting (this one), and since a scale was rarely being passed to that component, the code needed to reflect that it should only be added when a non-full-size scale existed.

packages/perseus/src/util/graphie.ts
mark-fitzgerald: I like that! Thanks!!

catandthemachines (approved):

packages/perseus/src/components/svg-image.tsx
mark-fitzgerald: Done.

PR #1541, opened 2024-08-21
Add a finite point question to gallery

nicolecomputer:
## Summary:
Adds a finite point question so its easier to test infinite points vs. finite points.

Pasted_Image_8_21_24__1_26 PM

Issue: LEMS-1816

## Test plan:
- Visit the gallery
- **There should be a new question**


benchristel (approved):
LGTM! Thanks for adding this.

PR #1548, opened 2024-08-22
Fixing resize behavior for Expression Widget

catandthemachines:
## Summary:
Enabling rootBoundary mode for Popover implementation in Expression Widget. This ensures the popover container does not snap depending on the viewport and instead positions it's contend based on where there is room in the page document. This change is needed an intended to ensure the Expression Widget is accessible to users in 400% or very small screen sizes. 

Issue: LEMS-2085

## Test plan:

1. Viewing disabled viewport functionality.
   - Go to /?path=/docs/perseus-widgets-expression--docs and open an expression widget keypad.
   - Notice that the keypad does not move to stay in the viewport when scrolling up and down the document.
2. Viewing improved 400% zoom experience
    - Go to /iframe.html?args=&id=perseus-widgets-expression--desktop&viewMode=story
    - Increase screen zoom to 400% or minimize screen size to 320 x 256.
    - Notice that you can open the keypad and it does not move outside the page and you can scroll and reach content within the keypad even if it's larger the they viewport size.


handeyeco (approved):
Looks good to me. My only concerns:
- Failing TS check
- I wonder if we need to update all WB/WS at the same time?

catandthemachines:
> Looks good to me. My only concerns:
> 
> * Failing TS check
> * I wonder if we need to update all WB/WS at the same time?

Yeah the weird failing type check made me think I needed to update all of them at the same time. Will keep digging into why it's mis-behaving, and see if it's needed.

catandthemachines:
> > Looks good to me. My only concerns:
> > 
> > * Failing TS check
> > * I wonder if we need to update all WB/WS at the same time?
> 
> Yeah the weird failing type check made me think I needed to update all of them at the same time. Will keep digging into why it's mis-behaving, and see if it's needed.



> > Looks good to me. My only concerns:
> > 
> > * Failing TS check
> > * I wonder if we need to update all WB/WS at the same time?
> 
> Yeah the weird failing type check made me think I needed to update all of them at the same time. Will keep digging into why it's mis-behaving, and see if it's needed.



> > Looks good to me. My only concerns:
> > 
> > * Failing TS check
> > * I wonder if we need to update all WB/WS at the same time?
> 
> Yeah the weird failing type check made me think I needed to update all of them at the same time. Will keep digging into why it's mis-behaving, and see if it's needed.

Okay after wrangling with the typing I can summarize why it's there. The update to tooltip also appears to have included a dependency update to the color object Tooltip accepts. So to fix this I just needed to update the tokens package version. So the only packages updated are tooltip, popover, and tokens. 😊

PR #1552, opened 2024-08-24
[Perseus + Perseus Editor] Bump Wonder Blocks Form to 4.9.0 (new component: TextArea)

nishasy:
## Summary:
In order to get access to the new TextArea component, we need to update the
dependency on wonder-blocks-form.

Issue: none

## Test plan:
N/A


benchristel (approved):
LGTM!

PR #1554, opened 2024-08-27
Editor: Typing improvements

jeremywiebe:
## Summary:

This is a tiny PR that cleans up the `HintEditor`'s `render()` function by migrating to use an arrow function. This causes `this` to be bound properly and allows us to remove a bunch of `@ts-expect-error` annotations. 

It also removes some duplication around typing for the `images` key in `PerseusRenderer`.

Issue: LEMS-1809

## Test plan:

`yarn tsc`


benchristel (approved):
LGTM! Thanks!

PR #1555, opened 2024-08-27
HintEditor: define DefaultProps type

jeremywiebe:
## Summary:

This PR introduces an explicit DefaultProps type for the HintEditor and uses our conventional approach for the types (by using indexed access types)

Issue: LEMS-1809

## Test plan:

`yarn tsc`


packages/perseus-editor/src/hint-editor.tsx
```diff
@@ -45,6 +45,15 @@ type HintEditorProps = {
...
 };
 
+type DefaultHintEditorProps = {
+    className: HintEditorProps["className"];
+    content: HintEditorProps["content"];
+    replace: HintEditorProps["replace"];
+    showMoveButtons: HintEditorProps["showMoveButtons"];
+    showRemoveButton: HintEditorProps["showRemoveButton"];
+    showTitle: HintEditorProps["showTitle"];
+};
```
benchristel, line 55: Question: is this really the type that we want? These properties on `HintEditorProps` are optional and nullable. The corresponding default props should presumably not be nullable.

I'd prefer to keep the old type definition, with explicit types for the properties. The typechecker ensures that the duplicated types stay consistent.

benchristel (approved):
LGTM; left a comment inline.

packages/perseus-editor/src/hint-editor.tsx
jeremywiebe: I do like the visual connection to the Prop that is being defaulted. However, if I land this PR, I do need to tweak the main HintEditorProps to make the defaulted props non-undefined. Typescript seems smart enough to make them optional as soon as the same key exists in the `static defaultProps`. 

Further, it seems like we might not even need a type for it at all. Any prop that appears in `defaultProps` automatically makes the same-named prop in the component props optional. So maybe typing `defaultProps` is all just yak-shaving and not improving type safety at all.

packages/perseus-editor/src/hint-editor.tsx
```diff
@@ -30,11 +30,11 @@ type HintEditorProps = {
...
-    showMoveButtons?: boolean;
-    showRemoveButton?: boolean;
-    showTitle?: boolean;
-    content?: string | null | undefined;
-    replace?: boolean | null | undefined;
+    showMoveButtons: boolean;
+    showRemoveButton: boolean;
+    showTitle: boolean;
+    content: string | null;
+    replace: boolean | null;
```
jeremywiebe, line 37: Any prop that appears in `defaultProps` should be non-optional here as TypeScript will make it optional automatically.

PR #1556, opened 2024-08-27
Editor: stop passing 'replace' to Editor (that is a Hint property)

jeremywiebe:
## Summary:

Prior to this PR the `Editor` component accepted a `replace` property.  This prop is only relevant to hints and so I've removed it from the `Editor` and changed things so it's handled by the hint editor. This allows improved typing also.

Issue: LEMS-1809

## Test plan:

`yarn tsc`
`yarn test`


jeremywiebe:
The parent pull-request ([#1555](https://github.com/Khan/perseus/pull/1555)) has been merged into `feature/editor-preview-cleanup`, but this branch (`jer/editor-preview-3`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

packages/perseus-editor/src/hint-editor.tsx
```diff
@@ -89,7 +89,10 @@ export class HintEditor extends React.Component<HintEditorProps> {
     };
 
     serialize: (options?: any) => any = (options: any) => {
-        return this.editor.current?.serialize(options);
+        return {
+            ...this.editor.current!.serialize(options),
+            replace: this.props.replace ?? undefined,
```
benchristel, line 94: Why is `?? undefined` needed here?

benchristel (approved):
LGTM! Seems like you might need to change the base branch - there are commits from your other PRs in this one.

packages/perseus-editor/src/hint-editor.tsx
jeremywiebe: Because `this.props.replace` can be `null`, but a `Hint`'s `replace` is defined as `replace?: boolean;`. One of my gripes with so many places we have `X | null | undefined`. 

Is there a better way?

benchristel: No, this seems fine! Thanks for explaining.

PR #1557, opened 2024-08-27
Editor: use PerseusRenderer type for Props

jeremywiebe:
## Summary:

A small type fix. The `Editor` component receives the standard `PerseusRenderer` type, but that fact is obscured by it defining each piece of the `PerseusRenderer` type piecemeal. By changing the type to an intersection, it makes the connection clear.

Issue: LEMS-1809

## Test plan:

`yarn tsc`
`yarn test`


benchristel (approved):
LGTM!

PR #1558, opened 2024-08-27
HintEditor: migrate to React refs

jeremywiebe:
## Summary:

This PR migrates the HintEditor to use React refs instead of legacy string refs. This improves safety.

Issue: LEMS-1809

## Test plan:

`yarn tsc`
`yarn test`

### DIFF
--- a/packages/perseus-editor/src/hint-editor.tsx
+++ b/packages/perseus-editor/src/hint-editor.tsx
@@ -301,6 +301,7 @@ type CombinedHintsEditorProps = {
  */
 class CombinedHintsEditor extends React.Component<CombinedHintsEditorProps> {
     static HintEditor: typeof HintEditor = HintEditor;
+    hintEditors: Record<number, CombinedHintEditor | null> = {};
 
     static defaultProps: {
         highlightLint: boolean;
@@ -353,9 +354,7 @@ class CombinedHintsEditor extends React.Component<CombinedHintsEditorProps> {
         const hint = hints.splice(i, 1)[0];
         hints.splice(i + dir, 0, hint);
         this.props.onChange({hints: hints}, () => {
-            // eslint-disable-next-line react/no-string-refs
-            // @ts-expect-error - TS2339 - Property 'focus' does not exist on type 'ReactInstance'.
-            this.refs["hintEditor" + (i + dir)].focus();
+            this.hintEditors[i + dir]?.focus();
         });
     };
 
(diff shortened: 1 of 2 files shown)


packages/perseus-editor/src/hint-editor.tsx
```diff
@@ -363,40 +363,29 @@ class CombinedHintsEditor extends React.Component<CombinedHintsEditorProps> {
...
 
-    serializeHint: (index: number, options?: any) => string = (
-        index: number,
-        options: any,
-    ): string => {
-        // eslint-disable-next-line react/no-string-refs
-        // @ts-expect-error - TS2339 - Property 'serialize' does not exist on type 'ReactInstance'.
-        return this.refs["hintEditor" + index].serialize(options);
-    };
+    serializeHint(index: number, options?: any): PerseusRenderer {
```
benchristel, line 386: Refactoring `serializeHint` to be a method instead of an arrow function changes semantics - the value of `this` might be unbound.

packages/perseus-editor/src/hint-editor.tsx
```diff
@@ -363,40 +363,29 @@ class CombinedHintsEditor extends React.Component<CombinedHintsEditorProps> {
...
+                    ?.getSaveWarnings()
+                    .map((issue) => "Hint " + (i + 1) + ": " + issue);
             })
-            .flatten(true)
-            .value();
+            .flat();
     };
 
-    serialize: (options?: any) => ReadonlyArray<string> = (options: any) => {
+    serialize(options?: any): ReadonlyArray<PerseusRenderer> {
```
benchristel, line 380: Refactoring `serialize` to be a method instead of an arrow function changes semantics - the value of `this` might be unbound.

packages/perseus-editor/src/hint-editor.tsx
```diff
@@ -363,40 +363,29 @@ class CombinedHintsEditor extends React.Component<CombinedHintsEditorProps> {
...
-    serializeHint: (index: number, options?: any) => string = (
-        index: number,
-        options: any,
-    ): string => {
-        // eslint-disable-next-line react/no-string-refs
-        // @ts-expect-error - TS2339 - Property 'serialize' does not exist on type 'ReactInstance'.
-        return this.refs["hintEditor" + index].serialize(options);
-    };
+    serializeHint(index: number, options?: any): PerseusRenderer {
+        return this.hintEditors["hintEditor" + index]?.serialize(options);
```
benchristel, line 387:
```suggestion
        return this.hintEditors[index]?.serialize(options);
```

packages/perseus-editor/src/hint-editor.tsx
```diff
@@ -363,40 +363,29 @@ class CombinedHintsEditor extends React.Component<CombinedHintsEditorProps> {
...
-        return _.chain(this.props.hints)
+        return this.props.hints
             .map((hint, i) => {
-                return _.map(
-                    // eslint-disable-next-line react/no-string-refs
-                    // @ts-expect-error - TS2339 - Property 'getSaveWarnings' does not exist on type 'ReactInstance'.
-                    this.refs["hintEditor" + i].getSaveWarnings(),
-                    (issue) => "Hint " + (i + 1) + ": " + issue,
-                );
+                return this.hintEditors["hintEditor" + i]
```
benchristel, line 373: I don't think this will work if `this.hintEditors` is a `Record<number, ...>`. It should probably be:

```suggestion
                return this.hintEditors[i]
```

benchristel (changes requested):
I think there's a bug in the way the `hintEditors` record is accessed that will break serialization and save warnings.

packages/perseus-editor/src/hint-editor.tsx
```diff
@@ -405,7 +394,9 @@ class CombinedHintsEditor extends React.Component<CombinedHintsEditorProps> {
             (hint, i) => {
                 return (
                     <CombinedHintEditor
-                        ref={"hintEditor" + i}
+                        ref={(editor) => {
+                            this.hintEditors[i] = editor;
+                        }}
```
nishasy, line 399: Nice ref callback

nishasy (approved):
LGTM.

I was wondering if there's any way to test that the serialization still works as expected? Do we just not have any tests for the `serialize()` function?

packages/perseus-editor/src/hint-editor.tsx
jeremywiebe: `serializeHint` is only called from within this component and never passed. When possible, I've been trying to convert these arrow functions on classes to standard methods to reduce some of the noise.

packages/perseus-editor/src/hint-editor.tsx
jeremywiebe: :D

packages/perseus-editor/src/hint-editor.tsx
jeremywiebe: This `serialize` is called directly by the `EditorPage` in its `serialize()` function so it is safe. As with `serializeHint`, I think it's useful to "undo" these migrations to arrow functions so that it's more obvious we aren't passing these around by value.

benchristel (approved):
LGTM!

PR #1560, opened 2024-08-27
Editors: better type serialize() functions

jeremywiebe:
## Summary:

This PR improves typing of the `serialize()` methods on the various editors. This function is used to return the edited data to higher-up components (and ultimately out of Perseus itself). 

Better typing helps us to ensure that we're returning the right data in the right format/structure.

Issue: LEMS-1809

## Test plan:

`yarn tsc`
`yarn test`

### DIFF
--- a/packages/perseus-editor/src/editor-page.tsx
+++ b/packages/perseus-editor/src/editor-page.tsx
@@ -193,13 +193,15 @@ class EditorPage extends React.Component<Props, State> {
         return issues1.concat(issues2);
     }
 
-    serialize(options?: {keepDeletedWidgets?: boolean}): any | PerseusItem {
+    serialize(options?: {keepDeletedWidgets?: boolean}): PerseusItem {
         if (this.props.jsonMode) {
             return this.state.json;
         }
-        return _.extend(this.itemEditor.current?.serialize(options), {
-            hints: this.hintsEditor.current?.serialize(options),
-        });
+        return {
+            ...this.itemEditor.current!.serialize(options),
+            hints: this.hintsEditor.current!.serialize(options) ?? [],
+            answer: undefined,
+        };
     }
 
     handleChange: ChangeHandler = (toChange, cb, silent) => {
--- a/packages/perseus-editor/src/editor.tsx
+++ b/packages/perseus-editor/src/editor.tsx
@@ -826,11 +826,7 @@ class Editor extends React.Component<Props, State> {
         }
     };
 
-    serialize: (options?: any) => {
-        content: string;
-        images: any;
-        widgets: Record<any, any>;
-    } = (options: any) => {
+    serialize(options?: any): PerseusRenderer {
         // need to serialize the widgets since the state might not be
         // completely represented in props. ahem //transformer// (and
         // interactive-graph and plotter).
@@ -864,7 +860,7 @@ class Editor extends React.Component<Props, State> {
             images: this.props.images,
             widgets: widgets,
         };
-    };
+    }
 
     render(): React.ReactNode {
         let pieces;
--- a/packages/perseus-editor/src/hint-editor.tsx
+++ b/packages/perseus-editor/src/hint-editor.tsx
@@ -20,6 +20,7 @@ import type {
     ChangeHandler,
     DeviceType,
     ImageUploader,
+    PerseusRenderer,
 } from "@khanacademy/perseus";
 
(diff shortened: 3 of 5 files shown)


packages/perseus-editor/src/editor-page.tsx
```diff
@@ -193,13 +193,15 @@ class EditorPage extends React.Component<Props, State> {
...
         if (this.props.jsonMode) {
             return this.state.json;
         }
-        return _.extend(this.itemEditor.current?.serialize(options), {
-            hints: this.hintsEditor.current?.serialize(options),
-        });
+        return {
+            ...this.itemEditor.current!.serialize(options),
+            hints: this.hintsEditor.current!.serialize(options) ?? [],
+            answer: undefined,
```
nishasy, line 203: **Question**: Why would we need this `answer: undefined` here?

Would it not be the same as just not adding it at all here? Or are you trying to override whatever `answer` might be in `this.itemEditor.current`? If so, why is that the case? Is this different behavior than what was here before?

packages/perseus-editor/src/editor.tsx
```diff
@@ -826,11 +826,7 @@ class Editor extends React.Component<Props, State> {
         }
     };
 
-    serialize: (options?: any) => {
-        content: string;
-        images: any;
-        widgets: Record<any, any>;
-    } = (options: any) => {
+    serialize(options?: any): PerseusRenderer {
```
benchristel, line 829: Refactoring `serialize` to a method instead of an arrow function changes semantics - `this` might be unbound.

packages/perseus-editor/src/item-editor.tsx
```diff
@@ -74,16 +76,13 @@ class ItemEditor extends React.Component<Props> {
...
-        question: any;
+        answerArea: PerseusAnswerArea;
+        itemDataVersion: Version;
+        question: PerseusRenderer;
     } = (options: any) => {
         return {
-            question: this.questionEditor.current?.serialize(options),
-            answerArea: this.itemExtrasEditor.current?.serialize(),
+            question: this.questionEditor.current!.serialize(options),
+            answerArea: this.itemExtrasEditor.current!.serialize(),
```
nishasy, line 85: **Question**: Is there really never a case when these could be null? (Re: the `!.`)

packages/perseus-editor/src/hint-editor.tsx
```diff
@@ -89,12 +89,12 @@ export class HintEditor extends React.Component<HintEditorProps> {
         return this.editor.current?.getSaveWarnings();
     };
 
-    serialize: (options?: any) => any = (options: any) => {
+    serialize(options?: any): Hint {
```
benchristel, line 92: Refactoring `serialize` to a method instead of an arrow function changes semantics - `this` might be unbound.

packages/perseus-editor/src/editor-page.tsx
benchristel, line 203: What's the purpose of setting `answer` to `undefined` here? We weren't doing that before.

benchristel (approved):
Seems good - just left a few comments inline.

packages/perseus-editor/src/editor-page.tsx
jeremywiebe: I found that I needed to provide it. I _think_ it's because it is defined as `answer: any`. So TypeScript doesn't like that it's not defined as a key, but you can assign it the value `undefined` which then essentially removes it. 🤷‍♂️ 

Ultimately, I'd like to look at removing `answer` from `PerseusRenderer` as it is a deprecated field. I searched our English content and can only find one usage of this field and it is in a "staging lifeboat" (ie, not really used). I loaded that assessment item in teh editor and the JSON view doesn't even show it, which indicates to me that the editor doesn't even acknowledge that field any longer...so I am going to remove the field from the `PerseusRenderer` type!

jeremywiebe: I explained above to the same question Nisha posed. It's because the field is defined as `answer: any`, so apparently TypeScript requires it to be present in the returned object even if we specify it's value is `undefined`.

packages/perseus-editor/src/hint-editor.tsx
jeremywiebe: I've checked that this doesn't happen. 👍

PR #1561, opened 2024-08-27
IframeContentRenderer: strongly type sendNewData()

jeremywiebe:
## Summary:

This PR improves typing on the editor's `sendNewData` method. This method is used to push edits to the preview iframe. Prior to this PR the data being sent was not clearly typed and hard to understand. 

Issue: LEMS-1809

## Test plan:

`yarn tsc`
`yarn test`

### DIFF
--- a/packages/perseus-editor/src/editor-page.tsx
+++ b/packages/perseus-editor/src/editor-page.tsx
@@ -156,7 +156,7 @@ class EditorPage extends React.Component<Props, State> {
 
         this.itemEditor.current?.triggerPreviewUpdate({
             type: "question",
-            data: _({
+            data: {
                 item: this.serialize(),
                 apiOptions: deviceBasedApiOptions,
                 initialHintsVisible: 0,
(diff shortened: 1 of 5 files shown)


packages/perseus-editor/src/editor-page.tsx
```diff
@@ -156,7 +156,7 @@ class EditorPage extends React.Component<Props, State> {
 
         this.itemEditor.current?.triggerPreviewUpdate({
             type: "question",
-            data: _({
```
jeremywiebe, line 159: I hadn't seen this usage of `_` (underscore) before. It provides an "OO" wrapper for the object.

packages/perseus-editor/src/editor-page.tsx
```diff
@@ -166,17 +166,11 @@ class EditorPage extends React.Component<Props, State> {
...
+                    stack: [],
                 },
                 reviewMode: true,
                 legacyPerseusLint: this.itemEditor.current?.getSaveWarnings(),
-            }).extend(
-                _(this.props).pick(
-                    "workAreaSelector",
-                    "solutionAreaSelector",
-                    "hintsAreaSelector",
-                    "problemNum",
```
jeremywiebe, line 177: None of these are used anymore. I checked this repo as well as webapp.

packages/perseus-editor/src/iframe-content-renderer.tsx
```diff
@@ -14,6 +14,48 @@
...
+
+type ArticleData = {
+    apiOptions: APIOptions;
+    json: Partial<PerseusRenderer>;
+    useNewStyles: boolean;
+    linterContext: LinterContextProps;
+    legacyPerseusLint: ReadonlyArray<string>;
+};
+
+export type SendNewData =
```
benchristel, line 33: Could we call this type something like `NewDataMessage`? The verb `Send` makes it sound like a function type.

benchristel (approved):
LGTM!

packages/perseus-editor/src/iframe-content-renderer.tsx
jeremywiebe: Yes! `SendNewData` just didn't feel right, but I couldn't think of a better alternative. :) Thanks!

PR #1563, opened 2024-08-27
IframeContentRenderer: simplify iframeID assignment so typing can improve

jeremywiebe:
## Summary:

The IframeContentRenderer used to assign itself an `iframeID` in `componentDidMount`. There's no reason it can't assign that to itself when the component is contructed. I've moved the id generation to be in the member declaration, which means the type can be implicitly, but strongly, defined there. 

Issue: LEMS-1809

## Test plan:

`yarn tsc`
`yarn test`

### DIFF
--- a/packages/perseus-editor/src/iframe-content-renderer.tsx
+++ b/packages/perseus-editor/src/iframe-content-renderer.tsx
@@ -115,16 +115,13 @@ class IframeContentRenderer extends React.Component<Props> {
     _lastData: any;
     // @ts-expect-error - TS2564 - Property '_lastHeight' has no initializer and is not definitely assigned in the constructor.
     _lastHeight: number;
-    // @ts-expect-error - TS2564 - Property 'iframeID' has no initializer and is not definitely assigned in the constructor.
-    iframeID: number;
+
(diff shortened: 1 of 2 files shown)


packages/perseus-editor/src/iframe-content-renderer.tsx
```diff
@@ -115,17 +115,14 @@ class IframeContentRenderer extends React.Component<Props> {
     _lastData: any;
     // @ts-expect-error - TS2564 - Property '_lastHeight' has no initializer and is not definitely assigned in the constructor.
     _lastHeight: number;
-    // @ts-expect-error - TS2564 - Property 'iframeID' has no initializer and is not definitely assigned in the constructor.
-    iframeID: number;
+
+    iframeID = nextIframeID++;
```
nishasy, line 119: Would this work? Where is `nextIframeID` defined?

benchristel: It's defined as a `let` at the top level of this file, on line 59.

benchristel (approved):

benchristel (approved):
LGTM!

PR #1566, opened 2024-08-27
Publish NPM snapshots for draft PRs

benchristel:
We talked about the tradeoffs in Slack: https://khanacademy.slack.com/archives/C01AZ9H8TTQ/p1724281433448859

## Test plan:

I'm opening this PR as a draft to see if the snapshot gets published.

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -272,7 +272,7 @@ jobs:
         name: Publish npm snapshot
         # We don't publish snapshots on Changeset "Version Packages" PRs
         if: |
-          !startsWith(github.head_ref, 'changeset-release/')
+            !startsWith(github.head_ref, 'changeset-release/')
         runs-on: ${{ matrix.os }}
         strategy:
             matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/node-ci.yml
```diff
@@ -278,11 +278,8 @@ jobs:
 
     publish_snapshot:
         name: Publish npm snapshot
-        # We don't publish snapshots on draft PRs or
-        # on the main Changeset "Version Packages" PR
-        if: |
-            github.event.pull_request.draft == false
-            && !startsWith(github.head_ref, 'changeset-release/')
+        # We don't publish snapshots on Changeset "Version Packages" PRs
+        if: !startsWith(github.head_ref, 'changeset-release/')
```
benchristel, line 282: The snapshot build didn't get created for this PR when it was in draft state... I wonder if that's because the workflow config on `main` is always used? (that seems reasonable... otherwise it'd be a security risk for OSS projects)

jeremywiebe: I can't find the documentation for this, but I'm pretty sure you're correct. GitHub Actions use the base branch's configuration so you have to land the change before you can test/see it.

jeremywiebe (approved):

PR #1569, opened 2024-08-28
Run `sync-dependencies.ts` script

handeyeco:
## Summary:
We're running into TS issues that I think is caused by using incompatible WB/WS versions. So I ran `sync-dependencies.ts` to get us up-to-sync with Webapp.

See: https://khanacademy.slack.com/archives/C01AZ9H8TTQ/p1724797503623059


packages/math-input/package.json
```diff
@@ -56,29 +56,29 @@
...
         "perseus-build-settings": "^0.4.1",
         "prop-types": "15.6.1",
-        "react": ">= 18",
-        "react-dom": ">= 18",
-        "react-router": "^5.2.1",
-        "react-router-dom": "^5.3.0",
+        "react": "^18.2.0",
+        "react-dom": "^18.2.0",
+        "react-router": "5.3.4",
+        "react-router-dom": "5.3.4",
```
jeremywiebe, line 62: **thought**: I wonder if these are even used/needed by `math-input`?  We don't import them anywhere. I wonder if this is just cruft from when we moved Perseus back into this repo. No change necessary, just musing.

jeremywiebe (approved):
Cool.

PR #1567, opened 2024-08-27
Remove points: reducer

nicolecomputer:
## Summary:
This adds a reducer action for removing points

Issue: LEMS-2280

## Test plan:
- Run the tests

PR #1570, opened 2024-08-28
Remove unlimited points via keyboard

nicolecomputer:
## Summary:

https://github.com/user-attachments/assets/118cb3dc-4afe-4371-946b-f401b17c09e2

Remove unlimited points via keyboard.

- A point can be removed by hitting delete
- After a point is deleted the graph is focused

Issue: https://khanacademy.atlassian.net/browse/LEMS-2280

## Test plan:
- Run the tests
- Everything should pass


nicolecomputer:
The parent pull-request ([#1567](https://github.com/Khan/perseus/pull/1567)) has been merged into `main`, but this branch (`new/remove-points-keyboard`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -77,16 +80,24 @@ export const MafsGraph = (props: MafsGraphProps) => {
...
+                        height,
+                    }}
+                    onKeyUp={(event) => {
+                        if (event.key === "Backspace") {
+                            dispatch(actions.global.deleteIntent());
+                            graphRef.current?.focus();
+                        }
                     }}
+                    ref={graphRef}
+                    tabIndex={0}
```
benchristel, line 100: Should the graph be focusable with the tab key? If we want it to be programmatically focusable but not tabbable, we could set `tabIndex={-1}`. ([MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex))

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -77,16 +80,24 @@ export const MafsGraph = (props: MafsGraphProps) => {
...
                         pointerEvents: props.static ? "none" : "auto",
                         userSelect: "none",
+                        width,
+                        height,
+                    }}
+                    onKeyUp={(event) => {
+                        if (event.key === "Backspace") {
+                            dispatch(actions.global.deleteIntent());
+                            graphRef.current?.focus();
+                        }
```
benchristel, line 97: I'm still a little confused about why we have to handle this event on the graph component rather than on the point itself. I tried adding an `onKeyDown` handler to the `g[data-testid='movable-point']` element and it works in both Edge and Safari. The focused point's event handler is fired when you press a key.

The advantage of putting the event handler on the point would be that we wouldn't have to keep track of which point is focused in the useReducer state.

Am I missing something?

benchristel (commented):
The code looks good, and I love the way it works. But I'm wondering if there's a simpler way to accomplish this (with no added reducer state). I left comments inline.

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
nicolecomputer: @mark-fitzgerald and @nishasy thought it should be 0 because we expect it to be tab-able. Nisha has future work that uses that!

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
nicolecomputer: I'm definitely open to changing it but there are a couple of things I like:

1. This lets us add a delete intent to other kinds of graphs with no fuss. I think we'll use that for unlimited polygon (but that's a gut feeling).
2. This makes it natural to add other keyboard shortcuts and gives a natural place for them to live
3. I think I need to know the selected point regardless of this change. After a point is added we want to focus on the newly added point. So as a learner I tab to the "Add point" button, hit space, the point is selected and I can manipulate it with my keyboard. I think that means we'll need to keep track of the selected point so that after the add button is activated we select the right thing.

benchristel (approved):
LGTM!

PR #1571, opened 2024-08-28
The final 'Upload Coverage' step doesn't need local source or node_modules

jeremywiebe:
## Summary:

I noticed this last step in our CI runs takes a bit of time for what amounts to downloading 2 files and then uploading them to Codecov. 

This action uses only other Github shared actions and so it doesn't need to checkout the source or install node_modules. This should speed up this last step in our CI runs somewhat. 

Issue: "none"

## Test plan:

I'll need to land this and then watch the next PR that runs this action.


handeyeco (approved):
Nice!

PR #1576, opened 2024-08-29
PhET widget: testing Github actions

jeremywiebe:
## Summary:

This PR is just to kick off the Github Actions to make sure the latest changes work. We can land still review and land this PR as it removes some lint suppressions that we shouldn't be using in modern React/Perseus code. 

The changes are all just code moving around based on ESLint rules. 

Issue: "none"

## Test plan:


packages/perseus/src/widgets/phet-simulation.tsx
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable react/forbid-prop-types, react/sort-comp */
```
aemandine, line 1: Nice! Thank you for getting rid of this lint suppression!

aemandine (approved):
LGTM!

PR #1577, opened 2024-08-29
Remove mock/test widgets

handeyeco:
## Summary:

I don't see the point of having a bunch of test widgets when we could just use real widgets in tests.

Open to push back.


.changeset/plenty-crews-wink.md
```diff
@@ -0,0 +1,6 @@
+---
+"@khanacademy/perseus": major
+"@khanacademy/perseus-editor": major
+---
+
+Remove mock widgets and their editors
```
jeremywiebe, line 6: Maybe refer to "example" instead of "mock" as the mocks were really only for unit tests, right? 

```suggestion
Remove example widgets and their editors
```

packages/perseus/src/__tests__/server-item-renderer.test.tsx
```diff
@@ -288,16 +288,22 @@ describe("server item renderer", () => {
         // render tree created.
         // Finally we re-render and poke the asset status to loaded. At that
         // everything is loaded.
+        //
+        // TODO: is it possible to use an existing widget to test the same logic?
```
jeremywiebe, line 292: **thought**: I think this mocked asset loading widget might be one we want to keep.  The `image` widget would be a widget that hooks into the asset loading context, but it may be tricky to assert that the ServerItemRenderer does the right thing using the implementation of a production widget, and not risk regression. 

The implementation of that widget is extremely small. Also, I wonder if we could eliminate all the `@ts-expect-error` suppressions by simply adding the mock widget to the list of widgets, but mark it as `hidden: true`. Maybe that's a net negative though. 

What do you think?

jeremywiebe (approved):
I'm definitely in favour of removing the example widgets, but I wonder if we want to keep the two mock widgets (or at least asset loading one) that are used in some of the tests. If we tie these tests to real widgets, we run the (possibly low) risk of breaking unrelated tests if we change the widget.

packages/perseus/src/__tests__/server-item-renderer.test.tsx
handeyeco: I came to the same conclusion as you: keep the widget. I chose to mark it as deprecated though so that people would at least be hesitant to use it (I personally think we should bias towards using real widgets when possible); that meant that I had to suppress "no deprecated" errors when using it.

I'd be fine removing the TODOs and deprecated stuff if you think that's best.

PR #1580, opened 2024-09-03
[Actions Caching] Use actions/setup-node to manage yarn cache

jeremywiebe:
## Summary:

We've been having issues with our Github Actions failing with snapshot failures sporadically. We suspect this is because our node_modules cache is being restored with a non-matching cache. 

To address this, I'm refactoring our actions so that we use the Github-supplied action to setup Node and manage the `yarn` cache instead of our home-grown method that specifies cache keys. 

I've also separated out the management of the Cypress binary cache (https://docs.cypress.io/guides/guides/command-line#cypress-cache-command) from the management of Node/node_modules caches. They are unrelated and only needed when we run the Cypress commands. 

Issue: "none"

## Test plan:

Testing is always fun. I'll try using the `workflow_dispatch` option to run these actions manually, but we might have to land them and verify if the `workflow_dispatch` also uses the base branch's action files.


.github/actions/cypress-cache/action.yml
```diff
@@ -0,0 +1,16 @@
...
+description: "Install & cache Cypress binaries"
+
+# The steps this action runs.
+runs:
+    using: "composite"
+    steps:
+        - name: Find Cypress cache
+          id: find-cypress-cache
+          shell: bash
+          run: echo "CYPRESS_CACHE_PATH=$(./node_modules/.bin/cypress cache path)" >> "$GITHUB_OUTPUT"
```
jeremywiebe, line 11: https://docs.cypress.io/guides/guides/command-line#cypress-cache-path

jeremywiebe:
Closing as my testing is done.

PR #1583, opened 2024-09-03
Remove "sort-comp" exceptions

handeyeco:
## Summary:
The only thing this PR does is it removes `sort-comp` ESLint exceptions and fixes the sort order. See: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/sort-comp.md

I didn't do:
- `interactive-graph.tsx` since I didn't want to cause merge conflicts for that team
- Components that will get deleted in [my other PR](https://github.com/Khan/perseus/pull/1577)

### DIFF
--- a/.changeset/swift-suits-rest.md
+++ b/.changeset/swift-suits-rest.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/perseus": patch
+"@khanacademy/perseus-editor": patch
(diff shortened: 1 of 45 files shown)


jeremywiebe (approved):
+++ Thanks for getting rid of one more ESLint suppression!

PR #1586, opened 2024-09-04
Use setup-node to install and cache node_modules

benchristel:
## Summary:
We've been seeing intermittent CI test failures in that appear to be caused by incorrect
NPM packages being installed. Clearing the node_modules cache used by our GitHub actions
resolves the issue... temporarily.

In the case of a cache miss, we restore node_modules from the "wrong" cache and
`yarn install` on top of that. Usually, this installs the correct versions of any
packages that need updating, and everything works. But we think that sometimes, it might
not.

The setup-node action from GitHub implements node_modules caching, and doesn't restore
from fallback caches. This PR switches our CI jobs to use `setup-node`, which we hope
will resolve the caching issues.

## Test plan:

Look. See CI. See CI run. Run, CI, run.

### DIFF
--- a/.github/actions/shared-node-cache/action.yml
+++ b/.github/actions/shared-node-cache/action.yml
@@ -19,4 +19,5 @@ runs:
               cache: yarn
 
         - name: yarn install
+          shell: sh
           run: yarn install
--- a/.github/actions/shared-node-cache/action.yml
+++ b/.github/actions/shared-node-cache/action.yml
@@ -12,11 +12,11 @@ inputs:
 runs:
     using: "composite"
     steps:
-      - name: Set up node
-        uses: actions/setup-node@v4
-        with:
-          node-version: ${{ inputs.node-version }}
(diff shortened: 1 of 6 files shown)


.github/actions/shared-node-cache/action.yml
```diff
@@ -6,22 +6,18 @@ inputs:
...
         description: "Node version to use"
         required: true
-    ssh-private-key:
-        description: "A private SSH key so that we can obtain private dependencies like our event schema package"
-        required: true
 
 # The steps this action runs.
 # The order of the two steps below needs to be maintained.
-# The ssh-agent step needs to be before the Node.js/Install step because our package.json uses
-# git+ssh. As this requires a valid SSH key, the ssh-agent action must be before the yarn install.
```
handeyeco, line 16: Cool that we don't need this anymore

handeyeco (approved):
Nice, looks good to me

PR #1588, opened 2024-09-04
Remove Util.DeprecationMixin

benchristel:
## Summary:
`Util.DeprecationMixin` was only being used to remove a deprecated `showGraph`
prop from the InteractiveGraph widget and its editor component. `showGraph` no
longer appears anywhere in our content, so the "mixin" (which, given the
naming, is likely a relic from the `React.createClass` era) is not needed
anymore.

Issue: none

## Test plan:
Edit and save an interactive graph in the editor.


handeyeco (approved):
Nice, love to see it

PR #1589, opened 2024-09-04
Adjust sizing

aemandine:
## Summary:
* Make PhET widget smaller

Storybook editor page with a PhET widget that has approximately a 16:9 aspect ratio and that does not leave its bounding box

Issue: LEMS-2306

## Test plan:
* `yarn jest packages/perseus/src/widgets/phet-simulation/phet-simulation.test.ts`
* Verify that the widget seems to have an appropriate aspect ratio in Storybook


packages/perseus/src/widgets/phet-simulation/phet-simulation.tsx
```diff
@@ -240,9 +240,9 @@ const styles = StyleSheet.create({
         borderRadius: borderRadiusLarge,
         borderWidth: 1,
         borderColor: basicBorderColor,
-        padding: phoneMargin,
+        padding: spacing.medium_16,
         paddingBottom: 0,
-        width: 650,
+        width: 400 + spacing.medium_16 * 2,
```
SonicScrewdriver: Would `spacing.xLarge_32` make sense here, rather than multiplying?

SonicScrewdriver (approved):
LGTM

packages/perseus/src/widgets/phet-simulation/phet-simulation.tsx
aemandine: Brilliant, and I will include a comment explaining why the xLarge_32!

PR #1590, opened 2024-09-04
Move InteractiveGraphEditor to a directory

benchristel:
This is in preparation for splitting up the editor into multiple components,
which I think I'm going to have to do in order to type the `correct` prop
correctly (currently it has type `any`).

Issue: none

## Test plan:
CI should pass


packages/perseus-editor/src/widgets/interactive-graph-editor/index.tsx
```diff
@@ -5,6 +5,7 @@ import {
     InteractiveGraphWidget,
     SizingUtils,
     Util,
+    PerseusGraphType,
```
nishasy, line 8: Will this be used later? Looks like it's not being used here.

nishasy, line 8: Will this be used later? It looks like it's not being used here now.

nishasy (commented):
A big undertaking! How exciting.

One concern I have is that changing the name of `interactive-graph-editor.tsx` to `index.tsx` will make it harder to search for in an IDE. I'd prefer for it to remain `interactive-graph-editor.tsx`, even with it being nested inside `interactive-graph-editor/`.

I think Matthew had a [similar thought](https://khanacademy.slack.com/archives/C01AZ9H8TTQ/p1725383323480269?thread_ts=1725381544.129369&cid=C01AZ9H8TTQ) when he reorganized the widgets.

benchristel:
@nishasy I've updated this in response to your comments. Can you take another look?

PR #1591, opened 2024-09-04
Move some editor-only components into `perseus-editor`

handeyeco:
## Summary:
There were 4 components declared in `perseus` but only used in `perseus-editor` so I moved them.

Issue: LEMS-2321

## Test plan:
Nothing should change, implementation nit


jeremywiebe (approved):
Nice. Good eye catching these editor-only components!

PR #1592, opened 2024-09-04
Consolidate Interactive Graph tests and stories

benchristel:
This commit just moves files around. Specifically, it moves the
tests, test data, and stories for the Interactive Graph widget into
the `widgets/interactive-graphs` directory.

Issue: none

## Test plan:

`yarn test`


packages/perseus/src/widgets/interactive-graphs/interactive-graph-regression.stories.tsx
```diff
@@ -1,8 +1,8 @@
 import * as React from "react";
 
 import Renderer from "../../renderer";
-import {mockStrings} from "../../strings";
-import {interactiveGraphQuestionBuilder} from "../interactive-graphs/interactive-graph-question-builder";
+import {mockStrings} from "@khanacademy/perseus/strings";
```
handeyeco, line 4: I noticed my IDE did this too; will need to switch back to relative imports (I think it has to do with the fact that Perseus has a new `perseus/strings` export.

```
// from packages/perseus/package.json
    "exports": {
        ".": {
            "import": "./dist/es/index.js",
            "require": "./dist/index.js",
            "types": "./dist/index.d.ts",
            "source": "./src/index.ts"
        },
        "./strings": {
            "import": "./dist/es/strings.js",
            "require": "./dist/strings.js",
            "types": "./dist/strings.d.ts",
            "source": "./src/strings.ts"
        },
        "./styles.css": "./dist/index.css"
    },
```

handeyeco (approved):
Thanks for doing this! Just some IDE weirdness related to auto-changing import paths, but after that it looks good.

PR #1594, opened 2024-09-04
Delete Storybook stories that use randomly generated data

benchristel:
These aren't particularly useful - the widgets they generate often look broken,
and it's hard to tell if there's a bug in the widget or the input data is just
something a human would never create.

I believe this completes the effort (started by @handeyeco) to move code out of
`perseus/src/widgets` and into specific widget subdirectories.

Issue: none

## Test plan:

`yarn test`


handeyeco (approved):
Conceptually I like the idea, but I think we need something like "random exercise from prod data" or something (outside of the scope of this PR).

Thanks for taking the initiative on this.

PR #1598, opened 2024-09-06
Add remove point button

nicolecomputer:
## Summary:

image

Adds a remove point button so that touch users can remove a point after they have added it.

The remove point button also shows when a learner clicks to focus a point or tabs to focus it.

Unfortunately the button steals focus when it's clicked and causes the point to blur so to work around this we keep track of the last focused point and use that when the delete button is clicked.

Issue: XXX-XXXX

## Test plan:
- Run the tests
- **Everything should pass**


packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
```diff
@@ -764,5 +771,81 @@ describe("MafsGraph", () => {
                 [{type: ADD_POINT, location: [0, 0]}],
             ]);
         });
+
+        it("adds shows a remove point button when a point is focused", async () => {
```
catandthemachines, line 775: Great tests! 🥰

catandthemachines (commented):
The code looks very straight forward and clean. No notes, but I recommend having an interactive graph person review since I'm not as familiar with this code. 👩‍💻

catandthemachines (approved):

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -161,40 +165,69 @@ export const MafsGraph = (props: MafsGraphProps) => {
...
+                kind="secondary"
+                color="destructive"
+                style={{
+                    width: "100%",
+                    marginLeft: "20px",
+                }}
+                onClick={(event) => {
+                    props.dispatch(
+                        actions.pointGraph.removePoint(
+                            props.state.previouslyFocusedPointIndex!,
```
benchristel, line 209: I don't quite understand how this works. We are checking `props.state.focusedPointIndex !== null || props.state.previouslyFocusedPointIndex !== null` to decide whether to show the button, but the `previouslyFocusedPointIndex` is the one that's deleted. Should we omit the check for `props.state.focusedPointIndex !== null`?

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -161,40 +165,69 @@ export const MafsGraph = (props: MafsGraphProps) => {
...
-        kind="secondary"
+    <View
         style={{
-            width: "100%",
-            marginLeft: "20px",
-        }}
-        onClick={() => {
-            props.dispatch(actions.pointGraph.addPoint([0, 0]));
+            flexDirection: "row",
+            width: props.width,
```
benchristel, line 182: Why do we need to specify the `width` now? It seems like `width: 100%` should work, given that this container is `display: flex`.

benchristel (commented):
Looks great overall! I left a question inline about the logic for determining when to show the delete button and deciding which point gets deleted.

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
nicolecomputer: Width: 100% takes up the whole container width and spills outside the graph 😱. This seems to make things show up correctly

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
nicolecomputer: previouslyFocusedPointIndex gets cleared when the point is removed (it's in the reducer).

We need this because the button steals the focus and the point blurs and then we're not able to remove the point. I wish it was different :( I'd be happy to continue to iterate in a follow on (and pair on it!) if you'd like

benchristel: Yeah, I get why we need both `previouslyFocusedPointIndex` and `focusedPointIndex`. What I don't understand is why we always delete `previouslyFocusedPointIndex` here. Based on your explanation, I would expect logic like this:

```ts
actions.pointGraph.removePoint(
  // delete the focused point, or the previously focused point if no point is focused.
  props.state.focusedPointIndex ?? props.state.previouslyFocusedPointIndex,
)
```

I'm running your code locally, and the currently focused point _does_ get deleted when I click the button, as I would expect. So it's working. I just don't understand _why_ it works.

benchristel: Ohhh, I get it... the point is _always_ blurred, because the button gets focused when it's clicked. Okay. That might be worth a comment :)

benchristel (approved):
LGTM!

PR #1599, opened 2024-09-06
Type the `correct` prop passed to InteractiveGraphEditor

benchristel:
Previously, `correct` was typed as `any` to suppress an error. TypeScript
complained that the InteractiveGraph component might pass the wrong subtype of
`PerseusGraphType` to its `onChange` callback, causing us to create a
nonsensical chimera graph when we do stuff like `correct = {...correct,
...newGraph};`

This PR adds runtime checks (mostly using `tiny-invariant`) to ensure that
graphs have the type we expect.

In doing this refactoring, I discovered a second problem: the
InteractiveGraphEditor sometimes sets the `coords` of the graph to `null`,
which wasn't allowed by `PerseusGraphType` - only `undefined` was allowed.
This PR fixes that by updating `PerseusGraphType` to allow null coords. The
interactive graph code can (thankfully) already handle null values just fine.

Issue: none

## Test plan:
Edit an interactive graph in the exercise editor.

### DIFF
--- a/packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
+++ b/packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
@@ -1,26 +1,20 @@
 import {vector as kvector} from "@khanacademy/kmath";
-import type {
-    APIOptionsWithDefaults,
-    LockedFigure,
-    PerseusImageBackground,
-    PerseusInteractiveGraphWidgetOptions,
-} from "@khanacademy/perseus";
 import {
     components,
     InteractiveGraphWidget,
     interactiveSizes,
-    PerseusGraphType,
     SizingUtils,
     Util,
 } from "@khanacademy/perseus";
-import type {PropsFor} from "@khanacademy/wonder-blocks-core";
 import {View} from "@khanacademy/wonder-blocks-core";
 import {OptionItem, SingleSelect} from "@khanacademy/wonder-blocks-dropdown";
 import {Checkbox} from "@khanacademy/wonder-blocks-form";
 import {spacing} from "@khanacademy/wonder-blocks-tokens";
 import {LabelSmall} from "@khanacademy/wonder-blocks-typography";
+import {UnreachableCaseError} from "@khanacademy/wonder-stuff-core";
 import {StyleSheet} from "aphrodite";
 import * as React from "react";
+import invariant from "tiny-invariant";
 import _ from "underscore";
 
 import LabeledRow from "../../components/graph-locked-figures/labeled-row";
@@ -33,9 +27,15 @@ import SegmentCountSelector from "../../components/segment-count-selector";
 import StartCoordsSettings from "../../components/start-coords-settings";
 import {shouldShowStartCoordsUI} from "../../components/util";
 import {parsePointCount} from "../../util/points";
-import {PerseusGraphTypePolygon, PerseusGraphTypeAngle} from "@khanacademy/perseus/src/perseus-types";
-import invariant from "tiny-invariant";
-import {UnreachableCaseError} from "@khanacademy/wonder-stuff-core";
+
+import type {
+    APIOptionsWithDefaults,
+    LockedFigure,
+    PerseusImageBackground,
+    PerseusInteractiveGraphWidgetOptions,
+    PerseusGraphType,
+} from "@khanacademy/perseus";
+import type {PropsFor} from "@khanacademy/wonder-blocks-core";
 
 const {InfoTip} = components;
 const {containerSizeClass, getInteractiveBoxFromSizeClass} = SizingUtils;
@@ -58,6 +58,8 @@ const POLYGON_SIDES = _.map(_.range(3, 13), function (value) {
 });
 
 type Range = [min: number, max: number];
+type PerseusGraphTypePolygon = Extract<PerseusGraphType, {type: "polygon"}>;
+type PerseusGraphTypeAngle = Extract<PerseusGraphType, {type: "angle"}>;
 
 export type Props = {
     apiOptions: APIOptionsWithDefaults;
@@ -293,7 +295,7 @@ class InteractiveGraphEditor extends React.Component<Props> {
                     let correct = this.props.correct;
                     // TODO(benchristel): can we improve the type of onChange
                     // so this invariant isn't necessary?
-                    invariant(newGraph != null)
+                    invariant(newGraph != null);
                     if (correct.type === newGraph.type) {
                         correct = mergeGraphs(correct, newGraph);
                     } else {
@@ -381,7 +383,9 @@ class InteractiveGraphEditor extends React.Component<Props> {
                                 }
                                 placeholder=""
                                 onChange={(newValue) => {
-                                    invariant(this.props.graph?.type === "polygon")
+                                    invariant(
+                                        this.props.graph?.type === "polygon",
+                                    );
                                     const updates = {
                                         numSides: parsePointCount(newValue),
                                         coords: null,
@@ -392,8 +396,14 @@ class InteractiveGraphEditor extends React.Component<Props> {
                                     } as const;
 
                                     this.props.onChange({
-                                        correct: {...this.props.correct, ...updates},
-                                        graph: {...this.props.graph, ...updates},
+                                        correct: {
+                                            ...this.props.correct,
+                                            ...updates,
+                                        },
+                                        graph: {
+                                            ...this.props.graph,
+                                            ...updates,
+                                        },
                                     });
                                 }}
                                 style={styles.singleSelectShort}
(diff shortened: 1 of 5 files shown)


packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
```diff
@@ -119,7 +124,7 @@ export type Props = {
      * on the state of the interactive graph previewed at the bottom of the
      * editor page.
      */
-    correct: any; // TODO(jeremy)
+    correct: PerseusGraphType;
```
nishasy, line 127: Not sure if this is the right time to think about this, but do we want `correct` to be typed as `PerseusGraphType` long term?

Personally, I think it makes more sense for `correct` to only have the `type` and `coords` fields in it, and all the other field would be specific to `graph`. And in most of the places where `correct` is being used now, we would use `graph` instead. We could start with `PerseusGraphType` for now and make that change later?

Any thoughts on that? (@benchristel @jeremywiebe)

I could see this potentially turning into a larger discussion, so I think it could also be worth taking this offline.

packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
```diff
@@ -694,6 +751,54 @@ class InteractiveGraphEditor extends React.Component<Props> {
...
+function mergeGraphs(
+    a: PerseusGraphType,
+    b: PerseusGraphType,
+): PerseusGraphType {
+    if (a.type !== b.type) {
+        throw new Error(
+            `Cannot merge graphs with different types (${a.type} and ${b.type})`,
+        );
+    }
+    switch (a.type) {
```
nishasy, line 766: Do we need this switch statement if there's already a check for `a.type !== b.type` above?

I feel like we should just be able to return `{...a, ...b}` after that? Or is this a case of TypeScript complaining if you do that?

packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
```diff
@@ -294,17 +291,16 @@ class InteractiveGraphEditor extends React.Component<Props> {
...
                     let correct = this.props.correct;
-                    // @ts-expect-error - TS2532 - Object is possibly 'undefined'.
-                    if (correct.type === newProps.graph.type) {
-                        correct = {
-                            ...correct,
-                            ...newProps.graph,
-                        };
+                    // TODO(benchristel): can we improve the type of onChange
+                    // so this invariant isn't necessary?
+                    invariant(newGraph != null);
```
nishasy, line 298: Is it better to use `invariant` than to do a direct check for existence? Not asking for a change, I'm just curious because I haven't really seen this `invariant` pattern used in other places.

nishasy (approved):
A couple of questions, but LGTM.

Might also be good to run this by @jeremywiebe since I think he might have more knowledge than I do about how `correct` is used.

jeremywiebe (approved):
I can't comment on the lines I'll refer to as they weren't changed.  But I wonder if we could tighten up the `graph` and `correct` types further by making the types enforce that `.graph.type` and `.correct.type` are always the same.

Something like this definition of `PerseusInteractiveGraphWidgetOptions`.  Basically, making `graph` and `correct` keys a set of type unions where they always match: 

```typescript
export type PerseusInteractiveGraphWidgetOptions = {
    // Where the little black axis lines & labels (ticks) should render.
    // Also known as the tick step. default [1, 1]
    // NOTE(kevinb): perseus_data.go defines this as Array<number>
    step: [number, number];
    // Where the grid lines on the graph will render. default [1, 1]
    // NOTE(kevinb): perseus_data.go defines this as Array<number>
    gridStep: [number, number];
    // Where the graph points will lock to when they are dragged. default [0.5, 0.5]
    // NOTE(kevinb): perseus_data.go defines this as Array<number>
    snapStep: [number, number];
    // An optional image to use in the background
    backgroundImage?: PerseusImageBackground;
    /**
     * The type of markings to display on the graph.
     * - graph: shows the axes and the grid lines
     * - none: shows no markings
     */
    markings: "graph" | "grid" | "none";
    // How to label the X and Y axis.  default: ["x", "y"]
    // Whether to show the Protractor tool overlayed on top of the graph
    showProtractor: boolean;
    /**
     * Whether to show the Ruler tool overlayed on top of the graph.
     * @deprecated - no longer used by the InteractiveGraph widget. The
     * property is kept on this type to prevent its accidental reuse in future
     * features, since it may appear in production data.
     */
    showRuler?: boolean;
    // Whether to show tooltips on the graph
    showTooltips?: boolean;
    /**
     * The unit to show on the ruler.  e.g. "mm", "cm",  "m", "km", "in", "ft",
     * "yd", "mi".
     * @deprecated - no longer used by the InteractiveGraph widget. The
     * property is kept on this type to prevent its accidental reuse in future
     * features, since it may appear in production data.
     */
    rulerLabel?: string;
    /**
     * How many ticks to show on the ruler.  e.g. 1, 2, 4, 8, 10, 16. Must be
     * an integer.
     * @deprecated - no longer used by the InteractiveGraph widget. The
     * property is kept on this type to prevent its accidental reuse in future
     * features, since it may appear in production data.
     */
    rulerTicks?: number;
    // The X and Y coordinate ranges for the view of the graph.  default: [[-10, 10], [-10, 10]]
    // NOTE(kevinb): perseus_data.go defines this as Array<Array<number>>
    // TODO(kevinb): Add a transform function to interactive-graph.jsx to
    // rename `range` to `ranges` so that things are less confusing.
    range: GraphRange;
    // be moved by the user.
    lockedFigures?: ReadonlyArray<LockedFigure>;
} & (
    | {
          // The type of graph
          graph: PerseusGraphTypeAngle;
          // The correct kind of graph, if being used to select function type
          correct: PerseusGraphTypeAngle;
      }
    | {
          // The type of graph
          graph: PerseusGraphTypeCircle;
          // The correct kind of graph, if being used to select function type
          correct: PerseusGraphTypeCircle;
      }
    | {
          // The type of graph
          graph: PerseusGraphTypeLinear;
          // The correct kind of graph, if being used to select function type
          correct: PerseusGraphTypeLinear;
      }
    | {
          // The type of graph
          graph: PerseusGraphTypeLinearSystem;
          // The correct kind of graph, if being used to select function type
          correct: PerseusGraphTypeLinearSystem;
      }
    | {
          // The type of graph
          graph: PerseusGraphTypePoint;
          // The correct kind of graph, if being used to select function type
          correct: PerseusGraphTypePoint;
      }
    | {
          // The type of graph
          graph: PerseusGraphTypePolygon;
          // The correct kind of graph, if being used to select function type
          correct: PerseusGraphTypePolygon;
      }
    | {
          // The type of graph
          graph: PerseusGraphTypeQuadratic;
          // The correct kind of graph, if being used to select function type
          correct: PerseusGraphTypeQuadratic;
      }
    | {
          // The type of graph
          graph: PerseusGraphTypeRay;
          // The correct kind of graph, if being used to select function type
          correct: PerseusGraphTypeRay;
      }
    | {
          // The type of graph
          graph: PerseusGraphTypeSegment;
          // The correct kind of graph, if being used to select function type
          correct: PerseusGraphTypeSegment;
      }
    | {
          // The type of graph
          graph: PerseusGraphTypeSinusoid;
          // The correct kind of graph, if being used to select function type
          correct: PerseusGraphTypeSinusoid;
      }
);
```

packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
benchristel: Yeah, TypeScript can't figure it out.

packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
benchristel: I agree that would be better. I think it's a bigger effort though (I'd like to analyze our existing content to figure out what values for `correct` are in use). I'll add a TODO comment and create a ticket.

benchristel:
@jeremywiebe I agree it would be cool if we could correlate the `correct` and `graph` types like that. Unfortunately, I don't think TypeScript supports that. Narrowing the type of either `correct` or `graph` does not narrow the type of the other. Example: [typescript playground](https://www.typescriptlang.org/play/?ssl=13&ssc=1&pln=14&pc=1#code/C4TwDgpgBA4gTgQzACygXigbwFBT1AYwHsBbEogOwC4oBnYOASwoHMAabAXygDIoAKXPkwtEKGplCQaAIjBFmwGWyjzF8JMhoUAriQBGEOJxXE4cCAWASpEWWopKVD4AGEi5y9ai6DRztxCeAA+WKKaNuB2UDL6CHDKUHFwGuI+eobGph4WVpHSMcmJye6eeel+xpzYAJTY2AAmlgA28dAAZjoUVoyUUCwQwKnI-DU0w-XEFPT96P2Dw6P1tADujMAEIywAdOEo27Y1WEGECLTQcgqOMlQn+P27YsjbLhP39ztmucAvV245Xi42CAA).

PR #1600, opened 2024-09-06
[Interactive Graph] Add example function area for locked function settings

mark-fitzgerald:
## Summary:
Having a list of sample functions available to quickly copy/paste into the locked function settings can be very helpful for content creators. This PR adds that capability, along with an updatable file (locked-function-examples.ts) for adding/modifying example functions.

Issue: LEMS-1948

## Test plan:

2. Navigate to Perseus Editor => Editor => [Demo Interactive Graph](http://localhost:6006/?path=/story/perseuseditor-editor--demo-interactive-graph)
3. Open the "interactive-graph 1" widget settings
4. Click on the "Add locked figure" drop-down at the bottom of the editor and choose "Function"
5. Below the "domain min/max" fields there is an "examples" dropdown, and instructions to the right.
6. Choose an example category from the dropdown.
   * Note that the instructions are replaced by examples for that category.
   * Currently, there are only 2 examples per category, but if you edit the DOM and add another list item, you can see that the area's height is limited, and you can scroll to see the additional example(s).
7. Copy one of the examples by clicking on the copy icon to the left of the example that you want.
8. Paste the contents of the clipboard into the equation field (above the domain min/max section).

### New settings feature
Example Equations - Updated


packages/perseus-editor/src/components/__tests__/locked-function-settings.test.tsx
```diff
@@ -363,6 +373,138 @@ describe("Locked Function Settings", () => {
...
+                    <LockedFunctionSettings
+                        {...defaultProps}
+                        directionalAxis="y"
+                        expanded={true}
+                        onChangeProps={onChangeProps}
+                    />,
+                    {wrapper: RenderStateRoot},
+                );
+
+                // Assert - "y" axis means "range"
```
nishasy, line 404: I like your system of adding extra context to the arrange/act/assert comments. I'm gonna start doing this too.

packages/perseus-editor/src/components/graph-locked-figures/locked-function-settings.tsx
```diff
@@ -45,34 +49,37 @@ const LockedFunctionSettings = (props: Props) => {
         onRemove,
     } = props;
     const equationPrefix = directionalAxis === "x" ? "y=" : "x=";
+    const domainRangeText = directionalAxis === "x" ? "domain" : "range";
```
nishasy, line 52: Nice!

packages/perseus-editor/src/components/graph-locked-figures/locked-function-settings.tsx
```diff
@@ -200,6 +256,43 @@ const LockedFunctionSettings = (props: Props) => {
...
+    return (
+        <li key={`${category}-${index}`} className={css(styles.exampleRow)}>
+            <IconButton
+                icon={autoPasteIcon}
+                aria-label="paste example"
+                aria-describedby={exampleId}
+                onClick={() => pasteEquationFn("equation", example)}
+                size="medium"
+                style={styles.copyPasteButton}
+            />
```
nishasy, line 279: Optional suggestion: Since content authors usually use desktop + mouse/keyboard, I think it would be appropriate to use a `title` prop here so that someone can hover over the icon to learn what it does.

nishasy (approved):
This is so exciting!

packages/perseus-editor/src/components/graph-locked-figures/locked-function-settings.tsx
```diff
@@ -200,6 +256,43 @@ const LockedFunctionSettings = (props: Props) => {
...
+    const exampleId = useId();
+
+    return (
+        <li key={`${category}-${index}`} className={css(styles.exampleRow)}>
+            <IconButton
+                icon={autoPasteIcon}
+                aria-label="paste example"
+                aria-describedby={exampleId}
+                onClick={() => pasteEquationFn("equation", example)}
+                size="medium"
```
nishasy, line 277: I think if you use `size="small"` here, it'll have the icons remain the same size, and the padding around them becomes smaller. This might be worth trying and seeing if the spacing looks better inline here? Let me know what you think.

nishasy (approved):
One more comment!

packages/perseus-editor/src/components/graph-locked-figures/locked-function-settings.tsx
mark-fitzgerald: Since `<IconButton>` doesn't have a `title` property available, I'm unable to add this directly. So, leaving this off for now.

packages/perseus-editor/src/components/graph-locked-figures/locked-function-settings.tsx
mark-fitzgerald: Yes, specifying "small" does keep the icon the same size. However, I'm going to keep it as is in order to keep the target area the current size (40px) as a best practice. If we get word from content creators that they prefer more space for the equation text, then we can adjust at that time.

PR #1603, opened 2024-09-06
Switch from using useRef() to createRef()

jeremywiebe:
## Summary:

I noticed a test that was jumping through hoops with spies in order to manage a ref in tests. Using `React.createRef()` is simpler and avoids needing to mock a hook. 

Issue: "none"

## Test plan:

`yarn test` ✅


aemandine (approved):
LGTM and this makes sense, no re-renders so `createRef` won't do any unwanted overwriting! Thanks for doing this, got me to look up the difference between `useRef` and `createRef`:)

PR #1605, opened 2024-09-09
Add a "None"-type interactive graph that only displays locked figures

benchristel:
Content creators will now see an "Answer type" dropdown in the interactive graph widget editor, from which they can select "Angle", "Circle", etc... plus a new type, "None".

Graphs of type "None" do not have any interactive elements and cannot be answered. However, they _can_ contain locked figures. Their purpose is to be a more accessible and content-creator-friendly replacement for the static SVG images generated by graphie-to-png.

Issue: LEMS-2030

## Test plan:

Search for "none" in the dev UI gallery (https://localhost:5173). Copy the question into the flipbook (https://localhost:5173/#flipbook) and verify that it is marked correct when you click the "check answer" button.

### DIFF
--- a/packages/perseus/src/widgets/interactive-graph.tsx
+++ b/packages/perseus/src/widgets/interactive-graph.tsx
@@ -2377,7 +2377,7 @@ class InteractiveGraph extends React.Component<Props, State> {
                 earned: 0,
                 total: 0,
                 message: null,
-            }
+            };
         }
 
         // When nothing has moved, there will neither be coords nor the
--- a/packages/perseus/src/widgets/interactive-graph.tsx
+++ b/packages/perseus/src/widgets/interactive-graph.tsx
@@ -517,6 +517,10 @@ class LegacyInteractiveGraph extends React.Component<Props, State> {
         this.line?.remove();
     };
 
+    addNoneControls: () => void = () => {};
+
+    removeNoneControls: () => void = () => {};
+
     addLinearControls: () => void = () => {
         this.addLine("line");
     };
@@ -2177,6 +2181,10 @@ class InteractiveGraph extends React.Component<Props, State> {
         });
     }
 
+    static getNoneEquationString(): string {
+        return ""
+    }
+
     static getLinearEquationString(props: Props): string {
         const coords = InteractiveGraph.getLineCoords(props.graph, props);
         if (eq(coords[0][0], coords[1][0])) {
@@ -2655,6 +2663,8 @@ export function shouldUseMafs(
     }
 
     switch (graph.type) {
+        case "none":
+            return true;
         case "point":
             if (graph.numPoints === UNLIMITED) {
                 return Boolean(mafsFlags["unlimited-point"]);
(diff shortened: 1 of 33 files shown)


packages/perseus/src/widgets/interactive-graph.tsx
```diff
@@ -517,6 +517,10 @@ class LegacyInteractiveGraph extends React.Component<Props, State> {
         this.line?.remove();
     };
 
+    addNoneControls: () => void = () => {};
+
+    removeNoneControls: () => void = () => {};
```
nishasy, line 522: (No action needed) I don't like that we have to do this 😞 Are the mandatory `addXControls`, `removeXControls`, and `getXEquationString` something we can remove when the migration is complete?

Or at the very least refactor so that they're not being accessed by putting a string together 😬

nishasy (approved):
This is so exciting! Everything makes sense to me.

packages/perseus/src/widgets/interactive-graph.tsx
benchristel: Yes, we'll be able to remove these after the migration. This whole component should go away, in fact. The reason these definitions have to be here is that these function names are called dynamically 😱 by code that does stuff like this:

```ts
widget["add" + widgetType + "Controls"]();
```

nishasy:
One suggestion: It might be nice to add the None type graph to the [interactive graph editor stories](https://github.com/Khan/perseus/blob/dbac5d491386fc0daf813d583de02af55b02920f/packages/perseus-editor/src/__stories__/interactive-graph-editor.stories.tsx), probably named `InteractiveGraphNone`.

packages/perseus/src/widgets/interactive-graphs/interactive-graph-question-builder.ts
```diff
@@ -16,10 +17,12 @@ import type {
...
-export type LockedFunctionOptions = Omit<
-    Partial<LockedFunctionType>,
-    "type" | "equation"
->;
+export type LockedFunctionOptions = {
+    color?: LockedFigureColor;
+    strokeStyle?: LockedLineStyle;
+    directionalAxis?: "x" | "y";
+    domain?: Interval;
+};
```
mark-fitzgerald, line 25: Thank you for catching this.

nishasy:
I think the editor might need a couple of updates:
1. Remove the start coords UI for a none type graph (in [shouldShowStartCoordsUI()](https://github.com/Khan/perseus/blob/dbac5d491386fc0daf813d583de02af55b02920f/packages/perseus-editor/src/components/util.ts#L304))
2. Make sure that the locked figures settings [render in the editor](https://github.com/Khan/perseus/blob/dbac5d491386fc0daf813d583de02af55b02920f/packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx#L757-L759)

Screenshot 2024-09-13 at 1 55 34 PM

benchristel:
Thanks for pointing that out, @nishasy. I've updated the story and `shouldShowStartCoordsUI`. I refactored `shouldShowStartCoordsUI` to the structure that would have forced me to handle none-type graphs when I added them to PerseusGraphType.

mark-fitzgerald (approved):
Thanks for "nothing"

nishasy (approved):
Nice!! Thanks for the additional updates. This looks awesome 🎉

PR #1606, opened 2024-09-10
[🔥AUDIT🔥] Update snapshots

nishasy:
🖍 _This is an audit!_ 🖍

## Summary:
Looks like some snapshot updates fell through the cracks at some point.

They seem to be related to the progress spinner.

Issue: none

## Test plan:
`yarn jest`


nishasy:
This doesn't seem like the right solution for this. The tests are still failing, and clearing my local cache just causes the snapshot failures to come back.

PR #1607, opened 2024-09-10
[Interactive Graph Editor] Use Wonder Blocks TextArea in the graph description settings UI

nishasy:
## Summary:
Right now, we are using a native HTML <textarea> element inside the
graph description settings, since there were some issues importing
TextArea from wonder blocks.

It seems those issues have been resolved (or were local the whole
time and therefore nonexistent), so now we can update the element
we're using for the description input.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2332

## Test plan:
`yarn jest packages/perseus-editor/src/components/__tests__/interactive-graph-description.test.tsx`

Storybook
- http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-with-aria-label
- Confirm that the screenreader reads out the label and description as
  expected in Safari and Firefox.

image

### DIFF
--- a/.changeset/ninety-laws-decide.md
+++ b/.changeset/ninety-laws-decide.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": patch
+---
+
+[Interactive Graph Editor] Use Wonder Blocks TextArea in the graph description settings UI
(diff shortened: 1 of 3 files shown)


mark-fitzgerald (approved):
Easy-peasy!

PR #1610, opened 2024-09-10
[Interactive Graph Editor] Remove the start coords UI if the graph is static

nishasy:
## Summary:
When the graph is static, it uses the "correct" graph's coords to set
the coords on the now noninteractive graph. This makes the start coords
UI useless, as those coords will not be used.

- Remove the start coords UI if the graph is static

Issue: https://khanacademy.atlassian.net/browse/LEMS-2318

## Test plan:

Storybook
- http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-segment (and other graph types)
- Toggle between static and not static graphs
- Confirm that the start coords UI goes away when the graph is static, and the
  "correct" graph coords are used
- Confirm that the start coords UI is present when the graph is not static,
  and the start coords are used

https://github.com/user-attachments/assets/a613668b-c8fa-411e-8de0-6f924b6371e3


benchristel (approved):
🎉

nishasy:
The parent pull-request ([#1609](https://github.com/Khan/perseus/pull/1609)) has been merged into `main`, but this branch (`no-start-coords-static`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

PR #1612, opened 2024-09-11
Delete redundant throw statement

benchristel:
We were throwing an error with the message `"FIXME implement circle reducer"`
when `doMovePointInFigure` was called with a circle graph. The error message
didn't make sense because there's no need to implement `doMovePointInFigure`
for circles. The `throw` statement was redundant because without it, the circle
case just falls through to the `throw` below.

Issue: none

## Test plan:

`yarn test`


jeremywiebe (approved):
Makes sense!

PR #1615, opened 2024-09-12
Unlimited point graph align design for mouse/touch users

nicolecomputer:
## Summary:

https://github.com/user-attachments/assets/3ca36cc8-9cad-4827-a7d9-669b40f44278

Aligns the design to work for mouse and touch users

- The "Add point" button should never show
- When a point is clicked (and only when clicked) it should show the "remove point" button
- After the button is clicked the point is removed and the button is hidden
- The delete key should still work to remove a point

Issue: https://khanacademy.atlassian.net/browse/LEMS-1670

## Test plan:
- Run the tests, no special testing plan


packages/perseus/src/widgets/interactive-graphs/graphs/point.tsx
```diff
@@ -80,13 +81,19 @@ export function UnlimitedPointGraph(props: PointGraphProps) {
                     onMove={(destination) =>
                         dispatch(actions.pointGraph.movePoint(i, destination))
                     }
-                    onFocusChange={(isFocused) => {
+                    onFocusChange={(event, isFocused) => {
                         if (isFocused) {
                             dispatch(actions.pointGraph.focusPoint(i));
                         } else {
+                            if (event.relatedTarget?.id === REMOVE_BUTTON_ID) {
```
nicolecomputer, line 88: This prevents the button from stealing focus and lets us remove the `previouslySelectedPoint` field on the reducer. This makes this MUCH more tidy and straightforward to follow the logic. The downside is that this component needs to know about another components existence. To prevent these components from slipping away from each other I use a constant `REMOVE_BUTTON_ID` to make sure that should the id of that button ever change that this will keep working

nicolecomputer:
The main thing to look at is `packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts`. There's been a bunch of changes to the reducer including:
- Removing the previouslySelectedPoint field (I added a note about that too to this review)
- Adding a new event when a point is clicked (or tapped but I named it after the click because that's the event handler we bind to)
- Tweaking the logic of when the remove button shows and hides along with a new field that controls its presence on the page

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point-view.tsx
```diff
@@ -101,10 +101,14 @@ export const MovablePointView = forwardRef(
...
                 }
-                onFocus={() => {
-                    return getOnFocusChangeCallback(focusBehavior)(true);
+                onFocus={(event) => {
+                    const callback = getOnFocusChangeCallback(focusBehavior);
+                    callback && callback(event, true);
+                }}
+                onBlur={(event) => {
+                    const callback = getOnFocusChangeCallback(focusBehavior);
+                    callback && callback(event, false);
```
benchristel, line 110: Do you need to check for the existence of the callback before calling it? Seems like it will always be defined.

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.tsx
```diff
@@ -46,7 +48,10 @@ export const MovablePoint = (props: Props) => {
             color={color}
             dragging={dragging}
             focusBehavior={{type: "uncontrolled", tabIndex: 0, onFocusChange}}
-            onClick={() => elementRef.current?.focus()}
+            onClick={() => {
+                onClick && onClick();
```
benchristel, line 52: Same thing here, I think `onClick` will always be defined.

packages/perseus/src/widgets/interactive-graphs/graphs/point.tsx
benchristel: TIL about `relatedTarget`!

benchristel (approved):
Cool! LGTM! I left a couple suggestions inline.

One minor UI thing: I notice that when the delete button appears it causes the content below it to shift down. I wonder if we want to reserve space for the button with `visibility: hidden` or something, so it doesn't shift layout. Might be worth discussing with Caitlyn when she gets back — or we could just do it in her absence if you agree.

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point-view.tsx
nicolecomputer: image

Sooooo yes! I totally agree but typescript doesn't.

nicolecomputer:
I think I'll have a "cleanup" PR with internationalization and design tweaks. I'll address the button in that PR- I agree it's annoying to have things shift around

PR #1616, opened 2024-09-12
BUGFIX - Some Graphie Labels Were Misaligned

mark-fitzgerald:
## Summary:
The work done in LEMS-2022 addressed the scaling problems with graphie labels in one specific use-case. There are actually two use-cases for Graphie. The second use-case didn't include elements that the original solution relied upon. This bugfix accounts for the second use case, and adds in safeguards for any potential other use-case so that a default of 100% scale is used instead of a possible NaN.

Issue: LEMS-2373

## Test plan:
1. Launch Storybook
2. Review the following examples (labels should be in their proper location):
   * Grapher widget (specifically the [Linear Question](http://localhost:6006/?path=/story/perseus-widgets-grapher--linear-question), which doesn't use a background)
   * [Number Line](http://localhost:6006/?path=/story/perseus-widgets-number-line--question-1) widget
   * Plotter widget (though at this time, the Storybook entry doesn't render the widget, so will need to [test in an editor](http://localhost:6006/?path=/story/perseuseditor-editorpage--demo))
   * (Legacy) Interactive Graph widget - this applies to just the [unlimited points](http://localhost:6006/?path=/story/perseus-widgets-interactive-graph--point) and [polygon](http://localhost:6006/?path=/story/perseus-widgets-interactive-graph--polygon)
   * Measurer widget (though at this time, there is no Storybook entry, so will need to [test in an editor](http://localhost:6006/?path=/story/perseuseditor-editorpage--demo))
   * [Interaction](http://localhost:6006/?path=/story/perseus-widgets-interaction--question-1) widget

#### Before:
Grapher - Before

#### After:
Grapher - After

#### Before:
Number Line - Before

#### After:
Number Line - After

#### Before:
Plotter - Before

#### After:
Plotter - After

#### Before:
Interactive Graph - Before

#### After:
Interactive Graph - After

#### Before:
Interaction - Before

#### After:
Interaction - After

### DIFF
--- a/packages/perseus/src/util/graphie.ts
+++ b/packages/perseus/src/util/graphie.ts
@@ -1675,25 +1675,55 @@ const setLabelMargins = function (span: HTMLElement, size: Coord): void {
             marginTop: -height / 2 - y * scale,
         });
     } else {
-        const $container = $span.closest(".svg-image");
+        // We are using jQuery to collect information and calculate a scale
+        //     since we don't have a way to pass it to this function.
+        // We need the width of the container in order to calculate the scale to apply to the label.
+        // Unfortunately, the DOM tree leading to the label is not consistent.
+        // Therefore, we need to check different elements and different style properties to get the width of the container.
+        // Some graphie labels exist within an ".svg-image" container, and others do not.
+        // All graphie labels have a ".graphie" container.
+        // When a label has an ".svg-image" container,
+        //     the width information in the ".graphie" element is expressed as a percentage instead of pixels,
+        //     so we can't use it.
+        // Instead, the ".svg-image" container can be used to get the width information.
+        // The following lines determine which container element to use for the width measurement.
+        const $svgImage = $span.closest(".svg-image");
+        const $graphie = $span.closest(".graphie");
+        // If an ".svg-image" container exists in the DOM tree, then use it, otherwise use the ".graphie" container.
+        const $container = $svgImage.length > 0 ? $svgImage : $graphie;
+
         // Ensuring that the line-height of the text doesn't throw off placement of the text element.
         // Inherited line-height values can really mess up placement.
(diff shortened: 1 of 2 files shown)


packages/perseus/src/util/graphie.ts
```diff
@@ -1675,25 +1675,57 @@ const setLabelMargins = function (span: HTMLElement, size: Coord): void {
...
+        // All graphie labels have a ".graphie" container.
+        // When a label has an ".svg-image" container,
+        //     the width information in the ".graphie" element is expressed as a percentage instead of pixels,
+        //     so we can't use it.
+        // Instead, the ".svg-image" container can be used to get the width information.
+        // The following lines determine which container element to use for the width measurement.
+        const $svgImage = $span.closest(".svg-image");
+        const $graphie = $span.closest(".graphie");
+        // If an ".svg-image" container exists in the DOM tree, then use it, otherwise use the ".graphie" container.
+        const $container = $svgImage.length > 0 ? $svgImage : $graphie;
```
jeremywiebe, line 1693: I'm not certain it's cleaner or clearer, but what do you think about collapsing these three lines to simply:

```typescript
const $container = $span.closest(".svg-image, .graphie")
```

That should select the closest element that matches either selector from what I read.

jeremywiebe (approved):
That looks reasonable to me.  Thanks for testing across so many browsers etc.

packages/perseus/src/util/graphie.ts
mark-fitzgerald: I had that originally, but it doesn't give the needed results. Since there is always a `.graphie` ancestor that is "closer" than a possible `.svg-image`, the query would always return the `.graphie` container. We need the `.svg-container` if it exists, hence the need to split these queries out separately.

PR #1617, opened 2024-09-12
[Interactive Graph] Refactor the mafs graph so the locked labels layer is between the interactive layer and background layer ("Mafs sandwich")

nishasy:
## Summary:
The locked labels were showing up behind the whole graph. In order to allow the locked
labels to show up in front of the background but behind the interactive elements, we
need to separate all the layers and put the locked layer in between.

This also allows us to add a background (white, semi-transparent) to the locked label
so that it's more visible in front of the grid lines.

- Add a background to the locked labels
- Refactor the mafs graph
  - Move the locked labels layer after the (first) Mafs component
  - Remove the protractor and interactive elements from the (first) Mafs component
  - Create a new Mafs comopnent after the locked labels layer that contains the
    protractor and interactive elements.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2319

## Test plan:
`yarn jest`
`yarn typecheck`

Storybook
- Test locked labels: http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--mafs-with-locked-labels-flag
- Test for regressions: http://localhost:6006/?path=/docs/perseus-widgets-interactive-graph-visual-regression-tests--docs

## Screenshots:

Label in front of background elements:
image

Label behind interactive element:
image


packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -1,3 +1,14 @@
...
+ * Render the Mafs graph with the specified background and graph elements.
+ *
+ * Render order (back to front):
+ * - Grid
+ * - Axis Ticks, Axis Arrows, and Axis Labels
+ * - Locked Figures
+ * - Locked Labels
+ * - Protractor
+ * - Interactive Graph Elements
+ */
```
benchristel, line 11: Thanks for adding this comment! Makes me wonder if maybe each of the layers should be its own component, so the code can be self-documenting, e.g.:

```ts
<GridLayer state={state} {...props} />
<AxisLayer state={state} {...props} />
<LockedFiguresLayer state={state} {...props} />
<LockedLabelsLayer state={state} {...props} />
...
```

Maybe passing the state and props down explicitly wouldn't even be needed, since I think we have the graph context for that purpose? I don't remember exactly what's in there.

Anyway, no action needed. This refactoring could be done separately. I might just do it today.

benchristel (approved):
Looks beautiful! I love that the axis tick labels don't interfere with the gridlines now.

It works on Safari, right? 😬

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
nishasy: That would make sense to me!

PR #1618, opened 2024-09-12
feat(LEMS-2722): Add scientific option to basic keypad

anakaren-rojas:
#### Proof of concept: 
Adds scientific notation to basic keypad using toggle on expression editor

Issue: LEMS-2272

## Test plan:
1. [Navigate to expression editor story](https://650db21c3f5d1b2f13c02952-evogyhpohm.chromatic.com/?path=/docs/perseuseditor-widgets-expression-editor--docs)
2. Toggle `scientific` in the editor to see the scientific notation added to the basic keypad.

#### Note
When adding other keypads, if scientific notation is toggled, the scientific notation button will remain visible. This includes when scientific and pre-algebra are toggled simultaneously, causing there to be 2 scientific notation buttons displayed on the screen. 
If we decide to pursue this further - it would be key to identify the use cases when we would display and hide the button, despite the toggle being set to `true` 

## Screenshots

https://github.com/user-attachments/assets/55d7395d-bd93-4aad-8195-b321be71e376

### DIFF
--- a/packages/math-input/src/components/keypad/mobile-keypad-internals.tsx
+++ b/packages/math-input/src/components/keypad/mobile-keypad-internals.tsx
@@ -220,6 +220,7 @@ class MobileKeypadInternals
                                 containerWidth > expandedViewThreshold
                             }
                             showDismiss
+                            scientific={keypadConfig?.scientific}
                         />
                     ) : null}
                 </AphroditeCssTransitionGroup>
--- a/packages/perseus/src/components/__tests__/math-input.test.tsx
+++ b/packages/perseus/src/components/__tests__/math-input.test.tsx
@@ -4,17 +4,18 @@ import * as React from "react";
 
 import {testDependencies} from "../../../../../testing/test-dependencies";
 import * as Dependencies from "../../dependencies";
-import MathInput from "../math-input";
+import MathInput, {KeypadButtonSets} from "../math-input";
 
 import type {UserEvent} from "@testing-library/user-event";
 
-const allButtonSets = {
+const allButtonSets: KeypadButtonSets = {
     advancedRelations: true,
     basicRelations: true,
     divisionKey: true,
     logarithms: true,
     preAlgebra: true,
     trigonometry: true,
+    scientific: true,
 };
 
 describe("Perseus' MathInput", () => {
@@ -140,6 +141,33 @@ describe("Perseus' MathInput", () => {
 
         // Assert
         expect(mockOnChange).toHaveBeenLastCalledWith("1+2-3");
+    });
+
+    it("is possible to use the scientific keypad", async () => {
+        // Assemble
+        const mockOnChange = jest.fn();
+        render(
+            <MathInput
+                onChange={mockOnChange}
+                keypadButtonSets={{scientific: true}}
+                analytics={{onAnalyticsEvent: () => Promise.resolve()}}
+                convertDotToTimes={false}
+                value=""
+            />,
+        );
+        act(() => jest.runOnlyPendingTimers());
+
+        // Act
+        await userEvent.click(
+            screen.getByRole("button", {name: /open math keypad/}),
+        );
+        await userEvent.click(screen.getByRole("button", {name: "2"}));
+        await userEvent.click(screen.getByRole("button", {name: "Custom exponent"}));
+        await userEvent.click(screen.getByRole("button", {name: "2"}));
+        act(() => jest.runOnlyPendingTimers());
+
+        // Assert
+        expect(mockOnChange).toHaveBeenLastCalledWith("2^{2}");
     });
 
     it("is possible to use buttons with legacy props", async () => {
--- a/packages/perseus/src/components/__tests__/math-input.test.tsx
+++ b/packages/perseus/src/components/__tests__/math-input.test.tsx
@@ -4,8 +4,9 @@ import * as React from "react";
 
 import {testDependencies} from "../../../../../testing/test-dependencies";
 import * as Dependencies from "../../dependencies";
-import MathInput, {KeypadButtonSets} from "../math-input";
+import MathInput from "../math-input";
 
+import type {KeypadButtonSets} from "../math-input";
 import type {UserEvent} from "@testing-library/user-event";
 
 const allButtonSets: KeypadButtonSets = {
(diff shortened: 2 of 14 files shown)


packages/perseus/src/components/__tests__/math-input.test.tsx
```diff
@@ -31,7 +33,7 @@ describe("Perseus' MathInput", () => {
     });
 
     it("renders", () => {
-        // Assemble
+        // Arrange
```
handeyeco, line 36: Good catch, thanks for finding these.

jeremywiebe, line 36: Thanks for fixing these!

packages/perseus/src/perseus-types.ts
```diff
@@ -424,7 +425,7 @@ export type LegacyButtonSets = ReadonlyArray<
 export type PerseusExpressionWidgetOptions = {
     // The expression forms the answer may come in
     answerForms: ReadonlyArray<PerseusExpressionAnswerForm>;
-    // Different buttons sets that can show in the expression. Options are "basic", "basic+div", "trig", "prealgebra", "logarithms", "basic relations", "advanced relations"
+    // Different buttons sets that can show in the expression. Options are "basic", "basic+div", "trig", "prealgebra", "scientific", "logarithms", "basic relations", "advanced relations"
```
jeremywiebe, line 428: I would rather see us remove the entire list of options in this comment.  We use TypeScript to define the union of available values and re-declaring them here is just a recipe for getting things out of sync.

packages/perseus/src/widgets/expression/expression.test.tsx
```diff
@@ -613,3 +618,90 @@ describe("Expression Widget", function () {
         });
     });
 });
+
+describe("Keypad configuration", () => {
```
handeyeco, line 622: Cool, thanks for adding tests for this helper.

packages/perseus/src/widgets/expression/expression.test.tsx
```diff
@@ -613,3 +618,90 @@ describe("Expression Widget", function () {
...
+            times: false,
+            scientific: false,
+            extraKeys: ["PI"],
+        };
+
+        // Act
+        const result = keypadConfigurationForProps(widgetOptions);
+
+        // Assert
+        expect(result).toEqual(expected);
```
jeremywiebe, line 643: This may be preference, but I think it'd be much easier to trace this test if the expected were just declared here inline. Otherwise my brain has to backtrack up the test to find where it was defined (and it's not needed before this line): 

```suggestion
        expect(result).toEqual({
             keypadType: KeypadType.EXPRESSION,
             times: false,
             scientific: false,
             extraKeys: ["PI"],
         });
```

packages/perseus/src/widgets/expression/expression.tsx
```diff
@@ -745,7 +745,12 @@ const keypadConfigurationForProps = (
         extraKeys = ["PI"];
     }
 
-    return {keypadType, extraKeys, times: widgetOptions.times};
+    return {
+        keypadType,
+        extraKeys,
+        times: widgetOptions.times,
+        scientific: widgetOptions.buttonSets.includes("scientific"),
```
handeyeco, line 752: I forget, why do we have to add this for the `scientific` button set but not the rest of them? Should we comment why this particular button set is special?

packages/perseus/src/widgets/expression/expression.test.tsx
jeremywiebe: No action required, but I think it's almost the same for the `widgetOptions`.

packages/math-input/src/components/keypad/mobile-keypad-internals.tsx
```diff
@@ -220,6 +220,7 @@ class MobileKeypadInternals
                                 containerWidth > expandedViewThreshold
                             }
                             showDismiss
+                            scientific={keypadConfig?.scientific}
```
handeyeco, line 223: Small nit, wondering if this should be up with the rest of the button sets (even though it's triggered differently). Also, does this need to be `scientific={isExpression && keypadConfig?.scientific}`? Or do we expect this to be used on the fraction keypad too?

handeyeco (approved):
Looks good. I especially appreciate the amount of test coverage.

Just some nits/questions, but don't let me hold you up.

jeremywiebe (approved):
Nice! :)   I'll defer to Matthew on the specifics of the implementation.

catandthemachines (approved):

packages/perseus/src/widgets/expression/expression.tsx
anakaren-rojas: I'm having trouble remembering as well 🤔
I tested without this, and it doesn't _seem_ to have an impact, so I'll remove it, but I'm sure if we do end up implementing this, we will quickly figure out why this was needed 😅

packages/math-input/src/components/keypad/mobile-keypad-internals.tsx
anakaren-rojas: Since we weren't able to get full requirements from content, I'm not sure if they will ever have a use case of scientific + fraction, this question would definitely deserve some time to flesh out if we decide to implement.

packages/perseus/src/components/math-input.tsx
```diff
@@ -32,13 +32,14 @@ import type {Keys, MathFieldInterface} from "@khanacademy/math-input";
 
 type ButtonsVisibleType = "always" | "never" | "focused";
 
-type KeypadButtonSets = {
+export type KeypadButtonSets = {
```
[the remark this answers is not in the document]
anakaren-rojas: This export was mainly for supporting tests - I figured it would be better to use the existing `KeypadButtonSets` type than creating a new one in the test suite and having to remember to keep it in sync with the real one

packages/math-input/src/components/keypad/mobile-keypad-internals.tsx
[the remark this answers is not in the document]
anakaren-rojas: Yes! At least it did using the storybook "mobile experience"

PR #1619, opened 2024-09-13
Remove fullscreen button if the URL is broken

aemandine:
## Summary:
<Add your summary here>

Issue: LEMS-2278

## Test plan:

PR #1620, opened 2024-09-13
Add PhET widget code coverage

aemandine:
## Summary:
Add new tests to ensure maximal code coverage of the PhET widget. This is based on the new code from the [a11y update, PR #1619](https://github.com/Khan/perseus/pull/1619) and should be merged in after the a11y update is merged in.
* Check that fullscreen button renders when the simulation loads successfully
* Check that fullscreen button does not render when the simulation fails to load
* Check that the locale warning banner renders when the `kaLocale` is not available for the requested PhET simulation

Issue: LEMS-2277

## Test plan:
* `yarn jest packages/perseus/src/widgets/phet-simulation/phet-simulation.test.ts`

### DIFF
--- a/.changeset/thick-eels-wait.md
+++ b/.changeset/thick-eels-wait.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Add code coverage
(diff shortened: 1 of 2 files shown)


SonicScrewdriver (approved):
Lovely!! Thank you :D

PR #1622, opened 2024-09-14
Fix types for 'classnames' package

jeremywiebe:
## Summary:

We didn't have types for the `classnames` package. That led us to try to define the types for it's return value in-place. That led us to suppress Typescript errors incorrectly. 

I added the version of the types used by webapp for the same version of `classnames` package we're using and voila, fixes.

Issue: "none"

## Test plan:

`yarn tsc`
`yarn test`

### DIFF
--- a/packages/perseus/src/util.ts
+++ b/packages/perseus/src/util.ts
@@ -9,6 +9,7 @@ import type {Range} from "./perseus-types";
 import type {PerseusStrings} from "./strings";
 import type {Widget, PerseusScore} from "./types";
 import type {KEScore} from "@khanacademy/perseus-core";
+import type * as React from "react";
 
 type WordPosition = {
     start: number;
@@ -791,7 +792,7 @@ const supportsPassiveEvents: () => boolean = () => {
  * Pass this function as the touchstart for an element to
  * avoid sending the touch to the mobile scratchpad
  */
-function captureScratchpadTouchStart(e: TouchEvent) {
+function captureScratchpadTouchStart(e: React.TouchEvent) {
     e.stopPropagation();
 }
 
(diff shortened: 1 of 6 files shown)


packages/perseus/src/widgets/radio/base-radio.tsx
```diff
@@ -383,15 +382,13 @@ const BaseRadio = function (props: Props): React.ReactElement {
...
-                            // @ts-expect-error - TS2322 - Type 'HTMLLIElement | null' is not assignable to type 'null'.
                             ref={(e) => (listElem = e)}
-                            // @ts-expect-error - TS2322 - Type 'readonly string[]' is not assignable to type 'string'.
                             className={className}
-                            // @ts-expect-error - TS2322 - Type 'null' is not assignable to type 'MouseEventHandler<HTMLLIElement> | undefined'.
                             onClick={clickHandler}
-                            // @ts-expect-error - TS2322 - Type '((e: TouchEvent) => void) | null' is not assignable to type 'TouchEventHandler<HTMLLIElement> | undefined'.
                             onTouchStart={
-                                !labelWrap ? null : captureScratchpadTouchStart
+                                labelWrap
```
jeremywiebe, line 389: The React event handlers are defined as `React.TouchEvent<T> | undefined`.  If we assign `null`, then we have to suppress Typescript. So changing to use `undefined` instead of `null` fixes that.

handeyeco (approved):
Very nice

packages/perseus/src/util.ts
```diff
@@ -9,6 +9,7 @@ import type {Range} from "./perseus-types";
 import type {PerseusStrings} from "./strings";
 import type {Widget, PerseusScore} from "./types";
 import type {KEScore} from "@khanacademy/perseus-core";
+import type * as React from "react";
```
anakaren-rojas, line 12: We could also add `@types/react` as a dev dependency 
https://www.npmjs.com/package/@types/react

jeremywiebe: We do have that as a dependency in the root `package.json` (https://github.com/Khan/perseus/blob/main/package.json#L61) :)

PR #1626, opened 2024-09-16
Add noop validator

handeyeco:
## Summary:
Consolidate validation logic for widgets that don't have anything to validate

Issues:

- https://khanacademy.atlassian.net/browse/LEMS-2370
- https://khanacademy.atlassian.net/browse/LEMS-2365
- https://khanacademy.atlassian.net/browse/LEMS-2366
- https://khanacademy.atlassian.net/browse/LEMS-2361
- https://khanacademy.atlassian.net/browse/LEMS-2358
- https://khanacademy.atlassian.net/browse/LEMS-2359

## Test plan:
Nothing should change, just moving around some code.

### DIFF
--- a/packages/perseus/src/widgets/__shared__/noop-validator.tsx
+++ b/packages/perseus/src/widgets/__shared__/noop-validator.tsx
@@ -1,0 +1,20 @@
+import type {PerseusScore} from "../../types";
+
+/**
+ * Several widgets don't have "right"/"wrong" validation logic,
+ * so this just says to move on past those widgets
+ *
+ * TODO: widgets that use this probably shouldn't have any
+ * validation logic and the thing validating an exercise
+ * should just know to skip these
+ */
+function noopValidator(...args: any[]): PerseusScore {
+    return {
+        type: "points",
+        earned: 0,
+        total: 0,
+        message: null,
+    };
+}
+
+export default noopValidator;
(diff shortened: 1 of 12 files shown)


packages/perseus/src/widgets/__shared__/noop-validator.tsx
```diff
@@ -0,0 +1,20 @@
+import type {PerseusScore} from "../../types";
+
+/**
+ * Several widgets don't have "right"/"wrong" validation logic,
+ * so this just says to move on past those widgets
+ *
+ * TODO: widgets that use this probably shouldn't have any
+ * validation logic and the thing validating an exercise
+ * should just know to skip these
```
jeremywiebe, line 9: +++  I think there is already some logic in `renderer.tsx` that knows how to skip scoring a widget if it doesn't have `simpleValidate`. Maybe we should make this more explicit by adding a flag on `WidgetExports<T>` that declares if the widget is scorable.

https://github.com/Khan/perseus/blob/f083200340545f41275f0696dbfc967f45028b0c/packages/perseus/src/renderer.tsx#L1830

packages/perseus/src/widgets/__shared__/noop-validator.tsx
```diff
@@ -0,0 +1,20 @@
...
+ */
+function noopValidator(...args: any[]): PerseusScore {
```
jeremywiebe, line 11: **nit**: accepting weakly typed `args` like this feels like it will just be confusing at the call sites. I think it'd be clearer to not accept any arguments.

jeremywiebe (approved):

PR #1628, opened 2024-09-16
Split out validation logic for dropdown

handeyeco:
## Summary:
Split out validation logic for the Dropdown widget

Issue: LEMS-2357

## Test plan:

### DIFF
--- a/packages/perseus/src/widgets/dropdown/dropdown-validator.test.tsx
+++ b/packages/perseus/src/widgets/dropdown/dropdown-validator.test.tsx
@@ -1,0 +1,61 @@
+import dropdownValidator from "./dropdown-validator";
+import {question1} from "./dropdown.testdata";
+
+import type {UserInput} from "./dropdown.types";
+
+describe("dropdownValidator", () => {
+    it("returns invalid for user input of 0", () => {
+        // Arrange
+        const userInput: UserInput = {
+            value: 0,
+        };
+        const rubric = question1.widgets["dropdown 1"].options;
+
+        // Act
+        const result = dropdownValidator(userInput, rubric);
+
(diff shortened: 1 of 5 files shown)


packages/perseus/src/widgets/dropdown/dropdown-validator.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+        const rubric = question1.widgets["dropdown 1"].options;
+
+        // Act
+        const result = dropdownValidator(userInput, rubric);
+
+        // Assert
+        expect(result).toEqual({
+            message: null,
+            type: "invalid",
+        });
```
jeremywiebe, line 21: **thought**: As I'm reading through, I'm seeing that we'll be doing alot of assertions across all our validators. Today, we have some Jest custom matchers in `config/test/custom-matchers.ts` for the `Renderer`. I wonder if it would make our tests a little more verbose if we created a set of custom matchers for the validators also?  

`expect(result).toHaveInvalidInput()` 

or 

`expect(result).toHaveBeenAnsweredCorrectly()`

jeremywiebe (approved):

packages/perseus/src/widgets/dropdown/dropdown-validator.test.tsx
handeyeco: I'm usually not a huge fan of custom matchers (I think they can sometimes make tests a little opaque), but in this case I think it would make a lot of sense to have a `toBeCorrect`, `toBeIncorrect`, `toBeInvalid` or something.

https://khanacademy.atlassian.net/browse/LEMS-2395

PR #1629, opened 2024-09-16
Remove WidgetJsonifyDeprecated

handeyeco:
## Summary:
I was looking through static functions in widgets for SSS and I noticed this weird thing.

## Test plan:
Nothing should change, cleanup


packages/perseus/src/mixins/widget-jsonify-deprecated.ts
```diff
@@ -1,23 +0,0 @@
-/* Free implementation of getUserInput. This should be used sparingly, since it
- * just returns all the widget's props rather than picking out those which were
- * input by the user.
- */
-import _ from "underscore";
-
-import WIDGET_PROP_DENYLIST from "./widget-prop-denylist";
```
jeremywiebe, line 7: I'm guessing we can't _quite_ delete this file yet, right? `widget-prop-denylist.ts` I mean.

jeremywiebe (approved):
Thanks. Eventually I'd like to get rid of the `widget-prop-denylist` entirely, but I think that's a few steps away yet. This is an improvement!

packages/perseus/src/mixins/widget-jsonify-deprecated.ts
handeyeco: Looks like it's used by a few things, but I noted in our Futures document to investigate if it could be removed.

PR #1630, opened 2024-09-16
Move most widgets from PropTypes to TS

handeyeco:
## Summary:
Went through most widgets and switched from PropTypes to TS. Still need to do Measurer.

## Test plan:
No functionality changes (hopefully), just type changes.


packages/perseus-editor/src/widgets/table-editor.tsx
```diff
@@ -93,7 +93,7 @@ class TableEditor extends React.Component<Props> {
     };
 
     render(): React.ReactNode {
-        const tableProps = _.pick(
+        const tableProps: any = _.pick(
```
handeyeco, line 96: I don't think this inferred type was correct and it was causing an error.

packages/perseus/src/components/__stories__/sortable.stories.tsx
```diff
@@ -19,7 +19,7 @@ export const SortableHorizontalExample = (
 ): React.ReactElement => {
     return (
         <Sortable
-            layout={Layout.HORIZONTAL}
+            layout={"horizontal"}
```
handeyeco, line 22: We don't use enums anymore, so I made this a string union.

packages/perseus-editor/src/widgets/table-editor.tsx
jeremywiebe: tbh, I'd much rather we remove `tableProps` entirely and pass the props to the `<Table>` component on line 156 explicitly.

packages/perseus/src/widgets/cs-program/cs-program.tsx
```diff
@@ -15,10 +14,42 @@ import Util from "../../util";
...
+    status: Status;
+    message: string | null;
+};
+
+type Setting = {
+    name: string;
+    value: any;
+};
+
+type Props = {
```
jeremywiebe, line 33: EDIT: I got through the rest and see you've done this on other widgets. I'm guessing you had a reason why you didn't do it here here. I'll leave the comment for posterity. 

Could you work `PerseusCSProgramWidgetOptions` in here instead of defining the props all independently?  I realize that the TS compiler matches by type but using the widget options defined in `perseus-types.ts` helps to paint the picture of which options are part of the schema.

packages/perseus/src/widgets/cs-program/cs-program.tsx
```diff
@@ -121,15 +138,15 @@ class CSProgram extends React.Component<any> {
         return Changeable.change.apply(this, args);
     };
 
-    simpleValidate: (arg1: any) => any = (rubric) => {
+    simpleValidate(rubric): PerseusScore {
```
jeremywiebe, line 141: I _wish_ we could figure out how to enforce that our Perseus widget React component conform to an interface. 🤔

jeremywiebe (approved):
Nice! One day, we'll be able to drop the `prop-types` dependency!! :)

packages/perseus/src/widgets/cs-program/cs-program.tsx
handeyeco: No, it was just one of my early conversions and I hadn't picked up the right pattern yet. Fixed, good catch.

packages/perseus/src/widgets/cs-program/cs-program.tsx
handeyeco: Yeah or if we didn't have external methods on components 😅

PR #1632, opened 2024-09-16
Adding back a nested svg element to fix Interactive Graph regression

SonicScrewdriver:
## Summary:
While deploying Perseus, it was discovered that there was a regression related to the bounding of the interactive elements and the protractor in the Interactive Graph widget. 

Upon investigation, it seems like the nested svg (which is containing the overlow) was accidentally removed. This PR adds the nested svg back.

Issue: LEMS-2392

## Screenshots:
Screenshot 2024-09-16 at 4 11 33 PM

Screenshot 2024-09-16 at 4 12 21 PM

## Test plan:
- Manual Testing

### DIFF
--- a/.changeset/perfect-taxis-do.md
+++ b/.changeset/perfect-taxis-do.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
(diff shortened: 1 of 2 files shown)


mark-fitzgerald (approved):
Great catch!

PR #1636, opened 2024-09-17
Small tweaks to non-interactive validation logic

handeyeco:
## Summary:
- Split out Measurer validation logic
- Go back to other non-interactive widgets and detach `staticValidate` from `validate`

## Test plan:
Nothing should change, just some tweaks

### DIFF
--- a/packages/perseus/src/widgets/explanation/explanation.test.ts
+++ b/packages/perseus/src/widgets/explanation/explanation.test.ts
@@ -302,10 +302,7 @@ describe("Explanation", function () {
 
     describe("static validate", () => {
         it("should always return 0 points", async () => {
-            const result = ExplanationWidgetExports.widget.validate(
-                {},
-                question1.widgets["explanation 1"].options,
-            );
+            const result = ExplanationWidgetExports.widget.validate();
 
             expect(result).toMatchInlineSnapshot(`
                 {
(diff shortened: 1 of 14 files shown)


packages/perseus/src/widgets/explanation/explanation.test.ts
```diff
@@ -302,10 +302,7 @@ describe("Explanation", function () {
 
     describe("static validate", () => {
         it("should always return 0 points", async () => {
-            const result = ExplanationWidgetExports.widget.validate(
-                {},
-                question1.widgets["explanation 1"].options,
-            );
+            const result = ExplanationWidgetExports.widget.validate();
```
Myranae, line 305: Question: Should this be a call to the validator function instead? I know it eventually gets to it as it goes through `validate()`, but it sounded like we wanted to replace all of those with the validator function.

Myranae (approved):
Looks good!

packages/perseus/src/widgets/explanation/explanation.test.ts
jeremywiebe: My $0.02 worth. :) I think leaving it like this for now is good. Keeps our changes more incremental. We'll do another round of separating the React components completely from validation.

packages/perseus/src/widgets/measurer/measurer-validator.ts
```diff
@@ -0,0 +1,10 @@
+function measurerValidator() {
+    return {
+        type: "points",
+        earned: 1,
+        total: 1,
+        message: null,
+    };
```
jeremywiebe, line 7: Why couldn't my teachers grade like this when I was in school?? 😂  💯%

jeremywiebe (approved):

PR #1637, opened 2024-09-17
[Locked Figure Labels] View locked point labels

nishasy:
## Summary:

- Add feature flag for locked point labels
- Add `labels` to LockedPointType
- Update builder
- Update stories and tests

Issue: https://khanacademy.atlassian.net/browse/LEMS-2346

## Test plan:
- Go to the following story: http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--mafs-with-locked-point-labels-flag
- Confirm that there is a label "A" next to the locked point
- Got o the following story: http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--mafs-with-locked-labels-flag
- Confirm there is no label "A" next to the locked point

image


packages/perseus-editor/src/__stories__/interactive-graph-editor.stories.tsx
```diff
@@ -171,7 +171,18 @@ MafsWithLockedFiguresCurrent.parameters = {
...
 export const MafsWithLockedLabelsFlag = (): React.ReactElement => {
     return (
-        <EditorPageWithStorybookPreview question={segmentWithLockedFigures} />
+        <EditorPageWithStorybookPreview
+            apiOptions={{
+                flags: {
+                    mafs: {
+                        ...flags.mafs,
+                        "interactive-graph-locked-features-labels": true,
+                        "locked-point-labels": false,
```
benchristel, line 180: Is `locked-point-labels` supposed to be `false` here?

packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.tsx
```diff
@@ -15,9 +15,10 @@ import {getGradableGraph, getRadius} from "./reducer/interactive-graph-state";
 import type {InteractiveGraphProps, InteractiveGraphState} from "./types";
 import type {PerseusGraphType} from "../../perseus-types";
 import type {Widget} from "../../renderer";
+import type {APIOptions} from "../../types";
 
 export type StatefulMafsGraphProps = {
-    showLabelsFlag?: boolean;
+    flags?: APIOptions["flags"];
```
benchristel, line 21: I wonder if it would make sense to put the flags in a React context if we want to make them generally available to the various components. No action needed.

benchristel (approved):
LGTM!

packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.tsx
nishasy: YES! I wanted to do this during that one cooldown sprint, but never got around to it. I even think we have a [ticket for this](https://khanacademy.atlassian.net/browse/LEMS-2136), but it's marked as Won't Do.

packages/perseus-editor/src/__stories__/interactive-graph-editor.stories.tsx
nishasy: Yes, because it's for the story with general labels but not point labels. Just for some flag testing.

PR #1638, opened 2024-09-17
[Locked Figure Labels] Locked points: Add the locked point labels UI to the editor

nishasy:
## Summary:
- Add the ability to add/edit/delete locked point labels from locked point settings
- Pass down flags
- Add tests

Issue: https://khanacademy.atlassian.net/browse/LEMS-2346

## Test plan:
- Go to http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--mafs-with-locked-point-labels-flag
- Go to locked figures
- Open the locked point settings
- confirm that the label settings are there
- Play around with the label settings and confirm that it moves
  and updates with the point

image


packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-point-settings.tsx
```diff
@@ -77,8 +127,54 @@ const LockedPointSettings = (props: Props) => {
...
+                            const newLabel = {
+                                ...getDefaultFigureForType("label"),
+                                // Place the label at the same position
+                                // as the point. Add a small offset to
+                                // avoid crowding.
+                                coord: [
+                                    coord[0] + 0.5,
+                                    // Additional offset for each label so
+                                    // they don't overlap.
+                                    coord[1] - 1 * labels?.length,
```
benchristel, line 161: These coords are in graph units, right? I think these offsets might not work well for graphs that have ranges other than `[[-10, 10], [-10, 10]]`. I guess that's okay, though, because these values are just defaults?

benchristel (approved):
LGTM!

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-point-settings.tsx
nishasy: Hmmm... that's a good point. I'm not sure how much it would be worth calculating the offset, since I suspect authors would change the position in most cases anyway? I guess we could get feedback from them and then update accordingly in the future.

PR #1639, opened 2024-09-17
Rough out UserInput type

handeyeco:
## Summary:
I might have gone the wrong direction with this (as it is now).

I thought it would be beneficial to have widget Rubrics and UserInput in one place: `validation.types.ts`. This way we have the information we need from the learner and the information we need to determine if the user input is correct/incorrect.

However as I was going through this, I started to feel like we misnamed something: `WidgetProps<RenderProps, Rubric>`. For some of these widgets, they were the same thing! For instance definition:

``` TS
type RenderProps = PerseusDefinitionWidgetOptions;

type Rubric = PerseusDefinitionWidgetOptions;

type UserInput = Empty;

type DefinitionProps = WidgetProps<RenderProps, Rubric> & {
    widgets: PerseusRenderer["widgets"];
};
```

RenderProps is the same as Rubric and then we just stick on another random prop (not to mention we have UserInput even though it's empty). On top of this, when we use `WidgetProps<RenderProps, Rubric>`, we only use `Rubric` for `reviewModeRubric` which is only used by a couple of components. Here's my guess at what's going on:

- WidgetOptions: these were meant to represent stored configuration options for widgets - the stuff we keep in the database.
- RenderProps: these were meant to be the "transformed" configuration options that we use when rendering - we take the WidgetOptions and transform them to the props we hand the widget (for instance if we need to upgrade the widget version).
- Rubric: this is the stuff we need to grade the widget - Expression might need `times` and `buttonSets` to _display_ the widget, but it only needs `answerForms` to _grade_ the user input.
- My guess is `WidgetProps` has `reviewModeRubric` as a legacy way of configuring or grading widgets. My guess is we need:
  - `WidgetProps<RenderProps>`, removing/replacing `reviewModeRubric`
  - `RenderProps` should be scoped to just the stuff it needs (per widget) to render, relying on the types from the widget's options when possible
  - `WidgetOptions` per widget should continue to be focused on our external API
  - `Rubric` per widget should be focused on just the stuff we need to grade the widget
  - `UserInput` should be focused on just the things the learner sends for a widget to be graded

Expression is a good example:

``` TS
// The public API, what gets passed from the Perseus JSON
type PerseusExpressionWidgetOptions = {
    buttonSets: LegacyButtonSets;
    times: boolean;
    visibleLabel?: string;
    ariaLabel?: string;
    buttonsVisible?: "always" | "never" | "focused";
};

// The props it needs to render, note it drops `answerForms` and adds `keypadConfiguration`
type RenderProps = {
    buttonSets: PerseusExpressionWidgetOptions["buttonSets"];
    buttonsVisible?: PerseusExpressionWidgetOptions["buttonsVisible"];
    functions: PerseusExpressionWidgetOptions["functions"];
    times: PerseusExpressionWidgetOptions["times"];
    visibleLabel: PerseusExpressionWidgetOptions["visibleLabel"];
    ariaLabel: PerseusExpressionWidgetOptions["ariaLabel"];
    keypadConfiguration: ReturnType<typeof keypadConfigurationForProps>;
};

type Rubric = PerseusExpressionWidgetOptions; // <- how it is now is not specific enough
// It should only have the stuff it needs to grade with
type Rubric = {
}

// The user input type should be whatever the user sends for grading
type UserInput = string;

// Props should *only* be combining the widget-specific render props
// and the shared, universal widget props
// or a little more obvious IMO
type Props = WidgetProps & RenderProps;
// or very obvious by explicitly saying which props are used
type Props = RenderProps & {
  onBlur: WidgetProps["onBlur"];
{
```

The more I look at it, the more I think it's kind of a mess. Anyway this PR doesn't address any of this. But I _was_ intentional about creating a specific `UserInput` and `Rubric` type for most widgets even though they need further refinement.

Issue: LEMS-2328

## Test plan:


packages/perseus/src/widgets/categorizer/categorizer.tsx
```diff
@@ -17,12 +17,15 @@ import Util from "../../util";
...
 import type {PerseusCategorizerWidgetOptions} from "../../perseus-types";
 import type {PerseusStrings} from "../../strings";
 import type {PerseusScore, WidgetExports, WidgetProps} from "../../types";
+import type {
+    PerseusCategorizerRubric,
+    PerseusCategorizerUserInput,
+} from "../../validation.types";
 
-type Props = WidgetProps<RenderProps, Rubric> & {
+type Props = WidgetProps<RenderProps, PerseusCategorizerRubric> & {
```
jeremywiebe, line 28: **thought**: Originally, when Perseus lived in webapp, we built the types using a `Perseus` prefix.  Today with Perseus being in its own repo and npm package, I feel like we could begin moving away from prefixing everything with `Perseus`.  It's unrelated to this PR, but seeing all the `PerseusXyz` types brought it to my mind again.

packages/perseus/src/widgets/label-image/label-image.tsx
```diff
@@ -66,6 +66,7 @@ type Point = {
     y: number;
 };
 
+// TODO: should this be using WidgetProps / PerseusLabelImageWidgetOptions?
```
jeremywiebe, line 69: Yes.  Or... we use this approach for all widget `RenderProps` types where we explicitly define each key. I think it makes this more verbose but potentially clearer. @benchristel I think you've had thoughts about this before, haven't you? 

Although not part of this PR, this should also be named `RenderProps` for consistency.

jeremywiebe (approved):
Nice. Really like the standardization and I agree with moving all these types. When we split things more aggressively for server-side scoring, this will be extremely helpful.

packages/perseus/src/types.ts
```diff
@@ -552,6 +552,9 @@ export type FilterCriterion =
       ) => boolean);
 
 // NOTE: Rubric should always be the corresponding widget options type for the component.
+// TODO: in fact, is it really the rubric? WidgetOptions is what we use to configure the widget
+// (which is what this seems to be for)
+// and Rubric is what we use to score the widgets (which not all widgets need validation)
```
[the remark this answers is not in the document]
handeyeco: https://khanacademy.atlassian.net/browse/LEMS-2437

PR #1640, opened 2024-09-17
Split out validation logic for input-number

Myranae:
## Summary:
For the server-side scoring project, we are moving validation logic out of each widget component. This moves the validation logic for the input-number widget.

Issue: LEMS-2355

## Test plan:
- All tests should pass
- Nothing should change as code is just being moved into separate files


packages/perseus/src/widgets/input-number/input-number-validator.test.ts
```diff
@@ -0,0 +1,262 @@
...
+            inexact: false,
+            maxError: 0.1,
+            answerType: "number",
+        } as const;
+        if (!transform) {
+            throw new Error("transform not defined");
+        }
+        const widgetProps = transform(editorProps, mockStrings);
+        expect(_.has(widgetProps, "value")).toBe(false);
+    });
```
Myranae, line 239: This isn't related to scoring, but I didn't want to have all of the item info in two separate places. Feel free to let me know if that is preferred to having this test here though. Thanks!

jeremywiebe (commented):
Some duplicate comments from what Matthew already pointed out.  I wrote these up yesterday and forgot to submit my review. :)

packages/perseus/src/widgets/input-number/input-number-validator.ts
```diff
@@ -0,0 +1,100 @@
...
+    pi: {
+        name: "Numbers with pi",
+        forms: "pi",
+    },
+} as const;
+
+function inputNumberValidator(
+    state: {
+        currentValue: string;
+    },
```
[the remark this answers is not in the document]
Myranae: Question: Should I wait for that PR to land then to reference `UserInput` or just copy over the `PerseusInputNumberUserInput` type for now?

packages/perseus/src/widgets/input-number/input-number-validator.ts
```diff
@@ -0,0 +1,100 @@
...
+    rubric: Rubric,
+    strings: PerseusStrings,
+    onInputError: APIOptions["onInputError"] = () => {},
```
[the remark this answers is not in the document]
Myranae: I think we decided to handle this during a separate stage of the validation separation process, right? Should I just leave this here for now then? Maybe just adding a TODO to fix these when I find them would be the best thing for now

packages/perseus/src/widgets/input-number/input-number-validator.ts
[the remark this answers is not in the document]
handeyeco: Don't wait to land it. I can fix this in my PR.

packages/perseus/src/widgets/input-number/input-number-validator.ts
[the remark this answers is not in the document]
handeyeco: Don't do anything for now, we'll circle back. I think he was just thinking out loud.

packages/perseus/src/widgets/input-number/input-number-validator.test.ts
```diff
@@ -0,0 +1,145 @@
+import {testDependencies} from "../../../../../testing/test-dependencies";
+import * as Dependencies from "../../dependencies";
+import {mockStrings} from "../../strings";
+
+import inputNumberValidator from "./input-number-validator";
+
+import type {Rubric} from "./input-number.types";
+
+describe("static function validate", () => {
```
handeyeco, line 7:
```suggestion
describe("inputNumberValidator", () => {
```
It's not a static method in this context.

packages/perseus/src/widgets/input-number/input-number-validator.test.ts
```diff
@@ -0,0 +1,145 @@
...
+    beforeEach(() => {
+        jest.spyOn(Dependencies, "getDependencies").mockReturnValue(
+            testDependencies,
+        );
+    });
```
handeyeco: We might need this, but I would try to run the tests without it and if they pass then remove it.

packages/perseus/src/widgets/input-number/input-number.test.ts
```diff
@@ -12,12 +12,13 @@ import {mockStrings} from "../../strings";
...
 
 import InputNumber from "./input-number";
+import inputNumberValidator from "./input-number-validator";
 import {question3 as question} from "./input-number.testdata";
 
 import type {
     PerseusInputNumberWidgetOptions,
     PerseusRenderer,
-} from "../../perseus-types";
+} from "@khanacademy/perseus";
```
handeyeco: That's weird, I think this should have been a linting error. 🤔 I think we have a linter that prevents packages from importing from themselves like this.

I don't know why the linter isn't mad, but I would revert this.

handeyeco (approved):

packages/perseus/src/widgets/input-number/input-number-validator.test.ts
Myranae: Already removed it :D

packages/perseus/src/widgets/input-number/input-number.test.ts
Myranae: Already reverted it ^_^ Thank you! Trying to keep an eye out for things like this after you pointed out that strings import issue

PR #1641, opened 2024-09-18
Split out validation logic for categorizer

Myranae:
## Summary:
For the servier-side scoring project, we are moving validation logic out of each widget component. This moves the validation logic for the categorizer widget.

Issue: LEMS-2341

## Test plan:
- All tests should pass
- Nothing should chang as code is just being moved into separate files

### DIFF
--- a/packages/perseus/src/widgets/categorizer/categorizer-validator.test.ts
+++ b/packages/perseus/src/widgets/categorizer/categorizer-validator.test.ts
@@ -4,7 +4,7 @@ import categorizerValidator from "./categorizer-validator";
 
 import type {Rubric} from "./categorizer.types";
 
-describe("validating answers", () => {
+describe("categorizerValidator", () => {
     it("gives points when the answer is correct", () => {
         const rubric: Rubric = {
             items: ["Graph $1$", "Graph $2$"],
--- a/packages/perseus/src/widgets/categorizer/categorizer-validator.test.ts
+++ b/packages/perseus/src/widgets/categorizer/categorizer-validator.test.ts
@@ -1,98 +1,8 @@
-import {screen} from "@testing-library/react";
-import {userEvent as userEventLib} from "@testing-library/user-event";
-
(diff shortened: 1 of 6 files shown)


packages/perseus/src/widgets/categorizer/categorizer-validator.test.ts
```diff
@@ -0,0 +1,185 @@
...
+                "Nonlinear relationship",
+            ],
+            highlightLint: false,
+            static: false,
+        };
+
+        const useInput = {
+            values: [1, 3],
+        } as const;
+        const score = Categorizer.validate(useInput, rubric, mockStrings);
```
handeyeco, line 116: We need to replace all instances of `[Widget].validate` and `[Widget].simpleValidate` with the pure `widgetValidator` function. No need to go looking for uses, but in files we touch we should be making this change.

handeyeco (changes requested):
Same feedback as from input-number, need to completely decouple from React and ideally replace object snapshot tests with `.toEqual`.

Also replace component method calls with the pure validator where possible.

packages/perseus/src/widgets/categorizer/categorizer-validator.ts
```diff
@@ -0,0 +1,34 @@
+import type {Rubric, UserInput} from "./categorizer.types";
+import type {PerseusStrings} from "../../strings";
+import type {PerseusScore} from "@khanacademy/perseus";
```
handeyeco, line 3: We should avoid importing from `@khanacademy/perseus` when we're within `@khanacademy/perseus`. The solution is relative imports.

handeyeco (approved):
Just a small ask regarding the import

packages/perseus/src/widgets/categorizer/categorizer-validator.ts
Myranae: 🤦‍♀️ Thank you! Fixed!

PR #1643, opened 2024-09-18
Make Perseus widgets open by default

anniegallagher:
## Summary:
This PR sets the default state of Perseus widgets to be open. Additionally, it introduces a new optional prop to the EditorPage component, which is passed down to the WidgetEditor. This prop allows the visibility of all widgets on a page to be controlled with a single variable, enabling them to be opened or closed simultaneously.

This work is a precursor to planned work for the v2 exercise editor in Webapp in which a switch will be added to provide the ability to toggle widget visibility across the entire page.

Issue: [CP-8491](https://khanacademy.atlassian.net/browse/CP-8491) 

## Test plan:
Add a temporary switch to the editor-page which toggles the widget's visibility to ensure the widget opens/closes as expected. Test in Perseus storybook.

https://github.com/user-attachments/assets/b6b480ba-52a6-4bed-8648-98884acaf92a

Used the npm snapshot in webapp to verify this works with multiple widgets on the page.

https://github.com/user-attachments/assets/c08c88d5-f64e-4587-8314-de21517b1cc2

[CP-8491]: https://khanacademy.atlassian.net/browse/CP-8491?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.changeset/three-trees-deny.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": minor
+---
+
+Make widgets open by default. Add ability to open/close all widgets on a page externally.
```
jeremywiebe, line 5: I'd like to work a slight term change through this PR. The "widget" is what the learner sees. The "widget editor" or simply "editor" is what the content author uses to adjust the settings for a widget.

So, could we say "Make widget editors expanded by default. ..."?

Although we don't use the WB Accordion, I think using the same terms is useful.  They use the term "expanded" and "collapsed" for open/closed.

packages/perseus-editor/src/components/widget-editor.tsx
```diff
@@ -30,6 +30,7 @@ type WidgetEditorProps = {
     ) => unknown;
     onRemove: () => unknown;
     apiOptions: APIOptions;
+    widgetIsOpen?: boolean;
```
jeremywiebe, line 33: So with the terms I propose, what do you think of simply having this prop be `expanded?: boolean`?

jeremywiebe (approved):
I have a few comments, but once you address them, you're free to land this. There are a few other changes that will be part of this release, mostly in the interactive graph editor. These are fine to go out together. Usually someone from the Perseus team volunteers to do the release (land the "Version Packages" PR in the Perseus repo and then bump the version in webapp and deploy).  If you take the lead on that, just post in #team-lems-perseus checking if it's safe to deploy.

packages/perseus-editor/src/components/widget-editor.tsx
anniegallagher: Are you suggesting `expanded` be the prop from the `editor-page` down to the `widget-editor` or just for the `widget-editor` portion of the code?

jeremywiebe: I think I might be splitting hairs here.  I was thinking this component would use `expanded` but as we move up that means the other components should change to be consistent. Changing prop names is relatively easy so why don't we just leave things as you have them and land the PR?  

If we see that over time it's confusing or we find a better set of terms, we can change it.

anniegallagher: Okay, thanks!

PR #1644, opened 2024-09-18
[Locked Figure Labels] Add/edit/delete locked line labels

nishasy:
## Summary:
Add the ability to add/edit/delete locked line labels in the
interactive graph editor.

Note: locked labels within the points that define a locked line
are coming in a follow-up PR.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2347

## Test plan:
- Go to http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--mafs-with-locked-line-labels-flag
- Go to locked figures
- Open the locked line settings
- Confirm that the label settings are there
- Play around with the label settings and confirm that it moves
and updates with the line

image

### DIFF
--- a/packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-line-settings.tsx
+++ b/packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-line-settings.tsx
@@ -4,7 +4,7 @@
  *
  * Used in the interactive graph editor's locked figures section.
  */
-import {vector as kvector, line as kline} from "@khanacademy/kmath";
+import {vector as kvector} from "@khanacademy/kmath";
 import Button from "@khanacademy/wonder-blocks-button";
 import {View} from "@khanacademy/wonder-blocks-core";
 import {OptionItem, SingleSelect} from "@khanacademy/wonder-blocks-dropdown";
@@ -13,6 +13,7 @@ import {color as wbColor, spacing} from "@khanacademy/wonder-blocks-tokens";
 import {LabelMedium, LabelLarge} from "@khanacademy/wonder-blocks-typography";
 import plusCircle from "@phosphor-icons/core/regular/plus-circle.svg";
 import {StyleSheet} from "aphrodite";
+import {vec} from "mafs";
 import * as React from "react";
 
 import PerseusEditorAccordion from "../../../components/perseus-editor-accordion";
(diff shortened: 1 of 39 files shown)


packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-line-settings.tsx
```diff
@@ -167,6 +226,55 @@ const LockedLineSettings = (props: Props) => {
...
+                        onClick={() => {
+                            const midpoint = [
+                                ...kline.midpoint([
+                                    points[0].coord,
+                                    points[1].coord,
+                                ]),
+                            ] as Coord;
+                            // Additional vertical offset for each label so
+                            // they don't overlap.
+                            midpoint[1] -= 1 * labels.length;
```
benchristel, line 258: Why multiply by `1` here? Also, mutating the array is slightly sketch (though I see you made a copy above) - if for no other reason than because after the mutation it isn't the midpoint anymore. I might suggest using vector addition instead:

```ts
const offsetPerLabel: vec.Vector2 = [0, -1];

const labelLocation = vec.add(
  vec.scale(offsetPerLabel, labels.length),
  vec.midpoint(...),
);
  
```

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-line-settings.tsx
```diff
@@ -167,6 +226,55 @@ const LockedLineSettings = (props: Props) => {
...
+                    <Button
+                        kind="tertiary"
+                        startIcon={plusCircle}
+                        onClick={() => {
+                            const midpoint = [
+                                ...kline.midpoint([
+                                    points[0].coord,
+                                    points[1].coord,
+                                ]),
+                            ] as Coord;
```
benchristel, line 255: Same thing here.

```suggestion
                            const midpoint =
                                vec.midpoint(
                                    points[0].coord,
                                    points[1].coord,
                                );
```

benchristel (approved):
I left a few suggestions inline. LGTM overall, though!

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-line-settings.tsx
nishasy: Ah, I added the `* 1` to make it more clear that that's the offset I'm using, but it a bit silly in hindsight. I like your approach way better.

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-line-settings.tsx
```diff
@@ -69,12 +77,40 @@ const LockedLineSettings = (props: Props) => {
...
+            points[0].coord,
+            points[1].coord,
+        ]) as Coord;
+        const newMidpoint = [
+            ...kline.midpoint([newPoints[0].coord, newPoints[1].coord]),
+        ] as Coord;
+        const offset = [
+            newMidpoint[0] - oldMidpoint[0],
+            newMidpoint[1] - oldMidpoint[1],
+        ] as Coord;
```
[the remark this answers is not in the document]
benchristel: Oh, interesting... was that because we didn't want to worry about the Mafs versions getting out of sync between perseus and perseus-editor?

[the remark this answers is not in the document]
nishasy: I think @jeremywiebe might have thoughts on this?

PR #1645, opened 2024-09-18
[SSS] Lint validators for banned imports

jeremywiebe:
## Summary:

This PR is mostly to gather feedback about this approach.  We can extend the rules to cover other things we discover that shouldn't be used in the validators. 

We started talking about what patterns and types we'll have available to us in the new validators. In order to make it a bit easier nd have fewer things to remember, I tried out using some `no-restricted-syntax` ESLint rules to note symbols we won't/can't use in server-side scoring.

image

Issue: "none"

## Test plan:

Import the `APIOptions` type into the `packages/perseus/src/widgets/expression/expression-validator.ts` file. Note that ESLint complains.  Try the same by importing `react`.


handeyeco:
I like it. It feels like it might be a bit fragile (the `.validator.ts` name is a convention, not a rule), but I think it strikes a good balance between keeping the logic next to the components and having the validator be detached from React.

Myranae:
Definitely appreciate more linters to catch things that shouldn't be done. Agree with Matthew's comment on the naming convention, but not sure on a better approach.

jeremywiebe:
> I like it. It feels like it might be a bit fragile (the `.validator.ts` name is a convention, not a rule), but I think it strikes a good balance between keeping the logic next to the components and having the validator be detached from React.

Although not a hard rule, I'd want to have a really good reason not to name a file containing validators _not_ `.validator.ts`.  At that point we could amend the linter. 

I guess the only thing I can think of is needing to break down a large file.  But maybe then what we do is create a `validation` folder to hold all the validation logic.  That would be easy to check for in the linter also.

packages/perseus/src/types.ts
```diff
@@ -175,7 +175,7 @@ export const InteractiveGraphLockedFeaturesFlags = [
 export type APIOptions = Readonly<{
     isArticle?: boolean;
     onInputError?: (
-        widgetId: any,
+        widgetId: string | null,
```
jeremywiebe, line 178: Just a tiny clarification on this type.

jeremywiebe:
@Myranae @handeyeco do you have any final concerns before I land this?

Myranae:
No concerns from me! Looks like the checks want it to have a changeset though XD

handeyeco (approved):
Thank you!

PR #1646, opened 2024-09-18
Add Keyboard support for unlimited points

nicolecomputer:
## Summary:
Adds keyboard support for unlimited point graphs.

https://github.com/user-attachments/assets/2ae468e1-82e0-4127-b371-bc4ad42a2ee3

- Adds a new overlay when the graph is focused that prompts the learner to enter keyboard mode when shift+return is pressed
- When in keyboard mode a button to "Add point" exists
- There is no way to leave keyboard mode for a graph once it is entered (this is by design)
- The delete button and the add point button are never shown at the same time

Figma: https://www.figma.com/design/JroPokeCLBv2VRIw3GPvY5/LEMS%3A-Interactive-Graph-Widget?node-id=2285-24101&node-type=frame&t=qhAvSlmYoTKgv1IN-0

Issue: [LEMS-1669](https://khanacademy.atlassian.net/browse/LEMS-1669)

## Test plan:

[LEMS-1669]: https://khanacademy.atlassian.net/browse/LEMS-1669?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -254,6 +258,29 @@ export const MafsGraph = (props: MafsGraphProps) => {
...
+                                    padding: "16px 0",
+                                    boxShadow: "0px 8px 8px 0px #21242C14",
+
+                                    // This translates the box to the center of the
+                                    // graph Then backs it off by half of its
+                                    // overall height so it's perfectly centered
+                                    top: "50%",
+                                    transform: "translateY(-50%)",
+                                }}
+                            >
```
mark-fitzgerald, line 277: Does the `<LabelMedium>` element need to be wrapped by this `<View>` element, or can the styling here just be applied to the label itself?

mark-fitzgerald:
Interesting approach for keyboard navigation. I like how the button only shows for keyboard users, and how the instructions only show when the graph first gains focus. I wonder about whether additional instructions are needed for learners. For instance, do they know what it means to "interact with the graph" after pressing shift+enter?

I understand that using shift+enter is a user-initiated way of showing the "Add Point" button. I wonder if we could do the same by assuming that getting focus on the graph would indicate that the learner is indeed using the keyboard. So, tab to the graph and the button automatically shows (aria on the graph could reference the button for adding points).

nicolecomputer:
> Interesting approach for keyboard navigation. I like how the button only shows for keyboard users, and how the instructions only show when the graph first gains focus. I wonder about whether additional instructions are needed for learners. For instance, do they know what it means to "interact with the graph" after pressing shift+enter?

These are all great points and Caitlyn has done a lot of thinking about this. This is the design she arrived at https://www.figma.com/design/JroPokeCLBv2VRIw3GPvY5/LEMS%3A-Interactive-Graph-Widget?node-id=2285-24101&node-type=frame&t=qhAvSlmYoTKgv1IN-0 and what we're going to ship

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
nicolecomputer: Yes! We do need the View. It helps with layout. Here's without it:

image

packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.tsx
```diff
@@ -116,12 +116,19 @@ export const StatefulMafsGraph = React.forwardRef<
...
         // after the first render. This is an optimization, but also prevents
         // a bug where the graph would be marked "incorrect" during grading
         // even if the user never interacted with it.
         if (latestPropsRef.current !== originalPropsRef.current) {
-            dispatch(reinitialize(latestPropsRef.current));
+            dispatch(
+                reinitialize({
+                    ...latestPropsRef.current,
+                }),
+            );
```
benchristel, line 131: Does this section need to change? I don't think copying the props will have any effect.

packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.tsx
```diff
@@ -116,12 +116,19 @@ export const StatefulMafsGraph = React.forwardRef<
     const originalPropsRef = useRef(props);
     const latestPropsRef = useLatestRef(props);
     useEffect(() => {
+        if (state.hasBeenInteractedWith) {
+            return;
+        }
```
benchristel, line 121: I think this change might cause problems for graph previews in the content editor. Once you interact with the preview, it will never reflect changes to the graph settings (`numPoints`, `numSegments`, etc.) What's the intent of this change? Is there another way it could be accomplished?

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-action.ts
```diff
@@ -150,6 +155,35 @@ function clickPoint(index: number): ClickPoint {
...
+    return {
+        type: CHANGE_INTERACTION_MODE,
+        mode,
+    };
+}
+
+export const CHANGE_KEYBOARD_INVITATION_VISIBILITY =
+    "change-keyboard-interaction-invitation-visbility";
+
+export interface ChangeKeyboardInvitationVibility {
```
benchristel, line 174:
```suggestion
export interface ChangeKeyboardInvitationVisibility {
```

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-action.ts
```diff
@@ -150,6 +155,35 @@ function clickPoint(index: number): ClickPoint {
...
+}
+function changeInteractionMode(mode: InteractionMode): ChangeInteractionMode {
+    return {
+        type: CHANGE_INTERACTION_MODE,
+        mode,
+    };
+}
+
+export const CHANGE_KEYBOARD_INVITATION_VISIBILITY =
+    "change-keyboard-interaction-invitation-visbility";
```
benchristel, line 172:
```suggestion
    "change-keyboard-interaction-invitation-visibility";
```

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-action.ts
```diff
@@ -150,6 +155,35 @@ function clickPoint(index: number): ClickPoint {
...
+
+export interface ChangeKeyboardInvitationVibility {
+    type: typeof CHANGE_KEYBOARD_INVITATION_VISIBILITY;
+    shouldShow: boolean;
+}
+function changeKeyboardInvitationVibility(
```
benchristel, line 178:
```suggestion
function changeKeyboardInvitationVisibility(
```

benchristel (commented):
This is so cool! Thank you for all the work on this epic-scale feature.

I left some questions inline about the changes to `stateful-mafs-graph.tsx` - once those are resolved I think this will be good to go.

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
mark-fitzgerald: I think that's because the `<View>` component has `flexbox` automatically applied to it. I would think that adding flexbox properties to the `<LabelMedium>` would accomplish the desired effect. Just looking for ways to reduce `<div>` tags, not a blocker.

mark-fitzgerald (approved):
Ben found some interesting things to consider, but everything else that I reviewed looks great.

packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.tsx
nicolecomputer: I would love to work through some way to make this work. When reinitialize is called it wipes away the flag to show the keyboard invitation. The re-initialization is getting called as soon as the graph has focus and so the prompt doesn't get seen until the second time.

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -70,6 +74,17 @@ export function interactiveGraphReducer(
 ): InteractiveGraphState {
     switch (action.type) {
         case REINITIALIZE:
+            // For unlimited point graphs reinitialize is getting called when
```
nicolecomputer, line 77: @benchristel I moved this here. Does this feel better?

PR #1650, opened 2024-09-19
[Locked Figure Labels] View locked vector labels

nishasy:
## Summary:
- Add feature flag for locked vector labels
- Add `labels` to LockedVectorType
- Update builder
- Update stories and tests

Issue: https://khanacademy.atlassian.net/browse/LEMS-2348

## Test plan:
- Go to http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--mafs-with-locked-vector-labels-flag
- Confirm that the visible label is on the locked vector _only_
- Go to the point and line stories and confirm that the locked
  vector does not have a visible label (make sure the flags are
  separating the behaviors correctly)

image

### DIFF
--- a/.changeset/strange-houses-breathe.md
+++ b/.changeset/strange-houses-breathe.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/perseus": minor
+"@khanacademy/perseus-editor": minor
+---
+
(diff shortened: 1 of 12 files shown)


benchristel (approved):
LGTM!

PR #1651, opened 2024-09-19
Measurer and DeprecatedStandin to use noopValidator

handeyeco:
## Summary:
Just made sense to share this logic

Issue: LEMS-2371

## Test plan:
No logic should change


packages/perseus/src/widgets/__shared__/noop-validator.test.tsx
```diff
@@ -2,6 +2,26 @@ import noopValidator from "./noop-validator";
...
+    it("can be configured to be pointful", () => {
+        const result = noopValidator(1);
+        expect(result).toHaveBeenAnsweredCorrectly();
+
+        // This is is just to narrow the type
+        // the expect above checks correctness
+        if (result.type === "points") {
+            expect(result.earned).toBe(1);
+            expect(result.total).toBe(1);
+        }
```
Myranae, line 25: Praise: Nice! This is super handy. Definitely makes it more versatile.

Myranae (approved):
Looks good :)

packages/perseus/src/widgets/__shared__/noop-validator.tsx
```diff
@@ -8,11 +8,11 @@ import type {PerseusScore} from "../../types";
  * validation logic and the thing validating an exercise
  * should just know to skip these
  */
-function noopValidator(): PerseusScore {
+function noopValidator(points: number = 0): PerseusScore {
```
jeremywiebe, line 11: The comment of this validator now seems slightly wrong. With the addition of a parameter we can make a validator that always returns "correct" (points).

jeremywiebe (commented):
I like this. Just might want to circle back on the comment for the `noopValidator`.

PR #1654, opened 2024-09-19
Omit extra properties from InteractiveGraph onChange arg

benchristel:
Previously, we were passing the entire Mafs graph state to onChange.  Since
the onChange callback data is used by the Interactive Graph widget editor to
set the correct answer for the graph, a bunch of extra data like
`"hasBeenInteractedWith": true` was getting saved in the `correct` field of
the Widget data.

This extra data is not used, so its presence in datastore is potentially
confusing to future maintainers trying to understand the data. It might
also cause bugs in the future - e.g. if some code merges it into another
object in which those properties _are_ meaningful.

This PR ensures that we only save relevant data in the `correct` field of
interactive graph widgets.

Issue: none

## Test plan:

Edit, save and publish an interactive graph widget in an exercise. As a
learner, you should be able to complete the exercise successfully.

### DIFF
--- a/packages/perseus/src/types.ts
+++ b/packages/perseus/src/types.ts
@@ -9,7 +9,6 @@ import type {
 } from "./perseus-types";
 import type {PerseusStrings} from "./strings";
 import type {SizeClass} from "./util/sizing-utils";
-import type {InteractiveGraphState} from "./widgets/interactive-graphs/types";
 import type {KeypadAPI} from "@khanacademy/math-input";
 import type {AnalyticsEventHandlerFn} from "@khanacademy/perseus-core";
 import type {LinterContextProps} from "@khanacademy/perseus-linter";
--- a/packages/perseus/src/types.ts
+++ b/packages/perseus/src/types.ts
@@ -3,6 +3,7 @@ import type {Item} from "./multi-items/item-types";
 import type {
     Hint,
     PerseusAnswerArea,
+    PerseusGraphType,
     PerseusWidget,
     PerseusWidgetsMap,
 } from "./perseus-types";
@@ -90,7 +91,7 @@ export type ChangeHandler = (
         // perseus-all-package/widgets/grapher.jsx
         plot?: any;
         // Interactive Graph callback (see legacy: interactive-graph.tsx)
-        graph?: InteractiveGraphState;
+        graph?: PerseusGraphType;
     },
     callback?: () => unknown | null | undefined,
     silent?: boolean,
--- a/packages/perseus/src/widgets/interactive-graph.tsx
+++ b/packages/perseus/src/widgets/interactive-graph.tsx
@@ -1656,7 +1656,6 @@ class LegacyInteractiveGraph extends React.Component<Props, State> {
 
         $(this.angle).on("move", () => {
             this.onChange({
-                // @ts-expect-error Type '{ coords: any; type: "angle"; showAngles?: boolean | undefined; allowReflexAngles?: boolean | undefined; angleOffsetDeg?: number | undefined; snapDegrees?: number | undefined; match?: "congruent" | undefined; }' is not assignable to type 'InteractiveGraphState | undefined'.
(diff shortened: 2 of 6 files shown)


packages/perseus/src/types.ts
```diff
@@ -90,7 +91,7 @@ export type ChangeHandler = (
         // perseus-all-package/widgets/grapher.jsx
         plot?: any;
         // Interactive Graph callback (see legacy: interactive-graph.tsx)
-        graph?: InteractiveGraphState;
+        graph?: PerseusGraphType;
```
benchristel, line 94: This type was wrong - InteractiveGraphState is the internal state of our StatefulMafsGraph component - we don't want it to leak outside that component. `PerseusGraphType` is what gets saved in the widget JSON.

packages/perseus/src/widgets/interactive-graph.tsx
```diff
@@ -1656,7 +1656,6 @@ class LegacyInteractiveGraph extends React.Component<Props, State> {
 
         $(this.angle).on("move", () => {
             this.onChange({
-                // @ts-expect-error Type '{ coords: any; type: "angle"; showAngles?: boolean | undefined; allowReflexAngles?: boolean | undefined; angleOffsetDeg?: number | undefined; snapDegrees?: number | undefined; match?: "congruent" | undefined; }' is not assignable to type 'InteractiveGraphState | undefined'.
```
benchristel, line 1659: The incorrect type above was causing this error.

packages/perseus/src/widgets/interactive-graphs/mafs-state-to-interactive-graph.ts
benchristel, line 1: TypeScript forces this function to be unfortunately verbose, but all it's doing is converting the circle point coords to `center` and `radius`, and passing the `coords` and `type` through for all the other graph types.

packages/perseus/src/widgets/interactive-graphs/mafs-state-to-interactive-graph.ts
```diff
@@ -0,0 +1,55 @@
...
+// Converts the state of a StatefulMafsGraph back to the format used to
+// represent graph state in the widget JSON.
+//
+// Rather than be tightly bound to how data was structured in
+// the legacy interactive graph, this lets us store state
+// however we want and we just transform it before handing it off
+// the the parent InteractiveGraph.
+//
+// The transformed state is used in the interactive graph widget editor, to
+// set the `correct` field of the graph options.
```
benchristel, line 16: @jeremywiebe @nishasy I'd especially like your feedback on this comment. Is this correct? Does anything besides the editor handle widgets' `onChange` event?

I found some dead test prep code in webapp that was using `onChange`, but that doesn't seem worth mentioning here since I believe it's about to be deleted. I also had a vague memory that we saved learners' progress to the backend on change, but that didn't seem to be the case when I tested - perhaps that feature was only in test prep.

jeremywiebe: The `renderer` itself sends an `onChange` callback to widgets where it updates its state (widget state). 

I also think that some widgets that accept user input send an `onChange` event with updated prop values and expect to be re-rendered with these new values. For example, the `label-image` widget does this ([Github](https://github.com/Khan/perseus/blob/a6ec402c0e41dae1dbd980106265929ebe761bce/packages/perseus/src/widgets/label-image/label-image.tsx#L432)).

jeremywiebe: So I don't think the editor is the only thing that uses `onChange`. I am not sure as I've never dug through, but I've long wondered if we need such an all-encompassing `onChange` handler type or if we could break it apart by widget.

jeremywiebe (approved):
I think we could use a slight tweak to the comment (based on my comments in-line), but the code changes look good!

PR #1656, opened 2024-09-23
Split out validation logic for sorter

Myranae:
## Summary:
For the server-side scoring project, we are moving validation logic out of each widget component. This moves the validation logic for the sorter widget and updates types. It also adds tests for the validator function as no tests were present for the sorter widget at all previously.

Issue: LEMS-2369

## Test plan:
- Confirm all tests pass
- Nothing should change. Code is being moved and types are being updated to be more specific.


packages/perseus/src/widgets/sorter/sorter-validator.ts
```diff
@@ -0,0 +1,17 @@
+import _ from "underscore";
+
+import type {Rubric, UserInput} from "./sorter.types";
+import type {PerseusScore} from "../../types";
+
+function sorterValidator(userInput: UserInput, rubric: Rubric): PerseusScore {
+    const correct = _.isEqual(userInput.values, rubric.correct);
```
Myranae, line 7: Would love to remove this `_.isEqual`, but didn't know of or find a quick way to do so.

packages/perseus/src/widgets/sorter/sorter.tsx
```diff
@@ -42,15 +42,8 @@ class Sorter extends React.Component<Props, State> {
...
-    static validate(state: any, rubric: any): any {
-        const correct = _.isEqual(state.options, rubric.correct);
-
-        return {
-            type: "points",
-            earned: correct ? 1 : 0,
-            total: 1,
-            message: null,
-        };
+    static validate(userInput: UserInput, rubric: Rubric): PerseusScore {
```
Myranae, line 45: Let me know if I should leave the old types here. My brain was having trouble wrapping around the older types, so I just fixed it.

packages/perseus/src/widgets/sorter/sorter-validator.test.ts
```diff
@@ -0,0 +1,41 @@
+import sorterValidator from "./sorter-validator";
+
+import type {Rubric, UserInput} from "./sorter.types";
+
+describe("SorterValidator", () => {
```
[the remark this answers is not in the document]
jeremywiebe: Looking at that code, it depends in the React components `state`.  

What we could do is include `changed` in the `UserInput` and then move this logic from `simpleValidate` to the new static validator. 

What do you think about that approach?

packages/perseus/src/widgets/sorter/sorter-validator.test.ts
```diff
@@ -0,0 +1,41 @@
...
+    it("gives points when the answer is correct", () => {
```
jeremywiebe, line 6: Could we define these tests rather as the description of what makes the answer correct or incorrect? Maybe something like: 

```js
   it("is correct when the user input values are in the same order as the rubric defines"...
```

When I read test names, I think if I'd have some idea of what broke when it fails. Just stating "is correct" can be said for all widget validators and doesn't point to what the test is actually validating.

packages/perseus/src/widgets/sorter/sorter-validator.ts
jeremywiebe: It looks like our `deepEq()` function in https://github.com/Khan/perseus/blob/12b8e01bf89abbc88d6bfc9bf243d5a0e95b5ed3/packages/perseus/src/util.ts#L547 would do what we need.

packages/perseus/src/widgets/sorter/sorter.tsx
jeremywiebe: Nope, this is good how you've done this. We want to trend to stronger typing where possible (especially migrating away from `any` whenever we can). So I'm a 👍  to this.

jeremywiebe (approved):
Nice work Tamara!

packages/perseus/src/widgets/sorter/sorter-validator.test.ts
```diff
@@ -0,0 +1,41 @@
...
+            layout: "horizontal",
+            correct: ["$0.005$ kilograms", "$15$ grams", "$55$ grams"],
+        };
+        const result = sorterValidator(userInput, rubric);
+        expect(result).toEqual({
+            type: "points",
+            earned: 1,
+            total: 1,
+            message: null,
+        });
```
handeyeco, line 21:
```suggestion
        expect(result).toHaveBeenAnsweredCorrectly();
```

packages/perseus/src/widgets/sorter/sorter-validator.test.ts
```diff
@@ -0,0 +1,41 @@
...
+            layout: "horizontal",
+            correct: ["$0.005$ kilograms", "$15$ grams", "$55$ grams"],
+        };
+        const result = sorterValidator(userInput, rubric);
+        expect(result).toEqual({
+            type: "points",
+            earned: 0,
+            total: 1,
+            message: null,
+        });
```
handeyeco, line 39:
```suggestion
        expect(result).toHaveBeenAnsweredIncorrectly();
```

packages/perseus/src/widgets/sorter/sorter-validator.ts
Myranae: Wow! Thank you so much! Below is how I used it, and all tests pass!
```
import Util from "../../util";
...
const correct = Util.deepEq(userInput.values, rubric.correct);
```

packages/perseus/src/widgets/sorter/sorter.tsx
```diff
@@ -84,12 +77,9 @@ class Sorter extends React.Component<Props, State> {
...
     };
 
-    getUserInput: () => {
-        options: any;
-    } = () => {
-        // eslint-disable-next-line react/no-string-refs
+    getUserInput: () => UserInput = () => {
         // @ts-expect-error - TS2339 - Property 'getOptions' does not exist on type 'ReactInstance'.
-        return {options: this.refs.sortable.getOptions()};
+        return {values: this.refs.sortable.getOptions()};
```
handeyeco, line 82: I don't think we should be changing the names of things where possible. This is an external API and we don't know what it might be used by.

packages/perseus/src/widgets/sorter/sorter-validator.test.ts
Myranae: Thanks for pointing these both out! Updated. Will use the invalid one for the invalid test when I get it working :)

packages/perseus/src/widgets/sorter/sorter-validator.test.ts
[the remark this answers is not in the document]
Myranae: Sounds great! Was able to add it to UserInput and refactor code to include it. Also was able to move over the invalid logic and add a related test! Thanks~ :D

packages/perseus/src/widgets/sorter/sorter-validator.test.ts
Myranae: Updated the test names. Thank you! :)

packages/perseus/src/widgets/sorter/sorter.tsx
Myranae: Reverted the naming. Thanks! :)

packages/perseus/src/widgets/sorter/sorter.tsx
```diff
@@ -84,12 +77,14 @@ class Sorter extends React.Component<Props, State> {
...
     };
 
-    getUserInput: () => {
-        options: any;
-    } = () => {
+    getUserInput: () => UserInput = () => {
         // eslint-disable-next-line react/no-string-refs
         // @ts-expect-error - TS2339 - Property 'getOptions' does not exist on type 'ReactInstance'.
-        return {options: this.refs.sortable.getOptions()};
+        const options = this.refs.sortable.getOptions();
```
Myranae, line 83: Had to pull this out so it is right beneath the `@ts-expect-error`. Wasn't exactly sure how to fix this hour, but it seemed like fixing it would take some larger refactoring, so left it alone.

packages/perseus/src/widgets/sorter/sorter-validator.test.ts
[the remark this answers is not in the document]
Myranae: @handeyeco mentioned changing the name of `options` to `values` in `getUserInput` might be bad since this is an external API and we don't know how it's used. Could modifying the return object of `getUserInput` be similarly bad? If I'm just adding something that no other code is currently using, it should be fine, right?

packages/perseus/src/widgets/sorter/sorter.tsx
jeremywiebe: This is fine.  The reason for the `@ts-expect-error` is that `this.refs` is an old React ref system that TypeScript can't properly type. If we were to refactor, we'd make a field within this class named `sortableRef = React.createRef<Sortable>();` and then use that as the ref, which would then have proper types. 

I could pair with you this week to work through that refactoring if you'd like.

packages/perseus/src/widgets/sorter/sorter-validator.test.ts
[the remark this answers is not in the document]
jeremywiebe: I think you're safe if you're adding a field. There _may_ be call sites that call `Sorter.validate(...)` in webapp, but we'll need to visit those sites when we full remove `static validate` from our widgets so we'll need to ensure that a `changed` flag is passed through properly for this widget.

packages/perseus/src/widgets/sorter/sorter.tsx
```diff
@@ -101,21 +96,8 @@ class Sorter extends React.Component<Props, State> {
...
-        // to make a change before trying to move forward. This makes an
-        // assumption that the initial order isn't the correct order! However,
-        // this should be rare if it happens, and interacting with the list
-        // will enable the button, so they won't be locked out of progressing.
-        if (!this.state.changed) {
-            return {
-                type: "invalid",
-                message: null,
-            };
-        }
```
handeyeco, line 116: Ah, lame. I thought all the `simpleValidate`s just grabbed user input and called the static validator.

I think you made the right call on how to handle this. Thanks!

packages/perseus/src/widgets/sorter/sorter.tsx
```diff
@@ -101,21 +96,8 @@ class Sorter extends React.Component<Props, State> {
...
-
-        return Sorter.validate(this.getUserInput(), rubric);
+    simpleValidate: (arg1: Rubric) => PerseusScore = (rubric) => {
```
handeyeco, line 99: No action needed because I addressed this in my other PR, but in the future if you see:

``` TS
myMethod: (genericArgName: SomeArgType) => SomeReturnType = (goodArgName) => {
  return this.whatever;
}
```

I think we could benefit from cleaning it up:

``` TS
myMethod(goodArgName: SomeArgType): SomeReturnType {
  return this.whatever;
}
```

packages/perseus/src/widgets/sorter/sorter-validator.test.ts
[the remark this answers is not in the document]
handeyeco: Yeah, I'm trying to be cautious about changing the external API of things, but in this case I think y'all made the right decision.

packages/perseus/src/widgets/sorter/sorter.tsx
handeyeco: +1, we should always be fixing types if we think we can find a simple solution.

handeyeco (approved):
Looks good, thank you

packages/perseus/src/widgets/sorter/sorter.tsx
Myranae: Thank you for pointing this out! Does that change it to no longer be an arrow function though and does that matter? I think arrow functions have an effect on code using "this". Looks like arrow functions inherit their "this" from the parent scope ([link](https://www.codementor.io/@dariogarciamoya/understanding-this-in-javascript-with-arrow-functions-gcpjwfyuc)). I'm not really sure if that's relevant here, but since it does use "this," wanted to point it out. It does look _way_ cleaner the second way.

PR #1661, opened 2024-09-24
Remove onInputError

handeyeco:
## Summary:
I didn't see that it was actually being used except in the protowebapp repo...which is obviously legacy.

Issue: LEMS-2422

## Test plan:
Nothing should functionally change


benchristel (approved):
Thank you! LGTM.

jeremywiebe (approved):
Huh, so we just don't need it at all! Cool!

PR #1664, opened 2024-09-24
Reimplement mafsStateToInteractiveGraph

benchristel:
This change was previously implemented in PR #1654, then reverted in #1663 due
to a bug. This PR reimplements the change without the bug. The original PR
summary follows.

---

Previously, we were passing the entire Mafs graph state to onChange. Since
the onChange callback data is used by the Interactive Graph widget editor to
set the correct answer for the graph, a bunch of extra data like
"hasBeenInteractedWith": true was getting saved in the correct field of
the Widget data.

This extra data is not used, so its presence in datastore is potentially
confusing to future maintainers trying to understand the data. It might
also cause bugs in the future - e.g. if some code merges it into another
object in which those properties are meaningful.

This PR ensures that we only save relevant data in the correct field of
interactive graph widgets.

Issue: none

## Test plan:

`yarn dev; open http://localhost:5173`

Play around with the graphs on the dev UI gallery page. In particular:

- Unlimited point
- Polygons with angle and side labels
- Polygons with side snapping

You should not observe any issues.

Edit, save and publish an interactive graph widget in an exercise. As a
learner, you should be able to complete the exercise successfully.

### DIFF
--- a/packages/perseus/src/widgets/interactive-graph.tsx
+++ b/packages/perseus/src/widgets/interactive-graph.tsx
@@ -1655,7 +1655,6 @@ class LegacyInteractiveGraph extends React.Component<Props, State> {
 
         $(this.angle).on("move", () => {
             this.onChange({
-                // @ts-expect-error Type '{ coords: any; type: "angle"; showAngles?: boolean | undefined; allowReflexAngles?: boolean | undefined; angleOffsetDeg?: number | undefined; snapDegrees?: number | undefined; match?: "congruent" | undefined; }' is not assignable to type 'InteractiveGraphState | undefined'.
                 graph: {...graph, coords: this.angle?.getClockwiseCoords()},
             });
         });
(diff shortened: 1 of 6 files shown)


packages/perseus/src/widgets/interactive-graph.tsx
```diff
@@ -1656,7 +1656,6 @@ class LegacyInteractiveGraph extends React.Component<Props, State> {
 
         $(this.angle).on("move", () => {
             this.onChange({
-                // @ts-expect-error Type '{ coords: any; type: "angle"; showAngles?: boolean | undefined; allowReflexAngles?: boolean | undefined; angleOffsetDeg?: number | undefined; snapDegrees?: number | undefined; match?: "congruent" | undefined; }' is not assignable to type 'InteractiveGraphState | undefined'.
```
jeremywiebe, line 1659: Well, removing this is a sign that things are moving in the right direction! :)

jeremywiebe (approved):
This looks correct to me.

PR #1665, opened 2024-09-24
Translations and polish for unlimited points

nicolecomputer:
## Summary:

- Reserves space for "Remove point" button
- Translations for strings

Issue: https://khanacademy.atlassian.net/browse/LEMS-1670

## Test plan:
- Run tests
- All is well

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
@@ -16,6 +16,8 @@ import {UnreachableCaseError} from "@khanacademy/wonder-stuff-core";
 import {Mafs} from "mafs";
 import * as React from "react";
 
+import {usePerseusI18n} from "../../components/i18n-context";
+
 import AxisArrows from "./backgrounds/axis-arrows";
 import AxisLabels from "./backgrounds/axis-labels";
 import {AxisTicks} from "./backgrounds/axis-ticks";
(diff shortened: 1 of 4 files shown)


packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -74,7 +74,8 @@ export function interactiveGraphReducer(
 ): InteractiveGraphState {
     switch (action.type) {
         case REINITIALIZE:
-            return initializeGraphState(action.params);
+            return state;
+        // return initializeGraphState(action.params);
```
benchristel, line 78: Pretty sure this change shouldn't be here, right?

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -358,13 +368,19 @@ const renderGraphControls = (props: {
     state: InteractiveGraphState;
     dispatch: (action: InteractiveGraphAction) => unknown;
     width: number;
+    strings: PerseusStrings;
```
benchristel, line 371: Instead of adding a prop for the strings, could we call `usePerseusI18n` here too? Or is that not possible / advisable? It seems like using the context might be a bit more convenient.

benchristel (commented):
There's a change to the reducer that looks like it might have been accidental. Otherwise, this looks good!

I also left a non-blocking suggestion inline about replacing the `strings` prop with `usePerseusI18n`.

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
nicolecomputer: usePerseusI18n throws a lint error if it's not used in a component directly. :(

benchristel (approved):
LGTM! 🎉

PR #1666, opened 2024-09-25
Refactoring of all angle methods to be as dry and reusable as possible

SonicScrewdriver:
## Summary:
Initial work / prototype before switching tickets

Issue: LEMS-2302

## Test plan:


packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
```diff
@@ -405,6 +405,86 @@ class InteractiveGraphEditor extends React.Component<Props> {
                         />
                     </LabeledRow>
                 )}
+                {this.props.correct?.type === "angle" && (
```
SonicScrewdriver, line 408: This adds the missing options to the Angle Graph Editor for controlling "showAngles" and "allowReflexAngles"

packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-angle.tsx
```diff
@@ -19,18 +19,19 @@ type AngleCoords = [Coord, Coord, Coord];
...
+    const {startCoords, allowReflexAngles, onChange} = props;
     return (
         <>
             {/* Current equation */}
             <View style={styles.equationSection}>
                 <LabelMedium>Starting equation:</LabelMedium>
                 <BodyMonospace style={styles.equationBody}>
-                    {getAngleEquation(startCoords)}
+                    {getAngleEquation(startCoords, allowReflexAngles)}
                 </BodyMonospace>
```
SonicScrewdriver, line 35: We need the allowReflexAngles option in order to correctly calculate the angle equation for the start coordinates.

packages/perseus/src/index.ts
```diff
@@ -125,6 +125,9 @@ export {
     getQuadraticCoords,
     getAngleCoords,
 } from "./widgets/interactive-graphs/reducer/initialize-graph-state";
+// This export is to support necessary functionality in the perseus-editor package.
+// It should be removed if widgets and editors become colocated.
+export {getClockwiseAngle} from "./widgets/interactive-graphs/math";
```
SonicScrewdriver, line 131: After reviewing kmath and some other possible locations, it seemed best to simply export this. I feel like this is a relatively safe function to export and is using the same approach as the getAngleCoords. 

I'm happy to take any feedback on this however!

packages/perseus/src/util/graphie.ts
```diff
@@ -1771,7 +1771,14 @@ const GraphUtils = {
     },
 
     // Find the angle in degrees between two or three points
-    findAngle: function (point1: Coord, point2: Coord, vertex?: Coord) {
+    // This function is deprecated as it has several issues calculating
+    // correctly when dealing with reflexive angles or the position of point2
+    // (LEMS-2202) Remove this function while removing the Legacy Interactive Graph.
+    findAngleDeprecated: function (
```
SonicScrewdriver, line 1777: I can add a comment to LEMS-2202 to mention deleting this function once we strip out the Legacy Graph. First, I need to confirm if `interactive.ts` is used anywhere else as it's the only other consumer of this function.

packages/perseus/src/widgets/interactive-graph.tsx
```diff
@@ -2351,7 +2352,8 @@ class InteractiveGraph extends React.Component<Props, State> {
             throw makeInvalidTypeError("getAngleEquationString", "angle");
         }
         const coords = InteractiveGraph.getAngleCoords(props.graph, props);
-        const angle = GraphUtils.findAngle(coords[2], coords[0], coords[1]);
+        const allowReflexAngles = props.graph.allowReflexAngles || false;
+        const angle = getClockwiseAngle(coords, allowReflexAngles);
         return (
```
SonicScrewdriver, line 2357: These options needed to be added in order to update the graph previews while editing these options in the Content Editor.

packages/perseus/src/widgets/interactive-graph.tsx
```diff
@@ -2600,19 +2602,20 @@ class InteractiveGraph extends React.Component<Props, State> {
...
-                            coords[0],
-                            // @ts-expect-error - TS2532 - Object is possibly 'undefined'.
-                            coords[1],
+                        if (!coords) {
+                            return false;
+                        }
+                        const angle = getClockwiseAngle(
+                            coords,
+                            allowReflexAngles,
                         );
```
SonicScrewdriver, line 2617: While there aren't any graphs that currently make use of the reflexive angles, this update in validation logic is necessary to make sure that we can support them properly moving forward. 

I think it is safe to update the validator at this time as the Legacy Angle Graph is no longer being shown to users.

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.test.ts
```diff
@@ -94,10 +95,13 @@ describe("shouldDrawArcOutside", () => {
         ).toBe(true);
     });
 
+    // TODO: (third) Move this test to the math package
     it("should correctly calculate the angle for the given coordinates", () => {
```
SonicScrewdriver, line 99: Whoops missed a TODO, I'll get this updated tomorrow.

packages/perseus/src/widgets/interactive-graphs/math/angles.ts
```diff
@@ -20,19 +23,55 @@ export function polar(r: number | vec.Vector2, th: number): vec.Vector2 {
...
 }
-
-export const findAngle = (
-    point1: vec.Vector2,
-    point2: vec.Vector2,
-    vertex?: vec.Vector2,
+// This function calculates the angle between two points and an optional vertex.
+// If the vertex is not provided, the angle is measured between the two points.
+// This does not account for reflex angles or clockwise position.
+export const findAngleFromVertex = (
```
SonicScrewdriver, line 29: There was only one case I could find where we were actually making use of all 3 points for findAngle, which was the polygon graph. 

In order to simplify these functions, I've opted to scope this function to only calculate the angle from a single point and the vertex. I have updated the polygon graph to use the new getClockwiseAngle method below, which is able to be shared in several other places.

packages/perseus/src/widgets/interactive-graphs/math/angles.ts
```diff
@@ -20,19 +23,55 @@ export function polar(r: number | vec.Vector2, th: number): vec.Vector2 {
...
+        return 0;
     }
-    return findAngle(point1, vertex) - findAngle(point2, vertex);
+    return (180 + (Math.atan2(-y, -x) * 180) / Math.PI + 360) % 360;
+};
+
+// This function calculates the clockwise angle between three points,
+// and is used to generate the labels and equation strings of the
+// current angle for the interactive graph.
+export const getClockwiseAngle = (
```
SonicScrewdriver, line 44: I should note that this is also used in the validator now, in order to preemptively solve issues with calculating the angles of reflexive graphs.

packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.tsx
```diff
@@ -109,7 +109,12 @@ export const StatefulMafsGraph = React.forwardRef<
     const numPoints = graph.type === "point" ? graph.numPoints : null;
     const numSides = graph.type === "polygon" ? graph.numSides : null;
     const snapTo = graph.type === "polygon" ? graph.snapTo : null;
-    const showAngles = graph.type === "polygon" ? graph.showAngles : null;
+    const showAngles =
+        graph.type === "polygon" || graph.type === "angle"
+            ? graph.showAngles
+            : null;
+    const allowReflexAngles =
+        graph.type === "angle" ? graph.allowReflexAngles : null;
```
SonicScrewdriver, line 117: These options needed to be added in order to update the graph previews while editing these options in the Content Editor.

packages/perseus/src/widgets/interactive-graphs/interactive-graph-validator.ts
```diff
@@ -256,19 +257,16 @@ function interactiveGraphValidator(
...
-                        coords[0],
-                        // @ts-expect-error - TS2532 - Object is possibly 'undefined'.
-                        coords[1],
-                    );
-                    return (angle + 360) % 360;
+                    if (!coords) {
+                        return false;
+                    }
+                    const angle = getClockwiseAngle(coords, allowReflexAngles);
+                    return angle;
```
SonicScrewdriver, line 269: While I generally tried to avoid updating the validation logic, I think this is now necessary in order to properly score reflexive/non-reflexive graphs in the future. This will provide a much more stable and consistent experience for our users. 

EXAMPLE
To provide an example of the issue being solved here: Previously, the legacy graph would **_incorrectly_** become "reflexive" if you dragged the bottom ray (point2) down, even if reflexive angles were turned off. This has been a long-standing bug with our graphs. As a result, legacy graphs could only be scored correct unidirectionally. 

In our modern re-implementation, this issue has already been fixed visually. Now, a graph's ability to become reflexive is wholly controlled by the `allowReflexAngles` property. As a result, the graph should be scored correctly regardless of the clockwise order of the points. ie. [0,1] [0,0] [1,0] and [1,0][0,0][0,1] both create a 90* angle for a graph where `allowReflexAngles=false`.

packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/util.ts
```diff
@@ -146,17 +147,13 @@ export const getQuadraticEquation = (startCoords: [Coord, Coord, Coord]) => {
...
-};
-
-export const getAngleEquation = (startCoords: [Coord, Coord, Coord]) => {
-    const [point1, vertex, point2] = startCoords;
-
-    const angle = findAngle(point2, vertex) - findAngle(point1, vertex);
+export const getAngleEquation = (
+    startCoords: [Coord, Coord, Coord],
+    allowReflexAngles?: boolean,
+) => {
```
anakaren-rojas, line 153: another nit suggestion: adding the default param vs using an `||`, you can remove the declaration of `allowReflex`
```diff
- allowReflexAngles?: boolean,
+ allowReflexAngles: boolean = false,
....
- const allowReflex = allowReflexAngles || false;
- const angle = getClockwiseAngle(startCoords, allowReflex);
+ const angle = getClockwiseAngle(startCoords, allowReflexAngles);
```

packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/util.ts
```diff
@@ -146,17 +147,13 @@ export const getQuadraticEquation = (startCoords: [Coord, Coord, Coord]) => {
...
+    const vertex = startCoords[1];
+    const allowReflex = allowReflexAngles || false;
+    const angle = getClockwiseAngle(startCoords, allowReflex);
```
anakaren-rojas, line 156: also chaining might be helpful here, although idk how much that would improve readability 
```diff
- const angle = getClockwiseAngle(startCoords, allowReflex);
- const roundedAngle = angle.toFixed(0);
+ const roundedAngle = getClockwiseAngle(startCoords, allowReflex).toFixed(0);
```

anakaren-rojas (commented):
Overall lgtm - very fun to play with, you did a great job! 
Since I'm not as familiar with this feature though, I will leave the approving the those that are 😜

packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
```diff
@@ -405,6 +405,86 @@ class InteractiveGraphEditor extends React.Component<Props> {
...
+                            <Checkbox
+                                label={
+                                    <LabelSmall>Show angle measures</LabelSmall>
+                                }
+                                checked={
+                                    // Don't show indeterminate checkbox state
+                                    !!this.props.correct?.showAngles
+                                }
+                                onChange={() => {
+                                    if (this.props.graph?.type === "angle") {
```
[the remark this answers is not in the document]
SonicScrewdriver: Ooo good callout! They definitely could be separated out, but I think this PR is probably already large and long-lived enough as it is.😅 

That sounds like a good clean up ticket for Interactive Graph's backlog however! It seems like this is the existing design pattern in this file, and we should probably update _all_ of the checkbox logic together in order to maintain consistency.

packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/util.test.ts
```diff
@@ -318,12 +318,12 @@ describe("getAngleEquation", () => {
...
-        ${[0, 5]}   | ${[0, 0]} | ${[5, 0]}                                  | ${"-90° angle at (0, 0)"}
-        ${[-5, 5]}  | ${[0, 0]} | ${[5, 0]}                                  | ${"-135° angle at (0, 0)"}
-        ${[-5, -5]} | ${[0, 0]} | ${[5, 0]}                                  | ${"-225° angle at (0, 0)"}
-        ${[5, -5]}  | ${[0, 0]} | ${[5, 0]}                                  | ${"-315° angle at (0, 0)"}
+        ${[5, 0]}   | ${[0, 0]} | ${[-5, -5]}                                | ${"135° angle at (0, 0)"}
+        ${[5, 0]}   | ${[0, 0]} | ${[5, -5]}                                 | ${"45° angle at (0, 0)"}
+        ${[0, 5]}   | ${[0, 0]} | ${[5, 0]}                                  | ${"90° angle at (0, 0)"}
+        ${[-5, 5]}  | ${[0, 0]} | ${[5, 0]}                                  | ${"135° angle at (0, 0)"}
+        ${[-5, -5]} | ${[0, 0]} | ${[5, 0]}                                  | ${"135° angle at (0, 0)"}
+        ${[5, -5]}  | ${[0, 0]} | ${[5, 0]}                                  | ${"45° angle at (0, 0)"}
```
Myranae, line 326: Question: This change is because reflexive angles are not allowed by default now?

packages/perseus/src/index.ts
Myranae: Question: What would the downsides of exporting it be?

packages/perseus/src/util/graphie.ts
Myranae: Praise: Thanks for making it clear this function should not be used in the future!

packages/perseus/src/widgets/interactive-graph.tsx
```diff
@@ -2351,7 +2352,8 @@ class InteractiveGraph extends React.Component<Props, State> {
             throw makeInvalidTypeError("getAngleEquationString", "angle");
         }
         const coords = InteractiveGraph.getAngleCoords(props.graph, props);
-        const angle = GraphUtils.findAngle(coords[2], coords[0], coords[1]);
+        const allowReflexAngles = props.graph.allowReflexAngles || false;
```
Myranae, line 2355: Question: Do you still need this after updating the default to false from Anakaren's suggestion?

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.test.ts
Myranae: Question: Do you still need to move this test? :)

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.test.ts
```diff
@@ -94,10 +95,13 @@ describe("shouldDrawArcOutside", () => {
...
         const point1 = [2, 2] as vec.Vector2;
         const point2 = [2, 0] as vec.Vector2;
         const vertex = [0, 0] as vec.Vector2;
-        expect(findAngle(point1, point2, vertex)).toBe(45);
+        const coords: [Coord, Coord, Coord] = [point1, vertex, point2];
+        const allowReflexAngles = false;
```
Myranae, line 104: Question: Does this still need to be defined after the change Anakaren suggested where the default is false? Haven't used defaults much, but I think it should just be false if the value is undefined, right? I could see it just being more clear to state it outright in a test, but just wanted to check :)

packages/perseus/src/widgets/interactive-graphs/interactive-graph-validator.ts
```diff
@@ -256,19 +257,16 @@ function interactiveGraphValidator(
         ) {
             const guess = userInput.coords;
             const correct = rubric.correct.coords;
+            const allowReflexAngles = rubric.correct.allowReflexAngles || false;
```
Myranae, line 260: Comment: Another place where you might not need to have `|| false`

packages/perseus/src/widgets/interactive-graphs/interactive-graph-validator.ts
```diff
@@ -256,19 +257,16 @@ function interactiveGraphValidator(
...
-                        // @ts-expect-error - TS2532 - Object is possibly 'undefined'.
-                        coords[2],
-                        // @ts-expect-error - TS2532 - Object is possibly 'undefined'.
-                        coords[0],
-                        // @ts-expect-error - TS2532 - Object is possibly 'undefined'.
-                        coords[1],
-                    );
-                    return (angle + 360) % 360;
+                    if (!coords) {
+                        return false;
```
Myranae, line 266: Question: Is there a validator test that covers this response? I can see there's one for answering invalid if guess.coords is missing; does that cover this?

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.test.ts
```diff
@@ -462,7 +462,10 @@ describe("movePoint on an angle graph", () => {
         invariant(updated.type === "angle");
 
         expect(
-            findAngle(updated.coords[0], updated.coords[2], updated.coords[1]),
+            getClockwiseAngle(
+                updated.coords,
+                updated.allowReflexAngles || false,
```
Myranae, line 467: Comment: Can possibly be shortened. Apologies if you can't actually shorten this and I just keep pointing out out 😅

Myranae (approved):
Truly beautiful work, Third. Thank you so much for your dedication to getting this right and to cleaning up all the loose odds and ends! Left comments about cleaning some stuff up, adding a test maybe, and just some random questions, but nothing blocking :)

packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/util.test.ts
SonicScrewdriver: They always weren't allowed by default ... technically. Previously, when you moved the bottom arm, you could accidentally generate reflexive angles regardless of the value of the `allowReflexAngles` property. With the previous validation scoring, these angles would sometimes calculate as negative angles. Generating negative angles was a major sign of this historical bug and should never have been possible. 😅

Now that we're calculating these angles correctly, we're always being provided positive values. :)

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.test.ts
SonicScrewdriver: No it doesn't, I think you're right!

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.test.ts
SonicScrewdriver: Thank you for finding these!! I only updated the content editor function, and missed doing it on the getClockwiseAngle method itself. I've updated them all now, and it's much cleaner!

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -968,12 +978,12 @@ function boundAndSnapToPolygonAngle(
     });
 
     const getAngle = function (a: number, vertex, b: number) {
-        const angle = findAngle(
-            coordsCopy[rel(a)],
-            coordsCopy[rel(b)],
-            coordsCopy[rel(vertex)],
+        const allowReflexAngles = false; // Polygons do not have reflex angles
```
benchristel, line 981: I'm not sure this is right — polygons _can_ have reflex angles if they're concave. But it's not clear to me how this `getAngle` function is being used. I don't understand what `innerAngles` (below) is for. So this code might be doing the right thing.

What are the consequences of setting `allowReflexAngles` to true vs. false here?

SonicScrewdriver: I can confirm that the polygon graph works identically to current prod, and that the reflexAngle setting does not appear to affect this either way. I've added a video to help illustrate! 

However there IS a pre-existing weird bug in the original reducer logic that might be worth a separate ticket. 🤔 It seems like we sometimes switch between inner and outer angles. @Myranae might know more about this. 

https://github.com/user-attachments/assets/b0c4bd1d-d465-47ee-a952-013cfa50a2b2

Myranae: Looks like the `angle-indicators.tsx` file might have some logic relating to the polygon angles; looks like you added this file in July, Third. I know there is logic  to prevent the shape from crossing over itself, but I don't _think_ that's relevant here.

PR #1668, opened 2024-09-25
Remove SIR's scoreWidgets

handeyeco:
## Summary:
Please double check, but I didn't see that ServerItemRenderer's `scoreWidgets` was being used.


jeremywiebe (approved):
Well huh!  That's great we can narrow the API some.

PR #1669, opened 2024-09-25
Updating protractor to default position.

catandthemachines:
## Summary:
We're updating the protractor's default position from 5% above the bottom of the graph to 25%. This is necessary because with the new angle Mafs implementation, the angle graph cannot reach the bottom of the graph, due to saving room for multiple angle options. To improve this we can simply position the graph slightly higher to make it accessible for keyboard users, while still preserving the need for students to properly position the graph and protractor.

Before: (default position of protractor.)
image

After:
image

Issue: LEMS-2372

## Test plan:

1. Go to http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph-editor--with-mafs
2. Select `Answer type: Angle` and check `Show protractor`
3. See that the protractor is higher (25% from the bottom of the graph) and you can use your keyboard to position the angle over the protractor without having to move the protractor.

### DIFF
--- a/.changeset/breezy-rockets-greet.md
+++ b/.changeset/breezy-rockets-greet.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Updating protractor's default position within a graph.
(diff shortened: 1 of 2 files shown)


benchristel (approved):
LGTM!

PR #1670, opened 2024-09-25
Matrix: split out validator

handeyeco:
## Summary:
Split out validation logic from Matrix

Issue: LEMS-2363

## Test plan:
Nothing should change, just splitting out code

### DIFF
--- a/packages/perseus/src/widgets/matrix/matrix-validator.test.ts
+++ b/packages/perseus/src/widgets/matrix/matrix-validator.test.ts
@@ -10,14 +10,6 @@ import type {
 describe("matrixValidator", () => {
     it("can be answered correctly", () => {
         // Arrange
-        const userInput: PerseusMatrixUserInput = {
-            answers: [
-                [0, 1, 2],
-                [3, 4, 5],
-                [6, 7, 8],
-            ],
-        };
-
         const rubric: PerseusMatrixRubric = {
             prefix: "",
             suffix: "",
@@ -29,6 +21,10 @@ describe("matrixValidator", () => {
             cursorPosition: [],
             matrixBoardSize: [],
             static: false,
+        };
+
+        const userInput: PerseusMatrixUserInput = {
+            answers: rubric.answers,
         };
 
         // Act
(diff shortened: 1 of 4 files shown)


packages/perseus/src/widgets/matrix/matrix-validator.test.ts
```diff
@@ -0,0 +1,184 @@
...
+            // This is so we can check that it's considered incorrect
+            // if it has the wrong length, even though it otherwise
+            // would be a partial match.
+            answers: [...correctUserInput.answers, [8, 6, 7]],
+        };
+
+        const rubric: PerseusMatrixRubric = {
+            prefix: "",
+            suffix: "",
+            answers: [
```
catandthemachines, line 158: Nit, instead of writing out the correct answer, you could just pass in `correctUserInput.answers` since that const is meant to represent the valid answer.

packages/perseus/src/widgets/matrix/matrix-validator.ts
```diff
@@ -0,0 +1,85 @@
...
+    });
+
+    if (hasEmptyCell) {
+        return {
+            type: "invalid",
+            message: strings.fillAllCells,
+        };
+    }
+
+    if (incorrectSize) {
```
catandthemachines, line 68: nit: For efficiency it would make more sense to return this error at the top of the function instead of placing at the bottom for after the computation has run. Even though I doubt our complexity will get more than `n^2`. Just food for thought 😉

catandthemachines (approved):
Looks great! Just added a few thoughts on improvement. ❤️

packages/perseus/src/widgets/matrix/matrix-validator.ts
handeyeco: That's a good point. I didn't write this code, but I'm wondering if they did this to give...

``` TS
    if (hasEmptyCell) {
        return {
            type: "invalid",
            message: strings.fillAllCells,
        };
    }
```

...a chance to go first. They needed `incorrectSize` in that loop, but they needed to run the loop to get `hasEmptyCell`. Invalid scores don't count against learners, but incorrect scores do; so they waited to return the incorrect score until after they had a chance to send the invalid score.

I'm trying to not be too opinionated about the code I'm moving though since we're having to do this for all widgets, so I'm going to leave it as-is.

PR #1673, opened 2024-09-25
[Locked Figures Labels] Make labels optional to increase type safety

nishasy:
## Summary:
Since the `labels` field was added to the interactive graph
locked figures after they were released and used by content
authors, there are already exercises saved with locked figures
that don't have a `labels` field.

If we make this a required field, this can lead to issues with
the existing exercises in the future. To increase type safety,
I'm making `labels` optional. We should also do this with
new types added in the future. (Labels haven't been released
for use yet, so we can still fix this now without issue.)

Issue: none

## Test plan:
`yarn jest`
`yarn typecheck`

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graph-locked-labels-layer.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graph-locked-labels-layer.tsx
@@ -45,14 +45,14 @@ export default function GraphLockedLabelsLayer(props: Props) {
                     {figure.type === "line" && (
                         <>
                             {/* Point 1 labels */}
-                            {(figure.points[0].labels ?? []).map((label, k) => (
+                            {figure.points[0].labels?.map((label, k) => (
                                 <LockedLabel
                                     key={`locked-figure-${i}-point-1-label-${k}`}
                                     {...label}
                                 />
                             ))}
                             {/* Point 2 labels */}
-                            {(figure.points[1].labels ?? []).map((label, k) => (
+                            {figure.points[1].labels?.map((label, k) => (
                                 <LockedLabel
                                     key={`locked-figure-${i}-point-2-label-${k}`}
                                     {...label}
--- a/packages/perseus/src/widgets/interactive-graphs/graph-locked-labels-layer.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graph-locked-labels-layer.tsx
@@ -37,7 +37,7 @@ export default function GraphLockedLabelsLayer(props: Props) {
         ) {
             return (
                 <React.Fragment key={i}>
-                    {figure.labels.map((label, j) => (
+                    {(figure.labels ?? []).map((label, j) => (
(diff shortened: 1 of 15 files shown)


packages/perseus/src/widgets/interactive-graphs/graph-locked-labels-layer.tsx
```diff
@@ -37,22 +37,22 @@ export default function GraphLockedLabelsLayer(props: Props) {
         ) {
             return (
                 <React.Fragment key={i}>
-                    {figure.labels.map((label, j) => (
+                    {(figure.labels ?? []).map((label, j) => (
```
jeremywiebe, line 40: This works, but could we just do... 
```suggestion
                    {figure.labels?.map((label, j) => (
```

packages/perseus/src/widgets/interactive-graphs/interactive-graph-question-builder.ts
```diff
@@ -324,13 +324,15 @@ class InteractiveGraphQuestionBuilder {
...
-            })),
+            labels: options?.labels
+                ? options.labels.map((label) => ({
+                      type: "label",
+                      coord: label.coord ?? vec.midpoint(point1, point2),
+                      text: label.text,
+                      color: options?.color ?? "grayH",
+                      size: label.size ?? "medium",
+                  }))
+                : undefined,
```
jeremywiebe, line 335: We can use optional chaining here again for the same result: 

```suggestion
            labels: options?.labels?.map((label) => ({
                      type: "label",
                      coord: label.coord ?? vec.midpoint(point1, point2),
                      text: label.text,
                      color: options?.color ?? "grayH",
                      size: label.size ?? "medium",
                  })),
```

packages/perseus/src/widgets/interactive-graphs/interactive-graph-question-builder.ts
```diff
@@ -362,13 +364,15 @@ class InteractiveGraphQuestionBuilder {
...
             color: options?.color ?? "grayH",
             points: [tail, tip],
-            labels: (options?.labels ?? []).map((label) => ({
-                type: "label",
-                coord: label.coord ?? vec.midpoint(tail, tip),
-                text: label.text,
-                color: options?.color ?? "grayH",
-                size: label.size ?? "medium",
-            })),
+            labels: options?.labels
```
jeremywiebe, line 367: Same note about optional chaining...

jeremywiebe (approved):
I think you can tighten up quite a few lines in the question builder using optional chaining.  Otherwise, this looks great!

packages/perseus/src/widgets/interactive-graphs/graph-locked-labels-layer.tsx
nishasy: Huh... for some reason I was convinced that you couldn't use `map` on something that could be undefined 🤔 Good to know.

PR #1674, opened 2024-09-25
[Interactive Graph Editor] Save empty full graph aria label/description as undefined

nishasy:
## Summary:
For consistency's sake, I'm updating the fullGraphAriaLabel and
fullGraphAriaDescription to save as `undefined` instead of
`""` when they're cleared. This will be consistent with their
values when they haven't been edited yet.

Issue: none

Test plan:
`yarn jest packages/perseus-editor/src/widgets/interactive-graph-editor/components/interactive-graph-description.test.tsx`

- Go to http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--mafs-with-locked-figure-labels-all-flags
- Confirm that the title and description can still be
  updated and cleared without any errors.

### DIFF
--- a/.changeset/violet-icons-breathe.md
+++ b/.changeset/violet-icons-breathe.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": patch
+---
+
+[Interactive Graph Editor] Save empty full graph aria label/description as undefined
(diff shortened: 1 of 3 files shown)


catandthemachines (approved):
Looks straight forward. Thanks for adding tests too!

PR #1675, opened 2024-09-25
Adds getImagesWithoutAltData

MrNickBreen:
## Summary:
This PR adds logic for getImagesWithoutAltData which is supporting a series of webapp changes to add AI Generated img descriptions for exercise images that lack alt text. More information about those webapp changes can be found in this PR: https://github.com/Khan/webapp/pull/25853

Previously this logic was existing in webapp code so we're doing the more "perseus approved" method of having logic interacting with perseus data live in perseus. 

Issue: https://khanacademy.atlassian.net/browse/TUT-743

## Test plan:
- confirm tests pass via yarn test


.changeset/three-buses-vanish.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/perseus": minor
```
MrNickBreen, line 2: I hope I did this right by setting this a minor version bump? Let me know if this change should be a major version bump. I thought minor version bump because this PR isn't introducing any breaking changes.

packages/perseus/src/util/extract-perseus-data.ts
```diff
@@ -577,6 +577,47 @@ function injectWidgets(
...
+        const widget = perseusRenderer.widgets[widgetID];
+
+        if (!widget.options) {
+            return;
+        }
+
+        // Add to imagesWithoutAltData if img alt is missing and
+        // image has valid img url data.
+        if (
+            widget.type === "image" &&
```
MrNickBreen, line 606: Originally Jeremy suggested I do some fancy regex evaluation here (https://github.com/Khan/webapp/pull/25853#discussion_r1775705660), but other code in this file just directly checks this `widget.type`, so I followed the existing logic I found in similar functions here.

.changeset/three-buses-vanish.md
nicolecomputer: A minor bump is perfect for this! 👍

nicolecomputer (approved):
This looks great! Thank you for the tests!

packages/perseus/src/util/extract-perseus-data.ts
```diff
@@ -577,6 +577,47 @@ function injectWidgets(
...
+            widget.options.backgroundImage.url
+        ) {
+            imgsWithoutAltData.push({
+                widgetId: widgetID,
+                imgUrl: widget.options.backgroundImage.url,
+            });
+        }
+    });
+
+    return JSON.stringify(imgsWithoutAltData);
```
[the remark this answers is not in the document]
MrNickBreen: So, the reason why I stringified this, was to make the work in webapp more straightforward & in line with other assignments to prompt variables via perseus data. We'll be directly setting this imgsWithoutAltData Perseus data to a prompt variable imgsWithoutAltText similar to how we're setting context & correctanswer (https://github.com/Khan/webapp/blob/master/services/static/javascript/content-library/ui/content-renderers/exercise/use-exercise-guide-context.ts#L243-L244). If we returned a structured object we'd have to do this stringifying in webapp and I'd rather just have a straightforward: 
```
const imgsWithoutAltData = getImagesWithoutAltData(taskProgress.itemData.question);
setPromptVariable("imgsWithoutAltData", sanitizeTex(imgsWithoutAltData));
```

packages/perseus/src/util/extract-perseus-data.test.ts
```diff
@@ -89,3 +93,52 @@ describe("injectWidgets", () => {
 
     // Add more test cases for other widget types...
 });
+
+describe("getImagesWithoutAltData", () => {
+    describe("image", () => {
+        it("should return images with an img url that lack alt text", () => {
+            // Arrange
+            const widgets = {
+                "Image 1": {
```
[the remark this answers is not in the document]
MrNickBreen: Huh. Changing this to lowercase i `image 1` fails yarn `typecheck`: 
```
packages/perseus/src/util/extract-perseus-data.test.ts:130:17 - error TS2322: Type '{ "image 1": { type: string; options: { alt: string; backgroundImage: { url: string; }; }; }; "image 2": { type: string; options: { alt: string; backgroundImage: { url: string; }; }; }; "image 3": { type: string; options: { ...; }; }; }' is not assignable to type 'PerseusWidgetsMap'.
  Type '{ "image 1": { type: string; options: { alt: string; backgroundImage: { url: string; }; }; }; "image 2": { type: string; options: { alt: string; backgroundImage: { url: string; }; }; }; "image 3": { type: string; options: { ...; }; }; }' is not assignable to type '{ [x: `image ${number}`]: ImageWidget; }'.
    Property '"image 1"' is incompatible with index signature.
      Type '{ type: string; options: { alt: string; backgroundImage: { url: string; }; }; }' is not assignable to type 'ImageWidget'.
        Types of property 'type' are incompatible.
          Type 'string' is not assignable to type '"image"'.
```
Digging in a bit to understand more deeply.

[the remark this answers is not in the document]
MrNickBreen: Other examples in codebase suggest this should be reasonable: 
https://github.com/Khan/perseus/blob/main/packages/perseus/src/__testdata__/article-renderer.testdata.ts#L8
& 
https://github.com/Khan/perseus/blob/main/packages/perseus/src/__testdata__/article-renderer.testdata.ts#L219
& 
https://github.com/Khan/perseus/blob/main/packages/perseus/src/__testdata__/graphie.testdata.ts#L20

Okay, so maybe something odd about how I'm creating perseusRenderer. 🤔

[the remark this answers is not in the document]
MrNickBreen: Github copilot helped me discover a more thorough type definition for the mock data solves the typecheck 🎉 just committed that work.

packages/perseus/src/util/extract-perseus-data.ts
```diff
@@ -577,6 +577,47 @@ function injectWidgets(
...
+ */
+function getImagesWithoutAltData(perseusRenderer: PerseusRenderer): string {
+    if (!perseusRenderer.widgets) {
+        return "";
+    }
+
+    const imgsWithoutAltData: {imgUrl: string; widgetId: string}[] = [];
+
+    keys(perseusRenderer.widgets).forEach((widgetID) => {
+        const widget = perseusRenderer.widgets[widgetID];
```
handeyeco, line 597: Not a blocker, just a thought

```suggestion
    entries(perseusRenderer.widgets).forEach(([widgetID, widget]) => {
```

packages/perseus/src/util/extract-perseus-data.ts
```diff
@@ -577,6 +577,47 @@ function injectWidgets(
...
+            return;
+        }
+
+        // Add to imagesWithoutAltData if img alt is missing and
+        // image has valid img url data.
+        if (
+            widget.type === "image" &&
+            !widget.options.alt &&
+            widget.options.backgroundImage &&
+            widget.options.backgroundImage.url
```
handeyeco, line 609: Same, not a blocker either way
```suggestion
            widget.options.backgroundImage?.url
```

packages/perseus/src/util/extract-perseus-data.ts
```diff
@@ -577,6 +577,47 @@ function injectWidgets(
...
+        ) {
+            imgsWithoutAltData.push({
+                widgetId: widgetID,
+                imgUrl: widget.options.backgroundImage.url,
+            });
```
handeyeco, line 614: If you called it `widgetId` above.
```suggestion
            imgsWithoutAltData.push({
                widgetId,
                imgUrl: widget.options.backgroundImage.url,
            });
```

packages/perseus/src/util/extract-perseus-data.test.ts
```diff
@@ -89,3 +94,55 @@ describe("injectWidgets", () => {
 
     // Add more test cases for other widget types...
 });
+
+describe("getImagesWithoutAltData", () => {
+    describe("image", () => {
```
handeyeco, line 99: Is the idea that in the future we might have more `describe` blocks for different widgets? If so, I might rename this to `image widget`. Otherwise I wonder if we could just drop this layer.

packages/perseus/src/util/extract-perseus-data.test.ts
```diff
@@ -89,3 +94,55 @@ describe("injectWidgets", () => {
...
+        it("should return images with an img url that lack alt text", () => {
+            // Arrange
+            const widgets: {[key: `image ${number}`]: ImageWidget} = {
```
handeyeco, line 102: Would this work?

```suggestion
            const widgets: PerseusWidgetsMap = {
```

handeyeco (approved):
Just some thoughts, no blockers. Thanks!

packages/perseus/src/util/extract-perseus-data.ts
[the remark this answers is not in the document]
jeremywiebe: My one concern is that if someone else needs this somewhere else and start doing `JSON.parse()` of this return value. 

Let's leave it as it is, and we can refactor it if need be.

PR #1676, opened 2024-09-26
Delete unused component param of InteractiveGraph.validate()

benchristel:
This parameter appeared to be used to score point graphs. It was passed to
`getPointCoords()`. However, on closer inspection, there was no way this code
could have worked: `getPointCoords` takes the interactive graph props as an
argument, but we were passing it the entire component. This props param was
only being used if the rubric contained no coords. Maybe that's a realistic
case, or maybe it's not, but either way, it was throwing an error trying to
access `props.range[0]`.

Long story short, we lose nothing by deleting the `component` param and
explicitly throwing an error if the rubric has no coords.

Issue: LEMS-2448

## Test plan:

Run this script to copy point graph data to your clipboard:

```
grep -rl '"type":"point"' data/questions/ | xargs cat | pbcopy
```

Paste it into `http://localhost:5173#flipbook`. Solve some point graph
questions.

### DIFF
--- a/packages/perseus/src/widgets/interactive-graph.test.tsx
+++ b/packages/perseus/src/widgets/interactive-graph.test.tsx
@@ -256,12 +256,7 @@ describe("InteractiveGraph.validate on a point question", () => {
 
         const result = InteractiveGraph.widget.validate(guess, rubric);
 
-        expect(result).toEqual({
-            type: "points",
-            earned: 1,
-            total: 1,
-            message: null,
-        });
+        expect(result).toHaveBeenAnsweredCorrectly();
     });
 
     it("does not modify the `guess` data", () => {
--- a/packages/perseus/src/widgets/interactive-graph.test.tsx
+++ b/packages/perseus/src/widgets/interactive-graph.test.tsx
@@ -215,12 +215,7 @@ describe("InteractiveGraph.validate on a point question", () => {
 
         const result = InteractiveGraph.widget.validate(guess, rubric);
 
-        expect(result).toEqual({
-            type: "points",
-            earned: 0,
-            total: 1,
-            message: null,
-        });
+        expect(result).toHaveBeenAnsweredIncorrectly();
     });
 
     it("awards points if guess.coords is right", () => {
--- a/packages/perseus/src/widgets/interactive-graph.test.tsx
+++ b/packages/perseus/src/widgets/interactive-graph.test.tsx
@@ -184,10 +184,7 @@ describe("InteractiveGraph.validate on a point question", () => {
 
         const result = InteractiveGraph.widget.validate(guess, rubric);
 
-        expect(result).toEqual({
-            type: "invalid",
-            message: null,
-        });
+        expect(result).toHaveInvalidInput();
     });
 
     it("throws an exception if correct.coords is missing", () => {
--- a/packages/perseus/src/widgets/interactive-graph.test.tsx
+++ b/packages/perseus/src/widgets/interactive-graph.test.tsx
@@ -1,5 +1,7 @@
 import invariant from "tiny-invariant";
 
+import {clone} from "../../../../testing/object-utils";
+
 import InteractiveGraph, {shouldUseMafs} from "./interactive-graph";
 
 import type {
@@ -28,7 +30,7 @@ describe("InteractiveGraph.validate on a segment question", () => {
             ],
         });
 
-        const result = InteractiveGraph.widget.validate(guess, rubric, null);
+        const result = InteractiveGraph.widget.validate(guess, rubric);
 
         expect(result).toHaveInvalidInput();
     });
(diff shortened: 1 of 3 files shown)


packages/perseus/src/widgets/interactive-graph.test.tsx
```diff
@@ -192,6 +194,151 @@ describe("InteractiveGraph.validate on a segment question", () => {
...
+            type: "point",
+            coords: [[0, 0]],
+        });
+
+        const result = InteractiveGraph.widget.validate(guess, rubric);
+
+        expect(result).toEqual({
+            type: "invalid",
+            message: null,
+        });
```
handeyeco, line 210:
```suggestion
        expect(result).toHaveInvalidInput();
```

packages/perseus/src/widgets/interactive-graph.test.tsx
```diff
@@ -192,6 +194,151 @@ describe("InteractiveGraph.validate on a segment question", () => {
...
+
+        const result = InteractiveGraph.widget.validate(guess, rubric);
+
+        expect(result).toEqual({
+            type: "points",
+            earned: 0,
+            total: 1,
+            message: null,
+        });
```
handeyeco, line 246:
```suggestion
        expect(result).toHaveBeenAnsweredIncorrectly();
```

packages/perseus/src/widgets/interactive-graph.test.tsx
```diff
@@ -192,6 +194,151 @@ describe("InteractiveGraph.validate on a segment question", () => {
...
+
+        const result = InteractiveGraph.widget.validate(guess, rubric);
+
+        expect(result).toEqual({
+            type: "points",
+            earned: 1,
+            total: 1,
+            message: null,
+        });
```
handeyeco, line 266:
```suggestion
        expect(result).toHaveBeenAnsweredCorrectly();
```

packages/perseus/src/widgets/interactive-graph.test.tsx
```diff
@@ -192,6 +194,151 @@ describe("InteractiveGraph.validate on a segment question", () => {
...
```
handeyeco, line 292:
```suggestion
        expect(result).toHaveBeenAnsweredCorrectly();
```

packages/perseus/src/widgets/interactive-graph.tsx
```diff
@@ -2504,10 +2500,10 @@ class InteractiveGraph extends React.Component<Props, State> {
...
                 userInput.coords != null
             ) {
-                let correct = InteractiveGraph.getPointCoords(
-                    rubric.correct,
-                    component,
-                );
+                let correct = rubric.correct.coords;
+                if (correct == null) {
+                    throw new Error("Point graph rubric has null coords");
+                }
```
handeyeco, line 2506: Maybe this is fine, but I want to ask @jeremywiebe. This validation logic is moving to the back-end and I'm not sure we want to be throwing or...[[insert idea here]]. Because if this throws on the BE, the client gets a 500 unless we handle it cleverly. I wonder if it would be better to return "invalid" (or report the error and return a correct answer since this would be KA's fault?)

handeyeco (approved):
Looks good, just some ideas for using custom matchers and a question for you and @jeremywiebe 

On a side note, love the way you work. I asked you a question and ended up with dead code removal and a bunch of tests. You're the best!

packages/perseus/src/widgets/interactive-graph.tsx
handeyeco: I see Jeremy's out today. I don't think this needs to block the PR, we can circle back to it.

benchristel: It's possible that the work I'm going to do pursuant to [this slack thread on WidgetOptions validation](https://khanacademy.slack.com/archives/C067UM1QAR4/p1727307152791499) will solve this. Our parser can default `coords` to `[]` if it's not present. It should also log an error if it has to set any defaults, so we can at least detect the problem.

Or maybe it would be better to default `correct` to a special value like `"ungraded"`, so the learner isn't penalized? That would require changing this logic.

handeyeco: I guess my feeling is: do the bare minimum to remove `component` and if we think we need to circle back we can file a ticket.

jeremywiebe: I think we'll need to handle invalid data coming from the frontend no regardless of this situation so I don't think this too bad throwing here. We'll also want to wrap whatever our final validation/scoring logic in error handling to avoid the server just crashing when an error is thrown. 

To the specific point, it looks like many of the other graph types access the `rubric.correct.coords` optimistically so this `point` type now does something extract. I'm not sure it's worth it unless we apply the same check/logic to all other types that access `rubric.correct.coords`.

PR #1677, opened 2024-09-26
[Locked Figure Aria] Implement locked point aria label behavior on graph

nishasy:
## Summary:
- Add `locked-figures-aria` flag
- Add `ariaLabel` field to LockedPointType
- Add the aria label behavior to the LockedPoint on the mafs graph
- Minor updates to the screen reader behvaior

Issue: https://khanacademy.atlassian.net/browse/LEMS-2375

## Test plan:

Storybook
- Use a screen reader to navigate through the preview graph
- Using tab should _only_ navigate to the interactive elements, NOT the locked point
- Using the screenreader controls (VO + arrow keys) should be able to navigate to
  the locked point. It should read out as "Point A".

https://github.com/user-attachments/assets/939ae929-8569-410b-a25b-8b09e16ec97b

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/backgrounds/axis-ticks.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/backgrounds/axis-ticks.tsx
@@ -48,7 +48,7 @@ const YGridTick = ({y, range}: {y: number; range: [Interval, Interval]}) => {
     const showLabel = shouldShowLabel(y, range);
 
     return (
-        <g className="tick">
+        <g className="tick" aria-hidden={true}>
             <line x1={x1} y1={y1} x2={x2} y2={y2} className="axis-tick" />
             {showLabel && (
                 <text
@@ -106,7 +106,7 @@ const XGridTick = ({x, range}: {x: number; range: [Interval, Interval]}) => {
     const yPositionText = yPosition + yAdjustment;
 
     return (
-        <g className="tick">
+        <g className="tick" aria-hidden={true}>
             <line x1={x1} y1={y1} x2={x2} y2={y2} className="axis-tick" />
             {
                 <text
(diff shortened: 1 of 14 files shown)


packages/perseus/src/widgets/interactive-graphs/backgrounds/axis-ticks.tsx
```diff
@@ -48,7 +48,7 @@ const YGridTick = ({y, range}: {y: number; range: [Interval, Interval]}) => {
     const showLabel = shouldShowLabel(y, range);
 
     return (
-        <g className="tick">
+        <g className="tick" aria-hidden={true}>
```
catandthemachines, line 51: Why hide this content from screen readers? 🤔

packages/perseus/src/widgets/interactive-graphs/locked-figures/locked-point.tsx
```diff
@@ -4,21 +4,38 @@ import * as React from "react";
...
-                    fill: filled ? lockedFigureColors[color] : wbColor.white,
-                    stroke: lockedFigureColors[color],
-                    strokeWidth: spacing.xxxxSmall_2,
-                },
-            }}
-        />
+        <g
+            className="locked-point"
+            aria-label={hasAria ? ariaLabel : undefined}
+            aria-hidden={!hasAria}
```
catandthemachines, line 23: How does the experience sound when it's hidden vs. not hidden? Just want to make sure we're not hiding content from screen readers when it doesn't need to be.

catandthemachines (commented):
Code looks good. Just have a lot of questions around the usage of aria-hidden and it's purpose in the implementation.

packages/perseus/src/widgets/interactive-graphs/locked-figures/locked-point.tsx
nishasy: If it's not hidden, it just says "image" or "empty group" 😬

packages/perseus/src/widgets/interactive-graphs/backgrounds/axis-ticks.tsx
nishasy: It makes you go through all the axis tick labels, so it sounds like "1, 2, 3, 4, 5, 6, 7, 8, 9, 10, -1, -2, -3, -4, -5..." with no context for why it's just randomly saying numbers.

catandthemachines (approved):

PR #1678, opened 2024-09-26
Port tests to custom matcher

handeyeco:
## Summary:
We made a custom matcher to make score tests easier to grok. This ports some existing tests over to that.


Myranae (approved):
Looks good!

PR #1681, opened 2024-09-26
DO NOT REVIEW: ZND build with unlimited point always on

nicolecomputer:
## Summary:
DO NOT REVIEW

## Test plan:
- Build a ZND


jeremywiebe:
FWIW, we recently changed the snapshot build process to also build for drafts. So you could mark this as a draft if that makes things more obvious to prevent folks reviewing. :)

PR #1682, opened 2024-09-26
[Locked Figure Aria] Locked point aria label editor UI

nishasy:
## Summary:
- Create a new `locked-figure-aria.tsx` file that can
  be reused for each locked figure's settings.
- Use this `locked-figure-aria.tsx` component within
  LockedPointSettings.
- Write the function to auto-generate the locked
  point aria label with its coordinates and
  visible labels.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2375

## Test plan:
- `yarn jest packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-point-settings.test.tsx`
- `yarn jest packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-figure-aria.test.tsx`

Storybook
- Confirm that the locked point aria label field is already populated with "Point A" (passed in via builder)
- Confirm that updateing the aria label in the editor also updates the aria label on the point

image


catandthemachines:
General question: Do we want to auto generate a label that includes the coordinates? And why is it part of the autogenerated labels, and not the default one? 🤔

image

catandthemachines (approved):
The code looks good, but just had a question about the intention behind the autogenerated aria-labels.

nishasy:
@catandthemachines 
> General question: Do we want to auto generate a label that includes the coordinates? And why is it part of the autogenerated labels, and not the default one? 🤔

Since this auto-generation is just to aid the content authors (and not to actually have those labels on the figures automatically), I just wanted to add all the info that would be relevant. The content authors can edit this field as necessary. Some questions may require hiding that information, and some may not.

Also @catandthemachines What do you mean by "default labels"? The default label should be empty string (undefined when saved).

PR #1683, opened 2024-09-26
[Locked Figure Aria] Implement locked line aria label behavior on graph

nishasy:
## Summary:
- Add `ariaLabel` field to LockedLineType
- Add the aria label behavior to the LockedLine on the mafs graph

Issue: https://khanacademy.atlassian.net/browse/LEMS-2376

## Test plan:
`yarn jest packages/perseus/src/widgets/interactive-graphs/interactive-graph.test.tsx`

Storybook
- Go to http://localhost:6006/iframe.html?args=&id=perseuseditor-widgets-interactive-graph--mafs-with-locked-figure-labels-all-flags&viewMode=story
- Use a screen reader to navigate through the preview graph
- Using the screenreader controls (VO + arrow keys) should be able to navigate to
  the locked line. It should read out as "Point PQ".

https://github.com/user-attachments/assets/a6fafd5b-9d3e-4b1a-94ca-7404bb60e7ab

PR #1684, opened 2024-09-26
[Locked Figure Aria] Locked line aria label editor UI

nishasy:
## Summary:
- Use `locked-figure-aria.tsx` within LockedLineSettings.
- Write the function to auto-generate the locked line
  aria label with its coordinates and visible labels.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2376

## Test plan:
`yarn jest packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-line-settings.test.tsx`

Storybook
- Confirm that the locked line aria label field is already populated with "Line PQ" (passed in via builder)
- Confirm that updating the aria label in the editor also updates the aria label on the line

image

https://github.com/user-attachments/assets/36738b4f-6524-4034-adbb-52f6cc53cb94


packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-line-settings.tsx
```diff
@@ -69,6 +71,29 @@ const LockedLineSettings = (props: Props) => {
     // Check if the line has length 0.
     const isInvalid = kvector.equal(point1.coord, point2.coord);
 
+    function getPrepopulatedAriaLabel() {
+        let str = `${capitalizeKind} from (${point1.coord[0]}, ${point1.coord[1]}) to (${point2.coord[0]}, ${point2.coord[1]})`;
+
+        if (labels && labels.length > 0) {
```
benchristel, line 77:
```suggestion
        if (labels?.length > 0) {
```

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-line-settings.tsx
```diff
@@ -69,6 +71,29 @@ const LockedLineSettings = (props: Props) => {
...
+                str += "s";
+            }
+
+            for (let i = 0; i < labels.length; i++) {
+                // Separate additional labels with commas.
+                if (i > 0) {
+                    str += ",";
+                }
+                str += ` ${labels[i].text}`;
+            }
```
benchristel, line 91: Could this be simplified to

```suggestion
            str += ` ${labels.map((l) => l.text).join(", ")}`
```

or am I missing something?

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-line-settings.tsx
```diff
@@ -69,6 +71,29 @@ const LockedLineSettings = (props: Props) => {
     // Check if the line has length 0.
     const isInvalid = kvector.equal(point1.coord, point2.coord);
 
+    function getPrepopulatedAriaLabel() {
+        let str = `${capitalizeKind} from (${point1.coord[0]}, ${point1.coord[1]}) to (${point2.coord[0]}, ${point2.coord[1]})`;
```
benchristel, line 75: I imagine this formatting style will be reused in the aria labels for many different locked figure types, so... what do you think about extracting a function like

```ts
function formatLockedPoint(point: LockedPointType): string {
    const [x, y] = point.coord;
    return `(${x}, ${y})`;
}
```

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-line-settings.tsx
```diff
@@ -69,6 +71,29 @@ const LockedLineSettings = (props: Props) => {
...
+                // Separate additional labels with commas.
+                if (i > 0) {
+                    str += ",";
+                }
+                str += ` ${labels[i].text}`;
+            }
+        }
+
+        return str;
+    }
```
benchristel, line 95: I notice that the points of a line can also have their own labels (at least, according to the `LockedPointType`). Do we want to include those as well?

benchristel (approved):
I made a few non-blocking suggestions inline. LGTM overall!

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-line-settings.tsx
nishasy: I thought about that, but I think I'm going to forego it because the more I thought about it, the more it sounded like something the content authors should manually write the aria label for. Especially since each point can hypothetically have multiple labels.

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-line-settings.tsx
nishasy: I think the LockedPointType is only used in locked points and locked lines, so I personally don't think there's enough  usage of that yet to extract the coordinates out?

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-line-settings.tsx
nishasy: That's much cleaner, thank you!

nishasy:
The parent pull-request ([#1683](https://github.com/Khan/perseus/pull/1683)) has been merged into `main`, but this branch (`locked-line-aria-editor`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

PR #1687, opened 2024-09-26
[Locked Figure Aria] Locked function aria labels (graph + editor)

nishasy:
## Summary:

- Add ariaLabel field to LockedFunctionType
- Add the aria label behavior to the LockedFunction on the mafs graph
- Use LockedFigureAria within LockedFunctionSettings
- Write the function to auto-generate the locked function aria label
- with its coordinates and visible labels.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2380

## Test plan:
`yarn jest packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-function-settings.test.tsx`

### Storybook

Graph behavior

- Use a screen reader to navigate through the preview graph
- Using the screenreader controls (VO + arrow keys) should be able to navigate to the locked function. It should read out as "Function E".

Editor

- Confirm that the locked function aria label field is already populated with "Function  E" (passed in via builder)
- Confirm that updating the aria label in the editor also updates the aria label on the function

## Screenshots & Videos

Graph experience

https://github.com/user-attachments/assets/206ca154-d8f8-40f8-8502-be73449170af

Editor

Screenshot 2024-09-26 at 4 52 43 PM

https://github.com/user-attachments/assets/4d773464-4b5d-409d-8c1b-d185035de54c


packages/perseus/src/widgets/interactive-graphs/locked-figures/locked-function.tsx
```diff
@@ -34,7 +42,11 @@ const LockedFunction = (props: LockedFunctionType) => {
     }
 
     return (
-        <g className="locked-function">
+        <g
+            className="locked-function"
+            aria-label={hasAria ? props.ariaLabel : undefined}
+            aria-hidden={!hasAria}
```
benchristel, line 48: Why is `aria-hidden` needed here? Does the screenreader read something for this SVG element if there's no `aria-label`?

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-function-settings.test.tsx
```diff
@@ -652,5 +652,242 @@ describe("Locked Function Settings", () => {
...
+                        {...defaultProps}
+                        ariaLabel="Ellipse at (x, y)"
+                    />,
+                    {wrapper: RenderStateRoot},
+                );
+
+                const input = screen.getByRole("textbox", {name: "Aria label"});
+
+                // Assert
+                expect(input).toHaveValue("Ellipse at (x, y)");
```
benchristel, line 672: Maybe this label shouldn't contain `Ellipse` - just to avoid confusion :)

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-function-settings.tsx
```diff
@@ -80,6 +82,30 @@ const LockedFunctionSettings = (props: Props) => {
         ]);
     }, [domain]);
 
+    function getPrepopulatedAriaLabel() {
+        let str = `Function with equation ${equationPrefix}${equation}`;
+
+        // Add the domain/range constraints to the aria label
+        // if they are not the default values.
+        if (domain && !(domain[0] === -Infinity && domain[1] === Infinity)) {
+            str += `, ${domainRangeText} from ${domain[0]} to ${domain[1]}`;
```
benchristel, line 91: I think, rather than using `${domainRangeText}`, we always want this word to be `domain`. IIUC, the domain of a function is always the set of input values, so if the function has the form `x = f(y)`, the domain is a set of y coordinates.

Might be good to confirm this with math content creators, though.

packages/perseus/src/widgets/interactive-graphs/locked-figures/locked-function.tsx
nishasy: Yes! It says "empty group"

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-function-settings.tsx
nishasy: That is a good point! I'll ask

nishasy: I'll change it to `domain` here and in the editor UI. If a change is requested in the future, we can updated it. For now, I'd like to unblock this so I can keep landing all the PRs stacked on top of this.

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-function-settings.tsx
```diff
@@ -223,7 +248,7 @@ const LockedFunctionSettings = (props: Props) => {
                     tag="label"
                     style={[styles.dropdownLabel, styles.domainMin]}
```
anakaren-rojas, line 249: Does the label component for domain min also need an aria label like the domain max one?

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-function-settings.tsx
```diff
@@ -80,6 +81,30 @@ const LockedFunctionSettings = (props: Props) => {
         ]);
     }, [domain]);
 
+    function getPrepopulatedAriaLabel() {
+        let str = `Function with equation ${equationPrefix}${equation}`;
+
```
anakaren-rojas, line 86: Should there be an additional check to validate that there is text added to the labels? Or make the text a required field if the button to add a visible label is clicked?

anakaren-rojas:
Played with the "auto generate aria label" a little too long , its way cool

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-function-settings.tsx
nishasy: No, it's automatically labeled by the label itself. You can see in the test file that the aria labels successfully get set as `"domain min"` and `"domain max"`.

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-function-settings.tsx
nishasy: I did think about that, and that is a good point. I do agree with you. Unfortunately, validation in the content editor has been deprioritized in favor of just trusting the content authors and giving them ultimate freedom. The content authors should definitely be checking that the label makes sense before saving this on the exercise.

If this ends up being an issue and content authors ask for extra validation, we would definitely add that in the future.

benchristel (approved):
LGTM!

PR #1688, opened 2024-09-27
Split out validation logic for cs-program

Myranae:
## Summary:
For the server-side scoring project, we are moving validation logic out of each widget component. This moves the validation logic for the cs-program widget and adds tests for the validator function.

Issue: LEMS-2353

## Test plan:
- Confirm all tests pass
- Nothing should change. Code is only being moved.


packages/perseus/src/widgets/cs-program/cs-program-validator.ts
```diff
@@ -0,0 +1,29 @@
+import type {PerseusScore} from "../../types";
+import type {PerseusCSProgramUserInput} from "../../validation.types";
+
+export function csProgramValidator(
+    state: PerseusCSProgramUserInput,
+): PerseusScore {
```
Myranae, line 6: `rubric` was not being used, so it was removed as a parameter.

packages/perseus/src/widgets/cs-program/cs-program-validator.test.ts
```diff
@@ -0,0 +1,54 @@
...
+        // Act
+        const result = csProgramValidator(state);
+
+        // Assert
+        expect(result).toEqual({
+            type: "points",
+            earned: 0,
+            total: 1,
+            message: "Try again!",
+        });
```
handeyeco, line 36: If it's important that we check the message (and I'm not sure it is), I think we should update `toHaveBeenAnsweredIncorrectly` to optionally handle a message like `toHaveInvalidInput`. Otherwise I think we should switch this to `toHaveBeenAnsweredIncorrectly`.

packages/perseus/src/widgets/cs-program/cs-program-validator.ts
handeyeco: This is one of those sus things we're going to have to figure out in the future. Can you check to make sure there's a ticket for this? Something along the lines on "some widgets don't use rubric for validation"? I imagine all the program/iframe widgets do this.

handeyeco (approved):
Thanks!

packages/perseus/src/widgets/cs-program/cs-program-validator.test.ts
Myranae: 🤦‍♀️ Yep, totally meant to do that. Not sure how I skipped this. Thanks for catching it!

packages/perseus/src/widgets/cs-program/cs-program-validator.ts
Myranae: Here's the relevant ticket: https://khanacademy.atlassian.net/browse/LEMS-2440

packages/perseus/src/widgets/cs-program/cs-program.tsx
```diff
@@ -130,14 +108,11 @@ class CSProgram extends React.Component<Props> {
...
-            {
-                status: this.props.status,
-                message: this.props.message,
-            },
-            rubric,
-        );
+    simpleValidate(): PerseusScore {
+        return csProgramValidator({
+            status: this.props.status,
+            message: this.props.message,
```
jeremywiebe, line 114: The data we pass to the `csProgramValidator` should be extracted into a [`getUserInput` function](https://github.com/Khan/perseus/blob/f5af2437133d68dcbaa42830850341c46d7affee/packages/perseus/src/renderer.tsx#L148). This will help us as we split the scoring logic further out of the renderer system. 

You can see this pattern in play in many of our widgets (eg [`grapher`](https://github.com/Khan/perseus/blob/f326139ee6bf9680075eac6353b1e84c44fa3f77/packages/perseus/src/widgets/grapher/grapher.tsx#L544) or [`number-line`](https://github.com/Khan/perseus/blob/ae51ccdb820894f6fc5c1b23556823efdd4edba6/packages/perseus/src/widgets/number-line/number-line.tsx#L638))

Could you do this in a new PR?

Myranae: Here's the ticket for that: https://khanacademy.atlassian.net/browse/LEMS-2459

PR #1689, opened 2024-09-27
[Locked Figure Labels] View locked polygon labels

nishasy:
## Summary:
- Add feature flag for locked polygon labels
- Add labels to LockedPolygonType
- Update builder
- Update stories and tests

Issue: https://khanacademy.atlassian.net/browse/LEMS-2350

## Test plan:
- Go to http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--mafs-with-locked-polygon-labels-flag
- Confirm that the visible label is on the locked polygon only
- Go to the other locked figure stories and confirm that
  the locked polygon does not have a visible label (make sure
  that the flags are separating behaviors correctly)
  
image


benchristel (approved):
LGTM!

PR #1690, opened 2024-09-27
[Locked Figure Labels] Add/edit/delete locked polygon labels

nishasy:
## Summary:
Add the ability to add/edit/delete labels for locked polygons
within the interactive graph editor.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2350

## Test plan:
- Go to http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--mafs-with-locked-polygon-labels-flag
- Go to locked figures
- Confirm that the label settings are there
- Play around with the label settings and confirm that
  the label moves and updates with the polygon

image


packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-polygon-settings.tsx
```diff
@@ -55,34 +60,87 @@ const LockedPolygonSettings = (props: Props) => {
...
                 });
                 break;
         }
     }
 
+    function handleLabelChange(
+        updatedLabel: LockedLabelType,
+        labelIndex: number,
+    ) {
+        if (!labels) {
```
catandthemachines, line 121: nit: I'm personally against empty returns when simply wrapping the logic in an if statement would work just as effectively. Highly recommend the following pattern instead.
`
if (label) {
....
}
`

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-polygon-settings.tsx
```diff
@@ -55,34 +60,87 @@ const LockedPolygonSettings = (props: Props) => {
...
+        updatedLabels[labelIndex] = {
+            ...labels[labelIndex],
+            ...updatedLabel,
+        };
+
+        onChangeProps({labels: updatedLabels});
+    }
+
+    function handleLabelRemove(labelIndex: number) {
+        if (!labels) {
```
catandthemachines, line 135: nit: same here.

catandthemachines (approved):
Looks good, just a gentle `nit` on some if statements. 😉

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-polygon-settings.tsx
nishasy: Interesting! This might be an "agree to disagree" situation because I'm pretty solidly in opposition. I prefer returning early, because I find that makes the rest of the function easier to read. It generally reduces indentation, and I feel it makes it clear that we don't have to worry about the null case anymore past that point. I think the "return early" pattern is pretty widely used/taught in my (admittedly limited) experience.

catandthemachines: It's so funny because while the "return early" pattern was widely taught in my experience. The "empty return" was highly discouraged. Let's agree to disagree here 😉

PR #1694, opened 2024-09-27
[Bug Fix] Update shouldShowLabel logic to better handle decimal tickSteps 

SonicScrewdriver:
## Summary:
Decimal labels for the Axis Ticks in the Interactive Graph Widget are experiencing a bug in relation to the shouldShowLabel logic, where the -1 is being hidden when it should not be. Additionally, decimal tick steps would often result in overlapping labels. 

This PR updates the shouldShowLabel logic to always hide the negative tickStep on the y-axis if it is within the graph bounds. This should solve both of the issues above, and result in a cleaner graph experience. 

Issue: LEMS-2406

## Test plan:
- yarn jest
- Updated test to test the edge case

### DIFF
--- a/.changeset/tall-scissors-pay.md
+++ b/.changeset/tall-scissors-pay.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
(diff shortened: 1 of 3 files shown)


mark-fitzgerald (approved):
Simple update - Love it!!

PR #1695, opened 2024-09-27
BUGFIX - Leftmost number in Number Line widget misaligned

mark-fitzgerald:
## Summary:
The Number Line widget contains display logic that adds duplicate endpoint numbers, and then colors them blue if the widget is being used on mobile. This duplicate endpoint becomes misaligned when styling is accounting for the reduced viewport size. This bugfix corrects that misalignment, and removes the mobile logic in favor of standardizing across all platforms (both mobile and desktop now have blue endpoints).

Issue: LEMS-2383

## Test plan:

1. Launch Storybook
1. Review the [Number Line](http://localhost:6006/?path=/story/perseus-widgets-number-line--question-1) widget
   - Both endpoints should be blue
   - All numbers on the line should be positioned the same vertically

### Before
Number Line Endpoint - Before

### After
Number Line Endpoint - After


packages/perseus/src/widgets/number-line/number-line.tsx
```diff
@@ -160,34 +160,18 @@ const TickMarks: any = Graphie.createSimpleClass((graphie, props) => {
 
         const labelTicks = props.labelTicks;
         if (labelTicks || props.labelStyle === "decimal ticks") {
-            results.push(_label(graphie, props.labelStyle, x, x, base));
+            if (x === leftLabel || x === rightLabel) {
+                results.push(
+                    graphie.style({color: KhanColors.BLUE}, () =>
```
catandthemachines, line 165: Just confirming, this color has the appropriate color contrast right? The blue color looks a little light to me.

catandthemachines (approved):
The code looks good. Just want to check on the color for the numbers and you should be good to go!

packages/perseus/src/widgets/number-line/number-line.tsx
catandthemachines: Okay I checked the color out and it does pass AA so we should be good, I would love for it to be darker for the type of text it it. But non-blocking 😉
image

mark-fitzgerald: Agreed! That color was the first thing I checked, too.

PR #1696, opened 2024-09-28
Change Widget type to an interface

jeremywiebe:
## Summary:

All Perseus widgets implicitly conform to the `Widget` type defined in `renderer.tsx`. To date, this has been managed somewhat manually by visually checking if the parameter and return types are correct. 

Today I decided to poke at the idea of making the type an interface so all of our widget components could declare that they actually do conform to this contract. 

It worked! 

This is a large-ish PR, but most of it is applying the `implements Widget` designation to all widgets and then removing incorrect/redundant parameter/return types. 

Issue: "none" (except my OCD)

## Test plan:

`yarn tsc` ✅
`yarn test` ✅


packages/perseus/src/types.ts
```diff
@@ -24,11 +25,66 @@ export type Dimensions = {
...
+/**
+ * The Widget type represents the common API that the Renderer uses to interact
+ * with all widgets. All widgets must implement the methods in this API, unless
+ * they are marked as optional (?: ...).
+ *
+ * These methods are called on the widget ref and allow the renderer to
+ * communicate with the individual widgets to coordinate actions such as
+ * scoring, state serialization/deserialization, and focus management.
+ */
+export interface Widget {
```
jeremywiebe, line 41: I moved this type out of `renderer.tsx` into this `types.ts` file as it felt more natural here.

packages/perseus/src/perseus-types.ts
```diff
@@ -174,7 +174,7 @@ export const ItemExtras = [
 ] as const;
 export type PerseusAnswerArea = Record<(typeof ItemExtras)[number], boolean>;
 
-type Widget<Type extends string, Options> = {
```
jeremywiebe, line 177: This was really confusing. We had two types named `Widget` in our codebase: one represented the React component and the other the widget options. Renaming this will hopefully make things clearer.

packages/perseus/src/renderer.tsx
```diff
@@ -104,60 +106,6 @@ type SetWidgetPropsFn = (
     silent?: boolean,
 ) => void;
 
-type SerializedState = {
```
jeremywiebe, line 107: This whole block was moved to `types.ts`. The `Widget` type also becomes an `interface` in that file.

packages/perseus/src/renderer.tsx
```diff
@@ -1544,17 +1492,16 @@ class Renderer extends React.Component<Props, State> {
...
+            if (widget?.getDOMNodeForPath) {
+                return widget.getDOMNodeForPath(interWidgetPath);
             }
             if (interWidgetPath.length === 0) {
                 // @ts-expect-error - TS2345 - Argument of type 'Widget | null | undefined' is not assignable to parameter of type 'ReactInstance | null | undefined'.
                 return ReactDOM.findDOMNode(widget);
             }
         };
 
-    getGrammarTypeForPath: (path: FocusPath) => string | null | undefined = (
```
jeremywiebe, line 1557: We never return `null`!

packages/perseus/src/renderer.tsx
```diff
@@ -1544,17 +1492,16 @@ class Renderer extends React.Component<Props, State> {
             // beyond the widgetID, as a special case we just return the widget's
             // DOM node.
             const widget = this.getWidgetInstance(widgetId);
-            const getNode = widget && widget.getDOMNodeForPath;
-            if (getNode) {
-                return getNode(interWidgetPath);
+            if (widget?.getDOMNodeForPath) {
```
jeremywiebe, line 1495: Optional chaining makes for slightly less verbose (but equivalent) code

packages/perseus/src/types.ts
```diff
@@ -581,7 +637,7 @@ export type WidgetProps<
     questionCompleted?: boolean;
     onFocus: (blurPath: FocusPath) => void;
     onBlur: (blurPath: FocusPath) => void;
-    findWidgets: (arg1: FilterCriterion) => ReadonlyArray<Widget>;
+    findWidgets: (criterion: FilterCriterion) => ReadonlyArray<Widget>;
```
jeremywiebe, line 640: `arg1` just didn't say much about the arg. 🤔

packages/perseus/src/widgets/expression/expression.tsx
```diff
@@ -339,25 +344,19 @@ export class Expression extends React.Component<Props, ExpressionState> {
...
         return "expression";
-    };
+    }
 
-    setInputValue: (path: string, newValue: string, cb: () => void) => void = (
-        path: string,
-        newValue: string,
-        cb: () => void,
-    ) => {
+    setInputValue(path: FocusPath, newValue: string, cb: () => void) {
```
jeremywiebe, line 352: This is an example of how conforming to the actual interface helps clean things up. Also, this widget was incorrectly typing `path` as a string, when it's really a `FocusPath` (array of strings - widget ids).

packages/perseus/src/widgets/measurer/measurer.tsx
```diff
@@ -21,51 +17,62 @@ const defaultImage = {
...
-            rulerLength: 10,
-        };
-    },
-
-    getInitialState: function () {
-        return {};
-    },
-
-    componentDidMount: function () {
+type Props = {
```
jeremywiebe, line 20: This is just a standard migration from `createReactClass()` to a true `React.Component` class.

packages/perseus/src/widgets/measurer/measurer.tsx
```diff
@@ -76,8 +83,7 @@ const Measurer: any = createReactClass({
                 "rulerPixels",
                 "rulerLength",
             ],
-            function (prop) {
-                // @ts-expect-error - TS2683 - 'this' implicitly has type 'any' because it does not have a type annotation.
+            (prop) => {
```
jeremywiebe, line 86: Converting to an arrow function to preserve the `this` binding.

packages/perseus/src/widgets/measurer/measurer.tsx
```diff
@@ -181,12 +185,8 @@ const Measurer: any = createReactClass({
                 <div className="graphie" ref="graphieDiv" />
             </div>
         );
-    },
-});
-
-_.extend(Measurer, {
-    validate: noopValidator(1),
```
jeremywiebe, line 188: This moved to top of the class.

packages/perseus/src/widgets/python-program/python-program.tsx
```diff
@@ -24,12 +24,20 @@ type DefaultProps = {
...
+ * This renders the program in an iframe.
+ */
+class PythonProgram extends React.Component<Props> implements Widget {
     static defaultProps: DefaultProps = {
         height: 400,
     };
 
+    iframe = React.createRef<HTMLIFrameElement>();
+
+    getDOMNodeForPath(path: FocusPath) {
```
jeremywiebe, line 37: I'm debating if this is a good idea.  In reality, this is not used. But the `Widget` interface has only optional methods _and_ this widget doesn't implement any of them.  As a result, we're triggering Typescript's [weak type detection](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-4.html#weak-type-detection).  Add this function silences the warning without the need for `@ts-expect-error` which I deemed as even worse!

packages/perseus/src/widgets/measurer/measurer.tsx
handeyeco: Just FYI:
- https://khanacademy.atlassian.net/browse/LEMS-365
- https://khanacademy.atlassian.net/browse/LEMS-2393

packages/perseus/src/widgets/expression/expression.tsx
jeremywiebe: I tried that, and it seems like TypeScript is not built to do that automatically. 

image

packages/perseus/src/widgets/measurer/measurer.tsx
jeremywiebe: Thanks Matthew. I should have commented further... I needed to do the transform so that I could make a class that conforms to the Widget interface. I'd not have touched it otherwise. :)

jeremywiebe: I'll mark 2393 as done when I get this deployed.

packages/perseus/src/widgets/categorizer/categorizer.tsx
```diff
@@ -40,7 +45,10 @@ type State = {
     uniqueId: string;
 };
 
-export class Categorizer extends React.Component<Props, State> {
+export class Categorizer
+    extends React.Component<Props, State>
+    implements Widget
+{
```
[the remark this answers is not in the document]
jeremywiebe: It would have to be a generic type `WidgetComponent<Props, State> = React.Component<Props, State> implements Widget`. 🤔  I'm a bit concerned it'd be too restrictive.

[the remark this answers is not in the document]
jeremywiebe: Ok, I tried but it feels like I'm going down a rabbit hole. I can make an abstract class:

```ts
abstract class WidgetComponent<Props, State>
    extends React.Component<Props, State>
    implements Widget
{}
```

But then it complains that no methods from Widget are implemented in this abstract class.  If I add one of the functions and mark _it_ as abstract, then any widget that hasn't implemented that function now need to implement it, because it's abstract. 

I'm going to leave it as is, but at least we explored the idea.

packages/perseus/src/types.ts
```diff
@@ -24,11 +25,66 @@ export type Dimensions = {
 
 export type DeviceType = "phone" | "tablet" | "desktop";
 
-// TODO(CP-4839): Create a proper type for Widget
-// Is this the same as the Widget type in `renderer.jsx`?
-export type Widget = any;
+export type SerializedState = {
+    [id: string]: any;
+};
```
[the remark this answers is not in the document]
jeremywiebe: This is another non-obvious part of Perseus. The `Renderer` will return the widget's render props by default. But, each widget can provide its own `getSerializedState` and `restoreSerializedState` functions which return/consume anything. 

I think the best way to solve this would be to make `Widget` a generic that defined a type for what this pair of functions returns/consumes, but I'd like to avoid that for now. 

I think what I'll do for the short term is to change this type to be a `Record<string, any>` as that feels like the more generic/obvious way to say "it's an object".

packages/perseus/src/types.ts
```diff
@@ -24,11 +25,66 @@ export type Dimensions = {
...
+
+    simpleValidate?: (
+        options?: any,
+        onOutputError?: (
+            widgetId: any,
+            value: string,
+            message?: string | null | undefined,
+        ) => unknown | null | undefined,
+    ) => PerseusScore;
+    showRationalesForCurrentlySelectedChoices?: (options?: any) => void;
```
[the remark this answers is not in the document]
jeremywiebe: Yeah! Added to that, there's now also a prop `showSolutions` on the `Renderer` that does something similar (https://github.com/Khan/perseus/blob/fec2febaabe5a146464f4713d965ccf41be1db24/packages/perseus/src/renderer.tsx#L126).

packages/perseus/src/widgets/passage/passage.tsx
```diff
@@ -20,13 +20,17 @@ import type {
...
 } from "../../perseus-types";
-import type {WidgetExports, WidgetProps} from "../../types";
+import type {Widget, WidgetExports, WidgetProps} from "../../types";
 import type {
     PerseusPassageRubric,
     PerseusPassageUserInput,
 } from "../../validation.types";
 import type {SingleASTNode} from "@khanacademy/simple-markdown";
 
+function isPassageWidget(widget: any): widget is Passage {
```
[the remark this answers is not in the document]
jeremywiebe: Yes, thanks for the prompt!

PR #1699, opened 2024-09-30
Remove c8 suppressions

handeyeco:
## Summary:
I don't think they're used anymore.


Myranae (approved):
Looks good! Seems like c8 was what was used before codecov or something along those lines. Haven't heard of it before, and all tests still pass!

jeremywiebe (changes requested):
These lines tell Jest to not gather code coverage for the annotated lines. We use the `"v8"` `coverageProvider` ([here](https://github.com/Khan/perseus/blob/0a4e48da1d69ac057bdbcf285b1be69fd17f26d7/config/test/test.config.js#L85)). This provider uses different annotations to suppress coverage collection (https://jestjs.io/docs/configuration#collectcoverage-boolean). 

I have no idea why it's `c8` instead of `v8`. That's annoyingly inconsistent. But yeah, I think we want to keep these for now.

handeyeco:
They are being used: https://khanacademy.slack.com/archives/C01AZ9H8TTQ/p1727807134622889

PR #1700, opened 2024-09-30
Split out InteractiveGraph validator

handeyeco:
## Summary:
Split out the validation logic from the InteractiveGraph widget

Issue: LEMS-2356

## Test plan:
Nothing should change, mostly just moving code around


packages/perseus/src/widgets/interactive-graph.tsx
```diff
@@ -135,6 +135,24 @@ type DefaultProps = {
     graph: Props["graph"];
 };
 
+// TODO: there's another, very similar getSinusoidCoefficients function
+// they should probably be merged
+export function getSinusoidCoefficients(
```
handeyeco, line 133: There are two `getSinusoidCoefficients` functions that are nearly identical except one returns an array and one returns an object. I didn't want to go down the refactoring rabbit-hole, so I kept both but made this an exported function rather than a static method.

benchristel (approved):
LGTM!

packages/perseus/src/widgets/interactive-graph.tsx
```diff
@@ -135,6 +128,57 @@ type DefaultProps = {
...
+    const angularFrequency = Math.PI / (2 * (p2[0] - p1[0]));
+    const phase = p1[0] * angularFrequency;
+    const verticalOffset = p1[1];
+
+    return [amplitude, angularFrequency, phase, verticalOffset];
+}
+
+// TODO: there's another, very similar getQuadraticCoefficients function
+// they should probably be merged
+export function getQuadraticCoefficients(
```
handeyeco, line 151: Same here, one returns `QuadraticCoefficient` and the other returns `QuadraticCoefficient | undefined`

PR #1701, opened 2024-09-30
Focus on point after added

nicolecomputer:
## Summary:
This focus on the newly added point after its added so that learners who are using keyboard can quickly position points.

Issue: XXX-XXXX

## Test plan:

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/point.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/point.tsx
@@ -46,6 +46,16 @@ export function UnlimitedPointGraph(props: PointGraphProps) {
         height,
     ]);
     const [[left, top]] = useTransformVectorsToPixels([minX, maxY]);
+    const itemsRef = React.useRef<[SVGElement | null]>([]);
+
+    React.useEffect(() => {
+        const focusedIndex = props.graphState.focusedPointIndex;
+
+        if (focusedIndex !== null && itemsRef.current[focusedIndex] !== null) {
+            itemsRef.current[focusedIndex].focus();
+        }
+    }, [props.graphState.focusedPointIndex, itemsRef]);
+
     return (
         <>
             {/* This rect is here to grab clicks so that new points can be added */}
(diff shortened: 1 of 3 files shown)


packages/perseus/src/widgets/interactive-graphs/graphs/point.tsx
```diff
@@ -46,6 +46,16 @@ export function UnlimitedPointGraph(props: PointGraphProps) {
         height,
     ]);
     const [[left, top]] = useTransformVectorsToPixels([minX, maxY]);
+    const itemsRef = React.useRef<[SVGElement | null]>([]);
+
+    React.useEffect(() => {
+        const focusedIndex = props.graphState.focusedPointIndex;
+
+        if (focusedIndex !== null && itemsRef.current[focusedIndex] !== null) {
+            itemsRef.current[focusedIndex]?.focus();
```
nicolecomputer, line 55: You would think optional chaining isn't needed here because of the checks buuuuuut typescript _insists_ that its necessary. So 🤷‍♀️

packages/perseus/src/widgets/interactive-graphs/graphs/point.tsx
```diff
@@ -46,6 +46,16 @@ export function UnlimitedPointGraph(props: PointGraphProps) {
         height,
     ]);
     const [[left, top]] = useTransformVectorsToPixels([minX, maxY]);
+    const itemsRef = React.useRef<Array<SVGElement | null>>([]);
+
+    React.useEffect(() => {
```
benchristel, line 51: If the learner focuses something outside the graph, and then the graph re-renders, will this hook force focus back to the point? Or no, because we unset `graphState.focusedPointIndex` when the point blurs?

benchristel (approved):
I left a few questions and suggestions inline, but the overall approach looks good to me!

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -677,10 +683,18 @@ function doRemovePoint(
         return state;
     }
 
+    let nextFocusedPointIndex: number | null;
+    if (state.interactionMode === "mouse") {
+        nextFocusedPointIndex = null;
+    } else {
+        nextFocusedPointIndex =
+            state.coords.length > 1 ? state.coords.length - 2 : null;
```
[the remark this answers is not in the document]
nicolecomputer: The last is the one that's getting removed. This is the new last. (state is the previous state here)

packages/perseus/src/widgets/interactive-graphs/graphs/point.tsx
nicolecomputer: In experimenting with this it'll cause the point to blur and that will unfocus the point

packages/perseus/src/widgets/interactive-graphs/graphs/point.tsx
```diff
@@ -81,12 +91,21 @@ export function UnlimitedPointGraph(props: PointGraphProps) {
...
                             dispatch(actions.pointGraph.focusPoint(i));
                         } else {
                             if (event.relatedTarget?.id === REMOVE_BUTTON_ID) {
                                 return;
+                            } else if (
+                                event.relatedTarget?.classList.contains(
+                                    "movable-point",
+                                )
+                            ) {
+                                return;
```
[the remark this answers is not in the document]
nicolecomputer: If we blur when we move to another point then the delete button gets removed from the page and then quickly added back which looks like flashing to the learner. It's annoying and this just cleans it up a bit. I'll leave a note!

PR #1702, opened 2024-09-30
Split out NumberLineValidator

handeyeco:
## Summary:
Split out the validation logic for NumberLineValidator

Issue: LEMS-2478

## Test plan:

### DIFF
--- a/packages/perseus/src/widgets/number-line/number-line-validator.test.ts
+++ b/packages/perseus/src/widgets/number-line/number-line-validator.test.ts
@@ -1,0 +1,109 @@
+import numberLineValidator from "./number-line-validator";
+
+import type {
+    PerseusNumberLineRubric,
+    PerseusNumberLineUserInput,
+} from "../../validation.types";
+
+const baseInput: PerseusNumberLineUserInput = {
+    isTickCrtl: true,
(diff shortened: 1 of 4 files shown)


packages/perseus/src/widgets/number-line/number-line-validator.test.ts
```diff
@@ -0,0 +1,109 @@
...
+}
+
+function generateRubric(
+    extend?: Partial<PerseusNumberLineRubric>,
+): PerseusNumberLineRubric {
+    return {...baseRubric, ...extend};
+}
+
+describe("numberLineValidator", () => {
+    it(`is invalid when outside allowed range`, () => {
```
Myranae, line 45: Nit: Should the backticks/"`" be regular quotation marks in the tests?

Myranae (approved):
Thanks so much!

packages/perseus/src/widgets/number-line/number-line-validator.test.ts
handeyeco: Probably isn't a big deal either way. This is just an old weird habit of mine to use backticks in tests. Switched to quotes to be more consistent with our other tests.

PR #1705, opened 2024-09-30
Split out validation logic for iframe

Myranae:
## Summary:
For the server-side scoring project, we are moving validation logic out of each widget component. This moves the validation logic for the iframe widget and adds tests for the validator function.

Issue: LEMS-2480

## Test plan:
- Confirm all tests pass
- Confirm iframe stories still work in a ZND

### DIFF
--- a/packages/perseus/src/widgets/iframe/iframe.tsx
+++ b/packages/perseus/src/widgets/iframe/iframe.tsx
@@ -15,8 +15,10 @@ import {getDependencies} from "../../dependencies";
 import * as Changeable from "../../mixins/changeable";
 import Util from "../../util";
 
+import {iframeValidator} from "./iframe-validator";
+
 import type {PerseusIFrameWidgetOptions} from "../../perseus-types";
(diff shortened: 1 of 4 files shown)


packages/perseus/src/widgets/iframe/iframe.tsx
```diff
@@ -90,10 +96,9 @@ class Iframe extends React.Component<Props> {
         return Changeable.change.apply(this, args);
     };
 
-    simpleValidate: (arg1: any) => any = (rubric) => {
-        // @ts-expect-error - TS2339 - Property 'validate' does not exist on type 'typeof Iframe'.
-        return Iframe.validate(this.getUserInput(), rubric);
-    };
+    simpleValidate(): PerseusScore {
```
Myranae, line 99: Comment: I don't _think_ this has to be an arrow function based on how other `simpleValidate` methods are done, but I could be wrong.

handeyeco (approved):
Perfect, thank you!

PR #1707, opened 2024-10-01
Split out Matcher validator

handeyeco:
## Summary:
Splits out validation logic from Matcher

Issue: LEMS-2479

## Test plan:
Nothing should change, just moving some logic

### DIFF
--- a/packages/perseus/src/widgets/matcher/matcher.tsx
+++ b/packages/perseus/src/widgets/matcher/matcher.tsx
@@ -10,9 +10,11 @@ import {getDependencies} from "../../dependencies";
 import Renderer from "../../renderer";
 import Util from "../../util";
 
+import matcherValidator from "./matcher-validator";
+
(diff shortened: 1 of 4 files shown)


packages/perseus/src/widgets/matcher/matcher.tsx
```diff
@@ -63,6 +65,13 @@ export class Matcher extends React.Component<Props, State> {
         texRendererLoaded: false,
     };
 
+    static validator(
```
Myranae, line 68: Comment: These are usually `static validate`, right?

Myranae (approved):
Looks good :) Thanks!

packages/perseus/src/widgets/matcher/matcher.tsx
handeyeco: 😊 Yep whoops

PR #1708, opened 2024-10-01
Split out validation logic for orderer

Myranae:
## Summary:
For the server-side scoring project, we are moving validation logic out of each widget component. This moves the validation logic for the orderer widget, moves validation focused tests, and adds a test.

Issue: LEMS-2477

## Test plan:
- Confirm all tests pass
- Confirm widget still works as expected in Storybook


packages/perseus/src/widgets/orderer/orderer-validator.test.ts
```diff
@@ -0,0 +1,54 @@
+import {act} from "@testing-library/react";
+
+import {renderQuestion} from "../__testutils__/renderQuestion";
```
handeyeco, line 3: Ideally we would keep React-centric tests where they are and have non-React tests specifically for the validator.

Myranae: Will fix this for sure.

handeyeco (approved):
Thanks 🌈

PR #1710, opened 2024-10-01
Add renderer level tests for shuffled scoring

handeyeco:
## Summary:
We were worried about SSS and shuffling, but it seems Radio's `getUserInput` is smart about un-shuffling answers before scoring.

Issue: LEMS-2435

## Test plan:
This PR is just tests

### DIFF
--- a/packages/perseus/src/widgets/radio/__tests__/radio.test.ts
+++ b/packages/perseus/src/widgets/radio/__tests__/radio.test.ts
@@ -5,16 +5,20 @@ import {userEvent as userEventLib} from "@testing-library/user-event";
 import {clone} from "../../../../../../testing/object-utils";
 import {testDependencies} from "../../../../../../testing/test-dependencies";
 import * as Dependencies from "../../../dependencies";
+import {mockStrings} from "../../../strings";
 import {renderQuestion} from "../../__testutils__/renderQuestion";
 import PassageWidget from "../../passage";
+import radioValidator from "../radio-validator";
 
 import {
     questionAndAnswer,
     multiChoiceQuestionAndAnswer,
+    shuffledQuestion,
 } from "./radio.testdata";
 
 import type {PerseusRenderer} from "../../../perseus-types";
 import type {APIOptions} from "../../../types";
+import type {PerseusRadioUserInput} from "../../../validation.types";
 import type {UserEvent} from "@testing-library/user-event";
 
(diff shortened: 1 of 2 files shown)


packages/perseus/src/widgets/radio/__tests__/radio.test.ts
```diff
@@ -874,3 +879,108 @@ describe("multi-choice question", () => {
...
+            screen.getByRole("radio", {name: /None of the above$/}),
+        );
+
+        const userInput = renderer.getUserInput()[0] as PerseusRadioUserInput;
+        const rubric = shuffledNoneQuestion.widgets["radio 1"].options;
+        const score = radioValidator(userInput, rubric, mockStrings);
+
+        // Assert
+        expect(score).toHaveBeenAnsweredCorrectly();
+        expect(renderer).toHaveBeenAnsweredCorrectly();
```
jeremywiebe, line 961: Are you asserting the `renderer` here just to ensure they both say the same thing?

packages/perseus/src/widgets/radio/__tests__/radio.testdata.ts
```diff
@@ -323,3 +323,96 @@ export const multiChoiceQuestionAndAnswer: [
         [0, 1, 2, 3],
     ],
 ];
+
+export const shuffledQuestion: PerseusRenderer = {
```
jeremywiebe, line 327: **thought**: The `interactive-graph` team built a whole test data generator api using the builder pattern. I long for a day when we have builders for all types of widgets. 

```ts
const shuffledQuestion = new RendererBuilder()
    .withRadioWidget()
        .withIncorrectAnswer("Incorrect Choice 1")
        .withIncorrectAnswer("Incorrect Choice 2")
        .withCorrectAnswer("Correct Choice")
        .withIncorrectAnswer("Incorrect Choice 3")
        .randomized()
        .endWidget()
    .build()
```

No action required.

jeremywiebe (approved):
Nice. Thanks for verifying this and locking down the behaviour with tests. :)

packages/perseus/src/widgets/radio/__tests__/radio.test.ts
handeyeco: Yeah, I just figured "why not" - I already had the renderer since that was the simplest way I could think to test this.

PR #1713, opened 2024-10-01
Validator-related cleanup

handeyeco:
## Summary:
Just some small tweaks to clean things up

### DIFF
--- a/packages/perseus/src/validation.types.ts
+++ b/packages/perseus/src/validation.types.ts
@@ -33,6 +33,12 @@ import type {Relationship} from "./widgets/number-line/number-line";
 
 export type UserInputStatus = "correct" | "incorrect" | "incomplete";
 
+// TODO (LEMS-2396): remove validation logic from widgets that don't validate
+export type EmptyUserInput = Empty;
+
+// TODO (LEMS-2396): remove validation logic from widgets that don't validate
+export type NullUserInput = null;
+
 export type PerseusCategorizerRubric = PerseusCategorizerWidgetOptions;
 
 export type PerseusCategorizerUserInput = {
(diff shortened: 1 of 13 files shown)


packages/perseus/src/widgets/deprecated-standin/__tests__/deprecated-standin.test.ts
```diff
@@ -41,51 +41,24 @@ describe("Deprecated Standin widget", () => {
...
+    it("should return an empty object for getUserInput()", () => {
         // Arrange
         const {renderer} = renderQuestion(question);
 
         // Act
-        const score = renderer.scoreWidgets();
+        const userInput = renderer.getUserInput();
 
         // Assert
-        expect(score).toMatchInlineSnapshot(`
```
handeyeco, line 80: We had two tests that were basically the same, just written differently.

packages/perseus/src/validation.types.ts
```diff
@@ -33,6 +33,12 @@ import type {Relationship} from "./widgets/number-line/number-line";
 
 export type UserInputStatus = "correct" | "incorrect" | "incomplete";
 
+// TODO (LEMS-2396): remove validation logic from widgets that don't validate
+export type EmptyUserInput = Empty;
+
+// TODO (LEMS-2396): remove validation logic from widgets that don't validate
+export type NullUserInput = null;
```
handeyeco, line 40: Decided to consolidate these since they're the same and will hopefully be going away in an upcoming PR.

packages/perseus/src/widgets/deprecated-standin/index.tsx
```diff
@@ -22,13 +22,13 @@ class DeprecatedStandin extends React.Component<Props> {
         return noopValidator(1);
     }
 
-    getUserInput: () => UserInput = () => {
+    getUserInput(): UserInput {
```
Myranae, line 25: Question, nb: I think you were just removing the arrow functions here, but I noticed this has an empty userInput return. Would you prefer the return type to be `EmptyUserInput` similar to how you changed the other `getUserInput` methods in this PR? I know that type is part of `UserInput`, but was thinking we might want to be consistent with which way we do it.

Myranae (approved):
Thanks!

packages/perseus/src/widgets/deprecated-standin/index.tsx
handeyeco: Good catch, fixed

PR #1715, opened 2024-10-02
Remove simpleValidate and static validate

handeyeco:
## Summary:

The goal of this PR is to completely remove validators (`simpleValidate` and `static validate`) off of React components. For the most part this was pretty chill, with the exception of Group widgets ([LEMS-2415](https://khanacademy.atlassian.net/browse/LEMS-2415)) which required some reworking of the Renderer component.

The next step is to use these new functional methods in Webapp.

Issue: LEMS-2424

## Test plan:
Functionality shouldn't generally change, at least not from a learner perspective. To manually confirm, go on test everything and answer some questions.

[LEMS-2415]: https://khanacademy.atlassian.net/browse/LEMS-2415?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/perseus/src/__tests__/widgets.test.ts
```diff
@@ -12,19 +12,8 @@ describe("Widget API support", () => {
     // validate function. Not all widgets support this function so even though
     // this list looks exhaustive, it's not!
     it.each([
-        "deprecated-standin",
```
jeremywiebe, line 15: Over the last little while I've been thinking that perhaps our `WidgetExports<T>` should be where a widget can describe if it's scorable or not.  In doing so, the type would require all the necessary functions needed to do scoring (such as `getUserInput()`).  

No action needed and I think your idea of attaching the validators to each `WidgetExports<T>` gets us partially there. Just thinkin' aloud.

packages/perseus/src/widgets/expression/expression.tsx
```diff
@@ -228,34 +200,6 @@ export class Expression
...
-        const score = expressionValidator(
-            this.getUserInput(),
-            rubric,
-            this.context.strings,
-            this.context.locale,
-        );
-
-        // "scoring" is a flag that indicates when we are checking answers.
-        // otherwise, we may just be checking validity after changes.
-        if (scoring && score.type !== "invalid") {
```
jeremywiebe, line 243: Just to clarify, this means we'll lose this analytics event, right?  I'm not sure that's the worst thing as I don't know that we look at the analytics for this event ever.

config/test/custom-matchers.ts
```diff
@@ -6,7 +6,7 @@ declare global {
     // eslint-disable-next-line @typescript-eslint/no-namespace
     namespace jest {
         interface Matchers<R> {
-            toHaveBeenAnsweredCorrectly(): R;
+            toHaveBeenAnsweredCorrectly(shouldHavePoints?: boolean): R;
```
jeremywiebe, line 9: **nit**: This boolean is somewhat non-obvious when looking at the call sites:

```ts 
expect(renderer)toHaveBeenAnsweredCorrectly(false);
```

It makes for more verbose code, but what are your thoughts on:

```ts 
expect(renderer).toHaveBeenAnsweredCorrectly({shouldHavePoints: false});`
```

jeremywiebe (approved):
Nice! I especially like collapsing all the Empty/Null types into a single use.

packages/perseus/src/widgets/expression/expression.tsx
handeyeco: Yeah, I was going to ask you about that. I don't think it's being used, but if it is being used...it can't be here anyway because `simpleValidate` is going away.

packages/perseus/src/__tests__/renderer.test.tsx
```diff
@@ -1709,15 +1750,14 @@ describe("renderer", () => {
...
-            const userInput = renderer.getUserInputForWidgets();
+            const userInput = renderer.getUserInputMap();
 
             // Assert
             expect(userInput).toMatchInlineSnapshot(`
                 {
                   "dropdown 1": {
                     "value": 1,
                   },
-                  "image 1": null,
```
handeyeco, line 1720: I'm hoping this isn't a problem either. I don't see the point in storing "user data" for widgets that aren't interactive.

packages/perseus/src/types.ts
```diff
@@ -44,6 +49,9 @@ export type SerializedState = Record<string, any>;
  * scoring, state serialization/deserialization, and focus management.
  */
 export interface Widget {
+    // don't use getType; it's just a dummy method to help TypeScript's weak
+    // typing to recognize non-interactive widgets as Widgets
+    getType?: () => string;
```
handeyeco, line 54: I expect push-back on this one. Here's the problem: when I removed all the validation methods on non-interactive widgets I got TS errors to the effect of "there's no overlap between Passage and Widget." That's because non-interactive widgets don't implement _any_ of the optional methods on Widget and Widget only has optional methods.

By adding this silly method and having non-interactive widgets implement it, TS suddenly recognizes them as Widgets.

What I didn't want them to do is implement _real_ methods they didn't need to because I think that's confusing long-term. For instance if we made them all return `null` for `getUserInput`, suddenly we're having to maintain code that expects `null` as a possibility.

I wanted to make this a required method, but it broke:

```TS
type WidgetExports<T extends React.ComponentType<any> & Widget = React.ComponentType<any>>
```

I don't know why.

packages/perseus/src/widgets/deprecated-standin/deprecated-standin.tsx
```diff
@@ -0,0 +1,46 @@
...
+import {PerseusI18nContext} from "../../components/i18n-context";
+import noopValidator from "../__shared__/noop-validator";
+
+import type {Widget, WidgetExports} from "../../types";
+
+// The props are type `any` on purpose so that this can receive props
+// from any deprecated widget
+type Props = any;
+
+class DeprecatedStandin extends React.Component<Props> implements Widget {
```
handeyeco, line 13: It was bugging me that `DeprecatedStandin` was declared in `index.tsx` rather than `deprecated-standin.tsx`. It makes it hard to find and none of our other widgets do that.

packages/perseus/src/renderer-util.ts
```diff
@@ -0,0 +1,100 @@
...
+            );
+        }
+
+        if (score) {
+            return Util.scoreIsEmpty(score);
+        }
+    });
+}
+
+// WARNING: recursive in order to handle Group widgets
```
benchristel, line 53: Why does this need a warning?

packages/perseus/src/renderer.tsx
```diff
@@ -1706,7 +1700,24 @@ class Renderer extends React.Component<Props, State> {
                 return widget.getUserInput();
             }
         });
-    };
+    }
+
+    /**
+     * Returns an object of the widget `.getUserInput()` results
+     */
+    getUserInputMap(): UserInputMap {
```
benchristel, line 1708: Could this method be `private`?

packages/perseus/src/renderer.tsx
```diff
@@ -1706,7 +1700,24 @@ class Renderer extends React.Component<Props, State> {
...
+        const userInputMap = {};
+        this.widgetIds.forEach((id: string) => {
+            const widget = this.getWidgetInstance(id);
+            // Handle Groups, which have their own sets of widgets
+            if (widget?.getUserInputMap) {
+                userInputMap[id] = widget.getUserInputMap();
+            } else if (widget?.getUserInput) {
+                userInputMap[id] = widget.getUserInput();
+            }
```
benchristel, line 1717: Why does Group need to be different from the other widgets here? Is the issue that we can't change the interface of `getUserInput` because it needs to stay compatible with e.g. Khanmigo?

benchristel (approved):
This looks good! I'm super happy that we can make big changes to Perseus incrementally like this.

My one concern is that Group has a different interface from the other widgets, and I wonder if it needs to, and if the asymmetry will cause confusion and extra complexity in the future. It seems like maybe we could make everything symmetrical with the right abstraction. I don't have all the context, or any answers, though.

packages/perseus/src/validation.types.ts
```diff
@@ -206,12 +230,19 @@ export type UserInput =
...
     | PerseusRadioUserInput
     | PerseusSorterUserInput
-    | PerseusTableUserInput
-    | EmptyUserInput;
+    | PerseusTableUserInput;
+
+export type UserInputMap = {[widgetId: string]: UserInput | UserInputMap};
+
+/**
+ * deprecated prefer using UserInputMap
```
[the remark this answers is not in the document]
handeyeco: `@deprecated` is supported but then we have a lint rule that yells whenever you import something that is flagged as `@deprecated` so then you have to add a bunch of error suppressions but in TS error suppressions suppress _all_ errors not just specific one and at a certain point I said "eff it" and changed `@deprecated` to `deprecated`.

packages/perseus/src/widgets/__shared__/noop-validator.test.tsx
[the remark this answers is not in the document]
handeyeco: I already noted that I need to come back and change that. I just didn't want to do it in this PR because moving/renaming things while making changes makes it look like the whole file changed: https://khanacademy.atlassian.net/browse/LEMS-2520

packages/perseus/src/__tests__/renderer.test.tsx
```diff
@@ -1435,7 +1436,7 @@ describe("renderer", () => {
             expect(input).toStrictEqual([
                 {currentValue: "0"},
                 {currentValue: "1"},
-                null, // image widget doesn't implement getUserinput
+                undefined, // image widget doesn't implement getUserinput
```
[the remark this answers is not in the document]
handeyeco: This would require us to keep external-facing methods on React components that don't need those specific external-facing methods.

I know you mentioned [in another comment](https://github.com/Khan/perseus/pull/1715#discussion_r1790721458) that you think we should keep functionality even when it's a no-op. I can kind of see your point (even if I don't agree), but I definitely don't think we want to do that with React anti-patterns (attaching external-facing methods to React components).

I suppose we could do `.getUserInput() ?? null`. Maybe that would be better than leaving it as `undefined`? It would be the tiniest bit more complexity, but at least it'd be compatible with existing behavior.

packages/perseus/src/types.ts
handeyeco: I originally tried a required method, but this blew up and I didn't want to go down a rabbit hole:

``` TS
export type WidgetExports<
>
```
But I moved the widgets that needed it from "getType" to "isWidget".

> I'm not sure I agree. getUserInput is currently typed as...

This was changed in this PR to be `() => UserInputArray | UserInput | undefined` and the only reason `undefined` is there is because of Group (and I'd like to remove that).

packages/perseus/src/widgets/video/video.tsx
```diff
@@ -132,4 +117,6 @@ export default {
     defaultAlignment: "block",
     supportedAlignments: ["block", "float-left", "float-right", "full-width"],
     widget: Video,
+    // TODO: things that aren't interactive shouldn't need validators
+    validator: () => noopValidator(),
```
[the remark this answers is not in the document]
handeyeco: I don't _totally_ agree, but I could be sold on this. IMO having defaults that don't do anything (1) add cognitive noise and (2) makes it harder to tell what is/isn't be used. For instance if you were trying to remove the `accessible` property, it might have made it seem more widely used if every widget export implemented it. Granted you would see that it was declared and maybe not used, but we have a lot of misdirection in both Perseus itself and across repos that makes it difficult to see what is actually being used.

packages/perseus/src/renderer.tsx
handeyeco: IIRC, I couldn't change the format of `getUserInput` while `guessAndScore` exists. The next step is to break `guessAndScore` though, so I filed a ticket to followup on this once that's done: https://khanacademy.atlassian.net/browse/LEMS-2523

packages/perseus/src/__tests__/renderer.test.tsx
Myranae: Is there anywhere that userInput needs to reflect all the widgets that are available? I could see it possibly being useful for that, but it does seem silly for userInput to track widgets that will never have userInput.

packages/perseus/src/renderer-util.ts
```diff
@@ -0,0 +1,99 @@
...
+        if (!widget || widget.static) {
+            // Static widgets shouldn't count as empty
+            return false;
+        }
+
+        let score: PerseusScore | null = null;
+        const userInput = userInputMap[id];
+        const validator = getWidgetValidator(widget.type);
+
+        if (widget.type === "group") {
```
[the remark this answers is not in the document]
jeremywiebe: Another way to approach this would be to have `group` have a validator and have it call back into this function.  That would eliminate the custom code paths for just `group`.  We also have  `graded-group` (and maybe `graded-group-set`) that do similar tricks. Not sure if those need to be addressed here too.

packages/perseus/src/renderer.tsx
```diff
@@ -1696,8 +1687,11 @@ class Renderer extends React.Component<Props, State> {
 
     /**
      * Returns an array of the widget `.getUserInput()` results
+     *
+     * TODO: can we remove this?
+     * @deprecated use getUserInputMap
```
jeremywiebe, line 1692: I think we should try!  The return value here is an array that is very unfriendly (as we have no mapping back to widget ID for each entry).

packages/perseus/src/types.ts
```diff
@@ -44,6 +49,12 @@ export type SerializedState = Record<string, any>;
  * scoring, state serialization/deserialization, and focus management.
  */
 export interface Widget {
+    /**
+     * don't use isWidget; it's just a dummy property to help TypeScript's weak
+     * typing to recognize non-interactive widgets as Widgets
+     * @deprecated
```
jeremywiebe, line 55: Yeah 😞 , maybe we can eventually expand this to have a second interface like `InteractiveWidget` which contains all the functions that truly interactive widgets implement.

packages/perseus/src/types.ts
```diff
@@ -574,6 +567,18 @@ export type WidgetTransform = (
...
+export type WidgetValidatorFunction = (
+    // The user data needed to score
+    userInput: any,
+    // The scoring criteria to score against
+    rubric: any,
+    // Strings, for error messages in invalid widgets
+    string?: PerseusStrings,
+    // Locale, for math evaluation
+    // (1,000.00 === 1.000,00 in some countries)
+    locale?: string,
```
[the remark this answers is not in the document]
jeremywiebe: Adding it to user input feels very useful especially as we ship it to the server. The rubric will likely be requested from the content service (or whatever fetches it) by including the locale, so unsure about rubric.

packages/perseus/src/widgets/expression/expression.tsx
jeremywiebe: Given that a portion of these analytics were to compare usages of the different keypads, I think it's ok to lost this for now.  Perhaps it would fit in `getUserInput`? But I'm not sure how much that function is called (it might be more than once per "answer check").

jeremywiebe (commented):
I had a few thoughts that might be useful even though this PR is closed.

PR #1717, opened 2024-10-02
Remove inaccurate WidgetDict type and bespoke typings for `widgets` key

jeremywiebe:
## Summary:

After releasing the PR to migrate the `Widget` type to an interface, I noticed that we have several different ways of typing the `widgets` key in our main Perseus render type (`PerseusRenderer`) and throughout the codebase. 

Today, we have the really useful `PerseusWidgetsMap` type which maintains the correct widget options type for each widget. So this PR removes the less-specific `WidgetDict` type and other bespoke typings (such as `{[key: string]: Widget}`) in favour of the `PerseusWidgetsMap`.

Issue: "none"

## Test plan:

`yarn tsc`

### DIFF
--- a/.changeset/green-pants-hear.md
+++ b/.changeset/green-pants-hear.md
@@ -1,5 +1,5 @@
 ---
-"@khanacademy/perseus": patch
+"@khanacademy/perseus": major
 "@khanacademy/perseus-editor": patch
 ---
 
--- a/.changeset/green-pants-hear.md
+++ b/.changeset/green-pants-hear.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/perseus": patch
+"@khanacademy/perseus-editor": patch
+---
+
+Remove duplicate WidgetDict type and bespoke typings of `widgets` in favour of `PerseusWidgetsMap` type.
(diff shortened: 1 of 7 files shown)


.changeset/green-pants-hear.md
benchristel, line 1: Since this PR removes an export from `index.ts`, should we consider it a major release?

packages/perseus-editor/src/editor.tsx
```diff
@@ -687,8 +686,7 @@ class Editor extends React.Component<Props, State> {
 
         const newContent = newPrelude + widgetContent + newPostlude;
 
-        const newWidgets = _.clone(this.props.widgets);
-        // @ts-expect-error TS(2345) Type '"categorizer" | undefined' is not assignable to type '"deprecated-standin"'.
+        const newWidgets = {...this.props.widgets};
         newWidgets[id] = {
             options: Widgets.getEditor(widgetType)?.defaultProps,
             type: widgetType as PerseusWidget["type"],
```
benchristel, line 692: We can remove this unsafe cast now!

```suggestion
            type: widgetType,
```

benchristel (approved):
I left a couple comments inline, but LGTM overall! Thanks for doing this!

.changeset/green-pants-hear.md
jeremywiebe: Yeah, I guess we should. :)  I still feel stingy with major version numbers, but I guess it's better to stick to semver than worry about burning major version numbers.

PR #1719, opened 2024-10-02
New event logging to Interactive Graph.

catandthemachines:
## Summary:
Adding new analytic event to interactive graph when it renders while adding details on what type of graph is rendered.

Issue: LEMS-2324

## Test plan:

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
@@ -95,7 +95,7 @@ export const MafsGraph = (props: MafsGraphProps) => {
     const uniqueId = React.useId();
     const descriptionId = `interactive-graph-description-${uniqueId}`;
     const graphRef = React.useRef<HTMLElement>(null);
-    const dependencies = useDependencies();
+    const {analytics} = useDependencies();
 
     // Set up the SVG attributes for the nested SVGs that help lock
     // the grid and graph elements to the bounds of the graph.
@@ -117,11 +117,9 @@ export const MafsGraph = (props: MafsGraphProps) => {
     const {strings} = usePerseusI18n();
 
     useOnMountEffect(() => {
-        dependencies.analytics.onAnalyticsEvent({
+        analytics.onAnalyticsEvent({
             type: "perseus:interactive-graph-widget:rendered",
-            payload: {
-                type: type,
-            },
+            payload: {type},
         });
     });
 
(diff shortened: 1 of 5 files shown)


packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -87,12 +88,14 @@ export const MafsGraph = (props: MafsGraphProps) => {
...
         fullGraphAriaDescription,
     } = props;
+    const {type} = state;
     const [width, height] = props.box;
     const tickStep = props.step as vec.Vector2;
 
     const uniqueId = React.useId();
     const descriptionId = `interactive-graph-description-${uniqueId}`;
     const graphRef = React.useRef<HTMLElement>(null);
+    const dependencies = useDependencies();
```
jeremywiebe, line 98: A slight stylistic improvement that might make it easier for future readers to see what parts of `dependencies` are being used:

```suggestion
    const {analytics} = useDependencies();
```

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -113,6 +116,15 @@ export const MafsGraph = (props: MafsGraphProps) => {
 
     const {strings} = usePerseusI18n();
 
+    useOnMountEffect(() => {
+        dependencies.analytics.onAnalyticsEvent({
+            type: "perseus:interactive-graph-widget:rendered",
+            payload: {
+                type: type,
+            },
```
jeremywiebe, line 124: **suggestion**: We can shorten this shorthand property naming:

```suggestion
            payload: { type },
```

packages/perseus/src/widgets/interactive-graphs/types.ts
```diff
@@ -14,6 +14,32 @@ export type MafsGraphProps<T extends InteractiveGraphState> = {
...
+    | "LINEAR_SYSTEM"
+    | "LINEAR"
+    | "RAY"
+    | "NONE"
+    | "POLYGON"
+    | "POINT"
+    | "CIRCLE"
+    | "QUADRATIC"
+    | "SINUSOID";
+
```
jeremywiebe, line 42: **suggestion**: I'm guessing the `InteracativeGraphEnum` matches the enum values for the event in CEDAR?  In that case, I'd recommend that this mapping live in webapp's adapter rather than here. Otherwise we're pulling CEDAR concepts into the Perseus codebase. I think it's completely fine to use the `interactive-graph`'s graph type in the event and then map it to the CEDAR value in webapp.

jeremywiebe (approved):
This looks good, but I'd recommend moving the mapping of interactive graph type into webapp if my reading of this is correct (see comment). :)

benchristel (approved):
LGTM!

PR #1723, opened 2024-10-03
Fix to event details fired from interactive graph.

catandthemachines:
## Summary:
Missing two addtional event details fired from interactive graph

Issue: LEMS-2324

## Test plan:
N/A

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
@@ -119,7 +119,11 @@ export const MafsGraph = (props: MafsGraphProps) => {
     useOnMountEffect(() => {
         analytics.onAnalyticsEvent({
             type: "perseus:interactive-graph-widget:rendered",
-            payload: {type},
+            payload: {
+                type,
+                widgetType: "INTERACTIVE_GRAPH",
+                widgetId: "interactive-graph",
+            },
         });
     });
 
--- a/.changeset/lovely-ants-itch.md
+++ b/.changeset/lovely-ants-itch.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/perseus": patch
+"@khanacademy/perseus-core": patch
+---
+
+Updating event data sent within interactive graph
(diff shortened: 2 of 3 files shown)


SonicScrewdriver (approved):
Thank you! LGTM :)

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -119,7 +119,11 @@ export const MafsGraph = (props: MafsGraphProps) => {
     useOnMountEffect(() => {
         analytics.onAnalyticsEvent({
             type: "perseus:interactive-graph-widget:rendered",
-            payload: {type},
+            payload: {
+                type,
+                widgetType: "INTERACTIVE_GRAPH",
+                widgetId: "interactive-graph",
```
jeremywiebe, line 125: Oh, this isn't right.  The `widgetType` should be [`interactive-graph`](https://github.com/Khan/perseus/blob/3e16972293b757e8631f4c5d8c3fb121babf5324/packages/perseus/src/widgets/interactive-graph.tsx#L2410) (unfortunately, Perseus uses `name` and `widgetType` in different places, they're the same concept) and `widgetId` is typically something like `interactive-graph 1`. 

It looks like we [pass down](https://github.com/Khan/perseus/blob/3e16972293b757e8631f4c5d8c3fb121babf5324/packages/perseus/src/widgets/interactive-graph.tsx#L1861) the `widgetId` (which is a [part](https://github.com/Khan/perseus/blob/23f639e88f75c5678a810a6e440d0bb937facc4a/packages/perseus/src/types.ts#L643) of the [`WidgetProps`](https://github.com/Khan/perseus/blob/3e16972293b757e8631f4c5d8c3fb121babf5324/packages/perseus/src/widgets/interactive-graph.tsx#L124) type.  

We don't define that prop on the `StatefulMafsGraphProps` type, but it's definitely passed down.  

We'll need to update this to pass the actual `widgetId` so that our analytics data is correct.

jeremywiebe (commented):
@catandthemachines Sorry that I missed this before you landed it. We need to fix the values passed to `widgetType` and `widgetId` :)

PR #1725, opened 2024-10-04
Refine Categorizer Rubric and Input types

Myranae:
## Summary:
As part of the Server Side Scoring project, this refactors Categorizer's Rubric type to contain only what is needed for scoring. Also updates the UserInput type to reference the Rubric type. 

Issue: LEMS-2463

## Test plan:
- Confirm all tests pass
- Confirm Categorizer stories still work as expected


packages/perseus/src/validation.types.ts
```diff
@@ -39,10 +38,14 @@ export type EmptyUserInput = Empty;
...
-export type PerseusCategorizerRubric = PerseusCategorizerWidgetOptions;
+export type PerseusCategorizerRubric = {
+    // The correct answers where index relates to the items and value relates
+    // to the category.  e.g. [0, 1, 0, 1, 2]
+    values: ReadonlyArray<number>;
+};
 
 export type PerseusCategorizerUserInput = {
-    values: ReadonlyArray<number>;
+    values: PerseusCategorizerRubric["values"];
```
Myranae, line 48: This could instead just point directly to the Rubric type since the UserInput type is the same:
```
export type PerseusCategorizerUserInput = PerseusCategorizerRubric
```
Going with still showing the shape of the UserInput object, but could switch if anyone feels strongly about it.

packages/perseus/src/widgets/categorizer/categorizer-validator.test.ts
```diff
@@ -7,17 +7,7 @@ import type {PerseusCategorizerRubric} from "../../validation.types";
 describe("categorizerValidator", () => {
     it("gives points when the answer is correct", () => {
         const rubric: PerseusCategorizerRubric = {
-            items: ["Graph $1$", "Graph $2$"],
```
handeyeco, line 10: It's nice to see how much clearer this makes the types but differentiating (1) what we need for rendering vs (2) what we need from the learner vs (3) what we need to score. Love to see this.

packages/perseus/src/validation.types.ts
handeyeco: This makes sense to me 🤷

handeyeco (approved):
Thanks!

packages/perseus/src/widgets/categorizer/categorizer-validator.test.ts
jeremywiebe: Very much agree! This would be extremely useful even if we weren't doing SSS.

PR #1726, opened 2024-10-04
Refine CS-Program Rubric type

Myranae:
## Summary:
As part of the Server Side Scoring project, this refactors the CS-Program Rubric type to show it is not particularly needed and could most likely be removed. Adds a comment to come back to it with LEMS-2440.

Issue: LEMS-2464

## Test plan:
- Confirm tests pass
- Confirm CS-Program widgets still work locally or via znd if locally doesn't work

### DIFF
--- a/packages/perseus/src/validation.types.ts
+++ b/packages/perseus/src/validation.types.ts
@@ -1,7 +1,6 @@
 import type {Coord} from "./interactive2/types";
 import type {
     PerseusCategorizerWidgetOptions,
-    PerseusCSProgramWidgetOptions,
     PerseusDefinitionWidgetOptions,
     PerseusDropdownWidgetOptions,
     PerseusExplanationWidgetOptions,
@@ -45,7 +44,10 @@ export type PerseusCategorizerUserInput = {
     values: ReadonlyArray<number>;
 };
 
-export type PerseusCSProgramRubric = PerseusCSProgramWidgetOptions;
+// TODO(LEMS-2440): Can possibly be removed during 2440?
+// This is not used for grading at all. The only place it is used is to define
+// Props type in cs-program.tsx, but RenderProps already contains WidgetOptions
+export type PerseusCSProgramRubric = Empty;
 
 export type PerseusCSProgramUserInput = {
     status: UserInputStatus;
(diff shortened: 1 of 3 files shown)


packages/perseus/src/validation.types.ts
```diff
@@ -45,7 +44,11 @@ export type PerseusCategorizerUserInput = {
     values: ReadonlyArray<number>;
 };
 
-export type PerseusCSProgramRubric = PerseusCSProgramWidgetOptions;
+// TODO(LEMS-2440): Can possibly be removed during 2440?
+// This is not used for grading at all. The only place it is used is to define
+// Props type in cs-program.tsx, but RenderProps already contains WidgetOptions
+// and is already included in the Props type.
+export type PerseusCSProgramRubric = Empty;
```
Myranae, line 51: Wanted something here that would not add anything to the Props type in cs-program.tsx. Pretty sure `Empty` works for that, but definitely let me know if that's not true. Kind of a stop gap measure until all the iframe related widgets are reviewed.

packages/perseus/src/widgets/cs-program/cs-program.tsx
```diff
@@ -26,15 +26,11 @@ import type {
...
-    UserInputStatus,
 } from "../../validation.types";
 
 const {updateQueryString} = Util;
 
-type RenderProps = PerseusCSProgramWidgetOptions & {
-    status: UserInputStatus;
-    message: string | null;
-};
+type RenderProps = PerseusCSProgramWidgetOptions & PerseusCSProgramUserInput;
```
Myranae, line 33: Seemed like a small refactor and allows us to know where that object comes from

packages/perseus/src/validation.types.ts
handeyeco: (I know you know, but for others that might see this)

Several widgets are wrappers around complex, nebulous things: CS Program, iFrame, Phet. IMO it's questionable that these really should even be Perseus Widgets at all. But either way, in these instances, it's the things we wrap that determine correctness and we just pass along their "score."

`Empty` works for me. I'd also be fine with `null` or something, but I don't care because we're going to circle back to this in LEMS-2440.

handeyeco (approved):

PR #1730, opened 2024-10-07
Improve Grapher types

Myranae:
## Summary:
As part of the Server Side Scoring project, this refines Grapher's Rubric to be only what is needed for scoring. It also refines the UserInput type to refer to the rubric as the types are mostly the same. In addition, it cleans up some error expects and some related TODOs by adding lots more typing.

Issue: LEMS-2466

## Test plan:
- Confirm all tests pass
- Confirm widget works as intended via Storybook


packages/perseus/src/validation.types.ts
```diff
@@ -82,17 +81,26 @@ export type PerseusGradedGroupRubric = PerseusGradedGroupWidgetOptions;
...
+              type: "exponential";
+              asymptote: [Coord, Coord];
+              coords: [Coord, Coord];
+          }
+        | {type: "linear"; coords: [Coord, Coord]}
+        | {type: "logarithm"; asymptote: [Coord, Coord]; coords: [Coord, Coord]}
+        | {type: "quadratic"; coords: [Coord, Coord]}
+        | {type: "sinusoid"; coords: [Coord, Coord]}
+        | {type: "tangent"; coords: [Coord, Coord]};
+};
```
Myranae, line 97: Came up with a couple other ways to do this. I don't _think_ either of these are preferred, and since it works as is, just wanted to leave it. But if either of the options below work better, would love to know :)

Using an enum _(Thanks @anakaren-rojas for showing me this!)_ -


Using as const with object types -

packages/perseus/src/validation.types.ts
```diff
@@ -82,17 +81,26 @@ export type PerseusGradedGroupRubric = PerseusGradedGroupWidgetOptions;
...
 /**
  * TODO: this is kind of just a guess right now
  * based off of an old comment in grapher
  */
-export type PerseusGrapherUserInput = {
-    type: string;
-    asymptote: ReadonlyArray<Coord>;
-    coords: ReadonlyArray<Coord>;
-};
+export type PerseusGrapherUserInput = PerseusGrapherRubric["correct"];
```
Myranae, line 103: I was thinking this might be enough to remove the TODO for this section. Wanted to get some input before removing it though. Thoughts?

packages/perseus/src/widgets/grapher/grapher-validator.ts
```diff
@@ -7,10 +7,10 @@ import type {
 } from "../../validation.types";
 
 function grapherValidator(
-    state: PerseusGrapherUserInput,
+    userInput: PerseusGrapherUserInput,
```
Myranae, line 10: Was there a reason to keep userInput as state? 🤔 I vaguely remember reverting a change like this, but not can't remember the reasoning.

packages/perseus/src/validation.types.ts
handeyeco, line 97: I'm guessing this is copied from `PerseusGrapherWidgetOptions`, but it feels like maybe this should be a shared type between them.

```suggestion
// perseus-types.ts

// I don't know what the best name would be
type GrapherAnswerState =
        | {type: "absolute_value"; coords: [Coord, Coord]}
        | {
              type: "exponential";
              asymptote: [Coord, Coord];
              coords: [Coord, Coord];
          }
        | {type: "linear"; coords: [Coord, Coord]}
        | {type: "logarithm"; asymptote: [Coord, Coord]; coords: [Coord, Coord]}
        | {type: "quadratic"; coords: [Coord, Coord]}
        | {type: "sinusoid"; coords: [Coord, Coord]}
        | {type: "tangent"; coords: [Coord, Coord]};

export type PerseusGrapherWidgetOptions = {
    availableTypes: ReadonlyArray<
        | "absolute_value"
        | "exponential"
        | "linear"
        | "logarithm"
        | "quadratic"
        | "sinusoid"
        | "tangent"
    >;
    correct: GrapherAnswerState;
    graph: {
      // ...etc...
    }
}
    

// validation.types.ts
export type PerseusGrapherRubric = {
    correct: GrapherAnswerState;
};
```

handeyeco: That way you did it is correct (except that I think we should share a type instead of copying it). We try not to use enums and in TS I haven't noticed much of a difference between `TypeObject.Linear` and `"linear"`; that's more of something you'd see in Java.

packages/perseus/src/widgets/grapher/grapher-validator.ts
handeyeco: “What's in a name? That which we call a rose by any other name would smell just as sweet.”

I prefer `userInput` over `state` personally

packages/perseus/src/widgets/grapher/grapher-validator.ts
```diff
@@ -20,21 +20,36 @@ function grapherValidator(
...
 
     // Get new function handler for grading
-    const grader = functionForType(state.type);
-    const guessCoeffs = grader.getCoefficients(state.coords, state.asymptote);
-    const correctCoeffs = grader.getCoefficients(
-        rubric.correct.coords,
-        // @ts-expect-error - TS(2339) - Property 'asymptote' does not exist on type '{ type: "absolute_value"; coords: [Coord, Coord]; }'
-        rubric.correct.asymptote,
-    );
+    const grader = functionForType(userInput.type);
```
handeyeco, line 31: I was hoping we wouldn't be changing the validation logic so much as describing the validation logic with proper types. This looks correct, but how confident are we? Is there test coverage for this?

packages/perseus/src/widgets/grapher/grapher-validator.ts
```diff
@@ -20,21 +20,36 @@ function grapherValidator(
...
+    let guessCoeffs: number[];
+    if (userInput.type === "logarithm" || userInput.type === "exponential") {
+        guessCoeffs = grader.getCoefficients(
+            userInput.coords,
+            userInput.asymptote,
+        );
+    } else {
+        guessCoeffs = grader.getCoefficients(userInput.coords);
+    }
+    let correctCoeffs: number[];
```
handeyeco, line 41: **nit**
```suggestion
    }
    
    let correctCoeffs: number[];
```

packages/perseus/src/widgets/grapher/grapher-validator.ts
```diff
@@ -20,21 +20,36 @@ function grapherValidator(
...
+    if (
+        rubric.correct.type === "logarithm" ||
+        rubric.correct.type === "exponential"
+    ) {
```
handeyeco, line 45: By this point we know the user input and rubric types are the same. Rather than having two if/elses, could we just have one?

packages/perseus/src/validation.types.ts
handeyeco: Yeah, I think you could remove the TODO

handeyeco (approved):
I like how this is coming and wouldn't lose sleep if we landed as-is, but my asks:
1. Make a shared type for that complex object union rather than copying it
2. I almost prefer reverting the logic change and continue suppressing errors **unless** we feel like we have really good test coverage of that logic (and we feel confident that existing data conforms to the data shape we're testing)

packages/perseus/src/widgets/grapher/util.tsx
```diff
@@ -669,13 +722,15 @@ export const defaultPlotProps = (type: string, graph: any): any => {
     // widget before even reading the question; you can't lose, but you
     // might get a free win.
     const model = functionForType(type);
+    const {defaultAsymptote} =
+        "defaultAsymptote" in model ? model : {defaultAsymptote: null};
```
handeyeco, line 726: Does this work?

```suggestion
    const defaultAsymptote = "defaultAsymptote" in model ? model.defaultAsymptote : null;
```

handeyeco (approved):

packages/perseus/src/widgets/grapher/grapher-types.ts
```diff
@@ -0,0 +1,107 @@
...
+        asymptote: Coords,
+        graph: any,
+    ) => Coord;
+    allowReflectOverAsymptote: boolean;
+    getPropsForCoeffs: (coeffs: ReadonlyArray<number>) => {fn: any};
+    getCoefficients: (
+        coords: Coords,
+        asymptote: Coords,
+    ) => ReadonlyArray<number>;
+};
```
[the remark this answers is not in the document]
Myranae: I actually went down to just having 3 types last night, but it seemed like it lost a lot of the helpful granularity. So I went back to listing out all the different types to keep that. I could totally see creating a shared type for this (I called it asymptotic last night), but I think I like the consistency of just having the base type intersect with the options more specific to each type.

packages/perseus/src/widgets/grapher/grapher-types.ts
```diff
@@ -0,0 +1,107 @@
...
+    areEqual: (
+        coeffs1: ReadonlyArray<number>,
+        coeffs2: ReadonlyArray<number>,
+    ) => boolean;
+    Movable: any;
+};
+
+export type LinearType = BaseType & {
+    getPropsForCoeffs: (coeffs: ReadonlyArray<number>) => {fn: any};
+    getCoefficients: (coords: Coords) => ReadonlyArray<number> | undefined;
```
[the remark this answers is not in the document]
Myranae: I did try that, but it didn't work so well. That's why I ended up having getCoefficients be defined separately on each type. Specifically, I had it on the baseType, but then when I realized it also had to potentially return `undefined`, it caused a bunch more TypeScript errors. I prefer not defining things to have a potential undefined return if they don't actually have that possibility, so I kept the definitions separate so the bare minimum have a return type of `undefined`. I'd feel comfortable putting a TODO to re-evaluate this though; would that work?

packages/perseus/src/widgets/grapher/grapher-validator.ts
```diff
@@ -20,21 +21,35 @@ function grapherValidator(
...
 
     // Get new function handler for grading
-    const grader = functionForType(state.type);
-    const guessCoeffs = grader.getCoefficients(state.coords, state.asymptote);
-    const correctCoeffs = grader.getCoefficients(
-        rubric.correct.coords,
-        // @ts-expect-error - TS(2339) - Property 'asymptote' does not exist on type '{ type: "absolute_value"; coords: [Coord, Coord]; }'
-        rubric.correct.asymptote,
-    );
+    function getCoefficientsFunc(
```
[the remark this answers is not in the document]
Myranae: Renamed and moved out!

packages/perseus/src/widgets/grapher/grapher-types.ts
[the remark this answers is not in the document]
Myranae: Okay, never mind, went ahead and pulled it out.

packages/perseus/src/widgets/grapher/grapher-types.ts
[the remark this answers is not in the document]
Myranae: Also never mind, went ahead and pulled it out.

packages/perseus/src/widgets/grapher/util.tsx
```diff
@@ -479,34 +508,41 @@ const Logarithm = _.extend({}, PlotDefaults, {
...
         const a = coeffs[0],
             b = coeffs[1],
             c = coeffs[2];
         return a * Math.log(b * x + c);
     },
 
-    getEquationString: function (coords, asymptote) {
-        if (!asymptote) {
-            return null;
-        }
```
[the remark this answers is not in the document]
Myranae: I actually thought this was something I added when trying to make types work, so added it back.

packages/perseus/src/widgets/grapher/util.tsx
Myranae: This line is object destructuring, so I think the brackets are needed (model refers to the asymptotic graph objects). Tried your suggested change and TS errors popped up.

packages/perseus/src/widgets/grapher/util.tsx
```diff
@@ -479,34 +511,46 @@ const Logarithm = _.extend({}, PlotDefaults, {
...
-        const c = -inverseCoeffs[2] / inverseCoeffs[0];
-        const b = 1 / inverseCoeffs[0];
-        const a = 1 / inverseCoeffs[1];
-        return [a, b, c];
+        if (inverseCoeffs) {
+            const c = -inverseCoeffs[2] / inverseCoeffs[0];
+            const b = 1 / inverseCoeffs[0];
+            const a = 1 / inverseCoeffs[1];
+            return [a, b, c];
+        }
```
Myranae, line 532: Added this to an if statement because of TS errors stating `inverseCoeffs` could be undefined.

packages/perseus/src/widgets/grapher/util.tsx
handeyeco: That's interesting, when I pull your branch and make the change I don't get errors. I don't see the point of making an object just to pull `null` out of it.

Myranae: Hmm, oh, I think I didn't fully read the change the first time. Updating it :)

packages/perseus/src/perseus-types.ts
```diff
@@ -474,6 +474,56 @@ export type GraphRange = [
     y: [min: number, max: number],
 ];
 
+export type GrapherAnswerTypes =
+    | {
+          type: "absolute_value";
+          coords: [
+              // The vertex
+              Coord, // A point along one line of the absolute value "V" lines
```
jeremywiebe, line 482: **nit**: I _think_ this comment is for the second `Coord` and should be dropped onto a newline.

packages/perseus/src/widgets/grapher/grapher-validator.test.ts
jeremywiebe, line 1: Really cool to see the improvement in Rubric types clarify these tests a bunch! Thank-you!

packages/perseus/src/perseus-types.ts
Myranae: For sure. I just copied it exactly how it was from the previous location. I think I'll be going over the widgets again, so I can adjust it later if needed.

PR #1731, opened 2024-10-07
Input Number Deprecation and Conversion into Numeric Input

SonicScrewdriver:
## Summary:
This PR contains all the work necessary to deprecate our Input Number Widget, and convert it into the Numeric Input widget (sans actually deleting the Input Number Files, which will be handled in LEMS-2411).

The changes include: 

- Using replaceWidget/Editor to change the rendering of Input Number widgets to Numeric Input 
- New conversion util functions to automatically convert the json on the Editor pages for pre-existing Input Numbers in content, as an effort to avoid a backfill. 
- Cleanup of MANY tests across the repo that used Input Number for their test data 
- New PropUpgrades function for Numeric Input that converts Input Number widgetOptions into Numeric Input's format.

NOTE: There is a temporary wrapper div around the Numeric Input component in order to help QA be able to determine that the widget has been converted, as InputNumber/NumericInput are virtually identical otherwise. This will be removed prior to landing the ticket, and after QA testing has completed.  

Issue: LEMS-2408 & LEMS-2409

## Test plan:
- Extensive Manual Testing 
- New Jest Tests 
- QA Testing Round


packages/perseus-editor/src/editor-page.tsx
```diff
@@ -287,7 +296,7 @@ class EditorPage extends React.Component<Props, State> {
                     <ItemEditor
                         ref={this.itemEditor}
                         itemId={this.props.itemId}
-                        question={this.props.question}
+                        question={convertDeprecatedWidgets(this.props.question)}
                         answerArea={this.props.answerArea}
```
SonicScrewdriver, line 300: This could be a spicy move -- but it seems to work with editing, and I can't move the conversion logic any higher.

packages/perseus-editor/src/editor.tsx
```diff
@@ -31,7 +31,7 @@ import type {
     PerseusWidgetsMap,
 } from "@khanacademy/perseus";
 
-// like [[snowman input-number 1]]
+// like [[snowman numeric-input 1]]
 const widgetPlaceholder = "[[\u2603 {id}]]";
```
SonicScrewdriver, line 35: Not a critical change, but we might as well have our example use a widget name that will still exist after this work.

packages/perseus/src/index.ts
```diff
@@ -190,7 +190,8 @@ export type {
     Size,
     CollinearTuple,
     MathFormat,
-    InputNumberWidget, // TODO(jeremy): remove?
+    InputNumberWidget, // Used for usurpation of InputNumberWidget in perseus-editor
+    NumericInputWidget, // Used for usurpation of InputNumberWidget in perseus-editor
     // Widget configuration types
```
SonicScrewdriver, line 195: Note to self: I should be able to fix this if I use the PerseusWidgetMap type -- then we don't need to export these.

packages/perseus/src/widgets.ts
```diff
@@ -115,6 +118,8 @@ export const replaceDeprecatedEditors = () => {
     replaceEditor("sequence", "deprecated-standin");
     replaceEditor("simulator", "deprecated-standin");
     replaceEditor("unit-input", "deprecated-standin");
+
+    replaceEditor("input-number", "numeric-input");
 };
```
SonicScrewdriver, line 123: Note to self: Add comment here explaining the editor page logic so that someone can follow the trail, if this is where they start of their journey

packages/perseus/src/widgets.ts
```diff
@@ -176,9 +181,8 @@ export const getPublicWidgets = (): ReadonlyArray<WidgetExports> => {
     // @ts-expect-error - TS2740 - Type 'Pick<{ [key: string]: Readonly<{ name: string; displayName: string; getWidget?: (() => ComponentType<any>) | undefined; accessible?: boolean | ((props: any) => boolean) | undefined; hidden?: boolean | undefined; ... 10 more ...; widget: ComponentType<...>; }>; }, string>' is missing the following properties from type 'readonly Readonly<{ name: string; displayName: string; getWidget?: (() => ComponentType<any>) | undefined; accessible?: boolean | ((props: any) => boolean) | undefined; hidden?: boolean | undefined; ... 10 more ...; widget: ComponentType<...>; }>[]': length, concat, join, slice, and 18 more.
     return _.pick(
         widgets,
-        // @ts-expect-error - TS2345 - Argument of type '(name: string) => boolean | undefined' is not assignable to parameter of type 'Iteratee<string[], boolean, string>'.
         _.reject(_.keys(widgets), function (name) {
-            return widgets[name].hidden;
+            return widgets[name].hidden || name === "input-number";
         }),
```
SonicScrewdriver, line 186: We don't want to register the input number widget any more, in order to keep it out of the list of available widgets while still allowing us to replace the widget with numeric-input.

packages/perseus/src/widgets/numeric-input/numeric-input.tsx
```diff
@@ -243,7 +243,7 @@ export class NumericInput
         });
 
         return (
-            <div>
+            <div className="Im-just-a-sweet-lil-numeric-input">
```
SonicScrewdriver, line 246: Oops, need to remove this sweet lil' guy

packages/perseus-editor/src/article-editor.tsx
```diff
@@ -116,10 +122,13 @@ export default class ArticleEditor extends React.Component<Props, State> {
...
-        return Array.isArray(this.props.json)
-            ? this.props.json
-            : [this.props.json];
+    _sections(): ReadonlyArray<PerseusRenderer> {
+        const sections = Array.isArray(this.props.json)
+            ? this.props.json.map((section) => {
+                  return convertDeprecatedWidgets(section as PerseusRenderer);
+              })
+            : [convertDeprecatedWidgets(this.props.json as PerseusRenderer)];
+        return sections;
```
SonicScrewdriver, line 131: I'm not a big fan of this having to run so frequently. I wonder if it would be better to be converting the json in the main repo by exporting the util from the editor package.

packages/perseus/src/widgets/numeric-input/numeric-input.tsx
SonicScrewdriver: I'll actually be adding this back temporarily for aiding QA testing, and I will remove it prior to deploy.

packages/perseus-editor/src/article-editor.tsx
```diff
@@ -15,7 +15,12 @@ import SectionControlButton from "./components/section-control-button";
 import Editor from "./editor";
```
SonicScrewdriver, line 15: All changes to this file will be reverted now that we're performing the conversion in Webapp.

packages/perseus-editor/src/util/modernize-widgets-utils.ts
```diff
@@ -0,0 +1,202 @@
...
+
+type WidgetRenameMap = {
+    [oldKey: string]: string;
+};
+
+// This utils in this file are used to modernize our Perseus JSON structure,
+// so that we can convert deprecated widgets to their modern equivalents when
+// content creators use the Editor Page to update content containing these widgets.
+// Currently, we're only converting input-number to numeric-input,
+// but we can add more conversions here in the future.
```
mark-fitzgerald, line 15: Absolutely LOVING the extensive explanations!!! Thank you!

packages/perseus-editor/src/util/modernize-widgets-utils.ts
```diff
@@ -0,0 +1,202 @@
...
+    renameMap: WidgetRenameMap,
+): string => {
+    let newContent = json.content;
+    // Loop through the renameMap and replace all the old keys with the new keys
+    for (const oldKey of Object.keys(renameMap)) {
+        const newKey = renameMap[oldKey];
+        if (newKey) {
+            newContent = newContent.replace(oldKey, newKey);
+        }
+    }
```
mark-fitzgerald, line 55: NIT: This could be a `reduce()` operation:
```
return Object.keys(renameMap).reduce((newContent, oldKey) => { ... })
```

packages/perseus-editor/src/article-editor.tsx
```diff
@@ -120,10 +133,13 @@ export default class ArticleEditor extends React.Component<Props, State> {
...
-        return Array.isArray(this.props.json)
-            ? this.props.json
-            : [this.props.json];
+    _sections(): PerseusRenderer[] {
+        const sections = Array.isArray(this.state.json)
+            ? this.state.json
+            : [this.state.json];
+        return sections.filter(
+            (section): section is PerseusRenderer => section !== null,
+        );
```
mark-fitzgerald, line 142: Nit: Perhaps this could use nullish coalescence?
```
    _sections(): PerseusRenderer[] {
        return Array.isArray(this.state.json)
            ? (this.state.json ?? [])
            : [this.state.json];
    };
```

packages/perseus-editor/src/article-editor.tsx
```diff
@@ -350,7 +362,7 @@ export default class ArticleEditor extends React.Component<Props, State> {
         const newSection =
             i >= 0
                 ? {
-                      widgets: sections[i].widgets,
+                      widgets: sections![i].widgets,
```
mark-fitzgerald, line 365: Shouldn't this be a question mark instead of an exclamation point?

packages/perseus-editor/src/util/modernize-widgets-utils.ts
```diff
@@ -0,0 +1,205 @@
...
+
+// This utils in this file are used to modernize our Perseus JSON structure,
+// so that we can convert deprecated widgets to their modern equivalents when
+// content creators use the Editor Page to update content containing these widgets.
+// Currently, we're only converting input-number to numeric-input,
+// but we can add more conversions here in the future.
+// Modernize the json content of a PerseusRenderer object
+// by converting deprecated widgets to their modern equivalents
+export const convertDeprecatedWidgets = (
+    json: PerseusRenderer,
```
mark-fitzgerald, line 19: Since `json` can be null/undefined, shouldn't this technically be:
```
json: PerseusRenderer | null | undefined,
```

packages/perseus-editor/src/editor-page.tsx
```diff
@@ -83,15 +86,21 @@ class EditorPage extends React.Component<Props, State> {
     constructor(props: Props) {
         super(props);
 
+        // Convert any widgets that need to be converted to newer widget types
+        let convertedQuestionJson: PerseusRenderer = props.question;
+        if (props.question) {
+            convertedQuestionJson = convertDeprecatedWidgets(props.question);
+        }
```
mark-fitzgerald, line 93: Since `convertDeprecatedWidgets` takes null/undefined as a parameter option, couldn't this be simplified to:
```
const convertedQuestionJson = convertDeprecatedWidgets(props.question);
```

packages/perseus-editor/src/editor-page.tsx
```diff
@@ -83,15 +86,21 @@ class EditorPage extends React.Component<Props, State> {
...
             // @ts-expect-error - TS2322 - Type 'Pick<Readonly<Props> & Readonly<{ children?: ReactNode; }>, "hints" | "question" | "answerArea" | "itemDataVersion">' is not assignable to type 'PerseusJson'.
-            json: _.pick(
-                this.props,
-                "question",
-                "answerArea",
-                "hints",
-                "itemDataVersion",
-            ),
+            json: json,
+            question: json.question,
```
mark-fitzgerald, line 103: I'm not seeing where `this.state.question` is used elsewhere. Can you point me to where it is used? I'm also wonder if it needs to be updated in `changeJSON`.

packages/perseus-editor/src/util/modernize-widgets-utils.ts
```diff
@@ -0,0 +1,205 @@
...
+): PerseusRenderer => {
+    // If there's no json, or no input-number widgets, then we can return the json as is
```
mark-fitzgerald, line 21: Nit: Maybe generalize the explanation to "If there's no json, or no widgets to convert, then we can return the json as is"

packages/perseus-editor/src/util/modernize-widgets-utils.ts
```diff
@@ -0,0 +1,205 @@
...
+    return modernJson;
+};
+
+const conversionRequired = (json: PerseusRenderer): boolean => {
+    const inputNumberRegex = /input-number \d+/; // We can add more regexes here in the future
+
+    // Check the content string for any top-level input-number widgets
+    if (inputNumberRegex.test(json.content)) {
+        return true;
+    }
```
mark-fitzgerald, line 39: This might be over-engineering, but to aid in possible future conversions, what do you think about setting things up this way:
```
const widgetRegExes = [/input-number \d+/]; // Add your widget expression here

if (widgetRegExes.some(regex => regex.test(json.content))) {
    return true;
}
```

packages/perseus-editor/src/util/modernize-widgets-utils.ts
```diff
@@ -0,0 +1,205 @@
...
+
+    // If there's no input-numbers top-level, then check the widgets for any nested widgets
```
mark-fitzgerald, line 41: Nit: Generic comment here, too?
"If there's no deprecated widget in the top-level, then check for any nested widgets"

packages/perseus-editor/src/util/modernize-widgets-utils.ts
```diff
@@ -0,0 +1,205 @@
...
+const convertDeprecatedWidgetsInContent = (
+    json: PerseusRenderer,
+    renameMap: WidgetRenameMap,
+): string => {
+    return Object.keys(renameMap).reduce((newContent, oldKey) => {
+        const newKey = renameMap[oldKey];
+        if (newKey) {
+            return newContent.replace(oldKey, newKey);
+        }
+        return newContent;
```
mark-fitzgerald, line 65: Nit: Ternary operator?
```
return newKey ? newContent.replace(oldKey, newKey) : newContent;
```

packages/perseus-editor/src/util/modernize-widgets-utils.ts
```diff
@@ -0,0 +1,205 @@
...
+    }, json.content);
+};
+
+// Methods to convert input-number widgets to numeric-input widgets
+
```
mark-fitzgerald, line 70: What do you think about pulling out widget-specific conversion functions to their own files (i.e. `util/deprecations/input-number.ts`)? This would help localize conversion logic, and would prevent bloat in this file as we add more deprecated widgets.

packages/perseus-editor/src/util/modernize-widgets-utils.ts
```diff
@@ -0,0 +1,205 @@
...
+const inputNumberToNumericInput = (json: PerseusRenderer) => {
+    // First we need to create a map of the old input-number keys to the new numeric-input keys
+    // so that we can ensure we update the content and widgets accordingly
+    const renameMap = getInputNumberRenameMap(json);
+
+    // Then we can use this to update the JSON
+    const modernizedJson = convertInputNumberJson(json, renameMap);
+    return modernizedJson;
```
mark-fitzgerald, line 78: Can we just return the function result directly?

packages/perseus-editor/src/util/modernize-widgets-utils.ts
```diff
@@ -0,0 +1,205 @@
...
+                type: "numeric-input",
+            };
+            const newWidgetName = renameMap[key];
+            // Create the new key entry
+            widgets[newWidgetName] = upgradedWidget;
+
+            // We need to delete the old widget key
+            delete widgets[key];
+        }
+    }
```
mark-fitzgerald, line 163: Organization and accompanying explanations is wonderful!! Thank you!

packages/perseus-editor/src/util/modernize-widgets-utils.ts
```diff
@@ -0,0 +1,205 @@
...
+    const allContentStrings = json.content;
+
+    // Loop through the content strings to get all the input-number widgets
+    const renameMap: WidgetRenameMap = {};
+    const inputNumberMatches: string[] = [
+        ...(allContentStrings.match(inputNumberRegex) || []),
+    ];
+
+    // We want to count any pre-existing numeric-input widgets
+    // so that we can start the new ids at the next number
```
mark-fitzgerald, line 182: Great catch!!!

packages/perseus-editor/src/util/modernize-widgets-utils.ts
```diff
@@ -0,0 +1,205 @@
...
+    ];
+    let currentNumericCount = 1;
+    if (numericMatches.length > 0) {
+        for (const match of numericMatches) {
+            const id = parseInt(match.split(" ")[1], 10);
+            if (id >= currentNumericCount) {
+                currentNumericCount = id + 1; // We want to start at the next highest number
+            }
+        }
+    }
```
mark-fitzgerald, line 194: Nit: Maybe use `reduce`?
```
const currentNumericCount = numericMatches.reduce((count, match) => {
    const id = parseInt(match.split(" ")[1], 10);
    if (id >= currentNumericCount) {
        currentNumericCount = id + 1; // We want to start at the next highest number
    }
}, 1);
```

packages/perseus-editor/src/util/modernize-widgets-utils.ts
```diff
@@ -0,0 +1,205 @@
...
+
+    // Now that we have all the necessary information, we can create the renameMap
+    for (const match of inputNumberMatches) {
+        const oldKey = match;
+        const newKey = `numeric-input ${currentNumericCount}`;
+        renameMap[oldKey] = newKey;
+        currentNumericCount++;
+    }
```
mark-fitzgerald, line 202: Nit: Definitely a style thing...
```
inputNumberMatches.forEach((match, index) => {
    renameMap[match] = `numeric-input ${currentNumericCount + index}`;
});
```

I also appreciate the explicit-ness of the variable names and the clear steps, so definitely up to which appeals to you more.

packages/perseus-editor/src/article-editor.tsx
SonicScrewdriver: I'm fixing this one by getting rid if the silly local type that's faking the PerseusRenderer.

packages/perseus-editor/src/editor-page.tsx
SonicScrewdriver: Hmm open to feedback on this one! 

The approach I've taken has been to ensure we're only ever calling convertQuestionJson if we already have the data. Otherwise we need a TON of null checks across our code. I've updated convertDeprecatedWidgets to no longer return null

packages/perseus-editor/src/editor-page.tsx
SonicScrewdriver: Whoops, thank you for catching this! It's a vestige of an old approach to fixing the json. Removed!

packages/perseus-editor/src/editor-page.tsx
```diff
@@ -83,15 +86,20 @@ class EditorPage extends React.Component<Props, State> {
...
+        // Convert any widgets that need to be converted to newer widget types
+        let convertedQuestionJson: PerseusRenderer = props.question;
+        if (props.question) {
+            convertedQuestionJson = convertDeprecatedWidgets(props.question);
+        }
+
+        const json = {
+            ..._.pick(this.props, "answerArea", "hints", "itemDataVersion"),
+            question: convertedQuestionJson,
+        };
```
mark-fitzgerald, line 98: In an effort to remove the underscore library, could this be changed to:
```
const json = {
    answerArea: this.props.answerArea,
    hints: this.props.hints,
    itemDataVersion: this.props.itemDataVersion,
    question: convertedQuestionJson,
}
```

SonicScrewdriver: Yeah good call, let's do it!

mark-fitzgerald (approved):
Super well-organized!! Thanks for all the documentation, too.

PR #1741, opened 2024-10-11
refine uses of reviewModeRubric

handeyeco:
## Summary:
move what I can to reviewMode

Issue: LEMS-2437

## Test plan:

### DIFF
--- a/packages/perseus-editor/src/widgets/input-number-editor.tsx
+++ b/packages/perseus-editor/src/widgets/input-number-editor.tsx
@@ -4,8 +4,10 @@ import _ from "underscore";
 
 import BlurInput from "../components/blur-input";
 
-import type {ParsedValue, InputNumber} from "@khanacademy/perseus";
-import type {PropsFor} from "@khanacademy/wonder-blocks-core";
+import type {
+    ParsedValue,
+    PerseusInputNumberWidgetOptions,
+} from "@khanacademy/perseus";
 
 const {InfoTip} = components;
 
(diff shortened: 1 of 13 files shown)


packages/perseus/src/types.ts
```diff
@@ -673,7 +673,8 @@ export type WidgetProps<
     onFocus: (blurPath: FocusPath) => void;
     onBlur: (blurPath: FocusPath) => void;
     findWidgets: (criterion: FilterCriterion) => ReadonlyArray<Widget>;
-    reviewModeRubric: Rubric;
+    reviewModeRubric?: Rubric | null | undefined;
+    reviewMode: boolean;
```
handeyeco, line 677: As far as I could tell, the only widget that was really using the contents of `reviewModeRubric` was Radio. The other widgets were just using its existence to tell if they were in review mode or not. Then if you looked at how `reviewModeRubric` was set, it kind of seemed like needless misdirection:

``` TSX
        // renderer.tsx checks if we're in `reviewMode` then sets `reviewModeRubric`
        const reviewModeRubric =
            this.props.reviewMode && widgetInfo ? widgetInfo.options : null;

        // passage.tsx for example then converted that object into a bool
        const editable = !this.props.reviewModeRubric;
```

So the crux of this PR is: why don't we just use `reviewMode` everywhere that's using coercing `reviewModeRubric` to a flag so that it's clearer what's actually using `reviewModeRubric` when we start work on LEMS-2437.

packages/perseus/src/types.ts
```diff
@@ -673,7 +673,8 @@ export type WidgetProps<
     onFocus: (blurPath: FocusPath) => void;
     onBlur: (blurPath: FocusPath) => void;
     findWidgets: (criterion: FilterCriterion) => ReadonlyArray<Widget>;
-    reviewModeRubric: Rubric;
```
handeyeco, line 676: I think this was mistyped.

``` TS
        const reviewModeRubric =
            this.props.reviewMode && widgetInfo ? widgetInfo.options : null;
```

packages/perseus-editor/src/widgets/input-number-editor.tsx
```diff
@@ -46,14 +48,12 @@ const answerTypes = {
 
 type Props = {
     value: number;
-    simplify: PropsFor<typeof InputNumber.widget>["simplify"];
-    size: PropsFor<typeof InputNumber.widget>["size"];
-    inexact: PropsFor<typeof InputNumber.widget>["reviewModeRubric"]["inexact"];
```
handeyeco, line 51: TS was exploding since `reviewModeRubric` could have been null/undefined. This might be an unpopular opinion, but I'm not a big fan of the `PropsFor<typeof SomeReactComponent>` pattern anyway. The editor exists to set widget options, so I feel like this is more correct anyway?

jeremywiebe: I actually agree.  I think the widget options types should be the glue between the editors and their widgets. I think we've gone through iterations of understanding Perseus code and so this might have been me doing this even (probably because the widget option type wasn't exported (yet)).

jeremywiebe (approved):
Thanks. I'm in favour!

jeremywiebe (approved):
This is a great refactoring (and simplification)! Thank-you.

PR #1744, opened 2024-10-11
Some small cleanup tweaks

handeyeco:
## Summary:
Some small changes, code pruning and typing.

### DIFF
--- a/packages/perseus-editor/src/editor-page.tsx
+++ b/packages/perseus-editor/src/editor-page.tsx
@@ -17,7 +17,6 @@ import type {
     Version,
     PerseusItem,
 } from "@khanacademy/perseus";
-import type {KEScore} from "@khanacademy/perseus-core";
 
 const {HUD} = components;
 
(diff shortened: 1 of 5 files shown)


Myranae (approved):
Looks good! 

So `scorePreview` wasn't being used anymore? Do we know where it was used in the past? Just want to make sure this change won't affect anything in webapp.

handeyeco:
@Myranae As far as I can tell, it's not being used: https://github.com/search?q=org%3AKhan%20scorePreview&type=code

Not sure what it was for 🤷

packages/perseus-editor/src/editor-page.tsx
```diff
@@ -219,14 +218,6 @@ class EditorPage extends React.Component<Props, State> {
         this.props.onChange(newJson);
     };
 
-    scorePreview(): KEScore | null | undefined {
```
jeremywiebe, line 222: `git` history says we [used to use it](https://github.com/khan/webapp/commit/439f898e2019fdb64218f5329675dc3557fafcfd#diff-39dc4b00ccacbe995d660158ef60215369e7e2614fc67b57b4794fec1398769aL81) for demo pages in webapp... that are long gone!

PR #1746, opened 2024-10-12
[EXPERIMENT] Create a recursive descent version of KAS' parser

kevinb-khan:
## Summary:
The reason for creating this alternate version of the KAS parse is that I want to create a parser that outputs a MathBlocks AST for my hackathon project.  This project is heavily based on MathBlocks which is a suite of node modules I've developed as a side project to facilitate the development of interact math applications on the web.   I thought that it would be easier to create the parser for my hackathon project if I first had a recursive descent parser (as opposed to the current JISON-based parser).  It's much hard to see what's going on with the JISON parser because parts of code the make up the parser are in different files.  Also, the JISON parser doesn't produce type-safe code.

In the future, we may want replace KAS' current parser with this one.  I think it'll make maintenance a bit easier since it uses plain old TypeScript code which means that maintainers don't have to understand JISON.  The parser is implemented as a recursive descent parser which is one of the simplest types of parsers to write.  Because it uses TypeScript it's also typesafe which also makes it easier to work on.  The parser itself is also fewer lines 577 (shipped) vs 814 (shipped) + 214 (dev-only).

If we do want to adopt this parser we'll likely also want to convert the unit parser to be a recursive descent parser as well.  This would allow us to remove JISON from the code base completely (JISON is a dev dependency so this wouldn't have too much impact on bundle size).  Additionally, converting nodes.js to TypeScript and all of the node types to be ES6 classes would also help with maintainability, but would not be a requirement for adopting the new parser(s).

Issue: None

TODO:
- [ ] handle custom decimal separator

## Test plan:
- yarn test packages/kas/src

### DIFF
--- a/packages/kas/src/lexer.ts
+++ b/packages/kas/src/lexer.ts
@@ -129,11 +129,6 @@ const preparedRules: PreparedRule[] = rules.map((rule) => {
     return [regex, rule[1]];
 });
 
-// TODO:
-// - try each of the rules in order
-// - pick the longest one and return token
-// - update the input string so that we match the next token
-
 const next = (
     input: string,
     preparedRules: PreparedRule[],
--- a/packages/kas/src/lexer.ts
+++ b/packages/kas/src/lexer.ts
@@ -20,10 +20,10 @@ export type Token =
     | {kind: "RIGHT|"}
     | {kind: "SIGN"; value: string} // (in)equality
     | {kind: "FRAC"}
-    | {kind: "sqrt"}
-    | {kind: "abs"}
-    | {kind: "ln"}
-    | {kind: "log"}
+    | {kind: "SQRT"}
+    | {kind: "ABS"}
+    | {kind: "LN"}
+    | {kind: "LOG"}
     | {kind: "TRIG"; value: string}
     | {kind: "TRIGINV"; value: string}
     | {kind: "CONST"; value: string}
@@ -81,10 +81,10 @@ const rules: Rule[] = [
     ["\u2265", {kind: "SIGN", value: ">="}], // ge
     ["\\\\frac", {kind: "FRAC"}],
     ["\\\\dfrac", {kind: "FRAC"}],
-    ["sqrt|\\\\sqrt", {kind: "sqrt"}],
-    ["abs|\\\\abs", {kind: "abs"}],
-    ["ln|\\\\ln", {kind: "ln"}],
-    ["log|\\\\log", {kind: "log"}],
+    ["sqrt|\\\\sqrt", {kind: "SQRT"}],
+    ["abs|\\\\abs", {kind: "ABS"}],
+    ["ln|\\\\ln", {kind: "LN"}],
+    ["log|\\\\log", {kind: "LOG"}],
     ["sin|cos|tan", (value) => ({kind: "TRIG", value})],
     ["csc|sec|cot", (value) => ({kind: "TRIG", value})],
     ["sinh|cosh|tanh", (value) => ({kind: "TRIG", value})],
--- a/packages/kas/src/lexer.ts
+++ b/packages/kas/src/lexer.ts
@@ -1,0 +1,192 @@
+export type Token =
+    | {kind: "WHITESPACE"}
+    | {kind: "INT"; value: string}
+    | {kind: "FLOAT"; value: string}
+    | {kind: "^"}
+    | {kind: "*"}
+    | {kind: "/"}
+    | {kind: "-"}
+    | {kind: "+"}
+    | {kind: "("}
+    | {kind: ")"}
+    | {kind: "["}
+    | {kind: "]"}
+    | {kind: "{"}
+    | {kind: "}"}
+    | {kind: "|"}
+    | {kind: "!"} // unused
+    | {kind: "_"}
+    | {kind: "LEFT|"}
+    | {kind: "RIGHT|"}
+    | {kind: "SIGN"; value: string} // (in)equality
+    | {kind: "FRAC"}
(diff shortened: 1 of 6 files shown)


packages/kas/src/lexer.ts
```diff
@@ -0,0 +1,192 @@
...
+    ["\u03C6", {kind: "VAR", value: "phi"}], // phi
+    ["\\\\phi", {kind: "VAR", value: "phi"}],
+    ["[a-zA-Z]", (value) => ({kind: "VAR", value})], // TODO: post-process
+    [".", {kind: "INVALID"}],
+];
+
+type PreparedRule = [RegExp, Token | ((match: string) => Token)];
+
+const preparedRules: PreparedRule[] = rules.map((rule) => {
+    const regex = new RegExp("^(?:" + rule[0] + ")");
```
jeremywiebe, line 128: So each rule becomes a regex that's anchored to the beginning of the input string and is non-capturing. Am I reading this correctly?

kevinb-khan: Yes, that's correct.  It's a simplified version of https://github.com/zaach/jison-lex/blob/master/regexp-lexer.js#L10-L68 which is what `jison` uses internally for its lever.

kevinb-khan: I don't think the non-capturing group part is really necessary since we only ever look at the whole string that's matched. 🤷‍♂️

packages/kas/src/lexer.ts
```diff
@@ -0,0 +1,192 @@
...
+    return [regex, rule[1]];
+});
+
+// TODO:
+// - try each of the rules in order
+// - pick the longest one and return token
+// - update the input string so that we match the next token
```
kevinb-khan, line 135: This TODO list can be removed now.

packages/kas/src/parser.ts
```diff
@@ -0,0 +1,387 @@
...
+    Log,
+    Mul,
+    Pow,
+    Trig,
+    Var,
+} from "./nodes";
+
+type Expr = any;
+
+export class Parser {
```
kevinb-khan, line 19: TODO: Add some comments at the top here about this being a recursive descent parser and how the order that the methods are called determines the precedence of the operators.  Include a list of the current operators and their relative precedences.

packages/kas/src/parser.ts
```diff
@@ -0,0 +1,387 @@
...
+        }
+    }
+
+    power() {
+        const left = this.primitive();
+        const op = this.peek();
+        if (op.kind === "^") {
+            this.consume(); // ^
+            const right = this.negative();
+            return new Pow(left, right);
```
kevinb-khan, line 214: TODO: add a comment here about this being right associative and why that's the case.

packages/kas/src/parser.ts
```diff
@@ -0,0 +1,387 @@
...
+                        const arg = this.additive();
+                        this.expect("}");
+                        return Pow.sqrt(arg);
+                    }
+                    case "[": {
+                        const index = this.additive();
+                        this.expect("]");
+                        this.expect("{");
+                        const arg = this.additive();
+                        this.expect("}");
```
kevinb-khan, line 307: TODO: encapsulate this pattern in a helper function called `between()`.  Calling it will look something like this:
```ts
this.between("{", "additive", "}");
```

kevinb-khan:
This PR resulted in a minimal (+14 B) change in the bundle size because the new parser wasn't being included in bundle.

jeremywiebe (commented):
I have a few thoughts.  I've also asked @jaredly to take a look as I get out of my depth when we get to the parser and how it handles the token stream. :)

packages/kas/src/parser.ts
```diff
@@ -0,0 +1,445 @@
...
+        return next.value;
+    }
+
+    parse() {
+        return this.equation();
+    }
+
+    equation(): Expr {
+        if (this.peek().kind === "EOF") {
+            return new Add([]);
```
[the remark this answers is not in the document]
kevinb-khan: I don't know why we do this either, but this is what was in the original code.

packages/kas/src/parser.ts
```diff
@@ -0,0 +1,445 @@
...
+        this.consume();
+    }
+
+    // Checks if the next token's kind is the specified `tokenKind` and
+    // consumes the token and return the token's value.  Otherwise, it
+    // throws an error.
+    //
+    // This method can only be called with `tokenKind`s for tokens that
+    // have a `value` property.  See lexer.ts.
+    expectValue<Kind extends "FUNC" | "VAR" | "TRIG">(tokenKind: Kind): string {
```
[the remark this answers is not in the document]
kevinb-khan: There's probably a fancy way to compute this type from the `Token` type.  I'll give that a try.

packages/kas/src/lexer.ts
```diff
@@ -0,0 +1,192 @@
...
+                token = rule[1];
+            }
+        }
+    }
+
+    if (currentMatch === "") {
+        throw new Error(`No match for ${input}`);
+    }
+
+    return [token, input.slice(currentMatch.length)];
```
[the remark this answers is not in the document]
kevinb-khan: I don't think it would that much additional complexity.  I did it this way because that's what jison-lex does, see https://github.com/zaach/jison-lex/blob/master/regexp-lexer.js#L330.  I'll try keeping track of the current index instead, but to do so I think restructuring the code to use a class for the lexer will help.

kevinb-khan:
I'll reopen this in the future once I've written up an ADR for this.

PR #1747, opened 2024-10-13
Update KAS nodes.js to use ES6 classes

kevinb-khan:
## Summary:
I had tried this previously but failed because I tried to convert all of the classes to ES6 classes at once.  This time around I converted them one at a time and made sure that all of the tests were passing after each class was converted.  In the next PR I will convert this file to TypeScript.

Issue: None

## Test plan:
- yarn test packages/kas/src

PR #1748, opened 2024-10-13
Convert nodes.js to TypeScript

kevinb-khan:
## Summary:
After updating nodes.js to ues ES6 classes in #1747, the move to TypeScript became much more feasible.  While this PR mostly just adds types in a bunch of places, there are some changes to the code itself in order to minimize the number of casts and @ts-expect-errors that were require.  I've tried to keep these kinds of changes to a minimum.

I ended up converting a bunch of function expressions to arrow functions.  Hopefully that doesn't add too much additional overhead when reviewing.  There's still more cleanup that could be done (adding more missing types, converting all `var`s to either `const` or `let`, etc.), but I'm going to leave that for a future PR.

Issue: None

## Test plan:
- yarn test packages/kas/src


packages/kas/src/nodes.ts
```diff
@@ -2817,49 +2890,52 @@ export class Eq extends Expr {
...
+                  Mul.handleDivide(expr.terms[1], variable),
+              ];
 
         return Mul.handleDivide(a, b).simplify();
     }
 }
 
 /* abstract symbol node */
-class Symbol extends Expr {
+class Sym extends Expr {
```
kevinb-khan, line 2919: I renamed this to `Sym` so that I could use `Symbol` for the prefixes types down below.

packages/kas/src/nodes.ts
```diff
@@ -2817,49 +2890,52 @@ export class Eq extends Expr {
...
     needsExplicitMul() {
         return false;
     }
 
-    findGCD(factor) {
-        if (factor instanceof Symbol || factor instanceof Num) {
-            return this.equals(factor) ? this : Num.One;
+    findGCD(factor: Readonly<Expr>): Expr {
```
kevinb-khan, line 2924: There were a few places that I made params `Readonly<>`.  This was an attempt to fix some issues with TS thinking that subclasses weren't assignable to a superclass.  This change probably isn't necessary so I'm going to try removing.  In general though, I think making things `Readonly<>` by default I preferable.  Maybe that could happening in a more comprehensive way in a future PR.

packages/kas/src/nodes.ts
```diff
@@ -3002,31 +3083,40 @@ export class Const extends Symbol {
         return this;
     }
 
-    eval(vars, options) {
+    eval(vars?, options?): number {
```
kevinb-khan, line 3086: `vars` and `options` should probably be given real types.

packages/kas/src/nodes.ts
```diff
@@ -69,22 +54,82 @@ var randomFloat = function (min, max) {
...
 
     // an array of the arguments to this node's immediate constructor
-    args() {
-        return abstract();
+    args(): (number | string | Expr)[] {
+        throw new Error(
+            "Abstract method - must override for expr: " +
+                // eslint-disable-next-line @babel/no-invalid-this
+                this.print(),
+        );
```
kevinb-khan, line 132: Instead of throwing in abstract methods, I think we can use TypeScript to mark the method as abstract, e.g.
```ts
abstract args(): (number | string | Expr)[];
```

packages/kas/src/nodes.ts
```diff
@@ -3485,19 +3612,15 @@ export class Float extends Num {
...
-Num.Zero = new Int(0);
-Num.One = new Int(1);
-Num.Ten = new Int(10);
-
-// set identities here
-Add.prototype.identity = Num.Zero;
-Mul.prototype.identity = Num.One;
+const NumZero = new Int(0);
+const NumOne = new Int(1);
+const NumTen = new Int(10);
```
kevinb-khan, line 3623: This can't be added to the `Num` class as static properties because `Int` and `Rational` depend on `Num` being defined.  We could probably drop the `Num` prefix.  Then we could export `Zero` and `One` directly here instead of down below.

packages/kas/src/nodes.ts
kevinb-khan: Even though we could do something other than using `Symbol` for the prefix types, I'd like avoid shadowing built-ins like this.

packages/kas/src/nodes.ts
```diff
@@ -69,26 +54,77 @@ var randomFloat = function (min, max) {
...
+        if (iteratee(elem, key, ctx)) {
+            a.push(elem);
+        } else {
+            b.push(elem);
+        }
+    });
+    return [a, b];
+}
+
+function isExpr(arg: string | Expr): arg is Expr {
```
handeyeco, line 76: Not a blocker, just a thought: would it be worth having these as like static methods? Like `Expr.isInstance`? Then we could make it abstract and force children to implement it; that way each variation would have a builtin predicate? idk, maybe more trouble than its worth, I just noticed the `expr instanceof Mul` below which made me think of it.

packages/kas/src/nodes.ts
```diff
@@ -69,26 +54,77 @@ var randomFloat = function (min, max) {
...
         this.hints = {
             parens: false,
         };
     }
 
     // this node's immediate constructor
-    func() {
-        abstract();
-    }
+    abstract func: {new (...args: any[]): any; name: string};
```
handeyeco, line 121: I'm having a hard time deciphering this line. Would it be worth a comment explanation or something?

packages/kas/src/nodes.ts
```diff
@@ -98,25 +134,43 @@ class Expr {
     }
 
     // an abstraction for chainable, bottom-up recursion
-    recurse(method, ...passed) {
+    recurse(method: string, ...passed: any[]): this {
```
handeyeco, line 137: I see a lot of `any`s, is that because we don't understand the types or because inheritance dictates what types will be used by different implementations?

handeyeco (commented):
I'm going to have to circle back to this tomorrow. Just some initial thoughts for now.

packages/kas/src/nodes.ts
kevinb-khan: I'm not sure if type predicates can be generic.  I'll have to experiment with that.

packages/kas/src/nodes.ts
kevinb-khan: Will do.  The `new` part of the type let's TypeScript know that `func` should be a class.

packages/kas/src/nodes.ts
kevinb-khan: This method is highly dynamic which makes typing it correctly difficult. 😞  I think we may be able to make this better using method overloads, but I don't want to block these changes on that.  I'll add a TODO here to tackle this later.  The return type though at least gives us some amount of type safety.

packages/kas/src/nodes.ts
```diff
@@ -162,65 +212,54 @@ class Expr {
 
     // returns the name of this expression's constructor as a string
     // only used for testing and debugging (the ugly regex is for IE8)
```
handeyeco, line 214: I think we could get rid of the reference to the regex now

packages/kas/src/nodes.ts
```diff
@@ -98,25 +134,43 @@ class Expr {
     }
 
     // an abstraction for chainable, bottom-up recursion
-    recurse(method, ...passed) {
+    recurse(method: string, ...passed: any[]): this {
         var args = _.map(this.args(), function (arg) {
-            return _.isString(arg) ? arg : arg[method].apply(arg, passed);
+            return _.isString(arg) || _.isNumber(arg)
```
handeyeco, line 139: Could you talk a little about the switch from `isString` to `isString || isNumber`? I see that in a couple of places. For the most part we're just changing types, but this is changing logic right?

packages/kas/src/nodes.ts
```diff
@@ -162,65 +212,54 @@ class Expr {
...
 
     // canonically reorders all commutative elements
-    normalize() {
+    normalize(): Expr {
         return this.recurse("normalize");
     }
 
     // expands the expression
-    expand() {
+    expand(options?): Expr {
```
handeyeco, line 242: Since `options` isn't being used, can it be omitted?

packages/kas/src/nodes.ts
```diff
@@ -162,65 +212,54 @@ class Expr {
...
     // expand and collect until the expression no longer changes
-    simplify(options) {
-        options = _.extend(
-            {
-                once: false,
-            },
-            options,
-        );
-
+    simplify(options: Options = {once: false}) {
```
handeyeco, line 262: This seems like a logic change, was it intentional? In the original one, if you passed `simplify({hello: "world"})` options would be `{once: false, hello: "world"}`; in the updated one it would be `{hello: "world"}`. I'm sure you checked, just wanted to be double sure.

packages/kas/src/nodes.ts
```diff
@@ -293,7 +330,7 @@ class Expr {
 
     // raise this expression to a given exponent
     // most useful for eventually implementing i^3 = -i, etc.
-    raiseToThe(exp) {
+    raiseToThe(exp: Expr, options?: {preciseFloats?: boolean}): Expr {
```
handeyeco, line 333: Same here: do we need `options` if it's never used? Or do we need it in order to tell TS that this is implementing an abstract class or something?

packages/kas/src/nodes.ts
```diff
@@ -310,12 +347,12 @@ class Expr {
     }
 
     // whether this node needs an explicit multiplication sign if following a Num
-    needsExplicitMul() {
-        return this.args()[0].needsExplicitMul();
+    needsExplicitMul(): boolean {
+        return this.exprArgs()[0].needsExplicitMul();
```
handeyeco, line 351: Is this essentially the same, just more explicit?

packages/kas/src/nodes.ts
```diff
@@ -874,22 +904,24 @@ export class Mul extends Seq {
...
-            _.map(numbers, function (term) {
-                var hasDenom =
-                    term instanceof Rational && !(term instanceof Int);
+            _.map(numbers, (term) => {
                 var shouldPushDown =
                     !term.hints.fraction || inverses.length > 0;
-                if (hasDenom && shouldPushDown) {
+                if (
+                    term instanceof Rational &&
+                    !(term instanceof Int) &&
```
handeyeco, line 920: **nit:** I kind of like having `hasDenom` as I think it makes intent clearer. Wouldn't lose sleep either way though.

packages/kas/src/nodes.ts
```diff
@@ -949,153 +981,145 @@ export class Mul extends Seq {
...
             return term instanceof Num ? term.abs() : term.strip();
         });
         return new Mul(terms).flatten();
     }
 
     // expand numerator and denominator separately
-    expand() {
-        var isAdd = function (term) {
+    expand(): Expr {
+        const isAdd = function (term: Expr): term is Add {
```
handeyeco, line 993: `x instanceof Add` pops up a few times in the code. I wonder if it would be beneficial to make these predicates global and shared.

packages/kas/src/nodes.ts
```diff
@@ -949,153 +981,145 @@ export class Mul extends Seq {
...
         var factored = this.recurse("factor", options).flatten();
         if (!(factored instanceof Mul)) {
             return factored;
         }
 
         // Combine any factored out Rationals into one, but don't collect
-        var grouped = _.groupBy(factored.terms, function (term) {
-            return term instanceof Rational;
+        var grouped = _.groupBy(factored.terms, (term) => {
+            return term instanceof Rational ? "true" : "false";
```
handeyeco, line 1065: Why "true"/"false" vs true/false?

EDIT: oh I see, `groupBy` is using the strings as keys in a return object. I think this is confusing, but your solution (`grouped.true`) is less confusing than the original (`grouped[true]`). I'm just wondering if we should have true/"true"/false/"false" as property names at all.

packages/kas/src/nodes.ts
```diff
@@ -1203,29 +1227,30 @@ export class Mul extends Seq {
...
         var partitioned = this.partition();
-        var numbers = partitioned[0].terms;
+        // `partition` splits the terms into two Seqs - one containing
+        // only Nums and the all non-Num nodes.
+        var numbers = partitioned[0].terms as Num[];
         var fold =
             numbers.length &&
-            _.all(numbers, function (num) {
+            _.all(numbers, (num) => {
+                // TODO(kevinb): this is unsafe
```
handeyeco, line 1253: I think this comment could use more context. Is there something we need to change? Something we need to look out for?

handeyeco (commented):
Another batch of comments. Will continue looking at this after my meetings.

[ note to self, I left off at
```
1503                // expr.terms should always have at least two terms
1504                const last = _.last(expr.terms)!;
```
]

packages/kas/src/nodes.ts
kevinb-khan: Sometimes the args passed to node constructors can include numbers.  This is the case for `Num` and its subclasses.  I'm not sure why the existing code hasn't caused issues previously. 🤷‍♂️

packages/kas/src/nodes.ts
kevinb-khan: Yes it can.  I'll make this change.

packages/kas/src/nodes.ts
kevinb-khan: We need it here because one of the subclasses uses it.

packages/kas/src/nodes.ts
kevinb-khan: Without it TS thinks the return type is `any`.  Sometimes TS type inference is just really bad.  I tried turning on `noImplicitAny: true` in the tsconfig.json but then it was complaining about things that it was able to infer as a real type. 😞

packages/kas/src/nodes.ts
kevinb-khan: I'm a little hesitant to extract predicates into their own top-level function if they're only used in a single place.  I know there's a cost to redefining functions like this, but I don't believe that KAS performance is current an issue.

packages/kas/src/nodes.ts
kevinb-khan: I need to update this to use the `partition` helper function I added.

packages/kas/src/nodes.ts
kevinb-khan: I think this comment is stale.  I'll remove it.

packages/kas/src/nodes.ts
kevinb-khan: I hear you.  TS unfortunately isn't able to follow the logic when it's expressed that way.

packages/kas/src/nodes.ts
```diff
@@ -1976,13 +2026,17 @@ export class Pow extends Expr {
...
+    base: Expr;
+    power: Expr;
+
+    constructor(base: Expr, power: Expr) {
         super();
         this.base = base;
         this.power = power;
-        this.hints = _.extend(this.hints, {
+        this.hints = {
+            ...this.hints,
```
handeyeco, line 2037: These are hints from the parent class?

packages/kas/src/nodes.ts
```diff
@@ -2112,15 +2171,27 @@ export class Log extends Expr {
...
+    eval: (arg: number) => number;
+    codegen: string | ((argStr: string) => string);
+    tex: string;
+    expand?: () => Expr;
+};
+
 /* trigonometric functions */
 export class Trig extends Expr {
-    constructor(type, arg) {
+    type: string; // TODO(kevinb): Use an enum for this
```
handeyeco, line 2183: Do you mean an actual enum or a union? I was thinking we were generally trying to avoid enums.

packages/kas/src/nodes.ts
```diff
@@ -2387,27 +2433,33 @@ export class Trig extends Expr {
...
     }
 
-    tex(options) {
+    tex(): string {
+        var func = this.functions[this.type].tex;
+        var arg = "(" + this.arg.tex() + ")";
+        return func + arg;
+    }
+
+    texSplit(): [func: string, arg: string] {
```
handeyeco, line 2446: Good change 👍

packages/kas/src/nodes.ts
handeyeco: I'll defer to your judgement, but I'm a little confused by the "only used in a single place" part as `isAdd` is declared twice.

packages/kas/src/nodes.ts
```diff
@@ -2817,49 +2866,52 @@ export class Eq extends Expr {
...
-        }
+        const [a, b] = hasVar(expr.terms[0])
+            ? [
+                  Mul.handleNegative(expr.terms[1]),
+                  Mul.handleDivide(expr.terms[0], variable),
+              ]
+            : [
+                  Mul.handleNegative(expr.terms[0]),
+                  Mul.handleDivide(expr.terms[1], variable),
+              ];
```
handeyeco, line 2888: Same here, just seems like going without destructuring is clearer and more concise.

```suggestion
        const termHasVar = hasVar(expr.terms[0])
        const a = termHasVar
            ? Mul.handleNegative(expr.terms[1])
            : Mul.handleNegative(expr.terms[0])
        const b = termHasVar
            ? Mul.handleDivide(expr.terms[0], variable)
            : Mul.handleDivide(expr.terms[1], variable)
```

But again, just personal preference.

packages/kas/src/nodes.ts
```diff
@@ -3051,18 +3121,22 @@ export class Const extends Symbol {
 }
 
 /* abstract number node */
-class Num extends Expr {
+abstract class Num extends Expr {
+    n: number;
+
     constructor() {
         super();
+        this.n = 0;
```
handeyeco, line 3129: So this is just a placeholder because Num needs `n` for `getDecimalPlaces`? But really `n` gets set by its child classes? Just making sure I understand the motivation.

Also just curious, could this also be `n: number = 0;`?

packages/kas/src/nodes.ts
```diff
@@ -3553,13 +3615,15 @@ export const parse = function (input, options) {
         var expr = parser.parse(input).completeParse();
         return {parsed: true, expr: expr};
     } catch (e) {
-        return {parsed: false, error: e.message};
+        return {parsed: false, error: (e as Error).message};
```
handeyeco, line 3618: Is this possible? I feel like probably not or else I imagine you would have done this.

```TS
    } catch (e: Error) {
        return {parsed: false, error: e.message;
    }
```

packages/kas/src/nodes.ts
```diff
@@ -3744,42 +3809,36 @@ var siPrefixes = {
 };
 
 // Use these two values to mark a unit as either SI-prefixable or not.
-var hasPrefixes = {};
-var hasntPrefixes = {};
```
handeyeco, line 3748: 🤨 This is some big brain JS.

I think `Symbol` makes sense. Also wonder if a string literal would work?

```
const hasPrefixes = "HAS_PREFIXES";
const hasntPrefixes = "HASNT_PREFIXES";

type Prefixes = hasPrefixes | hasntPrefixes;
```

A possible nice side effect is that console logging would make it visible whether it's has/hasnt

packages/kas/src/nodes.ts
```diff
@@ -3744,42 +3809,36 @@ var siPrefixes = {
...
         numdenomStr[0]
             .split(" ")
-            .filter(function (x) {
-                return x !== "";
-            })
-            .map(function (x) {
-                numdenom.push(new Unit(x));
-            });
+            .filter((x) => x !== "")
+            .map((x) => numdenom.push(new Unit(x)));
```
handeyeco, line 3834: **nit against the original code:** I think using `map` here instead of `forEach` is semantically incorrect.

packages/kas/src/nodes.ts
```diff
@@ -3744,42 +3809,36 @@ var siPrefixes = {
...
         numdenomStr[1]
             .split(" ")
-            .filter(function (x) {
-                return x !== "";
-            })
-            .map(function (x) {
-                numdenom.push(new Pow(new Unit(x), Num.Div));
-            });
+            .filter((x) => x !== "")
+            .map((x) => numdenom.push(new Pow(new Unit(x), NumDiv)));
```
handeyeco, line 3841: Same here

handeyeco (approved):
Insomuch as I understand KAS, I think this PR makes sense. Thanks for doing all of this! I just ask that you don't land this right before going on sabbatical because I'm not sure any LEMS people understand this as much as you. 😂

packages/kas/src/nodes.ts
```diff
@@ -1582,7 +1620,7 @@ export class Pow extends Expr {
             // If Float, convert to a Rational to enable the logic below
             if (simplifiedExp instanceof Float) {
                 var num = simplifiedExp.n;
-                var decimals = (num - num.toFixed()).toString().length - 2;
+                var decimals = (num - +num.toFixed()).toString().length - 2;
```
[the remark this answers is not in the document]
kevinb-khan: That's correct.

packages/kas/src/nodes.ts
```diff
@@ -1886,19 +1938,15 @@ export class Pow extends Expr {
...
-            exp = factor.exp;
-        } else {
-            base = factor;
-            exp = Num.One;
-        }
+    findGCD(factor: Expr): Expr {
+        const [base, exp] =
+            factor instanceof Pow
+                ? [factor.base, factor.exp]
+                : [factor, NumOne];
```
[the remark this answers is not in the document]
kevinb-khan: I don't think TS will refine `factor` to be a `Pow` correctly if we do that.  We'd have to repeat the `factor instanceof Pow` in each ternary.

packages/kas/src/nodes.ts
kevinb-khan: That's correct.  I believe that `Expr` sets `parens: false` or something like that.

packages/kas/src/nodes.ts
kevinb-khan: I can go with a union type.

packages/kas/src/nodes.ts
kevinb-khan: You're right.  I'll hoist it.

packages/kas/src/nodes.ts
kevinb-khan: I can do this one.

packages/kas/src/nodes.ts
kevinb-khan: This is to make TypeScript happy.  We could do `n: number = 0;`, but I tend to prefer setting variables in the constructor when there is a constructor.  I find it hard to reason about the order things run when we're doing both.  In this case I don't see any harm in having `n: number = 0;` outside the constructor.

packages/kas/src/nodes.ts
kevinb-khan: Strings would be easier.  I'm not sure why the original author tried to be fancy with `hasPRefixes = {}`.  I'll switch these to strings.

packages/kas/src/nodes.ts
kevinb-khan: I didn't even notice that.  Good eye.  There really should be a lint rule for useless `.map()` calls.  I ran into the same thing in webapp in a couple of places as well.

PR #1751, opened 2024-10-15
Refine InputNumber's Rubric type

Myranae:
## Summary:
As part of the Server Side Scoring project, this refactors InputNumber's Rubric type to contain only what is needed for scoring. Also adds the UserInput type to the validator as it was defined already but not referenced.

Issue: LEMS-2467

## Test plan:
- Confirm all tests pass
- Confirm InputNumber still works as expected via Storybook


handeyeco (approved):
This is great, thanks Tamara

PR #1752, opened 2024-10-15
BUGFIX: [Number Line] - Tick labels not showing in some exercises

mark-fitzgerald:
## Summary:
A regression was introduced during the last bugfix ([PR 1695](https://github.com/Khan/perseus/pull/1695)) that caused the tick labels to not show for endpoints when the "Show label ticks" options is deselected. This PR fixes that regression by refactoring how the ticks and their labels are generated.

Issue: LEMS-2532

## Test plan:

1. Launch Storybook
1. Review the [Editor Demo](http://localhost:6006/?path=/story/perseuseditor-editorpage--demo) page
1. Add a Number line widget
   * A number line should show with the default 0 - 10 scale
   * Tick marks and labels are provided in steps of 2 (i.e. 0, 2, 4...)
1. Toggle the "Show label ticks" checkbox off
   * All the labels should now be removed except for the endpoints of zero (0) and ten (10)
1. Enter label values that are different than the endpoints
   * For instance, adding one (1) and seven (7)
   * The endpoint labels are now no longer showing
   * Instead, the newly specified labelled points are showing, along with their associated ticks

#### Before
Endpoints - Before

#### After
Endpoints - After

#### Before
Labels - Before

#### After
Labels - After

### DIFF
--- a/packages/perseus/src/widgets/number-line/number-line.test.ts
+++ b/packages/perseus/src/widgets/number-line/number-line.test.ts
@@ -4,7 +4,7 @@ import {testDependencies} from "../../../../../testing/test-dependencies";
 import * as Dependencies from "../../dependencies";
 import {renderQuestion} from "../__testutils__/renderQuestion";
 
-import {question1, snapshots} from "./number-line.testdata";
+import {question1} from "./number-line.testdata";
 
 import type {APIOptions} from "../../types";
 
@@ -21,18 +21,28 @@ describe("number-line widget", () => {
     });
 
     describe("snapshots", () => {
-        let originalOptions;
         let apiOptions: APIOptions;
+        const createNumberLineQuestionWithOptions = (widgetOptions) => {
+            return {
+                ...question1,
+                widgets: {
+                    ...question1.widgets,
+                    "number-line 1": {
+                        ...question1.widgets["number-line 1"],
+                        options: {
+                            ...question1.widgets["number-line 1"].options,
+                            range: [-10, 10],
+                            ...widgetOptions,
+                        },
+                    },
+                },
+            };
+        };
 
         beforeAll(() => {
-            originalOptions = snapshots.widgets["number-line 1"].options;
             apiOptions = {
                 isMobile: false,
             };
-        });
-
-        afterAll(() => {
-            snapshots.widgets["number-line 1"].options = originalOptions;
         });
 
         it("default", () => {
@@ -60,11 +70,12 @@ describe("number-line widget", () => {
 
         it(`only endpoints/labels show when "Show label ticks" is off`, () => {
             // Arrange - endpoints
-            let widgetOptions = {...originalOptions, labelTicks: false};
-            snapshots.widgets["number-line 1"].options = widgetOptions;
+            let question = createNumberLineQuestionWithOptions({
+                labelTicks: false,
+            });
 
             // Act
-            let {container} = renderQuestion(snapshots, apiOptions);
+            let {container} = renderQuestion(question, apiOptions);
 
             // Assert
             expect(container).toMatchSnapshot(
@@ -72,11 +83,13 @@ describe("number-line widget", () => {
             );
 
             // Arrange - labels
-            widgetOptions = {...widgetOptions, labelRange: [-4, 4]};
-            snapshots.widgets["number-line 1"].options = widgetOptions;
+            question = createNumberLineQuestionWithOptions({
+                labelTicks: false,
+                labelRange: [-4, 4],
+            });
 
             // Act
-            container = renderQuestion(snapshots, apiOptions).container;
+            container = renderQuestion(question, apiOptions).container;
 
             // Assert
             expect(container).toMatchSnapshot("show label ticks off (labels)");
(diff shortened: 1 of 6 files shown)


packages/perseus/src/widgets/number-line/number-line.test.ts
```diff
@@ -20,32 +20,137 @@ describe("number-line widget", () => {
...
+
+        beforeAll(() => {
+            originalOptions = snapshots.widgets["number-line 1"].options;
+            apiOptions = {
+                isMobile: false,
+            };
+        });
+
+        afterAll(() => {
+            snapshots.widgets["number-line 1"].options = originalOptions;
```
benchristel, line 35: It's a little unusual to mutate test data like this, and makes the tests harder to understand. If someone sees `snapshots` being used in a test, they might reasonably (but erroneously, in this case) conclude that it's the value being imported from `./number-line.testdata`.

I'd recommend copying the data with modifications instead:

```ts
const question = {
  ...snapshots,
  widgets: {
    ...snapshots.widgets,
    "number-line 1": {
      ...snapshots.widgets["number-line 1"],
      options: widgetOptions,
    },
  },
};

const {container} = renderQuestion(question, apiOptions);
```

You could create a `createNumberLineQuestionWithOptions()` helper function to de-duplicate the resulting code.

packages/perseus/src/widgets/number-line/number-line.tsx
```diff
@@ -124,17 +124,16 @@ const TickMarks: any = Graphie.createSimpleClass((graphie, props) => {
     const results: Array<any> = [];
 
     // For convenience, extract some props into separate variables
-    const range = props.range;
-    const labelRange = props.labelRange;
+    const {range, labelRange, labelStyle, labelTicks, tickStep} = props;
```
benchristel, line 127: Minor: why not destructure `numDivisions` along with the other props?

(I think adding `numDivisions` here would cause Prettier to break this statement into multiple lines, which... might not be worth the consistency. But I'm throwing it out there as an option.)

packages/perseus/src/widgets/number-line/number-line.tsx
```diff
@@ -144,36 +143,39 @@ const TickMarks: any = Graphie.createSimpleClass((graphie, props) => {
...
+    //    `.keys()` gets the index values for each marker placeholder
+    //    `.map()` converts the index values into actual tick numbers
+    const initialTicks: number[] = [...Array(props.numDivisions).keys()].map(
+        (index) => range[0] + index * tickStep,
     );
+    // Ensure that any label markers and range endpoints are included in the array
+    // Using `Set()` prevents duplication of tick numbers (and is quite performant)
+    const allTicks: number[] = [
+        ...new Set([...initialTicks, leftLabel, rightLabel, ...range]),
+    ].sort((current, next) => current - next);
```
benchristel, line 163: `current` and `next` are somewhat misleading names because the sort comparator callback is called with non-consecutive elements of the array. I might prefer something like:

```ts
// sort() comparator
const byNumericAscending = (a: number, b: number) => a - b

// ...

[/* ... */].sort(byNumericAscending);
```

to make the intention of this code explicit.

packages/perseus/src/widgets/number-line/number-line.tsx
```diff
@@ -144,36 +143,39 @@ const TickMarks: any = Graphie.createSimpleClass((graphie, props) => {
...
+
+    // Cycle through each tick number and add a tick line, and a label (if needed)
+    allTicks.forEach((tick) => {
+        const tickHasLabel = tick === leftLabel || tick === rightLabel;
```
benchristel, line 167: Based on how we use this variable, I think `tickIsEndpoint` might be a clearer name.

- A tick might have a label for reasons other than `tickHasLabel` being true.
- When this variable is true, we use `endpointLineStyle` and `endpointTextStyle`.

```suggestion
        const tickIsEndpoint = tick === leftLabel || tick === rightLabel;
```

packages/perseus/src/widgets/number-line/number-line.test.ts
```diff
@@ -4,7 +4,7 @@ import {testDependencies} from "../../../../../testing/test-dependencies";
 import * as Dependencies from "../../dependencies";
 import {renderQuestion} from "../__testutils__/renderQuestion";
 
-import {question1} from "./number-line.testdata";
+import {question1, snapshots} from "./number-line.testdata";
```
benchristel, line 7: This new import is just another question / PerseusItem, right? Why is it called `snapshots`? Why not `question3` (since `question2` is already defined)?

Also, why not use `question1` in the new tests? At a glance, I don't see much difference between `snapshots` and `question1`. If there is a significant difference, I'd prefer for it to be explicit in each test case where it matters (so, set explicitly when we create the `options` object).

benchristel (approved):
Thanks for the fix! And thanks for including screenshots in the PR description. Makes it super clear what's going on.

The logic looks good! I left a few style- and structure-oriented comments inline, but I'll leave it up to you whether to address them now or defer that to another PR.

packages/perseus/src/widgets/number-line/number-line.tsx
mark-fitzgerald: Good catch.

packages/perseus/src/widgets/number-line/number-line.tsx
benchristel: Sounds good to me!

PR #1753, opened 2024-10-16
Expose a way to get user input from ServerItemRenderer (#1753)

handeyeco:
## Summary:
[Webapp PR](https://github.com/Khan/webapp/pull/27079)

I'm trying to tread lightly, maybe being overly cautious at the expense of keeping too much old code around. I tried to make everything we can get rid of after the move with `@deprecated`.

Basically this should expose everything we need to move the actual scoring process out of ServerItemRenderer and the React tree. We now have `getUserInput` on `ServerItemRenderer` and `scorePerseusItem` which is a non-React, pure function that returns a score.

Next step is to replace uses of `scoreInput` in Webapp with `scorePerseusItem`; then we can come back and delete a lot of this legacy code 🤞

Issue: [LEMS-XXXX](https://khanacademy.atlassian.net/browse/LEMS-XXXX)

## Test plan:
After the swap in Webapp, we should be able to complete an exercise with `scorePerseusItem` and everything else will work the same.

[LEMS-2389]: https://khanacademy.atlassian.net/browse/LEMS-2389?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Author: handeyeco

Reviewers: jeremywiebe

Required Reviewers:

Approved By: jeremywiebe

Checks: ✅ Publish npm snapshot (ubuntu-latest, 20.x), ✅ Lint, Typecheck, Format, and Test (ubuntu-latest, 20.x), ✅ Check for .changeset entries for all changed files (ubuntu-latest, 20.x), ✅ Cypress (ubuntu-latest, 20.x), ✅ Publish Storybook to Chromatic (ubuntu-latest, 20.x), ✅ Check builds for changes in size (ubuntu-latest, 20.x), ✅ gerald

Pull Request URL: https://github.com/Khan/perseus/pull/1753


packages/perseus/src/server-item-renderer.tsx
```diff
@@ -298,13 +300,26 @@ export class ServerItemRenderer
         return this.props.item.hints.length;
     }
 
+    /**
+     * Returns an object of the widget `.getUserInput()` results
+     */
+    getUserInputMap(): UserInputMap {
```
jeremywiebe, line 306: Nit: can we just call this `getUserInput()`? We do type the return value and adding the return type as part of the function name just feels like 90's.

packages/perseus/src/renderer-util.test.ts
```diff
@@ -469,3 +479,42 @@ describe("scoreWidgetsFunctional", () => {
...
+        await userEvent.type(
+            screen.getByRole("textbox", {name: /nearest ten/}),
+            "230",
+        );
+        await userEvent.type(
+            screen.getByRole("textbox", {name: /nearest hundred/}),
+            "200",
+        );
+        const userInput = renderer.getUserInputMap();
+        const score = scorePerseusItem(question1, userInput, mockStrings, "en");
```
jeremywiebe, line 509: Scoring split from user input! A red-letter day in history! 🎉

packages/perseus/src/renderer-util.ts
```diff
@@ -50,11 +51,35 @@ export function emptyWidgetsFunctional(
...
 
+// TODO: combine scorePerseusItem with scoreWidgetsFunctional
+// once scorePerseusItem is the only one calling scoreWidgetsFunctional
+export function scorePerseusItem(
+    perseusRenderData: PerseusRenderer,
+    userInputMap: UserInputMap,
+    // TODO(LEMS-2461,LEMS-2391): these probably
+    // need to be removed before we move this to the server
+    strings: PerseusStrings,
+    locale: string,
```
jeremywiebe, line 62: I suspect we'll always need the locale for scoring, because of things like decimal separators.  But localized strings I agree with you on.

packages/perseus/src/renderer-util.ts
```diff
@@ -50,11 +51,35 @@ export function emptyWidgetsFunctional(
...
+): PerseusScore {
+    // There seems to be a chance that PerseusRenderer.widgets might include
+    // widget data for widgets that are not in PerseusRenderer.content,
+    // so this checks that the widgets are being used before scoring them
```
jeremywiebe, line 66: Yes. Our editor doesn't (or hasn't always) cleared out widget configs when they are removed from `content`... so there is quite a bit of content that's been published that has unused widget configs in them.

packages/perseus/src/renderer-util.ts
```diff
@@ -50,11 +51,35 @@ export function emptyWidgetsFunctional(
...
+    );
+    return Util.flattenScores(scores);
+}
+
 export function scoreWidgetsFunctional(
     widgets: PerseusWidgetsMap,
     // This is a port of old code, I'm not sure why
     // we need widgetIds vs the keys of the widgets object
-    widgetIds: Array<string>,
+    widgetIds: ReadonlyArray<string>,
```
jeremywiebe, line 82: 👍  Thanks.

packages/perseus/src/renderer.tsx
```diff
@@ -1717,46 +1717,42 @@ class Renderer extends React.Component<Props, State> {
     };
 
     /**
-     * Returns an object mapping from widget ID to perseus-style score.
-     * The keys of this object are the values of the array returned
-     * from `getWidgetIds`.
+     * Grades the content.
```
jeremywiebe, line 1720: **nit**: We've been asked to use "score" over "grade". :) 

EDIT: Oh, I see that was the original comment from below. Would you mind updating it still? 

```suggestion
     * Scores the content.
```

jeremywiebe (approved):
Yay!

packages/perseus/src/server-item-renderer.tsx
```diff
@@ -136,6 +139,16 @@ export class ServerItemRenderer
                 this.props.onRendered(true);
             }
         }
+
+        if (this.props.score && this.props.score !== prevProps.score) {
```
jeremywiebe, line 143: **question**: Should we use a deep equality check here instead of simply instance comparison? 🤔  If the score is identical, I guess it could still mean that there are different widgets that are empty and need to be highlighted.  I think I answered my own question. 😖

jeremywiebe (approved):

packages/perseus/src/server-item-renderer.tsx
```diff
@@ -60,10 +60,24 @@ type DefaultProps = Required<
 >;
 
 type State = {
+    /**
+     * questionCompleted is used by Radio and LabelImage to change what
+     * it renders after a learner has checked the correctness of an answer.
+     * For instance LabelImage disables inputs for correct answers and marks
+     * incorrect answers so learners can update what they submitted.
```
jeremywiebe, line 67: I like referencing widgets that use the flag as an example, but I think we can word this more generically (like, what _should_ this flag be used for). :) 

So this flag is used by the renderer to signal to widgets that the question is done and they can alter what they render to reveal explanations or other information about why the learner's input was wrong, etc. 

Your comment is basically good, I guess I would just suggest that we remove the wording that implies that only Radio and LabelImage use this (they do! but in the future they may not be the only ones).  Sorry, I'm probably splitting hairs.

jeremywiebe: Also, it might be good to add a comment something like this to the `WidgetProps<T>` type as it has this flag too and it'd be helpful within widgets to be able to see that prop's intended usage.

packages/perseus/src/server-item-renderer.tsx
```diff
@@ -60,10 +60,24 @@ type DefaultProps = Required<
...
+     */
     questionCompleted: boolean;
+    /**
+     * As far as I can tell, this is used to highlight empty widgets
+     * after a learner has clicked the "check" button. I don't think this could
+     * still be used though, because the "check" button is disabled while there
+     * are empty widgets.
+     */
     questionHighlightedWidgets: ReadonlyArray<any>;
```
jeremywiebe, line 76: That's correct.  And to be really clear, it's an array of widget IDs (strings). Could you try changing this to `ReadonlyArray<string>`

packages/perseus/src/types.ts
```diff
@@ -666,6 +666,11 @@ export type WidgetProps<
     problemNum: number | null | undefined;
     apiOptions: APIOptionsWithDefaults;
     keypadElement?: any;
+    /**
```
jeremywiebe, line 669: Thank-you!

PR #1754, opened 2024-10-16
Add getPromptJSON utility functions to Widget and Renderer

MikeKlemarewski:
## Summary:
For some background context, you can [read this Slack thread](https://khanacademy.slack.com/archives/C01AZ9H8TTQ/p1728591264497999) as well as [Matthew's comment on the ticket](https://khanacademy.atlassian.net/jira/software/c/projects/TUT/boards/443?assignee=712020%3Aff9ac7d7-fbbc-4108-9209-3c12cb651cd3&selectedIssue=TUT-930).

This adds a utility function called `getPromptJSON` to the renderer that can be called from webapp to get a JSON representation of the exercise data to pass as prompt context to Khanmigo.

Issue: [TUT-930](https://khanacademy.atlassian.net/browse/TUT-930?atlOrigin=eyJpIjoiYjcwOTdhMWU4NzYxNDUyYjk5NDhhOTRkNmU4MTk0ZjQiLCJwIjoiaiJ9)

## Test plan:
- Run the tests

[TUT-930]: https://khanacademy.atlassian.net/browse/TUT-930?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/perseus/src/widgets/radio/prompt-utils.ts
```diff
@@ -0,0 +1,27 @@
...
+    const {choices} = renderProps;
+
+    const options = choices.map((choice) => {
+        return {
+            value: choice.content,
+        };
+    });
+
+    return {
+        type: WidgetType.RADIO,
```
handeyeco, line 19: You can just use "radio" here. We use lower-kebab-case for widget types:

``` TS
// radio.ts
name: "radio",
```

``` TS
// numeric-input.tsx
name: "numeric-input",
```

``` suggestion
        type: "radio",
```

packages/perseus/src/prompt-types.ts
```diff
@@ -0,0 +1,32 @@
+export enum WidgetType {
+    RADIO = "RADIO",
+}
+
+export type BasicOption = {
```
handeyeco, line 5: If this is just used internally, I don't think we need to export it. Same with the others that aren't used outside this file.

In fact, since you made a separate file for the util, I wonder if these types should just be in that file?

For context, we've had a long history in Perseus of having several _really close_ types without understanding why they all existed. So if we could quarantine these types as much as possible I think we might be better off.

Also I have a hunch @jeremywiebe might swoop in and suggest we use existing types for these derivative types, ie:

``` TSX
export type RadioPromptJSON = {
  hasNoneOfTheAbove: PerseusRadioWidgetOptions["hasNoneOfTheAbove"];
}
```

Ultimately I'll defer to Jeremy since he's smarter than me, but I'm the one who suggested that we keep these types completely decoupled from our regular types because I thought it would be beneficial for TUT to have full control of the _output_ type as long as we kept full control of the _input_ type. I could totally be misguided with this suggestion.

packages/perseus/src/prompt-types.ts
```diff
@@ -0,0 +1,32 @@
...
+export type RendererPromptJSON = {
+    content: string;
+    widgets: {
+        [key: string]: WidgetPromptJSON;
+    };
+};
+
+export interface GetPromptJSONInterface {
+    getPromptJSON(): RendererPromptJSON;
+}
```
handeyeco, line 32: I bet we could just add this to `RendererInterface`

packages/perseus/src/renderer.tsx
```diff
@@ -1727,6 +1731,26 @@ class Renderer extends React.Component<Props, State> {
...
+    /**
+     * Returns a JSON representation of the content and widgets
+     * that can be passed to an LLM for prompt context.
+     */
+    getPromptJSON(): RendererPromptJSON {
+        const {content, widgets} = this.props;
+        const widgetJSON = {};
+
+        Object.keys(widgets).forEach((key) => {
+            const widget = this.getWidgetInstance(key);
```
handeyeco, line 1743: I'm not sure how much of an impact this would have, but this would be more consistent with `getUserInput`.

```suggestion
    getPromptJSON(): RendererPromptJSON {
        const {content} = this.props;
        const widgetJSON = {};

        this.widgetIds.forEach((id: string) => {
            const widget = this.getWidgetInstance(id);
```

packages/perseus/src/widgets/radio/prompt-utils.ts
```diff
@@ -0,0 +1,27 @@
+import {WidgetType, type RadioPromptJSON} from "../../prompt-types";
+
+import type {RenderProps} from "./radio-component";
+import type {PerseusRadioUserInput} from "../../validation.types";
+
+export const getPromptJSON = (
+    renderProps: RenderProps,
+    userInput: PerseusRadioUserInput,
+): RadioPromptJSON => {
+    const {choices} = renderProps;
```
handeyeco, line 10: Unless you want me to, I'm not really paying attention to what information you're getting (which I feel like is a TUT concern), but how you're getting it (which is a LEMS concern). lmk if you want more feedback here, but also I think it'll help to have [LEMS-2471](https://khanacademy.atlassian.net/browse/LEMS-2471) done so you have a clearer view of what you might need.

handeyeco (commented):
I think you're on the right track. The important thing from my perspective is that we really try to keep this prompt stuff from the rest of Perseus as much as possible, especially conceptually. I'd advocate for prefixes/suffixes, but I imagine writing `PromptRadioBasicChoicePromptDontUseForRegularPerseusTypesPrompt` would get old quickly; so we need to find a balance between isolation and absurdity.

packages/perseus/src/widgets/radio/prompt-utils.ts
MikeKlemarewski: I'm going to collaborate with one of our prompt engineers to check that we're getting the correct information. I'll be keeping an eye on that ticket. Thanks!

packages/perseus/src/widgets/radio/prompt-utils.ts
MikeKlemarewski: I updated the enum to be lower case

jeremywiebe (approved):
Matthew and I chatted and we're settled with this approach. :)  

Thanks for your patience @MikeKlemarewski

jeremywiebe:
Also, there's a missing changeset. You can add one by running `yarn changeset` and then defining if the change is major/minor/patch according to Semver and describing the change for the Changelog. That'll create a file that you can commit and push to this PR.

packages/perseus/src/widgets/numeric-input/numeric-input.tsx
```diff
@@ -42,7 +44,7 @@ const formExamples: {
     pi: (form, strings: PerseusStrings) => strings.piExample,
 };
 
-type ExternalProps = WidgetProps<
+export type ExternalProps = WidgetProps<
```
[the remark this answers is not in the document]
MikeKlemarewski: Good call. I actually can't tell why I was exporting this now. If I find a reason, then I'll update it to use `PropsFor`

PR #1758, opened 2024-10-16
Refine Radio's Rubric and UserInput types

Myranae:
## Summary:
As part of the Server Side Scoring project, this refactors Radio's Rubric type to only contain what is necessary for scoring and UserInput to only contain info about the user's input. Also adds the types to Radio's validator tests.

Issue: LEMS-2471

## Test plan:
- Confirm all checks pass
- Confirm Radio still works as expected via Storybook


packages/perseus/src/validation.types.ts
```diff
@@ -177,9 +177,14 @@ export type PerseusPlotterRubric = PerseusPlotterWidgetOptions;
...
-export type PerseusRadioRubric = PerseusRadioWidgetOptions;
+export type PerseusRadioRubric = {
+    // The choices provided to the user.
+    choices: ReadonlyArray<PerseusRadioChoice>;
+};
 
 export type PerseusRadioUserInput = {
+    /*TODO(LEMS-2541): countChoices seems to be necessary for rendering the
+       question, not scoring. Maybe move to renderProps?*/
     countChoices?: boolean;
```
handeyeco, line 188: So one way to look at the overall problem of RenderProps vs UserInput vs Rubric: is the thing in question used in the validator? Since `countChoices` isn't in the validator (or any of the functionality that the validator uses), it can't be in the Rubric (which is stuff we need from KA to validate the answer) or UserInput (which is stuff we need from the leaner to validate the answer). It's rendering logic, so RenderProps.

Then like `numCorrect` is a weird one:

- Is there anything the learner can do to affect `numCorrect`? No, so it's not UserInput.
- Is `numCorrect` something KA determines and is it used to score? Yes, so it's part of the Rubric.
- Is `numCorrect` something that determines how we render the component? Yes, so it needs to be part of RenderProps.

There's some overlap between all of this, but that's okay. What's important (for this leg of the work) is that we know exactly what shape the Rubric and UserInput is.

packages/perseus/src/widgets/radio/radio-validator.test.ts
```diff
@@ -162,13 +166,13 @@ describe("radioValidator", () => {
...
     it("can handle none of the above incorrect answer", () => {
-        const userInput = {
+        const userInput: PerseusRadioUserInput = {
             choicesSelected: [false, false, false, false, true],
             noneOfTheAboveSelected: true,
             noneOfTheAboveIndex: 4,
         };
 
-        const rubric = {
+        const rubric: PerseusRadioRubric = {
```
catandthemachines, line 175: I love all this typing. Thank you!

catandthemachines (approved):
Looks good to me!

packages/perseus/src/validation.types.ts
handeyeco: I went through the code to see where removing `countChoices` from Radio's user input would require a refactor, but didn't see it. I asked Jeremy to follow-up on this today so you have coverage while I'm gone.

Myranae: I think I was getting some weird errors the first time, but I was able to follow through with removing `countChoices` from the `UserInput` type, so that's been added here.

packages/perseus/src/widgets/radio/radio-validator.ts
```diff
@@ -21,7 +21,7 @@ function radioValidator(
         };
     }
 
-    // TODO: should numCorrect actually be on the rubric
+    // TODO(LEMS-2541): should numCorrect actually be on the rubric
```
handeyeco, line 24: Thanks for filing this ticket. I think `numCorrect`, `noneOfTheAboveIndex`, and `noneOfTheAboveSelected` are all problematic, but I don't think we have to address it in this PR. It should probably be a fast follow though.

PR #1759, opened 2024-10-16
Remove seemingly useless rubrics

handeyeco:
## Summary:
These rubrics weren't really useful, so I delete them.

Cleanup for: LEMS-2396

## Test plan:
Nothing should change, just some type tweaks


Myranae (approved):
Wow! So fast! Thanks for taking care of this! Did we think we would need the Rubric types for anything scoring related down the road? In other words, will each widget need a Rubric when our new scoring is set up? Maybe we don't know the answer to this yet, and I suppose we can always add these back after we get a better picture of how the widgets are set up.

catandthemachines (approved):
Looks good to me!

PR #1762, opened 2024-10-17
BUGFIX: [Number Line] - Some exercises with fractions would not render

mark-fitzgerald:
## Summary:
The code that generates the list of tick marks on the number line assumed an integer value for the number of ticks. Exercises with fractions could generate a seed value (`numDivisions`) that is a decimal value (i.e. 7.000001). This fix ensures that an integer value is used.

Issue: LEMS-2549

## Test plan:
1. Launch Storybook
1. Open the [Editor Page demo](http://localhost:6006/?path=/story/perseuseditor-editorpage--demo)
1. Add a Number Line widget
1. Configure the widget as shown in the image:
Number Line - Settings

1. The number line in the preview should be visible

### Before
Number Line - Before

### After
Number Line - After

### DIFF
--- a/.changeset/sour-horses-scream.md
+++ b/.changeset/sour-horses-scream.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+BUGFIX - [Number Line] - Some exercises with fractions wouldn't render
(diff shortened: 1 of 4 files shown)


SonicScrewdriver (approved):
LGTM! :)

PR #1763, opened 2024-10-18
Delete old jison-based parser

kevinb-khan:
## Summary:
#1746 introduces a new fully-typed recursive descent parser.  This PR deletes the old parser, replacing it with the new one.

The previous PR that implemented the new parser didn't included it in the bundle so the -3.22 kB (-8.29%) decrease in bundle size, while modes, is real savings.  

Issue: None

TODO:
- [ ] ~move the unit parser into its own file~ I looked into doing this and it's possible but in order for it to break the cyclic import between nodes.ts and parser.ts we need to make the `compare` methods in nodes.ts regular functions and move them into compare.ts with the rest of the comparison code.  I think this is a worthwhile endeavour, but it doesn't make sense in this PR.

## Test plan:
- yarn test packages/kas/src


packages/kas/src/nodes.ts
```diff
@@ -2931,18 +2933,17 @@ export class Func extends Sym {
...
             x: arg.eval(vars, options),
         });
         var parsedFunc = parse(func, options);
-        if (parsedFunc.parsed) {
-            return parsedFunc.expr.eval(newVars, options);
+        if (!parsedFunc.parsed) {
+            throw new Error(parsedFunc.error);
         }
-        // If parsedFunc isn't actually parsed, return its error
-        return parsedFunc;
```
kevinb-khan, line 2945: The expectation is that `eval()` always returns a number so I made this function throw instead.  We should probably have a variant of this that returns the `parsedFunc` object.  Then callers in other modules can call `safeEval()`.  We can restrict the use of unsafe `eval()` by making the method private.

TODO: Since that seems like a lot of work, I think for now, the best course of action would be to revert these changes, add a `@ts-expect-error`, and file a ticket.

packages/kas/src/nodes.ts
```diff
@@ -3820,9 +3765,13 @@ const makeAlias = function (str: string, prefixes: Prefixes) {
...
 
-    var coefficient = NumOne;
+    var coefficient: Expr = NumOne;
     if (coefficientStr !== "") {
-        coefficient = parse(coefficientStr).expr;
+        const result = parse(coefficientStr);
+        if (!result.parsed) {
+            throw new Error(result.error);
+        }
+        coefficient = result.expr;
```
kevinb-khan, line 3774: TODO: Revert this change and add a `@ts-expect-error`.  All of the calls to `makeAlias` happen on initialization.  Any issues should be caught by the unit tests.  I'll leave a comment about this.

PR #1765, opened 2024-10-18
Refine Sorter's Rubric type

Myranae:
## Summary:
As part of the Server Side Scoring project, this refactors Sorter's Rubric type to only contain data necessary for scoring. It also updates the tests to reflect this new type.

Issue: LEMS-2472

## Test plan:
- Confirm all checks pass
- Confirm Sorter widget still works as expected via Storybook


packages/perseus/src/validation.types.ts
```diff
@@ -188,7 +187,10 @@ export type PerseusRadioUserInput = {
     noneOfTheAboveSelected?: boolean;
 };
 
-export type PerseusSorterRubric = PerseusSorterWidgetOptions;
+export type PerseusSorterRubric = {
+    // Translatable Text; The correct answer (in the correct order). The user will see the cards in a randomized order.
```
handeyeco, line 191: This comment is a little confusing to me, is it pulled from somewhere?

handeyeco (approved):

packages/perseus/src/validation.types.ts
Myranae: Yes, this comment is included within the `PerseusSorterWidgetOptions`. I copy the whole block of options and then whittle it down to only what's needed. I could just remove the comment from here since it's still in the options object?

PR #1768, opened 2024-10-20
Update the Expression widget to support the 'Show Your Work' hackathon project

kevinb-khan:
## Summary:
This PR makes the following changes:
- adds a 'setValue' method on the Expression widget that can be used to set the value of the input field programmatically
- adds a 'disabled' prop which will be used to disable to widget when its rendered outside a Renderer
- adds a 'noBackground' prop which removes the 'background: white' styling that's present by default

The reason for rendering the Expression widget outside of the Renderer is so that we can access the new 'setValue' method on it.

Issue: None

## Test plan:
- install the PR published npm module in webapp and do some manual testing there to ensure that setValue is work as expected
- [ ] write some unit tests for this


kevinb-khan:
@jeremywiebe I decided to use the `MathInput` component from `@khanacademy/perseus` instead of the `Expression` widget.  It does everything I need it to do which means I don't need to make these modifications to the Expression widget.

PR #1770, opened 2024-10-21
Add Knip config

handeyeco:
## Summary:
The Knip config helps us ignore false positives from Knip

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,6 +97,7 @@
         "jest-environment-jsdom": "^29.7.0",
         "jest-serializer-html": "^7.1.0",
         "jest-specific-snapshot": "^5.0.0",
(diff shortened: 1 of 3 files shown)


mark-fitzgerald (approved):

package.json
```diff
@@ -97,6 +97,7 @@
         "jest-environment-jsdom": "^29.7.0",
         "jest-serializer-html": "^7.1.0",
         "jest-specific-snapshot": "^5.0.0",
+        "knip": "^5.33.3",
```
handeyeco, line 100: The top-level `package.json` doesn't differentiate between dev and non-dev deps since it's the package manager for the monorepo (not the shipped packages).

PR #1771, opened 2024-10-21
Knip 1: remove mock-unique-id

handeyeco:
## Summary:
Remove `getMockUniqueId`.

Part of my Knip experiment. Knip identified this as unused and I tried my best to verify that, but please double check!


benchristel (approved):
LGTM! `ag mock-unique-id` (which should catch all imports) finds only the test file.

jeremywiebe (approved):

PR #1774, opened 2024-10-21
Knip 4: remove MovablePoint duplicate export

handeyeco:
## Summary:
`MovablePoint` was being exported twice: `export class MovablePoint` and `export default MovablePoint`

Found by Knip.

### DIFF
--- a/packages/perseus/src/interactive2/movable-point.tsx
+++ b/packages/perseus/src/interactive2/movable-point.tsx
@@ -133,7 +133,7 @@ type State = {
 // tooltipResetFunctions, but never remove them.
 const tooltipResetFunctions: Array<() => void> = [];
 
-export class MovablePoint {
+class MovablePoint {
     graphie: Graphie;
     movable: Movable<Record<string, never>>;
     state: State;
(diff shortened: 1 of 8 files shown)


benchristel (commented):
I'd prefer to keep the named export and remove the default export.

Named exports make it easier to find where stuff is defined or used by grepping.

handeyeco:
@benchristel Thanks for the feedback. I don't really agree, but I made the change anyway. I think I see your point, that renaming imports is less explicit with default exports (`import Whatever from "something"` vs `import {Something as Whatever} from "something"`). I could certainly be convinced, I just think it's inconsistent with the rest of our code base.

No skin off my back either way though. I appreciate you taking the time to review all of these.

packages/perseus/src/interactive2/movable-point.tsx
```diff
@@ -801,5 +801,3 @@ function getKey(eventName: string, id: string): string {
 function getEventName(key: string): string {
     return key.split(":")[0];
 }
-
-export default MovablePoint;
```
jeremywiebe, line 805: We really should have a final call on whether we use default exports or named.  I find a single named export kinda weird, but it protects against an import changing the name and being confusing. 🤔

jeremywiebe (approved):

PR #1777, opened 2024-10-21
Knip 7: remove unused style constants

handeyeco:
## Summary:
I could see an argument for keeping these since they define a palette we could choose from. However, we're trying to more closely align with WB so I think we should think of that as our palette rather than maintaining our own.


benchristel (approved):
LGTM.

PR #1779, opened 2024-10-21
Knip 9: remove fake-react-native text

handeyeco:
## Summary:
Remove `Text`.

Part of my Knip experiment. Knip identified this as unused and I tried my best to verify that, but please double check!


benchristel (approved):
LGTM!

jeremywiebe:
I'd be happy to get rid of the entire `fake-react-native-web` folder. I have a sneaking suspicion we could replace its `View` component with a WB `View` and be done.

PR #1780, opened 2024-10-21
Knip 10: remove needless type exports

handeyeco:
## Summary:
If a type is only used in the file it's declared in, there's no need to export it.


benchristel (approved):
LGTM!

PR #1782, opened 2024-10-21
Knip 12: widget option refinement

handeyeco:
## Summary:
`PerseusSequenceWidgetOptions` and `PerseusSimulatorWidgetOptions` are obsolete due to the deprecated stand-in.

LabelImage and Measurer weren't using their widget options type, but I think they should be (the other widgets do).


handeyeco:
The parent pull-request ([#1770](https://github.com/Khan/perseus/pull/1770)) has been merged into `main`, but this branch (`knip-12`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

packages/perseus/src/widgets/label-image/label-image.tsx
```diff
@@ -67,23 +68,14 @@ type Point = {
     y: number;
 };
 
-// TODO: should this be using WidgetProps / PerseusLabelImageWidgetOptions?
 type LabelImageProps = ChangeableProps &
-    DependencyProps & {
+    DependencyProps &
+    // TODO: there's some weirdness in our types between
+    // PerseusLabelImageMarker and InteractiveMarkerType
```
jeremywiebe, line 74: I'm pretty sure the weirdness is that this component triggers `onChange` callbacks and "enhances" the items in the `markers` with what is essentially state. The `Renderer` then renders this widget with these enhanced `markers`. On first render, they will be of type `PerseusLabelImageMarker`, but after the first `onChange` they will be `InteractiveMarkerType` objects (which just add a few keys to the base `PerseusLabelImageMarker`).

jeremywiebe (approved):

packages/perseus/src/widgets/label-image/label-image.tsx
```diff
@@ -67,23 +68,14 @@ type Point = {
...
+    Omit<PerseusLabelImageWidgetOptions, "markers"> & {
```
benchristel, line 75: It looks like `MarkerType` in `label-image/types.ts` duplicates `PerseusLabelImageMarker` in `perseus-types.ts`. `InteractiveMarkerType` just adds some optional properties to `MarkerType`. Those optional properties seem like they're probably set when the widget emits state changes (through `onChange`).

packages/perseus/src/widgets/measurer/measurer.tsx
```diff
@@ -17,22 +18,10 @@ const defaultImage = {
...
     apiOptions: APIOptions;
-    box: [number, number];
-    image: {
-        url?: string;
-        top?: number;
-        left?: number;
-    };
-    showProtractor: boolean;
     protractorX: number;
     protractorY: number;
```
benchristel, line 24: Where do protractorX and protractorY come from, if not the widget options?

I suspect the default props are always used on first render, and then protractorX and protractorY get set by an onChange call.

This probably deserves a code comment.

benchristel (approved):
The way widget options represent both persisted config and transient state bothers me. But there's nothing we can do about that in this PR! LGTM.

packages/perseus/src/widgets/label-image/label-image.tsx
handeyeco: I think this is going to get addressed in https://khanacademy.atlassian.net/browse/LEMS-2440

PR #1784, opened 2024-10-21
Check if process exists before accessing it

handeyeco:
## Summary:
`process` is a NodeJS thing, so it's not guaranteed to be on the FE. This checks that it exists before accessing it.

### DIFF
--- a/packages/math-input/src/components/i18n-context.tsx
+++ b/packages/math-input/src/components/i18n-context.tsx
@@ -15,14 +15,11 @@ type I18nContextType = {
     locale: string;
 };
 
-const shouldMockStrings: boolean = Boolean(
-    process?.env?.NODE_ENV === "test" || process?.env?.STORYBOOK,
-);
-
 // @ts-expect-error - TS2322 - Type 'Context<{ strings: {}; locale: string; }>' is not assignable to type 'Context<I18nContextType>'.
 export const MathInputI18nContext: React.Context<I18nContextType> =
     React.createContext(
-        shouldMockStrings
+        typeof process !== "undefined" &&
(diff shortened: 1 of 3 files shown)


mark-fitzgerald (approved):

packages/math-input/src/components/i18n-context.tsx
```diff
@@ -15,10 +15,14 @@ type I18nContextType = {
     locale: string;
 };
 
+const shouldMockStrings: boolean = Boolean(
+    process?.env?.NODE_ENV === "test" || process?.env?.STORYBOOK,
```
benchristel, line 19: I don't think this will do what you want. If the `process` global variable doesn't exist, accessing it directly will fail. `process?.env` just returns undefined if `process` is nullish, but in order for that to work the variable `process` has to exist.

I think this will work:

```ts
typeof process !== "undefined" && (process?.env?.NODE_ENV === "test" || ... )
```

However, storing the result of a `process.env` check in a variable like this will prevent most build tools from doing dead code elimination. Vite, webpack, etc. will optimize an expression like:

```
process.env.NODE_ENV === "test" ? foo : bar
```

to `foo` in test environments and `bar` elsewhere. The env check won't actually appear in the built code. This behavior might be why this code currently works in browsers.

If the env check is abstracted at all, dead code elimination generally won't happen. I think that doesn't matter in this case, because the amount of code that would be eliminated here is tiny and not side-effecting, but I think as a general practice we'd want to keep env checks inline.

handeyeco: Yeah, it didn't feel write (and was breaking the SB build). Thanks for the feedback, fixed.

PR #1785, opened 2024-10-22
Parse PerseusItem data on input

benchristel:
See [ADR #773][1] for context.

NOTE TO REVIEWERS: The best place to start reading this code is probably `parse-perseus-json/parser-types.ts`, followed by the files in `parse-perseus-json/general-purpose-parsers`.

I have not written parsers for all widget types yet, but I want to get this foundational code merged because this PR is already way too big. The parser is not hooked up in production yet, so this change should have no effect.

If you have a suggestion for a term I should use other than "parse", I'd love to hear it. Some other ideas are listed [here](https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/3358883970/Public+notes+on+ADR+773+-+Validate+widget+data+on+input+in+Perseus#Words).

[1]: https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/3318349891/ADR+773+Validate+widget+data+on+input+in+Perseus

Issue: https://khanacademy.atlassian.net/browse/LEMS-2521

## Test plan:

CI should pass.


packages/perseus/src/util/parse-perseus-json.test.ts
benchristel, line 1: This file moved to `parse-perseus-json/parse-perseus-json.test.ts`.

packages/perseus/src/util/parse-perseus-json.ts
benchristel, line 1: This file moved to `parse-perseus-json/index.ts`.

packages/perseus/src/util/parse-perseus-json/error-tracking-parse-context.test.ts
benchristel, line 1: I made `ParseContext` an interface thinking that I would mock it in tests. That turned out to be a silly decision. I should get rid of the interface and rename this class to `ParseContext`.

packages/perseus/src/util/parse-perseus-json/parser-types.typetest.ts
```diff
@@ -0,0 +1,12 @@
...
+
+// summons a value of type T out of thin air!
+function summon<T>(): T {
+    return null as any as T;
+}
+
+summon<ParsedValue<typeof string>>() satisfies string;
+
+// @ts-expect-error - string parser returns a number
+summon<ParsedValue<typeof string>>() satisfies number;
```
benchristel, line 12: I should use this pattern in the other typetests. It's a lot less verbose.

packages/perseus/src/util/parse-perseus-json/general-purpose-parsers/array.test.ts
```diff
@@ -0,0 +1,82 @@
...
+            "ok",
+            "ok",
+            99, // index 2
+        ];
+
+        const result = arrayOfStrings(theArray, ctx());
+
+        expect(result).toEqual(
+            failure(
+                expect.objectContaining({
```
benchristel, line 47: This can be abbreviated `parseFailureWith({message: ..., path: ...})`

packages/perseus/src/util/parse-perseus-json/general-purpose-parsers/array.test.ts
```diff
@@ -0,0 +1,82 @@
...
+                    message: "expected string, but got 99",
+                    path: [2],
+                }),
+            ),
+        );
```
benchristel, line 52: This expected value can be abbreviated `parseFailureWith({message: ..., path: ...})`

packages/perseus/src/util/parse-perseus-json/general-purpose-parsers/array.ts
```diff
@@ -0,0 +1,15 @@
+import * as Result from "../result";
+
+import type {ParseContext, Parser} from "../parser-types";
+
+export function array<T>(elementParser: Parser<T>): Parser<T[]> {
+    return (array: unknown, ctx: ParseContext) => {
```
benchristel, line 6: TODO: rename `array` to `rawValue` to indicate its unknown type, and to avoid shadowing the array function.

packages/perseus/src/util/parse-perseus-json/general-purpose-parsers/compose-parsers.ts
benchristel, line 1: TODO: get rid of `composeParsers` and use [parserPipeline](packages/perseus/src/util/parse-perseus-json/general-purpose-parsers/parser-pipeline.ts) instead, which is more versatile.

packages/perseus/src/util/parse-perseus-json/general-purpose-parsers/constant.ts
```diff
@@ -0,0 +1,12 @@
...
+type Primitive = string | number | boolean | null | undefined;
+
+export function constant<T extends Primitive>(acceptedValue: T): Parser<T> {
+    return (rawValue, ctx) => {
+        if (rawValue !== acceptedValue) {
+            return ctx.failure(JSON.stringify(acceptedValue), rawValue);
+        }
+        return ctx.success(acceptedValue);
+    };
+}
```
benchristel, line 12: Possible bug: `JSON.stringify(undefined)` returns undefined, not a string. I think this might result in the correct overall behavior anyway, but it's not, as Adam Savage would say, a "positive fit".

packages/perseus/src/util/parse-perseus-json/general-purpose-parsers/enumeration.ts
```diff
@@ -0,0 +1,14 @@
...
+): Parser<T[number]> {
+    return (rawValue, ctx) => {
+        if (typeof rawValue === "string" && acceptedValues.includes(rawValue)) {
+            return ctx.success(rawValue as T[number]);
+        }
+        const expected =
+            "one of " + acceptedValues.map((v) => JSON.stringify(v)).join(", ");
+        return ctx.failure(expected, rawValue);
+    };
+}
```
benchristel, line 14: I think we might not need to check `typeof rawValue`.

packages/perseus/src/util/parse-perseus-json/general-purpose-parsers/object.ts
```diff
@@ -0,0 +1,24 @@
...
+        }
+
+        for (const [prop, propParser] of Object.entries(schema)) {
+            const result = propParser(rawValue[prop], ctx.forSubtree(prop));
+            if (result.type === "failure") {
+                return result;
+            }
+        }
+
+        return ctx.success(rawValue as {[K in keyof S]: ParsedValue<S[K]>});
```
benchristel, line 22: Bug: in order for defaulting and other processing functions to work, we need to return the parsed values here instead of rawValue.

```ts
const ret: {[K in keyof S]: ParsedValue<S[K]>} = {};
for (const [prop, propParser] of Object.entries(schema)) {
  const result = propParser(rawValue[prop], ctx.forSubtree(prop));
  if (isFailure(result)) {
    return result;
  } else {
    ret[prop] = result.value;
  }
}

return ctx.success(ret);
```

packages/perseus/src/util/parse-perseus-json/error-tracking-parse-context.test.ts
benchristel: On second thought, I'm going to keep the interface, because it allows the `types.ts` file to have _only_ types and not import any run-time code. That's important to prevent import cycles.

packages/perseus/src/util/parse-perseus-json/general-purpose-parsers/compose-parsers.ts
benchristel: hmm, parserPipeline is added in the next PR, and depends on `composeParsers`. So I guess I'll leave this for now.

packages/perseus/src/util/parse-perseus-json/general-purpose-parsers/enumeration.ts
benchristel: The `typeof` check is needed. `unknown` cannot be passed to either `includes` or `indexOf`.

packages/perseus/src/util/parse-perseus-json/general-purpose-parsers/object.ts
benchristel: This is fixed now!

jeremywiebe (approved):
Ben and I reviewed this in a pairing session!

PR #1786, opened 2024-10-22
Add regression tests for PerseusItem parsing, and fix some failures

benchristel:
These regression tests are generated from real production data.
What I did was:

- Download all English-language exercises
- Extract the `item_data`
- Group the items by their "shape" or schema - the
  structure of object properties and their types.
- Keep one item of each shape. This reduced the overall
  number of items in the dataset by around 90%.

The resulting dataset has around 70,000 Perseus items.
A subset of the items (those whose schema hash starts with 00) are included in this PR.

I have not even attempted to run all the tests yet.
There are still many failures.
The full dataset is large (hundreds of megabytes) so
it's unclear whether we would want to keep it in the Perseus repo
or move these tests somewhere else and only run them in CI.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2521

## Test plan:

CI should pass.


benchristel:
This PR is obsolete now. I am developing a more robust and storage-efficient way of generating regression tests so we don't have to commit hundreds of megabytes of test data to the repository.

PR #1787, opened 2024-10-22
[show-your-diff] Show a diff with the previous line

jaredly:
## Summary:
hacking around. There are definitely some parts that don't work, or don't work well enough.

Issue: https://khanacademy.atlassian.net/browse/FEI-5900

## Test plan:
storybook


kevinb-khan:
@jaredly thanks for putting up a PR this.  I'm going to close it for now, but may revisit it in the future.

PR #1789, opened 2024-10-22
Remove isMobile from Dependencies

handeyeco:
## Summary:
Remove isMobile from Dependencies since it's now on APIOptions


mark-fitzgerald (approved):
Lovely "fall" cleaning.

PR #1794, opened 2024-10-24
Remove unused optionRenderer prop on dropdown's Option component

jeremywiebe:
## Summary:

As I was working in the code I noticed that the `optionRenderer` prop is never passed to the Option nor OptionGroup components. So I'm removing it. 

Issue: "none"

## Test plan:

`yarn typecheck`
`yarn test`
`rg optionRenderer` (in hosting applications - webapp, mobile)

### DIFF
--- a/.changeset/perfect-guests-raise.md
+++ b/.changeset/perfect-guests-raise.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": patch
+---
+
+Remove unused prop, optionRenderer, from dropdown-option component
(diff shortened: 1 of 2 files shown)


handeyeco (approved):
Looks good, guessing from the pre-WB days?

jeremywiebe:
> Looks good, guessing from the pre-WB days?

There were some shared components that Perseus used that we copied in from webapp. I can see this option being used in webapp's version of this component, but Perseus doesn't so we can clean it up here.

PR #1795, opened 2024-10-24
Make `MovablePoint` use `useControlPoint`

benchristel:
Note to reviewers: it will probably be easiest to review this PR one commit at
a time.

When I started to add aria labels to movable points, I hit a snag: some of
our graph types make the movable point SVG itself tab-focusable, while other
graph types (e.g. segment) add a focusable dummy element so the visual
render order of the points can be different from the tabindex order. This
means that we'd want aria labels in different locations for each of these
graph types. Sometimes, the aria label would need to be on the movable point
itself, and sometimes it would need to be on the dummy element.

The solution implemented in this PR is to use the dummy element for all movable
points. That way we will only need to add accessible stuff to the dummy
element.

This PR is just refactoring; it doesn't add any features.

Issue: none

## Test plan:

`yarn dev`

Play with the graph types on `localhost:5173`. You should be able to control
each graph using either the mouse or the keyboard. A ring should always appear
on the focused point (whether it received focus from a mouse or keyboard
interaction).

### DIFF
--- a/.changeset/tame-insects-return.md
+++ b/.changeset/tame-insects-return.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Refactor MovablePoint to useControlPoint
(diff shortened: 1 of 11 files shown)


nishasy (approved):
Nice cleanup and refactoring here. I really appreciate the intentionality behind all the changes here.

PR #1797, opened 2024-10-24
Hackathon 2024 - Biome 2 - Apply formatting

jeremywiebe:
## Summary:

This is the second PR of setting up Biome. It formats all files using Biome's formatting rules (I've kept the rules as close to Prettier as possible to avoid too much disruption). 

The biggest changes are inlining some arrays that were split across multiple lines and adding trailing commas to some TSConfig files (we were inconsistent with trailing commas within these files... so I just picked one way and made them consistent). 
    
Issue: --hackathon-2024--

## Test plan:

`yarn biome --check`


jeremywiebe:
This is pretty stale now. I'll re-do it again if we decide we want to go the Biome direction.

PR #1802, opened 2024-10-25
Modernize Perseus' components folder stories

jeremywiebe:
## Summary:

I spent some of Hackathon 2024 looking into a Storybook upgrade (unfinished). During this time, I read up on how Storybook [recommends](https://storybook.js.org/docs/get-started/whats-a-story) writing stories. The new format ([CSF](https://storybook.js.org/docs/api/csf)) makes Typescript understand the stories much better. 

So, it turns out, with a few changes, the Storybook tooling becomes a bunch more helpful and reduces how much code we need to write in our stories. I'll outline the types of changes this PR contains so that it's easy to follow if you want to adjust other stories in this repo. 

Here is what a very, simple, modern Storybook file looks like: 

```ts
import MyComponent from "./my-component";

import type {Meta, StoryObj> from "@storybook/react";

const meta: Meta = {
    title: "Perseus/Components/MyComponent",
    component: MyComponent,
}
export default meta; 

type Story = StoryObj<typeof MyComponent>;

export const Default: Story = {};
```

1. Always define the default export as: 

   ```
   const meta: Meta = {
       title: "..path where this component should show up in storybook", 
       component: MyComponent
   }
   export default meta; 
   ```

   Two things are important here: 

     1. The default export is typed as `Meta` (from `@storybook/react`). This enables Storybook to infer the props this component has and enables "intellisense" on the different parameters to `meta`.
     1. We _do not_ use `as Meta` on this object definition. I'm unclear why, but that doesn't seem to work as well as the example above. 

1. Always define a type in the file as `type Story = StoryObj<typeof MyComponent>`. (again, the `StoryObj<T>` type comes from `@storybook/react`. 

3. Define stories as `export const MyStory: Story = {}`. By defining stories as simple data objects we get better type checking on the args we provide and reduce alot of boilerplate that our stories typically had. 

4. Use [`render`](https://github.com/Khan/perseus/pull/1802/files#diff-f821d3794fd8ed90901679eb1306920bc224c9b028b4f5bc8eae31f9b2e1fc62R15) and/or [`decorators`](https://github.com/Khan/perseus/pull/1802/files#diff-096ad8bf05853f7a13ab3a8d0ec4f50e7799d835babcf1275730e6571f8e0c7cR25) to reduce duplication. Often times we need to wrap our component in order for it to behave properly or to look better in Storybook. Instead of writing each story with this wrapper, move the common wrapping code into the `meta` object's `render` key. (note: I'm not 100% clear when to use the `render` key and when to use the `decorators`. I was able to accomplish similar things with both. I _suspect_ `decorators` is more useful if you have multiple things you want to adorn the component with but do not want to merge them together into a single `render` function.

Issue: "none"

## Test plan:

Review the stories in `Perseus/components` (`yarn start; open http://localhost:6006/?path=/docs/perseus-components-button-group--docs`)

### DIFF
--- a/packages/perseus/src/components/__stories__/graph.stories.tsx
+++ b/packages/perseus/src/components/__stories__/graph.stories.tsx
@@ -1,27 +1,22 @@
-import * as React from "react";
-
 import Graph from "../graph";
 
 import type {StoryObj, Meta} from "@storybook/react";
 
-type StoryArgs = StoryObj<Graph>;
-
-type Story = Meta<Graph>;
+type Story = StoryObj<typeof Graph>;
 
 const size = 200;
 
-export default {
+const meta: Meta = {
     title: "Perseus/Components/Graph",
-} as Story;
-
-export const SquareBoxSizeAndOtherwiseEmpty = (
-    args: StoryArgs,
-): React.ReactElement => {
-    return <Graph box={[size, size]} />;
+    component: Graph,
+    args: {
+        box: [size, size],
+    },
 };
+export default meta;
 
-export const LabeledSquaredBox = (args: StoryArgs): React.ReactElement => {
-    return (
-        <Graph box={[size, size]} labels={["First label", "Second label"]} />
-    );
+export const SquareBoxSizeAndOtherwiseEmpty: Story = {};
+
+export const LabeledSquaredBox: Story = {
+    args: {labels: ["First label", "Second label"]},
 };
--- a/packages/perseus/src/components/__stories__/graphie.stories.tsx
+++ b/packages/perseus/src/components/__stories__/graphie.stories.tsx
@@ -23,6 +23,10 @@ export default meta;
 
 export const SquareBoxSizeAndOtherwiseEmpty: Story = {};
 
+/**
+ * A demonstration of a Graphie rendered using the Perseus `Renderer` complete
+ * with overlaid labels and an image caption below.
+ */
 export const PieChartGraphieLabels = () => {
     return <ServerItemRendererWithDebugUI item={itemWithPieChart} />;
 };
--- a/packages/perseus/src/components/__stories__/graphie.stories.tsx
+++ b/packages/perseus/src/components/__stories__/graphie.stories.tsx
@@ -6,28 +6,23 @@ import Graphie from "../graphie";
 
 import type {StoryObj, Meta} from "@storybook/react";
 
-type StoryArgs = StoryObj<Graphie>;
-
-type Story = Meta<Graphie>;
+type Story = StoryObj<typeof Graphie>;
 
 const size = 200;
 
-export default {
+const meta: Meta = {
     title: "Perseus/Components/Graphie",
-} as Story;
-
-export const SquareBoxSizeAndOtherwiseEmpty = (
-    args: StoryArgs,
-): React.ReactElement => {
-    return (
-        <Graphie
-            box={[size, size]}
-            setDrawingAreaAvailable={() => {}}
-            setup={() => {}}
-        />
-    );
+    component: Graphie,
+    args: {
+        box: [size, size],
+        setup: () => {},
+        setDrawingAreaAvailable: () => {},
+    },
 };
+export default meta;
 
-export const PieChartGraphieLabels = (args: StoryArgs): React.ReactElement => {
+export const SquareBoxSizeAndOtherwiseEmpty: Story = {};
+
+export const PieChartGraphieLabels = () => {
     return <ServerItemRendererWithDebugUI item={itemWithPieChart} />;
 };
--- a/packages/perseus/src/components/__stories__/image-loader.stories.tsx
+++ b/packages/perseus/src/components/__stories__/image-loader.stories.tsx
@@ -17,6 +17,14 @@ const meta: Meta = {
         },
         onUpdate: () => {},
     },
+    parameters: {
+        chromatic: {
+            // This component only deals with loading images and providing a
+            // fallback if it fails. This is not very useful to snapshot so
+            // we're disabling it.
+            disableSnapshot: true,
+        },
+    },
 };
 export default meta;
 
--- a/packages/perseus/src/components/__stories__/image-loader.stories.tsx
+++ b/packages/perseus/src/components/__stories__/image-loader.stories.tsx
@@ -1,60 +1,40 @@
-/* eslint-disable @khanacademy/ts-no-error-suppressions */
 import * as React from "react";
 
-type StoryArgs = Record<any, any>;
-
-type Story = {
-    title: string;
-};
-
 import ImageLoader from "../image-loader";
+
+import type {Meta, StoryObj} from "@storybook/react";
 
 const svgUrl = "http://www.khanacademy.org/images/ohnoes-concerned.svg";
 const imgUrl = "https://www.khanacademy.org/images/hand-tree.new.png";
 
-export default {
+const meta: Meta = {
     title: "Perseus/Components/Image Loader",
-} as Story;
+    component: ImageLoader,
+    args: {
+        preloader: null,
+        imgProps: {
+            alt: "ALT",
+        },
+        onUpdate: () => {},
+    },
+};
+export default meta;
 
-export const SvgImage = (args: StoryArgs): React.ReactElement => {
-    return (
-        <ImageLoader
-            src={svgUrl}
-            preloader={null}
-            imgProps={{
-                alt: "ALT",
-            }}
-            onUpdate={() => {}}
-        />
-    );
+type Story = StoryObj<typeof ImageLoader>;
+
+export const SvgImage: Story = {
+    args: {
+        src: svgUrl,
+    },
 };
 
-export const PngImage = (args: StoryArgs): React.ReactElement => {
-    return (
-        <ImageLoader
-            src={imgUrl}
-            preloader={null}
-            imgProps={{
-                alt: "ALT",
-            }}
-            onUpdate={() => {}}
-        />
-    );
+export const PngImage: Story = {
+    args: {src: imgUrl},
 };
 
-export const InvalidImageWithChildrenForFailedLoading = (
-    args: StoryArgs,
-): React.ReactElement => {
-    return (
-        <ImageLoader
-            src="http://abcdefiahofshiaof.noway.badimage.com"
-            preloader={null}
-            imgProps={{
-                alt: "ALT",
-            }}
-            onUpdate={() => {}}
-        >
-            <span>You can see me! The image failed to load.</span>
-        </ImageLoader>
-    );
+export const InvalidImageWithChildrenForFailedLoading: Story = {
+    args: {
+        src: "http://abcdefiahofshiaof.noway.badimage.com",
+        children: <span>You can see me! The image failed to load.</span>,
+    },
 };
--- a/packages/perseus/src/components/__stories__/input-with-examples.stories.tsx
+++ b/packages/perseus/src/components/__stories__/input-with-examples.stories.tsx
@@ -1,4 +1,4 @@
-import {actions} from "@storybook/addon-actions";
+import {action} from "@storybook/addon-actions";
 
 import InputWithExamples from "../input-with-examples";
 
@@ -10,7 +10,7 @@ const meta: Meta = {
     args: {
         examples: [],
         id: "",
-        onChange: actions("onChange"),
+        onChange: action("onChange"),
         value: "",
     },
     argTypes: {
--- a/packages/perseus/src/components/__stories__/input-with-examples.stories.tsx
+++ b/packages/perseus/src/components/__stories__/input-with-examples.stories.tsx
@@ -15,7 +15,7 @@ const meta: Meta = {
     },
     argTypes: {
         onChange: {
-            table: {disable: true},
+            control: {type: null},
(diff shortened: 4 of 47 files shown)


packages/perseus/src/components/tooltip.tsx
```diff
@@ -233,6 +195,47 @@ type State = {
     height: number | null;
 };
 
+/**
+ * DEPRECATED! Use Wonder Blocks tooltip instead.
```
catandthemachines, line 199: Yess!!

catandthemachines (approved):
Looked through all the storybook pages updated and they are 10x better with the ability to see an interact with the props. Let's ship!

packages/perseus/src/components/__stories__/input-with-examples.stories.tsx
```diff
@@ -1,53 +1,48 @@
...
 
-type Story = {
-    title: string;
+const meta: Meta = {
+    title: "Perseus/Components/Input with Examples",
+    component: InputWithExamples,
+    args: {
+        examples: [],
+        id: "",
+        onChange: actions("onChange"),
```
nishasy, line 13:
```suggestion
        onChange: action("onChange"),
```

I don't think `actions` is the import you want here; I think you want `action`. Currently, it's not logging the `onChange` in the actions panel, and it's throwing an error in the console.

image

packages/perseus/src/components/__stories__/input-with-examples.stories.tsx
```diff
@@ -1,53 +1,48 @@
...
+        value: "",
+    },
+    argTypes: {
+        onChange: {
+            table: {disable: true},
+        },
```
nishasy, line 19: Why do you want to remove `onChange` from the table?

packages/perseus/src/components/__stories__/input-with-examples.stories.tsx
```diff
@@ -1,53 +1,48 @@
-import * as React from "react";
+import {actions} from "@storybook/addon-actions";
```
nishasy, line 1: I think you want `action` rather than `actions`.

```suggestion
import {action} from "@storybook/addon-actions";
```

https://storybook.js.org/docs/essentials/actions#exports

packages/perseus/src/components/__stories__/input-with-examples.stories.tsx
```diff
@@ -1,53 +1,48 @@
...
 
-export const DisabledInput = (args: StoryArgs): React.ReactElement => {
-    return (
-        <InputWithExamples
-            {...defaultObject}
-            disabled={true}
-            examples={testExamples}
-        />
-    );
+export const DisabledInput = {
```
nishasy, line 43:
```suggestion
export const DisabledInput: Story = {
```

packages/perseus/src/components/__stories__/math-input.stories.tsx
```diff
@@ -1,46 +1,53 @@
-import * as React from "react";
+import {actions} from "@storybook/addon-actions";
```
nishasy, line 1:
```suggestion
import {action} from "@storybook/addon-actions";
```

packages/perseus/src/components/__stories__/math-input.stories.tsx
```diff
@@ -1,46 +1,53 @@
...
+            advancedRelations: true,
+            basicRelations: true,
+            divisionKey: true,
+            logarithms: true,
+            preAlgebra: true,
+            trigonometry: true,
+        },
+        convertDotToTimes: false,
+        value: "",
+        onChange: actions("onChange"),
```
nishasy, line 21:
```suggestion
        onChange: action("onChange"),
```

packages/perseus/src/components/__stories__/multi-button-group.stories.tsx
```diff
@@ -2,65 +2,47 @@ import * as React from "react";
...
+    },
+    render: function WithState(props: PropsFor<typeof MultiButtonGroup>) {
+        const [values, updateValues] = React.useState(props.values);
+        return (
+            <MultiButtonGroup
+                {...props}
+                values={values}
+                onChange={updateValues}
+            />
+        );
```
nishasy, line 23: Oh I like this! I usually put the `render` separately in each story and add their states individually, but I never thought about putting it directly in the meta so it applies to all of them.

packages/perseus/src/components/__stories__/multi-button-group.stories.tsx
```diff
@@ -2,65 +2,47 @@ import * as React from "react";
...
+
+export const ButtonsWithNoTitles: Story = {
+    args: {
+        buttons: [
+            {value: "One", content: "Item #1"},
+            {value: "Two", content: "Item #2"},
+            {value: "Three", content: "Item #3"},
+        ],
+    },
 };
```
nishasy, line 39: **Optional:** This could be a good opportunity to add a story description using JS Docs right above the story definition.

```
/** 
 * Hover over the buttons to see their titles.
 */
```

packages/perseus/src/components/__stories__/number-input.stories.tsx
```diff
@@ -1,41 +1,54 @@
-import * as React from "react";
+import {action} from "@storybook/addon-actions";
```
nishasy, line 1: ✅

packages/perseus/src/components/__stories__/stub-tag-editor.stories.tsx
```diff
@@ -1,45 +1,47 @@
-import * as React from "react";
+import {actions} from "@storybook/addon-actions";
```
nishasy, line 1:
```suggestion
import {action} from "@storybook/addon-actions";
```

packages/perseus/src/components/__stories__/stub-tag-editor.stories.tsx
```diff
@@ -1,45 +1,47 @@
...
-type Story = {
-    title: string;
+import type {Meta, StoryObj} from "@storybook/react";
+
+const meta: Meta = {
+    title: "Perseus/Components/Stub Tag Editor",
+    component: StubTagEditor,
+    args: {
+        value: [],
+        onChange: actions("onChange"),
```
nishasy, line 12:
```suggestion
        onChange: action("onChange"),
```

packages/perseus/src/components/__stories__/text-input.stories.tsx
```diff
@@ -1,33 +1,37 @@
-import * as React from "react";
+import {actions} from "@storybook/addon-actions";
```
nishasy, line 1:
```suggestion
import {action} from "@storybook/addon-actions";
```

nishasy (changes requested):
THIS IS SO EXCITING! 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 

I'm requesting changes to make sure the actions work everywhere that's specified before shipping this, but everything else looks awesome.

I also had one question about why the action props are all removed from the control panel, but that's nonblocking. (Usually in wonder blocks, we keep all the function props in the table, but we set the control type to null so they're not interactable. But it's clear that they're still props on the component.)

packages/perseus/src/components/__stories__/image-loader.stories.tsx
```diff
@@ -1,60 +1,40 @@
...
+    },
+};
+export default meta;
+
+type Story = StoryObj<typeof ImageLoader>;
+
+export const SvgImage: Story = {
+    args: {
+        src: svgUrl,
+    },
```
nishasy, line 28: I'd disable chromatic on this, since I don't think the snapshot is working as intended.

```suggestion
    },
    
    parameters: {
        chromatic: {
            // Disabling because this isn't correctly testing the SVG, just a loading a spinner.
            disableSnapshot: true,
        },
    }
```

image

packages/perseus/src/components/__stories__/multi-button-group.stories.tsx
jandrade: Right? I also saw this and it's really neat and makes ton of sense!

packages/perseus/src/components/__stories__/graph.stories.tsx
```diff
@@ -1,27 +1,22 @@
...
     title: "Perseus/Components/Graph",
-} as Story;
-
-export const SquareBoxSizeAndOtherwiseEmpty = (
-    args: StoryArgs,
-): React.ReactElement => {
-    return <Graph box={[size, size]} />;
+    component: Graph,
+    args: {
+        box: [size, size],
```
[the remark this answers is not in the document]
jeremywiebe: I mostly started with nothing and saw errors like this: 

image

Also, looking at the `Props` vs `DefaultProps` reveals that only `box` is required, but not defaulted.

packages/perseus/src/components/__stories__/graphie.stories.tsx
```diff
@@ -6,28 +6,23 @@ import Graphie from "../graphie";
...
+    },
 };
+export default meta;
+
+export const SquareBoxSizeAndOtherwiseEmpty: Story = {};
 
-export const PieChartGraphieLabels = (args: StoryArgs): React.ReactElement => {
+export const PieChartGraphieLabels = () => {
     return <ServerItemRendererWithDebugUI item={itemWithPieChart} />;
 };
```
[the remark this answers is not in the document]
jeremywiebe: I'm not sure about that.  If we mark this as a `Story` and provide the `render` function, we're sill ignoring all of the data that a user might edit in the **Controls** area of Storybook. I feel like that'd be confusing. 

I'm almost wondering if this is even a good story. It's rendering components that are quite a bit higher-level than the `Graphie` component.  🤔

packages/perseus/src/components/__stories__/input-with-examples.stories.tsx
jeremywiebe: Ah, Thanks for catching that! Fixed!

packages/perseus/src/components/__stories__/input-with-examples.stories.tsx
jeremywiebe: It didn't seem useful because you can't edit it. Do you find in WonderBlocks that it's useful to see callback props in the Controls table?

packages/perseus/src/components/__stories__/input-with-examples.stories.tsx
jeremywiebe: Fixed across the board. :)

packages/perseus/src/components/__stories__/multi-button-group.stories.tsx
jeremywiebe: I'm [lazy](http://wiki.c2.com/?LazinessImpatienceHubris) and always look for the least amount of work for myself! 😂

packages/perseus/src/components/__stories__/graphie.stories.tsx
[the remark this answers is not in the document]
nishasy: Oh yeah, that's a good point about the controls. 🤦🏽 My bad. I think it's fine to keep the story for the context though. Maybe a docstring would help provide that context?

packages/perseus/src/components/__stories__/input-with-examples.stories.tsx
nishasy: Personally, I think it's helpful even if I can't change it in the controls table, because it still lets me know what props are available. I use the controls table for documentation in addition to the manual control.

packages/perseus/src/components/__stories__/multi-button-group.stories.tsx
jeremywiebe: giphy

packages/perseus/src/components/__stories__/zoomable.stories.tsx
```diff
@@ -32,18 +19,29 @@ const computeChildBounds = (
...
-            disableEntranceAnimation={!!args.disableEntranceAnimation}
-        >
+const meta: Meta = {
+    title: "Perseus/Components/Zoomable",
+    component: Zoomable,
+    args: {
+        computeChildBounds,
+    },
+    argTypes: {
+        children: {table: {disable: true}},
```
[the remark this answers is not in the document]
jeremywiebe: Hmm, in other places I removed props that I didn't think would be useful for folks to edit. But I can see leaving all props in the table just so people know all the props a component supports. I'll try that with all the ones I disabled.

packages/perseus/src/components/__stories__/image-loader.stories.tsx
jeremywiebe: Good call!  I've disabled this image-loader for snapshots, but added a delay to `SvgImage` as I think it's useful to see snapshots for that component.

packages/perseus/src/components/__stories__/input-with-examples.stories.tsx
jeremywiebe: Ok, that's convincing. I've updated all the stories I changed to just disable the controls instead of hiding them.

nishasy (approved):
Yayyyy thanks for the updates! This is SO EXCITING! And I learned a ton from your PR too 🎉

PR #1803, opened 2024-10-25
Scope exports of icon paths

handeyeco:
## Summary:
I'm just going through our `perseus` exports right now to see if I can make them more focused.

In this one:
- We were `export * as icons` which makes it hard figure out what's actually being used
- Some things that were exported were only used in `perseus`, so I stopped exporting them
- Some things were only used in `perseus-editor`, so I moved them
- Some things weren't being used at all, so I deleted them

## Test plan:
Nothing should change


packages/perseus/src/index.ts
```diff
@@ -47,7 +47,7 @@ export * as constants from "./components/constants";
 export * as globalStyles from "./styles/global-styles";
 // TODO(FEI-4504): Replace globalStyles with wonder-blocks where possible
 export * as globalConstants from "./styles/global-constants";
-export * as icons from "./icon-paths";
+export {iconChevronDown, iconTrash} from "./icon-paths";
```
jeremywiebe, line 50: I _wonder_ if we could just get rid of all these `icon-paths` and use http://phosphoricons.com 🤔  That might not be a simple replacement, but it might be relatively easy.

jeremywiebe (approved):
This is a good incremental improvement. As I commented elsewhere in this PR, it might be nice to just move to all Phosphor Icons.

packages/perseus/src/index.ts
catandthemachines: Hmm, what would the benefit to moving to these phosphoricons over our existing implementation. It seems like this would require adding another dependency onto our codebase.

catandthemachines (approved):
Looks good!

packages/perseus/src/index.ts
jeremywiebe: We already have phosphor icons as [a dependency](https://github.com/Khan/perseus/blob/main/packages/perseus/package.json#L76) and WB uses them as the official source for icon paths. 

https://khan.github.io/wonder-blocks/?path=/docs/packages-icon-phosphoricon--docs

PR #1804, opened 2024-10-25
Move constants.ts file

handeyeco:
## Summary:
It was defined in `perseus` but seemed to only be used in `perseus-editor`, so I moved it.


jeremywiebe (approved):
Nice! Fewer exports is great!

PR #1808, opened 2024-10-29
Add getPromptJSON for a batch of widgets

MikeKlemarewski:
## Summary:
This adds the getPromptJSON and tests for the following widgets:
- categorizer
- cs-program
- definition
- dropdown
- explanation
- expression

Issue: [TUT-930](https://khanacademy.atlassian.net/browse/TUT-930?atlOrigin=eyJpIjoiZTk2NzllN2M0ZmQ2NDczZDk3YzE1ZGJiM2ZiMWY1MDIiLCJwIjoiaiJ9)

## Test plan:
- `yarn test`
- Think if there are any other edge case tests

[TUT-930]: https://khanacademy.atlassian.net/browse/TUT-930?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/perseus/src/widgets/categorizer/prompt-utils.ts
```diff
@@ -0,0 +1,34 @@
...
+import type categorizer from "./categorizer";
+import type {PerseusCategorizerUserInput} from "../../validation.types";
+import type {PropsFor} from "@khanacademy/wonder-blocks-core";
+
+type WidgetProps = PropsFor<typeof categorizer.widget>;
+
+export type CategorizerPromptJSON = {
+    type: WidgetType;
+    options: {
+        items: WidgetProps["items"];
```
MikeKlemarewski, line 12: Any idea why these prop types come through as optional? They all have `| undefined` as part of the type, even though the props get [defined from this type](https://github.com/Khan/perseus/blob/main/packages/perseus/src/perseus-types.ts#L331), which isn't optional.

jeremywiebe: Hrm. Well, maybe `PropsFor<>` isn't the right call.  They are optional here because the `Categorizer` defines a default prop for `items`, meaning `PropsFor<>` calculates that `items` can be `undefined` because it's got a default.

MikeKlemarewski: Do you think using `React.ComponentProps<typeof categorizer.widget>;` would be a suitable alternative?

packages/perseus/src/widgets/dropdown/prompt-utils.test.ts
```diff
@@ -0,0 +1,29 @@
...
+
+        const resultJSON = getPromptJSON(renderProps, userInput);
+
+        expect(resultJSON).toEqual({
+            type: WidgetType.DROPDOWN,
+            options: {
+                items: renderProps.choices,
+            },
+            userInput: {
+                selectedIndex: userInput.value - 1,
```
handeyeco, line 25: I liked your comment on the other one
```suggestion
                // Offset to account for placeholder
                selectedIndex: userInput.value - 1,
```

packages/perseus/src/widgets/explanation/explanation.test.ts
```diff
@@ -265,6 +267,24 @@ describe("Explanation", function () {
...
+        const questionOptions = question1.widgets["explanation 1"].options;
+
+        // Act
+        const json = widget?.getPromptJSON?.() as ExplanationPromptJSON;
+
+        // Assert
+        expect(json).toEqual({
+            type: WidgetType.EXPLANATION,
+            showPrompt: questionOptions.showPrompt,
+            hidePrompt: questionOptions.hidePrompt,
```
handeyeco, line 283: You think y'all will need the text telling learners how to close the explanation popover?

packages/perseus/src/widgets/explanation/prompt-utils.ts
```diff
@@ -0,0 +1,22 @@
+import {WidgetType} from "../../prompt-types";
+
+import type explanation from "./explanation";
+import type {PropsFor} from "@khanacademy/wonder-blocks-core";
+
+export type ExplanationPromptJSON = {
+    type: WidgetType;
+    showPrompt: PropsFor<typeof explanation.widget>["showPrompt"];
```
handeyeco, line 8: If we decide to not go with my suggestion of writing out explicit, segregated types for the prompt JSON API, I'd rather we just export `Props` or something from Explanation and using that instead of `PropsFor<typeof explanation.widget>`

packages/perseus/src/widgets/explanation/explanation.test.ts
MikeKlemarewski: Hmmm, probably not now that I stop to mull it over.

packages/perseus/src/widgets/cs-program/prompt-utils.ts
```diff
@@ -0,0 +1,13 @@
+import {WidgetType} from "../../prompt-types";
+
+export type CSProgramPromptJSON = {
+    type: WidgetType;
+    isSupported: boolean;
+};
+
+export const getPromptJSON = (): CSProgramPromptJSON => {
```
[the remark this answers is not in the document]
MikeKlemarewski: Yeah I think you're right. I think I'll add that in my next branch. There's another unsupported widget in there.

packages/perseus/src/widgets/categorizer/prompt-utils.ts
jeremywiebe: That's a good thought Matthew. Typescript will still ensure the types compatible even if we define them manually. Given that `items` is just a `ReadonlyArray<string>` I think it's a good idea to not reference the Prop types in these LLM-specific types. 

@MikeKlemarewski would you mind adjusting `items` and `categories` to be defined manually? (ie. `{items: ReadonlyArray<string>, categories: ReadonlyArray<string>}`)

packages/perseus/src/widget-ai-utils/categorizer/prompt-utils.ts
```diff
@@ -0,0 +1,33 @@
+import type {WidgetType} from "../../prompt-types";
+import type {PerseusCategorizerUserInput} from "../../validation.types";
+import type categorizer from "../../widgets/categorizer/categorizer";
+import type {PropsFor} from "@khanacademy/wonder-blocks-core";
+
+type WidgetProps = PropsFor<typeof categorizer.widget>;
+
+export type CategorizerPromptJSON = {
+    type: WidgetType;
```
jeremywiebe, line 9: I think this should just be `type: "categorizer"` as it really shouldn't be any other type for this structure, should it?

packages/perseus/src/widget-ai-utils/dropdown/prompt-utils.ts
```diff
@@ -1,22 +1,19 @@
...
+    type: "dropdown";
     options: {
-        items: WidgetProps["choices"];
+        items: ReadonlyArray<string>;
     };
     userInput: {
-        selectedIndex: PerseusDropdownUserInput["value"];
+        selectedIndex: number;
     };
 };
```
jeremywiebe, line 13: Nice. Looking through this I like how we aren't using any Perseus types for these output types. Thanks for making this change.

jeremywiebe (approved):
This is looking good @MikeKlemarewski.  Thanks for going through a few iterations to get it here.

PR #1810, opened 2024-10-30
Add TeX support to dropdown

daniellewhyte:
## Summary:
This PR adds support for showing TeX in the dropdown

Issue: LIT-1425

## Test plan:
Inspect new story:
http://localhost:6006/?path=/story/perseus-widgets-dropdown--dropdown-with-math
- Valid TeX renders in OptionItems when dropdown is expanded
Screenshot 2024-12-13 at 1 41 39 PM

- Valid TeX renders when an option is selected
Screenshot 2024-12-13 at 1 41 45 PM

- aria-labels exist on the `mathjax-wrapper` -- HTML inspection. SR reads the labels fine in the list box, but I've had mixed results after an option is selected (works on Firefox, but not Safari). This is not unique to TeX content, and there are a couple issues to address it here: [LIT-1516](https://khanacademy.atlassian.net/browse/LIT-1516) [WB-1799](https://khanacademy.atlassian.net/browse/WB-1799)
Screenshot 2024-12-13 at 1 49 23 PM

[WB-1799]: https://khanacademy.atlassian.net/browse/WB-1799?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[LIT-1516]: https://khanacademy.atlassian.net/browse/LIT-1516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


daniellewhyte:
This will be ready for review once this update has been released to WB: https://github.com/Khan/wonder-blocks/pull/2354

packages/perseus/src/widgets/dropdown/dropdown.testdata.ts
```diff
@@ -1,6 +1,6 @@
 import type {PerseusRenderer} from "../../perseus-types";
 
-export const question1: PerseusRenderer = {
+export const basicDropdown: PerseusRenderer = {
```
mark-fitzgerald, line 3: Thank you!! Meaningful names are so hard to come by these days.

packages/perseus/src/widgets/dropdown/dropdown.tsx
```diff
@@ -74,13 +75,27 @@ class Dropdown extends React.Component<Props> implements Widget {
                 key="placeholder"
                 value="0"
                 disabled
-                label={this.props.placeholder || " "}
+                label={
+                    <Renderer
+                        content={this.props.placeholder}
+                        strings={this.context.strings}
+                        inline
+                    />
```
mark-fitzgerald, line 83: Using `<Renderer>` feels heavy for rendering TeX. Perhaps using the `<TeX>` component would be a lighter-weight implementation? [Access to this component is via `getDependencies()`](https://github.com/Khan/perseus/blob/341d316aa8727ebb9e7fde28fc4e2d8779aa3e82/packages/perseus/src/components/sortable.tsx#L777) instead of direct import. Can we give that a try?

catandthemachines:
I think there is a typo in your code somewhere, it reads out the same option `5x` for all the options.
image

catandthemachines:
> I think there is a typo in your code somewhere, it reads out the same option `5x` for all the options. image

This actually might be an issue with Chrome, as Safari doesn't do this. I would say this is non-blocking feedback, might go away if you use the <TeX> component.

packages/perseus/src/widgets/dropdown/dropdown.tsx
daniellewhyte: Looks like that component always renders everything as TeX.

Screenshot 2024-12-17 at 2 34 05 PM

mark-fitzgerald: Bummer. OK. Let's stick with the `<Renderer>` widget for now. Thanks for checking into it.

mark-fitzgerald (approved):

daniellewhyte:
@catandthemachines I think you're right that it's a VO+Chrome issue. I am able to recreate the same behavior on the MathJax stories. It doesn't read the square root.

Screenshot 2024-12-17 at 3 02 58 PM

packages/perseus/src/widgets/dropdown/dropdown.tsx
jeremywiebe: Mark and I chatted in DMs, but I'll circle back here.  I suggested the `TeX` component, but didn't realize there might be a mixture of standard content and TeX markup. In that case, `Renderer` is the correct tool. It handles parsing content and rendering TeX as needed. This is pretty common to use the `Renderer` in widgets to render content (even if widgets aren't supported).

PR #1811, opened 2024-10-30
Fix types in test for perseus-linter

benchristel:
While working on something else, I saw TypeScript errors coming from this file.
The error was:

```
packages/perseus-linter/src/__tests__/rule.test.ts:83:32 - error TS2339: Property 'check' does not exist on type 'never'.

83                 const lint = r.check(node, state, content);
```

I'm not sure why this just started erroring, but the types are definitely *weird*, so I'm
taking the opportunity to fix them.

There was also a test that depended on a previous test having been run. I fixed
that too.

Issue: none

## Test plan:

`yarn test`

### DIFF
--- a/packages/perseus-linter/src/__tests__/rule.test.ts
+++ b/packages/perseus-linter/src/__tests__/rule.test.ts
@@ -56,11 +56,6 @@ the previous heading was level ${previousHeading.level}`;
         return PureMarkdown.parse(markdown);
     }
 
-    it("makeRules() factory method", () => {
-        const rules: Rule[] = ruleDescriptions.map((r) => Rule.makeRule(r));
-        expect(rules).toHaveLength(ruleDescriptions.length);
-    });
-
     it("check() method", () => {
         const rules = ruleDescriptions.map((r) => Rule.makeRule(r));
         const tree = parseTree();
--- a/packages/perseus-linter/src/__tests__/rule.test.ts
+++ b/packages/perseus-linter/src/__tests__/rule.test.ts
@@ -52,19 +52,17 @@ the previous heading was level ${previousHeading.level}`;
         },
     ];
 
-    let rules: Array<never> | Array<Rule | any> = [];
-
     function parseTree() {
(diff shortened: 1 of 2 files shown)


packages/perseus-linter/src/__tests__/rule.test.ts
```diff
@@ -52,19 +52,17 @@ the previous heading was level ${previousHeading.level}`;
...
-
     function parseTree() {
         return PureMarkdown.parse(markdown);
     }
 
     it("makeRules() factory method", () => {
-        rules = ruleDescriptions.map((r) => Rule.makeRule(r));
+        const rules: Rule[] = ruleDescriptions.map((r) => Rule.makeRule(r));
         expect(rules).toHaveLength(ruleDescriptions.length);
-        rules.forEach((r) => expect(r instanceof Rule).toBeTruthy());
```
mark-fitzgerald, line 64: I wonder if this was intended to verify some sort of edge case. (just me being paranoid)

mark-fitzgerald (approved):
Thanks for cleaning things up. Great stewardship!

packages/perseus-linter/src/__tests__/rule.test.ts
benchristel: It's duplicating a check that the type system should be doing for us... unless the types are wrong? But it seems pretty clear that `Rule.makeRule` always returns a `Rule`:

```ts
    static makeRule(options: MakeRuleOptions): Rule {
        return new Rule(
            options.name,
            options.severity,
            options.selector ? Selector.parse(options.selector) : null,
            Rule.makePattern(options.pattern),
            options.lint || options.message,
            options.applies,
        );
    }
```

benchristel: ...and, now that I'm looking at it again, I see that `expect(rules).toHaveLength(ruleDescriptions.length);` is basically testing `Array.map`. So this whole test should probably be deleted.

PR #1812, opened 2024-10-30
Add aria-label to movable points

benchristel:
This is step one toward implementing screenreader interactions
for the Interactive Graph widget.

This PR does *not* implement everything described in LEMS-1725.  Notably
missing is the description of the entire graph (listing all points).
I will open another PR for that.

## Test plan:

- Use VoiceOver
- `yarn dev`
- `open https://localhost:5173`

Navigate to the Point graph and tab to focus the movable point.  VoiceOver
should read the coordinates.

Move the point with the arrow keys. VoiceOver should play a sound effect and
read the new coordinates with each keypress.

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/components/use-control-point.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/components/use-control-point.tsx
@@ -2,7 +2,6 @@ import * as React from "react";
 import {useState, useRef, useLayoutEffect} from "react";
 
 import {usePerseusI18n} from "../../../../components/i18n-context";
-
 import {snap, X, Y} from "../../math";
 import useGraphConfig from "../../reducer/use-graph-config";
 import {useDraggable} from "../use-draggable";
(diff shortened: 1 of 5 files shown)


packages/perseus/src/widgets/interactive-graphs/graphs/components/use-control-point.tsx
```diff
@@ -72,6 +75,12 @@ export function useControlPoint(params: Params): Return {
             className="movable-point__focusable-handle"
             tabIndex={disableKeyboardInteraction ? -1 : 0}
             ref={focusableHandleRef}
+            role="button"
+            aria-label={strings.srPointAtCoordinates({
+                x: String(point[X]),
+                y: String(point[Y]),
+            })}
+            aria-live="assertive"
```
nishasy, line 83: Did you add `aria-live="assertive"` for the unlimited points graph? This wouldn't do anything for a finite points graph, right?

nishasy: It might also be good to add a comment here about why we're using "assertive" instead of "polite"

benchristel: `aria-live="assertive"` causes the new location of the point to be announced whenever you move it. I'll add a comment!

nishasy (approved):
LGTM! Thanks for the comment. Looking forward to seeing this whole widget become screenreader accessible 🎉

PR #1813, opened 2024-10-31
OLD: Refine Interactive Graph Rubric type

Myranae:
## Summary:
As part of the Server Side Scoring project, this refines the Interactive Graph widget's rubric to only what is used during the validation project. Specifically, the type of the correct value was simplified to only what is used in validation.

Second iteration: https://github.com/Khan/perseus/pull/1818

Issue: LEMS-2473

## Test plan:
- Confirm all checks pass


packages/perseus/src/validation.types.ts
```diff
@@ -95,9 +97,38 @@ export type PerseusInputNumberUserInput = {
     currentValue: string;
 };
 
+// Would it be better just to use a question mark instead of having all of them also have undefined?
+// Are those two different types though? optional versus required with an undefined value?
+export type PerseusGraphCorrectType = {
```
jeremywiebe, line 102: I'd recommend using a union type here where each `type` is associated with the actual set of data that it requires.  

Eg: 
```ts
export type PerseusGraphCorrectType = {
    type: "angle";
    coords: [Coord, Coord, Coord];
    allowReflexAngles: boolean | undefined;
} | {
    type: "circle";
    ... whatever circle needs
} | ... other types. 
```

packages/perseus/src/widgets/interactive-graphs/interactive-graph-validator.test.ts
```diff
@@ -5,11 +5,28 @@ import {clone} from "../../../../../testing/object-utils";
...
+        "radius",
+        "match",
+        "allowReflexAngle",
+    ];
+    const correct = {};
+    for (const key in graph) {
+        if (correctKeys.includes(key)) {
+            correct[key] = graph[key];
+        }
+    }
```
jeremywiebe, line 28: Although this is an easy way to build out the `correct` key, I'm not sure it's the best way.  We'll probably want to switch on `graph.type` and then extract the correct set of values for that given graph type (see earlier comment about the `PerseusInteractiveGraphRubric` type).

packages/perseus/src/validation.types.ts
Myranae: This is exactly what I was working on before, so I'll just go back to that version then :) Thanks!

PR #1814, opened 2024-10-31
Wrap text node to reduce Google Translate bug

nedredmond:
[It looks like Google Translate doesn't play nicely with React.](https://martijnhols.nl/gists/everything-about-google-translate-crashing-react)
There is [a known issue that Google plans to address](https://issues.chromium.org/issues/41407169)... one day, maybe.
In the short term, it looks like if we wrap text nodes in elements, we should be good to go.

Issue: LEMS-2574

Testing:
Use Google Translate on any radio widget exercise and do something that causes the radio widgets to change state to render a text node (the easiest way was showing rationales). The red bar of doom doesn't show.

### DIFF
--- a/packages/perseus/src/widgets/radio/choice-icon/choice-icon.tsx
+++ b/packages/perseus/src/widgets/radio/choice-icon/choice-icon.tsx
@@ -41,7 +41,7 @@ function ChoiceInner(props: ChoiceInnerProps) {
     const letter = getChoiceLetter(pos, strings);
 
     if (!showCorrectness) {
-        return letter;
+        return <span>{letter}</span>;
     }
     if (correct) {
         return (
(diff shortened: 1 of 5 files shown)


jeremywiebe:
Hrm. I've looked at this Sentry issue many times and had no idea where it came from. I wonder if it's all down to Google Translate or other DOM manipulation tooling? 

https://khanacademyorg.sentry.io/issues/4737185317/?project=15744&query=is%3Aunresolved%20Perseus&referrer=issue-stream&statsPeriod=14d&stream_index=0

nedredmond:
@jeremywiebe seems likely!

packages/perseus/src/widgets/radio/choice-icon/choice-icon.tsx
```diff
@@ -68,14 +68,14 @@ function getDynamicStyles(
...
     borderColor: string;
     color: string;
     borderRadius: number;
 } {
-    let backgroundColor;
-    let borderColor;
-    let color;
+    let backgroundColor: string | undefined;
+    let borderColor: string;
+    let color: string;
```
benchristel, line 78: Thanks for fixing up these type issues!

benchristel (approved):
LGTM, thanks!

mark-fitzgerald (approved):
Ooof. Adding elements "just because" makes me sad.

Thanks for finding a simple work-around!

PR #1815, opened 2024-10-31
Let interactive graph components render a screenreader description

benchristel:
This refactoring will make it easy for each graph subtype (segment, point,
linear-system, etc.) to describe itself to a screenreader.

At this point, no graph actually renders a screenreader description.  Future
PRs will add descriptions for each graph type.

## Test plan:

`yarn test`


packages/perseus/src/widgets/interactive-graphs/graphs/angle.tsx
```diff
@@ -16,10 +16,25 @@ import {getIntersectionOfRayWithBox} from "./utils";
...
+    state: AngleGraphState,
+    dispatch: Dispatch,
+): InteractiveGraphElementSuite {
+    return {
+        graph: <AngleGraph graphState={state} dispatch={dispatch} />,
+        screenreaderDescription: null,
+    };
+}
+
 export function AngleGraph(props: AngleGraphProps) {
```
nishasy, line 38: Since all these graphs are no longer being rendered directly outside their respective files (`<AngleGraph />` ==> `renderAngleGraph()`), can we remove the `export` for all of these?

benchristel: There's an `index.ts` file in this directory that exports most of the graph components. Perhaps we should remove that. If we did, then I think this wouldn't need to be exported.

nishasy: Is the `index.ts` file only for exporting things that would be used outside this package? I feel like we could probably remove that, but I don't think that should block this in any case.

nishasy (approved):
Makes sense to me

packages/perseus/src/widgets/interactive-graphs/graphs/angle.tsx
benchristel: It seems that nothing imports from the index.ts file, so I'm going to remove it.

PR #1817, opened 2024-10-31
Add screenreader description for point graph elements

benchristel:
## Summary:
When a user focuses a whole interactive graph, they should be able to
hear a description of all the interactive elements. This PR implements
this feature for point graphs.

Test plan:

- Use VoiceOver
- `yarn dev`
- `open http://localhost:5173`
- Tab to the point graph. You should hear a description including the
  coordinates of the interactive point.

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/screenreader-text.test.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/screenreader-text.test.ts
@@ -1,12 +1,16 @@
 import {srFormatNumber} from "./screenreader-text";
 
 describe("srFormatNumber", () => {
-    it("trivially converts most integers to strings", () => {
+    it("trivially converts small integers to strings", () => {
         expect(srFormatNumber(3, "en")).toBe("3");
     });
 
     it("does not use thousands separators", () => {
         expect(srFormatNumber(1234567, "en")).toBe("1234567");
+    });
+
+    it("does not use scientific notation", () => {
+        expect(srFormatNumber(1e27, "en")).toBe("1000000000000000000000000000");
     });
 
     it("displays at most 3 decimal places", () => {
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/screenreader-text.test.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/screenreader-text.test.ts
@@ -1,0 +1,23 @@
+import {srFormatNumber} from "./screenreader-text";
+
+describe("srFormatNumber", () => {
+    it("trivially converts most integers to strings", () => {
+        expect(srFormatNumber(3)).toBe("3");
+    });
+
+    it("does not use thousands separators", () => {
+        expect(srFormatNumber(1234567)).toBe("1234567");
+    });
+
+    it("displays at most 3 decimal places", () => {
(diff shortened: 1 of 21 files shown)


packages/perseus/src/widgets/interactive-graphs/graphs/screenreader-text.ts
```diff
@@ -0,0 +1,12 @@
+const numberFormat = new Intl.NumberFormat(
+    undefined, // use default locale
```
benchristel, line 2: I'm not totally sure this is what we want. We might need to use the `locale` from the `PerseusI18nContext`.

packages/perseus/src/widgets/interactive-graphs/graphs/screenreader-text.test.ts
```diff
@@ -0,0 +1,27 @@
+import {srFormatNumber} from "./screenreader-text";
+
+describe("srFormatNumber", () => {
+    it("trivially converts most integers to strings", () => {
```
nishasy, line 4: Nit: I think the wording here would be more accurate if you said "floats" or simply "numbers" in this description since decimal numbers are not integers.

packages/perseus/src/widgets/interactive-graphs/graphs/screenreader-text.test.ts
```diff
@@ -0,0 +1,27 @@
...
+    it("rounds to 3 decimal places", () => {
+        expect(srFormatNumber(1.2345, "en")).toBe("1.235");
+    });
+
+    it("never displays a negative sign on zero", () => {
+        expect(srFormatNumber(-0, "en")).toBe("0");
+    });
+
+    it("uses a locale-appropriate decimal separator", () => {
+        expect(srFormatNumber(1.5, "de")).toBe("1,5");
```
nishasy, line 25: Neat test!

packages/perseus/src/widgets/interactive-graphs/types.ts
```diff
@@ -21,7 +21,7 @@ export type MafsGraphProps<T extends InteractiveGraphState> = {
 // end up in different sections of the DOM.
 export type InteractiveGraphElementSuite = {
     graph: ReactNode;
-    screenreaderDescription: ReactNode;
+    interactiveElementsDescription: ReactNode;
```
nishasy, line 24: ~~I'm curious about why this is a ReactNode rather than a string, which is what `describePointGraph` returns in point.tsx?~~

Is this a ReactNode so that you can use the `usePerseusI18n` hook? (Maybe a comment would be helpful for confused individuals like me 😅)

nishasy (approved):
image

packages/perseus/src/widgets/interactive-graphs/graphs/screenreader-text.test.ts
benchristel: This description is just for the first test case, which tests `3`. I guess "most" integers are not like `3` (they are large and would have thousands separators).

That reminds me: I should add a test to ensure that scientific notation is not used.

nishasy: OH 🤦🏽 I don't know why I thought this was a comment above it rather than the description of the test. Sorry about that.

PR #1818, opened 2024-11-01
Refine Interactive Graph Rubric type

Myranae:
## Summary:
As part of the Server Side Scoring project, this refines the Interactive Graph widget's rubric to only what is used during the validation project. It updates the typing of correct to a union of correct types for all types of graphs.

Initial iteration: https://github.com/Khan/perseus/pull/1813

Issue: LEMS-2473

## Test plan:
- Confirm all checks pass
- Confirm graphs of all types still work as expected


packages/perseus/src/perseus-types.ts
```diff
@@ -909,6 +909,125 @@ export type PerseusGraphTypeRay = {
     startCoords?: CollinearTuple;
 } & PerseusGraphTypeCommon;
 
+type AngleGraphCorrect = {
```
Myranae, line 912: Not sure if all these types and the object/arrays I add here should be in this file or another one. Was considering the types file in Interactive Graph, though there might be a better place for that.

Perhaps the object/arrays should just go directly in a test data file since it was mainly to create rubrics in the validator test? Wasn't sure if they might be helpful in the future.

packages/perseus/src/widgets/interactive-graphs/interactive-graph-validator.test.ts
```diff
@@ -1,17 +1,66 @@
...
 
 function createRubric(graph: PerseusGraphType): PerseusInteractiveGraphRubric {
-    return {graph, correct: graph};
+    const rubricCorrect = {};
+    for (const key of Object.keys(graph) as Array<keyof typeof graph>) {
+        if (allGraphCorrectKeys[graph["type"]].includes(key)) {
+            rubricCorrect[key] = graph[key];
+        }
+    }
+    return {graph, correct: rubricCorrect as PerseusGraphCorrectType};
```
benchristel, line 22: I think there's probably a more future-proof way to do this. In order for this code to continue to work correctly, we will need to remember to update `allGraphCorrectKeys` whenever we add a new field to one of the `*GraphCorrect` types. If we forget to update `allGraphCorrectKeys`, nothing will fail, but the `correct` return value will not contain all the keys it's supposed to.

In addition, I don't think there's anything to ensure the keys listed in `allGraphCorrectKeys` are actually keys of the corresponding `*GraphCorrect` type, nor that those keys also exist (with the same value types) on `PerseusGraphType`. I.e. all these types/values could get out of sync, and we wouldn't necessarily get any type errors forcing us to fix the problem.

One way to solve this is to switch on the graph type and construct the `correct` value in a more explicit way:

```ts
function createRubric(graph: PerseusGraphType): PerseusInteractiveGraphRubric {
  switch (graph.type) {
    case "circle":
      const {type, center, radius} = graph;
      return {graph, correct: {type, center, radius}};
    case "linear":
      const {type, coords} = graph;
      return {graph, correct: {type, coords}};
    // ...
    default:
      throw new UnreachableCaseError(graph);
  }
}
```

This seems like more code, but it's typesafe and future-proof. We can get rid of the `*GraphCorrectKeys` consts, too, so it might end up being around the same amount of code overall.

packages/perseus/src/widgets/interactive-graphs/interactive-graph-validator.ts
```diff
@@ -239,9 +234,7 @@ function interactiveGraphValidator(
             const guess = userInput.coords;
             const correct = rubric.correct.coords;
             if (
-                // @ts-expect-error - TS2532 - Object is possibly 'undefined'.
                 deepEq(guess[0], correct[0]) &&
-                // @ts-expect-error - TS2532 - Object is possibly 'undefined'. | TS2532 - Object is possibly 'undefined'.
```
benchristel, line 244: Yay for fixing `@ts-expect-error`s! 🎉

packages/perseus/src/perseus-types.ts
benchristel: I think it's fine to put them here. It's an easy decision to change later, in any case.

benchristel (commented):
The types look good! I think the function that constructs the rubrics should be future-proofed, though. I left a comment about that inline.

packages/perseus/src/widgets/interactive-graphs/interactive-graph-validator.test.ts
benchristel: Bonus: if the first thing you do is add the `default` branch with `throw new UnreachableCaseError(graph);` I think you can just fix the TS errors one by one and you'll end up with correct code.

packages/perseus/src/perseus-types.ts
```diff
@@ -909,6 +909,125 @@ export type PerseusGraphTypeRay = {
...
+type SegmentGraphCorrect = {
+    type: "segment";
+    // Expects a list of Coord tuples. Length should match the `numSegments` value.
+    coords: CollinearTuple[];
+};
+
+type SinusoidGraphCorrect = {
+    type: "sinusoid";
+    // Expects a list of 2 Coords
+    coords: ReadonlyArray<Coord>;
```
jeremywiebe, line 968: Could this also use `CollinearTuple`? Otherwise we technically allow 0 up to infinity coords.

packages/perseus/src/perseus-types.ts
```diff
@@ -909,6 +909,125 @@ export type PerseusGraphTypeRay = {
...
+
+type PolygonGraphCorrect = {
+    type: "polygon";
+    match: "similar" | "congruent" | "approx";
+    coords: ReadonlyArray<Coord>;
+};
+
+type QuadraticGraphCorrect = {
+    type: "quadratic";
+    // expects a list of 3 coords
```
jeremywiebe, line 955: I'm not totally convinced we need these comments. :)   The type should, and does, define how many coords we expect.  I think these comments would only be useful if our types don't allow us to be explicit enough (like a case where we _have_ to use `ReadonlyArray<Coord>` but we expect only 2 coords).

Myranae: Sounds good! I can remove them. They were just copied over from where the graph properties were defined.

packages/perseus/src/perseus-types.ts
Myranae: Updated it :)

packages/perseus/src/perseus-types.ts
```diff
@@ -909,6 +909,79 @@ export type PerseusGraphTypeRay = {
     startCoords?: CollinearTuple;
 } & PerseusGraphTypeCommon;
 
+type AngleGraphCorrect = {
+    type: "angle";
+    // Whether to show the angle measurements.  default: false
```
Myranae, line 914: Kept this comment so it's easy to tell what the default is.

packages/perseus/src/widgets/interactive-graphs/interactive-graph-validator.test.ts
```diff
@@ -172,14 +237,24 @@ describe("InteractiveGraph.validate on a segment question", () => {
...
+            },
+            correct: {
+                type: "angle",
+                coords: [
+                    [1, 1],
+                    [0, 0],
+                    [-1, -1],
+                ],
+            },
+        };
```
handeyeco, line 257: It really bugs me we have to duplicate this. **_I think we should probably keep it as is_** so we can move on, but:

1. `PerseusInteractiveGraphRubric` should probably be generic
2. I think a lot of these optional properties shouldn't be optional
3. This works even though I think it's technically (from a types perspective) incorrect:

``` TS
        const shared: AngleGraphCorrect = {
            type: "angle",
            coords: [
                [1, 1],
                [0, 0],
                [-1, -1],
            ],
        };
        const rubric: PerseusInteractiveGraphRubric = {
            graph: shared,
            correct: shared,
        };
```

But I'm fine moving on as long as we have some tickets for these problems.

packages/perseus/src/perseus-types.ts
handeyeco: `perseus-types.ts` is for external facing types. I think it's correct to have them here since the rubric will be an external data structure now.

packages/perseus/src/perseus-types.ts
benchristel: I think this comment might be meant for a different field? The line below this is `allowReflexAngles?: boolean;`, which controls whether angles of more than 180° are allowed.

handeyeco (approved):
Thanks for seeing this one through

packages/perseus/src/perseus-types.ts
```diff
@@ -909,6 +909,79 @@ export type PerseusGraphTypeRay = {
...
+};
+
+type QuadraticGraphCorrect = {
+    type: "quadratic";
+    coords: [Coord, Coord, Coord];
+};
+
+type SegmentGraphCorrect = {
+    type: "segment";
+    // Expects a list of Coord tuples. Length should match the `numSegments` value.
```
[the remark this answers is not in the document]
benchristel: There's no way for the learner to add or remove segments, so I don't know why we'd check `numSegments`. I'd just as soon delete this comment, since it's the kind of thing that's likely to go out of date (it refers to logic in another part of the codebase).

packages/perseus/src/widgets/interactive-graphs/interactive-graph-validator.test.ts
benchristel: To Matthew's point, I don't think the `coords` need to be present on any of the `graph` objects in these tests. The `graph.coords` aren't used for scoring.

I.e. in this test, the rubric could be:

```ts
        const rubric: PerseusInteractiveGraphRubric = {
            graph: {type: "angle"},
            correct: {
                type: "angle",
                coords: [
                    [1, 1],
                    [0, 0],
                    [-1, -1],
                ],
            },
        };
```

That would solve some of the duplication.

benchristel (approved):
This looks good to me! I left a few non-blocking suggestions inline, but I don't think they'll need another round of review (feel free to tag me again if you think differently, though).

packages/perseus/src/perseus-types.ts
Myranae: I think the default boolean is false. When the validator is scoring angle graphs, it sends the the value for `allowReflexAngles` to `getClockwiseAngle`. In that function, the default value for `allowReflexAngles` is false. So if the rubric does not set a value for it, it automatically goes to false. It seems like it would never be undefined though. Would you agree with that? Might be just a holdover from `PerseusGraphTypeAngle`
image

packages/perseus/src/perseus-types.ts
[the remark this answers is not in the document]
Myranae: Deleted :)

packages/perseus/src/widgets/interactive-graphs/interactive-graph-validator.test.ts
Myranae: Just for reference, decided not to use the helper function and instead just created the rubric object for each test.

packages/perseus/src/widgets/interactive-graphs/interactive-graph-validator.test.ts
Myranae: Removed the `coords` from each rubric's graph object. 

@handeyeco I am up for doing more work on this if it needs to be fixed while its fresh on my mind. If you would prefer to leave it for later, totally cool with that though. But, for the ones that are currently optional, they are most likely optional because the validator function has if/else statements where there is an else statement to cover when the property is undefined. If that's meant to be more of a catchall than a true option, I can just change the types to be required (I would prefer that anyway too if it's correct).

I'm not sure I understand how to make it generic, but I'll do a short bit of research to see if I can find an example.

Myranae: Here's my first attempt and what I'm testing out atm:
```
export interface PerseusInteractiveGraphRubric2<C, G> {
    correct: C;
    graph: G;
}
```

packages/perseus/src/perseus-types.ts
Myranae: Removed the comment as I made the field required for the rubric.

packages/perseus/src/widgets/interactive-graphs/interactive-graph-validator.test.ts
Myranae: Made the optional properties required, but might have to look back at making things generic. Will land it as it is now and come back if needed.

PR #1822, opened 2024-11-04
Add getPromptJSON utility functions for a second batch of widgets

MikeKlemarewski:
## Summary:
getPromptJSON utility functions for the following widgets:
- Grapher
- iFrame
- Image
- InputNumber
- Interaction
- LabelImage
- Matcher

Issue: TUT-930

## Test plan:
- Make sure the tests pass
- Check if any other test cases should be added


packages/perseus/src/widget-ai-utils/input-number/input-number.test.ts
```diff
@@ -0,0 +1,63 @@
...
+            graded: true,
+            options: {
+                maxError: 0.1,
+                inexact: false,
+                value: 0.5,
+                simplify: "optional",
+                answerType: "percent",
+                size: "small",
+            },
+        } as InputNumberWidget,
```
jeremywiebe, line 26: `as` causes us to lose [all type safety](https://tusharf5.com/posts/typescript-satisfies-operator/#:~:text=With%20the%20new%20satisfies%20operator,code%3B%20not%20the%20TypeScript%20compiler.). Does using `satisfies` work?

packages/perseus/src/widget-ai-utils/matcher/matcher.test.tsx
```diff
@@ -0,0 +1,89 @@
...
+            },
+        },
+    },
+};
+
+describe("matcher widget", () => {
+    beforeEach(() => {
+        jest.spyOn(Dependencies, "getDependencies").mockReturnValue({
+            ...testDependencies,
+            TeX: ({
```
jeremywiebe, line 46: Is this needed because we need to call the `onLoad` callback?  If so, I wonder if it'd be worth moving the call to `onLoad` into the `TeX` mock that lives in `testDependencies`?

jeremywiebe (approved):
Good stuff Mike.  I have a few requests for this PR and one for future PRs. 


## This PR
* avoid using `as` at all costs and use `satisfies` instead. `as` loses type safety and is very much like typing something as `any`. 
* specify the full expected value as literals (instead of using references to the input values). That type of test basically codes a tautology (http://codertom.com/2014/06/a-warning-against-tautological-tests/). :)  In the case of these tests, it's not the end of the world, but it avoids introducing cases where the test code can be wrong and the test become invalid. 

## Future PRs

* Could you break up the PRs into smaller chunks in the future?  I think in the case of these PRs, one-per-widget would have been a nice granularity.  It makes a bit more work on the submitters part, but OLC helps quite a bit with updating and landing PRs and it reduces the size so that PRs are quicker to get through. Thanks. :)

packages/perseus/src/widget-ai-utils/grapher/prompt-utils.ts
```diff
@@ -0,0 +1,44 @@
...
+export type GrapherPromptJSON = {
+    type: "grapher";
+    options: {
+        availableTypes: ReadonlyArray<string>;
+        range: [min: number, max: number][];
+        labels: ReadonlyArray<string>;
+        tickStep: [number, number];
+        gridStep?: [number, number];
+        snapStep?: [number, number];
+        backgroundImage?: string | null;
```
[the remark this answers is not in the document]
MikeKlemarewski: Oh yeah it totally should I. I must have miss typed that after my eyes glazed over for the day.

packages/perseus/src/widget-ai-utils/matcher/matcher.test.tsx
MikeKlemarewski: I think it's mocking the loading state for TeX. When I didn't have it, the matcher widget would just render a loading spinner. If I put it inside the testDependencies then this test passes. I'm not sure if that would have any other unintended side effects.

packages/perseus/src/widget-ai-utils/grapher/grapher.test.ts
```diff
@@ -0,0 +1,82 @@
...
+
+describe("grapher widget", () => {
+    it("should get prompt json which matches the state of the UI", async () => {
+        // Arrange
+        const {renderer} = renderQuestion(question);
+        const widget = renderer.getWidgetInstance("grapher 1");
+        const graphOptions = question.widgets["grapher 1"].options;
+
+        // Act
+        const json = widget?.getPromptJSON?.() as GrapherPromptJSON;
```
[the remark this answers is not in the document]
MikeKlemarewski: I'm on board with that!

packages/perseus/src/widget-ai-utils/matcher/matcher.test.tsx
MikeKlemarewski: I actually went back on this. After moving it there were a few tests that started failing because they were validating re-render counts. I think explicitly adding the onLoad where tests rely on TeX loading might be easier to catch than sneaky re-render issues.

PR #1825, opened 2024-11-06
[ExpressionEditor] Fixing typing and onChange behavior through explicit typing

catandthemachines:
## Summary:
Content creators notifed LEMS team of some odd save behavior happening in the PerseusEditor.
Upon investigation it became clear that the lack of typing resulted in some old code using
updateForm to update properties that were not contained within answerForm, resulting in
corrupt data appearing in the editor.

Issue: LEMS-2577

## Test plan:


packages/perseus-editor/src/widgets/expression-editor.tsx
```diff
@@ -509,11 +584,14 @@ class AnswerOption extends React.Component<
...
 
     toggleConsidered = () => {
+        // Update findNextIn to have stronger typing.
         const newVal = findNextIn(
             PerseusExpressionAnswerFormConsidered,
             this.props.considered,
         );
-        this.change({considered: newVal});
+        this.props.onChangeConsidered(
+            newVal as (typeof PerseusExpressionAnswerFormConsidered)[number],
```
jeremywiebe, line 593: As a general rule, we should avoid `as` as much as possible. `as` tells TypeScript that we know what the type should be and suppresses type checking. If you can use `satisfies` that'd be better. Finally, if the `findNExtIn()` adjustment in my previous comment works, you won't need this type "cast" at all.

packages/perseus-editor/src/widgets/expression-editor.tsx
```diff
@@ -571,9 +649,7 @@ class AnswerOption extends React.Component<
                         <Checkbox
                             label="Answer expression must have the same form."
                             checked={this.props.form}
-                            onChange={(value) => {
```
jeremywiebe, line 574: Really like these changes!  So much more explicit! Thanks

packages/perseus/src/index.ts
```diff
@@ -162,6 +162,7 @@ export {default as WIDGET_PROP_DENYLIST} from "./mixins/widget-prop-denylist";
 export type {PerseusOptions} from "./init";
 export type {ILogger, LogErrorOptions} from "./logging/log";
 export type {ServerItemRenderer as ServerItemRendererComponent} from "./server-item-renderer";
+export type {ExpressionWidgetProps} from "./widgets/expression/index";
```
jeremywiebe, line 165: As I said above, if we can use the `PerseusExpressionWidgetOptions` we won't need to export this type at all. :)

jeremywiebe (approved):
Several thoughts that will hopefully improve the typing here even more!  Happy to re-review when you've addressed the comments.

catandthemachines:
image
I can confirm the changes work. Content editors will still see that something has changed, but no more scrambled data <3

PR #1828, opened 2024-11-06
Add prompt utilities for InteractiveGraph

MikeKlemarewski:
## Summary:
Adds the getPromptJSON utility function and tests for the InteractiveGraph widget.
Updates the getUnsupportedPromptJSON function to take an optional message so
future interactive graph types can be reported as unsupported until we implement
a case in getPromptJSON for them.

Issue: TUT-930

## Test plan:
- Make sure tests pass
- Think if any other test cases should be added

### DIFF
--- a/packages/perseus/src/widget-ai-utils/interactive-graph/prompt-utils.test.ts
+++ b/packages/perseus/src/widget-ai-utils/interactive-graph/prompt-utils.test.ts
@@ -565,7 +565,7 @@ describe("getPromptJSON", () => {
         });
     });
 
-    it("should return JSON for an unsupported graph", () => {
+    it("should throw an exception for an unsupported graph", () => {
         const renderProps: any = {
             graph: {
                 type: "fake-graph-type",
@@ -586,12 +586,8 @@ describe("getPromptJSON", () => {
             ],
         };
 
-        const resultJSON = getPromptJSON(renderProps, userInput);
-
-        expect(resultJSON).toEqual({
-            type: "interactive-graph-unsupported",
-            isSupported: false,
-            message: "The graph type fake-graph-type is not supported.",
-        });
+        expect(() => {
+            getPromptJSON(renderProps, userInput);
+        }).toThrow("Unhandled case for 'fake-graph-type'");
     });
 });
--- a/packages/perseus/src/widget-ai-utils/interactive-graph/prompt-utils.ts
+++ b/packages/perseus/src/widget-ai-utils/interactive-graph/prompt-utils.ts
@@ -1,4 +1,4 @@
-import {getUnsupportedPromptJSON} from "../unsupported-widget";
+import {UnreachableCaseError} from "@khanacademy/wonder-stuff-core";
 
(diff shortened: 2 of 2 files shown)


packages/perseus/src/widget-ai-utils/interactive-graph/prompt-utils.ts
```diff
@@ -0,0 +1,295 @@
...
+        case "sinusoid":
+            input = {
+                coords: userInput.coords,
+            } as SinusoidUserInput;
+            break;
+        default:
+            return getUnsupportedPromptJSON(
+                "interactive-graph-unsupported",
+                `The graph type ${props.graph.type} is not supported.`,
+            );
```
jeremywiebe, line 282: I think we'd be better served leveraging TypeScript here and avoiding a message that I'm skeptical will be used. :) 

What if we just `throw new UnreachableCaseError(userInput.type);` here? You can see how it's used [here](https://github.com/Khan/perseus/blob/22d1c0229e52fc2dc6bc9f38a3ca400c6e4a3884/packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx#L645). 

Note that we might need to assign the type to a const before the switch due to how TypeScript narrowing works with object properties (or rather, how it doesn't work).

packages/perseus/src/widget-ai-utils/interactive-graph/prompt-utils.ts
```diff
@@ -0,0 +1,295 @@
...
+};
+
+export const getPromptJSON = (
+    props: React.ComponentProps<typeof interactiveGraph.widget>,
+    userInput: PerseusGraphType,
+): InteractiveGraphPromptJSON | UnsupportedWidgetPromptJSON => {
+    let input: UserInput;
+    let graphOptions: GraphOptions;
+
+    switch (props.graph.type) {
```
jeremywiebe, line 149: **thought**: If you extract these two `switch` statements into their own helper functions, could you have them return the value directly from each `case`?  That would avoid you needing all the `as XyzType`, I think. It would also make this function slightly more succinct. YMMV. :)

packages/perseus/src/widget-ai-utils/interactive-graph/prompt-utils.test.ts
```diff
@@ -0,0 +1,597 @@
...
+            type: "invalid",
+            coords: [
+                [0, 0],
+                [1, 1],
+            ],
+        };
+
+        const resultJSON = getPromptJSON(renderProps, userInput);
+
+        expect(resultJSON).toEqual({
```
jeremywiebe, line 591: Stated elsewhere too, but I think if you use the `UnreachableCaseError` you can rely on the type system to catch new types of graphs instead of delaying the check till runtime. It also means you don't have to accommodate this type of "failure" at all.

jeremywiebe (approved):
This looks great with the one note that I think you can let the type system handle unsupported graph types instead of doing it manually (`UnreachableCaseError()`).

PR #1831, opened 2024-11-06
Add prompt JSON utility functions for GradedGroup widget

MikeKlemarewski:
## Summary:
Adds getPromptJSON and tests for the GradedGroup widget.

Issue: TUT-930

## Test plan:
- Make sure tests pass
- Are there any other tests we should write?

### DIFF
--- a/packages/perseus/src/widget-ai-utils/graded-group/graded-group.test.ts
+++ b/packages/perseus/src/widget-ai-utils/graded-group/graded-group.test.ts
@@ -1,0 +1,197 @@
+import {act, screen} from "@testing-library/react";
+import {userEvent as userEventLib} from "@testing-library/user-event";
+
+import {testDependencies} from "../../../../../testing/test-dependencies";
+import * as Dependencies from "../../dependencies";
+import {renderQuestion} from "../../widgets/__testutils__/renderQuestion";
+
+import type {PerseusRenderer} from "../../perseus-types";
+import type {UserEvent} from "@testing-library/user-event";
+
+const question: PerseusRenderer = {
+    content: "---\n\n##Check your understanding!\n\n[[☃ graded-group 1]]\n\n",
+    images: {},
+    widgets: {
+        "graded-group 1": {
+            type: "graded-group",
+            alignment: "default",
+            static: false,
+            graded: true,
+            options: {
+                title: "Metabolic strategies of bacteria",
+                content:
+                    "1. **Which of the following statements about metabolic strategies of bacteria are true?**\n\n [[☃ categorizer 1]]",
+                images: {},
+                widgets: {
+                    "categorizer 1": {
+                        type: "categorizer",
+                        alignment: "default",
+                        static: false,
+                        graded: true,
+                        options: {
+                            static: false,
+                            items: [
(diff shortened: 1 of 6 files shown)


packages/perseus/src/widget-ai-utils/graded-group/graded-group.test.ts
```diff
@@ -0,0 +1,197 @@
...
+                        content:
+                            "Some bacteria synthesize their own fuel molecules/fix their own carbon (autotrophic), while others take in fixed carbon from their environments (heterotrophic).\n\nSome autotrophs use light energy to synthesize their own fuel molecules, while others extract energy from chemical sources.\n\nBacteria that extract energy from chemical sources and use it to fix carbon are called chemosynthetic organisms.  These bacteria may be essential to communities where light is not available, like those around deep-sea vents. They can form the base of the food chain (act as primary producers) in these ecosystems.\n\nSome bacteria have symbiotic (mutually beneficial) relationships with other organisms, living inside these organisms and providing them with nutrients.\n\n**The following statements about the metabolic strategies of bacteria are true:**\n\n[[☃ categorizer 1]]",
+                        widgets: {},
+                    },
+                },
+            },
+        });
+    });
+
+    it("should get prompt json which matches the state of the UI when the hint is expanded", async () => {
```
jeremywiebe, line 135: I don't see that the output varies if the hint is collapsed or expanded.  As a result, is the difference between these two tests relevant?

packages/perseus/src/widgets/graded-group/graded-group.tsx
```diff
@@ -118,6 +121,9 @@ export class GradedGroup
         answerBarState: "HIDDEN",
     };
 
+    rendererRef: Renderer | null | undefined;
+    hintRendererRef: Renderer | null | undefined;
```
jeremywiebe, line 125: Can you use `React.createRef<T>() here instead? 

```suggestion
    rendererRef = React.createRef<Renderer>();
    hintRendererRef = React.createRef<Renderer>();
```

Then you can pass them to the `ref` prop directly, like this `ref={this.rendererRef}`.  

This does mean that usages will need to change to access the `current` value (eg. `this.rendereRef.current`). 

Example: https://github.com/Khan/perseus/blob/d6381f7737ff44f1ec3d91c4ab25ce51f47b35b9/packages/perseus/src/components/math-input.tsx#L435

jeremywiebe (approved):

packages/perseus/src/widget-ai-utils/graded-group/graded-group.test.ts
MikeKlemarewski: So when the hint is collapsed, the widgets don't render, and [hint.widgets = {}](https://github.com/Khan/perseus/pull/1831/files#diff-d93802a06953a622491d7fb4ff6911e6e9a606700402929300694ed307f603edR128).
When expanded, [we see the rendered widget](https://github.com/Khan/perseus/pull/1831/files#diff-d93802a06953a622491d7fb4ff6911e6e9a606700402929300694ed307f603edR175-R191).

PR #1832, opened 2024-11-07
[Numeric Input] - Update styling to match modern inputs

mark-fitzgerald:
## Summary:
The styling of the numeric input widget is quite constrained in regards to its size, and it is difficult to know when it has focus. Increasing its height will make it more physically accessible (better touch target). Adding more internal padding and increasing the border width will help with visual accessibility regarding a user's recognition of it having focus. Also, matching the styling that is used for the Expression widget (a close relative of this widget) lessens the cognitive load on users who see these two inputs as being similar/same (they don't know that they are two different widgets).

Issue: LEMS-2426

## Test plan:
1. Navigate to Perseus => Widgets => NumericInput => [Question 1](http://localhost:6006/?path=/story/perseus-widgets-numericinput--question-1)
   * Note that the input field is larger
1. Click or tab into the field
   * Note that the border is thicker and that the cursor is more visible (not next to the border as it was before)

#### Before:
Before - Standard

#### After:
After - Standard

#### Before:
Before - Focused

#### After:
After - Focused


packages/perseus/src/widgets/numeric-input/numeric-input.tsx
```diff
@@ -235,29 +243,34 @@ export class NumericInput
...
             input: {
+                borderRadius: "3px",
+                borderWidth: this.state.isFocused ? "2px" : "1px",
+                display: "inline-block",
+                fontFamily: `Symbola, "Times New Roman", serif`,
+                fontSize: "18px",
+                height: "32px",
+                lineHeight: "18px",
+                padding: this.state.isFocused ? "4px" : "4px 5px", // account for added focus border thickness
                 textAlign: this.props.rightAlign ? "right" : "left",
```
SonicScrewdriver, line 254: Thank you for the comments to help explain the difference!

SonicScrewdriver (approved):
This looks wonderful! :) Thank you for the great improvements

PR #1833, opened 2024-11-07
Remove unnecessary MathJax 2 logic from Perseus.Init

SonicScrewdriver:
## Summary:
While investigating a regression on Webapp related to our Deprecated Stand-in Widget, we discovered we were initializing MathJax unnecessarily! This initialization is specifically for MathJax2 so it is safe to remove entirely. 

Issue: LEMS-2588

## Test plan:
- Manual testing 
- Knowledge that this code has not been running for 2 months, and we've had zero issues with our MathJax rendering
- Knowledge that this is unused due to being MathJax2

### DIFF
--- a/packages/perseus/src/init.ts
+++ b/packages/perseus/src/init.ts
@@ -2,57 +2,19 @@ import basicWidgets from "./basic-widgets";
 import extraWidgets from "./extra-widgets";
 import * as Widgets from "./widgets";
 
-declare const MathJax: any;
-
-export type PerseusOptions = {
-    // TODO(LEMS-1608): remove skipMathJax once we have completely removed the
-    // legacy MathJax 2 renderer from webapp.
-    skipMathJax: boolean;
-};
-
 /**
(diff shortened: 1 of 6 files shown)


packages/perseus/src/init.ts
```diff
@@ -2,57 +2,19 @@ import basicWidgets from "./basic-widgets";
...
  * This should be called by all clients, specifying whether extra widgets are
  * needed via `loadExtraWidgets`. It is idempotent, so it's not a problem to
  * call it multiple times.
- *
- * skipMathJax:
- *   If false/undefined, MathJax will be configured, and the
- *   promise will wait for MathJax to load (if it hasn't already).
  */
-const init = function (options: PerseusOptions): Promise<undefined> {
+const init = function (): Promise<undefined> {
```
jeremywiebe, line 10: Given that there's nothing async in this function anymore, can you make it just `const init = function() {`?

jeremywiebe (approved):
Awesome.

PR #1834, opened 2024-11-07
[Locked Figures] Use $ symbols to denote TeX within locked labels and locked figure labels

nishasy:
## Summary:
In the rest of the exercise editor, content authors need to surround text in `$` for it to be denoted as TeX.

For consistency, we want to add that functionality to the interactive graph editor as well.

- Update LockedLabelSettings to say "text" instead of "TeX" and have an infotip to explain it.
- Update LockedLabel so that it surrounds the non-$ part of the string with \text{} before
  wrapping the whole thing in <TeX>

Note: This is a prerequisite for auto-generating spoken math details with the MathJax speech engine
for the aria labels, because the non-TeX text needs to be differentiated in the label for the label
to be generated correctly for non-TeX text. If it's all TeX, the speech engine reads every letter
separately (example: "<TeX>square A</TeX>" ==> "s q u a r e upper A")

## Test plan:
`yarn jest packages/perseus/src/widgets/interactive-graphs/utils.test.ts`

Storybook
- Play around with the locked label and the locked figure labels

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/utils.test.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/utils.test.ts
@@ -1,4 +1,4 @@
-import {normalizePoints, normalizeCoords} from "./utils";
+import {normalizePoints, normalizeCoords, replaceOutsideTeX} from "./utils";
 
 import type {Coord} from "../../interactive2/types";
 import type {GraphRange} from "../../perseus-types";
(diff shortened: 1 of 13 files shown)


packages/perseus/src/widgets/interactive-graphs/utils.ts
```diff
@@ -70,3 +70,27 @@ export function isUnlimitedGraphState(
...
+        if (currentlyTeX) {
+            result += part;
+        } else {
+            result += `\\text{${part}}`;
+        }
+        currentlyTeX = !currentlyTeX;
+    }
+
+    return result;
+}
```
nishasy, line 96: I tried doing some regex stuff along with `mathString.replace()` but it got convoluted and kept letting edge cases slip through the cracks. I decided to go back to basics and stick to what works, and this was the resulting code.

packages/perseus/src/widgets/interactive-graphs/utils.test.ts
```diff
@@ -65,3 +65,58 @@ describe("normalizeCoords", () => {
         expect(result).toEqual(expected);
     });
 });
+
+describe("replaceOutsideTeX", () => {
+    test("no $s", () => {
```
anakaren-rojas, line 70: Since these tests do the same thing, using a table might make things easier since you won't need to redo the code for each test, just have a table of inputs
https://jestjs.io/docs/api#2-testeachtablename-fn-timeout

packages/perseus/src/widgets/interactive-graphs/utils.ts
```diff
@@ -70,3 +70,27 @@ export function isUnlimitedGraphState(
         (state.type === "polygon" && state.numSides === "unlimited")
     );
 }
+
+/**
+ * Replace all text outside of the $ TeX blocks with `\\text{...}`
+ * This way, the entire resulting string can be rendered within <TeX>
+ * and the text outside of the $ blocks will be non-TeX text.
+ */
+export function replaceOutsideTeX(mathString) {
```
anakaren-rojas, line 79: For type safety
```suggestion
export function replaceOutsideTeX(mathString: string) {
```

packages/perseus/src/widgets/interactive-graphs/utils.ts
```diff
@@ -70,3 +70,27 @@ export function isUnlimitedGraphState(
...
+
+    const splitString = mathString.split("$").filter((part) => part !== "");
+
+    for (let i = 0; i < splitString.length; i++) {
+        const part = splitString[i];
+        if (currentlyTeX) {
+            result += part;
+        } else {
+            result += `\\text{${part}}`;
+        }
```
anakaren-rojas, line 91:
```suggestion
result += currentlyTeX ? part : `\\text{${part}}`;
```

anakaren-rojas (approved):

packages/perseus/src/widgets/interactive-graphs/utils.test.ts
nishasy: I thought about that, but I wanted the description of each test to be explicit!

nishasy: Also I think the table would get really long with some of the descriptions + math strings. It would definitely be fewer lines, but I think I prefer the deliberateness of having them be separate tests in this particular case.

PR #1837, opened 2024-11-08
Use 'satisfies' instead of 'as' for WidgetExports<T> objects

jeremywiebe:
## Summary:

While working on some docs for Perseus, I noticed that we're still using `as` for our `WidgetExports` objects. `as` isn't great as we lost type safety when we use it. Instead, I've migrated them all to use `satisfies` which tells TypeScript that we think the object conforms to the type but that it should make sure. 

In doing this, I found a few places where our code/types were wrong and fixed them.

Issue: "none"

## Test plan:

### DIFF
--- a/packages/perseus/src/widgets/input-number/input-number.test.ts
+++ b/packages/perseus/src/widgets/input-number/input-number.test.ts
@@ -237,7 +237,7 @@ describe("input-number", function () {
         if (!transform) {
             throw new Error("transform not defined");
         }
-        const widgetProps = transform(editorProps, mockStrings);
+        const widgetProps = transform(editorProps);
         expect(_.has(widgetProps, "value")).toBe(false);
     });
 });
--- a/packages/perseus/src/widgets/numeric-input/numeric-input.test.ts
+++ b/packages/perseus/src/widgets/numeric-input/numeric-input.test.ts
@@ -93,9 +93,6 @@ describe("static function getOneCorrectAnswerFromRubric", () => {
         const singleAnswer =
             NumericInputWidgetExport.getOneCorrectAnswerFromRubric?.({
                 answers,
-                labelText: "",
-                size: "medium",
-                static: false,
                 coefficient: false,
             });
         expect(singleAnswer).toBe("12.2");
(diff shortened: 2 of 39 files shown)


packages/perseus/src/widgets/input-number/input-number.test.ts
```diff
@@ -237,7 +237,7 @@ describe("input-number", function () {
         if (!transform) {
             throw new Error("transform not defined");
         }
-        const widgetProps = transform(editorProps, mockStrings);
```
jeremywiebe, line 240: This function doesn't take a `strings` object at all.

packages/perseus/src/widgets/numeric-input/numeric-input.test.ts
```diff
@@ -93,9 +93,6 @@ describe("static function getOneCorrectAnswerFromRubric", () => {
         const singleAnswer =
             NumericInputWidgetExport.getOneCorrectAnswerFromRubric?.({
                 answers,
-                labelText: "",
-                size: "medium",
-                static: false,
```
jeremywiebe, line 98: `NumericInputWidgetExport.getOneCorrectAnswerFromRubric` only takes `answers` and `coefficient` props in the args object.

packages/perseus/src/widgets/input-number/input-number.test.ts
Myranae: When you go to the `WidgetTransform` type (by command clicking `transform` here in GoLand), it does have strings as a parameter.
image

But when you go to the transform for inputNumber, it just takes the input number widget options.
image

Should strings on the `WidgetTransform` type be optional? Why doesn't this cause an error?

jeremywiebe: You are correct that the `WidgetTransform` function type has a `strings` parameter. However, in this test we've picked the `transform` function from the `default` export of the `input-number.tsx` file. That function is compatible with the `WidgetTransform` function type, but it ignores the `strings` param. In Javascript/TypeScript, functions can chose to drop any arguments from the right side of the declaration and still be considered compatible. 

So this is valid:

```ts
type WidgetTransform = (options: WidgetOptions, strings: PerseusStrings, problemNum?: number) => any;

const transform: WidgetTransform = (options) => { return {...options} };
```

Myranae (approved):
Thanks for updating this!

PR #1839, opened 2024-11-08
[Locked Figure Labels] Util function to generate spoken math + use it within Locked Point aria labels

nishasy:
## Summary:
When auto-generating the aria labels for locked figures, we want it to use
words as if they were spoken rather than math expressions that might be
read incorrectly by the screen reader.

- Create a utility function using the MathJax speech engine that converts
  math details (TeX denoted by the `$...$` blocks) to spoken words.
- Use this utility within LockedPointSettings. (Async + useEffect because
  the speech engine is async. I had to use mocks in the tests for this.)

If this approach is given the okay, I'll go ahead and apply it to the
other locked figures as well in a following PR.

## Test plan:
`yarn jest packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/util.test.ts`
`yarn jest packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-point-settings.test.tsx`

Storybook
- `http://localhost:6006/iframe.html?args=&id=perseuseditor-widgets-interactive-graph--mafs-with-locked-figure-labels-all-flags&viewMode=story`
- Open the locked point settings
- Change the visible label to have a mix of TeX (with `$...$`) and non-TeX
- Press the "Auto-generate" button
- Confirm that the input changes to include spoken math words for the Tex
- Also try this with no labels and with multiple labels

image

### Loading state demo

https://github.com/user-attachments/assets/1ee80dee-3b10-46ab-b3f5-d296612b7a07


packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/util.ts
```diff
@@ -124,3 +125,20 @@ export function generateLockedFigureAppearanceDescription(
...
 }
+
+export async function generateSpokenMathDetails(mathString: string) {
+    const engine = await SpeechRuleEngine.setup("en");
+
+    // Replace everything between two $ signs with the spoken version
+    // of the math inside the $ signs.
+    // Example: "Circle with radius $\frac{1}{2}$" -> "Circle with radius one half"
+    const convertedSpeech = mathString.replace(
+        /\$(.*?)\$/g,
```
benchristel, line 136: This way of identifying TeX isn't going to work in all cases, because the TeX itself can contain dollar signs (either escaped, or in subexpressions).

For example: `This sandwich costs $\$50$.`

I might recommend using `@khanacademy/simple-markdown` (see perseus-markdown.tsx) to parse the string. You'll probably want to configure a separate parser instance in order to get TeX to render as speech. You could also use the parsing algorithm from `mathMatcher()`, defined in `pure-markdown/src/index.ts`

benchristel (commented):
Very cool! I left some suggestions inline for making the TeX parsing more robust.

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-point-settings.test.tsx
```diff
@@ -26,6 +26,14 @@ const defaultProps = {
 
 const defaultLabel = getDefaultFigureForType("label");
 
+// Mock the async function generateSpokenMathDetails
+jest.mock("./util", () => ({
+    ...jest.requireActual("./util"),
+    generateSpokenMathDetails: (input) => {
+        return `Spoken math details for ${input}`;
```
benchristel, line 33: Since the actual generateSpokenMathDetails function is async, do we want to return a Promise here to make sure the result is correctly being awaited? (Or maybe we trust TypeScript to tell us if we get that wrong?)

```suggestion
        return Promise.resolve(`Spoken math details for ${input}`);
```

packages/perseus/src/widgets/interactive-graphs/utils.ts
```diff
@@ -112,6 +112,10 @@ type ParsedNode = {
 function condenseTextNodes(nodes: Array<ParsedNode>): Array<ParsedNode> {
     const result: ParsedNode[] = [];
 
+    if (!nodes) {
```
benchristel, line 115: According to the types, this condition should never be met, because `nodes` is an array, and all arrays are truthy. I think we need to change the `nodes` parameter type to something like `ParsedNode[] | undefined` to match reality.

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-point-settings.tsx
```diff
@@ -104,28 +105,39 @@ const LockedPointSettings = (props: Props) => {
...
+    React.useEffect(() => {
+        /**
+         * Get a prepopulated aria label for the point, with the math
+         * details converted into spoken words.
+         *
+         * If the point has no labels, the aria label will just be
+         * "Point at (x, y)".
+         *
+         * If the point has labels, the aria label will be
+         * "Point at (x, y) with label1, label2, label3".
```
benchristel, line 120: I think this comment should be something like

```suggestion
         * If the point has labels, the aria label will be
         * "Point label1, label2, label3 at (x, y)".
```

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-point-settings.tsx
```diff
@@ -104,28 +105,39 @@ const LockedPointSettings = (props: Props) => {
...
-        return str;
-    }
+            const pointAppearance =
+                generateLockedFigureAppearanceDescription(pointColor);
+            str += pointAppearance;
+
+            return str;
+        }
+
+        getPrepopulatedAriaLabel().then(setPrepopulatedAriaLabel);
```
benchristel, line 139: There's a race condition here because the last `getPrepopulatedAriaLabel` promise to resolve might not be one from the most recent render. This could result in an aria label being generated with the wrong coords, labels, or pointColor.

I think we could guard against that race condition by doing this in the useEffect callback:

```ts
let canceled = false;
getPrepopulatedAriaLabel().then((label) => {
  !canceled && setPrepopulatedAriaLabel(label);
})

return () => canceled = true;
```

The annoying thing is that this is really hard to test.

What we really want here is something like a `useMemo`, but unfortunately `useMemo` doesn't work when the function we want to memoize is async. I don't think it would be too hard to write a `useAsyncMemo` hook ourselves, though.

Then we could just write this:

```ts
const prepopulatedAriaLabel = useAsyncMemo(
  /* initial value */ "",
  () => getPrepopulatedAriaLabel(coord, labels, pointColor),
  [coord, labels, pointColor],
);
```

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-point-settings.test.tsx
nishasy: That's a great point! I'll change it to the promise.

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-point-settings.tsx
nishasy: Good catch!

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-figure-aria.tsx
```diff
@@ -13,15 +13,23 @@ const {InfoTip} = components;
 
 type Props = {
     ariaLabel: string | undefined;
-    prePopulatedAriaLabel: string;
+    prePopulatedAriaLabel?: string;
+    getPrepopulatedAriaLabel?: () => Promise<string>;
     onChangeProps: (props: {ariaLabel?: string | undefined}) => void;
```
anakaren-rojas, line 18: Can this be moved up since `prePopulatedAriaLabel` and `getPrepopulatedAriaLabel` are optional?

anakaren-rojas (approved):

benchristel (approved):
LGTM!

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-figure-aria.tsx
nishasy: I believe we have a general style of putting function type props at the end of the props list in all the Khan repos.

PR #1842, opened 2024-11-08
Update function signature for getUserInputFromProps

MikeKlemarewski:
## Summary:
In my prior PR [where I updated this function signature](https://github.com/Khan/perseus/commit/614425b129dfbd04160215d9026a6666e03deddc), I wasn't aware that it was called outside of the component anywhere. It turns out in Web App, [there are calls to getUserInputFromProps](https://github.com/Khan/webapp/blob/master/services/static/javascript/coach-report-package/responses/utils.ts#L75) and this change to the function signature breaks those calls.

## Test plan:
- Ensure the tests pass

### DIFF
--- a/.changeset/tough-buses-tell.md
+++ b/.changeset/tough-buses-tell.md
@@ -2,4 +2,4 @@
 "@khanacademy/perseus": patch
 ---
 
-Fix function signature of getUserInputFromProps for radio component
+Fix function signature of getUserInputFromProps for radio widget
--- a/.changeset/tough-buses-tell.md
+++ b/.changeset/tough-buses-tell.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Fix function signature of getUserInputFromProps for radio component
--- a/packages/perseus/src/__tests__/widgets.test.ts
+++ b/packages/perseus/src/__tests__/widgets.test.ts
@@ -26,6 +26,158 @@ describe("Widget API support", () => {
             expect(Widget).toHaveProperty("getUserInputFromProps");
         },
     );
+
+    it("categorizer widget getUserInputFromProps should return the correct user input", () => {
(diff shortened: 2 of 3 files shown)


jeremywiebe (approved):
That Coach Reports package has caused us some amount of trouble. Going forward, we are working hard to not allow code outside Perseus to access the widget's directly.

.changeset/tough-buses-tell.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Fix function signature of getUserInputFromProps for radio component
```
jeremywiebe, line 5: **nit**: s/component/widget/ - in Perseus, widgets are a special thing that includes a React component, but conforms to a specific API. 
```suggestion
Fix function signature of getUserInputFromProps for radio widget
```

packages/perseus/src/__tests__/widgets.test.ts
```diff
@@ -26,6 +26,158 @@ describe("Widget API support", () => {
...
+
+    it("numeric-input widget getUserInputFromProps should return the correct user input", () => {
+        const Widget = Widgets.getWidget("numeric-input");
+
+        if (Widget && "getUserInputFromProps" in Widget) {
+            const props = {
+                currentValue: 42,
+            };
+
+            // @ts-expect-error - TS2339 - Property 'getUserInputFromProps' does not exist on type 'ComponentType<any>'.
```
jeremywiebe, line 142: Ouch, I guess because `getWidget` returns `React.ComponentType<any>`.  No changes needed, just wish we could type `getWidget` better.

jeremywiebe (approved):
Thanks for adding the tests.

packages/perseus/src/__tests__/widgets.test.ts
MikeKlemarewski: Yeah 😦

PR #1844, opened 2024-11-12
Udpate Radio's userInput and rubric to be as focused as possible

Myranae:
## Summary:
There were a number of properties on Radio's rubric and userInput objects that did not need to be there. This work refines the objects to only be what is required from the user and what is required for scoring. 

Issue: LEMS-2541

## Test plan:
- Confirm all checks pass


packages/perseus/src/widgets/radio/radio-validator.ts
```diff
@@ -21,23 +23,33 @@ function radioValidator(
...
-        userInput.numCorrect > 1 &&
-        numSelected !== userInput.numCorrect
-    ) {
+    const numCorrect: number = _.reduce(
+        rubric.choices,
+        function (memo, choice) {
+            return choice.correct ? memo + 1 : memo;
+        },
+        0,
+    );
```
handeyeco, line 32: When it's an option, I'd like to move away from underscore and use native JS; [reduce](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce) is native now.

```suggestion
    const numCorrect: number = rubric.choices.reduce(
        (acc, curr) => {
            return curr.correct ? acc + 1 : acc;
        },
        0,
    );
```

packages/perseus/src/widgets/radio/radio-validator.ts
```diff
@@ -21,23 +23,33 @@ function radioValidator(
...
+
+    if (numCorrect && numCorrect > 1 && numSelected !== numCorrect) {
```
handeyeco, line 34:
```suggestion
    if (numCorrect > 1 && numSelected !== numCorrect) {
```

I know you're just moving existing code, but `numCorrect` is a number and the only falsy number is 0 which is not greater than 1.

packages/perseus/src/widgets/radio/radio-validator.ts
```diff
@@ -21,23 +23,33 @@ function radioValidator(
...
+    let noneOfTheAboveSelected = false;
+    for (let i = 0; i < rubric.choices.length; i++) {
+        if (
+            rubric.choices[i].isNoneOfTheAbove &&
+            userInput.choicesSelected[i]
+        ) {
+            noneOfTheAboveSelected = true;
+            break;
+        }
+    }
```
handeyeco, line 50: Your code is clearer, but just because I like the new methods here is an alternative:

```suggestion
    let noneOfTheAboveSelected = rubric.choices.some((e, i) => (
         e.isNoneOfTheAbove && userInput.choicesSelected[i]
    )
```

handeyeco (approved):
Thanks Tamara, I like this. A good step forward both for our work and Perseus as a whole.

packages/perseus/src/widgets/radio/radio-validator.ts
Myranae: I also try to get rid of underscore stuff. Sometimes my brain refuses to realize I can update something I copy and pasted from another part of the codebase 😂 Thanks! Updated~

packages/perseus/src/widgets/radio/radio-validator.ts
Myranae: Fixed!

PR #1845, opened 2024-11-12
Add visible label and ARIA label to Dropdown widget

daniellewhyte:
## Summary:
This commit adds two new fields to the Dropdown widget: `ariaLabel` and `visibleLabel`. 

Issue: LIT-1424

## Test plan:

Screen reader walkthrough with `visibleLabel`:

https://github.com/user-attachments/assets/47f4ca21-09a6-4b82-b92a-81841c315f89

Screen reader walkthrough with `ariaLabel` only:
(This will probably be most instances as the dropdown widget is mostly used inline, and the visible label makes things too cramped.)

https://github.com/user-attachments/assets/ad1dfd76-5bc0-48e5-ab15-f57f2f92d0c3

Updates to widget editor:

Screenshot 2024-11-27 at 3 37 07 PM

You can also see examples in the course editor on this ZND: https://prod-znd-241122-danielle-dw2.khanacademy.org/devadmin/content/articles/dropdown-with-labels/x56e2ea23d30e405c


packages/perseus-editor/src/widgets/dropdown-editor.tsx
```diff
@@ -115,9 +123,41 @@ class DropdownEditor extends React.Component<Props> {
...
+                            recommend adding a label here to ensure your
+                            exercise is accessible. For more information on
+                            writing accessible labels, please see{" "}
+                            <a
+                                href="https://www.w3.org/WAI/tips/designing/#ensure-that-form-elements-include-clearly-associated-labels"
+                                target="_blank"
+                            >
+                                this article. If left blank, the value will
+                                default to "Select an answer".
+                            </a>
```
mark-fitzgerald, line 154: NIT: Perhaps linking the article this way:
```
Please see this
<a
    href="https://www.w3.org/WAI/tips/designing/#ensure-that-form-elements-include-clearly-associated-labels"
    target="_blank"
>
    article on writing accessible labels
</a>
for more information. If left blank, the value will default to "Select an answer".
```

packages/perseus/src/widgets/dropdown/dropdown.tsx
```diff
@@ -80,31 +86,58 @@ class Dropdown extends React.Component<Props> implements Widget {
...
+                            onChange={(value) =>
+                                this._handleChange(parseInt(value))
+                            }
+                            selectedValue={String(this.props.selected)}
+                            disabled={this.props.apiOptions.readOnly}
+                            aria-label={
+                                this.props.ariaLabel ||
+                                this.context.strings.selectAnAnswer
+                            }
+                            aria-labelledby={ids.get("dropdown-label")}
```
mark-fitzgerald, line 121: I wonder if this should be moved into the `aria-label` logic. In the case where both a visible label and an aria label are specified, the visible label would be read instead of the aria label (since `aria-labelledby` takes precedence). Perhaps `aria-label` could be changed to:
```
aria-label={
    this.props.ariaLabel ||
    this.props.visibleLabel ||
    this.context.strings.selectAnAnswer
}
```

packages/perseus-editor/src/widgets/dropdown-editor.tsx
```diff
@@ -115,9 +123,41 @@ class DropdownEditor extends React.Component<Props> {
...
                 </div>
+                <div className="dropdown-visible-label">
+                    <LabeledTextField
+                        label="Visible label"
+                        value={this.props.visibleLabel}
+                        onChange={this.onVisibleLabelChange}
+                    />
+                    <InfoTip>
+                        <p>Optional visible label</p>
+                    </InfoTip>
```
mark-fitzgerald, line 134: The tooltip icons seem oddly placed (being below the input field instead of after the label text). Instead of using `<LabeledTextField>`, would it be possible to have a separate label, followed by the `<InfoTip>` and then a regular `<TextField>`?

Tooltip Placement

mark-fitzgerald (approved):
Thanks so much for getting this accessible! Just a few comments.

PR #1847, opened 2024-11-13
[Locked Labels] Make `{$}` show up as regular symbols inside TeX

nishasy:
## Summary:
With the current implementation, all `$` symbols are removed when
they are used to denote TeX inside `$...$` blocks. There is no way
to denote something like `$50` within TeX with this crude implementation.

To remedy this, I'm using the simple markdown parser to correctly
and more sophisticatedly determine which blocks are math.

- `{$}` and `\$` both show up as a regular dollar sign in the labels

Issue: https://khanacademy.atlassian.net/browse/LEMS-2591

## Test plan:

Storybook
- Go to http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--mafs-with-locked-figure-labels-all-flags
- Use `{$}` within the locked label and locked figure labels,
  with a combination of TeX and non-TeX

Screenshot 2024-11-12 at 4 59 39 PM

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/utils.test.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/utils.test.ts
@@ -71,7 +71,10 @@ describe("replaceOutsideTeX", () => {
         const mathString = "x^2";
         const convertedString = replaceOutsideTeX(mathString);
 
-        expect(convertedString).toEqual("\\text{x^2}");
+        // For some reason, the parser starts a new text block once it hits
+        // the '^' character. The result is still the same though.
+        // This is functionally equal to "\\text{x^2}"
+        expect(convertedString).toEqual("\\text{x}\\text{^2}");
     });
 
     test("$s surrounding string", () => {
@@ -118,5 +121,24 @@ describe("replaceOutsideTeX", () => {
         expect(convertedString).toEqual(
             "\\text{Square }A\\text{ is }B\\text{ also}",
         );
+    });
+
+    test("with a real $ inside a regular string", () => {
+        const string = "This sandwich is \\$12";
+        const convertedString = replaceOutsideTeX(string);
+
+        // For some reason, the parser starts a new text block once it hits
+        // the '$' character. The result is still the same though.
+        // This is functionally equal to "\\text{This sandwich is $12}"
+        expect(convertedString).toEqual(
+            "\\text{This sandwich is }\\text{$}\\text{12}",
+        );
+    });
+
+    test("with a real $ inside a TeX string", () => {
+        const mathString = "This sandwich is ${$}12$";
+        const convertedString = replaceOutsideTeX(mathString);
+
+        expect(convertedString).toEqual("\\text{This sandwich is }{$}12");
     });
 });
--- a/packages/perseus/src/widgets/interactive-graphs/utils.test.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/utils.test.ts
@@ -71,10 +71,7 @@ describe("replaceOutsideTeX", () => {
         const mathString = "x^2";
         const convertedString = replaceOutsideTeX(mathString);
 
-        // For some reason, the parser starts a new text block once it hits
-        // the '^' character. The result is still the same though.
-        // This is functionally equal to "\\text{x^2}"
-        expect(convertedString).toEqual("\\text{x}\\text{^2}");
+        expect(convertedString).toEqual("\\text{x^2}");
     });
 
     test("$s surrounding string", () => {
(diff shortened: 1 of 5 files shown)


packages/perseus/src/widgets/interactive-graphs/utils.test.ts
```diff
@@ -119,4 +122,23 @@ describe("replaceOutsideTeX", () => {
...
         );
     });
+
+    test("with a real $ inside a regular string", () => {
+        const string = "This sandwich is \\$12";
+        const convertedString = replaceOutsideTeX(string);
+
+        // For some reason, the parser starts a new text block once it hits
+        // the '$' character. The result is still the same though.
+        // This is functionally equal to "\\text{This sandwich is $12}"
```
benchristel, line 132: I don't think screenreaders will read the text in exactly the same way if it's split up. [Specific example on the TeX playground](https://www.khanacademy.org/devadmin/tex#JTVDdGV4dCU3QlRoaXMlMjBzYW5kd2ljaCUyMGlzJTIwJTdEJTVDdGV4dCU3QiU1QyUyNCU3RCU1Q3RleHQlN0IxMiU3RCUyMCU1QyU1QyUwQSU1Q3RleHQlN0JUaGlzJTIwc2FuZHdpY2glMjBpcyUyMCU1QyUyNDEyJTdE)

VoiceOver on Edge reads "This sandwich is ispace dollar twelve" for the first line, and "This sandwich is dollar twelve" for the second line.

packages/perseus/src/widgets/interactive-graphs/utils.test.ts
```diff
@@ -119,4 +122,23 @@ describe("replaceOutsideTeX", () => {
...
+            "\\text{This sandwich is }\\text{$}\\text{12}",
+        );
+    });
+
+    test("with a real $ inside a TeX string", () => {
+        const mathString = "This sandwich is ${$}12$";
+        const convertedString = replaceOutsideTeX(mathString);
+
+        expect(convertedString).toEqual("\\text{This sandwich is }{$}12");
+    });
```
benchristel, line 143: How does this handle text with curly braces, backslashes or other TeX special characters in it? I think we might need to escape those:

```ts
    it("escapes curly braces", () => {
        const mathString = "Hello}{";
        const convertedString = replaceOutsideTeX(mathString);

        expect(convertedString).toEqual("\\text{Hello\\}\\{}");
    });

    it("escapes backslashes", () => {
        const mathString = "\\";
        const convertedString = replaceOutsideTeX(mathString);

        expect(convertedString).toEqual("\\text{\\\\}");
    });
```

nishasy: I think the parser assumes `{` and `}` to be part of TeX, so it separates it out before evaluating. It seems like the same thing with`\` and `^`. This results in some weird received strings.

For the second one, I think it only counts as one backslash, so the result would be `"\\text{\\}"`

nishasy: - `"Hello\\}\\{"` ==> `"\\text{Hello}\\text{}}\\text{{}"`
- `"\\"` ==> `"\\text{\\}"`

benchristel: I think both of those results are syntax errors in TeX:

- [example 1](https://www.khanacademy.org/devadmin/tex#JTVDdGV4dCU3QkhlbGxvJTdEJTVDdGV4dCU3QiU3RCU3RCU1Q3RleHQlN0IlN0IlN0Q=)
- [example 2](https://www.khanacademy.org/devadmin/tex#JTVDdGV4dCU3QiU1QyU3RA==)

We do need to escape the curly braces and backslashes to make these render correctly.

packages/perseus/src/widgets/interactive-graphs/utils.ts
```diff
@@ -1,3 +1,5 @@
+import * as SimpleMarkdown from "@khanacademy/pure-markdown";
```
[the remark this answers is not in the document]
anakaren-rojas: With this suggestion, line 86 would be changed to 
`const firstSection = parse(mathString)[0];`

benchristel (approved):
Thanks for the updates! This looks good to me.

packages/perseus/src/widgets/interactive-graphs/utils.ts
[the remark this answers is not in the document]
nishasy: It's true that that's the more streamlined way to do it, but I have a personal preference to keep it as `SimpleMarkdown.parse` because I feel that it makes it more discoverable in the IDE search. There are a bunch of different parse functions, so I like being very explicit about which package this one is coming from.

PR #1848, opened 2024-11-13
Rename widget validators to use scoring phrasing

Myranae:
## Summary:
In preparation of splitting out the valid/invalid logic from the scoring logic, this renames the widget validator functions and related code to use phrasing related to scoring instead of validation. The goal is to prevent confusion when we add actual validation logic to validate a user's input before sending to the server and right before scoring on the server.

Issue: LEMS-2593

## Test plan:
- Confirm all checks pass
- Confirm all widgets still work as expected via Storybook


packages/perseus/src/widgets/cs-program/score-cs-program.ts
```diff
@@ -1,9 +1,7 @@
 import type {PerseusScore} from "../../types";
 import type {PerseusCSProgramUserInput} from "../../validation.types";
 
-export function csProgramValidator(
-    state: PerseusCSProgramUserInput,
-): PerseusScore {
+export function scoreCSProgram(state: PerseusCSProgramUserInput): PerseusScore {
```
Myranae, line 4: Looks like this scoring file doesn't do the export default at the bottom. Should I update it to export like the others?

.eslintrc.js
jeremywiebe, line 1: Thanks for fixing the linting also!

packages/perseus/src/widgets/cs-program/score-cs-program.ts
jeremywiebe: Yes, please. We mix between `export default` and exporting named functions in Perseus, but we should at least be consistent across the scoring function/files.

packages/perseus/src/widgets/expression/score-expression.ts
```diff
@@ -28,13 +28,12 @@ import type {
  * - For each answer:
  *   ~ Try to validate the user's input against the answer. The answer is
  *     expected to parse.
- *   ~ If the user's input validates (the validator judges it "correct"), we've
- *     matched and can stop considering answers.
```
jeremywiebe, line 32: Did you delete this line by accident?

packages/perseus/src/widgets/__shared__/score-noop.test.ts
```diff
@@ -1,8 +1,8 @@
-import noopValidator from "./noop-validator";
+import scoreNoop from "./score-noop";
 
-describe("noop-validator", () => {
+describe("score-noop", () => {
     it("returns a pointless object", () => {
-        const result = noopValidator();
+        const result = scoreNoop();
```
handeyeco, line 5: nit: `score[Widget]` makes sense, but I don't think `scoreNoop` really makes sense. I think `noopScorer` is better?

packages/perseus/src/widgets/cs-program/score-cs-program.ts
handeyeco: Jeremy and I disagree; I think `export function` is better and he advocates for having it at the end. I think it's best to be consistent more than anything.

packages/perseus/src/widgets/expression/score-expression.ts
handeyeco, line 32: Did you mean to delete this line?

jeremywiebe (approved):
Thanks for slogging through this rename! :)

handeyeco (approved):

packages/perseus/src/widgets/expression/score-expression.ts
Myranae: Yep, I accidentally modified this comment before realizing it was related to khanAnswerType instead of the SSS validators. Returning it now!

Myranae: Nope, definitely was unintended. I accidentally modified this comment before realizing it was related to khanAnswerType instead of the SSS validators. Returning it now! Thanks for pointing it out!!

packages/perseus/src/widgets/__shared__/score-noop.test.ts
Myranae: I think I was going for consistency with the naming of the scoring related functions, but that seems fine with me. Might be a good idea to keep the file name the way it is to conform to the eslint rules though. What do you think?

Myranae: We talked in messages and decided to leave it as is so it is consistent and conforms to the eslint rules.

PR #1849, opened 2024-11-13
Table: Extract validation out of scoring

jeremywiebe:
## Summary:

This PR is a test run for extracting validation logic from scoring (using `table` as the test subject). I've used a pattern where the validator still returns a `PerseusScore`. This means that we can call the validator as the first step in scoring and return the score if it's `invalid`.  

Issue: LEMS-2606

## Test plan:

All tests should still pass.

### DIFF
--- a/packages/perseus/src/util/answer-types.ts
+++ b/packages/perseus/src/util/answer-types.ts
@@ -26,7 +26,7 @@ export type Score = {
     ungraded?: boolean;
 };
 
-/*
+/**
  * Answer types
  *
  * Utility for creating answerable questions displayed in exercises
@@ -69,7 +69,6 @@ export type Score = {
  * TODO(alpert): Think of a less-absurd name for createValidatorFunctional.
  *
  */
-
 const KhanAnswerTypes = {
     /*
      * predicate answer type
--- a/packages/perseus/src/util/answer-types.ts
+++ b/packages/perseus/src/util/answer-types.ts
@@ -11,7 +11,7 @@ import type {PerseusStrings} from "../strings";
 const MAXERROR_EPSILON = Math.pow(2, -42);
 
 type Guess = any;
-type Predicate = any;
+type Predicate = (guess: number, maxError: number) => boolean;
 
 // TOOD(kevinb): Figure out how this relates to KEScore in
 // perseus-all-package/types.js and see if there's a way to
@@ -624,11 +624,11 @@ const KhanAnswerTypes = {
         convertToPredicate: function (
             correctAnswer: string,
             options: any,
-        ): Predicate {
+        ): [predicate: Predicate, options: any] {
             const correctFloat = parseFloat(correctAnswer);
 
             return [
-                function (guess: Predicate, maxError: Predicate) {
+                function (guess, maxError) {
                     return Math.abs(guess - correctFloat) < maxError;
                 },
                 $.extend({}, options, {type: "predicate"}),
@@ -640,7 +640,6 @@ const KhanAnswerTypes = {
             strings: PerseusStrings,
         ): (arg1: Guess) => Score {
             return KhanAnswerTypes.predicate.createValidatorFunctional(
-                // @ts-expect-error - TS2556 - A spread argument must either have a tuple type or be passed to a rest parameter.
                 ...KhanAnswerTypes.number.convertToPredicate(
                     correctAnswer,
                     options,
--- a/packages/perseus/src/validation.types.ts
+++ b/packages/perseus/src/validation.types.ts
@@ -180,7 +180,14 @@ export type PerseusSorterUserInput = {
     changed: boolean;
 };
 
-export type PerseusTableRubric = PerseusTableWidgetOptions;
+export type PerseusTableRubric = {
+    // The number of rows to display
+    rows: number;
+    // The number of columns to display
+    columns: number;
+    // Translatable Text; A 2-dimensional array of text to populate the table with
+    answers: ReadonlyArray<ReadonlyArray<string>>;
+};
 
 export type PerseusTableUserInput = ReadonlyArray<ReadonlyArray<string>>;
 
--- a/packages/perseus/src/validation.types.ts
+++ b/packages/perseus/src/validation.types.ts
@@ -13,7 +13,6 @@ import type {
     PerseusOrdererWidgetOptions,
     PerseusPlotterWidgetOptions,
     PerseusRadioChoice,
-    PerseusTableWidgetOptions,
     PerseusGraphCorrectType,
 } from "./perseus-types";
 import type {InteractiveMarkerType} from "./widgets/label-image/types";
--- a/packages/perseus/src/validation.types.ts
+++ b/packages/perseus/src/validation.types.ts
@@ -180,10 +180,6 @@ export type PerseusSorterUserInput = {
 };
 
 export type PerseusTableRubric = {
-    // The number of rows to display
-    rows: number;
-    // The number of columns to display
-    columns: number;
     // Translatable Text; A 2-dimensional array of text to populate the table with
     answers: ReadonlyArray<ReadonlyArray<string>>;
 };
--- a/packages/perseus/src/widgets/table/score-table.ts
+++ b/packages/perseus/src/widgets/table/score-table.ts
@@ -13,16 +13,16 @@ import type {
 } from "../../validation.types";
 
 function scoreTable(
-    state: PerseusTableUserInput,
+    userInput: PerseusTableUserInput,
     rubric: PerseusTableRubric,
     strings: PerseusStrings,
 ): PerseusScore {
-    const validationResult = validateTable(state);
+    const validationResult = validateTable(userInput);
     if (validationResult.type === "invalid") {
         return validationResult;
     }
 
-    const supplied = filterNonEmpty(state);
+    const supplied = filterNonEmpty(userInput);
     const solution = filterNonEmpty(rubric.answers);
     if (supplied.length !== solution.length) {
         return {
--- a/packages/perseus/src/widgets/table/score-table.ts
+++ b/packages/perseus/src/widgets/table/score-table.ts
@@ -3,7 +3,7 @@ import _ from "underscore";
 import KhanAnswerTypes from "../../util/answer-types";
 
 import {filterNonEmpty} from "./utils";
-import {validateTable} from "./validate-table";
+import validateTable from "./validate-table";
 
 import type {PerseusStrings} from "../../strings";
 import type {PerseusScore} from "../../types";
--- a/packages/perseus/src/widgets/table/score-table.ts
+++ b/packages/perseus/src/widgets/table/score-table.ts
@@ -2,6 +2,9 @@ import _ from "underscore";
 
 import KhanAnswerTypes from "../../util/answer-types";
 
+import {filterNonEmpty} from "./utils";
+import {validateTable} from "./validate-table";
+
 import type {PerseusStrings} from "../../strings";
 import type {PerseusScore} from "../../types";
 import type {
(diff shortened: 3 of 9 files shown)


packages/perseus/src/util/answer-types.ts
```diff
@@ -26,7 +26,7 @@ export type Score = {
     ungraded?: boolean;
 };
 
-/*
+/**
```
jeremywiebe, line 29: Tiny change that surfaces all this comment goodness in VSCode and other editors when you use KhanAnswerTypes. 

## Before 

image


## After

image

packages/perseus/src/util/answer-types.ts
```diff
@@ -69,7 +69,6 @@ export type Score = {
  * TODO(alpert): Think of a less-absurd name for createValidatorFunctional.
  *
  */
-
```
jeremywiebe, line 72: This also needed to be removed so that the JSDoc comment was actually "attached" to `KhanAnswerTypes`.

packages/perseus/src/util/answer-types.ts
```diff
@@ -625,11 +624,11 @@ const KhanAnswerTypes = {
         convertToPredicate: function (
             correctAnswer: string,
             options: any,
-        ): Predicate {
```
jeremywiebe, line 628: This isn't returning a `Predicate` by itself. It's returning a tuple of two items (one being the Predicate and the other the options).

packages/perseus/src/util/answer-types.ts
```diff
@@ -641,7 +640,6 @@ const KhanAnswerTypes = {
             strings: PerseusStrings,
         ): (arg1: Guess) => Score {
             return KhanAnswerTypes.predicate.createValidatorFunctional(
-                // @ts-expect-error - TS2556 - A spread argument must either have a tuple type or be passed to a rest parameter.
```
jeremywiebe, line 644: Fixing the return type above, resolves this TypeScript error!

packages/perseus/src/validation.types.ts
```diff
@@ -180,7 +180,14 @@ export type PerseusSorterUserInput = {
     changed: boolean;
 };
 
-export type PerseusTableRubric = PerseusTableWidgetOptions;
+export type PerseusTableRubric = {
```
jeremywiebe, line 183: I "exploded" this type from `PerseusTableWidgetOptions` because `headers` aren't needed for validation or scoring. TypeScript will still ensure that objects of one sort are compatible with the other.

packages/perseus/src/widgets/table/score-table.ts
```diff
@@ -14,25 +17,13 @@ function scoreTable(
     rubric: PerseusTableRubric,
     strings: PerseusStrings,
 ): PerseusScore {
-    const filterNonEmpty = function (table: any) {
```
jeremywiebe, line 17: This function moved to `utils.ts` because it's needed by both the scoring and the new validation behaviour.

packages/perseus/src/widgets/table/score-table.ts
```diff
@@ -41,12 +32,13 @@ function scoreTable(
...
         };
     }
+
     const createValidator = KhanAnswerTypes.number.createValidatorFunctional;
     let message = null;
-    const allCorrect = _.every(solution, function (rowSolution) {
+    const allCorrect = solution.every(function (rowSolution) {
         for (let i = 0; i < supplied.length; i++) {
             const rowSupplied = supplied[i];
-            const correct = _.every(rowSupplied, function (cellSupplied, i) {
```
jeremywiebe, line 49: Tiny migrations away from underscore.

packages/perseus/src/util/answer-types.ts
Myranae: So interesting that it just needs that one asterisk! I suppose that indicates how to handle the comment, but definitely did not know this led to such a huge difference!

packages/perseus/src/widgets/table/validate-table.ts
```diff
@@ -0,0 +1,24 @@
+import {filterNonEmpty} from "./utils";
+
+import type {PerseusScore} from "../../types";
+import type {PerseusTableUserInput} from "../../validation.types";
+
+export function validateTable(state: PerseusTableUserInput): PerseusScore {
```
Myranae, line 6: When we refined types, we started referring to `state` as `userInput`. Would you be up for changing it here to follow that naming convention? 

I think you might have said at one point you preferred calling it state, though I could be misremembering, so if that is the better option, we might want to go back and change them all to be state to be consistent. `scoreTable` is still showing `state` because it got skipped during the refinement step (whoops)

packages/perseus/src/widgets/table/validate-table.test.ts
```diff
@@ -0,0 +1,33 @@
...
+        ];
+
+        // Act
+        const result = validateTable(userInput);
+
+        // Assert
+        expect(result).toHaveInvalidInput();
+    });
+
+    it("is valid, not not correct, if all cells are provided", () => {
```
Myranae, line 20:
```suggestion
    it("is valid, not correct, if all cells are provided", () => {
```

packages/perseus/src/validation.types.ts
Myranae: I don't think the other fields are used, so might be worth a try to remove them. You'd also have to remove them from the tests.

```suggestion
    // Translatable Text; A 2-dimensional array of text to populate the table with
    answers: ReadonlyArray<ReadonlyArray<string>>;
};
```

packages/perseus/src/validation.types.ts
```diff
@@ -180,7 +179,14 @@ export type PerseusSorterUserInput = {
...
+    // The number of rows to display
+    rows: number;
+    // The number of columns to display
+    columns: number;
+    // Translatable Text; A 2-dimensional array of text to populate the table with
+    answers: ReadonlyArray<ReadonlyArray<string>>;
+};
```
Myranae, line 189: I don't think the other fields are used, so might be worth a try to remove them. You'd also have to remove them from the tests. 

```suggestion
    
    // Translatable Text; A 2-dimensional array of text to populate the table with
    answers: ReadonlyArray<ReadonlyArray<string>>;
};
```

packages/perseus/src/widgets/table/validate-table.ts
jeremywiebe: If the type is `XyzUserInput` I think calling the variable or parameter `userInput` makes alot of sense. :)   I'll update this PR.

Myranae: Thank you! Also noticed that this isn't using a default export like the scoring functions do. Was that on purpose?

jeremywiebe: Oops, I'll fix that too! Nice catch.

packages/perseus/src/widgets/table/score-table.ts
```diff
@@ -41,12 +32,13 @@ function scoreTable(
             message: null,
         };
     }
+
     const createValidator = KhanAnswerTypes.number.createValidatorFunctional;
-    let message = null;
-    const allCorrect = _.every(solution, function (rowSolution) {
+    let message: string | null = null;
+    const allCorrect = solution.every(function (rowSolution) {
```
handeyeco, line 38: Love seeing the switch away from `underscore` when possible!

packages/perseus/src/widgets/table/score-table.ts
```diff
@@ -57,7 +49,6 @@ function scoreTable(
                 );
                 const result = validator(cellSupplied);
                 if (result.message) {
-                    // @ts-expect-error - TS2322 - Type 'string' is not assignable to type 'null'.
```
handeyeco, line 60: 🌈

packages/perseus/src/widgets/table/score-table.ts
```diff
@@ -10,29 +13,17 @@ import type {
...
-            return cell === "";
-        });
-    });
-    if (hasEmptyCell || !supplied.length) {
-        return {
-            type: "invalid",
-            message: null,
-        };
+    const validationResult = validateTable(userInput);
+    if (validationResult.type === "invalid") {
```
handeyeco, line 21: Micro nit, feel free to ignore: we're going to be doing this a lot. Worth `isInvalid(validationResult)`? Or needless misdirection?

packages/perseus/src/widgets/table/validate-table.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    if (hasEmptyCell || !supplied.length) {
+        return {
+            type: "invalid",
+            message: null,
+        };
+    }
+
+    // No points earned, but at least we're not invalid
+    return {type: "points", earned: 0, total: 0};
```
handeyeco, line 23: No action needed: this feels wrong, but I don't know what the alternative is. Maybe return `PerseusScore | null`?

packages/perseus/src/widgets/table/score-table.ts
```diff
@@ -41,12 +32,13 @@ function scoreTable(
...
-    let message = null;
-    const allCorrect = _.every(solution, function (rowSolution) {
+    let message: string | null = null;
+    const allCorrect = solution.every(function (rowSolution) {
         for (let i = 0; i < supplied.length; i++) {
             const rowSupplied = supplied[i];
-            const correct = _.every(rowSupplied, function (cellSupplied, i) {
+            const correct = rowSupplied.every(function (cellSupplied, i) {
                 const cellSolution = rowSolution[i];
                 const validator = createValidator(
```
handeyeco, line 43: Random thought unrelated to the PR: is calling this "validator" still correct? Or is it a scorer now? Or is it both still?

Myranae (approved):
Looks great! Was able to follow your example to set up validation for categorizer with this :) 

Thinking about validation, it seems like my brain expects the result of a validation check to be a boolean. I suppose that wouldn't work here because we need the validation function to work in multiple places and it will need to return the invalid message when it fails, but something I wanted to note.

packages/perseus/src/widgets/table/score-table.ts
Myranae: This validator is related to khanAnswerTypes I believe, not the scoring/validation logic we worked on. I think we decided not to update its name, though I wonder if it might still lead to confusion though.

packages/perseus/src/widgets/table/utils.ts
```diff
@@ -0,0 +1,14 @@
+import type {PerseusTableUserInput} from "../../validation.types";
+
+/**
+ * Filters the given table (modelled as a 2D array) to remove any rows that are
+ * completely empty.
+ *
+ * @returns A new table with only non-empty rows.
+ */
+export const filterNonEmpty = function (table: PerseusTableUserInput) {
```
Myranae, line 9: I noticed the input parameter here is `table`, not `userInput`. I think that was probably on purpose since this is a util function, but pointing it out just in case.

jeremywiebe: Good eye @Myranae.  In this case I named it "table" because we call it with both the use input but also the rubric's solution data. What I think should change is the type here to make it clearer that we aren't only accepting user input. 

image

packages/perseus/src/widgets/table/score-table.ts
jeremywiebe: I made a change based on your other comment about the return value for "success" (returning `null`). I think I like it that way.  I'll land this today but feel free to push back and we can easily try another pattern for returning success from these validators in a new PR. 

image

packages/perseus/src/widgets/table/validate-table.ts
jeremywiebe: See comment above. I'm trying returning `Extract<PerseusScore, {type: "invalid"}> | null`.

PR #1850, opened 2024-11-13
Export the WidgetPromptJSON type from the main package

MikeKlemarewski:
## Summary:
We'll want access to this type in web app. Export it so it is available.

## Test plan:
- Ensure tests pass


jeremywiebe (approved):
This makes sense to me.

PR #1851, opened 2024-11-13
KAS: Improve TypeScript types

jeremywiebe:
## Summary:

While working on splitting validation (empty checking) from scoring, I couldn't get types for `KAS` functions even though everything seemed to be in TypeScript. Well, the entry file `index.js` was still in JS and that blocked any TypeScript types from bubbling outside of the `@khanacademy/kas` package. I've renamed the entry file `index.js` => `index.ts` and that gets the types flowin' real good. I had to fix some type issues and a tiny bit of code but I think this is a big net improvement.

Issue: LEMS-2598

## Test plan:

`yarn typecheck`
`yarn test`


packages/kas/package.json
```diff
@@ -17,7 +17,7 @@
     },
     "module": "dist/es/index.js",
     "main": "dist/index.js",
-    "source": "src/index.js",
+    "source": "src/index.ts",
```
jeremywiebe, line 20: 🎉

packages/kas/src/__tests__/transforming.test.ts
```diff
@@ -9,29 +9,44 @@ expect.extend({
 
         return actual === expected
             ? {pass: true, message: () => ""}
-            : {pass: false, message: () => `${input} factors as ${reference}`};
+            : {
```
jeremywiebe, line 12: These custom assertions were stating that the input matches the expected value when they failed, which was super confusing. 

For some, I was able to use Jest's internal utils for printing the expected and actual values as well as providing a nice "over/under" diff. 

Ref: https://dev.to/zirkelc/improve-your-testing-with-custom-matchers-in-jest-2ibm

packages/kas/src/types.d.ts
```diff
@@ -17,6 +17,7 @@ export type ExpressionVars = {
     equalIgnoringCase: boolean;
 };
 
+// TODO: Convert to interface and make the various Expr's implement it?
 export type Expression = {
```
jeremywiebe, line 21: All of the custom classes in `nodes.ts` conform to this type, but implicitly. That's not obvious.

kevinb-khan (approved):
LGTM.  Thanks for changing `index.js` to `index.ts` and fixing all of the TypeScript errors that resulted from that change.

PR #1854, opened 2024-11-14
[Locked Figure Aria] Use spoken math in Locked Line settings autogen labels

nishasy:
## Summary:
When auto-generating the aria labels for locked figures, we want it to use
words as if they were spoken rather than math expressions that might be
read incorrectly by the screen reader.

- Use the `generateSpokenMathDetails` utility within LockedLineSettings,
  following the pattern used in LockedPointSettings for this.
- Update the locked line autogen function to include the labels
  on the locked points that define the locked line.

## Test plan:
- `yarn jest packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-line-settings.test.tsx`

Storybook
- Open the locked line settings
- Change the visible label to hvae a mix of TeX (with `$...$`) and non-TeX
- Press the "Auto-generate" button
- Confirm that the input changes to include spoken math words for the TeX
- Also try this with no labels, multiple labels, and the labels changed
  for the locked points that define the line

image


packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-line-settings.tsx
```diff
@@ -73,13 +74,34 @@ const LockedLineSettings = (props: Props) => {
...
-
-    function getPrepopulatedAriaLabel() {
+    /**
+     * Generate a prepopulated aria label for the line, with the math
+     * details converted into spoken words.
+     */
+    async function getPrepopulatedAriaLabel() {
         let visiblelabel = "";
+        let point1VisibleLabel = "";
+        let point2VisibleLabel = "";
```
benchristel, line 84: Why are these variables called `*VisibleLabel` if this function is generating an aria label?

...oh, I think I get it. The term "label" is overloaded. We're using "visible label" to contrast with "aria label". If we just called the point labels "labels" that might be confusing.

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-line-settings.tsx
```diff
@@ -73,13 +74,34 @@ const LockedLineSettings = (props: Props) => {
...
+
+        if (point2.labels && point2.labels.length > 0) {
+            point2VisibleLabel += ` ${point2.labels
+                .map((l) => l.text)
+                .join(", ")}`;
+        }
+
+        let str = await generateSpokenMathDetails(
+            `${capitalizeKind}${visiblelabel} from point${point1VisibleLabel} at (${point1.coord[0]}, ${point1.coord[1]}) to point${point2VisibleLabel} at (${point2.coord[0]}, ${point2.coord[1]})`,
+        );
```
benchristel, line 104: The duplication of the `labels.map((l) => l.text).join(", ")` logic makes me think there's a missing abstraction here. What exactly that abstraction should be probably depends on how labels are being constructed for other locked figure types. It would be nice if we could represent this logic in one place.

I'm imagining something like:

```ts
function describeLockedPoint(point: LockedPointType): string {
  const {coord, labels} = point;
  const visibleLabel = labels && labels.length > 0
    ? ` ${labels.map((l) => l.text).join(", ")}`
    : "";

  return `point${visibleLabel} at (${coord[0]}, ${coord[1]})`
}
```

but I'm not sure this would be suitable for all locked figures.

benchristel, line 104: I think there might be a bug here when the labels include dollar signs.

Because we're calling `generateSpokenMathDetails` on the concatenated string, an unescaped dollar sign in one label might pair with an unescaped dollar sign in another label to create a TeX expression where there shouldn't be one.

For example, if one point had the label `$1` and another point had the label `$2`, the string passed to `generateSpokenMathDetails` would be:

```
Line from point $1 at (0, 0) to point $2 at (1, 1)
```

And that would get interpreted as containing a TeX expression `1 at (0, 0) to point` which is not going to speechify nicely (because every letter will be interpreted as a separate variable).

We could fix this by calling generateSpokenMathDetails for each label individually. For the example above, that would mean we'd get TeX syntax errors for both labels due to the unescaped dollar signs.

That makes me wonder: how are we handling TeX syntax errors?

benchristel (commented):
I think there might be a bug when the labels contain incomplete TeX expressions (i.e. an odd number of unescaped dollar signs). I left some comments/questions inline.

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-line-settings.tsx
nishasy: Yeah, it's expressing the the locked (visible) label that's associated with this figure is being added to the aria label.

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-line-settings.tsx
nishasy: Alternatively, maybe we could have a function like `joinLockedFigureLabels` that takes in an array of LockedLabels and then gets used within here. I thought about abstracting away the whole description, but the description does change a bit for every figure, which is why I didn't end up going that way.

nishasy: A few points:
1. I don't think you can use unescaped dollar signs as labels, so I'm not sure how to recreate the case you're referring to.
2. I can still call `generateSpokenMathDetails` on each label individually. The reason I didn't do this to begin with is that if it's a math block, it doesn't include the space before, but if it's regular text, it is. So I'd have to add an extra bit of logic to handle this case.
3. TeX syntax errors show up really ugly in the preview with big red text. We don't handle them specifically.

image

catandthemachines: I would definitely be for creating a function for this. Especially since we could potentially re-use it in future implementations. Plus we can make calls to `joinLockedFigureLabels` and add a comment describing the string it's generating. 😊

catandthemachines: Interesting! Can we file a bug to at some point add an error message into the form for `Visible labels`? We don't need to inflate the scope of your work but I think this is a small change that would make this error more clear to content creators! ❤️

catandthemachines (changes requested):
Looking pretty good, would love to see an abstracted function for the label joining and a bug for the syntax errors <3

catandthemachines (approved):
Approved!

nishasy:
Note: I'm going to implement the feedback in a following PR so I can do it for all the different graph types without having to work with the stacked PRs.

PR #1856, opened 2024-11-14
[Locked Figure Labels] Fix bug that's causing the editor to crash when the label input is empty (and on load)

nishasy:
## Summary:
I need to add a null check after my $ TeX updates in https://github.com/Khan/perseus/pull/1847.

This is actually already added in a larger PR, but this fix needs to get out ASAP.

Issue: none

## Test plan:
- Go to http://localhost:6006/iframe.html?id=perseuseditor-widgets-interactive-graph--mafs-with-locked-figure-labels-all-flags&viewMode=story
- Backspace in the locked point visible label input field
- Confirm that there is no crash

### DIFF
--- a/.changeset/dry-moles-notice.md
+++ b/.changeset/dry-moles-notice.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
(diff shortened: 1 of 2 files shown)


benchristel (approved):
LGTM!

PR #1859, opened 2024-11-14
Addition of new alert when json is converted.

SonicScrewdriver:
## Summary:
This PR adds an alert banner to the editors to notify Content Editors when pre-existing content has been converted into modern widgets. (ie Input Number has become Numeric Input) 

## Screenshot:
Screenshot 2024-11-14 at 12 59 01 PM

## Test plan:
- Manual testing

### DIFF
--- a/packages/perseus-editor/src/article-editor.tsx
+++ b/packages/perseus-editor/src/article-editor.tsx
@@ -72,16 +72,21 @@ export default class ArticleEditor extends React.Component<Props, State> {
         super(props);
 
         // Check if the json needs to be converted
-        const conversionWarningRequired = conversionRequired(
-            props.json as PerseusRenderer,
-        );
+        const conversionWarningRequired =
+            props.json instanceof Array
+                ? props.json
+                      .map(conversionRequired)
+                      .reduce((p, v) => p && v, false)
+                : conversionRequired(props.json);
+
         let json = props.json;
 
         // Convert the json if needed
         if (conversionWarningRequired) {
-            json = Array.isArray(props.json)
-                ? props.json.map(convertDeprecatedWidgets)
-                : convertDeprecatedWidgets(props.json as PerseusRenderer);
+            json =
+                props.json instanceof Array
+                    ? props.json.map(convertDeprecatedWidgets)
+                    : convertDeprecatedWidgets(props.json);
         }
 
         this.state = {
@@ -439,7 +444,7 @@ export default class ArticleEditor extends React.Component<Props, State> {
                 {this.state.conversionWarningRequired && (
                     <div style={{marginBottom: 10}}>
                         <Banner
-                            text="Pre-existing Input Number Widgets have been converted to Numeric Inputs. Please review the changes before publishing."
+                            text="Deprecated Input Number Widgets were found, and have been automatically upgraded to Numeric Inputs. Please review the changes before publishing."
                             kind="warning"
                             layout="floating"
                         />
--- a/packages/perseus-editor/src/article-editor.tsx
+++ b/packages/perseus-editor/src/article-editor.tsx
@@ -6,6 +6,7 @@
 
 import {components, ApiOptions, iconTrash} from "@khanacademy/perseus";
 import {Errors, PerseusError} from "@khanacademy/perseus-core";
+import Banner from "@khanacademy/wonder-blocks-banner";
 import * as React from "react";
 import _ from "underscore";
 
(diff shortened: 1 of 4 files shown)


packages/perseus-editor/src/util/deprecated-widgets/modernize-widgets-utils.ts
```diff
@@ -14,17 +14,12 @@ const widgetRegExes = [/input-number \d+/]; // We can add more regexes here in t
 export const convertDeprecatedWidgets = (
     json: PerseusRenderer,
 ): PerseusRenderer => {
-    // If there's no widgets that require conversion, return the original json
-    if (!conversionRequired(json)) {
-        return json;
-    }
-
```
SonicScrewdriver, line 21: Don't need to check here any more, as we're doing it directly in the editors. This allows us to know when the json has been updated! :)

packages/perseus-editor/src/article-editor.tsx
```diff
@@ -64,11 +70,24 @@ export default class ArticleEditor extends React.Component<Props, State> {
...
+        );
+        let json = props.json;
+
+        // Convert the json if needed
+        if (conversionWarningRequired) {
+            json = Array.isArray(props.json)
+                ? props.json.map(convertDeprecatedWidgets)
+                : convertDeprecatedWidgets(props.json as PerseusRenderer);
+        }
+
```
SonicScrewdriver, line 86: I can clean this up

packages/perseus-editor/src/article-editor.tsx
```diff
@@ -64,11 +70,24 @@ export default class ArticleEditor extends React.Component<Props, State> {
 
     constructor(props: Props) {
         super(props);
+
+        // Check if the json needs to be converted
+        const conversionWarningRequired = conversionRequired(
+            props.json as PerseusRenderer,
+        );
```
jeremywiebe, line 77: This is probably not be safe. I'm pretty sure there are places in webapp where it passes in `ReadonlyArray<PerseusRenderer>` (essentially the `PeprseusArticle` type). I think it can be fixed fairly easily though with: 

```suggestion
        const conversionWarningRequired =
            props.json instanceof Array
                ? props.json
                      .map(conversionRequired)
                      .reduce((p, v) => p && v, false)
                : conversionRequired(props.json);
```

mark-fitzgerald (approved):

packages/perseus-editor/src/article-editor.tsx
```diff
@@ -64,11 +70,24 @@ export default class ArticleEditor extends React.Component<Props, State> {
...
+        let json = props.json;
+
+        // Convert the json if needed
+        if (conversionWarningRequired) {
+            json = Array.isArray(props.json)
```
jeremywiebe, line 82: If you use `props.json instanceof Array`. That will still do array checking but properly narrow `props.json` so you don't need the `as`.

packages/perseus-editor/src/article-editor.tsx
```diff
@@ -417,6 +436,15 @@ export default class ArticleEditor extends React.Component<Props, State> {
     render(): React.ReactNode {
         return (
             <div className="framework-perseus perseus-article-editor">
+                {this.state.conversionWarningRequired && (
+                    <div style={{marginBottom: 10}}>
+                        <Banner
+                            text="Pre-existing Input Number Widgets have been converted to Numeric Inputs. Please review the changes before publishing."
```
jeremywiebe, line 442: **nit**: "Pre-existing" threw me... where did they exist before? :) 

Perhaps "Deprecated Input Number widgets were found in this article! They have been automatically upgraded to Numeric Inputs. Please review the changes before publishing."

packages/perseus-editor/src/editor-page.tsx
```diff
@@ -246,6 +260,15 @@ class EditorPage extends React.Component<Props, State> {
 
         return (
             <div id="perseus" className={className}>
+                {this.state.conversionWarningRequired && (
+                    <div style={{marginBottom: 10}}>
+                        <Banner
+                            text="Pre-existing Input Number Widgets have been converted to Numeric Inputs. Please review the changes before publishing."
```
jeremywiebe, line 266: Same nit about "pre-existing".  😄

jeremywiebe (approved):
Makes sense. :)

packages/perseus-editor/src/article-editor.tsx
SonicScrewdriver: Oh great catch, thank you! I forgot to check for the sections as part of the conversionRequired check

PR #1861, opened 2024-11-14
[Numeric Input] - Show "Format Options" Tooltip as List

mark-fitzgerald:
## Summary:
Upon receiving focus, the Numeric Input field will show a tooltip indicating the format of the answer that it is expecting (assuming the input was configured with suggested formats). When multiple formats are indicated, they should be displayed as a list. Currently, the formats are indeed HTML list elements, but they don't display as list items (all formatting has been removed). This change fixes the visual formatting to display the items as a list, but only if there are two or more items. In the case of just one format example, the HTML list element is removed, and the content is displayed as regular text. Also, the first line in the tooltip is more of an introductory text than a format option, and therefore should not be included in the HTML list markup.

Issue: LEMS-2457

## Test plan:
**NOTE:** Only the HTML modifications can be testing locally (aka Storybook). The visual modifications must be verified in a ZND (or locally in Webapp containerized) because the styling that interferes with the list styling does not exist in Storybook.

### Storybook:
1. Launch Storybook
1. Navigate to Perseus Editor => Editor => [Demo](http://localhost:6006/?path=/story/perseuseditor-editorpage--demo)
1. Add a Numeric Input widget
1. Configure the widget to have just one format option
Storybook - 1 Format Option
1. Move focus to the input field in the preview area
   * The tooltip should **NOT** be a list
Storybook - 1 Format Tooltip
1. Configure the widget to have more than one format option
Storybook - 3 Format Option
1. Move focus to the input field in the preview area
   * The tooltip should be a list, with the first line of text **NOT** in the list
Storybook - 3 Format Tooltip

### Webapp (locally or in a [ZND](https://prod-znd-241118-markfitz-m06.khanacademy.org/)):
1. Log into the app with a valid account for use in editing exercises
1. Navigate to a Numeric Input example
   * Test Everything > Unit 2 > Lesson 2: Numeric input ([ZND](https://prod-znd-241118-markfitz-m06.khanacademy.org/internal-courses/test-everything/test-everything-2-without-mastery/te-numeric-input/e/numeric-input-exercise))
1. Edit the exercise
1. Configure the widget to have just one format option
Storybook - 1 Format Option
1. Move focus to the input field in the preview area (make sure the "Desktop" preview is selected)
   * The tooltip should **NOT** be a list
Webapp - 1 Format Tooltip
1. Configure the widget to have more than one format option
Storybook - 3 Format Option
1. Move focus to the input field in the preview area
   * The tooltip should be a list, with the first line of text **NOT** in the list
Webapp - 3 Format Tooltip


packages/perseus/src/widgets/numeric-input/numeric-input.test.ts
```diff
@@ -47,6 +47,20 @@ describe("numeric-input widget", () => {
...
+
+        // Act
+        await userEvent.type(
+            screen.getByRole("textbox", {hidden: true}),
+            incorrect,
+        );
+
+        // Assert
+        expect(renderer).toHaveBeenAnsweredIncorrectly();
+    });
```
mark-fitzgerald, line 62: This isn't new. I just re-organized the tests to group them better. If the first test was "accept correct answer", I would think the next test would be "reject an incorrect answer". This also enabled me to group the snapshot tests together.

packages/perseus/src/styles/perseus-renderer.less
```diff
@@ -541,24 +541,20 @@
     }
 }
 
-// TODO(joel) - remove?
-.perseus-tooltip {
```
mark-fitzgerald, line 545: This style wasn't really used, other than to be a base later on in `.perseus-formats-tooltip`. So, I just moved this styling to where it is used.

packages/perseus/src/styles/perseus-renderer.less
```diff
@@ -541,24 +541,20 @@
...
+.perseus-formats-tooltip {
     background: #fff;
+    color: #777;
     padding: 5px 10px;
     width: 240px;
 }
 
-// CSS is evil; let's overwrite some styles. T_T
```
mark-fitzgerald, line 551: No! CSS is not evil. Misunderstood, maybe. But definitely not evil.

packages/perseus/src/components/input-with-examples.tsx
```diff
@@ -146,9 +146,18 @@ class InputWithExamples extends React.Component<Props, State> {
...
-            return "- " + example;
-        }).join("\n");
+        const examplesContent =
+            this.props.examples.length <= 2
+                ? this.props.examples.join(" ") // A single item (with or without leading text) is not a "list"
+                : this.props.examples // 2 or more items should display as a list
+                      .map((example, index) => {
+                          // If the first example is bold, then it is most likely a heading/leading text.
+                          // So, it shouldn't be part of the list.
+                          return index === 0 && example.startsWith("**")
```
jeremywiebe, line 156: It looks like both remaining usages of `<InputWithExamples />` provides an examples list where the first item is indeed a bold heading ([here](https://github.com/Khan/perseus/blob/9c2289b3fb0e415c4cef979ab146dc583ef91b7d/packages/perseus/src/widgets/input-number/input-number.tsx#L181) and [here](https://github.com/Khan/perseus/blob/94eba15cacf904daabce378d080291613490c4a1/packages/perseus/src/widgets/numeric-input/numeric-input.tsx#L132)). 

Perhaps we can simplify this once `input-number` is truly gone because at that point we'll know exactly what is always passed in.

jeremywiebe (approved):
Nice!

packages/perseus/src/components/input-with-examples.tsx
mark-fitzgerald: That's the plan. We have follow-up tickets to refactor aspects of Numeric Input. Seeing as how `InputWithExamples` isn't used anywhere else, we are considering bringing this component into Numeric Input and simplifying the two code bases.

SonicScrewdriver (approved):
Seems straightforward - Thanks for all the cleanup work with the CSS!

packages/perseus/src/__tests__/__snapshots__/server-item-renderer.test.tsx.snap
```diff
@@ -82,13 +82,21 @@ exports[`server item renderer should snapshot: initial render 1`] = `
...
+                            >
+                              <strong>
+                                Your answer should be
+                              </strong>
+                               
+                            </div>
+                          </div>
+                          <div
+                            class="paragraph"
+                            data-perseus-paragraph-index="1"
```
catandthemachines, line 97: What is this `data-perseus-paragraph-index` attribute? Is it like our version of `tabIndex`?

catandthemachines (approved):
Looks pretty good. Just curious about the data-perseus attribute.

packages/perseus/src/__tests__/__snapshots__/server-item-renderer.test.tsx.snap
mark-fitzgerald: Question has been posted to the team.

PR #1862, opened 2024-11-14
Categorizer: Extract validation out of scoring

Myranae:
## Summary:
In order to complete scoring server-side, we need to separate out the validation logic from the scoring logic. This work separates those two functions and updates associated tests.

Issue: LEMS-2596

## Test plan:
- Confirm all checks pass
- Confirm the widget still behaves as it should.

### DIFF
--- a/packages/perseus/src/validation.types.ts
+++ b/packages/perseus/src/validation.types.ts
@@ -20,16 +20,14 @@ import type {Relationship} from "./widgets/number-line/number-line";
 
 export type UserInputStatus = "correct" | "incorrect" | "incomplete";
 
-export type PerseusCategorizerRubric = {
+export type PerseusCategorizerScoringData = {
     // The correct answers where index relates to the items and value relates
     // to the category.  e.g. [0, 1, 0, 1, 2]
     values: ReadonlyArray<number>;
-    // Translatable text; a list of items to categorize. e.g. ["banana", "yellow", "apple", "purple", "shirt"]
-    items: ReadonlyArray<string>;
-};
+} & PerseusCategorizerValidationData;
 
 export type PerseusCategorizerUserInput = {
-    values: PerseusCategorizerRubric["values"];
+    values: PerseusCategorizerScoringData["values"];
 };
 
 export type PerseusCategorizerValidationData = {
@@ -195,7 +193,7 @@ export type PerseusTableRubric = {
 export type PerseusTableUserInput = ReadonlyArray<ReadonlyArray<string>>;
 
 export type Rubric =
-    | PerseusCategorizerRubric
+    | PerseusCategorizerScoringData
     | PerseusCSProgramRubric
     | PerseusDropdownRubric
     | PerseusExpressionRubric
--- a/packages/perseus/src/validation.types.ts
+++ b/packages/perseus/src/validation.types.ts
@@ -24,6 +24,8 @@ export type PerseusCategorizerRubric = {
     // The correct answers where index relates to the items and value relates
     // to the category.  e.g. [0, 1, 0, 1, 2]
     values: ReadonlyArray<number>;
+    // Translatable text; a list of items to categorize. e.g. ["banana", "yellow", "apple", "purple", "shirt"]
+    items: ReadonlyArray<string>;
 };
 
 export type PerseusCategorizerUserInput = {
--- a/packages/perseus/src/validation.types.ts
+++ b/packages/perseus/src/validation.types.ts
@@ -31,6 +31,12 @@ export type PerseusCategorizerRubric = {
 export type PerseusCategorizerUserInput = {
     values: PerseusCategorizerRubric["values"];
 };
+
+export type PerseusCategorizerValidationData = {
+    // Translatable text; a list of items to categorize. e.g. ["banana", "yellow", "apple", "purple", "shirt"]
+    items: ReadonlyArray<string>;
+};
+
 // TODO(LEMS-2440): Can possibly be removed during 2440?
 // This is not used for grading at all. The only place it is used is to define
 // Props type in cs-program.tsx, but RenderProps already contains WidgetOptions
--- a/packages/perseus/src/widgets/categorizer/score-categorizer.ts
+++ b/packages/perseus/src/widgets/categorizer/score-categorizer.ts
@@ -1,3 +1,5 @@
+import validateCategorizer from "./validate-categorizer";
+
 import type {PerseusStrings} from "../../strings";
 import type {PerseusScore} from "../../types";
 import type {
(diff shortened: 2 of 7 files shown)


packages/perseus/src/validation.types.ts
```diff
@@ -24,6 +24,8 @@ export type PerseusCategorizerRubric = {
     // The correct answers where index relates to the items and value relates
     // to the category.  e.g. [0, 1, 0, 1, 2]
     values: ReadonlyArray<number>;
+    // Translatable text; a list of items to categorize. e.g. ["banana", "yellow", "apple", "purple", "shirt"]
+    items: ReadonlyArray<string>;
```
Myranae, line 28: At first, I was just using `values` for validation, but since `items` is already sent to the scoring function, I just added it back to `rubric` so `values` can be stripped away during the rubric stripping step and validation still works. Totally up for changing this however though.

packages/perseus/src/widgets/categorizer/validate-categorizer.ts
```diff
@@ -0,0 +1,28 @@
...
+            completed = false;
+        }
+    });
+    if (!completed) {
+        return {
+            type: "invalid",
+            message: strings.invalidSelection,
+        };
+    }
+    return null;
```
handeyeco, line 25: I know this is just moved code, but what about something like:

```suggestion
    let incomplete = rubric.items.some((_, i) => userInput.values[i] == null);
    
    if (incomplete) {
        return {
            type: "invalid",
            message: strings.invalidSelection,
        };
    }
    
    return null;
```

packages/perseus/src/widgets/categorizer/validate-categorizer.ts
```diff
@@ -0,0 +1,28 @@
+import type {PerseusStrings} from "../../strings";
+import type {PerseusScore} from "../../types";
+import type {
+    PerseusCategorizerRubric,
+    PerseusCategorizerUserInput,
+} from "../../validation.types";
+
+function validateCategorizer(
+    userInput: PerseusCategorizerUserInput,
+    rubric: PerseusCategorizerRubric,
```
handeyeco, line 10: CC @jeremywiebe 

I think this highlights a possible need for more even more clarification of types (or maybe just a helper):
- [Example using unions (seems cleaner)](https://www.typescriptlang.org/play/?#code/C4TwDgpgBASgrgIwE4EsDGUC8UDeAoKQqFYCAWwGcAuKAOzjIQiQG0BdAbjwF889RIUAIK0KAd2ZZcBImgD2SJBDTAa9Rs3Zde-cNHjJ0AdRIALEeMnYDqDADJhoiUi555o4FACGT5jQvOUvhEUPKKyqpQLAAcADRQAGzxAOzxAKzxAMzxAAzxAJxsPHzuFJ4AZhA26DTVGNjBRCTk1FFxiSnpWbkFRTqlnkx1tYi2JsDmvkhBMoTNlDQx8UlQqVAZUNlQeVCFsbOhCkoqi+0raxtbO4U8QA)
- [Example using utilities](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAIglgMwRAThAdgYwgHgCoA0UAqgHxQC8UACnJgNb5ECiAHpgDYCuAJrvRBAB7BFEJQBw0WVIBuAFDzQkKACUuAIxR1KUAN7yoRqHGAQAtgGcAXFHRdzG1AG0AuguNRMQlGkzBbe0cXd3kAX0VlaABBdEsAd1RdWgYcdS06IgAib18Ifyy5SPBodO1MAHVTAAshLmBYhKSqeCRUDGw0zXKiRsSUIvlvOOAoAEM4-ts+5v1DY1y-AKhnAA4iADYiAHYiAFYiAGYiAAYiAE5XcMVhy1HkMrpbR8qauobJ2YNPUwsbFfWUC2UF2UAOUGOUDOUEu1yGQhGUCcL2e3R0VG+xl+Vlsa02O32R1OF1cBHmRkW+WWeKBBLBRKhJPCQA)

handeyeco (approved):
I think this is good to ship as-is, but could you also get @jeremywiebe to look at this? Since this is one of the earlier ones we're doing, I want to make sure we're on the right track.

packages/perseus/src/widgets/categorizer/score-categorizer.ts
```diff
@@ -10,22 +12,17 @@ function scoreCategorizer(
     rubric: PerseusCategorizerRubric,
     strings: PerseusStrings,
 ): PerseusScore {
-    let completed = true;
+    const validationResult = validateCategorizer(userInput, rubric, strings);
+    if (validationResult) {
+        return validationResult;
+    }
```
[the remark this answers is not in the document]
Myranae: Updated!

packages/perseus/src/widgets/categorizer/validate-categorizer.test.ts
```diff
@@ -0,0 +1,23 @@
+import {mockStrings} from "../../strings";
+
+import validateCategorizer from "./validate-categorizer";
+
+import type {PerseusCategorizerRubric} from "../../validation.types";
+
+describe("validateCategorizer", () => {
+    it("tells the learner its not complete if not selected", () => {
```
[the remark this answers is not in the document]
Myranae: Added!

packages/perseus/src/widgets/categorizer/validate-categorizer.ts
Myranae: I'm all for simplifying. I'll try to simplify the validation logic in the future if it seems like a straightforward update. Was trying to leave as close to original as possible as we've mentioned not changing the logic, but I think just simplifying what's there is probably fine. Updated :)

Myranae:
@jeremywiebe @handeyeco This should now follow the format we discussed, with `rubric` now as `scoringData` and its type as the old rubric type unioned with the `validationData` type.

packages/perseus/src/validation.types.ts
jeremywiebe: We chatted on Slack, but we agreed to model the Rubric like this: 

```ts
export type PerseusCategorizerRubric = {
     // The correct answers where index relates to the items and value relates
     // to the category.  e.g. [0, 1, 0, 1, 2]
     values: ReadonlyArray<number>;
} PerseusCategorizerValidationData; 
```

We also agreed to rename the term `Rubric` to `ScoringData` in the Server-Side Scoring area in a future task. :)

jeremywiebe (approved):

packages/perseus/src/validation.types.ts
Myranae: The above comment I made was from an older version. I updated Rubric to be the following:
```
export type PerseusCategorizerScoringData = {
    // The correct answers where index relates to the items and value relates
    // to the category.  e.g. [0, 1, 0, 1, 2]
    values: ReadonlyArray<number>;
} & PerseusCategorizerValidationData;
```

I assumed we could change Rubric as we go and the ticket would be for updating any we missed. It helps my mental model to have the names updated while I'm working, but I can totally leave the name change to the very end if it's preferred to do them all at once?

jeremywiebe: I started renaming things to `ScoringData` in one of my PRs. Then as I thought about it, I thought it might be more confusing to have a partial migration state in `main`. I was going to leave things as "Rubric" until we're done this pass of validation splitting and then we could do a single pass over everything to update our terms from Rubric to ScoringData.  But I don't think it's a huge deal either way. :)

PR #1863, opened 2024-11-14
Enable dependabot updates for WB/WS (and group them)

jeremywiebe:
## Summary:

Juan and I were discussing keeping Perseus dependnecies up to date with webapp (especially WB and WS). This is a trial run to see if Dependabot will help us. It should open up a new PR any time WB or WS packages are released. And all WB or WS packages that received an update should be updated together in the same Perseus PR. 

Issue: "none"

## Test plan:

I suspect I'll need to land this and monitor it to test. Github action infra is notoriously difficult to test locally.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,3 +10,21 @@ updates:
       allow:
           - dependency-name: "@khanacademy/eslint-config"
           - dependency-name: "@khanacademy/eslint-plugin"
+      assignees:
+            - "@Khan/perseus"
+
+    # Grouped updates for Wonder Blocks and Wonder Stuff releases.
+    # This helps us to stay in sync with the latest releases of these packages.
+    - package-ecosystem: "npm"
+      directory: "/"
+      schedule:
+          interval: "daily"
+      assignees:
+          - "@Khan/perseus"
+      groups:
+          wonder-stuff:
+              patterns:
+                  - "@khanacademy/wonder-stuff-*"
+          wonder-blocks:
+              patterns:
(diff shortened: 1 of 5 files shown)


.github/dependabot.yml
```diff
@@ -10,3 +10,15 @@ updates:
...
+
+      # Grouped updates for Wonder Blocks and Wonder Stuff releases.
+      # This helps us to stay in sync with the latest releases of these packages.
+      groups:
+          wonder-stuff:
+              patterns:
+                  - "@khanacademy/wonder-stuff-*"
+          wonder-blocks:
+              patterns:
+                  - "@khanacademy/wonder-blocks-*"
```
jandrade, line 24: Can we filter out certain packages? I'm asking as we have a couple of WB deprecated packages that we usually skip when upgrading webapp (`wb-color` and `wb-spacing`).

It would be nice if we could skip them programmatically with dependabot so we never see these packages in here, as it would be noisy/counterproductive to have them included.

jeremywiebe: We don't have any usages of `wb-color` nor `wb-spacing` in Perseus today. I can exclude them here, but I think I'm going to add an ESLint rule to ensure it doesn't come back. :) 

image

```json
    "no-restricted-imports": [
        "error",
        "@khanacademy/wonder-blocks-color",
        "@khanacademy/wonder-blocks-spacing",
    ],
```

jandrade (approved):
This is great! also the Lint rule is super useful (I'm going to replicate it in `webapp`). 🚀

PR #1865, opened 2024-11-14
fix(LEMS-2617): Require `$` to display TeX in axis labels

anakaren-rojas:
## Summary:
- Requires use of `$` to render TeX by using the `replaceOutsideTeX` function. 
- Adds test to validate that both TeX and text are being rendered 

Issue: LEMS-2617

## Test plan:
1. Navigate to [Chromatic](https://650db21c3f5d1b2f13c02952-tqthmecfcf.chromatic.com/?path=/docs/perseuseditor-editorpage--docs)
2. In the x Label or y Label inputs under `Common Graph Settings` , add a fraction `1/2` and then replace with `$\frac{1}{2}$` 
### Expected Results
a. `1/2` renders as text
b. `$\frac{1}{2}$` renders as TeX
image

## Screen recording
Comparison with current functionality and updated functionality using original reported bug text
https://github.com/user-attachments/assets/aeea74d3-70f4-417a-8a52-012a12a29e88


packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
```diff
@@ -128,6 +128,17 @@ describe("MafsGraph", () => {
...
 
+    it("renders X axis label with TeX and Y axis with text", () => {
+        const basePropsWithTexLabels = {
+            ...getBaseMafsGraphProps(),
+            labels: ["$1/2$", "1/2"],
+        };
+
+        render(<MafsGraph {...basePropsWithTexLabels} />);
+        expect(screen.getByText("1/2")).toBeInTheDocument();
+        expect(screen.getByText("\\text{1/2}")).toBeInTheDocument();
```
benchristel, line 139: I was a little confused when I first read this test because `1/2` appears in both labels. It's hard to tell which label goes with which assertion.

I might prefer splitting this into two tests:

```ts
    it("renders TeX in axis labels", () => {
        const basePropsWithTexLabels = {
            ...getBaseMafsGraphProps(),
            labels: ["$1/2$", "$3/4$"],
        };

        render(<MafsGraph {...basePropsWithTexLabels} />);
        expect(screen.getByText("1/2")).toBeInTheDocument();
        expect(screen.getByText("3/4")).toBeInTheDocument();
    });

    it("renders plain text in axis labels", () => {
        const basePropsWithTexLabels = {
            ...getBaseMafsGraphProps(),
            labels: ["1/2", "3/4"],
        };

        render(<MafsGraph {...basePropsWithTexLabels} />);
        expect(screen.getByText("\\text{1/2}")).toBeInTheDocument();
        expect(screen.getByText("\\text{3/4}")).toBeInTheDocument();
    });
```

Alternatively, I think just making the X and Y labels different would help.

Up to you what you want to do here.

benchristel (approved):
LGTM! I left a suggestion for testing inline.

packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
anakaren-rojas: I was going back and forth between splitting them up and not splitting, but you're right, having separate tests to validate the functionality independently is probably the better move

PR #1867, opened 2024-11-15
[SR] Represent point order in interactive elements' aria labels

nishasy:
## Summary:
- Include the current point order in the points' aria labels

According to Caitlyn’s SRUX doc, each element on the interactive graph should be labeled in the following format:

```
Element Type, Order, Math Details, Visual Details, Instructions.
```

For example:

```
Point 1 at ( 1, 2 ). Angle approximately equal to 187°. Use the Arrow keys to move the element within the graph. Use the Delete key to remove the element from the graph.
```

Issue: https://khanacademy.atlassian.net/browse/LEMS-2620

## Test plan:
todo

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.tsx
@@ -8,6 +8,11 @@ import type {vec} from "mafs";
 
 type Props = {
     point: vec.Vector2;
+    /**
+     * Represents where this point stands in the overall point order.
+     * This is used to provide screen readers with context about the point.
+     */
+    currentPointOrder: number;
     onMove?: (newPoint: vec.Vector2) => unknown;
     onClick?: () => unknown;
     color?: string;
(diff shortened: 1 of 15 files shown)


packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.tsx
```diff
@@ -8,6 +8,11 @@ import type {vec} from "mafs";
 
 type Props = {
     point: vec.Vector2;
+    /**
+     * Represents where this point stands in the overall point order.
+     * This is used to provide screen readers with context about the point.
+     */
+    currentPointOrder: number;
```
benchristel, line 15: I'd prefer the name `sequenceNumber` for this prop. See: https://en.wikipedia.org/wiki/Sequence_number. `current` is confusing to me because props are always current.

Maybe the doc comment should also mention that this number is 1-based, and that the numbering is supposed to restart from 1 for each interactive figure in the graph. But I could also see an argument for not documenting any of that, since the MovablePoint component can't control how people use it. All we really need to say in this comment is that the number will be presented to screenreader users as-is.

benchristel (approved):
LGTM! I left a suggestion about naming and doc comments inline, but I don't think that needs another round of review.

PR #1869, opened 2024-11-15
Orderer: Extract validation out of scoring

Myranae:
## Summary:
To complete server-side scoring, we are separating out validation logic from scoring logic. This separates that logic and updates associated tests.

Issue: LEMS-2603

## Test plan:
- Confirm all checks pass

### DIFF
--- a/packages/perseus/src/widgets/orderer/score-orderer.test.ts
+++ b/packages/perseus/src/widgets/orderer/score-orderer.test.ts
@@ -6,7 +6,7 @@ import type {
     PerseusOrdererUserInput,
 } from "../../validation.types";
 
-describe("ordererValiator", () => {
+describe("scoreOrderer", () => {
     it("is correct when the userInput is in the same order and is the same length as the rubric's correctOption content items", () => {
         // Arrange
         const rubric: PerseusOrdererRubric =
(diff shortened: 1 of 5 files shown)


packages/perseus/src/widgets/orderer/score-orderer.test.ts
```diff
@@ -56,20 +56,4 @@ describe("ordererValiator", () => {
         // assert
         expect(result).toHaveBeenAnsweredIncorrectly();
     });
-
-    it("is invalid when the when the user has not started ordering the options and current is empty", () => {
```
jeremywiebe, line 60: I think it is worth retaining a unit test for the scorer that makes sure the score is `"invalid"` for this case. This ensures we're correctly calling and handling the result of the validator in the scorer.

jeremywiebe (approved):

handeyeco (approved):

packages/perseus/src/widgets/orderer/score-orderer.test.ts
Myranae: Added this back :) Do you think all scoring functions should retain a test confirming it can handle validation results? Also, I remember hearing about how it's ideal to make sure tests don't overlap so that multiple tests don't fail as a result of one piece of code failing (for example, validation). Is that why you mocked this out in your PRs?

Myranae: Updated the test to mock the validation and added a test confirming scoring handles passing validation with validation mocked.

PR #1871, opened 2024-11-15
Add and pass one regression test for radio widget parser

benchristel:
This PR is a "pilot" demonstrating how regression testing for the PerseusItem
parser will work. The regression test uses production data found by running the
[exhaustive test tool][1].

[1]: https://github.com/Khan/perseus/pull/1864

Issue: LEMS-2582

## Test plan:

`yarn test`


packages/perseus/src/util/parse-perseus-json/regression-tests/data/radio1.json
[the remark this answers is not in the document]
mark-fitzgerald: I would suggest a `comment` property, but since this tests for type checking, that would probably fail. Would it be worth adding an optional `comment` property as a valid field within all Perseus JSON - not in this PR, but maybe something at an opportune time?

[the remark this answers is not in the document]
benchristel: It would pass typechecking, actually. Extra object properties are ignored by the typechecker and passed through unchanged (since that is "safest" for forward compatibility, and in line with TypeScript's view of object types). My main worry is that the comment property might be confusing to future readers who might mistake it for part of the real data.

I changed the name of the file to be more descriptive. Do you think that's good enough?

packages/perseus/src/util/parse-perseus-json/regression-tests/parse-perseus-json-regression.test.ts
```diff
@@ -0,0 +1,20 @@
+import * as fs from "fs";
+import {join} from "path";
+
+import {parseAndTypecheckPerseusItem} from "../index";
+
+const dataFiles = fs.readdirSync(join(__dirname, "data"));
```
jeremywiebe, line 6: Nice. So adding a regression test in the future is as easy as creating a file in `regression-tests/data` and checking it in, right?

packages/perseus/src/util/parse-perseus-json/regression-tests/data/radio-missing-noneOfTheAbove.json
```diff
@@ -0,0 +1,48 @@
+{
```
jeremywiebe, line 1: Thinking out loud... I wonder if we want to attribute the source of this item? Perhaps not. We can always run a full regression pass to see if it's safe to remove some (old) parsing logic. 🤔

jeremywiebe (approved):

packages/perseus/src/util/parse-perseus-json/regression-tests/data/radio-missing-noneOfTheAbove.json
jeremywiebe: Doh, I see you and Mark basically discussed this above, I missed that. 

One option could be to use a JSONC parser (https://www.npmjs.com/package/jsonc-parser) which is built by Microsoft. That does start to diverge from how our production code does things though. 

I think the file naming convention is likely to serve us just fine for now.

PR #1872, opened 2024-11-16
ADR 773: Look for the shortest assessment item to repro each parse error

benchristel:
The exhaustive test tool for PerseusItem parsing outputs the `assessmentItem`s
that can't be parsed. The plan is to generate regression tests from these
assessment items.

Previously, the tool would output an arbitrary `assessmentItem` for each
failure.  This PR makes it keep the *shortest* `assessmentItem` instead. That
way, our regression test data won't be over-complicated.

Issue: none

## Test plan:

Run the exhaustive test tool on a subset of data.

### DIFF
--- a/packages/perseus/src/util/parse-perseus-json/perseus-parsers/perseus-item.ts
+++ b/packages/perseus/src/util/parse-perseus-json/perseus-parsers/perseus-item.ts
@@ -3,14 +3,11 @@ import {
     any,
     array,
     boolean,
-    constant,
     enumeration,
     number,
     object,
     record,
-    union,
 } from "../general-purpose-parsers";
-import {composeParsers} from "../general-purpose-parsers/compose-parsers";
 
 import {parseHint} from "./hint";
 import {parsePerseusRenderer} from "./perseus-renderer";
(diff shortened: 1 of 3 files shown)


packages/perseus/src/util/parse-perseus-json/perseus-parsers/perseus-item.ts
```diff
@@ -18,7 +21,13 @@ import type {Parser} from "../parser-types";
 export const parsePerseusItem: Parser<PerseusItem> = object({
     question: parsePerseusRenderer,
     hints: array(parseHint),
-    answerArea: record(enumeration(...ItemExtras), boolean),
+    answerArea: record(
+        enumeration(...ItemExtras),
+        composeParsers(
+            union(boolean).or(constant("multiple")).parser,
+            (val, ctx) => ctx.success(Boolean(val)),
```
benchristel, line 28: I couldn't find any code in Perseus or Webapp where an answerArea value of "multiple" had a special meaning. It always seems to be treated as a boolean.

jeremywiebe (approved):

PR #1873, opened 2024-11-16
[Bugfix] - Adjust number formatting string to not fail

mark-fitzgerald:
## Summary:
The formatting string for `srPointAtCoordinates` was missing an 's', which was causing a malformed string when extracted for use in Webapp.

### DIFF
--- a/packages/perseus/src/strings.ts
+++ b/packages/perseus/src/strings.ts
@@ -134,11 +134,11 @@ export type PerseusStrings = {
     removePoint: string;
     graphKeyboardPrompt: string;
     srPointAtCoordinates: ({
-        pointNum,
+        num,
         x,
         y,
     }: {
-        pointNum: number;
+        num: number;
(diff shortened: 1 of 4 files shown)


nishasy (approved):
Sorry for missing this!

packages/perseus/src/strings.ts
```diff
@@ -321,7 +321,7 @@ export const strings: {
     graphKeyboardPrompt: "Press Shift + Enter to interact with the graph",
     srPointAtCoordinates: {
         context: "Screenreader-accessible description of a point on a graph",
-        message: "Point %(pointNum) at %(x)s comma %(y)s",
+        message: "Point %(num)s at %(x)s comma %(y)s",
```
jeremywiebe, line 324: For posterity, the issue was the lack of trailing `s` and not the fact that the value was named `num`. :)

PR #1874, opened 2024-11-16
[Locked Figure Aria] Fix $ edge cases for spoken math aria labels

nishasy:
## Summary:
I received a couple suggestions in PR https://github.com/Khan/perseus/pull/1854.
I'm implmementing them here:

- Wrote a helper function that joins labels together so that we don't have to
  duplicate that code for every single locked figure.
- Stopped calling `generateSpokenMathDetails()` on the whole aria string
  for the locked figure. Instead, it's now being called within
  the new helper function so that the labels can individually be converted,
  and we don't have to worry about the edge case with two different labels
  having dollar signs and causing all the content in between to be counted
  as TeX when it shouldn't.
- Handled the case with a lone unescaped dollar sign, which should not count
  as TeX.

Note: The lone unescaped dollar isn't handled for the TeX inside the
graph itself yet. This will be in the next PR.

## Test plan:
`yarn jest`

Storybook
- Add two visible labels to the point, with one having "$1 and" as the label
  and the other having "$1" as the label.
- Press the autogenerate button and confirm that the "and" is normal and not "a n d".
  Also confirm that the dollar signs show up normally.
- Add two TeX labels and confirm that the auto-generated text is as expected.
- Repeat for all the other locked figures.


packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/util.ts
```diff
@@ -144,11 +151,59 @@ export async function generateSpokenMathDetails(mathString: string) {
...
+export async function joinLabelsAsSpokenMath(
+    labels: LockedLabelType[] | undefined,
+): Promise<string> {
+    if (!labels || labels.length === 0) {
+        return "";
+    }
+
+    const spokenLabelPromises = labels.map(async (label) => {
+        return await generateSpokenMathDetails(label.text);
+    });
```
benchristel, line 189: Minor: I don't think you need async/await here. You can just return the promise.

```suggestion
    const spokenLabelPromises = labels.map((label) => {
        return generateSpokenMathDetails(label.text);
    });
```

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/util.ts
```diff
@@ -144,11 +151,59 @@ export async function generateSpokenMathDetails(mathString: string) {
...
+ * Non-async mocked version of joinLabelsAsSpokenMath for tests.
+ */
+export function mockedJoinLabelsAsSpokenMathForTests(
+    labels: LockedLabelType[] | undefined,
+) {
+    if (!labels || labels.length === 0) {
+        return Promise.resolve("");
+    }
+
+    // Mock this so that each label's text says "joint" before it.
```
benchristel, line 206: Should this comment say "spoken"?

```suggestion
    // Mock this so that each label's text says "spoken" before it.
```

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/util.ts
```diff
@@ -144,11 +151,59 @@ export async function generateSpokenMathDetails(mathString: string) {
...
+            switch (piece.type) {
+                case "math":
+                    convertedSpeech += engine.texToSpeech(piece.content);
+                    break;
+                case "unescapedDollar":
+                    // If the unescaped dollar had a closing pair to define
+                    // math, it would have been caught by the "math" case above.
+                    // Since this unescaped dollar is caught here, we can
+                    // assume it is alone and used as as a literal dollar sign.
+                    convertedSpeech += "$";
```
benchristel, line 163: This change is from another PR, right? You might need to change the base branch for this PR in GitHub (or land the other PR and then rebase these changes on `main`).

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-vector-settings.test.tsx
```diff
@@ -30,9 +33,8 @@ const defaultLabel = getDefaultFigureForType("label");
 // Mock the async function generateSpokenMathDetails
 jest.mock("./util", () => ({
     ...jest.requireActual("./util"),
-    generateSpokenMathDetails: (input) => {
-        return Promise.resolve(`Spoken math details for ${input}`);
-    },
+    joinLabelsAsSpokenMath: (input) =>
+        mockedJoinLabelsAsSpokenMathForTests(input),
```
benchristel, line 37: I worry that this mocking makes the tests harder to understand, and will become tech debt once we enable our wrapped SpeechRuleEngine to read locale data from local files.

Can you add a TODO comment linked to https://khanacademy.atlassian.net/browse/LEMS-2616 so we remember to un-mock this?

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/util.ts
nishasy: There are two util functions that have a similar branch! The are stuff is in this file. The visual TeX stuff is in the other `utils` file.

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/util.ts
```diff
@@ -1,21 +1,20 @@
...
+    type LockedFigureFillType,
+    type LockedFigureType,
+    type LockedFunctionType,
+    type LockedLabelType,
+    type LockedLineType,
+    type LockedPointType,
+    type LockedPolygonType,
+    type LockedVectorType,
+    type LockedLineStyle,
+    mathOnlyParser,
```
benchristel, line 15: Why is the `mathOnlyParser` exported from `@khanacademy/perseus`? Could it live in `perseus-editor`?

benchristel (approved):
This is looking great! I'm a little confused about why `mathOnlyParser` is defined in the perseus package when it seems like we only need it in the editor. I commented on that inline.

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/util.ts
nishasy: We need it in both! It's in Perseus for the visual TeX itself, and we need it in the editor for the aria label util.

nishasy: (see [other PR](https://github.com/Khan/perseus/pull/1875))

catandthemachines (approved):
Approved!

PR #1876, opened 2024-11-18
Sorter: Extract validation out of scoring

Myranae:
## Summary:
To complete server-side scoring, we are separating out validation logic from scoring logic. This separates that logic for the sorter widget and updates associated tests.

Issue: LEMS-2605

## Test plan:
- Confirm all checks pass
- Confirm widget still works as expected via Storybook

### DIFF
--- a/packages/perseus/src/widgets/sorter/score-sorter.test.ts
+++ b/packages/perseus/src/widgets/sorter/score-sorter.test.ts
@@ -29,16 +29,4 @@ describe("scoreSorter", () => {
         const result = scoreSorter(userInput, rubric);
         expect(result).toHaveBeenAnsweredIncorrectly();
     });
-
-    it("is invalid when the user has not made any changes", () => {
-        const userInput: PerseusSorterUserInput = {
-            options: ["$15$ grams", "$55$ grams", "$0.005$ kilograms"],
-            changed: false,
-        };
-        const rubric: PerseusSorterRubric = {
-            correct: ["$0.005$ kilograms", "$15$ grams", "$55$ grams"],
-        };
-        const result = scoreSorter(userInput, rubric);
-        expect(result).toHaveInvalidInput();
-    });
 });
--- a/packages/perseus/src/widgets/sorter/score-sorter.test.ts
+++ b/packages/perseus/src/widgets/sorter/score-sorter.test.ts
@@ -8,6 +8,7 @@ import type {
 
 describe("scoreSorter", () => {
     it("is correct when the user input values are in the order defined in the rubric", () => {
+        // Arrange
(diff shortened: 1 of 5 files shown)


packages/perseus/src/widgets/sorter/score-sorter.test.ts
```diff
@@ -29,16 +29,4 @@ describe("scoreSorter", () => {
...
-        const userInput: PerseusSorterUserInput = {
-            options: ["$15$ grams", "$55$ grams", "$0.005$ kilograms"],
-            changed: false,
-        };
-        const rubric: PerseusSorterRubric = {
-            correct: ["$0.005$ kilograms", "$15$ grams", "$55$ grams"],
-        };
-        const result = scoreSorter(userInput, rubric);
-        expect(result).toHaveInvalidInput();
-    });
```
jeremywiebe, line 43: I think we could keep a variant of this test in each scorer's test suite.  But, we can use mocking to ensure that `validate` is called first... something like:

I think we could keep a variant of this test in each scorer's test suite.  But, we can use mocking to ensure that `validate` is called first... something like:

```ts
// Top of file 
import * as SorterValidator from './validate-sorter'

    // Inside describe() block...
    it("should score if validator passes", () => {
        // Mock validator saying "all good" 👍 
        const mockValidate = jest
            .spyOn(SorterValidator, "default")
            .mockReturnValue(null);


        const score = scoreSorter(userInput, rubric);

        // Assert
        expect(mockValidate).toHaveBeenCalled();
        expect(score).toHaveBeenAnsweredCorrectly();
    });

    it("should abort if validator returns invalid", () => {
        // Mock validator saying "all good" 👍 
        const mockValidate = jest
            .spyOn(SorterValidator, "default")
            .mockReturnValue({type: "invalid", message: null});

        const score = scoreSorter(userInput, rubric);

        // Assert
        expect(mockValidate).toHaveBeenCalled();
        expect(score).toHaveInvalidInput()
    });
```

jeremywiebe (approved):

handeyeco (approved):

packages/perseus/src/widgets/sorter/score-sorter.test.ts
```diff
@@ -7,38 +8,82 @@ import type {
...
+
+        // Act
         const result = scoreSorter(userInput, rubric);
+
+        // Assert
         expect(result).toHaveBeenAnsweredIncorrectly();
     });
 
-    it("is invalid when the user has not made any changes", () => {
+    it("should abort if validator returns invalid", () => {
```
Myranae, line 44: @jeremywiebe Do the below added tests look as expected? Thanks!

PR #1878, opened 2024-11-18
Expression: Extract validation from scorer

jeremywiebe:
## Summary:

Arguably, this validator is quite useless in that it does nothing. However, I was thinking that we'd be better off if all of our widgets had a validator, even if it does nothing, rather than have some widgets have a validator and some not. In my mind, we're better off with the same convention everywhere.

This PR introduces the concept of "validation data" that we discussed earlier today (November 18). This represents _only_ the data needed by the validator. It intersects that type into the ScoringData type (nee Rubric) to ensure that the scoring function has all the data it needs to score (but also validate).

  * `PerseusExpressionRubric` - the data needed to score the `expression` user input (Note: this will be renamed to `PerseusExpressionScoringData` in the near future)
  * `PerseusExpressionValidationData` - the data needed to validate the `expression` user input
  * `ExpressionWidgetOptions` - the data needed to render the `expression` widget 

In the `expression` widget case, the validation data is empty, but in other widgets, the data in this type would also need to be in both the rubric and widget options (as the widget will call the validator on the client with only the widget options in hand and the scorer will call the validator on the server with only the rubric in hand). 

Issue: LEMS-2598

## Test plan:

`yarn test`
`yarn typecheck`

### DIFF
--- a/packages/perseus/src/validation.types.ts
+++ b/packages/perseus/src/validation.types.ts
@@ -55,6 +55,8 @@ export type PerseusExpressionRubric = {
 
 export type PerseusExpressionUserInput = string;
 
+export type ExpressionValidationData = Empty;
+
 export type PerseusGroupRubric = PerseusGroupWidgetOptions;
 
 export type PerseusGradedGroupRubric = PerseusGradedGroupWidgetOptions;
--- a/packages/perseus/src/validation.types.ts
+++ b/packages/perseus/src/validation.types.ts
@@ -1,3 +1,32 @@
+/**
+ * This file contains types used for validation and scoring. The types abide by
+ * a naming convention so that they're easy to follow and that we remain
+ * consistent across all of the widgets.
+ *
+ * These types are:
+ *
+ * `Perseus<Widget>UserInput`: the data returned by the widget that the user
+ * entered. This is referred to as the 'guess' in some older parts of Perseus.
+ *
+ * `Perseus<Widget>ValidationData`: the data needed to do validation of the
+ * user input. Validation is the checks we can do both on the client-side,
+ * before submitting user input for scoring, and on the server-side when we
+ * score it. As such, it cannot contain any of the sensitive scoring data
+ * that would reveal the answer.
+ *
+ * `Perseus<Widget>ScoringData` (nee `Perseus<Widget>Rubric`): the data needed
+ * to score the user input. By convention, this type is defined as the set of
+ * sensitive answer data and then intersected with
+ * `Perseus<Widget>ValidationData`.
+ *
+ * For example:
+ * ```
+ * type Perseus<Widget>ScoringData = {
+ *     correct: string;
+ * } & Perseus<Widget>ValidationData;
+ * ```
+ */
+
 import type {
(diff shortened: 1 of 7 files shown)


packages/perseus/src/validation.types.ts
```diff
@@ -51,10 +51,12 @@ export type PerseusDropdownUserInput = {
 export type PerseusExpressionRubric = {
     answerForms: ReadonlyArray<PerseusExpressionAnswerForm>;
     functions: ReadonlyArray<string>;
-};
+} & PerseusExpressionValidationData;
 
 export type PerseusExpressionUserInput = string;
 
+export type PerseusExpressionValidationData = Empty;
```
handeyeco, line 58: I've been trying to remove things that are just aliases for `Empty`. I understand the impulse to have it though, so I'll ultimately defer to you (I just might forget and delete it in the future).

packages/perseus/src/widgets/expression/validate-expression.ts
```diff
@@ -0,0 +1,28 @@
...
+    validationData: PerseusExpressionValidationData,
+    strings: PerseusStrings,
+    locale: string,
+): Extract<PerseusScore, {type: "invalid"}> | null {
+    if (userInput === "") {
+        return {type: "invalid", message: null};
+    }
+
+    return null;
+}
```
handeyeco, line 26: Just my $0.02 CAD, my preference would be to make these consistent with what we did for the scorers - not include unused params in the definition:

``` TS
export type WidgetScorerFunction = (
    userInput: UserInput,
    rubric: Rubric,
    string?: PerseusStrings,
    locale?: string,
) => PerseusScore;

export type WidgetExports<
> = Readonly<{
    // ...
    scorer?: WidgetScorerFunction;
    // ...
}>;

function scoreSorter(
    userInput: PerseusSorterUserInput,
    rubric: PerseusSorterRubric,
): PerseusScore {
    // ...
}

export default {
    name: "sorter",
    displayName: "Sorter",
    widget: Sorter,
    isLintable: true,
    scorer: scoreSorter,
} satisfies WidgetExports<typeof Sorter>;
```

I just like that it's less noise which makes it easier to find uses of things and to get a tl;dr of what the function is doing by looking at the signature.

```suggestion
function validateExpression(
    userInput: PerseusExpressionUserInput,
): Extract<PerseusScore, {type: "invalid"}> | null {
    if (userInput === "") {
        return {type: "invalid", message: null};
    }

    return null;
}
```

Not a blocker and not something we need to figure out now.

handeyeco (approved):
Ship it

packages/perseus/src/validation.types.ts
```diff
@@ -1,3 +1,32 @@
...
+ * `Perseus<Widget>ValidationData`.
+ *
+ * For example:
+ * ```
+ * type Perseus<Widget>ScoringData = {
+ *     correct: string;
+ * } & Perseus<Widget>ValidationData;
+ * ```
+ */
+
```
Myranae, line 29: Love this! Thank you so much for this comment!

PR #1879, opened 2024-11-18
Fixing conflicts that arose from Input Number Conversion and Server Side Scoring projects

SonicScrewdriver:
## Summary:
While testing our latest changes prior to deploy, we discovered that `perseus_widgets.spec.ts` was suddenly failing. 

After reviewing the changes with @handeyeco, we determined that the scoring logic is now preceding the `PerseusRenderer` rendering. As a result, the json passed to `scoreNumericInput` has not been converted to remove deprecated widgets and results in attempts to access data that does not exist. 

This PR ensures that we're converting both our UserInput and Rubric data prior to passing it to our scorer functions. 

## Test plan:
- manual testing 
- Snapshot testing on webapp along with `perseus_widgets.spec.ts`


packages/perseus/src/util/deprecated-widgets/modernize-widgets-utils.ts
```diff
@@ -41,3 +41,20 @@ export const conversionRequired = (json: PerseusRenderer): boolean => {
...
+
+// Convert the user input data keys from input-number to numeric-input
+export const convertUserInputData = (
+    userInputMap: UserInputMap,
+): UserInputMap => {
+    const updatedUserInputMap = {...userInputMap};
+
+    for (const key of Object.keys(userInputMap)) {
+        if (key.includes("input-number")) {
+            const updatedKey = key.replace("input-number", "numeric-input");
```
SonicScrewdriver, line 53: Just realized this is input number specific. I'll move this logic to the appropriate file, and keep this function a little more generic for our future selves.

handeyeco (approved):

PR #1881, opened 2024-11-19
Table: Validation/Scoring add unit tests

jeremywiebe:
## Summary:

Adding two unit tests to ensure that our validation is called and handled properly in the scoring function.

Issue: LEMS-2598

## Test plan:

`yarn test`
`yarn typecheck`


.changeset/giant-vans-notice.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/perseus": minor
+---
+
+Introduces a validation function for the matrix widget (extracted from matrix scoring function).
```
Myranae, line 5: I think this might need to be updated :)

packages/perseus/src/widgets/matrix/validate-matrix.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Checks user input from the matrix widget to see if it is scorable.
+ *
+ * Note: The matrix widget cannot do any validation without the Scoring
+ * Data because of its use of KhanAnswerTypes as a core part of scoring.
+ *
+ * @see `scoreMatrix()` for more details.
+ */
+function validateMatrix(
```
handeyeco, line 15: Not going to lose sleep on it, but I feel like we should remove useless validators/scorers. I think it's just noise and at the call site we could just do:

``` TS
const validationError = validator?.(data, userInput)
if (validationError) {
    return beSad()
}
```

Alternatively we could make a `noopValidator` like we have a `noopScorer`, but honestly I'd like to remove `noopScorer` anyway.

handeyeco (approved):

.changeset/giant-vans-notice.md
jeremywiebe: Gah! I mixed commits from another PR.  I'll clean this PR up.

jeremywiebe:
> I'm confused between this and #1883

I made it confusing. 😞  There was a commit from 1883 mixed into this PR. I've removed it so this PR is now strictly some unit tests for the `table` widget.

PR #1883, opened 2024-11-19
Matrix: Extract validation from scorer

jeremywiebe:
## Summary:

This PR extracts validation from the `matrix`'s scorer function. In reality, it's an empty function, but it follows our conventions for having the scorer call a validator first. I've created the standard tests to ensure that scorer calls the validator.

Issue: LEMS-2602

## Test plan:

`yarn test`
`yarn typecheck`

### DIFF
--- a/packages/perseus/src/validation.types.ts
+++ b/packages/perseus/src/validation.types.ts
@@ -9,10 +9,10 @@
  * entered. This is referred to as the 'guess' in some older parts of Perseus.
  *
  * `Perseus<Widget>ValidationData`: the data needed to do validation of the
- * user input. Validation is the checks we can do both on the client-side,
- * before submitting user input for scoring, and on the server-side when we
- * score it. As such, it cannot contain any of the sensitive scoring data
- * that would reveal the answer.
+ * user input. Validation refers to the different checks that we can do both on
+ * the client-side (before submitting user input for scoring) and on the
+ * server-side (when we score it). As such, it cannot contain any of the
+ * sensitive scoring data that would reveal the answer.
  *
  * `Perseus<Widget>ScoringData` (nee `Perseus<Widget>Rubric`): the data needed
  * to score the user input. By convention, this type is defined as the set of
@@ -22,7 +22,8 @@
  * For example:
  * ```
  * type Perseus<Widget>ScoringData = {
- *     correct: string;
+ *     correct: string;  // Used _only_ for scoring
+ *     size: number;     // Used _only_ for scoring
  * } & Perseus<Widget>ValidationData;
  * ```
  */
--- a/packages/perseus/src/validation.types.ts
+++ b/packages/perseus/src/validation.types.ts
@@ -150,7 +150,9 @@ export type PerseusMatcherUserInput = {
 export type PerseusMatrixRubric = {
     // A data matrix representing the "correct" answers to be entered into the matrix
     answers: PerseusMatrixWidgetAnswers;
-};
+} & PerseusMatrixValidationData;
+
+export type PerseusMatrixValidationData = Empty;
 
 export type PerseusMatrixUserInput = {
     answers: PerseusMatrixRubric["answers"];
--- a/packages/perseus/src/widgets/matrix/score-matrix.test.ts
+++ b/packages/perseus/src/widgets/matrix/score-matrix.test.ts
@@ -1,6 +1,7 @@
 import {mockStrings} from "../../strings";
 
 import scoreMatrix from "./score-matrix";
+import * as MatrixValidator from "./validate-matrix";
 
 import type {
     PerseusMatrixRubric,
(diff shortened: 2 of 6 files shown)


packages/perseus/src/widgets/matrix/score-matrix.test.ts
```diff
@@ -1,13 +1,66 @@
...
     PerseusMatrixRubric,
     PerseusMatrixUserInput,
 } from "../../validation.types";
 
 describe("scoreMatrix", () => {
+    it("should be correctly answerable if validation passes", function () {
+        // Arrange
+        const mockValidator = jest
+            .spyOn(MatrixValidator, "default")
+            .mockReturnValue(null);
```
Myranae, line 16: Question: I'm curious: why do we need to spy on the validator and give a mock return value here? The validator is called as part of the`scoreMatrix` function, right? Ooo, or is this to bypass the actual validation function so the test doesn't take that function into account? Should this be done with all the other scoring tests?

packages/perseus/src/widgets/matrix/validate-matrix.ts
```diff
@@ -0,0 +1,30 @@
+import type {PerseusScore} from "../../types";
+import type {
+    PerseusMatrixUserInput,
+    PerseusMatrixValidationData,
+} from "../../validation.types";
+
+/**
+ * Checks user input from the matrix widget to see if it is scorable.
+ *
```
Myranae, line 9: Question: Would it be good to leave comments like this for the other validators or is this specifically because of the Note below this?

packages/perseus/src/widgets/matrix/validate-matrix.ts
```diff
@@ -0,0 +1,30 @@
...
+ * Note: The matrix widget cannot do much validation without the Scoring
+ * Data because of its use of KhanAnswerTypes as a core part of scoring.
+ *
+ * @see `scoreMatrix()` for more details.
+ */
+function validateMatrix(
```
Myranae, line 15: Question: Did you decide just to leave all the `createValidator` logic in the scoring function? Do you think this is something we might want to change in the future? Looks like some of it is for scoring, but some is for validation.

Myranae (approved):
Looks good! Just left some questions to consider :)

handeyeco (approved):
I like this PR better than #1881 😅

packages/perseus/src/widgets/matrix/score-matrix.test.ts
Myranae: [Jeremy left a comment about this on one of my PRs](https://github.com/Khan/perseus/pull/1876/files/a9351e96858891d2fa6bb178527ad5758429b83c#r1847397738)

packages/perseus/src/validation.types.ts
```diff
@@ -121,7 +121,9 @@ export type PerseusMatcherUserInput = {
 export type PerseusMatrixRubric = {
     // A data matrix representing the "correct" answers to be entered into the matrix
     answers: PerseusMatrixWidgetAnswers;
-};
+} & PerseusMatrixValidationData;
+
+export type PerseusMatrixValidationData = Empty;
```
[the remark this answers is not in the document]
jeremywiebe: I'll remove it (and others that were `Empty`). I've added a doc comment to the top of this file in [another PR](https://github.com/Khan/perseus/pull/1878/files#diff-016e5aa099f7f6d1600d8347f6227c3dcdd3fb9480581e7c6a9526f3d78a1037R1-R29).

packages/perseus/src/widgets/matrix/score-matrix.test.ts
```diff
@@ -1,13 +1,66 @@
 import {mockStrings} from "../../strings";
 
 import scoreMatrix from "./score-matrix";
+import * as MatrixValidator from "./validate-matrix";
```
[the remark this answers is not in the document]
jeremywiebe: Yes. When you want to mock a function on an imported module, you need to use the "star import" syntax so that you can spy on a single function from that module.

packages/perseus/src/widgets/matrix/validate-matrix.ts
jeremywiebe: 🤔  I think it's worth a short description.

packages/perseus/src/widgets/matrix/validate-matrix.ts
jeremywiebe: It's all left in the scoring function because there is no logic that can be moved into the validator that doesn't require the scoring data.

Myranae: It kind of looked like the `hasEmptyCell` check could possibly be moved out, but then you'd duplicate a lot of that.

PR #1884, opened 2024-11-19
Ensuring UserInput and Rubric widget keys match

SonicScrewdriver:
## Summary:
While performing regression testing on the latest changes related to a conflict between our Server Side Scoring and Input Number Conversion projects, I found an edge case bug that needed to be handled. 

This bug was only reproducible on a test course that had an `input-number 2` and no `input-number 1`. As a result of the key mismatch between the `perseusRenderData` and `userInputMap`, the exercise became blocked / locked and unable to be continued after attempting to score. (If you would like a link to this exercise, please reach out to me!)

This PR updates the conversion logic in `renderer-utils.ts` to update both the `perseusRenderData` and `userInputMap` at the same time, so that the widget keys are identical between the two formats. This should cover any additional edge cases (such as the presence of other numeric-inputs that might adjust the IDs of the convertedWidgets). 

Some additional updates were required to create the helper functions that can perform this work in tandem. 

Issue: LEMS-2624

## Test plan:
- manual testing 
- Snapshot testing in Webapp (fixes issue)

### DIFF
--- a/packages/perseus/src/renderer-util.ts
+++ b/packages/perseus/src/renderer-util.ts
@@ -1,8 +1,7 @@
 import Util from "./util";
 import {
     conversionRequired,
-    convertDeprecatedWidgets,
-    convertUserInputData,
+    convertDeprecatedWidgetsForScoring,
 } from "./util/deprecated-widgets/modernize-widgets-utils";
 import {getWidgetIdsFromContent} from "./widget-type-utils";
 import {getWidgetScorer} from "./widgets";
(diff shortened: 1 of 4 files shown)


packages/perseus/src/renderer-util.ts
```diff
@@ -66,14 +65,17 @@ export function scorePerseusItem(
...
-        : perseusRenderData;
-    const convertedUserInputMap = mustConvertData
-        ? convertUserInputData(userInputMap) // Convert deprecated user input data keys to their modern equivalents
-        : userInputMap;
+    if (mustConvertData) {
+        const {convertedRubric, convertedUserData} =
+            convertDeprecatedWidgetsForScoring(perseusRenderData, userInputMap);
+        convertedRenderData = convertedRubric;
+        convertedUserInputMap = convertedUserData;
+    }
```
SonicScrewdriver, line 79: I could do a ternary destruction here, but it looked really messy, imo. I preferred the simple if statement for simple legibility - but I'd be happy hear any alternative opinions/suggestions! 

```suggestion
    const {convertedRubric, convertedUserData} = conversionRequired(perseusRenderData)
        ? convertDeprecatedWidgetsForScoring(perseusRenderData, userInputMap)
        : {convertedRubric: perseusRenderData, convertedUserData: userInputMap};

    // There seems to be a chance that PerseusRenderer.widgets might include
    // widget data for widgets that are not in PerseusRenderer.content,
    // so this checks that the widgets are being used before scoring them
    const usedWidgetIds = getWidgetIdsFromContent(convertedRubric.content);
    const scores = scoreWidgetsFunctional(
        convertedRubric.widgets,
        usedWidgetIds,
        convertedUserData,
        strings,
        locale,
    );
```

handeyeco (approved):

PR #1888, opened 2024-11-20
Fixing issue with Input Numbers that have a value of 0 

SonicScrewdriver:
## Summary:
This PR fixes an issue with the Input Number Conversion project that specifically pertains to widgets that have a value of 0, as it results in a falsy situation. As a result of this issue, exercises that contain the offending widget settings would fail to render. 

This PR updates the logic to a safer key, and ensures that it is not undefined. 

# Test:
- Manual testing both in Storybook and Webapp to confirm that failing cases now render properly.

### DIFF
--- a/packages/perseus/src/widgets/numeric-input/numeric-input.tsx
+++ b/packages/perseus/src/widgets/numeric-input/numeric-input.tsx
@@ -348,7 +348,7 @@ const propUpgrades = {
         // If the initialProps has a value, it means we're upgrading from
         // input-number to numeric-input. In this case, we need to upgrade
         // the widget options accordingly.
-        if (initialProps.value) {
+        if (initialProps.simplify) {
             // If the answerType is not number or percent, we need to provide
             // the answer form for the numeric-input widget
             const provideAnswerForm =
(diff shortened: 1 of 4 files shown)


packages/perseus/src/widgets/numeric-input/numeric-input.tsx
```diff
@@ -345,10 +345,10 @@ const propsTransform = function (
 const propUpgrades = {
     /* c8 ignore next */
     "1": (initialProps: any): PerseusNumericInputWidgetOptions => {
-        // If the initialProps has a value, it means we're upgrading from
+        // If the initialProps has simplify, it means we're upgrading from
         // input-number to numeric-input. In this case, we need to upgrade
         // the widget options accordingly.
-        if (initialProps.value) {
+        if (initialProps.simplify) {
```
jeremywiebe, line 351: Given we were bitten by this type of "bare if()" check for `value`, I think it's much safer and clearer to use an explicit check against `undefined`.  It appears that `simplify` is never null or an empty string in our published content (at least `"en"`), but I think we should move away from these truthy/falsey type of checks given how dangerous they can be in the face of `0` and `""`.

jeremywiebe: BTW, I used this shell command against the downloaded content tree to check: 

```
$ rg '"type"\s*:\s*"input-number"' -l | xargs jq --slurp '.. | objects | select(has("widgets")) | .widgets[].options | select(has("value") and .value == 0) | .simplify' | sort | uniq
```

mark-fitzgerald (approved):
Subtle but important change. Great job tracking this down!

PR #1889, opened 2024-11-20
Turn on strictFunctionTypes in tsconfig and suppress errors

benchristel:
I noticed that I wasn't getting type errors when I thought I should, and it
turned out to be because of this setting.

Issue: none

## Test plan:

`yarn tsc`

### DIFF
--- a/packages/perseus/src/__tests__/renderer.test.tsx
+++ b/packages/perseus/src/__tests__/renderer.test.tsx
@@ -47,7 +47,9 @@ jest.mock("../translation-linter", () => {
 
 describe("renderer", () => {
     beforeAll(() => {
+        // @ts-expect-error: InputNumberExport is not assignable to type WidgetExports
         registerWidget("numeric-input", InputNumberExport);
+        // @ts-expect-error: RadioWidgetExport is not assignable to type WidgetExports
         registerWidget("radio", RadioWidgetExport);
     });
 
--- a/packages/perseus/src/widgets/categorizer/categorizer.tsx
+++ b/packages/perseus/src/widgets/categorizer/categorizer.tsx
@@ -325,5 +325,6 @@ export default {
         );
     },
     isLintable: true,
+    // @ts-expect-error: Type 'UserInput' is not assignable to type 'PerseusCSProgramUserInput'.
     scorer: scoreCategorizer,
 } satisfies WidgetExports<typeof Categorizer>;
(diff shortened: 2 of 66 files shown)


jeremywiebe (approved):
Thanks for enabling this. We have some work to do but I'd rather have this on with some suppressions then off. 👍

packages/perseus/src/__tests__/renderer.test.tsx
```diff
@@ -47,7 +47,9 @@ jest.mock("../translation-linter", () => {
 
 describe("renderer", () => {
     beforeAll(() => {
+        // @ts-expect-error: InputNumberExport is not assignable to type WidgetExports
         registerWidget("numeric-input", InputNumberExport);
+        // @ts-expect-error: RadioWidgetExport is not assignable to type WidgetExports
```
handeyeco, line 52: This is curious 🤔

packages/perseus/src/widgets/categorizer/categorizer.tsx
```diff
@@ -325,5 +325,6 @@ export default {
         );
     },
     isLintable: true,
+    // @ts-expect-error: Type 'UserInput' is not assignable to type 'PerseusCSProgramUserInput'.
```
handeyeco, line 328: This is a bummer but I think we saw this coming. I think we need to be using generics here instead of unions.

packages/perseus/src/widgets/table/score-table.ts
```diff
@@ -9,7 +9,7 @@ import type {PerseusStrings} from "../../strings";
 import type {PerseusScore} from "../../types";
 import type {
     PerseusTableRubric,
-    PerseusTableUserInput,
+    PerseusTableUserInput, UserInput,
```
handeyeco, line 12: This doesn't seem right, does it? I wonder why we're not getting a linting error or something..

tsconfig-common.json
```diff
@@ -17,7 +17,7 @@
         /* Type Checking */
         "strict": true,
         "strictNullChecks": true,
-        "strictFunctionTypes": false,
+        "strictFunctionTypes": true,
```
handeyeco, line 20: Do we know what the counter argument to this is? Like why would we have turned it off?

Also I think it's true by default so maybe we don't need this line (unless we want to explicitly list it).

benchristel: I don't know if there's any counterargument beyond "you get fewer type errors if you do less rigorous typechecking." IMO TS's strict settings should _always_ be enabled whenever possible.

https://www.typescriptlang.org/tsconfig/#strictFunctionTypes

Myranae (commented):
Thanks for figuring this out! I also agree with using all the typechecking that is possible with TypeScript, so glad to see it getting turned on.

PR #1890, opened 2024-11-20
[Locked Labels + Aria] Create math only parser to help parse TeX how we want

nishasy:
## Summary:
We want to be able to parse a string meant for TeX into its most basic
essential parts. To do this, we are creating a `mathOnlyParser()` parser.

Using the default `parse()` function from SimpleMarkdown breaks up
text into too many sections, and parsing from scratch can get very
complicated. The ideal way to do this is to use the existing `parserFor()`
function from SimpleMarkdown that lets us customize the parser.

This will be used in
- Generating the TeX for Locked Labels and Locked Figure labels so that
  it shows up correctly on the graph. Non-TeX should show up with regular
  text styling and allow spaces.
- Generating the aria label for locked figures with TeX labels. The speech
  rule engine will only be run on the math sections of the parsed output.

(and also https://khanacademy.atlassian.net/browse/LEMS-2591)

## Test plan:
`yarn jest packages/perseus/src/widgets/interactive-graphs/utils.test.ts`

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/utils.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/utils.ts
@@ -1,4 +1,5 @@
-import * as SimpleMarkdown from "@khanacademy/pure-markdown";
+import {parse, pureMarkdownRules} from "@khanacademy/pure-markdown";
+import SimpleMarkdown from "@khanacademy/simple-markdown";
 
 import {clampToBox, inset, MIN, size} from "./math";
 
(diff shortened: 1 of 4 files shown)


packages/perseus/src/widgets/interactive-graphs/utils.ts
```diff
@@ -81,7 +82,7 @@ export function isUnlimitedGraphState(
 export function replaceOutsideTeX(mathString: string) {
     // All the information we need is in the first section,
     // whether it's typed as "blockmath" or "paragraph"
-    const firstSection = SimpleMarkdown.parse(mathString)[0];
+    const firstSection = parse(mathString)[0];
```
nishasy, line 85: Not making any changes here, just updating the export. This whole function will be updated in a different PR.

anakaren-rojas (approved):
Overall lgtm

PR #1893, opened 2024-11-20
Removing createReactClass usage in 3 files.

catandthemachines:
## Summary:
Removing createReactClass usage in the following three files:
- drag-target.tsx
- sortable.tsx
- stateful-editor-page.tsx

Issue: LEMS-365

## Test plan:
Run "yarn test" and confirm no errors or regressions are introduced.
Run "yarn start" and look and view if there are any issues with any Perseus components.


packages/perseus-editor/src/components/sortable.tsx
```diff
@@ -117,57 +142,75 @@ const SortableArea = createReactClass({
...
             </ol>
         );
-    },
-});
+    }
+}
+
+type ItemProps = {
+    area: any;
+    component: any;
```
jeremywiebe, line 168: Can we type this as `React.ReactNode`?  As far as I can see there are no constraints on what this is except that it is something React can render.

packages/perseus-editor/src/components/sortable.tsx
```diff
@@ -117,57 +142,75 @@ const SortableArea = createReactClass({
...
-        }),
-        component: PT.node.isRequired,
-        dragging: PT.bool.isRequired,
-        draggable: PT.bool.isRequired,
-        index: PT.number.isRequired,
-    },
-    handleDragStart: function (e) {
+export class SortableItem extends React.Component<ItemProps> {
+    static defaultProps: DefaultItemProps = {
+        area: "",
```
jeremywiebe, line 185: I think it'd be better to not provide all these defaults. The `Sortable` that renders these items looks like it provides all of these props. By making defaults for these props, we make the prop optional, and force this component to have a good default.  In the case of `area`, the default is of the wrong type as it should be an object that provides three callback functions: `onDragEnter`, `onDragStart`, and `onDrop`.

packages/perseus-editor/src/stateful-editor-page.tsx
```diff
@@ -1,47 +1,49 @@
 /* eslint-disable react/no-unsafe */
-import createReactClass from "create-react-class";
-import PropTypes from "prop-types";
 import * as React from "react";
 import _ from "underscore";
 
 import EditorPage from "./editor-page";
 
+type Props = {
+    componentClass: any;
```
jeremywiebe, line 8: I think you can type this as `React.ComponentType`.

jeremywiebe (changes requested):
Most of this is great. 🎉  

Requesting changes for the comment about default props not being necessary for the `SortableItem` changes.

packages/perseus-editor/src/components/drag-target.tsx
```diff
@@ -22,48 +22,61 @@
...
+
+    handleDragEnter(e) {
         this.setState({dragHover: this.props.shouldDragHighlight(e)});
-    },
-    render: function () {
-        const opacity = this.state.dragHover ? {opacity: 0.3} : {};
+    }
+
+    render() {
+        const opacity = this.dragHover ? {opacity: 0.3} : {};
```
benchristel, line 79: This should be `this.state.dragHover`. Then you don't need the `dragHover: any;` instance variable above (which is never set).

```suggestion
        const opacity = this.state.dragHover ? {opacity: 0.3} : {};
```

packages/perseus-editor/src/components/drag-target.tsx
benchristel, line 1: Are you sure `DragTarget` isn't dead code? I only see one usage, in `perseus-editor/src/editor.tsx`. That usage doesn't pass the `component` prop, so it seems like that `DragTarget` would throw if it were ever rendered.

packages/perseus-editor/src/components/sortable.tsx
```diff
@@ -1,10 +1,24 @@
 import {css, StyleSheet} from "aphrodite";
-import createReactClass from "create-react-class";
-import PropTypes from "prop-types";
 import * as React from "react";
 import ReactDOM from "react-dom";
 
-const PT = PropTypes;
+// Need to update these values.
```
benchristel, line 5: Could this be a TODO comment? Or are you planning to resolve it before this PR lands?

```suggestion
// TODO(catjohnson): update these values.
```

packages/perseus-editor/src/components/sortable.tsx
```diff
@@ -1,10 +1,24 @@
...
+type Props = {
+    className?: string;
+    components: any[];
```
benchristel, line 8:
```suggestion
    components: React.ReactElement[];
```

packages/perseus-editor/src/components/sortable.tsx
```diff
@@ -1,10 +1,24 @@
...
+    verify: (i: any) => boolean;
+};
+
+type DefaultProps = {
+    verify: Props["verify"];
+};
+
+type State = {
+    dragging: number;
+    components: any[];
```
benchristel, line 20:
```suggestion
    components: React.ReactElement[];
```

packages/perseus-editor/src/components/sortable.tsx
```diff
@@ -13,54 +27,61 @@ const PT = PropTypes;
...
+
+    onDragEnter(enterIndex) {
         // When a label is first dragged it triggers a dragEnter with itself,
         // which we don't care about.
         if (this.state.dragging === enterIndex) {
             return;
         }
 
-        const newComponents = this.state.components.slice();
+        const newComponents = this.props.components.slice();
```
benchristel, line 84: Why did this line change from `this.state.components` to `this.props.components`? I think we might need to use the state's components to get the most up-to-date value (since `this.state.components` is set below on line 95).

packages/perseus-editor/src/components/sortable.tsx
```diff
@@ -117,9 +141,10 @@ const SortableArea = createReactClass({
             dragItem.removeEventListener("dragstart", this._listenEvent);
             dragItem.removeEventListener("drop", this._cancelEvent);
         }
-    },
-    render: function () {
-        const sortables = this.state.components.map((component, index) => (
+    }
+
+    render() {
+        const sortables = this.props.components.map((component, index) => (
```
benchristel, line 147: Same here - should we be using `this.state.components`?

benchristel (commented):
Thanks for doing this upgrade! I left a few questions and suggestions inline. In particular, I'd like to confirm that it's safe to use `this.props.components` instead of `this.state.components` in `sortable.tsx` before this lands.

benchristel (approved):
LGTM!

packages/perseus-editor/src/components/sortable.tsx
```diff
@@ -1,59 +1,84 @@
...
+    dragging: number;
+    components: React.ReactElement[];
+};
 
 /**
+ * TODO(LEMS-2667): 11/26/24, at the time of writing this comment
+ * it has been identified that this file has been broken long before
+ * the refactoring of createReactClass. Future implementation need
+ * to determine how to fix this functionality or deprecate it.
+ * *
```
jeremywiebe, line 27: Nice, thanks for leaving this comment here.

packages/perseus-editor/src/components/sortable.tsx
```diff
@@ -134,40 +163,45 @@ const SortableArea = createReactClass({
...
-        component: PT.node.isRequired,
-        dragging: PT.bool.isRequired,
-        draggable: PT.bool.isRequired,
-        index: PT.number.isRequired,
-    },
-    handleDragStart: function (e) {
+export class SortableItem extends React.Component<ItemProps> {
+    constructor(props) {
+        super(props);
+    }
```
jeremywiebe, line 181: When you are just calling `super(props);` in a constructor, you can elide it. 
```suggestion
```

packages/perseus-editor/src/components/drag-target.tsx
```diff
@@ -22,53 +22,75 @@
...
+
+type DefaultProps = {
+    component: Props["component"];
+    shouldDragHighlight: Props["shouldDragHighlight"];
+};
+
+type State = {
+    dragHover: boolean;
+};
+export class DragTarget extends React.Component<Props, State> {
```
jeremywiebe, line 44: This is already being exported as the `default` export. I think we should do one or the other. Maybe remove this and just leave the default export?  I'll leave it for you to decide. :)

packages/perseus-editor/src/components/sortable.tsx
```diff
@@ -134,40 +163,45 @@ const SortableArea = createReactClass({
...
-            onDragStart: PT.func.isRequired,
-            onDrop: PT.func.isRequired,
-        }),
-        component: PT.node.isRequired,
-        dragging: PT.bool.isRequired,
-        draggable: PT.bool.isRequired,
-        index: PT.number.isRequired,
-    },
-    handleDragStart: function (e) {
+export class SortableItem extends React.Component<ItemProps> {
```
jeremywiebe, line 178: This class is the `default` export so we don't need to export this as a named export also.

jeremywiebe (approved):
This is looking great. I'd recommend resolving the two comments about `default` vs. named exports, but otherwise you're good to go!  Thanks for all the cleanup here Cat!

PR #1894, opened 2024-11-20
[SR] Add comment about creating translation tickets

nishasy:
## Summary:
Add a comment to the `strings.ts` file so it's easy
to keep track of which strings will need to be
finalized before translation tickets are made.

We want to wait until all the interactive element
strings for interactive graph have been written
and confirmed before creating their respective
translation tickets.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2660

## Test plan:
N/A

### DIFF
--- a/packages/perseus/src/strings.ts
+++ b/packages/perseus/src/strings.ts
@@ -323,12 +323,17 @@ export const strings: {
     graphKeyboardPrompt: "Press Shift + Enter to interact with the graph",
     closePolygon: "Close shape",
     openPolygon: "Re-open shape",
+    srInteractiveElements: "Interactive elements: %(elements)s",
+    srNoInteractiveElements: "No interactive elements",
+
+    // TODO(LEMS-2660): The following strings are ones that will need
+    // translation tickets after all interactive graph SR strings have
+    // been finalized. Remove this comment after the tickets have been
+    // created.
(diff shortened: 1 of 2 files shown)


packages/perseus/src/strings.ts
```diff
@@ -323,12 +323,17 @@ export const strings: {
...
+    // translation tickets after all interactive graph SR strings have
+    // been finalized. Remove this comment after the tickets have been
+    // created.
     srPointAtCoordinates: {
         context: "Screenreader-accessible description of a point on a graph",
         message: "Point %(num)s at %(x)s comma %(y)s",
     },
-    srInteractiveElements: "Interactive elements: %(elements)s",
-    srNoInteractiveElements: "No interactive elements",
 };
```
nishasy, line 337: would it be better to just not put the strings here at all? I can remove the strings from this section so that they don't get auto-flagged for translations, and I can move this comment up to the `PerseusStrings` section instead so we can remember to bring them down here later?

nishasy: Turns out that it's fine if we do it this way, since strings don't get translated until they're specifically requested anyway.

catandthemachines (approved):
LGTM!

benchristel (approved):
LGTM! Thank you!

PR #1895, opened 2024-11-20
Migrate PerseusItem.answerArea, removing unused fields

benchristel:
`answerArea` sometimes includes `options` and `type` fields. `type` always
seems to be set to `multiple` if it's present. I couldn't find evidence in
Perseus or webapp that this data is used anywhere, so I am removing it.
This is necessary to make parsePerseusItem return a valid PerseusItem while
still being able to handle existing data that has the extra fields.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2582

## Test plan:

`yarn test`


packages/perseus/src/util/parse-perseus-json/general-purpose-parsers/pipe-parsers.typetest.ts
```diff
@@ -0,0 +1,31 @@
+// Test: pipeParsers()...then().parser returns the expected type
+
+import {pipeParsers} from "./pipe-parsers";
+import {string} from "./string";
+
+import type {Parser, PartialParser} from "../parser-types";
+
+const stringToNumber = summon<PartialParser<string, number>>();
```
jeremywiebe, line 8: Well that's a neat way to generate fakes for type testing. :) TIL

packages/perseus/src/util/parse-perseus-json/perseus-parsers/perseus-item.ts
```diff
@@ -6,23 +6,47 @@ import {
...
+//     "type": "multiple",
+//     "options": {
+//       "content": "",
+//       "images": {},
+//       "widgets": {}
+//     }
+//   }
+//
+// The "type" and "options" fields don't seem to be used anywhere. This
+// migration function removes them.
```
jeremywiebe, line 45: Yep!  My guess is that these were written by a buggy editor at some point.

packages/perseus/src/util/parse-perseus-json/perseus-parsers/perseus-item.ts
```diff
@@ -6,23 +6,47 @@ import {
...
-import type {Parser} from "../parser-types";
+import type {ParseContext, Parser, ParseResult} from "../parser-types";
 
 export const parsePerseusItem: Parser<PerseusItem> = object({
     question: parsePerseusRenderer,
     hints: array(parseHint),
-    answerArea: record(enumeration(...ItemExtras), boolean),
+    answerArea: pipeParsers(object({}))
+        .then(migrateAnswerArea)
+        .then(record(enumeration(...ItemExtras), boolean)).parser,
```
jeremywiebe, line 24: Just so I understand, without `migrateAnswerArea` the `enumeration()` parser tries to parse `type: string` and/or `options: object` and fails parsing?

packages/perseus/src/util/parse-perseus-json/general-purpose-parsers/index.ts
```diff
@@ -9,7 +9,9 @@ export * from "./number";
 export * from "./object";
 export * from "./optional";
 export * from "./pair";
+export * from "./pipe-parsers";
```
catandthemachines, line 12: This appears to be a barrel file, in that case I would encourage you not to add more content to this file and instead reference the file it's self when referencing resources in `pipe.parsers`.
This is s a hard as as I don't believe we've reached a consensus on this. But for reference here's the doc presented at the Frontend Guild Meeting.
https://docs.google.com/document/d/1uNV1mX4GpVsxTLwmt7GeBx8M43gemMjxqTiQv5u5L5k/edit?usp=sharing

packages/perseus/src/util/parse-perseus-json/general-purpose-parsers/pipe-parsers.test.ts
catandthemachines, line 1: Yay more tests!

packages/perseus/src/util/parse-perseus-json/general-purpose-parsers/pipe-parsers.ts
```diff
@@ -1,13 +1,25 @@
...
-    ParsedValue,
-    PartialParser,
     ParseContext,
+    ParsedValue,
     Parser,
+    PartialParser,
 } from "../parser-types";
 
-export function composeParsers<
+export function pipeParsers<T>(p: Parser<T>): ParserPipeline<T> {
```
catandthemachines, line 10: I am not familiar with these files, so forgive me if this is a newbie question. What is the purpose of adding this function and class to pipe-parsers for the Migration `answerArea` and removing unused fields. This looks like your cleaning up something else on top of this work, which isn't bad. But just curious how this loops back to the issue you're trying to fix.

catandthemachines (commented):
Had a few questions about the implementation.

packages/perseus/src/util/parse-perseus-json/perseus-parsers/perseus-item.ts
benchristel: Correct. `type` and `options` aren't listed in `ItemExtras`, so the enumeration parser will not match them.

packages/perseus/src/util/parse-perseus-json/general-purpose-parsers/pipe-parsers.ts
benchristel: `pipeParsers` is used here: https://github.com/Khan/perseus/pull/1895/files/a5735edc9b887a5a9bd1ebde7df37f89f9ab72f1#diff-f4f46cb44d17e630e444aaecb0252590dd3274953230681bcaecfc81aab0df91R22

It's useful for composing more than two parsers in a readable way.

PR #1896, opened 2024-11-20
Fix: Open unlimited polygon does not restrict crossing sides.

catandthemachines:
## Summary:
For keyboard users if they're drawing a polygon and it places a point at 0,0. There are cases where it could cross the sides of the polygon they are actively creating and cause the polygon to break. Mouse uses can get around this a bit. But it's not idea. This fix checks whether the polygon type allows intersecting sides and disables the rejection logic.

This PR also includes improvements to:
- Screen readers announcing instructions to enable graph (was not reading out before).
- Disabled buttons are not tabbable now.

Issue: LEMS-2618

## Test plan:
Go to /?path=/story/perseus-widgets-interactive-graph--unlimited-polygon-with-mafs in Storybook.
Use your keyboard to add points.
Move a few points in a way that a new point would cause the sides to intersect.
Add a new point and make sure you can still move points on the polygon.

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
@@ -420,6 +420,10 @@ const renderPolygonGraphControls = (props: {
     const shouldShowRemoveButton =
         showRemovePointButton && focusedPointIndex !== null;
 
+    // We want to disable the closePolygon button when
+    // there are not enough coords to make a polygon
+    const disableCloseButton = coords.length < 3;
+
     // If polygon is closed, show open button.
     // If polygon is open, show close button.
     const polygonButton = closedPolygon ? (
@@ -441,12 +445,12 @@ const renderPolygonGraphControls = (props: {
             kind="secondary"
             // Conditional disable when there are less than 3 points in
             // the graph
-            disabled={coords.length < 3}
+            disabled={disableCloseButton}
             style={{
                 width: "100%",
                 marginLeft: "20px",
             }}
-            tabIndex={coords.length < 3 ? -1 : 0}
+            tabIndex={disableCloseButton ? -1 : 0}
             onClick={() => {
                 props.dispatch(actions.polygon.closePolygon());
             }}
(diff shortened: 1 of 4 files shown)


packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -440,7 +446,7 @@ const renderPolygonGraphControls = (props: {
                 width: "100%",
                 marginLeft: "20px",
             }}
-            tabIndex={0}
+            tabIndex={coords.length < 3 ? -1 : 0}
```
nishasy, line 449: Can you add a comment explaining why the tab index depends on the coords length being less than 3?

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -483,8 +483,13 @@ function doMovePoint(
                 newValue: newValue,
             });
 
-            // Reject the move if it would cause the sides of the polygon to cross
-            if (polygonSidesIntersect(newCoords)) {
+            // Boolean value to tract whether we can let the polygon sides interact.
```
nishasy, line 486:
```suggestion
            // Boolean value to track whether we can let the polygon sides interact.
```

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
benchristel, line 1: The changes to this file seem unrelated to the rest of the PR. Are they supposed to be included?

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -301,6 +304,9 @@ export const MafsGraph = (props: MafsGraphProps) => {
                     {interactionPrompt && (
                         <View
                             style={{
+                                display: interactionPrompt
+                                    ? undefined
+                                    : "hidden",
```
nishasy, line 309: Nit: Can you add a brief note in your PR description saying that this change is also in your PR even though it's not the main focus? It's helpful for me to follow PR expectations that way, and I think it's nice to document for posterity.

nishasy (approved):
Neat! A couple of non-blocking suggestions, but overall looks good to me.

benchristel (commented):
What happens when closing the polygon is what causes the sides to cross? Can you then untangle the points using the keyboard, or are they stuck?

I think we might need some logic like "if the polygon's sides currently cross, don't reject a move that results in the sides still crossing."

benchristel (approved):
Discussed with Cat in Slack. It sounds like Caitlyn is aware of the "lines crossing" behavior and is okay with it for now. LGTM!

PR #1897, opened 2024-11-21
[Storybook] Disable snapshots globally + enable for select few stories

nishasy:
## Summary:
We want to disable snapshots globally, since most of our stories
don't need visual snapshot testing anyway, and we keep going over
our monthly chromatic snapshot limit.

Instead, we should deliberately set snapshots to true for stories we
want chromatic snapshots for.

- turn off chromatic snapshots globally
- turn on chromatic snapshots for interactive graph stories

Issue: https://khanacademy.atlassian.net/browse/LEMS-2579

Test plan:
Look at the Chromatic snapshots number on this PR below.

| Before | After |
| --- | --- |
| Screenshot 2024-11-20 at 4 11 35 PM | Screenshot 2024-11-20 at 4 11 41 PM |

### DIFF
--- a/.changeset/lucky-clouds-hang.md
+++ b/.changeset/lucky-clouds-hang.md
@@ -1,4 +1,5 @@
 ---
+"@khanacademy/math-input": patch
 "@khanacademy/perseus": patch
 "@khanacademy/perseus-editor": patch
 ---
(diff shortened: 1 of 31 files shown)


nishasy:
It turns out that we were only supposed to be capturing snapshots for regression stories anyway. For some reason, other stories were still being captured. That means we don't need all the extra enabled stories in this PR, so I'm going to create a new PR for the minimal necessary changes:

https://github.com/Khan/perseus/pull/1911

PR #1898, opened 2024-11-21
Dropdown: Extract validation out of scoring

Myranae:
## Summary:
To complete server-side scoring, we are separating out validation logic from scoring logic. This PR separates that logic and updates associated tests.

Issue: LEMS-2597

## Test plan:
- Confirm all checks pass


packages/perseus/src/widgets/dropdown/score-dropdown.test.ts
```diff
@@ -7,22 +7,6 @@ import type {
...
-        const rubric: PerseusDropdownRubric = {
-            choices: question1.widgets["dropdown 1"].options.choices,
-        };
-
-        // Act
-        const result = scoreDropdown(userInput, rubric);
-
-        // Assert
-        expect(result).toHaveInvalidInput();
-    });
```
Myranae, line 24: Didn't keep a check for this in scoring test since it looks like we might be doing validation before scoring (love that)

packages/perseus/src/widgets/dropdown/validate-dropdown.test.ts
```diff
@@ -0,0 +1,31 @@
+import validateDropdown from "./validate-dropdown";
+
+import type {PerseusDropdownUserInput} from "../../validation.types";
+
+describe("validateDropdown", () => {
+    it("returns invalid for invalid input (user input of 0)", () => {
```
jeremywiebe, line 6: Tiny improvement suggestion as `0` is the selected index. 🤷‍♂️  You decide. :) 

```suggestion
    it("returns invalid when nothing selected (user input of 0)", () => {
```

packages/perseus/src/widgets/dropdown/validate-dropdown.ts
```diff
@@ -0,0 +1,22 @@
+import type {PerseusScore} from "../../types";
+import type {PerseusDropdownUserInput} from "../../validation.types";
+
+/**
+ * Checks if the user has selected an item from the dropdown before scoring.
+ * This is shown with a userInput value / index other than 0.
+ * @param userInput
```
jeremywiebe, line 7: I would suggest removing these two lines:

The first one documents the internal implementation of the function (how it checks validation) which you can see by reading the code. It also presents something that can easily get out of sync with the code. A great guide for comments that I was told one time was "document why, not what". :) 

The `@param userInput` doesn't really add any value.

jeremywiebe (approved):

PR #1899, opened 2024-11-21
Plotter: Extract validation out of scoring

Myranae:
## Summary:
To complete server-side scoring, we are separating out validation logic from scoring logic. This PR separates that logic and associated tests for the plotter widget.

Issue: LEMS-2604

## Test plan:
- Confirm checks pass

### DIFF
--- a/packages/perseus/src/widgets/plotter/validate-plotter.ts
+++ b/packages/perseus/src/widgets/plotter/validate-plotter.ts
@@ -11,8 +11,7 @@ const {deepEq} = Util;
 /**
  * Checks user input to confirm it is not the same as the starting values for the graph.
  * This means the user has modified the graph, and the question can be scored.
- * @param userInput
- * @param validationData
+ *
  * @see 'scorePlotter' for more details on scoring.
  */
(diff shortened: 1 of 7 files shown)


packages/perseus/src/widgets/plotter/validate-plotter.ts
```diff
@@ -0,0 +1,31 @@
...
+    PerseusPlotterValidationData,
+} from "../../validation.types";
+
+const {deepEq} = Util;
+
+/**
+ * Checks user input to confirm it is not the same as the starting values for the graph.
+ * This means the user has modified the graph, and the question can be scored.
+ * @param userInput
+ * @param validationData
```
jeremywiebe, line 15: Let's remove these. They don't add any new information.

jeremywiebe (approved):

packages/perseus/src/widgets/plotter/validate-plotter.ts
Myranae: Removed!

PR #1901, opened 2024-11-21
Number Line: Extract validation out of scoring

Myranae:
## Summary:
To complete server-side scoring, we are separating out validation logic from scoring logic. This PR separates that logic and associated tests for the number line widget.

Issue: LEMS-2608

## Test plan:
- Confirm checks pass
- Confirm changing the flow of validation and scoring does not affect user experience


packages/perseus/src/widgets/number-line/score-number-line.ts
```diff
@@ -1,50 +1,36 @@
...
-        rubric.correctX || 0,
+        userInput.numLinePosition,
+        scoringData.correctX || 0,
     );
-    const outsideAllowedRange =
-        state.numDivisions > divisionRange[1] ||
-        state.numDivisions < divisionRange[0];
 
-    // TODO: I don't think isTickCrtl is a thing anymore
-    if (state.isTickCrtl && outsideAllowedRange) {
```
Myranae, line 27: I'm guessing when exercises allow users to set their own interval, then this check was necessary in case they put in an invalid number of divisions.

I checked the number line editor, and it still has a checkbox for `isTickCrtl` and the json still has the field (though it defaults to not being there). I tried removing `state.isTickCrtl` from line 27 and all the tests still pass. It doesn't look like it is referenced anywhere other than tests and this line here. Wouldn't this always be false if `isTickCrtl` is not in use? Maybe this validation check hasn't been working?

Or maybe this check is not necessary if we are no longer allowing the user to set their own tick interval. It _is_ possible to turn this to `true` in the editor, so I think we should only remove it if that checkbox ever gets removed, just in case it has been set to true in the past or in case anyone sets it to true in the future.

I don't think this property should be on userInput though. Should we move it to validationData?

packages/perseus/src/widgets/number-line/score-number-line.ts
```diff
@@ -1,50 +1,36 @@
...
             message: null,
         };
     }
-    if (state.numLinePosition === start && state.rel === startRel) {
-        // We're where we started.
-        return {
-            type: "invalid",
-            message: null,
-        };
-    }
```
Myranae, line 47: I don't think it's necessary to do this check after check if the user got it right. If the user tries to check their answer and they have not moved the dot, then it should probably be invalid. This might not be the case if the first tick could possibly be the right answer. Might be good to check with content before landing this change.

packages/perseus/src/widgets/number-line/score-number-line.ts
Myranae: Maybe moving it can be a later fix if needed. Just left it for now as the only logic moved into the validation function.

packages/perseus/src/widgets/number-line/score-number-line.ts
```diff
@@ -1,44 +1,41 @@
...
+    if (correctPos && correctRel === userInput.rel) {
         return {
             type: "points",
             earned: 1,
             total: 1,
             message: null,
         };
     }
-    if (state.numLinePosition === start && state.rel === startRel) {
+    if (userInput.numLinePosition === start && userInput.rel === startRel) {
```
Myranae, line 38: This check could possibly be moved to the validation function at a later time if the flow can change so this check occurs first. Only possible if the answer is never the starting position though.

packages/perseus/src/widgets/number-line/score-number-line.ts
Myranae: Decided to leave this validation logic in the scoring function since it occurs after scoring logic.

handeyeco (approved):
Thanks!

PR #1902, opened 2024-11-21
Radio: Extract validation out of scoring

Myranae:
## Summary:
To complete server-side scoring, we are separating out validation logic from scoring logic. This PR separates validation logic that does not depend on answer information and also separates associated tests for the radio widget.

Issue: LEMS-2594

## Test plan:
- Confirm checks pass
- Confirm the validation logic remaining in the scoring function cannot be removed from scoring

### DIFF
--- a/packages/perseus/src/widgets/radio/score-radio.ts
+++ b/packages/perseus/src/widgets/radio/score-radio.ts
@@ -1,3 +1,5 @@
+import validateRadio from "./validate-radio";
+
 import type {PerseusStrings} from "../../strings";
 import type {PerseusScore} from "../../types";
 import type {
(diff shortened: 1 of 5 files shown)


packages/perseus/src/widgets/radio/score-radio.ts
```diff
@@ -10,17 +12,15 @@ function scoreRadio(
...
         return sum + (selected ? 1 : 0);
     }, 0);
 
-    if (numSelected === 0) {
-        return {
-            type: "invalid",
-            message: null,
-        };
-    }
-
```
Myranae, line 23: The remaining validation checks rely on data in the scoringData object, so I don't think we can move those to the validation function. If it's possible, would love to know so we can move all that over!

jeremywiebe (approved):

PR #1903, opened 2024-11-21
Interactive Graph: Extract validation out of scoring

Myranae:
## Summary:
To complete server-side scoring, we are separating out validation logic from scoring logic. This PR separates that logic and associated tests for the Interactive Graph widget.

Issue: LEMS-2595

## Test plan:
- Confirm checks pass
- Confirm validation check location works

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/score-interactive-graph.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/score-interactive-graph.ts
@@ -13,6 +13,7 @@ import {
 } from "../interactive-graph";
 
 import {getClockwiseAngle} from "./math/angles";
+import validateInteractiveGraph from "./validate-interactive-graph";
 
 import type {PerseusScore} from "../../types";
 import type {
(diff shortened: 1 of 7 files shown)


packages/perseus/src/widgets/interactive-graphs/score-interactive-graph.ts
```diff
@@ -280,14 +281,11 @@ function scoreInteractiveGraph(
...
-        // We're where we started.
-        return {
-            type: "invalid",
-            message: null,
-        };
+    // The input wasn't correct, so check if it's a blank input (validate) or if
+    // it's actually just wrong
+    const validationError = validateInteractiveGraph(userInput, scoringData);
+    if (validationError) {
+        return validationError;
```
Myranae, line 288: The validation function was put here since this is the only place where validation occurs. It seems like it should be okay to check if the user input is blank or if the user hasn't changed the graph yet at the start of scoring, but since that's not what was done and there is a comment specifically about doing this after confirming the input was not correct, I left it here.

Myranae: Realized that the none type graph check does need to happen before validation because all none type graphs will have empty input. Will need to review this before allowing validation to occur before scoring.

Myranae:
Not moving forward with this work for now. Please see the associated ticket for details.

PR #1906, opened 2024-11-23
feat(LEMS-2664): update use control point and movable point with optional props

anakaren-rojas:
## Summary:
- Updates _MovablePoint_ component and _useControlPoint_ method to include optional props for `aria label`, `aria described by`, `aria live`, and `sequence number`
- Sets default for aria live to `polite`; based on design doc, only the removal of a point or trying to move a point of out bounds warrants the use of `assertive`
- Set default for sequence number - if aria label and sequence number are not provided - sequence will default to 1 
- Creates logic for optional aria label - custom aria label will take precedence over sequence number

Issue: LEMS-2664

## Test plan:
- Using [chromatic](https://650db21c3f5d1b2f13c02952-jhskbowqxd.chromatic.com/?path=/docs/perseuseditor-editorpage--docs) link, selecting an interactive graph point reads out the point's coordinates 

## Screen recording

https://github.com/user-attachments/assets/f2414756-b0bf-48bf-98ed-08f30c41d1a1


packages/perseus/src/widgets/interactive-graphs/graphs/components/use-control-point.tsx
```diff
@@ -47,6 +51,9 @@ export function useControlPoint(params: Params): Return {
     const {
         point,
         sequenceNumber,
+        ariaDescribedBy,
+        ariaLabel,
+        ariaLive = "polite",
```
anakaren-rojas, line 56: Only removing point or moving out of bounds should be `assertive`

packages/perseus/src/widgets/interactive-graphs/graphs/components/use-control-point.tsx
```diff
@@ -26,13 +25,18 @@ type Params = {
      * interactive figure on the graph.
      */
     sequenceNumber: number;
```
nishasy, line 27: How would you feel about making `sequenceNumber` optional, and then defaulting it to `1` below?

If this point has special aria label, it could be confusing to also require passing in a sequence number if it's not going to be used.

packages/perseus/src/widgets/interactive-graphs/graphs/components/use-control-point.tsx
```diff
@@ -76,6 +83,15 @@ export function useControlPoint(params: Params): Return {
         constrainKeyboardMovement: constrain,
     });
 
+    // if custom aria label is not provided, will use default of sequence number and point coordinates
+    const pointAriaLabel =
+        ariaLabel ||
+        strings.srPointAtCoordinates({
+            num: sequenceNumber,
```
nishasy, line 90: From my comment above, maybe this line could be

```
num: sequenceNumber ?? 1,
```

anakaren-rojas: Defaulted the prop in the de-structuring section

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.tsx
```diff
@@ -16,14 +23,11 @@ type Props = {
      * Note: This number is 1-indexed, and should restart from 1 for each
      * interactive figure on the graph.
      */
-    sequenceNumber: number;
-    onMove?: (newPoint: vec.Vector2) => unknown;
+    sequenceNumber?: number;
+    onBlur?: ((event: React.FocusEvent) => unknown) | undefined;
```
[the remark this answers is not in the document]
anakaren-rojas: Nope, I kept them as they were, but it would be safe to remove the `| undefined`

PR #1908, opened 2024-11-23
Migrate expression widget to latest version in PerseusItem parser

benchristel:
This demonstrates how we can upgrade/migrate widget options in the parsing
code, obviating the need for `propUpgrades`.

The fact that `version` is an object makes this really annoying because
TypeScript can't discriminate unions based on nested properties like
`version.major`. Given that constraint, I think the approach I came up with is
reasonable, but I'm certainly open to feedback on how to improve it.

Issue: LEMS-2582

## Test plan:

`yarn test`

### DIFF
--- a/packages/perseus/src/util/parse-perseus-json/perseus-parsers/expression-widget.ts
+++ b/packages/perseus/src/util/parse-perseus-json/perseus-parsers/expression-widget.ts
@@ -36,6 +36,29 @@ const parseAnswerForm: Parser<PerseusExpressionAnswerForm> = object({
     ).parser,
 });
 
+const parseExpressionWidgetLatest: Parser<ExpressionWidget> = parseWidget(
+    constant("expression"),
+    object({
+        answerForms: array(parseAnswerForm),
+        functions: array(string),
+        times: boolean,
+        visibleLabel: optional(string),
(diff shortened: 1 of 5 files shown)


packages/perseus/src/util/parse-perseus-json/perseus-parsers/expression-widget.ts
```diff
@@ -51,3 +58,58 @@ export const parseExpressionWidget: Parser<ExpressionWidget> = parseWidget(
...
+                    simplify: options.simplify,
+                    value: options.value,
+                },
+            ],
+        },
+    })
+}
+
+export const parseExpressionWidget: Parser<ExpressionWidget> =
+    union(pipeParsers(parseExpressionWidgetV0).then(migrateV0ToLatest).parser)
```
jeremywiebe, line 114: So this will attempt to parse using the v0 and migration parsers and fall back to the `parseExpressionWidgetLatest` if that fails, right? 

This might be extremely over-the-top optimization, but I wonder if we should order these in the opposite order given we'll more likely get a latest version set of options rather than v0 options?  🤔

jeremywiebe (approved):
Nice. I like how this PR is starting to work out the upgrade/migration ideas.

PR #1911, opened 2024-11-25
[Chromatic] Disable snapshots globally, only keep regression stories

nishasy:
## Summary:
Right now, Chromatic snapshots are only supposed to run on files
that end with `.regression.stories.tsx`, but they seem to be running
on other stories as well in the github checks.

Instead of using the .yml to specify this, I'm disabling chromatic
using the `disableSnapshot` chromatic parameter globally, and
manually enabling it in the one regression stories file we have.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2579

## Test plan:
Check the chromatic info in the github checks below.

### DIFF
--- a/.github/workflows/storybook.yml
+++ b/.github/workflows/storybook.yml
@@ -43,4 +43,3 @@ jobs:
                   token: ${{ secrets.GITHUB_TOKEN }}
                   projectToken: ${{ secrets.CHROMATIC_APP_CODE }}
                   autoAcceptChanges: "main"
-                  onlyStoryFiles: "**/*regression.stories.tsx"
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -23,7 +23,14 @@ const preview: Preview = {
         (Story) => (
             <RenderStateRoot>
                 <DependenciesContext.Provider value={storybookDependenciesV2}>
-                    <Story />
+                    {/* Most of our components have an expectation to be
+                        rendered inside of a .framework-perseus container.
+                        We want to make sure we can include it here, since it
+                        can also affect the styling.
+                    */}
+                    <div className="framework-perseus">
+                        <Story />
+                    </div>
                 </DependenciesContext.Provider>
             </RenderStateRoot>
         ),
@@ -58,6 +65,12 @@ const preview: Preview = {
                 name,
                 value,
             })),
+        },
+        // Disabling Chromatic snapshots across all stories, since we have
+        // a limited snapshot budget. Set this explicitly to false on
+        // stories that we want snapshots for in their respective files.
+        chromatic: {
+            disableSnapshot: true,
         },
     },
 };
--- a/.changeset/wise-dragons-prove.md
+++ b/.changeset/wise-dragons-prove.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+[Chromatic] Disable snapshots globally, only keep regression stories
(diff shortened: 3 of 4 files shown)


nishasy:
The new "test" number on the dashboard matches the expected snapshot number in the checks! I think this means it's working.

image
image

.storybook/preview.tsx
```diff
@@ -23,7 +23,14 @@ const preview: Preview = {
...
                 <DependenciesContext.Provider value={storybookDependenciesV2}>
-                    <Story />
+                    {/* Most of our components have an expectation to be
+                        rendered inside of a .framework-perseus container.
+                        We want to make sure we can include it here, since it
+                        can also affect the styling.
+                    */}
+                    <div className="framework-perseus">
+                        <Story />
+                    </div>
```
nishasy, line 33: NOTE: This is causing some trivial snapshot updates within the UI tests for the regression stories.

.github/workflows/storybook.yml
```diff
@@ -43,4 +43,3 @@ jobs:
                   token: ${{ secrets.GITHUB_TOKEN }}
                   projectToken: ${{ secrets.CHROMATIC_APP_CODE }}
                   autoAcceptChanges: "main"
-                  onlyStoryFiles: "**/*regression.stories.tsx"
```
jeresig, line 46: So I think the issue is that `onlyStoryFiles` is an option inside the `chromatic.config.json` file, the option is here: https://www.chromatic.com/docs/configure/#options

nishasy: Ah, that's good to know!

I think we're gonna enable a few key stories other than the regression stories soon, so I think it would still be worth removing this from the .yml file.

jeremywiebe: It seems like this is working.  This setting is documented as being supported in the GitHub Action: 

image

On an [example PR](https://github.com/Khan/perseus/actions/runs/12017642792/job/33500417832?pr=1906) I checked, it seems like it is obeying the setting: 

```
19:50:09.030 Starting partial build
19:50:09.030     → Snapshots will be limited to story files matching '**/*regression.stories.tsx'

...

19:50:58.649 Build 5134 passed!
                 → Tested 412 stories across 130 components; captured 27 snapshots in 39 seconds
```

It captured only 27 snapshots.

nishasy: @jeremywiebe It does say "captured 27 snapshots" on all the PRs. I checked a bunch of older unrelated Perseus PRs to check this, and it's true that that messaging is consistent. However, the PRs would _also_ show other snapshots that aren't part of these 27 in the "UI test" check. It also says there are 130 tests on the chromatic dashboard for other PRs when there should only be 27:

image


I'd like to move away from using the `onlyStoryFiles: "**/*regression.stories.tsx"` because I think there's a disconnect somewhere that isn't super obvious.

nishasy: I'm going to retract my last comment. Just for a confidence check, I put up another PR that would have had a ton of UI test updates if my assertion was correct. It only reported changes in the regression stories, so I think my statements before this were all false, and I just don't know what "tests" means on the chromatic dashboard.

nishasy: @jeremywiebe we can still continue with this PR if we want to enable other stories to have snapshots later, rather than just limiting to regression stories.

nishasy:
It turns out that the `onlyStoryFiles: "**/*regression.stories.tsx"` line in the .yml file is working as expected, and I was just misunderstanding the reports in the chromatic dashboard. We don't need to enable snapshots on any other stories at this time, so I'm closing this PR.

PR #1912, opened 2024-11-25
Use EditableControlledInput body within ScrolllessNumberTextField

nishasy:
## Summary:
todo

Issue: https://khanacademy.atlassian.net/browse/LEMS-2303

## Test plan:
todo

### DIFF
--- a/packages/perseus-editor/src/components/scrollless-number-text-field.tsx
+++ b/packages/perseus-editor/src/components/scrollless-number-text-field.tsx
@@ -1,13 +1,32 @@
-import {TextField} from "@khanacademy/wonder-blocks-form";
+import {
+    addStyle,
+    type PropsFor,
+    type StyleType,
+} from "@khanacademy/wonder-blocks-core";
+import {border, color, spacing} from "@khanacademy/wonder-blocks-tokens";
+import {styles as typographyStyles} from "@khanacademy/wonder-blocks-typography";
+import {StyleSheet} from "aphrodite";
 import * as React from "react";
 
-import type {PropsFor} from "@khanacademy/wonder-blocks-core";
+type Props = Omit<PropsFor<"input">, "value" | "onInput"> & {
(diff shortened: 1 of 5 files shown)


packages/perseus-editor/src/components/scrollless-number-text-field.tsx
```diff
@@ -39,7 +49,33 @@ const ScrolllessNumberTextField = (props: PropsFor<typeof TextField>) => {
...
+                setWipValue(newValue);
+                onChange(newValue);
+            }}
+            onFocus={(e) => {
+                setWipValue(value);
+                setFocused(true);
+
+                if (props.onFocus) {
+                    props.onFocus(e);
+                }
```
benchristel, line 67: Could this be simplified to

```suggestion
                props.onFocus?.(e);
```

packages/perseus-editor/src/components/scrollless-number-text-field.tsx
```diff
@@ -39,7 +49,33 @@ const ScrolllessNumberTextField = (props: PropsFor<typeof TextField>) => {
...
+            }}
+            onBlur={(e) => {
+                setFocused(false);
+
+                if (props.onBlur) {
+                    props.onBlur(e);
+                }
```
benchristel, line 74:
```suggestion
                props.onBlur?.(e);
```

benchristel (approved):
Cool!

It's slightly awkward that we now have two different input components with similar behavior. I wonder if `EditableControlledInput` should be rewritten to use `TextField` internally. Then `ScrolllessNumberTextField` could use `EditableControlledInput`. It doesn't have to block this change, though.

nishasy:
@benchristel I think it could make sense to do that. I think we'd have to move it to the Perseus package so we can import it where necessary, and we'd have to make sure that it forwards refs.

PR #1913, opened 2024-11-25
check how many snapshots update with this change

nishasy:
## Summary:
<Add your summary here>

Issue: XXX-XXXX

## Test plan:


nishasy:
This was just a test. It proved that only the regression stories are running chromatic snapshots.

PR #1914, opened 2024-11-26
Add tests for propUpgrades functions

handeyeco:
## Summary:
Only 3 widgets seem to use the `propUpgrades` mechanism:
- Radio
- Expression
- Measurer

None of them seemed to have tests (at least not direct tests), so this adds them.

I also took out underscore after I wrote the tests.

## Test plan:
This is the test plan

### DIFF
--- a/packages/perseus/src/widgets/expression/expression.test.tsx
+++ b/packages/perseus/src/widgets/expression/expression.test.tsx
@@ -523,4 +523,33 @@ describe("Expression Widget", function () {
             ).toBeNull();
         });
     });
+
+    describe("propUpgrades", () => {
+        it("can upgrade from v0 to v1", () => {
+            const v0props = {
+                times: false,
+                buttonSets: ["basic"],
+                functions: [],
+                form: false,
+                simplify: false,
+                value: "42",
+            };
+
+            const result = ExpressionWidgetExport.propUpgrades["1"](v0props);
+
+            expect(result).toEqual({
+                times: false,
+                buttonSets: ["basic"],
+                functions: [],
+                answerForms: [
+                    {
+                        considered: "correct",
+                        form: false,
+                        simplify: false,
+                        value: "42",
+                    },
+                ],
+            });
+        });
+    });
 });
--- a/packages/perseus/src/widgets/expression/expression.test.tsx
+++ b/packages/perseus/src/widgets/expression/expression.test.tsx
@@ -552,7 +552,8 @@ describe("Expression Widget", function () {
                 ],
             };
 
-            const result = ExpressionWidgetExport.propUpgrades["1"](v0props);
+            const result: PerseusExpressionWidgetOptions =
+                ExpressionWidgetExport.propUpgrades["1"](v0props);
 
             expect(result).toEqual(expected);
         });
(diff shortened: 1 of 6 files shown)


packages/perseus/src/widgets/expression/expression.test.tsx
```diff
@@ -523,4 +523,33 @@ describe("Expression Widget", function () {
...
+            const v0props = {
+                times: false,
+                buttonSets: ["basic"],
+                functions: [],
+                form: false,
+                simplify: false,
+                value: "42",
+            };
+
+            const result = ExpressionWidgetExport.propUpgrades["1"](v0props);
```
jeremywiebe, line 538: **suggestion**: Typically, I'm against explicit typing for variables, but in this case I think it's a good additional validation to type this `result` variable as `PerseusExpressionWidgetOptions` so that TypeScript ensures that the prop upgrade it typed correctly.

packages/perseus/src/widgets/measurer/measurer.test.tsx
```diff
@@ -0,0 +1,39 @@
...
+                showRuler: false,
+                rulerLabel: "test",
+                rulerTicks: 4,
+                rulerPixels: 4,
+                rulerLength: 4,
+                box: [4, 4],
+                static: false,
+            };
+
+            const result = MeasurerWidgetExport.propUpgrades["1"](v0props);
```
jeremywiebe, line 20: Same note about typing the `result` variable by the widget options type as above.

packages/perseus/src/widgets/measurer/measurer.tsx
```diff
@@ -183,18 +183,16 @@ class Measurer extends React.Component<Props> implements Widget {
...
+        const {imageUrl, imageTop, imageLeft, ...rest} = v0props;
+
+        return {
+            ...rest,
+            image: {
+                url: imageUrl,
+                top: imageTop,
+                left: imageLeft,
+            },
+        };
```
jeremywiebe, line 195: So much simpler and clearer than the underscore variant!

packages/perseus/src/widgets/radio/radio.ts
```diff
@@ -127,22 +127,18 @@ const transform = (
...
-        let choices;
-        let hasNoneOfTheAbove;
+        const {noneOfTheAbove, ...rest} = v0props;
 
-        if (!v0props.noneOfTheAbove) {
-            choices = v0props.choices;
-            hasNoneOfTheAbove = false;
-        } else {
+        if (noneOfTheAbove) {
             throw new Error(
```
jeremywiebe, line 133: 🤔  I wonder what happens if this throws in prod. 🤔 🤔 🤔 

No action required!

jeremywiebe (approved):
Thanks for adding these tests!

packages/perseus/src/widgets/expression/expression.test.tsx
handeyeco: I went back and forth on this. This won't work:

``` TS
const result: WhateverWidgetOptions = WhateverWidgetExport.propUpgrades["1"](v0props);
```

...because the `propUpgrades` function takes `any` and modifies it (which basically makes the return value `any`). Also because 2/3 `propUpgrades` returned `any` (I just changed that).

So I did a mix:

``` TS
const expected: WhateverWidgetOptions = {}
const result: WhateverWidgetOptions = WhateverWidgetExport.propUpgrades["1"](v0props);
expect(result).toEqual(expected)
```

Not sure how helpful it'll be though.

PR #1919, opened 2024-11-26
remove findDOMNode from text-input

handeyeco:
## Summary:
`findDOMNode` is deprecated and React whines about it constantly when running tests, so I'm trying to chip away at removing uses of it.

### DIFF
--- a/packages/perseus/src/components/text-input.tsx
+++ b/packages/perseus/src/components/text-input.tsx
@@ -2,7 +2,6 @@
 import {Errors, PerseusError} from "@khanacademy/perseus-core";
 import {TextField} from "@khanacademy/wonder-blocks-form";
 import * as React from "react";
-import ReactDOM from "react-dom";
 
 import type {StyleType} from "@khanacademy/wonder-blocks-core";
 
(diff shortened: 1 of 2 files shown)


packages/perseus/src/components/text-input.tsx
```diff
@@ -47,45 +48,46 @@ class TextInput extends React.Component<Props> {
...
     blur: () => void = () => {
-        // @ts-expect-error - TS2531 - Object is possibly 'null'. | TS2339 - Property 'blur' does not exist on type 'Element | Text'.
-        ReactDOM.findDOMNode(this).blur();
+        this._getInput().blur();
     };
 
     getValue: () => string | null | undefined = () => {
-        // @ts-expect-error - TS2339 - Property 'value' does not exist on type 'Element | Text'.
-        return ReactDOM.findDOMNode(this)?.value;
+        return this.inputRef.current?.value;
```
handeyeco, line 71: I didn't use `_getInput` here because the original code is actually checking to see if the node exists before trying to access it.

jeremywiebe (approved):

PR #1923, opened 2024-11-27
Convert backgroundImage dimensions to numbers during parsing

benchristel:
Issue: LEMS-2582

## Test plan:

`yarn test`


packages/perseus/src/util/parse-perseus-json/perseus-parsers/perseus-image-background.ts
```diff
@@ -3,21 +3,26 @@ import {
...
     union,
 } from "../general-purpose-parsers";
+import {stringToNumber} from "../general-purpose-parsers/string-to-number";
 
 import type {Parser} from "../parser-types";
 import type {PerseusImageBackground} from "@khanacademy/perseus";
 
+const numericToNumber = pipeParsers(union(number).or(string).parser).then(
+    stringToNumber,
+).parser;
```
jeremywiebe, line 17: I don't understand why we need to `union` `number` or `string` and then pass that result through `stringToNumber`.  Doesn't `stringToNumber` handle parsing a string or number within itself already?

packages/perseus/src/util/parse-perseus-json/general-purpose-parsers/string-to-number.ts
```diff
@@ -0,0 +1,17 @@
+import type {PartialParser} from "../parser-types";
+
+export const stringToNumber: PartialParser<string | number, number> = (
```
jeremywiebe, line 3: Sadly, there is more than one place where Perseus data contains both actual numbers and stringified numbers. 

No action needed.

jeremywiebe (approved):

packages/perseus/src/util/parse-perseus-json/perseus-parsers/perseus-image-background.ts
benchristel: Parsing with a union ensures that if we get something that's _neither_ a string nor a number, the parse will fail instead of doing something we didn't want (like converting null to 0).

packages/perseus/src/util/parse-perseus-json/general-purpose-parsers/string-to-number.ts
```diff
@@ -0,0 +1,17 @@
...
+    rawValue,
+    ctx,
+) => {
+    if (typeof rawValue === "number") {
+        return ctx.success(rawValue);
+    }
```
benchristel, line 9: I don't think these 3 lines are needed.

PR #1924, opened 2024-11-27
Add and pass more regression tests for PerseusItem parser

benchristel:
Issue: LEMS-2582

## Test plan:

`yarn test`

### DIFF
--- a/packages/perseus/src/util/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
+++ b/packages/perseus/src/util/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
@@ -285,7 +285,7 @@ export const parseInteractiveGraphWidget: Parser<InteractiveGraphWidget> =
             snapStep: optional(pairOfNumbers),
             backgroundImage: optional(parsePerseusImageBackground),
             markings: enumeration("graph", "grid", "none"),
-            labels: defaulted(array(string), () => []),
+            labels: optional(array(string)),
             showProtractor: boolean,
             showRuler: optional(boolean),
             showTooltips: optional(boolean),
--- a/packages/perseus/src/util/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
+++ b/packages/perseus/src/util/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
@@ -13,6 +13,7 @@ import {
     trio,
     union,
 } from "../general-purpose-parsers";
+import {defaulted} from "../general-purpose-parsers/defaulted";
 
 import {parsePerseusImageBackground} from "./perseus-image-background";
 import {parseWidget} from "./widget";
@@ -276,11 +277,11 @@ export const parseInteractiveGraphWidget: Parser<InteractiveGraphWidget> =
         constant("interactive-graph"),
         object({
             step: pairOfNumbers,
-            gridStep: pairOfNumbers,
-            snapStep: pairOfNumbers,
+            gridStep: optional(pairOfNumbers),
+            snapStep: optional(pairOfNumbers),
             backgroundImage: optional(parsePerseusImageBackground),
             markings: enumeration("graph", "grid", "none"),
-            labels: array(string),
+            labels: defaulted(array(string), () => []),
             showProtractor: boolean,
             showRuler: optional(boolean),
             showTooltips: optional(boolean),
--- a/packages/perseus/src/util/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
+++ b/packages/perseus/src/util/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
@@ -288,7 +288,12 @@ export const parseInteractiveGraphWidget: Parser<InteractiveGraphWidget> =
             rulerLabel: optional(string),
             rulerTicks: optional(number),
             range: pair(pairOfNumbers, pairOfNumbers),
-            graph: parsePerseusGraphType,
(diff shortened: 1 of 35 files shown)


packages/perseus/src/util/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
```diff
@@ -276,18 +277,23 @@ export const parseInteractiveGraphWidget: Parser<InteractiveGraphWidget> =
         constant("interactive-graph"),
         object({
             step: pairOfNumbers,
-            gridStep: pairOfNumbers,
-            snapStep: pairOfNumbers,
+            gridStep: optional(pairOfNumbers),
+            snapStep: optional(pairOfNumbers),
```
benchristel, line 281: I'd prefer it if we could default `gridStep` and `snapStep` here, but the defaulting logic used in the `InteractiveGraph` component is fairly complicated and depends on the graph bounds. I guess there should be a TODO comment here.

packages/perseus/src/util/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
```diff
@@ -276,18 +277,23 @@ export const parseInteractiveGraphWidget: Parser<InteractiveGraphWidget> =
...
             backgroundImage: optional(parsePerseusImageBackground),
             markings: enumeration("graph", "grid", "none"),
-            labels: array(string),
+            labels: defaulted(array(string), () => []),
```
benchristel, line 284: This default isn't right... `labels` is for the X and Y axis labels, not locked labels. The default labels are currently assigned in the default props of the InteractiveGraph component.

I think the correct fix is to let `labels` be `optional` here and in the `PerseusInteractiveGraphWidgetOptions` type. The props type for InteractiveGraph should not use `PerseusInteractiveGraphWidgetOptions`.

packages/perseus/src/util/parse-perseus-json/perseus-parsers/widgets-map.ts
```diff
@@ -175,6 +175,10 @@ const parseWidgetsMapEntry: (
             return parseAndAssign(`table ${id}`, parseTableWidget);
         case "video":
             return parseAndAssign(`video ${id}`, parseVideoWidget);
+        case "sequence":
+            // sequence is a deprecated widget type, and the corresponding
+            // widget component no longer exists.
+            return ctx.success(null);
```
benchristel, line 181: This will cause the sequence widget options not to be added to the `widgets` map. I should confirm that content with sequence widgets will still render the deprecated widget placeholder if those options aren't present. We might need to add a placeholder widget to the `widgets` map here in place of the sequence widget to prevent the renderer from crashing.

packages/perseus/src/widgets/numeric-input/score-numeric-input.ts
```diff
@@ -111,7 +111,7 @@ function scoreNumericInput(
 
     const normalizedAnswerExpected = rubric.answers
         .filter((answer) => answer.status === "correct")
-        .every((answer) => Math.abs(answer.value) <= 1);
+        .every((answer) => answer.value == null || Math.abs(answer.value) <= 1);
```
benchristel, line 114: I had to change this in order to convince TypeScript that we're handling answers with no `value`, which exist in our content.

...though now that I'm reading the code again, I think I did it wrong. `Math.abs(null)` is 0, but `Math.abs(undefined)` is NaN, and NaN is not `<= 1`. So the existing behavior is to set `normalizedAnswerExpected` to `false` if any `answer.value` is undefined, but to `true` if every `answer.value` is either null or at most 1. My change allows `normalizedAnswerExpected` to be true when both `null` and `undefined` values are present.

I'm not sure how much this detail matters given that the current logic is extremely weird already (see the characterization tests above) but I don't want to break existing exercises.

packages/perseus/src/widgets/interactive-graph.tsx
```diff
@@ -118,11 +122,77 @@ const makeInvalidTypeError = (
...
+    snapStep?: [x: number, y: number];
+    // An optional image to use in the background
+    backgroundImage?: PerseusImageBackground;
+    /**
+     * The type of markings to display on the graph.
+     * - graph: shows the axes and the grid lines
+     * - grid: shows only the grid lines
+     * - none: shows no markings
+     */
+    markings: "graph" | "grid" | "none";
```
jeremywiebe, line 144: This is slightly OCD, but would you mind making all of these JSDoc-style (`/** */`) comments? Some are, but the ones that aren't won't surface their information at call sites in VSCode and Goland. It's a tiny thing but is handy when accessing the prop in code to see its comment. 

image

packages/perseus/src/util/parse-perseus-json/perseus-parsers/widgets-map.ts
```diff
@@ -34,8 +34,12 @@ import {parseRadioWidget} from "./radio-widget";
 import {parseSorterWidget} from "./sorter-widget";
 import {parseTableWidget} from "./table-widget";
 import {parseVideoWidget} from "./video-widget";
+import {parseWidget} from "./widget";
 
-import type {PerseusWidgetsMap} from "../../../perseus-types";
+import type {
+    AutoCorrectWidget,
```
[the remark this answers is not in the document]
benchristel: `AutoCorrectWidget` should probably be called `DeprecatedStandin`.

PR #1928, opened 2024-11-27
[SR] Add aria labels to interactive Circle graph

nishasy:
## Summary:
Make the Circle interactive graph screen reader accessible as outlined by the [Circle SRUX doc](https://khanacademy.atlassian.net/wiki/spaces/LC/pages/3455221794/Circle).

- Add aria label, aria-described-by, and aria-live for the outer circle
- Add aria label, aria-described-by, and aria-live for the radius point on the edge of the circle
- Add all the label and description strings to the `strings.ts` file for translations later

## Test plan:
- http://localhost:6006/iframe.html?args=&globals=&id=perseuseditor-widgets-interactive-graph--interactive-graph-circle&viewMode=story (Or use the storybook publish below)
- test the screen reader experience
- initial focus should include the additional info that's a part of the aria-describedby
- updating the circle position should read out the new info that part of the aria-label
- updating the radius point position should read out the new info that part of the aria-label


packages/perseus/src/widgets/interactive-graphs/graphs/circle.test.tsx
nishasy, line 1: Unfortunately, I don't think I can test the changes to the aria properties on update with jest. I don't even think I can test that the radius point's aria live value state updates when the circle is moved.

Someone please tell me if I'm wrong. I'm open to suggestions for how to test this stuff.

catandthemachines (approved):
Approved! If you can give the test one extra try to trigger a move event and then check that the aria-live changes. Other than that you look good. ❤️

packages/perseus/src/widgets/interactive-graphs/graphs/circle.test.tsx
catandthemachines: This article might have what you're looking for, the tool is called `fireEvent`:
https://vinayak-hegde.medium.com/js-react-and-customevent-testing-using-jest-and-react-testing-library-abb247fd758a

packages/perseus/src/widgets/interactive-graphs/graphs/circle.tsx
```diff
@@ -37,31 +38,101 @@ function CircleGraph(props: CircleGraphProps) {
     const {dispatch, graphState} = props;
     const {center, radiusPoint} = graphState;
 
+    const {strings} = usePerseusI18n();
+    const [radiusPointAriaLive, setRadiusPointAriaLive] = React.useState<
+        "off" | "polite"
```
anakaren-rojas, line 43: I'd recommend using the aria live type in
packages/perseus/src/widgets/interactive-graphs/types.ts

packages/perseus/src/widgets/interactive-graphs/graphs/circle.tsx
```diff
@@ -37,31 +38,101 @@ function CircleGraph(props: CircleGraphProps) {
...
                 cursor="ew-resize"
                 onMove={(newRadiusPoint) => {
+                    setRadiusPointAriaLive("polite");
                     dispatch(actions.circle.moveRadiusPoint(newRadiusPoint));
                 }}
             />
-        </>
+            {/* Hidden elements to provide the descriptions for the
+                circle and radius point's `aria-describedby` properties. */}
+            <g id={radiusId} style={{display: "hidden"}}>
```
anakaren-rojas, line 117: If I remember correctly, using `display: "hidden"` also hides it from the screen reader. Did you have any issues when testing this? 
Using `a11y.srOnly` might be an alternative if you do have issues 
packages/perseus/src/util/a11y.ts

anakaren-rojas (approved):

packages/perseus/src/widgets/interactive-graphs/graphs/circle.tsx
nishasy: I used `display: hidden` because I don't want the screen reader (or any user) to ever actually access this element. It's only there for its text and ID so that other stuff can refer to it. It seemed to work fine in my tests, so I think this should be okay?

packages/perseus/src/widgets/interactive-graphs/graphs/circle.test.tsx
nishasy: Added tests!

PR #1904, opened 2024-11-21
Upgrade Storybook v8.x

mark-fitzgerald:
## Summary:
Nothing is broken. Everything is happy.

## Test plan:
`yarn start` is happy
`yarn build-storybook; python3 -m http.server -d storybook-static;` -> open `http://localhost:8000` in a browser
`yarn build-storybook --stats-json` works and generates `/storybook-static/preview-stats.json`
`yarn dev` is happy
`yarn cypress:ci` is happy

PR #1929, opened 2024-11-28
Update TypeScript to version 5.7.2

kevinb-khan:
## Summary:
Wonder Blocks is updating Storybook to 8.x which requires using "moduleResolution": "bundler" in its tsconfig, see https://stackoverflow.com/questions/72193784/why-doesnt-the-exports-field-of-npm-work-in-typescript for details.  To fix this, I decided to update TypeScript in Wonder Blocks.  Instead of updating to the minimally required version I figured I may as well upgrade to the most recent version since I'm upgrading anyways.

Although we probably don't need to update TypeScript in all of our repos at the same time (unless we start using new syntax), we may as well.  That way, if a new Wonder Blocks package is published with new TS syntax it its types, Perseus will be ready.

In theory, this shouldn't change the output of any of the packages since we're using babel for compilation.

Issue: None

## Test plan:
- yarn typecheck

### DIFF
--- a/package.json
+++ b/package.json
@@ -120,7 +120,7 @@
         "storybook": "^7.6.17",
         "style-loader": "^3.3.3",
         "tiny-invariant": "^1.3.1",
-        "typescript": "^5.4.2",
+        "typescript": "5.7.2",
         "typescript-coverage-report": "^0.7.0",
         "vite-plugin-istanbul": "^5.0.0",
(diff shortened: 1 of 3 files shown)


jeremywiebe (commented):
Nice.  We're moving to Storybook 8 next week also: https://github.com/Khan/perseus/pull/1904  

Thanks for this!

jeremywiebe (approved):

PR #1930, opened 2024-11-28
Remove unused types for widgets that no longer exist

jeremywiebe:
## Summary:

I was working in the `perseus-types.ts` file a bit and noticed a few types that defined options for widgets that have been deleted. This PR cleans up the types. 

Issue: "none"

## Test plan:

`yarn typecheck` is fine

### DIFF
--- a/.changeset/strong-flies-thank.md
+++ b/.changeset/strong-flies-thank.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": major
+---
+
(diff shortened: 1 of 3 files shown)


.changeset/strong-flies-thank.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/perseus": major
```
jeremywiebe, line 2: This feels almost like overkill, but we're removing types from the exported package, so that constitutes a major version bump.

SonicScrewdriver (approved):
Looks good to me, and I don't see any usages in webapp either. Thank you for cleaning these up!

PR #1931, opened 2024-11-29
Allow WB/WS packages otherwise groups don't match anything

jeremywiebe:
## Summary:

I set up Dependabot updates for WonderBlocks and WonderStuff packages last week and set them so they'd be updated in groups. However, I missed adding them to the `allow` list, which prevented any updates to them. This PR adds them to the `allow` list so that Dependabot actually works. 

Issue: "none"

## Test plan:

Land and watch tonights Dependabot run.


jandrade (approved):
LGTM, good catch!

PR #1934, opened 2024-11-29
Improve some comments

jeremywiebe:
## Summary:

As it says, just some comments that I wanted to improve as I was working on SSS.

Issue: "none"

## Test plan:

Read the comments.

### DIFF
--- a/packages/perseus/src/renderer.tsx
+++ b/packages/perseus/src/renderer.tsx
@@ -1595,7 +1595,10 @@ class Renderer
     };
 
     /**
-     * Returns an array of widget ids that are empty.
+     * Returns an array of widget ids that are empty (meaning widgets where the
+     * learner has not interacted with the widget yet or has not filled in all
+     * fields).  For example, the `interactive-graph` widget is considered
+     * empty if the graph is in the starting state.
      */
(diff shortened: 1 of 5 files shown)


packages/perseus/src/renderer.tsx
```diff
@@ -1594,6 +1594,9 @@ class Renderer
         return state;
     };
 
+    /**
+     * Returns an array of widget ids that are empty.
+     */
     emptyWidgets(): ReadonlyArray<string> {
```
SonicScrewdriver, line 1600: To confirm my understanding of this one, this is determining if the user hasn't input or selected a value for a scorable widget, correct? 

I wonder if it would be helpful to provide a little explanation about what "empty" means just to make it easier to grok at a quick glance, but this is a totally ignorable nit.

SonicScrewdriver (approved):
Looks good, thanks for the cleanup!

packages/perseus/src/renderer.tsx
jeremywiebe: Good idea.  I've expanded it.  We can always keep clarifying if it's still not good enough. :)

PR #1935, opened 2024-11-29
Fix reviewers for dependabot - 'assignees' is only for individuals

jeremywiebe:
## Summary:

*sigh* I can't read. 

`assignees` => https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#assignees

> Use assignees to specify individual assignees for all pull requests raised for a package manager.

`reviewers` ==> https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#reviewers

> Use reviewers to specify individual reviewers or teams of reviewers for all pull requests raised for a package manager. You must use the full team name, including the organization, as if you were @mentioning the team.

Issue: "none"

## Test plan:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,8 +12,8 @@ updates:
           - dependency-name: "@khanacademy/eslint-plugin"
           - dependency-name: "@khanacademy/wonder-blocks-*"
           - dependency-name: "@khanacademy/wonder-stuff-*"
-      assignees:
-          - "@Khan/perseus"
+      reviewers:
+          - "Khan/perseus"
 
       # Grouped updates for Wonder Blocks and Wonder Stuff releases.
       # This helps us to stay in sync with the latest releases of these packages.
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -12,8 +12,8 @@ updates:
           - dependency-name: "@khanacademy/eslint-plugin"
           - dependency-name: "@khanacademy/wonder-blocks-*"
```
jandrade, line 13: would it be useful adding `Khan/wonder-blocks` for reviewing these changes? I'm not sure if that would be a good signal for us to know when we can update the `webapp` repo to have synced deps? or do you recommend something different?

How do you think the process should work now that dependabot is connected for these type of updates?

jeremywiebe: I can add that team, for sure.  If it gets noisy we can change it

mark-fitzgerald (approved):

PR #1936, opened 2024-11-29
Migrate PerseusWidgetsMap to use a merged interface to represent all available widgets

jeremywiebe:
## Summary:

Perseus is designed to be able to support widgets registered into it at runtime. These widgets can be built in a separate npm package or even in the hosting application. As a result, the way that we've defined the `widgets` prop of `PerseusRenderer` is too rigid and poses long-term maintenance issues. 

As I was working on slitting out the bits of validation from scoring that we'll still be able to do on the client-side, I faced the same challenge (we will have a map of widget IDs for validation). 

Working with Kevin Barabash, we figured out a way to define an interface that represents all registered widgets and then use some TypeScript types to generate the `PerseusWidgetsMap`. This allows us to extend the "known set" of widgets from outside of Perseus. 

Although this is not an immediate concern, it will become more and more troublesome to build maps of widget information that are hard-coded to our current set of widgets. 

Issue: "none"

## Test plan:

`yarn typecheck`
`yarn test`

I'll import this npm snapshot into webapp to check that all types are fine there also.

EDIT: I installed the snapshot package in webapp and it type checks! ✅


SonicScrewdriver (approved):
This is WAY better than our first attempt, thank you greatly for this work! This seems considerably more maintainable and readable. I'm not seeing anything that should cause any issues with the Extract Perseus Data logic, and if there's no type errors, this gets the all clear from me. 

Thank you for adding some truly excellent comments as well. :)

packages/perseus/src/perseus-types.ts
```diff
@@ -11,72 +11,83 @@ export type Size = [number, number];
...
-    [key in `python-program ${number}`]: PythonProgramWidget;
-} & {
-    [key in `radio ${number}`]: RadioWidget;
-} & {
-    [key in `sorter ${number}`]: SorterWidget;
-} & {
-    [key in `table ${number}`]: TableWidget;
-} & {
-    [key in `video ${number}`]: VideoWidget;
+    [Property in keyof PerseusWidgetTypes as `${PerseusWidgetTypes[Property]["type"]} ${number}`]: PerseusWidgetTypes[Property];
```
kevinb-khan, line 90: I think it would be good to document what the keys of this object look like, e.g. `radio 1`, `expression 2` etc. for people who aren't that familiar with all of the fancy TypeScript types that are being used here.

packages/perseus/src/perseus-types.ts
```diff
@@ -11,72 +11,83 @@ export type Size = [number, number];
...
+ *     }
+ * }
+ *
+ * type MyAwesomeNewWidget = WidgetOptions<'awesomeness', MyAwesomeNewWidgetOptions>;
+ * ```
+ *
+ * This interface can be extended through the magic of TypeScript "Declaration
+ * merging". Specifically, we augment this module and extend this interface.
+ *
+ * @see {@link https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation}
```
kevinb-khan, line 42: Thanks for adding this docstring.  I think it might also be worth mentioning that `MyAwesomeNewWidget` needs to be registered and what that looks like, but that might make more sense in the README.md.

kevinb-khan (approved):
LGTM.  Thanks for doing the due diligence to verify that this plan will work.  I left a few documentation suggestions in the comments – nothing blocking. 🚢

packages/perseus/src/perseus-types.ts
jeremywiebe: Good idea. I've extended the JSDoc comment

packages/perseus/src/index.ts
```diff
@@ -222,7 +222,9 @@ export type {
     PerseusRenderer,
     PerseusWidget,
     PerseusWidgetsMap,
+    PerseusWidgetTypes,
     MultiItem,
+    WidgetOptions,
```
handeyeco, line 227: Do you know that these will be used in Webapp or are we preemptively exporting them? If not, I wonder if it would be better to keep them internal until we have a use for them externally.

packages/perseus/src/perseus-types.ts
```diff
@@ -11,72 +11,89 @@ export type Size = [number, number];
...
+ * a new widget, an entry should be added to this interface. Note that this
+ * only registers the widget options type, you'll also need to register the
+ * widget so that it's available at runtime (@see
+ * {@link file://./widgets.ts#registerWidget}).
+ *
+ * Importantly, the key is not important and is not used anywhere outside of
+ * this interface so it is arbitrary.
+ *
+ * If you define the widget outside of this package, you can still add the new
+ * widget to this interface by writing the following in that pacakge that
```
handeyeco, line 28:
```suggestion
 * widget to this interface by writing the following in that package that
```

packages/perseus/src/perseus-types.ts
```diff
@@ -11,72 +11,89 @@ export type Size = [number, number];
...
+    PassageRefTarget: RefTargetWidget;
+    PassageRef: PassageRefWidget;
+    Passage: PassageWidget;
+    PhetSimulation: PhetSimulationWidget;
+    Python: PythonProgramWidget;
+    Plotter: PlotterWidget;
+    Radio: RadioWidget;
+    Sorter: SorterWidget;
+    Table: TableWidget;
+    Video: VideoWidget;
```
handeyeco, line 80: Not related to your PR and no action needed: can we leverage this or something like this to get rid of `basic-widgets`, `extra-widgets`, and `all-widgets`?

packages/perseus/src/perseus-types.ts
```diff
@@ -11,72 +11,89 @@ export type Size = [number, number];
...
+ *
+ * type MyAwesomeNewWidget = WidgetOptions<'awesomeness', MyAwesomeNewWidgetOptions>;
+ * ```
+ *
+ * This interface can be extended through the magic of TypeScript "Declaration
+ * merging". Specifically, we augment this module and extend this interface.
+ *
+ * @see {@link https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation}
+ */
+export interface PerseusWidgetTypes {
```
handeyeco, line 47: I think this is a step in the right direction, I just wish we could get to a place where the Renderer-layer of Perseus is completely unaware of the Widget-layer data which would mean finding ways to eliminate these type of widget union/map things.

packages/perseus/src/perseus-types.ts
jeremywiebe: I added a short note about registering widgets in this comment.

packages/perseus/src/index.ts
jeremywiebe: `PerseusWidgetTypes` needs to be exported so that it can be extended from anywhere outside of Perseus. @kevinb-khan has already had a use case for this while building a hackathon widget inside of webapp. 

`WidgetOptions` is a very useful helper type for defining a widget's widget options structure. 

Both are forward-looking, but it seemed reasonable to export them as part of this PR given the "widgets can be created outside of the @khanacademy/perseus" package nature.

packages/perseus/src/perseus-types.ts
jeremywiebe: I don't think so.  This interface is a registry of widget option types whereas the things you mentioned are the "runtime registries".  We could certainly merge all those three items together though.

packages/perseus/src/perseus-types.ts
jeremywiebe: Agreed! I've spent a decent amount of time thinking about how we could make the lower level types opaque to the Renderers, I don't have a solution yet.

PR #1937, opened 2024-11-30
Enable "react/recommended" lint ruleset and fix/suppress violations

jeremywiebe:
## Summary:

While running tests I saw warnings that React components rendered in a list were missing `key`s. This lead me down a path of "is there an ESLint rule for this?" Turns out, there's a nice ruleset for the React ESLint plugin that includes a rule for ensuring JSX elements have a `key` when in a list. 

So, this PR enables the `react/recommended` ESLint rules and fixes/suppresses things. 

Issue: "none"

## Test plan:

`yarn lint`

Also, I opened the affected stories to double-check things.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -49,6 +49,7 @@ function banImportExtension(extension) {
 module.exports = {
     extends: [
         "@khanacademy",
+        "plugin:react/recommended",
         // This config includes rules from storybook to enforce story best
         // practices
         "plugin:storybook/recommended",
@@ -308,6 +309,10 @@ module.exports = {
         /**
          * react
          */
+        "react/no-string-refs": "off", // on in recommended, but #legacy-code
+        "react/no-find-dom-node": "off", // on in recommended, but #legacy-code
+        "react/display-name": "off", // on in recommended, but we have many components that don't have a displayName
+        "react/no-unescaped-entities": "off", // on in recommended, but #legacy-code
         // This rule results in false-positives when using some types of React
         // components (such as functional components or hooks). Since
         // TypeScript is already checking that components are only using props
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -309,10 +309,10 @@ module.exports = {
         /**
          * react
          */
-        "react/no-string-refs": "off", // on in recommended, but #legacy-code
-        "react/no-find-dom-node": "off", // on in recommended, but #legacy-code
-        "react/display-name": "off", // on in recommended, but we have many components that don't have a displayName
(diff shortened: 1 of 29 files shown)


.eslintrc.js
```diff
@@ -308,6 +309,10 @@ module.exports = {
         /**
          * react
          */
+        "react/no-string-refs": "off", // on in recommended, but #legacy-code
+        "react/no-find-dom-node": "off", // on in recommended, but #legacy-code
+        "react/display-name": "off", // on in recommended, but we have many components that don't have a displayName
+        "react/no-unescaped-entities": "off", // on in recommended, but #legacy-code
```
jeremywiebe, line 315: I've turned off a few of the `react/recommended` rules because they were pretty pervasive. Some adventurous person can pick these off if that sounds like fun. :)

packages/perseus-editor/src/__stories__/content-preview.stories.tsx
```diff
@@ -4,7 +4,7 @@ import {
 } from "@khanacademy/perseus";
 import {View} from "@khanacademy/wonder-blocks-core";
 import {spacing} from "@khanacademy/wonder-blocks-tokens";
-import {useState} from "react";
+import * as React from "react";
```
jeremywiebe, line 7: Not importing React was violating the `react/react-in-jsx-scope` rule. React 18 apparently uses the new JSX runtime which doesn't require this, but I'm not 100% sure if we are using the new JSX runtime (we are in Storybook/Vite, but not sure about in our production builds). I'm going to add this here for now instead of disabling the rule.

.eslintrc.js
```diff
@@ -308,6 +309,10 @@ module.exports = {
         /**
          * react
          */
+        "react/no-string-refs": "off", // on in react/recommended, but we have #legacy-code
+        "react/no-find-dom-node": "off", // on in react/recommended, but we have #legacy-code
+        "react/display-name": "off", // on in react/recommended, but doesn't seem that useful to fix
+        "react/no-unescaped-entities": "off", // on in react/recommended, but we have #legacy-code
```
benchristel, line 315: It seems like unescaped HTML entities should be easy to fix, though it also seems low-value to fix them if they're in code that we're not actively working on.

benchristel (approved):
LGTM! Thanks for fixing these warnings!

.eslintrc.js
jeremywiebe: They were easy to fix. I enabled this rule, took 20 min and fixed them all. Thanks for the nudge. :)

PR #1940, opened 2024-12-02
feat(LEMS-2662): handle angle aria label updates

anakaren-rojas:
## Summary:
- Adds SR functionality for angle sides and vertex
- Adds variations of 6 variations of angle strings - initial side copy (x2), updated side copy (x2), initial vertex copy, updated vertex copy

Issue: LEMS-2662

### Testing Steps
1. Navigate to [chromatic](https://650db21c3f5d1b2f13c02952-gnmgrjkyab.chromatic.com/?path=/docs/perseuseditor-editorpage--docs)
2. Add an interactive graph widget and update the answer type to `Angle`
3. With screen reader on - move any of the points on the graph 
4. Enable `Show angle measures` and move points 
5. Enable `Allow reflex angles` and move points

### Expected behavior
1. Screen Reader reads point coordinates (always)
2. When `Show angle measures` is not enabled, screen reader does not read out angle measure 
3. When `Show angle measures` is enabled, angle measure is read out with vertex 
4. When `Show angle measures` and `Allow reflex angles` are enabled, Screen Reader reads out reflex angles

### Without angle measures
https://github.com/user-attachments/assets/505dae92-34d4-416b-ac36-d4e072526233

### With angle measures
https://github.com/user-attachments/assets/d727aefd-52a0-4e78-84d8-cca8cff18b9a


packages/perseus/src/widgets/interactive-graphs/graphs/angle.tsx
```diff
@@ -95,6 +100,97 @@ function AngleGraph(props: AngleGraphProps) {
...
+    const handleVertexMove = (coordIndex: number, newPoint: vec.Vector2) => {
+        updateVertexAriaLabel(newPoint);
+        return dispatch(actions.angle.movePoint(coordIndex, newPoint));
+    };
+
+    React.useEffect(() => {
+        updateVertexAriaLabel(coords[1]);
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [showAngles]);
+
```
benchristel, line 193: I think we could avoid using the `useState` and `useEffect` hooks here if we:

- compute aria labels from the `graphState` prop
- use aria-live to notify the user of changes to the graph

This would ensure that the SR view and the visible view never get out of sync, since both would be pure functions of `graphState`. Currently, all the individual graph components are stateless (or nearly so — there are a couple `useState`s for hover and dragging effects) and I'd like to keep it that way.

Wonder Blocks is [planning an API for sending aria-live notifications](https://khanacademy.atlassian.net/wiki/spaces/WB/pages/3418095646/Announcer+Implementation+Spec), but since it's not built yet, I think what we could do is write a minimal implementation of the same API and swap in the WB one when it's ready. I'd be down to pair or consult on that if you want.

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.test.ts
```diff
@@ -104,3 +108,66 @@ describe("shouldDrawArcOutside", () => {
...
+        const coords: [Coord, Coord, Coord] = [
+            [0, 0],
+            [7, 0.5],
+            [12.5, 2.5],
+        ];
+        const vertex = coords[1];
+        expect(getWholeAngleMeasure(coords, vertex)).toBe(184);
+        expect(getWholeAngleMeasure(coords, vertex)).not.toBe(
+            184.08561677997488,
+        );
```
benchristel, line 145: This second assertion will always pass if the first assertion passes, so I think we don't need it. If you want to document what the value _would_ be if it weren't rounded, I think a comment would be fine.

```suggestion
```

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.tsx
```diff
@@ -144,24 +145,44 @@ interface AngleProps {
...
+    vertex: vec.Vector2,
+    allowReflexAngles: boolean = false,
+) {
+    // Check if the points are clockwise or not, depending on whether we allow reflex angles
+    const areClockwise = clockwise([...coords, vertex]);
+    const shouldReverseCoords = areClockwise && !allowReflexAngles;
+
+    // Reverse the coordinates accordingly to ensure the angle is calculated correctly
+    return shouldReverseCoords ? coords : coords.reverse();
+}
```
benchristel, line 168: Thanks for refactoring this code!

It seems like this duplicates some of the logic from `getClockwiseAngle` in `interactive-graphs/math/angles.ts`. Could we use that function to calculate the angle measure here? `getClockwiseAngle` is used for scoring, and it would be nice if we could use the same function to calculate the angle label.

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.test.ts
benchristel, line 1: I like these tests!

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.test.ts
```diff
@@ -104,3 +108,66 @@ describe("shouldDrawArcOutside", () => {
...
+            [1, 0],
+            [0, 1],
+        ];
+
+        expect(getClockwiseCoords(coords, coords[0], true)).toEqual([
+            [0, 1],
+            [1, 0],
+            [0, 0],
+        ]);
+    });
```
benchristel, line 172: I'm a bit confused by this. The function is called `getClockwiseCoords` but it can return counterclockwise coords?

benchristel (changes requested):
This is exciting!

Requesting changes because we're ultimately going to want a different approach to aria-live notifications (see my inline comment about avoiding useState and building for the upcoming Wonder Blocks API) and I'd prefer to start designing for that now.

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.tsx
anakaren-rojas: oh yes, you're right, it looks like it combines both the `getClockwiseCoords` and the `getWholeAngleMeasure` methods into one! let me test it out

anakaren-rojas: ok, so I tested this out and using the `getClockwiseAngle` function introduces a regression in the graph 😢
image
The very small difference between how we determine whether the coordinates should be reversed is the culprit: 
In `getClockwiseAngle`
`const shouldReverseCoords = !areClockwise || allowReflexAngles;`
In `getWholeAngleMeasure` 
`const shouldReverseCoords = areClockwise && !allowReflexAngles;`

PR #1945, opened 2024-12-04
[Storybook] Update styles to be more accurate to prod

nishasy:
## Summary:
Perseus Storybook in its current state does not accurately
reflect how these component look in prod.

A couple things are added here to fix this:
1. Stories are wrapped with the `framework-perseus` class.
2. Webapp styles are imported
  - `box-sizing: border-box` in particular affects all the
    spacing (margins and padding)

Issue: none

## Test plan:
- Go to http://localhost:6006/?path=/docs/perseus-components-number-input--docs
- Confirm that the inputs have rounded bordersa, and the
  "mini", "small", and "normal" stories show different sizes.
- Go to http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-ray
- Scroll to the start coordinates panels
- Use the dev tools to confirm that the width is 308px
  (previously 336px)

## Screenshots:

| Before | After |
| --- | --- |
| Screenshot 2024-12-03 at 6 03 03 PM| Screenshot 2024-12-03 at 6 02 37 PM | 
| Screenshot 2024-12-03 at 6 02 56 PM | Screenshot 2024-12-03 at 6 02 47 PM |


.storybook/story-wrapper.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from "react";
+
+import "./webapp-styles.less";
+
+/**
+ * Wrapper for Storybook stories that
+ * 1. Renders the story inside of a .framework-perseus container
+ * 2. Includes the global styles from webapp
```
jeremywiebe, line 8: I would love it if we could isolate Perseus from these types of resets that trickle down into Perseus from outside. It makes sense to try to align styling with that of our primary consumer (webapp), but, if possible, could we make these styles/resets explicit in our Perseus styling system instead of mimicking them in Storybook?

.storybook/webapp-styles.less
```diff
@@ -0,0 +1,136 @@
+/**
+ * This file contains styles that are used in webapp that ultimately
```
jeremywiebe, line 2: I'd like to avoid, if possible, references to projects outside of Perseus. I realize that might be pedantic, but Perseus is open source and although webapp is our primary consumer, I think it's worthwhile to stay at least somewhat agnostic of where Perseus is used.

Also, as stated in `story-wrapper.tsx`, I'm loathe to copy this much styling from an outside consumer just in Storybook. It gives us a false sense of what final styling will be in any given location if we've duplicated the styles only in Storybook, but not in our production styles.

jeremywiebe (commented):
I'm going to hold off approving. I'd love to find a way to set these styles as part of the Perseus bundle we deliver instead of mimicking things in Storybook.

.storybook/global.less
```diff
@@ -0,0 +1,42 @@
...
+ * .storybook/preview.js file so that the components can have the same
+ * styles as prod when viewed within Storybook.
+ */
+
+// Variables
+@offBlack: #21242c;
+@wonderBlocksFontFamily: "Lato", "Noto Sans", "Helvetica", "Corbel", sans-serif;
+
+// Layers
+@layer reset, shared;
```
mark-fitzgerald, line 13: I think this needs to include `legacy`. The rest of our PROD site show:
```
@layer reset, shared, legacy;
```

.storybook/global.less
```diff
@@ -0,0 +1,42 @@
...
+        height: 100%;
+    }
+
+    body {
+        font-family: @wonderBlocksFontFamily, "Helvetica", "Corbel", sans-serif;
+        font-size: 14px;
+        margin: 0;
+        color: @offBlack;
+        line-height: 1.4;
+        min-width: 0 !important;
```
mark-fitzgerald, line 32: Does this really need `!important`?

.storybook/global.less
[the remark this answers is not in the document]
nishasy: @jeremywiebe but maybe not in the `.storybook` folder, right? I can move it where `perseus-renderer.less` is?

.storybook/story-wrapper.tsx
```diff
@@ -0,0 +1,22 @@
...
+ */
+function StoryWrapper(props) {
+    // Most of our components have an expectation to be
+    // rendered inside of a .framework-perseus container.
+    // We want to make sure we can include it here, since it
+    // can also affect the styling.
+    return (
+        <div className="framework-perseus box-sizing-border-box-reset">
+            {props.children}
+        </div>
```
mark-fitzgerald, line 18: Do we need a separate file for this? Could we not just wrap the `preview.tsx` content in this `<div>` directly?

.storybook/global.less
nishasy: That's true, but I left it out since `@legacy` isn't defined anywhere in perseus styles. Should I add it anyway?

.storybook/global.less
nishasy: Ah, that was already there in webapp with the comment

```
/* Needs !important because we might load shared.css later */
```

But I don't think that's relevant anymore? I don't see a `shared.css` anywhere. I'll remove the `!important`.

mark-fitzgerald (approved):
I like the call-outs that Jeremy made. I added a couple additional items, but all else looks good.

.storybook/global.less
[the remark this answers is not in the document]
jeremywiebe: Correct.  This will become a part of our production Perseus CSS so it should go alongside the `perseus-renderer.less` file.

jeremywiebe (approved):
Thanks for doing this Nisha. Hoping this will reduce inconsistencies between Storybook and prod even further. 👍

.storybook/global.less
mark-fitzgerald: Yes, please add it. Although `legacy` is last in the list, and "shouldn't" affect anything when imported due to definition order, we should be consistent with how it is declared in other parts in PROD.

That said, since this is the first time that we have declared and used layers in Perseus, we should verify how this comes across in PROD. Perhaps a ZND would help? I want to make sure that we aren't making sub-layers due to how these styles are imported.

.storybook/story-wrapper.tsx
nishasy: Ah yeah. Now that I removed the import, I'll change it back to how it was before.

PR #1946, opened 2024-12-04
SSS: Move group scoring into group scorer

jeremywiebe:
## Summary:

This PR refactors the scoring around `group` so that it follows the same pattern as all the other widgets. This means that our main `scoreWidgetsFunctional` can be simplified to remove the special case for the `group` widget. 

All tests still pass!

Issue: LEMS-2561

## Test plan:

`yarn test`
`yarn typecheck`

### DIFF
--- a/.changeset/stupid-apricots-retire.md
+++ b/.changeset/stupid-apricots-retire.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": patch
+---
+
+Remove debugging call in GraphSettings component
--- a/packages/perseus-editor/src/components/graph-settings.tsx
+++ b/packages/perseus-editor/src/components/graph-settings.tsx
@@ -560,8 +560,9 @@ const GraphSettings = createReactClass({
                             />
                             <InfoTip>
                                 <p>
-                                    Create an image in graphie, or use the "Add
-                                    image" function to create a background.
+                                    Create an image in graphie, or use the
+                                    &quot;Add image&quot; function to create a
+                                    background.
                                 </p>
                             </InfoTip>
                         </div>
--- a/packages/perseus-editor/src/components/graph-settings.tsx
+++ b/packages/perseus-editor/src/components/graph-settings.tsx
@@ -160,7 +160,6 @@ const GraphSettings = createReactClass({
 
         // @ts-expect-error - TS2531 - Object is possibly 'null'. | TS2339 - Property 'value' does not exist on type 'Element | Text'.
         const url = ReactDOM.findDOMNode(this.refs["bg-url"]).value; // eslint-disable-line react/no-string-refs
-        url; //  ?
         if (url) {
             Util.getImageSize(url, (width, height) => {
                 if (this._isMounted) {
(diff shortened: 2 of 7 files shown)


packages/perseus/src/renderer-util.ts
```diff
@@ -53,27 +53,13 @@ export function emptyWidgetsFunctional(
...
-            score = Util.flattenScores(scores);
-        } else if (scorer) {
-            score = scorer(
-                userInput as UserInput,
-                widget.options,
-                strings,
-                locale,
-            );
-        }
+        const score = scorer?.(
```
Myranae, line 57: Question: I don't remember seeing a function called like this before. So you can do a null check before calling a function by adding a question mark and a period?

packages/perseus/src/renderer-util.ts
```diff
@@ -53,27 +53,13 @@ export function emptyWidgetsFunctional(
...
-            );
-            score = Util.flattenScores(scores);
-        } else if (scorer) {
-            score = scorer(
-                userInput as UserInput,
-                widget.options,
-                strings,
-                locale,
-            );
-        }
```
Myranae, line 76: Question: It looks like you are pulling out the special case for group from the emptyWidgetsFunctional function here, but the summary for the PR says the goal was to remove this from the scoreWidgetsFunctional one. Did you want to remove it from there as well?

Myranae (commented):
Added a couple questions, one about where you removed the special case for group. I think it looks good otherwise though. You're basically pulling out the special part of the functional scoring functions and putting that in as the scorer for group, right?

handeyeco (approved):
Looks good to me! Can't _think_ of any hidden implications. 🤷

Tests pass and it makes the code cleaner. 🚀

packages/perseus-editor/src/components/graph-settings.tsx
```diff
@@ -160,7 +160,6 @@ const GraphSettings = createReactClass({
 
         // @ts-expect-error - TS2531 - Object is possibly 'null'. | TS2339 - Property 'value' does not exist on type 'Element | Text'.
         const url = ReactDOM.findDOMNode(this.refs["bg-url"]).value; // eslint-disable-line react/no-string-refs
-        url; //  ?
```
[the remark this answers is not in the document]
jeremywiebe: That's a special comment that the Wallaby.js tooling uses to emit the value to the right of the comment when running tests. 

So the `url; // ?` line was only ever for debugging.

image

.changeset/stupid-apricots-retire.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": patch
+---
+
+Remove debugging call in GraphSettings component
```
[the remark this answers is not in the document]
jeremywiebe: It's the changes in `perseus-editor` which was just removing a debugging statement.

packages/perseus/src/renderer-util.ts
jeremywiebe: Yes.  It's call an ["optional chaining"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) operator.  

With it, you can safely access a property on an object, even if the object is null/undefined at runtime. The result of the call will short-circuit and be `undefined` at runtime. It avoids 2 extra lines of code for an `if (x != null) {` and the closing `}`.

packages/perseus/src/renderer-util.ts
jeremywiebe: Good catch!  I've fixed `scoreWidgetsFunctional` now too!

PR #1953, opened 2024-12-05
[Dropdown] [BUGFIX] Update aria-label logic

daniellewhyte:
## Summary:
This PR contains two fixes for the aria-label in the dropdown widget: 
1. Change the logic of the aria label
Follow up from Mark's comment on #1845 
This commit ensures that the visible label does not override the aria label. 
Motivation: There may be additional info for the screen reader in the aria label that may not be in the visible label.

This changes also removes the aria-labelledby from the Dropdown which was causing the screen reader output to be garbled in Safari. 

2. Pass in a space `(" ")` for empty placeholder.
Wonder Blocks adds a nbsp when an empty string is passed in as one of the option items. This does not work well with screen readers. When testing with Voice Over on Safari and Firefox, it reads out the space and the aria label as a series of characters (instead of words). 

https://github.com/user-attachments/assets/6ed58cfe-e849-4c96-a661-7d8c25f25617

The root cause for this should be fixed in Wonder Blocks and I’ve started a discussion with the WB team [here](https://khanacademy.slack.com/archives/C4PE1QM5Y/p1733761447231789) (I haven't been able to find a solution in WB yet). Given that this bug exists on prod in webapp, it feels appropriate to go ahead and implement this fix in Perseus. 

Issue: LIT-1476, LIT-1482

## Test plan:

Test with the new [Dropdown with empty placeholder story](http://localhost:6007/?path=/story/perseus-widgets-dropdown--dropdown-with-empty-placeholder)
- Dropdown displays inline when nothing is selected
- Screen reader does not say "space" when no option is selected. And does not spell out aria-label

https://github.com/user-attachments/assets/b4ee0524-50e7-4000-a66e-2bb0590d5c3f


mark-fitzgerald (approved):
Thanks for following up with the WB team.

PR #1955, opened 2024-12-05
Remove MultiRenderer

handeyeco:
## Summary:
GTP is dead and MultiRenderer existed for GTP. Removing it simplifies things quite a bit.

Issue: [LEMS-2675](https://khanacademy.atlassian.net/browse/LEMS-2675)

## Test plan:
Shouldn't affect anything, no longer in use

[LEMS-2675]: https://khanacademy.atlassian.net/browse/LEMS-2675?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/perseus-editor/src/i18n.ts
jeremywiebe, line 1: I don't see these used anywhere. Can we delete these too?

jeremywiebe (approved):
Exciting to drop support for these MultiItems. 

Not sure if we want to, but I think we can delete the `data/item-missing-answerArea.json` parsing test file and its associated snapshot too then.

PR #1961, opened 2024-12-06
Remove deprecated/unused support for examples in Renderer (keeping for numeric-input though)

jeremywiebe:
## Summary:

While working on SSS, I came across the `examples()` function on the `Renderer` that was causing me issues (with threading client-side validation in). 

I checked and it is no longer used anywhere. Note that the `input-number` and `numeric-input` use their local `examples()` functions still to pass a set of examples to the `InputWithExamples` component, which we're keeping.

Issue: LEMS-2561

## Test plan:

`yarn test`
`yarn typecheck`
Checked in host webapp and the function is not used.


packages/perseus/src/widgets/numeric-input/numeric-input.tsx
```diff
@@ -123,8 +123,11 @@ export class NumericInput
         isFocused: false,
     };
 
-    // TODO(Nicole, Jeremy): This is maybe never used and should be removed
```
jeremywiebe, line 126: This _is_ used, but only internal to the widget.

handeyeco (approved):
Nice, love to see a smaller external API

PR #1962, opened 2024-12-06
Move scoring out of general Util into scoring util file

jeremywiebe:
## Summary:

Like #1948, I was working in `utils.ts` and found it annoying how there were so many different utils in one file (graphing support, scoring, word/text manipulation). So this just pulls scoring-related behaviour out into a new file. This means these symbols are no longer on the `Utils` object, but I checked and they aren't used in Perseus nor in consuming apps.

Issue: LEMS-2561

## Test plan:

`yarn test`
`yarn typecheck`


packages/perseus/src/__tests__/util.test.ts
```diff
@@ -1,21 +1,4 @@
-import Util, {isCorrect} from "../util";
-
-describe("isCorrect", () => {
```
jeremywiebe, line 3: All deleted tests here just moved to `scoring.test.ts`

packages/perseus/src/multi-items/multi-renderer.tsx
```diff
@@ -47,7 +47,7 @@ import {DependenciesContext} from "../dependencies";
 import HintsRenderer from "../hints-renderer";
 import {Log} from "../logging/log";
 import Renderer from "../renderer";
-import Util from "../util";
+import {combineScores, keScoreFromPerseusScore} from "../util/scoring";
```
jeremywiebe, line 50: Also switched to ES6 exports in the new file.

Myranae (approved):
Looks good! Just have to remove that unused import for the failing check :)

.changeset/clever-cars-eat.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/perseus": minor
```
handeyeco, line 2: I'd argue this is a major change. `Util` is exported from `index.ts`:

``` TS
export {default as Util} from "./util";
```

You'll see that this breaks Webapp because I used `Util.keScoreFromPerseusScore` in my recent change. So:

1. Probably a major change
2. I'd double check the utils that you moved to make sure they're not used in Webapp (and export the ones that are)

handeyeco (approved):
I think you'll need to make changes to `index.ts` to make sure the utils that are used in Webapp are still accessible. Otherwise a great change. 👍

.changeset/clever-cars-eat.md
jeremywiebe: Agreed. I didn't catch the one usage of `keScoreFromPerseusScore` outside of Perseus.  Fixing.

PR #1963, opened 2024-12-06
Fix Jest inline snapshots

jeremywiebe:
## Summary:

While writing some tests, I wanted to use inline snapshots, but Jest complained that inline snapshots aren't supported on Prettier v3 (which we use). 

The solution is to install Prettier v2 beside v3 and customize Jest to use v2. 

Support for Prettier v3 has already [landed](https://github.com/jestjs/jest/pull/14566) and will be in Jest 30 but the last I saw was an [issue comment](https://github.com/jestjs/jest/issues/14305#issuecomment-1784898007) from Oct 2023 and even then the author was unsure when it would be released. I'll just use the workaround for now.

Issue: "none"

## Test plan:

Find a test that uses `.toMatchInlineSnapshot()` and delete the snapshot string passed to the function.
Save the file
`yarn test`
Note that the snapshot string is back!

https://github.com/user-attachments/assets/c33d2ac3-adae-43e1-8414-092880536d1d

### DIFF
--- a/config/test/test.config.js
+++ b/config/test/test.config.js
@@ -42,6 +42,9 @@ const pkgMap = fs
 /** @type {import('jest').Config} */
 module.exports = {
     rootDir: path.join(__dirname, "../../"),
+    // Jest does not support Prettier v3!
+    // See: https://jestjs.io/docs/configuration/#prettierpath-string
+    prettierPath: require.resolve("prettier-2"),
     transform: {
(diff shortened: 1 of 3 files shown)


jeremywiebe:
Closing. This doesn't work as the `prettier-2` instance gets installed in `node_modules/.bin` which then causes formatting issues when we check formatting with `yarn prettier --check` in GitHub Actions. 

Abandoning for now. Wait for Jest v30. 😓

PR #1965, opened 2024-12-06
Refactor LabelImage to separate out answers from userInput into scoringData

Myranae:
## Summary:
This PR updates LabelImage so that answers are no longer available in the userInput object. This allows the scoring function to have both a userInput parameter and a scoringData parameter to keep answers separate from the user's input with the goal of supporting server side scoring.

PerseusLabelImageMarker and MarkerType contained the same properties. As such, I simplified the code a bit and removed PerseusLabelImageMarker.

In addition, several locations were referencing the wrong types, so those were updated to reference the correct ones. Also, new tests were added confirming the output of `getUserInput` does not contain answers, that `scorePerseusItem` returns the correct results, and that the widget renders correctly if answers are not present in the JSON blob.

Issue: LEMS-2440

## Test plan:
- Confirm all checks pass


packages/perseus/src/validation.types.ts
```diff
@@ -138,12 +141,12 @@ export type PerseusInteractiveGraphRubric = {
...
-/* TODO(LEMS-2440): Should be removed or refactored. Grading info may need
-    to be moved to the rubric from userInput. */
-export type PerseusLabelImageRubric = Empty;
+export type PerseusLabelImageScoringData = {
+    markers: ReadonlyArray<MarkerAnswers>;
+};
 
 export type PerseusLabelImageUserInput = {
-    markers: ReadonlyArray<InteractiveMarkerType>;
+    markers: ReadonlyArray<Omit<InteractiveMarkerType, "answers">>;
```
Myranae, line 149: These are the main changes. Changed `PerseusLabelImageUserInput` so it no longer includes `answers` in the array. Then I moved the parts of `InteractiveMarkerType` needed for scoring to PerseusLabelImageScoringData`.

packages/perseus/src/widgets/label-image/label-image.tsx
```diff
@@ -72,8 +72,6 @@ type Point = {
 
 type LabelImageProps = ChangeableProps &
     DependencyProps &
-    // TODO: there's some weirdness in our types between
-    // PerseusLabelImageMarker and InteractiveMarkerType
```
Myranae, line 76: Was able to resolve this by removing `PerseusLabelImageMarker`

packages/perseus/src/widgets/label-image/score-label-image.ts
```diff
@@ -43,15 +43,21 @@ export function scoreMarker(
...
+    const combinedData = userInput.markers.map((marker, index) => {
+        return {
+            ...marker,
+            answers: scoringData.markers[index].answers,
+        };
+    });
+    const allMarkerData = {markers: combinedData};
 
-    for (const marker of userInput.markers) {
+    for (const marker of allMarkerData.markers) {
```
Myranae, line 60: Data was combined here so that `marker` can have all the available information as `scoreMarker` requires an object of `InteractiveMarkerType`, which includes answers.

packages/perseus/src/widgets/label-image/label-image.tsx
```diff
@@ -318,8 +316,10 @@ export class LabelImage
         return _getPromptJSON(this.props, this.getUserInput());
     }
 
-    // TODO(LEMS-2544): Investigate impact on scoring; possibly pull out &/or remove rubric parameter.
-    showRationalesForCurrentlySelectedChoices(rubric: PerseusLabelImageRubric) {
+    // TODO(LEMS-2544): Investigate impact on scoring; possibly pull out &/or remove scoringData parameter.
+    showRationalesForCurrentlySelectedChoices(
```
Myranae, line 320: This is a bit of a problem as it requires scoringData on the client-side it seems like. Since this has a TODO associated with it, I left it for now.

handeyeco: This is a known issue: https://khanacademy.atlassian.net/browse/LEMS-2544

Luckily AX won't use rationales so we can punt on this for now.

packages/perseus/src/validation.types.ts
handeyeco, line 149: It's a little higher-risk, but I think we should make this test case pass:

``` TS
    // label-image.test.ts

    describe("getUserInput", () => {
        it("doesn't include answer in getUserInput", async () => {
            // render component
            const {renderer} = renderQuestion(textQuestion);

            const userInput = renderer.getUserInputMap();

            expect(userInput).toEqual({
                "label-image 1": {
                    markers: [
                        {
                            label: "The fourth unlabeled bar line.",
                            x: 25,
                            y: 17.7,
                        },
                        {
                            label: "The third unlabeled bar line.",
                            x: 25,
                            y: 35.3,
                        },
                        {
                            label: "The second unlabeled bar line.",
                            x: 25,
                            y: 53,
                        },
                        {
                            label: "The first unlabeled bar line.",
                            x: 25,
                            y: 70.3,
                        },
                    ],
                },
            });
        });
    });
```

I think there should be another set of tests too:
1. Render a LabelImage widget using ItemData that has no answers (to simulate how it will be on the FE)
2. Answer the LabelImage (both wrong and right)
3. Use `getUserInput` on the Renderer to get the user input
4. Use `scorePerseusItem` to score the user input

More than anything, I'm just surprised that this ticket ended up being more of a types change whereas I was thinking it would end up being more of a logic change. Not saying you're wrong, it just makes me worried we're missing something.

.changeset/many-penguins-hug.md
```diff
@@ -0,0 +1,6 @@
+---
+"@khanacademy/perseus": minor
```
handeyeco, line 2: Since you changed PerseusTypes in a way that's not backwards compatible, I'd call this a major/breaking change.

packages/perseus/src/widgets/label-image/score-label-image.ts
handeyeco: No action needed, but my gut feeling is that this is needless misdirection. I kind of feel we should just change `scoreMarker`:

``` TS
function scoreMarker(
    userInput: APieceOfUserInputType,
    scoringData: APieceOfScoringDataType,
): InteractiveMarkerScore;
```

I'm fine punting on this though.

packages/perseus/src/perseus-types.ts
```diff
@@ -1059,17 +1060,6 @@ export type PerseusLabelImageWidgetOptions = {
     static: boolean;
 };
 
-export type PerseusLabelImageMarker = {
```
handeyeco, line 1062: @jeremywiebe is particular when it comes to changing `perseus-types.ts`, I would have him sign off first.

handeyeco (commented):
I think this is the right direction. I'm wondering though:
1. Will LabelImage render if we provide it with PerseusData sans-`answers`?
2. Will `getUserInput` work sans-`answers`?
Knowing how good you are about manual testing, I'm guessing the answer is "yes." I'd just like to see some automated tests to confirm.

packages/perseus/src/widgets/label-image/types.ts
```diff
@@ -1,19 +1,21 @@
 // Base marker, with the props that are set by the editor.
-export type MarkerType = {
-    // The list of correct answers expected for the marker.
+export type MarkerAnswers = {
```
handeyeco, line 2: These types feel off to me. I'm wondering if it's actually something like:

``` TS
// stuff to render
type LabelImageMarker = {
    // Reveal the correctness state of the user selected answers for the marker.
    showCorrectness: "correct" | "incorrect";
    focused: boolean;
    // Translatable Text; The text to show for the marker. Not displayed directly to the user
    label: string;
    // The marker coordinates on the question image as percent of image size.
    x: number;
    y: number;
}

type LabelImageMarkerScoringData = {
    // The list of correct answers expected for the marker. Often only one but can have multiple
    answers: ReadonlyArray<string>;
}

type LabelImageMarkerUserInput = {
    // The user selected list of answers, used to grade the question.
    selected: ReadonlyArray<string>;
}

type LabelImageFullMarker = LabelImageMarker & LabelImageMarkerScoringData & LabelImageMarkerUserInput;
```

Myranae: Updated! Still need to add recommended tests and do additional manual testing though.

packages/perseus/src/widgets/label-image/score-label-image.ts
Myranae: Fixed!

packages/perseus/src/perseus-types.ts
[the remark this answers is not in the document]
jeremywiebe: Further, because TypeScript types match if they are the same shape, I was thinking about ScoringData and ValidationData for each widget as a matching subset of the widget options in this file.

jeremywiebe (commented):
Overall I like the direction of this PR. One change I'd like to see is that we keep the type defined in `perseus-types.ts` and then define the ScoringData for it in `validation.types.ts`.

packages/perseus/src/perseus-types.ts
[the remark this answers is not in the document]
Myranae: Do you mean use the types in this file to define the ScoringData and ValidationData types?

packages/perseus-editor/src/widgets/label-image-editor.tsx
```diff
@@ -28,7 +31,7 @@ type Props = {
     imageWidth: number;
     imageHeight: number;
     // The list of label markers on the question image.
-    markers: ReadonlyArray<MarkerType>;
+    markers: ReadonlyArray<LabelImageMarker & LabelImageMarkerScoringData>;
```
catandthemachines, line 34: Nit: I notice your creating this union in a lot of files. Can you combine them in the types object so you don't have to keep joining them in every file?

catandthemachines (approved):
The code looks good to me! Though I would encourage you to get an approval from Matthew/Jeremy as they have more context into the scoring code 😉

packages/perseus-editor/src/widgets/label-image-editor.tsx
Myranae: Great suggestion! I'll set that up. :)

packages/perseus/src/widgets/label-image/label-image.tsx
```diff
@@ -310,21 +305,27 @@ export class LabelImage
...
 
     getUserInput(): PerseusLabelImageUserInput {
-        const {markers} = this.props;
-        return {markers};
+        return {
+            markers: this.props.markers.map((marker) => ({
+                selected: marker.selected,
+                label: marker.label,
+            })),
+        };
```
jeremywiebe, line 313: @handeyeco for visibility, Tamara and I re-wrote this function to build a return value by explicitly picking the pieces we need instead of simply return a value out of props. This help stop guarantee we're really only using values in the type and not depending elsewhere on keys that may get stripped when we do the Scoring Data stripping stuff.

packages/perseus/src/widgets/label-image/score-label-image.test.ts
```diff
@@ -8,11 +8,23 @@ const emptyMarker = {
...
+    selected: [],
+    x: 0,
+    y: 0,
+} as const;
+
+const emptyScoringData = {
+    label: "",
+    answers: [],
+    x: 0,
+    y: 0,
```
jeremywiebe, line 22: I think we dropped `x` and `y` from this type, didn't we?

packages/perseus/src/widgets/label-image/score-label-image.test.ts
```diff
@@ -8,11 +8,23 @@ const emptyMarker = {
     y: 0,
 } as const;
 
+const emptyUserInput = {
+    label: "",
+    selected: [],
+    x: 0,
+    y: 0,
```
jeremywiebe, line 15: I think we dropped `x` and `y` from this type, didn't we?

packages/perseus/src/widgets/label-image/score-label-image.test.ts
```diff
@@ -73,138 +77,224 @@ describe("scoreMarker", function () {
 
 describe("scoreLabelImage", function () {
     it("should not grade non-interacted widget", function () {
-        const state = {
+        const userInput = {
             markers: [
                 {
-                    ...emptyMarker,
+                    ...emptyUserInput,
```
jeremywiebe, line 83: I think once we update this object to only have what's in the user input type, we might not need it. It might be more clear to inline the single `selected: []` bit here instead of spreading this object at that point. I'll leave it to you to decide which way you want to go. :)

jeremywiebe (approved):

packages/perseus/src/widgets/label-image/score-label-image.test.ts
Myranae: Updated the tests to remove these empty types.

packages/perseus/src/widgets/label-image/__tests__/label-image.test.ts
```diff
@@ -709,4 +714,110 @@ describe("LabelImage", function () {
             });
         });
     });
+
+    describe("getUserInput", () => {
+        it("doesn't include answer in getUserInput from getUserInputMap", () => {
```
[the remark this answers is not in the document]
Myranae: I think the focus _was_ supposed to be making sure answer was not included, but it's also correct to confirm it's just returning the correct userInput, especially since I didn't write any expect statements specifically checking that answers are absent. I do think it's important to emphasize that the goal is to have answers out of user input, but I guess that will be obvious from the shape of UserInput.

packages/perseus/src/widgets/label-image/__tests__/label-image.test.ts
```diff
@@ -709,4 +714,110 @@ describe("LabelImage", function () {
...
+    it("should render a text question without answers", () => {
+        // @ts-expect-error - Type not assignable to PerseusRenderer (answers removed)
+        renderQuestion(textWithoutAnswersQuestion);
+
+        const text = screen.getByText(
+            "Carol created a chart and a bar graph to show how many of each type of vehicle were in her supermarket parking lot.",
+        );
+
+        // expect(container).toMatchSnapshot();
+        expect(text).toBeInTheDocument();
```
handeyeco, line 821: This is closer but not quite what I was suggesting:

> - Render a LabelImage widget using ItemData that has no answers (to simulate how it will be on the FE)
> - Answer the LabelImage (both wrong and right)
> - Use getUserInput on the Renderer to get the user input
> - Use scorePerseusItem to score the user input

So you're rendering a LabelImage widget without answers, but only asserting that the content around the widget renders (the text you're getting isn't a part of the widget). What I'm proposing is that we repeat "can be answered correctly when correct option is picked for the marker" and "can be answered incorrectly when incorrect option picked for the marker" using `textWithoutAnswersQuestion`.

Myranae: We ended up deciding to handle rendering without answers with future related work. Kept this test to show that rendering doesn't rely on having answer data (the answer arrays are still in the test data, but they are empty. Can modify or remove this test when we start fully handling rendering with no answer property.

PR #1966, opened 2024-12-06
[Numeric Input] - BUGFIX - Adjust color contrast for tooltip text

mark-fitzgerald:
## Summary:
The text in the numeric input tooltip was slightly out of spec for accessibility. This change ensures that the text has a minimum 4.5:1 contrast ratio.

## Test plan:
1. Launch Storybook
2. Navigate to Perseus Editor => Editor => [Demo](http://localhost:6006/?path=/story/perseuseditor-editorpage--demo)
3. Add a Numeric Input widget
4. Configure the widget to have any number of format options
5. Place focus in the input field in the preview for the widget
6. Check the color value in the browser inspector for the text in the tooltip (it should be greater than 4.5:1)

### Before
Color Contrast - Before

### After
Color Contrast - After

### DIFF
--- a/packages/perseus/src/styles/perseus-renderer.less
+++ b/packages/perseus/src/styles/perseus-renderer.less
@@ -544,7 +544,7 @@
 
 .perseus-formats-tooltip {
     background: #fff;
-    color: #777;
+    color: #717378;
     padding: 5px 10px;
     width: 240px;
 }
--- a/packages/perseus/src/styles/perseus-renderer.less
+++ b/packages/perseus/src/styles/perseus-renderer.less
@@ -544,7 +544,7 @@
 
 .perseus-formats-tooltip {
     background: #fff;
-    color: #717378;
+    color: @textSecondary;
     padding: 5px 10px;
     width: 240px;
 }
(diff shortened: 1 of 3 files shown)


packages/perseus/src/styles/perseus-renderer.less
```diff
@@ -544,7 +544,7 @@
 
 .perseus-formats-tooltip {
     background: #fff;
-    color: #777;
+    color: #717378;
```
jeremywiebe, line 547: Is this a WB color?  Even though we're in a .less file, could you use (or define and use) a variable from `variables.less`?

jeremywiebe (approved):

packages/perseus/src/styles/perseus-renderer.less
mark-fitzgerald: Yes it is. I've moved that color to `variables.less`.

nishasy:
I trust this works perfectly, but for sake of PR completion and posterity, could you put a before and after screenshot here @mark-fitzgerald?

mark-fitzgerald:
> I trust this works perfectly, but for sake of PR completion and posterity, could you put a before and after screenshot here @mark-fitzgerald?

Added.

nishasy (approved):
Thank you! 🎉

PR #1968, opened 2024-12-07
Switch from Gerald to use native CODEOWNERS file

jeremywiebe:
## Summary:

I would like to start using Dependabot to help keep some deps up to date. We already have Dependabot configured to update two Khan Academy lint packages, but currently the Gerald check fails because it needs access to an org-wide secret (which it doesn't have access to). The good news is that this repo doesn't really use Gerald very much (only two rules). These rules are easily re-created with the [`CODEOWNERS`](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file) file instead. 

Issue: "none"

## Test plan:

We'll need to land this PR and watch the next few PRs and see if it tags the team correctly. The failure mode is just that we won't be tagged as a team.


SonicScrewdriver (approved):
Lovely, thank you!

PR #1973, opened 2024-12-10
Remove `scoreInput`

handeyeco:
## Summary:
Remove `scoreInput` from ServerItemRenderer. This was marked as deprecated and uses of it in Webapp have been removed in favor of `scorePerseusItem`.

## Test plan:
This was an external-facing API and usage was already removed. At this point it's just dead code, so I'm not sure what would need to be tested.


packages/perseus/src/__tests__/server-item-renderer.test.tsx
```diff
@@ -118,31 +118,6 @@ describe("server item renderer", () => {
         expect(screen.getByRole("textbox")).toBeVisible();
     });
 
-    it("should be invalid if no input provided", async () => {
```
handeyeco, line 121: I don't like deleting tests, but these seem to be specifically focused on testing SIR's ability to score.

packages/perseus/src/widgets/expression/score-expression.ts
```diff
@@ -165,9 +165,6 @@ function scoreExpression(
     // We matched a graded answer form, so we can now tell the user
     // whether their input was correct or incorrect, and hand out
     // points accordingly
-    // TODO(eater): Seems silly to translate result to this
```
handeyeco, line 168: idk, I think we're past the point of doing anything about this since this is from the Eater era.

jeremywiebe (approved):

PR #1975, opened 2024-12-10
Document perseus-types.ts file

jeremywiebe:
## Summary:

The `perseus-types.ts` file is special in the Perseus project. It represents the data that is saved out of editors and consumed by the Perseus rendering system. As a result, we need to be extremely careful when making changes to it. I've added a JSDoc comment to the top of the file that (hopefully) explains the file and some cautions to take into account when making changes. 
 
Issue: LEMS-2561

## Test plan:

Read the comments and let me know if they make sense and are clear. I'm happy to make further edits. :)


packages/perseus/src/perseus-types.ts
```diff
@@ -1,14 +1,40 @@
...
 
-// Range is replaced within this file with Interval, but it is used elsewhere
-// and exported from the package, so we need to keep it around.
+export type Coord = [x: number, y: number];
+export type Interval = [min: number, max: number];
+export type Vector2 = Coord; // Same name as Mafs
 export type Range = Interval;
-export type Size = [number, number];
-export type CollinearTuple = [vec.Vector2, vec.Vector2];
+export type Size = [width: number, height: number];
```
jeremywiebe, line 36: These changes are to "internalize" all types into this file instead of importing them from "outside". This means that if we upgrade Mafs and a type changes, we won't be implicitly breaking the schema. We'd still have type mismatches here _in_ Perseus, but we'd then be able to rationalize the differences between our schema (this file) and the changed types.

packages/perseus/src/perseus-types.ts
```diff
@@ -1,14 +1,40 @@
...
+ *   a) the changes are backwards compatible. If they are not, old data from
+ *      previous versions of the "schema" could become unrenderable, or worse,
+ *      introduce hard-to-diagnose bugs.
+ *   b) the parsing code (`util/parse-perseus-json/`) is updated to handle
+ *      the new format _as well as_ the old format.
+ */
 
-import type {Coord} from "./interactive2/types";
-import type {Interval, vec} from "mafs";
+// TODO(FEI-4010): Remove `Perseus` prefix for all types here
```
nishasy, line 30: Do we even still want to remove the `Perseus` prefix for all the types here?

nishasy (commented):
LGTM, but I'll let other people weigh in too

packages/perseus/src/perseus-types.ts
jeremywiebe: I believe we still could, yes.  Today, some of these type names get quite wordy. 

But, I don't think we need to deal with that now.

packages/perseus/src/perseus-types.ts
```diff
@@ -1,14 +1,40 @@
-// TODO(FEI-4010): Remove `Perseus` prefix for all types here
-// TODO(FEI-4011): Use types generated by https://github.com/jaredly/generate-perseus-flowtypes
+/**
+ * The Perseus "data schema" file.
+ *
+ * This file, and the types in it, represents the "data schema" that Perseus
+ * uses. The @khanacademy/perseus-editor package edits and produces objects
+ * that conform to the types in this file. Similarly, the top-level renderers
+ * in @khanacademy/perseus, consume objects that conform to these types.
```
handeyeco, line 7: My only feedback is that this seems to confident. It _does_ represent all permutations of the data or it _should_?

But if you feel confident then just 🚢

handeyeco (approved):

packages/perseus/src/perseus-types.ts
jeremywiebe: It's aspirational. :)  The new parsing code validates against these types and so I think it's a fair statement.

PR #1978, opened 2024-12-11
[Polygon] Close polygon if last point is moved to overlap the first point

nishasy:
## Summary:
todo

Issue: https://khanacademy.atlassian.net/browse/LEMS-2711

## Test plan:
todo

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
@@ -493,6 +493,23 @@ function doMovePoint(
                 return state;
             }
 
+            // If the last point is being moved to the same position as
+            // the first point, close the polygon.
+            if (
+                !state.closedPolygon && // not already close
+                action.index === newCoords.length - 1 && // last point
+                state.coords[0][0] === newValue[0] && // same x as first point
+                state.coords[0][1] === newValue[1] // same y as first point
+            ) {
+                // Remove the last point so it isn't a duplicate of the
+                // first point int he polygon representation. This is
+                // consistent with the cases in which the "Close shape"
+                // button is clicked or the first point is clicked to
+                // close the polygon.
+                newCoords.pop();
+                state.closedPolygon = true;
+            }
+
             return {
                 ...state,
                 hasBeenInteractedWith: true,
--- a/packages/perseus-editor/src/__stories__/flags-for-api-options.ts
+++ b/packages/perseus-editor/src/__stories__/flags-for-api-options.ts
@@ -13,6 +13,8 @@ export const flags = {
         ray: true,
         point: true,
         none: true,
+        "unlimited-point": true,
+        "unlimited-polygon": true,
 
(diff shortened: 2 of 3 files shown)


packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -200,8 +201,15 @@ function doClickPoint(
 
 function doClosePolygon(state: InteractiveGraphState): InteractiveGraphState {
     if (isUnlimitedGraphState(state) && state.type === "polygon") {
+        // We want to remove any duplicate points when closing the polygon to
```
catandthemachines, line 204: Love these comments!

packages/perseus/src/widgets/interactive-graphs/graphs/utils.ts
```diff
@@ -44,3 +45,20 @@ export const getIntersectionOfRayWithBox = (
 function isBetween(x: number, low: number, high: number) {
     return x >= low && x <= high;
 }
+
+export function getArrayWithoutDuplicates(array: Array<Coord>): Array<Coord> {
+    const returnArray: Array<Coord> = [];
```
catandthemachines, line 50: Might I suggest using the spread operator or the filter method to remove the duplicates. I think they have more efficiencies built in that will make this run with higher performance.

https://medium.com/@collettemathieu/what-is-the-fastest-way-to-remove-duplicates-from-an-array-in-javascript-9e5b4d3f55e1

catandthemachines: Additional caviat, our arrays will likely never be more than 20 points long if that. So we don't necessarily need to worry about it. But it would be nicer to have this code a little shorter and I think this article can help with it.

catandthemachines (approved):
Approved! Tested it out and it looks so sharp and really solves a lot of additional edge cases that make the polygon really great!

Added a comment about a way to condense the duplication function and make it more efficient but we should be good to go!

packages/perseus/src/widgets/interactive-graphs/graphs/utils.ts
nishasy: I was going to just create a set similar to that, but my understanding is that that doesn't work for array whose elements are also arrays or objects, because it doesn't consider two separate arrays equivalent even if their elements are the same.

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -493,6 +493,23 @@ function doMovePoint(
                 return state;
             }
 
+            // If the last point is being moved to the same position as
+            // the first point, close the polygon.
+            if (
```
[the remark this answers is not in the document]
nishasy: I think I ended up removing this code because this function should only be marking the polygon as closed? I noticed in the tests that the polygon's state changed to close even if it had an empty coords array, so I removed these lines to stay consistent with that behavior.

PR #1983, opened 2024-12-11
New approach to svg-image

SonicScrewdriver:
## Summary:
This PR updates our approach to loading graphie images for svg-image. 

An error will be thrown if the labels are not found, but should not impede users. Many tests were added using the failed cases provided on the ticket, in order to ensure that this new approach works for nearly all use cases. 

Issue: LEMS-2524

## Test plan:
- New svg-image tests 
- New graphie-utils tests that include (most) of the test cases provided.

### DIFF
--- a/packages/perseus/src/components/__tests__/svg-image.test.tsx
+++ b/packages/perseus/src/components/__tests__/svg-image.test.tsx
@@ -1,14 +1,56 @@
-import {render, screen} from "@testing-library/react";
+import {act, render} from "@testing-library/react";
 import * as React from "react";
 
+import {testDependencies} from "../../../../../testing/test-dependencies";
+import * as Dependencies from "../../dependencies";
+import {typicalCase} from "../../util/graphie-utils.testdata";
 import SvgImage from "../svg-image";
 
 describe("SvgImage", () => {
+    const images: Array<Record<any, any>> = [];
+    let originalImage;
+
+    beforeEach(() => {
+        jest.spyOn(Dependencies, "getDependencies").mockReturnValue(
+            testDependencies,
+        );
+        originalImage = window.Image;
+        // Mock HTML Image so we can trigger onLoad callbacks and see full
+        // image rendering.
+        // @ts-expect-error - TS2322 - Type 'Mock<Record<string, any>, [], any>' is not assignable to type 'new (width?: number | undefined, height?: number | undefined) => HTMLImageElement'.
+        window.Image = jest.fn(() => {
+            const img: Record<string, any> = {};
+            images.push(img);
+            return img;
+        });
+    });
+
+    afterEach(() => {
+        window.Image = originalImage;
+    });
+
+    // Tells the image loader 1, or all, of our images loaded
+    const markImagesAsLoaded = (imageIndex?: number) => {
+        if (imageIndex != null) {
+            const img = images[imageIndex];
+            if (img?.onload) {
+                act(() => img.onload());
+            }
+        } else {
+            images.forEach((i) => {
+                if (i?.onload) {
+                    act(() => i.onload());
+                }
+            });
+        }
+    };
     it("should render a spinner initially", () => {
+        // Act
         const {container} = render(
             <SvgImage src="http://localhost/sample.png" alt="png image" />,
         );
 
+        // Assert
         expect(
             // eslint-disable-next-line testing-library/no-container, testing-library/no-node-access
             container.querySelector("div[class*=spinnerContainer]"),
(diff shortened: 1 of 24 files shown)


packages/perseus/src/util/graphie-utils.ts
```diff
@@ -0,0 +1,253 @@
...
+
+// For offline exercises in the mobile app, we download the graphie data
+// (svgs and localized data files) and serve them from the local file
+// system (with file://). We replace urls that start with `web+graphie`
+// in the perseus json with this `file+graphie` prefix to indicate that
+// they should have the `file://` protocol instead of `https://`.
+const svgLocalLabelsRegex = /^file\+graphie:/;
+const hashRegex = /\/([^/]+)$/;
+
+/**
```
SonicScrewdriver, line 18: These types needed to come along to support the parsing. I'm exporting them for Graphie2000, but I do wonder which side should be in charge of these.

packages/perseus/src/util/graphie-utils.test.ts
```diff
@@ -0,0 +1,43 @@
...
+        // Act
+        parseDataFromJSONP(typicalCase.hash, "invalid jsonp", errorCallback);
+
+        // Assert
+        expect(errorCallback).toHaveBeenCalled();
+    });
+
+    // also test the array of edge cases from the testdata file
+    edgeCases.forEach((edgeCase, i) => {
+        it(`should parse the edge case ${i}`, () => {
```
SonicScrewdriver, line 31: There's a lot of edge cases, and some of them are quite similar. I could reduce them down, potentially, but I did like that we could cover 99% of the cases outlined in the ticket as failing to parse. Happy to take feedback on this one!

packages/perseus/src/util/graphie-utils.ts
```diff
@@ -0,0 +1,253 @@
...
+            retrieveData(Util.getDataUrl(url), (x, status, error) => {
+                Log.error("Data load failed for svg-image", Errors.Service, {
+                    cause: error,
+                    loggedMetadata: {
+                        dataUrl: Util.getDataUrl(url),
+                        status,
+                    },
+                });
+            });
+        }
```
mark-fitzgerald, line 251: I understand that this is simply copy-pasted from the `svg-image.tsx` file, but I think a bit of simplification might be helpful here. Perhaps the `shouldUseLocalizedData()` conditional block could be written as follows:
```
const dataErrorHandler = (x, status, error) => {
    Log.error("Data load failed for svg-image", Errors.Service, {
        cause: error,
        loggedMetadata: {
            dataUrl: Util.getDataUrl(url),
            status,
        },
});

if (shouldUseLocalizedData()) {
    retrieveData(getLocalizedDataUrl(url), (x, status, error) => {
        cacheData.localized = false;

        // If there is isn't any localized data, fall back to
        // the original, unlocalized data
        retrieveData(Util.getDataUrl(url), dataErrorHandler)
} else {
    retrieveData(Util.getDataUrl(url), dataErrorHandler)
}
```

packages/perseus/src/util/graphie-utils.ts
jeremywiebe: I think it makes sense for Perseus to own them. They are a part of how `SvgImage` works and consumes graphie images.

packages/perseus/src/util/graphie-utils.test.ts
jeremywiebe: I think the collection of edge cases you have is probably good. The ones that will fail (like that one with a `u''` prefix on all strings, is likely broken in prod too so I'd guess it would need to be re-written to work anyways.

packages/perseus/src/components/__tests__/svg-image.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+        const {container} = render(
+            <SvgImage src={typicalCase.url} alt="svg image" />,
+        );
+
+        markImagesAsLoaded(); // Tell the ImageLoader that our images are loaded
+
+        // Assert
+        expect(container).toMatchSnapshot();
+    });
+});
```
jeremywiebe, line 92: This is a great set. I do think it'd be good to add one or two tests to cover the path of loading a localized Graphie.  There's branching in there where it tries to load the localized version and falls back to `en` if it fails. I think if we add a test for localized loading and the fallback, that'd be awesome.

jeremywiebe (approved):
This is great. I had one idea for a couple of more tests to cover localized graphies, but otherwise this is looking great!  Thank-you!

mark-fitzgerald (approved):
Thank you for tightening all this up!

PR #1984, opened 2024-12-11
[FIX] React hooks causing unlimited polygon to crash in editor.

catandthemachines:
## Summary:
When toggling to unlimited polygon option in the interactive graph editor, the experience would crash. This is because the unlimited polygon code was utilizing React hooks improperly by only using some hooks under certain conditions. This is not allowed within React applications as it relies on the order of hook calls not to change.

To solve this problem we needed to move the hooks up one level of the component hierarchy and pass down the state to the child elements (LimitedPolygon and UnlimitedPolygon)

To ensure all our unlimited experiences follow this format, I've also applied these changes to Point Graph.

Issue: LEMS-2686

## Test plan:
Go to "http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-polygon"
Scroll down and change the number for sides to `unlimited`
Notice that the page does not crash nor are there any errors in the console.


packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx
```diff
@@ -68,7 +87,63 @@ const LimitedPolygonGraph = (props: Props) => {
...
     });
 
-    const lastMoveTime = React.useRef<number>(0);
+    // Hover/focus state for Limited Polgyon effects.
+    const [hovered, setHovered] = React.useState(false);
+    // This is more so required for the re-rendering that occurs when state
+    // updates; specifically with regard to line weighting and polygon focus.
+    const [focusVisible, setFocusVisible] = React.useState(false);
+
+    // This useEffect is to handle the focus snapping for Unlimited Polygon.
```
nishasy, line 96: Thank you for this comment!

packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx
```diff
@@ -31,35 +33,52 @@ export function renderPolygonGraph(
...
+const PolygonGraph = (props: Props) => {
     const {dispatch} = props;
+    const {numSides, coords, snapStep, snapTo = "grid"} = props.graphState;
+    const graphConfig = useGraphConfig();
+
+    // Ref to implement the dragging behavior on a Limited/Closed Polygon.
+    const polygonRef = React.useRef<SVGPolygonElement>(null);
+    // Ref to manage dynamic focus for the points in Unlimited Polygon
+    const pointsRef = React.useRef<Array<SVGElement | null>>([]);
+    // Ref to manage the last move tim for a Limited Polygon.
```
nishasy, line 61:
```suggestion
    // Ref to manage the last move time for a Limited Polygon.
```

packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx
```diff
@@ -31,35 +33,52 @@ export function renderPolygonGraph(
...
+    const lastMoveTimeRef = React.useRef<number>(0);
```
nishasy, line 62: Interesting usage! I don't think I've seen a ref used for anything other than an HTML element before, but I see what you're doing here.

nishasy (approved):
Nice sleuthing!

packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx
catandthemachines: Me neither 🤣

nishasy: Alternatively, I suppose you could make this a state value, and you can pass in the value and its setter into the limited and unlimited polygon graphs. But that seems more convoluted.

PR #1988, opened 2024-12-11
One weird trick that delights screen readers

SonicScrewdriver:
## Summary:
This is a simple PR that adds an aria-hidden tag to our graphie container so that we don't read out axis labels or "Made by Raphael" on screen readers. As a result, screen readers should have an easier time navigating the website. 

While this may hide some labels within graphie, we've taken the position that the alt tags should be the sole source of truth for our screen readers. 

Issue: LEMS-2712

## Test plan:
- manual testing

### DIFF
--- a/.changeset/cool-moons-reply.md
+++ b/.changeset/cool-moons-reply.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": minor
+---
+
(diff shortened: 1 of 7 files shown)


mark-fitzgerald (approved):
Simple. Always a winner.

PR #1995, opened 2024-12-12
Swap out deprecated input-number with numeric-input in some tests

jeremywiebe:
# Subject 

The `input-number` widget is deprecated and was causing issues in the work I'm doing for validation and SSS. Swapping them out for the `numeric-input` so that I have a path forward. 

# Test Plan 

`yarn test`
`yarn typecheck`

Issue: LEMS-2561

### DIFF
--- a/.changeset/spicy-cups-join.md
+++ b/.changeset/spicy-cups-join.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+TESTS: swap input-number out of renderer tests as it is deprecated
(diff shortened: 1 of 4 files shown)


Myranae (commented):
Looks like these changes are just in test related content, right? So it shouldn't touch the user experience at all? Does this conflict with the efforts to automatically swap these two widgets? My guess is no, but wanted to see if that needed to be considered at all.

handeyeco (approved):

jeremywiebe:
> Looks like these changes are just in test related content, right? So it shouldn't touch the user experience at all? 

Correct! They are strictly tests and getting rid of a widget that is deprecated and scheduled for removal. 

> Does this conflict with the efforts to automatically swap these two widgets? My guess is no, but wanted to see if that needed to be considered at all.

No, that initiative is currently paused as we look at a backfill instead.

PR #1996, opened 2024-12-12
Remove unused rubric type for iFrame

Myranae:
## Summary:
The initial goal was to separate out answers from the userInput, but iFrame is a unique widget in that the answers aren't actually contained in what is sent to the scoring function. Scoring happens via the iFrame, and the scoring function is told if the user is correct, incorrect, and incomplete. It may be useful to change what this parameter is called as it is not actually the user's input either, but a summary of the results of checking the user's input. Additionally, validation happens after both types of scoring, so it most likely will not be possible to have a validation function for this widget either.

Issue: LEMS-2440

## Test plan:
- Confirm all checks pass
- Confirm widget still works as expected via a ZND (iframe tends not to be testable locally)

### DIFF
--- a/.changeset/proud-ghosts-learn.md
+++ b/.changeset/proud-ghosts-learn.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Remove unused iframe rubric type
--- a/packages/perseus/src/validation.types.ts
+++ b/packages/perseus/src/validation.types.ts
@@ -102,9 +102,6 @@ export type PerseusGrapherRubric = {
 
 export type PerseusGrapherUserInput = PerseusGrapherRubric["correct"];
 
-// TODO(LEMS-2440): Can possibly be removed during 2440; only userInput used
-export type PerseusIFrameRubric = Empty;
-
 export type PerseusIFrameUserInput = {
     status: UserInputStatus;
     message: string | null;
@@ -244,7 +241,6 @@ export type Rubric =
     | PerseusGradedGroupRubric
     | PerseusGradedGroupSetRubric
     | PerseusGrapherRubric
-    | PerseusIFrameRubric
     | PerseusInputNumberRubric
     | PerseusInteractiveGraphRubric
     | PerseusLabelImageRubric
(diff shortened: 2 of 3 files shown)


packages/perseus/src/validation.types.ts
```diff
@@ -102,9 +102,6 @@ export type PerseusGrapherRubric = {
 
 export type PerseusGrapherUserInput = PerseusGrapherRubric["correct"];
 
-// TODO(LEMS-2440): Can possibly be removed during 2440; only userInput used
-export type PerseusIFrameRubric = Empty;
```
Myranae, line 106: Could removing this now cause problems down the road? I think it's pretty easy to add it back in, but not sure if we need a rubric and userInput for every widget. I'm guessing it's okay, but wanted to double check.

.changeset/proud-ghosts-learn.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
```
Myranae, line 2: This seems to just be removing an unused type, so I put it as patch, but it does touch the UserInput type and the Props type, so wasn't sure if that meant it should be minor. Leaning towards `patch`, but wanted to confirm :)

packages/perseus/src/validation.types.ts
jeremywiebe: I agree to remove it now.  We already have other widgets where there is no Rubric type.

.changeset/proud-ghosts-learn.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Remove unused iframe rubric type
```
jeremywiebe, line 5: I don't see any changes to our main package file (`index.ts`) which means that this type was not exported from Perseus, so a `patch` release sounds right to me.

jeremywiebe (approved):
> Would it be useful to change what the scoreIframe parameter is called as it is not actually the user's input, but a summary of the results of checking the user's input? 

No, please don't change the scoring function parameter names. They should all conform to the `WidgetScorerFunction` type: 

```typescript
export type WidgetScorerFunction = (
    // The user data needed to score
    userInput: UserInput,
    // The scoring criteria to score against
    rubric: Rubric,
    // Strings, for error messages in invalid widgets
    string?: PerseusStrings,
    // Locale, for math evaluation
    // (1,000.00 === 1.000,00 in some countries)
    locale?: string,
) => PerseusScore;
```

We can omit parameters from the end of the list (all the up to having a scoring function that takes no parameters). Also, TypeScript doesn't care about the parameter names, but for consistency let's keep the argument list and argument names the same as in the `WidgetScorerFunction`.  :)

PR #1997, opened 2024-12-12
Remove unused rubric type for CS Program

Myranae:
## Summary:
This removes an unused rubric type. The iframe tells the widget if the user's input is correct or not, so a rubric is not needed.
Additionally, validation happens after scoring, so it won't be possible to create a validation function for this widget at this time either.

Issue: LEMS-2440

## Test plan:
- Confirm all checks pass
- Confirm widget still works as expected via a ZND (CS Program works via iFrame, which tens not to be testable locally)

### DIFF
--- a/.changeset/fifty-laws-hear.md
+++ b/.changeset/fifty-laws-hear.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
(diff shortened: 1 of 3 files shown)


jeremywiebe (approved):
That makes sense. We'll just trust the frontend in this case for correctness.

PR #2000, opened 2024-12-13
SSS: Hook emptyWidgets() up to validation functions

jeremywiebe:
## Summary:

This PR begins connecting the validation functions that have been built to the existing `emptyWidgets()` function that already exists on the `Renderer` (as well as `emptyWidgesFunctional()`). This function powers the `APIOptions.answerableCallback` which is what the exercise chrome using Perseus uses to detect if a question is at a point where it can be scored. 

Note that some widget's "emptiness" checks depend on scoring data and so those checks are not included in the empty checking now. 
 

Issue: LEMS-2561

## Test plan:

yarn test
yarn typecheck

### DIFF
--- a/.changeset/quiet-adults-look.md
+++ b/.changeset/quiet-adults-look.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": minor
+---
+
+Change empty widgets check in Renderer to depend only on data available (and not on scoring data)
--- a/packages/perseus/src/__tests__/renderer.test.tsx
+++ b/packages/perseus/src/__tests__/renderer.test.tsx
@@ -1,6 +1,6 @@
 import {describe, beforeAll, beforeEach, it} from "@jest/globals";
 import {Errors} from "@khanacademy/perseus-core";
-import {act, logRoles, screen, waitFor, within} from "@testing-library/react";
+import {act, screen, waitFor, within} from "@testing-library/react";
 import {userEvent as userEventLib} from "@testing-library/user-event";
 import * as React from "react";
 
--- a/packages/perseus/src/__tests__/renderer.test.tsx
+++ b/packages/perseus/src/__tests__/renderer.test.tsx
@@ -1,6 +1,6 @@
 import {describe, beforeAll, beforeEach, it} from "@jest/globals";
 import {Errors} from "@khanacademy/perseus-core";
-import {act, screen, waitFor, within} from "@testing-library/react";
+import {act, logRoles, screen, waitFor, within} from "@testing-library/react";
 import {userEvent as userEventLib} from "@testing-library/user-event";
 import * as React from "react";
 
@@ -15,6 +15,7 @@ import {
     definitionItem,
     mockedRandomItem,
     mockedShuffledRadioProps,
+    question3,
 } from "../__testdata__/renderer.testdata";
 import * as Dependencies from "../dependencies";
 import {registerWidget} from "../widgets";
@@ -1605,35 +1606,36 @@ describe("renderer", () => {
         it("should return all empty widgets", async () => {
             // Arrange
             const {renderer} = renderQuestion({
-                ...question2,
+                ...question3,
                 content:
-                    "Input 1: [[☃ numeric-input 1]]\n\n" +
-                    "Input 2: [[☃ numeric-input 2]]",
+                    "Input 1: [[☃ expression 1]]\n\n" +
+                    "Input 2: [[☃ expression 2]]",
                 widgets: {
-                    ...question2.widgets,
-                    "numeric-input 2": question2.widgets["numeric-input 1"],
+                    ...question3.widgets,
+                    "expression 2": question3.widgets["expression 1"],
                 },
             });
             await userEvent.type(screen.getAllByRole("textbox")[0], "150");
+            act(() => jest.runOnlyPendingTimers());
 
             // Act
             const emptyWidgets = renderer.emptyWidgets();
 
             // Assert
-            expect(emptyWidgets).toStrictEqual(["numeric-input 2"]);
+            expect(emptyWidgets).toStrictEqual(["expression 2"]);
         });
 
         it("should not return static widgets even if empty", () => {
             // Arrange
             const {renderer} = renderQuestion({
-                ...question2,
+                ...question3,
                 content:
-                    "Input 1: [[☃ numeric-input 1]]\n\n" +
-                    "Input 2: [[☃ numeric-input 2]]",
+                    "Input 1: [[☃ expression 1]]\n\n" +
+                    "Input 2: [[☃ expression 2]]",
                 widgets: {
-                    ...question2.widgets,
-                    "numeric-input 2": {
-                        ...question2.widgets["numeric-input 1"],
+                    ...question3.widgets,
+                    "expression 2": {
+                        ...question3.widgets["expression 1"],
                         static: true,
                     },
                 },
@@ -1643,7 +1645,7 @@ describe("renderer", () => {
             const emptyWidgets = renderer.emptyWidgets();
 
             // Assert
-            expect(emptyWidgets).toStrictEqual(["numeric-input 1"]);
+            expect(emptyWidgets).toStrictEqual(["expression 1"]);
         });
 
         it("should return widget ID for group with empty widget", () => {
@@ -1663,7 +1665,7 @@ describe("renderer", () => {
                 JSON.stringify(simpleGroupQuestion),
             );
             simpleGroupQuestionCopy.widgets["group 1"].options.widgets[
-                "numeric-input 1"
+                "expression 1"
             ].static = true;
             const {renderer} = renderQuestion(simpleGroupQuestionCopy);
 
@@ -1678,6 +1680,7 @@ describe("renderer", () => {
             // Arrange
             const {renderer} = renderQuestion(simpleGroupQuestion);
             await userEvent.type(screen.getByRole("textbox"), "99");
+            act(() => jest.runOnlyPendingTimers());
 
             // Act
             const emptyWidgets = renderer.emptyWidgets();
--- a/packages/perseus/src/renderer-util.ts
+++ b/packages/perseus/src/renderer-util.ts
@@ -1,12 +1,20 @@
 import {mapObject} from "./interactive2/objective_";
 import {scoreIsEmpty, flattenScores} from "./util/scoring";
 import {getWidgetIdsFromContent} from "./widget-type-utils";
-import {getWidgetScorer, upgradeWidgetInfoToLatestVersion} from "./widgets";
+import {
+    getWidgetScorer,
+    getWidgetValidator,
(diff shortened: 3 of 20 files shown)


packages/perseus/src/__tests__/renderer.test.tsx
```diff
@@ -15,6 +15,7 @@ import {
     definitionItem,
     mockedRandomItem,
     mockedShuffledRadioProps,
+    question3,
```
jeremywiebe, line 18: The changes in this file were needed because: 

  * the `input-number` widget is deprecated 
  * the `numeric-input` widget does not have any validation it can do that does not require scoring data, so it can't be used for empty widget tests. 
  *

packages/perseus/src/renderer-util.ts
```diff
@@ -33,31 +41,31 @@ export function getUpgradedWidgetOptions(
...
     strings: PerseusStrings,
     locale: string,
 ): ReadonlyArray<string> {
-    const upgradedWidgets = getUpgradedWidgetOptions(widgets);
-
     return widgetIds.filter((id) => {
-        const widget = upgradedWidgets[id];
-        if (!widget || widget.static) {
+        const widget = widgets[id];
+        if (!widget || widget.static === true) {
```
jeremywiebe, line 60: This use of `static` here means that both our frontend data and scoring data will need to have the `static` field kept. Just observing this. I'm working through this in an upcoming type PR

packages/perseus/src/types.ts
```diff
@@ -578,6 +578,13 @@ export type WidgetTransform = (
 
 export type ValidationResult = Extract<PerseusScore, {type: "invalid"}> | null;
 
+export type WidgetValidatorFunction = (
+    userInput: any, // STOPSHIP
```
jeremywiebe, line 582: Future PR

packages/perseus/src/validation.types.ts
```diff
@@ -286,3 +287,60 @@ export type UserInputMap = {[widgetId: string]: UserInput | UserInputMap};
 export type UserInputArray = ReadonlyArray<
     UserInputArray | UserInput | null | undefined
 >;
+export interface ValidationDataTypes {
```
jeremywiebe, line 290: This uses the same construct as the new `PerseusWidgetTypes` does. However, the fact that not all widgets have validation data concerns me because it means that this interface won't have an entry for all widgets. 🤔 I'm not sure how to tackle that yet.

packages/perseus/src/widgets/categorizer/categorizer.tsx
```diff
@@ -328,4 +329,7 @@ export default {
     // TODO(LEMS-2656): remove TS suppression
     // @ts-expect-error: Type 'UserInput' is not assignable to type 'PerseusCSProgramUserInput'.
     scorer: scoreCategorizer,
+    // TODO(LEMS-2656): remove TS suppression
+    // @ts-expect-error: Type 'UserInput' is not assignable to type 'PerseusCSProgramUserInput'.
```
jeremywiebe, line 333: This is the same issue as with the `scorer` entry. Each of our validation functions accepts the specific user input and validation data type for the widget, but our function definition for it on `WidgetExports` takes a type union of _any_ widget's user input and validation data type. 

The solution, I think, is to introduce more generic types on WidgetExports<T> but that gets out of hand quickly each time I try.  I'm going to leave this here for now and come back to it after this PR.

packages/perseus/src/widgets/group/validate-group.ts
```diff
@@ -0,0 +1,31 @@
...
+    PerseusGroupValidationData,
+} from "../../validation.types";
+
+function validateGroup(
+    userInput: PerseusGroupUserInput,
+    validationData: PerseusGroupValidationData,
+    strings: PerseusStrings,
+    locale: string,
+): ValidationResult {
+    const emptyWidgets = emptyWidgetsFunctional(
```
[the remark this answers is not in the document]
jeremywiebe: Also, the nice thing about how it's implemented is that we don't care what widgets inside are empty, just that _any_ are.

handeyeco, line 16: This makes me happy that we're able to reuse the functional helpers like this rather than having to access methods on components.

.changeset/quiet-adults-look.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/perseus": minor
```
handeyeco, line 2: Overall, I think this is right. Some of the TS is a little "big-brain" for me to understand and it looks like you might still be tidying up a little bit, but I think it's mostly good to go.

My concern/observation: this does change the _behavior_ of `emptyWidgets`. Before we only enabled the "Check" button after a deep validation check (including the validation that requires the answers). Now we'll enable the "Check" button after a shallow validation check (only checking if the answer is scorable). 

This is inevitable and baked into the project. It just occurs to me that we haven't really shared this with content and support advocates. I think it'll hardly be noticeable, but what do I know.

packages/perseus/src/widgets/group/group.testdata.ts
```diff
@@ -159,32 +159,24 @@ export const simpleGroupQuestion: PerseusRenderer = {
         "group 1": {
             graded: true,
             options: {
-                content: "[[☃ numeric-input 1]]",
+                content: "[[☃ expression 1]]",
```
handeyeco, line 162: It's weird to me that test data for `renderer.test.ts` is in `group.test.ts`. Not your fault, maybe even likely that I did that. It's just weird.

packages/perseus/src/widgets/categorizer/categorizer.tsx
handeyeco: Might be a fun "wind down" task next week. Might not. idk, I'm fine with error suppression for now.

packages/perseus/src/validation.types.ts
```diff
@@ -278,11 +280,73 @@ export type UserInput =
...
+    // grapher: PerseusGrapherValidationData;
+    // "graded-group-set": PerseusGradedGroupSetValidationData;
+    // "graded-group": PerseusGradedGroupValidationData;
+    group: PerseusGroupValidationData;
+    // iframe: PerseusIFrameValidationData;
+    // image: PerseusImageValidationData;
+    // "input-number": PerseusInputNumberValidationData;
+    // interaction: PerseusInteractionValidationData;
+    // "interactive-graph": PerseusInteractiveGraphValidationData;
+    // "label-image": PerseusLabelImageValidationData;
```
handeyeco, line 307: I'm guessing you're going to do something with this commented code right?

handeyeco (approved):

packages/perseus/src/validation.types.ts
Myranae: What are the potential issues with the interface not having an entry for all widgets?

packages/perseus/src/validation.types.ts
```diff
@@ -278,11 +280,73 @@ export type UserInput =
     | PerseusSorterUserInput
     | PerseusTableUserInput;
 
-export type UserInputMap = {[widgetId: string]: UserInput | UserInputMap};
```
Myranae, line 281: Wasn't this here because some of the utility functions were recursive and might have nesting going on? Is that not the case anymore? May not have understood how it was used originally, so just checking :)

packages/perseus/src/renderer-util.ts
```diff
@@ -33,31 +41,31 @@ export function getUpgradedWidgetOptions(
...
-    widgets: PerseusWidgetsMap,
+    widgets: ValidationDataMap,
     // This is a port of old code, I'm not sure why
     // we need widgetIds vs the keys of the widgets object
     widgetIds: Array<string>,
     userInputMap: UserInputMap,
     strings: PerseusStrings,
     locale: string,
 ): ReadonlyArray<string> {
-    const upgradedWidgets = getUpgradedWidgetOptions(widgets);
```
Myranae, line 45: Question: What makes this line no longer necessary? I looked through the PR, but not sure what's related

Myranae (commented):
Excited to have this set up :) Just added some comments and questions.

packages/perseus/src/validation.types.ts
jeremywiebe: Today I'm not totally sure yet. At least, it would give us an inaccurate view of what the different types of validation data we support (we might have a type defined, but not in this "master list"). Ideally in future PRs we can derive some extra checks to ensure these interface type registries stay accurate. 🤞

packages/perseus/src/widgets/group/group.testdata.ts
jeremywiebe: The interactive-graph project ended up creating a builder pattern for ig options. That'd be amazing to have for all Perseus widget options.  But instead I think we sometimes cheated and used renderer configs from the Renderer tests.

packages/perseus/src/validation.types.ts
jeremywiebe: Yes, just not sure what. 

https://github.com/Khan/perseus/pull/2000#discussion_r1883083926

packages/perseus/src/__tests__/renderer.test.tsx
```diff
@@ -1605,35 +1606,36 @@ describe("renderer", () => {
...
+                    "Input 2: [[☃ expression 2]]",
                 widgets: {
-                    ...question2.widgets,
-                    "numeric-input 2": question2.widgets["numeric-input 1"],
+                    ...question3.widgets,
+                    "expression 2": question3.widgets["expression 1"],
                 },
             });
             await userEvent.type(screen.getAllByRole("textbox")[0], "150");
+            act(() => jest.runOnlyPendingTimers());
```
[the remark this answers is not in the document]
jeremywiebe: This is because we are now using the `expression` widget. I couldn't figure out why the `userEvent.type()` call above wasn't working. In other places of Perseus where we test `expression` we also do this. I _think_ the reason is that it takes a render cycle for the typing to show up in the expression widget. `expression` is based on MathInput which uses MathQuill under the hood. Probably related to that.

packages/perseus/src/validation.types.ts
jeremywiebe: The only widget that did that was the `group`.  I added a proper type for the `group`'s user input (which is `export type PerseusGroupUserInput = UserInputMap;`). So it makes this type a bit more accurate and the code that uses the `UserInput` easier to work with. I have a future PR almost ready where this type will be further improved.

packages/perseus/src/renderer-util.ts
jeremywiebe: This is not needed because this empty widgets call can (should?) only ever be called in a situation where the widget options have already been upgraded.  When we are on the client, this data will be widget options from the Renderer which have already been upgraded. When we are scoring on the server, it will be scoring data that is already upgraded. Basically, I don't think this function should be exported outside of Perseus.

PR #2002, opened 2024-12-13
SSS: Improve types for validation

jeremywiebe:
## Summary:

This PR begins working out the types for Scoring, Validation, and how they relate to our full widget options types. There are currently three "trees" of types that are shaped as a map of `widgetId` to something. They are: 

  * Full widget options (starting from `PerseusRenderer`)
  * Scoring data - used to score the learner's guess (user input)
  * Validation data - a shared subset (of Render and Scoring data) used to do empty widget checking (aka validation). This helps the frontend to know if the question is scorable yet. 

Finally, there is also a widget map known as User Input. This map is a map of widget ids from the item to the user input the learner has entered so far. 

Issue: LEMS-2561

## Test plan:

`yarn typecheck` (especially, the new `validation.typetest.ts` file!)
`yarn test` (just to be sure)

### DIFF
--- a/packages/perseus/src/__tests__/validation.typetest.ts
+++ b/packages/perseus/src/__tests__/validation.typetest.ts
@@ -9,8 +9,15 @@
 import type {PerseusRenderer} from "../perseus-types";
 import type {ScoringDataMap, ValidationDataMap} from "../validation.types";
 
+/**
+ * An utility type that verifies that the given type `E` extends the type `T`.
+ * This is useful for asserting that one type remains a compatible subset of
+ * the other.
+ */
+type Extends<T, E extends T> = (T) => E;
+
 // We can use a 'widgets' map from a PerseusRenderer as a ValidationDataMap
-0 as any as PerseusRenderer["widgets"] satisfies ValidationDataMap;
+type _ = Extends<ValidationDataMap, PerseusRenderer["widgets"]>;
 
 // We can use a ScoringDataMap as a ValidationDataMap
-0 as any as ScoringDataMap satisfies ValidationDataMap;
+type __ = Extends<ValidationDataMap, ScoringDataMap>;
--- a/packages/perseus/src/__tests__/validation.typetest.ts
+++ b/packages/perseus/src/__tests__/validation.typetest.ts
@@ -2,9 +2,11 @@
  * This file contains TypeScript type "tests" which ensure that types needed
  * for scoring and validation stay in sync with other types in the system.
  *
- * If you make a change and one of these "satisfies" assertions fails, that
- * will usually mean you've made a change that will cause runtime breakages in
- * scoring or validation.
+ * If you make a change and `Extends<>` starts to complain, that will usually
+ * mean you've made a change that will cause runtime breakages in scoring or
+ * validation. ie. The types that should be compatible are no longer
+ * compatible. Read the TypeScript error message closely and it should point
+ * you in the right direction.
  */
 import type {PerseusRenderer} from "../perseus-types";
 import type {ScoringDataMap, ValidationDataMap} from "../validation.types";
--- a/packages/perseus/src/__tests__/validation.typetest.ts
+++ b/packages/perseus/src/__tests__/validation.typetest.ts
@@ -1,0 +1,16 @@
+/**
+ * This file contains TypeScript type "tests" which ensure that types needed
+ * for scoring and validation stay in sync with other types in the system.
+ *
+ * If you make a change and one of these "satisfies" assertions fails, that
+ * will usually mean you've made a change that will cause runtime breakages in
+ * scoring or validation.
+ */
+import type {PerseusRenderer} from "../perseus-types";
+import type {RubricMap, ValidationDataMap} from "../validation.types";
+
+// We can use a 'widgets' map from a PerseusRenderer as a ValidationDataMap
+0 as any as PerseusRenderer["widgets"] satisfies ValidationDataMap;
+
+// We can use a RubricMap as a ValidationDataMap
+0 as any as RubricMap satisfies ValidationDataMap;
--- a/packages/perseus/src/__tests__/validation.typetest.ts
+++ b/packages/perseus/src/__tests__/validation.typetest.ts
@@ -7,10 +7,10 @@
  * scoring or validation.
  */
 import type {PerseusRenderer} from "../perseus-types";
-import type {RubricMap, ValidationDataMap} from "../validation.types";
+import type {ScoringDataMap, ValidationDataMap} from "../validation.types";
 
 // We can use a 'widgets' map from a PerseusRenderer as a ValidationDataMap
 0 as any as PerseusRenderer["widgets"] satisfies ValidationDataMap;
 
-// We can use a RubricMap as a ValidationDataMap
-0 as any as RubricMap satisfies ValidationDataMap;
+// We can use a ScoringDataMap as a ValidationDataMap
+0 as any as ScoringDataMap satisfies ValidationDataMap;
--- a/packages/perseus/src/perseus-types.ts
+++ b/packages/perseus/src/perseus-types.ts
@@ -37,10 +37,46 @@ export type Size = [width: number, height: number];
 export type CollinearTuple = [Vector2, Vector2];
 export type ShowSolutions = "all" | "selected" | "none";
 
-// A utility type that constructs a widget map from a "registry interface".
-// The keys of the registry should be the widget type (aka, "categorizer" or
-// "radio", etc) and the value should be the option type stored in the value
-// of the map.
+/**
+ * A utility type that constructs a widget map from a "registry interface".
+ * The keys of the registry should be the widget type (aka, "categorizer" or
+ * "radio", etc) and the value should be the option type stored in the value
+ * of the map.
+ *
+ * You can think of this as a type that generates another type. We use
+ * "registry interfaces" as a way to keep a set of widget types to their data
+ * type in several places in Perseus. This type then allows us to generate a
+ * map type that maps a widget id to its data type and keep strong typing by
+ * widget id.
+ *
+ * For example, given a fictitious registry such as this:
+ *
+ * ```
+ * interface DummyRegistry {
+ *     categorizer: { categories: ReadonlyArray<string> };
+ *     dropdown: { choices: ReadonlyArray<string> }:
+ * }
+ * ```
+ *
+ * If we create a DummyMap using this helper:
+ *
+ * ```
+ * type DummyMap = MakeWidgetMap<DummyRegistry>;
+ * ```
+ *
+ * We'll get a map that looks like this:
+ *
+ * ```
+ * type DummyMap = {
+ *     `categorizer ${number}`: { categories: ReadonlyArray<string> };
+ *     `dropdown ${number}`: { choices: ReadonlyArray<string> };
+ * }
+ * ```
+ *
+ * We use interfaces for the registries so that they can be extended in cases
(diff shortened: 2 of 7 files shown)


packages/perseus/src/validation.types.ts
```diff
@@ -236,58 +238,105 @@ export type PerseusTableRubric = {
...
-    | PerseusMatrixRubric
-    | PerseusNumberLineScoringData
-    | PerseusNumericInputRubric
-    | PerseusOrdererRubric
-    | PerseusPlotterScoringData
-    | PerseusRadioRubric
-    | PerseusSorterRubric
-    | PerseusTableRubric;
-
-export type UserInput =
```
jeremywiebe, line 262: This is moved down below all of the Rubric/Scoring Data types together with the various widget UserInput types. #organizing

packages/perseus/src/validation.types.ts
```diff
@@ -236,58 +238,105 @@ export type PerseusTableRubric = {
...
-    | PerseusPlotterUserInput
-    | PerseusRadioUserInput
-    | PerseusSorterUserInput
-    | PerseusTableUserInput;
-
-export type UserInputMap = {[widgetId: string]: UserInput};
+export interface RubricRegistry {
+    categorizer: PerseusCategorizerScoringData;
+    "cs-program": PerseusCSProgramRubric;
+    // definition: PerseusDefinitionScoringData;
```
jeremywiebe, line 244: I really am not a fan of how this works right now. The fact that this is not a full listing of all widgets because we don't score all widgets worries me, but I don't think it makes sense to add a type for all widgets just to make this "whole". 

I'm tempted to leave it like this for now and as we work further hopefully something resolves to a better solution. 

Also open to ideas.

packages/perseus/src/__tests__/validation.typetest.ts
```diff
@@ -0,0 +1,16 @@
...
+ * scoring or validation.
+ */
+import type {PerseusRenderer} from "../perseus-types";
+import type {RubricMap, ValidationDataMap} from "../validation.types";
+
+// We can use a 'widgets' map from a PerseusRenderer as a ValidationDataMap
+0 as any as PerseusRenderer["widgets"] satisfies ValidationDataMap;
+
+// We can use a RubricMap as a ValidationDataMap
+0 as any as RubricMap satisfies ValidationDataMap;
```
Myranae, line 16: May want to update these to `ScoringMap` since we're removing the `rubric` terminology.

packages/perseus/src/perseus-types.ts
```diff
@@ -37,6 +37,14 @@ export type Size = [width: number, height: number];
 export type CollinearTuple = [Vector2, Vector2];
 export type ShowSolutions = "all" | "selected" | "none";
 
+// A utility type that constructs a widget map from a "registry interface".
+// The keys of the registry should be the widget type (aka, "categorizer" or
+// "radio", etc) and the value should be the option type stored in the value
+// of the map.
```
Myranae, line 43: Could you provide an example of the output from this function? Or an example usage?

packages/perseus/src/validation.types.ts
```diff
@@ -236,58 +238,105 @@ export type PerseusTableRubric = {
...
-    | PerseusNumberLineUserInput
-    | PerseusNumericInputUserInput
-    | PerseusOrdererUserInput
-    | PerseusPlotterUserInput
-    | PerseusRadioUserInput
-    | PerseusSorterUserInput
-    | PerseusTableUserInput;
-
-export type UserInputMap = {[widgetId: string]: UserInput};
+export interface RubricRegistry {
```
Myranae, line 241: Should this be ScoringDataRegistry? Doesn't have the same ring, but we're trying to move away from the `rubric` vocab, right? Also, should the Rubric types below be changed to ScoringData types or will we just handle that after landing my renaming PR when we get back? Might make sense to plan the order in which we land PRs, though it might make sense just to end the week with a fully merged feature branch with all our changes combined.

packages/perseus/src/validation.types.ts
Myranae: If not all widgets have ScoringData, why don't we just remove the inner comments and add a comment to the whole interface (just above line 241) saying that? We could keep a list of which widgets currently don't have scoring data in that comment so it doesn't muddy the interface so to speak. Or is there a reason why we _should_ have an entry for every widget in this interface?

packages/perseus/src/validation.types.ts
```diff
@@ -236,58 +238,105 @@ export type PerseusTableRubric = {
...
+/**
+ * A map of scoring data (previously referred to as "rubric"), keyed by
+ * `widgetId`. This data is used to score a learner's guess for a PerseusItem.
+ *
+ * NOTE:  The value in this map is intentionally a subset of WidgetOptions<T>.
+ * By using the same shape (minus any unneeded render data), we are able to
+ * share functionality that understands how to traverse maps of `widget id` to
+ * `options`.
+ */
+export type RubricMap = {
```
handeyeco, line 289: So overall the point of this PR seems to be aimed at more closely matching widget-specific data to the widget ID?

Doesn't this reinforce the thing you and Nicole were against - tightly binding the widget `type` and the widget `id`? Like if we were to decide tomorrow that we were going to give all widgets UUIDs (which we should as we've seen in the GradedGroup issue), wouldn't that break this code?

I'm having trouble aligning the description of the PR and the ticket with the actual changes. This is like pre/post work for LEMS-2561?

packages/perseus/src/validation.types.ts
```diff
@@ -236,58 +238,105 @@ export type PerseusTableRubric = {
...
+}
+
+// A union type of all the widget user input types
+export type UserInput = UserInputRegisry[keyof UserInputRegisry];
+
+/**
+ * A map of widget IDs to user input types (strongly typed based on the format
+ * of the widget ID).
+ */
+export type UserInputMap = MakeWidgetMap<UserInputRegisry>;
```
Myranae, line 331: Do you only use the asterisk comments in certain situations? Just wondering when it should be the `//` version or the `/**` version.

packages/perseus/src/validation.types.ts
```diff
@@ -236,58 +238,105 @@ export type PerseusTableRubric = {
...
+        options: RubricRegistry[Property];
+    };
+};
+
+export type Rubric = RubricRegistry[keyof RubricRegistry];
+
+// This is an interface so that it can be extended if a widget is created
+// outside of this Perseus package. See `PerseusWidgetTypes` for a full
+// explanation.
+interface UserInputRegisry {
```
handeyeco, line 302:
```suggestion
interface UserInputRegistry {
```

packages/perseus/src/__tests__/validation.typetest.ts
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * If you make a change and one of these "satisfies" assertions fails, that
+ * will usually mean you've made a change that will cause runtime breakages in
+ * scoring or validation.
+ */
+import type {PerseusRenderer} from "../perseus-types";
+import type {RubricMap, ValidationDataMap} from "../validation.types";
+
+// We can use a 'widgets' map from a PerseusRenderer as a ValidationDataMap
+0 as any as PerseusRenderer["widgets"] satisfies ValidationDataMap;
```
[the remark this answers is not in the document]
Myranae: ++

packages/perseus/src/perseus-types.ts
jeremywiebe: Yes. So I've started developing the idea of a type registry (because we have full widget options types, scoring data types, frontend rendering types). Each of these is map of widget type to the option type for that widget.

If you look at the usage in this PR, we use the `PerseusWidgetTypes` interface to create the `PerseusWidgetsMap` type. 

`PerseusWidgetTypes` looks like this (shortened a bit here): 

```typescript
export interface PerseusWidgetTypes {
    categorizer: CategorizerWidget;
    "cs-program": CSProgramWidget;
    definition: DefinitionWidget;
    ...
}
```

We use the helper: 

```typescript
export type PerseusWidgetsMap = MakeWidgetMap<PerseusWidgetTypes>;
```

And out comes the following type:

```typescript
type PerseusWidgetsMap = {
    [x: `categorizer ${number}`]: CategorizerWidget;
    [x: `cs-program ${number}`]: CSProgramWidget;
    [x: `definition ${number}`]: DefinitionWidget;
    ...
}
```

So you can think of this helper as a sort of type "generator". It takes a registry of widget types and generates a map type where the ID of the map is a widget id and the value is the correct option type.

packages/perseus/src/validation.types.ts
jeremywiebe: I plan to land this PR after your scoring name changes... so I'll be sure to update the instances of Rubric in this PR before landing.

packages/perseus/src/validation.types.ts
jeremywiebe: Oops, I should be consistent. I use `/** */` for comments that would be useful to see if I used the type somewhere else. 

Eg. 
image

packages/perseus/src/__tests__/validation.typetest.ts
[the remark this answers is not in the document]
jeremywiebe: This is a pattern I saw in Ben's parsing code changes. 

I've tried a different approach that will hopefully be clearer. My concern is that the two types being tested here (`PerseusRenderer["widgets"]` and `ScoringDataMap`) _must_ remain compatible with `ValidatioDataMap`. 

I don't another way to do that.

packages/perseus/src/validation.types.ts
jeremywiebe: Sorry, this PR drifted from what I initially typed out as the PR description. I'll update the description. It's basically type work to make the types we already use (full widget options, Scoring Data, Validation Data) a bit more clear and well-typed. 

I agree that I don't like us buying further into the mapping of widget ID to data in our various maps. Maybe we can chat about that in a future PR? I'm also happy to have pushback and do more work in this PR to isolate the Renderer and outer code knowing so much about widget option details.

packages/perseus/src/perseus-types.ts
Myranae: Thank you! I was thinking we could add an example to the comments, but that seems a bit long to add XD Thank you for writing that all out though. Helps to visualize it better ^_^

packages/perseus/src/__tests__/validation.typetest.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * This file contains TypeScript type "tests" which ensure that types needed
+ * for scoring and validation stay in sync with other types in the system.
+ *
+ * If you make a change and one of these "satisfies" assertions fails, that
```
Myranae, line 5: May want to update this line since it no longer uses `satisifies` assertions, though I suppose you could still think of them that way.

packages/perseus/src/perseus-types.ts
```diff
@@ -37,6 +37,50 @@ export type Size = [width: number, height: number];
...
+ * type DummyMap = MakeWidgetMap<DummyRegistry>;
+ * ```
+ *
+ * We'll get a map that looks like this:
+ *
+ * ```
+ * type DummyMap = {
+ *     `categorizer ${number}`: { categories: ReadonlyArray<string> };
+ *     `dropdown ${number}`: { choices: ReadonlyArray<string> };
+ * }
```
Myranae, line 73: Love this! Thanks for adding a more detailed example!

Myranae (approved):
Looks good to me! Thanks for your updates :)

PR #2004, opened 2024-12-13
Fix for svg-image loading 

SonicScrewdriver:
## Summary:
This PR updates our approach to loading graphie images for svg-image + a fix for a bug that was introduced while addressing PR feedback. 

An error will be thrown if the labels are not found, but should not impede users. Many tests were added using the failed cases provided on the ticket, in order to ensure that this new approach works for nearly all use cases. 

Issue: [LEMS-2524](https://khanacademy.atlassian.net/browse/LEMS-2524)

## Test plan:
- New svg-image tests + test mocking error logger 
- New graphie-utils tests that include (most) of the test cases provided.

[LEMS-2524]: https://khanacademy.atlassian.net/browse/LEMS-2524?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/perseus/src/util/graphie-utils.test.ts
+++ b/packages/perseus/src/util/graphie-utils.test.ts
@@ -1,7 +1,12 @@
 import {Dependencies} from "..";
 import {testDependencies} from "../../../../testing/test-dependencies";
+import {Log} from "../logging/log";
 
-import {parseDataFromJSONP, getLocalizedDataUrl} from "./graphie-utils";
+import {
+    parseDataFromJSONP,
+    getLocalizedDataUrl,
+    loadGraphie,
+} from "./graphie-utils";
 import {typicalCase, edgeCases} from "./graphie-utils.testdata";
 
 describe("graphie utils", () => {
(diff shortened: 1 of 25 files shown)


packages/perseus/src/util/graphie-utils.test.ts
```diff
@@ -0,0 +1,103 @@
...
+                });
+            }
+        }) as jest.Mock;
+
+        // Mock the error logger to confirm no errors are thrown
+        jest.spyOn(Log, "error").mockImplementation(
+            jest.fn(() => {
+                // Do nothing
+            }),
+        );
```
SonicScrewdriver, line 91: Mocking the error logger here to protect against an issue that was found from the previous deploy.

packages/perseus/src/util/graphie-utils.test.ts
```diff
@@ -0,0 +1,103 @@
...
+        // Act
+        await loadGraphie(typicalCase.url, (data, localized) => {
+            // Assert
+            expect(data).toEqual({labels: [], range: null});
+            expect(localized).toEqual(false);
+        });
+
+        // Assert
+        expect(Log.error).not.toHaveBeenCalled();
+    });
```
SonicScrewdriver, line 102: We shouldn't throw errors when we fall back to english, even though we hit a fetch error.

mark-fitzgerald (approved):

PR #2006, opened 2024-12-13
Rename usages of rubric to scoringData

Myranae:
## Summary:
This is to rename anything related to scoring that still uses rubric to scoringData instead to be more in alignment with the vocabulary for server side scoring.

Issue: LEMS-2657

## Test plan:
- Confirm all checks pass
- Confirm all checks pass in a webapp draft 
- Check for issues with widgets affected

### DIFF
--- a/packages/perseus/src/validation.types.ts
+++ b/packages/perseus/src/validation.types.ts
@@ -97,11 +97,11 @@ export type PerseusGradedGroupScoringData = PerseusGradedGroupWidgetOptions;
 export type PerseusGradedGroupSetScoringData =
     PerseusGradedGroupSetWidgetOptions;
 
-export type PerseusGrapherRubric = {
+export type PerseusGrapherScoringData = {
     correct: GrapherAnswerTypes;
 };
 
-export type PerseusGrapherUserInput = PerseusGrapherRubric["correct"];
+export type PerseusGrapherUserInput = PerseusGrapherScoringData["correct"];
 
 // TODO(LEMS-2440): Can possibly be removed during 2440; only userInput used
 export type PerseusIFrameRubric = Empty;
@@ -244,7 +244,7 @@ export type Rubric =
     | PerseusGroupScoringData
     | PerseusGradedGroupScoringData
     | PerseusGradedGroupSetScoringData
-    | PerseusGrapherRubric
+    | PerseusGrapherScoringData
     | PerseusIFrameRubric
     | PerseusInputNumberRubric
     | PerseusInteractiveGraphRubric
--- a/packages/perseus/src/validation.types.ts
+++ b/packages/perseus/src/validation.types.ts
@@ -94,7 +94,8 @@ export type PerseusGroupUserInput = UserInputMap;
 
 export type PerseusGradedGroupScoringData = PerseusGradedGroupWidgetOptions;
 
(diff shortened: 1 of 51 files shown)


packages/perseus/src/widgets/cs-program/score-cs-program.ts
```diff
@@ -1,23 +1,23 @@
 import type {PerseusScore} from "../../types";
 import type {PerseusCSProgramUserInput} from "../../validation.types";
 
-function scoreCSProgram(state: PerseusCSProgramUserInput): PerseusScore {
+function scoreCSProgram(userInput: PerseusCSProgramUserInput): PerseusScore {
```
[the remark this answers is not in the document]
Myranae: For reference, here's the[ PR comment where Jeremy mentions we should keep the parameter names the same as WidgetScorerFunction.](https://github.com/Khan/perseus/pull/1996#pullrequestreview-2503244464)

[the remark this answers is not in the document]
jeremywiebe: It is semantically incorrect, but I think for consistency sake it makes sense to send the correctness info inside `UserInput` for these types of widgets. 

It's not technically user input, but I don't think it's worth creating another way to send it. "User input" is close enough for now.

packages/perseus/src/validation.types.ts
```diff
@@ -228,35 +229,35 @@ export type PerseusSorterUserInput = {
...
     // Translatable Text; A 2-dimensional array of text to populate the table with
     answers: ReadonlyArray<ReadonlyArray<string>>;
 };
 
 export type PerseusTableUserInput = ReadonlyArray<ReadonlyArray<string>>;
 
-export type Rubric =
+export type ScoringData =
     | PerseusCategorizerScoringData
     | PerseusCSProgramRubric
```
jeremywiebe, line 241: There's some types here that haven't been renamed from *Rubric to *ScoringData. Is there a reason for that?

jeremywiebe (approved):

packages/perseus/src/validation.types.ts
Myranae: Oh, yes, these are changed in one of my other PRs

packages/perseus/src/widgets/radio/radio-component.tsx
```diff
@@ -268,12 +268,12 @@ class Radio extends React.Component<Props> implements Widget {
      */
     showRationalesForCurrentlySelectedChoices: (
```
handeyeco, line 269: My only reservation with these changes are touching some of the things we need to address in the future: `showRationalesForCurrentlySelectedChoices`, `reviewModeRubric`, etc. I think it's fine to ship the changes, there's just something in the back of my mind worried that we'll lose some context by renaming things around these APIs before we actually work on them.

I have no justification for this concern, so no action needed.

Myranae: @handeyeco I can go through and leave a link to this PR on the relevant tickets and just point out that the naming may have changed and here's where you can see where that happened. I think I might also suggest that if it hasn't been renamed that it get renamed in the course of whatever ticket work happens. Does that sound like a good way to help prevent some loss of context? 

I am also 100% okay with putting out another PR reverting any naming changes if we decide after I've landed this that we don't actually want some of these things renamed.

PR #2009, opened 2024-12-14
[Unlimited Polygon] Fix for creating concave polygons.

catandthemachines:
## Summary:
Moving the Polygon component lower in the stack to ensure the rect component is on top. This is necessary to allow users to draw a polygon that is concaved, because the Polyline blocks the click events.

https://github.com/user-attachments/assets/181a987f-c0a2-4ff9-bc08-cbf57df44fe9

Issue: LEMS-2679

## Test plan:
Go to: http://localhost:6006/?path=/story/perseus-widgets-interactive-graph--unlimited-polygon-with-mafs.
Draw a polygon that is concave and notice it should allow you to draw it.

### DIFF
--- a/.changeset/slow-rules-unite.md
+++ b/.changeset/slow-rules-unite.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Fixing bug in creating concave shapes in unlimited polygons.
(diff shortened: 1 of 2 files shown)


nishasy (approved):
Oh nice!! Glad this was an easy fix!

PR #2014, opened 2024-12-16
Migrate off deprecated ID APIs

somewhatabstract:
## Summary:
This migrates Perseus off the deprecated ID APIs and onto the `useId`-based approach.

The lint rules remain in place until Wonder Blocks is released without the old APIs and Perseus has been updated to that version of Wonder Blocks. This won't happen until webapp has been migrated off the old APIs.

Issue: WB-1812

## Test plan:
`yarn typecheck`
`yarn lint`
`yarn test`


mark-fitzgerald (approved):
Thank you for doing this conversion!

PR #2016, opened 2024-12-16
Label-image: Extract validation out of scoring

Myranae:
## Summary:
To complete server-side scoring, we are separating out validation logic from scoring logic. This PR separates that logic and updates associated tests.

Issue: LEMS-2609

## Test plan:
- Confirm checks pass


packages/perseus/src/widgets/label-image/validate-label-image.ts
```diff
@@ -0,0 +1,25 @@
...
+function validateLabelImage(
+    userInput: PerseusLabelImageUserInput,
+): Extract<PerseusScore, {type: "invalid"}> | null {
+    let numAnswered = 0;
+    for (let i = 0; i < userInput.markers.length; i++) {
+        const userSelection = userInput.markers[i].selected;
+        if (userSelection && userSelection.length > 0) {
+            numAnswered++;
+        }
+    }
```
Myranae, line 13: I re-worked the logic a bit here so as not to need scoringData. This should be the same logic; it just doesn't use `scoreMarker`.

packages/perseus/src/widgets/label-image/score-label-image.ts
```diff
@@ -53,21 +54,14 @@ function scoreLabelImage(
...
 
-    // We expect all question markers to be answered before grading.
-    if (numAnswered !== userInput.markers.length) {
-        return {
-            type: "invalid",
-            message: null,
-        };
+    const validationError = validateLabelImage(userInput);
+    if (validationError) {
+        return validationError;
```
handeyeco, line 64: I think this is going to need to be moved to the beginning of this function. It shouldn't have any impact on the code because we're iterating over the markers twice anyway.

packages/perseus/src/widgets/label-image/validate-label-image.test.ts
```diff
@@ -0,0 +1,29 @@
+import validateLabelImage from "./validate-label-image";
+
+import type {PerseusLabelImageUserInput} from "../../validation.types";
+
+describe("scoreLabelImage", function () {
```
handeyeco, line 5: Micro nit that's more personal taste than anything and I know it's just copied from the original, but in this context I prefer arrow functions.

```suggestion
describe("scoreLabelImage", () => {
```

handeyeco (approved):
Looks good. I like how focused the tests become when splitting out logic by concern. It's nice to see.

packages/perseus/src/widgets/label-image/score-label-image.ts
Myranae: Moved! Thanks for pointing this out :)

packages/perseus/src/widgets/label-image/validate-label-image.test.ts
Myranae: Fixed :)

PR #2017, opened 2024-12-16
[SR] Minor improvements to Circle graph SR strings

nishasy:
## Summary:
- Update the variable names to start with `sr` prefix.
- Use `srFormatNumber()` to make sure the numbers are
  formatted correctly.
- Update the coordinate types to `string` in the strings
  instead of `number` since they are now using
  `srFormatNumber()` (which returns a string)

## Test plan:
`yarn jest`

- Go to http://localhost:6006/iframe.html?globals=&args=&id=perseuseditor-widgets-interactive-graph--interactive-graph-circle&viewMode=story
- Confirm nothing has changed from before

### DIFF
--- a/.changeset/shiny-houses-promise.md
+++ b/.changeset/shiny-houses-promise.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+[SR] Minor improvements to Circle graph SR strings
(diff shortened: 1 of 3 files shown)


catandthemachines (approved):
Looks good to me 😊

PR #2022, opened 2024-12-17
Update to pre-release WB that deletes the ID stuff

somewhatabstract:
## Summary:
This updates Perseus to a version of Wonder Blocks where the custom ID generation API has been deleted.

This is going to be released just after the WB release, then webapp will be updated with both and deployed.

Issue: WB-1812

## Test plan:
`yarn typecheck`
`yarn lint`


anakaren-rojas (approved):

handeyeco (approved):
Thanks Jeff!

PR #2025, opened 2024-12-18
[SR] Linear graph - add grab handle description and aria lives

nishasy:
## Summary:
The [SRUX doc](https://khanacademy.atlassian.net/wiki/spaces/LC/pages/3460366348/Linear) still needs a label for the grab handle, but I tried my best in the meantime.

- Add a label and describedby for the grab handle.
- Add aria-live states for the different interactive elements
  so they don't override each other.

Issue: https://khanacademy.atlassian.net/browse/LEMS-1726

## Test plan:
`yarn jest packages/perseus/src/widgets/interactive-graphs/graphs/linear.test.tsx`

Storybook
- http://localhost:6006/iframe.html?id=perseuseditor-widgets-interactive-graph--interactive-graph-linear&viewMode=story
- Try all the different slopes and intercepts
- Move different elements and confirm that the updated aria-label
  is what is read out, and none of the other elements override the
  currently focused one.

https://github.com/user-attachments/assets/b528c467-dde0-457e-8b53-56fa8493e06c


packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-line.tsx
```diff
@@ -150,9 +188,16 @@ const Line = (props: LineProps) => {
...
+                aria-label={ariaLabel}
+                aria-describedby={ariaDescribedBy}
+                aria-live={ariaLive}
                 className="movable-line"
                 data-testid="movable-line"
                 style={{cursor: dragging ? "grabbing" : "grab"}}
+                // Indicate that this element is interactive.
+                // As a bonus, giving this group a non-group role makes
+                // the screen reader skip over its empty children.
+                role="button"
```
anakaren-rojas, line 200: I re-read the whole graph section of the SRUX doc earlier and noticed I missed this 

> An appropriate semantic element or role for the graph as a whole is [figure](https://khanacademy.atlassian.net/wiki/spaces/LC/pages/3261071420/Proposal+Screen+Reader+User+Experience+SRUX#Semantics).

packages/perseus/src/widgets/interactive-graphs/graphs/linear.test.tsx
```diff
@@ -50,21 +56,36 @@ describe("Linear graph screen reader", () => {
         );
     });
 
-    test("should have aria labels for both points on the line", () => {
+    test("should have aria labels and describedbys for both points and grab handle on the line", () => {
```
anakaren-rojas, line 59: this might need to be 4 separate tests: point + aria label, point + description, grab handle + aria label, grab handle + description

anakaren-rojas (approved):

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-line.tsx
nishasy: 💀 I missed that too.

I think this line is still `button` since it's referring to the grab handle, but I think I'll have to add `figure` to the parent layer.

PR #2030, opened 2024-12-18
[SR] Linear System - add screen reader support for Linear System interactive graph

nishasy:
## Summary:
Add the aria label and descriptions for the full graph and the
interactive elements in the Linear System graph, based on the
[SRUX doc](https://khanacademy.atlassian.net/wiki/spaces/LC/pages/3460366359/Linear+Systems).

Issue: https://khanacademy.atlassian.net/browse/LEMS-1727

## Test plan:
`yarn jest packages/perseus/src/widgets/interactive-graphs/graphs/linear-system.test.tsx`

Storybook
- http://localhost:6006/iframe.html?globals=&args=&id=perseuseditor-widgets-interactive-graph--interactive-graph-linear-system&viewMode=story

View Storybook publish in the checks below to try it yourself.

https://github.com/user-attachments/assets/007a3418-5dcb-470f-a3fd-45479cc3c4d2


nishasy:
The parent pull-request ([#2025](https://github.com/Khan/perseus/pull/2025)) has been merged into `main`, but this branch (`sr-linear-system`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

packages/perseus/src/widgets/interactive-graphs/graphs/linear-system.tsx
```diff
@@ -28,12 +31,70 @@ const LinearSystemGraph = (props: LinearSystemGraphProps) => {
...
+                locale,
+            ),
+            slopeDescription: getSlopeStringForLine(line, strings),
+        };
+    });
+
     return (
-        <>
+        <g
+            aria-label={strings.srLinearSystemGraph}
```
catandthemachines, line 60: This aria-label is not getting read out when navigating to the whole graph. This is because the whole graph descriptions needs to happen at the mafs-graph.tsx level.

catandthemachines (changes requested):
The experience is working for the individual points and lines, but I think we need to circle back and make sure it's working at the all up graph level.

packages/perseus/src/strings.ts
```diff
@@ -478,6 +516,13 @@ export const strings: {
     srAngleGraphAriaLabel: "An angle on a coordinate plane.",
     srAngleGraphAriaDescription:
         "The angle measure is %(angleMeasure)s degrees with a vertex at %(vertexX)s comma %(vertexY)s, a point on the starting side at %(startingSideX)s comma %(startingSideY)s and a point on the ending side at %(endingSideX)s comma %(endingSideY)s",
+    srLinearSystemGraph: "Two lines on a coordinate plane.",
+    srLinearSystemPoints:
+        "Line %(lineSequence)s has two points, point 1 at %(point1X)s comma %(point1Y)s and point 2 at %(point2X)s comma %(point2Y)s.",
```
benchristel, line 521: `%(lineSequence)s` made me think this is a placeholder for a description of a sequence of lines. Translators might find it confusing as well. I might prefer `lineSequenceNumber` or maybe just `lineNumber`.

```suggestion
        "Line %(lineSequenceNumber)s has two points, point 1 at %(point1X)s comma %(point1Y)s and point 2 at %(point2X)s comma %(point2Y)s.",
```

packages/perseus/src/widgets/interactive-graphs/graphs/linear-system.test.tsx
```diff
@@ -0,0 +1,298 @@
...
+        // Act
+        const linearSystemGraph = screen.getByLabelText(
+            "Two lines on a coordinate plane.",
+        );
+
+        // Assert
+        expect(linearSystemGraph).toBeInTheDocument();
+        expect(linearSystemGraph).toHaveAttribute(
+            "aria-describedby",
+            ":r1:-line1-points :r1:-line1-intercept :r1:-line1-slope :r1:-line2-points :r1:-line2-intercept :r1:-line2-slope",
```
benchristel, line 66: I wish we could assert on the description text instead of the IDs of the description elements. That way we'd be testing what the user actually sees. If React Testing Library doesn't have any matchers for this (I didn't find any with a cursory search) maybe we should write our own. Then we could do something like:

```ts
expect(linearSystemGraph).toHaveAriaDescription("line 1 through 0 comma 0 and 5 comma 5 ...")
```

Our custom matchers for Jest are defined in `config/test/custom-matchers.ts`.

packages/perseus/src/widgets/interactive-graphs/graphs/linear-system.test.tsx
```diff
@@ -0,0 +1,298 @@
...
+                expect(linearSystemGraph).toHaveTextContent(
+                    interceptDescription,
+                );
+            },
+        );
+
+        test.each`
+            case                 | coords              | slopeDescription
+            ${"positive slope"}  | ${[[1, 1], [3, 3]]} | ${`Its slope increases from left to right.`}
+            ${"negative slope"}  | ${[[3, 3], [1, 6]]} | ${`Its slope decreases from left to right.`}
```
benchristel, line 113: Maybe we should get a content creator to weigh in on this wording. Strictly speaking, I don't think the _slope_ increases or decreases; the slope is constant and it's the y-coordinate that increases or decreases.

Maybe "Its slope is positive/negative/zero/undefined" would be a good way to phrase this? In any case, I'd defer to Charlie or Nick.

packages/perseus/src/widgets/interactive-graphs/graphs/linear-system.tsx
```diff
@@ -28,12 +34,70 @@ const LinearSystemGraph = (props: LinearSystemGraphProps) => {
...
     const {coords: lines} = props.graphState;
 
+    const {strings, locale} = usePerseusI18n();
+    const id = React.useId();
+
+    const linesAriaInfo = lines.map((line, i) => {
+        return {
+            pointsDescriptionId: id + `-line${i + 1}-points`,
+            interceptDescriptionId: id + `-line${i + 1}-intercept`,
+            slopeDescriptionId: id + `-line${i + 1}-slope`,
```
benchristel, line 44: Minor: as long as we're using a template string, we could include `id` in it too:

```suggestion
            pointsDescriptionId: `${id}-line${i + 1}-points`,
            interceptDescriptionId: `${id}-line${i + 1}-intercept`,
            slopeDescriptionId: `${id}-line${i + 1}-slope`,
```

packages/perseus/src/widgets/interactive-graphs/graphs/linear-system.tsx
```diff
@@ -56,7 +120,82 @@ const LinearSystemGraph = (props: LinearSystemGraphProps) => {
...
+                            key={interceptDescriptionId}
+                            id={interceptDescriptionId}
+                            style={{display: "hidden"}}
+                        >
+                            {interceptDescription}
+                        </g>
+                        <g
+                            key={slopeDescriptionId}
+                            id={slopeDescriptionId}
+                            style={{display: "hidden"}}
```
benchristel, line 150: What is the purpose of `display: hidden` here? I don't think `hidden` is a valid value for `display` ([MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/display)) and either `display: none` or `visibility: hidden` [will hide the content from screenreaders](https://developer.mozilla.org/en-US/docs/Web/CSS/visibility#accessibility).

I think we probably want to apply an `srOnly` set of styles here.

packages/perseus/src/widgets/interactive-graphs/graphs/linear.tsx
```diff
@@ -49,57 +50,17 @@ const LinearGraph = (props: LinearGraphProps, key: number) => {
...
-        interceptString = hasXIntercept
-            ? strings.srLinearGraphXOnlyIntercept({
-                  xIntercept: srFormatNumber(xIntercept, locale),
-              })
-            : strings.srLinearGraphYOnlyIntercept({
-                  yIntercept: srFormatNumber(yIntercept, locale),
-              });
-    }
+    const slopeString = getSlopeStringForLine(line, strings);
+    const interceptString = getInterceptStringForLine(line, strings, locale);
```
benchristel, line 54: 👍 for refactoring this!

packages/perseus/src/widgets/interactive-graphs/graphs/utils.ts
```diff
@@ -62,3 +65,49 @@ export function getArrayWithoutDuplicates(array: Array<Coord>): Array<Coord> {
 
     return returnArray;
 }
+
+export function getSlopeStringForLine(line: PairOfPoints, strings): string {
+    const slope = (line[1][1] - line[0][1]) / (line[1][0] - line[0][0]);
+    if (slope === Infinity || slope === -Infinity) {
```
benchristel, line 71: FWIW, you could use `Number.isFinite` here:

```suggestion
    if (!Number.isFinite(slope)) {
```

That will also cause `srLinearGraphSlopeVertical` to be returned if the slope is `NaN`. (A slope of `NaN` shouldn't be possible, since we prevent the two control points of the line from having the same coordinates, so this is more paranoia than anything.)

packages/perseus/src/widgets/interactive-graphs/graphs/utils.ts
```diff
@@ -62,3 +65,49 @@ export function getArrayWithoutDuplicates(array: Array<Coord>): Array<Coord> {
...
+export function getInterceptStringForLine(
+    line: PairOfPoints,
+    strings,
+    locale,
+): string {
+    const slope = (line[1][1] - line[0][1]) / (line[1][0] - line[0][0]);
+    const xIntercept = (0 - line[0][1]) / slope + line[0][0];
+    const yIntercept = line[0][1] - slope * line[0][0];
+    const hasXIntercept = xIntercept !== Infinity && xIntercept !== -Infinity;
+    const hasYIntercept = yIntercept !== Infinity && yIntercept !== -Infinity;
```
benchristel, line 93: `Number.isFinite` would be a nice simplification here, too. Actually, I think there's a bug that isFinite would fix, because `NaN` _is_ a possibility here. E.g. if `line[0][1]` is 0 and the slope is also 0, the `xIntercept` will be `NaN` (since the line lies on the x-axis). Might be worth proving that with a test.

```suggestion
    const hasXIntercept = Number.isFinite(xIntercept);
    const hasYIntercept = Number.isFinite(yIntercept);
```

packages/perseus/src/widgets/interactive-graphs/graphs/linear-system.test.tsx
nishasy: That's a good idea. Personally, I don't think that should be a blocker, I think we should update the wording when we do our strings pass before creating translation tickets.

packages/perseus/src/widgets/interactive-graphs/graphs/linear-system.tsx
nishasy: omg

Thank you for pointing this out!

nishasy: I think this also means we can use `innerText` in the SR tree like you suggested rather than `textContent`.

packages/perseus/src/widgets/interactive-graphs/graphs/utils.ts
nishasy: Oh hmm. You're right that the line crossing the x-axis causes it to be weird. Right now, it says it crosses the x-axis at [some number] comma 0, which isn't really correct.

I'll ask Caitlyn if we should add a custom string for that case, and same thing for Y axis.

packages/perseus/src/widgets/interactive-graphs/graphs/linear-system.tsx
catandthemachines: @nishasy , can you remind me of the resolution we decided on this? I remember we discussed this before the holidays.

nishasy: I added the full graph description ("Interactive elements: ...") to this PR. You can see the commit below as "Add full graph description of all interactive elements" - [7d96885](https://github.com/Khan/perseus/pull/2030/commits/7d96885d5e33daa1ddbb6e466cff07980ca55184)

packages/perseus/src/widgets/interactive-graphs/graphs/linear-system.tsx
```diff
@@ -18,7 +23,9 @@ export function renderLinearSystemGraph(
 ): InteractiveGraphElementSuite {
     return {
         graph: <LinearSystemGraph graphState={state} dispatch={dispatch} />,
-        interactiveElementsDescription: null,
+        interactiveElementsDescription: (
+            <LinearSystemGraphDescription state={state} />
```
anakaren-rojas, line 27: should all of our graphs be updating this?

packages/perseus/src/widgets/interactive-graphs/graphs/linear-system.tsx
```diff
@@ -56,7 +121,82 @@ const LinearSystemGraph = (props: LinearSystemGraphProps) => {
...
+            )}
+        </g>
     );
 };
+
+function LinearSystemGraphDescription({
+    state,
+}: {
+    state: LinearSystemGraphState;
+}) {
```
anakaren-rojas, line 166: a nit since I still don't quite understand the purpose of the destructuring that's happening here
```suggestion
function LinearSystemGraphDescription(
    state: LinearSystemGraphState) {
```

packages/perseus/src/widgets/interactive-graphs/graphs/utils.ts
```diff
@@ -62,3 +65,54 @@ export function getArrayWithoutDuplicates(array: Array<Coord>): Array<Coord> {
 
     return returnArray;
 }
+
+export function getSlopeStringForLine(line: PairOfPoints, strings): string {
```
anakaren-rojas, line 69: can `strings` either be renamed to `perseusStrings` or the type added?

packages/perseus/src/widgets/interactive-graphs/graphs/utils.ts
```diff
@@ -62,3 +65,54 @@ export function getArrayWithoutDuplicates(array: Array<Coord>): Array<Coord> {
...
+
+    return slope > 0
+        ? strings.srLinearGraphSlopeIncreasing
+        : strings.srLinearGraphSlopeDecreasing;
+}
+
+export function getInterceptStringForLine(
+    line: PairOfPoints,
+    strings,
+    locale,
```
anakaren-rojas, line 87:
```suggestion
    locale : string,
```

anakaren-rojas (approved):

packages/perseus/src/widgets/interactive-graphs/graphs/linear-system.tsx
nishasy: @anakaren-rojas Yes! I went back and added it for Circle and Linear. We'll have to double check and make sure all the graphs have this.

packages/perseus/src/widgets/interactive-graphs/graphs/linear-system.tsx
nishasy: This is a component! The `{state}` is the props. It needs to be like that because it gets called like `<LinearSystemGraphDescription state={state} />`.

catandthemachines (approved):

packages/perseus/src/widgets/interactive-graphs/graphs/linear-system.tsx
anakaren-rojas: Oh I seeee
There's an alternative way to declare functional components that might make the distinction between a FC and a regular function more apparent to the layman like me 
```ts
type LinearSystemGraphDescriptionProps = {
 state: LinearSystemGraphState
}

const LinearSystemGraphDescription: React.FC<LinearSystemGraphDescriptionProps> =({state}) => {
```

PR #2032, opened 2024-12-18
feat(LEMS-1733): adds aria labels to line segment

anakaren-rojas:
## Summary:
- Adds aria label and description to whole line segment graph 
- Adds copy that distinguishes between single line segment on a coordinate plane and multiple line segments on a coordinate plane  
- Adds aria label for individual end points of line segment

Issue: LEMS-1733

## Test plan:
1. Navigate to [Chromatic](https://650db21c3f5d1b2f13c02952-crydbfdmmp.chromatic.com/?path=/docs/perseuseditor-editorpage--docs) 
2. Select Line Segment Interactive Graph from drop down
3. Validate screen reader can read line segment

#### Single Segment 
1. Screen reader reads aria label for whole graph: `A line segment on a coordinate plane.` 
2. Screen reader reads aria description for whole graph: `A line segment on a coordinate plane. Endpoint 1 at ${point1X} comma ${point1Y}. Endpoint 2 ${point2X} comma ${point2Y}. Segment length %(length)s units.`

#### Multiple Segments 
1. Screen reader reads aria label for whole graph: `{Number of segments} segment on a coordinate plane.` 
2. Screen reader reads aria description for whole graph: `{Number of segments} segment on a coordinate plane. Segment {N}: Endpoint 1 at ${point1X} comma ${point1Y}. Endpoint 2 ${point2X} comma ${point2Y}. Segment length %(length)s units.`

#### Individual Points 
1. When selected, screen reader reads aria label for end points: `Endpoint ${endpointNumber} at ${x} comma ${y}`
2. When moving a point, aria labels and descriptions adjust to include new end points and length, as appropriate

### Single Segment on a Coordinate Plane - Aria Label and Description
image

### Multiple segments on a coordinate plane - Aria Label and Description
image

### Individual Segment Aria Label - Description
image


packages/perseus/src/widgets/interactive-graphs/graphs/segment.tsx
```diff
@@ -24,12 +28,36 @@ export function renderSegmentGraph(
...
+    const wholeSegmentAriaLabel = strings.srSegmentGraphAriaLabel;
+    const wholeSegmentAriaDescription =
+        segments[0] &&
+        strings.srSegmentGraphAriaDescription({
+            point1X: srFormatNumber(segments[0][0][X], locale),
+            point1Y: srFormatNumber(segments[0][0][Y], locale),
+            point2X: srFormatNumber(segments[0][1][X], locale),
+            point2Y: srFormatNumber(segments[0][1][Y], locale),
+            length: srFormatNumber(lengthOfSegment, locale),
+        });
```
nishasy, line 46: EDIT: I know why this is happening. You have to use unique IDs to fix this. You can see how I did this for circle and linear if you need a reference.

I'm not really sure what's going on, but it seems like this is somehow using the correct answer or default coordinates instead of the ones that are on the graph? I'm extra confused because your implementation looks right to me... 🤔 

It has the right values for the points themselves, just not for the aria describedby.

https://github.com/user-attachments/assets/7977804d-3853-431e-87dc-028b5952292e

packages/perseus/src/widgets/interactive-graphs/graphs/segment.tsx
```diff
@@ -49,8 +77,23 @@ const SegmentGraph = (props: SegmentProps) => {
...
+                        point2AriaLabel: formatSegment(
+                            2,
+                            segment[1][X],
+                            segment[1][Y],
+                        ),
+                    }}
                 />
             ))}
-        </>
+            <g id="segment-description" style={{display: "hidden"}}>
```
nishasy, line 94: Please use `React.useId()` to create unique IDs. If there are ever multiple segment interactive graphs on the screen, they'd both read the description for the first one.

OHH!!! I JUST FIGURED OUT THE PROBLEM FROM THE ABOVE COMMENT WHILE WRITING THIS!

nishasy (commented):
A couple of small requests.

packages/perseus/src/widgets/interactive-graphs/graphs/segment.tsx
```diff
@@ -31,19 +32,23 @@ type SegmentProps = MafsGraphProps<SegmentGraphState>;
...
-        strings.srSegmentGraphAriaDescription({
-            point1X: srFormatNumber(segments[0][0][X], locale),
-            point1Y: srFormatNumber(segments[0][0][Y], locale),
-            point2X: srFormatNumber(segments[0][1][X], locale),
-            point2Y: srFormatNumber(segments[0][1][Y], locale),
-            length: srFormatNumber(lengthOfSegment, locale),
+
+    function getLengthOfSegment(segment: PairOfPoints) {
+        return kpoint.distanceToPoint(...segment);
+    }
```
nishasy, line 38: Nit (Optional): Since this component is not using stuff inside the component (everything it needs is passed into the parameters), you can move it outside the component, maybe at the top of this file. That way it won't re-initialize on every render.

This won't really make a difference in performance in this particular case, so it's really up to you, but I believe that's considered good practice.

packages/perseus/src/widgets/interactive-graphs/graphs/segment.tsx
```diff
@@ -54,46 +59,52 @@ const SegmentGraph = ({dispatch, graphState}: SegmentProps) => {
     }
 
     return (
-        <g
-            aria-label={wholeSegmentAriaLabel}
-            aria-describedby="segment-description"
-        >
+        <>
```
nishasy: I think we might still want a label on the overall segments graph saying something like "[Num] segments" so the user knows what to expect before going through each segment individually.

packages/perseus/src/widgets/interactive-graphs/graphs/segment.tsx
```diff
@@ -54,46 +59,52 @@ const SegmentGraph = ({dispatch, graphState}: SegmentProps) => {
...
-                            2,
-                            segment[1][X],
-                            segment[1][Y],
-                        ),
-                    }}
-                />
+                        }}
+                    />
+                    <g
+                        id={`segment-description-${i}`}
```
nishasy, line 136: You still need to use `React.useId()` to make sure that it's referring to the correct description if there are multiple interactive graphs on the page.

anakaren-rojas: Oh I see, you're talking about having multiple instances of the line segment graph on a single page, right? 
I wonder if the ID creation should be bubbled up to the Mafs graph to ensure that we're not getting the same ID in multiple instances for all the graphs, then the individual graphs can consume the IDs. 
It looks like we already have a [unique ID being generated there](https://github.com/Khan/perseus/blob/main/packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx#L95), we would just need to consume it

nishasy: You can bubble it down, but I think just adding another `React.useId()` at the top of the current file is much less work. Then you don't have to bubble it down into everything and keep passing it around everywhere even when not strictly necessary.

I'm not aware of performance issues with `React.useId()` so I'd personally go with that approach myself, but it's up to you.

nishasy (approved):
Nice! This is really coming together.

Note: We'll still have to figure out the aria label/description for the middle section (missing from the SRUX doc I think) after the holidays.

packages/perseus/src/strings.ts
```diff
@@ -478,6 +520,16 @@ export const strings: {
     srAngleGraphAriaLabel: "An angle on a coordinate plane.",
     srAngleGraphAriaDescription:
         "The angle measure is %(angleMeasure)s degrees with a vertex at %(vertexX)s comma %(vertexY)s, a point on the starting side at %(startingSideX)s comma %(startingSideY)s and a point on the ending side at %(endingSideX)s comma %(endingSideY)s",
+    srSingleSegmentGraphAriaLabel: "A line segment on a coordinate plane.",
+    srMultipleSegmentGraphAriaLabel:
+        "%(countOfSegments)s segments on a coordinate plane.",
+    srIndividualSegmentAriaLabel: "PLACEHOLDER: PLEASE UPDATE ME",
```
anakaren-rojas, line 526: @nishasy FYI - I added a placeholder for inner aria label, which should be updated when the design copy is finalized

packages/perseus/src/widgets/interactive-graphs/interactive-graph.test.tsx
```diff
@@ -1382,7 +1383,7 @@ describe("Interactive Graph", function () {
 
         it("should not have an aria-label or description if they are not provided", async () => {
             // Arrange
-            const {container} = renderQuestion(segmentQuestion, apiOptions);
```
catandthemachines, line 1385: @nishasy I'm assuming Anakaren used the "nonQuestion" testdata because segmentQuestions now have an aria-label?

catandthemachines (approved):
LGTM!

packages/perseus/src/widgets/interactive-graphs/interactive-graph.test.tsx
nishasy: Ah, I just changed that. Yes, it's because the graph container now has a description (and will for all graph types eventually), so I think only the None type graph might not have one.

PR #2033, opened 2024-12-18
[Interactive Graph][Axis labels] Make default x & y axis labels TeX on load

nishasy:
## Summary:
Right now, the x and y labels when an interactive graph is
first created show up as text. They should be showing up as
text.

- Add `$` around the "x" and "y" to make them show up as TeX

Issue: none

## Test plan:
`yarn jest packages/perseus/src/widget-ai-utils/interactive-graph/interactive-graph-ai-utils.test.ts`
`yarn jest packages/perseus-editor/src/widgets/interactive-graph-editor/components/interactive-graph-settings.test.tsx`

Before
Screenshot 2024-12-18 at 3 16 30 PM

After
Screenshot 2024-12-18 at 3 17 11 PM

### DIFF
--- a/.changeset/soft-lizards-lie.md
+++ b/.changeset/soft-lizards-lie.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/perseus": patch
+"@khanacademy/perseus-editor": patch
+---
+
+[Interactive Graph][axis labels] Make default x & y axis labels TeX on load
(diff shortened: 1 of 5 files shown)


jeremywiebe (approved):
Not that we _have_ to keep everything as Graphie did it, but Graphie's axis labels were always interpreted as TeX. So perhaps an easier solution is to force them to be rendered as TeX instead of making content authors wrap the labels in `$`?

image


You can try it out on the old `grapher` widget: https://www.khanacademy.org/devadmin/content/articles/survey-of-grapher/xe97b6553bdb0d1f7

catandthemachines (approved):
Looks good to me!

nishasy:
@jeremywiebe We don't want to wrap everything in TeX, because then a label like "Time" will be read by screen readers as "upper T i m e" (like T times I times M times E). Lone characters in TeX are treated as variables, so it makes sense for just "x" or just "y" but not for much more than that.

PR #2034, opened 2024-12-18
Remove final usage of createReactClass and dependency.

catandthemachines:
## Summary:
Removing the final usage of createReactClass and create-react-class dependency from Perseus.

Issue: LEMS-2668

## Test plan:
Run yarn test.
Go to ?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-polygon (or any widget editor)
Select `Developer JSON Mode`
Confirm you can edit and modify the json object.


.changeset/brown-eagles-shout.md
```diff
@@ -0,0 +1,6 @@
+---
+"@khanacademy/perseus": patch
+"@khanacademy/perseus-editor": patch
+---
+
+Removing final usage of createReactApp.
```
jeremywiebe, line 6: 😁  `s/App/Class` 

```suggestion
Removing final usage of `createReactClass`.
```

packages/perseus-editor/src/components/graph-settings.tsx
```diff
@@ -28,70 +28,111 @@ const defaultBackgroundImage = {
...
-            showRuler: false,
-            showTooltips: false,
-            rulerLabel: "",
-            rulerTicks: 10,
-        };
-    },
-
-    getInitialState: function () {
+type Props = {
+    editableSettings: any[];
```
jeremywiebe, line 32: The prototype looks like this was just `ReadonlyArray<"canvas" | "graph" | "snap" | "image" | "measure">`.  Can you try swapping `any` for that?

packages/perseus-editor/src/widgets/interaction-editor/interaction-editor.tsx
```diff
@@ -59,6 +46,21 @@ type DefaultProps = {
     graph: Props["graph"];
 };
 
+const defaultInteractionProps: DefaultProps = {
```
jeremywiebe, line 49: Given this variable is only used once, I'd be in favour if you wanted to just inline it.

packages/perseus-editor/src/components/json-editor.tsx
```diff
@@ -1,30 +1,62 @@
...
+        value: 0,
+    };
+
+    constructor(props) {
+        super(props);
+        this.state = this.getInitialState();
+
+        this.handleBlur = this.handleBlur.bind(this);
+        this.handleChange = this.handleChange.bind(this);
+        this.handleKeyDown = this.handleKeyDown.bind(this);
```
jeremywiebe, line 34: We haven't used this pattern much (if at all) in Perseus. Usually for these types of things we make the method an arrow function. That said, both approaches accomplish the same thing. 

Ideally we'd eventually move these to be functional components (this one would be relatively easy I think - in a future PR).

packages/perseus-editor/src/components/json-editor.tsx
```diff
@@ -1,30 +1,62 @@
...
+type State = {
+    currentValue: string | undefined;
+    valid: boolean | undefined;
+};
+
+class JsonEditor extends React.Component<Props, State> {
+    static displayName: "JsonEditor";
+
+    static defaultProps: DefaultProps = {
+        value: 0,
```
jeremywiebe, line 25: This seems a bit odd.  We `JSON.stringify()` `value` in other parts, so the default thing we'd edit is `"0"`?

packages/perseus-editor/src/components/graph-settings.tsx
```diff
@@ -317,7 +358,8 @@ const GraphSettings = createReactClass({
         const scale = Util.scaleFromExtent(ranges[i], this.props.box[i]);
         if (this.validRange(ranges[i]) === true) {
             step[i] = Util.tickStepFromExtent(ranges[i], this.props.box[i]);
-            gridStep[i] = Util.gridStepFromTickStep(step[i], scale);
+            // don't like this use of as. :(
```
jeremywiebe, line 361: Yeah. I guess it's that `gridStepFromTickStep` could return `null | undefined`. This is the way it is already so I think it's ok as we're kinda acknowledging that the old code didn't take that into account.

packages/perseus-editor/src/components/graph-settings.tsx
```diff
@@ -28,70 +28,111 @@ const defaultBackgroundImage = {
...
         this._isMounted = true;
 
         this.changeGraph = _.debounce(this.changeGraph, 300);
-    },
+    }
 
-    UNSAFE_componentWillReceiveProps: function (nextProps) {
+    UNSAFE_componentWillReceiveProps(nextProps) {
+        // Make sure that state updates when switching
+        // between different items in a multi-item editor.
```
jeremywiebe, line 135: Is this still relevant?  Matthew recently removed multi items entirely.

packages/perseus-editor/src/components/graph-settings.tsx
```diff
@@ -28,70 +28,111 @@ const defaultBackgroundImage = {
...
+        backgroundImage: defaultBackgroundImage,
+        markings: "graph",
+        rulerLabel: "",
+        rulerTicks: 10,
+        showProtractor: false,
+        showRuler: false,
+        showTooltips: false,
+    };
+
+    _isMounted: any;
```
jeremywiebe, line 103: I'd guess this is a `boolean`.  Can we just initialize it like this instead?

```suggestion
    _isMounted = false;
```

packages/perseus-editor/src/components/graph-settings.tsx
```diff
@@ -28,70 +28,111 @@ const defaultBackgroundImage = {
...
-
-    getInitialState: function () {
+type Props = {
+    editableSettings: any[];
+    box: readonly number[];
+    labels: readonly string[];
+    range: Coords;
+    step: readonly number[];
+    gridStep: readonly number[];
+    snapStep: number[];
```
jeremywiebe, line 38: I think most (all except `range`) are really a two-number tuple, right?  Does it accept if you type these as `[number, number]`?

jeremywiebe (approved):
Exciting work.  I saw a few type improvements, but nothing blocking as we can keep improving the types now that at least we aren't using `createReactClass()`!  

Thank-you!

PR #2037, opened 2024-12-19
Add prepublish checks to block releases of non-snapshot versionsfirst one

somewhatabstract:
## Summary:
This PR implements a simple preventative measure for the race condition described by https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/3571646568/Race+condition+breaks+Perseus+release whereby an in progress release triggered via merging a changesets PR and a snapshot release triggered by a PR update can lead to a snapshot releasing an actual versioned release.

The events that cause this look something like:
1. Version Packages merged, updating versions in `main`
2. Someone updates their PR with a merge of `main`, updating its versions to the ones in `main` that are not yet published
3. The PR action to publish a snapshot runs, but the main release isn't done yet so the new releases aren't present in NPM
4. The snapshot release tries to publish those packages before the main release has tried

This change should prevent that last step; failing the snapshot release.

We could look at trying other ways to prevent this, such as making the `release.yml`` workflow responsible for snapshot releases too and limit concurrency. However, even that won't fully prevent this since the merge of the "Version Packages" PR does not guarantee the run order of the release workflow and the snapshot workflow.

So, this helps prevent the incorrect publish occurring without impacting the official release process - only PR snapshot releases get affected in the cases where this might occur (which seem to be rare, for now).

This update also modifies our pre-publish checks to look for all errors before quitting, instead of quitting on the first one. This is a quality of life change for devs that modify this script and need to check all packages are passing without having to run, then fix, then run, then fix, repeatedly.

Issue: XXX-XXXX

## Test plan:
I ran `SNAPSHOT_RELEASE=1 npm publish --dry-run` on a package that had a non-snapshot release version, and it failed as expected. I also ran it on a package that had the correct `0.0.0-PR...` format, and it succeeded. I also ran a `npm publish --dry-run` on a package without the `SNAPSHOT_RELEASE` env var, and it succeeded.


utils/internal/pre-publish-utils.ts
```diff
@@ -2,27 +2,47 @@
...
+    // check that we are running our pre-publish check for this package
+    if (
+        !scripts.prepublishOnly ||
+        !scripts.prepublishOnly.includes("utils/package-pre-publish-check.sh")
+    ) {
+        console.error(
+            `ERROR: ${name} must have a "prepublishOnly" script that runs "utils/package-pre-publish-check.sh".`,
+        );
+        returnCode = false;
+    }
```
jeremywiebe, line 40: Yay. Thanks for adding this check here too!

packages/perseus-core/package.json
```diff
@@ -30,4 +31,4 @@
     },
     "peerDependencies": {},
     "keywords": []
-}
+}
```
jeremywiebe, line 34: OCD-level 10 comment:  looks like all the package.json lost their newline at end of file. I'm not sure if that matters, but Github flags it so... 🤷‍♂️

utils/internal/pre-publish-utils.ts
somewhatabstract, line 40: **note:** New check to make sure all the packages we publish are running our prepublish check to avoid snapshot releases usurping real releases.

jeremywiebe (approved):

utils/internal/pre-publish-utils.ts
```diff
@@ -32,29 +52,29 @@ const checkField = (pkgJson, field, value) => {
...
-        if (pkgJson[field] !== value) {
-            console.error(
-                `ERROR: ${
-                    pkgJson.name
-                } must have a "${field}" set to ${JSON.stringify(value)}.`,
-            );
-            process.exit(1);
+            returnCode = false;
         }
+    } else if (pkgJson[field] !== value) {
```
somewhatabstract, line 57: **note:** I just changed that nested `if` in an `else` to an `else if`.

utils/pre-publish-check-ci.ts
```diff
@@ -16,16 +16,25 @@ import {
...
+                !checkSource(pkgJson)
+            ) {
+                allPassed = false;
             }
         }
+
+        // Exit only after we've processed all the packages.
+        if (!allPassed) {
+            process.exit(1);
+        }
```
somewhatabstract, line 38: **note:** This means we check all the packages for all the things before we quit, rather than quitting on the first error. Makes it easier to do changes to the overall checks and ensure compliance.

packages/perseus-core/package.json
somewhatabstract: It doesn't make a difference. I just saved like usual - no idea why this means it doesn't get a newline.

PR #2038, opened 2024-12-19
Fix some naming discrepancies and simplify Matcher scoring type

Myranae:
## Summary:
Noticed some discrepanies in naming relating to calling the validation function and also noticed some scoring types still reference their widgetOptions. The scoring types for the group widgets still references widgetOptions, but here I updated the scoring type for Matcher to be more specific to what is actually used.

Issue: LEMS-2734

## Test plan:
- Confirm all checks pass
- Will do more manual testing after merging to the feature branch

### DIFF
--- a/.changeset/nine-planes-relax.md
+++ b/.changeset/nine-planes-relax.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/perseus": minor
+"@khanacademy/perseus-editor": minor
+---
+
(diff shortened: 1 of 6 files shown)


.changeset/nine-planes-relax.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/perseus": minor
```
handeyeco, line 2: nit, no action needed: I think this would be a patch

handeyeco (approved):

PR #2042, opened 2024-12-19
fix(LEMS-2738): Convert Expression Editor Answer keys to strings

anakaren-rojas:
## Summary:
- Converts expression editor keys to strings
- Updates `onDelete` handler to use index instead of key

Issue: LEMS-2738

## Test plan:

https://github.com/user-attachments/assets/60424f5a-f654-4afc-901f-de47505e053c

### DIFF
--- a/packages/perseus-editor/src/widgets/expression-editor.tsx
+++ b/packages/perseus-editor/src/widgets/expression-editor.tsx
@@ -211,7 +211,13 @@ class ExpressionEditor extends React.Component<Props, State> {
     handleRemoveForm: (answerKey: number) => void = (i) => {
         const answerForms = this.props.answerForms.slice();
         answerForms.splice(i, 1);
-        this.change({answerForms});
+
+        const updatedAnswerForms = answerForms.map((form, index) => ({
+            ...form,
+            key: index,
+        }));
+
+        this.change({answerForms: updatedAnswerForms});
     };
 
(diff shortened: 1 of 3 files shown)


packages/perseus-editor/src/widgets/expression-editor.tsx
```diff
@@ -565,6 +570,7 @@ class AnswerOption extends React.Component<
 
     handleImSure = () => {
         this.props.onDelete();
+        this.handleCancelDelete();
```
catandthemachines, line 573: Ooo, what's this do and why is it here?

catandthemachines (approved):
I have one comment on a line that looked odd that I have questions about. But once that's resolved it looks good to go.

packages/perseus-editor/src/widgets/expression-editor.tsx
anakaren-rojas: Its an existing function that closes the "I'm sure" button, I added it so it wouldn't stay open once someone deletes the current item

anakaren-rojas: Here is the before: 
https://github.com/user-attachments/assets/8838f716-dfd4-4aa7-995d-b15f4c5aa22a

Here is the after: 

https://github.com/user-attachments/assets/8766dcda-036d-4298-906f-018aec5b0859

PR #2044, opened 2024-12-19
Check for in progress releases before trying to do snapshots

somewhatabstract:
## Summary:
This PR modifies our node-ci workflow so that we don't bother trying to do snapshots if we see a release workflow in progress.

Issue: FEI-6062

## Test plan:
To test this out, we will need to:

1. Have a Version Packages release ready to merge (DO NOT MERGE YET)
2. Have a PR ready that will trigger a snapshot (a simple change in an existing package, like a comment - we're not actually landing the PR)
3. Merge that release
4. Immediately `git pull && git merge origin/main && git push` on the PR
5. Watch the workflows. The release workflow should succeed and publish the packages. The snapshot workflow should fail with a message about a release in progress, and it should update the PR comment saying the snapshot was not created.


.github/workflows/node-ci.yml
```diff
@@ -214,8 +214,24 @@ jobs:
...
+            # PR comment later.
+            - name: Check for release
+              id: check-release
+              run: |
+                  release_count=$(gh run list --workflow release.yml --json status --jq '[.[] | select(.status == "in_progress")] | length')
+                  echo "release_count=$release_count" >> $GITHUB_OUTPUT
+                  if [ "$release_count" -ne 0 ]; then
+                    echo "Error: There are $release_count releases in progress."
+                    exit 1
+                  fi
```
somewhatabstract, line 229: **note:** This checks for in-progress release workflows and fails the job if there are any.

.github/workflows/node-ci.yml
```diff
@@ -214,8 +214,24 @@ jobs:
...
+
             # We need to checkout all history, so that the changeseat tool can diff it
             - name: Checkout current commit
+              # Only bother if there are no releases in progress
+              if: steps.check-release.outputs.release-count == '0'
```
somewhatabstract, line 234: **note:** The job failure should mean this doesn't run, but no harm in being explicit.

.github/workflows/node-ci.yml
```diff
@@ -226,11 +242,15 @@ jobs:
                   git checkout $REF
 
             - name: Install & cache node_modules
+              # Only bother if there are no releases in progress
+              if: steps.check-release.outputs.release-count == '0'
```
somewhatabstract, line 246: **note:** The job failure should mean this doesn't run, but no harm in being explicit.

.github/workflows/node-ci.yml
```diff
@@ -226,11 +242,15 @@ jobs:
...
               uses: ./.github/actions/shared-node-cache
               with:
                   node-version: ${{ matrix.node-version }}
 
             - name: Publish Snapshot Release to npm
+              # Only bother if there are no releases in progress
+              if: steps.check-release.outputs.release-count == '0'
```
somewhatabstract, line 253: **note:** The job failure should mean this doesn't run, but no harm in being explicit.

.github/workflows/node-ci.yml
```diff
@@ -249,6 +269,8 @@ jobs:
             # Note: these two actions are locked to the latest version that were
             # published when I created this yml file (just for security).
             - name: Find existing comment
+              # Even if we're failing, we want to update the comments.
+              if: always()
```
somewhatabstract, line 273: **note:** We want to update the PR comment if we never even published, so we always run this step.

.github/workflows/node-ci.yml
```diff
@@ -295,3 +317,17 @@ jobs:
                       Oh noes!! We couldn't find any changesets in this PR (${{
                       steps.short-sha.outputs.short_sha }}). As a result, we did not
                       publish an npm snapshot for you.
+
+            - name: Create or update npm snapshot comment - failure, concurrent with release
+              if: always() && steps.publish-snapshot.outputs.npm_snapshot_tag == '' && steps.check-release.outputs.release-count != '0'
```
somewhatabstract, line 322: **note:** We want to update the PR comment if we never even published, so we always run this step.

.github/workflows/node-ci.yml
```diff
@@ -295,3 +319,17 @@ jobs:
...
```
jeremywiebe, line 324: 🤔  Won't this skip in cases where we didn't get a snapshot tag but the `release-count` is `0`?

Wouldn't this be what we want?

```suggestion
              if: failure() || steps.publish-snapshot.outputs.npm_snapshot_tag == '' || steps.check-release.outputs.release-count != '0'
```

jeremywiebe (approved):
Thanks for working on this Jeff.  I think the `if` check in the failure comment step needs tweaking.  Otherwise just some comments/opinions. :)

.github/workflows/node-ci.yml
somewhatabstract: Yeah, that's what just happened in my test. 🤦

And yes, we can do `failure()` too :)

.github/workflows/node-ci.yml
somewhatabstract: Yeah, that's reasonable. I added these before I changed to failing hard and I just didn't want to do the work of deleting it. Totally lazy. I'll just delete these. :grin:

.github/workflows/node-ci.yml
somewhatabstract: Hmm, this still didn't work...I must've messed something up here

somewhatabstract: Oh, it was the or vs and 🤦🏻

somewhatabstract: > 🤔 Won't this skip in cases where we didn't get a snapshot tag but the release-count is 0?

Yes, because this is solely for when we have an active release. There is no message for if the publish didn't happen and we hit failure. There's no path for that; never was.

somewhatabstract: I just spotted the bug.

`release-count`. It's output as `release_count` in the step. 🤦🏻

PR #2053, opened 2024-12-20
[Interactive Graph] New Axes type for graph markings.

catandthemachines:
## Summary:
At request from content creators they would like to have a new marking type for graph that will show the axes without requiring a full grid.

image
image

Issue: LEMS-2713

## Test plan:
Go to: http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-linear or any other graph editor.
Select `Axes` option and notice new axes graph markings.


packages/perseus-editor/src/widgets/interactive-graph-editor/components/interactive-graph-settings.tsx
```diff
@@ -75,11 +75,12 @@ type Props = {
 
     /**
      * The type of markings to display on the graph.
+     * - axes: shows the axes without the gride lines
      * - graph: shows the axes and the grid lines
      * - grid: shows only the grid lines
      * - none: shows no markings
      */
```
benchristel, line 82: Since the possible values are not actually defined on this line anymore, it seems like this comment might drift out of sync with the code. How would you feel about documenting the different values where `MarkingsType` is defined (in perseus-types.ts)?

benchristel (commented):
The code looks good! I think we should probably have a visual test for the `axes` option, in `interactive-graph-regression.stories.ts`.

benchristel (approved):
LGTM! Thanks for the updates.

Don't forget to review and accept the snapshot test changes on Chromatic (via the link in the build status) :)

PR #2054, opened 2024-12-20
[Dropdown] Bugfix - Render options and placeholder inline

mark-fitzgerald:
## Summary: ##
The dropdown widget was not rendering content properly. Punctuation and other non-alphanumeric characters were causing the content to break to their own lines, even without TeX markup. Removing the `inline` setting for the `<Renderer>` usages resolved this.

Issue: LEMS-2741

## Test plan:
1. Open Storybook
1. Navigate to the Editor Demo page
1. Add a Dropdown widget
1. Type in TeX content for one of the option
1. Add another option with regular text (all alphanumeric characters)
1. Add another option with punctuation or other non-alphanumeric characters ($%^<>,.:)
1. Viewing the dropdown options in the preview section should show all of the options on their own lines (single line for each)
1. Repeat these steps for the Placeholder portion of the widget

### DIFF
--- a/.changeset/hungry-taxis-clap.md
+++ b/.changeset/hungry-taxis-clap.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+[Dropdown] Bugfix - Render options and placeholder inline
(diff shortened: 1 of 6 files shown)


jeremywiebe (approved):
Looks good.  This is what we paired through and I think it's correct.

jeremywiebe (commented):
Looks good with adding the `className`

jeremywiebe:
> Looks good with adding the `className`

Mark and I discovered that the attempted fix with `className` did _not_ work. I'm reverting the commits that introduced it and we'll deal with the Articles layout issues in `dropdown` in 2025!

PR #2060, opened 2024-12-24
[SR] Circle - Add the interactive elements circle description to the whole graph container

nishasy:
## Summary:
Add the interactive Circle graph description to the full graph container.

This adds the "Interactive elements: Circle..." description to the outermost graph container.

## Test plan:
- Go to https://650db21c3f5d1b2f13c02952-iupnsmdbhn.chromatic.com/iframe.html?globals=&args=&id=perseuseditor-widgets-interactive-graph--interactive-graph-circle&viewMode=story
- Use a screen reader to get to the outermost graph container
- Confirm that it has the circle description after "Interactive elements:"

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/circle.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/circle.tsx
@@ -31,7 +31,9 @@ export function renderCircleGraph(
 ): InteractiveGraphElementSuite {
     return {
         graph: <CircleGraph graphState={state} dispatch={dispatch} />,
-        interactiveElementsDescription: CircleGraphDescription({state}),
+        interactiveElementsDescription: (
+            <CircleGraphDescription state={state} />
+        ),
     };
 }
 
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/circle.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/circle.tsx
@@ -16,6 +16,7 @@ import {
     useTransformVectorsToPixels,
 } from "./use-transform";
 
+import type {PerseusStrings} from "../../../strings";
 import type {
     AriaLive,
     CircleGraphState,
@@ -30,7 +31,7 @@ export function renderCircleGraph(
 ): InteractiveGraphElementSuite {
     return {
         graph: <CircleGraph graphState={state} dispatch={dispatch} />,
-        interactiveElementsDescription: null,
+        interactiveElementsDescription: CircleGraphDescription({state}),
     };
 }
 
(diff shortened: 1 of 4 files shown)


packages/perseus/src/widgets/interactive-graphs/graphs/circle.tsx
```diff
@@ -221,3 +207,59 @@ function crossProduct<A, B>(as: A[], bs: B[]): [A, B][] {
     }
     return result;
 }
+
+function CircleGraphDescription({state}: {state: CircleGraphState}) {
+    // CircleGraphDescription needs to the `usePerseusI18n`, hook so it has]
```
anakaren-rojas, line 212: ?
```suggestion
    // CircleGraphDescription needs the `usePerseusI18n`, hook so it has
```

packages/perseus/src/widgets/interactive-graphs/graphs/circle.tsx
```diff
@@ -221,3 +207,59 @@ function crossProduct<A, B>(as: A[], bs: B[]): [A, B][] {
     }
     return result;
 }
+
+function CircleGraphDescription({state}: {state: CircleGraphState}) {
```
anakaren-rojas, line 211: this is the same as `state: CircleGraphState` right? Or is there an additional benefit to the destructing in this case?

packages/perseus/src/widgets/interactive-graphs/graphs/circle.tsx
```diff
@@ -221,3 +207,59 @@ function crossProduct<A, B>(as: A[], bs: B[]): [A, B][] {
...
+    const i18n = usePerseusI18n();
+    const strings = describeCircleGraph(state, i18n);
+
+    return strings.srCircleInteractiveElement;
+}
+
+// Exported for testing
+export function describeCircleGraph(
+    state: CircleGraphState,
+    i18n: {strings: PerseusStrings; locale: string},
```
anakaren-rojas, line 223: theres a `I18nContextType` that might be worth exporting to use here and around in the other graphs as needed. Alternatively, we could create a type in a utility file and destructure 
`{strings: PerseusStrings; locale: string}: I18ContextType` remove line 225

packages/perseus/src/widgets/interactive-graphs/graphs/circle.tsx
nishasy: I think that works too. I _think_ the reason it was done this way (I was following what Ben did for the point description) is so it follows the prop structure the way they're standardly written for other functional components.

packages/perseus/src/widgets/interactive-graphs/graphs/circle.tsx
```diff
@@ -30,7 +31,7 @@ export function renderCircleGraph(
 ): InteractiveGraphElementSuite {
     return {
         graph: <CircleGraph graphState={state} dispatch={dispatch} />,
-        interactiveElementsDescription: null,
+        interactiveElementsDescription: CircleGraphDescription({state}),
```
benchristel, line 34: I'm slightly surprised that this works. Since `CircleGraphDescription` is calling a hook, I'd expect it to have to be used as a JSX element:

```tsx
<CircleGraphDescription state={state} />
```

If the component calling `renderCircleGraph` does so conditionally, e.g.

```ts
if (graphType === "circle") {
  return renderCircleGraph(state, dispatch)
}
```

And `graphType` changes between renders, this code will break the Rules of Hooks. The `usePerseusI18n` hook will be called on one render and not called on the next.

I think it's working (mostly?) because the graph type rarely changes between renders. It probably only does so in the content editor.

packages/perseus/src/widgets/interactive-graphs/graphs/circle.tsx
```diff
@@ -51,40 +52,25 @@ function CircleGraph(props: CircleGraphProps) {
...
-        point4X: srFormatNumber(center[0], locale),
-        point4Y: srFormatNumber(center[1] - radius, locale),
-    });
+    const {
+        srCircleGraph,
+        srCircleShape,
+        srCircleRadiusPoint,
+        srCircleRadius,
+        srCircleOuterPoints,
+    } = describeCircleGraph(graphState, {strings, locale});
```
benchristel, line 61: I like this way of splitting up the logic! Makes it much easier to read and test :)

packages/perseus/src/widgets/interactive-graphs/graphs/circle.tsx
nishasy: Ack! My mistake. I 100% meant to use your code for Point as the blueprint for this, which means I was supposed to write `<CircleGraphDescription state={state} />`.

packages/perseus/src/widgets/interactive-graphs/graphs/circle.tsx
nishasy: Yeah, here's the real reason:

https://github.com/Khan/perseus/pull/2060/files#r1905985319

benchristel (approved):
LGTM!

PR #2068, opened 2025-01-03
Redesign discriminated union type parser

benchristel:
Previously, the `discriminatedUnion` parser required you to specify the
discriminant via an object parser, which was confusing and verbose. Now you only
have to specify the discriminant key once, and provide the values for each
branch.

Issue: LEMS-2582

## Test plan:

`yarn test`

### DIFF
--- a/packages/perseus/src/util/parse-perseus-json/general-purpose-parsers/discriminated-union.test.ts
+++ b/packages/perseus/src/util/parse-perseus-json/general-purpose-parsers/discriminated-union.test.ts
@@ -2,83 +2,118 @@ import {parse} from "../parse";
 import {failure, success} from "../result";
 
 import {constant} from "./constant";
-import {discriminatedUnion} from "./discriminated-union";
+import {discriminatedUnionOn} from "./discriminated-union";
 import {number} from "./number";
 import {object} from "./object";
 
-describe("a discriminatedUnion with one variant", () => {
-    const unionParser = discriminatedUnion(
-        object({type: constant("ok")}),
-        object({type: constant("ok"), value: number}),
-    ).parser;
+describe("a discriminatedUnion with no variants", () => {
+    const parseUnion = discriminatedUnionOn("shape").parser;
 
-    it("parses a valid value", () => {
-        const input = {type: "ok", value: 3};
-
-        expect(parse(input, unionParser)).toEqual(success(input));
-    });
-
-    it("rejects a value with the wrong `type`", () => {
-        const input = {type: "bad", value: 3};
-
-        expect(parse(input, unionParser)).toEqual(
-            failure(`At (root).type -- expected "ok", but got "bad"`),
+    it("fails appropriately given a non-object", () => {
+        expect(parse(true, parseUnion)).toEqual(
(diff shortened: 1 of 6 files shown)


packages/perseus/src/util/parse-perseus-json/general-purpose-parsers/discriminated-union.typetest.ts
```diff
@@ -0,0 +1,65 @@
...
+    const parser = discriminatedUnionOn("shape")
+        .withBranch("circle", parseCircle)
+        .withBranch("rectangle", parseRectangle)
+        .withBranch("square", parseSquare).parser;
+
+    parser satisfies Parser<Figure>;
+
+    // Guard against implicit 'any' type
+    // @ts-expect-error - Type '{ shape: "circle"; radius: number; }' is not assignable to type 'string'.
+    parser satisfies Parser<string>;
```
jeremywiebe, line 40: I understand all the tests in the file except this one. Why would this potentially be implicit any? Or what does this test protect against?

packages/perseus/src/util/parse-perseus-json/general-purpose-parsers/discriminated-union.test.ts
```diff
@@ -2,83 +2,118 @@ import {parse} from "../parse";
...
+        expect(parse({}, parseUnion)).toEqual(
+            failure(
+                "At (root).shape -- expected a valid value, but got undefined",
+            ),
+        );
+    });
 
-        expect(parse(input, unionParser)).toEqual(
+    it("fails appropriately given an object with an invalid discriminant", () => {
+        expect(parse({shape: "squarle"}, parseUnion)).toEqual(
```
jeremywiebe, line 96: A, my favourite geometric shape, the squarle! 😂  🐿️

jeremywiebe (approved):
Nice, thanks for that improvement!

packages/perseus/src/util/parse-perseus-json/general-purpose-parsers/discriminated-union.typetest.ts
benchristel: Sometimes, when you do complicated things with types, TypeScript can't figure out what the type of an expression is so it just infers `any` 😱 . I was worried that the type of `parser` would be inferred as `any`.

`parser satisfies Parser<Figure>;` asserts that `parser` is assignable to a variable of type `Parser<Figure>;`. This assertion will succeed if `parser` is actually a `Parser<Figure>`, but it will _also_ succeed if `parser` has type `any`.

To guard against that case, I always include a negative assertion in my type tests as well: `parser` should not be assignable to `Parser<string>`. If the type is `any`, it _will_ be assignable, and we'll get an error about the `@ts-expect-error` being unused.

I wish TypeScript had a way to compare types exactly, so I didn't have to do this.

benchristel: TIL: https://stackoverflow.com/a/53808212

PR #2071, opened 2025-01-03
Bugfix - Ensure that all strings are double escaped.

SonicScrewdriver:
## Summary:
After our migration to Lingui (which included a new approach to loading in the Perseus strings in the `extract-perseus-strings.tsx` file), it was discovered that we now need to ensure that we always double escape our tex. 

This is a simple PR that adds these double escapes, and a comment for our future selves. 

Issue: LEMS-2746

## Test plan:
- manual testing locally and on a znd

### DIFF
--- a/.changeset/hot-eels-arrive.md
+++ b/.changeset/hot-eels-arrive.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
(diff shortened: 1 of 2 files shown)


mark-fitzgerald (approved):
Great catch! Thank you.

PR #2072, opened 2025-01-06
Mock Widget for cleaning up Tests

SonicScrewdriver:
## Summary:
This work was done as part of the Numeric Input project. 

This PR introduces a new Mock Widget to be used for any tests that doesn't require testing the capabilities of a specific widget. I have updated as many tests and/or testdatas as I could find, except for a few key situations:

- The test requires that we test a specific widget, such as extract-perseus-data. (Although I would like to recommend that we break up and move these tests into the individual widget tests as all we need to do is import a single function.) 
- The testdata is being used for storybook

I have also updated as many occurrences of any remaining uses of `input-number `in our tests to `numeric-input`, in order to help smooth the path forward for the eventual return to the Input Number Conversion project.   

I have opted not to include the widget in our extra-widgets / general registration for now as it would require that I also create an editor for it, but a caveat of this approach means that this widget is not accessible for Webapp tests. I would be happy to hear any thoughts regarding this approach, as I wasn't sure if I should be globally registering this widget. 

Issue: LEMS-2615

## Test plan:
- ensure all tests pass


packages/perseus/src/validation.types.ts
```diff
@@ -235,6 +235,14 @@ export type PerseusTableRubric = {
 
 export type PerseusTableUserInput = ReadonlyArray<ReadonlyArray<string>>;
 
+export type PerseusMockWidgetRubric = {
```
SonicScrewdriver, line 238: Note to self: I should ensure these are ordered alphabetically

packages/perseus/src/perseus-types.ts
```diff
@@ -330,6 +331,8 @@ export type MoleculeRendererWidget = WidgetOptions<'molecule-renderer', PerseusM
 export type RefTargetWidget = WidgetOptions<'passage-ref-target', PerseusPassageRefTargetWidgetOptions>;
 // prettier-ignore
 export type VideoWidget = WidgetOptions<'video', PerseusVideoWidgetOptions>;
+// prettier-ignore
+export type MockWidget = WidgetOptions<'mock-widget', MockWidgetOptions>;
```
SonicScrewdriver, line 335: Note to self: I should ensure these are ordered alphabetically

packages/perseus/src/widgets/mock-widgets/score-mock-widget.ts
```diff
@@ -0,0 +1,38 @@
+import KhanAnswerTypes from "../../util/answer-types";
```
SonicScrewdriver, line 1: This will probably end up getting moved, or possible removed, as part of the Server Side Scoring project - which is fine!

jeremywiebe (approved):
I love this! Moving away from using production widgets for tests that aren't testing that widget should make for much less churn as we maintain/improve each widget!

packages/perseus/src/__tests__/renderer-api.test.tsx
```diff
@@ -153,13 +156,13 @@ describe("Perseus API", function () {
 
     describe("CSS ClassNames", function () {
         describe("perseus-focused", function () {
-            it("should be on an input-number exactly when focused", async function () {
+            it("should be on an mock-widget exactly when focused", async function () {
```
mark-fitzgerald, line 159: Nit: "an" => "a"

mark-fitzgerald (approved):
Thank you for all the work to clean up these loose ends (aka input number widget)!!

packages/perseus/src/__tests__/renderer.test.tsx
```diff
@@ -1718,20 +1714,20 @@ describe("renderer", () => {
...
                         static: true,
                     },
                 },
             });
             const cb = jest.fn();
 
             // Act
-            act(() => renderer.setInputValue(["input-number 2"], "1000", cb));
+            act(() => renderer.setInputValue(["mock-widget 2"], "1000", cb));
             act(() => jest.runOnlyPendingTimers());
```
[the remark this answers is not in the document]
SonicScrewdriver: That's actually one of the few calls that I _cannot_ remove the jest call from, but I WAS able to remove them from 6 other locations in this file, plus the call in afterEach. 

The tests all seems to run reliably without the ones I removed, but the remaining ones are all necessary for tests to pass, usually due to the use of _.defer() in how we handle blurs in the renderer.

packages/perseus/src/index.ts
```diff
@@ -223,6 +223,7 @@ export type {
     CollinearTuple,
     MathFormat,
     InputNumberWidget, // TODO(jeremy): remove?
+    NumericInputWidget,
```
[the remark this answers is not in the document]
SonicScrewdriver: Oh I should have caught this, thank you! 

I changed some `testdata` in `perseus-editor` from `InputNumber` to `NumericInput`. It's nice to have the widget types there, but completely unneccessary.

In fact, that change also means we no longer need to export `InputNumberWidget` either, so I can remove both! :) (I made sure to check Webapp and we're not using it there either.)

PR #2082, opened 2025-01-08
Enable parsePerseusItem to handle all published Perseus content

benchristel:
This PR fixes the remaining cases where the parser couldn't handle some data in
our content corpus -- notably, in articles and international content. After this
PR is merged, we will be able to use the parser in Webapp!

Note that running the exhaustive test tool still produces some failures.
However, I suspect the failing content isn't published, because it either
doesn't render (crashes the page) or can't be scored (throws an exception when
you click the "check answer" button). We'll find out when we start logging
parser errors in production whether I'm right about this.

The remaining errors are:

```
(root).question.widgets["grapher N"].options.correct.coords -- expected array of length 2; got []
(root).question.widgets["matcher N"].options -- expected object; got undefined
(root).question.widgets["graded-group N"].options.widgets["numeric-input N"].options.answers[N].answerForms[N] -- expected "integer", "mixed", "improper", "proper", "decimal", "percent", "pi"; got "number"
(root).question.widgets["example-graphie-widget N"] -- expected a valid widget type; got "example-graphie-widget"
(root).question.widgets["image N"]["(widget key)"][1] -- expected a string representing a positive integer; got "0"
(root).question.widgets["explanation N"]["(widget key)"][1] -- expected a string representing a positive integer; got "0"
```

Issue: LEMS-2582

## Test plan:

`yarn test`

### DIFF
--- a/packages/perseus/src/util/parse-perseus-json/perseus-parsers/grapher-widget.ts
+++ b/packages/perseus/src/util/parse-perseus-json/perseus-parsers/grapher-widget.ts
@@ -11,7 +11,6 @@ import {
     string,
     union,
 } from "../general-purpose-parsers";
-import {defaulted} from "../general-purpose-parsers/defaulted";
 import {discriminatedUnionOn} from "../general-purpose-parsers/discriminated-union";
 
 import {parseWidget} from "./widget";
--- a/packages/perseus/src/util/parse-perseus-json/perseus-parsers/grapher-widget.ts
+++ b/packages/perseus/src/util/parse-perseus-json/perseus-parsers/grapher-widget.ts
@@ -42,7 +42,7 @@ export const parseGrapherWidget: Parser<GrapherWidget> = parseWidget(
                 "absolute_value",
                 object({
                     type: constant("absolute_value"),
-                    coords: pairOfPoints,
+                    coords: nullable(pairOfPoints),
                 }),
             )
             .withBranch(
@@ -50,21 +50,14 @@ export const parseGrapherWidget: Parser<GrapherWidget> = parseWidget(
                 object({
                     type: constant("exponential"),
                     asymptote: pairOfPoints,
-                    coords: pairOfPoints,
+                    coords: nullable(pairOfPoints),
                 }),
             )
             .withBranch(
                 "linear",
                 object({
                     type: constant("linear"),
-                    coords: defaulted(
-                        pairOfPoints,
-                        () =>
-                            [
-                                [-5, 5],
-                                [5, 5],
-                            ] as [[number, number], [number, number]],
-                    ),
+                    coords: nullable(pairOfPoints),
                 }),
             )
             .withBranch(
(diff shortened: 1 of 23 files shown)


packages/perseus/src/perseus-types.ts
```diff
@@ -550,11 +550,9 @@ export type GraphRange = [
...
     | {
           type: "absolute_value";
-          coords: [
-              // The vertex
-              Coord, // A point along one line of the absolute value "V" lines
-              Coord,
-          ];
+          // If `coords` is null, the graph will not be gradable. All answers
+          // will be scored as invalid.
+          coords: null | [vertex: Coord, secondPoint: Coord];
```
benchristel, line 555: There's no way to default coords sensibly if it's missing, since it represents the correct answer. The widget renders fine with null coords, but scoring threw an exception. I've fixed the exception by returning a score of "invalid" if correct.coords is null (see score-grapher.ts below).

packages/perseus/src/util/parse-perseus-json/perseus-parsers/grapher-widget.ts
```diff
@@ -42,58 +41,51 @@ export const parseGrapherWidget: Parser<GrapherWidget> = parseWidget(
...
                 object({
                     type: constant("linear"),
-                    coords: defaulted(
-                        pairOfPoints,
-                        () =>
-                            [
-                                [-5, 5],
-                                [5, 5],
-                            ] as [[number, number], [number, number]],
-                    ),
```
benchristel, line 67: I previously thought we could default `coords` here, but thought better of it. This default isn't appropriate if the graph's X and Y ranges aren't the defaults ([-10, 10]).

packages/perseus/src/util/parse-perseus-json/perseus-parsers/measurer-widget.ts
```diff
@@ -17,7 +17,14 @@ import type {Parser} from "../parser-types";
 export const parseMeasurerWidget: Parser<MeasurerWidget> = parseWidget(
     constant("measurer"),
     object({
-        image: parsePerseusImageBackground,
+        // The default value for image comes from measurer.tsx.
```
jeremywiebe, line 20: Thanks for adding these comments in!  I think they'll be really helpful going forward.

packages/perseus/src/util/parse-perseus-json/perseus-parsers/widgets-map.test.ts
```diff
@@ -34,6 +34,29 @@ describe("parseWidgetsMap", () => {
         expect(result).toEqual(anyFailure);
     });
 
+    it("rejects a key with ID 0", () => {
```
jeremywiebe, line 37: We usually refer to this value as the widget Id and not the "key". I think it's a useful differentiation because it avoids being conflated with a React component key.

jeremywiebe (approved):

packages/perseus/src/util/parse-perseus-json/perseus-parsers/widgets-map.test.ts
benchristel: What are the parts of a widget ID called, then? If the widget ID is something like `radio 1`, what is the `radio` part called and what is the `1` part called?

benchristel: I think I'm just going to call it the "widget number".

jeremywiebe: `radio` is the widget type (in some places its referred to as the widget name). 

`radio 1` is the widget id and I've been working hard to avoid folks thinking of them as being "made up" of a type and number, but rather that we push for them to be opaque identifiers.

benchristel: Hmm, well, they're definitely not treated as opaque currently, because if the number is 0 the entire page crashes. I've tried to avoid referring directly to the number in the tests and variable names.

jeremywiebe: Agreed, we do parse these widget ids in places, but I feel it's a thing we should avoid and remove when we can. Interpreting IDs as informative structures has bit me many times in the past when needs cause us to change the ID in some way. We hit this during Goliath when we assumed that KAIDs were of a specific format (legacy IDs weren't of that format).

PR #2085, opened 2025-01-08
Export Item and Article parsing functions from the 'perseus' package

benchristel:
This is the public API for Perseus JSON parsing that we'll call in Webapp.
See [ADR 773][1] for context on why this is needed.

[1]: https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/3318349891/ADR+773+Validate+widget+data+on+input+in+Perseus

Issue: https://khanacademy.atlassian.net/browse/LEMS-2774

## Test plan:

`yarn test`

### DIFF
--- a/packages/perseus/src/util/parse-perseus-json/perseus-parsers/perseus-image-background.ts
+++ b/packages/perseus/src/util/parse-perseus-json/perseus-parsers/perseus-image-background.ts
@@ -10,8 +10,8 @@ import {
 import {convert} from "../general-purpose-parsers/convert";
 import {stringToNumber} from "../general-purpose-parsers/string-to-number";
 
-import type {PerseusImageBackground} from "@khanacademy/perseus-core";
 import type {Parser} from "../parser-types";
+import type {PerseusImageBackground} from "@khanacademy/perseus-core";
 
 function emptyToZero(x: string | number): string | number {
     return x === "" ? 0 : x;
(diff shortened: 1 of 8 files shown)


handeyeco:
I haven't looked at this yet, but my first reaction is that I think this will need to live outside of `perseus` if we ever want to use it on the server. My current work is moving logic we'll need on the server out of `perseus` and this seems like something we'll want in the medium-to-long term.

My guess is we'll want this in `perseus-core`. CC @jeremywiebe

benchristel:
@handeyeco do you mean that all the parsing code will have to live in `perseus-core`? Let me double-check that it doesn't have any runtime dependencies on `perseus`...

handeyeco:
`perseus-types.ts` in `perseus` is moving to `data-schema.ts` in `perseus-core`: https://github.com/Khan/perseus/pull/2086

We're going to have two registries, one on the BE (scoring, validation, upgrading) and one on the FE (rendering)

packages/perseus/src/util/parse-perseus-json/perseus-parsers/perseus-image-background.ts
```diff
@@ -10,8 +10,8 @@ import {
 import {convert} from "../general-purpose-parsers/convert";
 import {stringToNumber} from "../general-purpose-parsers/string-to-number";
 
+import type {PerseusImageBackground} from "../../../perseus-types";
 import type {Parser} from "../parser-types";
-import type {PerseusImageBackground} from "@khanacademy/perseus";
```
jeremywiebe, line 14: I've seen "self" imports like this sneak into our codebase before. I _think_ they work, but it'd be nice if we could lint for them and prevent them.

jeremywiebe (approved):

packages/perseus/src/util/parse-perseus-json/perseus-parsers/perseus-image-background.ts
handeyeco: I've noticed this too. I thought we have a lint rule, but maybe it's not working for TS?

PR #2088, opened 2025-01-09
Start moving some utils out of Perseus

handeyeco:
## Summary:
One of a few PRs where I'll be moving utils we need on the server out of `perseus` and into `perseus-core`

Issue: LEMS-2737


testing/object-utils.ts
```diff
@@ -1,9 +0,0 @@
-// Performs a deep copy of the given object. If there are cycles in the object
```
jeremywiebe, line 1: I'm not certain we want to move this into `Perseus-core`. It's only used in tests. Also, if we want to do this move, I'd want to see if we can consolidate all the different ways we clone (today, we have this `clone`, then we have [`deepClone()`](https://github.com/Khan/perseus/blob/341d316aa8727ebb9e7fde28fc4e2d8779aa3e82/packages/perseus/src/util.ts#L718), another [`clone()`](https://github.com/Khan/perseus/blob/f5a2cf521291180dbbd448adc97700f7c52c8b50/packages/perseus-linter/src/__tests__/tree-transformer.test.ts#L6), and [`_.clone()`](https://github.com/Khan/perseus/blob/368e222a6577dff38143d1584d6773129e8abbd7/packages/kas/src/nodes.ts#L573): all of which look to do very similar things).

jeremywiebe (approved):

testing/object-utils.ts
handeyeco: This was a total mistake from me being on auto-pilot. Thanks for catching it.

handeyeco:
The parent pull-request ([#2086](https://github.com/Khan/perseus/pull/2086)) has been merged into `main`, but this branch (`LEMS-2737-2/utils`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

PR #2089, opened 2025-01-09
Move most of `perseus - util.ts` into `perseus-core - util.ts`

handeyeco:
## Summary:
I couldn't move the whole of `util.ts` from `perseus` to `perseus-core` because of some of its dependencies, but I moved as much as I could. Most of this is just changing imports.

Issue: LEMS-2737


packages/perseus-core/src/utils/util.ts
```diff
@@ -0,0 +1,688 @@
...
+    strongEncodeURIComponent,
+    touchHandlers,
+    resetTouchHandlers,
+    extractPointerLocation,
+    supportsPassiveEvents,
+    textarea,
+    unescapeMathMode,
+    random,
+    deepClone,
+} as const;
```
jeremywiebe, line 686: What's your appetite like for switching these to simple named exports instead of bundling them in an object first? We could still export them as a `Util` namespace in the top-level `index.ts` so the imports don't change...

packages/perseus-core/src/utils/util.ts
```diff
@@ -0,0 +1,688 @@
...
+type QueryParams = {
+    [param: string]: string;
+};
+
+export type Position = {
+    top: number;
+    left: number;
+};
+
+type TouchHandlers = {
```
jeremywiebe, line 40: Having references to UI concepts in `perseus-score` raises a warning flag for me.  What do you think of leaving anything to do with UI (touch, event handling, anything touching `window`).

packages/perseus-core/src/utils/util.ts
```diff
@@ -0,0 +1,688 @@
...
+/**
+ * Query String Parser
+ *
+ * Original from:
+ * http://stackoverflow.com/questions/901115/get-querystring-values-in-javascript/2880929#2880929
+ */
+function parseQueryString(query: string): QueryParams {
+    // TODO(jangmi, CP-3340): Use withLocation to access SSR safe location.
+    // eslint-disable-next-line no-restricted-syntax
+    query = query || window.location.search.substring(1);
```
jeremywiebe, line 443: I think this is safe to move, but we should probably drop the `|| window...` bit as that won't be safe in a non-browser env.

packages/perseus-core/src/utils/util.ts
```diff
@@ -0,0 +1,688 @@
...
+ */
+function moveCursor(textarea: HTMLTextAreaElement, pos: number): void {
+    textarea.selectionStart = pos;
+    textarea.selectionEnd = pos;
+}
+
+const textarea = {
+    getWordBeforeCursor,
+    moveCursor,
+} as const;
```
jeremywiebe, line 624: This all feels like it should stay in `perseus`.

jeremywiebe (approved):
Looks good. I'd just suggest moving some UI functions back into Perseus if we can.

handeyeco:
The parent pull-request ([#2088](https://github.com/Khan/perseus/pull/2088)) has been merged into `main`, but this branch (`LEMS-2737-3/utils2`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

PR #2092, opened 2025-01-10
Create helper to build public widget options for categorizer

Myranae:
## Summary:
This adds a function that takes categorizer's full widget options and filters out answer data. It also adds this function to the widget's widget export and adds a test confirming the function does what we expect.

Issue: LEMS-2756

## Test plan:
- Confirm all checks pass
- Confirm categorizer still works as expected


packages/perseus/src/types.ts
```diff
@@ -542,6 +543,11 @@ export type WidgetScorerFunction = (
     locale?: string,
 ) => PerseusScore;
 
+/**
+ * A union type of all the functions to get the public options for each widget.
+ */
+export type PublicWidgetOptionsFunctions = typeof getCategorizerPublicOptions;
```
Myranae, line 549: This currently only has categorizer's public options function, but as I create more, I'll keep adding to this. I wanted to do some kind of generic here, but that spiraled a bit, so staying with what seemed to work,

packages/perseus/src/types.ts
```diff
@@ -15,6 +15,7 @@ import type {
     UserInputMap,
 } from "./validation.types";
 import type {WidgetPromptJSON} from "./widget-ai-utils/prompt-types";
+import type getCategorizerPublicOptions from "./widgets/categorizer/categorizer.util";
```
Myranae, line 18: Not sure if it's the right move to have these in their own file and, if that is the right move, wasn't sure about the filename being `categorizer.util.ts`. If anyone has thoughts on where the function and type should be, let me know :)

packages/perseus/src/widgets/categorizer/categorizer.util.ts
```diff
@@ -0,0 +1,25 @@
...
+type CategorizerPublicWidgetOptions = {
+    // a list of items to categorize. e.g. ["banana", "yellow", "apple", "purple", "shirt"]
+    items: PerseusCategorizerWidgetOptions["items"];
+    // a list of categories. e.g. ["fruits", "colors", "clothing"]
+    categories: PerseusCategorizerWidgetOptions["categories"];
+    // Whether the items should be randomized
+    randomizeItems: PerseusCategorizerWidgetOptions["randomizeItems"];
+    // Whether this widget is displayed with the results and immutable
+    static: PerseusCategorizerWidgetOptions["static"];
+};
```
Myranae, line 12: I wasn't 100% sure on what should be kept here. I took out `values` since that is where the answer is. I also took out `highlightLint` and `linterContext` as those seemed related to the editors (and editors will get the full widget options). Let me know if those should go back though.

.changeset/moody-numbers-move.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
```
Myranae, line 2: I chose patch because I'm just adding content that isn't being used yet, but I am modifying a widget's widgetExport object, so let me know if that means it should be a minor change.

packages/perseus/src/widgets.ts
```diff
@@ -141,6 +142,12 @@ export const getWidgetScorer = (name: string): WidgetScorerFunction | null => {
     return widgets[name]?.scorer ?? null;
 };
 
+export const getWidgetPublicOptionsFunction = (
+    name: string,
+): PublicWidgetOptionsFunctions | null => {
+    return widgets[name]?.getPublicOptions ?? null;
+};
+
```
Myranae, line 150: We talked about adding this, so I threw it in, but not sure if there's somewhere I need to add it yet. I think this will be used on the assessment editor side, but not positive. Although I know Matthew is splitting up the widget exports so maybe we don't need this right now.

packages/perseus/src/widgets/categorizer/categorizer.util.test.ts
```diff
@@ -0,0 +1,32 @@
+import getCategorizerPublicOptions from "./categorizer.util";
+
+import type {PerseusCategorizerWidgetOptions} from "../../perseus-types";
+
+describe("getCategorizerPublicOptions", () => {
```
Myranae, line 5: Would love to add more tests if anyone can think of other useful things to check :)

packages/perseus/src/types.ts
```diff
@@ -589,6 +595,12 @@ export type WidgetExports<
      */
     scorer?: WidgetScorerFunction;
 
+    /**
+     * A function that provides a public version of the widget options that can
+     * be shared with the client.
+     */
+    getPublicOptions?: PublicWidgetOptionsFunctions;
```
Myranae, line 602: Would this be more clear if it was `getPublicWidgetOptions`? I think I was trying to make it a bit shorter, but after looking at it again not sure if it's clear enough.

packages/perseus/src/widgets.ts
```diff
@@ -141,6 +142,12 @@ export const getWidgetScorer = (name: string): WidgetScorerFunction | null => {
     return widgets[name]?.scorer ?? null;
 };
 
+export const getWidgetPublicOptionsFunction = (
+    name: string,
+): PublicWidgetOptionsFunctions | null => {
+    return widgets[name]?.getPublicOptions ?? null;
```
handeyeco, line 148: I think it should probably return an identity function if it can't find a splitter:

```
    return widgets[name]?.getPublicOptions ?? i => i;
```

That way at the call site we don't have to check if the function exists:

```
const splitStuff = getWidgetPublicOptionsFunction("expression")(stuffToSplit)
```

handeyeco (commented):
I think this looks good, but I'm going to leave it to Jeremy to approve since the two of you have been looking at this more than I have and we're still thinking about the final design.

.changeset/moody-numbers-move.md
Myranae: Updated!

packages/perseus/src/types.ts
Myranae: It feels a little incorrect to call it "splitCategorizer" as we're not really splitting the whole widget in the same way we would score or validate it, so I changed it to "splitCategorizerWidgetOptions" and updated the rest accordingly. Also didn't want it to be "splitter" similar to how we do "scorer". 😅

packages/perseus/src/types.ts
Myranae: This was updated to `splitWidgetOptions`.

packages/perseus/src/widgets.ts
Myranae: Added!

packages/perseus/src/widgets/categorizer/categorizer.util.ts
jeremywiebe: I think you are on the right track. We want to remove everything that's in the associated `ScoringData` type if we can. I think we've found some widgets where the widget depends on some bits of scoring data to function and so we'll have to untangle those as we get to them. 

Removing `highlightLint` and `linterContext` are a good move. In fact, I suspect that those are not actually something we should be saving as part of the data, but rather props that are used in the editor at runtime (not something you need to fix now).

.changeset/moody-numbers-move.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/perseus": minor
+---
+
+Create a function to filter out answer data from categorizer's widget options and add to its widget export
```
jeremywiebe, line 5: Since this is where we introduce this concept, I'm going to suggest a few small wording improvements. Feel free to edit further. :) 

```suggestion
Introduce a widget export function to filter out scoring data from widget options. Implement this function for the categorizer widget.
```

packages/perseus/src/types.ts
```diff
@@ -542,6 +543,11 @@ export type WidgetScorerFunction = (
     locale?: string,
 ) => PerseusScore;
 
+/**
+ * A union type of all the functions to get the public options for each widget.
+ */
+export type SplitWidgetOptionsFunctions = typeof splitCategorizerWidgetOptions;
```
jeremywiebe, line 549: Naming: as we're thinking more about what these functions will do, I think they're more about removing than splitting. I think in recent discussions we considered that we might only need to provide a widget options with only public data and then provide the full item for scoring and other server-side work. 

What do you think of simply naming these functions `removePrivateWidgetOptions`, and so this type would be `removePrivateWidgetOptionsFunction` (I think we want this to be singular because it is a union of all the functions, but it represents a single instance of any of them - sorry, it's nitpicky but plural, to me, speaks of an array).

packages/perseus/src/types.ts
```diff
@@ -589,6 +595,12 @@ export type WidgetExports<
      */
     scorer?: WidgetScorerFunction;
 
+    /**
+     * A function that provides a public version of the widget options that can
+     * be shared with the client.
+     */
+    splitWidgetOptions?: SplitWidgetOptionsFunctions;
```
jeremywiebe, line 602: I think the doc string here reinforces my point above about needing to rename this concept/field. What do you think of this suggested naming?

```suggestion
    /**
     * A function that provides a public version of the widget options that can
     * be shared with the client.
     */
    removePublicWidgetOptions?: RemovePublicWidgetOptionsFunction;
```

packages/perseus/src/widgets/categorizer/categorizer.util.test.ts
```diff
@@ -0,0 +1,32 @@
...
+                paths: ["paths"],
+                stack: ["stack"],
+            },
+        };
+
+        const splitOptions = splitCategorizerWidgetOptions(
+            categorizerTestWidgetOptions,
+        );
+
+        expect(splitOptions).toEqual({
```
jeremywiebe, line 25: I can't think of more tests to write here. We just want to make sure that the private data is not in here, and `.toEqual()` does that.

packages/perseus/src/widgets/categorizer/categorizer.util.ts
```diff
@@ -0,0 +1,25 @@
...
+type CategorizerSplitWidgetOptions = {
+    // a list of items to categorize. e.g. ["banana", "yellow", "apple", "purple", "shirt"]
+    items: PerseusCategorizerWidgetOptions["items"];
+    // a list of categories. e.g. ["fruits", "colors", "clothing"]
+    categories: PerseusCategorizerWidgetOptions["categories"];
+    // Whether the items should be randomized
+    randomizeItems: PerseusCategorizerWidgetOptions["randomizeItems"];
+    // Whether this widget is displayed with the results and immutable
+    static: PerseusCategorizerWidgetOptions["static"];
+};
```
jeremywiebe, line 12: I'm of two minds on documenting these option fields. On the one hand, it might be useful for folks coming to this type to see what it does, but I think the risk of this comment getting out of sync with the main field in `PerseusCategorizerWidgetOptions` is greater. 

So I think I'd recommend removing all field documentation in this type and maybe adding a JSDoc string to the entire structure pointing back to the full options type for full details. 

Lastly, I think the `Split` term should be `Public` here also (so `CategorizerPublicWidgetOptions`). 

```suggestion
/** 
 * The `categorizer` widget's options that are public (ie. safe to 
 * pass to the browser without divulging the answer).
 * 
 * @see {@link PerseusCategorizerWidgetOptions} for full documentation of each option.
 */ 
type CategorizerPublicWidgetOptions = {
    items: PerseusCategorizerWidgetOptions["items"];
    categories: PerseusCategorizerWidgetOptions["categories"];
    randomizeItems: PerseusCategorizerWidgetOptions["randomizeItems"];
    static: PerseusCategorizerWidgetOptions["static"];
};
```

jeremywiebe (approved):
I like the direction this is going!  I've left some comments but I think once you feel they're addressed, we're ready to start into other widgets. 

One thing I'd suggest to see if you have the type correct: try to use the new "public" widget options type as the source for the `categorizer` widget's render props. This will tell you whether you have added all the keys that it needs, or if you've removed too much.

packages/perseus/src/types.ts
Myranae: For sure. It started out as `PublicWidgetOptionsFunctions` but changed it to "split" after Matthew's feedback. I kind of think it should be in sync with the rest of the naming for this update, which would mean all the other aspects would also change to `removePrivateWidgetOptions`. I do agree that it's not really splitting anymore though. I'm totally cool changing it to `RemovePrivateWidgetOptionsFunction`,  but then I might want to change others like `removePrivateWidgetOptions` as the name on the export and `removeCategorizerPrivateWidgetOptions`. What do you think?

packages/perseus/src/types.ts
Myranae: Hmm, I think you mean `removePrivateWidgetOptions`, right? It started out as `getPublicWidgetOptions` before it was changed to "split", which is why I use that phrasing in the docstring.

jeremywiebe: 🤦‍♂️ Yes! "private" not "public"! :)

packages/perseus/src/types.ts
jeremywiebe: I agree.  I think we should be as consistent as possible.

PR #2097, opened 2025-01-14
[SR][Locked Figures] Give all locked figures "img" role

nishasy:
## Summary:
Right now, when traversing through locked figures with a
screen reader, it reads out the label and then identifies
the figure as "empty group." This is an extremely
unhelpful description.

To give a more accurate description of the function the
locked figures are serving, we need to give them a
role of "img." This way, the screen reader will read
the aria label and then say "image" after.

Issue: none

## Test plan:
Storybook
- http://localhost:6006/iframe.html?globals=&args=&id=perseuseditor-widgets-interactive-graph--mafs-with-locked-figures&viewMode=story
- Go through this story before and after the change. It should say image
  instead of empty group now.

### Before

https://github.com/user-attachments/assets/245cd639-a64b-4abf-bcc3-9354a6381c8d

### After

https://github.com/user-attachments/assets/68c67c9b-160e-4949-a123-209e28759b4e


SonicScrewdriver (approved):
Nice, thank you! :)

PR #2106, opened 2025-01-14
Move Radio scorer/validator

handeyeco:
## Summary:
Move Radio scorer/validator to `perseus-score`

Issue: LEMS-2737

## Test plan:


benchristel (approved):
LGTM!

packages/perseus/src/strings.ts
```diff
@@ -700,17 +700,24 @@ export const mockStrings: PerseusStrings = {
...
-    [ErrorCodes.WRONG_LETTER_ERROR]: strings.WRONG_LETTER_ERROR,
-    [ErrorCodes.MULTIPLICATION_SIGN_ERROR]: strings.MULTIPLICATION_SIGN_ERROR,
-    [ErrorCodes.INVALID_SELECTION]: strings.invalidSelection,
+// This type helps us make sure all error codes are mapped to strings
+type ErrorStringMap = {
+    [K in keyof typeof ErrorCodes]: string;
 };
+
+const errorToString: ErrorStringMap = {
+    MISSING_PERCENT_ERROR: strings.MISSING_PERCENT_ERROR as string,
```
jeremywiebe, line 709: Instead of being forced to `as string` each of these, I think we can fix this more cleanly by moving the type annotation for `strings` from:

```typescript
export const strings: {
    [key in keyof PerseusStrings]:
        | string
        | {context?: string; message: string}
        | {context?: string; one: string; other: string};
} = {
    closeKeypad: "close math keypad",
    ...
};
```

to:

```typescript 
export const strings = {
    closeKeypad: "close math keypad",
    ...
} satisfies {
    [key in keyof PerseusStrings]:
        | string
        | {context?: string; message: string}
        | {context?: string; one: string; other: string};
};
```

Using `satisfies` instead of `: mytype =` means that we allow TypeScript to infer the type of each key in `strings`, but the entire structure must conform to the satisfied type.  So TypeScript can now infer that `strings` has a `closeKeypad` key with a string value instead of a value that is one of `string` `{context..}` etc.

jeremywiebe (approved):

PR #2108, opened 2025-01-14
Move Numeric Input's UI related code to a functional component 

SonicScrewdriver:
## Summary:
We have a lot of old patterns in Perseus, and we would like to start to updating to more modern methods. This PR updates the Numeric Input logic in the following ways:

1. Moves all UI Related Numeric Input logic to a functional component, and creates a new `numeric-input.class.tsx` file for housing the static / class methods. 
2. Removes all string refs related to Numeric Input in both the InputWithExamples, SimpleKeypadInput, and Tooltip components
3. Adds a little more specificity to method parameters 

Issue: LEMS-2443

## Test plan:
- Manual Testing 
- Automated Tests 
- Landing onto a feature branch for future QA Regression pass


packages/perseus/src/widgets/numeric-input/numeric-input.tsx
```diff
@@ -1,419 +1,123 @@
...
-            labelText = this.context.strings.yourAnswerLabel;
-        }
+        const handleFocus = (): void => {
+            props.onFocus([]);
+            setIsFocused(true);
+        };
 
-        // To right align a custom keypad we need to wrap it.
-        const maybeRightAlignKeypadInput = (
-            keypadInput: React.ReactElement<
```
SonicScrewdriver, line 237: I felt that this was needlessly complicating the code readability, so I've opted to simply add the div around both left/right aligned inputs that gains a class if it needs to be right aligned.

packages/perseus/src/widgets/numeric-input/numeric-input.stories.tsx
```diff
@@ -3,7 +3,7 @@ import * as React from "react";
 
 import {RendererWithDebugUI} from "../../../../../testing/renderer-with-debug-ui";
 
-import {NumericInput} from "./numeric-input";
+import {NumericInput} from "./numeric-input.class";
```
mark-fitzgerald, line 6: Should these be pulling in from `./index`? Eventually when we remove the class files, we would need to update all of these imports. Just wondering out loud regarding preference.

packages/perseus/src/widgets/numeric-input/numeric-input.tsx
```diff
@@ -1,419 +1,123 @@
...
+                textAlign: props.rightAlign ? "right" : "left",
+                width: props.size === "small" ? 40 : 80,
             },
         });
 
+        // (mobile-only) If the custom keypad is enabled, use the SimpleKeypadInput component
+        if (props.apiOptions.customKeypad) {
+            const alignmentClass = props.rightAlign
+                ? "perseus-input-right-align"
+                : "";
```
mark-fitzgerald, line 92: Nit: Instead of an empty class, perhaps use `null` so that the `class` attribute doesn't show on left-aligned input?

packages/perseus/src/widgets/numeric-input/utils.ts
```diff
@@ -0,0 +1,71 @@
+import _ from "underscore";
+
+import type {PerseusStrings} from "../../strings";
+import type {PerseusNumericInputAnswerForm} from "@khanacademy/perseus-core";
+
```
mark-fitzgerald, line 5: Should there be unit tests for these utility functions?

packages/perseus/src/widgets/numeric-input/numeric-input.class.tsx
```diff
@@ -0,0 +1,268 @@
...
+
+    /**
+     * Returns the JSON representation of the prompt for this widget.
+     * This is used by the AI to determine the prompt for the widget.
+     */
+    getPromptJSON(): NumericInputPromptJSON {
+        return _getPromptJSON(this.props, this.getUserInput());
+    }
+
+    render(): React.ReactNode {
```
mark-fitzgerald, line 147: I wonder if we should add some sort of boilerplate comment that explains that we are migrating from class-based to function-based components.

mark-fitzgerald (approved):
I can feel the "clean"!

packages/perseus/src/components/tooltip.tsx
```diff
@@ -356,8 +356,6 @@ class Tooltip extends React.Component<Props, State> {
                     {/* The contents of the tooltip */}
                     <div
                         className={this.props.className}
-                        // eslint-disable-next-line react/no-string-refs
-                        ref="tooltipContent"
                         style={{
```
SonicScrewdriver, line 359: This ref was not being used anywhere.

packages/perseus/src/widgets/numeric-input/utils.ts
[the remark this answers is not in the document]
mark-fitzgerald: I think "utils" is fine.

PR #2113, opened 2025-01-15
Move scorers: Plotter and Sorter

handeyeco:
## Summary:
I moved these together because they both depended on `Utils.deepEq` which I moved to `_.isEqual` ([which is deep equality](https://underscorejs.org/#isEqual)).

Issue: LEMS-2737

## Test plan:
Nothing should change


packages/perseus-score/src/widgets/plotter/score-plotter.ts
```diff
@@ -20,7 +18,7 @@ function scorePlotter(
     }
     return {
         type: "points",
-        earned: deepEq(userInput, scoringData.correct) ? 1 : 0,
+        earned: _.isEqual(userInput, scoringData.correct) ? 1 : 0,
```
benchristel, line 21: I vaguely remember we were talking about trying to get rid of underscore at some point. Was that only for functions like `_.map` that have native equivalents now?

benchristel (approved):
LGTM! Left one question inline about underscore.

packages/perseus-score/src/widgets/plotter/score-plotter.ts
jeremywiebe: Where native equivalents exist, I think we should switch away from `underscore`. For cases like this, it'd be nice if we had a single deep equality check. I _think_ we actually want to revert this change because our "in-house" `deepEq` uses an approximate equals check for numeric values.

```suggestion
        earned: deepEq(userInput, scoringData.correct) ? 1 : 0,
```

image

handeyeco: Yeah, that's right. I need to go back and undo the change to underscore. I think what I'll do is add another PR to the top because I also renamed `eq` and `deepEq` to make it clearer that they're approximate.

handeyeco: https://github.com/Khan/perseus/pull/2125

handeyeco: Just to add another thought.

> Was that only for functions like _.map that have native equivalents now?

I think we should absolutely not use jQuery/Underscore for things that have native equivalents (`map`, `each`, etc).

I also don't think we should be reinventing complex logic that's not native and provided through something like Underscore.

In this case it wasn't clear to me that these were approximate equality checks, so I didn't see the need to reinvent the wheel when we depend on Underscore and Underscore has `isEqual`. Now that I know that these are fairly specific helpers, I know now we should have kept them. I renamed them to `approximateDeepEqual` and `approximateEqual` to prevent people from making my mistake in the future.

PR #2114, opened 2025-01-15
Move scorer: Orderer

handeyeco:
## Summary:

Move scoring logic for Orderer. I updated the tests a little bit so as to not rely on `orderer.testdata.ts` which I didn't want to split and which I think is overkill for these tiny pure functions.

Issue: LEMS-2737

## Test plan:
Nothing should change, moving code


benchristel (approved):
LGTM!

PR #2115, opened 2025-01-15
Move scorerer: LabelImage

handeyeco:
## Summary:
Move LabelImage scoring logic

Issue: LEMS-2737

## Test plan:


packages/perseus-score/src/widgets/label-image/score-label-image.ts
```diff
@@ -13,7 +13,7 @@ type InteractiveMarkerScore = {
     isCorrect: boolean;
 };
 
-export function scoreMarker(
+export function labelImageScoreMarker(
```
benchristel, line 16: Why the rename? Could this function be called `scoreLabelImageMarker` or `scoreMarkerOnLabelImage`? That reads more like English.

benchristel (approved):
LGTM! Left an optional suggestion inline about naming.

packages/perseus-score/src/widgets/label-image/score-label-image.ts
handeyeco: It's getting exported from `perseus-score` and imported into `perseus` now. Since `scoreMarker` isn't very descriptive and it's exported just for one place in `perseus` I thought I'd give it a more descriptive name.

You're right that I picked the wrong name. I was reading it as a noun (it's not "a score marker"), but it was a verb (it's something that "scores markers").

PR #2116, opened 2025-01-15
Move scoring logic: Matrix

handeyeco:
## Summary:

Move Matrix scoring logic. This one required a few other changes:

- It uses a string, so that needed to be converted to an error code
- Matrix and MatrixEditor had a duplicate function that I consolidated

Issue: LEMS-2737

## Test plan:
Nothing should change, just code shoveling


benchristel (approved):
LGTM!

packages/perseus/src/widgets/matrix/matrix.tsx
```diff
@@ -79,30 +79,6 @@ const getRefForPath = function (path: FocusPath) {
     return "answer" + row + "," + column;
 };
 
-export function getMatrixSize(matrix: ReadonlyArray<ReadonlyArray<number>>) {
```
jeremywiebe, line 82: Ouch, more duplication. Thanks for cleaning it up.

jeremywiebe (approved):

PR #2118, opened 2025-01-16
Move scoring logic: Expression

handeyeco:
## Summary:
This is one of the trickier ones:

- Expression scoring logic depended on `Log` which is attached to `getDependencies` which is attached to React stuff. Jeremy and I met about this and decided to use `throw` instead which we can catch/log higher up in the code. We think it's unlikely we'll reach this code on the server anyway and KhanAnswerTypes also throws.
- I don't want to change the type for the scorers right now which is: `(userInput: UserInput, rubric: Rubric, string?: PerseusStrings, locale?: string) => PerseusScore;` but I do need to remove `PerseusStrings` AND I need to keep `locale`. So I just made `string: any` and gave it a dummy object in tests.
- Consolidated two declarations of `getDecimalSeparator` which required some subtle logic changes.

Issue: LEMS-2737

## Test plan:
Nothing should change, potential areas that could be affected:
- Expression widget decimal separator based on locale
- Logging/Throwing for unparsable data


packages/perseus-score/src/widgets/expression/score-expression.ts
```diff
@@ -38,7 +39,8 @@ import type {
 function scoreExpression(
     userInput: PerseusExpressionUserInput,
     rubric: PerseusExpressionRubric,
-    strings: PerseusStrings,
+    // TODO: remove strings as a param for scorers
+    strings: any,
```
benchristel, line 43: Since `strings` isn't used, we could give it type `unknown` instead of `any`.
```suggestion
    strings: unknown,
```

benchristel (approved):
LGTM!

jeremywiebe (approved):

PR #2119, opened 2025-01-16
Move scorer: Grapher

handeyeco:
## Summary:
This was a messier one, because some of the scoring logic was also mapped to React components. I tried to comment on the highlights, but most of the changes are just splitting Grapher utils and moving shared stuff to `perseus-core`.

Issue: LEMS-2737

## Test plan:
Everything should continue to function as usual. Grapher is obviously the major affected widget.


packages/perseus-score/src/widgets/grapher/score-grapher.ts
```diff
@@ -16,7 +14,7 @@ function getCoefficientsByType(
         return undefined;
     }
     if (data.type === "exponential" || data.type === "logarithm") {
-        const grader = functionForType(data.type);
+        const grader = GrapherUtil.functionForType(data.type);
```
handeyeco, line 17: `functionForType` is the reason for the size of the PR. Basically it was designed in a pattern similar to WidgetExports where a bunch of functionality and some React classes were thrown together and accessed via a central getter using the `type`.

This explains two things:

1. I had to move that getter and all associated code out of `perseus` because `perseus-score` can't import from `perseus`. `perseus` and `perseus-editor` still needed access to this logic, so I moved it to `perseus-core`.
2. `perseus-core` can't depend on React/JSX, so I had to split the data (similar to what we're doing with the scorers and WidgetExports); but I still needed access to the Movable component in Grapher, so I made the `movableToGraphieMovable` mapper.

packages/perseus/src/widgets/grapher/util.tsx
```diff
@@ -1,661 +1,35 @@
 import {point as kpoint} from "@khanacademy/kmath";
+import {GrapherUtil} from "@khanacademy/perseus-core";
```
handeyeco, line 2: In #2089 (which will likely be closed without merging) I felt like I went too far moving utilities out of `perseus`. So this time I _only_ moved what I had to move to get the Grapher scorer onto `perseus-score`.

packages/perseus-core/src/utils/grapher-types.ts
```diff
@@ -1,4 +1,4 @@
-import type {Coord} from "@khanacademy/perseus";
+import type {Coord} from "../data-schema";
```
handeyeco, line 1: I consolidated some `Coord` types.

packages/perseus-core/src/utils/grapher-util.ts
```diff
@@ -0,0 +1,644 @@
...
+    AbsoluteValueType,
+    Coords,
+} from "./grapher-types";
+import type {Coord} from "../data-schema";
+
+export const MOVABLES = {
+    PLOT: "PLOT",
+    PARABOLA: "PARABOLA",
+    SINUSOID: "SINUSOID",
+};
```
handeyeco, line 19: When I moved this, I couldn't move `Graphie` which was used by these objects to map `Graphie` components to their `Movable` properties. Instead I made this enough, changed `Movable: Graphie.Plot` to `movable: MOVABLES.PLOT`, and mapped them back to Graphie widgets using `movableToGraphieMovable`.

packages/perseus-core/src/utils/grapher-util.ts
```diff
@@ -0,0 +1,644 @@
...
+}
+
+const PlotDefaults = {
+    areEqual: function (
+        coeffs1: ReadonlyArray<number>,
+        coeffs2: ReadonlyArray<number>,
+    ): boolean {
+        return _.isEqual(coeffs1, coeffs2);
+    },
+    movable: MOVABLES.PLOT,
```
handeyeco, line 100: Like this one.

packages/perseus/src/widgets/grapher/util.tsx
```diff
@@ -1,661 +1,35 @@
...
-    url: "https://ka-perseus-graphie.s3.amazonaws.com/e23d36e6fc29ee37174e92c9daba2a66677128ab.png",
-
-    defaultCoords: [
-        [0.5, 0.5],
-        [0.75, 0.75],
-    ],
+type MovableMap = {
+    [K in keyof typeof GrapherUtil.MOVABLES]: any;
+};
+export const movableToGraphieMovable: MovableMap = {
```
benchristel, line 16: I was slightly confused by the name of this object. Could it be called something like `movableTypeToComponent` or `movableToComponentMap`? I guess the `movable` vs. `GraphieMovable` distinction confused me.

packages/perseus-editor/src/widgets/grapher-editor.tsx
```diff
@@ -7,6 +7,7 @@ import {
     containerSizeClass,
     getInteractiveBoxFromSizeClass,
 } from "@khanacademy/perseus";
+import {GrapherUtil as CoreGrapherUtil} from "@khanacademy/perseus-core";
```
benchristel, line 10: Why do these "util" functions need to be in core? Previously, I think `perseus-core` didn't have any widget-specific code. Could the utils move to `perseus-score` instead? I guess I'm confused about the purposes of `perseus-core` vs. `perseus-score`.

jeremywiebe (approved):

PR #2120, opened 2025-01-16
Move scorer: Interactive Graph

handeyeco:
## Summary:
This one was the most painful. First off, IG is a beast. Secondly, IG is in the process of migrating from the old to the new and in the meantime there are a lot of duplicate function/variable/type declarations that are subtly different that made it really hard to do find/replaces.

This required:

- Moving logic out of `perseus`:
  - Mathy logic to `kmath`
  - Shared functions to `perseus-core`
  - Scoring logic to `perseus-score`

Issue: LEMS-2737

## Test plan:
Nothing should change, IG should continue to be scorable. Just moving code.


packages/kmath/src/angles.ts
```diff
@@ -1,22 +1,21 @@
-import {clockwise} from "../../../util/geometry";
+// This file contains helper functions for working with angles.
 
-import type {Coord} from "@khanacademy/perseus";
-import type {vec} from "mafs";
```
handeyeco, line 4: `angles` needs to be moved out of `perseus` for scoring reasons. `maf` can't be in part of the process because it depends on React.

packages/kmath/src/coefficients.ts
```diff
@@ -0,0 +1,62 @@
...
+export type NamedSineCoefficient = {
+    amplitude: number;
+    angularFrequency: number;
+    phase: number;
+    verticalOffset: number;
+};
+
+// TODO: there's another, very similar getSinusoidCoefficients function
+// they should probably be merged
+export function getSinusoidCoefficients(
```
handeyeco, line 13: There's two of each of these functions. These are the ones used for scoring.

packages/kmath/src/geometry.ts
```diff
@@ -2,14 +2,16 @@
...
+} from "@khanacademy/perseus-core";
 import _ from "underscore";
 
-import Util from "../util";
-
-import type {Coord, Line} from "../interactive2/types";
+import {number as knumber, point as kpoint, sum} from "@khanacademy/kmath";
 
-const {eq, deepEq} = Util;
+type Line = [Coord, Coord];
```
handeyeco, line 14: There's multiple `type Line` declarations, at least 1 for 2D lines and 1 for nD lines.

packages/kmath/src/geometry.ts
```diff
@@ -21,12 +23,9 @@ export type SineCoefficient = [
     number, // verticalOffset
 ];
 
-// a, b, c
-export type QuadraticCoefficient = [number, number, number];
-
 // Given a number, return whether it is positive (1), negative (-1), or zero (0)
 export function sign(val: number): 0 | 1 | -1 {
-    if (eq(val, 0)) {
+    if (approximateEqual(val, 0)) {
```
handeyeco, line 28: I found out (unfortunately pretty late) that `eq` and `deepEq` aren't just standard equality checks, they're APPROXIMATE equality checks 😬 I thought it was worth being clear about that distinction.

packages/perseus/src/util.ts
```diff
@@ -761,7 +688,6 @@ const Util = {
     textarea,
     unescapeMathMode,
     random,
-    deepClone,
```
handeyeco, line 764: These three were moved to `perseus-core` since we needed them within the scorer.

packages/perseus/src/widgets/interactive-graphs/graphs/sinusoid.tsx
```diff
@@ -97,7 +91,7 @@ export const computeSine = function (
 
 export const getSinusoidCoefficients = (
     coords: ReadonlyArray<Coord>,
-): SineCoefficient | undefined => {
+): NamedSineCoefficient | undefined => {
```
handeyeco, line 94: There were two `SineCoefficient` types and they were different, so I renamed one of them.

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.test.ts
```diff
@@ -1,6 +1,6 @@
 import invariant from "tiny-invariant";
 
-import {getClockwiseAngle} from "../math/angles";
+import {getClockwiseAngle} from "../../../../../kmath/src/angles";
```
jeremywiebe, line 3: Can you fix this to import `@khanademy/kmath`?

jeremywiebe (approved):
Looks good, just need to fix that one import.

PR #2121, opened 2025-01-16
Modernization and Migration of InputWithExamples to NumericInput folder

SonicScrewdriver:
## Summary:
This PR is part of the Numeric Input Project work. It is being landed onto the `feature/numeric-dx-refactor` branch. 

This PR contains the following changes:
1. Moves the InputWithExamples component to the NumericInput folder
2. Modernizes InputWithExamples to be a functional component 
3. Addition of some comments 

Video Example of Snapshot Testing:

https://github.com/user-attachments/assets/ca917778-50b0-46d2-89d8-dad95d1dadf2

Issue: LEMS-2785

## Test plan:
- Ensure all tests pass 
- Manual testing with PR Snapshot in upstream consumer 
- Landing onto feature branch that will see full QA regression pass before deployment


packages/perseus/src/widgets/numeric-input/input-with-examples.tsx
```diff
@@ -0,0 +1,194 @@
...
+        linterContext = PerseusLinter.linterContextDefault,
+        className = "",
+    } = props;
+
+    const context = React.useContext(PerseusI18nContext);
+    const inputRef = React.useRef<TextInput>(null);
+    const [state, setState] = React.useState<State>({
+        focused: false,
+        showExamples: false,
+    });
```
SonicScrewdriver, line 72: These two variables are always set together, so I've opted to keep this pairing after the shift to the functional component. I wasn't sure of a good name to explain both of them so I went "generic", but I could easily see an argument to rename this to something like "setFocusStates".

SonicScrewdriver: ...Wait. If these are always set the exact same thing in multiple places ... why do we even need both? I bet we could just use "focused".

SonicScrewdriver: I've updated this to just use a single `inputFocused` state.

packages/perseus/src/widgets/numeric-input/input-with-examples.tsx
SonicScrewdriver, line 1: I opted to keep this a separate file as it seemed like too long a component to append to the numeric-input file.

packages/perseus/src/widgets/numeric-input/numeric-input.tsx
```diff
@@ -31,15 +31,16 @@ export const NumericInputComponent = forwardRef(
 
         // Pass the focus and blur methods to the Numeric Input Class component
         useImperativeHandle(ref, () => ({
+            current: inputRef.current,
             focus: () => {
                 if (inputRef.current) {
-                    inputRef.current.focus();
+                    inputRef.current?.focus();
```
mark-fitzgerald, line 37: Is the optional (`?`) needed when we have the code within a conditional that checks for its existence? I thought that Typescript was able to understand that.

mark-fitzgerald (approved):
Thank you! Great progress on all this clean-up work.

packages/perseus/src/widgets/numeric-input/numeric-input.tsx
SonicScrewdriver: Oooo I actually meant to replace the if conditional! Thank you for catching this

PR #2122, opened 2025-01-16
[Client Validation] Fixes after merging main

jeremywiebe:
## Summary:

I recently merged `main` and the introduction of `MockWidget` caused some tests to break. I've fixed them in this PR and adjusted the MockWidget to be simpler (ie. not use KhanAnswerTypes) as I don't _think_ that's needed.

I also fixed up some related types so that the MockWidget does not exist in our production Widget types, only in tests. The new `MockWidget` illustrates this "test-only" expansion of the widget types.

Issue: LEMS-2561

## Test plan:

`yarn test`
`yarn typecheck`

### DIFF
--- a/packages/perseus-core/src/data-schema.ts
+++ b/packages/perseus-core/src/data-schema.ts
@@ -144,7 +144,6 @@ export interface PerseusWidgetTypes {
     matcher: MatcherWidget;
     matrix: MatrixWidget;
     measurer: MeasurerWidget;
-    "mock-widget": MockWidget;
     "molecule-renderer": MoleculeRendererWidget;
     "number-line": NumberLineWidget;
     "numeric-input": NumericInputWidget;
@@ -346,8 +345,6 @@ export type MatrixWidget = WidgetOptions<'matrix', PerseusMatrixWidgetOptions>;
 // prettier-ignore
 export type MeasurerWidget = WidgetOptions<'measurer', PerseusMeasurerWidgetOptions>;
 // prettier-ignore
-export type MockWidget = WidgetOptions<'mock-widget', MockWidgetOptions>;
-// prettier-ignore
 export type NumberLineWidget = WidgetOptions<'number-line', PerseusNumberLineWidgetOptions>;
 // prettier-ignore
 export type NumericInputWidget = WidgetOptions<'numeric-input', PerseusNumericInputWidgetOptions>;
@@ -400,7 +397,6 @@ export type PerseusWidget =
     | MatcherWidget
     | MatrixWidget
     | MeasurerWidget
-    | MockWidget
     | MoleculeRendererWidget
     | NumberLineWidget
     | NumericInputWidget
@@ -1718,11 +1714,6 @@ export type PerseusPhetSimulationWidgetOptions = {
 export type PerseusVideoWidgetOptions = {
     location: string;
     static?: boolean;
-};
-
-export type MockWidgetOptions = {
-    static?: boolean;
-    value: string;
 };
 
 export type PerseusInputNumberWidgetOptions = {
--- a/packages/perseus-core/src/data-schema.ts
+++ b/packages/perseus-core/src/data-schema.ts
@@ -181,6 +181,18 @@ export interface PerseusWidgetTypes {
 export type PerseusWidgetsMap = MakeWidgetMap<PerseusWidgetTypes>;
 
 /**
+ * PerseusWidget is a union of all the different types of widget options that
+ * Perseus knows about.
+ *
+ * Thanks to it being based on PerseusWidgetTypes interface, this union is
+ * automatically extended to include widgets used in tests without those widget
+ * option types seeping into our production types.
+ *
+ * @see MockWidget for an example
+ */
+export type PerseusWidget = PerseusWidgetTypes[keyof PerseusWidgetTypes];
+
+/**
  * A "PerseusItem" is a classic Perseus item. It is rendered by the
  * `ServerItemRenderer` and the layout is pre-set.
  *
@@ -376,8 +388,6 @@ export type RefTargetWidget = WidgetOptions<'passage-ref-target', PerseusPassage
 export type VideoWidget = WidgetOptions<'video', PerseusVideoWidgetOptions>;
 //prettier-ignore
 export type DeprecatedStandinWidget = WidgetOptions<'deprecated-standin', object>;
-
-export type PerseusWidget = PerseusWidgetTypes[keyof PerseusWidgetTypes];
 
 /**
  * A background image applied to various widgets.
--- a/packages/perseus-core/src/data-schema.ts
+++ b/packages/perseus-core/src/data-schema.ts
@@ -102,7 +102,7 @@ export type MakeWidgetMap<TRegistry> = {
  * `PerseusWidgets` with the one defined below.
  *
  * ```typescript
- * declare module "@khanacademy/perseus" {
+ * declare module "@khanacademy/perseus-core" {
  *     interface PerseusWidgetTypes {
  *         // A new widget
  *         "new-awesomeness": MyAwesomeNewWidget;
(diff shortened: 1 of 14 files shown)


packages/perseus-core/src/data-schema.ts
```diff
@@ -144,7 +144,6 @@ export interface PerseusWidgetTypes {
     matcher: MatcherWidget;
     matrix: MatrixWidget;
     measurer: MeasurerWidget;
-    "mock-widget": MockWidget;
```
jeremywiebe, line 147: Removing here as we don't want the MockWidget to be part of the public set of widget types. It is an internal, test-only widget.

packages/perseus-core/src/data-schema.ts
```diff
@@ -346,8 +345,6 @@ export type MatrixWidget = WidgetOptions<'matrix', PerseusMatrixWidgetOptions>;
 // prettier-ignore
 export type MeasurerWidget = WidgetOptions<'measurer', PerseusMeasurerWidgetOptions>;
 // prettier-ignore
-export type MockWidget = WidgetOptions<'mock-widget', MockWidgetOptions>;
```
jeremywiebe, line 349: I moved all types related to the MockWidget into its directory so that we don't have test pieces intermingled/shipped with our production code.

packages/perseus-core/src/data-schema.ts
```diff
@@ -380,42 +377,7 @@ export type VideoWidget = WidgetOptions<'video', PerseusVideoWidgetOptions>;
...
-    | PhetSimulationWidget
-    | PlotterWidget
-    | PythonProgramWidget
-    | RadioWidget
-    | RefTargetWidget
-    | SorterWidget
-    | TableWidget
-    | VideoWidget
-    | DeprecatedStandinWidget;
+export type PerseusWidget = PerseusWidgetTypes[keyof PerseusWidgetTypes];
```
jeremywiebe, line 380: Now that we have the PerseusWidgetTypes interface (which can be properly extended in test code and/or externally), we can derive this union from that (meaning we don't have to maintain two official set of widget types.

packages/perseus/src/__testdata__/server-item-renderer.testdata.ts
```diff
@@ -40,7 +41,7 @@ export const itemWithNumericInput: PerseusItem = {
                     labelText: "What's the answer?",
                     size: "normal",
                 },
-            } as NumericInputWidget,
+            } satisfies NumericInputWidget,
```
jeremywiebe, line 44: `satisfies` 

> The new satisfies operator lets us validate that the type of an expression matches some type, without changing the resulting type of that expression. 

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#the-satisfies-operator

Compare to `as` which does some type checking but is not as robust and can end up hiding test data that doesn't actually fully match the `as` type.

packages/perseus/src/widgets/mock-widgets/mock-widget-types.ts
```diff
@@ -0,0 +1,23 @@
...
+export type PerseusMockWidgetUserInput = {
+    currentValue: string;
+};
+
+// Extend the widget registries for testing
+declare module "@khanacademy/perseus-core" {
+    export interface PerseusWidgetTypes {
+        "mock-widget": MockWidget;
+    }
+}
```
jeremywiebe, line 23: This is the magic sauce that extends the set of `PerseusWidgetTypes` to include the mock widget. Anything that ends up importing this file will "see" the `mock-widget`... meaning tests, but not production code!

packages/perseus-core/src/data-schema.ts
```diff
@@ -380,43 +389,6 @@ export type VideoWidget = WidgetOptions<'video', PerseusVideoWidgetOptions>;
 //prettier-ignore
 export type DeprecatedStandinWidget = WidgetOptions<'deprecated-standin', object>;
 
-export type PerseusWidget =
```
jeremywiebe, line 383: Moved up to Line 183.

packages/perseus/src/widgets/mock-widgets/mock-widget.tsx
```diff
@@ -6,18 +6,15 @@ import * as React from "react";
 import {getPromptJSON as _getPromptJSON} from "../../widget-ai-utils/mock-widget/prompt-utils";
 
 import scoreMockWidget from "./score-mock-widget";
+import validateMockWidget from "./validate-mock-widget";
```
handeyeco, line 9: Not related to your PR, but I wonder if all this mock widget stuff should even be in `/packages` when it could possibly live in `/testing`.

packages/perseus/src/widgets/mock-widgets/mock-widget-types.ts
handeyeco: I don't really understand the code and after looking it up I still don't understand the code. I wonder if would be worth adding a link to something that explains the goal here.

packages/perseus/src/widget-ai-utils/mock-widget/prompt-utils.ts
```diff
@@ -1,5 +1,5 @@
-import type {PerseusMockWidgetUserInput} from "../../widgets/mock-widgets/mock-widget";
 import type mockWidget from "../../widgets/mock-widgets/mock-widget";
+import type {PerseusMockWidgetUserInput} from "../../widgets/mock-widgets/mock-widget-types";
```
handeyeco, line 2: Why do we need AI prompt utils for a mock widget?

packages/perseus-core/src/data-schema.ts
```diff
@@ -181,6 +180,18 @@ export interface PerseusWidgetTypes {
...
+ * PerseusWidget is a union of all the different types of widget options that
+ * Perseus knows about.
+ *
+ * Thanks to it being based on PerseusWidgetTypes interface, this union is
+ * automatically extended to include widgets used in tests without those widget
+ * option types seeping into our production types.
+ *
+ * @see MockWidget for an example
+ */
+export type PerseusWidget = PerseusWidgetTypes[keyof PerseusWidgetTypes];
```
handeyeco, line 193: Nice

handeyeco (approved):

packages/perseus/src/widgets/mock-widgets/mock-widget-types.ts
jeremywiebe: I'll expand the comment.  Honestly, I wasn't able to find a really clear description for the case we're using this for.  The bottom line is that when this file is imported  (should be from tests only), the `PerseusWidgetTypes` type from `@khanacademy/perseus-core` is extended to include the `MockWidget`.  Our other types that define the unions of widget options types, etc. are all based on `PerseusWidgetTypes`. So once we extend `PerseusWidgetTypes` with this mock widget type, all the other widget option unions know about the mock widget also. This means that our tests and test data understand that `mock-widget 1` refers to a mock widget, but if you were to try it in webapp, it wouldn't know about mock widgets (rightfully).

packages/perseus/src/widget-ai-utils/mock-widget/prompt-utils.ts
jeremywiebe: I think just to "round out" the implementation of the mock widget. I don't think we actually use it.  Perhaps it'd be better if we deleted this to avoid confusion and have folks think it serves some purpose.

jeremywiebe: Changed my mind. There's tests that exercise the `Renderer` calling the widget's `getPromptJSON()` using the mock widget. I'll leave it in place as that feels like a valid use of the mock widget. I'd also like to avoid doing this sort of refactoring in this PR as I'm trying to stick to just dealing with after effects from merging `main`. :)

PR #2123, opened 2025-01-16
Make `values` optional on `PerseusCategorizerWidgetOptions` type

benchristel:
This is a proof of concept of how we might change the WidgetOptions types so we
can avoid sending "private data" like hints and answers to the client.

Issue: LEMS-2562

## Test plan:

`yarn test`

### DIFF
--- a/packages/perseus-core/src/data-schema.ts
+++ b/packages/perseus-core/src/data-schema.ts
@@ -423,7 +423,7 @@ export type PerseusCategorizerWidgetOptions = {
     // Whether this widget is displayed with the results and immutable
     static: boolean;
     // The correct answers where index relates to the items and value relates to the category.  e.g. [0, 1, 0, 1, 2]
-    values: ReadonlyArray<number>;
+    values?: ReadonlyArray<number>;
     // Whether we should highlight i18n linter errors found on this widget
     highlightLint?: boolean;
     // Internal editor configuration. Can be ignored by consumers.
--- a/packages/perseus/src/util/extract-perseus-data.ts
+++ b/packages/perseus/src/util/extract-perseus-data.ts
@@ -59,9 +59,9 @@ function getAnswersFromWidgets(
                     categorizer.options?.items &&
                     categorizer.options?.values
                 ) {
-                    const categories = categorizer.options?.categories;
-                    const items = categorizer.options?.items;
-                    const values = categorizer.options?.values;
+                    const categories = categorizer.options.categories;
+                    const items = categorizer.options.items;
+                    const values = categorizer.options.values;
                     answers.push(
                         ...values.map(
                             (value, index) =>
--- a/packages/perseus/src/widgets/categorizer/categorizer.tsx
+++ b/packages/perseus/src/widgets/categorizer/categorizer.tsx
@@ -28,13 +28,13 @@ import type {CategorizerPromptJSON} from "../../widget-ai-utils/categorizer/cate
 import type {PerseusCategorizerWidgetOptions} from "@khanacademy/perseus-core";
 
 type Props = WidgetProps<RenderProps, PerseusCategorizerScoringData> & {
-    values: ReadonlyArray<string>;
+    values: ReadonlyArray<number>;
 };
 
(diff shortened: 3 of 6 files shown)


packages/perseus/src/util/extract-perseus-data.ts
```diff
@@ -59,9 +59,9 @@ function getAnswersFromWidgets(
                     categorizer.options?.items &&
                     categorizer.options?.values
                 ) {
-                    const categories = categorizer.options?.categories;
-                    const items = categorizer.options?.items;
-                    const values = categorizer.options?.values;
+                    const categories = categorizer.options.categories;
+                    const items = categorizer.options.items;
+                    const values = categorizer.options.values;
```
benchristel, line 64: These `?.` nullish coalescing operators were not needed, since we check the nullishness of each of these fields above.

packages/perseus/src/widgets/categorizer/categorizer.tsx
```diff
@@ -28,13 +28,13 @@ import type {CategorizerPromptJSON} from "../../widget-ai-utils/categorizer/cate
 import type {PerseusCategorizerWidgetOptions} from "@khanacademy/perseus-core";
 
 type Props = WidgetProps<RenderProps, PerseusCategorizerScoringData> & {
-    values: ReadonlyArray<string>;
+    values: ReadonlyArray<number>;
```
benchristel, line 31: This should have been `ReadonlyArray<number>` all along. Fixing this let me remove the `@ts-expect-error` below.

packages/perseus-core/src/data-schema.ts
```diff
@@ -423,7 +423,7 @@ export type PerseusCategorizerWidgetOptions = {
     // Whether this widget is displayed with the results and immutable
     static: boolean;
     // The correct answers where index relates to the items and value relates to the category.  e.g. [0, 1, 0, 1, 2]
-    values: ReadonlyArray<number>;
+    values?: ReadonlyArray<number>;
```
benchristel, line 426: Potential concern here: with this change, the `data-schema` types no longer represent the JSON that actually gets persisted. The JSON should always have the `values` field, but the _client_ won't always receive it.

If we wanted to fix this, we would have to decouple the widget prop types from these data-schema types. Though I think maybe they are already pretty much decoupled (in theory) by the `transform` and `staticTransform` functions. `transform` renders the widget with public-only data and `staticTransform` renders with public+private data.

packages/perseus/src/widgets/categorizer/categorizer.tsx
Myranae: I thought this might be the case, but I couldn't figure out why it was added at all. It's already included in renderProps, so I thought it might have been for another reason. Glad we can make it number! Do we even need that added on if it's already added via RenderProps?

packages/perseus-core/src/data-schema.ts
Myranae: I think if the JSON should always have the values field, then it might be worth it to fix this. It seems like the data-schema and our contract with users is pretty important, and I'd rather it be as accurate as possible. Aren't we creating different types for the JSON versus what the client is getting though?

packages/perseus/src/widgets/categorizer/categorizer.tsx
```diff
@@ -28,13 +28,13 @@ import type {CategorizerPromptJSON} from "../../widget-ai-utils/categorizer/cate
...
 };
 
 type DefaultProps = {
     items: Props["items"];
     categories: Props["categories"];
-    values: Props["values"];
+    values: ReadonlyArray<number>;
```
Myranae, line 37: Is this necessary if you updated the type of values to be this anyway?

packages/perseus-core/src/data-schema.ts
benchristel: Yeah, I agree it's important for the data-schema to accurately reflect the JSON. You are right, the JSON types and the prop types are going to be different in a SSS world. I'll post in Slack with more thoughts.

PR #2126, opened 2025-01-17
Fix a React warning about non-unique component keys in the editor

benchristel:
## Summary:
WidgetOptions has a `key` field (which appears to be unused, but that's
a different discussion) which is often not present in the JSON data.
This field, when absent from the JSON, was getting set to `undefined` by
`parseAndMigratePerseusItem()`. The `undefined` value then overrode the
`key` prop which we intended to set on the WidgetEditor component, which
caused React to log an error about non-unique keys.

I've changed the order of the props so the keys we want to set
explicitly will override any fields parsed out of the widget JSON. This
is what we want anyway, since the JSON might contain absolutely
anything, and we don't want it to be able to set arbitrary editor props.

I discovered this bug while hooking up `parseAndMigratePerseusItem` in
webapp (https://github.com/Khan/webapp/pull/28551)

Issue: https://khanacademy.atlassian.net/browse/LEMS-2774

Test plan:

`yarn test`

### DIFF
--- a/.changeset/slimy-scissors-smash.md
+++ b/.changeset/slimy-scissors-smash.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": patch
+---
+
+Fix a React warning about non-unique component keys in the exercise editor.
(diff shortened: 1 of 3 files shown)


jeremywiebe (approved):
Thanks for this. I have seen these `key` fields in widget options as well. My hunch is that these were introduced by some widget editor being broken and letting react component keys escape into the serialized data.

PR #2128, opened 2025-01-17
Refactoring Numeric Input helper functions to remove underscore

SonicScrewdriver:
## Summary:
This PR is part of the Numeric Input Project, and will be landed onto the feature branch for a full QA pass. 

The intended goal was to remove all cases of underscore in the Numeric Input component, and to improve the commenting / documentation of the code. 

Some things to note: 
- I've changed the logic of `generateExamples` to match `shouldShowExamples`, as we were generating a list of all possible examples and simply not displaying it. This seemed unnecessary and we can exit both functions early.
- I've added more specific types for `PerseusNumericInputWidgetOptions.simplify` 

Issue: LEMS-2446

## Test plan:
- New tests 
- Manual testing in storybook


packages/perseus-core/src/data-schema.ts
```diff
@@ -1210,7 +1212,7 @@ export type PerseusNumericInputAnswer = {
     // NOTE: perseus_data.go says this is non-nullable even though we handle null values.
     maxError: number | null | undefined;
     // Unsimplified answers are Ungraded, Accepted, or Wrong. Options: "required", "correct", or "enforced"
-    simplify: string | null | undefined;
+    simplify: PerseusNumericInputSimplify;
```
SonicScrewdriver, line 1215: This could be a spicy move, but I think it's good to be more specific with our types here. While it resulted in some minor downstream updates, I think greater specificity will help us.

packages/perseus/src/widgets/numeric-input/numeric-input.class.tsx
```diff
@@ -157,79 +156,34 @@ export class NumericInput
...
-};
-
 type RenderProps = {
-    answerForms: ReadonlyArray<{
-        simplify: "required" | "correct" | "enforced" | null | undefined;
-        name: "integer" | "decimal" | "proper" | "improper" | "mixed" | "pi";
-    }>;
-    labelText: string;
-    size: "normal" | "small";
+    answerForms: ReadonlyArray<PerseusNumericInputAnswerForm>;
```
SonicScrewdriver, line 160: We already have this type, no need to force ourselves to maintain two locations.

packages/perseus/src/widgets/numeric-input/numeric-input.test.ts
```diff
@@ -376,35 +374,3 @@ describe("Numeric input widget", () => {
         );
     });
 });
-
-describe("unionAnswerForms utility function", () => {
```
SonicScrewdriver, line 380: I created better tests over in `utils.test.ts` for this logic.

packages/perseus/src/widgets/numeric-input/utils.test.ts
```diff
@@ -15,7 +16,7 @@ describe("generateExamples", () => {
                 simplify: "required",
             },
         ];
-        const strings: Partial<PerseusStrings> = {
+        const fakePerseusStrings: Partial<PerseusStrings> = {
```
SonicScrewdriver, line 19: Mostly a self nit, but it seemed good to be really specific that I'm faking the strings here.

packages/perseus/src/widgets/numeric-input/utils.ts
```diff
@@ -35,37 +33,105 @@ export const generateExamples = (
...
+
+    return !allFormsAccepted;
+};
+
+/**
+ * uniqueAnswerForms takes a list of answer forms and returns a list of unique
+ * answer forms. This is useful for ensuring that we don't show duplicate examples
+ * to the user.
+ */
+const getUniqueAnswerForms = function (
```
mark-fitzgerald, line 84: I wonder if it would be easier to implement something like [how number line tick marks are handled](https://github.com/Khan/perseus/blob/main/packages/perseus/src/widgets/number-line/number-line.tsx#L172)? Perhaps something like:
```
return [...new Set(list)];
```
Maybe even just use that syntax in place of calls to `getUniqueAnswerForms()`?

packages/perseus/src/widgets/numeric-input/utils.ts
```diff
@@ -35,37 +33,105 @@ export const generateExamples = (
...
-            : Object.keys(NumericExampleStrings).map((name) => {
-                  return {
-                      name: name,
-                      simplify: "required",
-                  } as PerseusNumericInputAnswerForm;
-              });
+    // If the Content Creator has not specified any answer forms,
+    // we do not need to show any examples.
+    if (answerForms.length === 0) {
+        return [];
```
SonicScrewdriver, line 39: I wanted to call out that I've technically changed this functionality. 

Previously we would return a string with all of the example strings that we just never displayed. This seemed like unnecessary logic/computations so I've updated it to simply exit early in such a case.

packages/perseus/src/widgets/numeric-input/utils.ts
SonicScrewdriver: I modded my approach slightly but you're right, this could be simplified even more! :)

SonicScrewdriver: This did make me think deeper about which examples we're filtering: the way we've historically generated these examples seems antithetical to how we intend Content Creators to create answers. 

If I make a wrong answer that's strictly set to match a decimal, and a right answer set to strictly match a fraction, the provided tooltip will read: "Your answer should be - a simplified proper fraction - an exact decimal like 0.75". This would directly mislead the user as their answer specifically _shouldn't_ be a decimal.🤔

This is how the logic works currently so it's not a regression. It does make me wonder if we should only be providing the _**correct**_ answers to the tooltip however. Perhaps a ticket for emergent work?

packages/perseus-core/src/data-schema.ts
jeremywiebe: Merging the `| null | undefined` into the type feels like it hides something important. Without referring to the type, this field looks required, but you find out that it's optional only when you inspect the new `PerseusNumericInputSimplify`. 

I'm not sure if we have a precedent, but I think I'd prefer we kept the new type as the union of string values and then add `| null | undefined` here so that the "optionality" stays in situ.

packages/perseus/src/components/input-with-examples.tsx
```diff
@@ -93,7 +93,7 @@ class InputWithExamples extends React.Component<Props, State> {
         const ariaId = `aria-for-${id}`;
         // Generate text from a known set of format options that will read well in a screen reader
         const examplesAria =
-            this.props.examples.length === 7
+            this.props.examples.length === 0
```
jeremywiebe, line 96: The original `7` feels very magic. Why are we now changing it to `0`?

packages/perseus/src/widgets/numeric-input/numeric-input.class.tsx
```diff
@@ -157,79 +156,34 @@ export class NumericInput
...
                         simplify: answer.simplify,
                         name: form,
                     };
                 });
             }),
         ),
-    });
+    };
 
+    delete rendererProps.answers;
```
jeremywiebe, line 186: Instead of needing to jump through this typing hoop, what do you think of restructuring the `answers` away? 

```typescript
   const {answers: _, ...rendererProps} = {
         ...widgetOptions,
         answerForms: unionAnswerForms(
             widgetOptions.answers.map((answer) => {
                 return (answer.answerForms || []).map((form) => {
                     return {
                         simplify: answer.simplify,
                         name: form,
                     };
                 });
             }),
         ),
     };
```

packages/perseus/src/widgets/numeric-input/utils.test.ts
jeremywiebe: **question**: could we not use the mock strings from https://github.com/Khan/perseus/blob/9cabe689a7aa143f95adf4556bf5c10d654a66ae/packages/perseus/src/strings.ts#L551 ?

jeremywiebe (commented):
Sorry... drive by review comments. 🫶

packages/perseus/src/widgets/numeric-input/utils.ts
mark-fitzgerald: Agreed. Perhaps we can get confirmation from Content folks that only correct answers should be included in the tooltip? Though, I wonder if the scoring logic would fail when a "wrong" answer matches a "correct" answer (i.e. the decimal equivalent of a fraction).

packages/perseus/src/components/input-with-examples.tsx
SonicScrewdriver: Previously we were looping through to return all of the example strings when the teacher had not selected any `answerForms`. Then we would never show these strings as `shouldShowExamples` would hide them in this case. 

There are 6 example strings, and then we prepend a "Your answer should be" as the first string (for a total of 7). In the case where teachers have selected all 6 answerForms, we still don't show these answerForms as they are effectively useless. 

```     
 const forms =
            this.props.answerForms?.length !== 0
                ? this.props.answerForms
                : Object.keys(formExamples).map((name) => {
                      return {
                          name: name,
                          simplify: "required",
                      } as PerseusNumericInputAnswerForm;
                  });
```

That seemed a little silly to me, so I've updated the function to exit early in such a case, returning an empty array. That meant that I could update this section to make more sense. :)

jeremywiebe: Ohhhh... Nice. Thanks for cleaning that up!

packages/perseus/src/widgets/numeric-input/numeric-input.class.tsx
SonicScrewdriver: Much better, thank you!

mark-fitzgerald (approved):
Look at those cobwebs go! So much cleaner - thank you!!

PR #2130, opened 2025-01-20
Disable Chromatic snapshots globally

jeremywiebe:
## Summary:

This PR disables Chromatic snapshots globally (in `preview.tsx`) and then enables it in the one `-regression.stories.tsx` file.

Going forward, we can enable snapshots per-file.

We are making this change to address the fact that hte Perseus repo has been burning through our organization-wide Turbosnap budget because even skipped tests contribute to the Turbosnap count when snapshots are enabled (see the support thread that Juan had with Chromatic: https://khanacademy.slack.com/archives/CEA6W0F6F/p1737389339991279?thread_ts=1737389278.405609&cid=CEA6W0F6F). Specifically, in our Perseus repo we use `onlyStoryFiles`:

> OnlyStoryNames, OnlyStoryFiles work exactly like TurboSnap. The bypassed snapshots (skipping 366 tests) are counted and charged as TurboSnaps.

| Before | After |
| ------ | ----- |
| original | after | 

Issue: "none"

## Test plan:

Land the PR and then watch the build and make sure it doesn't write out that it skipped +300 tests.


packages/perseus/src/widgets/interactive-graphs/interactive-graph-regression.stories.tsx
```diff
@@ -6,261 +6,245 @@ import {mockStrings} from "../../strings";
 import {interactiveGraphQuestionBuilder} from "./interactive-graph-question-builder";
 
 import type {PerseusRenderer} from "@khanacademy/perseus-core";
+import type {Meta, StoryObj} from "@storybook/react";
 
-type StoryArgs = Record<any, any>;
+type Story = StoryObj<typeof MafsQuestionRenderer>;
 
-export default {
+const meta: Meta<typeof MafsQuestionRenderer> = {
```
jeremywiebe, line 13: I switched this entire file to CSF v3 so that we'd have typing on the default export as well as trim down the size of all stories (and have better typing on them too!).

jandrade (approved):
Looks great! thanks for the quick fix and for adding the comments on why we have to disable snapshots this way 🚢

PR #2135, opened 2025-01-22
[LEMS-2737/default-expression] move expression upgrade logic

handeyeco:
## Summary:
<Add your summary here>

Issue: LEMS-2737

## Test plan:

### DIFF
--- a/packages/perseus-core/src/index.ts
+++ b/packages/perseus-core/src/index.ts
@@ -28,5 +28,7 @@ export * from "./data-schema";
 
 export {pluck, mapObject} from "./utils/objective_";
 
-export * as expressionUpgrade from "./widgets/expression";
+export {default as expressionLogic} from "./widgets/expression";
 export type {ExpressionDefaultWidgetOptions} from "./widgets/expression";
+
+export type * from "./widgets/logic-export.types";
--- a/packages/perseus-core/src/index.ts
+++ b/packages/perseus-core/src/index.ts
@@ -27,3 +27,6 @@ export {PerseusError} from "./error/perseus-error";
 export * from "./data-schema";
 
 export {pluck, mapObject} from "./utils/objective_";
+
+export * as expressionUpgrade from "./widgets/expression";
+export type {ExpressionDefaultWidgetOptions} from "./widgets/expression";
--- a/packages/perseus-core/src/widgets/expression.ts
+++ b/packages/perseus-core/src/widgets/expression.ts
@@ -1,0 +1,36 @@
+import type {PerseusExpressionWidgetOptions} from "../data-schema";
+
+export const currentVersion = {major: 1, minor: 0};
+
+export const widgetOptionsUpgrades = {
+    /* c8 ignore next */
+    "1": (v0props: any): PerseusExpressionWidgetOptions => ({
+        times: v0props.times,
+        buttonSets: v0props.buttonSets,
+        functions: v0props.functions,
+        buttonsVisible: v0props.buttonsVisible,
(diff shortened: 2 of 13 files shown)


packages/perseus-core/src/index.ts
```diff
@@ -27,3 +27,6 @@ export {PerseusError} from "./error/perseus-error";
 export * from "./data-schema";
 
 export {pluck, mapObject} from "./utils/objective_";
+
+export * as expressionUpgrade from "./widgets/expression";
```
jeremywiebe, line 31: * I think this could simply be called `expression` (it's not strictly related to upgrades... well, maybe kinda... meh... it's a name)
* I would feel more comfortable if the pieces we're exporting per-widget had some type that defined what we export. Kinda like a subset of `WidgetExports<T>`. 🤔

packages/perseus-core/src/widgets/expression.ts
```diff
@@ -0,0 +1,36 @@
+import type {PerseusExpressionWidgetOptions} from "../data-schema";
+
+export const currentVersion = {major: 1, minor: 0};
+
+export const widgetOptionsUpgrades = {
+    /* c8 ignore next */
```
jeremywiebe, line 6: Let's drop this comment.  It's from the original stabilization where we were focused on reaching a certain percentage of test coverage and didn't plan to include this upgrade logic in that coverage calculation.

packages/perseus-core/src/widgets/expression.ts
```diff
@@ -0,0 +1,36 @@
...
+                considered: "correct",
+                form: v0options.form,
+                simplify: v0options.simplify,
+                value: v0options.value,
+            },
+        ],
+    }),
+} as const;
+
+export type ExpressionDefaultWidgetOptions = Pick<
```
jeremywiebe, line 26: Nice use of `Pick<>` here.

packages/perseus-editor/src/widgets/expression-editor.tsx
```diff
@@ -94,12 +91,8 @@ type State = {
 class ExpressionEditor extends React.Component<Props, State> {
     static widgetName = "expression" as const;
 
-    static defaultProps: DefaultProps = {
-        answerForms: [],
-        times: false,
-        buttonSets: ["basic"],
-        functions: ["f", "g", "h"],
-    };
+    static defaultProps: ExpressionDefaultWidgetOptions =
```
jeremywiebe, line 94: **thought**: Just thinking aloud... After our chat about widget options vs. props, this feels to blur the line again a bit. Although, keeping both the `DefaultProps` type and `ExpressionDefaultWidgetOptions` doesn't feel useful/reasonable. I'm not sure there's any action here unless this triggers an idea in your head.  

No action needed.

jeremywiebe (approved):
Yeah, this looks in line with what we talked about while pairing!

packages/perseus-editor/src/widgets/expression-editor.tsx
handeyeco: I was thinking the same thing. In my head, my justification is: the editors are creating the JSON and the JSON is then transformed into the widget props for the preview. So I was thinking the editor was a little closer to the data than the widget's UI. I don't feel strongly about it either way and I'd be fine having a separate type.

packages/perseus-core/src/index.ts
handeyeco: I think that makes sense if long-term we want to move the scorers into core.

PR #2137, opened 2025-01-22
Create helper to build public widget options for orderer

Myranae:
## Summary:
Adds a function that takes orderer's full widget options and filters out answer data. It also adds this function to the widget's widget export and adds a test confirming the function does as expected. 

Issue: LEMS-2763

## Test plan:
- Confirm all checks pass
- Confirm orderer still works as expected


packages/perseus/src/widgets/orderer/orderer.util.ts
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * For details on the individual options, see the
+ * PerseusOrdererWidgetOptions type
+ */
+type OrdererPublicWidgetOptions = {
+    options: ReadonlyArray<PerseusRenderer>;
+    height: "normal" | "auto";
+    layout: "horizontal" | "vertical";
+};
```
Myranae, line 14: Left out correctOptions and otherOptions.

For contrast, here are the regular widget options:
```
export type PerseusOrdererWidgetOptions = {
    // All of the options available to the user. Place the cards in the correct order. The same card can be used more than once in the answer but will only be displayed once at the top of a stack of identical cards.
    options: ReadonlyArray<PerseusRenderer>;
    // The correct order of the options
    correctOptions: ReadonlyArray<PerseusRenderer>;
    // Cards that are not part of the answer
    otherOptions: ReadonlyArray<PerseusRenderer>;
    // "normal" for text options.  "auto" for image options.
    height: "normal" | "auto";
    // Use the "horizontal" layout for short text and small images. The "vertical" layout is best for longer text (e.g. proofs).
    layout: "horizontal" | "vertical";
};
```

benchristel (approved):
LGTM!

packages/perseus/src/widgets/orderer/orderer.util.ts
```diff
@@ -0,0 +1,30 @@
...
+} from "@khanacademy/perseus-core";
+
+/**
+ * For details on the individual options, see the
+ * PerseusOrdererWidgetOptions type
+ */
+type OrdererPublicWidgetOptions = {
+    options: ReadonlyArray<PerseusRenderer>;
+    height: "normal" | "auto";
+    layout: "horizontal" | "vertical";
```
jeremywiebe, line 13: I'm not sure if this was discussed elsewhere, but it might be useful to build these types off of the main widget options type. This will help keep it clear that this type is a subset of the full widget options. 

```suggestion
    height: PerseusOrdererWidgetOptions["height"];
    layout: PerseusOrdererWidgetOptions["layout"];
```

packages/perseus/src/widgets/orderer/orderer.util.ts
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Given a PerseusOrdererWidgetOptions object, return a new object with only
+ * the public options that should be exposed to the client.
+ */
+function getOrdererPublicWidgetOptions(
+    options: PerseusOrdererWidgetOptions,
+): OrdererPublicWidgetOptions {
+    return {
+        options: options.options,
```
jeremywiebe, line 24: I have a sense that we should "deep copy" these public variants. As this is, the `.options` array will be the same one between the original widget options and this public one we're returning. Likely never an issue, but making a copy feels safer. 

@benchristel do you have thoughts?

```suggestion
        options: [...options.options],
```

jeremywiebe (approved):
Nice.  One thing I was thinking as I read through this. We will almost certainly need to call this in a server environment and so I think a better place to build out these helpers is in `@khanacademy/perseus-core`.  Otherwise, shortly after we build these all we'll be moving them anyways. :)

packages/perseus/src/widgets/orderer/orderer.util.ts
Myranae: Updated :)

packages/perseus/src/widgets/orderer/orderer.util.ts
Myranae: Ben, didn't you mention that with Categorizer I think we update the values field based on user input? Would that be an instance of modifying these options? I think the main goal of these public widget options is just to render the clientside widgets, but if the object gets modified to track user input, copying might be needed.

jeremywiebe: I can't see these functions ever being used directly with React. They'll be used in an editing environment and the publish pipeline most so the issue of mutation after copy doesn't matter here. 

So let's avoid copying.

PR #2138, opened 2025-01-22
[its-a-numeric-story] Improving Numeric Input Storybook Stories

SonicScrewdriver:
## Summary:

The purpose of this PR is to improve our storybook setup for our Numeric Input Widget. This includes the following work:
- Modernizing story structure 
- Hooking up argTypes with descriptions 
- Updating RendererWithDebugUI to also set customKeypad to the same value as isMobile 
     - This allows us to ensure that our Widgets that use MathInput are properly updating when toggled into Mobile view
- Adjustments to SideBySide 
     - Automatically collapse the Perseus JSON view. 
     - Moved the PerseusJSON view below the Renderer View
     - Rename to SplitView to better encapsulate the new design 
     - Updated variable names to match

[Current (Live) Storybook Example](https://khan.github.io/perseus/?path=/docs/perseus-widgets-numericinput--docs) | [PR Storybook Example](https://650db21c3f5d1b2f13c02952-osexoxinde.chromatic.com/?path=/docs/perseus-widgets-numeric-input--docs)

Issue: LEMS-2449

## Video Example:
https://github.com/user-attachments/assets/69f6dbfb-1fda-445b-a06f-90a178f9dbeb

## Test plan:
- Ensure all tests pass + manual testing

### DIFF
--- a/packages/perseus/src/widgets/numeric-input/numeric-input.stories.tsx
+++ b/packages/perseus/src/widgets/numeric-input/numeric-input.stories.tsx
@@ -11,6 +11,27 @@ import type {
 } from "@khanacademy/perseus-core";
 import type {Meta} from "@storybook/react";
 
+// We're using this format as storybook was not able to infer the type of the options.
+// It also gives us a lovely hover view of the JSON structure.
+const answerFormsArray: string = `[
+    {
+        simplify: string;
+        name: string;
+    }
+]`;
+
+const answersArray: string = `[
+    {
+        message: string;
+        value: number;
+        status: string;
+        answerForms: array<string>;
+        strict: boolean;
+        maxError: number;
+        simplify: string;
+    }
+]`;
+
 const meta: Meta<typeof NumericInput> = {
     component: NumericInput,
     title: "Perseus/Widgets/Numeric Input",
@@ -29,30 +50,23 @@ const meta: Meta<typeof NumericInput> = {
                 message: "",
             },
         ],
-        labelText: "What's the answer?",
-        answerForms: [],
+        answerForms: [
+            {simplify: "required", name: "decimal"},
+            {simplify: "required", name: "integer"},
+            {simplify: "required", name: "mixed"},
+            {simplify: "required", name: "percent"},
+            {simplify: "required", name: "pi"},
+        ],
     },
     argTypes: {
         answers: {
             control: {type: "object"},
             description:
                 "A list of all the possible correct and incorrect answers",
-            type: {
-                name: "array",
-                value: {
-                    name: "object",
-                    value: {
-                        message: {name: "string", required: true},
-                        value: {name: "number"},
-                        status: {name: "string", required: true},
-                        answerForms: {
-                            name: "array",
-                            value: {name: "string"},
-                        },
-                        strict: {name: "boolean", required: true},
-                        maxError: {name: "number"},
-                        simplify: {name: "string"},
-                    },
+            table: {
+                type: {
+                    summary: "array",
+                    detail: answersArray,
                 },
             },
         },
@@ -60,14 +74,10 @@ const meta: Meta<typeof NumericInput> = {
             control: {type: "object"},
             description:
                 "Used by examples, maybe not used and should be removed in the future",
-            type: {
-                name: "array",
-                value: {
-                    name: "object",
-                    value: {
-                        simplify: {name: "string"},
-                        name: {name: "string"},
-                    },
+            table: {
+                type: {
+                    summary: "array",
+                    detail: answerFormsArray,
                 },
             },
         },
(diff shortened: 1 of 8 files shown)


testing/side-by-side.tsx
```diff
@@ -4,34 +4,32 @@ import {HeadingMedium} from "@khanacademy/wonder-blocks-typography";
...
-    leftTitle: React.ReactNode;
-    left: React.ReactNode;
-    rightTitle?: string;
+    rendererTitle: React.ReactNode;
+    renderer: React.ReactNode;
+    JSONTitle?: string;
     jsonObject: any;
 };
 
-const SideBySide = ({
```
SonicScrewdriver, line 16: After some discussion, it was determined that we really don't like this side-by-side view. I've renamed this to "SplitView" to be a little more generic, and renamed the variables accordingly. 

This moves the PerseusJSON below the Widget, which gives both the Widget and the JSON more room to breathe.

testing/renderer-with-debug-ui.tsx
```diff
@@ -40,9 +40,15 @@ export const RendererWithDebugUI = ({
     const [isMobile, setIsMobile] = React.useState(false);
     const {strings} = usePerseusI18n();
 
+    const controlledAPIOptions = {
+        ...apiOptions,
+        isMobile,
+        customKeypad: isMobile, // Use the mobile keypad for mobile
+    };
```
SonicScrewdriver, line 47: Many of our stories would not respond to the "Mobile" toggle. This addition now ensures that all Widgets that use MathInput will properly render their mobile views.

testing/side-by-side.tsx
```diff
@@ -4,34 +4,32 @@ import {HeadingMedium} from "@khanacademy/wonder-blocks-typography";
...
-            <View style={styles.rightPanel}>
-                <HeadingMedium>{rightTitle}</HeadingMedium>
+            <View>
+                <HeadingMedium>{JSONTitle}</HeadingMedium>
                 <ReactJson
                     style={{marginTop: "10px"}}
                     quotesOnKeys={false}
                     enableClipboard={false}
-                    collapsed={5}
+                    collapsed={true}
```
SonicScrewdriver, line 32: This auto-collapses the PerseusJSON view so that it doesn't fill the page, and push the Storybook Controls "below the fold".

packages/perseus/src/widgets/numeric-input/numeric-input.testdata.ts
```diff
@@ -134,6 +175,7 @@ export const multipleAnswersWithDecimals: PerseusRenderer = {
                         value: 12.2,
                         simplify: "required",
                         message: "",
+                        answerForms: ["decimal"],
```
SonicScrewdriver, line 178: Realized that these should be specifying their answerForms, just for consistency.

packages/perseus/src/widgets/numeric-input/numeric-input.stories.tsx
```diff
@@ -1,99 +1,188 @@
...
-                Small:
-                <NumericInput {...smallProps} />
-            </label>
-            <label>
-                Normal:
-                <NumericInput {...normalProps} />
-            </label>
-        </div>
-    );
+export const Default = (
```
SonicScrewdriver, line 170: After setting up all the Storybook controls, I wasn't really sure if we needed any more stories as we can control most of these options on the fly now.

SonicScrewdriver: Ah! Perhaps a story for each answerForm, as it IS a little frustrating to edit the json views for that. If possible, it'd be good to get an example of the pi answerForm with a note about how to trigger the "It seems like you've approximated pi" message.

testing/split-view.tsx
```diff
@@ -4,34 +4,36 @@ import {HeadingMedium} from "@khanacademy/wonder-blocks-typography";
...
 
-const SideBySide = ({
-    leftTitle = "Renderer",
-    left,
-    rightTitle = "Perseus JSON",
+/**
+ * This component is used in Storybook to render both the renderer and the JSON object in a split view.
+ * Currently, the renderer is displayed above the JSON object.
+ */
+const SplitView = ({
```
SonicScrewdriver, line 18: There was some discussion about how the team would like to restructure the Storybook views. By moving the PerseusJSON view below the Renderer (Widgets), we can give both elements more room to breathe. This is particularly noticeable on the Number Line Widget, or with the Graded Group JSON. 

As a result of this move, "SideBySide" no longer felt like an accurate name for this component. I've renamed it to "SplitView" to be more generic, just in case we ever wish to reverse this decision. I've also updated variables accordingly.

packages/perseus/src/widgets/numeric-input/numeric-input.stories.tsx
```diff
@@ -1,99 +1,318 @@
...
+            {
+                status: "correct",
+                maxError: null,
+                strict: false,
+                value: 1252,
+                simplify: "required",
+                message: "",
+            },
+        ],
+        // We're including all the answer forms to make it easier to edit in storybook.
```
mark-fitzgerald, line 62: What about "proper" and "improper"?

SonicScrewdriver: Whoops thank you!

PR #2139, opened 2025-01-22
Create helper to build public widget options for expression

Myranae:
## Summary:
Adds a function that takes expression's full widget options and filters out answer data. It also adds this function to the widget's widget export and adds a test confirming the function does what is expected.

Issue: LEMS-2758

## Test plan:
- Confirm all checks pass
- Confirm expression still works as expected


packages/perseus/src/widgets/expression/expression.util.ts
```diff
@@ -0,0 +1,36 @@
...
+ * PerseusOrdererWidgetOptions type
+ */
+type ExpressionPublicWidgetOptions = {
+    buttonSets: LegacyButtonSets;
+    functions: ReadonlyArray<string>;
+    times: boolean;
+    visibleLabel?: string;
+    ariaLabel?: string;
+    buttonsVisible?: "always" | "never" | "focused";
+};
```
Myranae, line 17: Left out answerFrom from the below widget options:
```export type PerseusExpressionWidgetOptions = {
    // The expression forms the answer may come in
    buttonSets: LegacyButtonSets;
    // Variables that can be used as functions.  Default: ["f", "g", "h"]
    // Use x for rendering multiplication instead of a center dot.
    times: boolean;
    // visible label associated with the MathQuill field
    visibleLabel?: string;
    // aria label for screen readers attached to MathQuill field
    ariaLabel?: string;
    // Controls when buttons for special characters are visible when using a
    // desktop browser.  Defaults to "focused".
    // NOTE: This isn't listed in perseus-format.js or perseus_data.go, but
    // appears in item data in the datastore.
    buttonsVisible?: "always" | "never" | "focused";
```

And here is PerseusExpressionAnswerForm. I don't think I need to pick out value from this, but if someone could confirm for me, that would be great.
```
value: string     
form: boolean     
simplify: boolean     
considered: typeof PerseusExpressionAnswerFormConsidered[number]     
key?: string 
```

packages/perseus/src/widgets/expression/expression.util.test.ts
```diff
@@ -0,0 +1,23 @@
+import {expressionItem2} from "./expression.testdata";
+import getExpressionPublicWidgetOptions from "./expression.util";
+
+import type {PerseusExpressionWidgetOptions} from "@khanacademy/perseus-core";
+
+describe("getExpressionPublicWidgetOptions", () => {
+    it("should return the correct public options without any answer data", () => {
+        // Arrange
+        const options: PerseusExpressionWidgetOptions =
+            expressionItem2.question.widgets["expression 1"].options;
```
benchristel, line 10: It looks like `visibleLabel` and `ariaLabel` are not included in the test data? I'd prefer to define the input inline rather than relying on shared data from `expression.testdata`.

Sharing data between tests makes them harder to understand (you have to look at multiple files to understand the test setup) and prone to false passes if the test data changes later.

```suggestion
        const options: PerseusExpressionWidgetOptions = {
            buttonSets: ["basic"],
            functions: ["f", "g", "h"],
            times: false,
            visibleLabel: "the visible label",
            ariaLabel: "the aria label",
            buttonsVisible: "always",
        };
```

packages/perseus/src/widgets/expression/expression.util.ts
```diff
@@ -0,0 +1,36 @@
...
+ * PerseusExpressionWidgetOptions type
+ */
+type ExpressionPublicWidgetOptions = {
+    buttonSets: LegacyButtonSets;
+    functions: ReadonlyArray<string>;
+    times: boolean;
+    visibleLabel?: string;
+    ariaLabel?: string;
+    buttonsVisible?: "always" | "never" | "focused";
+};
```
handeyeco, line 17: PublicWidgetOptions will always be a subset of WidgetOptions, so I think it would make sense to derive one from the other.

```suggestion
type ExpressionPublicWidgetOptions = {
    buttonSets: ExpressionWidgetOptions["buttonSets"];
    functions: ExpressionWidgetOptions["functions"];
    times: ExpressionWidgetOptions["times"];
    visibleLabel?: ExpressionWidgetOptions["visibleLabel"];
    ariaLabel?: ExpressionWidgetOptions["ariaLabel"];
    buttonsVisible?: ExpressionWidgetOptions["buttonsVisible"];
};
```

This will get more complicated for nested data. For instance if we're trying to remove `correct` from `options.choices.correct`, we can't do this exactly; but for a lot of this stuff I think we'd be fine.

If someone else (or past me) pitched a reason this wouldn't work, feel free to ignore.

packages/perseus/src/widgets/expression/expression.util.test.ts
handeyeco: I think this makes sense. I know Ben advocates the builder method and I like the generator method, but either way it's nice to have a shared logic to generate an object that can just be overwritten for the parts that are most relevant to the test. For instance we have `generateTestExpressionWidget` in `test-utils.ts` which can give you a skeleton that you can then modify without risking changes to other tests.

handeyeco (approved):
Thanks Tamara

packages/perseus/src/widgets/expression/expression.util.ts
handeyeco: Yeah, I _think_ we'll just want to pull out `answerForms` entirely, so we won't need to modify `PerseusExpressionAnswerForm`

packages/perseus/src/widgets/expression/expression.util.test.ts
Myranae: Is the previously created test data just something to ignore moving forward then? When should it be used? Will update to one of the above methods for sure though 😅

Myranae: Updated :)

packages/perseus/src/widgets/expression/expression.util.ts
Myranae: Updated :)

jeremywiebe (approved):

PR #2143, opened 2025-01-23
Move Passage widgets default props

handeyeco:
## Summary:
Tidied up some versions and move Passage widget stuff that we'll need to be able to use on the server.

- Passage
- PassageRef
- PassageRefTarget

Issue: LEMS-2737

## Test plan:
Nothing should functionally change, just moving code


packages/perseus/src/widgets/interactive-graphs/interactive-graph-question-builder.ts
```diff
@@ -91,10 +91,6 @@ class InteractiveGraphQuestionBuilder {
                         lockedFigures: this.lockedFigures,
                     },
                     type: "interactive-graph",
-                    version: {
-                        major: 0,
-                        minor: 0,
-                    },
```
handeyeco, line 97: 1. `version` is optional
2. Version 0.0.0 won't get upgraded

packages/perseus/src/widgets/mock-widgets/mock-asset-loading-widget.tsx
```diff
@@ -18,7 +18,6 @@ export const mockedAssetItem: PerseusItem = {
                 static: false,
                 graded: true,
                 options: Object.freeze({value: ""}),
-                version: {major: 1, minor: 0},
```
handeyeco, line 21: I think this was a mistake. This is a new widget, I don't think we've moved from 0 to 1 already. Also I don't think we need to upgrade a mock widget.

packages/perseus/src/widgets/passage-ref-target/passage-ref-target.tsx
```diff
@@ -67,7 +67,6 @@ export default {
     transform: (editorProps: any): any => {
         return _.pick(editorProps, "content");
     },
-    version: {major: 0, minor: 0},
```
handeyeco, line 70: There's no point in having a 0.0 version I don't think since `version` itself is optional.

packages/perseus/src/widgets/interactive-graphs/interactive-graph-question-builder.ts
jeremywiebe: Pedantically, widgets [default to v0.0.0](https://github.com/Khan/perseus/blob/7a984eba6f1cec3df314ec245d2176f5db190548/packages/perseus/src/widgets.ts#L232). Data missing a version also defaults to v0.0.0 and so does not need an upgrade.

packages/perseus/src/widgets/mock-widgets/mock-asset-loading-widget.tsx
jeremywiebe: Agreed. Looking at `widgets.ts` we have some [protective code](https://github.com/Khan/perseus/blob/7a984eba6f1cec3df314ec245d2176f5db190548/packages/perseus/src/widgets.ts#L239) that just skips if the data (v1.0) is _newer_ than the widget's version (v0.0).

But agree it's safe to remove this.

jeremywiebe (approved):

PR #2144, opened 2025-01-23
Batch 1 of simple default widget option transfers

handeyeco:
## Summary:
I couldn't justify doing a PR per widget because all of these are really straight-forward; there's no prop upgrades or versions or anything. It's just moving "default props" from editors into "default widget options" in core so we can do minor upgrades on the server.

[LEMS-2737/default-collection] matcher
[LEMS-2737/default-collection] interaction
[LEMS-2737/default-collection] input-number
[LEMS-2737/default-collection] image
[LEMS-2737/default-collection] iframe
[LEMS-2737/default-collection] group
[LEMS-2737/default-collection] graded-group
[LEMS-2737/default-collection] explanation
[LEMS-2737/default-collection] dropdown
[LEMS-2737/default-collection] definition
[LEMS-2737/default-collection] cs-program
[LEMS-2737/default-collection] categorizer

Issue: LEMS-2737

## Test plan:


packages/perseus-core/src/widgets/image/index.ts
```diff
@@ -0,0 +1,31 @@
...
+    "title" | "range" | "box" | "backgroundImage" | "labels" | "alt" | "caption"
+>;
+
+const defaultBoxSize = 400;
+const defaultRange = [0, 10] as any;
+
+const defaultWidgetOptions: ImageDefaultWidgetOptions = {
+    title: "",
+    range: [defaultRange, defaultRange],
+    box: [defaultBoxSize, defaultBoxSize],
```
jeremywiebe, line 15: I suspect this is how it was defined originally, but it would be much easier to read this if these weren't variables.

jeremywiebe (approved):

PR #2146, opened 2025-01-23
[LEMS-2737/default-grapher] untangle grapher

handeyeco:
## Summary:
<Add your summary here>

Issue: LEMS-2737

## Test plan:

### DIFF
--- a/packages/perseus-core/src/utils/grapher-default.ts
+++ b/packages/perseus-core/src/utils/grapher-default.ts
@@ -1,0 +1,88 @@
+import {
+    functionForType,
+    maybePointsFromNormalized,
+    snapStepFromGridStep,
+    type FunctionTypeMappingKeys,
+} from "./grapher-util";
+
+import type {Coord} from "../data-schema";
+
+/* Given a plot type, return the appropriate default value for a grapher
+ * widget's plot props: type, default coords, default asymptote. */
+export const defaultPlotProps = (
+    type: FunctionTypeMappingKeys,
+    graph: any,
+): any => {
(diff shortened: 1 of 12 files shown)


packages/perseus-core/src/utils/grapher-default.ts
```diff
@@ -0,0 +1,88 @@
...
+    // therefore be considered empty and ineligible for grading. The user
+    // *can* move the coords from the default position and then back if
+    // they really want to submit the default coords as their answer, but
+    // we currently don't write questions that require this.
+    //
+    // We *do* write questions in which the asymptote should be left in
+    // the default position. For this reason, we fill in the default
+    // asymptote rather than leaving it null; if the user moves the coords
+    // but not the asymptote, the widget is non-empty and eligible for
+    // grading.
```
jeremywiebe, line 27: This feels related to a discussion we had in huddle recently. It'd be nice if we separated out the widget options used to render from user input state. That would resolve this almost entirely (when user interacts, we build a user input object... if it's empty, user hasn't interacted). Having the widget options transformed to props and then intermingled with user input state is so challenging.

packages/perseus/src/util.ts
```diff
@@ -270,10 +270,6 @@ function getGridStep(
     });
 }
 
-function snapStepFromGridStep(gridStep: [number, number]): [number, number] {
```
jeremywiebe, line 273: Nice. I've always thought this `util.ts` file was too much of a grab-bag of stuff.  Nice to start pulling things apart into more focused files.

jeremywiebe (approved):

PR #2147, opened 2025-01-23
Move default editor props for NumberLine

handeyeco:
## Summary:
This one is a little tricky because it's called `showTooltip` in the data schema but was called `showTooltips` in the editor. I think it was a bug? So I tried to strengthen the types a little.

Issue: LEMS-2737

## Test plan:
NumberLine should continue to be editable and the widgets the editor makes should be rendered correctly.


packages/perseus-core/src/widgets/number-line/index.ts
```diff
@@ -0,0 +1,44 @@
...
+    divisionRange: [1, 12],
+    numDivisions: 5,
+    snapDivisions: 2,
+
+    tickStep: null,
+    correctRel: "eq",
+    correctX: null,
+    initialX: null,
+
+    showTooltip: false,
```
handeyeco, line 36: So this was originally `showTooltips`, but in `PerseusNumberLineWidgetOptions` that's `showTooltip`. I think this might have been a bug of some sort due to lack of typing? So I was hoping someone could double check my thinking.

packages/perseus-editor/src/widgets/number-line-editor.tsx
```diff
@@ -15,52 +20,32 @@ const bound = (x: number, gt: number, lt: number): number =>
 
 const EN_DASH = "\u2013";
 
-type Props = any;
```
handeyeco, line 18: Since I was concerned with the type, I converted PropTypes to TS

packages/perseus-editor/src/widgets/number-line-editor.tsx
```diff
@@ -389,9 +374,9 @@ class NumberLineEditor extends React.Component<Props> {
                     {!this.props.static && (
                         <Checkbox
                             label="Show tooltips"
-                            checked={this.props.showTooltips}
+                            checked={this.props.showTooltip}
                             onChange={(value) => {
-                                this.props.onChange({showTooltips: value});
+                                this.props.onChange({showTooltip: value});
```
handeyeco, line 379: This makes me nervous.

packages/perseus-editor/src/widgets/__tests__/number-line-editor.test.tsx
```diff
@@ -117,7 +117,7 @@ describe("number-line-editor", () => {
             screen.getByRole("checkbox", {name: "Show tooltips"}),
         );
 
-        expect(onChangeMock).toBeCalledWith({showTooltips: true});
+        expect(onChangeMock).toBeCalledWith({showTooltip: true});
```
handeyeco, line 120: Nervous about this too

packages/perseus-core/src/widgets/number-line/index.ts
jeremywiebe: The prop that the `NumberLine` component receives is `showTooltips` though, and the plural form is used in the React code, so my thought is that the data schema and other instances that reference `showTooltip` are wrong. 

image

image

I grepped through the entire content bucket and I can't find instances of `number-line` using the singular `showTooltip` either.

@benchristel I see the [parser](https://github.com/Khan/perseus/blob/bc3d955b57e847a379328fcc7cf276f42e0874dd/packages/perseus/src/util/parse-perseus-json/perseus-parsers/number-line-widget.ts#L43) uses singular, so I wonder if because it's marked as `optional()`, it succeeded (ie. it never saw a value, but passed because it's optional). We should fix the parser here too.

packages/perseus-editor/src/widgets/number-line-editor.tsx
```diff
@@ -15,52 +20,32 @@ const bound = (x: number, gt: number, lt: number): number =>
...
+    correctRel: "lt" | "gt" | "le" | "ge" | "eq";
+    correctX: number;
+    initialX: number;
+    isTickCtrl?: boolean;
 
-        static: PropTypes.bool,
-        showTooltips: PropTypes.bool,
-    };
+    static?: boolean;
+    showTooltip: boolean;
```
jeremywiebe, line 41: Note this should be plural.

packages/perseus-editor/src/widgets/number-line-editor.tsx
jeremywiebe: Rightly. :)  It should be singular as noted above. :)

jeremywiebe (approved):

PR #2148, opened 2025-01-23
Plotter defaults

handeyeco:
## Summary:
This actually includes a logic change, `get picUrl` seemed over-engineered and unnecessary.

Issue: LEMS-2737

## Test plan:
Plotter can be edited, Plotters edited with this change can be rendered.


packages/perseus-editor/src/widgets/plotter-editor.tsx
```diff
@@ -89,30 +71,8 @@ const formatNumber = (num) => "$" + knumber.round(num, 2) + "$";
...
-
-        // @ts-expect-error - TS2322
-        get picUrl() {
-            const staticUrl = Dependencies.getDependencies().staticUrl;
-            if (staticUrl) {
-                return staticUrl("/images/badges/earth-small.png");
-            }
-
-            return null;
-        },
```
handeyeco, line 114: I just felt like this was a little much just to have a placeholder image.

packages/perseus-core/src/widgets/plotter/index.ts
```diff
@@ -0,0 +1,45 @@
...
+    correct: [1],
+    starting: [1],
+
+    type: "bar",
+    labels: ["", ""],
+    categories: [""],
+
+    picSize: 30,
+    picBoxHeight: 36,
+    plotDimensions: [275, 200],
```
jeremywiebe, line 34: Woah, that's some random size. 😂  Our graphs are usually 400x400. 

No action needed.

packages/perseus-editor/src/widgets/plotter-editor.tsx
jeremywiebe: I've always really disliked that default props has behaviour embedded in it via a getter! 

What does the Plotter show by default in the editor with this now defaulting to `null`.

jeremywiebe (approved):

packages/perseus-editor/src/widgets/plotter-editor.tsx
handeyeco: Turns out this was kind of a hard question to answer.

1. I added a story, but SB didn't properly resolve the URL for the image; it looked the same in `main` and this branch
2. I tried to look in the content editor, but Plotter is a hidden widget so I couldn't make new widget to test

I think I might just merge this PR since the widget is hidden anyway. Shouldn't change the editing experience since they'll never see an empty Plotter widget right?

jeremywiebe: > I think I might just merge this PR since the widget is hidden anyway. Shouldn't change the editing experience since they'll never see an empty Plotter widget right?

Good point!  For hidden widgets, they can never be "created" in the editor anyways, so defaults for these widgets feel like a fading concern.

PR #2149, opened 2025-01-23
Table defaults

handeyeco:
## Summary:
Move Table's default widget options

Issue: LEMS-2737

## Test plan:
Mostly just moving code. Swap out some logic with some more modern JS.


packages/perseus-core/src/widgets/table/index.ts
```diff
@@ -0,0 +1,30 @@
...
+>;
+
+const defaultRows = 4;
+const defaultColumns = 1;
+
+// initialize a 2D array
+// (defaultRows x defaultColumns) of empty strings
+const answers = new Array(defaultRows)
+    .fill(0)
+    .map(() => new Array(defaultColumns).fill(""));
```
handeyeco, line 16: I think this is functionally equivalent as:

``` TS
function stringArrayOfSize(size: number): ReadonlyArray<string> {
    return _(size).times(function () {
        return "";
    });
}

const blankAnswers = _(defaultRows).times(function () {
    return Util.stringArrayOfSize(defaultColumns);
});
```

And this way I don't have to use `_` or move a util.

jeremywiebe: Looks synonymous, yeah.  👍

jeremywiebe (approved):

PR #2150, opened 2025-01-23
InteractiveGraph defaults

handeyeco:
## Summary:
This one is a little tricky. I'll add comments.

Issue: LEMS-2737

## Test plan:
Interactive Graph should be editable and edited widgets should be renderable.


packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
```diff
@@ -178,15 +164,11 @@ class InteractiveGraphEditor extends React.Component<Props> {
...
-        ...InteractiveGraph.defaultProps,
+    static defaultProps: InteractiveGraphDefaultWidgetOptions & {
+        valid: boolean;
+    } = {
+        ...interactiveGraphLogic.defaultWidgetOptions,
         valid: true,
-        backgroundImage: defaultBackgroundImage,
-        showTooltips: false,
-        correct: {
-            type: InteractiveGraph.defaultProps.graph.type,
```
handeyeco, line 187: Rather than depending on the default props from the IG widget, I copied those values. I think that's fine? Alternatively I could update IG to pull from this file? `perseus-core` can't depend on `perseus`.

Ultimately though I think this should be fine?

packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
```diff
@@ -178,15 +164,11 @@ class InteractiveGraphEditor extends React.Component<Props> {
     displayName = "InteractiveGraphEditor";
     className = "perseus-widget-interactive-graph";
 
-    static defaultProps: DefaultProps = {
-        ...InteractiveGraph.defaultProps,
+    static defaultProps: InteractiveGraphDefaultWidgetOptions & {
+        valid: boolean;
+    } = {
+        ...interactiveGraphLogic.defaultWidgetOptions,
```
handeyeco, line 170: I think this might be a mistake from a different PR. `defaultProps` for all the other widget editors is a subset of widget options - which makes sense since the editors are for generating widget options. This one uses `valid` though, which isn't part of `PerseusInteractiveGraphWidgetOptions`.

When doing these kind of large-scale reorgs, I try not to be too opinionated; so I kept it in a way that I think highlights that it's not part of widget options.

jeremywiebe: I bet it's state again. However, isn't it `boolean | string`?

image

jeremywiebe: I think it originates from the `InteractiveGraphSettings` component: https://github.com/Khan/perseus/blob/5173c2e43bf939159f420dcd448b90691d52353b/packages/perseus-editor/src/widgets/interactive-graph-editor/components/interactive-graph-settings.tsx#L436

packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
jeremywiebe: I think that's good. The default for the editor is a much different thing that the default for runtime anyways. In fact, I'd hope we _never_ have data that's published where `type` is missing... that feels like invalid data for learner-facing content.

jeremywiebe (approved):

packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
handeyeco: That makes me think it's actually `valid: true | string` right?

> // a string message, not false

PR #2151, opened 2025-01-23
[shifting-static] docs(changeset): Moving static function for Numeric

SonicScrewdriver:
## Summary:

All this PR does is move a single static function to a util file. 

Note for Mark: I think I would recommend that we simply set this ticket to "Won't Do", as there's not a lot to be gained here anymore! We would need to essentially duplicate a single, very simple, function as it seems that `getUserInputFromProps` is also called from the renderer, even though `getUserInput` does the exact same thing. 

I'll keep the PR as-is for reference. 

Issue: LEMS-2444

## Test plan:
- tests pass

### DIFF
--- a/.changeset/gorgeous-dolls-promise.md
+++ b/.changeset/gorgeous-dolls-promise.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Moving static function for Numeric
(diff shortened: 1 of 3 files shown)


mark-fitzgerald:
Agreed. The value proposition has decreased significantly. Thanks for looking into it.

PR #2152, opened 2025-01-23
Update locations of public widget options functions: categorizer, orderer, and expression

Myranae:
## Summary:
Moves the util files related to creating public widget options for categorizer, orderer, and expression into their respective widget folders in the perseus-core package. Previously they were located within their respective widget folders in the perseus package.

Issue: none

## Test plan:
- Confirm all checks pass
- Confirm widgets act as expected


benchristel (approved):
LGTM!

PR #2153, opened 2025-01-24
Rename all instances of the term ScoringData with Rurbic

jeremywiebe:
## Summary:

TLDR: lots of files changed, but this is a straightforward rename of a bunch of symbols. 

Not long ago we decided to rename the term `Rubric` with `ScoringData` because we felt `ScoringData` better captured the spirit of the data that will be removed for Server-Side Scoring.

However, we aren't able to rename every instance of "Rubric" in the codebase and so we were left with both terms. I'm reverting us back to use "Rubric" for now. If we decide its a worthwhile change, we can re-introduce it at a later point in time. 

I made this change using the following [Fish shell](http://fishshell.com) script: 

```fish
for f in (rg -i ScoringData -l) 
    sed -i '' 's/ScoringData/Rubric/g; s/scoringData/rubric/g;' $f 
end
```

Issue: "none"

## Test plan:

`yarn typecheck`
`yarn test`


.changeset/silent-coats-check.md
```diff
@@ -0,0 +1,6 @@
+---
+"@khanacademy/perseus": major
+"@khanacademy/perseus-score": major
```
Myranae, line 3: What makes this a major change? The extent of the change?

Myranae (approved):
Looks good! The only place I could still find "scoringdata" in any form was in changesets. Looks like it might need a prettier format for the checks to pass, but thanks for handling this! Going to look into Fish Invocations; seems handy :)

.changeset/silent-coats-check.md
jeremywiebe: We are changing the name of types that are exported in a non-backwards compatible way. According to semver, this is a major change:

> 1. MAJOR version when you make incompatible API changes
> 2. MINOR version when you add functionality in a backward compatible manner
> 3. PATCH version when you make backward compatible bug fixes

jeremywiebe:
> Looks good! The only place I could still find "scoringdata" in any form was in changesets. Looks like it might need a prettier format for the checks to pass, but thanks for handling this! Going to look into Fish Invocations; seems handy :)


I should be a bit clearer (I've updated the description). the "Fish Invocation" is a [Fish Shell](https://fishshell.com) script I ran from the command-line. I use the Fish Shell as my daily terminal shell. It has simpler syntax and less foot-guns than Bash, but it comes at the expense of not being the same as Bash.

PR #2154, opened 2025-01-24
Create helper to build public widget options for Sorter

Myranae:
## Summary:
Adds a function that takes Sorter's full widget options and filters out answer data. It also adds this function to Sorter's widget export and adds a test confirming the function does what is expected.

Issue: LEMS-2765

## Test plan:
- Confirm all checks pass
- Confirm Sorter still works as expected


packages/perseus-core/src/widgets/sorter/sorter-util.test.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import type {PerseusSorterWidgetOptions} from "../../data-schema";
+
+describe("getSorterPublicWidgetOptions", () => {
+    it("should return the correct public options without any answer data", () => {
+        // Arrange
+        const options: PerseusSorterWidgetOptions = {
+            correct: ["$0.005$ kilograms", "$15$ grams", "$55$ grams"],
+            layout: "horizontal",
+            padding: true,
```
Myranae, line 11: @benchristel It seems like correct is the source of the card information in this widget, since there are no other fields with that data. Maybe sorting here would work as well? It looks like the first thing the widget does to render is shuffle the correct options anyway:
```
    render(): React.ReactNode {
        const options = shuffle(
            this.props.correct,
            this.props.problemNum as number,
            /* ensurePermuted */ true,
        );
```

benchristel: +1 for sorting. Could we also use `ReadonlyArray` in the public options type? I think that should work.

Myranae: Ah, you're right! Looks like it does. Probably because we do the slice and I just didn't update it. Will fix that, and then there's no issue. Though I was planning on changing the "correct" key to "cards" in the public type; would that be an issue for backwards compatibility between the types?

benchristel: If we change `correct` to `cards`, then the widget will have to be updated to use the `correct` prop when the answer has been revealed and the `cards` prop when it hasn't been. So it would complicate things.

This might be a good excuse to test out the migration facilities in the new perseus JSON parser. That would let us rename `correct` to `cards` on the `PerseusSorterWidgetOptions` type while remaining backwards-compatible with existing data. However, we can't do that until we have confirmed that the parsers are actually working in webapp, and it might be a sprint or two before that's done.

I would also support just adding a comment to the `correct` field to explain the situation.

jeremywiebe: I think this is something that doesn't need to be figured out now. We can shuffle the data for the public widget options, but keep it named `correct` (and document it clearly as Ben suggests).

This means that in different types, `correct` will mean different things (in the public widget options, it's the initial state, in the rubric (full) type, it's the correct answer). 

This isn't great, but I think results in the least churn/uncertainty right now (no logic changes needed). 

I've noted this issue in the [Content Issues for Backfill](https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/3524690040/Content+Issues+for+Backfill) doc.

jeremywiebe (approved):

packages/perseus-core/src/widgets/sorter/sorter-util.test.ts
Myranae: Sounds good! I'll leave it as correct with a note, and we can come back to it later. Thanks so much, everyone!

PR #2155, opened 2025-01-24
Move Perseus JSON parsers to perseus-core

benchristel:
## Summary:
These parsers will need to run in a NodeJS service for server-side
scoring. The server needs to be able to import them without also pulling
in JSX.

Moving them to perseus-core means we can't validate
dynamically-registered widgets as strictly, but that seems like a small
sacrifice given that a) dynamic widgets are not currently used in
production, and b) the validation we can do on them would not be
particularly strict in any case.

Issue: LEMS-2774

Test plan:

`yarn test`


benchristel:
@handeyeco this is kind of jumping the gun on our architecture discussion next week, but I figured I'd put this up as a proposal for what I think we should do.

benchristel:
> **Size Change:** -126 B (-0.01%)
> 
> **Total Size:** 1.48 MB
> 
> Filename	Size	Change	
> `packages/perseus-core/dist/es/index.js`	42.4 kB	+11.9 kB (+39.03%)	🚨
> `packages/perseus/dist/es/index.js`	385 kB	-12 kB (-3.03%)	
> ℹ️ **View Unchanged**

I feel pretty good that our parser infrastructure + all the Perseus-specific parsers are only 200b bigger than [superstruct](https://bundlephobia.com/package/superstruct@2.0.2).

packages/perseus-core/src/parse-perseus-json/perseus-parsers/categorizer-widget.ts
```diff
@@ -11,8 +11,8 @@ import {defaulted} from "../general-purpose-parsers/defaulted";
 
 import {parseWidget} from "./widget";
 
+import type {CategorizerWidget} from "../../data-schema";
```
handeyeco, line 14: It's a two-way door so we don't have to make the change in this PR, but now that we have a folder for each widget from my changes I wonder if we could put individual widget parsers in those folders to better colocate widget logic.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/widgets-map.test.ts
```diff
@@ -75,16 +74,27 @@ describe("parseWidgetsMap", () => {
...
         };
 
         const result = parse(widgetsMap, parseWidgetsMap);
 
+        expect(result).toEqual(success(widgetsMap));
+    });
+
+    it("rejects an unknown widget whose type doesn't match its ID", () => {
+        const widgetsMap: unknown = {
+            "transmogrifier 1": {type: "bork"},
```
handeyeco, line 90: I'm not sure how much we want a test like this. We try to keep our understanding of widget IDs to be opaque blobs with the ultimate goal of having arbitrary IDs in the future (`widget 1` >> `895h-789e-39j2-4359`). In practice we don't do this, but I don't think we want to move towards reinforcing the idea that IDs contain widget types.

packages/perseus/src/index.ts
```diff
@@ -109,11 +109,28 @@ export {
     getImagesWithoutAltData,
 } from "./util/extract-perseus-data";
 export {
+    /**
+     * @deprecated - import this function from perseus-core instead
```
handeyeco, line 113: My vote would be to make this a major release for `perseus` and remove these from the exports.

packages/perseus/src/index.ts
```diff
@@ -190,7 +207,7 @@ export type {
     SharedRendererProps,
 } from "./types";
 export type {ParsedValue} from "./util";
-export type {Result, Success, Failure} from "./util/parse-perseus-json/result";
+export type {Result, Success, Failure} from "../../perseus-core/src/parse-perseus-json/result";
```
handeyeco, line 210: I think this might have been a mistake from your IDE auto-updating imports. We shouldn't be importing from other packages this way. I think our linter for this is broken.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/categorizer-widget.ts
benchristel: I think the proposed change might make sense. But really, the parsers are most tightly coupled to the widget options types in data-schema.ts. Unless we also move those types out of data-schema.ts into the individual widget folders, I'm inclined to keep this as-is for now.

packages/perseus/src/index.ts
benchristel: > Are these the "pre-parsing" types?

Not sure I understand. `parsePerseusItem` is currently used in webapp, and I want to avoid causing any headaches when we do the perseus release this week.

PR #2156, opened 2025-01-24
Create helper to build public widget options for Matcher

Myranae:
## Summary:
Adds a function that takes a Matcher widget's full widget options and removes correct order information. Some additional shuffling logic has been added to allow shuffling to occur server-side.

It also adds this function to the widget's widget export and adds a test confirming the function works as expected.

Issue: LEMS-2761

## Test plan:
- Confirm all checks pass
- Confirm Matcher still works as expected.


packages/perseus-core/src/widgets/matcher/matcher-util.test.ts
```diff
@@ -0,0 +1,52 @@
...
+                "The life cycle of medium-sized stars includes a red giant stage and ends in a whimper as a white dwarf",
+                "Rapid escalation of greenhouse gas emissions",
+                "The current trajectory of the Milky Way galaxy and those in its immediate proximity",
+            ],
+            left: [
+                "Our Sun will run out of fuel and die in around 5 billion years ",
+                "Plate tectonics will rearrange the continents: the Pacific will narrow, bringing Australia closer to the Americas, and the Atlantic will expand to form the largest of the oceans ",
+                "Our Sun will run out of hydrogen, swell into a red giant, gobble up the inner rocky planets, and then collapse and die ",
+                "Average global temperatures will rise ",
+                "In 3 to 4 billion years, our galaxy will begin a slow collision with its closest large neighbor, Andromeda ",
```
Myranae, line 24: Should I simplify these to not be so much text? I just copied the data from the matcher test data file.

packages/perseus-core/src/widgets/matcher/matcher-util.ts
```diff
@@ -0,0 +1,31 @@
+import type {PerseusMatcherWidgetOptions} from "@khanacademy/perseus-core";
+
+/**
+ * For details on the individual options, see the
+ * PerseusMatcherWidgetOptions type
+ */
+type MatcherPublicWidgetOptions = {
+    labels: PerseusMatcherWidgetOptions["labels"];
+    left: Array<string>;
+    right: Array<string>;
```
Myranae, line 10: Will this be a problem to still work with the original widget options? @jeremywiebe, I remember you talking about how it should be backward compatible in a way so that the public widget options can still pass where the widget options are expected, or something like that.

packages/perseus-core/src/widgets/matcher/matcher-util.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Given a PerseusMatcherWidgetOptions object, return a new object with only
+ * the public options that should be exposed to the client.
+ */
+function getMatcherPublicWidgetOptions(
+    options: PerseusMatcherWidgetOptions,
+): MatcherPublicWidgetOptions {
+    return {
+        labels: options.labels,
+        left: options.left.slice().sort(),
+        right: options.right.slice().sort(),
```
Myranae, line 25: Thoughts on a different/better way to remove order information?

packages/perseus-core/src/widgets/matcher/matcher-util.test.ts
benchristel: +1. I think smaller data in tests is better. It makes the tests easier to read and change.

packages/perseus-core/src/widgets/matcher/matcher-util.ts
benchristel: If you use `ReadonlyArray` instead of `Array` here, do you get a type error? If you can use `ReadonlyArray`, I agree that we might want to prefer that for compatibility with the other types.

packages/perseus-core/src/widgets/matcher/matcher-util.test.ts
jeremywiebe: Agree with Ben.  Unless the quantity of data is needed for the test, let's trim it down. It doesn't even have to be actual real data. I know when we started writing tests for each widget, we pulled actual assessment items so that we'd know what the data looked like.  Today, I don't think we need to do that anymore.

packages/perseus-core/src/widgets/matcher/matcher-util.ts
jeremywiebe: Where possible, I like using `ReadonlyArray<>` because these types represent data that Perseus will receive. The idea that it could be mutable and mutated should be private to the widget. So yes, I'd agree with Ben to try switching these to `ReadonlyArray<>`.  If that causes type issues, then I think we should look to fix the function code and not the type.

packages/perseus-core/src/widgets/matcher/matcher-util.ts
jeremywiebe: 🤔 So we [shuffle these options](https://github.com/Khan/perseus/blob/29a1c656ee7f74b6eba8ce95fa9c239b2f328813/packages/perseus/src/widgets/matcher/matcher.tsx#L164-L172) when the widget renders the options. 

I'm inclined to extract that logic out to a helper for matcher and use it here for now. This means that the rubric type and this public options type will have a property that is named the same, but is conceptually different (one defines the correct answer, the other is the initial render state).

jeremywiebe (approved):
Nice. Please check my comments as one of them (how to remove the order information) should be resolved before landing. Thanks!

packages/perseus-core/src/widgets/matcher/matcher-util.test.ts
Myranae: Shortened! :)

packages/perseus-core/src/widgets/matcher/matcher-util.ts
Myranae: Switched back to reference the original widget options type :)

packages/perseus-core/src/widgets/matcher/matcher-util.ts
Myranae: It looks like the shuffle and seededRNG utils are part of the Perseus package. I don't think we want to use part of the Perseus package in Perseus Core, so I'm not sure I'll be able to pull it out to a helper function without copying a lot of logic. I might be missing something though.

jeremywiebe: Extracting the `seededRNG()` and `shuffle()` functions to `perseus-core` would be fine. I think we'll find more and more of our "core logic" moving into `perseus-core` and `perseus` becoming more and more a UI-only library.

packages/perseus-core/src/widgets/matcher/matcher-util.ts
```diff
@@ -0,0 +1,61 @@
...
+
+type MatcherInfo = {
+    left: ReadonlyArray<string>;
+    right: ReadonlyArray<string>;
+    orderMatters: boolean;
+    problemNum: number | null | undefined;
+};
+
+export const matcherShuffle = (
+    props: MatcherInfo,
```
Myranae, line 13: I could instead update to use the original widget options type, but it seemed like specifically listing which items are needed would be more useful, but I could change to be the below code instead.
```
props: PerseusMatcherWidgetOptions & {
        problemNum: number | null | undefined;
    }
```

packages/perseus-core/src/widgets/matcher/matcher-util.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    let left;
+    if (!props.orderMatters) {
+        // If the order doesn't matter, don't shuffle the left column
+        left = props.left;
+    } else {
+        left = CoreUtil.shuffle(props.left, rng, /* ensurePermuted */ true);
+    }
+
+    const right = CoreUtil.shuffle(props.right, rng, /* ensurePermuted */ true);
```
benchristel, line 26: This `ensurePermuted = true` actually leaks information about the correct answer. We are revealing that the correct answer is _not_ the permutation sent to the client. A cheater could answer the question with better than chance accuracy by choosing some other permutation at random.

Fun fact: a variant of this type of information leak [allowed the Allies to break the Enigma cipher in World War II](https://en.wikipedia.org/wiki/Cryptanalysis_of_the_Enigma)!

> A major weakness of the system, however, was that no letter could be enciphered to itself.

packages/perseus-core/src/widgets/matcher/matcher-util.ts
```diff
@@ -0,0 +1,61 @@
...
+};
+
+/**
+ * Given a PerseusMatcherWidgetOptions object, return a new object with only
+ * the public options that should be exposed to the client.
+ */
+function getMatcherPublicWidgetOptions(
+    options: PerseusMatcherWidgetOptions,
+): MatcherPublicWidgetOptions {
+    const {left, right} = matcherShuffle({...options, problemNum: undefined});
```
benchristel, line 50: Why do we force `problemNum` to be undefined here?

benchristel (commented):
Cool! The code looks good in general. ~~I don't fully understand why we need to use the seeded random shuffle when getting the public options, though. It seems like sorting the "right" and "left" arrays would conceal the answer just as well, and would be simpler.~~ EDIT: just saw Jeremy's comment.

packages/perseus-core/src/widgets/matcher/matcher-util.ts
benchristel: I think what you have is okay! I might lean toward passing the `seed` as a separate argument instead of adding a `problemNum` field to the options object.

```
export const matcherShuffle = (
  matcherOptions: PerseusMatcherWidgetOptions,
  seed: number,
) => // ...
```

Naming the parameter `problemNum` makes it seem like it has something inherently to do with the `problemNum` passed to the Renderer. Really, though, it's just the seed for the random number generator, which can be any arbitrary number (as evidenced by the rubric stripping code where you hardcode it). `matcherShuffle` is only concerned with shuffling, so its parameters should be at that level of abstraction.

packages/perseus-core/src/widgets/matcher/matcher-util.ts
```diff
@@ -0,0 +1,61 @@
...
+    right: ReadonlyArray<string>;
+    orderMatters: boolean;
+    problemNum: number | null | undefined;
+};
+
+export const matcherShuffle = (
+    props: MatcherInfo,
+): {left: ReadonlyArray<string>; right: ReadonlyArray<string>} => {
+    // Use the same random() function to shuffle both columns sequentially
+    const rng = CoreUtil.seededRNG(props.problemNum as number);
```
benchristel, line 16: Instead of doing an unsafe cast of `problemNum` to `number` here, could we actually provide a default?

```suggestion
    const rng = CoreUtil.seededRNG(props.problemNum ?? 0);
```

`undefined` becomes `0` when converted to an integer using bitwise operations like `<<`, so I'm pretty sure defaulting to 0 preserves behavior.

packages/perseus-core/src/widgets/matcher/matcher-util.ts
benchristel: Hmm, I see that this line was copied from the component. I think we should keep it as-is to avoid changing the UX.

packages/perseus/src/widgets/matcher/matcher.tsx
```diff
@@ -158,18 +160,7 @@ export class Matcher extends React.Component<Props, State> implements Widget {
...
-        let left;
-        if (!this.props.orderMatters) {
-            // If the order doesn't matter, don't shuffle the left column
-            left = this.props.left;
-        } else {
-            left = shuffle(this.props.left, rng, /* ensurePermuted */ true);
-        }
-
-        const right = shuffle(this.props.right, rng, /* ensurePermuted */ true);
+        const {left, right} = matcherShuffle(this.props);
```
benchristel, line 163: On third thought, I think the interaction of this (clientside) shuffle with the serverside shuffle during rubric stripping will cause a regression. Since we `ensurePermuted` when shuffling the columns serverside, shuffling again on the client may result in the original (correct) permutation being shown to the learner.

To resolve this, I think we might want to avoid shuffling on the client, and just use the `left` and `right` props as-is. But then we'll have to figure out something to use as the seed on the server (where `problemNum` isn't available) so the shuffled order isn't the same every time.

@jeremywiebe does this agree with your understanding of how this widget should work?

packages/perseus-core/src/widgets/matcher/matcher-util.ts
Myranae: Hmm, this logic was pulled directly from the widget. If I change it to use a seed, then I would end up changing the widget's logic. Inside the widget, it is using the problemNum. When used with the public widget options logic, it wouldn't be though. I think if we keep this shuffle logic pulled out, we probably shouldn't change it.

packages/perseus-core/src/widgets/matcher/matcher-util.ts
Myranae: The shuffle usage within the widget uses an actual problemNum, so I just use undefined here to provide the value to the function and get it to shuffle for the public widget options.

packages/perseus-core/src/widgets/matcher/matcher-util.ts
Myranae: The cast was already there. I'm hesitant to change widget logic during this project.

packages/perseus/src/widgets/matcher/matcher.tsx
Myranae: I feel like it was simpler the way it was before and would like to switch back to that if possible. Where this widget shuffling logic was left in the widget and we were just sorting. If that still works, it seems much simpler.

jeremywiebe: Ok.  Sorry it took so long to reply!  

I was looking at this more closely. Randomizing these lists at any point other than when we "publish" the item is going to force us to have a synchronized random number generator on the client and server... and like Ben eluded to, that could be really tricky. 

I wonder if the right solution is to do the randomization when we generate the public widget options. This would alleviate any Perseus JSON "schema" changes (other than the new public widget options). 

I think this would mean some scoring adjustments but I don't think they are too complicated and they alleviate the need to randomize at any other place. 

This does presume that having the same starting order every time the question is posed is acceptable (it might not be!). 

I feel like an in-person chat might be helpful here. Let's pair/huddle tomorrow about this.

packages/perseus-core/src/utils/util.ts
```diff
@@ -0,0 +1,76 @@
+/* Note(tamara): Brought over from the perseus package packages/perseus/src/util.ts file.
+May be useful to bring other perseus package utilities here. */
```
Myranae, line 2: The functions here are still used in other widgets as well, specifically categorizer, radio, and sorter, so I figured I should leave them here for now.

packages/perseus-core/src/widgets/matcher/matcher-util.test.ts
```diff
@@ -0,0 +1,54 @@
...
+        // Assert
+        expect(publicWidgetOptions).toEqual({
+            labels: ["**Claims**", "**Evidence**"],
+            padding: true,
+            orderMatters: true,
+            right: ["Milky Way", "Stars", "Life", "Earth", "Rapid"],
+            left: ["Billion", "Fuel", "Hydrogen", "Plate", "Average"],
+        });
+        expect(publicWidgetOptions.right).not.toEqual(options.right);
+        expect(publicWidgetOptions.left).not.toEqual(options.left);
```
Myranae, line 52: I added these expect lines to emphasize what we are looking for in this test. I think these check what we need to check regarding making sure the public widget options don't have the same order as the original ones, but please let me know if there's another test I can add for better coverage!

packages/perseus-core/src/widgets/matcher/matcher-util.ts
```diff
@@ -0,0 +1,138 @@
...
+    problemNum: number | null | undefined;
+};
+
+type MatcherShuffleInfo = {
+    left: ReadonlyArray<string>;
+    right: ReadonlyArray<string>;
+    orderMatters: boolean;
+};
+
+// Note: Previously called cyrb128
```
Myranae, line 19: @benchristel Should I link to that stack overflow page in the code? Not sure about attribution within code 🤔

packages/perseus-core/src/widgets/matcher/matcher-util.ts
```diff
@@ -0,0 +1,138 @@
...
+    // Use the same random() function to shuffle both columns sequentially
+    let left;
+    if (!data.orderMatters) {
+        // If the order doesn't matter, don't shuffle the left column
+        left = data.left;
+    } else {
+        left = CoreUtil.shuffle(data.left, seed, /* ensurePermuted */ true);
+    }
+
+    const right = CoreUtil.shuffle(data.right, seed, /* ensurePermuted */ true);
```
Myranae, line 100: @benchristel Is there still a concern with these `ensurePermuted` arguments leaking data? Is that something we can fix?

packages/perseus-core/src/utils/util.ts
jeremywiebe, line 1: This might be a good opportunity to make this filename more focused and avoid the dreaded "util file that's 10,000 lines of unrelated functionality" problem. :) 

Maybe `random_utils.ts`?

In that case, we could also update your `Note` that you added to be a comment describing what types of functions this file has (and should have).

packages/perseus-core/src/widgets/matcher/matcher-util.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+describe("getMatcherPublicWidgetOptions", () => {
+    it("should return shuffled right array when order doesn't matter", () => {
+        // Arrange
+        const options: PerseusMatcherWidgetOptions = {
+            labels: ["**Claims**", "**Evidence**"],
+            padding: true,
+            orderMatters: false,
+            right: ["Stars", "Earth", "Life", "Rapid", "Milky Way"],
+            left: ["Fuel", "Plate", "Hydrogen", "Average", "Billion"],
```
jeremywiebe, line 13: **nit**: This is probably OCD of the highest level, but I'd order this with `left` first as we read right-to-left in English and so I'd expect `left` before `right`.

packages/perseus-core/src/widgets/matcher/matcher-util.test.ts
```diff
@@ -0,0 +1,54 @@
...
+            labels: ["**Claims**", "**Evidence**"],
+            padding: true,
+            orderMatters: false,
+            right: ["Milky Way", "Stars", "Life", "Earth", "Rapid"],
+            left: ["Fuel", "Plate", "Hydrogen", "Average", "Billion"],
+        });
+        expect(publicWidgetOptions.right).not.toEqual(options.right);
+    });
+
+    it("should return shuffled left and right arrays when order matters", () => {
```
jeremywiebe, line 30: I read your Github comment further down this test. Given that, I wonder if we could be more accurate with this in saying `should not return the left or right arrays in answer order`?

packages/perseus-core/src/widgets/matcher/matcher-util.ts
jeremywiebe, line 19: You could just make it part of the doc comment. :) 

**nit**: Also, more and more I find myself questioning the value of the `get`/`calculate`/`compute` prefixes we tend to give function names. 

* `getHash(myString)`
* `hash(myString)`

I _think_ the second is just as clear. What do you think? (not asking for a change unless this resonates with you...)

```suggestion
/**
 * Returns the hash value of the given input string using the CYRB128 hashing function.
 *
 * Using the implementation found at: https://stackoverflow.com.... 
 */
```

packages/perseus-core/src/widgets/matcher/matcher-util.ts
jeremywiebe: Agree. Cryptographically it's leaking information, but I think we don't want to ever render the correct answer as the starting point.

packages/perseus-core/src/widgets/matcher/matcher-util.ts
```diff
@@ -0,0 +1,138 @@
...
+        a = b ^ (b >>> 9);
+        b = (c + (c << 3)) | 0;
+        c = (c << 21) | (c >>> 11);
+        c = (c + t) | 0;
+        return (t >>> 0) / 4294967296;
+    };
+}
+
+// TODO(LEMS-2841): Should be able to remove once getPublicWidgetOptions is hooked up
+export const matcherShuffle = (
```
jeremywiebe, line 68: Sorry, I realize I'm focusing alot on naming. Typically, I'd organize a function name as a verb  or verb+noun.  noun+verb tends to sound Yoda-ish. 

`shuffleMatcher()`
vs
`matcherShuffle()` you must.. 😁

packages/perseus/src/util.ts
```diff
@@ -1,10 +1,10 @@
+import {type Range} from "@khanacademy/perseus-core";
 import {KhanAnswerTypes} from "@khanacademy/perseus-score";
 import _ from "underscore";
 
 import * as GraphieUtil from "./util.graphie";
 
 import type {PerseusStrings} from "./strings";
-import type {Range} from "@khanacademy/perseus-core";
```
jeremywiebe, line 7: I'm surprised this moved.. could we move it back to the `type` imports section here?

packages/perseus/src/widgets/categorizer/categorizer.tsx
```diff
@@ -92,8 +92,12 @@ export class Categorizer
         const isMobile = this.props.apiOptions.isMobile;
         let indexedItems = this.props.items.map((item, n) => [item, n]);
         if (this.props.randomizeItems) {
-            // @ts-expect-error - TS4104 - The type 'readonly (string | number)[][]' is 'readonly' and cannot be assigned to the mutable type '(string | number)[][]'. | TS2345 - Argument of type 'number | null | undefined' is not assignable to parameter of type 'number | RNG'.
-            indexedItems = Util.shuffle(indexedItems, this.props.problemNum);
+            // @ts-expect-error - TS4104 - The type 'readonly (string | number)[][]' is 'readonly' and cannot be assigned to the mutable type '(string | number)[][]'.
+            indexedItems = CoreUtil.shuffle(
+                indexedItems,
+                // @ts-expect-error - TS2345 - Argument of type 'number | null | undefined' is not assignable to parameter of type 'number | RNG'.
+                this.props.problemNum,
```
jeremywiebe, line 99: Instead of a `ts-expect-error`, could you use a cast to `any`? Polaris work is going to disallow `ts-expect-error` for many reasons. Casting to `any` should suppress the error but is not quite so risky. 

```suggestion
                this.props.problemNum as any,
```

jeremywiebe (approved):
Whew, I had a bunch of nitpicky comments. The approach you did her is good though! Thanks for taking this through a few rounds as we all figure out how we want to handle this.

packages/perseus-core/src/widgets/matcher/matcher-util.ts
benchristel: I think we should keep `ensurePermuted = true`. It's necessary to preserve the existing behavior. I think it would be confusing (and maybe even blocking?) to learners if we ever displayed the correct answer as the initial state, since we don't enable the "submit" button until the learner changes the state of every widget.

packages/perseus-core/src/widgets/matcher/matcher-util.ts
Myranae: Linking to this resource in another PR - https://github.com/Khan/perseus/pull/2215

packages/perseus-core/src/utils/util.ts
Myranae: Renamed to `random_utils.ts` added `Contains utility functions
    and types used across multiple widgets.` to the comment. I kind of see it as a catch all for utility functions used across widgets, but let me know if it should be more focused than that.

packages/perseus-core/src/widgets/matcher/matcher-util.test.ts
Myranae: I think you mean we read left to right, right? I'll put left above right :)

packages/perseus-core/src/widgets/matcher/matcher-util.ts
Myranae: Updated :)

packages/perseus-core/src/widgets/matcher/matcher-util.ts
Myranae: Changed the comments to doc comments with references :) I feel like the function name change would be going against how I usually see these functions named. If we all decide we want to move away from it, I'd be fine with it, but I think I most often see these kinds of functions with the verb in front, so I'll probably just leave it for now if that's cool :)

packages/perseus/src/util.ts
Myranae: I think it's because the `type` keyword got moved inside the brackets. I think this was because I had imported something else from core at one point that was not a type, so I moved the type keyword inside the brackets and the linter moved it as a result. Moving it out allowed it to stay back where it was before.

packages/perseus/src/widgets/categorizer/categorizer.tsx
Myranae: Removed the expect error comment and cast it to any :)

packages/perseus-core/src/widgets/matcher/matcher-util.test.ts
jeremywiebe: 🤔  Yeah, I guess "negation" in test names is kinda like naming a variable like `notEnabled`.  Boolean logic with it is a nightmare `if (!notEnabled) { /* enabled! */ }`.

packages/perseus-core/src/utils/util.ts
jeremywiebe: That sounds good. We'd only put utility functions in here that relate to randomization and shuffling.  Other utilities could be moved into `perseus-core` under a different filename (for example, there are a bunch of grid-related utilities in `packages/perseus/src/util.ts` which we could move to `grid_utils.ts` if we need to move them). Thanks for updating!

Myranae: Ooo, okay, you meant more like _randomization_ instead of a bunch of random utilities. That makes more sense :D I'll make that more clear too. Thanks!

packages/perseus-core/src/widgets/matcher/matcher-util.test.ts
Myranae: I updated it to `should return a shuffled right array when order doesn't matter to remove the correct order information` which I think is more clear as it gets rid of the negation while still indicating the goal is to remove correct order info

packages/perseus-core/src/widgets/matcher/matcher-util.test.ts
```diff
@@ -0,0 +1,41 @@
...
+            left: ["Fuel", "Plate", "Hydrogen", "Average", "Billion"],
+            right: ["Stars", "Earth", "Life", "Rapid", "Milky Way"],
+        };
+
+        // Act
+        const publicWidgetOptions = getMatcherPublicWidgetOptions(options);
+
+        // Assert
+        expect(publicWidgetOptions.right).not.toEqual(options.right);
+        expect(publicWidgetOptions.left).toEqual(options.left);
```
Myranae, line 21: Now that we're using Math.random(), we can't expect a specific outcome because the shuffle should be different each time. To achieve this, I did have to multiply the random output by 100 though.

packages/perseus-core/src/widgets/matcher/matcher-util.ts
```diff
@@ -0,0 +1,94 @@
...
+        left = CoreUtil.shuffle(
+            data.left,
+            Math.random() * 100,
+            /* ensurePermuted */ true,
+        );
+    }
+
+    const right = CoreUtil.shuffle(
+        data.right,
+        Math.random() * 100,
```
Myranae, line 59: Multiplied the Math.random() here by 100 to get the shuffle to be random each time. Without multiplying, the shuffle is the same each time, though it is different than the answer. I was expecting it to be random, so updated to get that response.

packages/perseus-core/src/utils/random_utils.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    return function () {
+        // Robert Jenkins' 32 bit integer hash function.
+        let seed = randomSeed;
+        seed = (seed + 0x7ed55d16 + (seed << 12)) & 0xffffffff;
+        seed = (seed ^ 0xc761c23c ^ (seed >>> 19)) & 0xffffffff;
+        seed = (seed + 0x165667b1 + (seed << 5)) & 0xffffffff;
+        seed = ((seed + 0xd3a2646c) ^ (seed << 9)) & 0xffffffff;
+        seed = (seed + 0xfd7046c5 + (seed << 3)) & 0xffffffff;
+        seed = (seed ^ 0xb55a4f09 ^ (seed >>> 16)) & 0xffffffff;
```
handeyeco, line 20: This isn't your fault, but I strongly dislike having this kind of indecipherable bit manipulation code in our code base; borderline want to make a ticket to replace it with a small dep...but I guess it's fine for now.

packages/perseus-core/src/widgets/matcher/matcher-util.ts
handeyeco: Would it be worth making this a comment in the code?

packages/perseus/src/widgets/categorizer/categorizer.tsx
```diff
@@ -92,8 +92,11 @@ export class Categorizer
         const isMobile = this.props.apiOptions.isMobile;
         let indexedItems = this.props.items.map((item, n) => [item, n]);
         if (this.props.randomizeItems) {
-            // @ts-expect-error - TS4104 - The type 'readonly (string | number)[][]' is 'readonly' and cannot be assigned to the mutable type '(string | number)[][]'. | TS2345 - Argument of type 'number | null | undefined' is not assignable to parameter of type 'number | RNG'.
-            indexedItems = Util.shuffle(indexedItems, this.props.problemNum);
+            // @ts-expect-error - TS4104 - The type 'readonly (string | number)[][]' is 'readonly' and cannot be assigned to the mutable type '(string | number)[][]'.
+            indexedItems = CoreUtil.shuffle(
+                indexedItems,
+                this.props.problemNum as any,
+            );
```
handeyeco, line 99: It feels like there could be an opportunity to tighten up our types here. Like can `problemNum` really be `null | undefined`? Would that break `shuffle`?

But if we don't want to go down that rabbit hole, maybe:

``` TS
        let indexedItems: ReadonlyArray<Readonly<[string, number]>> =
            this.props.items.map((item, n) => [item, n]);
        if (this.props.randomizeItems) {
            indexedItems = Util.shuffle(
                indexedItems,
                this.props.problemNum as number,
            );
        }
```

packages/perseus/src/widgets/matcher/matcher.tsx
```diff
@@ -157,18 +159,13 @@ export class Matcher extends React.Component<Props, State> implements Widget {
...
-        if (!this.props.orderMatters) {
-            // If the order doesn't matter, don't shuffle the left column
-            left = this.props.left;
-        } else {
-            left = shuffle(this.props.left, rng, /* ensurePermuted */ true);
-        }
-
-        const right = shuffle(this.props.right, rng, /* ensurePermuted */ true);
+        /* TODO(LEMS-2841):
+            Once the getRadioPublicWidgetOptions function gets connected to the
```
handeyeco, line 163: `getRadioPublicWidgetOptions`? Not `getMatcherPublicWidgetOptions`?

handeyeco (approved):
My main concerns are:

1. `ts-expect-error`, but I realize that comes from existing code so whatevs
2. Changing the `package.json` seems like it needs to be reverted
3. I think `random-utils` should be exported as `RandomUtils`

I'm not sure I fully grasp what we're doing though, so if my suggestions conflict with Ben/Jeremy I'd defer to them.

packages/perseus-core/src/widgets/matcher/matcher-util.test.ts
Myranae: Will update the Math.random() call to just Math.random, which will still allow the tests here to pass

packages/perseus-core/src/widgets/matcher/matcher-util.ts
Myranae: Ben pointed out why this is happening and how I just need to pass in Math.random instead, so will update that here.

packages/perseus-core/src/index.ts
```diff
@@ -138,3 +138,6 @@ export {default as getTablePublicWidgetOptions} from "./widgets/table/table-util
 export {default as getIFramePublicWidgetOptions} from "./widgets/iframe/iframe-util";
 export {default as getMatrixPublicWidgetOptions} from "./widgets/matrix/matrix-util";
 export {default as getPlotterPublicWidgetOptions} from "./widgets/plotter/plotter-util";
+export {default as getMatcherPublicWidgetOptions} from "./widgets/matcher/matcher-util";
+export {shuffleMatcher} from "./widgets/matcher/matcher-util";
```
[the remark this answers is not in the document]
Myranae: Yep! Thanks! I thought it usually pointed these out, but maybe it was a different error that helped me see these. Fixed :)

packages/perseus-core/src/index.ts
```diff
@@ -138,3 +138,6 @@ export {default as getTablePublicWidgetOptions} from "./widgets/table/table-util
...
+export {default as CoreUtil} from "./utils/random_utils";
```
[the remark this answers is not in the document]
Myranae: Updated :)

package.json
```diff
@@ -153,5 +153,6 @@
     "nyc": {
         "report-dir": "coverage/cypress/"
     },
-    "dependencies": {}
+    "dependencies": {},
+    "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
```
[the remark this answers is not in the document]
benchristel: The `yarn@1.22.22` is concerning. We should be using pnpm now. Are you using `pnpm install` to update dependencies?

packages/perseus/src/widgets/categorizer/categorizer.tsx
benchristel: Could we avoid the `as any` by doing this?

```ts
indexedItems = CoreUtil.shuffle(
    indexedItems,
    Number(this.props.problemNum),
)
```

or this?

```ts
indexedItems = CoreUtil.shuffle(
    indexedItems,
    this.props.problemNum ?? 0,
)
```

Either of those will handle `null` and `undefined` values, converting them to 0. The problem with using `as any` is that it's unsafe; `CoreUtils.shuffle` says it needs a number as the second argument, not null or undefined. The `as any` is only needed here because we are telling `shuffle` to do something that it explicitly says it does not support.

Alternatively, we could change the types of `CoreUtils.shuffle` to accept `null | undefined` for the second argument.

package.json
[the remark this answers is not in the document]
Myranae: Apologies, I figured it would notify me when it needed to change, but I needed to do it myself it seems. This all should update and actually add some pnpm stuff now.

packages/perseus-core/src/widgets/matcher/matcher-util.ts
Myranae: Updated :)

packages/perseus-core/src/utils/random_utils.ts
```diff
@@ -0,0 +1,77 @@
...
+        return (randomSeed = seed & 0xfffffff) / 0x10000000;
+    };
+};
+
+// Shuffle an array using a given random seed or function.
+// If `ensurePermuted` is true, the input and output are guaranteed to be
+// distinct permutations.
+function shuffle<T>(
```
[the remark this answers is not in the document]
Myranae: Shuffle is exported as part of RandomUtil. So when it's used, it's either RandomUtil.shuffle or it can be destructured like `const {shuffle} = RandomUtil. This should be how it was before I moved these functions to perseus-core.
```
export const RandomUtil = {
    seededRNG,
    shuffle,
    random,
} as const;

export default RandomUtil;
```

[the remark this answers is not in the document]
Myranae: Ah, I see your comment below. You're suggesting we change the way it's exported and used? I'd need to update all the locations that use it as well. Totally can if this is an undesirable pattern here.

packages/perseus-core/src/widgets/matcher/matcher-util.ts
```diff
@@ -0,0 +1,94 @@
...
+): {left: ReadonlyArray<string>; right: ReadonlyArray<string>} => {
+    // Use the same random() function to shuffle both columns sequentially
+    const rng = CoreUtil.seededRNG(props.problemNum as number);
+
+    let left;
+    if (!props.orderMatters) {
+        // If the order doesn't matter, don't shuffle the left column
+        left = props.left;
+    } else {
+        left = CoreUtil.shuffle(props.left, rng, /* ensurePermuted */ true);
```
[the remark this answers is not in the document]
Myranae: Since this function is used in Radio, which is an assessments widget, I'd prefer not to change it. I think it's a little strange to have a comment like that, but it does provide the information in a way available to users outside of an IDE, so I would vote for just leaving it as it is for now. Totally cool if I'm overruled here though, but it sounds like it's more of a nit :) Definitely let me know though.

packages/perseus-core/src/utils/random_utils.ts
Myranae: Yea, it seems a bit excessive, but it works so 🤷‍♀️

packages/perseus/src/widgets/matcher/matcher.tsx
Myranae: Eeek, maybe a copy paste error? Did I even work on radio? lol. 

Updated. Thanks! :D

packages/perseus-core/src/utils/random_utils.ts
```diff
@@ -0,0 +1,77 @@
+/* Note(tamara): Brought over from the perseus package packages/perseus/src/util.ts file.
+    May be useful to bring other perseus package utilities here. Contains utility functions
+    and types used across multiple widgets for randomization and shuffling. */
+
+import _ from "underscore";
+
+type RNG = () => number;
+
+const seededRNG: (seed: number) => RNG = function (seed: number): RNG {
```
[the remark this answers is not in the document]
Myranae: Updated~

packages/perseus-core/src/utils/random_utils.ts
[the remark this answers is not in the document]
Myranae: Updated~

packages/perseus-core/src/utils/random_utils.ts
```diff
@@ -0,0 +1,77 @@
...
+
+const random: RNG = seededRNG(new Date().getTime() & 0xffffffff);
+
+export const CoreUtil = {
+    seededRNG,
+    shuffle,
+    random,
+} as const;
+
+export default CoreUtil;
```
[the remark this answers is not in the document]
Myranae: Removed~

packages/perseus-core/src/index.ts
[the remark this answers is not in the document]
Myranae: Hmm, I may need to update the names to make them more unique since we're exporting them directly now.

package.json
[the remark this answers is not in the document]
Myranae: Looks like we are still on yarn. I will remove this file from the PR, but I'm going to go back to using yarn for now 😅

packages/perseus-core/src/utils/random_utils.ts
```diff
@@ -0,0 +1,77 @@
...
+            shuffled[newEnd] = shuffled[top - 1];
+            // @ts-expect-error - TS2542 - Index signature in type 'readonly T[]' only permits reading.
+            shuffled[top - 1] = temp;
+        }
+    } while (ensurePermuted && _.isEqual(array, shuffled));
+
+    return shuffled;
+}
+
+const random: RNG = seededRNG(new Date().getTime() & 0xffffffff);
```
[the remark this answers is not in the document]
Myranae: Updated

packages/perseus/src/widgets/categorizer/categorizer.tsx
Myranae: I worry about changing what is passed into the function to zero, mostly because I don't know how it handled receiving undefined/null answers before. I also worry about changing it from `as any` to as `as number` but that seems like it's more type focused, so I went with that suggestion :)

Did a little bit of testing, and if it receives undefined it does shuffle it, but the shuffle doesn't change. If it receives null, it does a different shuffle, but it doesn't change either. Passing in `0` gave the same result as `null`.

packages/perseus-core/src/widgets/matcher/matcher-util.ts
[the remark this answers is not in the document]
handeyeco: I still disagree, but it's fine to ship for now and we can talk as a team later.

Mainly I think it's needlessly noisy.

Also if we make a change to shuffle...

``` TS
function shuffle<T>(
    array: ReadonlyArray<T>,
    randomSeed: number | RNG,
    shuffleTwice = false,
    ensurePermuted = false,
): ReadonlyArray<T>
```

...every place we have that comment will be wrong. IMO we should be depending on our IDEs / documentation.

PR #2157, opened 2025-01-24
Take 2: Grapher default widget options

handeyeco:
## Summary:
This is an alternative for: https://github.com/Khan/perseus/pull/2146

In #2146 I tried to a direct move of logic from `perseus` to `perseus-core` to get the default widget options for Grapher. That required `defaultPlotProps` which required `maybePointsFromNormalized` which eventually required `kmath`. It was gross for a number of reasons:

1. I had to move a lot of utilities
2. I had to give `perseus-core` a dependency on `kmath` (and `perseus-core` probably shouldn't have dependencies)
3. There was a lot of misdirection and logic all to make one static object

Whatever, I was okay hacking my way through that. _However_ we ended up with a circular dependency: `kmath` depends on `perseus-core` which was now depending on `kmath`.

At this point I reflected on my life choices and thought "why are we using all of these utilities just to compute an unchanging object? Why am I moving all this logic just to generate 2 properties?

So I just logged the default object and copied it to `perseus-core`.

Issue: LEMS-2737

## Test plan:
Nothing should change, just copy/pasta-ing code.

### DIFF
--- a/.changeset/dry-fans-study.md
+++ b/.changeset/dry-fans-study.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/perseus-core": minor
+"@khanacademy/perseus-editor": patch
+---
+
+Move default widget options for Grapher to Perseus Core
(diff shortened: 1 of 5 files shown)


jeremywiebe (approved):
👍  This feels much cleaner than all the moving and shaping that was needed in https://github.com/Khan/perseus/pull/2146 just to generate the same default object in all situations! 

Thanks for re-thinking this.

PR #2160, opened 2025-01-27
Bugfix: error messages aren't displayed as translated strings

handeyeco:
## Summary:
Proposing a bug fix for error messages not showing up as translated: https://khanacademy.slack.com/archives/C06FULVQLSV/p1738014199191129?thread_ts=1738012970.693379&cid=C06FULVQLSV

**What I thought we were doing:** Originally I thought `strings` were getting replaced during the translation process (I don't know why I thought that). So we were mapping error codes to error strings directly off of `strings`.

**What we were actually doing:** `strings` isn't actually changed, it's used to make a new object that has all of the translated strings in it. So we were pulling the English strings off of the base object that was used to create a separate translated object.

**What this PR does:** instead of pulling from `strings`, we instead pass the translated strings into `mapErrorToString`. Then we map an error code to a key that we use to access the translated message from the passed in object.

## Test plan:
- Go into a non-English language (like Portuguese)
- Go to a Radio widget that requires multiple selections
- Select one right answer, but not all right answers
- Submit question
- There should be an error message displayed
  - broken: that message will be English
  - patched: that message will be Portuguese

### DIFF
--- a/packages/perseus/src/strings.test.ts
+++ b/packages/perseus/src/strings.test.ts
@@ -1,0 +1,18 @@
+import {clone} from "../../../testing/object-utils";
+
+import {mapErrorToString, mockStrings} from "./strings";
+
+describe("mapErrorToString", () => {
+    it("handles translated strings", () => {
+        // Assemble
+        const translated = clone(mockStrings);
+        translated.MISSING_PERCENT_ERROR =
+            "pretend this is a different language";
+
+        // Act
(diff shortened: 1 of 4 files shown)


jeresig (approved):
Looks ok to me, thank you!

packages/perseus/src/strings.test.ts
catandthemachines, line 1: Yay for future proofing!

catandthemachines (approved):

PR #2164, opened 2025-01-28
Remove deprecated exports from the `perseus` package

benchristel:
## Summary:
They have been moved to `perseus-core`.

Issue: none

Test plan:

CI should be green.

### DIFF
--- a/.changeset/poor-masks-return.md
+++ b/.changeset/poor-masks-return.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": major
+---
+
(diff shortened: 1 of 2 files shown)


.changeset/poor-masks-return.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/perseus": major
+---
+
+Removes the following deprecated exports from `@khanacademy/perseus`: `parsePerseusItem`, `parseAndMigratePerseusItem`, `parseAndMigratePerseusArticle`, `isSuccess`, `isFailure`, `Result`, `Success`, `Failure`. Clients should import these from `@khanacademy/perseus-core` instead.
```
jeremywiebe, line 5: 👍  Thanks for pointing to the fix in here also.

jeremywiebe (approved):

PR #2167, opened 2025-01-29
[editor-pi] [Interactive Graph] Support tick labels that are multiples of pi

nishasy:
## Summary:
Currently, if a graph has tick labels that are multiples of pi,
such as in a sinusoid graph, content authors have to set the
graph markings to "none" and load in a graphie background image
that has the graph with the desired tick markings.

This is not only inconvenient for content authors, it also means
that the screen reader will not read the correct coordinates for
these interactive graphs. It might say "2" when it should be
saying "2 pi."

Implementing a fix for this to improve the lives of our content
authors and our screen reader users.

- Editor
  - Check if the value loaded into NumberInput is a multiple
    of pi, and not already written in the "pi" format.
  - If so, and if the `allowPiTrunacation` prop is true,
    divide this number by pi and append "π" to the value.
  - This will make it so that the value shown in the text
    input will show as pi format even if loaded in without it.
    (6.28... --> "2π")
- Mafs graph
  - In the AxisTicks component, check if the tick step is a
    multiple of pi.
  - If so, divide all the ticks by pi and append pi to the value
    (special cases for 0, 1π, and -1π).
  - This will result in the ticks showing as integers * pi.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2059

## Test plan:
`yarn jest packages/perseus/src/components/__tests__/number-input.test.tsx`
`yarn jest packages/perseus/src/util/math-utils.test.ts`
`yarn jest packages/perseus/src/widgets/interactive-graphs/backgrounds/axis-ticks.test.ts`

Storybook
- http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-sinusoid

image

### DIFF
--- a/packages/perseus/src/util/math-utils.ts
+++ b/packages/perseus/src/util/math-utils.ts
@@ -1,0 +1,15 @@
+export function isPiMultiple(value: number): boolean {
+    // Early return for integers that clearly aren't multiples
+    // of pi to save some computation.
+    if (Number.isInteger(value)) {
+        return false;
+    }
+
+    return (
+        value % Math.PI === 0 ||
+        value % (Math.PI / 2) === 0 ||
+        value % (Math.PI / 3) === 0 ||
+        value % (Math.PI / 4) === 0 ||
+        value % (Math.PI / 6) === 0
+    );
+}
--- a/packages/perseus/src/widgets/interactive-graphs/backgrounds/axis-ticks.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/backgrounds/axis-ticks.tsx
@@ -14,10 +14,13 @@ const YGridTick = ({
     y,
     range,
     tickStep,
+    showPi,
 }: {
     y: number;
     range: [Interval, Interval];
     tickStep: number;
+    // Whether to show the tick label as a multiple of pi
+    showPi: boolean;
 }) => {
     // If the graph requires out-of-bounds labels, we want to make sure to set the
     // coordinates to the edge of the visible range of the graph. Otherwise,
@@ -55,6 +58,8 @@ const YGridTick = ({
     // to hide the label at -1 on the y-axis to prevent overlap with the x-axis label
     const showLabel = shouldShowLabel(y, range, tickStep);
 
+    const yLabel = showPi ? divideByAndShowPi(y) : y.toString();
+
     return (
         <g className="tick" aria-hidden={true}>
             <line x1={x1} y1={y1} x2={x2} y2={y2} className="axis-tick" />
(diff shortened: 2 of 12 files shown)


packages/perseus/src/util/math-utils.ts
```diff
@@ -0,0 +1,15 @@
+export function isPiMultiple(value: number): boolean {
+    // Early return for integers that clearly aren't multiples
+    // of pi to save some computation.
+    if (Number.isInteger(value)) {
+        return false;
+    }
```
SonicScrewdriver, line 6: Love it!

packages/perseus/src/widgets/interactive-graphs/backgrounds/axis-ticks.tsx
```diff
@@ -66,14 +71,23 @@ const YGridTick = ({
                     x={xPositionText}
                     y={yPositionText}
                 >
-                    {y.toString()}
+                    {yLabel}
```
SonicScrewdriver, line 74: I like that this got slightly cleaner as well.

packages/perseus/src/widgets/interactive-graphs/backgrounds/axis-ticks.tsx
```diff
@@ -209,6 +242,9 @@ export const AxisTicks = () => {
                             key={`y-grid-tick-${y}`}
                             range={range}
                             tickStep={tickStep[Y]}
+                            // Show the tick labels as multiples of pi
+                            // if the tick step is a multiple of pi.
+                            showPi={tickStep[Y] % Math.PI === 0}
```
SonicScrewdriver, line 247: Ooo this looks great. My brain's a little tired at the end of the day - is there a reason we're not using the `isPiMultiple` function you created here?

SonicScrewdriver (approved):
This looks great, thanks for solving this very tricky problem!

packages/perseus/src/widgets/interactive-graphs/backgrounds/axis-ticks.tsx
nishasy: I didn't use that function because isPiMultiple returns true for some pi fractions. At the moment, I don't think the ticks can handle things like pi/2 or pi/4, so I opted for a direct pi modulo.

I could use the function and add an option to only do integer multiples of pi?

SonicScrewdriver: Naw, that makes sense now! Thank you for the explanation! :D

benchristel:
Tiny detail: from the screenshot, it looks like pi is not rendering in the MathJax font. Do we want it to? I think it should use the MathJax font, since that's the font we use everywhere else for pi.

nishasy:
@benchristel It looks like the font is falling back to the Mafs font for all axis labels, so I'm going to create a new ticket for that.

PR #2170, opened 2025-01-29
Remove z-index tokens, constants, and classes that place SOME widgets above the doodlepad.

nedredmond:
## Summary
I did this in two passes -- one commit to remove `above scratchpad`, and a second for `interactive`, which seems to be used for the same purpose. Both place widgets above the ~scratchpad~ doodlepad, which has long led to many bugs.
- **remove all references to "above scratchpad"**
- **remove z-index from more widgets**

Issue: TUT-1059

### DIFF
--- a/packages/perseus/src/components/graph.tsx
+++ b/packages/perseus/src/components/graph.tsx
@@ -416,7 +416,7 @@ class Graph extends React.Component<Props> {
 
         return (
             <div
-                className="graphie-container above-scratchpad"
+                className="graphie-container blank-background"
                 style={{
                     width: this.props.box[0],
                     height: this.props.box[1],
--- a/packages/perseus/src/perseus-api.tsx
+++ b/packages/perseus/src/perseus-api.tsx
@@ -168,10 +168,6 @@ export const ClassNames = {
         SELECTED: "perseus-radio-selected",
         OPTION_CONTENT: "perseus-radio-option-content",
     },
-    /**
-     * @deprecated This used to add a z-index of 3. Now it does nothing.
-     */
-    INTERACTIVE: "perseus-interactive",
     CORRECT: "perseus-correct",
     INCORRECT: "perseus-incorrect",
     UNANSWERED: "perseus-unanswered",
--- a/packages/perseus/src/perseus-api.tsx
+++ b/packages/perseus/src/perseus-api.tsx
@@ -168,6 +168,9 @@ export const ClassNames = {
         SELECTED: "perseus-radio-selected",
         OPTION_CONTENT: "perseus-radio-option-content",
     },
+    /**
+     * @deprecated This used to add a z-index of 3. Now it does nothing.
+     */
     INTERACTIVE: "perseus-interactive",
     CORRECT: "perseus-correct",
     INCORRECT: "perseus-incorrect",
--- a/packages/perseus/src/widgets/measurer/measurer.tsx
+++ b/packages/perseus/src/widgets/measurer/measurer.tsx
@@ -161,7 +161,7 @@ class Measurer extends React.Component<Props> implements Widget {
             <div
                 className={
                     "perseus-widget perseus-widget-measurer " +
-                    "graphie-container above-scratchpad"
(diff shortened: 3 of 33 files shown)


nedredmond:
I'll do some testing once I get an npm artifact.

packages/perseus/src/components/graph.tsx
```diff
@@ -416,7 +416,7 @@ class Graph extends React.Component<Props> {
 
         return (
             <div
-                className="graphie-container above-scratchpad"
+                className="graphie-container blank-background"
```
nedredmond, line 419: Swapped because `above-scratchpad` adds `blank-background` styles in [perseus-renderer.less](https://github.com/Khan/perseus/pull/2170/files#diff-ff79a578a7a0678d666d4b522615c60cf052a99f70b64109568dd476d9f8e4e4L49)

packages/perseus/src/widgets/measurer/measurer.tsx
```diff
@@ -161,7 +161,7 @@ class Measurer extends React.Component<Props> implements Widget {
             <div
                 className={
                     "perseus-widget perseus-widget-measurer " +
-                    "graphie-container above-scratchpad"
+                    "graphie-container blank-background"
```
nedredmond, line 164: Here too, ensuring we don't lose [nested styles](https://github.com/Khan/perseus/pull/2170/files#diff-ff79a578a7a0678d666d4b522615c60cf052a99f70b64109568dd476d9f8e4e4L49)

nedredmond:
whoops my branch is completely mis-named. ignore that!

nedredmond:
Works as expected!


https://github.com/user-attachments/assets/750c50c2-40ac-4df5-a38f-70e32e0662f3

nedredmond:
@MikeKlemarewski added you mostly as FYI

packages/perseus/src/perseus-api.tsx
```diff
@@ -168,6 +168,9 @@ export const ClassNames = {
         SELECTED: "perseus-radio-selected",
         OPTION_CONTENT: "perseus-radio-option-content",
     },
+    /**
+     * @deprecated This used to add a z-index of 3. Now it does nothing.
+     */
     INTERACTIVE: "perseus-interactive",
```
SonicScrewdriver, line 174: I'm a little less familiar with this file, but would it be possible to remove this style now? If the style is doing nothing now, it seems like it could be a fairly safe thing to clean up! 

The only consumers of it are the widgets already in this PR, along with two more: 

- Input with Examples (A sub component of Numeric Input / Input Number that is used for desktop views) 
- Plotter

SonicScrewdriver (approved):
I'm not currently able to test this in webapp myself due to some dev env issues with bumping Perseus, but this looks sound to me! 

I only had one nit about the possibility of fully deleting the deprecated style. 

It would be good to test the Numeric Input / Input Number widgets to make sure everything is working as expected (as they did use the Interactive style), but I think we're good to go otherwise. :) I don't think we have many, if any, uses of the Plotter widget anymore. 

Thank you for a great improvement! Not sure why we ever tried to show widgets above the scratchpad myself.

packages/perseus/src/perseus-api.tsx
nedredmond: This is in the API so I was hesitant to make any breaking changes.

I wasn't sure about the purpose of this enum. Could this be used as a selector? If so, maybe I should restore the class and just delete the styles.

nedredmond: I'll explore uses of this enum, but it would be a breaking change, which would lead to a major version bump.

nedredmond: I don't see any consumers of ClassNames so I think it's safe to remove it.

nedredmond: I guess the Perseus API is internal? At least this was. Removed it.

nedredmond: Excellent call, Sarah

PR #2171, opened 2025-01-29
Create a function to get the public options for LabelImage

benchristel:
Issue: LEMS-2769

## Test plan:

`yarn test`

### DIFF
--- a/.changeset/forty-meals-teach.md
+++ b/.changeset/forty-meals-teach.md
@@ -1,4 +1,5 @@
 ---
+"@khanacademy/perseus": minor
 "@khanacademy/perseus-core": minor
 ---
 
--- a/.changeset/forty-meals-teach.md
+++ b/.changeset/forty-meals-teach.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-core": minor
+---
+
+Create a function to get the public options for the LabelImage widget
--- a/packages/perseus-core/src/widgets/label-image/label-image-util.ts
+++ b/packages/perseus-core/src/widgets/label-image/label-image-util.ts
@@ -1,0 +1,41 @@
+import type {
+    PerseusLabelImageMarker,
+    PerseusLabelImageWidgetOptions,
+} from "@khanacademy/perseus-core";
+
+/**
+ * For details on the individual options, see the
+ * PerseusLabelImageWidgetOptions type
+ */
+type LabelImagePublicWidgetOptions = {
+    choices: PerseusLabelImageWidgetOptions["choices"];
+    imageUrl: PerseusLabelImageWidgetOptions["imageUrl"];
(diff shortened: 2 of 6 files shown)


packages/perseus-core/src/widgets/label-image/label-image-util.ts
```diff
@@ -0,0 +1,41 @@
...
+type LabelImagePublicWidgetOptions = {
+    choices: PerseusLabelImageWidgetOptions["choices"];
+    imageUrl: PerseusLabelImageWidgetOptions["imageUrl"];
+    imageAlt: PerseusLabelImageWidgetOptions["imageAlt"];
+    imageHeight: PerseusLabelImageWidgetOptions["imageHeight"];
+    imageWidth: PerseusLabelImageWidgetOptions["imageWidth"];
+    markers: ReadonlyArray<LabelImageMarkerPublicData>;
+    hideChoicesFromInstructions: boolean;
+    multipleAnswers: boolean;
+    static: boolean;
```
Myranae, line 19: Curious why these ones are listed as boolean while the others reference the original widget options. Wouldn't it be good to reference the original type for all of the fields that are unchanged?

packages/perseus-core/src/widgets/label-image/label-image-util.ts
```diff
@@ -0,0 +1,41 @@
...
+        ...options,
+        markers: options.markers.map(getLabelImageMarkerPublicData),
+    };
+}
+
+function getLabelImageMarkerPublicData(
+    marker: PerseusLabelImageMarker,
+): LabelImageMarkerPublicData {
+    const {answers: _, ...publicData} = marker;
+    return publicData;
```
Myranae, line 40: So here, instead of building out the wanted object, you're using destructuring to pull out answers, is that correct? What does using the underscore do? Is that just to help signify that we aren't interested in the answers here? Looks like it's needed to satisfy ESLint.

.changeset/forty-meals-teach.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/perseus-core": minor
```
Myranae, line 2: Should there be a line here for `perseus` since there are changes there as well?

Myranae (approved):
Looks good! Added some comments/questions, but nothing blocking. Looks like there are some prettier formatting issues in the `label-image.tsx` file though

packages/perseus-core/src/widgets/label-image/label-image-util.ts
benchristel: > What does using the underscore do? Is that just to help signify that we aren't interested in the answers here?

Correct!

.changeset/forty-meals-teach.md
benchristel: Good catch! I guess this is one reason not to generate the changeset until the PR is reviewed 🤔

PR #2174, opened 2025-01-30
Create helper to build public widget options for Numeric Input

Myranae:
## Summary:
Adds a function that takes a Numeric Input widget's full widget options and removes correct answer information. It also adds the function to the widget's widget export and adds a test confirming the function works as expected.

Issue: LEMS-2767

## Test plan:
- Confirm all checks pass
- Confirm the Numeric Input widget still acts as expected

### DIFF
--- a/.changeset/dry-fans-study.md
+++ b/.changeset/dry-fans-study.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/perseus-core": minor
+"@khanacademy/perseus-editor": patch
+---
+
+Move default widget options for Grapher to Perseus Core
--- a/packages/perseus-core/src/index.ts
+++ b/packages/perseus-core/src/index.ts
@@ -63,6 +63,8 @@ export {default as gradedGroupLogic} from "./widgets/graded-group";
 export type {GradedGroupDefaultWidgetOptions} from "./widgets/graded-group";
 export {default as gradedGroupSetLogic} from "./widgets/graded-group-set";
 export type {GradedGroupSetDefaultWidgetOptions} from "./widgets/graded-group-set";
+export {default as grapherLogic} from "./widgets/grapher";
+export type {GrapherDefaultWidgetOptions} from "./widgets/grapher";
 export {default as groupLogic} from "./widgets/group";
 export type {GroupDefaultWidgetOptions} from "./widgets/group";
 export {default as iframeLogic} from "./widgets/iframe";
--- a/packages/perseus-core/src/widgets/grapher/index.ts
+++ b/packages/perseus-core/src/widgets/grapher/index.ts
@@ -1,0 +1,38 @@
+import type {PerseusGrapherWidgetOptions} from "../../data-schema";
+import type {WidgetLogic} from "../logic-export.types";
+
+export type GrapherDefaultWidgetOptions = Pick<
+    PerseusGrapherWidgetOptions,
+    "graph" | "correct" | "availableTypes"
+>;
+
+const defaultWidgetOptions: GrapherDefaultWidgetOptions = {
+    graph: {
+        labels: ["x", "y"],
+        range: [
+            [-10, 10],
+            [-10, 10],
+        ],
+        step: [1, 1],
+        backgroundImage: {
+            url: null,
+        },
+        markings: "graph",
+        rulerLabel: "",
+        rulerTicks: 10,
+        valid: true,
+        showTooltips: false,
+    },
+    correct: {
+        type: "linear",
+        coords: null,
+    },
+    availableTypes: ["linear"],
+};
+
+const grapherWidgetLogic: WidgetLogic = {
+    name: "grapher",
+    defaultWidgetOptions,
+};
+
+export default grapherWidgetLogic;
--- a/packages/perseus-editor/src/widgets/grapher-editor.tsx
+++ b/packages/perseus-editor/src/widgets/grapher-editor.tsx
@@ -7,21 +7,21 @@ import {
     containerSizeClass,
     getInteractiveBoxFromSizeClass,
 } from "@khanacademy/perseus";
-import {GrapherUtil as CoreGrapherUtil} from "@khanacademy/perseus-core";
+import {
+    GrapherUtil as CoreGrapherUtil,
+    grapherLogic,
+} from "@khanacademy/perseus-core";
 import * as React from "react";
 import _ from "underscore";
 
 import GraphSettings from "../components/graph-settings";
 
+import type {GrapherDefaultWidgetOptions} from "@khanacademy/perseus-core";
+
 const {InfoTip, MultiButtonGroup} = components;
 const Grapher = GrapherWidget.widget;
-const {
-    DEFAULT_GRAPHER_PROPS,
-    chooseType,
-    defaultPlotProps,
-    getEquationString,
-    typeToButton,
-} = GrapherUtil;
+const {chooseType, defaultPlotProps, getEquationString, typeToButton} =
+    GrapherUtil;
 
 type Props = any;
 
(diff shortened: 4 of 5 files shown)


packages/perseus/src/types.ts
```diff
@@ -544,6 +545,7 @@ export type WidgetScorerFunction = (
  * A union type of all the functions that provide public widget options.
  */
 export type PublicWidgetOptionsFunction =
+    | typeof getNumericInputPublicWidgetOptions
```
Myranae, line 548: Is this file the best place for this union? It is used to type the `getPublicWidgetOptionsFunction` in the `widgets.ts` file, which was created to be the function used in the editor to access the correct `getPublicWidgetOptions` function.

packages/perseus-core/src/widgets/numeric-input/numeric-input-util.ts
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Given a PerseusNumericInputWidgetOptions object, return a new object with only
+ * the public options that should be exposed to the client.
+ */
+function getNumericInputPublicWidgetOptions(
+    options: PerseusNumericInputWidgetOptions,
+): NumericInputPublicWidgetOptions {
+    const {answers: _, answerForms: __, ...publicWidgetOptions} = options;
+    return publicWidgetOptions;
```
Myranae, line 23: Is there any reason this would be preferable or not preferable to building out the specific object we're looking for like I was doing previously? ex. {size: options.size, static: options.static, etc.}

I could see this being less brittle as it makes it easier to add fields to the widgetOptions and have them be automatically included. If this is preferable, should I go through and update the functions that do this the previous way?

packages/perseus-core/src/widgets/numeric-input/numeric-input-util.ts
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * For details on the individual options, see the
+ * PerseusNumericInputWidgetOptions type
+ */
+type NumericInputPublicWidgetOptions = {
+    labelText?: PerseusNumericInputWidgetOptions["labelText"];
+    size: PerseusNumericInputWidgetOptions["size"];
+    coefficient: PerseusNumericInputWidgetOptions["coefficient"];
+    rightAlign?: PerseusNumericInputWidgetOptions["rightAlign"];
+    static: PerseusNumericInputWidgetOptions["static"];
```
Myranae, line 12: I left out both `answers` and `answerForms` as they both appear to contain information related to the correct answer. 'answerForms` even has a note that it's supposedly used in examples, but might not even be used anymore.

benchristel: I think answerForms should actually be considered public information. See e.g. the `examples()` method on `NumericInput`, which apparently generates learner-facing instructions based on the `answerForms` that are accepted.

packages/perseus-core/src/widgets/numeric-input/numeric-input-util.ts
benchristel: I don't think it matters too much one way or the other. The tests ensure that we include the correct fields either way.

I guess one way of approaching it might be to ask, "when we add fields to the widget options, are they more likely to be public or private?" If the code is written the way you have it here, we won't need to change it when adding public options, but we _will_ need to change it when adding private options. On the other hand, if we explicitly list the public fields, we won't need to change this code when adding private options, but we will need to change it when adding public options.

Then there's the question of what's easier to read. The current approach has the advantage of being short, but it might be a bit obscure.

Ultimately, it's a judgment call, and a two-way door, so we can just pick an approach and, worst case, change it later.

packages/perseus/src/types.ts
benchristel: I think this is fine!

benchristel (changes requested):
Requesting changes because I think we want to include `answerForms` in the public data, since it is used to display instructions to the learner. Other than that, this looks good!

packages/perseus-core/src/widgets/numeric-input/numeric-input-util.ts
Myranae: Thanks for sharing all that! I didn't think about the difference between adding public versus private options and how either way may require a change. I like this way as it shows what is getting pulled out much more clearly.

packages/perseus-core/src/widgets/numeric-input/numeric-input-util.ts
handeyeco: Related to this, I added @SonicScrewdriver as a reviewer (and pinged her) since she's just been looking at this.

Myranae: My concern was that answerForms has information that could tell the user what to do. Specifically, the answerFrom names field: 

```
    answerForms: ReadonlyArray<{
        simplify: "required" | "correct" | "enforced" | null | undefined;
        name: "integer" | "decimal" | "proper" | "improper" | "mixed" | "pi";
    }>;
```

Maybe that information is supposed to be available to the user?

benchristel: Yes, the answer forms are displayed to the learner. Example:

Screen Shot 2025-01-31 at 8 43 59 AM

SonicScrewdriver: Yes! The answer forms are displayed to the learner, and should already be specified to the user through how the question is written. 

> For example: If the question wants a fraction answer, the user will be prompted about this in a secondary fashion as well, either in the title of their course ("Improper Fractions") or directly in the question text. ("Please write your answer as an improper fraction"). 

The tool tips are just helpful reminders for users, particularly young ones. I don't think this could be something that anyone could use to cheat or get ahead on a question, as the user should never be surprised about which answer format the question requires. 

As a result, I think it's safe to keep the answerForms public!

Myranae: Awesome! Thank you both! Just added them back in :D

SonicScrewdriver:
Once we have answerForms, I think this looks great! Thank you very much for sorting this out. :) 

I'm a little less familiar with the server-side-scoring setup for this: will this usurp the `transform` method? 

If so, I have a bugfix on the Numeric Input branch that will need to be added to the `numeric-input-util.ts` file. It's a simple fix: just a line that filters for the correct answers before generating the `answerForms` key, as we don't want to suggest that users enter the `answerForms` for the _wrong_ answers! 

```typescript
const propsTransform = function (
    widgetOptions: PerseusNumericInputWidgetOptions,
): RenderProps {
    const {answers: _, ...rendererProps} = {
        ...widgetOptions,
        answerForms: unionAnswerForms(
            widgetOptions.answers
                .filter((answer) => answer.status === "correct")    <--- Just for context 
                .map((answer) => {
                    return (answer.answerForms || []).map((form) => {
                        return {
                            simplify: answer.simplify,
                            name: form,
                        };
                    });
                }),
        ),
    };

    return rendererProps;
};
```

Myranae:
> I'm a little less familiar with the server-side-scoring setup for this: will this usurp the `transform` method?

I'm not sure if it will overwrite anything, but the transform method on the numeric-input widget export _is_ this propsTransform function. Or are you referring to something else?

```
export default {
    name: "numeric-input",
    displayName: "Numeric input",
    defaultAlignment: "inline-block",
    accessible: true,
    widget: NumericInput,
    transform: propsTransform,
...
```

Also when we actually hook these functions up, we'll have to do some more work to make sure the widget is getting all the information it needs on the frontend without revealing the answer, so I think that's when this would be addressed, if I am understanding correctly.

benchristel (approved):
LGTM!

PR #2176, opened 2025-01-30
Add -all option to 'yarn clean' 

jeremywiebe:
## Summary:

We've had instances where folks' local repo gets into a bad state. It's assumed that `yarn clean` will resolve this, but it hasn't in all cases because it doesn't delete `node_modules`.

This PR adds a `--all` option to `yarn clean` to demolish all `node_modules` folders in the local repo.

Issue: "none" 

## Test plan:

`yarn`
`yarn clean`
`ls node_modules` # still exists
`yarn clean --all`
`find . -name node_modules -and -type d` # no results!

### DIFF
--- a/utils/clean.sh
+++ b/utils/clean.sh
@@ -13,7 +13,21 @@ MYPATH=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 # ROOT is the root directory of our project.
 ROOT="$MYPATH/.."
 
-pushd "$ROOT"
+pushd "$ROOT" >/dev/null 2>&1
+
+CLEAN_ALL=false
+while [[ $# -gt 0 ]]; do
+  case $1 in
+    -a|--all)
+      CLEAN_ALL=true
+      shift # past argument
+      ;;
+    *)
+      echo "Unknown option: $1"
+      exit 1
+      ;;
+  esac
+done
 
 rm -rf packages/*/dist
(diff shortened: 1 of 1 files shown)


utils/clean.sh
```diff
@@ -13,11 +13,30 @@ MYPATH=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 # ROOT is the root directory of our project.
 ROOT="$MYPATH/.."
 
-pushd "$ROOT"
+pushd "$ROOT" >/dev/null 2>&1
+
+CLEAN_ALL=false
+while [[ $# -gt 0 ]]; do
+  case $1 in
+    -a|--all)
```
jeremywiebe, line 21: We could probably make an argument for not having an option at all and just deleting all `node_modules` folder always. 

What say you?

SonicScrewdriver: Tbh I'm personally fine with nuking it all the time! It's fairly quick to install and I find I need to clear out node_modules often enough that I might as well do it all the time. :)

SonicScrewdriver (approved):
Love it! Thank you for a small but great dev experience improvement. :)

utils/clean.sh
jeremywiebe: I'm going to remove the `--all` flag and just always do it.  After all, if you're running this, it's typically understood that "clean" means... delete everything that isn't in git.

PR #2178, opened 2025-01-31
Add a function to get the public options for a cs-program widget

benchristel:
CSProgram doesn't have any private options, so this is just the identity
function.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2770

## Test plan:

`yarn test`


Myranae (approved):
Looks good :) Thanks!

PR #2180, opened 2025-01-31
Restore deprecated exports of `@khanacademy/perseus`

benchristel:
The removal of these functions was causing problems for people who wanted to
use a dev build of Perseus in webapp, because webapp is still importing them.
We need to update webapp to import these functions from perseus-core instead,
and then we can remove them from perseus.

Issue: none

## Test plan:

`yarn test`


jeremywiebe (approved):
Thanks Ben.

👍

PR #2185, opened 2025-02-03
[Bugfix] Zoomable content within a collapsed parent not showing when expanded (Mobile only)

mark-fitzgerald:
## Summary:
Content that is intended to be zoomable needs to have a context for its available space. It does this by using the `offsetWidth` of its container node to calculate a `scale` value. However, if the container is within a collapsible widget (such as the Explanation widget), then the container node has a zero width, which results in a scale of zero making it unavailable to the learner.

This bugfix adds a check for the `offsetWidth` value and loops up the DOM tree for an ancestor with the `perseus-renderer` class. This ancestor should have a valid width. If no ancestors can be found with a valid width (aka we run out of `perseus-renderer` nodes), then we break out of the loop and allow for the scale to be zero (instead of causing an infinite loop or error message).

Issue: LEMS-2805

## Test plan:

1. Go to the Test Everything Course (Settings > Debug Settings).
2. Go to Unit 2: > Radio > “Radio, definition, explanation exercise”.
3. Answer a question correctly.
4. Click the link in the popup (e.g. “See how we answered this”) to view the hint/rationale.
   * (If needed) Scroll up to the Explanation widget.
5. Toggle open the Explanation widget.
6. Observe that the open widget renders the content in a scaled-down mode (so that there is no scrolling).
   * Observe that the content can be scaled up by tapping once on the content.


packages/perseus/src/components/zoomable.tsx
```diff
@@ -186,9 +186,26 @@ class Zoomable extends React.Component<Props, State> {
...
+        // Therefore, we check to see if the node has a valid width, and if it doesn't,
+        //      then we traverse up the DOM tree and look for the closest "perseus-renderer"
+        //      (which should have a width).
+        // To be cautious, we also account for the possibility that we can't find any parents with a valid width
+        //      (not that they don't exist, just that we can't find them).
+        let parentNode = this._node;
+        let currentNode: HTMLElement | null = parentNode;
+        while (currentNode && currentNode.offsetWidth === 0) {
+            currentNode =
+                currentNode.parentElement?.closest(".perseus-renderer") ?? null;
```
jeremywiebe, line 201: Why do we need to access `parentElement` here before calling `.closest()`?  If we don't need to, then `closest()` would return and would alleviate us from having to use the bullish coalescing operator on the end also.

jeremywiebe (approved):
The fix makes sense... and thanks for documenting why we're doing it right there with the fix.

packages/perseus/src/components/zoomable.tsx
mark-fitzgerald: `closest()` [also checks the element on which it is called](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest#:~:text=The%20closest%20ancestor%20Element%20or%20itself). So, when we find an ancestor with `.perseus-renderer`, if we need to do another call for `closest()`, we would end up with the same element, resulting in an infinite loop (speaking from personal experience, unfortunately). Referencing the parent element ensures that we continue traversing up the DOM tree.

jeremywiebe: Ouch!  Can you note that in a comment?  That was the first thing I wondered and would probably think, "I'll just simplify this code some by refactoring this... " 😂

mark-fitzgerald: Comment added.

jeremywiebe: Thanks!

PR #2187, opened 2025-02-04
[scrollin-aint-right] Fix right align scrollbars in Numeric Input 

SonicScrewdriver:
## Summary:
This PR is part of the Numeric Input project. 

While investigating the scrollbar issue for mobile views, I discovered that we have a lot of hacks in place to make sure that moving the cursor does not move the surrounding numbers/elements. However, this logic was causing a bug that would show a tiny scrollbar when a user first interacts with a Math Input that was set to be right aligned. 

We can fix this without causing any regressions by slightly increasing the padding. 

NOTE: This PR also includes some quick requests regarding changes to aria labels. :) 

Issue: LEMS-2817

## Videos:
Left Aligned:

https://github.com/user-attachments/assets/35e23365-19ef-422b-9d3f-e148b03a76d0

Right Aligned:

https://github.com/user-attachments/assets/7b087a8c-48de-4506-a211-2f7642f7b5e8

## Test plan:
- Manual testing in webapp.


packages/perseus/src/widgets/numeric-input/input-with-examples.tsx
```diff
@@ -95,21 +95,20 @@ const InputWithExamples = forwardRef<
         const ariaId = `aria-for-${id}`;
 
         // Generate the provided examples in simple language for screen readers.
-        // If all examples are provided, do not provide them to the screen reader.
-        const examplesAria =
-            props.examples.length === 0
-                ? ""
-                : `${props.examples[0]}
+        const examplesAria = props.shouldShowExamples
```
SonicScrewdriver, line 98: We already have this variable that accomplished the same thing earlier, so we might as well use it! :)

mark-fitzgerald (approved):
Thanks for the additional accessibility related adjustments, as well.

PR #2188, opened 2025-02-04
Add a function to get the public widget options for an iframe widget

Myranae:
## Summary:
IFrame does not have private options to remove, so all widget options are public. The IFrame get public widget options function is an identity function.

Issue: LEMS-2760

## Test plan:
- Confirm all checks pass
- Confirm widget acts as expected


benchristel (approved):
LGTM!

PR #2189, opened 2025-02-04
Create a function to get the public options for a Table widget

benchristel:
Issue: https://khanacademy.atlassian.net/browse/LEMS-2766

## Test plan:

`yarn test`

### DIFF
--- a/packages/perseus-core/src/widgets/table/table-util.ts
+++ b/packages/perseus-core/src/widgets/table/table-util.ts
@@ -1,0 +1,13 @@
+import type {PerseusTableWidgetOptions} from "@khanacademy/perseus-core";
+
+type TableWidgetPublicOptions = Pick<
+    PerseusTableWidgetOptions,
+    "headers" | "rows" | "columns"
+>;
+
+export default function getTablePublicWidgetOptions(
+    options: PerseusTableWidgetOptions,
+): TableWidgetPublicOptions {
+    const {answers: _, ...publicOptions} = options;
+    return publicOptions;
+}
--- a/packages/perseus-core/src/widgets/table/table-util.ts
+++ b/packages/perseus-core/src/widgets/table/table-util.ts
@@ -1,13 +1,13 @@
 import type {PerseusTableWidgetOptions} from "@khanacademy/perseus-core";
 
-type TableWidgetPublicOptions = Pick<
+type TablePublicWidgetOptions = Pick<
     PerseusTableWidgetOptions,
     "headers" | "rows" | "columns"
 >;
 
 export default function getTablePublicWidgetOptions(
     options: PerseusTableWidgetOptions,
-): TableWidgetPublicOptions {
+): TablePublicWidgetOptions {
     const {answers: _, ...publicOptions} = options;
     return publicOptions;
 }
--- a/packages/perseus/src/widgets/table/table.tsx
+++ b/packages/perseus/src/widgets/table/table.tsx
@@ -17,6 +17,7 @@ import type {
     PerseusTableRubric,
     PerseusTableUserInput,
 } from "@khanacademy/perseus-score";
+import {getTablePublicWidgetOptions} from "@khanacademy/perseus-core";
 
 const {assert} = InteractiveUtil;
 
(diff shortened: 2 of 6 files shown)


packages/perseus/src/widgets/table/table.tsx
```diff
@@ -18,6 +18,7 @@ import type {
     PerseusTableRubric,
     PerseusTableUserInput,
 } from "@khanacademy/perseus-score";
+import {getTablePublicWidgetOptions} from "@khanacademy/perseus-core";
```
Myranae, line 21: Might want to merge this import with the other `perseus-core` ones. My IDE gives me warnings about this sometimes, but not always. I don't know if I checked this for your other PRs, so those might have two as well.

packages/perseus-core/src/widgets/table/table-util.ts
```diff
@@ -0,0 +1,13 @@
+import type {PerseusTableWidgetOptions} from "@khanacademy/perseus-core";
+
+type TableWidgetPublicOptions = Pick<
```
Myranae, line 3: I believe we've been naming them `<widget>PublicWidgetOptions` so this would be `TablePublicWidgetOptions`

Myranae (approved):
Looks good to me with a couple things you may want to update. I think Table will definitely need some work once we actually set it up to only receive the public widget options. Excited to get to that next step soon!

packages/perseus/src/widgets/table/table.tsx
benchristel: It looks like the only other import from `@khanacademy/perseus-core` is types-only. I think our linter requires type and value imports to be separated like this? The alternative would be:

```ts
import {
    getTablePublicWidgetOptions,
    type PerseusTableWidgetOptions,
```

packages/perseus-core/src/widgets/table/table-util.ts
benchristel: Good catch, thanks!

packages/perseus/src/widgets/table/table.tsx
Myranae: I usually write it the alternative way you show here. If both are correct though, then totally okay with either. I felt it would be good to keep imports from the same place together if possible, so I combine them when I notice.

packages/perseus-core/src/widgets/table/table-util.ts
```diff
@@ -0,0 +1,13 @@
...
+
+type TablePublicWidgetOptions = Pick<
+    PerseusTableWidgetOptions,
+    "headers" | "rows" | "columns"
+>;
+
+export default function getTablePublicWidgetOptions(
+    options: PerseusTableWidgetOptions,
+): TablePublicWidgetOptions {
+    const {answers: _, ...publicOptions} = options;
```
jeremywiebe, line 11: Random question: if a new key is added to `PerseusTableWidgetOptions` that is answer data, will type checking still allow this spread to pass (incorrectly)?

It appears so: https://www.typescriptlang.org/play/?#code/C4TwDgpgBA8mwEsD2A7AzlAvFA3gKCkKkWABsIAuKNYAJwRQHMAaKAogQ3QHcJaqa9Js3aFRUCAA86HAII8+AugxZQA9GqgA5AKIB1NgF88eUJCgAFAK4AjUggDGcRKgzYLjgNYAeZ8nSsAEQk5IEAfCYAZlYoDi4oUIwQwNZ2jn6uABRI8P5oVBnoAJRUqfZOua644g6uwLhcaLz8UAD6rAB0XWC25YVohlhQOfFoANxs4rTJVrQJPWkVo3iGQA

benchristel: Type checking will pass, but the test will fail.

packages/perseus/src/widgets/table/table.tsx
jeremywiebe: I think when the linter auto fixes/auto imports, it splits them... but the linter doesn't force that separation.  I tend to prefer type imports separated, but that's personal preference. :)

PR #2190, opened 2025-02-04
[kmath-radian-to-degree] Add a `convertRadiansToDegrees` function to kmath

nishasy:
## Summary:
Currently kmath has a `convertDegreesToRadians` function, but not the other way around.

- Add `convertRadiansToDegrees` function to kmath
- Replace all the identical function in other repos with this new one

Issue: none

## Test plan:
`yarn jest packages/kmath/src/angles.test.ts`
`yarn jest`


SonicScrewdriver (approved):
LOVE IT. Thank you for cleaning up our duplicated math functions!

packages/kmath/src/angles.ts
```diff
@@ -8,6 +8,12 @@ export function convertDegreesToRadians(degrees: number): number {
     return (degrees / 180) * Math.PI;
 }
 
+export function convertRadiansToDegrees(radians: number): number {
```
jeremywiebe, line 11: No action needed, but I was musing on naming in this file.  We have `calculate`, `convert`, `get`, and bare action naming. 

I think I'd be inclined to remove the prefixes as they don't really add much value (is there a difference between "calculating" or "converting" or "getting" a value given another value? It's all a transformation/calculation. 

That would leave us with an angles API like this: 

* `degreesToRadians(degrees: number): number`
* `radiansToDegrees(radians: number): number`
* `angleToDegrees([x, y]: Coord): number`
* `angleFromVertex...`
* `clockwiseAngle...`

Not asking you to change anything, but I feel having a consistent naming convention for these types of things is useful. :) Dig in if you feel motivated and agree though!

PR #2191, opened 2025-02-05
[SR] Polygon (unlimited) - Add screen reader experience

nishasy:
## Summary:
Add the aria labels and descriptions for the full graph and interactive elements in the Unlimited Polygon graph, based on the [SRUX doc](https://khanacademy.atlassian.net/wiki/spaces/LC/pages/3455713294/Unlimited+Polygon).

Issue: https://khanacademy.atlassian.net/browse/LEMS-2487

## Test plan:
`packages/perseus/src/widgets/interactive-graphs/graphs/polygon.test.tsx`

Storybook
http://localhost:6006/iframe.html?globals=&args=&id=perseuseditor-widgets-interactive-graph--interactive-graph-unlimited-polygon&viewMode=story

### DIFF
--- a/packages/perseus/src/strings.ts
+++ b/packages/perseus/src/strings.ts
@@ -427,7 +427,9 @@ export type PerseusStrings = {
     srPolygonGraph: string;
     srPolygonGraphCoordinatePlane: string;
     srPolygonGraphPointsNum: ({num}: {num: number}) => string;
+    srPolygonGraphPointsOne: string;
     srPolygonElementsNum: ({num}: {num: number}) => string;
+    srPolygonElementsOne: string;
     srPolygonPointAngleApprox: ({angle}: {angle: string}) => string;
     srPolygonPointAngle: ({angle}: {angle: number}) => string;
     srPolygonSideLength: ({
@@ -444,6 +446,7 @@ export type PerseusStrings = {
         pointNum: number;
         length: string;
     }) => string;
+    srUnlimitedPolygonEmpty: string;
     // The above strings are used for interactive graph SR descriptions.
 };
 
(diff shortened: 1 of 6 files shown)


packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -180,7 +180,7 @@ export const MafsGraph = (props: MafsGraphProps) => {
                         interactiveElementsDescription &&
                             interactiveElementsDescriptionId,
                         isUnlimitedGraphState(state) &&
-                            "unlimited-graph-keyboard-prompt",
```
nishasy, line 183: This wasn't doing anything, since the actual prompt's ID is `unlimitedGraphKeyboardPromptId`, which includes the unique ID from `React.useId()`. Fixed it here so it actually gets the correct ID.

packages/perseus/src/strings.ts
```diff
@@ -1005,14 +1011,17 @@ export const mockStrings: PerseusStrings = {
     srPolygonGraph: "A polygon.",
     srPolygonGraphCoordinatePlane: "A polygon on a coordinate plane.",
     srPolygonGraphPointsNum: ({num}) => `The polygon has ${num} points.`,
+    srPolygonGraphPointsOne: "The polygon has 1 point.",
     srPolygonElementsNum: ({num}) => `A polygon with ${num} points.`,
+    srPolygonElementsOne: "A polygon with 1 point.",
```
SonicScrewdriver, line 1016: [Just a comment] It's actually interesting semantically if this is still a "polygon". I think this is the right call given it's specifically a polygon graph, but I thought it was interesting. :)

SonicScrewdriver (approved):
Love it, more great work! Thank you for persevering through these arduous graph types.

packages/perseus/src/strings.ts
nishasy: It's definitely weird! I asked about this in the SRUX doc while working on this. I may have to update this in the future depending on the response I get.

PR #2194, opened 2025-02-05
Create a function to get the public options for an InteractiveGraph widget

benchristel:
Issue: https://khanacademy.atlassian.net/browse/LEMS-2755

## Test plan:

`yarn test`

### DIFF
--- a/packages/perseus-core/src/widgets/interactive-graph/interactive-graph-util.test.ts
+++ b/packages/perseus-core/src/widgets/interactive-graph/interactive-graph-util.test.ts
@@ -1,0 +1,40 @@
+import getInteractiveGraphPublicWidgetOptions from "./interactive-graph-util";
+
+import type {PerseusInteractiveGraphWidgetOptions} from "@khanacademy/perseus-core";
+
+describe("getInteractiveGraphPublicWidgetOptions", () => {
+    it("removes the `correct` field", () => {
+        const options: PerseusInteractiveGraphWidgetOptions = {
+            correct: {
+                type: "angle",
+                coords: [
+                    [0, 0],
+                    [1, 1],
+                    [2, 2],
+                ],
+                match: "congruent",
+            },
+            graph: {type: "angle"},
+            markings: "none",
+            range: [
+                [0, 10],
+                [0, 10],
+            ],
+            showProtractor: false,
+            step: [1, 1],
+        };
+
+        const publicOptions = getInteractiveGraphPublicWidgetOptions(options);
+
+        expect(publicOptions).toEqual({
+            graph: {type: "angle"},
+            markings: "none",
+            range: [
+                [0, 10],
+                [0, 10],
+            ],
+            showProtractor: false,
+            step: [1, 1],
+        });
+    });
+});
--- a/.changeset/curvy-donuts-applaud.md
+++ b/.changeset/curvy-donuts-applaud.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/perseus": minor
+"@khanacademy/perseus-core": minor
+---
+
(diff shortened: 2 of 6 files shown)


Myranae:
Looks good, but it reminded me of this [ticket that hasn't been done yet related to the correct field](https://khanacademy.atlassian.net/browse/LEMS-2344?atlOrigin=eyJpIjoiMGFkNGRjMWZmYzg4NGM3NWFkOWNlOTNkYzMyODkxNjgiLCJwIjoiaiJ9). I don't think that actually affects us at all, but I remember that the correct field was frequently used as the source of information for various graphs. Do you know if that's still the case?

benchristel:
@Myranae I don't _think_ the issue described in that ticket will come into play for SSS. We're removing the entire `correct` field from the options, so its internal structure doesn't really matter for our purposes.

Myranae (approved):
Looks good! From what I can remember, `correct` was sometimes used in place of `graph` to render the graphs, so might need to look into that before hooking these functions up. Created a [follow-up ticket](https://khanacademy.atlassian.net/browse/LEMS-2838).

benchristel:
Oh! Are you talking about static mode? That's going to be an issue for several widgets, not just interactive graph. I know Radio has a static mode too.

Since the `static` field is on the `WidgetOptions` type, which is common to all widgets, what I imagine we'll do is something like this during publish:

```ts
function getPublicWidgetData<W extends WidgetOptions<any, any>>(widget: W): W {
    // Don't remove the answers from static widgets. Static widgets are not
    // interactive; instead they alwats display the correct answer.
    if (widget.static) {
        return widget;
    }

    // Remove answers and hints from non-static widgets.
    const publicOptions = getWidgetExport(widget.type)
        .getPublicWidgetOptions(widget.options);
    
    return {
        ...widget,
        options: publicOptions,
    };
}
```

packages/perseus-core/src/widgets/interactive-graph/interactive-graph-util.test.ts
```diff
@@ -0,0 +1,40 @@
+import getInteractiveGraphPublicWidgetOptions from "./interactive-graph-util";
+
+import type {PerseusInteractiveGraphWidgetOptions} from "@khanacademy/perseus-core";
+
+describe("getInteractiveGraphPublicWidgetOptions", () => {
+    it("removes the `correct` field", () => {
+        const options: PerseusInteractiveGraphWidgetOptions = {
+            correct: {
+                type: "angle",
```
jeremywiebe, line 9: Sorry, I'm late to this PR... 

Should we have one more test here for `static: true`?  When that's the case, we need to keep the `correct` field as that's used in the render: https://github.com/Khan/perseus/blob/9cabe689a7aa143f95adf4556bf5c10d654a66ae/packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.tsx#L136

benchristel: The static field doesn't exist on `PerseusInteractiveGraphWidgetOptions`, only `WidgetOptions`, so we'll have to do this check elsewhere as in my comment above. The static field on `WidgetOptions` is passed to each widget by the renderer: https://github.com/Khan/perseus/blob/b07f2936fccb94cde5cd34ce258973581611416a/packages/perseus/src/renderer.tsx#L588.

jeremywiebe: 🤦‍♂️  Yikes, sorry! I totally missed the comments above relating to this already! I switched straight to the file changes and didn't see them.

PR #2195, opened 2025-02-05
[fool-tips] Tooltip is Dead. Long Live Tooltip.

SonicScrewdriver:
## Summary:

This PR implements the Wonderblocks ToolTip, while finally deleted the much loathed Perseus Tooltip component. As a result, we have a much more accessible tooltip experience! This solves the issue in LEMS-2815, as the Wonderblocks Tooltip has already solved this issue. 

I have also adjusted the default storybook story to show as many answer forms as possible, to help show the list view for testing. 

Test out the [new experience here.](https://650db21c3f5d1b2f13c02952-naypxoohhw.chromatic.com/?path=/docs/perseus-widgets-numeric-input--docs) 

## Screenshots: 
List View:
Screenshot 2025-02-05 at 10 00 54 AM

Single Answer Form: 
Screenshot 2025-02-05 at 10 01 05 AM

Single Answer Form with small viewport: 
Screenshot 2025-02-05 at 10 01 21 AM

Issue: LEMS-2815

## Test plan:
- tests pass 
- manual testing in Storybook


packages/perseus/src/styles/widgets/numeric.less
```diff
@@ -0,0 +1,19 @@
+.input-with-examples-tooltip {
```
SonicScrewdriver, line 1: While these styles used to live in perseus-renderer.less, I think it makes more sense in a numeric file now. We're going to be deleting Input Number, and Expression no longer shows answer formats. 

Unfortunately, the tooltip has a reset, so we need to redefine a couple of the text styles here.

packages/perseus/src/styles/perseus-renderer.less
```diff
@@ -527,13 +528,6 @@
     }
 }
 
-.perseus-formats-tooltip {
```
SonicScrewdriver, line 530: As these styles are really only relevant to the numeric input file now, I've added them over to a widget specific file.

mark-fitzgerald (approved):
Fantastic! Thanks for finding a way to use the Wonder Blocks Tooltip, and cleaning up the unused code.

PR #2196, opened 2025-02-05
[SR] Segment - fix 2 strings with typos

nishasy:
## Summary:
Two segment strings are missing the word "at"

Fixing them here.

Issue: none

## Test plan:
N/A - will have to confirm on the webapp PR.


Myranae (approved):
👍

PR #2198, opened 2025-02-05
Create a function to get the public options for a Grapher widget

benchristel:
Issue: https://khanacademy.atlassian.net/browse/LEMS-2759

## Test plan:

`yarn test`


Myranae (approved):
Looks good! I do worry about leaving in the available types as that seems like information that could be related to the answer, but that seems to be necessary for setting defaults and providing buttons. I imagine if there are multiple options, then it would only show the possibilities and not the correct one, so that's probably fine. Just need to check out those failed checks, though they seem unrelated.

PR #2201, opened 2025-02-05
[SR] Refactor - Update graphs to pass in i18n context

nishasy:
## Summary:
When getting the graphs' description strings via `renderGraphElements` in `mafs-graph.tsx`,
we're getting a component with a string with each component separately running the
Perseus i18n hook.

Instead, we can just gave `renderGraphElements` return a string, and we can pass down the
i18n info from the parent where the hook is called once.

This should be easier for people to follow when looking at the code in the future.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2800

## Test plan:
`yarn jest` to run the specific graph tests (`[graphType].test.tsx`)

Storybook
- Check each graph's SR tree to confirm that the "Interactive elements: ..." text
  is still correct.

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
@@ -692,19 +692,19 @@ const renderGraphElements = (props: {
         case "angle":
             return renderAngleGraph(state, dispatch);
         case "segment":
-            return renderSegmentGraph(state, dispatch);
+            return renderSegmentGraph(state, dispatch, i18n);
         case "linear-system":
-            return renderLinearSystemGraph(state, dispatch);
+            return renderLinearSystemGraph(state, dispatch, i18n);
         case "linear":
-            return renderLinearGraph(state, dispatch);
+            return renderLinearGraph(state, dispatch, i18n);
         case "ray":
-            return renderRayGraph(state, dispatch);
+            return renderRayGraph(state, dispatch, i18n);
         case "polygon":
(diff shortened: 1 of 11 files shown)


packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -692,19 +692,19 @@ const renderGraphElements = (props: {
         case "angle":
```
nishasy, line 692: Angle doesn't have the overall graph description yet, so the i18n context would go unused here. We can add it in the PR where we add the Angle "interactive elements" description.

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -692,19 +692,19 @@ const renderGraphElements = (props: {
...
             return renderPolygonGraph(state, dispatch, i18n, markings);
         case "point":
-            return renderPointGraph(state, dispatch);
+            return renderPointGraph(state, dispatch, i18n);
         case "circle":
-            return renderCircleGraph(state, dispatch);
+            return renderCircleGraph(state, dispatch, i18n);
         case "quadratic":
             return renderQuadraticGraph(state, dispatch, i18n);
         case "sinusoid":
```
nishasy, line 710: Same thing for sinusoid.

catandthemachines (approved):
Amazing! Thank you for cleaning up this formate and making the interactive graph widget more efficient. 🎉

PR #2202, opened 2025-02-05
Migrate Perseus to pnpm

jeremywiebe:
## Summary:

This PR migrates this repo from `yarn` to [pnpm](pnpm.io). The version of Yarn we are using is very outdated and is no longer supported. Khan Academy has made an engineering-wide decision to adopt pnpm as its "package manager of choice" and so let's adopt!

Notable changes:

  * Use `workspace:` syntax for in-repo dependencies
  * Use `catalog:` syntax for peer/dev dependencies (easier "global" upgrade of these) - and Github _just_ added support for `catalog:` support Dependabot [_yesterday_](https://github.blog/changelog/2025-02-04-dependabot-now-supports-pnpm-workspace-catalogs-ga/)!!
  * Build system is simplified because we don't have to worry about build order
  * 

Issue: LEMS-2825

## Test plan:

`pnpm build`
`pnpm dev`
`pnpm start` # storybook
`pnpm cypress:ci`


mark-fitzgerald (approved):
Storybook/Vite/Aphrodite adjustment looks good. Thanks for keeping it along for the configuration changes.

config/build/rollup.config.js
```diff
@@ -34,74 +32,6 @@ const rootDir = ancesdir(__dirname);
  *               from the default behavior.
  */
 
-// Kahn's algorithm
-// https://en.wikipedia.org/wiki/Topological_sorting#Kahn%27s_algorithm
-const topoSort = (yarnWorkspacesOutput) => {
```
jeremywiebe, line 39: Not needed.  WonderBlocks doesn't build in any dependency order, and after thinking about it a bit, we don't need to either. The build doesn't depend on build artifacts of other in-repo packages, so this simplifies our build a bunch!

config/build/rollup.config.js
```diff
@@ -225,10 +162,7 @@ const createConfig = (
                 },
             }),
             babel({
-                babelHelpers:
-                    platform === "browser" && format === "esm"
-                        ? "runtime"
-                        : "bundled",
+                babelHelpers: "runtime",
```
jeremywiebe, line 161: WonderBlocks uses [`runtime`](https://github.com/Khan/wonder-blocks/blob/9bbbb01d19e7da25ad7eadc30883e73b2f1473f9/build-settings/rollup.config.mjs#L44) for both ES and CJS builds, so we'll do the same!

https://www.npmjs.com/package/@rollup/plugin-babel#babelhelpers

vite.config.ts
```diff
@@ -0,0 +1,61 @@
+import {readFileSync} from "node:fs";
```
jeremywiebe, line 1: Most of this file is from the file that was moved from `dev/vite.config.ts`

vendor/hubble/package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "hubble",
+  "private": true,
```
jeremywiebe, line 3: Juuuuust so we don't try to publish this package!

.github/workflows/node-ci.yml
```diff
@@ -153,13 +158,20 @@ jobs:
             - name: Checking out latest commit
               uses: actions/checkout@v4
 
+            # MUST be before we install node_modules as that depends on finding
+            # Cypress binaries in this cache!
+            - uses: actions/cache@v4
+              with:
+                  path: ~/.cache/Cypress
+                  key: cypress-${{ runner.os }}-${{ hashFiles('pnpm-lock.yaml') }}
+
```
jeremywiebe, line 167: Cypress runs a post-install script (which we've allowed in `package.json`), but it doesn't seem to include downloading the Cypress binaries if we've cached `node_modules` (next step). By caching this [directory](https://docs.cypress.io/app/references/advanced-installation#Binary-cache), the post-install script is able to find a cached binary and connect it. 

On a clean install (or Cypress upgrade) the `node_modules` cache will be busted and the cypress npm package properly downloads the correct binary and caches it in `~/.cache/Cypress`.

.storybook/main.ts
```diff
@@ -1,4 +1,4 @@
-import viteConfig from "../dev/vite.config";
```
jeremywiebe, line 1: I moved some vite configs around. It always felt weird that we used `dev/vite.config.ts` as our base.  Now we have a shared base config in the root of the repo and both Storybook and `dev/` leverage it. 

This will pave the way in the future for us to use Vitest for testing (faster!!)

.storybook/main.ts
```diff
@@ -72,7 +72,7 @@ const config: StorybookConfig = {
             },
             // Fix from: https://github.com/storybookjs/storybook/issues/25256#issuecomment-1866441206
             assetsInclude: ["/sb-preview/runtime.js"],
-            plugins: [...viteConfig.plugins, lessWrapper],
+            plugins: [...(viteConfig.plugins ?? []), lessWrapper],
```
jeremywiebe, line 75: No guarantee that there are plugins defined in the base config.

.storybook/preview.tsx
```diff
@@ -1,9 +1,11 @@
 import * as React from "react";
 import {color} from "@khanacademy/wonder-blocks-tokens";
 import {RenderStateRoot} from "@khanacademy/wonder-blocks-core";
-import {Dependencies} from "@khanacademy/perseus";
```
jeremywiebe, line 4: The root package doesn't depend on one of its monorepo packages, so this needed to change to a relative path now that we're using pnpm.

dev/package.json
```diff
@@ -14,27 +14,27 @@
...
-        "vite": "^5.1.0",
-        "@vitejs/plugin-react": "^4.3.3"
+        "@khanacademy/kas": "workspace:*",
+        "@khanacademy/kmath": "workspace:*",
+        "@khanacademy/math-input": "workspace:*",
+        "@khanacademy/perseus-core": "workspace:*",
+        "@khanacademy/perseus-linter": "workspace:*",
+        "@khanacademy/pure-markdown": "workspace:*",
+        "@khanacademy/simple-markdown": "workspace:*",
+        "@khanacademy/wonder-blocks-banner": "catalog:",
```
jeremywiebe, line 24: `workspace:*` for in-repo dependencies, `catalog:` for the deps that we'll be provided at runtime (peer deps).

This should make managing peer and dev dependency versions so much easier!  Plus, Dependabot (Github) [just announced](https://github.blog/changelog/2025-02-04-dependabot-now-supports-pnpm-workspace-catalogs-ga/) full support for pnpm catalogs a few _days_ ago!

package.json
```diff
@@ -55,7 +55,6 @@
         "@testing-library/jest-dom": "^6.4.6",
         "@testing-library/react": "^16.0.0",
         "@testing-library/user-event": "^14.5.2",
-        "@types/classnames": "^2.3.1",
```
jeremywiebe, line 58: pnpm kept telling me that this library ships with its types built-in now... so we don't need this dep.

package.json
```diff
@@ -123,22 +125,24 @@
         "tiny-invariant": "^1.3.1",
         "typescript": "5.7.2",
         "typescript-coverage-report": "^0.7.0",
+        "vite": "5.1.0",
```
jeremywiebe, line 128: Moved up from `dev/`

package.json
```diff
@@ -123,22 +125,24 @@
...
         "vite-plugin-istanbul": "^5.0.0",
         "winston": "^3.7.2"
     },
     "scripts": {
-        "gen:parsers": "yarn --cwd packages/kas gen:parsers",
-        "prebuild": "yarn gen:parsers",
+        "preinstall": "npx -y only-allow pnpm",
```
jeremywiebe, line 133: This catches if someone runs `yarn...`

package.json
```diff
@@ -153,5 +157,13 @@
     "nyc": {
         "report-dir": "coverage/cypress/"
     },
-    "dependencies": {}
+    "packageManager": "pnpm@10.0.0+sha512.b8fef5494bd3fe4cbd4edabd0745df2ee5be3e4b0b8b08fa643aa3e4c6702ccc0f00d68fa8a8c9858a735a0032485a44990ed2810526c875e416f001b17df12b",
+    "pnpm": {
+        "onlyBuiltDependencies": [
```
jeremywiebe, line 162: This flags packages that we explicitly allow to run `post-install` scripts during `pnpm install`. A nice safety feature to protect us from supply-chain attacks.

packages/kas/package.json
```diff
@@ -23,19 +23,18 @@
     ],
     "scripts": {
         "prepublishOnly": "../../utils/package-pre-publish-check.sh",
-        "gen:parsers": "node src/parser-generator.ts",
-        "test": "bash -c 'yarn --silent --cwd \"../..\" test ${@:0} $($([[ ${@: -1} = -* ]] || [[ ${@: -1} = bash ]]) && echo $PWD)'"
```
jeremywiebe, line 27: I removed this. As far as I know, nobody cd's into the various package directories and runs tests. It's just as easy to run from the root, and you can just as easily run `pnpm test packages/kas` from the root!

packages/math-input/less/main.less
```diff
@@ -1,2 +1,2 @@
 @import "./overrides.less";
-@import (inline) "../../../node_modules/mathquill/build/mathquill.css";
+@import (inline) "../node_modules/mathquill/build/mathquill.css";
```
jeremywiebe, line 2: pnpm installs direct package dependencies into the package's local `node_modules` folder, so these relative paths now make more sense.

pnpm-workspace.yaml
```diff
@@ -0,0 +1,40 @@
+packages:
+    - "config/build"
+    - "dev"
+    - "packages/*"
+    - "vendor/*"
+
+catalog:
```
jeremywiebe, line 7: The magic of `pnpm` catalogs.  We can define the version of each peer dependency here and then update _all_ of our Perseus packages to the same version in one fell swoop!

jeremywiebe:
> **Size Change:** -191 kB (-12.89%) 👏
> 
> **Total Size:** 1.29 MB
> 
> Filename	Size	Change	
> `packages/kas/dist/es/index.js`	38.9 kB	-88 B (-0.23%)	
> `packages/kmath/dist/es/index.js`	10.4 kB	-76.4 kB (-87.99%)	🏆
> `packages/math-input/dist/es/index.js`	77.5 kB	-141 B (-0.18%)	
> `packages/perseus-core/dist/es/index.js`	27.7 kB	-18.6 kB (-40.12%)	🎉
> `packages/perseus-editor/dist/es/index.js`	688 kB	-158 B (-0.02%)	
> `packages/perseus-linter/dist/es/index.js`	22.1 kB	-114 B (-0.51%)	
> `packages/perseus-score/dist/es/index.js`	20.1 kB	-94.8 kB (-82.5%)	🏆
> `packages/perseus/dist/es/index.js`	379 kB	-273 B (-0.07%)	
> `packages/pure-markdown/dist/es/index.js`	3.54 kB	-118 B (-3.23%)	
> ℹ️ **View Unchanged**

I checked these size changes. We were bundling jQuery which we've marked as a dev/peer dep (meaning that we expect the host app to supply it). We no longer are. There are a few reductions in that we're using `babelHelpers: "runtime"` to allow us to use the host app's babel (which is what Wonder Blocks also does).

config/build/rollup.config.js
jeremywiebe: @kevinb-khan I tagged you mostly for these changes. I'm pretty sure we don't need to build in any dependency order.

.eslintrc.js
```diff
@@ -138,7 +138,7 @@ module.exports = {
             },
         },
         {
-            files: ["config/**", "utils/**", "testing/*"],
+            files: ["config/**", "utils/**", "testing/*", "dev/*"],
```
benchristel, line 141: Why did this change?

.github/workflows/node-ci.yml
```diff
@@ -86,55 +91,55 @@ jobs:
...
+                  pnpm -s prettier --check .
 
             - id: js-files
               name: Find .js(x)/.ts(x) changed files
               uses: Khan/actions@filter-files-v1
               with:
                   changed-files: ${{ steps.changed.outputs.files }}
                   extensions: ".js,.jsx,.ts,.tsx"
-                  files: "yarn.lock"
+                  files: "pnpm-lock.yarml"
```
benchristel, line 102:
```suggestion
                  files: "pnpm-lock.yaml"
```

.github/workflows/node-ci.yml
```diff
@@ -86,55 +91,55 @@ jobs:
 
             - name: Check formatting
               run: |
-                  yarn -s prettier --check .
+                  pnpm -s prettier --check .
```
benchristel, line 94: I don't think `-s` has any effect for `pnpm`. At least, it didn't seem to make any difference to the output for this command. `pnpm` is non-noisy by default.

.storybook/preview.tsx
benchristel: By "one of its monorepo packages," you mean `@khanacademy/perseus`? I think the resolutions for these monorepo package imports are defined in `vite.config.ts`. Maybe there's a problem with how that's being done?

config/build/rollup.config.js
benchristel: Does it still build if you reverse the topological sort? If it does, I think we're probably okay.

config/test/test.config.js
```diff
@@ -86,4 +86,9 @@ module.exports = {
     // Only output log messages on test failure. From:
     // https://github.com/facebook/jest/issues/4156#issuecomment-490764080
     // reporters: ["<rootDir>/config/test/log-on-fail-reporter.js"],
+
+    // We turn off Prettier as Prettier v3 is incompatible with Jest v29.
+    // Once they release Jest v30 we can switch to that:
+    // https://github.com/jestjs/jest/issues/14305
```
benchristel, line 92: Is there a Jira issue for this Jest upgrade? We should probably link to it here in a TODO.

vite.config.ts
```diff
@@ -0,0 +1,61 @@
...
+        const pkg = JSON.parse(readFileSync(packageJsonPath, "utf-8"));
+        packageAliases[pkg.name] = join(dirname(packageJsonPath), pkg.source);
+    },
+);
+
+export default defineConfig({
+    define: {
+        // This is used to determine if we are running in a
+        // Dev/Storybook environment.
+        "process.env.STORYBOOK": JSON.stringify(true),
```
benchristel, line 21: This is a funny way of spelling `"true"` 😂 

No change needed.

benchristel (changes requested):
Very cool! Thanks for doing this upgrade.

Requesting changes to fix a typo that will break things (noted inline). I have some other optional suggestions too, but in general this looks great!

.github/workflows/node-ci.yml
handeyeco: That's nice, I imagine that would speed things up a bit.

pnpm-workspace.yaml
handeyeco: Do we also need to update `utils/sync-dependencies.ts`?

config/build/rollup.config.js
jeremywiebe: Yep. I did a manual ordering in the reverse of what `topoSort()` produces and it built fine.

.github/workflows/node-ci.yml
jeremywiebe: On a random sampling of one other PR, [this workflow took](https://github.com/Khan/perseus/actions/runs/13247696841/job/36978127948) 1m 34s. 

On this PR, [this same workflow](https://github.com/Khan/perseus/actions/runs/13210137302/job/36881957372?pr=2202) took 1m 10s.  

So, yes, 24s faster!

.storybook/preview.tsx
jeremywiebe: > By "one of its monorepo packages," you mean @khanacademy/perseus?

Yes

> I think the resolutions for these monorepo package imports are defined in vite.config.ts. Maybe there's a problem with how that's being done?

We did have aliases. But with pnpm, we don't need them anymore as pnpm correctly sets up in-repo dependencies. I removed the aliases from the Vite config and this was the only thing that needed adjustment. I figured it was cleaner to have fewer things (and customizations) in Vite config if pnpm solves the same problem. Changing this to a relative path import worked (we were already importing the `DependenciesContext` using a relative path in this file).

pnpm-workspace.yaml
jeremywiebe: Oh yes! Thank-you for thinking of that! I'll update this PR.

.github/workflows/node-ci.yml
jeremywiebe: Ouch. Thank-you for catching that!

config/test/test.config.js
jeremywiebe: Added: https://khanacademy.atlassian.net/browse/LEMS-2858

vite.config.ts
jeremywiebe: Haha! True. It does quote it though... so there's that.

image

package.json
```diff
@@ -30,11 +25,16 @@
...
-        "@khanacademy/wonder-blocks-spacing": "^4.0.1",
-        "@popperjs/core": "^2.10.2",
+        "@khanacademy/wonder-blocks-button": "catalog:",
+        "@khanacademy/wonder-blocks-core": "catalog:",
+        "@khanacademy/wonder-blocks-icon": "catalog:",
+        "@khanacademy/wonder-blocks-layout": "catalog:",
+        "@khanacademy/wonder-blocks-pill": "catalog:",
+        "@khanacademy/wonder-blocks-switch": "catalog:",
+        "@khanacademy/wonder-blocks-tokens": "catalog:",
+        "@khanacademy/wonder-blocks-typography": "catalog:",
```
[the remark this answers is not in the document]
jeremywiebe: Thanks for taking notice of this. These were a mixture of a few things: 

1. Packages within this repo must now declare the dependencies they use (such as `wonder-blocks-switch`) because pnpm is stricter in how it resolves dependencies. A repo package cannot simply depend on some package that another package in the repo was already depending on. (we got by with this with `yarn`, especially in our dev tooling, because it sets everything up in the root `node_modules` - with a few exceptions).
2. The `testing/` and `.storybook/` folders are a special case in that they don't have their own `package.json`. So we need to define their deps in the root `package.json` (that's why there are a few `@khanacademy/wonder-blocks-*` packages still listed in the root package.json.

vite.config.ts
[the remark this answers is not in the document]
jeremywiebe: Good observation. I decided to be more careful and moved this `define` block into `dev/vite.config.ts` and `.storybook/main.ts`. I'm really hopeful to move to Vitest in the future and in that environment we probably don't want to define this constant.

benchristel (approved):
LGTM!

vite.config.ts
benchristel: All it does it convert it to a string. The following two expressions are equivalent:

```ts
JSON.stringify(true);
"true";
```

PR #2203, opened 2025-02-05
[whole-fraction-fix] Ensure that whole numbers are not accepted answers for Numeric Inputs that require Improper fractions.

SonicScrewdriver:
## Summary:
This PR is part of the Numeric Input project. 

This small addition ensures that users cannot be marked correctly for entering whole numbers when required to enter Improper fractions. 

Issue: LEMS-2802

## Test plan:
- Tests pass 
- Creation of two new tests to ensure bug fix works, and there's no regressions for tex

### DIFF
--- a/packages/perseus-score/src/widgets/numeric-input/score-numeric-input.test.ts
+++ b/packages/perseus-score/src/widgets/numeric-input/score-numeric-input.test.ts
@@ -214,6 +214,56 @@ describe("scoreNumericInput", () => {
         expect(score).toHaveBeenAnsweredCorrectly();
     });
 
+    it("rejects a strict improper and whole number answer", () => {
+        const rubric: PerseusNumericInputRubric = {
+            answers: [
+                {
+                    value: 3,
+                    status: "correct",
+                    maxError: 0.2,
+                    simplify: "optional",
+                    strict: true,
+                    answerForms: ["improper"],
+                    message: "",
+                },
+            ],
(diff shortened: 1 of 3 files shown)


packages/perseus-score/src/widgets/numeric-input/score-numeric-input.test.ts
```diff
@@ -214,6 +214,56 @@ describe("scoreNumericInput", () => {
...
+                    strict: true,
+                    answerForms: ["improper"],
+                    message: "",
+                },
+            ],
+            coefficient: true,
+        };
+
+        const userInput = {
+            currentValue: "\\frac{9}{3}",
```
mark-fitzgerald, line 259: Should other variations of fractions be checked, as well? (i.e. "9/3" and "9\3")?

mark-fitzgerald (approved):
Always love simple fixes. Thank you!

packages/perseus-score/src/widgets/numeric-input/score-numeric-input.test.ts
SonicScrewdriver: Good callout - I've added a test for 9/3! I'm removing the "nice" option (9\3) as the test revealed that we're hitting an "extra symbols error" before we get to that point, so it isn't helpful! Hopefully the error helps indicate to users what the issue is. I don't think this is something worth a larger refactor right now.

PR #2207, opened 2025-02-06
Create a function to get the public options for a Matrix widget

benchristel:
Issue: https://khanacademy.atlassian.net/browse/LEMS-2762

## Test plan:

`yarn test`

### DIFF
--- a/packages/perseus-core/src/widgets/matrix/matrix-util.ts
+++ b/packages/perseus-core/src/widgets/matrix/matrix-util.ts
@@ -1,8 +1,13 @@
 import type {PerseusMatrixWidgetOptions} from "@khanacademy/perseus-core";
 
+type MatrixPublicWidgetOptions = Pick<
+    PerseusMatrixWidgetOptions,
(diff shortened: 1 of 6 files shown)


packages/perseus-core/src/widgets/matrix/matrix-util.ts
```diff
@@ -0,0 +1,8 @@
+import type {PerseusMatrixWidgetOptions} from "@khanacademy/perseus-core";
+
+export default function getMatrixPublicWidgetOptions(
+    options: PerseusMatrixWidgetOptions,
+) {
```
Myranae, line 5: I think we need the return type added to this file and to this function :) So here it would be `MatrixPublicWidgetOptions`

Myranae (approved):
Looks good to me! Answers is used for styling and for building out the matrix though, so we will have a good chunk of work to do when hooking it up. Might have to process answers on the backend for the information the widget needs on the frontend and then send that along. Probably will have to do that for a lot of widgets though.

PR #2208, opened 2025-02-06
[SR Tree] Handle SR Tree when multiple editors are on the page

nishasy:
## Summary:
Right now, the SR Tree just gets the first mafs-graph container on the page and renders its SR content.
However, this does not work if there are also hint editors on the page. The hints will just show the
SR content for the first graph instead of reflecting their respective graph.

To fix this, I added a unique ID to the editor and passed that into the graph container and the SR Tree.
And SR Tree now finds by this unique ID rather than just the first graph on the page.

Issue: none

## Test plan:
`yarn jest packages/perseus-editor/src/widgets/interactive-graph-editor/components/interactive-graph-sr-tree.test.tsx`

Storybook
- Go to http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-ray
- Click "Add a hint" at the bottom
- Copy the interactive graph question
- Paste into the hint editor
- Confirm that the ray SR tree shows ray info
- Change the hint graph type
- Confirm that the hint SR tree no longer shows ray info

### Screenshots of the SR tree for a HINT none-type graph
| Before | After |
| --- | --- |
| image | Screenshot 2025-02-06 at 12 32 07 PM |

### DIFF
--- a/packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
+++ b/packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
@@ -16,7 +16,7 @@ import {
     type InteractiveGraphDefaultWidgetOptions,
     interactiveGraphLogic,
 } from "@khanacademy/perseus-core";
-import {View} from "@khanacademy/wonder-blocks-core";
+import {Id, View} from "@khanacademy/wonder-blocks-core";
 import {OptionItem, SingleSelect} from "@khanacademy/wonder-blocks-dropdown";
 import {Checkbox} from "@khanacademy/wonder-blocks-form";
 import {spacing} from "@khanacademy/wonder-blocks-tokens";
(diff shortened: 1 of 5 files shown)


packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
nishasy, line 1: This is showing as a bunch of changes, but it's really just one big indent.

packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
```diff
@@ -344,495 +344,559 @@ class InteractiveGraphEditor extends React.Component<Props> {
...
-                    <>
-                        <View style={styles.row}>
-                            <Checkbox
-                                label={
-                                    <LabelSmall>Show angle measures</LabelSmall>
-                                }
-                                checked={
-                                    // Don't show indeterminate checkbox state
-                                    !!this.props.correct?.showAngles
+            <Id>
```
nishasy, line 347: Using wonder-blocks-core `Id` here since it's a class component. If it were a functional component, I would have used the `React.useId()` hook and avoided all this indentation 🥲

benchristel (approved):
LGTM!

Protip: you can review changes that indent a lot of code using the "hide whitespace" feature.

Screen Shot 2025-02-07 at 8 52 28 AM

PR #2212, opened 2025-02-07
[LX] Add hairlines when point has focus

nishasy:
## Summary:
Right now, we have hairlines showing when a user moves a point using a mouse, but the hairlines
don't show up when using keyboard.

Adding hairlines when the point is focused in this PR.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2308

## Test plan:
`yarn jest packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.test.tsx`

Storybook
- Move the points around with mouse and confirm that that behavior has not changed - hairlines
  should still be present.
- Tab to the points with keyboard and confirm that the hairlines now show up.
- Switch the graph to "none" markings, and confirm that the hairlines no longer show up.

### Screenshots when point is focused

| Before | After |
| --- | --- |
| Screenshot 2025-02-06 at 4 15 36 PM | Screenshot 2025-02-06 at 4 15 43 PM |

### DIFF
--- a/.changeset/pretty-spies-tan.md
+++ b/.changeset/pretty-spies-tan.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+[LX] Add hairlines when point has focus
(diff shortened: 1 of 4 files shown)


SonicScrewdriver:
Screenshot 2025-02-10 at 1 15 31 PM

Love it, thank you! 

It's not related to this PR, but I do feel like we may need a solution for ensuring these hairlines are visible when they overlap with the axis lines.

SonicScrewdriver (approved):

PR #2213, opened 2025-02-07
[SR] Angle - Add the "interactive elements" overall graph description

nishasy:
## Summary:
This got missed when doing the angle SR work. We have to add the “interactive elements” description to the overall graph.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2839

## Test plan:
`yarn jest packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx`

Storybook
- Go to http://localhost:6006/iframe.html?globals=&id=perseuseditor-widgets-interactive-graph--interactive-graph-angle&viewMode=story
- Confirm that the "Interactive elements: ..." section is there in the screen reader tree.
- Change the values of the correct graph and confirm that the SR tree updates to reflect them in the interactive elements.

image

### DIFF
--- a/.changeset/calm-cats-ring.md
+++ b/.changeset/calm-cats-ring.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+[SR] Refactor - Update graphs to pass in i18n context
--- a/.changeset/neat-beds-smash.md
+++ b/.changeset/neat-beds-smash.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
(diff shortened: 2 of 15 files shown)


SonicScrewdriver (approved):
Nice simple little PR but another great win for A11Y! :) 

Question: Should we add the actual angle to the interactive elements description if the "Show angle measures" option is on?

nishasy:
@SonicScrewdriver

> Should we add the actual angle to the interactive elements description if the "Show angle measures" option is on?

I was thinking about that, but then I thought maybe it's more info than necessary just for listing the interactive elements, especially if they're going to come across it anyway when traversing through the angle 🤔

PR #2215, opened 2025-02-07
Update Sorter's public widget option function to use custom seed generation for shuffle

Myranae:
## Summary:
To shuffle server-side, we need to set up a seed other than the problemNum. We decided just to use Random.math for the seed. Once Sorter is receiving public widget options from the server, we can remove the client-side shuffle to ensure answer data is not transmitted to the client.

Issue: LEMS-2846

## Test plan:
- Confirm all checks pass
- Confirm Sorter still works as expected.


packages/perseus-core/src/utils/util.ts
```diff
@@ -21,6 +21,54 @@ const seededRNG: (seed: number) => RNG = function (seed: number): RNG {
...
+    return [
+        (h1 ^ h2 ^ h3 ^ h4) >>> 0,
+        (h2 ^ h1) >>> 0,
+        (h3 ^ h1) >>> 0,
+        (h4 ^ h1) >>> 0,
+    ];
+}
+
+// Note: Previously called sfc32
+function getSeedFunctionFromHashes(
```
benchristel, line 51: Since this function returns a random number generator, could it be called something like `createRandomNumberGenerator` or `seedRandomNumberGenerator`?

I don't think I'd draw the correct conclusions from the name `getSeedFunction` if I saw this being used.

packages/perseus-core/src/utils/util.ts
```diff
@@ -21,6 +21,54 @@ const seededRNG: (seed: number) => RNG = function (seed: number): RNG {
...
+    h4 = Math.imul(h2 ^ (h4 >>> 19), 2716044179);
+    return [
+        (h1 ^ h2 ^ h3 ^ h4) >>> 0,
+        (h2 ^ h1) >>> 0,
+        (h3 ^ h1) >>> 0,
+        (h4 ^ h1) >>> 0,
+    ];
+}
+
+// Note: Previously called sfc32
```
benchristel, line 50: It might be nice to link to the source on Stack Overflow in this comment.

packages/perseus-core/src/utils/util.ts
```diff
@@ -21,6 +21,54 @@ const seededRNG: (seed: number) => RNG = function (seed: number): RNG {
     };
 };
 
+// Note: Previously called cyrb128
```
benchristel, line 24: It might be nice to link to the source on Stack Overflow in this comment.

packages/perseus-core/src/utils/util.ts
```diff
@@ -21,6 +21,54 @@ const seededRNG: (seed: number) => RNG = function (seed: number): RNG {
...
+function getHashes(str: string): [number, number, number, number] {
```
benchristel, line 25:
```suggestion
function getHash(str: string): [number, number, number, number] {
```

I think this function should be called `getHash`. Semantically, it returns a _single_ 128-bit hash, but since JavaScript doesn't have a 128-bit number type, the value is split into four 32-bit integers.

packages/perseus-core/src/widgets/sorter/sorter-util.ts
```diff
@@ -17,13 +20,23 @@ type SorterPublicWidgetOptions = {
...
+    const correctToString = options.correct.join("");
+    const hashes = CoreUtil.getHashes(correctToString);
+    const getSeed = CoreUtil.getSeedFunctionFromHashes(...hashes);
+    const seedFromCorrect = getSeed();
+
+    const shuffledCorrect = CoreUtil.shuffle(
+        options.correct,
+        seedFromCorrect,
+        /* ensurePermuted */ true,
+    );
```
benchristel, line 32: Looking at this, I think it could be simplified. `CoreUtil.shuffle` has its own seeded random number generator, so we might not need the sfc32 algorithm.

```suggestion
    const [seed] = CoreUtil.getHashes(correctToString);

    const shuffledCorrect = CoreUtil.shuffle(
        options.correct,
        seed,
        /* ensurePermuted */ true,
    );
```

packages/perseus-core/src/widgets/sorter/sorter-util.ts
```diff
@@ -17,13 +20,23 @@ type SorterPublicWidgetOptions = {
 function getSorterPublicWidgetOptions(
     options: PerseusSorterWidgetOptions,
 ): SorterPublicWidgetOptions {
+    const correctToString = options.correct.join("");
```
benchristel, line 23: Minor: if we really want to ensure that different `correct` values result in different seeds, we should use `JSON.stringify()` instead of `join` here. The problem with `join` is that a value like `["1", "23"]` will produce the same seed as `["12", "3"]`.

This is unlikely to matter much in practice, but I'd prefer to set a good example and use `JSON.stringify` instead, in case this idea is reused somewhere where it _does_ matter.

packages/perseus/src/widgets/sorter/sorter.tsx
```diff
@@ -103,6 +103,10 @@ class Sorter extends React.Component<Props, State> implements Widget {
     };
 
     render(): React.ReactNode {
+        // TODO(LEMS-2841): Remove client-side shuffle once receiving public
+        //     options. The correct field will already be shuffled.
+        //     Probably easiest to replace "options" with "this.props.correct" when
+        //     setting up Sortable below. Or use the assignment to rename for clarity.
```
benchristel, line 109: I'm trying to think through how the sequencing of these changes will work. We might end up in a situation where a single version of Perseus needs to be able to handle both shuffled and unshuffled answers here. In order to distinguish between the shuffled and unshuffled cases, we'd need to include some additional information in the public options, like an `isCorrectShuffled` boolean.

The alternative is to coordinate the release of Perseus that removes the clientside shuffling with the change to webapp that starts consuming the public widget options. I think that kind of coordination limits our maneuverability and can get expensive, so I'd lean toward preferring the boolean. But I could be convinced otherwise.

packages/perseus-core/src/utils/util.ts
Myranae: Okay, I was thinking about that, but I wasn't sure how appropriate that would be. I'm aways thinking about things affecting cheating now 😅

benchristel: Now that you mention it... I suppose that a cheater could still get the correct answer from the shuffled options using something like the following procedure:

- List all permutations of the options listed in the `correct` field.
- For each permutation _P_, compute the hash that would result if that permutation were the original (correct) one. Then shuffle the options based on that hash.
- If the shuffled options match the `correct` value sent from the server, then _P_ is the correct answer!

To block this cheating vector, we would need the seed to be a random value that is stored in the database and _not_ revealed to the client. That would require a change to the Sorter editor to generate and store this seed. We would also have to backfill a seed for existing Sorter widgets.

packages/perseus-core/src/utils/util.ts
Myranae: Updated to `createRandomNumberGenerator`

packages/perseus-core/src/utils/util.ts
Myranae: Add the resource to both functions that were borrowed

packages/perseus-core/src/utils/util.ts
Myranae: Added the reference :)

packages/perseus-core/src/utils/util.ts
Myranae: Wow, that sounds like a lot, @benchristel. Is that something we need to do? That sounds like more work than just solving the problem XD

packages/perseus-core/src/utils/util.ts
Myranae: Updated! :)

packages/perseus-core/src/utils/util.ts
benchristel: It is more work, until someone makes a browser extension to automate it. That's how almost all cheating happens currently, AFAIK.

Maybe there's already some other piece of data which is stored in the database but not revealed to the client? If so, we could incorporate that into the seed to make it secure. I'm imagining something like the publish time of the content item. Though piping that info through to the options splitting code would be a pain.

packages/perseus-core/src/widgets/sorter/sorter-util.ts
Myranae: So then the strings will not be a string of all the answers, but a super literal string of the array (or arrays for Matcher) made into a strings and then combined to a string? 
image

packages/perseus/src/widgets/sorter/sorter.tsx
jeremywiebe: I like the idea of adding a field to be explicit about the state of the data (shuffled or not). That in itself doesn't reveal any information really.

packages/perseus-core/src/widgets/sorter/sorter-util.ts
Myranae: So then this would just grab the first number of the hash and use that in the shuffle? Does that decrease the security level of the hash? Is there a greater chance that it will be the same as a hash for another problem?

packages/perseus-core/src/widgets/sorter/sorter-util.ts
Myranae: Updated in this PR and the Matcher PR, but let me know if my understanding is incorrect :) Thanks!

packages/perseus/src/widgets/sorter/sorter.tsx
Myranae: I can add that. I'd just be adding it as a field in the public widget options type and then setting it to true in the get function. I think I would need to update the original widget options in the data-scheme.ts file if we want it there all the time, but I'm guess we can check for it and if it's true do something and if it's not there / false do something else.

packages/perseus-core/src/widgets/sorter/sorter-util.test.ts
```diff
@@ -15,10 +15,6 @@ describe("getSorterPublicWidgetOptions", () => {
         const publicWidgetOptions = getSorterPublicWidgetOptions(options);
 
         // Assert
-        expect(publicWidgetOptions).toEqual({
-            correct: ["$0.005$ kilograms", "$15$ grams", "$55$ grams"],
-            layout: "horizontal",
-            padding: true,
-        });
+        expect(publicWidgetOptions.correct).not.toEqual(options.correct);
```
benchristel, line 18: Minor: this assertion is a little loose / nonspecific. For example, it would pass if `correct` were set to an empty array.

I might prefer something like this:

```suggestion
        expect(publicWidgetOptions.correct).not.toEqual(options.correct);        
        expect(new Set(publicWidgetOptions.correct)).toEqual(new Set(options.correct));
```

That ensures that all the original elements are still there, but their order is different.

I wrote [a blog post about this topic](https://bensguide.substack.com/p/testing-randomness-with-loose-assertions)!

packages/perseus-core/src/widgets/sorter/sorter-util.ts
```diff
@@ -17,13 +21,22 @@ type SorterPublicWidgetOptions = {
 function getSorterPublicWidgetOptions(
     options: PerseusSorterWidgetOptions,
 ): SorterPublicWidgetOptions {
+    const shuffledCorrect = CoreUtil.shuffle(
+        options.correct,
+        Math.random() * 100,
```
benchristel, line 26: Why multiply by 100 here?

I think you can just do

```suggestion
        Math.random,
```

to use `Math.random` as the random number generator for shuffling.

benchristel (approved):
I left a couple suggestions inline, but overall this looks great!

Thanks for all the iteration on this. It turned out to be a lot more complicated than I thought it would be at first!

packages/perseus-core/src/widgets/sorter/sorter-util.ts
Myranae: This is based off of Matcher's PR, so there's probably more details there, but I mentioned in sync that I couldn't get the order to randomize on every test run unless I multiplied by 100. But that was with the parenthesis. Maybe it's different if I just pass Math.random. Will test!

benchristel: The reason multiplying by 100 makes it random is that `seededRNG` uses bitwise operations like `<< n` (left shift by n bits) and `& n` (bitwise and with n) which convert the operands to integers. `Math.random()` returns floats in the range [0, 1), and when those are converted to integers, the result is always 0.

If you multiply by 100, though, you get a random number in the range [0, 100), so there are 100 distinct seeds.

Even then, though, the results of shuffling won't be very random, because 100 different seeds is not that many. There are over 3 million possible permutations of a 10-element array, and if you only have 100 different seeds for your RNG, only 100 of those possible permutations will ever be selected.

This might not seem like that big a deal, but _it leaks information to cheaters_ because they can then infer the original order of the array from the shuffled order with better-than-chance accuracy.

In summary, I think passing `Math.random` as the RNG function, like I suggested above, is the way to go.

packages/perseus-core/src/widgets/sorter/sorter-util.test.ts
jeremywiebe: Just make sure we keep the `.not` in that assertion. :) 

```
expect(new Set(publicWidgetOptions.correct)).not.toEqual(new Set(options.correct));
```

Myranae: It looks like you don't add the ".not" here because this new expect is just to confirm that the two sets have the same elements.

packages/perseus-core/src/widgets/sorter/sorter-util.ts
Myranae: Thanks so much for pointing this out! Will update in Matcher too

packages/perseus-core/src/widgets/sorter/sorter-util.test.ts
benchristel: @Myranae is right, we want the sets of elements to be the same, but the arrays (the order) to be different.

PR #2217, opened 2025-02-07
[LEMS-2190] Move coord reverse logic into scoring logic

SonicScrewdriver:
## Summary:
This PR is part of the Interactive Graph project. 

It contains the work required to move the coordinate reversal logic from the `getGradeableGraph` function and move it directly into our scoring logic. This logic was initially left in `getGradeableGraph` while we waited for the legacy Angle graph to no longer be in use. 

The reason for this was that there were 2 conflicting bugs in the Angle graph previously:

1.  Bug 1 allowed graphs to be reflexive if the bottom coordinate was moved, even if `allowReflexiveAngles` was false.
2. Bug 2 failed to properly reverse the coordinates in the same situation, resulting in the answer being marked as incorrect. 

Unfortunately, the combination of these bugs meant that we we needed to keep the scoring logic unchanged until the feature flags for the new Angle graph were removed. 

Issue: LEMS-2190

## Test plan:
- moved tests to new location 
- tests pass

### DIFF
--- a/packages/perseus-core/src/data-schema.ts
+++ b/packages/perseus-core/src/data-schema.ts
@@ -1017,7 +1017,7 @@ export type PerseusGraphTypeRay = {
 type AngleGraphCorrect = {
     type: "angle";
     allowReflexAngles: boolean;
-    match: "congruent";
+    match?: "congruent";
     coords: [Coord, Coord, Coord];
 };
 
--- a/packages/perseus-score/src/widgets/interactive-graph/score-interactive-graph.test.ts
+++ b/packages/perseus-score/src/widgets/interactive-graph/score-interactive-graph.test.ts
@@ -358,3 +358,131 @@ describe("InteractiveGraph scoring on a point question", () => {
         expect(rubric).toEqual(rubricClone);
     });
 });
+
+describe("InteractiveGraph scoring on an angle question", () => {
+    it("marks the answer invalid if guess.coords is missing", () => {
+        // Arrange
+        const guess: PerseusGraphType = {type: "angle"};
+        const rubric: PerseusInteractiveGraphRubric = {
+            graph: {
+                type: "angle",
+            },
+            correct: {
+                type: "angle",
+                allowReflexAngles: false,
+                coords: [
+                    [-5, 0],
+                    [0, 0],
+                    [5, 5],
+                ],
+            },
+        };
+
+        // Act
+        const result = scoreInteractiveGraph(guess, rubric);
+
+        // Assert
+        expect(result).toHaveInvalidInput();
+    });
+
+    it("properly reverses coordinates if angle graph is reflexive when not allowed", () => {
+        // Arrange
+        const guess: PerseusGraphType = {
+            type: "angle",
+            coords: [
+                [-5, 0],
+                [0, 0],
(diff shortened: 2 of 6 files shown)


packages/perseus-score/src/widgets/interactive-graph/score-interactive-graph.ts
```diff
@@ -247,10 +248,28 @@ function scoreInteractiveGraph(
...
             rubric.correct.type === "angle"
         ) {
-            const guess = userInput.coords;
+            const coords = userInput.coords;
             const correct = rubric.correct.coords;
             const allowReflexAngles = rubric.correct.allowReflexAngles;
 
+            // While the angle graph should always have 3 points, our types
+            // technically allow for null values. We'll check for that here.
+            if (!coords) {
```
SonicScrewdriver, line 257: There's really no possible way for coords to ever be less than 3 coordinates, due to how the graphs are set up. 

I will make another ticket for setting a stricter type for the Angle Graph Coords, as it will require updating several tests that have only partially mocked data.

packages/perseus-core/src/data-schema.ts
```diff
@@ -1017,7 +1017,7 @@ export type PerseusGraphTypeRay = {
 type AngleGraphCorrect = {
     type: "angle";
     allowReflexAngles: boolean;
-    match: "congruent";
+    match?: "congruent";
```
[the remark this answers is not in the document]
nishasy: I think `match: "congruent"` is pretty odd and confusing. If there's a match at all, it's always "congruent"? I feel like this should be a boolean...

In any case, I think making it optional to match the schema makes sense here. Also it doesn't make sense for it to _not_ be optional, because then match is always there, and it's always "congruent."

packages/perseus-score/src/widgets/interactive-graph/score-interactive-graph.ts
nishasy: Could you make the ticket now and update this comment to start with a TODO(LEMS-XXXX)?

packages/perseus-score/src/widgets/interactive-graph/score-interactive-graph.ts
```diff
@@ -264,11 +283,9 @@ function scoreInteractiveGraph(
                 match = approximateEqual(...angles);
             } else {
                 /* exact */
-                match = // @ts-expect-error - TS2532 - Object is possibly 'undefined'. | TS2532 - Object is possibly 'undefined'.
```
nishasy, line 267: 🎉

packages/perseus-score/src/widgets/interactive-graph/score-interactive-graph.test.ts
```diff
@@ -358,3 +358,131 @@ describe("InteractiveGraph scoring on a point question", () => {
         expect(rubric).toEqual(rubricClone);
     });
 });
+
+describe("InteractiveGraph scoring on an angle question", () => {
+    it("marks the answer invalid if guess.coords is missing", () => {
```
nishasy, line 363: Thank you for adding this test!

nishasy (approved):
I like it!

PR #2220, opened 2025-02-07
[SR] Stop graph from exposing random empty image to screen readers

nishasy:
## Summary:
When using a screen reader in the interactive graph widget, it would read
an empty image after the axis labels. This is bad practice, as it can be
confusing to screen reader users to hear random elements with no context.

To remove this, I had to wrap the Mafs graph in a `View` and add the
`aria-hidden` property, but only when the graph doesn't have locked figures.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2721

## Test plan:
Storybook
- On Chrome or Firefox, go to http://localhost:6006/iframe.html?globals=&args=&id=perseuseditor-widgets-interactive-graph--interactive-graph-segment&viewMode=story
- Use a screen reader to go through an interactive graph
- Note that after reading "x, math" and "y, math" (the axis labels), it does NOT read an empty image.

- On Chrome or Firefox, go to http://localhost:6006/iframe.html?globals=&args=&id=perseuseditor-widgets-interactive-graph--locked-figures&viewMode=story
- Use a screen reader to go through an interactive graph
- Note that it still reads all the locked figure aria labels

### DIFF
--- a/.changeset/funny-jeans-bake.md
+++ b/.changeset/funny-jeans-bake.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+[SR] Stop graph from exposing random empty image to screen readers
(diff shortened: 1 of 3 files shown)


SonicScrewdriver (approved):
Love it, thank you for the fix! :D

PR #2221, opened 2025-02-08
[SR] Add aria-labels for x- and y-axis labels

nishasy:
## Summary:
Right now if a person using a screen reader traverses through the graph, they hear "x, math" and "y, math" for the axis labels, or whatever the x and y labels are updated to by the content authors.

We need to add a description for this so people know what x and y are referring to.

- Added "X-axis" and "Y-axis" strings as `aria-label`s for the `AxisLabel` elements.
- This means that it reads the actual label as a child of this axis label group
- This is my preferred approach so that it already handles math speech wehen reading the
  TeX label, and we don't have to manually use the speech rule engine again redundantly

Issue: https://khanacademy.atlassian.net/browse/LEMS-2844

## Test plan:
`yarn jest packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx`

Storybook
- Go to any interactive graph editor story (http://localhost:6006/iframe.html?globals=&args=&id=perseuseditor-widgets-interactive-graph--interactive-graph-segment&viewMode=story)
- Use the screen reader to confirm that the axis labels now say "X-axis, group"
- Continue using the screen reader and make sure that the actual axis label is read
- Update the x and y labels to have some kind of math TeX like `\frac{1}{2}`
- Repeat the above steps and confirm that the screen reader reads the math correctly

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
@@ -17,6 +17,8 @@ import type {InteractiveGraphState} from "./types";
 import type {GraphRange} from "@khanacademy/perseus-core";
 import type {UserEvent} from "@testing-library/user-event";
 
+const baseMafsProps = getBaseMafsGraphPropsForTests();
+
 function expectLabelInDoc(label: string) {
     expect(screen.getByLabelText(label)).toBeInTheDocument();
 }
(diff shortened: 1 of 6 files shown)


packages/perseus/src/widgets/interactive-graphs/mafs-graph.test.tsx
```diff
@@ -17,6 +17,8 @@ import type {InteractiveGraphState} from "./types";
 import type {GraphRange} from "@khanacademy/perseus-core";
 import type {UserEvent} from "@testing-library/user-event";
 
+const baseMafsProps = getBaseMafsGraphPropsForTests();
```
nishasy, line 20: While I was in this file, I took this as an opportunity to reduce the number of function calls being made.

SonicScrewdriver (approved):
Much better, thank you!

PR #2223, opened 2025-02-10
[Bugfix] Interactive Graph crashes in editor when setting domain for locked function

mark-fitzgerald:
## Summary:
When editing an Interactive Graph widget and specifying only one of the domain values for a locked function, the exercise is then unable to be loaded in the editor once it has been saved and re-opened. The problem was that the logic handling the domain values (an array) wasn't set up to handle null values within the array.

Issue: CP-8932

## Test plan:
1. Open an exercise editor
2. Add an interactive graph widget
3. Add a locked function
4. Set the value of one of the domain options to some number (leave the other one blank)
5. Click the "Save All" button for the editor
6. When the page reloads, the page should render (previously, it was showing the "Oops. Something went wrong" message.


benchristel (dismissed):
Thanks for fixing this! I think there's a remaining bug, though, when the elements of `domain` are 0. See my comment inline for details.

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-function-settings.tsx
```diff
@@ -122,7 +129,7 @@ const LockedFunctionSettings = (props: Props) => {
             dedicated code to convert empty to Infinity.
      */
     function handleDomainChange(limitIndex: number, newValueString: string) {
-        const newDomainEntries = [...domainEntries];
+        const newDomainEntries = [...domainEntries] satisfies [string, string];
```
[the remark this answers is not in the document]
mark-fitzgerald: No reason, other than that's where my mind was at. I'll correct this.

packages/perseus-core/src/data-schema.ts
```diff
@@ -871,7 +871,7 @@ export type LockedFunctionType = {
     strokeStyle: LockedLineStyle;
     equation: string; // This is the user-defined equation (as it was typed)
     directionalAxis: "x" | "y";
-    domain?: Interval;
+    domain?: [min: number | null, max: number | null];
```
mark-fitzgerald, line 874: As per the referenced ticket, `null` values are being passed into the widget (and editor) because of the use of JSON and how it is dropping `Infinity` values.

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-function-settings.tsx
```diff
@@ -57,35 +56,38 @@ const LockedFunctionSettings = (props: Props) => {
...
     //     entering a negative value, as well as representing Infinity as an empty string.
     // This variable is used when specifying the values of the input fields.
-    const [domainEntries, setDomainEntries] = useState([
-        domain && domain[0] !== -Infinity ? domain[0].toString() : "",
-        domain && domain[1] !== Infinity ? domain[1].toString() : "",
-    ]);
+    const getDomainStringValues = (domain): [string, string] => {
+        return [
+            domain && Number.isFinite(domain[0]) ? domain[0].toString() : "",
+            domain && Number.isFinite(domain[1]) ? domain[1].toString() : "",
```
mark-fitzgerald, line 74: `Number.isFinite()` handles both `Infinity` and `null` values - returns false. So, no need to also check for truthiness of the individual domain values.

packages/perseus/src/widgets/interactive-graphs/interactive-graph.test.tsx
```diff
@@ -1046,100 +1046,186 @@ describe("Interactive Graph", function () {
...
-            // Act
-            // eslint-disable-next-line testing-library/no-container, testing-library/no-node-access
-            const functionPlots = container.querySelectorAll(
-                ".locked-function path",
-            );
+                // Act
+                // eslint-disable-next-line testing-library/no-container, testing-library/no-node-access
+                const functionPlots = container.querySelectorAll(
+                    ".locked-function path",
+                );
```
mark-fitzgerald, line 1061: The extra spacing is due to moving this code within a `describe()` block.

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-function-settings.test.tsx
```diff
@@ -766,6 +836,133 @@ describe("Locked Function Settings", () => {
...
+                await userEvent.click(autoGenButton);
+
+                // Assert
+                expect(onChangeProps).toHaveBeenCalledWith({
+                    ariaLabel:
+                        "Function with equation y=x^2. Appearance solid gray.",
+                });
+            });
+
+            test("aria label auto-generates with partial domain information", async () => {
```
nishasy, line 865: Nit: I think the expected practice at Khan is to write multiple tests for each specific case. Can you break this up into separate cases for each test so that each has only one "Arrange," "Act," and "Assert"?

Alternatively, you can also use `test.each` and pass in the domain values and expected labels.

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-function-settings.tsx
```diff
@@ -362,13 +363,13 @@ const LockedFunctionSettings = (props: Props) => {
                         ...getDefaultFigureForType("label"),
                         // Vertical offset for each label so they
                         // don't overlap.
-                        coord: [0, -(labels?.length ?? 0)],
+                        coord: [0, -labels.length],
```
nishasy, line 366: Much nicer! Thank you for streamlining the labels code

packages/perseus/src/widgets/interactive-graphs/interactive-graph.test.tsx
```diff
@@ -1046,100 +1046,186 @@ describe("Interactive Graph", function () {
...
+                // Assert
+                expect(PlotOfXMock).toHaveBeenCalledWith(
+                    expect.objectContaining({
+                        ...expectedParameters,
+                        domain: [-Infinity, 3],
+                    }),
+                    {},
+                );
+
+                // Act - no lower limit specified
```
nishasy, line 1173: Extending my above nit to all the tests. Can you break this up into two separate tests? A "plots the equation with partially supplied domains (no upper limit specified)" case and a "plots the equation with partially supplied domanis (no lower limit specified)" case.

nishasy (approved):
Looks good to me! Just one nit about test structure.

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-function-settings.test.tsx
mark-fitzgerald: I keep forgetting about `test.each`. I do that - thanks!

PR #2224, opened 2025-02-11
[SR] Add minimal instructions for how to interact with graph

nishasy:
## Summary:
Adding instructions for how to interact with the interactive elements in the graph
as per the [SRUX Doc](https://khanacademy.atlassian.net/wiki/spaces/LC/pages/3261071420/Proposal+Screen+Reader+User+Experience+SRUX?focusedCommentId=3566666164).

Note: More detailed instructions for specific graph types are part of phase 2.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2783

## Test plan:

Storybook
http://localhost:6006/?path=/docs/perseuseditor-widgets-interactive-graph--docs


packages/perseus/src/widgets/interactive-graphs/interactive-graph.testdata.ts
```diff
@@ -1014,6 +1017,7 @@ export const staticGraphQuestionWithAnotherWidget: () => PerseusRenderer =
 
 export const interactiveGraphWithAriaLabel: PerseusRenderer =
     interactiveGraphQuestionBuilder()
+        .withNoInteractiveFigure()
```
nishasy, line 1020: Updating this so that the test for full graph aria labels can focus on just the graph without any interactive elements.

packages/perseus/src/widgets/interactive-graphs/interactive-graph.test.tsx
```diff
@@ -73,6 +76,10 @@ const updateWidgetState = (renderer: Renderer, widgetId: string, update) => {
     update(state[widgetId]);
     renderer.restoreSerializedState(state);
 };
+const commonInstructions =
+    "Use the Tab key to move through the interactive elements in the graph. When an interactive element has focus, use the Arrow keys to move it.";
+const unlimitedInstructions =
+    "Press Shift + Enter to interact with the graph. Use the Tab key to move through the interactive elements in the graph and access the graph Action Bar. When an interactive element has focus, use the Arrow keys to move it or use the Delete key to remove it from the graph. Use the buttons in the Action Bar to add or adjust elements within the graph.";
```
SonicScrewdriver, line 82: Nit: Would it be possible to use the exported mock strings here instead of redeclaring them in the test? I could see us updating the instructions over time, and it would require us to update these strings in multiple places.

SonicScrewdriver (approved):
Love it! It's really exciting to see some actual instructions for users on how to use our widgets!!

packages/perseus/src/widgets/interactive-graphs/interactive-graph.test.tsx
nishasy: That's true, but I was thinking of not using the exported mock strings in case they change _unexpectedly_ and we'd want to catch them in the test. Thoughts?

nishasy: I guess the mocks are mocks anyway, and they wouldn't reflect if the real strings are changed unexpectedly 🤔

SonicScrewdriver: Hmm great question! I would think that an unexpected change to the real strings is far more likely to be intentional if they also update the mock strings identically. 🤔 I could be misunderstanding the purpose for the mockStrings, but my impression was that they were to help write the tests. I'm a little out of the loop on our string set up now.

nishasy: I think the mock strings are what get used by the renderer in local dev, and the actual strings only get used in prod? I was under the impression that that's why we don't catch string errors until we bump Perseus in webapp.

packages/perseus/src/widgets/interactive-graphs/graphs/angle.test.ts
nishasy, line 1: Not deleted, just renamed to allow rendering React components.

PR #2227, opened 2025-02-11
Accept 'scientific' as a buttonSets value when parsing Expression widgets

benchristel:
Issue: https://khanacademy.atlassian.net/browse/LEMS-2861

## Test plan:

`yarn test`

### DIFF
--- a/packages/perseus-core/src/parse-perseus-json/general-purpose-parsers/test-helpers.ts
+++ b/packages/perseus-core/src/parse-perseus-json/general-purpose-parsers/test-helpers.ts
@@ -24,3 +24,10 @@ export function parseFailureWith(
 export function summonParsedValue<P extends Parser<any>>(): ParsedValue<P> {
     return "fake summoned value" as any;
 }
+
+// Summons a value of type T out of thin air!*
+// *Terms and conditions apply. Value will not actually be returned.
+// For use in typetests only.
+export function summon<T>(): T {
+    return "fake summoned value" as any;
+}
--- a/packages/perseus-core/src/parse-perseus-json/perseus-parsers/legacy-button-sets.ts
+++ b/packages/perseus-core/src/parse-perseus-json/perseus-parsers/legacy-button-sets.ts
@@ -1,0 +1,21 @@
+import {array, enumeration} from "../general-purpose-parsers";
+import {defaulted} from "../general-purpose-parsers/defaulted";
+
+export const parseLegacyButtonSet = enumeration(
+    "basic",
+    "basic+div",
+    "trig",
+    "prealgebra",
+    "logarithms",
+    "basic relations",
+    "advanced relations",
+    "scientific",
+);
+
+export const parseLegacyButtonSets = defaulted(
+    array(parseLegacyButtonSet),
+    // NOTE(benchristel): I copied the default buttonSets from
+    // expression.tsx. See the parse-perseus-json/README.md for
+    // an explanation of why we want to duplicate the default here.
+    () => ["basic", "trig", "prealgebra", "logarithms"] as const,
+);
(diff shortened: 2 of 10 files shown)


packages/perseus-core/src/parse-perseus-json/general-purpose-parsers/test-helpers.ts
```diff
@@ -24,3 +24,10 @@ export function parseFailureWith(
 export function summonParsedValue<P extends Parser<any>>(): ParsedValue<P> {
     return "fake summoned value" as any;
 }
+
+// Summons a value of type T out of thin air!*
+// *Terms and conditions apply. Value will not actually be returned.
+// For use in typetests only.
```
Myranae, line 30: 😂

packages/perseus-core/src/parse-perseus-json/regression-tests/parse-perseus-json-regression.test.ts
```diff
@@ -15,6 +15,9 @@ describe("parseAndTypecheckPerseusItem", () => {
 
         // This strange-looking assertion style results in the failure message
         // being printed if parsing fails, so the test is easier to debug.
+        expect((result as any).detail?.message).toBe(undefined);
```
Myranae, line 18: Does this work because if there is a message, then it will say something like "expected undefined but got <message>"? And the only time there is a message is when it fails?

packages/perseus-core/src/parse-perseus-json/regression-tests/parse-perseus-json-regression.test.ts
```diff
@@ -15,6 +15,9 @@ describe("parseAndTypecheckPerseusItem", () => {
...
+        // Backstop assertion in case the format of failures changes so
+        // `detail.message` doesn't exist anymore.
         expect(result).toEqual(expect.objectContaining({type: "success"}));
```
Myranae, line 21: How does this double expect work?

packages/perseus-core/src/parse-perseus-json/perseus-parsers/legacy-button-sets.ts
```diff
@@ -0,0 +1,21 @@
...
+    "advanced relations",
+    "scientific",
+);
+
+export const parseLegacyButtonSets = defaulted(
+    array(parseLegacyButtonSet),
+    // NOTE(benchristel): I copied the default buttonSets from
+    // expression.tsx. See the parse-perseus-json/README.md for
+    // an explanation of why we want to duplicate the default here.
+    () => ["basic", "trig", "prealgebra", "logarithms"] as const,
```
Myranae, line 20: So this shouldn't contain `scientific` because it represents old content and needs to not change over time to prevent old exercises from becoming "unsolvable or otherwise misconfigured," is that correct? Want to make sure I read the right part of the readme 😂

Myranae (approved):
Looks good! Got a little lost in the parsing functions, but I can see you're adding `scientific` and have some typetests to make sure everything stays in sync. Thanks for the bug fix! Make sure to wait until the deploy block is released before landing~

packages/perseus-core/src/parse-perseus-json/perseus-parsers/legacy-button-sets.typetest.ts
```diff
@@ -0,0 +1,13 @@
+import {summon} from "../general-purpose-parsers/test-helpers";
+
+import type {parseLegacyButtonSet} from "./legacy-button-sets";
+import type {LegacyButtonSets} from "../../data-schema";
+import type {ParsedValue} from "../parser-types";
+
+type ParsedButtonSet = ParsedValue<typeof parseLegacyButtonSet>;
```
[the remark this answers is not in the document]
benchristel: - `ParsedValue<P>` gets the result type of some parser type `P`.
- `typeof parseLegacyButtonSet` is the type of the `parseLegacyButtonSet` parser.
- so this whole line is getting result type of the `parseLegacyButtonSet` parser, which _should_ be identical to `LegacyButtonSets` from data-schema.ts. If there's any difference (e.g. extra or missing enum values on either side) this typetest will fail.

packages/perseus-core/src/parse-perseus-json/regression-tests/parse-perseus-json-regression.test.ts
benchristel: Yep!

PR #2230, opened 2025-02-12
[SR] Add screen reader support for pi-based numbers

nishasy:
## Summary:
The Sinusoid interactive graph type often uses pi-based ticks. We have support for this
visually, but the screen reader still reads out these numbers as decimal points.

Adding support for the screen reader to read pi-based numbers here in the `srFormatNumber()`
function, through which all coordinates should already be passed.

NOTE: The screen reader unfortunately does not read this with the math speech rule engine,
but that adds another layer of complexity that we may want to come back to later.
As is, it's not ideal, but it's workable. Sub-note: Manually adding the words for these
specific cases would be a nightmare for translations.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2484

## Test plan:
`yarn jest packages/perseus/src/widgets/interactive-graphs/graphs/screenreader-text.test.ts`

Storybook
http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-sinusoid-with-pi-ticks

### DIFF
--- a/.changeset/chatty-spoons-laugh.md
+++ b/.changeset/chatty-spoons-laugh.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/perseus": patch
+"@khanacademy/perseus-editor": patch
+---
+
+[SR] Add screen reader support for pi-based numbers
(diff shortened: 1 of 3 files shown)


SonicScrewdriver (approved):
Gorgeous work

PR #2231, opened 2025-02-12
[LEMS-2850] Answerless Dropdown

handeyeco:
## Summary:
Nothing interesting really. Dropdown didn't need the answers at all, so I just made sure it wouldn't have access and added tests to make sure it was interactive with/without answer data.

Issue: LEMS-2850

## Test plan:
Nothing should change. Dropdown should continue to be interactive and scorable.


packages/perseus/src/widgets/dropdown/dropdown.test.ts
```diff
@@ -127,4 +134,144 @@ describe("Dropdown widget", () => {
         // Assert
         expect(screen.getByLabelText("Select an answer")).toBeInTheDocument();
     });
+
+    describe("interactive: full vs answerless", () => {
```
handeyeco, line 138: Pretty much all tests and some minor cleanup in this PR

benchristel (approved):
LGTM!

packages/perseus/src/widgets/dropdown/dropdown.test.ts
```diff
@@ -127,4 +133,94 @@ describe("Dropdown widget", () => {
...
+        beforeAll(() => {
+            registerAllWidgetsForTesting();
```
Myranae, line 139: Is this required to use the score and split functions?

PR #2233, opened 2025-02-12
[LEMS-2849/radio-answerless] add broken test for answerful requirement

handeyeco:
## Summary:
[LEMS-2849/radio-answerless] test deriveNumCorrect
[LEMS-2849/radio-answerless] all tests passing
[LEMS-2849/radio-answerless] derive numCorrect in upgrader

Issue: LEMS-2849

## Test plan:


packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
```diff
@@ -47,3 +53,107 @@ export const parseRadioWidget: Parser<RadioWidget> = parseWidget(
...
+        version: {major: 2, minor: 0},
+        options: {
+            ...options,
+            numCorrect: deriveNumCorrect(options),
+        },
+    };
+}
+
+const version0 = optional(object({major: constant(0), minor: number}));
+const parseRadioWidgetV0: Parser<RadioWidget> = parseWidgetWithVersion(
```
handeyeco, line 107: So Radio version 0 and 1 already existed, but for some reason there wasn't an existing migration for Radio. The parser got upset with me when I tried to add a migration from 1 to 2 without having one for 0 to 1.

packages/perseus/src/widgets/radio/radio.ts
```diff
@@ -106,19 +105,16 @@ const transform = (
 ): RenderProps => {
     const choices = _choiceTransform(widgetOptions, strings, problemNum);
 
-    const numCorrect: number = _.reduce(
-        widgetOptions.choices,
-        function (memo, choice) {
-            return choice.correct ? memo + 1 : memo;
-        },
-        0,
-    );
```
handeyeco, line 115: The main idea is to pull this logic out of `transform` and call it during the upgrade process. That way the widget can do all the pre-scoring things it needs to without answers (ie. render and be answerable).

packages/perseus/src/widgets/radio/__tests__/answerless-radio.test.tsx
```diff
@@ -0,0 +1,177 @@
...
+            screen.getByRole("checkbox", {name: "(Choice A) Correct 1"}),
+        );
+        await userEvent.click(
+            screen.getByRole("checkbox", {name: "(Choice B) Correct 2"}),
+        );
+
+        // assert that functionality previous based on answers still works
+        expect(screen.getByRole("group", {name: "Choose 2 answers:"}));
+
+        const userInput = renderer.getUserInputMap();
```
handeyeco, line 171: This is important because we can do everything we need to do pre-scoring with the answerless data now as far as I can tell.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
benchristel: I didn't write a migration for v0 -> v1 because it's fairly trivial and I don't think `hasNoneOfTheAbove` is even used anymore.

```
    "1": (v0props: any): PerseusRadioWidgetOptions => {
        const {noneOfTheAbove, ...rest} = v0props;

        if (noneOfTheAbove) {
            throw new Error(
                "radio widget v0 no longer supports auto noneOfTheAbove",
            );
        }

        return {
            ...rest,
            hasNoneOfTheAbove: false,
        };
    },
```

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
```diff
@@ -47,3 +53,107 @@ export const parseRadioWidget: Parser<RadioWidget> = parseWidget(
...
+    widget: ParsedValue<typeof parseRadioWidgetV1>,
+): RadioWidget {
+    const {options} = widget;
+    return {
+        ...widget,
+        version: {major: 1, minor: 0},
+        options: {
+            ...options,
+        },
+    };
```
benchristel, line 151: Do we even need to copy the options here?

```suggestion
    return {
        ...widget,
        version: {major: 1, minor: 0},
    };
```

Alternatively, maybe we should remove `noneOfTheAbove` and set `hasNoneOfTheAbove` to `false`, like in the original upgrade function?

packages/perseus/src/widgets/radio/__tests__/answerless-radio.test.tsx
```diff
@@ -0,0 +1,177 @@
...
+        const score = scorePerseusItem(fullItem, userInput, "en");
+
+        // Assert
+        expect(score).toHaveBeenAnsweredCorrectly();
+    });
+
+    it("is interactive with stripped widget options", async () => {
```
benchristel, line 130: We're using a few different terms for this concept:

- public options
- stripped options
- answerless options

I'd prefer to standardize on one, for the sake of future maintainers.

packages/perseus-core/src/widgets/radio/radio-upgrade.ts
```diff
@@ -1,8 +1,31 @@
...
 
-export const currentVersion = {major: 1, minor: 0};
+    return (
+        numCorrect ??
+        choices.reduce(
+            (acc: number, curr: PerseusRadioChoice) =>
+                curr.correct ? acc + 1 : acc,
+            0,
+        )
+    );
```
benchristel, line 18: I see this algorithm was copied from the component code. I think there's a simpler way. I suspect the JIT can optimize this:

```ts
choices.filter((c) => c.correct).length
```

to avoid creating the intermediate array. Even if it doesn't... 🤷

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
```diff
@@ -1,23 +1,29 @@
...
-import type {Parser} from "../parser-types";
+import type {ParsedValue, Parser} from "../parser-types";
 
-export const parseRadioWidget: Parser<RadioWidget> = parseWidget(
+const version2 = optional(object({major: constant(2), minor: number}));
+const parseRadioWidgetV2: Parser<RadioWidget> = parseWidgetWithVersion(
+    version2,
     constant("radio"),
     object({
+        numCorrect: optional(number),
```
benchristel, line 26: This makes me slightly nervous. `numCorrect` only needs to be optional because we'll be running this parser on two different inputs:

- data loaded from datastore, where `numCorrect` is usually not present (and maybe never needs to be present, since it can be derived?)
- data sent to the the client, where `numCorrect` _must_ be present if the data is answerless.

I worry that if the parsers accumulate a lot of irregularities of this kind, we'll be back in the position of not really understanding our data schema.

But I also think that it would be a pain in the butt to maintain separate parsers for answerless and answerful data.

One idea I'm toying with is letting the `ParseContext` keep track of whether we're parsing answerless or answerful data. Then we could write widget options parsers like this:

```
object({
  numCorrect: ifAnswerful(optional(number)).ifAnswerless(number),
})
```

But maybe that's overkill? I guess one thing I don't understand is whether the plan is to persist `numCorrect` in the answerful data or not.

benchristel (approved):
I left a few minor comments and suggestions inline, but nothing blocking. LGTM!

jeremywiebe (approved):
Nice. I like the approach of removing `countChoices` in get `getPublicWidgetOptions`.

packages/perseus/src/widgets/radio/__tests__/answerless-radio.test.tsx
jeremywiebe: The functions are all named `getPublicWidgetOptions`... so it feels like we've gone with "public".  I'm a bit averse to "stripped" after starting with it. It feels aggressive. "Answerless" isn't always 100% correct because there are situations where we remove data that isn't the answer, but could reveal info about the answer. 🤷‍♂️

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
handeyeco: For NumericInput v2 `numCorrect` will:

- always be in the data store
- always be in the full widget options
- will conditionally be in the "public" widget options

I [added a commit](https://github.com/Khan/perseus/pull/2233/commits/dc6d95b84f08371a776f25d4b196090d70dc569d#diff-4000a1441bb38cf8c1c2b7104152b3463d66f1a603f30fb17b3ffc44fa981217) since your comment that changes things. Maybe this is resolved now?

packages/perseus/src/widgets/radio/__tests__/answerless-radio.test.tsx
handeyeco: I personally think "public" is the most incorrect. Answerless/answerful are both public, just at different times. But I can change it for this PR.

packages/perseus-core/src/data-schema.ts
```diff
@@ -1343,6 +1343,9 @@ export type PerseusRadioWidgetOptions = {
     // If multipleSelect is enabled, Specify the number expected to be correct.
     // NOTE: perseus_data.go says this is required even though it isn't necessary.
     countChoices?: boolean;
+    // How many of the choices are correct, which is conditionally used to tell
+    // learners ahead of time how many options they'll need.
+    numCorrect?: number;
```
Myranae, line 1348: Does updating this file require a stakeholder update? Apologies for pointing this out each time, but the memories of what Ned went through and Jeremy always talking about protecting this file are strong 😅

packages/perseus-core/src/widgets/radio/radio-util.ts
```diff
@@ -42,16 +43,37 @@ function getRadioChoicePublicData(
...
+ *
+ * So only include/use numCorrect when we know it's useful.
+ */
+export function usesNumCorrect(
+    multipleSelect: PerseusRadioWidgetOptions["multipleSelect"],
+    countChoices: PerseusRadioWidgetOptions["countChoices"],
+    numCorrect: PerseusRadioWidgetOptions["numCorrect"],
+) {
+    return multipleSelect && countChoices && numCorrect;
+}
```
benchristel, line 60: It feels like we could better leverage the type system's ability to model the different possibilities here. Something like:

```ts
type PerseusRadioWidgetOptions = {
  responseFormat: SingleSelect | MultiSelect,
  // ...
}

type SingleSelect =  {type: "single-select"};

type MultiSelect = {
  type: "multi-select",
  /**
   * If present, the number of choices the learner must select for
   * their response to be accepted. Responses with a different number
   * of selected choices are marked invalid rather than incorrect. The
   * number of correct choices is displayed in the widget instructions.
   *
   * If null, the learner can submit a response with any number of 
   * selected options. Responses with the wrong number of options 
   * selected are marked incorrect.
   */
  numCorrect: number | null,
}
```

This would require a major version bump of the widget options format, so perhaps it's not something we want to take on right now. It might be worth migrating to this in the future.

No action needed.

packages/perseus/src/widgets/radio/__tests__/answerless-radio.test.tsx
benchristel: +1 for keeping things consistent for now. A temporarily unclear name is better than temporary inconsistency. Find-replace is easy, but gets harder when naming isn't consistent.

packages/perseus-core/src/data-schema.ts
handeyeco: Thanks for bringing this up: https://khanacademy.slack.com/archives/C49296Q7P/p1740496295273589

packages/perseus-editor/src/widgets/__tests__/radio-editor.test.tsx
```diff
@@ -176,4 +176,61 @@ describe("radio-editor", () => {
...
+         */
+        const choices = screen.getAllByTestId(
+            "choice-icon__library-choice-icon",
+        );
+
+        // switch an incorrect answer into a correct answer
+        await userEvent.click(choices[0]);
+
+        // now there should be 3 correct answers
+        expect(options?.numCorrect).toBe(3);
```
jeremywiebe, line 234: Stylistically, I find using an `onChange` `jest.fn()` for these types of things simpler to grok. With a var we set in the callback you have a to trace up through the test to understand that the `onChange` callback sets it. 

With an `onChange` mock, we could use:

```ts
expect(onChange).toHaveBeenCalledWith(expect.objectContaining({
    numCorrect: 3
});
```

In my mind that's less logic to support the test than having a `let options` we manage. 🤷‍♂️

handeyeco: Not a hill I'll die on, but I'm not totally sure I agree. I think...

``` TS
expect(options?.numCorrect).toBe(3);
```

...is much more legible than...

``` TS
expect(onChangeMock).toHaveBeenCalledWith(
    expect.objectContaining({
        numCorrect: 3,
    }),
);
```

I also think from a KISS perspective that a simple callback is easier to follow/debug than a Jest abstraction. But it's not a big deal in this particular case, so I made the change.

PR #2234, opened 2025-02-12
[LEMS-2563] Primary widget splitter

handeyeco:
## Summary:
Create a central place to pass in a PerseusItem and return a version with all the widget options split of answer data.

Mostly tests and some moving of logic. The main important thing is `splitPerseusItem`.

Issue: LEMS-2563

## Test plan:
This just adds the logic, still not using it. Just unit tests right now.


packages/perseus-core/src/utils/split-perseus-item.ts
```diff
@@ -0,0 +1,36 @@
...
+import {getPublicWidgetOptionsFunction} from "../widgets/core-widget-registry";
+
+import type {PerseusRenderer} from "../data-schema";
+
+export default function splitPerseusItem(
+    originalItem: PerseusRenderer,
+): PerseusRenderer {
+    const item = _.clone(originalItem);
+
+    if (!item.widgets || Object.keys(item.widgets).length === 0) {
```
handeyeco, line 12: I forgot to upgrade widgets which I probably need to do first?

benchristel: Is this early return really needed? I might prefer to do something like

```
const upgradedWidgets = getUpgradedWidgetOptions(item.widgets ?? {});
```

which I think will be equivalent.

[Returning early just because data is empty can lead to bugs](https://bensguide.substack.com/i/150330330/consequential).

packages/perseus-core/src/utils/split-perseus-item.ts
```diff
@@ -0,0 +1,39 @@
...
+        );
+
+        splitWidgets[id] = {
+            ...widget,
+            // Not all widgets have a splitting function
+            // so just pass options through for those that don't
+            // TODO(LEMS-2870): make publicWidgetOptionsFun generic
+            options: publicWidgetOptionsFun
+                ? publicWidgetOptionsFun(widget.options as any)
+                : widget.options,
```
benchristel, line 32: Maybe `getPublicWidgetOptionsFunction` should return the identity function if a widget doesn't have private options? Then we wouldn't need this logic and comment.

EDIT: oh! Looks like it's already doing that! This ternary can go away, I think.

packages/perseus-core/src/utils/split-perseus-item.ts
```diff
@@ -0,0 +1,39 @@
...
+        };
+    }
+
+    item.widgets = splitWidgets;
+
+    return item;
```
benchristel, line 38: If we did this:

```suggestion
    return {...item, widgets: splitWidgets};
```

we wouldn't need to `clone` above. It might also be clearer what parts of `item` are being changed.

packages/perseus-core/src/utils/split-perseus-item.ts
```diff
@@ -0,0 +1,39 @@
+import _ from "underscore";
+
+import {getPublicWidgetOptionsFunction} from "../widgets/core-widget-registry";
+import {getUpgradedWidgetOptions} from "../widgets/upgrade";
+
+import type {PerseusRenderer} from "../data-schema";
+
+export default function splitPerseusItem(
```
benchristel, line 8: Should this function be called `splitPerseusRenderer`? Or perhaps `getRendererPublicData` for consistency with the `get*PublicWidgetOptions` functions?

```suggestion
export default function getRendererPublicData(
```

benchristel (approved):
Looks good! I left a few non-blocking suggestions inline.

packages/perseus-core/src/utils/split-perseus-item.ts
handeyeco: I'm not sure I fully agree with you. One subtle difference:

``` TS
// my code
splitPerseusItem({}) // {}

// your code
splitPerseusItem({}) // { widgets: [] }
```

To me this feels wrong, but I guess so does continuing on.

Maybe the correct thing would be to throw and/or report the issue, but I think it's highly unlikely we hit this case anyway.

packages/perseus-core/src/utils/split-perseus-item.ts
handeyeco: I did `splitPerseusItem` to be consistent with `scorePerseusItem` and `validatePerseusItem`. If it's all the same to you, I think I'll keep it as-is. As an aside, I think both `Renderer` and `PublicData` are misnomers.

- `PerseusRenderer` isn't a renderer, it's an object that describes an article/exercise that's passed to a renderer. Most of Webapp calls it `item` or `itemData` which is a little better IMO but also a little more vague.
- Both the full data and `PublicOptions` are public data, they're just public at different times. I think `StrippedData` or `AnswerlessData` would be more appropriate, but they both have their own issues.

jeremywiebe: I think `item` and `itemData` are usually a slightly higher-level structure that includes hints and some other rarely-used data. We actually have a type in Perseus called [`PerseusItem`](https://github.com/Khan/perseus/blob/e63f83d0d89fd5b8e7aee3ab7248bcb19ec9be8a/packages/perseus-core/src/data-schema.ts#L202). 

So I agree `PerseusRenderer` is a misnomer... I wonder what would be a better name that isn't "item" (as that name is taken).

PR #2236, opened 2025-02-12
[LX] Add hairlines when Circle center has focus

nishasy:
## Summary:
Right now, all Interactive Graph movable points have hairlines when
they're being focused or drags. The exception being the center of a
Circle graph because it's not a MovablePoint.

Adding the Circle center hairlines here.

I also refactored the hairlines to be a separate component so they
can be reused in multiple places.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2310

## Test plan:
`yarn jest packages/perseus/src/widgets/interactive-graphs/graphs/circle.test.tsx`

Storybook
http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-circle

image


packages/perseus/src/widgets/interactive-graphs/graphs/components/hairlines.tsx
```diff
@@ -0,0 +1,42 @@
+import {color} from "@khanacademy/wonder-blocks-tokens";
```
SonicScrewdriver, line 1: LOVE that this is a separate file, this makes way more sense while also making this code far easier to locate.

SonicScrewdriver (approved):
This works great! Big improvement by moving out the hairlines too imo.

PR #2237, opened 2025-02-12
Remove unused functions from interactive-graph.tsx

benchristel:
Issue: none

## Test plan:

`yarn test`


benchristel:
Turns out these are actually used... I didn't read `getEquationString` carefully.

PR #2238, opened 2025-02-13
[unlimited-focus] [Bug Fix] Ensure that Unlimited Graphs move focus properly when deleting points out of order.

SonicScrewdriver:
## Summary:
This PR is part of the Interactive Graph Project.

While recreating LEMS-2551, it was discovered that focus was not being set for mouse-based interaction. This PR fixes both that issue, and the issue from LEMS-2551 where deleting the points out of order would eventually result in a failure to move the focus to the next point.

## Video example:
https://github.com/user-attachments/assets/acf582ff-e564-4950-a997-204c01b75e5a

Issue: LEMS-2551

## Test plan:
tests pass

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/point.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/point.tsx
@@ -52,7 +52,12 @@ function PointGraph(props: Props) {
         if (focusedIndex != null) {
             pointsRef.current[focusedIndex]?.focus();
         }
-    }, [props.graphState.focusedPointIndex, pointsRef]);
+    }, [
+        props.graphState.focusedPointIndex,
+        // Ensure that we re-focus if a point is deleted.
+        props.graphState.coords.length,
+        pointsRef,
+    ]);
 
     const statefulProps = {
         ...props,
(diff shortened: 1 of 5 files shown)


packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -722,19 +722,14 @@ function doRemovePoint(
         return state;
     }
 
-    let nextFocusedPointIndex: number | null;
-    if (state.interactionMode === "mouse") {
```
SonicScrewdriver, line 726: I confirmed with Design that we still want the focus to move to the next item for mouses as well.

packages/perseus/src/widgets/interactive-graphs/graphs/point.tsx
```diff
@@ -52,7 +52,12 @@ function PointGraph(props: Props) {
         if (focusedIndex != null) {
             pointsRef.current[focusedIndex]?.focus();
         }
-    }, [props.graphState.focusedPointIndex, pointsRef]);
+    }, [
+        props.graphState.focusedPointIndex,
+        // Ensure that we re-focus if a point is deleted.
+        props.graphState.coords.length,
```
SonicScrewdriver, line 58: While testing, this seemed to work well for the bug we were facing without causing any additional refreshes for other interactions. 

It seemed like a better solution than refactoring all of the Polygon and Point graph types/logic, but I'm open to discussion on it!

nishasy (approved):
Awesome!!! I love that you were able to find a simple solution.

PR #2242, opened 2025-02-13
Deprecate `metadata` in renderer, hint, and Group widget data schemas

benchristel:
We aren't using `metadata`. As near as I can tell, [it was added in 2014][1] so content
creators could add "tags" to content, but I don't know why it was added or if it was
ever used.

We are seeing Perseus JSON parser errors in webapp because `metadata` is null in some
exercises. This PR fixes the errors by changing the type of `metadata` to `any`, marking
it deprecated, and removing references to it.

[1]: https://phabricator.khanacademy.org/D13929

Issue: none

## Test plan:

- Install the build of Perseus from this PR
- Visit https://www.khanacademy.dev/internal-courses/test-everything/test-everything-1/te-group/a/group-article
- You should be able to view and edit the article without errors.

### DIFF
--- a/.changeset/tall-llamas-crash.md
+++ b/.changeset/tall-llamas-crash.md
@@ -1,7 +1,7 @@
 ---
 "@khanacademy/perseus": minor
-"@khanacademy/perseus-core": patch
-"@khanacademy/perseus-editor": patch
+"@khanacademy/perseus-core": minor
+"@khanacademy/perseus-editor": minor
 ---
 
-Delete 'metadata' from renderer, hint, and Group widget data schemas
+Deprecate the `metadata` field in renderer, hint, and Group widget data schemas.
--- a/.changeset/tall-llamas-crash.md
+++ b/.changeset/tall-llamas-crash.md
@@ -1,0 +1,7 @@
+---
+"@khanacademy/perseus": minor
+"@khanacademy/perseus-core": patch
+"@khanacademy/perseus-editor": patch
+---
+
+Delete 'metadata' from renderer, hint, and Group widget data schemas
--- a/packages/perseus-core/src/parse-perseus-json/perseus-parsers/hint.ts
+++ b/packages/perseus-core/src/parse-perseus-json/perseus-parsers/hint.ts
@@ -1,5 +1,5 @@
 import {
-    array,
+    any,
     boolean,
     object,
     optional,
@@ -17,6 +17,7 @@ export const parseHint: Parser<Hint> = object({
     replace: optional(boolean),
     content: string,
     widgets: defaulted(parseWidgetsMap, () => ({})),
-    metadata: optional(array(string)),
     images: parseImages,
+    // deprecated
+    metadata: any,
 });
--- a/packages/perseus-editor/src/widgets/group-editor.tsx
+++ b/packages/perseus-editor/src/widgets/group-editor.tsx
@@ -19,7 +19,6 @@ class GroupEditor extends React.Component<Props> {
         content: PropTypes.string,
         widgets: PropTypes.object,
         images: PropTypes.object,
-        metadata: PropTypes.any,
         apiOptions: ApiOptions.propTypes,
     };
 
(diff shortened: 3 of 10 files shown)


packages/perseus-editor/src/widgets/group-editor.tsx
```diff
@@ -50,19 +38,12 @@ class GroupEditor extends React.Component<Props> {
     };
 
     serialize: () => any = () => {
-        return _.extend({}, this.editor.current?.serialize(), {
-            metadata: this.props.metadata,
-        });
+        return _.extend({}, this.editor.current?.serialize());
```
handeyeco, line 41: Could this be simplified?

```suggestion
        return this.editor.current?.serialize();
```

packages/perseus/src/types.ts
```diff
@@ -200,6 +200,9 @@ export type APIOptions = Readonly<{
         keypadHeight?: number,
         focusedElement?: HTMLElement,
     ) => unknown;
+    /**
+     * @deprecated - metadata is no longer used by the Group widget
+     */
```
handeyeco, line 205: We're not able to just get rid of it?

packages/perseus-core/src/parse-perseus-json/perseus-parsers/hint.ts
```diff
@@ -17,6 +17,7 @@ export const parseHint: Parser<Hint> = object({
     replace: optional(boolean),
     content: string,
     widgets: defaulted(parseWidgetsMap, () => ({})),
-    metadata: optional(array(string)),
     images: parseImages,
+    // deprecated
+    metadata: any,
```
handeyeco, line 22: We can't just ignore it? It has to be parsed?

.changeset/tall-llamas-crash.md
```diff
@@ -0,0 +1,7 @@
+---
+"@khanacademy/perseus": minor
+"@khanacademy/perseus-core": patch
```
handeyeco, line 3: Isn't this a breaking change in `perseus-core`? If someone had defined a `PerseusRenderer` with `metadata` and upgraded `perseus-core` they would get a TS error and would be required to change their code?

handeyeco (approved):
Nothing blocking on my end. Looks like a very promising change.

packages/perseus-editor/src/widgets/group-editor.tsx
benchristel: Yeah, I was being cautious by not messing with it. Given that it's making a copy and I don't trust the editors not to mutate objects, I'd prefer:

```ts
return {...this.editor.current?.serialize()};
```

packages/perseus/src/types.ts
benchristel: This is probably way too paranoid, but I want to avoid the situation where

- we delete `metadata` from the schema and parser.
- 5 years from now, someone decides it would be cool if `PerseusRenderer` had a `metadata` field.
- they get errors in production because legacy content already has `metadata`, and it's shaped totally differently from whatever new metadata format they came up with.

I guess the parser regression tests should stop them before they get to step 3, though?

packages/perseus-core/src/parse-perseus-json/perseus-parsers/hint.ts
benchristel: This is probably way too paranoid, but I want to avoid the situation where

- we delete `metadata` from the schema and parser.
- 5 years from now, someone decides it would be cool if `PerseusRenderer` had a `metadata` field.
- they get errors in production because legacy content already has `metadata`, and it's shaped totally differently from whatever new metadata format they came up with.

I guess the parser regression tests should stop them before they get to step 3, though?

.changeset/tall-llamas-crash.md
benchristel: Yeah, I guess that's true!

benchristel:
I've confirmed that Group widgets can still be rendered and edited after this change.

.changeset/tall-llamas-crash.md
benchristel: I've deprecated `metadata` rather than removing it outright, and changed all of these version bumps to `minor`.

jeremywiebe:
Thanks. Glad we can start deprecating and cleaning up some of these really ancient fields that are no longer used.

PR #2243, opened 2025-02-14
[SR] Angle - Update strings to reflect changes to SRUX specs

nishasy:
## Summary:
The SRUX spec changed some vocabulary. Specifically, "terminal side" is now
"ending side" and "initial side" is now "starting side."

While here, I also refactored the code to match the other graph types'
aria labels getter functions, and I moved the tests over from the
mafs-graph test file to the angle test file.

Issue: none

## Test plan:
`yarn jest packages/perseus/src/widgets/interactive-graphs/graphs/angle.test.tsx`

Storybook
http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-angle


packages/perseus/src/strings.ts
```diff
@@ -226,18 +226,10 @@ export type PerseusStrings = {
...
-        point: number;
-        side: string;
-        x: string;
-        y: string;
-    }) => string;
-    srAngleVertexAtCoordinatesWithAngleMeasure: ({
+    srAngleStartingSide: ({x, y}: {x: string; y: string}) => string;
+    srAngleEndingSide: ({x, y}: {x: string; y: string}) => string;
+    srAngleVertex: ({x, y}: {x: string; y: string}) => string;
+    srAngleVertexWithAngleMeasure: ({
```
nishasy, line 232: I updated this to be multiple different strings for each point, so that passing in the `side` substring does cause errors for translations.

packages/perseus/src/widgets/interactive-graphs/graphs/angle.test.ts
nishasy, line 1: Not deleted. Just renamed to `angle.test.tsx` so that it can render React components.

packages/perseus/src/widgets/interactive-graphs/graphs/angle.test.tsx
nishasy, line 1: Tests moved over from https://github.com/Khan/perseus/pull/2224/commits/7e7f371953b337018ff184dcb9ab47fcbdd73547 and also updated to reflect new values.

SonicScrewdriver (approved):
We chatted about this elsewhere, and this looks great to me!

PR #2244, opened 2025-02-14
[Interactive Graph] Replace string-based function call with switch statement

nishasy:
## Summary:
There's a part in the interactive graph code where the equation string
getter function is called by a code that concatonates a string together
and uses that string as a function name to call.

For the sake of type safety and the wellbeing of our engineers, updating
this to a switch statement.

Issue: none

## Test plan:

Storybook


benchristel (approved):
Yay! LGTM.

PR #2246, opened 2025-02-14
Fix Perseus JSON parser issues identified via production logs

benchristel:
There are a few lingering issues with Perseus data that somehow didn't
show up in my content data dump. This PR updates the parser to handle them.

Issue: none

## Test plan:

- Deploy to webapp
- Search Sentry for "Perseus parser"
- There should be fewer errors after the deploy.


packages/perseus-core/src/data-schema.ts
```diff
@@ -406,8 +406,7 @@ export type PerseusImageBackground = {
     // The scale of the image
     // NOTE: perseus_data.go says this is required, but nullable, even though
     // it isn't necessary at all.
-    // Yikes, production data as this as both a number (1) and string ("1")
-    scale?: number | string;
+    scale?: number;
```
benchristel, line 409: We convert `scale` to a number in the parser now, so this type awkwardness can go away.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/perseus-answer-area.ts
benchristel, line 1: This code moved from `perseus-item.ts`, below.

packages/perseus-core/src/data-schema.ts
handeyeco: So is `data-schema` now the representation of the _parsed_ JSON? Because it was the representation of the _stored_ JSON, so the parser wouldn't affect that right?

packages/perseus-core/src/parse-perseus-json/perseus-parsers/numeric-input-widget.ts
```diff
@@ -40,7 +40,7 @@ export const parseNumericInputWidget: Parser<NumericInputWidget> = parseWidget(
                 // the data, simplify this.
                 value: optional(nullable(number)),
                 status: string,
-                answerForms: optional(array(parseMathFormat)),
+                answerForms: defaulted(array(parseMathFormat), () => undefined),
```
handeyeco, line 43: So is the different here that originally it was possible for the `answerForms` key to not exist but now it must exist but it's value is possibly `undefined`?

packages/perseus-core/src/parse-perseus-json/perseus-parsers/perseus-answer-area.ts
```diff
@@ -0,0 +1,37 @@
...
+
+// Some answerAreas have extra, bogus fields, like:
+//
+//   "answerArea": {
+//     "type": "multiple",
+//     "options": {},
+//     "version": null,
+//     "static": false,
+//     "graded": false,
+//     "alignment": "",
```
handeyeco, line 19: That's interesting, probably a bug with serialization. Some of our serialization is based on a blocklist of properties rather than an allowlist, which occasionally introduces some accidental properties.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/widgets-map.test.ts
```diff
@@ -37,9 +37,10 @@ describe("parseWidgetsMap", () => {
         );
     });
 
-    it("rejects a widget ID numbered 0", () => {
-        // Widget IDs with 0 currently cause a full-page crash when the
-        // exercise is rendered in webapp!
+    it("accepts a widget ID numbered 0", () => {
+        // Widget IDs with 0 cause a full-page crash when an exercise is
```
handeyeco, line 41: Do we need a ticket to address this?

packages/perseus-core/src/parse-perseus-json/perseus-parsers/widgets-map.ts
```diff
@@ -211,14 +211,18 @@ const parseDeprecatedWidget: Parser<DeprecatedStandinWidget> = parseWidget(
...
 );
 
-const parseStringToPositiveInt: Parser<number> = (rawValue, ctx) => {
-    if (typeof rawValue !== "string" || !/^[1-9][0-9]*$/.test(rawValue)) {
+const parseStringToNonNegativeInt: Parser<number> = (rawValue, ctx) => {
+    // The article renderer seems to allow the numeric part of a widget ID to
+    // be 0, at least for image widgets. However, if widget IDs in an exercise
+    // contain 0, the exercise renderer will blow up. We allow 0 here for
+    // compatibility with articles.
+    if (typeof rawValue !== "string" || !/^(0|[1-9][0-9]*)$/.test(rawValue)) {
```
handeyeco, line 219:
```suggestion
    if (typeof rawValue !== "string" || !/^(\d+)$/.test(rawValue)) {
```

I'm guessing we can't do this because we don't want `01`?

I'm a little worried about codifying the idea that a widget ID is `widget 1` whereas where we're trying to get to is widget ID is something like `/[a-z0-9 -]+/` (a uuid) - it just happens that they are made of a widget type and a number now.

I know `parseStringToNonNegativeInt` is more generic than just an ID, but I thought I'd bring it up since the comment brings up widget IDs.

packages/perseus-core/src/parse-perseus-json/regression-tests/item-data/radio-with-null-widgets.json
```diff
@@ -0,0 +1,2 @@
...
...nts out how “studies show that this focus on conformity can cause particular problems for certain students.” But the text doesn’t mention many people fighting to eliminate schools’ focus on conformity.\n","correct":false,"isNoneOfTheAbove":false,"widgets":null},{"content":"There have been many people fighting to impose new rules within the American public school system.","clue":"This statement isn’t likely true. The text doesn’t mention people fighting to impose new rules. \n","correct":false,"isNoneOfTheAbove":false,"widgets":null}],"noneOfTheAbove":false,"hasNoneOfTheAbove":false,"countChoices":false,"deselectEnabled":false,"randomize":false,"multipleSelect":false,"onePerLine":false,"displayCount":null}}},"replace":null,"metadata":null,"images":{}},"hints":[],"answerArea":{"type":"","static":false,"graded":false,"alignment":"","id":null,"key":null,"version":null,"zTable":false,"chi2Table":false,"tTable":false,"calculator":false,"periodicTable":false,"periodicTableWithKey":false,"financialCalculatorMonthlyPayment":false,"financialCalculatorTotalAmount":false,"financialCalculatorTimeToPayOff":false,"options":null},"_multi":null,"itemDataVersion":{"major":0,"minor":1},"answer":null}
```
handeyeco, line 2: Is there a reason this didn't get formatted? Feels like it would be nicer if it were across multiple lines.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/perseus-answer-area.test.ts
```diff
@@ -0,0 +1,53 @@
...
+        chi2Table: false,
+        financialCalculatorMonthlyPayment: false,
+        financialCalculatorTotalAmount: false,
+        financialCalculatorTimeToPayOff: false,
+        periodicTable: false,
+        periodicTableWithKey: false,
+        tTable: false,
+    };
+
+    it("fills in missing fields with `false`", () => {
```
handeyeco, line 21: No action needed, but noting that we don't test a mix between false/true values:

```
const result = parsePerseusAnswerArea({
  calculator: true,
  chi2Table: false,
}, ctx());
```

handeyeco (approved):

packages/perseus-core/src/data-schema.ts
benchristel: `data-schema` _should_ represent the _latest_ version of the stored JSON. The purpose of the parser is to migrate past versions of the JSON to match what's in data-schema. Really, data-schema never represented "the" stored JSON, because we have multiple versions of JSON stored.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/numeric-input-widget.ts
benchristel: The difference is that `null` was previously rejected as invalid, but now it's accepted and converted to `undefined`.

`optional(T)` allows either `T` or `undefined`, and rejects `null`.

`defaulted(T, () => U)` will accept `T`, `undefined`, or `null` as input, and convert `null` and `undefined` to `U`.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/widgets-map.test.ts
benchristel: Eh, probably not? We're not writing content with IDs of 0 into datastore anymore. And Jeremy has talked about wanting the IDs to be completely opaque strings, so addressing the issue with 0 separately seems like maybe a waste of effort.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/widgets-map.ts
benchristel: > I'm guessing we can't do this because we don't want 01?

Yeah, a leading `0` causes the number to be parsed as octal, so e.g. `011` will parse to 9.

> I'm a little worried about codifying the idea that a widget ID is widget 1 whereas where we're trying to get to is widget ID is something like /[a-z0-9 -]+/ (a uuid) - it just happens that they are made of a widget type and a number now.

It's easy to relax constraints, but harder to tighten them (since we'd have to make sure that our existing data adheres to the tighter constraint). I think it will be straightforward to allow more widget ID formats in the future.

packages/perseus-core/src/parse-perseus-json/regression-tests/item-data/radio-with-null-widgets.json
benchristel: Good call!

packages/perseus-core/src/parse-perseus-json/perseus-parsers/perseus-answer-area.ts
```diff
@@ -0,0 +1,37 @@
...
+//   }
+//
+// This function filters the fields of an answerArea object, keeping only the
+// known ones, and converts `undefined` and `null` values to `false`.
```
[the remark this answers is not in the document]
benchristel: `answerArea` is used to select which auxiliary features of the exercise UI to display. AFAICT code that uses it treats the values as booleans, so missing/undefined values are treated the same as `false`.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/perseus-answer-area.test.ts
benchristel: We're also not testing that each input field is mapped to the correct output field. As I've written the test, there's nothing stopping us from scrambling the fields when we assign them.

I'm inclined not to test for such things; tests provide value when they catch our mistakes, so a test for a mistake we're unlikely to ever make is unlikely to provide value.

PR #2249, opened 2025-02-18
[LEMS-2851] Answerless NumericInput

handeyeco:
## Summary:

**🚨🚨🚨 Will likely be closed in favor of: https://github.com/Khan/perseus/pull/2261**

---

NumericInput needed to be adjusted for answerless rendering because it was using `answers.answerforms` to generate a set of answer forms it could use when generating text for a tooltip that's displayed pre-scoring (see the ticket for more information).

This PR adds `PerseusNumericInputWidgetOptions.fullAnswerForms` (`PerseusNumericInputWidgetOptions.answerForms` was already taken although it's deprecated), a new NumericInput widget version, and upgrade functionality to derive `fullAnswerForms` from `answers.answerforms`. I am totally willing to rename `fullAnswerForms`, it's just a placeholder as I work. In case you're confused:

- `PerseusNumericInputWidgetOptions.answerForms`: deprecated but might still exist for old data (though I don't think it's used)
- `PerseusNumericInputAnswer.answerForms`: current way we derive `answerForms` data
- `PerseusNumericInputWidgetOptions.fullAnswerForms`: what I add in this PR so we can get `answerForms` without answers

If you're still confused, I'm happy to do a tour.

Issue: LEMS-2851

## Test plan:
- Go to a NumericInput widget that is known to show the tooltip that starts with "Your answer should be..."
- That tooltip should continue to be visible
- NumericInput should be renderable, interactive, and answerable.


packages/perseus-core/src/data-schema.ts
```diff
@@ -1205,6 +1205,7 @@ export type PerseusNumericInputWidgetOptions = {
     // Used by examples, maybe not used and should be removed in the future
     // see TODO in numeric-input
     answerForms?: ReadonlyArray<PerseusNumericInputAnswerForm>;
+    fullAnswerForms?: ReadonlyArray<PerseusNumericInputAnswerForm>;
```
handeyeco, line 1208: I'm expecting that this will change (or something will need to change). Here's what I think happened:

1. Originally NumericInput used `PerseusNumericInputWidgetOptions.answerForms`
2. Then NumericInput was updated to use `PerseusNumericInputAnswer.answerForms` and `PerseusNumericInputWidgetOptions.answerForms` was deprecated (see: "maybe not used and should be removed in the future")

Now `PerseusNumericInputWidgetOptions.answers` is being removed and I think we need _something like_ `PerseusNumericInputWidgetOptions.answerForms` again **but** `PerseusNumericInputWidgetOptions.answerForms` _might exist on old data_ so I called the new thing `fullAnswerForms` pending more research.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/numeric-input-widget.ts
```diff
@@ -36,60 +38,173 @@ const parseSimplify = enumeration(
...
+        ...widget,
+        version: {major: 1, minor: 0},
+        options: {
+            ...options,
+            fullAnswerForms: deriveAnswerForms(options),
+        },
+    };
+}
+
+export const parseNumericInputWidget: Parser<NumericInputWidget> =
```
handeyeco, line 198: Upgrader in Ben's parser

packages/perseus-core/src/widgets/numeric-input/numeric-input-upgrade.ts
```diff
@@ -0,0 +1,37 @@
+import {deriveAnswerForms} from "./util";
+
+import type {PerseusNumericInputWidgetOptions} from "../../data-schema";
+
+export const currentVersion = {major: 1, minor: 0};
+
+export const widgetOptionsUpgrades = {
```
handeyeco, line 7: Upgrader using the `widgetOptionsUpgrades` API (which will likely be replaced by Ben's parser)

packages/perseus/src/widgets/numeric-input/numeric-input.class.tsx
```diff
@@ -171,25 +170,12 @@ const propsTransform = function (
...
-                    });
-                }),
-        ),
-    };
+    const {answers: _, fullAnswerForms = [], ...rendererProps} = widgetOptions;
 
-    return rendererProps;
+    return {
+        ...rendererProps,
+        answerForms: fullAnswerForms,
```
handeyeco, line 177: Like I said, we might not keep it named `fullAnswerForms` so for now I'm just renaming it to `answerForms` so I don't have to update the widget code which I'd be fine doing once we have an established name.

packages/perseus/src/widgets/numeric-input/numeric-input.test.ts
```diff
@@ -453,47 +453,12 @@ describe("transform", () => {
...
+            fullAnswerForms: [
                 {
                     simplify: "required",
                     name: "proper",
                 },
             ],
-        });
-    });
-
-    it("only uses answer forms from correct answers", () => {
```
handeyeco, line 471: This got moved to tests for `deriveAnswerForms`

packages/perseus/src/widgets/numeric-input/utils.ts
```diff
@@ -42,8 +46,16 @@ export const generateExamples = (
     // Generate a list of the unique answer forms.
     const uniqueForms = getUniqueAnswerForms(answerForms);
 
+    // Sort them by the order they appear in the `formExamples` list.
+    const formExampleKeys = Object.keys(NumericExampleStrings);
+    const sortedForms = uniqueForms.sort((a, b) => {
```
handeyeco, line 51: So originally we were sorting `answerForms` when deriving the data, but:

1. I think that's a UI concern, not a data concern
2. I didn't want to have to move `NumericExampleStrings` into `perseus-core`

So I moved the sorting logic and added a test.

packages/perseus/src/widgets/numeric-input/utils.ts
```diff
@@ -75,47 +87,3 @@ export const shouldShowExamples = (
 
     return !allFormsAccepted;
 };
-
-/**
- * uniqueAnswerForms takes a list of answer forms and returns a list of unique
- * answer forms. This is useful for ensuring that we don't show duplicate examples
- * to the user.
- */
-const getUniqueAnswerForms = function (
```
handeyeco, line 84: These were moved to `perseus-core` because I needed them for deriving answer forms.

packages/perseus/src/widgets/numeric-input/numeric-input.test.ts
```diff
@@ -467,45 +478,162 @@ describe("transform", () => {
...
+        await userEvent.type(screen.getByRole("textbox", {hidden: true}), "42");
+
+        const userInput = renderer.getUserInputMap();
+        const score = scorePerseusItem(fullItem, userInput, "en");
+
+        // Assert
+        expect(score).toHaveBeenAnsweredCorrectly();
+    });
+
+    it("is interactive with answerless widget options", async () => {
```
handeyeco, line 554: Here's the important test. We:

- Render the widget with the stripped data
- Assert that UI that previously required `answers` still renders
- Answer and score the widget

packages/perseus-core/src/widgets/numeric-input/util.ts
```diff
@@ -0,0 +1,64 @@
...
+    // We use a Set to keep track of the forms we've already seen.
+    const foundForms = new Set<string>();
+    return list.filter((form) => {
+        // If we've already seen this form, skip it.
+        if (foundForms.has(form.name)) {
+            return false;
+        }
+        // Otherwise, add it to the set and return true.
+        foundForms.add(form.name);
+        return true;
```
benchristel, line 23: Is it correct to not take `simplify` into account when de-duplicating answer forms? E.g. if we have

```ts
const answerForms = [
  {name: "proper", simplify: "required"},
  {name: "proper", simplify: "optional"},
];
```

Only `{name: "proper", simplify: "required"}` will be present in the output. Whether `simplify` is `"required"` or `"optional"` in the output depends on the order in which the answers were originally listed, which is awkward, since I don't think the order is otherwise important.

benchristel: I noticed there's a very similar function, `unionAnswerForms`, in `numeric-input.tsx`. I think yours is better (better-typed, easier to read, possibly faster) but the two functions should probably be combined if they represent the same concept.

packages/perseus-core/src/widgets/numeric-input/util.ts
```diff
@@ -0,0 +1,64 @@
...
+    if (options.fullAnswerForms) {
+        return options.fullAnswerForms;
+    }
+
+    return unionAnswerForms(
+        // Filter out the correct answers and map them to the answer forms
+        // so that we can generate the examples for the widget.
+        (options.answers || [])
+            .filter((answer) => answer.status === "correct")
+            .map((answer) => {
```
benchristel, line 55: If we used `flatMap` here I don't think we'd need the call to `flat()` in `unionAnswerForms`. `unionAnswerForms` would then be a trivial wrapper around `getUniqueAnswerForms`, so it could be removed.

```suggestion
            .flatMap((answer) => {
```

packages/perseus-core/src/parse-perseus-json/perseus-parsers/numeric-input-widget.test.ts
```diff
@@ -0,0 +1,60 @@
+import {parse} from "../parse";
+import {success} from "../result";
+
+import {parseNumericInputWidget} from "./numeric-input-widget";
+
+describe("parseExpressionWidget", () => {
```
handeyeco, line 6: Copy pasta mistake

PR #2251, opened 2025-02-19
[SR] Update Linear strings

nishasy:
## Summary:
We're doing a final pass on strings now at the end of the phase 1
interactive graph screen reader work.

- Update the linear graph's grab handle to say it's going through
  the points, rather than it being from one point to the other
  (as that would make it a line segment instead).

Issue: https://khanacademy.atlassian.net/browse/LEMS-2782

## Test plan:
`yarn jest packages/perseus/src/widgets/interactive-graphs/graphs/linear.test.tsx`

Storybook
- http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-linear

### DIFF
--- a/.changeset/fair-humans-tie.md
+++ b/.changeset/fair-humans-tie.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+[SR] Update Linear strings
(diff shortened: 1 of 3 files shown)


catandthemachines (approved):
LGTM!

PR #2254, opened 2025-02-19
Enable interactive graphs to take answerless props

benchristel:
To prevent cheating, we want to avoid sending exercise answers to the frontend
before the learner submits their response. This PR ensures that interactive
graphs can render without answers.

The types prove that `props.correct` is not used unless the `static` prop is
true. Static graphs are not answerable or interactable, and always display the
data in the `correct` prop.

Issue: LEMS-2853

## Test plan:

`yarn test`


packages/perseus/src/widgets/interactive-graph.tsx
```diff
@@ -193,6 +194,11 @@ type DefaultProps = {
     PerseusInteractiveGraphRubric
 > satisfies PropsFor<typeof InteractiveGraph>;
 
+0 as any as WidgetProps<
+    InteractiveGraphPublicWidgetOptions,
+    PerseusInteractiveGraphRubric
+> satisfies PropsFor<typeof InteractiveGraph>;
```
benchristel, line 200: This is probably the most important line of code in this PR. It ensures that answerless data can be assigned to the widget props.

packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.tsx
```diff
@@ -127,7 +127,7 @@ export const StatefulMafsGraph = React.forwardRef<
 
     // If the graph is static, it always displays the correct answer. This is
     // standard behavior for Perseus widgets (e.g. compare the Radio widget).
-    if (props.static) {
+    if (props.static && props.correct) {
```
benchristel, line 130: This is the only place we use `correct` inside the interactive graph components.

packages/perseus/src/widgets/interactive-graphs/interactive-graph.test.tsx
benchristel, line 1: This file has a lot of indentation changes because I removed a rendundant `describe` call. I recommend using the "hide whitespace" feature to review it.

packages/perseus-core/src/index.ts
```diff
@@ -130,7 +130,10 @@ export {default as getCategorizerPublicWidgetOptions} from "./widgets/categorize
 export {default as getCSProgramPublicWidgetOptions} from "./widgets/cs-program/cs-program-util";
 export {default as getExpressionPublicWidgetOptions} from "./widgets/expression/expression-util";
 export {default as getGrapherPublicWidgetOptions} from "./widgets/grapher/grapher-util";
-export {default as getInteractiveGraphPublicWidgetOptions} from "./widgets/interactive-graph/interactive-graph-util";
+export {
+    default as getInteractiveGraphPublicWidgetOptions,
+    InteractiveGraphPublicWidgetOptions,
```
handeyeco, line 135: I could be wrong, but for some reason I was thinking we needed `export type` for types specifically in `index.ts` files because of some weirdness with how things get exported from packages.

packages/perseus/src/widgets/interactive-graphs/interactive-graph.test.tsx
```diff
@@ -102,310 +102,318 @@ describe("Interactive Graph", function () {
...
-                // TODO(jeremy): This test is disabled because it fails
-                // sporadically (especially on slower/lower-end computers, like
-                // CI). Will work on a fix after the React 18 release.
-                it.skip("rejects incorrect answer", async () => {
-                    // Arrange
-                    const {renderer} = renderQuestion(question, blankOptions);
+                const answerlessQuestion = splitPerseusItem(question);
 
-                    await userEvent.tab();
+                renderQuestion(answerlessQuestion, blankOptions);
```
handeyeco, line 275: We don't need an assertion? Just rendering?

handeyeco (approved):
Can't believe it was so straight-forward!

In the Dropdown PR, even though it didn't use answer data [I added RTL tests](https://github.com/Khan/perseus/pull/2231/files#diff-cfc86712f3e6850ec71a02ff417c86807f4b1ce801aaea8641a38f958ce4ac65) to make sure that it...

- Rendered as expected
- Was interactive
- Was answerable
- Was scorable

...for both answerful/answerless data. I think it could be beneficial to have something like that for IG to make sure we don't regress as we work.

packages/perseus-core/src/index.ts
benchristel: Good call! This export broke Storybook, and changing it to a type export fixed it.

packages/perseus/src/widgets/interactive-graphs/interactive-graph.test.tsx
benchristel: None of the other interactive graph tests check for specific elements being rendered. They're just testing that rendering doesn't throw an error. I didn't want to shave that yak in this PR, so what I decided to do instead is add a visual regression test in Storybook to make sure the elements render. We have several such tests for Interactive Graph already.

PR #2257, opened 2025-02-20
[SR] Allow using control + shift + arrow keys for NVDA

nishasy:
## Summary:
NVDA takes control of inputs in its default mode, so users can't use just the arrow keys to move the interactive elements on the graph. We want to allow using modifier key + control + shift + arrow keys to move interactive elements on the graph.

Note: This already works with the modifier key + shift + arrow keys (no control), but holding down shift makes it so that the amount the element moves is 4x more than without shift.

Note 2: When using NVDA, you can get the same keyboard experience as not using a screen reader (i.e. move the points with arrow keys only) if you go into focus mode (NVDA key + space bar).

Implemented:
- Added support for the control key press
  - stops shift from increasing the move difference if control is held down
- Added "modifier key + control + shift +" to the instruction strings

Issue: https://khanacademy.atlassian.net/browse/LEMS-2874

## Test plan:

Storybook
http://localhost:6006/?path=/docs/perseuseditor-widgets-interactive-graph--docs


catandthemachines:
Question, does this solution still work for screen reader users who have different modifier keys set or Voiceover? 🤔

PR #2259, opened 2025-02-20
Move version-embed fn and embed in each package

jeremywiebe:
## Summary:

As we are working on Server-Side Scoring, the fact that the `addLibraryVersionToPerseusDebug` function lived in `@khanacdemy/perseus-core` was becoming problematic. All packages need to depend on `perseus-core` because they all need to use this function.

This PR moves the function out to a top-level directory and then each package simply imports it using a relative-path import. This violates our lint rules so we suppress it and because we do this, each package will be a few hundred bytes larger. But it completely eliminates the need for cross-package dependencies just for this one function. 

Issue: LEMS-2881

## Test plan:

I ran `yarn build` and inspected the files in each package's `dist/` folder. They had the built version number properly embedded in their output. They also had the `addLibraryVersionToPerseusDebug` function in them (the fn was copied into the bundle successfully).


packages/kas/src/shared-utils
```diff
@@ -0,0 +1 @@
+../../../utils/shared
```
Myranae, line 1: What does this file and line do?

packages/kas/src/version.ts
```diff
@@ -2,7 +2,11 @@
 // version number during the release build.
 // In dev, you'll never see the version number.
 
-import {addLibraryVersionToPerseusDebug} from "@khanacademy/perseus-core";
+// ESLint's import/no-relative-packages doesn't understand symlinks! If this
+// lint suppression is removed, it autofixes the import path to an invalid
+// package+path entry.
+// eslint-disable-next-line import/no-relative-packages
```
Myranae, line 8: I was wondering if maybe these comments should explain why this is being done since I can imagine something maybe trying to fix this down the road and moving it back to perseus-core. Maybe this PR is enough to prevent that though  XD

packages/kas/src/shared-utils
jeremywiebe: Exactly that. That's how git represents symlinks.  By using a directory that's symlinked in... we can share the code, but it's just copied into each bundle instead of being referenced as a package dependency.

packages/kas/src/version.ts
jeremywiebe: I edited these comments to hopefully be clearer. :)

Myranae: Thank you! Looks great :D

PR #2261, opened 2025-02-20
[LEMS-2851] Answerless NumericInput (Take 2)

handeyeco:
## Summary:
Alternative to https://github.com/Khan/perseus/pull/2249

Instead of doing a widget upgrade (the old PR), just split `answers` more selectively so that the data we need for rendering pre-scoring is still available.

Issue: LEMS-2851

## Test plan:

- Go to a NumericInput widget that is known to show the tooltip that starts with "Your answer should be..."
- That tooltip should continue to be visible
- NumericInput should be renderable, interactive, and answerable.


benchristel (approved):
LGTM!

packages/perseus-core/src/widgets/numeric-input/numeric-input-util.ts
```diff
@@ -10,18 +18,36 @@ type NumericInputPublicWidgetOptions = {
...
+    answer: PerseusNumericInputAnswer,
+): NumericInputAnswerPublicData {
+    const {answerForms, simplify, status} = answer;
+    return {
+        answerForms,
+        simplify,
+        status,
+    };
+}
+
```
benchristel, line 38: I like this approach! Much simpler, and this way we don't have to denormalize the data at rest.

packages/perseus-core/src/data-schema.ts
```diff
@@ -1202,9 +1202,6 @@ export type PerseusNumericInputWidgetOptions = {
     rightAlign?: boolean;
     // Always false.  Not used for this widget
     static: boolean;
-    // Used by examples, maybe not used and should be removed in the future
-    // see TODO in numeric-input
-    answerForms?: ReadonlyArray<PerseusNumericInputAnswerForm>;
```
Myranae, line 1207: Since this is our contract file, are there folks we need to inform about changing it? Or this isn't actually used anywhere anymore so it shouldn't really cause an impact?

packages/perseus-core/src/widgets/numeric-input/numeric-input-util.ts
```diff
@@ -10,18 +18,36 @@ type NumericInputPublicWidgetOptions = {
...
+ * to the learner regarding the format of accepted answers
+ */
+function getNumericInputAnswerPublicData(
+    answer: PerseusNumericInputAnswer,
+): NumericInputAnswerPublicData {
+    const {answerForms, simplify, status} = answer;
+    return {
+        answerForms,
+        simplify,
+        status,
```
Myranae, line 35: So although `status` says whether this specific answer is correct, wrong, or ungraded, it doesn't provide enough info to cheat since we don't include the `value` field` from `answer`. Is that correct?

packages/perseus/src/widgets/numeric-input/numeric-input.class.tsx
```diff
@@ -35,7 +35,7 @@ export type NumericInputProps = ExternalProps & {
     rightAlign: NonNullable<ExternalProps["rightAlign"]>;
     apiOptions: NonNullable<ExternalProps["apiOptions"]>;
     coefficient: NonNullable<ExternalProps["coefficient"]>;
-    answerForms: NonNullable<ExternalProps["answerForms"]>;
+    answerForms: ReadonlyArray<PerseusNumericInputAnswerForm>;
```
Myranae, line 38: I was wondering, what does removing `answerForms` from the widget options but keeping it here accomplish?

packages/perseus-core/src/data-schema.ts
handeyeco: In this case, it shouldn't have an impact. The data still can have `answerForms`, we just won't be using it. During a backfill, we'll want to remove it from the data.

packages/perseus/src/widgets/numeric-input/numeric-input.class.tsx
handeyeco: I didn't add either; `answerForms` on widget options was deprecated and unused, `answerForms` on answers existed already and is used.

PR #2262, opened 2025-02-20
Create new Perseus linter rule to catch markdown images added to content without a URL

Myranae:
## Summary:
A bug was found where an infinite spinner occurs when a Markdown image is added to content without an URL. The page couldn't load the image, so the loading spinner stayed present and prevented interactions with the page. This adds a linter rule to catch when an image is added without a URL to prevent this bug from happening in the future.

Before:
image

After:
image

Issue: LEMS-2727

## Test plan:
- Confirm tests pass
- Confirm the expected warning appears when an image without a URL is added to content via the content editor


packages/perseus-linter/src/__tests__/rules.test.ts
```diff
@@ -496,6 +497,25 @@ describe("Individual lint rules tests", () => {
...
     });
 
+    expectWarning(
+        imageUrlEmptyRule,
+        [
+            "![alt-text]()", // empty URL
+            "![alt-text](  )", // empty URL with spaces
+            "![alt-text](\n)", // empty URL with newline
+        ],
+        null,
```
Myranae, line 507: Other similar tests add a ts-expect-error (`// @ts-expect-error - TS2554 - Expected 3 arguments, but got 2.`), but I know those are not ideal, so just added `null` instead. There is no context needed as far as I can tell.

jeremywiebe: That's fine. I think the type for those `expectPass()` and `expectFail()` functions should make `context` be optional.  You could try adding a `?` at the end of `context` and seeing if that works.

packages/perseus-linter/src/rules/image-url-empty.ts
```diff
@@ -0,0 +1,23 @@
...
+        const url = image.target;
+
+        // If no URL is provided, an infinite spinner will be shown in articles
+        // overlaying the page where the image should be. This prevents the page
+        // from fully loading. As a result, we check for URLS with all images.
+        if (!url || !url.trim()) {
+            return "Images should have a URL";
+        }
+
+        // NOTE(TB): Ideally there would be a check to confirm the URL works
```
jeremywiebe, line 18: Nice, thanks for leaving this here!

jeremywiebe (approved):

packages/perseus-linter/src/__tests__/rules.test.ts
Myranae: Wow, that's a beautiful fix! I need to change my mindset from, "Don't change the existing code unrelated to your change" to "Fix the things!" :D Thanks!

PR #2263, opened 2025-02-21
[Interactive Graph] Update protractor to new SVG

nishasy:
## Summary:
The old protractor looked awful. We got a request to use a protractor
with two rows of numbers going in opposite directions (previously
there was only one direction).

- Update the protractor to a new SVG with numbers going both ways

Note: The numbers are kind of hard to read over the axes. I tried to add
white backgrounds to the numbers, but it's unfortunately not as easy as
it looks. Since this is not a regression, I left it as is.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2863

## Test plan:
Storybook
- https://khan.github.io/perseus/?path=/story/perseus-widgets-interactive-graph-visual-regression-tests--mafs-with-protractor

| Before | After |
| --- | --- |
| image | image |


jeremywiebe (approved):
Nice!  I had two thoughts (neither of which are blocking):

1. This increases the bundle size by about 45kb. Previously we loaded the protractor image remotely which meant that the image size didn't increase the main Perseus bundle size. Perhaps you could touch base with FEI to see their thoughts on something like this? I think it makes sense to embed the SVG ... and perhaps the gzip compression we apply when serving the JS bundles will negate the increase. 
2. Should we change the mouse cursor to the move handle instead the default arrow? I notice for other movables in the graph we do that, but not when moving the protractor. This seems like a new change, but it felt "off" that the mouse cursor didn't change to show I could drag the protractor (but it did for the double-ended arrow that allows me to rotate the protractor)

nishasy:
[Latest storybook publish](https://650db21c3f5d1b2f13c02952-menogthlrg.chromatic.com/?path=/story/perseus-widgets-interactive-graph-visual-regression-tests--mafs-with-protractor)

jeremywiebe (approved):
Review Round II. Still looks good. 😁

PR #2265, opened 2025-02-24
Export the ContentPreview component from perseus-editor

Myranae:
## Summary:
The ContentPreview component was created previously during our work to try to remove the iframe-based preview system. That work did not roll out, but access has been requested to this component for other work. As such, we are now exporting the component from the perseus-editor package.

Issue: LEMS-2886

## Test plan:
- Confirm all tests and checks pass

### DIFF
--- a/packages/perseus-editor/src/index.ts
+++ b/packages/perseus-editor/src/index.ts
@@ -8,13 +8,13 @@ export {default as ItemDiff} from "./diffs/item-diff";
 export {default as EditorPage} from "./editor-page";
 export {default as Editor} from "./editor";
 export {default as IframeContentRenderer} from "./iframe-content-renderer";
+export {default as ContentPreview} from "./content-preview";
 
 import "./styles/perseus-editor.less";
 
 // eslint-disable-next-line import/order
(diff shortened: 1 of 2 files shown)


packages/perseus-editor/src/index.ts
```diff
@@ -14,11 +14,12 @@ import "./styles/perseus-editor.less";
 // eslint-disable-next-line import/order
 import {Widgets, widgets} from "@khanacademy/perseus";
 import AllEditors from "./all-editors";
+import ContentPreview from "./content-preview";
```
jeremywiebe, line 17: Not a required change, but we can export this component in a single line:

```ts
export {default as ContentPreview} from "./content-preview";
```

jeremywiebe (approved):
Thanks for taking care of this Tamara!

packages/perseus-editor/src/index.ts
Myranae: Whoops! Used the wrong export section. I can see now that section below is there because some work is done to `AllEditors` and `widgets` before exporting. Moved the ContentPreview export to the top with the other exports. Thanks!

PR #2266, opened 2025-02-24
[LEMS-2777] Bugfix: KAS handling functions around plain numbers

handeyeco:
## Summary:
See the ticket for the full break-down of the issue.

What I think was happening:
- KAS saw `f(2)`
- KAS looked for a variable, which it doesn't have (`f` is a function)
- KAS said "cool, I'll take a shortcut and just treat this as a number then"
- It tried to do math on `f(2)` like it was a Rational

I just removed the part of the code that was trying to take a shortcut.

Issue: LEMS-2777

## Test plan:
- Added unit tests
- Repro steps for the UI are in the ticket

### DIFF
--- a/packages/kas/src/nodes.ts
+++ b/packages/kas/src/nodes.ts
@@ -425,11 +425,6 @@ abstract class Expr {
             );
         };
 
-        // if no variables, only need to evaluate once
-        if (!varList.length && !this.has(Unit) && !other.has(Unit)) {
-            return equalNumbers(this.eval(), other.eval());
-        }
-
         // collect here to avoid sometimes dividing by zero, and sometimes not
         // it is better to be deterministic, e.g. x/x -> 1
         // TODO(alex): may want to keep track of assumptions as they're made
--- a/packages/kas/src/nodes.ts
+++ b/packages/kas/src/nodes.ts
@@ -425,6 +425,19 @@ abstract class Expr {
             );
         };
 
+        // If no variables, only need to evaluate once.
+        // note(matthew) Seems to be an optimization for simple cases like `2+2=4`
+        // where there are no variables / functions.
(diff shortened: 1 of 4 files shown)


packages/kas/src/nodes.ts
```diff
@@ -425,11 +425,6 @@ abstract class Expr {
             );
         };
 
-        // if no variables, only need to evaluate once
-        if (!varList.length && !this.has(Unit) && !other.has(Unit)) {
-            return equalNumbers(this.eval(), other.eval());
-        }
```
handeyeco, line 431: I have very low confidence that this change won't introduce a bug. KAS has always been very tricky to me. However all new and old tests are passing, I'm going to encourage a thorough QE run, and I would love if I could get a thoughts/feels from other devs.

handeyeco: As an alternative, this also passes tests:

``` TS
        const varAndFuncList = _.union(
            this.getVars(/* excludeFunc */ false),
            other.getVars(/* excludeFunc */ false),
        );
        // if no variables, only need to evaluate once
        if (!varAndFuncList.length && !this.has(Unit) && !other.has(Unit)) {
            return equalNumbers(this.eval(), other.eval());
        }
```

kevinb-khan: From the comment it makes me think that this was an optimization to avoid unnecessary computation.  Out of curiosity, did this `return` statement ever get hit when running the tests?

kevinb-khan: I think the alternative you've proposed makes sense.

kevinb-khan (approved):
LGTM.  I think you should go with the alternative you suggested, but both approaches work.

PR #2268, opened 2025-02-25
[numeric-refactor-feedback-cleanup] Numeric Input Refector Feedback

SonicScrewdriver:
## Summary:
This is a little PR that addresses some additional feedback points from the Numeric Input Refactor. 

This includes:
- Using useId() for generating ids in InputWithExamples 
- Removing underscores from InputWithExamples after it was converted to a functional component as part of the Numeric Input Refactor work. 
- Removing unnecessary constructors 
- Rename of `getTooltipContent` to `renderTooltipContent`
- Moving RenderProps to top of file for numeric-input.class.tsx 
- Removing unnecessary data from test data 
- Updating test name in utils.test.ts 
- Simplifying types from `NumericExampleStrings` after they were moved into the util file. 
- Commenting the reason for the `ariaLabel` setting to `undefined`.
- New tests in answer-types

## Test plan:
- Tests Pass
- Manual testing in Storybook (Link to ZND here)

### DIFF
--- a/packages/perseus/src/components/text-input.tsx
+++ b/packages/perseus/src/components/text-input.tsx
@@ -105,6 +105,10 @@ class TextInput extends React.Component<Props> {
 
         const formattedValue = value === null ? "" : value.toString();
 
+        // Some of our content was saved with empty strings as the label text,
+        // and we don't want to render an empty aria-label attribute.
+        const ariaLabel = labelText || undefined;
+
         return (
             <TextField
                 ref={this.inputRef}
@@ -113,7 +117,7 @@ class TextInput extends React.Component<Props> {
                 id={this.id}
                 value={formattedValue}
                 type="text"
-                aria-label={labelText}
+                aria-label={ariaLabel}
                 aria-describedby={this.props["aria-describedby"]}
                 onChange={(value) => this.props.onChange(value)}
                 placeholder={placeholder}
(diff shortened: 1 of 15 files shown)


packages/perseus/src/widgets/numeric-input/input-with-examples.tsx
```diff
@@ -90,9 +87,22 @@ const InputWithExamples = forwardRef<
...
+    };
+
+    const handleBlur = () => {
+        onBlur();
+        setInputFocused(false);
+    };
+
+    const getUniqueId = () => {
+        return `input-with-examples-${id}`;
+    };
```
mark-fitzgerald, line 102: Does this need to be a function, now that it's simpler? Could we just have the following up around line 63:
```
const id = useId();
```

packages/perseus/src/components/text-input.tsx
```diff
@@ -113,7 +113,7 @@ class TextInput extends React.Component<Props> {
                 id={this.id}
                 value={formattedValue}
                 type="text"
-                aria-label={labelText || undefined}
+                aria-label={labelText}
```
mark-fitzgerald, line 116: I think the `undefined` was there so that there is no `aria-label` attribute when `labelText` is blank. Otherwise, we end up with `aria-label=""`.

mark-fitzgerald (approved):

PR #2269, opened 2025-02-25
Default nulls to +/-Infinity when parsing the domain of a locked function

benchristel:
This fixes a Perseus JSON parser error observed in production. A null value for
in the JSON the upper or lower bound means infinity/unbounded, so we now convert nulls
to infinities during parsing. Note that the infinite values are converted back to
`null` when the JSON is serialized, so this preserves behavior.

Issue: none

## Test plan:

`pnpm test`

Verify that locked functions with a domain of `[null, null]` display correctly
by running `pnpm storybook` and visiting the [locked function docs page][1].

Test that when you edit the domain of a locked function the correct JSON is
generated, and the graph displays correctly.

[1]: http://localhost:6006/?path=/docs/perseus-widgets-interactive-graph-locked-functions--docs

### DIFF
--- a/packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
+++ b/packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
@@ -253,16 +253,22 @@ const parseLockedPolygonType: Parser<LockedPolygonType> = object({
     ariaLabel: optional(string),
 });
 
+// Exported for testing.
+export const parseLockedFunctionDomain = defaulted(
+    pair(
+        defaulted(number, () => -Infinity),
+        defaulted(number, () => Infinity),
+    ),
+    (): [number, number] => [-Infinity, Infinity],
+);
+
 const parseLockedFunctionType: Parser<LockedFunctionType> = object({
     type: constant("function"),
     color: parseLockedFigureColor,
     strokeStyle: parseLockedLineStyle,
     equation: string,
     directionalAxis: enumeration("x", "y"),
-    domain: defaulted(
-        pair(nullable(number), nullable(number)),
-        (): [null, null] => [null, null],
-    ),
+    domain: parseLockedFunctionDomain,
     // TODO(benchristel): default labels to empty array?
     labels: optional(array(parseLockedLabelType)),
     ariaLabel: optional(string),
--- a/packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
+++ b/packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
@@ -259,7 +259,8 @@ const parseLockedFunctionType: Parser<LockedFunctionType> = object({
     strokeStyle: parseLockedLineStyle,
     equation: string,
     directionalAxis: enumeration("x", "y"),
-    domain: optional(pairOfNumbers),
+    // TODO(benchristel): default domain to [null, null]?
+    domain: optional(pair(nullable(number), nullable(number))),
     // TODO(benchristel): default labels to empty array?
     labels: optional(array(parseLockedLabelType)),
     ariaLabel: optional(string),
--- a/packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
+++ b/packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
@@ -259,8 +259,10 @@ const parseLockedFunctionType: Parser<LockedFunctionType> = object({
     strokeStyle: parseLockedLineStyle,
     equation: string,
     directionalAxis: enumeration("x", "y"),
-    // TODO(benchristel): default domain to [null, null]?
-    domain: optional(pair(nullable(number), nullable(number))),
+    domain: defaulted(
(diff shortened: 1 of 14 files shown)


packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
```diff
@@ -259,7 +259,8 @@ const parseLockedFunctionType: Parser<LockedFunctionType> = object({
     strokeStyle: parseLockedLineStyle,
     equation: string,
     directionalAxis: enumeration("x", "y"),
-    domain: optional(pairOfNumbers),
+    // TODO(benchristel): default domain to [null, null]?
+    domain: optional(pair(nullable(number), nullable(number))),
```
benchristel, line 263: @nishasy would it be useful to default the domain to `[null, null]` if it's not present? Then the React components wouldn't need to handle the case where `domain` is `undefined`.

nishasy: I think @mark-fitzgerald recently updated this type, so he might have a better answer for you.

mark-fitzgerald: @benchristel - That sounds like a good plan. Thank you!

benchristel: @mark-fitzgerald Updated!

handeyeco (approved):
Looks good to me. Do we need actual tests for this or is banking on snapshots enough?

packages/perseus/src/widgets/interactive-graphs/interactive-graph.test.tsx
```diff
@@ -1169,41 +1169,6 @@ describe("Interactive Graph", function () {
...
-                    expect(PlotOfXMock).toHaveBeenCalledWith(
-                        expect.objectContaining({
-                            ...expectedParameters,
-                            domain: domainExpected,
-                        }),
-                        {},
-                    );
-                },
-            );
-
```
benchristel, line 1206: These tests aren't needed anymore, because the elements of `domain` can no longer be `null` once they get past the parser!

packages/perseus/src/widgets/interactive-graphs/locked-figures/locked-function.tsx
```diff
@@ -16,16 +16,14 @@ const LockedFunction = (props: LockedFunctionType) => {
...
-          ]
-        : undefined;
+    const domain: [min: number, max: number] = [
+        Number.isFinite(props.domain[0])
+            ? (props.domain[0] as number)
+            : -Infinity,
+        Number.isFinite(props.domain[1])
+            ? (props.domain[1] as number)
+            : Infinity,
+    ];
```
benchristel, line 26: This normalization can probably go away, since it's happening in the parser now 🤔

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-function-settings.test.tsx
```diff
@@ -92,38 +92,6 @@ describe("Locked Function Settings", () => {
...
-                wrapper: RenderStateRoot,
-            },
-        );
-
-        // Assert
-        inputField = screen
-            .getByText("domain min")
-            // eslint-disable-next-line testing-library/no-node-access
-            .querySelector("input");
-        expect(inputField?.value).toEqual("");
```
benchristel, line 126: These tests can go away because the elements of `domain` can no longer be `null` here!

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-function-settings.test.tsx
```diff
@@ -836,32 +804,6 @@ describe("Locked Function Settings", () => {
...
-                });
-                await userEvent.click(autoGenButton);
-
-                // Assert
-                expect(onChangeProps).toHaveBeenCalledWith({
-                    ariaLabel:
-                        "Function with equation y=x^2. Appearance solid gray.",
-                });
-            });
-
```
benchristel, line 864: This test isn't needed anymore either. We have a similar test for the case where the `domain` is `[-Infinity, Infinity]`.

benchristel:
@mark-fitzgerald I made some significant updates to this. Now, instead of defaulting the domain to `[null, null]` in the parser, we default to `[-Infinity, Infinity]` to avoid needing another layer of defaults in the React components. Does this look okay to you?

@handeyeco Thanks for the nudge. I've added tests!

packages/perseus/src/widgets/interactive-graphs/interactive-graph.test.tsx
mark-fitzgerald, line 1: I know you didn't add this (I think I accidentally left it in the prior PR for this issue). Would you mind removing the comments on lines 1065 - 1072?
```
            // const domain = (
            //     min: number | null,
            //     max: number | null,
            // ): [number | null, number | null] => {
            //     return [min, max];
            // };
            // type domain = [min: number | null, max: number | null];
```

mark-fitzgerald (approved):
Great job cleaning up the `domain` definitions - thank you!!

PR #2270, opened 2025-02-25
[Interactive Graph] Polygon side snapping is keyboard accessible.

catandthemachines:
## Summary:
Updating Polygon interactive graph implementation to have keyboard accessible side snapping. This is done by hooking up the side snapping behavior into a constraint function to make to keyboard arrow keys.

Issue: LEMS-2243

## Test plan:

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx
@@ -179,7 +179,6 @@ const LimitedPolygonGraph = (statefulProps: StatefulProps) => {
         showSides,
         range,
         snapTo = "grid",
-        coords,
     } = statefulProps.graphState;
     const {disableKeyboardInteraction} = graphConfig;
     const {strings, locale} = usePerseusI18n();
@@ -329,7 +328,7 @@ const LimitedPolygonGraph = (statefulProps: StatefulProps) => {
                             point={point}
                             sequenceNumber={i + 1}
                             onMove={(destination: vec.Vector2) => {
-                                console.log("Movement!");
+                                //console.log("Movement!");
 
                                 const now = Date.now();
                                 const targetFPS = 40;
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx
@@ -1,6 +1,8 @@
-import {angles} from "@khanacademy/kmath";
+import {angles, geometry, vector as kvector} from "@khanacademy/kmath";
+import {approximateEqual} from "@khanacademy/perseus-core";
 import {Polygon, Polyline, vec} from "mafs";
 import * as React from "react";
+import _ from "underscore";
 
 import {
     usePerseusI18n,
@@ -10,7 +12,7 @@ import a11y from "../../../util/a11y";
 import {snap} from "../math";
 import {actions} from "../reducer/interactive-graph-action";
 import useGraphConfig from "../reducer/use-graph-config";
-import {TARGET_SIZE} from "../utils";
+import {bound, isInBound, TARGET_SIZE} from "../utils";
 
 import {PolygonAngle} from "./components/angle-indicators";
 import {MovablePoint} from "./components/movable-point";
@@ -25,6 +27,7 @@ import {
     getSideLengthsFromPoints,
 } from "./utils";
 
+import type {KeyboardMovementConstraint} from "./use-draggable";
 import type {Coord} from "../../../interactive2/types";
 import type {GraphConfig} from "../reducer/use-graph-config";
 import type {
@@ -36,8 +39,11 @@ import type {
     PolygonGraphState,
 } from "../types";
 import type {CollinearTuple} from "@khanacademy/perseus-core";
+import type {Interval} from "mafs";
 
-const {convertRadiansToDegrees} = angles;
+const {getAngleFromVertex, polar, convertRadiansToDegrees} = angles;
+
+const {ccw, lawOfCosines, magnitude, sign, vector} = geometry;
 
 export function renderPolygonGraph(
     state: PolygonGraphState,
@@ -65,7 +71,7 @@ type StatefulProps = MafsGraphProps<PolygonGraphState> & {
     top: number;
     dragging: boolean;
     points: Coord[];
-    constrain: (p: any) => any;
+    constrain: KeyboardMovementConstraint;
     hovered: boolean;
     setHovered: React.Dispatch<React.SetStateAction<boolean>>;
     focusVisible: boolean;
@@ -96,7 +102,19 @@ const PolygonGraph = (props: Props) => {
 
     // Logic to build the dragging experience. Primarily used by Limited Polygon.
     const dragReferencePoint = points[0];
-    const constrain = (p) => snap(snapStep, p);
+    const constrain: KeyboardMovementConstraint = (p) => snap(snapStep, p);
+
+    // switch (snapTo) {
+    //     case "sides":
+    //         // Values are BS, need to fix.
+    //         constrain = getSideSnapConstraint(snapStep, coords[0], 12);
+    //         break;
+    //     case "angles":
+    //     case "grid":
+    //     default:
+    //         constrain = (p) => snap(snapStep, p);
+    //         break;
+    // }
 
     const {dragging} = useDraggable({
         gestureTarget: polygonRef,
@@ -179,6 +197,7 @@ const LimitedPolygonGraph = (statefulProps: StatefulProps) => {
         showSides,
         range,
         snapTo = "grid",
+        snapStep,
     } = statefulProps.graphState;
     const {disableKeyboardInteraction} = graphConfig;
     const {strings, locale} = usePerseusI18n();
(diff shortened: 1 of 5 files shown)


packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx
```diff
@@ -96,9 +99,8 @@ const PolygonGraph = (props: Props) => {
 
     // Logic to build the dragging experience. Primarily used by Limited Polygon.
     const dragReferencePoint = points[0];
-    const constrain = ["angles", "sides"].includes(snapTo)
-        ? (p) => p
-        : (p) => snap(snapStep, p);
+    const constrain: KeyboardMovementConstraint =
```
nishasy, line 102: Nit: Can this be a function named `getConstraintFn` that includes a switch statement for the snapTo type? I think that might be easier to follow for engineers looking at this in the future.

Something like 
```
function getConstraintFn(): KeyboardMovementConstraint {
  switch(snapTo) {
    case "angles":
      return p => p
    case "sides":
      return getSideSnapConstraint(...)
    default: // "grid"
      return p => snap(snapStep, p);
  }
}
``

And then you can pass just this function into `constrain` below.

File: packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx [hunk 2]
```diff
@@ -683,3 +689,60 @@ function describePolygonGraph(
         srPolygonInteractiveElements,
     };
 }
+
+export function getSideSnapConstraint(
```
nishasy, line 693: Is there a test for this?

packages/perseus/src/widgets/interactive-graphs/utils.ts
```diff
@@ -66,6 +66,22 @@ export function bound({
     return clampToBox(boundingBox, point);
 }
 
+// Returns true if the point is within the range of the graph.
+export function isInBound({
```
nishasy, line 70: Is there a test for this?

nishasy:
Question: I see this is allowing approximate sides too. Is that intended?

image

catandthemachines:
> Question: I see this is allowing approximate sides too. Is that intended?
> 
> image

Yes, this is how the functionality works currently. It will show that when it's not exactly a whole number, but it's very close.

packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx
```diff
@@ -96,9 +99,8 @@ const PolygonGraph = (props: Props) => {
 
     // Logic to build the dragging experience. Primarily used by Limited Polygon.
     const dragReferencePoint = points[0];
-    const constrain = ["angles", "sides"].includes(snapTo)
-        ? (p) => p
-        : (p) => snap(snapStep, p);
+    const constrain: KeyboardMovementConstraint =
+        snapTo === "angles" ? (p) => p : (p) => snap(snapStep, p);
```
benchristel, line 103: Could we use a switch statement to get the `constrain` value? Currently, the logic is split between two ternaries that are a couple hundred lines apart, and I think that will confuse future readers.

```ts
function getKeyboardMovementConstraintForPoint(
  pointIndex: number,
): KeyboardMovementConstraint {
  switch (snapTo) {
    case "grid":
      return (p) => snap(snapStep, p);
    case "angles":
      return (p) => p;
    case "sides":
      return getSideSnapConstraint(points, pointIndex, graphConfig.range);
    default:
      throw new UnreachableCaseError(snapTo);
  }
}
```

Complication: it looks like the `constrain` on line 102 is used in the `useDraggable` below, which I think is for moving the whole polygon. So we'd also need:

```ts
function getConstrainFunctionForWholePolygon(): (point: vec.Vector2) => vec.Vector2 {
  return snapTo === "angles" ? (p) => p : (p) => snap(snapStep, p);
}
```

benchristel (changes requested):
This looks good overall! Thanks for this huge accessibility improvement!

Requesting changes because `coords` is being mutated in the reducer; we should make a copy instead. I left a few other suggestions inline too.

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -1070,23 +1070,39 @@ function boundAndSnapToSides(
...
+    destinationPoint: vec.Vector2,
+    range: [Interval, Interval],
+    coords: Coord[],
+    index: number,
+    startingPoint: vec.Vector2,
+) {
     // Takes the destination point and makes sure it is within the bounds of the graph
     // SnapStep is [0, 0] because we don't want to snap to the grid
-    coordsCopy[index] = bound({
+    coords[index] = bound({
```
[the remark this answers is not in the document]
catandthemachines: resolved offline.

packages/perseus/src/widgets/interactive-graphs/utils.ts
```diff
@@ -66,6 +66,22 @@ export function bound({
...
+}: {
+    range: [Interval, Interval];
+    point: vec.Vector2;
+}): boolean {
+    return (
+        point[0] >= range[0][0] &&
+        point[0] <= range[0][1] &&
+        point[1] >= range[1][0] &&
+        point[1] <= range[1][1]
+    );
```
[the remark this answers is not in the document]
catandthemachines: Moved it to `box.ts` and using constants.

packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx
catandthemachines: resolved offline. Built two constrain functions one for individual points and one for whole graph movement.

packages/perseus/src/widgets/interactive-graphs/__snapshots__/interactive-graph.test.tsx.snap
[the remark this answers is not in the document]
catandthemachines: Looked through the file change and it appears to have reordered the DOM a bit. I wonder if this is because of the new constraint functionality. Either way it looks fine. Let em know if you guys notice any issues.

benchristel (approved):
Thanks for the updates! LGTM!

I left a couple small questions and suggestions inline, but nothing blocking.

packages/perseus/src/widgets/interactive-graphs/interactive-graph.testdata.ts
```diff
@@ -302,7 +302,7 @@ export const polygonWithFourSidesSnappingQuestion: PerseusRenderer =
         .withGridStep(0.5, 0.5)
         .withSnapStep(0.25, 0.25)
         .withTickStep(0.5, 0.5)
-        .withMarkings("none")
+        .withMarkings("graph")
```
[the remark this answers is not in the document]
catandthemachines: Oh okay, I can revert them. I updated them because I found debugging very challenging for these testdata examples and I'm curious why we don't want it to have any markings?

packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx
```diff
@@ -690,6 +693,40 @@ function describePolygonGraph(
     };
 }
 
+function getKeyboardMovementConstraintForPoint(
```
nishasy, line 696: Yay, thanks for this!

packages/perseus/src/widgets/interactive-graphs/math/box.test.ts
```diff
@@ -81,3 +82,21 @@ describe("inset", () => {
...
+describe("isInBound", () => {
+    it("returns true whe the point is within the bounds of the graph", () => {
+        const interval: [Interval, Interval] = [
+            [-10, 10],
+            [-10, 10],
+        ];
+        expect(isInBound({range: interval, point: [0, 0]})).toBeTruthy();
+    });
+
+    it("returns false whe the point is outside the bounds of the graph", () => {
```
nishasy, line 95: Can you make this an `it.each` that includes a point outside each bounding side of the graph? Like on on the left, one on the right, one above, and one below? Just in case any of the inequality checks get messed up somehow in the future, then the test would catch it.

nishasy (approved):
LGTM! Just left one small suggestion, but this looks great.

packages/perseus/src/widgets/interactive-graphs/math/box.test.ts
```diff
@@ -81,3 +82,21 @@ describe("inset", () => {
         ]);
     });
 });
+
+describe("isInBound", () => {
+    it("returns true whe the point is within the bounds of the graph", () => {
```
nishasy, line 87:
```suggestion
    it("returns true when the point is within the bounds of the graph", () => {
```

packages/perseus/src/widgets/interactive-graphs/math/box.test.ts
nishasy, line 95:
```suggestion
    it("returns false when the point is outside the bounds of the graph", () => {
```

catandthemachines: Love that idea!

packages/perseus/src/widgets/interactive-graphs/interactive-graph.testdata.ts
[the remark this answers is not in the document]
catandthemachines: Yay! I'll add them back in a follow-up PR.

PR #2272, opened 2025-03-03
[comment-api] add some comments to external APIs

handeyeco:
## Summary:
<Add your summary here>

Issue: LEMS-XXXX

## Test plan:

### DIFF
--- a/packages/perseus-core/src/utils/split-perseus-item.ts
+++ b/packages/perseus-core/src/utils/split-perseus-item.ts
@@ -5,6 +5,11 @@ import {getUpgradedWidgetOptions} from "../widgets/upgrade";
 
 import type {PerseusRenderer} from "../data-schema";
 
+/**
+ * get a copy of a Perseus item with pre-scoring data removed (ie answers)
+ *
+ * @param originalItem - the full answer data, includes the correct answer
+ */
 export default function splitPerseusItem(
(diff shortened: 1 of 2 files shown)


packages/perseus-core/src/utils/split-perseus-item.ts
```diff
@@ -5,6 +5,11 @@ import {getUpgradedWidgetOptions} from "../widgets/upgrade";
 
 import type {PerseusRenderer} from "../data-schema";
 
+/**
+ * get a copy of a Perseus item with pre-scoring data removed (ie answers)
```
jeremywiebe, line 9: I think we agreed to call this rubric data. 
```suggestion
 * Return a copy of a Perseus item with rubric data removed (ie answers)
```

packages/perseus-core/src/utils/split-perseus-item.ts
```diff
@@ -5,6 +5,11 @@ import {getUpgradedWidgetOptions} from "../widgets/upgrade";
...
+ *
+ * @param originalItem - the full answer data, includes the correct answer
```
jeremywiebe, line 11:
```suggestion
 * @param originalItem - the original, full Perseus item (which includes the rubric - aka answer data)
```

jeremywiebe (approved):

PR #2273, opened 2025-03-03
Add a basic FreeResponse widget

aag:
## Summary:
This commit adds a basic version of the new FreeResponse widget. This
widget allows users to enter free text into an input field. The initial
use case is for "short answer" type questions.

The UX of the widget is not complete in this commit. Instead, the
intention is to get all of the files in place with a basic UI which will
then be iterated on with more widget options and more sophisticated
rendering.

Issue: https://khanacademy.atlassian.net/browse/LIT-1661

## Test plan:
- View the editor and rendered widget stories in Storybook.
- Run the tests.

### DIFF
--- a/packages/perseus-core/src/widgets/free-response/index.ts
+++ b/packages/perseus-core/src/widgets/free-response/index.ts
@@ -1,0 +1,18 @@
+import type {PerseusFreeResponseWidgetOptions} from "../../data-schema";
+import type {WidgetLogic} from "../logic-export.types";
+
+export type FreeResponseDefaultWidgetOptions = Pick<
+    PerseusFreeResponseWidgetOptions,
+    "question"
+>;
+
+const defaultWidgetOptions: PerseusFreeResponseWidgetOptions = {
+    question: "",
+};
+
+const freeResponseWidgetLogic: WidgetLogic = {
+    name: "free-response",
+    defaultWidgetOptions,
+};
+
+export default freeResponseWidgetLogic;
--- a/packages/perseus-editor/src/widgets/free-response-editor.tsx
+++ b/packages/perseus-editor/src/widgets/free-response-editor.tsx
@@ -1,0 +1,46 @@
+import {freeResponseLogic} from "@khanacademy/perseus-core";
+import {LabeledTextField} from "@khanacademy/wonder-blocks-form";
+import {spacing} from "@khanacademy/wonder-blocks-tokens";
+import * as React from "react";
+
+import type {PerseusFreeResponseWidgetOptions} from "@khanacademy/perseus-core";
+
+type Props = PerseusFreeResponseWidgetOptions & {
+    onChange: (arg1: {question?: Props["question"]}) => void;
+};
+
+class FreeResponseEditor extends React.Component<Props> {
+    static defaultProps: PerseusFreeResponseWidgetOptions =
+        freeResponseLogic.defaultWidgetOptions;
+
+    static widgetName = "free-response" as const;
+
+    serialize(): PerseusFreeResponseWidgetOptions {
+        return {
+            question: this.props.question,
+        };
+    }
+
+    getSaveWarnings(): Array<string> {
+        const warnings: Array<string> = [];
+        if (!this.props.question) {
+            warnings.push("The question is empty");
+        }
+        return warnings;
+    }
+
+    render(): React.ReactNode {
+        return (
+            <LabeledTextField
+                label={"Question"}
+                value={this.props.question}
+                onChange={(question: string) => this.props.onChange({question})}
+                style={{
+                    marginBottom: spacing.large_24,
+                }}
+            />
+        );
+    }
+}
+
+export default FreeResponseEditor;
--- a/packages/perseus-score/src/validation.types.ts
+++ b/packages/perseus-score/src/validation.types.ts
@@ -225,6 +225,11 @@ export type PerseusNumericInputUserInput = {
     currentValue: string;
 };
 
+export type PerseusFreeResponseUserInput = {
+    currentValue: string;
+    question: string;
+};
+
 export type PerseusOrdererRubric = PerseusOrdererWidgetOptions;
 
 export type PerseusOrdererUserInput = {
@@ -321,6 +326,7 @@ interface UserInputRegistry {
     "cs-program": PerseusCSProgramUserInput;
     dropdown: PerseusDropdownUserInput;
     expression: PerseusExpressionUserInput;
+    "free-response": PerseusFreeResponseUserInput;
     grapher: PerseusGrapherUserInput;
     group: PerseusGroupUserInput;
     iframe: PerseusIFrameUserInput;
--- a/packages/perseus/src/widgets/free-response/free-response.test.tsx
+++ b/packages/perseus/src/widgets/free-response/free-response.test.tsx
@@ -1,0 +1,94 @@
+import {screen} from "@testing-library/react";
+import {userEvent as userEventLib} from "@testing-library/user-event";
+
+import {testDependencies} from "../../../../../testing/test-dependencies";
+import * as Dependencies from "../../dependencies";
+import {renderQuestion} from "../__testutils__/renderQuestion";
+
+import {question} from "./free-response.testdata";
+
+import type {APIOptions} from "../../types";
+import type {UserEvent} from "@testing-library/user-event";
+
+describe("free-response widget", () => {
+    let userEvent: UserEvent;
+
+    beforeEach(() => {
+        userEvent = userEventLib.setup({
+            advanceTimers: jest.advanceTimersByTime,
+        });
+
+        jest.spyOn(Dependencies, "getDependencies").mockReturnValue(
+            testDependencies,
+        );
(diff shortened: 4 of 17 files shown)


packages/perseus-core/src/widgets/free-response/index.ts
```diff
@@ -0,0 +1,18 @@
...
+import type {WidgetLogic} from "../logic-export.types";
+
+export type FreeResponseDefaultWidgetOptions = Pick<
+    PerseusFreeResponseWidgetOptions,
+    "question"
+>;
+
+const defaultWidgetOptions: PerseusFreeResponseWidgetOptions = {
+    question: "",
+};
```
benchristel, line 11: What's the purpose of this default? It looks like `question` is required in the widget options, so when would the default ever be used?

packages/perseus-editor/src/widgets/free-response-editor.tsx
```diff
@@ -0,0 +1,46 @@
+import {freeResponseLogic} from "@khanacademy/perseus-core";
+import {LabeledTextField} from "@khanacademy/wonder-blocks-form";
+import {spacing} from "@khanacademy/wonder-blocks-tokens";
+import * as React from "react";
+
+import type {PerseusFreeResponseWidgetOptions} from "@khanacademy/perseus-core";
+
+type Props = PerseusFreeResponseWidgetOptions & {
+    onChange: (arg1: {question?: Props["question"]}) => void;
```
benchristel, line 9: I think the type of the onChange arg should be `Partial<PerseusFreeResponseWidgetOptions>`.

packages/perseus-score/src/validation.types.ts
```diff
@@ -225,6 +225,11 @@ export type PerseusNumericInputUserInput = {
     currentValue: string;
 };
 
+export type PerseusFreeResponseUserInput = {
+    currentValue: string;
+    question: string;
```
benchristel, line 230: Why include the question here?

benchristel (commented):
Looks very nice! Thank you!

One thing you'll need to do before landing this is create a parser for this widget in `packages/perseus-core/src/parse-perseus-json/perseus-parsers`, and use it in the PerseusWidgetsMap parser in `widgets-map.ts`.

That said, I don't think anything will fail (for now) if the parsers aren't updated. The parser currently serves two purposes:

- it alerts us (via Sentry logs) if it encounters data that doesn't match the schema.
- it migrates old widget options formats to the latest version.

The parser allows widgets of an unrecognized type (since they might have been registered dynamically), so it won't fail if you feed it a free-response widget as-is. However, we probably don't want to rely on that behavior for first-party widgets.

packages/perseus-core/src/widgets/free-response/index.ts
aag: I was following the pattern I see in other widgets, e.g. [PassageWidget](https://github.com/Khan/perseus/blob/basic-free-response-widget/packages/perseus-core/src/widgets/passage/index.ts#L11). I think it makes sense to have `""` be the default `question` prop value in `FreeResponseEditor`, but I'm not sure how else the `freeResponseLogic` is used. Do you know what else it's used for?

packages/perseus-score/src/validation.types.ts
aag: Good question. That was an initial attempt to make it easy to extract the question from the widget, but we've realized it's not needed, so I've removed it.

packages/perseus/src/widgets/free-response/free-response.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+import {View, Id} from "@khanacademy/wonder-blocks-core";
+import * as React from "react";
+
+import type {Widget, WidgetExports, WidgetProps} from "../../types";
+import type {PerseusFreeResponseWidgetOptions} from "@khanacademy/perseus-core";
+import type {UserInput} from "@khanacademy/perseus-score";
+
+type RenderProps = PerseusFreeResponseWidgetOptions;
+type Props = WidgetProps<RenderProps, PerseusFreeResponseWidgetOptions>;
```
handeyeco, line 16: You're catching us in a moment of self-reflection where we're trying to decide if `RenderProps === WidgetOptions`. `WidgetOptions` is basically the "data at rest" which should include pretty much everything needed to render/interact/score/etc. `RenderProps` is kind of a subset of `WidgetOptions`, but it also can have its own data that's not stored in `WidgetOptions`; it's specifically "what do we need to put a widget on the screen?"

For instance you were talking about adding something to `WidgetOptions` that will not be sent to the learner but will be sent to the LLM. So _right now_ `RenderProps === WidgetOptions`, but you already have an idea for why that won't be the case in the medium term.

So anyway:

```suggestion
type RenderProps = Pick<PerseusFreeResponseWidgetOptions, "question">;
type Props = WidgetProps<RenderProps>;
```

It's the same result but with a clearer motive.

packages/perseus/src/widgets/free-response/free-response.tsx
```diff
@@ -0,0 +1,67 @@
...
+{
+    state: State = {
+        currentValue: "",
+    };
+
+    handleChange = (event: React.ChangeEvent<HTMLTextAreaElement>) => {
+        this.setState({currentValue: event.target.value});
+    };
+
+    getUserInput(): UserInput {
```
handeyeco, line 34:
```suggestion
    getUserInput(): PerseusFreeResponseUserInput {
```

packages/perseus/src/widgets/free-response/free-response.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+        expect(userInput).toMatchObject({
+            "free-response 1": {
+                currentValue: "test-answer",
+            },
+        });
+    });
+
+    // TODO(agoforth): Create a custom validator for the free-response widget
+    //   that will cause this test to pass.
+    // it("should be included in the empty widgets list if no text has been input yet", async () => {
```
handeyeco, line 82: Ideally we wouldn't leave commented code around. If you plan to fix this before the feature branch lands to main, all good. If you plan to fix this _someday_, I think this would be better in a ticket.

packages/perseus/src/widgets/free-response/free-response.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+        // Assert
+        expect(screen.getByLabelText("test-question")).toBeVisible();
+    });
+
+    it("should return the correct user input", async () => {
+        // Arrange
+        const {renderer} = renderQuestion(question);
+        await userEvent.type(
+            screen.getByLabelText("test-question"),
```
handeyeco, line 65: I don't think it matters much, but Testing Library [docs](https://testing-library.com/docs/queries/about/#priority) state a preference for `getByRole` over `getByLabelText` (though just by a hair).

```suggestion
            screen.getByRole("textbox", {name: "test-question"}),
```

handeyeco (approved):
This is amazingly comprehensive and well-written. I don't see any blockers, especially since this is landing into a feature branch and not main.

I still have some architectural questions to answer before we land the feature branch, but I think this makes sense as y'all are roughing out an idea of how the widget will work.

packages/perseus-score/src/validation.types.ts
```diff
@@ -225,6 +225,10 @@ export type PerseusNumericInputUserInput = {
     currentValue: string;
 };
 
+export type PerseusFreeResponseUserInput = {
+    currentValue: string;
+};
```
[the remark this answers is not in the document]
aag: We don't know of any other user input fields, but I like the flexibility of returning an object in case something else comes up, so I'm going to leave this as it is.

packages/perseus/src/widgets/free-response/free-response.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+        jest.spyOn(Dependencies, "getDependencies").mockReturnValue(
+            testDependencies,
+        );
+    });
+
+    it("should snapshot", () => {
+        // Arrange
+        const apiOptions: APIOptions = {
+            isMobile: false,
+        };
```
jeremywiebe, line 30: I suspect this was done by following other tests. Ideally we rely less and less on the `isMobile` flag as that causes our code to diverge for mobile vs desktop. I think you could remove the test that checks for snapshotting on mobile and just remove the `isMobile` flag from APIOptions here. This widget, especially, feels like something that won't vary between desktop and mobile.. I could be wrong though.

packages/perseus/src/widgets/free-response/free-response.tsx
```diff
@@ -0,0 +1,75 @@
...
+}
+
+export default {
+    name: "free-response",
+    accessible: true,
+    displayName: "Free Response",
+    widget: FreeResponse,
+    // Hides widget from content creators until full release
+    hidden: true,
+} as WidgetExports<typeof FreeResponse>;
```
jeremywiebe, line 67: All other widgets use `satisfies WidgetExports<...` for these objects. We've found that `satisfies` is better at ensuring the literal object type checks properly.

PR #2277, opened 2025-03-04
[SR] Add aria-hidden to individual components of interactive graph elements

nishasy:
## Summary:
When using Safari + VoiceOver, the screen reader reads out every single internal part of every single element on the graph. For example, the arrow on a line is read out as "image" as well as the segment composing the line. This means that when traversing the graph in Safari, there are a whole bunch of empty "image"s being read out with no context. The context is provided by specific element's aria labels.

To remedy this, I added `aria-hidden` to all the visual elements that should not be read as "image" in all the graph types.

Issue: none

## Test Plan:
Storybook
- http://localhost:6006/?path=/docs/perseuseditor-widgets-interactive-graph--docs
- Go to every graph type and traverse through it with a screen reader
- Confirm there are no random empty "image" calls

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/quadratic.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/quadratic.tsx
@@ -81,7 +81,16 @@ function QuadraticGraph(props: QuadraticGraphProps) {
             aria-label={srQuadraticGraph}
             aria-describedby={`${quadraticDirectionId} ${quadraticVertexId} ${quadraticInterceptsId}`}
         >
-            <Plot.OfX y={y} color={color.blue} />
+            <Plot.OfX
+                y={y}
+                color={color.blue}
+                svgPathProps={{
+                    // Use aria-hidden to hide the line from screen readers
+                    // so it doesn't read as "image" with no context.
+                    // This is okay because the graph has its own aria-label.
+                    "aria-hidden": true,
+                }}
+            />
             {coords.map((coord, i) => {
                 const srQuadraticPoint = getQuadraticPointString(
                     i + 1,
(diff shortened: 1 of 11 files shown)


catandthemachines:
Does this have any adverse effects on other screen readers? I know this is pretty bad on Safari, but I want to make sure we're not over correcting for issues that should be logged against Safari/VoiceOver.

nishasy:
@catandthemachines This doesn't affect the experience in any other browsers, it just corrects the issue in Safari. The other browsers didn't read out those elements in the first place, so this is just the status quo for those.

packages/perseus/src/widgets/interactive-graphs/graphs/quadratic.tsx
```diff
@@ -81,7 +81,16 @@ function QuadraticGraph(props: QuadraticGraphProps) {
             aria-label={srQuadraticGraph}
             aria-describedby={`${quadraticDirectionId} ${quadraticVertexId} ${quadraticInterceptsId}`}
         >
-            <Plot.OfX y={y} color={color.blue} />
+            <Plot.OfX
+                y={y}
+                color={color.blue}
+                svgPathProps={{
+                    // Use aria-hidden to hide the line from screen readers
```
SonicScrewdriver, line 88: Ooo fun I never looked into this but I'm glad mafs allowed us to pass props through.

SonicScrewdriver (approved):
Looks good to me! Thanks for this great fix. :) 

Note: I tried testing what I currently have in storybook and I wasn't able to recreate it with Safari. 🤔 I wonder if this is a more recent bug that they've introduced, as I'm a few versions behind.

catandthemachines (approved):
No impact to other screen readers, lets go! 😎

nishasy:
Update: Third was able to repro when using control + option + arrow keys.

PR #2278, opened 2025-03-04
Remove jQuery from answer types

handeyeco:
## Summary:
I'm kind of just trying this out, but I think it would be good to remove jQuery where possible anyway.

See: https://khanacademy.slack.com/archives/CDHPTMSTB/p1741115861977089


packages/perseus-score/src/util/answer-types.ts
```diff
@@ -230,13 +235,20 @@ const KhanAnswerTypes = {
...
                 proper: function (text) {
-                    return $.map(fractionTransformer(text), function (o) {
-                        // All fractions that are less than 1
-                        if (Math.abs(o.value) < 1) {
-                            return [o];
-                        }
-                        return [];
-                    });
+                    const transformed = fractionTransformer(text);
+                    return transformed.reduce(
```
handeyeco, line 239: jQuery's `map` seemingly does some special flattening/filtering that's not as straight-forward as JS' `map`: https://github.com/HubSpot/youmightnotneedjquery/issues/355

packages/perseus-score/src/util/answer-types.ts
```diff
@@ -442,17 +461,17 @@ const KhanAnswerTypes = {
 
                 // Numbers with percent signs
                 percent: function (text) {
-                    text = $.trim(text);
+                    text = String(text).trim();
```
handeyeco, line 464: jQuery's `trim` coerces numbers to strings and apparently our code was banking on that.

packages/perseus-score/src/util/answer-types.ts
jeremywiebe: Agreed.  jQuery's `map` says for the mapping callback:

```
The function can return:

* the translated value, which will be mapped to the resulting array
* null or undefined, to remove the item
* an array of values, which will be flattened into the full array
```

That last point comes into play.  I think you could just use [`flatMap()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap) here instead of `reduce()`

image

jeremywiebe (approved):
I'd love to see us build more unit tests around this file. It's pretty complex and has a pretty light set of tests. :)  Not blocking this PR though!

packages/perseus-score/src/util/answer-types.ts
handeyeco: Good call out, thank you

PR #2279, opened 2025-03-04
[SR] Tweak strings for Segment and Angle graphs

nishasy:
## Summary:
Segment
- Update segment so that a singular segment doesn't have multiple layers of redundant labels (keep this for multiple segments)

Angle
- Update angle so the vertex point always reads the angle, not only when "show angles" is turned on.

Issue: none

## Test plan:
`yarn jest packages/perseus/src/widgets/interactive-graphs/graphs/segment.test.tsx`
`yarn jest packages/perseus/src/widgets/interactive-graphs/graphs/angle.test.tsx`

Storybook
- http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-segment
- http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-angle

### DIFF
--- a/.changeset/calm-dolls-dance.md
+++ b/.changeset/calm-dolls-dance.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+[SR] Tweak strings for Segment and Angle graphs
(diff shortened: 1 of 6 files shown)


catandthemachines (approved):
LGTM. Ship it!

PR #2281, opened 2025-03-04
[Interactive Graph] Improve Polygon angle snapping is for keyboard users.

catandthemachines:
## Summary:
Updating Polygon interactive graph implementation to have a better keyboard accessible angle snapping. This is done by hooking up the angle snapping behavior into a constraint function to make the keyboard arrow keys more effective in locating the next best point and angle combination.

Issue: LEMS-2893

## Test plan:
- Go to: /?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-polygon
- Set **Snap to:** to **angle measures**
- Use your keyboard (tab and arrow keys) to move the various polygon points.
- Notice the movement is smaller to have a better chance of getting the angles correct!


packages/perseus/src/widgets/interactive-graphs/interactive-graph.testdata.ts
catandthemachines, line 1: Per discussion in our previous PR, I am removing these `.withMarkings("none")` type as it's unnecessary for testing and make it harder to debug for developers.

https://github.com/Khan/perseus/pull/2270#discussion_r1980175969

packages/perseus/src/widgets/interactive-graphs/graphs/polygon.test.tsx
```diff
@@ -605,3 +609,51 @@ describe("getSideSnapConstraint", () => {
...
+            [0, 0],
+            [0, 2],
+            [2, 2],
+            [2, 0],
+        ];
+
+        // We're moving the third point in the top right corner of the polygon (square).
+        const constraint = getAngleSnapConstraint(points, 2, range);
+
+        expect(constraint).toEqual({
```
nishasy, line 629: Can you leave a quick comment here about what angle these new points are supposed to represent?

packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx
```diff
@@ -706,7 +709,7 @@ function getKeyboardMovementConstraintForPoint(
         case "sides":
             return getSideSnapConstraint(points, index, range);
         case "angles":
-            return (p) => p;
+            return getAngleSnapConstraint(points, index, range);
```
nishasy, line 712: 🎉

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -972,6 +972,22 @@ function boundAndSnapToPolygonAngle(
 ) {
     const startingPoint = coords[index];
 
+    return calculateAngleSnap(
+        destinationPoint,
+        range,
+        coords,
+        index,
+        startingPoint,
+    ) as vec.Vector2;
```
nishasy, line 981: Generally we'd want to avoid `as` since that blows away typechecking and. If you need to keep it, you can add a comment explaining why.

```suggestion
    ) satisfies vec.Vector2;
```

nishasy (changes requested):
When trying stuff out in [this PR's storybook publish](https://650db21c3f5d1b2f13c02952-qcigaahhrt.chromatic.com/?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-polygon), I observed 2 supposed bugs.

1. I held down the arrow key on a point with interior angle snapping and eventually the whole graph crashed. I'm having a hard time reproducing it now though.

Screenshot 2025-03-04 at 5 19 42 PM

2. It looks like the point gets stuck on the edges of the graph when snapping to interior angles.

https://github.com/user-attachments/assets/82e97f88-800f-4349-ad6c-340c12bce021

nishasy:
@catandthemachines

> I held down the arrow key on a point with interior angle snapping and eventually the whole graph crashed. I'm having a hard time reproducing it now though.

Never mind, I can repro it consistently now using the configuration in this video.

https://github.com/user-attachments/assets/391ad2e8-ec3a-458a-bd15-ee4149bd89a8

catandthemachines:
> @catandthemachines
> 
> > I held down the arrow key on a point with interior angle snapping and eventually the whole graph crashed. I'm having a hard time reproducing it now though.
> 
> Never mind, I can repro it consistently now using the configuration in this video.
> 
>  Screen.Recording.2025-03-05.at.4.22.36.PM.mov

Okay I think I resolved the issue? I'll still file the bug that we saw just in case with some context but hopefully it's gone?

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.tsx
```diff
@@ -45,8 +45,17 @@ export const PolygonAngle = ({
     const b = vec.dist(centerPoint, endPoints[1]);
     const c = vec.dist(endPoints[0], endPoints[1]);
 
+    let equation = (a ** 2 + b ** 2 - c ** 2) / (2 * a * b);
```
nishasy, line 48: Super duper tiny nit: rename `equation` to `expression` since there is no equal sign inside it.

I think if you want to get super technical, you can call it `lawOfCosinesRadicand`?? I tried looking it up and it doesn't seem to have a specific name 🤔

packages/perseus/src/widgets/interactive-graphs/graphs/utils.ts
```diff
@@ -269,8 +269,17 @@ export function getAngleFromPoints(points: Coord[], i: number) {
     const b = vec.dist(point, pt2);
     const c = vec.dist(pt1, pt2);
 
+    let equation = (a ** 2 + b ** 2 - c ** 2) / (2 * a * b);
```
nishasy, line 272: Same tiny nit here.

nishasy (approved):
While manually testing in the current storybook publish, I confirmed that the NaN and the getting stuck on the border bugs are gone!

Unfortunately, it did explode on me again :( But it's so hard to repro and it seems like a pretty weird and rare edge case. I'm in favor of filing a bug to come back to it later.

catandthemachines:
Bug to fix the weird exploding: https://khanacademy.atlassian.net/browse/LEMS-2901

PR #2282, opened 2025-03-05
[its-a-drag-to-tap-out] docs(changeset): Ensure that dragging a point also moves the focus for Interactive Graph

SonicScrewdriver:
## Summary:
This PR is part of the Interactive Graph project. 

The purpose of the PR is to ensure that dragging a point with a touch event also moves the focus to that point. This is to ensure that the focus always remains on the last element interacted with, which avoids confusion when the user tries to delete points or move them with the keyboard. 

NOTE: I wasn't able to fix a bug related to the hover state sizing of the control points, as I simply cannot seem to find where this is occurring. As a result, touch-dragging a point without any delay can result in the point not getting "bigger". However, the focus outline, hairlines, and focus itself all move to the correct element. I've attached a video below to demonstrate this.

## Video:
https://github.com/user-attachments/assets/59e26360-b54e-4340-b49b-c4f9c32c4baf

Issue: LEMS-2475

## Test plan:
- Manual testing

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/components/use-control-point.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/components/use-control-point.tsx
@@ -103,6 +103,7 @@ export function useControlPoint(params: Params): Return {
         if (dragging && !focused) {
             // If the point is being dragged, focus the focusable handle so that
             // users can continue to interact with the point using the keyboard or buttons.
+            // This particular focus call ensures that the focus ring and hairlines are visible.
             focusableHandleRef.current?.focus();
         }
     }, [dragging, focused]);
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/components/use-control-point.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/components/use-control-point.tsx
@@ -68,6 +68,7 @@ export function useControlPoint(params: Params): Return {
 
     const [focused, setFocused] = useState(false);
     const focusableHandleRef = useRef<SVGGElement>(null);
+
     useDraggable({
         gestureTarget: focusableHandleRef,
         point,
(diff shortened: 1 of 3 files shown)


packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -507,6 +507,7 @@ function doMovePoint(
             return {
                 ...state,
                 hasBeenInteractedWith: true,
+                focusedPointIndex: action.index,
```
SonicScrewdriver, line 510: This ensures that our reducer is aware of the current focused point for future interactions, such as deleting points.

packages/perseus/src/widgets/interactive-graphs/graphs/components/use-control-point.tsx
```diff
@@ -92,10 +93,20 @@ export function useControlPoint(params: Params): Return {
...
     useLayoutEffect(() => {
         setForwardedRef(forwardedRef, focusableHandleRef.current);
     }, [forwardedRef]);
 
+    // If the point is being dragged and is not focused, focus the focusable handle.
+    useLayoutEffect(() => {
+        if (dragging && !focused) {
+            // If the point is being dragged, focus the focusable handle so that
+            // users can continue to interact with the point using the keyboard or buttons.
+            focusableHandleRef.current?.focus();
```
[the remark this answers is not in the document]
nishasy: You can also add this as a comment in the code.

packages/perseus/src/widgets/interactive-graphs/graphs/components/use-control-point.tsx
```diff
@@ -92,10 +93,20 @@ export function useControlPoint(params: Params): Return {
...
+        }
```
nishasy, line 107: Question: Should there also be a `setFocused(true)` here to update the stored state after setting the focus?

nishasy (approved):
Looks great when manually testing! Just left one small question.

packages/perseus/src/widgets/interactive-graphs/graphs/components/use-control-point.tsx
SonicScrewdriver: I thought so too, but I tested it out and by manually triggering the focus, we also trigger the onFocus event on that element that calls the setFocused value already!

PR #2286, opened 2025-03-06
Add scoring criteria to Free Response Widget

aag:
## Summary:
This commit adds a new scoringCriteria option to the Free Response
Widget editor. It allows users to input one or more scoring criteria
texts, which can be used when grading the answers to a Free Response
question.

This commit also changes the "Question" input to be a textarea instead
of an input element. This allows longer questions to wrap and be fully
visible even when the widget editor is narrow.

Issue: https://khanacademy.atlassian.net/browse/LIT-1663

## Test plan:
- Run `pnpm test packages/perseus-editor/src/widgets/__tests__/free-response-editor.test.tsx`
- Open Storybook.
- Open the Free Response Editor stories.
- Verify that the rendered stories look right.
- Interact with the "Editable" story, verifying that it's possible to
  add new criteria and delete existing ones.
- Verify that the "Delete" button is grayed out on the "One Criterion" story.
image

### DIFF
--- a/packages/perseus-core/src/parse-perseus-json/perseus-parsers/free-response-widget.ts
+++ b/packages/perseus-core/src/parse-perseus-json/perseus-parsers/free-response-widget.ts
@@ -1,4 +1,4 @@
-import {constant, object, string} from "../general-purpose-parsers";
+import {array, constant, object, string} from "../general-purpose-parsers";
 
 import {parseWidget} from "./widget";
 
@@ -9,5 +9,10 @@ export const parseFreeResponseWidget: Parser<FreeResponseWidget> = parseWidget(
     constant("free-response"),
     object({
         question: string,
+        scoringCriteria: array(
+            object({
+                text: string,
+            }),
+        ),
     }),
 );
--- a/packages/perseus-core/src/widgets/free-response/free-response-util.ts
+++ b/packages/perseus-core/src/widgets/free-response/free-response-util.ts
@@ -2,7 +2,7 @@ import type {PerseusFreeResponseWidgetOptions} from "@khanacademy/perseus-core";
 
 /**
  * For details on the individual options, see the
- * PerseusFreeResponseWidgetOptions type
+ * {@link PerseusFreeResponseWidgetOptions} type
  */
 export type FreeResponsePublicWidgetOptions = {
     question: PerseusFreeResponseWidgetOptions["question"];
--- a/packages/perseus-core/src/widgets/free-response/free-response-util.ts
+++ b/packages/perseus-core/src/widgets/free-response/free-response-util.ts
@@ -1,0 +1,23 @@
+import type {PerseusFreeResponseWidgetOptions} from "@khanacademy/perseus-core";
+
+/**
+ * For details on the individual options, see the
+ * PerseusFreeResponseWidgetOptions type
+ */
+export type FreeResponsePublicWidgetOptions = {
+    question: PerseusFreeResponseWidgetOptions["question"];
+};
+
+/**
+ * Given a FreeResponsePublicWidgetOptions object, return a new object with only
+ * the public options that should be exposed to the client.
+ */
+function getFreeResponsePublicWidgetOptions(
+    options: PerseusFreeResponseWidgetOptions,
+): FreeResponsePublicWidgetOptions {
+    return {
+        question: options.question,
+    };
+}
+
+export default getFreeResponsePublicWidgetOptions;
--- a/packages/perseus-editor/src/widgets/free-response-editor.tsx
+++ b/packages/perseus-editor/src/widgets/free-response-editor.tsx
@@ -39,7 +39,7 @@ class FreeResponseEditor extends React.Component<Props> {
         return warnings;
     }
 
-    handleCriterionUpdate = (
+    handleUpdateCriterion = (
         index: number,
         criterion: PerseusFreeResponseWidgetScoringCriterion,
     ) => {
@@ -55,7 +55,7 @@ class FreeResponseEditor extends React.Component<Props> {
         });
     };
 
-    handleCriterionDelete = (index: number) => {
+    handleDeleteCriterion = (index: number) => {
         this.props.onChange({
             scoringCriteria: this.props.scoringCriteria.filter(
                 (_, i) => i !== index,
@@ -77,8 +77,8 @@ class FreeResponseEditor extends React.Component<Props> {
                     index={index}
                     key={index}
                     numCriteria={this.props.scoringCriteria.length}
-                    onChange={this.handleCriterionUpdate}
-                    onDelete={this.handleCriterionDelete}
+                    onChange={this.handleUpdateCriterion}
+                    onDelete={this.handleDeleteCriterion}
                 />
             );
         });
--- a/packages/perseus-editor/src/widgets/free-response-editor.tsx
+++ b/packages/perseus-editor/src/widgets/free-response-editor.tsx
@@ -130,7 +130,7 @@ const CriterionEditor = function (props: {
     return (
         <View style={styles.criterionContainer}>
             <textarea
-                data-testid={`criterion-input-${props.index}`}
+                aria-label={`Criterion ${props.index + 1}`}
                 onChange={(e) =>
                     props.onChange(props.index, {
                         text: e.target.value,
@@ -140,8 +140,8 @@ const CriterionEditor = function (props: {
             />
             <View style={styles.deleteButtonContainer}>
                 <Button
+                    aria-label={`Delete criterion ${props.index + 1}`}
                     color="destructive"
-                    data-testid={`criterion-delete-button-${props.index}`}
                     disabled={isOnlyCriterion}
                     kind="tertiary"
                     onClick={() => props.onDelete(props.index)}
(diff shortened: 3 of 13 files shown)


rgpass (commented):
LGTM, just one idea.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/free-response-widget.ts
```diff
@@ -9,5 +9,10 @@ export const parseFreeResponseWidget: Parser<FreeResponseWidget> = parseWidget(
     constant("free-response"),
     object({
         question: string,
+        scoringCriteria: array(
+            object({
+                text: string,
+            }),
+        ),
```
benchristel, line 16: Have any free-response widgets been created in production? Any existing free-response widgets won't have `scoringCriteria`, so this parser will fail for them.

If you need to be compatible with existing data, I'd suggest defaulting to an empty array:

```ts
scoringCriteria: defaulted(
    array(
        object({
            text: string,
        }),
    ),
    () => [],
),
```

And you'd also want to add a regression test to `parse-perseus-json/regression-tests/item-data` to demonstrate the need for the default.

packages/perseus-core/src/widgets/free-response/free-response-util.ts
```diff
@@ -0,0 +1,23 @@
+import type {PerseusFreeResponseWidgetOptions} from "@khanacademy/perseus-core";
+
+/**
+ * For details on the individual options, see the
+ * PerseusFreeResponseWidgetOptions type
```
benchristel, line 5: Thanks for adding this doc comment! To make it even more helpful, you could make the reference a link:

```suggestion
 * {@link PerseusFreeResponseWidgetOptions} type
```

packages/perseus-editor/src/widgets/free-response-editor.tsx
```diff
@@ -32,18 +39,145 @@ class FreeResponseEditor extends React.Component<Props> {
...
+        criterion: PerseusFreeResponseWidgetScoringCriterion,
+    ) => void;
+    onDelete: (index: number) => void;
+}): React.ReactNode {
+    const isOnlyCriterion = props.numCriteria <= 1;
+
+    return (
+        <View style={styles.criterionContainer}>
+            <textarea
+                data-testid={`criterion-input-${props.index}`}
```
handeyeco, line 133: I wonder if rather than using test IDs if we could use descriptive labels. "Criterion 1", "Delete Criterion 1", etc. We try our best to avoid test IDs when possible although we recognize that isn't always possible, just because if it's hard for us to find it with accessible names that means it's hard to find it with a11y tools.

packages/perseus-editor/src/widgets/free-response-editor.tsx
```diff
@@ -32,18 +39,145 @@ class FreeResponseEditor extends React.Component<Props> {
...
+                        text: e.target.value,
+                    })
+                }
+                value={props.criterion.text}
+            />
+            <View style={styles.deleteButtonContainer}>
+                <Button
+                    color="destructive"
+                    data-testid={`criterion-delete-button-${props.index}`}
+                    disabled={isOnlyCriterion}
```
handeyeco, line 145: I don't know which is preferred by designers: not showing the button when it can't be used or disabling the button when it can't be used. 🤔

packages/perseus-editor/src/widgets/free-response-editor.tsx
```diff
@@ -32,18 +39,145 @@ class FreeResponseEditor extends React.Component<Props> {
...
+        <View style={styles.criterionContainer}>
+            <textarea
+                data-testid={`criterion-input-${props.index}`}
+                onChange={(e) =>
+                    props.onChange(props.index, {
+                        text: e.target.value,
+                    })
+                }
+                value={props.criterion.text}
+            />
```
benchristel, line 140: To be accessible, this textarea should have a unique label, like "Text of Criterion 1". If it had such a label, we wouldn't need the test ID.

packages/perseus-editor/src/widgets/free-response-editor.tsx
```diff
@@ -32,18 +39,145 @@ class FreeResponseEditor extends React.Component<Props> {
...
+                <Button
+                    color="destructive"
+                    data-testid={`criterion-delete-button-${props.index}`}
+                    disabled={isOnlyCriterion}
+                    kind="tertiary"
+                    onClick={() => props.onDelete(props.index)}
+                    size="small"
+                    startIcon={trashIcon}
+                >
+                    Delete
```
benchristel, line 151: To be accessible, this button's text should be unique, e.g. "Delete Criterion 1". It might also be okay to put the unique text in an aria-label and not have it visible... I'm not an accessibility expert.

If the buttons had unique labels, we wouldn't need testIds.

benchristel (commented):
Looks good overall! I left a few comments and questions inline. In short:

- the parser might need a default for the criteria - depends on whether production data exists for this widget yet.
- the editor could be made more accessible (and testable!)

packages/perseus-editor/src/widgets/free-response-editor.tsx
benchristel: Is there a way to allow deleting all criteria in the editor but block the item from being saved if the widget has no criteria. The "no criteria" state could be valuable as a transient state, e.g. if a content creator wants to remove and replace all criteria for a widget.

(as an aside, it irks me in general when tools prevent you from temporarily entering a state that isn't globally "valid". An example that's near and dear to many of us is how `go` won't compile your code if it has unused variables, so if you temporarily comment something out you often have to remove a bunch of variable declarations as well.)

packages/perseus-core/src/parse-perseus-json/perseus-parsers/free-response-widget.ts
rgpass: There are no FRW's (Free-Response Widget) in prod. The feature branch that we're merging into will be the introduction of this widget.

packages/perseus-editor/src/widgets/free-response-editor.tsx
rgpass: > I don't know which is preferred by designers: not showing the button when it can't be used or disabling the button when it can't be used. 🤔

The Radio widget editor hides it. I'll change that.

> Is there a way to allow deleting all criteria in the editor but block the item from being saved if the widget has no criteria? The "no criteria" state could be valuable as a transient state, e.g. if a content creator wants to remove and replace all criteria for a widget.

I'm good with adding that functionality. Is there code I can emulate to do that?

aag: > The Radio widget editor hides it. I'll change that.

I think this is the wrong choice. It makes the software less predictable if actions appear and disappear as you use it. I think it's better for the user if buttons get disabled when they aren't currently available.

I'm open to the possibility of allowing deletion of the last item, but then not allowing saving until another item is added, though I do think it's more confusing. Personally, I would prefer the current experience, because the difference between clicking the "Delete" button and deleting the text out of the last box is not very large, and the current UI prevents the user from getting into an invalid state.

I think a bigger deal is that we allow deletion without confirmation or undo. You might spend several minutes writing a great criteria item, but then accidentally click the "Delete" button and delete it with no way to get it back. The easiest solution is to add a confirmation step (either inline or through a `prompt()` modal).

rgpass: Do we typically consult a designer for widget editors? If not, how do we proceed?

packages/perseus-editor/src/widgets/free-response-editor.tsx
```diff
@@ -32,18 +39,148 @@ class FreeResponseEditor extends React.Component<Props> {
...
+        });
+    };
+
+    renderCriteriaList = () => {
+        return this.props.scoringCriteria.map((criterion, index) => {
+            return (
+                <CriterionEditor
+                    criterion={criterion}
+                    index={index}
+                    key={index}
```
handeyeco, line 78: I get nervous when I see [key set to index](https://react.dev/learn/rendering-lists#why-does-react-need-keys), but I imagine this might be unavoidable because in the editor maybe we can't use `criterion.text`? Because (1) it might be an empty string and (2) we might have multiple empty strings until we fill out the form?

packages/perseus-editor/src/widgets/free-response-editor.tsx
```diff
@@ -32,18 +39,148 @@ class FreeResponseEditor extends React.Component<Props> {
...
+                    numCriteria={this.props.scoringCriteria.length}
```
handeyeco, line 79: Ultra tiny nit but I feel like telling CriterionEditor how many criteria there are is TMI. Really what you're saying is something along the lines of `deletable`.

```suggestion
                    deletable={this.props.scoringCriteria.length <= 1}
```

Go with your gut on this one; feel free to ignore.

handeyeco (approved):
Ultra tiny, non-blocking feedback. LGTM

packages/perseus-editor/src/widgets/free-response-editor.tsx
rgpass: That's right. We can do `${index}-${criterion.text}`, though.

packages/perseus-editor/src/widgets/free-response-editor.tsx
rgpass: > It's an internal tool and the users of it have a direct line to us if they don't like something.

Given this and that we have time before this rolls out, I'm going to merge this as-is. It's a simple enough change to make later.

PR #2287, opened 2025-03-07
[LX] Make static graphs' interactive elements visually disabled

nishasy:
## Summary:
Right now, a static graph looks exactly the same as a regular graph that can be interacted with.

To distinguish graphs as disabled visually when in static mode, update the color so that the
graph elements use Wonder Blocks `offBlack50` when in static mode.

I also did some refactoring to make the way we code colors uniform across all the graphs.

NOTE: You may notice that `offBlack50` is not the best for contrast. However, it needs to be distinguishable
from our regular interactive blue so it's clearly disabled. When using a darker gray, it looks the same as
our WB blue in grayscale. We had to make a call here, and we decided to make it lighter and more
distinguishable even if it sacrifices color contrast.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2784

## Test plan:
`yarn jest packages/perseus/src/widgets/interactive-graphs/interactive-graph-static.test.tsx`

Storybook
- Go to http://localhost:6006/?path=/docs/perseuseditor-widgets-interactive-graph--docs
- For all graph types, use the "Static" toggle at the top of the editor to set the
  graph into static mode.
- Confirm that the interactive elements are no longer blue.

| Interactive | Static |
| --- | --- |
| Screenshot 2025-03-06 at 5 54 08 PM | Screenshot 2025-03-06 at 5 54 14 PM |

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/angle.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/angle.tsx
@@ -45,7 +45,7 @@ export function renderAngleGraph(
 
 function AngleGraph(props: AngleGraphProps) {
     const {dispatch, graphState} = props;
-    const {graphDimensionsInPixels} = useGraphConfig();
+    const {graphDimensionsInPixels, interactiveColor} = useGraphConfig();
     const i18n = usePerseusI18n();
     const id = React.useId();
     const descriptionId = id + "-description";
@@ -83,14 +83,15 @@ function AngleGraph(props: AngleGraphProps) {
                     start={startPtPx}
                     end={endPtPx}
                     style={{
-                        stroke: "var(--movable-line-stroke-color)",
+                        stroke: interactiveColor,
                         strokeWidth: "var(--movable-line-stroke-weight)",
                     }}
+                    testId="angle-graph__line"
                 />
                 <Vector
                     tail={angleLines[i][1]}
                     tip={endExtend}
-                    color={"var(--movable-line-stroke-color)"}
+                    testId="angle-graph__vector"
                 />
             </g>
         );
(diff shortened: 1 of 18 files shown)


packages/perseus/src/widgets/interactive-graphs/graphs/angle.tsx
```diff
@@ -83,14 +83,15 @@ function AngleGraph(props: AngleGraphProps) {
...
-                        stroke: "var(--movable-line-stroke-color)",
+                        stroke: interactiveColor,
                         strokeWidth: "var(--movable-line-stroke-weight)",
                     }}
+                    testId="angle-graph__line"
                 />
                 <Vector
                     tail={angleLines[i][1]}
                     tip={endExtend}
-                    color={"var(--movable-line-stroke-color)"}
```
nishasy, line 93: Since the color defaults to the interactive color inside the Vector component, I removed all the redundant color props being passed in.

packages/perseus/src/widgets/interactive-graphs/graphs/angle.tsx
```diff
@@ -83,14 +83,15 @@ function AngleGraph(props: AngleGraphProps) {
...
+                    testId="angle-graph__vector"
```
nishasy, line 94: I had to add test IDs to find the non-interactive elements within graphs for tests. I favor this over using query selectors.

packages/perseus/src/widgets/interactive-graphs/mafs-styles.css
```diff
@@ -217,12 +216,16 @@
...
+.MafsView .angle-arc-interactive {
     stroke: var(--mafs-blue);
     stroke-width: 0.1px;
 }
 
+.MafsView .angle-arc-static {
+    stroke: var(--static-gray);
+    stroke-width: 0.1px;
+}
+
```
nishasy, line 228: I could not for the life of me figure out how to pass the `stroke` value into the angle paths in the `angle-indictors.tsx` file, so I used CSS classes/styles for that.

Unfortunately, it means this is the only place where the way I handled the color is inconsistent with the rest.

SonicScrewdriver (approved):
Looks great!

PR #2289, opened 2025-03-07
make ServerItemRendererWithDebugUI optionally render answerless data

handeyeco:
## Summary:
<Add your summary here>

Issue: LEMS-2868

## Test plan:

### DIFF
--- a/packages/perseus/src/widgets/expression/expression.stories.tsx
+++ b/packages/perseus/src/widgets/expression/expression.stories.tsx
@@ -23,12 +23,14 @@ type WrappedKeypadContextProps = {
     item: PerseusItem;
     customKeypad: boolean;
     isMobile?: boolean;
+    answerless?: boolean;
 };
 
 const WrappedKeypadContext = ({
     item,
     customKeypad,
     isMobile = false,
+    answerless = false,
 }: WrappedKeypadContextProps) => {
     return (
         <TestKeypadContextWrapper>
@@ -37,6 +39,7 @@ const WrappedKeypadContext = ({
                     return (
                         <ServerItemRendererWithDebugUI
                             item={item}
+                            answerless={answerless}
                             keypadElement={keypadElement}
                             // Hardcoding the V2 Keypad to true as the Storybook Args
                             // were not working.
@@ -118,6 +121,16 @@ export const ExpressionItem3 = (args: StoryArgs): React.ReactElement => {
         <WrappedKeypadContext
             item={expressionItem3}
             customKeypad={args.customKeypad}
+        />
+    );
+};
+
+export const AnswerlessExpression = (args: StoryArgs): React.ReactElement => {
+    return (
+        <WrappedKeypadContext
+            item={expressionItem3}
+            customKeypad={false}
+            answerless
         />
     );
 };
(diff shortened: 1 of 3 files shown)


packages/perseus/src/widgets/expression/expression.testdata.ts
```diff
@@ -4,11 +4,12 @@ import {
     type Version,
     type PerseusItem,
     type PerseusAnswerArea,
+    expressionLogic,
 } from "@khanacademy/perseus-core";
 
 const createItemJson = (
     widgetOptions: PerseusExpressionWidgetOptions,
-    version: Version,
```
Myranae, line 11: It seems like changing version to widgetVersion, adding extra keys, and modifying item 3 below is all that was done here. Did you add an item without answers somewhere? I _think_ the big red and green chunks were just moved code, but I might have missed soemthing.

packages/perseus/src/widgets/expression/expression.stories.tsx
```diff
@@ -122,6 +125,16 @@ export const ExpressionItem3 = (args: StoryArgs): React.ReactElement => {
     );
 };
 
+export const AnswerlessExpression = (args: StoryArgs): React.ReactElement => {
```
jeremywiebe, line 128: Where is this used? I don't see it referenced in this PR at all.

packages/perseus/src/widgets/expression/expression.testdata.ts
```diff
@@ -4,11 +4,12 @@ import {
...
+    widgetVersion: Version = expressionLogic.version as Version,
```
jeremywiebe, line 12: Can we use `satisfies` instead of `as` here? Also, why do we need the cast at all? :)

Myranae:
I'm not sure this is working as we think it is. I could be missing something, but should the Perseus JSON section reflect the result of splitPerseusItem? In other words, should the answerful data be missing from the Perseus JSON area? In my testing, so far that's not the case. I assumed that was because we only need the render data to be missing the answers, but how do we confirm what data it's rendering with? I'm still testing since I'm working on this too, but curious if you have ideas or I missed something in regards to this.

packages/perseus/src/widgets/expression/expression.stories.tsx
handeyeco: This is how Storybook works, but exporting the function in a `.stories.tsx` file it shows up in SB.

Screenshot 2025-03-21 at 9 23 17 AM

packages/perseus/src/widgets/expression/expression.testdata.ts
handeyeco: `expressionLogic` is of type `WidgetLogic` and not all widgets have versions so `version?: Version`

You're right thought, it was overkill. It works with `widgetVersion = expressionLogic.version`

packages/perseus/src/widgets/expression/expression.stories.tsx
jeremywiebe: 🤦‍♂️  I didn't take note this was in a `.stories.tsx` file!

PR #2293, opened 2025-03-10
[Analytics] Removing PerseusErrorOccurred code in favor of PerseusErrorOccurredTI.

catandthemachines:
## Summary:
To add the new "message" metric to the PerseusErrorOccurred event, LEMS team promissed our data team to move from a shared khan metric (that they must manage) to a team internal error event for us to own and control our own event with less overhead.

Issue: LEMS-2826

## Test plan:
Run pnpm test
All tests should pass.


SonicScrewdriver (approved):
Love it! Thank you for cleaning this up :D

nishasy (approved):
LGTM!

jeremywiebe:
This ship has sailed, but I wanted to register that I don't agree with this change and the need for `:ti` events in Perseus. The mapping to CEDAR events happens outside of Perseus and so there's no need to for Perseus to be impacted by these team-specific event requirements: rather, than should happen in any analytics adapter in webapp or whatever app is hosting Perseus. :)

PR #2294, opened 2025-03-10
[Polygon] Show right angle markers when showAngles is on

nishasy:
## Summary:
There was a small bug where the 90 degree angles had the arc instead of the square marker
in the Polygon interactive graph, but only when `showAngles` was on.

Fixed it so that it uses the correct function and shows it appropriately.

Issue: none

## Test plan:
- Start up `yarn dev`
- Look at the polygon question with the square and the angle labels showing
- Confirm that it shows the square indicators now

| Before | After |
| --- | --- |
| Screenshot 2025-03-10 at 4 01 30 PM | Screenshot 2025-03-10 at 4 01 41 PM |

### DIFF
--- a/.changeset/small-gifts-float.md
+++ b/.changeset/small-gifts-float.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+[Polygon] Show right angle markers when showAngles is on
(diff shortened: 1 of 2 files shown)


catandthemachines (approved):
Looks fixed. Thanks Nisha!!

PR #2297, opened 2025-03-12
[Free Response Widget] Add ability to customize placeholder

rgpass:
## Summary:
This PR adds the ability to set the placeholder text in a Free Response Widget component.

Issue: [LIT-1668](https://khanacademy.atlassian.net/browse/LIT-1668)

## Test plan:
- Open Storybook
- See the placeholder can be changed in the editor
- See the placeholder can be changed in the renderer

Screenshot 2025-03-12 at 8 47 49 AM

Screenshot 2025-03-12 at 8 48 04 AM

**NOTE:** The screenshots use something I typed in. The default value is "Please provide response here"

[LIT-1668]: https://khanacademy.atlassian.net/browse/LIT-1668?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/perseus-core/src/widgets/free-response/index.ts
+++ b/packages/perseus-core/src/widgets/free-response/index.ts
@@ -5,12 +5,12 @@ import type {WidgetLogic} from "../logic-export.types";
 
 export type FreeResponseDefaultWidgetOptions = Pick<
     PerseusFreeResponseWidgetOptions,
-    "question" | "scoringCriteria" | "placeholder"
+    "placeholder" | "question" | "scoringCriteria"
 >;
 
 const defaultWidgetOptions: FreeResponseDefaultWidgetOptions = {
-    question: "",
     placeholder: "Please provide response here",
+    question: "",
     // Always display one criterion, since the user will always have to input
     // at least one.
     scoringCriteria: [
--- a/packages/perseus-core/src/widgets/free-response/index.ts
+++ b/packages/perseus-core/src/widgets/free-response/index.ts
@@ -5,12 +5,12 @@ import type {WidgetLogic} from "../logic-export.types";
 
 export type FreeResponseDefaultWidgetOptions = Pick<
     PerseusFreeResponseWidgetOptions,
-    "question" | "scoringCriteria"
+    "question" | "scoringCriteria" | "placeholder"
 >;
 
 const defaultWidgetOptions: FreeResponseDefaultWidgetOptions = {
     question: "",
-
+    placeholder: "Please provide response here",
     // Always display one criterion, since the user will always have to input
     // at least one.
     scoringCriteria: [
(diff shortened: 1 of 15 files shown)


rgpass:
Do I need to add a changeset if merging into a feature branch?

handeyeco:
> Do I need to add a changeset if merging into a feature branch?

@rgpass The important thing is that there's a changeset when landing into main **_and_** that changeset includes all the packages that changed. If a package is changed and it's not included in the changeset, it might not get a new version in `npm` and thus can't be referenced in `package.json`.

We don't use feature branches a lot in Perseus, so that's why the changeset check isn't smarter about looking at your base branch. IMO it's fine to land a PR without a changeset as long as you're not landing into main.

You can also do `pnpm changeset --empty` and it'll make a blank changeset which will quiet the check.

packages/perseus-core/src/widgets/free-response/index.ts
```diff
@@ -5,12 +5,12 @@ import type {WidgetLogic} from "../logic-export.types";
 
 export type FreeResponseDefaultWidgetOptions = Pick<
     PerseusFreeResponseWidgetOptions,
-    "question" | "scoringCriteria"
+    "placeholder" | "question" | "scoringCriteria"
 >;
 
 const defaultWidgetOptions: FreeResponseDefaultWidgetOptions = {
+    placeholder: "Please provide response here",
```
handeyeco, line 12: I think this is probably the right way to do this. My initial gut concern was that this needed to be marked for translation in Perseus (vs CP), but I don't think that's correct.

1. "Please provide response here" will be the default in the editor and content creators can write their own string
2. On publish this string will get sent to CP and their parser will need to know that this is a translatable string
3. It'll get published and we'll receive the translated string when rendering

That seems fine. If we were storing the placeholder as something like `string | null | undefined` in the schema and defaulting to "Please provide response here" at render (vs at publish) then we would need that string in `perseus/src/strings.ts`.

Going this path does allow for less flexibility in the future if we want to change the default string:

1. If we use a default placeholder **when publishing** (what this PR is doing), we have to run a backfill to update existing content
2. If we use a default placeholder **when rendering**, we can change the string without running a backfill since we could just change the string in `perseus/src/string.ts` and request a new translation for that one string

Hmm...now that I've typed all this up I'm wondering if this should change. @benchristel what do you think?

benchristel: I don't know the product context for this feature, so it's hard for me to say what the right choice is. TBH, I don't quite understand why content creators want the ability to customize the placeholder.

In the absence of more information, I would lean toward the current approach of defaulting the placeholder at content-creation time. That way, even if we change the default in the future, learners looking at old content will always see a placeholder that makes sense in context.

rgpass: > That way, even if we change the default in the future, learners looking at old content will always see a placeholder that makes sense in context.

That was our thinking as well.

Given that, are there any changes needed?

benchristel (approved):
LGTM!

PR #2300, opened 2025-03-13
Allow the 'key' field of Radio widgets to be null

benchristel:
This fixes a bug in the parser that caused it to incorrectly reject some production data.

Issue: none

## Test plan:

`yarn test`

After deployment, the number of errors on Sentry mentioning `At
(root).question.widgets["radio 1"].key -- expected number, but got null`
should decrease sharply.

### DIFF
--- a/.changeset/early-swans-brush.md
+++ b/.changeset/early-swans-brush.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-core": patch
+---
+
+Bugfix: allow the 'key' field of Radio widgets to be null when parsing Perseus JSON"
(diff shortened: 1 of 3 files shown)


Myranae (approved):
Thanks for fixing this! Glad we'll have fewer errors :)

handeyeco:
I just realized I made a mistake skipping a test in the parser (I didn't mean to land this): https://github.com/Khan/perseus/pull/2233/files#diff-68d645e728532aa287208fa9587f48a0e0f8fcabd727a85d72f1d2e337b1024eR17

I think it's related to this work?

PR #2301, opened 2025-03-13
Fix number line input outline to meet accessible contrast standards

ivyolamit:
## Summary:
This PR will fix the input outline in number line widget to meet the accessible contrast standards, changing the border color from `#ccc` to `#909296`, which is the same border color of any input e.g. numeric input.

Issue: LEMS-2855

## Test plan:
1. Navigate to the Number Line widget story
2. Select "Show Tick Controller"
- [ ] Confirm that the input border now matches to `#909296` color instead of `#ccc`

## Affected UI:
| Before | After |
| ------ | ----- |
| Screenshot 2025-03-13 at 4 45 59 PM | Screenshot 2025-03-13 at 4 45 24 PM |


catandthemachines (approved):
This all looks good to me @ivyolamit ! 

I do have a non blocking question about his, do you know of any effort to update our less files and/or utilize WonderBlocks stylings more? I think the recommended guidance is to use our defined color palette and avoid using one-off color codes for consistency within the Khan Academy site.

@mark-fitzgerald are you aware of any effort here?

ivyolamit:
> This all looks good to me @ivyolamit !
> 
> I do have a non blocking question about his, do you know of any effort to update our less files and/or utilize WonderBlocks stylings more? I think the recommended guidance is to use our defined color palette and avoid using one-off color codes for consistency within the Khan Academy site.
> 
> @mark-fitzgerald are you aware of any effort here?

Thanks @catandthemachines for the code review.
Regarding the color, that's a great point! Looks like this is a customize color which is patterned from this definition [here](https://github.com/Khan/perseus/blob/0d5ab0b2e5d079b8ac805077a2abad6bc5b2132d/packages/perseus/src/widgets/interactive-graphs/mafs-styles.css#L27) it has a good ℹ️ in it 😅 

I'm not sure how to go about it, but if we are going to use the color code, we can define it in packages/perseus/src/styles/variables.less where all the color codes for .less files are defined.
@mark-fitzgerald any recommendations?

ivyolamit:
@catandthemachines pointing out the WB styling was a good call! Turns out we have the definition in the docs related to the borders [here](https://khan.github.io/wonder-blocks/?path=/docs/foundations-using-color--docs#border). I have updated the PR to reflect it. When you have time, please have another look 🙏🏼

catandthemachines (approved):
Whoohoo love this and the additional attention to adding the WB color. Approved and thanks for the attention to detail @ivyolamit !

PR #2305, opened 2025-03-18
Dropdown story with answerless data

Myranae:
## Summary:
Create a Storybook story for the Dropdown widget that renders with answerless data, is interactive, and is scorable.

Issue: LEMS-2930

## Test plan:
- Confirm all checks pass
- Confirm story renders, is interactive, and can be scored

### DIFF
--- a/packages/perseus-core/src/utils/split-perseus-item.ts
+++ b/packages/perseus-core/src/utils/split-perseus-item.ts
@@ -5,6 +5,10 @@ import {getUpgradedWidgetOptions} from "../widgets/upgrade";
 
 import type {PerseusRenderer} from "../data-schema";
 
+/**
+ * Upgrades widget options and removes answerful data for all the widgets in a
+ * Perseus item.
+ */
 export default function splitPerseusItem(
     originalItem: PerseusRenderer,
 ): PerseusRenderer {
--- a/packages/perseus/src/widgets/dropdown/dropdown.stories.tsx
+++ b/packages/perseus/src/widgets/dropdown/dropdown.stories.tsx
@@ -41,3 +41,7 @@ export const DropdownWithEmptyPlaceholder = (
 ): React.ReactElement => {
     return <RendererWithDebugUI question={dropdownWithEmptyPlaceholder} />;
 };
+
+export const DropdownWithNoAnswers = (args: StoryArgs): React.ReactElement => {
+    return <RendererWithDebugUI question={basicDropdown} answerless={true} />;
+};
--- a/packages/perseus/src/widgets/dropdown/dropdown.stories.tsx
+++ b/packages/perseus/src/widgets/dropdown/dropdown.stories.tsx
@@ -1,3 +1,4 @@
+import {splitPerseusItem} from "@khanacademy/perseus-core";
 import * as React from "react";
 
 import {RendererWithDebugUI} from "../../../../../testing/renderer-with-debug-ui";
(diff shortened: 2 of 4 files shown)


jeremywiebe (approved):

testing/renderer-with-debug-ui.tsx
```diff
@@ -72,13 +79,14 @@ export const RendererWithDebugUI = ({
...
                             ref={ref}
-                            content={question.content}
+                            content={renderedQuestion.content}
                             images={question.images}
                             widgets={question.widgets}
                             problemNum={0}
                             apiOptions={controlledAPIOptions}
                             reviewMode={reviewMode}
                             strings={strings}
+                            answerless={answerless}
```
benchristel, line 89: What does the `answerless` prop do here? I don't see it defined in the `Renderer` prop types.

packages/perseus/src/widgets/dropdown/dropdown.stories.tsx
```diff
@@ -41,3 +41,7 @@ export const DropdownWithEmptyPlaceholder = (
 ): React.ReactElement => {
     return <RendererWithDebugUI question={dropdownWithEmptyPlaceholder} />;
 };
+
+export const DropdownWithNoAnswers = (args: StoryArgs): React.ReactElement => {
+    return <RendererWithDebugUI question={basicDropdown} answerless={true} />;
```
benchristel, line 46: Suggestion: rather than add an `answerless` prop to `RendererWithDebugUI`, I might prefer to call `splitPerseusItem` right here:

```suggestion
    return <RendererWithDebugUI question={splitPerseusItem(basicDropdown)} />;
```

That way, anyone reading the stories doesn't have to guess what `answerless` does (or dig into `RendererWithDebugUI` to figure it out) — they can see exactly what's happening. Granted, it would be clearer if `splitPerseusItem` were called something like `removeAnswers`...

I don't feel very strongly about this suggestion.

testing/renderer-with-debug-ui.tsx
Myranae: Removed!

testing/renderer-with-debug-ui.tsx
```diff
@@ -46,6 +49,10 @@ export const RendererWithDebugUI = ({
         customKeypad: isMobile, // Use the mobile keypad for mobile
     };
 
+    const renderedQuestion: PerseusRenderer = answerless
```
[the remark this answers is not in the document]
Myranae: Removed!

packages/perseus-core/src/utils/split-perseus-item.ts
```diff
@@ -5,6 +5,10 @@ import {getUpgradedWidgetOptions} from "../widgets/upgrade";
 
 import type {PerseusRenderer} from "../data-schema";
 
+/**
+ * Upgrades widget options and removes answerful data for all the widgets in a
+ * Perseus item.
+ */
```
benchristel, line 11: Thanks for adding this comment!

benchristel (approved):
LGTM!

packages/perseus/src/widgets/dropdown/dropdown.stories.tsx
Myranae: Ooo, I think there's a problem doing it this way though. The problem no longer has access to the answer at all. When we use the answerless prop and pass the full question in, the answer is available for when the Check button is clicked. I forgot to confirm the Check button worked correctly. Going to revert back to the previous way, which allows the Check button to work.

testing/renderer-with-debug-ui.tsx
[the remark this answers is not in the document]
Myranae: Using this again, so to explain, we're taking the `question` variable and modifying it for only specific situations, so we don't want to overwrite `question` with its answerless version. We want to render the answerless version and provide the full version for logic.

PR #2307, opened 2025-03-18
Bug fix: Fix Multiselect + Randomization only show correct for proper answers

SonicScrewdriver:
## Summary:
This quick bug-fix temporarily reverts a recent one-line change to fix a P1 bug for the Radio widget. 

### Bug History: 
The issue we were facing was that additional answers were showing as "correct" upon score even though they were incorrect. 

This bug is largely due to logic in the `radio-component.tsx` file on line 423 that uses the rubric to set the correct answer if it is not provided — as is often the case with our older content. 

`correct:  choice.correct === undefined
                            ? !!reviewChoice && !!reviewChoice.correct
                            : choice.correct ` 

**The problem with this logic is that our choices have already been shuffled, but our rubric has not.** As a result of this index mismatch, any questions that have turned on both Multi Selection AND Randomization _could_ result in an additional answer or two being visually styled as if it is correct. 

The bug has always existed, it was merely uncovered after a minor change from the Server Side Scoring project. Previously, we were never hitting the first condition of the ternary as `radio.ts` was explicitly setting `choice.correct` to a boolean value at an earlier stage. We're now hitting this condition after Server Side Scoring removed that line. 

### Temporary Solution:
This PR adds the removed line back to `radio.ts` so that we can solve the bug as quickly as possible — while we work with the Server Side Scoring team to come up with a better long-term solution. 

One _alternative_ proposal to this current PR would be to change the logic from `radio-component.tsx` (outlined above) to:

`correct: Boolean(choice.correct)`

This change is effectively the same as the one proposed this PR. As we were unsure if Server Side Scoring might require the reviewChoice logic in the future, we've opted for a simple revision. We're happy to adjust our approach however! 

Issue: LEMS-2909

## Test plan:
- manual testing
- tests pass

### DIFF
--- a/.changeset/old-pens-hope.md
+++ b/.changeset/old-pens-hope.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Revert SSS temporarily for Radio to ensure Multiselect + Random answer correctness is accurate
--- a/packages/perseus/src/widgets/radio/radio.ts
+++ b/packages/perseus/src/widgets/radio/radio.ts
@@ -75,6 +75,7 @@ const _choiceTransform = (
             return {
                 ...choice,
                 originalIndex: i,
+                correct: Boolean(choice.correct),
             };
         });
 


SonicScrewdriver:
We want to implement some better tests that will take a tiny bit more time than the quick bug fix. Given this is blocking users, we're going to land this now and then add the tests later today with [LEMS-2931](https://khanacademy.atlassian.net/browse/LEMS-2931)

PR #2308, opened 2025-03-19
Numeric Input: Treat unrecognized `simplify` values as "required"

benchristel:
Context: I'm seeing errors in Sentry because the perseus JSON parser doesn't
understand some of the values of `simplify` that exist in production data. This
PR ensures we accept and handle those values.

## Summary:
- I added tests to characterize the behavior of numeric input scoring for all
  the values of `simplify` that are in production.
- I updated the perseus JSON parser to convert nonstandard `simplify` values
  to equivalent ones.

Issue: none

## Test plan:

`yarn test`

### DIFF
--- a/packages/perseus-core/src/parse-perseus-json/perseus-parsers/numeric-input-widget.ts
+++ b/packages/perseus-core/src/parse-perseus-json/perseus-parsers/numeric-input-widget.ts
@@ -32,7 +32,7 @@ const parseMathFormat = enumeration(
     "pi",
 );
 
-const parseSimplify = pipeParsers(
+export const parseSimplify = pipeParsers(
     union(constant(null))
         .or(constant(undefined))
         .or(boolean)
--- a/packages/perseus-core/src/parse-perseus-json/perseus-parsers/numeric-input-widget.ts
+++ b/packages/perseus-core/src/parse-perseus-json/perseus-parsers/numeric-input-widget.ts
@@ -16,7 +16,10 @@ import {defaulted} from "../general-purpose-parsers/defaulted";
 
 import {parseWidget} from "./widget";
 
-import type {NumericInputWidget} from "../../data-schema";
+import type {
+    NumericInputWidget,
+    PerseusNumericInputSimplify,
+} from "../../data-schema";
 import type {Parser} from "../parser-types";
 
 const parseMathFormat = enumeration(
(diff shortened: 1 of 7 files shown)


packages/perseus-core/src/parse-perseus-json/perseus-parsers/numeric-input-widget.ts
```diff
@@ -53,20 +85,7 @@ export const parseNumericInputWidget: Parser<NumericInputWidget> = parseWidget(
                 // TODO(benchristel): simplify should never be a boolean, but we
                 // have some content where it is anyway. If we ever backfill
                 // the data, we should simplify `simplify`.
-                simplify: optional(
-                    nullable(
-                        union(parseSimplify).or(
-                            pipeParsers(boolean).then(
-                                convert((value) => {
-                                    if (typeof value === "boolean") {
-                                        return value ? "required" : "optional";
```
benchristel, line 62: @SonicScrewdriver I changed the behavior here to match what the code was doing in production before this parser existed. It seems like the scoring logic treats both `true` and `false` as `"required"`, so I removed the conditional that converts `false` into `"optional"`. Please let me know if I'm missing something here!

packages/perseus-core/src/parse-perseus-json/perseus-parsers/numeric-input-widget.ts
```diff
@@ -29,12 +32,41 @@ const parseMathFormat = enumeration(
...
+): PerseusNumericInputSimplify {
+    switch (simplify) {
+        case "enforced":
+        case "required":
+        case "optional":
+            return simplify;
+        // NOTE(benchristel): "accepted", "correct", true, false, undefined, and
+        // null are all treated the same as "required" during scoring, so we
+        // convert them to "required" here to preserve behavior. See the tests
+        // in score-numeric-input.test.ts
```
SonicScrewdriver, line 65: Greatly appreciate this comment as it is very helpful to remember all of the silly values allowed for this field.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/numeric-input-widget.ts
SonicScrewdriver: I dug deep into the history for the validator and it seems like this has been this way for longer than 8 years - so I think this is a fairly safe change! 

It feels completely unintuitive for false to not be "optional" — but if that's how it's been, that's how it's been. :P

SonicScrewdriver (approved):

PR #2309, opened 2025-03-19
[SR] Update graph elements so their aria-describedby is read in non-Chrome browsers

nishasy:
## Summary:
Descriptions are not working in Safari or Firefox.

After a number of Slack discussions, it was determined that this was caused by the description blocks being defined within SVGs.

With a small amount of refactoring, we can make is so that the screen reader reads as expected across browsers.

Note: This does not include any changes for the outer graph descriptions, just the interactive elements and their surrounding container inside the graph SVG.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2914

## Test plan:
`yarn jest`

Storybook
- http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-segment
- Go through every graph type in Storybook
- Confirm that the screen reader experience matches what's in the SR tree

### DIFF
--- a/.changeset/early-dodos-wink.md
+++ b/.changeset/early-dodos-wink.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+[SR] Update graph elements so their aria-describedby is read in non-Chrome browsers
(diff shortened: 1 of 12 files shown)


mark-fitzgerald (approved):
I like the generic `<SRDescInSVG>` approach. Thank you!

PR #2313, opened 2025-03-19
[LX] Stop locked functions from memory leaking

nishasy:
## Summary:
It looks like locked functions - specifically locked exponential functions (y = e^x) - cause a memory leak and completely brick the webpage.

One solution is to add bounds behind the scenes. Instead of -Infinity to Infinity, we can clamp the bounds to the range of the graph.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2899

## Test plan:
`yarn jest packages/perseus/src/widgets/interactive-graphs/locked-figures/locked-function.test.ts`

Storybook
- All locked function stories: http://localhost:6006/?path=/docs/perseus-widgets-interactive-graph-locked-functions--docs
- Story that was causing storybook to freeze: http://localhost:6006/?path=/story/perseus-widgets-interactive-graph-locked-functions--exponent

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/locked-figures/locked-function.test.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/locked-figures/locked-function.test.ts
@@ -1,0 +1,25 @@
+import {clampedDomain} from "./locked-function";
+
+describe("clampedDomain", () => {
+    test.each`
+        domain                   | graphXBounds | expected
+        ${[-Infinity, Infinity]} | ${[-10, 10]} | ${[-10, 10]}
+        ${[-20, Infinity]}       | ${[-10, 10]} | ${[-10, 10]}
+        ${[-10, Infinity]}       | ${[-10, 10]} | ${[-10, 10]}
+        ${[-5, Infinity]}        | ${[-10, 10]} | ${[-5, 10]}
+        ${[-Infinity, 20]}       | ${[-10, 10]} | ${[-10, 10]}
+        ${[-Infinity, 10]}       | ${[-10, 10]} | ${[-10, 10]}
+        ${[-Infinity, 5]}        | ${[-10, 10]} | ${[-10, 5]}
+        ${[-Infinity, -5]}       | ${[-10, 10]} | ${[-10, -5]}
+        ${[-Infinity, -10]}      | ${[-10, 10]} | ${[-10, -10]}
+        ${[10, -10]}             | ${[-10, 10]} | ${[-10, 10]}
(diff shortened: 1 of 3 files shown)


packages/perseus/src/widgets/interactive-graphs/locked-figures/locked-function.tsx
```diff
@@ -19,7 +22,7 @@ const LockedFunction = (props: LockedFunctionType) => {
     const plotProps = {
         color: lockedFigureColors[color],
         style: strokeStyle,
-        domain,
+        domain: clampedDomain(domain, range[0]),
```
mark-fitzgerald, line 25: Is this (`range[0]`) still valid when the `directionalAxis` of the graph is along the y-axis?

packages/perseus/src/widgets/interactive-graphs/locked-figures/locked-function.test.ts
```diff
@@ -0,0 +1,25 @@
...
+        ${[-5, Infinity]}        | ${[-10, 10]} | ${[-5, 10]}
+        ${[-Infinity, 20]}       | ${[-10, 10]} | ${[-10, 10]}
+        ${[-Infinity, 10]}       | ${[-10, 10]} | ${[-10, 10]}
+        ${[-Infinity, 5]}        | ${[-10, 10]} | ${[-10, 5]}
+        ${[-Infinity, -5]}       | ${[-10, 10]} | ${[-10, -5]}
+        ${[-Infinity, -10]}      | ${[-10, 10]} | ${[-10, -10]}
+        ${[10, -10]}             | ${[-10, 10]} | ${[-10, 10]}
+        ${[-10, -20]}            | ${[-10, 10]} | ${[-10, 10]}
+        ${[10, 10]}              | ${[-10, 10]} | ${[10, 10]}
+        ${[-10, 10]}             | ${[-10, 10]} | ${[-10, 10]}
```
benchristel, line 18: What happens when the `domain` doesn't overlap the `graphXBounds`? We might need to return some special value like `null` and avoid rendering the locked function at all in that case.

```suggestion
        ${[-10, 10]}             | ${[-10, 10]} | ${[-10, 10]}
        ${[-Infinity, -9]}       | ${[-10, 10]} | ${???}
```

benchristel (approved):
LGTM! Thanks for this fix!

If you're interested, it might be nice to tell Steven (the Mafs maintainer) about this bug. The best place to reach him is [on the Mafs Discord](https://discord.com/invite/mafs).

mark-fitzgerald (approved):
Thanks for researching and resolving this tricky issue.

PR #2314, opened 2025-03-20
Undo Label Location work 

SonicScrewdriver:
## Summary:
Undoing the Label Location work as something seems to have gone strange after deploying to webapp. We will revisit this work separately afterwards. 

This will also require some changes to webapp to remove the new property from the go structs. 

## Test plan:
tests pass


nishasy (approved):
So sorry about this 😞

PR #2321, opened 2025-03-21
Allow `itemDataVersion` to be null when parsing Perseus assessment item JSON

benchristel:
## Summary:
There are a few content items that have `itemDataVersion: null`, and these are
causing parser errors in production because the parser doesn't like `null`.
This PR updates the parser to accept `null` values for `itemDataVersion`.
`itemDataVersion` is already marked deprecated, and can be `undefined`, so it
seems harmless to allow `null` as well.

Issue: none

## Test plan:

`pnpm test`


Myranae (approved):
Makes sense to me! If it can be `undefined` already, then allowing `null` also seems fine. I don't know the current uses for itemDataVersion, maybe some sort of upgrading? Would be nice to confirm any usages still work with null, but I don't _think_ it would really be any different than undefined, so may not be necessary.

packages/perseus-core/src/parse-perseus-json/regression-tests/item-data/item-with-null-itemDataVersion.json
```diff
@@ -0,0 +1,111 @@
...
...23. I gulped. I’d expected we’d start slow, maybe with some sprints. \n\n24. *You’ve got this,* I coached myself as I walked toward the center line. *You’re ready. Even Malik said so.* \n\n25. When I arrived at the center line and saw the three boys between me and the end zone, a sense of calm settled over me. Strategy flooded my mind as I thought of everything I’d observed from the bushes. Marcus, on the right, was big but slow. Raj, in the center, had a tendency to cover the left side of his area. Hiro, on the left, was the biggest threat–he was fast and clever, and, as he smirked at me, I knew immediately what my best play would be.\n\n26. As Coach blew his whistle, I raced hard towards the left, ignoring the obvious opening on the right side of the field. No one who knew him ever tried to get past Hiro, so that’s exactly what I would do. I barely saw Hiro’s smirk change into a look of surprise as I approached him, sliding under his outstretched arms. I leapt into the end zone with a whoop of delight. \n\n27. Coach looked reluctantly impressed when I trotted back to him. “Well, that wasn’t half-bad. I think we might give you a shot after all.”\n\n[[☃ explanation 1]]\n\n\n=====",
```
handeyeco, line 3: I don't really love how much noise we get from using full content. It feels a little counter-productive because I'm not sure what parts of this are actually important to the PR. Maybe it'll make more sense when we start making changes to the snapshots, but right now it just feels like a distraction.

handeyeco (approved):
It's cool you're finding these things. I wonder if it's worth filing tickets as we run into things like this that could likely just be removed (or maybe add it to the Confluence page for things to address during a backfill).

PR #2322, opened 2025-03-21
Change version injection to be back in a package

jeremywiebe:
## Summary:

A while ago, we decided it would be useful to inject the built version of each package into the package itself. This getrs surfaced on the `__perseus_debug__` object attached to the `window` (`globalThis`) object at runtime. This is handy!

Originally we implemented it in `@khanacademy/perseus-core` as that was a small, no-deps package. Over time, we've started moving more and more functionality into perseus-core (especially as part of server-side scoring). This ended up causing a circular dependency. 

To solve this, we moved the injection code inot the `/utils` folder and symlinking it into each package. That worked... until we needed to enable the Typescript parser for eslint (see: https://github.com/typescript-eslint/typescript-eslint/issues/2987). 

So this PR changes things once more. I've re-introduced a package which we can use for small, focused utilities. I know utils can sometimes become dumping grounds, but we can also be disciplined. :) 

Issue: --none--

## Test plan:

Run `pnpm build`

Build size check is stable (in fact, the bundles shrink a bit because things aren't integrated/bundled until the consuming app bundles things now).


config/build/rollup.config.js
```diff
@@ -170,8 +174,15 @@ const createConfig = (
                 browser: platform === "browser",
                 extensions,
             }),
-            autoExternal({
+            nodeExternals({
```
jeremywiebe, line 177: Switched to this plugin as this is was wonder-blocks uses, its actively maintained (autoExternals hadn't been touched in 5 years), and it allows us to selectively force bundling.

packages/perseus-linter/package.json
```diff
@@ -29,6 +29,8 @@
     },
     "devDependencies": {
         "@khanacademy/pure-markdown": "workspace:*",
+        "perseus-build-settings": "workspace:*",
```
jeremywiebe, line 32: This was missing the build-settings dep, which it should have!

handeyeco (commented):
Thanks for looking into this! Couple of thoughts:

1. The bundle sizes seemed to get bigger (`perseus-editor` grew 3MB?) which makes me think there's a bug here
2. I worry about the long-term maintainability of this; it feels like we're doing back flips trying to avoid copy/pasting ~50 lines of code
3. Are people using this feature? Do people know about this feature?
4. What if one of us just made an `npm` package that handled this for us; it could be an `index.js` file and a `index.test.js` file living in `npm` that we could depend on. Could that help us reduce the complexity in Perseus?

Open to push-back and will approve after you double-check the bundle size alerts, but I'm feeling like this feature is drifting from the KISS principle.

config/build/rollup.config.js
jeremywiebe: Reverted this now that we're just making this a regular old package.

tatianasnook:
How does moving this logic to a package fix the ESLint issue?

packages/perseus-score/package.json
```diff
@@ -27,7 +27,9 @@
     "dependencies": {
         "@khanacademy/kas": "workspace:*",
         "@khanacademy/kmath": "workspace:*",
-        "@khanacademy/perseus-core": "workspace:*"
+        "@khanacademy/perseus-core": "workspace:*",
+        "@khanacademy/perseus-utils": "workspace:*",
+        "perseus-build-settings": "workspace:*"
```
handeyeco, line 32: I noticed that sometimes this is a dev dep and here it's a regular dep. Is one preferred over the other?

packages/perseus-utils/README.md
```diff
@@ -0,0 +1,4 @@
+# @khanacademy/perseus-utils
+
+This is a container package for utilities that are used by all, or at least
+most Perseus packages.
```
handeyeco, line 4: Worth mentioning that it can't depend on other Perseus packages?

packages/perseus-utils/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "publishConfig": {
+        "access": "public"
+    },
+    "repository": {
+        "type": "git",
+        "url": "https://github.com/Khan/perseus.git",
+        "directory": "packages/perseus-utils"
+    },
+    "bugs": {
+        "url": "https://github.com/Khan/perseus/issues"
```
handeyeco, line 16: Random thought: do we even have the issues tab enabled?

packages/perseus-score/package.json
jeremywiebe: `purseus-build-settings` should be a dev dep!  I'll review and fix where it's wrong.

jeremywiebe: I also caught one of `perseus-utils` in the wrong spot and fixed that too!

packages/perseus-utils/package.json
jeremywiebe: We do! Not sure that's good or bad. I'll leave this here for now. :)

jeremywiebe:
> How does moving this logic to a package fix the ESLint issue?

Previously, this shared logic was included in each package's source tree as a symlinked folder. ([bug report](https://github.com/typescript-eslint/typescript-eslint/issues/2234))

This PR changes things so that the shared code is in a regular npm package that all of the other packages use.

PR #2324, opened 2025-03-25
Adding button roles to the Image Widget to help improve A11Y.

SonicScrewdriver:
## Summary:
Our A11Y Partner has requested that the button role be added to our Image Widget, as images within it may be interactive if they require a zoom. 

This PR adds the role, fixes a related story for testing purposes, and updates several related tests. 

Issue: LEMS-2697

## Test plan:
- Manual testing on Desktop/Mobile 
- Tests Pass

### DIFF
--- a/packages/perseus/src/__tests__/renderer.test.tsx
+++ b/packages/perseus/src/__tests__/renderer.test.tsx
@@ -409,7 +409,9 @@ describe("renderer", () => {
             markImagesAsLoaded();
 
             // Assert
-            const imageNodes = screen.queryAllByRole("img");
+            const imageNodes = screen.queryAllByAltText(
+                "This image has dimensions",
+            );
             expect(imageNodes).toHaveLength(1);
 
             // eslint-disable-next-line testing-library/no-node-access
@@ -433,7 +435,9 @@ describe("renderer", () => {
             markImagesAsLoaded();
 
             // Assert
-            const imageNodes = screen.queryAllByRole("img");
+            const imageNodes = screen.queryAllByAltText(
+                "This image doesn't have dimensions",
+            );
             expect(imageNodes).toHaveLength(1);
 
             // image didn't have dimensions provided so it is not rendered as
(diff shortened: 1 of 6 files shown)


packages/perseus/src/__tests__/renderer.test.tsx
```diff
@@ -409,7 +409,9 @@ describe("renderer", () => {
             markImagesAsLoaded();
 
             // Assert
-            const imageNodes = screen.queryAllByRole("img");
+            const imageNodes = screen.queryAllByAltText(
```
SonicScrewdriver, line 412: Sadly we can't query for img roles anymore, so I've updated all of these to use the appropriate alttext.

packages/perseus/src/components/image-loader.tsx
```diff
@@ -142,11 +142,18 @@ class ImageLoader extends React.Component<Props, State> {
         }
         const staticUrl = getDependencies().staticUrl;
 
+        // If the image is interactive, it should have a role of "button"
+        // to indicate that it is interactive.
+        const imageRole = imgProps.onClick !== null ? "button" : "img";
```
SonicScrewdriver, line 147: I'm not positive that the onClick is ever null, but it seemed good to have this check in case there is a situation where the image ISN'T interactive. This matches the null check in the onClick logic further up in the file.

packages/perseus/src/components/image-loader.tsx
```diff
@@ -142,11 +142,18 @@ class ImageLoader extends React.Component<Props, State> {
...
+
         return (
-            // eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions -- TODO(LEMS-2871): Address a11y error
+            // (LEMS-2871) NOTE: This image IS interactive if it requires zooming, and LevelAccess specifically requested
+            // that the role and tabIndex be set here. When users interact with the image on mobile, it will zoom in.
+            // eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions
```
SonicScrewdriver, line 152: We have a ticket to fix this A11Y exceptions, but this is one case that I think we're going to want to keep. It did make me wonder about whether this linter rule could be updated as we're having to write a lot of exceptions for this interactive element.

mark-fitzgerald (approved):

PR #2326, opened 2025-03-25
[SR] Remove instructions for static graphs, mark interactive elements as disabled

nishasy:
## Summary:
Right now, when an interactive graph is not interactive (static graphs or after the question has been answered), there is no way for a screen
reader user to know that. The graph instructions are still there, and
the interactive elements sound the same.

Fixing that in this PR
- When a graph is not interactive, remove the interaction instructions
- When a graph is not interactive, give all interactive elements the
  "aria-disabled={true}" property.

| Before | After |
| --- | --- |
| Screenshot 2025-03-25 at 12 50 14 PM | Screenshot 2025-03-25 at 12 49 34 PM |

Issue: none

## Test plan:
`yarn jest packages/perseus/src/widgets/interactive-graphs/interactive-graph-static.test.tsx`

Storybook
- https://khan.github.io/perseus/?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-segment
- For each graph type, go through and make sure everything remains as expected by default
- Set the graph to static
- For each graph type, go through and make sure all interactive elements read as "dimmed"

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point-view.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point-view.tsx
@@ -1,4 +1,3 @@
-import {color as WBColor} from "@khanacademy/wonder-blocks-tokens";
 import Tooltip from "@khanacademy/wonder-blocks-tooltip";
 import * as React from "react";
 import {forwardRef} from "react";
@@ -15,7 +14,6 @@ import type {ForwardedRef} from "react";
 
 type Props = {
     point: vec.Vector2;
-    color?: string | undefined;
     dragging: boolean;
     focused: boolean;
     showFocusRing: boolean;
@@ -32,10 +30,14 @@ const hitboxSizePx = 48;
 // on an interactive graph.
 export const MovablePointView = forwardRef(
     (props: Props, hitboxRef: ForwardedRef<SVGGElement>) => {
-        const {markings, showTooltips, interactiveColor} = useGraphConfig();
+        const {
+            markings,
+            showTooltips,
+            interactiveColor,
+            disableKeyboardInteraction,
+        } = useGraphConfig();
         const {
             point,
-            color = interactiveColor,
             dragging,
             focused,
             cursor,
@@ -43,11 +45,15 @@ export const MovablePointView = forwardRef(
             onClick = () => {},
         } = props;
 
-        // WB Tooltip requires a color name for the background color.
-        // Since the color in props is a hex value, a reverse lookup is needed.
-        const wbColorName = (Object.entries(WBColor).find(
-            ([_, value]) => value === color,
-        )?.[0] ?? "blue") as keyof typeof WBColor;
+        // WB Tooltip requires a WB color name for the background color.
+        // We use the blue color when the points are interactive, and
+        // offBlack64 when they are disabled.
+        // Note: The disabled point color is offBlack50 to contrast with
+        // the interactive blue color, but the tooltip background has to be
+        // darker to contrast with its white text - using offBlack64.
+        const wbColorName = disableKeyboardInteraction
+            ? "fadedOffBlack64" // not see-through
+            : "blue";
 
         const pointClasses = classNames(
             "movable-point",
@@ -68,7 +74,9 @@ export const MovablePointView = forwardRef(
                 aria-hidden={true}
                 ref={hitboxRef}
                 className={pointClasses}
-                style={{"--movable-point-color": color, cursor} as any}
+                style={
+                    {"--movable-point-color": interactiveColor, cursor} as any
+                }
                 data-testid="movable-point"
                 onClick={onClick}
             >
@@ -85,7 +93,7 @@ export const MovablePointView = forwardRef(
                     className="movable-point-center"
                     cx={x}
                     cy={y}
-                    style={{fill: color}}
+                    style={{fill: interactiveColor}}
                     data-testid="movable-point__center"
                 />
             </g>
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.test.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.test.tsx
@@ -106,7 +106,7 @@ describe("MovablePoint", () => {
             expect(Tooltip).not.toHaveBeenCalled();
         });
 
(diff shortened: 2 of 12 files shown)


packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point-view.tsx
```diff
@@ -15,7 +14,6 @@ import type {ForwardedRef} from "react";
 
 type Props = {
     point: vec.Vector2;
-    color?: string | undefined;
```
nishasy, line 18: Removing this prop because it's not used by anything.

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point-view.tsx
```diff
@@ -32,22 +30,30 @@ const hitboxSizePx = 48;
...
             cursor,
             showFocusRing,
             onClick = () => {},
         } = props;
 
-        // WB Tooltip requires a color name for the background color.
-        // Since the color in props is a hex value, a reverse lookup is needed.
-        const wbColorName = (Object.entries(WBColor).find(
-            ([_, value]) => value === color,
-        )?.[0] ?? "blue") as keyof typeof WBColor;
```
nishasy, line 50: Since MovablePoint is never any color except blue or gray, I removed this backwards search and made it more explicit.

packages/perseus/src/widgets/interactive-graphs/graphs/components/use-control-point.tsx
```diff
@@ -53,7 +53,6 @@ export function useControlPoint(params: Params): Return {
         ariaDescribedBy,
         ariaLabel,
         ariaLive = "polite",
-        color,
```
nishasy, line 56: Removed this because the prop is not being used anywhere. All interactive colors are coming from the graph config now.

packages/perseus/src/widgets/interactive-graphs/mafs-styles.css
```diff
@@ -34,7 +34,6 @@
     --movable-point-hover-expansion: 2px;
     --movable-point-focus-ring-offset: 2px;
 
-    --movable-line-stroke-color: var(--mafs-blue);
```
nishasy, line 37: This wasn't being used anywhere.

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.test.tsx
catandthemachines, line 1: Just to confirm we're certain that we don't have any experiences where we want to override the color correct?

packages/perseus/src/widgets/interactive-graphs/interactive-graph-static.test.tsx
catandthemachines, line 1: Just to confirm, `aria-disabled` is only ment to be set to `true` when the graph is non-interactive. When it is interactive (a normal graph) they should be `aria-disabled=false` right?

packages/perseus/src/widgets/interactive-graphs/interactive-graph-static.test.tsx
```diff
@@ -313,4 +326,48 @@ describe.each`
...
+            // Act
+            renderQuestion(question, blankOptions);
+            const interactiveElements = screen.getAllByRole("button");
+
+            // Assert
+            expect(interactiveElements).toHaveLength(expectedNumOfElements);
+            for (const element of interactiveElements) {
+                expect(element).toHaveAttribute(
+                    "aria-disabled",
+                    `${staticMode}`, // aria-disabled is a string
```
catandthemachines, line 368: Same here is it meant to be set to true? False? Just want to make sure this is clear in the test.

catandthemachines (approved):
The code looks nice an clean. The only remaining question I have is just clarifying when is `aria-disabled` meant to be false or true.

mark-fitzgerald (approved):

packages/perseus/src/widgets/interactive-graphs/interactive-graph-static.test.tsx
nishasy: That's correct. 

You can see that being explicitly set in the context provider in mafs-graph.tsx:

https://github.com/Khan/perseus/blob/main/packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx#L154

https://github.com/Khan/perseus/blob/main/packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx#L170

When you see me referring to `disableKeyboardInteraction` from `userGraphConfig()` in the code, it means it's that value, which is explicitly true or false.

packages/perseus/src/widgets/interactive-graphs/interactive-graph-static.test.tsx
nishasy: Yes. This test is inside a `describe.each` that has a `staticMode` value of true and then false.

packages/perseus/src/widgets/interactive-graphs/interactive-graph-static.test.tsx
```diff
@@ -10,6 +10,19 @@ import type {PerseusRenderer} from "@khanacademy/perseus-core";
...
+    circle: interactiveGraphQuestionBuilder().withCircle(),
+    quadratic: interactiveGraphQuestionBuilder().withQuadratic(),
+    sinusoid: interactiveGraphQuestionBuilder().withSinusoid(),
+    polygon: interactiveGraphQuestionBuilder().withPolygon(),
+    angle: interactiveGraphQuestionBuilder().withAngle(),
+};
+
 describe.each`
     type             | staticMode | expectedColor
     ${"Interactive"} | ${false}   | ${"var(--mafs-blue)"}
```
nishasy, line 28: @catandthemachines This is where `staticMode` is explicitly set to true/false.

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.test.tsx
nishasy: I'm certain, since MovablePoint is an interactive element, and all interactive elements on Khan are blue.

If we do ever want to change the color (like maybe Khanmigo theme or something for example), I think it should be done via the `interctiveColor` value from the config, rather than passing color into each individual interactive element.

PR #2327, opened 2025-03-25
Upgrade Rollup and all plugins and libs to latest

jeremywiebe:
## Summary:

Adapt to be compatible with Rollup v4. Note that there are some build size changes, but they won't be terribly relevant once https://github.com/Khan/perseus/pull/2331 is landed (which is based on this PR's branch). When we only generate ES Modules, consuming bundlers can tree-shake away unused code and so our Perseus packages' bundle sizes will be less relevant.

Issue: LEMS-2954

## Test plan:


config/build/rollup.config.js
```diff
@@ -55,6 +55,12 @@ const createOutputConfig = (pkgName, format, targetFile) => ({
     sourcemap: true,
     format,
 
+    // These two settings are to keep the builds as similar to pre-Rollup v4 as
+    // possible until we get rid of CJS builds.
+    // See: https://rollupjs.org/migration/#changed-defaults
+    esModule: true,
+    interop: "compat",
```
handeyeco, line 62: So we can get rid of this after https://github.com/Khan/perseus/pull/2331 lands?

handeyeco (approved):
Thanks!

config/build/rollup.config.js
jeremywiebe: Yes, and you can verify its deleted in the PR here: https://github.com/Khan/perseus/pull/2331/files#diff-45f213140a165c743b42e38a00d97984fcec3f841195e8587d03a861d402ca17R63-R75

PR #2328, opened 2025-03-26
Fix self-imports and eslint rule for type imports

jeremywiebe:
## Summary:

I noticed that we have quite a few files that are importing symbols from the package their in using the package name instead of a relative import. This is confusing when looking at the source and will also cause issues in future versions of Rollup (our build tool).

Issue: "none"

## Test plan:

`yarn lint`
`yarn build`
`yarn typecheck`

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -197,7 +197,6 @@ module.exports = {
         "max-lines": "off",
         "new-cap": "off",
         "no-invalid-this": "off",
-        "@typescript-eslint/no-this-alias": "off",
         "no-unused-expressions": "off",
         "no-restricted-imports": [
             "error",
@@ -388,6 +387,10 @@ module.exports = {
          * typescript
          */
         "@typescript-eslint/no-empty-function": "off",
-        "@typescript-eslint/consistent-type-imports": "error",
+        "@typescript-eslint/no-this-alias": "off",
+        "@typescript-eslint/consistent-type-imports": [
+            "error",
+            {prefer: "type-imports"},
+        ],
     },
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -390,7 +390,10 @@ module.exports = {
         "@typescript-eslint/no-this-alias": "off",
         "@typescript-eslint/consistent-type-imports": [
             "error",
-            {prefer: "type-imports"},
+            {
+                prefer: "type-imports",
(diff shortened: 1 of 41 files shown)


.eslintrc.js
```diff
@@ -197,7 +197,6 @@ module.exports = {
         "max-lines": "off",
         "new-cap": "off",
         "no-invalid-this": "off",
-        "@typescript-eslint/no-this-alias": "off",
```
jeremywiebe, line 200: Moved this down to be with the other `@typescript-eslint` rules.

.eslintrc.js
```diff
@@ -388,6 +387,13 @@ module.exports = {
...
         "@typescript-eslint/no-empty-function": "off",
-        "@typescript-eslint/consistent-type-imports": "error",
+        "@typescript-eslint/no-this-alias": "off",
+        "@typescript-eslint/consistent-type-imports": [
+            "error",
+            {
+                prefer: "type-imports",
+                fixStyle: "separate-type-imports",
+            },
+        ],
```
jeremywiebe, line 397: We've been discussing how it seems like this rule hasn't been working lately. I _think_ adding some configs has helped for this PR so we can see if this makes it work again (better).

packages/kmath/src/geometry.ts
```diff
@@ -9,7 +9,9 @@ import {
 } from "@khanacademy/perseus-core";
 import _ from "underscore";
 
-import {number as knumber, point as kpoint, sum} from "@khanacademy/kmath";
```
jeremywiebe, line 12: We're _in_ the kmath package!

packages/perseus/src/widgets/interactive-graphs/graphs/angle.test.tsx
```diff
@@ -1,9 +1,8 @@
 import {render, screen} from "@testing-library/react";
 import * as React from "react";
 
-import {Dependencies} from "@khanacademy/perseus";
-
 import {testDependencies} from "../../../../../../testing/test-dependencies";
+import * as Dependencies from "../../../dependencies";
```
jeremywiebe, line 5: This is a `*` import because that's how we export it in `packages/perseus/src/index.ts`.

benchristel (approved):
Thanks for fixing these! LGTM.

PR #2329, opened 2025-03-26
Create stories for Assessment's 5 scorable widgets that use answerless data

Myranae:
## Summary:
In order to confirm Assessment's main widgets can render with answerless data and then still score after that, we are adding stories for each of the widgets that render with answer less data.

Issue: LEMS-2869

## Test plan:
- Confirm all answerless stories start out with answerless data and then switch to answerful data in order to score. 
- Confirm the widgets in the stories score correctly.
- Confirm behavior has no regressios and is as expected.
- Confirm all checks pass.

### DIFF
--- a/.changeset/ninety-wasps-love.md
+++ b/.changeset/ninety-wasps-love.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": minor
+---
+
+Update and add stories where Dropdown, Interactive Graph, Numeric Input, Expression, and Radio widgets render using answerless data
--- a/packages/perseus/src/server-item-renderer.tsx
+++ b/packages/perseus/src/server-item-renderer.tsx
@@ -460,8 +460,6 @@ export class ServerItemRenderer
             />
         );
 
-        console.log("item question sent to renderer", this.props.item.question);
-
         return (
             <DependenciesContext.Provider value={this.props.dependencies}>
                 <div>
--- a/packages/perseus/src/server-item-renderer.tsx
+++ b/packages/perseus/src/server-item-renderer.tsx
@@ -460,6 +460,8 @@ export class ServerItemRenderer
             />
         );
 
+        console.log("item question sent to renderer", this.props.item.question);
+
         return (
             <DependenciesContext.Provider value={this.props.dependencies}>
                 <div>
--- a/packages/perseus/src/widgets/dropdown/dropdown.stories.tsx
+++ b/packages/perseus/src/widgets/dropdown/dropdown.stories.tsx
@@ -44,5 +44,5 @@ export const DropdownWithEmptyPlaceholder = (
 };
 
 export const DropdownWithNoAnswers = (args: StoryArgs): React.ReactElement => {
-    return <RendererWithDebugUI question={splitPerseusItem(basicDropdown)} />;
+    return <RendererWithDebugUI question={basicDropdown} answerless={true} />;
 };
--- a/packages/perseus/src/widgets/dropdown/dropdown.stories.tsx
+++ b/packages/perseus/src/widgets/dropdown/dropdown.stories.tsx
@@ -1,14 +1,13 @@
-import {splitPerseusItem} from "@khanacademy/perseus-core";
 import * as React from "react";
 
-import {RendererWithDebugUI} from "../../../../../testing/renderer-with-debug-ui";
+import {ServerItemRendererWithDebugUI} from "../../../../../testing/server-item-renderer-with-debug-ui";
 
 import {
-    basicDropdown,
-    dropdownWithEmptyPlaceholder,
-    dropdownWithMath,
-    dropdownWithVisibleLabel,
-    inlineDropdownWithVisibleLabel,
+    basicDropdownItem,
+    dropdownWithEmptyPlaceholderItem,
+    dropdownWithMathItem,
+    dropdownWithVisibleLabelItem,
+    inlineDropdownWithVisibleLabelItem,
 } from "./dropdown.testdata";
 
 export default {
@@ -18,31 +17,92 @@ export default {
 type StoryArgs = Record<any, any>;
 
 export const BasicDropdown = (args: StoryArgs): React.ReactElement => {
-    return <RendererWithDebugUI question={basicDropdown} />;
+    return <ServerItemRendererWithDebugUI item={basicDropdownItem} />;
 };
 
 export const DropdownWithMath = (args: StoryArgs): React.ReactElement => {
-    return <RendererWithDebugUI question={dropdownWithMath} />;
+    return <ServerItemRendererWithDebugUI item={dropdownWithMathItem} />;
 };
 
 export const DropdownWithVisibleLabel = (
     args: StoryArgs,
 ): React.ReactElement => {
-    return <RendererWithDebugUI question={dropdownWithVisibleLabel} />;
+    return (
+        <ServerItemRendererWithDebugUI item={dropdownWithVisibleLabelItem} />
+    );
 };
 
 export const InlineDropdownWithVisibleLabel = (
     args: StoryArgs,
 ): React.ReactElement => {
-    return <RendererWithDebugUI question={inlineDropdownWithVisibleLabel} />;
+    return (
+        <ServerItemRendererWithDebugUI
+            item={inlineDropdownWithVisibleLabelItem}
+        />
+    );
 };
 
 export const DropdownWithEmptyPlaceholder = (
     args: StoryArgs,
 ): React.ReactElement => {
-    return <RendererWithDebugUI question={dropdownWithEmptyPlaceholder} />;
+    return (
+        <ServerItemRendererWithDebugUI
+            item={dropdownWithEmptyPlaceholderItem}
+        />
+    );
 };
 
-export const DropdownWithNoAnswers = (args: StoryArgs): React.ReactElement => {
-    return <RendererWithDebugUI question={basicDropdown} answerless={true} />;
+export const AnswerlessBasicDropdown = (
+    args: StoryArgs,
+): React.ReactElement => {
+    return (
+        <ServerItemRendererWithDebugUI
+            item={basicDropdownItem}
+            startAnswerless={true}
+        />
+    );
+};
+
+export const AnswerlessDropdownWithMath = (
+    args: StoryArgs,
+): React.ReactElement => {
+    return (
+        <ServerItemRendererWithDebugUI
+            item={dropdownWithMathItem}
+            startAnswerless={true}
+        />
+    );
+};
+
+export const AnswerlessDropdownWithVisibleLabel = (
+    args: StoryArgs,
+): React.ReactElement => {
+    return (
+        <ServerItemRendererWithDebugUI
+            item={dropdownWithVisibleLabelItem}
+            startAnswerless={true}
+        />
+    );
+};
+
+export const AnswerlessInlineDropdownWithVisibleLabel = (
+    args: StoryArgs,
+): React.ReactElement => {
+    return (
+        <ServerItemRendererWithDebugUI
+            item={inlineDropdownWithVisibleLabelItem}
+            startAnswerless={true}
+        />
+    );
+};
+
+export const AnswerlessDropdownWithEmptyPlaceholder = (
+    args: StoryArgs,
+): React.ReactElement => {
+    return (
+        <ServerItemRendererWithDebugUI
+            item={dropdownWithEmptyPlaceholderItem}
+            startAnswerless={true}
+        />
+    );
 };
--- a/packages/perseus/src/widgets/interactive-graphs/interactive-graph.testdata.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/interactive-graph.testdata.ts
@@ -2,9 +2,36 @@ import {interactiveGraphQuestionBuilder} from "./interactive-graph-question-buil
 
 import type {LockedFunctionOptions} from "./interactive-graph-question-builder";
 import type {Coord} from "../../interactive2/types";
-import type {PerseusRenderer, RadioWidget} from "@khanacademy/perseus-core";
+import type {
+    PerseusItem,
+    PerseusRenderer,
+    RadioWidget,
+} from "@khanacademy/perseus-core";
 
 // Data for the interactive graph widget
+
+const createInteractiveGraphItem = (question: PerseusRenderer): PerseusItem => {
+    return {
+        question: question,
+        answer: null,
+        itemDataVersion: {
+            major: 0,
+            minor: 1,
+        },
+        hints: [],
+        answerArea: {
+            calculator: false,
+            chi2Table: false,
+            financialCalculatorMonthlyPayment: false,
+            financialCalculatorTotalAmount: false,
+            financialCalculatorTimeToPayOff: false,
+            periodicTable: false,
+            periodicTableWithKey: false,
+            tTable: false,
+            zTable: false,
+        },
+    };
+};
 
 export const angleQuestion: PerseusRenderer = interactiveGraphQuestionBuilder()
     .withContent(
@@ -29,6 +56,8 @@ export const angleQuestion: PerseusRenderer = interactiveGraphQuestionBuilder()
     })
     .build();
 
+export const angleItem: PerseusItem = createInteractiveGraphItem(angleQuestion);
+
 export const angleQuestionWithDefaultCorrect: PerseusRenderer =
     interactiveGraphQuestionBuilder().withAngle().build();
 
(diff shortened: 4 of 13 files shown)


packages/perseus/src/widgets/radio/__tests__/radio.testdata.ts
```diff
@@ -205,16 +234,16 @@ export const choicesWithImages: PerseusRenderer = {
...
 export const multiChoiceQuestion: PerseusRenderer = {
     content:
         "**Select all input values for which $g(x)=2$.**\n\n[[\u2603 radio 1]]\n\n ![](web+graphie://ka-perseus-graphie.s3.amazonaws.com/4613e0d9c906b3053fb5523eed83d4f779fdf6bb)",
-    images: {
-        "web+graphie://ka-perseus-graphie.s3.amazonaws.com/4613e0d9c906b3053fb5523eed83d4f779fdf6bb":
-            {
-                width: 428,
-                height: 428,
-            },
-    },
```
Myranae, line 217: This content was removed because the image is already in the content section and having an image here seems to cause a style issue. This is referenced in the investigation ticket [LEMS-2948](https://khanacademy.atlassian.net/browse/LEMS-2948). Changing this caused an update to the snapshots as well.

jeremywiebe (approved):
Thanks for adding these. I have a few comments, but nothing major. :)

.changeset/ninety-wasps-love.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/perseus": minor
+---
+
+Update and add stories where Dropdown, Interactive Graph, Numeric Input, Expression, and Radio widgets render using answerless data
```
handeyeco, line 5: For future PRs, it would be beneficial to have separate PRs. In this case it could have been 5 PRs.

packages/perseus/src/widgets/__testutils__/test-data-util.ts
```diff
@@ -0,0 +1,32 @@
...
+            financialCalculatorMonthlyPayment: false,
+            financialCalculatorTotalAmount: false,
+            financialCalculatorTimeToPayOff: false,
+            periodicTable: false,
+            periodicTableWithKey: false,
+            tTable: false,
+            zTable: false,
+        },
+    };
+};
```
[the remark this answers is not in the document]
handeyeco: We also have `generateTestPerseusItem` in the `perseus` package. LEMS needs to devote a sprint or two to combining our test utils.

Also might be worth looking into how Ned handled `answerArea`:

``` TS
    answerArea: Object.fromEntries(
        ItemExtras.map((extra) => [extra, false]),
    ) as PerseusAnswerArea,
```

packages/perseus/src/widgets/dropdown/dropdown.stories.tsx
```diff
@@ -43,6 +48,57 @@ export const DropdownWithEmptyPlaceholder = (
...
 };
 
-export const DropdownWithNoAnswers = (args: StoryArgs): React.ReactElement => {
-    return <RendererWithDebugUI question={splitPerseusItem(basicDropdown)} />;
+export const AnswerlessBasicDropdown = (
+    args: StoryArgs,
+): React.ReactElement => {
+    return (
+        <ServerItemRendererWithDebugUI
+            item={basicDropdownItem}
```
handeyeco, line 56: I think it's a little overkill to duplicate each story (really we just needed one interactive answerless story per widget). Having a bunch of stories that are nearly identical might be confusing to people especially in a couple of months when all stories should start answerless. Not too worried about this, but maybe we need a post-SSS ticket to come back and simplify the stories.

What I think is definitely overkill is exporting/importing what is essentially the same test data. Rather than redeclare the test data, just wrap the existing test data.

```suggestion
            item={createPerseusItem(basicDropdownItem)}
```

packages/perseus/src/widgets/interactive-graphs/interactive-graph.testdata.ts
```diff
@@ -29,6 +35,8 @@ export const angleQuestion: PerseusRenderer = interactiveGraphQuestionBuilder()
     })
     .build();
 
+export const angleItem: PerseusItem = createPerseusItem(angleQuestion);
```
handeyeco, line 38: Same here, we don't need to be redeclaring a bunch of test data when we could just be wrapping the existing test data in `createPerseusItem`

packages/perseus/src/widgets/radio/__stories__/answerlessRadio.stories.tsx
```diff
@@ -0,0 +1,90 @@
+// NOTE(Tamara): ServerItemRenderWithDebugUI does not work with the same props
```
handeyeco, line 1: Rather than make a new file, we need to update `radio.stories.tsx` to use `ServerItemRendererWithDebugUI`. The problem isn't that we need stories for both `ServerItemRendererWithDebugUI` and `RendererWithDebugUI`, the problem is that `RendererWithDebugUI` needs to be deprecated and replaced with `ServerItemRendererWithDebugUI`

packages/perseus/src/widgets/radio/__tests__/radio.testdata.ts
handeyeco: This style issue is unrelated to LEMS-2948, likely just a problem with styling in the fake Perseus Chrome we use in SB. Removing the image only sweeps the issue under the rug, we need a separate ticket to look into the styles causing this.

packages/perseus/src/widgets/radio/__tests__/radio.testdata.ts
```diff
@@ -126,6 +129,9 @@ export const questionWithPassage: PerseusRenderer = {
     },
 };
 
+export const questionWithPassageItem: PerseusItem =
```
handeyeco, line 132: Same here, don't need a new declaration.

testing/server-item-renderer-with-debug-ui.tsx
```diff
@@ -15,7 +15,7 @@ import * as Perseus from "../packages/perseus/src/index";
 import {keScoreFromPerseusScore} from "../packages/perseus/src/util/scoring";
 
 import KEScoreUI from "./ke-score-ui";
-import SideBySide from "./split-view";
+import SplitView from "./split-view";
```
handeyeco, line 18: Thanks!

.changeset/ninety-wasps-love.md
Myranae: For sure. I thought each was small enough to be combined into one, but I can still separate them out if you think that would be useful?

packages/perseus/src/widgets/radio/__tests__/radio.testdata.ts
Myranae: Created a separate ticket here: https://khanacademy.atlassian.net/browse/LEMS-2955

.changeset/ninety-wasps-love.md
handeyeco: The main goal (IMO) of small PRs is to make review easier. Since you already have the reviews, I don't think it'd be worth the trouble.

packages/perseus/src/widgets/__testutils__/test-data-util.ts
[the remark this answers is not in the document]
jeremywiebe: I _think_ this also points to something Matthew brought up to me in conversation recently, that we have the ServerItemRenderer UI and Renderer UI... they overlap alot and I _think_ we could get away with only the Renderer UI component.  I'll look into that this week yet.

packages/perseus/src/widgets/__testutils__/test-data-util.ts
```diff
@@ -0,0 +1,32 @@
+import type {PerseusItem, PerseusRenderer} from "@khanacademy/perseus-core";
+
+/**
+ * Creates a Perseus item with the provided question/PerseusRenderer data.
+ * Useful for converting test data used with RendererWithDebugUI to test data
+ * for ServerItemRendererWithDebugUI.
+ *
+ * @param {PerseusRenderer} question - The question to be included in the Perseus item.
+ * @returns {PerseusItem} The created Perseus item.
+ */
```
jeremywiebe, line 10: Regarding docs.  For things that are self-evident, I prefer to avoid adding a doc comment.  So I think I'd just recommend this (as the `question` param is quite obvious as is the return value): 

```suggestion
/**
 * Creates a Perseus item with the provided question/PerseusRenderer data.
 * Useful for converting test data used with RendererWithDebugUI to test data
 * for ServerItemRendererWithDebugUI.
 */
```

packages/perseus/src/widgets/__testutils__/test-data-util.ts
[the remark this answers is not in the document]
Myranae: Ah, yea when I talked about this with Matthew, we thought we were leaning towards keeping the ServerItemRenderer and moving away from the Renderer one. We can definitely make things work either way, but that's why the answerless stories are using the ServerItemRenderer component.

packages/perseus/src/widgets/numeric-input/numeric-input.stories.tsx
```diff
@@ -180,6 +182,29 @@ const updateWidgetOptions = (
     };
 };
 
+const createNumericInputItem = (question: PerseusRenderer): PerseusItem => {
```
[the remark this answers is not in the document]
Myranae: These functions should now be gone in favor of using the previously created utility that Matthew pointed out.

packages/perseus/src/widgets/__testutils__/test-data-util.ts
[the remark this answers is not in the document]
Myranae: Removed this duplicate utility function creation and used the one that was already defined.

packages/perseus/src/widgets/dropdown/dropdown.stories.tsx
Myranae: Updated the stories to do inline item creation and reduced the overall number of stories, though I kept one for each type of interactive graph.

packages/perseus/src/widgets/__testutils__/test-data-util.ts
Myranae: Thanks! Ended up removing this file, but glad to know my thought to remove the parameter definitions was correct.

packages/perseus/src/widgets/radio/__stories__/answerlessRadio.stories.tsx
Myranae: I wanted to preserve Controls functionality for the answerful stories and wasn't sure how to do that with ServerItemRendererWithDebugUI. So I changed it from converting all stories to ServerItem to just the answerless stories in the new file. Just took a second attempt and was able to preserve the Controls functionality for all stories. The answerless ones just need to turn `startAnswerless` off or click the Check button first to get full data.

packages/perseus/src/widgets/__testutils__/test-data-util.ts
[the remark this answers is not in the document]
handeyeco: > we thought we were leaning towards keeping the ServerItemRenderer and moving away from the Renderer one

That would be my vote anyway, let's get close to what consumers are mostly using which is ServerItemRenderer

packages/perseus/src/widgets/radio/__stories__/radio.stories.tsx
```diff
@@ -70,24 +81,55 @@ const buildApiOptions = (args: StoryArgs): APIOptions => ({
...
+        }),
+    },
+};
+
+// NOTE(Tamara): The answerless radio stories currently lose the user's input
+// upon switching to answerless data initially. It will remember the user's
+// input after clicking the Check button a second time.
+// TODO(LEMS-2948): After investigating a solution, confirm this issue is fixed
+
+export const AnswerlessSingleSelect = {
```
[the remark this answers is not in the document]
Myranae: I think the way I wrote this was confusing. I updated the comment. User input is lost with the first click. If the user selects an option after that and then clicks Check again, it will remember the input. It's just that first time when it switches to answerful data.

packages/perseus/src/widgets/radio/__stories__/radio.stories.tsx
```diff
@@ -1,26 +1,29 @@
...
 type StoryArgs = {
     // Story Option
-    question: PerseusRenderer;
+    item: PerseusItem;
     // Radio Options
     static: boolean;
     // API Options
     crossOutEnabled: boolean;
+    // Renderer Options
+    startAnswerless: boolean;
```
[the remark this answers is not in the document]
Myranae: Updated!

packages/perseus/src/server-item-renderer.tsx
```diff
@@ -46,7 +46,7 @@ type OwnProps = {
     item: PerseusItem;
     score?: KEScore | null;
     problemNum?: number;
-    reviewMode?: boolean;
+    reviewMode?: boolean | null | undefined;
```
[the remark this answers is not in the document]
Myranae: For sure! That makes sense. Instead, I updated the props for `ServerItemRendererWithDebugUI` to be in alignment with `reviewMode?: boolean;`.

testing/server-item-renderer-with-debug-ui.tsx
```diff
@@ -79,10 +88,13 @@ export const ServerItemRendererWithDebugUI = ({
                         item={renderedItem}
                         dependencies={storybookDependenciesV2}
                         keypadElement={keypadElement}
+                        reviewMode={reviewMode}
+                        {...rest}
```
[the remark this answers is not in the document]
Myranae: For sure. Pulled out the last prop I know was needed and removed the rest spread.

testing/server-item-renderer-with-debug-ui.tsx
```diff
@@ -25,17 +26,25 @@ type Props = {
     item: PerseusItem;
     apiOptions?: APIOptions;
     keypadElement?: KeypadAPI | null | undefined;
-    answerless?: boolean;
+    // Temporary measure testing rendering with answerless data;
+    // only exists until all widgets are renderable with answerless data
+    startAnswerless?: boolean;
+    reviewMode?: boolean | null | undefined;
+    showSolutions?: ShowSolutions;
```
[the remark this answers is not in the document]
Myranae: This only works with Radio because showSolutions is undefined with other widgets. I updated it a bit so that it works with other widgets and also so that the `startAnswerless` toggle actually toggles between the data! Not sure how to make the static toggle do this though.

testing/server-item-renderer-with-debug-ui.tsx
```diff
@@ -58,7 +67,16 @@ export const ServerItemRendererWithDebugUI = ({
...
+    // `reviewMode` and `showSolutions` require answerful data by definition,
+    // so only use answerless data if those are not enabled. Also updates the
+    // startAnswerless toggle so it actually switches between answerless and answerful.
+    const shouldUseAnswerless =
+        useAnswerless &&
+        !reviewMode &&
+        (showSolutions === "none" || !showSolutions) &&
+        startAnswerless;
+
+    const renderedQuestion: PerseusRenderer = shouldUseAnswerless
```
handeyeco, line 79:
```suggestion
    // `reviewMode` and `showSolutions` require answerful data by definition,
    // so only use answerless data if those are not enabled.
    const shouldUseAnswerless =
        useAnswerless &&
        !reviewMode &&
        (showSolutions === "none" || !showSolutions);

    const renderedQuestion: PerseusRenderer = shouldUseAnswerless
```

You don't want to directly use `startAnswerless` because `useAnswerless` is the stateful version of `startAnswerless`. This also doesn't update `startAnswerless` so the comment is a little confusing.

Just realizing `useAnswerless` should probably be renamed so it's not confused with a hook 🤦 That was my bad. Maybe just `answerless`/`setAnswerless` or something. This is getting a little confusing, probably due to bad design decisions on my part...but it's test code, so we can always fix it later.

handeyeco (approved):
`startAnswerless` should be removed from the `shouldUseAnswerless`, but otherwise looks good

testing/server-item-renderer-with-debug-ui.tsx
Myranae: @handeyeco Okay, I renamed `startAnswerless` to `answerless/setAnswerless`. 

Regarding `startAnswerless` being there (on L77 of the highlighted code), when it is there, toggling `startAnswerless` in the Controls for answerless stories immediately changes the data from answerless to answerful. It actually acts as a functioning toggle. It doesn't seem to work if `startAnswerless` starts out being false though. Those need a page refresh. Should I still take `startAnswerless` out from here though?

Myranae: I'll just remove the `startAnswerless` from the conditional for now, but it can be added back or refactored if we want the toggle to work.

PR #2331, opened 2025-03-26
Switch to only generate ES Modules for packages

jeremywiebe:
## Summary:

We've had a long-standing TODO to update the Perseus build system to only generate ES Modules. Today, we build both CommonJS (cjs) and ES Modules (esm). CJS builds go into `dist/` and the esm build goes into `dist/es`. 

This makes for a more complicated `package.json` as well as a complicated build, test, and Storybook/Cypress setup. 

This PR simplifies everything and sets the build system to only generate an ES Module build. It also marks all published packages as being Nodejs v18+ (Node 18 reaches [end-of-life on April 30, 2025](https://github.com/nodejs/Release?tab=readme-ov-file#release-schedule) and so that shouldn't be a problem for any consumers). I've also used the [Pure ESM Package](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c) gist for other inspiration in creating more modern packages.

It also adopts the modern approach to exporting symbols using the `"exports"` key in `package.json`. This has the added benefit of making private everything in the package that isn't specifically mentioned in the `"exports"` key. 

Some final thoughts from around the web confirm we're at a good time to go ESM-only: https://antfu.me/posts/move-on-to-esm-only

Issue: LEMS-2247

## Test plan:

All local tooling works: build, test, cypress, storybook, IG dev tool
Integrate into webapp and build/test. Also check ZND


jeremywiebe:
> **Size Change:** -259 kB (-35.24%) 🎉
> 
> **Total Size:** 476 kB
> 
> Filename	Size	Change	
> `packages/kas/dist/index.js`	25.8 kB	-13.6 kB (-34.52%)	🎉
> `packages/keypad-context/dist/index.js`	416 B	-557 B (-57.25%)	🏆
> `packages/kmath/dist/index.js`	6.51 kB	-4.16 kB (-38.97%)	🎉
> `packages/math-input/dist/index.js`	59.2 kB	-18.2 kB (-23.48%)	🎉
> `packages/math-input/dist/strings.js`	1.65 kB	-206 B (-11.08%)	👏
> `packages/perseus-core/dist/index.js`	20 kB	-11.8 kB (-37.17%)	🎉
> `packages/perseus-editor/dist/index.js`	100 kB	-34.6 kB (-25.63%)	🎉
> `packages/perseus-linter/dist/index.js`	7.69 kB	-16 kB (-67.49%)	🏆
> `packages/perseus-score/dist/index.js`	9.99 kB	-10.8 kB (-52.06%)	🏆
> `packages/perseus-utils/dist/index.js`	462 B	-334 B (-41.96%)	🎉
> `packages/perseus/dist/index.js`	228 kB	-140 kB (-37.98%)	🎉
> `packages/perseus/dist/strings.js`	6.32 kB	-968 B (-13.28%)	👏
> `packages/pure-markdown/dist/index.js`	1.38 kB	-2.28 kB (-62.34%)	🏆
> `packages/simple-markdown/dist/index.js`	7.49 kB	-5.44 kB (-42.09%)	🎉

Note that these sizes are all changed because these `dist/index.js` files are now the ES Module builds (previously they lived in `dist/es/index.js`). Overall the bundle size is less important now because we only export ESM and so any consuming application's bundler should be able to tree-shake away any unused functions. These are still somewhat useful to give us a sense of how much code is in each package.

config/build/babel.config.js
```diff
@@ -1,19 +1,38 @@
...
-    presets: createBabelPresets({
-        platform: "browser",
-        format: "cjs",
-    }),
-    plugins: createBabelPlugins({
-        platform: "browser",
-        format: "cjs",
-        coverage: process.env["BABEL_COVERAGE"],
-    }),
+    presets: [
```
handeyeco, line 6: Love that there's less indirection

config/build/rollup.config.js
```diff
@@ -44,62 +28,20 @@ const makePackageBasedPath = (pkgName, pkgRelPath) => {
...
-    const values =
-        arg != null && arg.length > 0
-            ? arg
-                  .split(",")
-                  .map((p) => p.trim())
-                  .filter(Boolean)
-            : [];
-    return new Set(values.length ? values : defaults);
+    return path.normalize(
+        path.join("packages", pkgName, pkgJson.exports["."].source),
```
handeyeco, line 32: Not sure I get what's happening here with `.exports["."]`

config/build/rollup.config.js
```diff
@@ -200,77 +149,30 @@ const createConfig = (
...
-                    version: pkgJson.version,
-                    format: "cjs",
-                    platform: "browser",
-                    inputFile: exportConfig.source,
-                    file: exportConfig.require,
-                    plugins: [],
-                });
-            }
+    if (!pkgJson.exports) {
+        console.log(pkgName);
```
handeyeco, line 163: We don't need anything else here? Just log the name?

config/cypress/cypress.config.ts
```diff
@@ -2,28 +2,24 @@ import fs from "fs";
...
-    const pkgJson = require(pkgPath);
-    aliases["@khanacademy/" + name] = path.join(
-        __dirname,
-        "../../packages",
-        name,
-        pkgJson.source,
-    );
-});
+
+fg.globSync(path.join(__dirname, "../../packages/*/package.json")).forEach(
```
handeyeco, line 12: Nice

utils/internal/pre-publish-utils.ts
```diff
@@ -33,58 +37,56 @@ const checkPublishConfig = ({
...
-export {checkPublishConfig, checkEntrypoints, checkSource, checkPrivate};
+    let returnValue = true;
+    returnValue =
+        checkExport(pkgJson, "import", "./dist/index.js") && returnValue;
+    returnValue =
+        checkExport(pkgJson, "types", "./dist/index.d.ts") && returnValue;
+    returnValue =
+        checkExport(pkgJson, "source", "./src/index.ts") && returnValue;
+
+    return returnValue;
```
handeyeco, line 91: I feel like it would be clearer and more concise to just use multiple variables.

```suggestion
    const hasTranspiled = checkExport(pkgJson, "import", "./dist/index.js");
    const hasTypes = checkExport(pkgJson, "types", "./dist/index.d.ts");
    const hasSrc = checkExport(pkgJson, "source", "./src/index.ts");

    return hasTranspiled && hasTypes && hasSrc;
```

utils/pre-publish-check-ci.ts
```diff
@@ -8,33 +8,27 @@ import fg from "fast-glob";
...
+    const pkgJson = require(path.relative(__dirname, pkgPath));
 
-        // Exit only after we've processed all the packages.
-        if (!allPassed) {
-            process.exit(1);
-        }
-    },
-);
+    // allPassed is at the end of the chain because of short-circuiting
+    allPassed =
```
handeyeco, line 24: Nice

config/build/rollup.config.js
handeyeco: Ah, I understand now. Each package.json will be in the shape of:

```
{
    "exports": {
        ".": {
            "import": "./dist/index.js",
            "types": "./dist/index.d.ts",
            "source": "./src/index.ts"
        }
    }
}
```

> When using the ["exports"](https://nodejs.org/api/packages.html#exports) field, custom subpaths can be defined along with the main entry point by treating the main entry point as the "." subpath

So I guess we don't get a say in what we call `"."`.

handeyeco (approved):
My brain is fried this morning, but insomuch as I understand it I think it looks good. That's for doing this, it's an exciting step.

config/build/rollup.config.js
jeremywiebe: Correct.  There's a simpler form of the "exports" field like this: `"exports": "./dist/index.js"`, but then we can't specify source and types files. "." is the "alias" for the main export so we gotta life with this.

config/build/rollup.config.js
jeremywiebe: Thanks for catching this. I've changed it to throw. We don't want to allow older-style package.json's when we build/publish.

config/build/babel.config.js
jeremywiebe: Much of the inspiration for these changes is form the WB repo.

utils/pre-publish-check-ci.ts
```diff
@@ -8,33 +8,27 @@ import fg from "fast-glob";
...
+for (const pkgPath of pkgPaths) {
+    const pkgJson = require(path.relative(__dirname, pkgPath));
 
-        // Exit only after we've processed all the packages.
-        if (!allPassed) {
-            process.exit(1);
-        }
-    },
-);
+    // allPassed is at the end of the chain because of short-circuiting
```
benchristel, line 23: Why does `allPassed` have to be at the end of the chain? Do the `check*` functions have side effects that need to happen for each `package.json`? Note that some of the later `check`s might still not be called if the earlier ones return false.

benchristel: To make sure all the checks run, I'd suggest doing something like this:

```ts
function isFalsey(value: unknown): boolean {
  return !value
}

const results = pkgPaths.flatMap((pkgPath) => {
  const pkgJson = require(path.relative(__dirname, pkgPath));
  return [
    checkPrivate(pkgJson),
    checkPublishConfig(pkgJson),
    checkExports(pkgJson),
  ];
})

if (results.some(isFalsey)) {
  process.exit(1);
}
```

jeremywiebe: I like this refactor!  This code was in place originally and when I started looking at it for this PR, I found that it wasn't even working (different callers had a different sense of what the `boolean` return value was - some interpreted `true` as a pass and some as a failure). 

I'll work in the suggestion you made as I think it makes things even clearer and avoids dealing with short-circuiting.

PR #2332, opened 2025-03-27
[Free Response Widget] Add optional character limit support

rgpass:
## Summary:
This PR adds the ability for content creators to set a character limit for the answers. If the user goes over that limit, then the field visually shows it as invalid.

The creator can choose to allow for unlimited characters if they would prefer.

Issue: LIT-1666

## Test plan:

- Open Storybook
- In the editor
  - modify the character limit
  - toggle unlimited characters
- In the renderer
  - type some characters to see the count update
  - type enough to go over the limit, see the error state
  - type a lot into the unlimited, see no error state

### DIFF
--- a/packages/perseus-core/src/data-schema.ts
+++ b/packages/perseus-core/src/data-schema.ts
@@ -1675,6 +1675,10 @@ export type PerseusFreeResponseWidgetScoringCriterion = {
 };
 
 export type PerseusFreeResponseWidgetOptions = {
+    // Whether to allow the user to enter an unlimited number of characters.
+    allowUnlimitedCharacters: boolean;
+    // The maximum number of characters that the user can enter.
+    characterLimit: number;
     // The placeholder text that will be displayed to the user in the text input field.
     placeholder: string;
     // The question text that will be displayed to the user.
--- a/packages/perseus-editor/src/widgets/__tests__/free-response-editor.test.tsx
+++ b/packages/perseus-editor/src/widgets/__tests__/free-response-editor.test.tsx
@@ -31,6 +31,41 @@ describe("free-response editor", () => {
         expect(screen.getByDisplayValue("test-question")).toBeInTheDocument();
     });
 
+    it("calls onChange when the allow unlimited characters is changed", async () => {
+        // Arrange
+        const onChangeMock = jest.fn();
+
+        // Act
+        render(
+            <FreeResponseEditor
+                allowUnlimitedCharacters={false}
+                onChange={onChangeMock}
+            />,
+        );
+
+        await userEvent.click(
+            screen.getByRole("checkbox", {name: /Allow unlimited characters/i}),
+        );
+
+        // Assert
+        expect(onChangeMock).toBeCalledWith({allowUnlimitedCharacters: true});
+    });
+
+    it("calls onChange when the character limit is changed", async () => {
+        // Arrange
+        const onChangeMock = jest.fn();
+
+        // Act
+        render(
+            <FreeResponseEditor characterLimit={0} onChange={onChangeMock} />,
+        );
+
+        await userEvent.type(screen.getByLabelText(/Character limit/), "1");
+
+        // Assert
+        expect(onChangeMock).toBeCalledWith({characterLimit: 1});
+    });
+
     it("calls onChange when the question is changed", async () => {
         // Arrange
         const onChangeMock = jest.fn();
--- a/packages/perseus-editor/src/widgets/__tests__/free-response-editor.test.tsx
+++ b/packages/perseus-editor/src/widgets/__tests__/free-response-editor.test.tsx
@@ -51,7 +51,7 @@ describe("free-response editor", () => {
         expect(onChangeMock).toBeCalledWith({allowUnlimitedCharacters: true});
     });
 
-    it("calls onChange when the character limit is changed", async () => {
+    it("calls onChange when the character limit is changed to a number", async () => {
         // Arrange
         const onChangeMock = jest.fn();
 
@@ -64,6 +64,21 @@ describe("free-response editor", () => {
 
         // Assert
         expect(onChangeMock).toBeCalledWith({characterLimit: 1});
+    });
+
+    it("does not call onChange when the character limit is changed to a non-number", async () => {
+        // Arrange
+        const onChangeMock = jest.fn();
+
+        // Act
+        render(
+            <FreeResponseEditor characterLimit={5} onChange={onChangeMock} />,
+        );
+
+        await userEvent.type(screen.getByLabelText(/Character limit/), "e");
+
+        // Assert
+        expect(onChangeMock).not.toBeCalled();
     });
 
     it("calls onChange when the question is changed", async () => {
--- a/packages/perseus-editor/src/widgets/free-response-editor.tsx
+++ b/packages/perseus-editor/src/widgets/free-response-editor.tsx
@@ -121,14 +121,11 @@ class FreeResponseEditor extends React.Component<Props> {
                         }
                     />
                 </label>
-                <label
-                    className={css(styles.textOptionWithLabelContainer)}
-                    htmlFor="allow-unlimited-characters"
-                >
+                {/* eslint-disable-next-line jsx-a11y/label-has-associated-control -- TODO(LEMS-2871): Address a11y error */}
+                <label className={css(styles.textOptionWithLabelContainer)}>
                     <HeadingSmall>Allow unlimited characters</HeadingSmall>
                     <Checkbox
                         checked={this.props.allowUnlimitedCharacters}
-                        id="allow-unlimited-characters"
                         onChange={(val) =>
                             this.props.onChange({
                                 allowUnlimitedCharacters: val,
--- a/packages/perseus-editor/src/widgets/free-response-editor.tsx
+++ b/packages/perseus-editor/src/widgets/free-response-editor.tsx
@@ -14,6 +14,7 @@ import type {
     FreeResponseDefaultWidgetOptions,
     PerseusFreeResponseWidgetScoringCriterion,
 } from "@khanacademy/perseus-core";
+import type {ChangeEventHandler} from "react";
 
 type Props = PerseusFreeResponseWidgetOptions & {
     onChange: (options: Partial<PerseusFreeResponseWidgetOptions>) => void;
@@ -42,6 +43,15 @@ class FreeResponseEditor extends React.Component<Props> {
         }
         return warnings;
     }
+
+    handleUpdateCharacterLimit: ChangeEventHandler<HTMLInputElement> = (e) => {
+        const val = parseInt(e.target.value);
+        if (isNaN(val)) {
+            return;
+        }
+
+        this.props.onChange({characterLimit: Math.max(1, val)});
+    };
 
     handleUpdateCriterion = (
         index: number,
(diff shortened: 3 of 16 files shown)


packages/perseus-editor/src/widgets/free-response-editor.tsx
```diff
@@ -108,6 +111,37 @@ class FreeResponseEditor extends React.Component<Props> {
...
+                </label>
+                {!this.props.allowUnlimitedCharacters && (
+                    <label className={css(styles.textOptionWithLabelContainer)}>
+                        <HeadingSmall>Character limit</HeadingSmall>
+                        <input
+                            type="number"
+                            min={1}
+                            value={this.props.characterLimit}
+                            onChange={(e) => {
+                                const val = parseInt(e.target.value);
```
aag, line 137: I think we need to check if `val` is not `NaN`, and either not process the change or call `props.onChange` with `1`. If we don't do that, we'll pass through `NaN` to React for rendering, which will cause errors.

packages/perseus-editor/src/widgets/free-response-editor.tsx
```diff
@@ -108,6 +121,32 @@ class FreeResponseEditor extends React.Component<Props> {
...
+                    className={css(styles.textOptionWithLabelContainer)}
+                    htmlFor="allow-unlimited-characters"
+                >
+                    <HeadingSmall>Allow unlimited characters</HeadingSmall>
+                    <Checkbox
+                        checked={this.props.allowUnlimitedCharacters}
+                        id="allow-unlimited-characters"
+                        onChange={(val) =>
+                            this.props.onChange({
+                                allowUnlimitedCharacters: val,
```
aag, line 134: Might it make sense to update the `characterLimit` when `allowUnlimitedCharacters` changes? I could imagine setting it to `0` when `allowUnlimitedCharacters === true`, and setting it to the default when `allowUnlimitedCharacters === false`.

packages/perseus/src/strings.ts
```diff
@@ -4,6 +4,7 @@ import type {ErrorCodes} from "@khanacademy/perseus-score";
  * The translated strings that are used to render Perseus.
  */
 export type PerseusStrings = {
+    characterCount: ({used, num}: {used: number; num: number}) => string;
```
aag, line 7: It might be useful to add a comment explaining that `num` has to be named `num` to indicate that its value controls pluralization.

packages/perseus/src/strings.ts
```diff
@@ -516,6 +517,10 @@ export type PerseusStrings = {
  * !! Note: Ensure that all escape sequences are double-escaped. (e.g. `\\text` -> `\\\\text`)
  */
 export const strings = {
+    characterCount: {
+        one: "%(used)s / %(num)s Character",
```
aag, line 521: Also here: a comment explaining the name `num` could be helpful for maintenance.

aag (commented):
LGTM! I made a few minor suggestions, but I'll leave the final approval to one of the LEMS developers.

packages/perseus-core/src/data-schema.ts
```diff
@@ -1675,6 +1675,10 @@ export type PerseusFreeResponseWidgetScoringCriterion = {
 };
 
 export type PerseusFreeResponseWidgetOptions = {
+    // Whether to allow the user to enter an unlimited number of characters.
+    allowUnlimitedCharacters: boolean;
+    // The maximum number of characters that the user can enter.
+    characterLimit: number;
```
handeyeco, line 1681: Maybe this is a bad idea, but would it make sense to make `characterLimit: number | null`? `null` means "allow unlimited characters" and a number is the the char limit?

packages/perseus-editor/src/widgets/__tests__/free-response-editor.test.tsx
```diff
@@ -31,6 +31,56 @@ describe("free-response editor", () => {
...
+    it("does not call onChange when the character limit is changed to a non-number", async () => {
+        // Arrange
+        const onChangeMock = jest.fn();
+
+        // Act
+        render(
+            <FreeResponseEditor characterLimit={5} onChange={onChangeMock} />,
+        );
+
+        await userEvent.type(screen.getByLabelText(/Character limit/), "e");
```
handeyeco, line 78: I mean `e` [is a number](https://en.wikipedia.org/wiki/E_(mathematical_constant)) 😆 

(ignore me)

packages/perseus-editor/src/widgets/free-response-editor.tsx
```diff
@@ -40,6 +44,15 @@ class FreeResponseEditor extends React.Component<Props> {
         return warnings;
     }
 
+    handleUpdateCharacterLimit: ChangeEventHandler<HTMLInputElement> = (e) => {
+        const val = parseInt(e.target.value);
+        if (isNaN(val)) {
+            return;
+        }
+
+        this.props.onChange({characterLimit: Math.max(1, val)});
```
handeyeco, line 53: I think this makes sense from the technical stand-point. I wonder if we also need some kind of linter that warns for suspiciously low/high numbers - like if someone meant to write 100 and accidentally wrote 1...yeah that's _technically_ fine but do we really want a free response widget that only accepts one character?

packages/perseus-editor/src/widgets/free-response-editor.tsx
handeyeco: This is kind of related to my suggestion of merging `allowUnlimitedCharacters` and `characterLimit` so that `characterLimit === null` means "allow unlimited characters"; it wouldn't allow for a state where `characterLimit` looks like it's being used but it's actually not.

packages/perseus-editor/src/widgets/free-response-editor.tsx
```diff
@@ -108,6 +121,32 @@ class FreeResponseEditor extends React.Component<Props> {
                         }
                     />
                 </label>
+                <label
+                    className={css(styles.textOptionWithLabelContainer)}
+                    htmlFor="allow-unlimited-characters"
```
handeyeco, line 126: Do you need the `id` and `htmlFor` if the label is wrapping the checkbox? If not, I'd advocate for removing them.

packages/perseus/src/strings.ts
```diff
@@ -818,6 +827,10 @@ export const strings = {
  * Mock strings for the Perseus package, to be used for tests and Storybook.
  */
 export const mockStrings: PerseusStrings = {
+    characterCount: ({used, num}) =>
+        num === 1
+            ? `${used} / ${num} Character`
+            : `${used} / ${num} Characters`,
```
handeyeco, line 833: Yeah, this is related to my question of technical correctness vs UX correctness. Do we want to allow a free response input to limit to only one character? Like we're going to force a learner to fill out a response with just "e" and then we're going to send that to an LLM? Maybe so! I don't know the requirements.

If not though, we're adding code/translations that will never be touched.

packages/perseus/src/widgets/free-response/free-response.tsx
```diff
@@ -45,17 +54,64 @@ export class FreeResponse
...
+    };
+
+    isOverLimit = () => {
+        return (
+            !this.props.allowUnlimitedCharacters &&
+            this.characterCount() > this.props.characterLimit
+        );
+    };
+
+    renderCharacterCount(): React.ReactNode {
```
handeyeco, line 68: Just a style nit, but this feels a little inconsistent.

```TS
    isOverLimit = () => {
        // ...whatever
    }

    renderCharacterCount(): React.ReactNode {
        // ...whatever
    }
```

Is there a reason we couldn't stick to one format?

```suggestion
    isOverLimit() {
        return (
            !this.props.allowUnlimitedCharacters &&
            this.characterCount() > this.props.characterLimit
        );
    };

    renderCharacterCount(): React.ReactNode {
```

packages/perseus/src/widgets/free-response/free-response.tsx
```diff
@@ -31,12 +37,15 @@ export class FreeResponse
...
+    declare context: React.ContextType<typeof PerseusI18nContext>;
+
     state: State = {
         currentValue: "",
     };
 
-    handleChange = (event: React.ChangeEvent<HTMLTextAreaElement>) => {
-        this.setState({currentValue: event.target.value});
+    characterCount = () => {
+        return this.state.currentValue.replace(/\n/g, "").length;
```
handeyeco, line 48: Just an observation, likely not an issue:

``` TS
const test1 = "hello world"
const test2 = "hello\nworld"

console.log(test1.replace(/\n/g, "")) // hello world
console.log(test2.replace(/\n/g, "")) // helloworld
console.log(test1.replace(/\n/g, "").length) // 11
console.log(test2.replace(/\n/g, "").length) // 10
```

handeyeco (approved):
Just some nits

packages/perseus-editor/src/widgets/free-response-editor.tsx
aag: Personally, I prefer a dedicated `allowUnlimitedCharacters` variable, for two reasons:
1. It's more explicit. If a developer looks at the `characterLimit` and sees it's `null`, does that mean it's just not set (maybe accidentally) or does it have a valid meaning?
2. It maps more clearly onto the editing UI. We want the user to have to separately choose yes or no for unlimited characters, and the data reflects that.

I'd be open to allowing `null` for the `characterLimit` and nulling it out when the content editor selects `allowUnlimitedCharacters === false`, but I'd prefer to keep the two values separate.

packages/perseus-editor/src/widgets/__tests__/free-response-editor.test.tsx
rgpass: 2.718 character limit 😂

packages/perseus-core/src/data-schema.ts
rgpass: I had a similar idea originally, the only difference being `0` translated to "unlimited characters" -- there was a discussion with Peter to go with a checkbox for unlimited instead.

rgpass: But in terms of data, it seemed clearer (to me at least) to have this always be a number.

packages/perseus/src/strings.ts
rgpass: It would be rare to have only one, but there are languages that have several different ways of addressing plurality. This approach allows us to address that when more languages are supported.

packages/perseus-editor/src/widgets/free-response-editor.tsx
rgpass: That's an interesting idea. I'm curious to see what character limits the content team ends up creating. Do you know if we get insight into this kind of data?

packages/perseus-editor/src/widgets/free-response-editor.tsx
rgpass: Interesting. I added that to silence the [jsx-a11y/label-has-associated-control](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/label-has-associated-control.md#case-my-label-and-input-components-are-custom-components) ESLint issue.

There are a few places where this is ignored and those reference a TODO of [LEMS-2871](https://khanacademy.atlassian.net/browse/LEMS-2871). I added the same comment here.

PR #2334, opened 2025-03-27
add no-restricted-imports rule

handeyeco:
## Summary:
Makes it slightly harder to introduce circular dependencies.

Issue: LEMS-2794


packages/kas/.eslintrc.js
```diff
@@ -11,5 +11,23 @@ module.exports = {
                 includeTypes: true,
             },
         ],
+        // To avoid circular dependencies,
+        // we only allow imports from specific Perseus packages.
+        // This is overly strict to prevent accidental interdependencies,
+        // so feel free to update if you know what you're doing.
+        "no-restricted-imports": [
+            "error",
+            "@khanacademy/kas",
```
handeyeco, line 20: Basically for each of these I added every Perseus package, ran eslint, and removed packages that were already being used.

packages/kas/.eslintrc.js
```diff
@@ -11,5 +11,23 @@ module.exports = {
                 includeTypes: true,
             },
         ],
+        // To avoid circular dependencies,
+        // we only allow imports from specific Perseus packages.
+        // This is overly strict to prevent accidental interdependencies,
+        // so feel free to update if you know what you're doing.
+        "no-restricted-imports": [
```
jeremywiebe, line 18: Not blocking, but I wonder if we should use `@typescript-eslint/no-restricted-imports` instead for better TS support?

jeremywiebe (approved):
I like this... but I also feel like it might be a bit of a maintenance burden.  Let's try it and see how it goes though!

PR #2336, opened 2025-03-28
Fixing the Shift functions Error by resolving a falsey bug with the Grapher Widget's asymptotes. 

SonicScrewdriver:
## Summary:
There's been a long standing bug regarding Graphers in hints, which has been making quite a bit of noise in our reports. (This bug has been causing nearly 70% of our rendering errors!) 

After investigating with @Myranae, we discovered that the issue was stemming from a bug in the truthy logic when determining whether to render an asymptote. Normally, the asymptote property is null for graph types that do not require it. However, I believe that our `staticTransform` method _may_ have been changing the asymptote to an empty array, which was resolving as "truthy". 

I've separated out this logic into a new property to properly handle this case. I've opted not to add new tests or to dig too much deeper into the `staticTransform` method as this was a time-boxed bugfix for a deprecated widget that is scheduled to be replaced with our Interactive Graph Widget in the near future. 
 
(PR completed with @Myranae)

Issue: LEMS-2717

## Screenshot:
Screenshot 2025-03-28 at 1 47 44 PM

## Test plan:
- tests pass 
- manual testing upstream to ensure Shift Functions page loads correctly

### DIFF
--- a/.changeset/tough-stingrays-cross.md
+++ b/.changeset/tough-stingrays-cross.md
@@ -2,4 +2,4 @@
 "@khanacademy/perseus": patch
 ---
 
-logging
+Fix Grapher Widget so that static graphs without asymptotes load correctly.
--- a/.changeset/tough-stingrays-cross.md
+++ b/.changeset/tough-stingrays-cross.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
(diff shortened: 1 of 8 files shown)


Myranae (approved):
Thank you so much for pairing with me and getting us that last final step! So excited!

PR #2341, opened 2025-03-31
Fix `each` vs `forEach` bug

handeyeco:
## Summary:
In https://github.com/Khan/perseus/pull/2278 I removed jQuery from code we needed to run on the server. Unfortunately there are some subtle nuances between jQuery and native JS functions.

One of the differences is that `each` can exit the loop early while `forEach` cannot. I happened to trip on some untested code that was banking on that jQuery feature.

Issue: LEMS-2962

## Test plan:
- Go to the [EditorPage in Storybook](https://khan.github.io/perseus/?path=/story/perseuseditor-editorpage--demo)
- Add a NumericInput widget
- In the input labelled "User input:" under "Answers" type "6/8"
- Note the title of the card "Correct answer:"
    - BROKEN: It will say "Correct answer: 6"
    - FIXED: It will say "Correct answer 3/4"

### DIFF
--- a/packages/perseus-score/src/util/answer-types.ts
+++ b/packages/perseus-score/src/util/answer-types.ts
@@ -563,50 +563,63 @@ const KhanAnswerTypes = {
                     guess: guess,
                 };
 
-                // iterate over all the acceptable forms, and if one of the
-                // answers is correct, return true
-                acceptableForms.forEach((form) => {
-                    const transformed = forms[form](guess);
-                    for (let j = 0, l = transformed.length; j < l; j++) {
-                        const val = transformed[j].value;
-                        const exact = transformed[j].exact;
-                        const piApprox = transformed[j].piApprox;
-                        // If a string was returned, and it exactly matches,
-                        // return true
-                        if (predicate(val, options.maxError)) {
-                            // If the exact correct number was returned,
+                // Iterate over all the acceptable forms
+                // and exit if one of the answers is correct.
+                //
+                // HACK: The anonomous function is a bug fix from LEMS-2962;
+                // after a transition from jQuery's `each` to JS's `forEach`
+                // we realized this code was banking on the ability to exit early
+                // from nested loops which is tricky to do outside of a function.
+                (function () {
+                    // WARNING: Don't use `forEach` without additional refactoring
+                    // code needs to be able to exit early
+                    for (const form of acceptableForms) {
+                        const transformed = forms[form](guess);
+                        for (let j = 0, l = transformed.length; j < l; j++) {
+                            const val = transformed[j].value;
+                            const exact = transformed[j].exact;
+                            const piApprox = transformed[j].piApprox;
+                            // If a string was returned, and it exactly matches,
                             // return true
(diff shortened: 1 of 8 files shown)


packages/perseus-score/src/util/answer-types.ts
```diff
@@ -563,50 +563,63 @@ const KhanAnswerTypes = {
...
+                                    if (options.simplify !== "enforced") {
+                                        score.empty = true;
+                                    }
+                                    score.message =
+                                        ErrorCodes.NEEDS_TO_BE_SIMPLIFIED_ERROR;
                                 }
+                                // HACK: The return false below stops the looping of the
+                                // callback since predicate check succeeded.
+                                // No more forms to look to verify the user guess.
+                                return false;
```
handeyeco, line 610: This line was the problem. In [the PR](https://github.com/Khan/perseus/pull/2278/files#diff-70d36ecf9980a6f939d859313d00c2d6cad1ebf1615ed995e9046d4cb9ffc1e4R568) that introduced the bug, I moved from jQuery's `each` to the native JS function `forEach`. One of the subtle differences is that `each` allows for exiting from a loop early while `forEach` does not allow for exiting early.

My solution is itself pretty hacky (although hopefully well commented and with a test): use a `for` loop wrapped in an anonymous function to allow for exiting early. I tried to extract this logic into its own regular function, but that feels even more hacky/risky because of how the surrounding code is structured and poorly typed. Since there's some urgency to get a fix out and because this code isn't super well tested, I tried to be as surgical as possible.

Myranae (approved):
Looks good! Thanks for removing unused code. Was able to test your fix and saw it update to `3/4`! Thanks so much!

packages/perseus-score/src/util/answer-types.ts
```diff
@@ -563,50 +563,63 @@ const KhanAnswerTypes = {
                     guess: guess,
                 };
 
-                // iterate over all the acceptable forms, and if one of the
-                // answers is correct, return true
-                acceptableForms.forEach((form) => {
```
jeremywiebe, line 568: This feels like it should be `acceptableForms.find()`.  We're searching through `acceptableForms` and the first one that matches we're returning `true` (ie. FOUND!). 

It's a bit weird in that we're updating the `score` as we loop but once we find a match, we'd update the score and returning `true` to break out of the search.

handeyeco: Yeah, I thought about using something like that too. I just feel like using something like `find` is semantically incorrect since this is mutating external variables.

I also thought about updating the logic to be something like:

```
const score = acceptableForms.reduce(
    () => { /* ...whatever... */},
    {
        empty: guess === "",
        correct: false,
        message: null as string | null | undefined,
        guess: guess,
    }
)
```

But I kind of feel like I'm walking on eggshells in this code.

handeyeco:
Asking for reviews from @SonicScrewdriver and @mark-fitzgerald specifically since I added a RTL test for this logic and needed to do a small change to the NumericInput editor.

PR #2345, opened 2025-04-01
Switch storybook to use SWC

jeremywiebe:
## Summary:

This PR moves Storybook to use [SWC](http://swc.rs) for building/bundling stories. SWC is much faster than Babel and is already used in many other places at Khan Academy (and within this repo).

Issue: LEMS-2967

## Test plan:

`pnpm storybook`
Stories load and render


dev/package.json
```diff
@@ -25,8 +25,8 @@
         "@khanacademy/wonder-blocks-button": "catalog:",
         "@khanacademy/wonder-blocks-core": "catalog:",
         "@khanacademy/wonder-blocks-dropdown": "catalog:",
-        "@khanacademy/wonder-blocks-icon-button": "catalog:",
         "@khanacademy/wonder-blocks-icon": "catalog:",
+        "@khanacademy/wonder-blocks-icon-button": "catalog:",
```
jeremywiebe, line 29: pnpm re-ordered this automatically.

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-function-settings.test.tsx
```diff
@@ -29,12 +29,11 @@ jest.mock("./util", () => ({
...
-const exampleEquationsMock = {
-    foo: ["bar", "zot"],
-};
 jest.mock("./locked-function-examples", () => ({
     __esModule: true,
-    default: exampleEquationsMock,
+    default: {
+        foo: ["bar", "zot"],
+    },
 }));
```
jeremywiebe, line 37: Both babel and swc hoist `jest.mock()` calls to ensure they work correctly. Apparently with SWC though, the hoisting is stricter. This was complaining that it couldn't find the variable `exampleEquationsMock`. Because this was the only usage, I simply inlined it.

utils/publish-snapshot.sh
```diff
@@ -84,7 +84,7 @@ check_for_changes() {
 }
 
 pre_publish_check() {
-    SWCRC=true node -r @swc-node/register "$ROOT/utils/pre-publish-check-ci.ts"
```
jeremywiebe, line 87: This started complaining about a @swc/helpers "MODULE_NOT_FOUND". I'm not 100% sure why, but we already use `@swc-node/register` _inside_ this script and it's already `+x` so we can just call it.

dev/vite.config.ts
```diff
@@ -10,5 +9,4 @@ export default mergeConfig(viteConfig, {
         // Dev/Storybook environment.
         "process.env.STORYBOOK": "true",
     },
-    plugins: [react({jsxRuntime: "automatic"})],
```
jeremywiebe, line 13: see the `vite.config.ts` at the root of the repo - we include the react plugin there now.

handeyeco (approved):

PR #2346, opened 2025-04-01
Vendor in mathquill release to work around Github issue

jeremywiebe:
## Summary:

Github has started returning a `401 Unauthorized` error for our mathquill dependency. It is referenced using a url to the release asset. 

image

Over in webapp, they've vendored in the tarball and just install from it locally. This feels like a more reliable solution as we know we'll have the tarball if we clone the repo!

Issue: "none"

## Test plan:

Clear pnpm cache
`rm -rf $(pnpm store path)`

`pnpm clean`
`pnpm install --frozen-lockfile`

Should all succeed


SonicScrewdriver (approved):
Ooo thanks for fixing this so quickly!

catandthemachines (approved):
Sounds like a plan to me. Approved!

PR #2347, opened 2025-04-02
Switch Jest to use SWC

jeremywiebe:
## Summary:

This PR moves Jest to use [SWC](http://swc.rs) for building/bundling stories. SWC is much faster than Babel and is already used in many other places at Khan Academy (and within this repo).

This PR doesn't yet remove babel deps. I'll do that in the final SWC cleanup PR.

Issue: LEMS-2967

## Test plan:

`pnpm test`


config/test/test.config.js
```diff
@@ -35,13 +35,20 @@ const pkgMap = fg
         };
     }, {});
 
-console.log(pkgMap);
+// NOTE: We need to use this plugin in order to turn the module exports
+// into module.exports. This will make it so that we can mock exports
+// correctly.
+const swcrc = JSON.parse(fs.readFileSync(path.join(root, ".swcrc"), "utf8"));
+(((swcrc.jsc ??= {}).experimental ??= {}).plugins ??= []).push([
```
jeremywiebe, line 42: See https://github.com/magic-akari/swc_mut_cjs_exports 

The funky `??= {}` is a supported JS syntax that defaults an object if the key is undefined, BUT, it also ASSIGNS that default to the key. 

See [Nullish coalescing assignment (??=)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment)

handeyeco: Fancy! Kind of hard to read, but I couldn't think of an alternative so I think I just need to retrain my brain to read this as-is; seems like a handy tool to have on the tool belt.

package.json
```diff
@@ -47,8 +47,10 @@
         "@storybook/addon-viewport": "^8.6.11",
         "@storybook/react": "^8.6.11",
         "@storybook/react-vite": "^8.6.11",
-        "@swc-node/register": "^1.6.6",
-        "@swc/core": "^1.3.68",
+        "@swc-node/register": "^1.10.10",
+        "@swc/core": "1.11.13",
+        "@swc/helpers": "^0.5.15",
+        "@swc/jest": "^0.2.37",
```
handeyeco, line 53: You don't seem to be importing these, I'm guessing it's just behind-the-scenes implicit magic somewhere?

handeyeco (approved):

package.json
jeremywiebe: `@swc/jest` is used in `test.config` as a Jest `transform`. 
I've removed `@swc/helpers` (I think I added it while trying to get something working).

config/test/test.config.js
jeremywiebe: Updated!

PR #2348, opened 2025-04-02
Switch ESLint to use SWC

jeremywiebe:
## Summary:

This PR moves ESLint to use [SWC](http://swc.rs) for building/bundling stories. SWC is much faster than Babel and is already used in many other places at Khan Academy (and within this repo).

There were a few Babel rules we were using, but I found equivalent ones in `@typescript-eslint`.

This PR doesn't yet remove babel deps. I'll do that in the final SWC cleanup PR.

Issue: LEMS-2967

## Test plan:

`pnpm lint`


.eslintrc.js
```diff
@@ -226,14 +220,6 @@ module.exports = {
         "object-curly-spacing": "off",
         semi: "off",
 
-        /**
-         * @babel
-         */
-        // "@babel/new-cap": "error",
-        "@babel/no-invalid-this": "error",
-        "@babel/object-curly-spacing": "error",
```
jeremywiebe, line 234: This is a formatting rule - prettier handles all formatting!

.eslintrc.js
```diff
@@ -226,14 +220,6 @@ module.exports = {
...
-        "@babel/semi": "error",
```
jeremywiebe, line 235: These are formatting rules - prettier handles all formatting!

.eslintrc.js
```diff
@@ -214,10 +209,9 @@ module.exports = {
         },
     ],
     rules: {
-        "max-lines": "off",
```
jeremywiebe, line 217: I removed this - unsure if that's a great idea.  There are a handful of files that trip the 1000 line limit (which is defined in our shared KA linter config). 

I think it's worth at least a warning if you go over 1,000 lines in a file. Anyone is free to eslint-ignore it, but at least it causes some pause to think about breaking up the file. 🤷‍♂️

packages/simple-markdown/src/index.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines */
 /* eslint-disable prefer-spread, no-regex-spaces, guard-for-in, no-console, no-var */
```
handeyeco, line 2:
```suggestion
/* eslint-disable max-lines, prefer-spread, no-regex-spaces, guard-for-in, no-console, no-var */
```
Feel like we should be consistent, either they're all multi line or all one line

.eslintrc.js
handeyeco: I guess 🤷‍♂️ Worth giving it a shot to see if anyone cares either way.

handeyeco (approved):

PR #2349, opened 2025-04-02
Switch Rollup to use SWC

jeremywiebe:
## Summary:

This PR moves Rollup to use [SWC](http://swc.rs) for building our production files. SWC is much faster than Babel and is already used in many other places at Khan Academy (and within this repo).

This uses a plugin for the SWC compilation, but it's a plugin that is built by the Rollup team and it lives in the Rollup monorepo so I'd expect it to be well-supported. 

This PR also removes all babel deps as they are no longer needed! 

Issue: LEMS-2967

## Test plan:

`pnpm clean; pnpm install; pnpm build`

The built files can be found in `packages/*/dist/*`. They will be somewhat smaller as I've turned on minification (something we couldn't do previously because `terser` broke the content editor tools). I'll double-check they continue to work before landing this PR.


config/build/rollup.config.js
```diff
@@ -179,11 +176,6 @@ const createConfig = (
             autoExternal({
                 packagePath: makePackageBasedPath(name, "./package.json"),
             }),
-            // TODO(FEI-4557): Figure out how to make this plugin work so that
-            // @khanacademy/perseus-editor works in webapp.  If we enable this
-            // plugin right now, the content editor pages in webapp will fail
-            // with the following error: `TypeError: Object(...) is not a function`
-            // terser(),
```
handeyeco, line 186: wut

handeyeco (approved):
https://www.youtube.com/watch?v=ybtl9qVFAjc

config/build/rollup.config.js
jeremywiebe: This was never set up properly when we moved Perseus back to this repo.  Thankfully, today with SWC the minification can be re-enabled and I was able to verify in the editor in webapp that there were no `TypeError`s. 

image

Izzy-CodersPlanet (approved):
Looks good! ✅

PR #2356, opened 2025-04-04
[Expression Editor] Removing sort functionality, fixing delete functionality, and updating change handlers to cleaner format.

catandthemachines:
## Summary:
Updating the expression editor to remove/update several sections of problematic code such as:
- Removing the sort functionality for expression answers.
- Fixing the delete functionality.
- Updating change handlers to no longer use Changeable.

Issue: LEMS-2739

## Test plan:
Run `pnpm test`


packages/perseus-editor/src/widgets/expression-editor.tsx
```diff
@@ -171,44 +133,55 @@ class ExpressionEditor extends React.Component<Props, State> {
...
-        const newKey = _makeNewKey(this.props.answerForms);
-        return {
+    // Gets random key of the format "answer_(1-1000000)"
+    getRandomKey() {
+        return `answer_${Math.round(Math.random() * 1000000)}`;
+    }
+
+    newAnswer: () => void = () => {
+        const answerForms = this.props.answerForms.slice();
+        let newKey = this.getRandomKey();
```
catandthemachines, line 143: Would love thoughts or feedback on the random key generation and if you guys have another suggestion. I tried using `useId()` from React but can't call a hook within this function. Because we just need a random key for likely 1-5 answers I feel like generating a random number from 1 in 1000000 is likely sufficient for this code (with an added check in case it picks the same number). But let me know your thoughts here.

catandthemachines (commented):
Here are some areas of code I would love feedback on!

packages/perseus-editor/src/widgets/expression-editor.tsx
```diff
@@ -20,23 +20,20 @@ import lens from "hubble";
...
-
 const {InfoTip} = components;
 
 type Props = {
-    widgetId?: any;
+    widgetId?: string;
     value?: string;
-} & Omit<PerseusExpressionWidgetOptions, "buttonsVisible"> &
-    Changeable.ChangeableProps;
+    onChange: (any) => void;
```
jeremywiebe, line 35: Is there any way we can type this?  I'd expect that it would be something like this: 

```typescript
    onChange: (newValues: Partial<PerseusExpressionWidgetOptions>) => void;
```

This would mean we can't support the form of passing a string and new value, but I think with literal object syntax it's almost exactly the same "size"

ie. `this.props.onChange("buttonSets", buttonSets)` is actually more than `this.props.onChange({buttonSets})`

packages/perseus-editor/src/widgets/expression-editor.tsx
jeremywiebe: We could go even lower-tech.  Just have a field in the class named `nextId = 0`.  

Then in the `getRandomKey()` (e would want to name it `getNextKey()`). Then it could just be:

```typescript
nextId = 0;

getNextKey() {
    return `answer_${this.nextId++}
}
```

This would mean you don't need to check if the key's already in use either.

jeremywiebe (approved):
Looks good!  This is a nice cleanup to this editor!  Thanks.

packages/perseus-editor/src/widgets/expression-editor.tsx
catandthemachines: Would this break if someone re-rendered the expression-editor (refreshed the page) and it started from 0 again and throwing a React error?

I was thinking I can grab the last answer, see what the id is and start from there or start from 0 if there are no answers. Thoughts?

packages/perseus-editor/src/widgets/__tests__/expression-editor.test.tsx
```diff
@@ -246,23 +235,18 @@ describe("expression-editor", () => {
...
-            },
-            undefined,
-        );
+        expect(mathRandomSpy).toBeCalled();
+        expect(onChangeMock).toBeCalledWith({
+            answerForms: [
+                {
+                    considered: "correct",
+                    form: false,
+                    key: "answer_0",
```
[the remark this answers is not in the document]
catandthemachines: The other tests don't add a new key, it just a hard coded/starting value (I can update them to this new format). I picked this format because I wanted to move away with conflating the `key` value with the `index` of the answerForm value.

And yes we generate the keys using `Math.random()` I just mocked the return so it could be controlled for the test. 🥰

packages/perseus-editor/src/widgets/expression-editor.tsx
```diff
@@ -292,109 +249,104 @@ class ExpressionEditor extends React.Component<Props, State> {
...
                         key={ans.key}
-                        draggable={true}
                         considered={ans.considered}
                         expressionProps={expressionProps}
                         form={ans.form}
                         simplify={ans.simplify}
                         onDelete={() => this.handleRemoveForm(index)}
                         onChangeSimplify={(simplify) =>
-                            this.changeSimplify(key, simplify)
+                            this.changeSimplify(index, simplify)
```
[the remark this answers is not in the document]
catandthemachines: That is correct. The big change here was separating out a someone wanting to delete an answer at a specific index vs. the key value we use to track the react widget.

packages/perseus-editor/src/widgets/expression-editor.tsx
```diff
@@ -171,44 +133,55 @@ class ExpressionEditor extends React.Component<Props, State> {
...
     // This function is designed to update the answerForm property
     // with new data. This function should not be used to update any
     // other properties within ExpressionEditor except extraKeys
     // which is derived from answerForms
-    updateAnswerForm(i: number, props: AnswerForm) {
+    updateAnswerForm(index: number, props: AnswerForm) {
         const answerForms = lens(this.props.answerForms)
-            .merge([i], props)
+            .merge([index], props)
             .freeze();
```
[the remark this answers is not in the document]
catandthemachines: Love it! I'll follow-up with another PR to rip `hubble` out 😉

packages/perseus-editor/src/widgets/expression-editor.tsx
jeremywiebe: That's a good point.  Ok, so instead of using `Math.random()` what if we used a  better random value generator?  There's a function [`crypto.randomUUID()`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) that will generate a new, unique string on each call.  If we use these values we don't need to worry about key collisions and it'll work on page refreshes. 

Looks like this API is also well-supported in all browsers we support.

jeremywiebe: Example... called it 5 times:

image

packages/perseus-editor/src/widgets/__tests__/expression-editor.test.tsx
```diff
@@ -77,15 +77,9 @@ describe("expression-editor", () => {
...
-                answerForms: [],
-                buttonSets: ["basic"],
-                functions: ["f", "g", "h"],
-                times: true,
-            },
-            undefined,
-        );
+        expect(onChangeMock).toBeCalledWith({
+            times: true,
+        });
```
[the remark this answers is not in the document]
catandthemachines: Created a ZND to test the implementation and it looks to be working as expected and seems to have fixed the weird mysterious `static: false` value that has plagued the editor. 🥳

https://prod-znd-250411-31445-dcc446c.khanacademy.org/

packages/perseus-editor/src/widgets/__tests__/expression-editor.test.tsx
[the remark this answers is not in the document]
catandthemachines: Updated the keys in the tests.

PR #2360, opened 2025-04-07
ghange Aphrodite to inline styles for snapshots

jeremywiebe:
## Summary:

The Wonder Blocks repository uses a setting that causes Aphrodite to inline styles when generating snapshots. This is extremely useful as it makes it possible to compare what's changed in a style when a snapshot test breaks. This is compared to our current setting which simply adds a "mangled" style name in place which tells you nothing of what the actual styling is.

Issue: "none"

## Test plan:

Review the `.snap` file updates. Note that you can now see the exact styling applied at each location. Future PRs that affect these styles will be much clearer to review.


jeremywiebe:
Closing. We _may_ never get to this and it's just cluttering things up.

PR #2368, opened 2025-04-09
Create helper for DINER emptiness

handeyeco:
## Summary:
**The Problem**
1. `emptyWidgetsFunctional` considers a NumericInput as non-empty with a value of `""`. That's because in some contexts, `""` _is_ a correct answer (when `coefficient` is `true` and one of the answers is `1`). But we can't check that the answer is 1 on the FE, because the FE can't have answers. While we need to preserve this behavior for KA Classic, AX doesn't care about the coefficient edge-case; they just want `""` to be considered empty.
2. Our original workaround was to switch from `emptyWidgetsFunctional` to `scorePerseusItem`. However, `scorePerseusItem` does _too much_ validation for AX, which is the root of LEMS-3000.

**Proposed Solution**
For now, this is a quick and dirty hack: just copy what `emptyWidgetsFunctional` is doing and we can adjust it to meet AX's needs.

Issue: LEMS-3000


packages/perseus-score/src/empty-diner-check.ts
```diff
@@ -0,0 +1,72 @@
...
+    const usedWidgetIds = getWidgetIdsFromContent(itemData.question.content);
+    const widgets = itemData.question.widgets;
+
+    for (const widgetId of usedWidgetIds) {
+        const widget = widgets[widgetId];
+        const input = userInputMap[widgetId];
+
+        switch (widget.type) {
+            case "dropdown": {
+                // see `validateDropdown`
```
handeyeco, line 31: I thought about using the validate functions directly, but I figured we've had some issues juggling AX and KA Classic so I figured I'd just keep them separate for now.

packages/perseus-core/src/data-schema.ts
```diff
@@ -1253,7 +1253,7 @@ export type PerseusNumericInputAnswer = {
     // A range of error +/- the value
     // NOTE: perseus_data.go says this is non-nullable even though we handle null values.
     maxError: number | null | undefined;
-    // Unsimplified answers are Ungraded, Accepted, or Wrong. Options: "required", "correct", or "enforced"
+    // Unsimplified answers are Ungraded, Accepted, or Wrong.
```
handeyeco, line 1256: The comment was wrong and there's now a real TS type for this.

packages/perseus-score/src/empty-diner-check.ts
```diff
@@ -0,0 +1,72 @@
...
+                // IG doesn't do client-side validation
+                break;
+            }
+            case "numeric-input": {
+                // this is possibly a source of bugs,
+                // really it should be:
+                // empty input && not coefficient && answer is not 1
+                // but we don't have answers here
+                // (see: `scoreNumericInput`)
+                if (!input.currentValue && !widget.options.coefficient) {
```
handeyeco, line 47: I could also check for "answer === 1" here, but we can't have that in the future so I don't see the point of adding it now.

packages/perseus-score/src/widgets/radio/validate-radio.ts
```diff
@@ -12,11 +12,7 @@ import type {
  * @see `scoreRadio` for the additional validation logic and the scoring logic.
  */
 function validateRadio(userInput: PerseusRadioUserInput): ValidationResult {
-    const numSelected = userInput.choicesSelected.reduce((sum, selected) => {
-        return sum + (selected ? 1 : 0);
-    }, 0);
-
-    if (numSelected === 0) {
+    if (!userInput.choicesSelected.includes(true)) {
```
handeyeco, line 15: I think this is the same...

packages/perseus-core/src/data-schema.ts
jeremywiebe: Nice.

packages/perseus-score/src/empty-diner-check.test.ts
```diff
@@ -0,0 +1,288 @@
+import emptyDinerCheck from "./empty-diner-check";
+
+import type {UserInputMap} from "./validation.types";
+import type {PerseusItem, PerseusRenderer} from "@khanacademy/perseus-core";
+
+function generatePerseusItem(question: PerseusRenderer): PerseusItem {
```
jeremywiebe, line 6: Non-blocking, but would you mind making a TODO to come back and merge this with the one in Perseus?  Or maybe it's not that big a deal to have two similar test helpers in different packages. 

https://github.com/Khan/perseus/blob/c8e383b469426182a1392ca6ad2cde21b61e2f40/packages/perseus/src/util/test-utils.ts#L61

packages/perseus-score/src/empty-diner-check.test.ts
```diff
@@ -0,0 +1,288 @@
...
+                currentValue: "",
+            },
+        };
+
+        expect(emptyDinerCheck(generatePerseusItem(question), userInput)).toBe(
+            true,
+        );
+    });
+
+    it("passes an unanswered numeric-input (coefficient edge case)", () => {
```
jeremywiebe, line 205: This is protective, but we don't expect these to show up for projects that exclusively use the DINER widgets, right?

packages/perseus-score/src/empty-diner-check.ts
```diff
@@ -0,0 +1,72 @@
...
+ * - InteractiveGraph
+ * - NumericInput
+ * - Expression
+ * - Radio
+ *
+ * @param {PerseusItem} itemData
+ * @param {UserInputMap} userInputMap
+ * @returns {boolean} true if there's an empty widget, otherwise false
+ */
+function emptyDinerCheck(
```
jeremywiebe, line 18: This function name doesn't immediately reveal the meaning of the answer. When going through the unit tests, I had to think about what `true` and `false` meant. What do you think of `hasEmptyDinerWidgets()`?  Or `hasEmptyWidgetsDiner()`?

packages/perseus-score/src/empty-diner-check.ts
jeremywiebe: I think that's fair.  I'd love for this to be a short-term solution that we can remove once we settle on what to do with the `coefficient` flag and empty `numeric-input` widgets .

packages/perseus-score/src/empty-diner-check.ts
```diff
@@ -0,0 +1,72 @@
...
+        switch (widget.type) {
+            case "dropdown": {
+                // see `validateDropdown`
+                if (input.value === 0) {
+                    return true;
+                }
+                break;
+            }
+            case "interactive-graph": {
+                // IG doesn't do client-side validation
```
adinardi, line 38: So there's no sort of "must interact" with it or anything else? It's just always good to go?

packages/perseus-score/src/widgets/radio/validate-radio.ts
jeremywiebe: Lol, I believe that's right.  We just want to know if _any_ choices are selected (ie `true`).

jeremywiebe (approved):

packages/perseus-score/src/empty-diner-check.ts
adinardi: Do we need to document anything in SSS reqs or some tickets or anything? I'd just want to make sure we continue to have the AX cases handled in an SSS world.

adinardi (approved):
seems like it'll probably work! I had a question about graphs.

packages/perseus-score/src/empty-diner-check.ts
handeyeco: I think the important thing is to have AX unit tests for all cases that are considered important.

timmcca-be:
Screenshot 2025-04-09 at 2 10 46 PM

I figured it out eventually ([DINER refers to the class of widgets we use in assessments](https://khanacademy.atlassian.net/wiki/spaces/LC/pages/3276734623/Project+Plan+for+SSS#Rough-overview-of-upcoming-AX-work)), but thought this was funny

packages/perseus-score/src/empty-diner-check.test.ts
handeyeco: I'll file a ticket. I just couldn't import from `perseus` into `perseus-score`. I think we just need a central place to help construct objects.

PR #2369, opened 2025-04-09
Update pnpm-lock.yaml

benchristel:
Issue: none

## Test plan:
Run `pnpm install`. There should be no change to the lockfile.


jeremywiebe (approved):
Weird... I wonder what got missed.  Thanks for fixing!

PR #2370, opened 2025-04-09
Change team-lems-perseus to eng-perseus in Github Actions

Myranae:
## Summary:
There was a recent name change to the `#team-lems-perseus` channel to reflect the diversity of teams currently working in Perseus. Because we send notifications to this Slack channel when publishing Perseus, the channel name needed to be updated.

## Test plan:
- Confirm Slack messages are sent to the `#eng-perseus` channel for future publishes of Perseus after this change is released (hopefully during the release of this change too)
- Nothing else should change regarding the product or the release process


jeremywiebe (approved):
Thanks for updating this Tamara.

PR #2371, opened 2025-04-09
Add typesafe parsers for the UserInput types of DINER widgets

benchristel:
DINER = Dropdown, Interactive Graph, Numeric Input, Expression, and Radio.
These are the widgets needed for the Assessments pilot.

Having parsers for these types will allow us to detect misformatted input in
calls to the the scoring API, and hopefully recover from any bugs or
integration issues more quickly.

Issue: LEMS-3014

See also: LEMS-3015, LEMS-3016, LEMS-3017, LEMS-3018, LEMS-2968

## Test plan:

`pnpm tsc`


packages/perseus-core/src/validation.types.ts
benchristel, line 1: This file moved without changes. @handeyeco I'm looking for feedback about where these types should live. Should they go in `types.ts`, or maybe even `data-schema.ts`?

jeremywiebe (commented):
This looks good to me.  I'll leave this open so Matthew has chance to reply to your question.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
```diff
@@ -157,9 +156,7 @@ const parsePerseusGraphTypeSinusoid: Parser<PerseusGraphTypeSinusoid> = object({
     coord: optional(pairOfNumbers),
 });
 
-const parsePerseusGraphType: Parser<PerseusGraphType> = discriminatedUnionOn(
```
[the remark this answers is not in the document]
benchristel: We currently have a DX problem where some changes to the data-schema don't force corresponding changes to the parser. E.g. suppose we add a new optional property to a widget's options:

```ts
type FooWidgetOptions = {
  bar?: number
}
```

A parser that ignores or strips out the `bar` field will still satisfy this type, because `bar` is optional. For example, this parser will work:

```
const parseFooWidgetOptions: Parser<FooWidgetOptions> = object({});
```

An empty object is always a valid `FooWidgetOptions`, even though it's clearly not what the authors of the code had in mind!

A similar issue occurs when we add a variant to a union type:

```diff
  type FooWidgetOptions = {
    union:
      | "one"
      | "two"
+     | "new"
  }
```

Type checking won't fail if the parser outputs `{union: "one" | "two"}`. But the parser will reject `{union: "new"}`!

To fix this, I'm planning to add typetests to ensure the inferred return types of our parsers are mutually assignable with the data-schema types, e.g.:

```ts
const Parsed = ParsedValue<typeof parseFooWidgetOptions>;
summon<Parsed>() satisfies FooWidgetOptions;
summon<FooWidgetOptions>() satisfies Parsed;
```

However, if we explicitly type the parsers, `ParsedValue<typeof parseFooWidgetOptions>` will be the explicit type rather than the inferred type — which makes typetests like the above tautological. To enforce this, we could have the linter ban imports from `data-schema.ts` in parser files.

packages/perseus-core/src/validation.types.ts
benchristel, line 1: @handeyeco I'm looking for feedback on where this file should live. I think it's fine to put it in `perseus-core`. It's slightly awkward that it's mostly used by scoring code, but I think that's a minor issue.

Weren't we talking about merging `perseus-score` into `perseus-core`? What did we decide about that? If we merged the two packages, this question would be moot.

handeyeco: I think it's fine for these to live in `perseus-core` and eventually we'll probably end up moving `perseus-score` into `perseus-core`.

Especially the user input types, since that's scoring but also user interaction (`getUserInput` will always want to use the type in `perseus`).

Rubric maybe less so; we use rubric types for WidgetProps:

``` TS
export type WidgetProps<
    RenderProps,
    Rubric = Empty, // <- here
    TrackingExtraArgs = Empty,
>
```

However that's only used for `reviewModeRubric?: Rubric | null | undefined;` and [we want to remove reviewModeRubric](https://khanacademy.atlassian.net/browse/LEMS-2804). So Rubric is a subset of WidgetOptions that really is _just_ for scoring.

I personally don't super care. I'm getting to the point where I'm wondering if we shouldn't just have a `perseus-types` package where we could just dump everything.

handeyeco: Like I could be wrong, but I think this would be fine (as an example that we really don't need Rubrics everywhere): https://github.com/Khan/perseus/pull/2382

handeyeco (approved):
Conditional approval in that I don't think we want to export the same stuff from two packages.

Otherwise, I think it makes sense. If it were me, I'd probably keep the rubric types in `perseus-score` and move the user input types. It makes sense you had to move those since we want the parser in `perseus-core` and importing those types from `perseus-score` would have caused a circular dependency.

packages/perseus-score/src/index.ts
```diff
@@ -1,7 +1,64 @@
...
+    RubricMap,
+    Rubric,
+    UserInput,
+    UserInputMap,
+    UserInputArray,
+    ValidationDataMap,
+    ValidationData,
+    RubricRegistry,
+    ValidationDataTypes,
+} from "@khanacademy/perseus-core";
```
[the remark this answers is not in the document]
benchristel: I kept the perseus-score exports for backward compatibility. Maybe I'm misremembering, but I seem to recall that incompatibilies have caused us pain when deploying Perseus into webapp. This is especially true when we're on a restricted release schedule with more changes piling up between releases, and the person doing the webapp deploy is unlikely to be the person who wrote the breaking change. 

The future I'm imagining is:

- we release this PR and deploy it into webapp
- we update webapp to import these types from `perseus-core`
- we release a new major version of perseus-score that doesn't re-export these types.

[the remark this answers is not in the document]
handeyeco: We have an AX deploy branch we could work on. So you could land this with a breaking change, cut a Perseus release, and then bump the Perseus version and update the imports in the AX deploy branch; then we don't have duplicate exports and the release engineers don't have to deal with it themselves.

PR #2373, opened 2025-04-10
Don't export mediaQueries - not used (and shouldn't be) outside of Perseus

jeremywiebe:
## Summary:

I noticed in webapp that the Perseus `mediaQueries` object is being used. It was unrelated to Perseus and so I fixed it. But it pointed to the fact that there's no reason for this object to be exported. 

I checked other packages in this repo and none of them use `mediaQueries`, so I'm changing things so that we don't export it anymore; it's an internal thing.

Issue: "none"

## Test plan:

`pnpm typecheck`
Check consuming apps and ensure none use `mediaQueries`

### DIFF
--- a/.changeset/new-trains-swim.md
+++ b/.changeset/new-trains-swim.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": major
+---
+
+No longer export 'mediaQueries' object (it is internal-only)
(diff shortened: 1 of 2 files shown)


mark-fitzgerald (approved):
I am now motivated to remove the `styles/media-queries` file all together!

PR #2374, opened 2025-04-10
Narrow onAnalyticsEvent type to only those events that Keypad component actually generates

kevinb-khan:
## Summary:
I'm doing some refactoring in webapp to limit where we need to use 'ResolvedContentPath' types.  By narrowing the type of onAnalyticsEvent to only the event types that the Keypad actually generates it means that we don't have to use 'ResolveContentPath' when using the Keypad component in the AI Guide.  See https://khanacademy.slack.com/archives/C06P903SXV3/p1744302526954149?thread_ts=1744290664.479029&cid=C06P903SXV3 for more details.

Issue: None

## Test plan:
- pnpm typecheck


packages/math-input/src/components/keypad/keypad.tsx
```diff
@@ -39,7 +36,12 @@ type Props = {
     scientific?: boolean;
 
     onClickKey: ClickKeyCallback;
-    onAnalyticsEvent: AnalyticsEventHandlerFn;
+    onAnalyticsEvent: (
+        event: Extract<
+            PerseusAnalyticsEvent,
+            {type: "math-input:keypad-opened" | "math-input:keypad-closed"}
+        >,
+    ) => void;
```
jeremywiebe, line 44: Yeah, this makes sense.

jeremywiebe (approved):

PR #2377, opened 2025-04-10
Issues Panel (#2377)

tatianasnook:
This PR introduces the creation of the Issues Panel Container(LEMS-2925). The primary goal of this container is to provide a UI that displays a list of warnings with relevant details. Each warning includes title, description, impact, and message to help content creators understand potensial issues. The panel also uses icons to clearly indicate whether there are accessibility concerns or if all checks have passed.

Author: tatianasnook

Reviewers: mark-fitzgerald, tatianasnook, catandthemachines

Required Reviewers:

Approved By: catandthemachines, mark-fitzgerald

Checks: ✅ 18 checks were successful

Pull Request URL: https://github.com/Khan/perseus/pull/2377

Issue: LEMS-2925


packages/perseus-editor/src/issues-panel.tsx
```diff
@@ -0,0 +1,154 @@
...
+    ];
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const [violations, setViolations] = useState(hardCodedViolations); // Update type based on new approach
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const [incompletes, setIncompletes] = useState(hardCodedIncompletes); // Update type based on new approach
+
+    const [showPanel, setShowPanel] = useState(false);
+    const toggleIcon = showPanel ? caretDown : caretRight;
+    const togglePanel = () => setShowPanel((prev) => !prev);
```
mark-fitzgerald, line 50: Should this be:
```
const togglePanel = () => setShowPanel(! showPanel);
```

packages/perseus-editor/src/issues-panel.tsx
```diff
@@ -0,0 +1,154 @@
...
+                    />
+                    <span>Issues</span>
+                </div>
+                {issuesCount}
+                <PhosphorIcon icon={icon} size="medium" color={iconColor} />
+            </div>
+            {showPanel && (
+                <div className="perseus-widget-editor-panel">
+                    <div className="perseus-widget-editor-content">
+                        {hardCodedViolations.map((issue, index) => (
```
mark-fitzgerald, line 91: This probably needs to reference `violations` instead of `hardCodedViolations`.

packages/perseus-editor/src/issues-panel.tsx
```diff
@@ -0,0 +1,154 @@
...
+                            <IssueDetails
+                                key={issue.id}
+                                issue={issue}
+                                issueType="violation"
+                            />
+                        ))}
+                        {hardCodedIncompletes.map((issue, index) => (
```
mark-fitzgerald, line 98: This probably needs to reference `incompletes` instead of `hardCodedIncompletes`.

packages/perseus-editor/src/issues-panel.tsx
```diff
@@ -0,0 +1,154 @@
...
+        </div>
+    );
+};
+
+type IssueProps = {
+    issueType: "violation" | "incomplete";
+    issue: any; // Update type based on new approach
+};
+
+const IssueDetails = (props: IssueProps) => {
```
mark-fitzgerald, line 117: This should probably be pulled out and I can add it when I do the accessibility scan implementation. That way, you can make a separate file named something like `WidgetIssueDetails` that displays the information specific to your use case, and import it into this panel to be used for the issues that you track. I will make a separate file for the axe-core issue details and will update the panel to pull that in whenever those issues exist.

packages/perseus-editor/src/issues-panel.tsx
tatianasnook: Yes. I will change.

packages/perseus-editor/src/issues-panel.tsx
tatianasnook: Agree

packages/perseus-editor/src/issues-panel.tsx
```diff
@@ -0,0 +1,154 @@
...
+            help: "This is an incomplete help message.",
+            impact: "Medium",
+            message: "This is an incomplete message.",
+        },
+    ];
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const [violations, setViolations] = useState(hardCodedViolations); // Update type based on new approach
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const [incompletes, setIncompletes] = useState(hardCodedIncompletes); // Update type based on new approach
```
catandthemachines, line 46: What's the difference between `violations` vs. `incompletes`, my hunch is incompletes refers to axecore and linter logic. With the warning messages we're starting with I think we only need to worries about `violations`.
@mark-fitzgerald how does this sound?

catandthemachines (commented):
Great first stab at the Issues Panel code I love all the code you extracted from Marks initial proof of concept. One thing I see missing from this POC is how someone (you, me, or another developer) can send violations into this issues panel? It has logic around tracking and counting violations which made sense if it handles calculating axe-core violations. But for this initial implementation the Issues Panel needs to be sent warnings/violations from outside of it's self.

packages/perseus-editor/src/issues-panel.tsx
mark-fitzgerald: That makes sense. When the axe-core issues are supplied, the "incompletes" can easily be placed into the "warnings" (or whatever name is used).

packages/perseus-editor/src/__tests__/issues-panel.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+        // Arrange
+        render(
+            <IssuesPanel
+                violations={[makeIssue("v1"), makeIssue("v2")]}
+                incompletes={[makeIssue("inc1")]}
+            />,
+        );
+
+        // Assert
+        expect(screen.getByText("3 issues")).toBeInTheDocument();
```
mark-fitzgerald, line 86: Should this test also verify that the error icon is showing since there is a combination of errors and warnings? (which is different from testing the warning icon with only 1 violation supplied)

packages/perseus-editor/src/issues-panel.tsx
tatianasnook: @mark-fitzgerald @catandthemachines Thanks so much for the suggestions. I’ve removed incompletes and renamed violations to warnings based on our discussion.

packages/perseus-editor/src/__tests__/issues-panel.test.tsx
```diff
@@ -0,0 +1,107 @@
...
+    helpUrl: "http://example.com",
+    impact,
+    message: "Example message",
+});
+
+// Normally, we avoid direct node access per Testing Library best practices.
+// But <PhosphorIcon> renders as a <span> with no test-friendly props (e.g., no data-testid or role),
+// and attempts like getByTestId, getByText().nextSibling, or adding a test ID weren’t reliable.
+// So to verify the icon (e.g., data-icon-type="check-circle-fill.svg"), direct access was the most
+// consistent option.
```
catandthemachines, line 19: nit: let's move this comment to the first instance of this usage so it's next to the implementation it's discussing. 😊

packages/perseus-editor/src/__tests__/issues-panel.test.tsx
```diff
@@ -0,0 +1,107 @@
...
+        expect(screen.getByText("0 issues")).toBeInTheDocument();
+
+        // Assert that the icon is a passing icon
+        const icon = screen
+            .getByText("0 issues")
+            // eslint-disable-next-line testing-library/no-node-access
+            .parentElement?.querySelector(
+                '[data-icon-type="check-circle-fill.svg"]',
+            );
+        expect(icon).toBeInTheDocument();
```
catandthemachines, line 43: Idea: Can you also grab the warning icon and confirm it is not in the document within this test.

packages/perseus-editor/src/__tests__/issues-panel.test.tsx
```diff
@@ -0,0 +1,107 @@
...
+        expect(screen.getByText("1 issue")).toBeInTheDocument();
+
+        // Assert that the icon is caution icon
+        const icon = screen
+            .getByText("1 issue")
+            // eslint-disable-next-line testing-library/no-node-access
+            .parentElement?.querySelector(
+                '[data-icon-type="warning-fill.svg"]',
+            );
+        expect(icon).toBeInTheDocument();
```
catandthemachines, line 60: Same here: Can you also grab the passing icon and confirm it is not in the document within this test.

packages/perseus-editor/src/__tests__/issues-panel.test.tsx
catandthemachines, line 1: Yay love these tests!

packages/perseus-editor/src/issue-details.tsx
catandthemachines, line 1: Looks great, just surfacing our offline discussion that we want to get Aishwarya's put here on the design of the issue details 😊

catandthemachines (approved):
Everything here is looking really sharp, just added a few nit to clean up in the test.

Per our offline conversation let's get Aishwarya's feedback before committing this change and when we do let's push this to a feature branch so we can get the issues panel and perseus warning message complete before releasing to webapp.

mark-fitzgerald (approved):
Thanks for taking this on! This is going to help consolidate editor issues as we add more capabilities to check the content.

PR #2379, opened 2025-04-10
[Locked functions] Treat null domain values as infinity/-infinity

nishasy:
## Summary:
Todo

Issue: https://khanacademy.atlassian.net/browse/LEMS-3013

## Test plan:
todo


nishasy:
This parser already covers this, so the value isn't ever null.

PR #2380, opened 2025-04-10
Resolve some TODO comments in the Perseus JSON parsing code

benchristel:
Issue: none

## Test plan:

`pnpm test`

### DIFF
--- a/packages/perseus-core/src/data-schema.ts
+++ b/packages/perseus-core/src/data-schema.ts
@@ -771,7 +771,7 @@ export type PerseusInteractiveGraphWidgetOptions = {
     correct: PerseusGraphType;
     // Shapes (points, chords, etc) displayed on the graph that cannot
     // be moved by the user.
-    lockedFigures?: LockedFigure[];
+    lockedFigures: LockedFigure[];
     // Aria label that applies to the entire graph.
     fullGraphAriaLabel?: string;
     // Aria description that applies to the entire graph.
@@ -817,7 +817,7 @@ export type LockedPointType = {
     coord: Coord;
     color: LockedFigureColor;
     filled: boolean;
-    labels?: LockedLabelType[];
+    labels: LockedLabelType[];
     ariaLabel?: string;
 };
 
@@ -829,7 +829,7 @@ export type LockedLineType = {
     lineStyle: LockedLineStyle;
     showPoint1: boolean;
     showPoint2: boolean;
-    labels?: LockedLabelType[];
+    labels: LockedLabelType[];
     ariaLabel?: string;
 };
 
@@ -837,7 +837,7 @@ export type LockedVectorType = {
     type: "vector";
     points: [tail: Coord, tip: Coord];
     color: LockedFigureColor;
-    labels?: LockedLabelType[];
+    labels: LockedLabelType[];
     ariaLabel?: string;
 };
 
@@ -857,7 +857,7 @@ export type LockedEllipseType = {
     color: LockedFigureColor;
     fillStyle: LockedFigureFillType;
     strokeStyle: LockedLineStyle;
-    labels?: LockedLabelType[];
+    labels: LockedLabelType[];
     ariaLabel?: string;
 };
 
@@ -868,7 +868,7 @@ export type LockedPolygonType = {
     showVertices: boolean;
     fillStyle: LockedFigureFillType;
     strokeStyle: LockedLineStyle;
-    labels?: LockedLabelType[];
+    labels: LockedLabelType[];
     ariaLabel?: string;
 };
 
@@ -890,7 +890,7 @@ export type LockedFunctionType = {
      * allowed. Note that infinite values are serialized as `null` in JSON.
      */
     domain: [min: number, max: number];
-    labels?: LockedLabelType[];
+    labels: LockedLabelType[];
     ariaLabel?: string;
 };
 
--- a/packages/perseus-core/src/data-schema.ts
+++ b/packages/perseus-core/src/data-schema.ts
@@ -917,12 +917,6 @@ export type PerseusGraphType =
     | PerseusGraphTypeSegment
     | PerseusGraphTypeSinusoid;
 
-type PerseusGraphTypeCommon = {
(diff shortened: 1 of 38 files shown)


packages/perseus-core/src/parse-perseus-json/exhaustive-test-tool/index.ts
```diff
@@ -195,7 +195,7 @@ function typeName(value: unknown): string {
 async function writeFileIfShorterThanExisting(
     path: string,
     content: string,
-    encoding: "utf-8", // TODO(benchristel): add more encodings if needed
+    encoding: "utf-8", // FUTURE(benchristel): add more encodings if needed
```
jeremywiebe, line 198: Do you anticipate this ever being needed? I wonder if we should just remove this.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
```diff
@@ -71,16 +71,12 @@ const parsePerseusGraphTypeCircle: Parser<PerseusGraphTypeCircle> = object({
             radius: number,
         }),
     ),
-    // TODO: remove coord? it's legacy.
-    coord: optional(pairOfNumbers),
```
jeremywiebe, line 75: **question**: We can remove this because the parser correctly upgrades this field, right? Or is it because we've fully migrated to the Mafs version of the interactive-graph?

packages/perseus-core/src/parse-perseus-json/perseus-parsers/measurer-widget.ts
```diff
@@ -32,7 +32,5 @@ export const parseMeasurerWidget: Parser<MeasurerWidget> = parseWidget(
         rulerPixels: number,
         rulerLength: number,
         box: pair(number, number),
-        // TODO(benchristel): static is not used. Remove it?
-        static: defaulted(boolean, () => false),
```
jeremywiebe, line 36: Nice!  My guess is that this snuck into the object data at some point by accident.

packages/perseus/src/widgets/interactive-graph.tsx
```diff
@@ -163,7 +163,7 @@ type RenderProps = {
      * Shapes (points, chords, etc) displayed on the graph that cannot be moved
      * by the user.
      */
-    lockedFigures?: ReadonlyArray<LockedFigure>;
+    lockedFigures: ReadonlyArray<LockedFigure>;
```
jeremywiebe, line 166: Love that these parsers are giving us the ability to make the production code safer/cleaner!

jeremywiebe (approved):
Really happy to see how the parsers are letting us improve things. Thanks for that! 

I noticed a few linter violations at the bottom of the file listing. Not sure if they're valid... 


image

packages/perseus-core/src/data-schema.ts
```diff
@@ -1192,9 +1192,6 @@ export type PerseusMeasurerWidgetOptions = {
     rulerLength: number;
     // Containing area [width, height]
     box: [number, number];
-    // TODO(benchristel): static is not used. Remove it?
-    // Always false.  Not used for this widget
-    static: boolean;
```
handeyeco, line 1197: Nice, `static` is so annoying

handeyeco (approved):
Cool!

packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
benchristel: Oof, good catch! I checked, and `coord` is actually still used for Point graphs. It seems to be unused for all the others. Typechecking and parsing were succeding with this line removed because the `coord` field was typed as optional. I really want those mutual assignability typetests... 😅

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -321,24 +316,19 @@ export const MafsGraph = (props: MafsGraphProps) => {
...
                             </Mafs>
                         </View>
-                        {props.lockedFigures && (
-                            <GraphLockedLabelsLayer
-                                lockedFigures={props.lockedFigures}
-                            />
-                        )}
+                        <GraphLockedLabelsLayer
+                            lockedFigures={props.lockedFigures}
+                        />
```
Myranae, line 331: So before when `lockedFigures` was optional, if it was falsy in some way, then this `GraphLockedLabelsLayer` element wouldn't be included. Now it's always included. Could that cause any side effects with interaction or the UI?

Myranae (approved):
Wanted to remind you to add back `coords` in interactive graph and to point out some checks need to be fixed. Also, since data-schema changed, we might need to inform CP, iirc? Just one NB question, but looks good otherwise :)

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
nishasy: I don't think this would render in the DOM if `lockedFigures` length is 0, because the outermost layer in `packages/perseus/src/widgets/interactive-graphs/graph-locked-labels-layer.tsx` is a React fragment. It only renders the actual labels themselves if there are any.

So I think this is fine.

packages/perseus-core/src/data-schema.ts
```diff
@@ -782,7 +782,7 @@ export type PerseusInteractiveGraphWidgetOptions = {
     correct: PerseusGraphType;
     // Shapes (points, chords, etc) displayed on the graph that cannot
     // be moved by the user.
-    lockedFigures?: ReadonlyArray<LockedFigure>;
```
nishasy, line 785: Very exciting! The rest of the code for making `lockedFigures` required makes sense to me. Same for `labels`.

Only one question for my own understanding: Is there any possibility of this causing issues for old exercise items that saved this as `null` or `undefined`? Or would that be fixed by the parser?

nishasy (approved):
Nice! Looks good to me.

I added one question for my own understanding, but I don't anticipate it being a problem.

PR #2382, opened 2025-04-11
Remove some unneeded Rubric uses

handeyeco:
## Summary:
See: https://github.com/Khan/perseus/pull/2371/files#r2038337346
See: https://khanacademy.atlassian.net/browse/LEMS-2804

So basically we have this:

``` TS
export type WidgetProps<
    RenderProps,
    Rubric = Empty,
    TrackingExtraArgs = Empty,
> = RenderProps & {
    // ...stuff
    reviewModeRubric?: Rubric | null | undefined;
    // ...more stuff
}
```

AFAICT, Radio is the only widget that uses `reviewModeRubric`. The team working on Radio right now is planning to look at LEMS-2804, but in the meantime I thought it would be good to remove `Rubric` from `WidgetProps` in places where it didn't seem to be needed.


packages/perseus/src/widgets/graded-group/graded-group.tsx
```diff
@@ -105,7 +102,7 @@ type State = {
 // via defaultProps.
 0 as any as WidgetProps<
     PerseusGradedGroupWidgetOptions,
-    PerseusGradedGroupRubric
+    null
```
benchristel, line 105: Since the default for this type parameter is `Empty`, using `null` here is slightly inconsistent with the other widget types. I don't think it really matters, though. Maybe the default type arg should be `null` or `undefined`?

benchristel (approved):
LGTM! I left one comment inline about default type arguments.

packages/perseus/src/widgets/graded-group/graded-group.tsx
handeyeco: I was thinking the same thing, I just always hate defining something as undefined. I'll make the change.

PR #2385, opened 2025-04-11
Stop bundling _any_ mathquill styles/fonts - let host app do that!

jeremywiebe:
## Summary:

Fixes Perseus so that it doesn't ship Mathquill CSS with absolute font paths (ie `/fonts/Symbola.ttf`). This includes a bump to a new version of Mathquill.

This PR also updates the `@khanacademy/math-input` package to bundle all of mathquill (fonts, CSS, etc). This makes bundling things down the line much simpler. I couldn't find any apps consuming Perseus that actually used Mathquill directly, so this also aligns with that usage (consumer doesn't have to do anything special to accommodate Mathquill). 

Also verified Storybook loads the Symbola font properly!

image

Issue: LEMS-3019

## Test plan:

Update webapp with a test version of Perseus from this PR. 
Build webapp - OK!
Run type checking and tests - OK!
Check webapp CSS and build output to ensure the fonts are "bundled" as assets and the CSS has correct paths for font files - OK!


packages/math-input/less/main.less
```diff
@@ -1,2 +1,2 @@
 @import "./overrides.less";
-@import (inline) "../node_modules/mathquill/build/mathquill.css";
+@import "mathquill/build/mathquill.css";
```
jeremywiebe, line 2: This is the correct import syntax to resolve a CSS/Less file out of an dependant npm package.

packages/math-input/package.json
```diff
@@ -42,8 +42,7 @@
     "dependencies": {
         "@khanacademy/keypad-context": "workspace:*",
         "@khanacademy/perseus-core": "workspace:*",
-        "@khanacademy/perseus-utils": "workspace:*",
-        "mathquill": "file:../../vendor/mathquill-v1.0.0.tgz"
```
jeremywiebe, line 46: We move this to be in `devDependencies` because we now bundle everything from mathquill _into_ `@khanacademy/math-input`. This relieves client apps from needing to depend on `mathquill` directly (most client apps don't use `mathquill` directly if they're using `@khanacademy/math-input`)

packages/perseus/src/styles/perseus-renderer.less
```diff
@@ -714,4 +714,4 @@ div.graphie {
     padding-top: 0;
 }
 
-@import "../../../math-input/less/main.less";
+@import (reference) "@khanacademy/math-input/styles.css";
```
jeremywiebe, line 717: This becomes a `(reference)` now to avoid the situation where an app using `math-input` directly would "double" the math-input CSS because it's integrated into the Perseus CSS also. 

This means that consuming apps will need to ensure that math-input CSS is properly imported, but that's a one-line change and allows for deduplication of math-input CSS.

vendor/mathquill-v1.0.2.tgz
jeremywiebe, line 1: This is not the final build yet. I'll bump this PR once more when the [Mathquill PR](https://github.com/Khan/mathquill/pull/30) lands and a new release is done.

vendor/mathquill-v1.0.0.tgz
jeremywiebe, line 1: We can just reference the Git tag for the release on Github.

packages/math-input/.eslintrc.js
```diff
@@ -8,10 +8,6 @@ module.exports = {
             "error",
             {
                 packageDir: [__dirname, path.join(__dirname, "../../")],
-                // note(matthewc): I changed this in a time pinch and I'm not sure
-                // it's right so I made a ticket to circle back to it
-                // https://khanacademy.atlassian.net/browse/LC-864
-                devDependencies: ["**/*.stories.tsx", "**/*.test.tsx"],
```
jeremywiebe, line 14: @handeyeco Weird. I can remove this in this PR and all linting passes. 🤷‍♂️

config/build/rollup.config.js
```diff
@@ -149,12 +150,26 @@ const createConfig = (
...
+                    publicPath: "./assets",
+                    // This resolve is for the fonts in Mathquill. We want to
+                    // bundle them into the math-input package so that
+                    // consumers of @khanacademy/math-input don't need to
+                    // _also_ depend on the mathquill package.
+                    resolve(inputUrl, baseDir) {
+                        // Some inputUrl's have a hash (eg.
+                        // font/Symbola.svg#Symbola). We use the URL object to
+                        // ensure we only have the file path and nothing else.
+                        const url = new URL(inputUrl, "http://example.com");
```
jeremywiebe, line 164: I used the `URL` object here instead of string searching just to ensure we handle things correctly.

config/build/rollup.config.js
```diff
@@ -149,12 +150,26 @@ const createConfig = (
...
+                        const targetFile = path.join(baseDir, url.pathname);
+                        return {
+                            from: targetFile,
+                            source: fs.readFileSync(targetFile),
```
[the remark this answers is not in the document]
jeremywiebe: It's safe to throw here as Rollup will capture the Error and fail the build with a sensible error.

package.json
```diff
@@ -41,6 +41,7 @@
         "@storybook/react-vite": "^8.6.11",
         "@swc-node/register": "^1.10.10",
         "@swc/core": "1.11.13",
+        "@swc/helpers": "^0.5.17",
```
jeremywiebe, line 44: I saw warnings that a few files couldn't resolve this module. I'm not sure why it didn't happen earlier, but adding this to ensure our builds are correct.

jeresig (approved):
Great - thank you for reworking this, these changes look good to me!

.changeset/tame-grapes-ring.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/math-input": minor
+---
+
+Stop bundling mathquill CSS - enables consuming app to control bundling and font resolution
```
handeyeco, line 5: I'm guessing we need to let Learning Equality know?

packages/math-input/.eslintrc.js
handeyeco: 🤷

.changeset/tame-grapes-ring.md
handeyeco: Actually I'm a little confused by this changeset. Here you're saying "enables consuming app to control bundling and font resolution" but in the comment below you say "This resolve is for the fonts in Mathquill. We want to bundle them into the math-input package so that consumers of @khanacademy/math-input don't need to _also_ depend on the mathquill package." The two comments seem at odds, right?

handeyeco (approved):

.changeset/tame-grapes-ring.md
jeremywiebe: It gives consumers control by bundling the Mathquill fonts into the `@khanacademy/math-input` package complete with correct CSS files point to the bundled font files using relative paths. 

The custom resolve was for Mathquill fonts, but after you asked, I found I was able to simplify things and didn't need a custom resolver (the built-in one handles things fine). And that was exactly the motivation for updating Mathquill! Thanks.

PR #2386, opened 2025-04-11
Add flags in APIOption, this will be used for feature flags that can be passed from consuming application.

ivyolamit:
## Summary:
Add flags in APIOption, this will be used for feature flags that can be passed from consuming application.

Issue: LEMS-3039

## Test plan:

### DIFF
--- a/packages/perseus-core/src/feature-flags.ts
+++ b/packages/perseus-core/src/feature-flags.ts
@@ -8,4 +8,6 @@
  * 3. When no active feature flag is available value will be:
  *    export const PerseusFeatureFlags = [];
  */
-export const PerseusFeatureFlags = ["new-radio-widget"];
+const PerseusFeatureFlags = ["new-radio-widget"];
+
+export default PerseusFeatureFlags;
--- a/packages/perseus-core/src/feature-flags.ts
+++ b/packages/perseus-core/src/feature-flags.ts
@@ -1,7 +1,11 @@
 /**
- * IMPORTANT: Do not delete this file or the variable below.
- * This will be the definition of the active feature flags in Perseus
- * that will hide features/changes under development. And will be used
- * by flags of APIOptions type.
+ * IMPORTANT:
+ * 1. Do not delete this file or the variable below.
+ *    This will be the definition of the active feature flags in Perseus
+ *    that will hide features/changes under development. This will be used
+ *    by flags of APIOptions type.
+ * 2. Do cleanup unused feature flags
+ * 3. When no active feature flag is available value will be:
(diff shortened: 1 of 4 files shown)


SonicScrewdriver (approved):
This looks great, thanks for getting this set up so quickly! :)

ivyolamit:
@SonicScrewdriver update the PerseusFeatureFlags comments to have more detailed information/guide.

packages/perseus-core/src/feature-flags.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * IMPORTANT:
+ * 1. Do not delete this file or the variable below.
+ *    This will be the definition of the active feature flags in Perseus
+ *    that will hide features/changes under development. This will be used
+ *    by flags of APIOptions type.
+ * 2. Do cleanup unused feature flags
+ * 3. When no active feature flag is available value will be:
+ *    export const PerseusFeatureFlags = [];
```
mark-fitzgerald, line 9: "Thank you!" - from our future selves.

packages/perseus-core/src/feature-flags.ts
```diff
@@ -0,0 +1,11 @@
...
+ */
+export const PerseusFeatureFlags = ["new-radio-widget"];
```
mark-fitzgerald, line 11: Since this file is called "feature-flags", and we are only dealing with Perseus feature flags, perhaps we have this as a default export?

mark-fitzgerald (approved):

packages/perseus-core/src/feature-flags.ts
ivyolamit: that make sense to me, updated.

packages/perseus/src/types.ts
```diff
@@ -296,6 +296,11 @@ export type APIOptions = Readonly<{
      * only after a good few seconds.
      */
     editorChangeDelay?: number;
+    /**
+     * Feature flags that can be passed from consuming application.
+     * Define the feature flag name in packages/perseus-core/src/feature-flags.ts
+     */
+    flags?: Record<string, boolean>;
```
jeremywiebe, line 303: @ivyolamit @mark-fitzgerald @SonicScrewdriver  I'm late to this, but I'd recommend that we model this as `Record<"flag1" | "flag2" | "flag3", boolean>` instead.  Otherwise, we:

* Never know from the outside what flags Perseus supports at any given time
* Within Perseus, we have no knowledge of what flags the host application may push in.  

We had the same design initially with Mafs flags and moved to be explicit instead for exactly these reasons.

PR #2387, opened 2025-04-11
Update Matrix to render with answerless data

Myranae:
## Summary:
Updates the Matrix widget to render using MatrixPublicWidgetOptions instead of the full widget options. Adds a Storybook story showing the widget can render, is interactive, and is scorable when rendered with answerless data. Answerful data is still required for the actual scoring. Tests showing this functionality are also included.

The render section of Matrix still references `this.props.answers`, but the `transform` function takes a widgetOptions object without answers and adds in an answers array shaped after the board size. The default props are set for an empty nested array, so I think that's all good. The answers field is also what carries the user input.

Issue: LEMS-2975

## Test plan:
- Confirm all checks pass
- Confirm widget still acts as expected
- No other changes should be noticeable


packages/perseus/src/util/test-utils.ts
```diff
@@ -162,3 +165,28 @@ export function generateTestNumericInputWidget(): NumericInputWidget {
...
+        widgets,
+    };
+}
+
+export function getAnswerlessItem<T extends keyof PerseusWidgetTypes>(
+    widgetType: T,
+    options: PerseusWidgetTypes[T]["options"],
+): PerseusRenderer {
+    return splitPerseusItem(getAnswerfulItem(widgetType, options));
+}
```
Myranae, line 192: I was trying out Cursor with generalizing these functions, but it is basically what I was trying to accomplish. It helped a bit with the TS stuff, so if it looks off, please let me know :)

packages/perseus/src/widgets/matrix/matrix.tsx
```diff
@@ -537,16 +533,27 @@ class Matrix extends React.Component<Props, State> implements Widget {
     }
 }
 
-const propTransform: (arg1: any) => any = (editorProps) => {
+// NOTE: Answers here in RenderProps does not contain answers from the rubric.
+// It's an empty array of the same length and depth used to render the matrix.
```
Myranae, line 537: Let me know if this comment doesn't seem helpful. I always take a second look when I see the word "answers", so wanted to clarify.

packages/perseus/src/widgets/matrix/matrix.tsx
```diff
@@ -537,16 +533,27 @@ class Matrix extends React.Component<Props, State> implements Widget {
...
+type RenderProps = MatrixPublicWidgetOptions & {
+    answers: ReadonlyArray<ReadonlyArray<string>>;
+};
```
Myranae, line 540: Not sure if this should be up with the other props. ExternalProps doesn't have the same optional properties as MatrixPublicWidgetOptions, so I didn't think I should add RenderProps to it. The typing for answers is a little strange; transform converts it to a string, so the matrix gets rendered with an array of strings, but there's other logic using it as numbers. The matrix editor creates the array with numbers, but then all the interaction is with strings, until you get back to scoring.

packages/perseus/src/widgets/matrix/matrix.tsx
```diff
@@ -537,16 +533,27 @@ class Matrix extends React.Component<Props, State> implements Widget {
...
+        widgetOptions,
+        "matrixBoardSize",
+        "prefix",
+        "suffix",
+    );
+    return _.extend(widgetOptions, {
         answers: blankAnswers,
     });
-};
+}
```
Myranae, line 556: The changes here include change to function syntax and renaming editorProps to widgetOptions. Shouldn't have any changes to logic.

packages/perseus/src/widgets/matrix/matrix.test.ts
```diff
@@ -91,4 +97,58 @@ describe("matrix widget", () => {
...
+        let userEvent: UserEvent;
+        beforeEach(() => {
+            userEvent = userEventLib.setup({
+                advanceTimers: jest.advanceTimersByTime,
+            });
+
+            jest.spyOn(Dependencies, "getDependencies").mockReturnValue(
+                testDependencies,
+            );
+        });
```
Myranae, line 115: I just noticed this is available at the top level describe. I can take this chunk out of the nested describe, right?

packages/perseus/src/widgets/matrix/matrix.tsx
```diff
@@ -537,16 +533,27 @@ class Matrix extends React.Component<Props, State> implements Widget {
...
-    return _.extend(editorProps, {
+    widgetOptions = _.pick(
+        widgetOptions,
+        "matrixBoardSize",
+        "prefix",
+        "suffix",
+    );
+    return _.extend(widgetOptions, {
         answers: blankAnswers,
     });
```
benchristel, line 555: As long as you're refactoring this, could we move away from underscore?

```suggestion
    const {matrixBoardSize, prefix, suffix} = widgetOptions;
    return {
        matrixBoardSize,
        prefix,
        suffix,
        answers: blankAnswers,
    }
```

packages/perseus/src/widgets/matrix/matrix.tsx
benchristel: I belive the difference between `RenderProps` and `ExternalProps` is that `RenderProps` is supposed to be what appears inside the component (after `defaultProps` have been applied) and `ExternalProps` is what callers of this component need to pass in. But TBH I am a little confused about why we need both types here. It seems like maybe we could define `type RenderProps = ExternalProps` for this component? I will look into this more after lunch.

benchristel (approved):
LGTM! I left some suggestions inline, but nothing blocking.

packages/perseus/src/util/test-utils.ts
handeyeco: I think it's fine to use AI to some extent as long as:

1. You understand the code it generates enough to explain it to others
2. The code is well commented (so you don't have to explain it to others!)
3. We have ample testing around it

I personally have trouble grokking the TS in this PR (that's on me to _some_ extent) but it would certainly help if there were comments/tests.

packages/perseus/src/widgets/matrix/matrix.tsx
```diff
@@ -537,16 +533,27 @@ class Matrix extends React.Component<Props, State> implements Widget {
...
+// It's an empty array of the same length and depth used to render the matrix.
+type RenderProps = MatrixPublicWidgetOptions & {
+    answers: ReadonlyArray<ReadonlyArray<string>>;
+};
+
+function transform(widgetOptions: MatrixPublicWidgetOptions): RenderProps {
     // Remove answers before passing to widget
-    const blankAnswers = _(editorProps.matrixBoardSize[0]).times(function () {
-        return stringArrayOfSize(editorProps.matrixBoardSize[1]);
+    const blankAnswers = _(widgetOptions.matrixBoardSize[0]).times(function () {
```
handeyeco, line 544: I made a helper for this in my PR: https://github.com/Khan/perseus/pull/2396/files#diff-5d41841bf6867bd9e265d0ff7631b205e25c0d4ea16386b1bce446e324724575

You could rebase on top of mine or wait until I land into `main` (maybe tomorrow since I don't understand what's happening with releases right now)

packages/perseus/src/widgets/matrix/matrix.tsx
handeyeco: tbh I was very confused by props in `Table` too. I think for some of these older widgets, the architecture wasn't as clear.

There's some info here: https://github.com/Khan/perseus/blob/main/docs/architecture.md

I _think_ it's:

- `PerseusMatrixWidgetOptions`: all the data we store about the specific widget
- `MatrixPublicWidgetOptions`: all the answerless data for the pre-scoring rendering
- `RenderProps`: the output of the `transform` which takes `PerseusMatrixWidgetOptions` (and in the future `MatrixPublicWidgetOptions`) as input
- `ExternalProps`: wraps `RenderProps` with additional stuff from `WidgetOptions`*
- `DefaultProps`: is the defaults for things that are optional in `ExternalProps`
- `Props`: is the combination of `ExternalProps` and `DefaultProps` - it's everything in `ExternalProps` except optional fields are filled in from `DefaultProps`

---

*`WidgetOptions` is its own thing:

``` TS
/**
 * The type representing the common structure of all widget's options. The
 * `Options` generic type represents the widget-specific option data.
 */
export type WidgetOptions<Type extends string, Options> = {
    // The "type" of widget which will define what the Options field looks like
    type: Type;
    // Whether this widget is displayed with the values and is immutable.  For display only
    static?: boolean;
    // Whether a widget is scored.  Usually true except for IFrame widgets (deprecated)
    // Default: true
    graded?: boolean;
    // The HTML alignment of the widget.  "default" or "block"
    alignment?: string;
    // Options specific to the type field of the widget.  See Perseus*WidgetOptions for more details
    options: Options;
    // Only used by interactive child widgets (line, point, etc) to identify the components
    key?: number | null;
    // The version of the widget data spec.  Used to differentiate between newer and older content data.
    version?: Version;
};
```

packages/perseus/src/widgets/matrix/matrix.tsx
handeyeco: Yeah, I think we could rename this from `answers`. Without looking, I suspect it's also mutate with user input like Table is? Hm...but like Table has `rows`/`columns`, Matrix has `matrixBoardSize`. So I bet it's getting dimensions from one and storing user input in another?

handeyeco (commented):
I don't feel super comfortable approving until the AI generated code has tests/comments/etc, but I also don't want to block you. So I'll just leave some comments.

packages/perseus/src/util/test-utils.ts
Myranae: Added comments and tests! Let me know what you think :)

packages/perseus/src/widgets/matrix/matrix.tsx
Myranae: It's more along the lines of what @handeyeco describes. Answers is the box that holds the user's answers. It uses a function called `getMatrixSize` to iterate over the matrix and get the size based on which rows and cells actually contain data. I think this can just be changed to use `matrixBoardSize` since rows and cells are not supposed to be null/empty... but I'm not sure. I made the change and there is a snapshot change. I'll investigate a bit more tomorrow and also try renaming `answers` to `userInput` or something, at least in the places where it represents that. 🤔

packages/perseus/src/util/test-utils.ts
```diff
@@ -162,3 +165,28 @@ export function generateTestNumericInputWidget(): NumericInputWidget {
...
+        type: widgetType,
+        options,
+    };
+    const widgets = {};
+    widgets[widgetName] = widget;
+    return {
+        content: `[[☃ ${widgetName}]]`,
+        images: {},
+        widgets,
+    };
```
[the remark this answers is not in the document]
Myranae: This leads to a TS error that is a bit hard to get around. Seems to be having trouble with the map types we have. I'll just leave it as it is since this is a test util.

packages/perseus/src/widgets/categorizer/categorizer.test.ts
```diff
@@ -165,38 +167,25 @@ describe("categorizer widget", () => {
...
-                        },
-                    },
-                },
-            };
-        }
-
-        function getAnswerlessItem(): PerseusRenderer {
-            return splitPerseusItem(getAnswerfulItem());
-        }
+        const options = {
```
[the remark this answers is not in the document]
Myranae: Fixing!

packages/perseus/src/widgets/matrix/matrix.tsx
Myranae: Updated it to use your helper. :)

packages/perseus/src/widgets/matrix/matrix.tsx
Myranae: Alright, I didn't make any changes to ExternalProps or Props. Just added in the RenderProps type. I did notice that getUserInput returns an object with an answers property, but changing the naming there requires a lot of changes, so if that's important, might do that in another PR. Otherwise, will be landing this tomorrow :)

PR #2390, opened 2025-04-14
remove answer from PerseusItem

handeyeco:
## Summary:
<Add your summary here>

Issue: LEMS-XXXX

## Test plan:

### DIFF
--- a/packages/perseus-core/src/data-schema.ts
+++ b/packages/perseus-core/src/data-schema.ts
@@ -214,10 +214,6 @@ export type PerseusItem = {
      * @deprecated Not used.
      */
     itemDataVersion: any;
-    /**
-     * @deprecated Superseded by per-widget answers.
-     */
-    answer: any;
 };
 
 /**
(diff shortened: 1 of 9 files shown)


packages/perseus-core/src/data-schema.ts
```diff
@@ -214,10 +214,6 @@ export type PerseusItem = {
      * @deprecated Not used.
      */
     itemDataVersion: any;
```
handeyeco, line 216: Wondering if I should just remove `itemDataVersion` too... 🤔

handeyeco: https://github.com/Khan/perseus/pull/2391

.changeset/serious-jeans-greet.md
```diff
@@ -0,0 +1,2 @@
+---
```
handeyeco, line 1: This will be collapsed into: https://github.com/Khan/perseus/pull/2382

jeremywiebe (approved):
Perseus-side of this is good. We'll need to make sure CP and webapp are good.

PR #2392, opened 2025-04-14
Update APIOptions flags type to make it more explicit to Perseus

ivyolamit:
## Summary:
Update APIOptions flags type to make it more explicit to Perseus

Issue: LEMS-3039

## Test plan:

### DIFF
--- a/packages/perseus-core/src/feature-flags.ts
+++ b/packages/perseus-core/src/feature-flags.ts
@@ -8,6 +8,6 @@
  * 3. When no active feature flag is available value will be:
  *    export const PerseusFeatureFlags = [];
  */
-const PerseusFeatureFlags = ["new-radio-widget"];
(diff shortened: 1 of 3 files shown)


SonicScrewdriver (approved):
Looks great to me, thanks for the updates!

packages/perseus-core/src/feature-flags.ts
```diff
@@ -8,6 +8,6 @@
  * 3. When no active feature flag is available value will be:
  *    export const PerseusFeatureFlags = [];
```
jeremywiebe, line 9: Can you add the `as const` to the example just so somebody doesn't accidentally remove it when they're following the instructions.

jeremywiebe (approved):
Great. Thanks for adjusting this!

PR #2396, opened 2025-04-15
Answerless Table

handeyeco:
## Summary:
I probably could have kept cleaning this widget up, but...
- Table is `hidden`
- This display name for Table is: "Table (deprecated - use markdown table instead)"
- I did more than what I needed for SSS

The only thing it was really using `PerseusTableWidgetOptions.answers` for was to determine the number of rows/columns, but `PerseusTableWidgetOptions` has `rows: number` and `columns: number` so I just used those. Before switching things over however, I wrote some unit tests for current behavior and then made sure those tests passed with answerless data.

Issue: LEMS-2982

### DIFF
--- a/packages/perseus/src/util.test.ts
+++ b/packages/perseus/src/util.test.ts
@@ -14,7 +14,7 @@ describe("stringArrayOfSize", () => {
 
 describe("stringArrayOfSize2D", () => {
     it("makes a 2D array of strings", () => {
-        expect(Util.stringArrayOfSize2D(2, 4)).toEqual([
+        expect(Util.stringArrayOfSize2D({rows: 2, columns: 4})).toEqual([
             ["", "", "", ""],
             ["", "", "", ""],
         ]);
--- a/packages/perseus/src/util.test.ts
+++ b/packages/perseus/src/util.test.ts
@@ -5,3 +5,18 @@ describe("firstNumericalParse", () => {
         expect(Util.firstNumericalParse("6/8")).toBe(0.75);
     });
 });
+
+describe("stringArrayOfSize", () => {
+    it("makes an array of strings", () => {
+        expect(Util.stringArrayOfSize(2)).toEqual(["", ""]);
+    });
+});
+
+describe("stringArrayOfSize2D", () => {
+    it("makes a 2D array of strings", () => {
+        expect(Util.stringArrayOfSize2D(2, 4)).toEqual([
+            ["", "", "", ""],
+            ["", "", "", ""],
+        ]);
+    });
+});
--- a/packages/perseus/src/util.ts
+++ b/packages/perseus/src/util.ts
@@ -146,9 +146,15 @@ function firstNumericalParse(text: string): ParsedValue | null | undefined {
 }
 
 function stringArrayOfSize(size: number): ReadonlyArray<string> {
-    return _(size).times(function () {
-        return "";
-    });
+    return Array(size).fill("");
+}
+
+function stringArrayOfSize2D(
+    rows: number,
+    columns: number,
+): ReadonlyArray<ReadonlyArray<string>> {
+    const rowArr = stringArrayOfSize(rows);
+    return rowArr.map(() => stringArrayOfSize(columns));
 }
 
 /**
@@ -593,6 +599,7 @@ const Util = {
     split,
     firstNumericalParse,
     stringArrayOfSize,
+    stringArrayOfSize2D,
     gridDimensionConfig,
     getGridStep,
     snapStepFromGridStep,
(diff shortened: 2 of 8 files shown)


packages/perseus/src/util.test.ts
```diff
@@ -5,3 +5,18 @@ describe("firstNumericalParse", () => {
...
+
+describe("stringArrayOfSize", () => {
+    it("makes an array of strings", () => {
+        expect(Util.stringArrayOfSize(2)).toEqual(["", ""]);
+    });
+});
+
+describe("stringArrayOfSize2D", () => {
+    it("makes a 2D array of strings", () => {
+        expect(Util.stringArrayOfSize2D(2, 4)).toEqual([
```
benchristel, line 17: It seems slightly weird to have the number of rows be the first argument, since coordinates are normally `(x, y)`. I might prefer to use an object to name the arguments: `Util.stringArrayOfSize({rows: 2, cols: 4})`.

Alternatively, an API like this might be more transparent and reusable:

```ts
times(2, () => times(4, () => ""))
```

packages/perseus/src/util.ts
```diff
@@ -146,9 +146,15 @@ function firstNumericalParse(text: string): ParsedValue | null | undefined {
...
-        return "";
-    });
+    return Array(size).fill("");
+}
+
+function stringArrayOfSize2D(
+    rows: number,
+    columns: number,
+): ReadonlyArray<ReadonlyArray<string>> {
+    const rowArr = stringArrayOfSize(rows);
```
benchristel, line 156: You can do
```suggestion
    const rowArr = Array(rows).fill();
```

which might be slightly clearer (since you don't need the strings)

packages/perseus/src/widgets/table/table.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+        userEvent = userEventLib.setup({
+            advanceTimers: jest.advanceTimersByTime,
+        });
+    });
+
+    [
+        {optionsMode: "answerful", renderItem: getFullItem()},
+        {optionsMode: "answerless", renderItem: getSplitItem()},
+    ].forEach(({optionsMode, renderItem}) => {
+        it(`${optionsMode}: renders`, () => {
```
benchristel, line 35: Using template strings and `forEach` instead of `it.each` here might prevent Jest from finding these tests by name. I think that will break IDE integration with Jest — I.e. you won't be able to run these tests individually from your editor.

benchristel (approved):
LGTM. My paranoid side is worried that the `rows` and `columns` props might not always agree with the dimensions of `answers`.

I wonder if it would be simpler to change `getTablePublicWidgetOptions` so it replaces every element of `answers` with an empty string, instead of removing `answers` entirely. Would that work?

packages/perseus/src/util.ts
handeyeco: `fill` requires an argument. I could do `.fill(null)` (which I'm not sure is clearer than `.fill("")`) but I like leveraging the existing helper.

packages/perseus/src/widgets/table/table.test.tsx
handeyeco: I originally tried `it.each` but it wouldn't let me have separate test cases within it. Seems like `describe.each` works though so I switched to that.

packages/perseus/src/widgets/table/table.tsx
```diff
@@ -302,17 +312,16 @@ class Table extends React.Component<Props> implements Widget {
...
     // Remove answers before passing to widget
-    const rows = editorProps.answers.length;
-    const columns = editorProps.answers[0].length;
-    const blankAnswers = _(rows).times(function () {
-        return Util.stringArrayOfSize(columns);
-    });
-    return _.extend({}, editorProps, {
+    const rows = options.rows;
+    const columns = options.columns;
+    const blankAnswers = Util.stringArrayOfSize2D({rows, columns});
```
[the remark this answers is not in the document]
benchristel: I would lean the other way: given that Table is deprecated and we're not going to be improving it, I would lean toward making as few changes as possible to reduce the risk that we break legacy content.

I'm good with whatever you decide, though.

packages/perseus/src/widgets/table/table.stories.tsx
```diff
@@ -0,0 +1,32 @@
...
+};
+export default meta;
+
+type Story = StoryObj<typeof ServerItemRendererWithDebugUI>;
+
+const tableItem: PerseusItem = generateTestPerseusItem({
+    question: generateTableRenderer(),
+});
+
+export const AnswerfulTable: Story = {
```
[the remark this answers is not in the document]
handeyeco: > All stories should be answerful, unless we are specifically making one to test answerless rendering, right?

In the future, all Stories will probably start Answerless and become Answerful based on interaction (just like Webapp). So eventually there won't be specific Answerless/Answerful stories. Just splitting them right now so that we stay out of the way of other Perseus devs while we work.

PR #2397, opened 2025-04-15
When parsing Perseus JSON, optionalize object properties that can be undefined

benchristel:
## Summary:
Previously, it was not possible for the return types of our parsers to have
optional properties, like `{foo?: string}`. This meant that the data-schema types
that used that syntax were not mutually assignable with the types inferred from their parser.

E.g. if you have a parser like this:

```ts
const foo = object({foo: optional(string)});
```

The type `ParsedValue<typeof foo>` is equivalent to `{foo: string |
undefined}`. It is assignable to `{foo?: string}`, but `{foo?: string}`
is not assignable to `ParsedValue<typeof foo>`.

This lack of mutual assignability made it hard to write robust typetests for
the parsers, which in turn made it hard to ensure that parsers would be kept
up to date when changes were made to data-schema.

For example, consider the case of adding a variant to a union type in data-schema.

```ts
type MyUnion = "a" | "b" | "new-variant";
```

Suppose the parser for this type looks like this:

```ts
const parseMyUnion = enumeration("a", "b");
```

You won't get any type errors after adding `"new-variant"` to the type, because
`"a" | "b"` is assignable to `"a" | "b" | "new-variant"`. Yet the parser will
reject `"new-variant"`, which is certainly not what you want. In order to catch
this bug statically, you need typetests that ensure the data-schema types can
be assigned to the parser return types.

This PR takes a first step toward making that possible, by standardizing on the
syntax `{foo?: T}` for optional properties, rather than `{foo: T | undefined}`.

Issue: LEMS-3048

## Test plan:

`pnpm tsc`

### DIFF
--- a/packages/perseus-core/src/parse-perseus-json/general-purpose-parsers/object-types.ts
+++ b/packages/perseus-core/src/parse-perseus-json/general-purpose-parsers/object-types.ts
@@ -1,0 +1,17 @@
+/**
+ * These types were inspired by:
+ * - https://stackoverflow.com/questions/56863875/typescript-how-do-you-filter-a-types-properties-to-those-of-a-certain-type
+ *   (public domain)
+ * - https://github.com/ianstormtaylor/superstruct/blob/e414c8afd3b69f6bc0173b8ee25f71d8e5694f01/src/utils.ts
+ *   (MIT license)
+ */
+
+export type OptionalizeProperties<T extends object> = Omit<
+    T,
+    OptionalKeysOf<T>
+> &
+    Partial<Pick<T, OptionalKeysOf<T>>>;
+
+export type OptionalKeysOf<T extends object> = {
+    [K in keyof T]-?: undefined extends T[K] ? K : never;
+}[keyof T];
--- a/packages/perseus-core/src/parse-perseus-json/general-purpose-parsers/object.ts
+++ b/packages/perseus-core/src/parse-perseus-json/general-purpose-parsers/object.ts
@@ -2,13 +2,20 @@ import {failure, isSuccess} from "../result";
 
 import {isObject} from "./is-object";
 
+import type {OptionalizeProperties} from "./object-types";
 import type {Mismatch, ParsedValue, Parser} from "../parser-types";
 
 type ObjectSchema = Record<keyof any, Parser<any>>;
 
-export function object<S extends ObjectSchema>(
+export function objectWithAllPropertiesRequired<S extends ObjectSchema>(
     schema: S,
 ): Parser<{[K in keyof S]: ParsedValue<S[K]>}> {
+    return object(schema) as any;
+}
+
+export function object<S extends ObjectSchema>(
+    schema: S,
+): Parser<OptionalizeProperties<{[K in keyof S]: ParsedValue<S[K]>}>> {
     return (rawValue, ctx) => {
         if (!isObject(rawValue)) {
             return ctx.failure("object", rawValue);
--- a/packages/perseus-core/src/parse-perseus-json/perseus-parsers/widget.ts
+++ b/packages/perseus-core/src/parse-perseus-json/perseus-parsers/widget.ts
@@ -3,6 +3,7 @@ import {
     nullable,
     number,
     object,
+    objectWithAllPropertiesRequired,
     optional,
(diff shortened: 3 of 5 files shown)


packages/perseus-core/src/parse-perseus-json/general-purpose-parsers/object.ts
```diff
@@ -2,13 +2,20 @@ import {failure, isSuccess} from "../result";
...
+import type {OptionalizeProperties} from "./object-types";
 import type {Mismatch, ParsedValue, Parser} from "../parser-types";
 
 type ObjectSchema = Record<keyof any, Parser<any>>;
 
-export function object<S extends ObjectSchema>(
+export function objectWithAllPropertiesRequired<S extends ObjectSchema>(
     schema: S,
 ): Parser<{[K in keyof S]: ParsedValue<S[K]>}> {
+    return object(schema) as any;
```
benchristel, line 13: This seems weird and makes me a little nervous, but I think it's safe. By equating `objectWithAllPropertiesRequired` and `object`, we are essentially saying that we don't care about the difference between `{foo: string | undefined}` and `{foo?: string}`. The difference according to TS is that `{foo: string | undefined}` requires `foo` to be present, so `{}` is not assignable to it.

packages/perseus-core/src/parse-perseus-json/general-purpose-parsers/object-types.ts
```diff
@@ -0,0 +1,17 @@
...
+ *   (public domain)
+ * - https://github.com/ianstormtaylor/superstruct/blob/e414c8afd3b69f6bc0173b8ee25f71d8e5694f01/src/utils.ts
+ *   (MIT license)
+ */
+
+export type OptionalizeProperties<T extends object> = Omit<
+    T,
+    OptionalKeysOf<T>
+> &
+    Partial<Pick<T, OptionalKeysOf<T>>>;
```
benchristel, line 13: Ugh, I hate Prettier's whitespace choices here.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/widget.ts
```diff
@@ -33,13 +34,13 @@ export function parseWidget<Type extends string, Options>(
...
     Type extends string,
-    Options,
+    Options extends object,
 >(
     parseVersion: Parser<Version>,
     parseType: Parser<Type>,
     parseOptions: Parser<Options>,
 ): Parser<WidgetOptions<Type, Options>> {
-    return object({
+    return objectWithAllPropertiesRequired({
```
benchristel, line 43: We have to use `objectWithAllPropertiesRequired` because the type parameters `Type` and `Options` interact poorly with `OptionalizeProperties`. Since the concrete types of `Type` and `Options` aren't known, I think TS gives up and assumes they're `any`, which means `undefined` is assignable to them. This causes them to be incorrectly treated as optional. I don't know why TS doesn't look at the `extends` constraints for these type parameters, but it seems not to.

Note that this means we will have to use the `foo: string | undefined` syntax for the properties of `WidgetOptions`. Hopefully that's not too big a deal.

jeremywiebe: I was trying to figure this out as I read the PR.  If I make incremental types, I see that the return type seems to be missing the `"type"` key entirely. 🤔  

image

jeremywiebe (approved):

packages/perseus-core/src/parse-perseus-json/general-purpose-parsers/object.ts
```diff
@@ -2,13 +2,20 @@ import {failure, isSuccess} from "../result";
 
 import {isObject} from "./is-object";
 
+import type {OptionalizeProperties} from "./object-types";
 import type {Mismatch, ParsedValue, Parser} from "../parser-types";
 
 type ObjectSchema = Record<keyof any, Parser<any>>;
 
-export function object<S extends ObjectSchema>(
+export function objectWithAllPropertiesRequired<S extends ObjectSchema>(
```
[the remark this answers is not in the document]
benchristel: Right!

PR #2398, opened 2025-04-15
Remove readonly types from data-schema.ts

benchristel:
## Summary:
See the [slack discussion] for more details. TL;DR:

- The `readonly` modifier in TypeScript simply removes mutating operations like
  `Array.push` and `obj.foo = 1` property assignment from a type.
- As such, it's best to think of it as an application of the [Interface
  Segregation Principle].
- That implies that `readonly` types are most appropriate in function parameters.
  There, they act as a promise that the function won't mutate its arguments.
- They may also be appropriate in return types **if** the function returns a
  frozen object or array, because frozen objects actually disallow mutation at
  runtime.
- In general, readonly types are **not** appropriate in the return types of
  functions, because they unnecessarily restrict the caller from mutating an
  object to which no one else holds a reference.
- That means readonly types are probably **not** appropriate for shared data types
  which may be used in _either_ parameters or return values.
- We weren't using readonly types consistently in `data-schema.ts` to begin with.
  We used them for arrays, but not for tuples or objects.
- For the sake of consistency, and to allow our `data-schema` types to be reused
  easily in different contexts, this PR replaces `ReadonlyArray` with `Array` in
  `data-schema.ts` and fixes the type errors that cascade out of that.

[slack discussion]: https://khanacademy.slack.com/archives/C01AZ9H8TTQ/p1744325813358959
[Interface Segregation Principle]: https://en.wikipedia.org/wiki/Interface_segregation_principle

Issue: LEMS-XXXX

## Test plan:

`pnpm tsc`

### DIFF
--- a/packages/perseus-core/src/data-schema.ts
+++ b/packages/perseus-core/src/data-schema.ts
@@ -56,8 +56,8 @@ export type ShowSolutions = "all" | "selected" | "none";
  *
  * ```
  * interface DummyRegistry {
- *     categorizer: { categories: ReadonlyArray<string> };
- *     dropdown: { choices: ReadonlyArray<string> }:
+ *     categorizer: { categories: string[] };
+ *     dropdown: { choices: string[] }:
  * }
  * ```
  *
@@ -71,8 +71,8 @@ export type ShowSolutions = "all" | "selected" | "none";
  *
  * ```
  * type DummyMap = {
- *     `categorizer ${number}`: { categories: ReadonlyArray<string> };
- *     `dropdown ${number}`: { choices: ReadonlyArray<string> };
+ *     `categorizer ${number}`: { categories: string[] };
+ *     `dropdown ${number}`: { choices: string[] };
  * }
  * ```
  *
@@ -206,7 +206,7 @@ export type PerseusItem = {
     // The details of the question being asked to the user.
     question: PerseusRenderer;
     // A collection of hints to be offered to the user that support answering the question.
-    hints: ReadonlyArray<Hint>;
+    hints: Hint[];
     // Details about the tools the user might need to answer the question
     answerArea: PerseusAnswerArea | null | undefined;
 };
@@ -215,7 +215,7 @@ export type PerseusItem = {
  * A "PerseusArticle" is an item that is meant to be rendered as an article.
  * This item is never scored and is rendered by the `ArticleRenderer`.
  */
-export type PerseusArticle = PerseusRenderer | ReadonlyArray<PerseusRenderer>;
+export type PerseusArticle = PerseusRenderer | PerseusRenderer[];
 
 export type Version = {
     // The major part of the version
@@ -427,15 +427,15 @@ export type AxisLabelLocation = "onAxis" | "alongEdge";
 
 export type PerseusCategorizerWidgetOptions = {
     // Translatable text; a list of items to categorize. e.g. ["banana", "yellow", "apple", "purple", "shirt"]
-    items: ReadonlyArray<string>;
+    items: string[];
     // Translatable text; a list of categories. e.g. ["fruits", "colors", "clothing"]
-    categories: ReadonlyArray<string>;
+    categories: string[];
     // Whether the items should be randemized
     randomizeItems: boolean;
     // Whether this widget is displayed with the results and immutable
     static: boolean;
     // The correct answers where index relates to the items and value relates to the category.  e.g. [0, 1, 0, 1, 2]
-    values: ReadonlyArray<number>;
+    values: number[];
     // Whether we should highlight i18n linter errors found on this widget
     highlightLint?: boolean;
     // Internal editor configuration. Can be ignored by consumers.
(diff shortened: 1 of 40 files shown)


packages/perseus-core/src/data-schema.ts
```diff
@@ -642,7 +642,7 @@ export type GrapherAnswerTypes =
       };
 
 export type PerseusGrapherWidgetOptions = {
-    availableTypes: ReadonlyArray<
+    availableTypes: Array<
```
jeremywiebe, line 645: For the sake of consistency, should we write these as:

```typescript 
    availableTypes: (
        | "absolute_value"
        | "exponential"
        | "linear"
        | "logarithm"
        | "quadratic"
        | "sinusoid"
        | "tangent"
     )[]
```

packages/perseus-editor/src/widgets/label-image-editor.tsx
```diff
@@ -25,7 +25,7 @@ import type {
 
 type Props = {
     // List of answer choices to label question image with.
-    choices: ReadonlyArray<string>;
```
jeremywiebe, line 28: This is in a Props type... won't we want to move our React component props towards Readonly-ness?

packages/perseus/src/widgets/expression/expression.test.tsx
```diff
@@ -277,12 +278,10 @@ describe("Expression Widget", function () {
 
         it("should return undefined if rubric.value is null/undefined", () => {
             // Arrange
-            const rubric = {
+            const rubric: PerseusExpressionRubric = {
```
jeremywiebe, line 281: Is there a need to explicitly type `rubric` here?  Doesn't the call to `getOneCorrectAnswerFromRubric()` type check the input value?

jeremywiebe (approved):
Overall looks good.  One place where I'm not sure we want to embrace mutable arrays is in React props. I suspect that was one place where things "rippled out".  Perhaps we can address that in a future PR?  I wonder if we could just wrap all of the final Props types with `Readonly<>` where they're used in a React component declaration.

packages/perseus-core/src/data-schema.ts
benchristel: I used `Array` here because I don't like holding readers in suspense unnecessarily. The problem with `(` is that when you see it, you don't know what's on the other end, which requires you to hold more in memory (and identify the matching `)`) as you're mentally parsing the code. By contrast, when you see `Array<` you know to expect exactly one type argument, and you know what that argument means.

Possibly relevant:
- https://github.com/benchristel/benchristel.github.io/wiki/FunctionalPipeline
- https://github.com/benchristel/benchristel.github.io/wiki/LongBranchLast

packages/perseus-editor/src/widgets/label-image-editor.tsx
benchristel: The problem, IIRC, is that `choices` gets passed to an `onChange` callback that expects `data-schema` types.

We should probably be making a copy the data before passing it to the callback. Then we could use readonly types here.

Do you think I should make that change as part of this PR?

packages/perseus/src/widgets/expression/expression.test.tsx
benchristel: If `rubric` doesn't have a type declaration, TS will infer the types of `answerForms` and `functions` to be `never[]`. That prevents them from being passed to `getOneCorrectAnswerFromRubric`. Previously `as const` was used to make the types `readonly never[]`, which is assignable to any readonly array type. But since `getOneCorrectAnswerFromRubric` now requires read-write arrays, that's no longer an option.

packages/perseus-editor/src/widgets/label-image-editor.tsx
jeremywiebe: The `onChange` callback types are quite a mess currently.  We defined a single `onChange` in `WidgetExports<T>` because we pass it down into each widget. And so `onChange` is the union of all types of data all widgets pass up (https://github.com/Khan/perseus/blob/c922913b6289d34bf0fecb7bef96c0a6be45d8e7/packages/perseus/src/types.ts#L583).  

To your question, I feel like this is another area that we could address across the board as we figure out how we want to handle React props ideally.

benchristel: I've been thinking about this a bit more. IMO adding `readonly` to React prop types seems like it won't add a lot of value. React's philosophy of immutable props and state is already ingrained in every React developer. At the same time, various libraries (looking at you, Mafs) use mutable types in their APIs, even when they do not mutate the values. That would cause an impedance mismatch with our code if we switched to using `readonly` type everywhere.

I think we would get more value out of a linter that banned syntax like the following:

```js
a[b] = c
a.b = c
Object.assign(a, ...)
array.push()
array.splice()
array.sort()
```

With exceptions for OOP:

```js
this.a = b // ok
```

That would make it much harder to mutate values accidentally. Interop with libraries would still be easy.

jeremywiebe: So something like this? https://github.com/eslint-functional/eslint-plugin-functional/blob/main/docs/rules/immutable-data.md

I get your point about our code choices causing challenges in integration. I still do like the idea of making our expectations explicit. In the places we've marked things as readonly I haven't found the interop to be that difficult. 

But like I said, I think if we can tackle the mutation question in _some_ way, it's better than our current state. I know we've had mutation bugs in the past (I can't recall exactly where, sorry), so adding some helpful tooling sounds like a good approach.

PR #2403, opened 2025-04-17
Add tests to ensure Perseus JSON parsers are idempotent

benchristel:
## Summary:
I realized that the type of the `defaulted` parser encouraged non-idempotent
parsing. Essentially, a default value could be supplied that would cause a parse
error on a second pass through the parser. Example:

```ts
const parseMyObject = object({
  myField: defaulted(number, () => "bork"),
})
```

The inferred type `ParsedValue<typeof parseMyObject>` would be:

```ts
type MyObject = {myField: number | string}
```

However, the parser doesn't consider a myField value of type `string` to be valid!
This contradicts the types! You will get an error if you try to parse

```ts
parseMyObject({myField: "bork"}, ctx)
```

because the parser for `myField` only accepts a number, null, or undefined as
input.

This is clearly not how we want parsers to behave. Parsing should be idempotent.

This PR fixes the types of `defaulted` so we don't encourage non-idempotent
parsing. It also adds tests to verify that our existing parsers are idempotent
when run on example data.

Issue: LEMS-3055

## Test plan:

`pnpm test`

### DIFF
--- a/packages/perseus-core/src/parse-perseus-json/general-purpose-parsers/defaulted.ts
+++ b/packages/perseus-core/src/parse-perseus-json/general-purpose-parsers/defaulted.ts
@@ -2,10 +2,10 @@ import {success} from "../result";
 
 import type {Parser} from "../parser-types";
 
-export function defaulted<T, Default>(
+export function defaulted<T>(
     parser: Parser<T>,
-    fallback: (missingValue: null | undefined) => Default,
-): Parser<T | Default> {
+    fallback: (missingValue: null | undefined) => NoInfer<T>,
+): Parser<T> {
     return (rawValue, ctx) => {
         if (rawValue == null) {
             return success(fallback(rawValue));
--- a/packages/perseus-core/src/parse-perseus-json/perseus-parsers/hint.ts
+++ b/packages/perseus-core/src/parse-perseus-json/perseus-parsers/hint.ts
@@ -1,11 +1,17 @@
-import {any, boolean, object, string} from "../general-purpose-parsers";
+import {
+    any,
+    boolean,
+    object,
+    optional,
+    string,
+} from "../general-purpose-parsers";
 import {defaulted} from "../general-purpose-parsers/defaulted";
 
 import {parseImages} from "./images-map";
 import {parseWidgetsMap} from "./widgets-map";
 
 export const parseHint = object({
-    replace: defaulted(boolean, () => undefined),
+    replace: defaulted(optional(boolean), () => undefined),
     content: string,
     widgets: defaulted(parseWidgetsMap, () => ({})),
     images: parseImages,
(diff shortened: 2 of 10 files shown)


packages/perseus-core/src/parse-perseus-json/general-purpose-parsers/defaulted.ts
```diff
@@ -2,10 +2,10 @@ import {success} from "../result";
 
 import type {Parser} from "../parser-types";
 
-export function defaulted<T, Default>(
+export function defaulted<T>(
     parser: Parser<T>,
-    fallback: (missingValue: null | undefined) => Default,
-): Parser<T | Default> {
+    fallback: (missingValue: null | undefined) => NoInfer<T>,
```
jeremywiebe, line 7: TIL about `NoInfer<T>` (I read https://www.totaltypescript.com/noinfer).  

So this blocks the fallback's return value from being included in the type inference for `T`, right? So the return value's `T` will be only inferred from the `parser` type.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/hint.ts
```diff
@@ -1,11 +1,17 @@
...
+    string,
+} from "../general-purpose-parsers";
 import {defaulted} from "../general-purpose-parsers/defaulted";
 
 import {parseImages} from "./images-map";
 import {parseWidgetsMap} from "./widgets-map";
 
 export const parseHint = object({
-    replace: defaulted(boolean, () => undefined),
+    replace: defaulted(optional(boolean), () => undefined),
```
jeremywiebe, line 14: I'm not clear why you needed to wrap `optional()` here? Is it because the `replace` can be `undefined` and if we didn't wrap it in `optional()` here, we'd have a resulting type of `boolean` because the fallback is ignored for type inference?

packages/perseus-core/src/parse-perseus-json/regression-tests/parse-perseus-json-regression.test.ts
```diff
@@ -34,6 +37,23 @@ describe("parseAndMigratePerseusItem", () => {
             assertSuccess(result);
             expect(result.value).toMatchSnapshot();
         });
+
+        it("is not changed by a second pass through the parser", () => {
+            // This test ensures that the parser is idempotent, i.e. running it
+            // once is the same as running it many times. Idempotency is
+            // valuable because it means e.g. that if we run the parser on data
+            // before saving it to datastore, it won't be changed by being
```
jeremywiebe, line 45: 💯 💯

jeremywiebe (approved):

packages/perseus-core/src/parse-perseus-json/general-purpose-parsers/defaulted.ts
benchristel: Correct! FWIW writing this code made me happy because I think it's a _perfect_ place to use `NoInfer`.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/hint.ts
benchristel: Yep!

PR #2407, opened 2025-04-17
[SR] Prepare strings for translation - Add context to interactive graph strings

nishasy:
## Summary:
We held off on translations until now, beacuse the Interactive Graph strings were
still in flux. Now that they're finalized, we can send them in for translations.

I added context to all the strings that need to be translated so that
- it's clear for translators looking at the translation ticket what
  the context is for each individual string.
- it'll be easy for us to create the translation ticket in the first place.
  Since these are going to count as new strings, they'll trigger the
  action on the PR that will create a translation ticket for us that's
  pre-populated with all new strings.

NOTE: This PR is probably easier to review with the "Hide whitespace" option.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2660

## Test plan:
- I looked at the PR diff, and then did a page search for every string on the left side and confirmed it's the exact same on the right side to make sure that no strings accidentally got changed in this PR.
- We'll have to make sure that the strings are detected in the webapp PR so we can create a translation ticket with the new context.

### DIFF
--- a/packages/perseus/src/strings.ts
+++ b/packages/perseus/src/strings.ts
@@ -134,7 +134,6 @@ export type PerseusStrings = {
     simulationLoadFail: string;
     simulationLocaleWarning: string;
     selectAnAnswer: string;
-    // The following strings are used for interactive graph SR descriptions.
     srGraphInstructions: string;
     srUnlimitedGraphInstructions: string;
     addPoint: string;
@@ -505,7 +504,6 @@ export type PerseusStrings = {
         point2X: string;
         point2Y: string;
     }) => string;
-    // The above strings are used for interactive graph SR descriptions.
 };
 
 /**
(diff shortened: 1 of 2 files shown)


packages/perseus/src/strings.ts
```diff
@@ -675,133 +673,434 @@ export const strings = {
...
+            "Ray with endpoint %(point1X)s comma %(point1Y)s going through point %(point2X)s comma %(point2Y)s.",
+    },
+    srRayEndpoint: {
+        context:
+            "Aria label for the defining point of the Ray in the interactive graph widget that sits on the tail of the Ray.",
+        message: "Endpoint at %(x)s comma %(y)s.",
+    },
+    srRayTerminalPoint: {
+        context:
+            "Aria label for the defining point the Ray goes through in the interactive graph widget that determines the direction of the Ray.",
```
benchristel, line 934:
```suggestion
            "Aria label for the point that determines the direction of the Ray in the interactive graph widget. The ray passes through this point.",
```

packages/perseus/src/strings.ts
```diff
@@ -675,133 +673,434 @@ export const strings = {
...
+        message: "Through point at %(x)s comma %(y)s.",
+    },
+    srQuadraticGraph: {
+        context:
+            "Aria label for the container containing a Quadratic function in the interactive graph widget.",
+        message: "A parabola on a 4-quadrant coordinate plane.",
```
benchristel, line 940:
```suggestion
        message: "A parabola.",
```

Why mention the number of quadrants shown in the graph? I don't think this is actually correct; e.g. the graph might only show quadrant 1.

packages/perseus/src/strings.ts
```diff
@@ -675,133 +673,434 @@ export const strings = {
...
+        message: "A sinusoid function on a coordinate plane.",
+    },
+    srSinusoidRootPoint: {
+        context:
+            "Aria label for the Point defining the midline intersection of the Sinusoid function in the interactive graph widget.",
+        message: "Midline intersection at %(x)s comma %(y)s.",
+    },
+    srSinusoidMaxPoint: {
+        context:
+            "Aria label for the Point defining max/min of the Sinusoid function in the interactive graph widget when it's above 0.",
```
benchristel, line 1079:
```suggestion
            "Aria label for the Point defining the maximum of the Sinusoid function in the interactive graph widget.",
```

I don't think the maximum being > 0 is a condition for this message to be shown, right? It's whenever the point is a local maximum (rather than a local minimum, or neither a min nor a max).

packages/perseus/src/strings.ts
```diff
@@ -675,133 +673,434 @@ export const strings = {
...
+        message: "Maximum point at %(x)s comma %(y)s.",
+    },
+    srSinusoidMinPoint: {
+        context:
+            "Aria label for the Point defining the max/min of the Sinusoid function in the interactive graph widget when it's below 0.",
```
benchristel, line 1084:
```suggestion
            "Aria label for the Point defining the minimum of the Sinusoid function in the interactive graph widget.",
```

packages/perseus/src/strings.ts
```diff
@@ -675,133 +673,434 @@ export const strings = {
...
+        message: "Minimum point at %(x)s comma %(y)s.",
+    },
+    srSinusoidFlatPoint: {
+        context:
+            "Aria label for the Point defining the max/min of the Sinusoid function in the interactive graph widget when it's at 0.",
```
benchristel, line 1089:
```suggestion
            "Aria label for the Point defining the amplitude of the Sinusoid function in the interactive graph widget when the amplitude is 0.",
```

benchristel (commented):
Wow, this is a lot of strings! I found a few things that don't look quite right, but most of this LGTM.

packages/perseus/src/strings.ts
```diff
@@ -505,7 +504,6 @@ export type PerseusStrings = {
         point2X: string;
         point2Y: string;
     }) => string;
-    // The above strings are used for interactive graph SR descriptions.
```
SonicScrewdriver, line 508: I actually liked having this context, as the organization made it easier for our future selves to quickly find the chunk of strings! I'm not fussed either way, however as I know we have some great context on the strings themselves. :)

SonicScrewdriver (approved):
The contexts you've added to the strings are GREAT! Absolutely love it, thank you!

packages/perseus/src/strings.ts
nishasy: I think we added "4-quadrant" this time because so many of the related strings refer to the quadrant number that something is in. Even if the vertex doesn't show up on the graph, it'll read the quadrant that it's in.

It's a fair point though. Since this would require copy updates from a designer, I'm going to punt it to the phase 2 string updates.

PR #2408, opened 2025-04-21
Fix formatting of eslintrc file

Myranae:
## Summary:
I noticed recently that my IDE has not been autofixing lint errors on save. After asking Cursor, I tested its solution, and it allowed for autofixing. Confirmed the syntax with the eslint documentation: https://eslint.org/docs/latest/rules/no-restricted-imports#paths

## Test plan:
- Confirm linting works correctly
- Confirm all checks pass
- Confirm syntax follows documentation


packages/perseus/.eslintrc.js
```diff
@@ -17,8 +17,9 @@ module.exports = {
         // so feel free to update if you know what you're doing.
         "@typescript-eslint/no-restricted-imports": [
             "error",
-            "@khanacademy/perseus",
-            "@khanacademy/perseus-editor",
+            {
```
jeremywiebe, line 20: That looks to match what's documented here: https://eslint.org/docs/latest/rules/no-restricted-imports#paths

PR #2410, opened 2025-04-21
Bugfix - Redundant reset.css styling was causing font display issues

mark-fitzgerald:
## Summary:
Revert some of the changes in PR-1945 to ensure that webapp styling is not included in Perseus stying build. Styling changes from that PR are now localized to Storybook.

Issue: LEMS-3057

## Test plan:
- Verify prior Storybook changes are still in effect (taken from PR-1945)
   1. Go to http://localhost:6006/?path=/docs/perseus-components-number-input--docs
   1. Confirm that the inputs have rounded bordersa, and the "mini", "small", and "normal" stories show different sizes.
   1. Go to http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-ray
   1. Scroll to the start coordinates panels
   1. Use the dev tools to confirm that the width is 308px (previously 336px)
- Verify that CSS is not duplicated in Webapp
   1. Go to [a ZND with the changes](https://prod-znd-250421-markfitz-m01.khanacademy.org/)
   2. Ensure that the fonts look correct (Using Lato and not Times)
   3. In the inspector tool, ensure that there aren't any nested layers (like `shared.reset`) on any particular item


.storybook/global.less
```diff
@@ -0,0 +1,32 @@
+/**
+ * This file contains styles that are used in prod that ultimately
+ * get passed onto Perseus components. This file is imported in the
+ * .storybook/preview.js file so that the components can have the same
+ * styles as prod when viewed within Storybook.
```
jeremywiebe, line 5: Thanks for explaining this!

jeremywiebe (approved):

.storybook/global.less
mark-fitzgerald: I can't take credit for that - the thanks belong to @nishasy.

nishasy (approved):
Looks good to me. I also confirmed that the box-sizing and the perseus styles look right in storybook.

PR #2413, opened 2025-04-22
Update Wonder Blocks dependencies: Tokens as CSS variables

jandrade:
## Summary:

This PR switches Wonder Blocks to use CSS variables for theming instead of the
previous approach of using JS tokens. This change is part of a larger effort
to improve the theming capabilities of Wonder Blocks to make it easier to
customize the appearance of components.

The following tokens have been converted to CSS variables:

- `semanticColor`
- `sizing`
- `border`

Issue: "none"

## Test plan:

Verify that widgets and components that use WB are rendering correctly
and that the CSS variables are being applied as expected.

Screenshot 2025-04-22 at 1 10 19 PM


catandthemachines (approved):
LGTM!

PR #2414, opened 2025-04-22
ESLint config cleanup

jeremywiebe:
## Summary:

This PR cleans up a bunch of ESLint-related configuration. In it, we:

  * Remove (now) unnecessary configs from each package dir's `.eslintrc.js` (its now properly covered by the root `.eslintrc.js`)
  * Update `package.json` files to properly include dependencies that they use
  * Move some `.d.ts` files from the shared `types` folder into the vendored package folder (and then expose them using the `"types"` key in the `package.json file)

Issue: "none"

## Test plan:

`pnpm lint`
Open your favourite code editor and check if there aren't any invalid lint errors (also try adding an import into a production `.ts` file for a package that isn't in the package's dependencies list and see if the linter flags it)


dev/package.json
```diff
@@ -37,6 +37,11 @@
         "@khanacademy/wonder-blocks-tooltip": "catalog:",
         "@khanacademy/wonder-stuff-core": "catalog:",
         "@phosphor-icons/core": "catalog:",
-        "aphrodite": "catalog:"
+        "aphrodite": "catalog:",
+        "react": "catalog:",
+        "react-dom": "catalog:"
```
jeremywiebe, line 42: These dips are used in this area but weren't explicit. This helps ESLint out when checking extraneous deps.

packages/kas/.eslintrc.js
```diff
@@ -1,16 +1,7 @@
...
 
 module.exports = {
     rules: {
-        "import/no-extraneous-dependencies": [
-            "error",
-            {
-                packageDir: [__dirname, path.join(__dirname, "../../")],
-                includeTypes: true,
-            },
-        ],
```
jeremywiebe, line 13: This is a core part of this PR. We can get rid of this rule in each package's `.eslintrc.js`, making our linting setup simpler (I couldn't figure out how to get rid of the circular imports rules below, so this file needs to stay here still).

packages/perseus-editor/package.json
```diff
@@ -45,7 +45,8 @@
         "@khanacademy/perseus-utils": "workspace:*",
         "@khanacademy/pure-markdown": "workspace:*",
         "katex": "0.11.1",
-        "mafs": "^0.19.0"
+        "mafs": "^0.19.0",
+        "tiny-invariant": "catalog:"
```
jeremywiebe, line 49: More dependencies used by this package that weren't explicit.

vendor/jsdiff/package.json
```diff
@@ -2,5 +2,6 @@
   "name": "jsdiff",
   "private": true,
   "version": "1.0.0",
-  "main": "jsdiff.js"
+  "main": "jsdiff.js",
+  "types": "jsdiff.d.ts"
```
jeremywiebe, line 6: For both of these `vendor/` changes, I've moved the associated `.d.ts` file into the package's folder and defined the `"types"` key in the `package.json.  This allows use to provide types for a vendored package without them being in our shared `types/` folder (eventually, it'd be nice to get rid of that as it's a bandaid for packages that don't ship types and is symlinked to all package folders).

jeremywiebe:
@mark-fitzgerald @Myranae I've tagged you both for review as I have recently worked with you on ESLint tooling or problems.  Would you mind giving this branch a test in your environments and see if it behaves in your IDE, please?

Myranae:
Hmm, I ran `pnpm lint` and had no issues in GoLand. It looks like the `kas` package does not have `perseus-core` as a dependency, so I imported something from `perseus-core` into a file in the `kas` package, but the linter didn't flag it. Maybe I did that wrong?

jeremywiebe:
> Hmm, I ran `pnpm lint` and had no issues in GoLand. It looks like the `kas` package does not have `perseus-core` as a dependency, so I imported something from `perseus-core` into a file in the `kas` package, but the linter didn't flag it. Maybe I did that wrong?

Seems to work for me. 

image

mark-fitzgerald (approved):
Spring cleaning is so refreshing!

Thank you for the PR comments, also. So helpful!!

PR #2417, opened 2025-04-23
Answerless Group

handeyeco:
## Summary:
For `splitPerseusItem` Group needs to call part of the logic from `splitPerseusItem` since it's widget options are a PerseusRenderer. So I split out `splitPerseusRenderer` from `splitPerseusItem` and had `getGroupPublicWidgetOptions` call that. Unfortunately this caused a circular dependency issue that I resolved by wrapping the PerseusCore registration logic in a function (like what we do for Perseus).

Issue: LEMS-2986

## Test plan:
- Unit tests
- Answerless/Answerful stories
- This widget and the content that uses it is deprecated, so I'm not worried about it


packages/perseus-core/src/utils/split-perseus-item.ts
```diff
@@ -3,7 +3,39 @@ import _ from "underscore";
...
+    const upgradedWidgets = getUpgradedWidgetOptions(originalWidgets);
+    const splitWidgets = {};
+
+    for (const [id, widget] of Object.entries(upgradedWidgets)) {
+        if (widget.type === "group") {
+            splitWidgets[id] = {
+                ...widget,
+                options: splitPerseusRenderer(widget.options),
+            };
+        } else {
```
handeyeco, line 23: So I'm leaning toward calling this task a "Won't Do" and here's why:

1. Group is deprecated and the content is deprecated. I don't think learners ever see Group really. I pushed to remove it and [got push back](https://khanacademy.slack.com/archives/C04EA1QAKN0/p1745337743067339).
2. Originally my plan was to make a `getGroupPublicWidgetOptions` that called `splitPerseusRenderer` (since a Group is just a PerseusRenderer). This was pretty clean and out of the way, so I didn't mind doing this work even though it was for a widget no one would ever see. Unfortunately this resulted in an import circular dependency: "can't access lexical declaration before initialization."
3. After talking it through with Anakaren, I thought up the current suggestion - rather than Group having its own `getGroupPublicWidgetOptions` just have `splitPerseusRenderer` handle group widgets by calling itself when it encounters one. No circular imports! Tests pass! Everything's working! I still really don't like it though; having special logic in a high-level function for an unused widget feels really gross.

What I think we should do: **don't remove answers from Group widgets.** Since we're not sure we want to remove Group AND since learners don't see Group AND since Group is tricky to handle, let's just do nothing!

packages/perseus-core/src/utils/split-perseus-item.ts
```diff
@@ -3,18 +3,18 @@ import _ from "underscore";
...
  *
- * @param originalItem - the original, full Perseus item (which includes the rubric - aka answer data)
+ * @param original - the original, full PerseusRenderer (which includes the rubric - aka answer data)
  */
-export default function splitPerseusItem(
-    originalItem: PerseusItem,
-): PerseusItem {
-    const item = _.clone(originalItem);
-    const originalWidgets = item.question.widgets ?? {};
+export function splitPerseusRenderer(
```
benchristel, line 13: Could `splitPerseusRenderer` be in its own file? It's slightly strange for it to be exported from a file called `split-perseus-item.ts`. Moving it to its own file would also limit the blast radius of the dependency cycle with group-util.ts

packages/perseus-core/src/utils/split-perseus-item.ts
```diff
@@ -3,18 +3,18 @@ import _ from "underscore";
...
+    original: PerseusRenderer,
+): PerseusRenderer {
+    const clone = _.clone(original);
```
benchristel, line 16: It's not clear to me why we need to clone here, but I don't think we need to address that in this PR.

packages/perseus-core/src/utils/test-utils.ts
```diff
@@ -1,11 +1,19 @@
-import type {PerseusItem} from "../data-schema";
+import type {PerseusItem, PerseusRenderer} from "../data-schema";
+
+const genericPerseusRenderer: PerseusRenderer = {
```
benchristel, line 3: I might call this `blankPerseusRenderer` since it doesn't have any information. `generic` makes me think it takes type arguments or something.

```suggestion
export const blankPerseusRenderer: PerseusRenderer = {
```

I also think it might be slightly clearer to export this value directly rather than exporting a `generateTestPerseusRenderer` function. Consider the difference between

```ts
const testRenderer = {
    ...blankPerseusRenderer,
    widgets: {"group 1": groupWidget},
}
```

and

```ts
const testRenderer = generateTestPerseusRenderer({
    widgets: {"group 1": groupWidget},
})
```

The function call looks like it might be hiding something sophisticated. The object merge obviously has nothing up its sleeve.

packages/perseus/src/widgets/group/group.testdata.ts
```diff
@@ -185,3 +191,38 @@ export const simpleGroupQuestion: PerseusRenderer = {
...
+                type: "group",
+                options: groupRenderer,
+            },
+        },
+    });
+
+    return generateTestPerseusItem({question: itemRenderer});
+}
+
+export function getSplitGroupTestItem(): PerseusItem {
```
benchristel, line 226: "Split" means "answerless" and "full" means "answerful", right?

I wonder if these exports should be const values rather than functions.

packages/perseus/src/widgets/group/group.test.tsx
```diff
@@ -562,4 +572,113 @@ describe("group widget", () => {
...
+            },
+            {
+                content: "Correct",
+            },
+        ]);
+    });
+
+    describe.each([
+        ["answerless", getFullGroupTestItem()],
+        ["answerful", getSplitGroupTestItem()],
```
benchristel, line 594: Tamara created `getAnswerfulItem` and `getAnswerlessItem` functions for this purpose — could those be used here? I think that would reduce the amount of Group-specific test helper code we need.

See: https://github.com/Khan/perseus/pull/2387/files#diff-5da0a8814b21acafa0ef44b34f1f8def4e5f43eb8a55710c5c909fee26ef1d3bR3

benchristel (approved):
LGTM! I left a few suggestions inline.

packages/perseus-core/src/utils/split-perseus-item.ts
handeyeco: Just overly cautious on my part trying to avoid mutating incoming objects.

packages/perseus-core/src/utils/test-utils.ts
handeyeco: Renamed from `generic` to `blank`.

I prefer factories over object literals because it allows us to control how objects are created in one central place. For instance if we wanted to change this from a shallow to a deep merge, we'd just update the factory.

My real intent though is to keep one test from mutating the test data another test depends on. Unfortunately I didn't really do it right here, did I? Because you can still mutate nested objects...

I'll fix that.

packages/perseus/src/widgets/group/group.test.tsx
handeyeco: I was aware of those. They're a little inflexible for what I need here though since it makes whole PerseusItems and I need nested PerseusRenderers in an item.

PR #2418, opened 2025-04-23
Inaccessible widget warning

tatianasnook:
(LEMS-2972) This PR displays a warning in the Issues Panel when a content creator selects or pastes a widget that is marked as accessible: false.

Changes include:
Detects widgets with accessible: false
Displays a clear warning in the Issues Panel to alert content creators

### DIFF
--- a/packages/perseus-editor/src/editor.tsx
+++ b/packages/perseus-editor/src/editor.tsx
@@ -31,6 +31,9 @@ import TexErrorView from "./tex-error-view";
 
 import type {ChangeHandler, ImageUploader} from "@khanacademy/perseus";
 import type {PerseusWidget, PerseusWidgetsMap} from "@khanacademy/perseus-core";
+import { getWidgetExport, isAccessible } from "../../perseus/src/widgets";
+import { Console } from "console";
+import type { Issue }  from "./issues-panel";
 
 // like [[snowman numeric-input 1]]
 const widgetPlaceholder = "[[\u2603 {id}]]";
@@ -129,6 +132,7 @@ type Props = Readonly<{
     widgetIsOpen?: boolean;
     imageUploader?: ImageUploader;
     onChange: ChangeHandler;
+    onAddWarning?: (warning: Issue) => void;
 }>;
 
 type DefaultProps = {
@@ -731,6 +735,8 @@ class Editor extends React.Component<Props, State> {
             return;
         }
 
+        this._checkAccessibilityAndWarn(widgetType);
+
         this._addWidgetToContent(
             this.props.content,
             [textarea.selectionStart, textarea.selectionEnd],
@@ -738,6 +744,32 @@ class Editor extends React.Component<Props, State> {
         );
         textarea.focus();
     };
+
+    _checkAccessibilityAndWarn(
+        widgetType: string,
+    ) {
+        const widgetInfo = getWidgetExport(widgetType);
+        const widgetClass = widgetInfo?.widget;
+
+        const maybeAccessibleWidget = widgetClass as typeof React.Component & {
+            accessible?: boolean;
+        };
+
+        // If .accessible === false, then isAccessible will be false
+        // If .accessible === true or .accessible === undefined, then isAccessible will be true
+        // const isAccessible = maybeAccessibleWidget.accessible !== false;
+        const isAccessible = false; //just fot testing
+
+        if (!isAccessible && this.props.onAddWarning) {
+            this.props.onAddWarning({
+                id: `${widgetType} inaccessible`,
+                message: `${widgetType} is marked as inaccessible.`,
+                help: "Please choose an accessible alternative.",
+                helpUrl: "https://khanacademy.org/accessibility-guidelines",
+                impact: "Medium",
+            });
+        }
+    }
 
     addTemplate: (e: React.SyntheticEvent<HTMLSelectElement>) => void = (
         e: React.SyntheticEvent<HTMLSelectElement>,
--- a/packages/perseus-editor/src/editor.tsx
+++ b/packages/perseus-editor/src/editor.tsx
@@ -6,6 +6,7 @@ import {
     PerseusMarkdown,
     Util,
     Widgets,
+    isAccessible,
 } from "@khanacademy/perseus";
 import {
     CoreWidgetRegistry,
@@ -24,12 +25,10 @@ import "./katex-mhchem";
 import * as React from "react";
 import _ from "underscore";
 
-// eslint-disable-next-line import/no-relative-packages
-import {isAccessible} from "../../perseus/src/widgets";
-
 import DragTarget from "./components/drag-target";
 import WidgetEditor from "./components/widget-editor";
 import WidgetSelect from "./components/widget-select";
+import {WARNINGS} from "./messages";
 import TexErrorView from "./tex-error-view";
 
 import type {Issue} from "./issues-panel";
@@ -134,6 +133,7 @@ type Props = Readonly<{
     imageUploader?: ImageUploader;
     onChange: ChangeHandler;
     onAddWarning?: (warning: Issue) => void;
+    onRemoveWarning?: (warningId: string) => void;
 }>;
 
 type DefaultProps = {
@@ -181,6 +181,7 @@ class Editor extends React.Component<Props, State> {
         textAreaValue: this.props.content,
     };
 
+    _prevWidgetIds: string[] = [];
     componentDidMount() {
         // See componentDidUpdate() for how this flag is used
         this.lastUserValue = null;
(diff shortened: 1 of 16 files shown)


packages/perseus-editor/src/editor.tsx
```diff
@@ -129,6 +133,7 @@ type Props = Readonly<{
     widgetIsOpen?: boolean;
     imageUploader?: ImageUploader;
     onChange: ChangeHandler;
+    onAddWarning?: (warning: Issue) => void;
```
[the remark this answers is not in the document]
tatianasnook: I’ve added an onRemoveWarning?: (warningId: string) => void prop to the Editor and wired it up in _checkAccessibilityAndWarn. Now we keep track of the previous widgetIds array, diff it against the current array on each render, and call onRemoveWarning(warningId) for any widget that’s gone (or resolved). The parent’s handleRemoveWarning will then filter it out of the IssuesPanel.

packages/perseus-editor/src/editor.tsx
```diff
@@ -739,6 +744,24 @@ class Editor extends React.Component<Props, State> {
         textarea.focus();
     };
 
+    _checkAccessibilityAndWarn(widgetInfo: any) {
+        const widgetType = widgetInfo?.type;
+        const widgetAccessibility = isAccessible(widgetInfo);
+
+        if (!widgetAccessibility && this.props.onAddWarning) {
+            this.props.onAddWarning({
```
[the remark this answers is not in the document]
tatianasnook: I’ve added an onRemoveWarning?: (warningId: string) => void prop to the Editor and wired it up in _checkAccessibilityAndWarn. Now we keep track of the previous widgetIds array, diff it against the current array on each render, and call onRemoveWarning(warningId) for any widget that’s gone (or resolved). The parent’s handleRemoveWarning will then filter it out of the IssuesPanel.

packages/perseus-editor/src/messages.ts
catandthemachines, line 1: Can you add a comment to the beginning of this file stating what it's purpose is. Like:
>This file is meant to contain any and all perseus-editor messages we want to show within the issues panel.

packages/perseus/src/index.ts
```diff
@@ -24,6 +24,7 @@ export {default as Renderer} from "./renderer";
  */
 export * as Widgets from "./widgets";
 export {default as widgets} from "./all-widgets";
+export {isAccessible} from "./widgets";
```
catandthemachines, line 27: @handeyeco or @mark-fitzgerald , do you see any concerns exporting this function from Perseus package? I'm assuming we probably want this function (or another function) exported from `perseus-core` or maybe this should be moved to the `extract-perseus-data` file within the perseus package? 🤔

packages/perseus-editor/src/item-editor.tsx
```diff
@@ -90,6 +98,25 @@ class ItemEditor extends React.Component<Props> {
         };
     };
 
+    handleAddWarning = (warning: Issue) => {
+        this.setState((prevState) => {
+            const alreadyExists = prevState.warnings.some(
+                (w) => w.id === warning.id,
+            );
+            return alreadyExists
+                ? null
```
catandthemachines, line 107: Instead of returning null, we probably want to return `[]` since that is the default state of warnings.

packages/perseus-editor/src/editor.tsx
```diff
@@ -987,6 +1028,10 @@ class Editor extends React.Component<Props, State> {
             this.widgetIds = _.keys(widgets);
             widgetsDropDown = <WidgetSelect onChange={this._addWidget} />;
 
+            // Iterate through all rendered widgets and trigger a warning
+            // in the issues panel for any widget marked as inaccessible.
+            this._checkAccessibilityAndWarn(this.widgetIds);
```
catandthemachines, line 1033: Hmmm, I'm having a few thoughts about this implementation. It works in practice because we go through all the widgetIds here, compare against a local `previousWidgetIds` and then from there add or remove a warning message. This isn't necessarily bad but this implementation feels like it's adding an extra step on top of the natural onChange functionality that get's passed down from `editor-page.tsx`. I almost think maybe this logic should live in `editor-page.tsx` and be called within the `handleChange` and or `changeJSON` function as that is the central point where all changes to widgets and widget states get called back up to and we need this to get triggered by changes in both of those areas.

@jeremywiebe and @mark-fitzgerald , what do you guys think? Do you have other thoughts or suggestions on where this logic should live within the editor?

catandthemachines (commented):
The changes are coming along great. I think the two big open areas for this implementation is feedback on how we want to export `isAccessible` and clarity where is the right location and home for this `_checkAccessibilityAndWarn` logic, because we don't want this functionality to accidentally get hidden for future engineers.

packages/perseus-editor/src/editor.tsx
```diff
@@ -739,6 +745,41 @@ class Editor extends React.Component<Props, State> {
...
+            const widgetType = widgetInfo?.type;
+            const widgetAccessibility = isAccessible(widgetInfo);
+
+            if (!widgetAccessibility && this.props.onAddWarning) {
+                this.props.onAddWarning(
+                    WARNINGS.inaccessibleWidget(widgetType),
+                );
+            }
+        });
+        this._prevWidgetIds = currentWidgetIds;
```
mark-fitzgerald, line 780: Isn't this already in place on line 756?

packages/perseus-editor/src/issue-details.tsx
```diff
@@ -6,8 +6,9 @@ import PerseusEditorAccordion from "./components/perseus-editor-accordion";
 
 type Issue = {
     id: string;
-    help: string;
+    description: string;
     helpUrl: string;
+    help: string;
     impact: string;
     message: string;
 };
```
mark-fitzgerald, line 14: Since this is defined in `issues-panel.tsx`, I think we can just import it instead.

packages/perseus-editor/src/issues-panel.tsx
```diff
@@ -46,17 +51,25 @@ const IssuesPanel = ({warnings = []}: IssuesPanelProps) => {
                         kind="secondary"
                         size="small"
                         onClick={togglePanel}
-                        style={{marginRight: 0, flexGrow: 0}}
+                        disabled={!hasWarnings}
+                        style={{
+                            marginRight: 0,
+                            // flexGrow: 0,
```
mark-fitzgerald, line 57: Is `flexGrow` no longer needed? If not, the commented code can be removed.

packages/perseus-editor/src/issues-panel.tsx
```diff
@@ -46,17 +51,25 @@ const IssuesPanel = ({warnings = []}: IssuesPanelProps) => {
                         kind="secondary"
                         size="small"
                         onClick={togglePanel}
-                        style={{marginRight: 0, flexGrow: 0}}
+                        disabled={!hasWarnings}
+                        style={{
+                            marginRight: 0,
```
mark-fitzgerald, line 56: Nit: Even though we don't have to handle right-to-left (RTL) languages in the editor, I recommend that we get in the habit of using new RTL styling techniques. In this case, using `marginInlineEnd: 0` would be the replacement property to use.

packages/perseus-editor/src/item-editor.tsx
```diff
@@ -90,6 +98,25 @@ class ItemEditor extends React.Component<Props> {
...
+                : {warnings: [...prevState.warnings, warning]};
+        });
+    };
+
+    handleRemoveWarning = (warningId: string) => {
+        this.setState((prevState) => {
+            return {
+                warnings: prevState.warnings.filter((w) => w.id !== warningId),
```
mark-fitzgerald, line 115: Does this handle the situation where multiple instances of an inaccessible widget exist, but only one of them is removed?

packages/perseus-editor/src/editor.tsx
mark-fitzgerald: My preference is that component logic & UI reside in their own files. The `editor.tsx` file is already quite long (in my opinion). I'm thinking something like an `inaccessible-widget-warnings.tsx` file that is passed as a child element of the `<IssuesPanel>` in the `item-editor.tsx` file. There could be an exported util function in that file that could be pulled into any ancestor file to determine if a warning should be added/removed.

packages/perseus-editor/src/issues-panel.tsx
tatianasnook: It was actually a mistake on my part to comment out flexGrow: 0. It’s still needed to ensure the layout behaves correctly. I’ll uncomment it and fix it.

packages/perseus-editor/src/issues-panel.tsx
tatianasnook: Hi Mark! Thanks for the tip! I’ll update marginRight to marginInlineEnd to follow the RTL best practice.

packages/perseus-editor/src/item-editor.tsx
tatianasnook: Thanks for raising this! In our current implementation, the warning IDs include the full widget ID (like "matcher 1 inaccessible" and "matcher 2 inaccessible"), so each widget has its own unique warning. Therefore, this code will correctly remove only the warning for the specific widget that was deleted.

packages/perseus-editor/src/issue-details.tsx
tatianasnook: Thanks for the suggestion, Mark! I’ll remove the duplicate Issue type definition from issue-details.tsx and import it from issues-panel.tsx.

packages/perseus-editor/src/item-editor.tsx
tatianasnook: Thanks for the feedback, Cat! That makes sense — I’ll update the code to return an empty array [] instead of null to keep the warnings state consistent.

packages/perseus-editor/src/editor.tsx
tatianasnook: You're right, Mark — that line was duplicated. I've removed the redundant one. Thanks for catching it!

packages/perseus/src/index.ts
tatianasnook: Thanks for the discussion! I've moved the isAccessible function to perseus-core/utils as suggested, following @jeremywiebe's earlier guidance.

packages/perseus-editor/src/issues-panel.tsx
```diff
@@ -46,17 +51,25 @@ const IssuesPanel = ({warnings = []}: IssuesPanelProps) => {
...
+                            // flexGrow: 0,
+                            border: "none",
+                            backgroundColor: "transparent",
+                            cursor: hasWarnings ? "pointer" : "not-allowed",
+                        }}
                     />
                     <span>Issues Panel</span>
```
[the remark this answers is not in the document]
tatianasnook: You're right — that makes sense. I've updated the title from "Issues Panel" to just "Issues" for clarity and to avoid redundancy. Thanks for the suggestion!

packages/perseus-editor/src/editor.tsx
```diff
@@ -739,6 +745,41 @@ class Editor extends React.Component<Props, State> {
         textarea.focus();
     };
 
+    // Tracks previous widget IDs to diff out removals and keep the Issues Panel in sync
+    _checkAccessibilityAndWarn(currentWidgetIds: string[]) {
```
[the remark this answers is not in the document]
tatianasnook: Hi Jeremy! Thank you for reviewing. I saw that you're suggesting building checkAccessibilityAndWarn as a lint rule. I was under the impression that linter rules run before the program executes—not during runtime. Am I understanding that correctly?

tatianasnook:
Closing this PR in favor of a new one. The implementation has changed significantly, and it's cleaner to start fresh with a consolidated set of updates. The new PR will reflect all current changes and updated context.

PR #2419, opened 2025-04-23
remove references to propUpgrade

handeyeco:
## Summary:
<Add your summary here>

Issue: LEMS-3065

## Test plan:

### DIFF
--- a/packages/perseus-core/src/widgets/upgrade.ts
+++ b/packages/perseus-core/src/widgets/upgrade.ts
@@ -52,11 +52,7 @@ export const upgradeWidgetInfoToLatestVersion = (
         return oldWidgetInfo;
     }
 
-    // We do a clone here so that it's safe to mutate the input parameter
-    // in propUpgrades functions (which I will probably accidentally do at
-    // some point, and we would like to not break when that happens).
-    // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
-    let newEditorOptions = _.clone(oldWidgetInfo.options) || {};
+    let newEditorOptions = _.clone(oldWidgetInfo.options) ?? {};
 
     const upgradePropsMap = getWidgetOptionsUpgrades(type);
 
--- a/packages/perseus/src/types.ts
+++ b/packages/perseus/src/types.ts
@@ -20,7 +20,6 @@ import type {
     PerseusWidgetsMap,
     AnalyticsEventHandlerFn,
     Version,
-    WidgetOptionsUpgradeMap,
     Rubric,
     UserInput,
     UserInputArray,
@@ -495,7 +494,7 @@ export type WidgetExports<
     hidden?: boolean;
     /**
      * The widget version. Any time the _major_ version changes, the widget
-     * should provide a new entry in the propUpgrades map to migrate from the
+     * should provide a new entry in the widget parser to migrate from the
      * older version to the current (new) version. Minor version changes must
      * be backwards compatible with previous minor versions widget options.
      *
@@ -520,19 +519,6 @@ export type WidgetExports<
     getOneCorrectAnswerFromRubric?: (
         rubric: Rubric,
     ) => string | null | undefined;
-
-    /**
-     * A map of major version numbers (as a string, eg "1") to a function that
-     * migrates from the _previous_ major version.
-     *
-     * Example:
-     * ```
-     * propUpgrades: {'1': (options) => ({...options})}
-     * ```
-     *
-     * This configuration would migrate options from major version 0 to 1.
-     */
-    propUpgrades?: WidgetOptionsUpgradeMap;
 }>;
 
 export type FilterCriterion =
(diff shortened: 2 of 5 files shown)


packages/perseus-core/src/utils/split-perseus-item.test.ts
```diff
@@ -175,6 +175,7 @@ describe("splitPerseusItem", () => {
                             },
                         ],
                         buttonSets: ["basic"],
+                        extraKeys: ["PI"],
```
handeyeco, line 178: Since ServerItemRenderer doesn't do propUpgrade now (it's expected data is upgraded by the parser outside of ServerItemRenderer), test data had to be upgraded. We could probably modify the `renderQuestion` test util to do this, but I kind of like having the data up-to-date as it shows how the JSON will actually look.

packages/perseus-core/src/widgets/core-widget-registry.ts
```diff
@@ -37,14 +37,14 @@ import type {
...
+const widgets: Record<string, WidgetLogic> = {};
 
 function registerWidget(type: string, logic: WidgetLogic) {
     widgets[type] = logic;
 }
 
-export function isWidgetRegistered(type: string) {
-    const widgetLogic = widgets[type];
+export function isWidgetRegistered(type: string): boolean {
+    const widgetLogic: any = widgets[type];
```
handeyeco, line 47: I don't know what's going on with this. ESLint thinks that widgetLogic can't be undefined (it can be) so it doesn't like that I try to do a type coercion on the next line.

packages/perseus-core/src/widgets/expression/index.ts
```diff
@@ -1,18 +1,25 @@
...
+    PerseusExpressionWidgetOptions,
+    "answerForms" | "times" | "buttonSets" | "functions"
+>;
+
+export const defaultWidgetOptions: ExpressionDefaultWidgetOptions = {
+    answerForms: [],
+    times: false,
+    buttonSets: ["basic"],
+    functions: ["f", "g", "h"],
+};
```
handeyeco, line 18: Just moved

packages/perseus-core/src/widgets/measurer/index.ts
```diff
@@ -1,17 +1,34 @@
...
+export const defaultWidgetOptions: MeasurerDefaultWidgetOptions = {
+    box: [480, 480],
+    image: {} as any,
+    showProtractor: true,
+    showRuler: false,
+    rulerLabel: "",
+    rulerTicks: 10,
+    rulerPixels: 40,
+    rulerLength: 10,
+};
```
handeyeco, line 27: Moved

packages/perseus-core/src/widgets/radio/derive-num-correct.ts
```diff
@@ -0,0 +1,7 @@
+import type {PerseusRadioWidgetOptions} from "../../data-schema";
+
+export function deriveNumCorrect(options: PerseusRadioWidgetOptions) {
```
handeyeco, line 3: Moved

packages/perseus-core/src/widgets/radio/index.ts
```diff
@@ -1,18 +1,34 @@
...
+
+export const defaultWidgetOptions: RadioDefaultWidgetOptions = {
+    choices: [{}, {}, {}, {}] as any,
+    displayCount: null,
+    randomize: false,
+    hasNoneOfTheAbove: false,
+    multipleSelect: false,
+    countChoices: false,
+    deselectEnabled: false,
+};
```
handeyeco, line 27: Moved

packages/perseus-core/src/widgets/upgrade.ts
```diff
@@ -52,50 +51,7 @@ export const upgradeWidgetInfoToLatestVersion = (
...
-                            fromMajorVersion: nextVersion - 1,
-                            toMajorVersion: nextVersion,
-                            oldWidgetInfo: JSON.stringify(oldWidgetInfo),
-                        },
-                    },
-                );
-            }
-        }
-    }
+    let newEditorOptions = _.clone(oldWidgetInfo.options) ?? {};
```
handeyeco, line 54: So as of this writing, I'm keeping `upgradeWidgetInfoToLatestVersion` because it seems to do other things besides purely upgrading widget options. I need to rename it.

benchristel (approved):
LGTM! Thanks for doing this!

ivyolamit:
removing reviewers from project-radio-widget team since there's no direct impact to our project on this change.

packages/perseus-core/src/widgets/radio/index.ts
```diff
@@ -1,18 +1,34 @@
...
-    defaultWidgetOptions,
-    widgetOptionsUpgrades,
-} from "./radio-upgrade";
 import getRadioPublicWidgetOptions from "./radio-util";
 
+import type {PerseusRadioWidgetOptions} from "../../data-schema";
 import type {WidgetLogic} from "../logic-export.types";
 
-export type {RadioDefaultWidgetOptions} from "./radio-upgrade";
+export const currentVersion = {major: 2, minor: 0};
```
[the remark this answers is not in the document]
handeyeco: Hm... 🤔 I wonder if it would be worth doing a followup ticket. The type `WidgetLogic` has `version`, maybe somewhere in the code we could throw if

- a parser doesn't have a migration path to the latest major version OR
- a parser has a migration path to a version that's above the `WidgetLogic` version

PR #2424, opened 2025-04-24
Ensure answerless options are assignable to NumberLine props

benchristel:
## Summary:
This PR adds tests for the answerless case, and updates the types to reflect
what actually constitutes valid input.

The `numDivisions` prop is now required to be a number. The code for `NumericInput` is written in an (overly) clever way that treats 0 and null the same, but this is brittle and TypeScript doesn't like it. To clarify the code, we default `numDivisions` to 0 in the parser. 

Issue: LEMS-2976

## Test plan:

TODO


packages/perseus/src/widgets/number-line/number-line.tsx
```diff
@@ -191,8 +199,8 @@ const TickMarks: any = Graphie.createSimpleClass((graphie, props) => {
 
 // TODO: most widgets use some like Widget<Something, PerseusNumberLineWidgetOptions>
 // should this one?
-type Props = ChangeableProps & {
-    range: [number, number];
+type Props = {
+    range: number[];
```
handeyeco, line 203: I thought `[number, number]` was preferred when there was a fixed number of elements?

packages/perseus/src/widgets/number-line/number-line.tsx
```diff
@@ -207,19 +215,26 @@ type Props = ChangeableProps & {
     onBlur: (arg1: any) => void;
     onChange: (arg1: any, arg2?: () => void | null | undefined) => void;
     apiOptions: APIOptions;
-    keypadElement: HTMLElement | null | undefined;
+    keypadElement?: HTMLElement | undefined;
```
handeyeco, line 218: This couldn't be:

```suggestion
    keypadElement?: HTMLElement;
```

handeyeco (approved):
Cool, fairly small PR and I feel like I learned a lot.

benchristel:
@handeyeco turns out this change breaks the Storybook examples. I'm going to rework this.

packages/perseus/src/widgets/number-line/number-line.tsx
benchristel: The `data-schema` types have `range: number[]`, and the parser (which is derived from data-schema) does too. Since we apparently don't guarantee that there will always be two elements, I opted to change this type to be compatible with data-schema rather than vice-versa.

PR #2425, opened 2025-04-24
Add `isInequality` to the `PerseusNumberLineWidgetOptions` type.

benchristel:
The NumberLine component was using this field, but it wasn't represented in the types.

Issue: none

## Test plan:

`pnpm test`

Review the Number Line stories in Storybook.


handeyeco (commented):
> The NumberLine component was using this field, but it wasn't represented in the types.

How was that possible? Seems like it should have been a TS error. Should we also fix what was burying the error?

benchristel:
> How was that possible? Seems like it should have been a TS error. Should we also fix what was burying the error?

@handeyeco The issue is that we don't have typechecking at the boundary where the Renderer passes props to the widgets. The fix is to assert that the widget options are assignable to the component's external props, as I've done here: https://github.com/Khan/perseus/pull/2424/files#diff-6daf6fc96214fb21eec64ecf94b7293819455980404b4d1f0f5a71c7e253b3cfR224-R230

In other words, the fix for the underlying issue is in another PR.

PR #2426, opened 2025-04-24
Answerless InputNumber

handeyeco:
## Summary:
Pretty simple. Had to add `getInputNumberPublicWidgetOptions` but InputNumber wasn't using `value` within the widget, so there wasn't much else to do besides add tests/stories.

Issue: LEMS-2987

## Test plan:
InputNumber should continue to work as normal. Now we can confirm it will work sans-answers due to the tests and stories.


packages/perseus-core/src/widgets/input-number/input-number-util.ts
```diff
@@ -0,0 +1,29 @@
...
+    | "simplify"
+    | "size"
+    | "customKeypad"
+>;
+
+/**
+ * Given a PerseusInputNumberWidgetOptions object, return a new object with only
+ * the public options that should be exposed to the client.
+ */
+function getInputNumberPublicWidgetOptions(
```
handeyeco, line 22: For some reason this didn't exist, so I made one.

packages/perseus-core/src/widgets/logic-export.types.ts
```diff
@@ -30,23 +31,24 @@ export type WidgetOptionsUpgradeMap = {
...
 export type PublicWidgetOptionsFunction =
-    | typeof getMatcherPublicWidgetOptions
-    | typeof getPlotterPublicWidgetOptions
-    | typeof getIFramePublicWidgetOptions
-    | typeof getRadioPublicWidgetOptions
-    | typeof getNumericInputPublicWidgetOptions
-    | typeof getDropdownPublicWidgetOptions
     | typeof getCategorizerPublicWidgetOptions
-    | typeof getOrdererPublicWidgetOptions
+    | typeof getCSProgramPublicWidgetOptions
```
handeyeco, line 35: Just sorted

packages/perseus-core/src/widgets/input-number/input-number-util.ts
Myranae: I think we (or I) was under the assumption at the time that all input number would be replaced with numeric input. I think that's not how it turned out, though I know some work was done towards that goal. Thanks for adding these!

packages/perseus/src/widgets/input-number/input-number.stories.tsx
```diff
@@ -108,3 +109,23 @@ export const Percent = (args: InputNumberOptions): React.ReactElement => {
...
+
+export const Answerless = (): React.ReactElement => {
+    const item = getAnswerfulItem("input-number", {
+        simplify: "optional",
+        size: "normal",
+        value: 42,
+    });
+    item.question.content = `The answer is 42\n${item.question.content}`;
+    return <ServerItemRendererWithDebugUI item={item} startAnswerless />;
+};
```
Myranae, line 131: I think Jeremy must have skipped this widget for converting the stories to the new CSF format. Would be nice if these stories were in that format as well as we're trying to switch them all over to it, but looks like you'd have to convert the older stories for that to work, so totally up to you.

packages/perseus/src/widgets/input-number/input-number.tsx
```diff
@@ -19,42 +18,42 @@ import type {
     PerseusInputNumberUserInput,
 } from "@khanacademy/perseus-core";
 
-const formExamples = {
-    integer: function (options, strings: PerseusStrings) {
+type FormExampleFunction = (options: Props, strings: PerseusStrings) => string;
+
+const formExamples: Record<string, FormExampleFunction> = {
+    integer: function (_, strings) {
```
Myranae, line 24: Love the typing/cleanup here!

packages/perseus/src/widgets/input-number/input-number.stories.tsx
```diff
@@ -108,3 +109,23 @@ export const Percent = (args: InputNumberOptions): React.ReactElement => {
...
 };
 Percent.args = question3.widgets["input-number 1"].options;
+
+export const Answerful = (): React.ReactElement => {
+    const item = getAnswerfulItem("input-number", {
+        simplify: "optional",
+        size: "normal",
+        value: 42,
+    });
+    item.question.content = `The answer is 42\n${item.question.content}`;
```
Myranae, line 119: I think having the answer included in the content is a great idea. Could possibly even say what the answer field is called to help with QE testing, since that takes the longest to figure out without any context.

Myranae (approved):
Looks great! I did notice that answerless stories actually end up with more Perseus JSON data than answerful, not including the answer. Not sure that's important since these are just stories, but wanted to point out.

benchristel (approved):
LGTM!

packages/perseus/src/widgets/input-number/input-number.stories.tsx
handeyeco: I talked to him and I guess there were reasons for not doing InputNumber

PR #2427, opened 2025-04-24
Jest test correctness fixes

jeremywiebe:
## Summary:

This PR fixes a bunch of Jest tests to be more reliable. It improves:

  * Switches to using `findByXyz()` functions when using `userEvent` functions on the result. I've found that in some environments we'll see the dreaded `An update to XyzComponent inside a test was not wrapped in act(...).` error.
  * Move away from using the `done` callback in Jest tests (it's more portable to [return a `Promise`](https://jestjs.io/docs/asynchronous#promises) that just resolves when the test is done). 
  * Fixed a format option in a `test.each()` test (`%b` is invalid)
  * Wrapping operations that mutate the DOM in [`act()`](https://testing-library.com/docs/react-testing-library/api/#act) calls.

Issue: n/a

## Test plan:

`pnpm test` <-- all tests pass!


packages/perseus/src/widgets/image/__snapshots__/image.test.ts.snap
```diff
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`image widget - isMobile %b should snapshot: first render 1`] = `
```
jeremywiebe, line 3: These snapshots got re-ordered because I fixed the boolean format.  I think the `true` version used to be first here, but now `false` is.

packages/perseus/src/widgets/input-number/input-number.test.ts
```diff
@@ -335,18 +335,22 @@ describe("focus state", () => {
...
 
         // Assert
+        expect(screen.getByRole("textbox")).toHaveFocus();
         expect(gotFocus).toBe(true);
     });
 
     it("supports blurring", async () => {
         //  Arrange
         const {renderer} = renderQuestion(question);
+        await act(() => renderer.focus());
```
jeremywiebe, line 345: As we move forward with new versions of testing-library and React, it becomes more picky about DOM updates not done within `act()`. These calls definitely mutate the DOM so they need to be wrapped!

.changeset/six-pants-brush.md
```diff
@@ -0,0 +1,2 @@
+---
```
jeremywiebe, line 1: This PR is testing changes only, so no changeset entry because there are no production changes.

packages/perseus/src/components/number-input.tsx
```diff
@@ -226,12 +226,10 @@ class NumberInput extends React.Component<any, any> {
...
-            // eslint-disable-next-line @typescript-eslint/no-unused-vars
-            onFormatChange,
-            // eslint-disable-next-line @typescript-eslint/no-unused-vars
-            checkValidity,
-            // eslint-disable-next-line @typescript-eslint/no-unused-vars
-            useArrowKeys,
+            onFormatChange: _,
+            checkValidity: __,
+            useArrowKeys: ___,
+            allowPiTruncation: ____,
```
jeremywiebe, line 232: Just getting rid of some eslint suppressions. ☺️

Myranae (approved):
Looks good! And all tests passed :) Thanks for these fixes!

PR #2431, opened 2025-04-25
Add support for CSS Modules

mark-fitzgerald:
## Summary:
In our effort to modernize our approach to and toolset for styling, this change adds the ability to use standard CSS syntax in a `*.module.css` file for any of our widgets and renderers. This will enable us to migrate our styling away from Aphrodite and onto a standardized format. This is accomplished via an update to our build process. The prior Rollup `styles` plugin has been replaced with the `postcss` plugin since the [LESS files have since been converted to standard CSS](https://github.com/Khan/perseus/pull/2489) (no "building" is necessary).

The conversion of Aphrodite code to standard CSS is accomplished by the addition of a script that can be run against any TSX file that contains Aphrodite code. This script file uses various `@babel` plugins to generate an AST parsing of the code files, and then procedurally extracts the CSS settings and reformats them into a separate CSS file. This script does the heavy-lifting of moving and reformatting the styling, but it is expected that there are nuances in the JavaScript code that will necessitate adjustments to complete the transformation. For instance, in places where style information was passed to a `style` attribute, it will now need to reference the generated class name in the `className` attribute. This script is intended to be temporary, and only last as long as needed to convert and remove Aphrodite.

Issue: LEMS-3008

## Test plan:
1. Review the `<fieldset>` element in the Radio widget ([line 211 in radio/base-radio.new.tsx](https://github.com/Khan/perseus/pull/2431/files#diff-1e7272f36753eeee0138a5375f4434db42e6701e75782f0ab1638b99a53d2602R211)).
   * Note the addition of a class (`cssModulesTest`) pulled from a CSS file (radio/radio-test.module.css)
1. Launch Storybook
1. Navigate to the [new Radio Widget](http://localhost:6006/?path=/story/perseus-widgets-radionew--single-select) (this only shows when the Feature Flag is on)
1. Inspect the instructions text "Choose 1 answer:"
   * Note that the `<fieldset>` parent element in the DOM now includes a class called something like `_css-modules-test_6k65e_1` (<underscore><my-class-name><underscore><8-digit-hash>)
   * Note that in PROD, the class name will be `perseus_<8-digit-hash>`


packages/perseus/src/widgets/radio/radio-test.module.css
```diff
@@ -0,0 +1,3 @@
+.css-modules-test {
+    opacity: 0.99;
```
mark-fitzgerald, line 2: This is for demonstration purposes only (to show that CSS Modules can be imported). It will be replaced with widget-specific styling from the Aphrodite conversion in a future PR.

utils/extract-css.js
```diff
@@ -0,0 +1,604 @@
+const fs = require("fs");
+const path = require("path");
+
+const {parse} = require("@babel/parser");
+
```
mark-fitzgerald, line 6: This file is much longer that I would normally like things like this to be. It is intended to be temporary, but if folks prefer that this be broken up into smaller files, I'm happy to oblige. I'm happy to pair with anyone to explain what is happening in any section of the code. Also, since this is intended to be temporary, I didn't make it a TypeScript file, nor did I write any tests for it. It is intended to be a tool to aid in our conversion process. Let me know if you have strong feelings for a different approach.

utils/extract-css.js
```diff
@@ -0,0 +1,604 @@
...
+const fileDirectory = path.dirname(filePath);
+const fileNameParts = path.basename(filePath).split(".");
+const archivedFileName = `${fileNameParts[0]}.OLD.${fileNameParts[1]}`;
+const archivedFilePath = path.join(fileDirectory, archivedFileName);
+const cssFileName = `${fileNameParts[0]}.module.css`;
+const cssFilePath = path.join(fileDirectory, cssFileName);
+const indentation = "    ";
+const wbColorValues = {
+    // from node_modules/@khanacademy/wonder-blocks-tokens/dist/css/index.css
+    "#21242c": "--wb-semanticColor-text-primary",
```
ivyolamit, line 32: are we planning in the future to use directly WB Semantic colors instead of having these const here?
both css variables and tokens are now available in WB.

source: https://khan.github.io/wonder-blocks/?path=/docs/packages-tokens-semantic-color--docs

ivyolamit: ^ ah disregard the comment above, i just realized this is the extract script.

mark-fitzgerald: Future CSS should always reference semantic values (when available). This is only here to aid in the conversion of literal values (i.e. hex colors) to their semantic counterpart.

config/build/rollup.config.js
```diff
@@ -146,17 +148,26 @@ const createConfig = (
...
-                    // in this file. If you change one, you should ensure that
-                    // a new build correctly places font files in the right dir
-                    // and that the generated CSS contains valid relative urls
-                    // to those fonts!
-                    publicPath: "assets",
+                sourceMap: true,
+                plugins: [postcssImport()],
+                modules: {
+                    localsConvention: "camelCase",
+                    generateScopedName: function (name, filename, css) {
```
jeremywiebe, line 158: Could you add a comment here explaining what this is doing? I think it's building a CSS classname that is prefixed with `perseus_` and then is the hash of the CSS file contents... but it'd be good to explain for me in the future. :)

pnpm-workspace.yaml
```diff
@@ -73,6 +73,7 @@ catalogs:
         tiny-invariant: 1.3.1
         underscore: 1.4.4
         vite: 5.1.0
+        vite-css-modules: 1.8.6
```
jeremywiebe: We don't really need to put this in the catalog if it's only used in one `package.json`.  The catalog really shines where we have the same dep showing up in multiple places in the repo and we want to keep them all on the same version.

utils/extract-css.js
jeremywiebe: I think at a minimum it'd be good to put a JSDoc comment on the top explaining how to use/invoke it and what it does (reads a component file and dumps a .css file?)

utils/extract-css.js
jeremywiebe, line 1: Some usage comments (none of which should block this PR and may all be fine to leave as-is because of the short lifespan of this tool): 

* I ran `node utils/extract-css.js packages/perseus/src/components/hud.tsx`:
  * It didn't add an import that was needed: `import {css} from "aphrodite";` 
  * It left an unused import in place: `import * as constants from "../styles/constants";`
* The export from the CSS module file doesn't seem to have any type info which will mean that TypeScript won't help us verify that we have typed class names correctly. This is probably something we'll want to investigate and figure out in the near future.
* IMHO, you could just over-write the original file being processed. This is a git repo and so `git diff` will show exactly the changes. 🤷‍♂️ 
* For an stress-test, I ran this across all .tsx files in the repo 👿  It hit several issues in the script (several "cannot access XYZ on undefined" issues). `git ls-files packages/**/*.tsx | xargs -n 1 -t node ./utils/extract-css.js`

config/build/rollup.config.js
mark-fitzgerald: Added.

pnpm-workspace.yaml
mark-fitzgerald: Moved back.

utils/extract-css.js
mark-fitzgerald: 1. JSDoc comment added to file.
2. RE: Ran on `hud.tsx`
   * I anticipate the need to add more functionality as other files are converted.
   * I don't remove imports of external files since it's possible that those objects/values are used elsewhere. I might add a search function in the future to check to see if such imports are no longer used.
3. Type info added. Typescript now understands the imported CSS Modules and will cause a failure if code is trying to reference a class name that doesn't exist.
4. Added an `--archive` option to the script command. File archiving is "off" by default.
5. RE: Stress test
   * See item 2 (I'll address issues as they arrive).

vite.config.ts
```diff
@@ -48,10 +48,8 @@ export default defineConfig({
         ],
     },
     css: {
-        preprocessorOptions: {
-            less: {
-                math: "always",
-            },
```
jeremywiebe, line 54: Oh right.  Less is gone!

utils/extract-css.js
mark-fitzgerald: Done.

jeremywiebe (approved):

PR #2432, opened 2025-04-25
More test cleanup

jeremywiebe:
## Summary:

This PR forms another small set of unit test cleanup/fixes. It includes: 

  * Mock WB ToolTip differently so that it's a bit more dependable (avoids needing shared variables)
  * Fix test names for expression-editor
  * Add component names so they show up in React stack traces properly
  * Fix graded-group tests to be a bit more accurate

Issue: "none"

## Test plan:

`pnpm test`

### DIFF
--- a/packages/perseus-editor/src/widgets/__tests__/expression-editor.test.tsx
+++ b/packages/perseus-editor/src/widgets/__tests__/expression-editor.test.tsx
@@ -154,9 +154,8 @@ describe("expression-editor", () => {
         });
     });
 
-        const onChangeMock = jest.fn();
     it("should toggle logarithms checkbox", async () => {
-
+        const onChangeMock = jest.fn();
         render(<ExpressionEditor onChange={onChangeMock} />);
         act(() => jest.runOnlyPendingTimers());
 
@@ -171,9 +170,8 @@ describe("expression-editor", () => {
         });
     });
 
-        const onChangeMock = jest.fn();
     it("should toggle basic relations checkbox", async () => {
-
+        const onChangeMock = jest.fn();
         render(<ExpressionEditor onChange={onChangeMock} />);
         act(() => jest.runOnlyPendingTimers());
 
@@ -188,8 +186,8 @@ describe("expression-editor", () => {
         });
     });
 
-        const onChangeMock = jest.fn();
     it("should toggle advanced relations checkbox", async () => {
+        const onChangeMock = jest.fn();
 
         render(<ExpressionEditor onChange={onChangeMock} />);
         act(() => jest.runOnlyPendingTimers());
--- a/packages/perseus-editor/src/widgets/__tests__/expression-editor.test.tsx
+++ b/packages/perseus-editor/src/widgets/__tests__/expression-editor.test.tsx
@@ -154,8 +154,8 @@ describe("expression-editor", () => {
         });
     });
 
-    it("should toggle prealgebra checkbox", async () => {
         const onChangeMock = jest.fn();
+    it("should toggle logarithms checkbox", async () => {
 
         render(<ExpressionEditor onChange={onChangeMock} />);
         act(() => jest.runOnlyPendingTimers());
@@ -171,8 +171,8 @@ describe("expression-editor", () => {
         });
     });
 
-    it("should toggle prealgebra checkbox", async () => {
         const onChangeMock = jest.fn();
+    it("should toggle basic relations checkbox", async () => {
 
         render(<ExpressionEditor onChange={onChangeMock} />);
         act(() => jest.runOnlyPendingTimers());
@@ -188,8 +188,8 @@ describe("expression-editor", () => {
         });
     });
 
-    it("should toggle prealgebra checkbox", async () => {
         const onChangeMock = jest.fn();
+    it("should toggle advanced relations checkbox", async () => {
 
         render(<ExpressionEditor onChange={onChangeMock} />);
         act(() => jest.runOnlyPendingTimers());
--- a/packages/perseus/src/widgets/__testutils__/renderQuestion.tsx
+++ b/packages/perseus/src/widgets/__testutils__/renderQuestion.tsx
@@ -92,7 +92,7 @@ const RendererWrapper = React.forwardRef<
         apiOptions: APIOptions;
         extraProps?: PropsFor<typeof Perseus.Renderer>;
     }
->((props, ref) => {
+>(function RendererWithDependencies(props, ref) {
     const dependencies = useDependencies();
     return (
         <UserInputManager widgets={props.question.widgets} problemNum={0}>
--- a/packages/perseus/src/widgets/graded-group/__snapshots__/graded-group.test.ts.snap
+++ b/packages/perseus/src/widgets/graded-group/__snapshots__/graded-group.test.ts.snap
@@ -125,7 +125,7 @@ exports[`graded-group should snapshot: initial render (mobile: false) 1`] = `
                                       <div
                                         class="paragraph"
                                       >
-                                         Some bacteria conduct photosynthesis and produce oxygen, much like plants.
+                                        Some bacteria conduct photosynthesis and produce oxygen, much like plants.
                                       </div>
                                     </div>
                                   </div>
@@ -557,7 +557,7 @@ exports[`graded-group should snapshot: initial render (mobile: true) 1`] = `
                                       <div
                                         class="paragraph"
                                       >
-                                         Some bacteria conduct photosynthesis and produce oxygen, much like plants.
+                                        Some bacteria conduct photosynthesis and produce oxygen, much like plants.
                                       </div>
                                     </div>
                                   </div>
--- a/packages/perseus/src/widgets/graded-group/graded-group.testdata.ts
+++ b/packages/perseus/src/widgets/graded-group/graded-group.testdata.ts
@@ -23,7 +23,7 @@ export const question1: PerseusRenderer = {
                         options: {
                             static: false,
                             items: [
-                                " Some bacteria conduct photosynthesis and produce oxygen, much like plants.",
+                                "Some bacteria conduct photosynthesis and produce oxygen, much like plants.",
                                 "Bacteria are always autotrophic but they may get energy from either light or chemical sources.",
                                 "Some chemosynthetic bacteria introduce energy and fixed carbon into communities where photosynthesis is not possible (e.g., deep-sea vents).",
                                 "Some bacteria live symbiotically inside of host organisms and provide the host with nutrients.",
@@ -48,7 +48,7 @@ export const question1: PerseusRenderer = {
                             options: {
                                 static: false,
                                 items: [
-                                    " Some bacteria conduct photosynthesis and produce oxygen, much like plants.",
+                                    "Some bacteria conduct photosynthesis and produce oxygen, much like plants.",
                                     "Bacteria are always autotrophic but they may get energy from either light or chemical sources.",
                                     "Some chemosynthetic bacteria introduce energy and fixed carbon into communities where photosynthesis is not possible (e.g., deep-sea vents).",
                                     "Some bacteria live symbiotically inside of host organisms and provide the host with nutrients.",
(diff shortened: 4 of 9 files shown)


packages/perseus/src/widgets/graded-group/__snapshots__/graded-group.test.ts.snap
```diff
@@ -125,7 +125,7 @@ exports[`graded-group should snapshot: initial render (mobile: false) 1`] = `
                                       <div
                                         class="paragraph"
                                       >
-                                         Some bacteria conduct photosynthesis and produce oxygen, much like plants.
+                                        Some bacteria conduct photosynthesis and produce oxygen, much like plants.
```
jeremywiebe, line 128: The test data had a leading space. I removed it which caused this snapshot update.

packages/perseus/src/widgets/graded-group/graded-group.testdata.ts
```diff
@@ -23,7 +23,7 @@ export const question1: PerseusRenderer = {
                         options: {
                             static: false,
                             items: [
-                                " Some bacteria conduct photosynthesis and produce oxygen, much like plants.",
+                                "Some bacteria conduct photosynthesis and produce oxygen, much like plants.",
```
jeremywiebe, line 26: Just removing a leading space

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.test.tsx
```diff
@@ -48,66 +67,51 @@ describe("MovablePoint", () => {
...
-
-        /*  NOTE: When a WonderBlocks Tooltip is added to an element, it isn't rendered with the element.
-                  Instead, it is rendered elsewhere in the document (usually the bottom of the <body>),
-                        and only when the user interacts with the element.
-                  This makes checking for the Tooltip element difficult to manage in a test.
-                  Therefore, these tests mock the Tooltip component and check if/how it is called.
-         */
-        beforeEach(() => {
-            // @ts-expect-error // TS2339: Property mockImplementation does not exist on type typeof Tooltip
-            Tooltip.mockImplementation(TooltipMock);
```
jeremywiebe, line 76: Mocking the tooltip in `beforeEach()` can cause problems when tests are run in parallel. Plus, all we really want to do is verify that the content of the tooltip was rendered and so we can mock the module for all tests (see higher up in this file).

packages/perseus-editor/src/widgets/__tests__/expression-editor.test.tsx
jeremywiebe, line 1: These are all copy-pasta fixes.

packages/perseus/src/widgets/numeric-input/input-with-examples.tsx
```diff
@@ -45,7 +45,7 @@ type Props = {
  * examples of how to input the selected answer forms.
  */
 const InputWithExamples = forwardRef<Focusable, Props>(
-    (
+    function InputWithExamples(
```
jeremywiebe, line 48: Another change so that this component shows up nicely named in React dev tools (and stack traces)

benchristel:
@jeremywiebe these look like good improvements. Any reason this PR shouldn't be opened for review?

jeremywiebe:
> @jeremywiebe these look like good improvements. Any reason this PR shouldn't be opened for review?

I had extracted these tests out of a spike investigating moving to Vitest but forgot to open it for review.  I'll rebase it onto a current version of `main` and then open for review! Thanks for the prompt!

packages/perseus/src/widgets/__testutils__/renderQuestion.tsx
```diff
@@ -91,7 +91,7 @@ const Renderer = React.forwardRef<
         apiOptions: APIOptions;
         extraProps?: PropsFor<typeof Perseus.Renderer>;
     }
->((props, ref) => {
+>(function RendererWithDependencies(props, ref) {
```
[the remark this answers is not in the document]
Myranae: Yay easier stack traces!

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.test.tsx
```diff
@@ -18,20 +17,41 @@ jest.mock("@khanacademy/wonder-blocks-tooltip", () => {
     const originalModule = jest.requireActual(
         "@khanacademy/wonder-blocks-tooltip",
     );
+
     return {
-        __esModule: true,
         ...originalModule,
-        default: jest.fn(),
+        __esModule: true,
```
Myranae: Moving the __esModule like to after the spread means it overwrites the property brought in from the spread, right? I'm guessing that's what you want now though?

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.test.tsx
```diff
@@ -18,20 +17,41 @@ jest.mock("@khanacademy/wonder-blocks-tooltip", () => {
...
+        default: function MockToolTip({children, ...props}) {
+            /*
+            NOTE: When a WonderBlocks Tooltip is added to an element, it isn't
```
Myranae: Is it possible to interact with the tooltip so it appears and then search for whatever content is in the tooltip? If the tests are specifically for Tooltip, it seems not as helpful to mock it 🤔 I haven't written any tooltip tests though, so not sure.

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.test.tsx
```diff
@@ -18,20 +17,41 @@ jest.mock("@khanacademy/wonder-blocks-tooltip", () => {
...
 });
 
-const TooltipMock = ({children}) => {
-    return children;
-};
+function mockGraphConfig(config) {
+    return jest.spyOn(ReducerGraphConfig, "default").mockReturnValue(config);
+}
+
+function mockDraggingState(state) {
```
Myranae: I love how these functions make the following tests so much clearer!

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.test.tsx
jeremywiebe: Thanks for the push.  I checked the WB Tooltip tests and found out how to interact with the tooltip to make it show! :) 

Now we don't need the mock.

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.test.tsx
jeremywiebe: You have the right idea, but I figured out I don't need to set it at all!  (I suspect because we're spreading the original module and so it already is an ES Module.)

packages/perseus-editor/src/widgets/__tests__/expression-editor.test.tsx
```diff
@@ -188,7 +186,7 @@ describe("expression-editor", () => {
         });
     });
 
-    it("should toggle prealgebra checkbox", async () => {
+    it("should toggle advanced relations checkbox", async () => {
```
handeyeco, line 189: I wonder if there's a lint rule we could use to prevent test names from being duplicates? I've seen issues with copy/pasta in other tests too.

packages/perseus/src/widgets/__testutils__/renderQuestion.tsx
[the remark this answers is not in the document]
handeyeco: I feel like if I saw this and lacked context, I would immediately think to switch it back to an arrow function. I wonder:

- Should we add a comment as to why it's a named function
- Should we move it up to be a top-level function? (rather than an inline declaration?)

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.test.tsx
```diff
@@ -48,66 +40,72 @@ describe("MovablePoint", () => {
...
                         sequenceNumber={1}
                         onMove={() => {}}
                     />
-                    ,
                 </Mafs>,
             );
-            expect(Tooltip).toHaveBeenCalled();
+
+            // Act
+            await userEvent.hover(await screen.findByTestId("movable-point"));
```
handeyeco: Now that we've gone through all the IG a11y work, shouldn't there be a way to use `findByRole` vs `findByTestId`? Ideally we would never use `*ByTestId` but I know that's now always achievable.

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.test.tsx
```diff
@@ -48,66 +40,72 @@ describe("MovablePoint", () => {
...
                         sequenceNumber={1}
                         onMove={() => {}}
                     />
-                    ,
                 </Mafs>,
             );
-            expect(Tooltip).not.toHaveBeenCalled();
+
+            // Act
+            await userEvent.hover(await screen.findByTestId("movable-point"));
```
handeyeco: Unrelated to my other message and mostly just curiosity, I tend to have a bias towards `get` vs `find`; is there a reason you chose `find` here?

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.test.tsx
```diff
@@ -138,25 +151,32 @@ describe("MovablePoint", () => {
...
+            mockGraphConfig(baseGraphConfigContext);
+            mockDraggingState({dragging: true});
             const {container} = render(
                 <Mafs width={200} height={200}>
                     <MovablePoint
                         point={[0, 0]}
                         sequenceNumber={1}
                         onMove={() => {}}
                     />
-                    ,
```
handeyeco, line 159: 🤔 wut

packages/perseus-editor/src/widgets/__tests__/expression-editor.test.tsx
jeremywiebe: Viest was the reason I found this.  It logged an error because of the duplicate test name.  I'm not sure it's a huge deal... I'd be open to linting for it, but I don't think this duplicate name has caused us any hassle. :)

packages/perseus/src/widgets/__testutils__/renderQuestion.tsx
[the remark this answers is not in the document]
jeremywiebe: > I feel like if I saw this and lacked context, I would immediately think to switch it back to an arrow function. I wonder:
> 
> * Should we add a comment as to why it's a named function
> * Should we move it up to be a top-level function? (rather than an inline declaration?)

I found that we've actually disabled the lint rule that would catch this (it was commented that it wasn't helpful, but I really believe it is!).  I enabled it and there's about 10 violations so I'm going to enable it in a new PR and fix the violations.

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.test.tsx
jeremywiebe: You're right!  We can use `findByRole("button")`.  Thanks!

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.test.tsx
jeremywiebe: When trying out Vitest, I found that some uses of `getBy()` would hang the test if they were intermixed with other async APIs from testing library.  That's an annoyance.  I believe I also ran into this when trying out Jest v30 (recently released) so I think it makes sense to keep them.

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.test.tsx
jeremywiebe: Yeah... I think this was added/left here in error. Probably didn't affect the test so it was never noticed.

PR #2433, opened 2025-04-28
Suppress type errors in numeric-input.class.tsx

benchristel:
These type errors appeared intermittently when running `tsc --watch`. I suspect
there's a cleaner way to deal with these refs. In the short term, suppressing the
errors in code is preferable to tolerating them.

Issue: none

## Test plan:

`pnpm tsc -w`

### DIFF
--- a/packages/perseus/src/widgets/numeric-input/numeric-input.class.tsx
+++ b/packages/perseus/src/widgets/numeric-input/numeric-input.class.tsx
@@ -99,16 +99,25 @@ export class NumericInput
     }
 
     focus: () => boolean = () => {
-        this.inputRef.current?.focus();
+        // TODO(benchristel): type `inputRef` properly.
+        // Typechecking fails here intermittently, probably because of nondeterminism
+        // in the interpretation of `&` types.
+        (this.inputRef.current as any)?.focus();
         return true;
     };
 
     focusInputPath: () => void = () => {
-        this.inputRef.current?.focus();
+        // TODO(benchristel): type `inputRef` properly
+        // Typechecking fails here intermittently, probably because of nondeterminism
+        // in the interpretation of `&` types.
+        (this.inputRef.current as any)?.focus();
     };
 
     blurInputPath: () => void = () => {
-        this.inputRef.current?.blur();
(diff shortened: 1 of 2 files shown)


packages/perseus/src/widgets/numeric-input/numeric-input.class.tsx
```diff
@@ -99,16 +99,25 @@ export class NumericInput
     }
 
     focus: () => boolean = () => {
-        this.inputRef.current?.focus();
+        // TODO(benchristel): type `inputRef` properly.
+        // Typechecking fails here intermittently, probably because of nondeterminism
```
handeyeco, line 103: I've never had trouble with this, is it possibly a local configuration issue? Are other people seeing this? The fact that TS fails it _sometimes_ seems like a red flag to me. Is there a ticket to follow up on this?

SonicScrewdriver: Oh yeah, I was a little confused about this one. It was working perfectly fine on my machine, but I believe other people on the team were seeing some errors. It passed the PR checks fine as well, at the time.

SonicScrewdriver: Still seems fine for me right now. (I checked in both VSCode and Cursor in case Cursor isn't set up properly.) 

Screenshot 2025-05-07 at 9 49 14 AM

jeremywiebe: The way that we've defined this ref to bridge between two different components feels wrong to me.  I can repro the type check failures by using `pnpm tsc --watch` and then making a change in this file. 

I think what we should do is have the ref defined using an interface with the common functions defined in that interface.

jeremywiebe: Does this fix things for you @benchristel 

https://github.com/Khan/perseus/pull/2452

jeremywiebe: #2452 has landed. :)

PR #2434, opened 2025-04-28
Verify that NumberLine widgets work with answerless data

benchristel:
Issue: LEMS-2976

## Test plan:

```
pnpm storybook
open 'http://localhost:6006/?path=/docs/perseus-widgets-number-line--docs'
```


packages/perseus/src/widgets/number-line/number-line.test.ts
```diff
@@ -183,41 +188,91 @@ describe("number-line widget", () => {
...
-        act(() => numberLine.movePosition(3.5));
-        const score = scorePerseusItemTesting(
-            question1,
-            renderer.getUserInputMap(),
-        );
+    describe.each([
+        ["answerless", getAnswerlessItem("number-line", numberLineOptions)],
+        ["answerful", getAnswerfulItem("number-line", numberLineOptions)],
+    ])("given %s options", (_, {question}) => {
+        const answerKey = getAnswerfulItem(
```
handeyeco, line 218: I don't think I've seen the use of "key" used for widget options / rubric / whatever. It was a little confusing at first.

packages/perseus/src/widgets/number-line/number-line.test.ts
```diff
@@ -183,41 +188,91 @@ describe("number-line widget", () => {
...
+            "number-line",
+            numberLineOptions,
+        ).question;
+
+        it("can be answered correctly", () => {
+            // Arrange
+            const apiOptions: APIOptions = {
+                isMobile: false,
+            };
```
handeyeco, line 227: `isMobile` defaults to `false`, so I don't think you need to define your own APIOptions for these tests.

packages/perseus/src/widgets/number-line/number-line.test.ts
```diff
@@ -183,41 +188,91 @@ describe("number-line widget", () => {
...
+            const {renderer} = renderQuestion(question, apiOptions);
 
-        // Assert
-        expect(score).toHaveBeenAnsweredIncorrectly();
+            // Act
+            const [numberLine] = renderer.findWidgets("number-line 1");
+            act(() => numberLine.movePosition(-2.5));
```
handeyeco, line 232: I'm guessing there isn't an accessible handle for NumberLine? Like we have to do this vs `getByRole`?

handeyeco (approved):

PR #2435, opened 2025-04-28
remove deprecated options for perseus parser version 3

anakaren-rojas:
## Summary:
<full summary>

Issue: LEMS-3061 <url or "none">

## Test plan:
<see below>

# --
# For 'Issue:', A link (JIRA, Slack, etc) to the issue that
#   drives this change or "none" if there isn't a link
# For 'Test Plan:', list the commands you ran or process you
#   followed to test this change.
# If this commit is adding to an existing review, delete the
# 'Test Plan:' lines and replace them with 'Review: Dxxx'.


packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.test.ts
```diff
@@ -2,59 +2,29 @@ import {anySuccess} from "../general-purpose-parsers/test-helpers";
...
+    migrateV2toV3,
+    parseRadioWidget,
+} from "./radio-widget";
 
 describe("parseRadioWidget", () => {
-    it("migrates v1 options to v2", () => {
+    it("migrates to the latest version", () => {
         const widget = {
             type: "radio",
-            graded: true,
```
anakaren-rojas, line 11: I'm not sure why the latest migration function gets rid of graded

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.test.ts
```diff
@@ -2,59 +2,48 @@ import {anySuccess} from "../general-purpose-parsers/test-helpers";
...
-                            content: "Incorrect",
-                            correct: false,
-                        },
-                    ],
-                    numCorrect: 2,
-                },
-                version: {
-                    major: 2,
-                    minor: 0,
+                    choices: [],
```
benchristel, line 40: Don't we want to include at least one choice here, to make sure choices are migrated correctly?

packages/perseus-core/src/data-schema.ts
```diff
@@ -1364,15 +1364,6 @@ export type PerseusRadioWidgetOptions = {
...
     deselectEnabled?: boolean;
-    // deprecated
-    // NOTE: perseus_data.go says this is required even though it isn't necessary.
-    onePerLine?: boolean;
-    // deprecated
-    // NOTE: perseus_data.go says this is required even though it isn't necessary.
-    displayCount?: any;
-    // v0 props
-    // `noneOfTheAbove` is still in use (but only set to `false`).
-    noneOfTheAbove?: false;
```
benchristel, line 1375: 🎉

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.test.ts
```diff
@@ -145,3 +134,129 @@ describe("parseRadioWidget", () => {
...
+                hasNoneOfTheAbove: false,
+                multipleSelect: false,
+                numCorrect: 1,
+                randomize: false,
+            },
+            type: "radio",
+            version: {major: 3, minor: 0},
+        });
+    });
+});
```
benchristel, line 262: Thanks for adding these tests! I think testing the migration functions individually is an interesting approach; it ensures we won't have to update the tests when we add new migrations.

However, I think we'll lose some of the documentation power of tests if we don't also test the full upgrade paths: v1 -> latest, v2 -> latest, etc. What I want to know, as a reader/reviewer of these tests, is whether the different input formats will produce the expected output when run through all the necessary migrations. One way to test that without introducing duplication might be:

```ts
describe.each([
  ["with no version", radioNoVersion],
  ["v0", radioV0],
  ["v1", radioV1],
  ["v2", radioV2],
  ["v3", radioV3],
])("Radio %s", (_, inputData) => {
  it("upgrades to the latest version", () => {
    expect(parse(inputData, parseRadioWidget)).toEqual({
      // data that every input will upgrade to
    });
  });
});
```

I.e. have a single expected output that is reused for all tests, and craft the input data for each version so it will upgrade to that output.

benchristel (commented):
The removal of the deprecated props LGTM! Did you also want to rename `clue` in this PR? My concern is that if we land this PR to `main` separately, it might be released, and then we'll have to do another migration for the rename.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.test.ts
anakaren-rojas: thats a great idea, love that

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.test.ts
anakaren-rojas: yes definitely, you caught me a little early in my testing 😆

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.test.ts
anakaren-rojas: @benchristel I added the table test above and created a separate mock data file to avoid crowding this test file with the mock data.

packages/perseus/src/widgets/explanation/__snapshots__/explanation.test.ts.snap
```diff
@@ -21,7 +21,7 @@ exports[`Explanation should snapshot when expanded: expanded 1`] = `
             aria-controls=":r1:"
             aria-disabled="false"
             aria-expanded="true"
-            class="button_vr44p2-o_O-shared_11fo0od-o_O-default_1dlh9ep-o_O-small_1rcykji-o_O-inlineStyles_1s8anjv"
+            class="button_vr44p2-o_O-shared_11fo0od-o_O-default_1la56ro-o_O-small_1rcykji-o_O-inlineStyles_1s8anjv"
```
benchristel, line 24: I don't think these snapshots should have changed. Do you maybe need to `pnpm install`?

anakaren-rojas: oh yes, good callout, thanks Ben!

packages/perseus-core/src/data-schema.ts
SonicScrewdriver: It's so beautiful.🥹

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.test.ts
```diff
@@ -2,62 +2,29 @@ import {anySuccess} from "../general-purpose-parsers/test-helpers";
...
-            }),
-        );
+    describe.each([
+        ["v0", v0Widget],
+        ["v1", v1Widget],
+        ["v2", v2Widget],
+        ["v3", v3Widget],
+    ])("Radio %s", (_, inputData) => {
+        it(`migrates to the latest version, ${LATEST_VERSION.major}.${LATEST_VERSION.minor}`, () => {
+            expect(parse(inputData, parseRadioWidget)).toEqual(
```
SonicScrewdriver, line 24: This is much cleaner, nice work!

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.test.ts
```diff
@@ -86,7 +53,7 @@ describe("parseRadioWidget", () => {
             key: null,
             graded: true,
             version: {
-                major: 2,
+                major: 3,
```
SonicScrewdriver, line 56: I wonder if we should make this a silly large number just so that we don't have to update the test every time we get a new widget version. Although, maybe it's a nice canary to let us know.🤔

packages/perseus/src/widgets/radio/radio.new.tsx
```diff
@@ -279,9 +279,9 @@ const RadioComponent = (props: RadioComponentProps) => {
                         ? !!reviewChoice && !!reviewChoice.correct
                         : choice.correct,
                 disabled: readOnly,
-                hasRationale: !!choice.clue,
+                hasRationale: !!choice.rationale,
                 // Render the rationale for the choice, as it may contain subwidgets as well.
-                rationale: renderRenderer(choice.clue),
+                rationale: renderRenderer(choice.rationale),
                 showRationale: rationaleShown,
```
SonicScrewdriver, line 285: I love this so much, thank you!

SonicScrewdriver (approved):
LOVE IT, thank you! 

I made sure to test with a nested passage-ref component (due to the removal of the `widgets` property) and it DOES still work, even on EditorPage. 

Screenshot 2025-05-14 at 10 26 37 AM

packages/perseus-editor/src/styles/perseus-editor.less
```diff
@@ -838,15 +838,15 @@ code {
         }
     }
 
-    .clue-editor {
+    .rationale-editor {
```
ivyolamit, line 841: ❤️ 💅🏼

ivyolamit (approved):
nice cleanup! 🎉 🫶 
LGTM 👍🏼

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.test.ts
anakaren-rojas: Ohh, I have a constant declared with the latest version number, I should use that here instead

anakaren-rojas:
getting webapp PR ready for review

packages/perseus-core/src/parse-perseus-json/regression-tests/item-data/hint-with-null-replace.json
```diff
@@ -17,14 +17,14 @@
           "choices": [
             {
               "content": "crwdns2693864:0crwdne2693864:0",
-              "clue": "",
+              "rationale": "",
```
jeremywiebe, line 20: This test data should not be updated.  @benchristel can explain it better, but these are v1.0 configurations and so will be migrated automatically by the parser.

What we probably want is some v3 versions of radio widgets that can form teh basis of the parser validation for v3 radio widgets.

packages/perseus-core/src/widgets/radio/radio-util.test.ts
```diff
@@ -10,44 +10,31 @@ describe("getRadioPublicWidgetOptions", () => {
...
                 {
                     content: "There is no such input value.",
                     isNoneOfTheAbove: true,
                     correct: true,
-                    clue: "This is a none of the above option.",
+                    rationale: "This is a none of the above option.",
                     // Note(tamara): Choices in a Radio don't use the widgets key, but
                     // the key exists in prod so we want to make sure we don't
                     // break things (unknown things) by confirming it passes
                     // through the getPulicWidgetOptions function unchanged.
```
Myranae, line 30: It seems like this comment was for the chunk that is getting deleted. I honestly can't remember adding it, but does it make sense to keep it? Maybe it makes sense to radio folks at the moment though 😆

PR #2436, opened 2025-04-30
Type NumberLine widget props correctly

benchristel:
The way our widget props work is way too complicated, IMO.

- The Renderer passes the WidgetOptions parsed from Perseus JSON to a
  transform function exported by the widget file. `staticTransform` is used
  if the assessment item is in "static" mode (with answers revealed); otherwise,
  `transform` is used.
- The result of the transform function is called `RenderProps`.
- The Renderer then merges the `RenderProps` with some other props, which I'm calling
  `UniversalWidgetProps` in this PR.
- The combined props ("external props") are passed to the widget
- The widget component class then applies `defaultProps`. The result is the widget's
  "internal props". The type name for internal props is usually `Props`.

This PR hopefully makes all of this clearer for the NumberLine widget, though I think
there's still a lot of room for improvement.

Issue: LEMS-2976

## Test plan:

`pnpm tsc`


packages/perseus/src/types.ts
```diff
@@ -141,6 +142,14 @@ export type ChangeHandler = (
...
         // Interactive Graph callback (see legacy: interactive-graph.tsx)
         graph?: PerseusGraphType;
+        // widgets/number-line.tsx
+        divisionRange?: number[];
+        // widgets/number-line.ts
+        numDivisions?: number;
+        // widgets/number-line.ts
+        numLinePosition?: number;
+        // widgets/number-line.ts
+        rel?: Relationship;
```
handeyeco, line 152: Not a result of your PR, but this is really gross. I get the sense that `ChangeHandler` was supposed to take a PerseusItem or PerseusRenderer or something and then other things were just thrown in...

packages/perseus-core/src/data-schema.ts
```diff
@@ -1260,7 +1262,7 @@ export type PerseusNumberLineWidgetOptions = {
     // This controls the number (and position) of the tick marks; you can either set the number of divisions (2 divisions would split the entire range in two halves), or the tick step (the distance between ticks) and the other value will be updated accordingly. Note:  There is no check to see if labels coordinate with the tick marks, which may be confusing for users if the blue labels and black ticks are off-step.
     tickStep?: number | null;
     // The correct relative value. default: "eq". options: "eq", "lt", "gt", "le", "ge"
-    correctRel?: string | null;
+    correctRel?: "eq" | "lt" | "gt" | "le" | "ge";
```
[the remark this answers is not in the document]
jeremywiebe: This has opened up an idea I hadn't thought of before, which is: where should Perseus widgets define their default option values? 

Historically, we've used React's default props to define the fallback (default) values for missing options. But as we move further away from React being the only place we have Perseus logic (ie. the SSS project), it becomes less and less sensible to use React concepts for Perseus logic.

I wonder if the approach we should be taking is to define default values in the parsers (I believe we (mostly) do this now) and then our Perseus widget components not provide _any_ default props (because they should be filled in by the parser by the time the component sees them).

packages/perseus-core/src/data-schema.ts
```diff
@@ -1260,7 +1262,7 @@ export type PerseusNumberLineWidgetOptions = {
     // This controls the number (and position) of the tick marks; you can either set the number of divisions (2 divisions would split the entire range in two halves), or the tick step (the distance between ticks) and the other value will be updated accordingly. Note:  There is no check to see if labels coordinate with the tick marks, which may be confusing for users if the blue labels and black ticks are off-step.
     tickStep?: number | null;
     // The correct relative value. default: "eq". options: "eq", "lt", "gt", "le", "ge"
```
jeremywiebe, line 1264: Can you also remove the part of the comment that defines the options? I find these types of comments are just ripe to age poorly and become inaccurate. Plus, with the addition of the actual union definition, it becomes extremely redundant.

packages/perseus/src/types.ts
jeremywiebe: At the start, it seemed like change props across our widgets used the same type signature. In reality, each widget's `onChange` prop should be defined as the set of options that the widget handles (and I think in some cases it's even more custom than that). 

In reality, the `Renderer` _does_ pass down the same `onChange` value to all widgets, so they do share the same type, and we've gotten away with it because [in the Renderer](https://github.com/Khan/perseus/blob/4ac22bdadcdad844ed9964485ef727de53c6dfa7/packages/perseus/src/renderer.tsx#L602) there's no typing on the function (lots of `any` in this area!).

packages/perseus/src/types.ts
```diff
@@ -561,7 +570,21 @@ export type WidgetProps<
     // Defines the arguments that can be passed to the `trackInteraction`
     // function from APIOptions for this widget.
     TrackingExtraArgs = Empty,
-> = RenderProps & {
+> = RenderProps & UniversalWidgetProps<Rubric, TrackingExtraArgs>;
+
+/**
+ * The props passed to every widget, regardless of its `type`.
+ */
+export type UniversalWidgetProps<
```
jeremywiebe, line 578: Perhaps this should be called something like `WidgetPropsFromRenderer` or `RendererWidgetProps`?

jeremywiebe (approved):
I love the refinement. I'd really like too see `docs/architecture.md` updated to reflect what we see as the idiomatic type names and what they mean. 

I'd also love it if we could converge on fewer types. I know they're a source of confusion, but I think they expose the complexity we have in Perseus that isn't obvious to those new to it. 

The flow of widget options from datastore through to the final widget's React component is complex. Maybe there's a way we can use TypeScript utility types to further reduce the number types each widget needs to define. 🤔  

Thanks for thinking and working on this!

packages/perseus-core/src/types.ts
```diff
@@ -44,8 +44,7 @@ export type InteractiveMarkerType = MarkerType & {
 };
 
 // Used for NumberLine
-// TODO: can this be merged with PerseusNumberLineWidgetOptions.correctRel?
-export type Relationship = "lt" | "gt" | "le" | "ge";
+export type Relationship = "eq" | "lt" | "gt" | "le" | "ge";
```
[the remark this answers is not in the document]
handeyeco: Couldn't we use `PerseusNumberLineWidgetOptions` and remove `Relationship`?

PR #2438, opened 2025-05-02
Use eslint-plugin-functional to ban mutating objects and arrays

benchristel:
We generally don't want to mutate objects and arrays, since React assumes that these
values are immutable. This PR adds a linter rule that blocks mutating operations.

See this [slack discussion] for context.

[slack discussion]: https://khanacademy.slack.com/archives/C01AZ9H8TTQ/p1744325813358959

ADR for adding the new dependency: [ADR #836]

[ADR #836]: https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/4016767174/ADR+836+Use+eslint-plugin-functional+to+ban+mutable+data+in+Perseus

Issue: LEMS-3084

## Test plan:

`pnpm lint`

### DIFF
--- a/.changeset/wise-pandas-perform.md
+++ b/.changeset/wise-pandas-perform.md
@@ -1,0 +1,15 @@
+---
+"@khanacademy/perseus-dev-ui": patch
+"@khanacademy/kas": patch
+"@khanacademy/kmath": patch
+"@khanacademy/math-input": patch
+"@khanacademy/perseus": patch
+"@khanacademy/perseus-core": patch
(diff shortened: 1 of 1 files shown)


ivyolamit:
Removed reviewers from radio widget team, PR doesn't affect on what we're doing.

handeyeco (commented):
I don't see anything intrinsically wrong with this. I wasn't involved with the ADR though and it seems like something Jeremy would want to look at, so I'll defer to him for the approval.

PR #2439, opened 2025-05-05
Make Perseus JSON parsers accept cross-realm objects

benchristel:
## Summary:
While looking at Sentry logs this morning, I noticed a weird error coming from
the Perseus parser:

```
At (root) -- expected object, but got {"answerArea":{"calculator":true,"chi2Table":false, ... }}
```

Expected object, but got an object? Huh??

The only thing I could think of is that the object is coming from another realm
(e.g. an iframe).  Previously our test for objectiness rejected cross-realm
objects.

This PR fixes the parser so cross-realm objects are recognized as objects.

Issue: none

## Test plan:

`pnpm test`

### DIFF
--- a/packages/perseus-core/src/parse-perseus-json/general-purpose-parsers/is-plain-object.ts
+++ b/packages/perseus-core/src/parse-perseus-json/general-purpose-parsers/is-plain-object.ts
@@ -3,6 +3,13 @@ const objectConstructorString = Object.prototype.constructor.toString();
 
 const functionToString = Function.prototype.toString;
 
+/**
+ * Checks whether a value is a plain JS object (e.g. one constructed with `{}`
+ * literal syntax or `Object.create(null)`. Returns true given a plain object
+ * from another realm.
+ *
+ * @param x - the value to check for objectiness
+ */
 export function isPlainObject(x: unknown): x is Record<keyof any, unknown> {
     if (x == null) {
         return false;


packages/perseus-core/src/parse-perseus-json/general-purpose-parsers/is-object.test.ts
```diff
@@ -0,0 +1,118 @@
...
+    it("is false for an array", () => {
+        expect(isObject([])).toBe(false);
+    });
+
+    it("is false for a class instance", () => {
+        // Even if our class is named Object, it's not the same as the global
+        // Object.
+        class Object {}
+        expect(isObject(new Object())).toBe(false);
+    });
```
handeyeco, line 22: I think the naming of `isObject` is a little confusing although I can't think of anything better off the top of my head. An array **_is_** an object, right? An instance of a class _**is**_ an object? Which is why I imagine we're not able to use `_.isObject` here (besides also trying to use less underscore).

`isPlainObject`? `isSerializableObject`? idk

packages/perseus-core/src/parse-perseus-json/general-purpose-parsers/is-object.ts
```diff
@@ -1,3 +1,18 @@
...
+        return false;
+    }
+    const prototype = Object.getPrototypeOf(x);
+    if (prototype == null) {
+        return true;
+    }
+    return (
+        typeof prototype.constructor === "function" &&
+        functionToString.call(prototype.constructor) === objectConstructorString
+    );
```
handeyeco, line 17: It's not a lot of code, but it is big-brain code (I rarely think of prototypes on a day-to-day basis anymore). I feel like it would be beneficial to have inline comments to reinforce intent.

handeyeco (approved):

jeremywiebe:
My gut says this originates from the Purseus iframe preview in the editor. To preview, the editor sets the edited item onto a global and then the preview iframe reaches up into the parent root frame to pull down the Perseus item. In my mind, that'd be exactly the kind of cross-realm issue you describe.

PR #2440, opened 2025-05-06
docs(changeset): New feature for PhET Widget to provide an option for full screen on mobile devices.

SonicScrewdriver:
## Summary:
This PR creates a fake fullscreen view in the mobile application for the PhET Widget, as we're currently unable to use the fullscreen API. It also adds a new apiOption for isMobileApp, which will allow Perseus to know when we're being served in one of the mobile applications. 

This PR will require updates in webapp in order to pass down the new isMobileApp variable as part of the apiOptions.

Additional support for the fullscreen API could _**possibly**_ be added to the Mobile application in the future, but there is currently no set release date or work for the next Mobile release. Given that, we've opted for a bit of a "hacky" temporary solution that will allow us to get the widget a little larger to improve usability. 

Unfortunately, full screen will not be supported in iOS web at this time. 

This direction was chosen due to several restrictions:

1. We're locked into a z-index in an upstream container that we cannot change. 
2. We have many variable headers, banners, and footers in the mobile web version that interfere with allowing this fix to also be served to mobile web. 
3. We can't inject the widget at the document root to work around the z-index issue as it would result in the loss of any actions that had been performed in the PhET iframe. 
4. This is a low usage widget. If we were to see increased usage and partnership, we could discuss improving the communication of state with PhET so that alternative solutions could be used in the future. 

Issue: LEMS-2922

## Test plan:
- Manual testing in mobile application 
- New test that confirms isMobileApp logic is being used.

### DIFF
--- a/packages/perseus/src/perseus-api.tsx
+++ b/packages/perseus/src/perseus-api.tsx
@@ -97,6 +97,9 @@ export const ApiOptions = {
         // Indicates whether or not to use mobile styling.
         isMobile: PropTypes.bool,
 
+        // Indicates whether or not to use mobile app styling.
+        isMobileApp: PropTypes.bool,
+
         // A function, called with a bool indicating whether use of the
         // drawing area (scratchpad) should be allowed/disallowed.
         // Previously handled by `Khan.scratchpad.enable/disable`
@@ -136,6 +139,7 @@ export const ApiOptions = {
     defaults: {
         isArticle: false,
         isMobile: false,
+        isMobileApp: false,
         onFocusChange: function () {},
         GroupMetadataEditor: StubTagEditor,
         showAlignmentOptions: false,
--- a/packages/perseus/src/widgets/phet-simulation/phet-simulation.stories.tsx
+++ b/packages/perseus/src/widgets/phet-simulation/phet-simulation.stories.tsx
@@ -1,6 +1,14 @@
+import {
+    generateTestPerseusItem,
+    type PerseusRenderer,
+} from "@khanacademy/perseus-core";
+import * as React from "react";
+
+import {ServerItemRendererWithDebugUI} from "../../../../../testing/server-item-renderer-with-debug-ui";
+
 import {PhetSimulation} from "./phet-simulation";
 
-import type {Meta, StoryObj} from "@storybook/react";
+import type {Meta} from "@storybook/react";
 
 const meta: Meta<typeof PhetSimulation> = {
     component: PhetSimulation,
@@ -8,11 +16,43 @@ const meta: Meta<typeof PhetSimulation> = {
 };
 
 export default meta;
-type Story = StoryObj<typeof PhetSimulation>;
 
-export const Primary: Story = {
-    args: {
-        url: "https://phet.colorado.edu/sims/html/projectile-data-lab/latest/projectile-data-lab_all.html",
-        description: "Projectile Data Lab",
-    },
+// Create a simple renderer with a PhET simulation widget
+const createPhETSimulation = (
+    url: string,
+    description: string,
+): PerseusRenderer => {
+    return {
+        content: "[[☃ phet-simulation 1]]",
+        images: {},
+        widgets: {
+            "phet-simulation 1": {
+                type: "phet-simulation",
+                graded: true,
+                static: false,
+                options: {
+                    url,
+                    description,
+                },
+                version: {
+                    major: 0,
+                    minor: 0,
+                },
+            },
+        },
+    };
+};
+
+export const Primary = (args: any): React.ReactElement => {
+    const question = createPhETSimulation(
+        "https://phet.colorado.edu/sims/html/projectile-data-lab/latest/projectile-data-lab_all.html",
+        "Projectile Data Lab",
+    );
+
+    return (
+        <ServerItemRendererWithDebugUI
+            item={generateTestPerseusItem({question})}
+            title="PhET Simulation"
+        />
+    );
 };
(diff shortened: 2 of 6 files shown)


packages/perseus/src/widgets/phet-simulation/phet-simulation.test.ts
```diff
@@ -138,4 +146,36 @@ describe("phet-simulation widget", () => {
...
+            await userEvent.click(fullscreenButton);
+        });
+
+        // Assert
+        await waitFor(() => {
+            const closeButton = screen.getByRole("button", {
+                name: "Exit fullscreen",
+            });
+            // Check that the mobile app fullscreen close button is present
+            expect(closeButton).toBeInTheDocument();
```
SonicScrewdriver, line 178: I COULD verify that the styles are set, but I wasn't sure if that would be necessary as they'll be applied using the same check that the `closeButton` is using.

packages/perseus/src/widgets/phet-simulation/phet-simulation.stories.tsx
```diff
@@ -1,18 +1,34 @@
...
 
-export const Primary: Story = {
+// Story showing the PhetSimulation component directly
+export const Default: Story = {
     args: {
         url: "https://phet.colorado.edu/sims/html/projectile-data-lab/latest/projectile-data-lab_all.html",
         description: "Projectile Data Lab",
+        apiOptions: {
+            ...ApiOptions.defaults,
+            isMobileApp: false,
```
jeremywiebe, line 31: Since the default value is `false`... why do we need to specify it here?

packages/perseus/src/widgets/phet-simulation/phet-simulation.tsx
```diff
@@ -165,15 +173,60 @@ export class PhetSimulation
...
+        // Use our fake fullscreen implementation for mobile
+        this.setState((prevState) => ({
+            isFullScreen: !prevState.isFullScreen,
+        }));
+    };
+
     render(): React.ReactNode {
+        // Extract state and props we'll use
+        const {isFullScreen, banner, url} = this.state;
+        const {isMobileApp} = this.props.apiOptions;
```
jeremywiebe, line 204: I wonder if we can side-step needing to create a new API Option and instead just use duck typing. 

I tried this in Safari on my iPhone and it appears that the `requestFullScreen` API doesn't exist there.  Would this work? 

```typescript
if (!this.iframeRef.requestFullScreen) {
    // Set up the fake full-screen stuff 
}
```

jeremywiebe (approved):
Approving, but before landing, please consider just checking for `requestFullScreen` on the DOM element instead of adding a new APIOption.  These new options live a long time and if we can avoid adding it, that'd be good.

packages/perseus/src/widgets/phet-simulation/phet-simulation.tsx
SonicScrewdriver: We're actually _**specifically**_ not supporting iOS web due to several issues. (Android web supports the fullscreen API). 

The problem is that using this solution on the mobile web results in a ton of visual glitches due to how webapp explicitly setting the z-index on a parent container, which means that we're locked behind all of the headers, and footers...without actually being bound by their positions/sizes. Notification banners seem to be at a lower z-index, which results in a very glitchy experience. :( 

image
Here's a quick example, 

The usual approach to fixing THAT issue is injecting the content into a new container at document root. Sadly, that won't work here, as we're using an iFrame to an external website — which would mean the simulation would lose all context of the user adjustments/settings when they switch between fullscreen and in-situ modes.

packages/perseus/src/widgets/phet-simulation/phet-simulation.test.ts
jeremywiebe: I think you could check if the overlay div was present (but I agree that checking styling might be overkill).  If the button is _inside_ the overlay div then you're fine with this.

packages/perseus/src/perseus-api.tsx
```diff
@@ -136,6 +139,7 @@ export const ApiOptions = {
     defaults: {
         isArticle: false,
         isMobile: false,
+        isMobileApp: false,
         onFocusChange: function () {},
```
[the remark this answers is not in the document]
jeremywiebe: I don't think that's needed.  This file is mostly the definition of APIOptions and its types... so I don't feel unit tests would protect that much.

packages/perseus/src/widgets/phet-simulation/phet-simulation.stories.tsx
SonicScrewdriver: Ooo good callout, we don't!

packages/perseus/src/widgets/phet-simulation/phet-simulation.tsx
jeremywiebe: For sure. I'm not arguing against special treatment of iOS, just that I think we can use a check for the existence of `iframeRef.requestFullScreen` as the thing we detect instead of adding a new API Option.

packages/perseus/src/widgets/phet-simulation/phet-simulation.tsx
```diff
@@ -256,6 +331,37 @@ const styles = StyleSheet.create({
...
     },
+    // Mobile app fullscreen implementation styles
+    appFullScreenWidgetContainer: {
+        position: "fixed",
+        top: 0,
+        left: 0,
+        right: 0,
+        // Ensure that the Check Answer bar does not cover the bottom of the simulation
+        bottom: MOBILE_APP_BOTTOM_BAR_HEIGHT,
+        width: "100%",
```
mark-fitzgerald, line 342: Is this needed since the `left` and `right` position values are specified?

mark-fitzgerald (approved):
Thanks for taking on this tough issue to provide a better experience for mobile users.

PR #2441, opened 2025-05-06
Answerless Orderer

Myranae:
## Summary:
Update the Orderer widget for answerless rendering. Small updates to remove correctOptions and confirm types are correct. Added tests and answerless story.

Issue: LEMS-2979

## Test plan:
- Confirm all checks pass
- Confirm widget still acts as expected
- No changes should be noted in user experience

### DIFF
--- a/packages/perseus/src/widgets/orderer/orderer.test.ts
+++ b/packages/perseus/src/widgets/orderer/orderer.test.ts
@@ -1,13 +1,31 @@
-import {act} from "@testing-library/react";
+import {act, screen} from "@testing-library/react";
 
 import {testDependencies} from "../../../../../testing/test-dependencies";
 import * as Dependencies from "../../dependencies";
-import {scorePerseusItemTesting} from "../../util/test-utils";
+import {getAnswerfulItem, getAnswerlessItem, scorePerseusItemTesting} from "../../util/test-utils";
 import {renderQuestion} from "../__testutils__/renderQuestion";
 
 import {question2} from "./orderer.testdata";
 
 import type {APIOptions} from "../../types";
+import {PerseusOrdererWidgetOptions} from "@khanacademy/perseus-core";
+import {scorePerseusItem} from "@khanacademy/perseus-score";
+
+const ordererOptions: PerseusOrdererWidgetOptions = {
+    otherOptions: [],
+    layout: "horizontal",
+    options: [
+        {content: "1", images: {}, widgets: {}},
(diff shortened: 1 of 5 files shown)


packages/perseus/src/widgets/orderer/orderer.test.ts
```diff
@@ -42,60 +60,100 @@ describe("orderer widget", () => {
...
+            // Act
+            const [orderer] = renderer.findWidgets("orderer 1");
+            act(() => orderer.setListValues(["37"]));
+            const userInput = renderer.getUserInputMap();
+
+            // Assert
+            expect(userInput).toEqual({"orderer 1": {current: ["37"]}});
+        });
+
+        it("can given an invalid score", () => {
```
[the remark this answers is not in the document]
Myranae: Fixed!

packages/perseus/src/widgets/orderer/orderer.test.ts
```diff
@@ -1,13 +1,31 @@
-import {act} from "@testing-library/react";
+import {scorePerseusItem} from "@khanacademy/perseus-score";
+import {act, screen} from "@testing-library/react";
 
 import {testDependencies} from "../../../../../testing/test-dependencies";
 import * as Dependencies from "../../dependencies";
-import {scorePerseusItemTesting} from "../../util/test-utils";
```
handeyeco, line 5: Now that the scorers don't need strings, I feel like we should just get rid of `scorePerseusItemTesting`. Kind of pointless now, so good impulse to just use `scorePerseusItem`.

packages/perseus/src/widgets/orderer/orderer.tsx
```diff
@@ -527,41 +526,31 @@ class Orderer
...
-                },
-                this,
-            );
+            list.forEach((opt, i) => {
+                const card = ReactDOM.findDOMNode(
+                    // eslint-disable-next-line react/no-string-refs
+                    this.refs["sortable" + i],
+                );
+                // @ts-expect-error - TS2769 - No overload matches this call. | TS2339 - Property 'outerWidth' does not exist on type 'JQueryStatic'.
+                const outerWidth = $(card).outerWidth(true);
```
handeyeco, line 535: It might be good to remove some of these `@ts-expect-error` when possible since they bury **all** problems. I think this could have been something like:

```suggestion
                const outerWidth = ($(card) as any).outerWidth(true);
```

handeyeco (approved):

PR #2444, opened 2025-05-06
Ensure Orderer Editor updates Options property in change callback

SonicScrewdriver:
## Summary:
We were facing a bug with the Orderer Editor where the state was always a step out-of-date. Thanks to @handeyeco 's insights, it sounds like we've deprecated `serialize`, and we're now handling everything in the onChange events. Unfortunately, the logic for `onOptionsChange` did not handle the updating of the dynamic options property, thus resulting in it being out of sync while editing pre-existing content. 

This PR adds the missing logic to `onOptionsChange`, while also modernizing it to remove underscore. 

Issue: LEMS-3082

## Test plan:
- New tests for onChange response AND sort order 
- Confirmed fixes work upstream via snapshot testing

### DIFF
--- a/packages/perseus-editor/src/widgets/orderer-editor.tsx
+++ b/packages/perseus-editor/src/widgets/orderer-editor.tsx
@@ -61,11 +61,8 @@ class OrdererEditor extends React.Component<Props> {
         const updatedOptions = [
             ...new Set(allOptions.map((item) => item.content)),
         ]
-            // filter out empty strings
             .filter((content) => content !== "")
-            // Alphabetical sort
             .sort()
-            // Category sort
             .sort((a, b) => {
                 const getCategoryScore = (content) => {
                     // 1. Any content that contains numbers
--- a/packages/perseus-editor/src/widgets/orderer-editor.tsx
+++ b/packages/perseus-editor/src/widgets/orderer-editor.tsx
@@ -58,20 +58,14 @@ class OrdererEditor extends React.Component<Props> {
         const allOptions = [...correctOptionsToUse, ...otherOptionsToUse];
 
         // Get unique content items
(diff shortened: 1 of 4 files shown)


packages/perseus-editor/src/widgets/orderer-editor.tsx
```diff
@@ -41,6 +41,50 @@ class OrdererEditor extends React.Component<Props> {
         props[whichOptions] = _.map(options, function (option) {
             return {content: option};
         });
+
+        // Get content from correctOptions (either updated or existing)
+        const correctOptionsToUse =
```
handeyeco, line 46: `onOptionsChange` seems to be doing a lot and it's a little tricky to test because it's attached to a React component. Also now the general concept is duplicated between this function and `serialize`.

What if we extracted most of the logic into a helper (could even stay in this file) so that we can have more complete test coverage without having to muck about with React or PerseusItems. Then we could do something like:

```TS
    onOptionsChange: (
        arg1: "correctOptions" | "otherOptions",
        arg2: any,
        arg3: any,
    ) => any = (whichOptions, options, cb) => {
        const next = generateNextOptions(
            whichOptions,
            options,
            this.props.correctOptions,
            this.props.otherOptions
        )
        this.props.onChange(next, cb);
    };
```

Benefits:
1. Cleaner component
2. Less duplication
3. Easier to test

Not a blocker, just a suggestion.

handeyeco (approved):

packages/perseus-editor/src/widgets/orderer-editor.tsx
SonicScrewdriver: Makes perfect sense to me, I'll get right on it!

PR #2447, opened 2025-05-07
Minor test cleanup

handeyeco:
## Summary:
I was looking at something for someone and noticed that this test could be cleaner.

### DIFF
--- a/packages/perseus/src/__tests__/server-item-renderer.test.tsx
+++ b/packages/perseus/src/__tests__/server-item-renderer.test.tsx
@@ -188,7 +188,7 @@ describe("server item renderer", () => {
     it("should call the answerable callback when no widgets are empty", async () => {
         // Arrange
         const answerableCallback = jest.fn();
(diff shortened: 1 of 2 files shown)


packages/perseus/src/__tests__/server-item-renderer.test.tsx
```diff
@@ -201,23 +201,11 @@ describe("server item renderer", () => {
...
-                    }}
-                    item={itemWithMockWidget}
-                    problemNum={1} // to force componentDidUpdate
-                    reviewMode={false}
-                    dependencies={testDependenciesV2}
-                />
-                ,
-            </RenderStateRoot>,
-        );
+        await userEvent.type(screen.getByRole("textbox"), "-42");
```
jeremywiebe, line 208: Oh, we don't have to re-render!  Yeah, that makes sense.

jeremywiebe (approved):

PR #2449, opened 2025-05-07
Expression Editor: UI update to clarify how things work

jeremywiebe:
## Summary:

I spent an hour cleaning up a bunch of UI stuff in the `expression` editor. Namely: 

  * Change the 'considered' (`correct`, `incorrect`, `ungraded`) options to be a button group so it's more obvious that you can actually change it
  * Moved tooltip `?` to be at end of field label instead of trailing input
  * Added a gap between answers
  * Stretch "Add new answer" to full width so it looks nicer

| Before | After | 
| --- | --- | 
| image  |  image |

Issue: "none" 

## Test plan:

Review UI changes and code

### DIFF
--- a/packages/perseus-editor/src/widgets/expression-editor.tsx
+++ b/packages/perseus-editor/src/widgets/expression-editor.tsx
@@ -358,75 +358,94 @@ class ExpressionEditor extends React.Component<Props, State> {
 
                 <div className={css(styles.paddedY)}>
                     <LabeledTextField
-                        label="Visible label"
+                        label={
+                            <>
+                                Visible label
+                                <InfoTip>
+                                    <p>
+                                        Optional visible text; strongly
+                                        encouraged to help learners using
+                                        dictation software, but can be omitted
+                                        if the surrounding content provides
+                                        enough context.
+                                    </p>
+                                </InfoTip>
+                            </>
+                        }
                         value={this.props.visibleLabel || ""}
                         onChange={this.handleVisibleLabel}
                     />
-                    <InfoTip>
-                        <p>
-                            Optional visible text; strongly encouraged to help
-                            learners using dictation software, but can be
-                            omitted if the surrounding content provides enough
-                            context.
-                        </p>
-                    </InfoTip>
                 </div>
 
                 <div className={css(styles.paddedY)}>
                     <LabeledTextField
-                        label="Aria label"
-                        value={this.props.ariaLabel || ""}
-                        onChange={this.handleAriaLabel}
-                    />
-                    <InfoTip>
-                        <p>
-                            Label text that&apos;s read by screen readers.
-                            Highly recommend adding a label here to ensure your
-                            exercise is accessible. For more information on
-                            writting accessible labels, please see{" "}
-                            <a
-                                href="https://www.w3.org/WAI/tips/designing/#ensure-that-form-elements-include-clearly-associated-labels"
+                        label={
+                            <>
+                                Aria label
+                                <InfoTip>
+                                    <p>
+                                        Label text that&apos;s read by screen
+                                        readers. Highly recommend adding a label
+                                        here to ensure your exercise is
+                                        accessible. For more information on
+                                        writting accessible labels, please see{" "}
+                                        <a
+                                            href="https://www.w3.org/WAI/tips/designing/#ensure-that-form-elements-include-clearly-associated-labels"
                                 target="_blank"
                                 rel="noreferrer"
                             >
                                 this article.
                             </a>
                         </p>
-                    </InfoTip>
+                                </InfoTip>
+                            </>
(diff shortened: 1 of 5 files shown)


packages/perseus-editor/src/widgets/expression-editor.tsx
```diff
@@ -436,19 +447,18 @@ class ExpressionEditor extends React.Component<Props, State> {
 
                 <HeadingSmall>Answers</HeadingSmall>
 
-                <p style={{margin: "4px 0"}}>
+                <Caption style={{fontStyle: "italic"}}>
```
jeremywiebe, line 450: nit: Just moving to use WB typography.

packages/perseus-editor/src/widgets/expression-editor.tsx
```diff
@@ -539,64 +549,64 @@ class AnswerOption extends React.Component<
...
-                                will render this response &quot;Your answer is
-                                not fully expanded and simplified.&quot;
-                            </p>
-                        </InfoTip>
-                    </div>
-
-                    <div className={css(styles.buttonRow, styles.paddedY)}>
-                        {removeButton}
-                    </div>
+                <ButtonGroup
```
jeremywiebe, line 554: This is the core of the change. Instead of a coloured div that you have a to know to click on to change, we present a button group where all options are visible and the selected option is highlighted! 

image

packages/perseus/src/components/button-group.tsx
```diff
@@ -66,12 +73,13 @@ class ButtonGroup extends React.Component<Props> {
                 <button
                     title={button.title}
                     type="button"
-                    id={"" + i}
```
jeremywiebe, line 69: I can't imagine a world where having HTML elements id'd by a single number being useful or correct.  These buttons are being ID'd as `<button id="0" .../>` which is just ripe for ID collisions so I'm 99.99999% that this is never used and legacy!  As soon as we have more than one `<ButtonGroup>` on a page we'd have id collisions. 

Per [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/id):
> The id [global attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes) defines an identifier (ID) which must be unique in the whole document.

packages/perseus-editor/src/widgets/expression-editor.tsx
```diff
@@ -353,80 +355,89 @@ class ExpressionEditor extends React.Component<Props, State> {
...
                 <HeadingSmall>Global Options</HeadingSmall>
 
                 <div className={css(styles.paddedY)}>
                     <LabeledTextField
-                        label="Visible label"
+                        label={
+                            <>
+                                Visible label
+                                <InfoTip>
+                                    Optional visible text; strongly encouraged
```
ivyolamit, line 367: Not sure what's our convention on texts. Are we not doing translations with our editor?

This is probably a preference, it would help to make the code look cleaner if we have these text defined somewhere.

```suggestion
                                <InfoTip>{optionalTextInfo}</InfoTip>
```

packages/perseus-editor/src/widgets/expression-editor.tsx
ivyolamit: Is there a reason why we don't use `StyleSheet` definition here and use inline styles instead?

packages/perseus-editor/src/widgets/expression-editor.tsx
jeremywiebe: We don't do localization in the editor so I think leaving this inline is more readable/maintainable than adding the indirection of a variable containing the text.  If we needed to localize, then I'd definitely make changes as you suggest.

packages/perseus-editor/src/widgets/expression-editor.tsx
```diff
@@ -353,80 +355,89 @@ class ExpressionEditor extends React.Component<Props, State> {
...
-                            Optional visible text; strongly encouraged to help
-                            learners using dictation software, but can be
-                            omitted if the surrounding content provides enough
-                            context.
-                        </p>
-                    </InfoTip>
                 </div>
 
                 <div className={css(styles.paddedY)}>
                     <LabeledTextField
```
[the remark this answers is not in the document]
jeremywiebe: Yes... but... I'd like to avoid that as it's a new package we don't depend on yet and I don't want to expand this PR that much. :)

packages/perseus-editor/src/widgets/expression-editor.tsx
jeremywiebe: Nope, just didn't think of it. I'll make that change.

ivyolamit (approved):

PR #2453, opened 2025-05-07
Render LaTeX in Free Response Widget questions

aag:
## Summary:
This commit adds code to render LaTeX formatted questions in the
Free Response Widget. The specific use case we want to support is
including Math expressions in the question. We use the same Renderer
that is used to render LaTeX elsewhere, but we don't want to support
nested widgets in the question, so we detect that as an issue when
saving in the editor.

Note that the spacing around the question is changed in this commit and
it doesn't match the designs, but that will be fixed in a later change.

Issue: https://khanacademy.atlassian.net/browse/AX-964

## Test plan:
- View the Free Response Widget in Storybook and verify that the
  "Question with Katex" story renders properly.

### DIFF
--- a/packages/perseus-editor/src/widgets/__stories__/free-response-editor.stories.tsx
+++ b/packages/perseus-editor/src/widgets/__stories__/free-response-editor.stories.tsx
@@ -45,6 +45,12 @@ export const ThreeCriteria: Story = {
     },
 };
 
+export const QuestionWithWidget: Story = {
+    args: {
+        question: "[[\u2603 radio 1]]",
+    },
+};
+
 type State = Partial<PropsFor<typeof FreeResponseEditor>>;
 
 const WithState = () => {
--- a/packages/perseus/src/widgets/free-response/free-response.stories.tsx
+++ b/packages/perseus/src/widgets/free-response/free-response.stories.tsx
@@ -36,3 +36,13 @@ export const UnlimitedCharacters: Story = {
         question: "What is the theme of the essay?",
     },
 };
+
+export const QuestionWithKatex: Story = {
+    args: {
+        allowUnlimitedCharacters: true,
+        characterLimit: 500,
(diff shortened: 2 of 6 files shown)


packages/perseus/src/widgets/free-response/free-response.stories.tsx
```diff
@@ -36,3 +36,13 @@ export const UnlimitedCharacters: Story = {
         question: "What is the theme of the essay?",
     },
 };
+
+export const QuestionWithKatex: Story = {
```
jeremywiebe, line 40: KaTeX isn't used anymore. We've been settling into using simply `TeX` to refer to math markup.

packages/perseus-editor/src/widgets/__stories__/free-response-editor.stories.tsx
```diff
@@ -45,6 +45,12 @@ export const ThreeCriteria: Story = {
     },
 };
 
+export const QuestionWithWidget: Story = {
```
jeremywiebe, line 48: **suggestion**: add a comment to explain why we have this story and what to expect. Reading this the first time through I couldn't understand why you had a radio widget in here and why it was special. 

```suggestion
/** 
  * An example story showing how the editor looks when 
  * a widget is detected in the free response widget configuration. Widgets are not allowed in the free 
  * response widget. 
  */
export const QuestionWithWidget: Story = {
```

jeremywiebe (commented):
Good stuff. I mentioned in a comment, but we don't use KaTeX at Khan Academy anymore. We also try to refer to math simply as ["TeX"](https://khanacademy.atlassian.net/wiki/spaces/LC/pages/2200469646/LEMS+KaTeX+-+MathJax+migration#TeX) because LaTeX is tooling, not the markup language. I know, I'm being pedantic. 😬

jeremywiebe (approved):

PR #2456, opened 2025-05-07
Remove pre-install check which warns if you aren't using pnpm - super slow!

jeremywiebe:
## Summary:

When we first moved to `pnpm`, we added a `preinstall` script which ran an npm package that complained loudly if you weren't using `pnpm` (ie. you might have "muscle memory" and just run `yarn install` without thinking). 

This extra check adds about 4 seconds to every `pnpm install`. I think we're at a point where we're mostly used to `pnpm`, so I'm removing it to save everyone this time.

#### With check 

```shell
$ hyperfine 'pnpm install'
Benchmark 1: pnpm install 
  Time (mean ± σ):      4.589 s ±  0.063 s    [User: 5.116 s, System: 1.032 s]
  Range (min … max):    4.474 s …  4.671 s    10 runs
```

#### Without check (this PR)

```shell
$ hyperfine 'pnpm install'
Benchmark 1: pnpm install 
  Time (mean ± σ):      1.041 s ±  0.025 s    [User: 1.426 s, System: 0.325 s]
  Range (min … max):    1.013 s …  1.081 s    10 runs
```

Issue: "none"

## Test plan:

`pnpm install` -> works!


handeyeco (approved):

Myranae (approved):
`pnpm install` worked just fine. :) Thanks for updating this!

PR #2457, opened 2025-05-08
Proof-of-concept: tighten up prop types for Categorizer

benchristel:
## Summary:
It's come to my attention that our prop types for many widgets don't quite
reflect reality, due to the internal and external props of class components
not being handled appropriately. E.g. in some places, we are deriving the
widget's internal props from the `RenderProps` returned from the `transform`
and `staticTransform` functions, which isn't correct — the _external_ props
need to match `RenderProps`.

This PR is a proof of concept for what a fix might look like.
The proposed design is currently documented in the linked issue, [LEMS-3086].
I will document it in architecture.md if we end up moving forward with this.

[LEMS-3086]: https://khanacademy.atlassian.net/browse/LEMS-3086

Issue: LEMS-3086

## Test plan:

`pnpm tsc`

### DIFF
--- a/packages/perseus/src/widgets/categorizer/categorizer.tsx
+++ b/packages/perseus/src/widgets/categorizer/categorizer.tsx
@@ -33,6 +33,8 @@ interface InternalProps extends UniversalWidgetProps {
 
 type ExternalProps = PropsFor<typeof Categorizer>;
 
+type RenderProps = Omit<ExternalProps, keyof UniversalWidgetProps>;
+
 type DefaultProps = Pick<
     InternalProps,
     "items" | "categories" | "values" | "linterContext"
@@ -293,8 +295,6 @@ const styles = StyleSheet.create({
         color: "#888",
     },
 });
-
-type RenderProps = Omit<ExternalProps, keyof UniversalWidgetProps>;
 
 export default {
     name: "categorizer",
--- a/packages/perseus/src/widgets/categorizer/categorizer.tsx
+++ b/packages/perseus/src/widgets/categorizer/categorizer.tsx
@@ -15,7 +15,12 @@ import mediaQueries from "../../styles/media-queries";
 import sharedStyles from "../../styles/shared";
 import {getPromptJSON as _getPromptJSON} from "../../widget-ai-utils/categorizer/categorizer-ai-utils";
 
-import type {UniversalWidgetProps, Widget, WidgetExports} from "../../types";
+import type {
+    ChangeFn,
+    UniversalWidgetProps,
+    Widget,
+    WidgetExports,
+} from "../../types";
 import type {CategorizerPromptJSON} from "../../widget-ai-utils/categorizer/categorizer-ai-utils";
 import type {
     PerseusCategorizerWidgetOptions,
@@ -68,8 +73,7 @@ export class Categorizer
         return {values: props.values};
     }
 
-    change: (...args: ReadonlyArray<unknown>) => any = (...args) => {
-        // @ts-expect-error - TS2345 - Argument of type 'readonly unknown[]' is not assignable to parameter of type 'any[]'.
+    change: (...args: Parameters<ChangeFn>) => any = (...args) => {
         return Changeable.change.apply(this, args);
     };
 
--- a/packages/perseus/src/widgets/categorizer/categorizer.tsx
+++ b/packages/perseus/src/widgets/categorizer/categorizer.tsx
@@ -15,31 +15,35 @@ import mediaQueries from "../../styles/media-queries";
 import sharedStyles from "../../styles/shared";
 import {getPromptJSON as _getPromptJSON} from "../../widget-ai-utils/categorizer/categorizer-ai-utils";
 
-import type {Widget, WidgetExports, WidgetProps} from "../../types";
+import type {UniversalWidgetProps, Widget, WidgetExports} from "../../types";
 import type {CategorizerPromptJSON} from "../../widget-ai-utils/categorizer/categorizer-ai-utils";
 import type {
     PerseusCategorizerWidgetOptions,
     PerseusCategorizerUserInput,
     CategorizerPublicWidgetOptions,
 } from "@khanacademy/perseus-core";
+import type {PropsFor} from "@khanacademy/wonder-blocks-core";
 
-type Props = WidgetProps<RenderProps> & {
-    values: ReadonlyArray<string>;
-};
+interface InternalProps extends UniversalWidgetProps {
+    values: number[];
+    items: PerseusCategorizerWidgetOptions["items"];
+    categories: PerseusCategorizerWidgetOptions["categories"];
+    randomizeItems: PerseusCategorizerWidgetOptions["randomizeItems"];
+}
 
-type DefaultProps = {
-    items: Props["items"];
-    categories: Props["categories"];
-    values: Props["values"];
-    linterContext: Props["linterContext"];
-};
+type ExternalProps = PropsFor<typeof Categorizer>;
+
+type DefaultProps = Pick<
+    InternalProps,
+    "items" | "categories" | "values" | "linterContext"
+>;
 
 type State = {
     uniqueId: string;
 };
 
 export class Categorizer
-    extends React.Component<Props, State>
+    extends React.Component<InternalProps, State>
     implements Widget
 {
     static contextType = PerseusI18nContext;
@@ -56,7 +60,9 @@ export class Categorizer
         uniqueId: _.uniqueId("perseus_radio_"),
     };
 
-    static getUserInputFromProps(props: Props): PerseusCategorizerUserInput {
+    static getUserInputFromProps(
+        props: InternalProps,
+    ): PerseusCategorizerUserInput {
         return {values: props.values};
     }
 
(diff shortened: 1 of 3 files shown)


packages/perseus/src/widgets/categorizer/categorizer.tsx
```diff
@@ -15,31 +15,42 @@ import mediaQueries from "../../styles/media-queries";
...
-    linterContext: Props["linterContext"];
-};
+type ExternalProps = PropsFor<typeof Categorizer>;
+
+type RenderProps = Omit<ExternalProps, keyof UniversalWidgetProps>;
+
+type DefaultProps = Pick<
+    InternalProps,
+    "items" | "categories" | "values" | "linterContext"
+>;
```
jeremywiebe, line 46: This is an idea that goes past this PR, but I wanted to note it. I suspect that for most widgets, the _only_ reason we use/need default props is to support usages in the editors. 

If we moved the default behaviour _into_ each widget's editor component, we could remove the defaulting of props from widgets entirely. Any defaulting at runtime would be handled by the parser and our types could become a bit simpler here.

packages/perseus/src/widgets/categorizer/categorizer.tsx
```diff
@@ -15,31 +15,42 @@ import mediaQueries from "../../styles/media-queries";
...
     PerseusCategorizerWidgetOptions,
     PerseusCategorizerUserInput,
     CategorizerPublicWidgetOptions,
 } from "@khanacademy/perseus-core";
+import type {PropsFor} from "@khanacademy/wonder-blocks-core";
 
-type Props = WidgetProps<RenderProps> & {
-    values: ReadonlyArray<string>;
-};
+interface InternalProps extends UniversalWidgetProps {
```
jeremywiebe, line 32: I have a small reaction to renaming this to `InternalProps`. I realize that's the usage of this type, but it feels a bit like we're breaking convention (a vast majority of code both inside Perseus and in other codebases simply refers to this type as `Props`). 

I'm undecided which way I land but thought I'd comment in case other's feel more strongly.

jeremywiebe: For metrics, in Perseus, we the top 5 most popular type name for Props is: 

  * 700 - `Props`
  * 420 - `ChangeProps`
  * 130 - `DefaultProps`
  * 125 - `MafsGraphProps`
  * 66 - `WidgetProps`

In webapp, that numbers are:

 * 5512 - `Props`
 * 493 - `ExportProps`
 * 312 - `OwnProps`
 * 269 - `DefaultProps`
 * 224 - `WithApolloProps`

jeremywiebe (approved):

packages/perseus/src/widgets/categorizer/categorizer.tsx
```diff
@@ -15,31 +15,42 @@ import mediaQueries from "../../styles/media-queries";
...
 
-type DefaultProps = {
-    items: Props["items"];
-    categories: Props["categories"];
-    values: Props["values"];
-    linterContext: Props["linterContext"];
-};
+type ExternalProps = PropsFor<typeof Categorizer>;
+
+type RenderProps = Omit<ExternalProps, keyof UniversalWidgetProps>;
```
handeyeco, line 41: I'm having trouble grokking this (no more than usual though). Here's my understanding...

1. We define `InternalProps` which is a combination of shared widget props, a partial of widget options, and added properties from `transform`
2. We give `InternalProps` to the component
3. We pull the props from the component into `ExternalProps`
4. We take `ExternalProps` and remove shared which props which makes `RenderProps`

Is the idea that React applies `DefaultProps` to `InternalProps`? Which is why we can't use `type ExternalProps = InternalProps`? (But that doesn't really make sense because the External props would be the ones without defaults...so that must not be it...)

IF that's the case and IF we can't remove `DefaultProps` altogether, I wonder if it wouldn't be clearer to do something like:

``` TS
// ChatGPT gave me this helper function, I don't know if it works
type ApplyDefaults<Props, Defaults> = Omit<Props, keyof Defaults> & Defaults;

// The type in the datastore
type WidgetOptions = {
  somethingFromWidgetOptions?: number;
}

// The result from the transform
type RenderProps = Pick<WidgetOptions, "somethingFromWidgetOptions"> & {
  somethingFromTransform: boolean;
}

// Adding shared widget props
type ExternalProps =  UniversalWidgetProps & RenderProps

// The type for the component's `defaultProps`
type DefaultProps = {
  somethingFromWidgetOptions: number;
}

// The type within the component (although wouldn't React apply defaults to props?)
type InternalProps = ApplyDefaults<ExternalProps, DefaultProps>
```

Okay, I'm more confused not than when I started...

handeyeco: Even _more_ confused after looking at `data-schema` because `PerseusCategorizerWidgetOptions.items` isn't optional, so why does it have a default? I'm guessing that's what Jeremy was saying about the editing experience which might not be relevant anymore?

handeyeco: ChatGPT gave me another helper if we didn't want to define `DefaultProps` explicitly

``` TS
type NonOptional<T, K extends keyof T> = Omit<T, K> & Required<Pick<T, K>>;

// The type for the component's `defaultProps`
type DefaultProps = {
  somethingFromWidgetOptions: NonOptional<ExternalProps, "somethingFromWidgetOptions">;
}
```

handeyeco: Maybe it'd be nice to take out some of the leg work...

``` TS
// ChatGPT gave me this helper function, I don't know if it works
type ApplyDefaults<Props, Defaults> = UniversalWidgetProps & Omit<Props, keyof Defaults> & Defaults;

// The type coming out of the parser
type WidgetOptions = {
  somethingFromWidgetOptions?: number;
}

// The result from the transform
type TransformedProps = Pick<WidgetOptions, "somethingFromWidgetOptions"> & {
  somethingFromTransform: boolean;
}

// The type for the component's `defaultProps`
type DefaultProps = {
  somethingFromWidgetOptions: NonOptional<TransformedProps, "somethingFromWidgetOptions">;
}

// The type within the component
type Props = ApplyDefaults<TransformedProps, DefaultProps>

class MyWidget extends React.Component<Props> {
```

handeyeco: > Maybe we should consider converting our class components to functional

I think that would be time better spent for sure

packages/perseus/src/widgets/categorizer/categorizer.tsx
jeremywiebe: So thinking a bit more about this I think a descriptive name makes sense especially for class components. I think you're addressing the fact that `InternalProps` are what the component sees from inside and `ExternalProps` are what consumers of the component see (mainly differing because of DefaultProps), right? 

Further, for functional components, this can go away because we define `Props` as what the consumer sees and then the defaulting happens when we destructure in the function declaration, relieving us from needing a separately defined type for the internal view of Props. 

TLDR; I'm good with special naming for our class components.

packages/perseus/src/widgets/categorizer/categorizer.tsx
jeremywiebe: > > Maybe we should consider converting our class components to functional

> I think that would be time better spent for sure

I agree with Matthew, but the thread gave me two ideas...

  1. How would this look if we wrapped each widget's React component that was a class component in a functional one. That would give us a path to having cleaner types but relieve us from having to port _all_ widget components to functional _right now_. 
  Also, then all the remaining widget components that are classes would not have _any_ optional props as they'd be filled in by defaults in the wrapper functional component. 

  2. Are we making things more complicated by trying to use so many TypeScript utility types (`Omit<>`, `Pick<>`, etc). Would we be better off to just define each transitional type explicitly?
  For example:
  ```typescript 
  type WidgetOptions = {
    requiredField: string;
    optionalField?: number;
  } 

  type RenderProps = {
    requiredField: string;
    optionalField?: number;
    propFromTransform: number;
  }

  type Props = {
    requiredField: string;
    optionalField: number;
    propFromTransform: number;
  } & UniversalWidgetProps;

  type DefaultProps = {
    optionalField: number;
  } 
  ```

  Ultimately, I feel like all the type "spreading" across the different types is giving us a false sense that we're keeping things in sync, when really we end up fighting the optionality alot. 

Lastly, these types are complicated, but they reflect the complexity of the pipeline our data goes through to be rendered. So I think there's a certain amount of necessary complexity here.

jeremywiebe: > Do you mean that props that have defaultProps appear to be required when they're actually optional? That's one of the problems this new approach aims to solve, actually.

Yes, that's what I meant. In my mind, avoiding utility types would be another way to solve that.  But I don't see this as a deal-breaker so I'm good with crafting some utility types that help us to define these different pieces consistently.

jeremywiebe: > However, in the case of widget props, utility types are a workaround for missing type coverage in the Renderer. Utility types let us ensure our types are consistent with each other, since the typechecker isn't able to do that for us in this case.
> 
> I know I sound like a broken record but I'm going to keep repeating it: we don't have type safety at the boundary where the renderer passes props to the widget. That's why we have to jump through all these hoops to make sure our type definitions don't get out of sync.

100% agree! I think the parser helps here and ideally the Renderer would only care about the props an types that _it_ provides. The widget options coming from `data-schema.ts` would be trusted as accurate in the Renderer. 

So, it seems we have two options here:

1) Wrap all React class component widget's in a functional component
2) Figure out these type conventions. 

I think I'm leaning to do #1 and avoid trying to fix these types for class components altogether. 

Thoughts?

PR #2458, opened 2025-05-08
Answerless Grapher

handeyeco:
## Summary:
Making sure Grapher can work in an Answerless world. For the most part this PR:

- Updates all Cypress tests to test both answerful and aswerless¹
- Adds an Answerless story
- Removes Underscore from Grapher

¹ Grapher is an inaccessible widget that requires drag-and-drop behavior that [can't be tested with React Testing Library](https://github.com/testing-library/user-event/issues/440#issuecomment-685010755), which is why we use Cypress

Issue: LEMS-2985

## Test plan:
Grapher should continue to work as it's always worked.

### DIFF
--- a/packages/perseus/src/widgets/grapher/grapher.cypress.ts
+++ b/packages/perseus/src/widgets/grapher/grapher.cypress.ts
@@ -1,3 +1,8 @@
+import {
+    generateTestPerseusItem,
+    splitPerseusItem,
+} from "@khanacademy/perseus-core";
+
 import renderQuestionWithCypress from "../../../../../testing/render-question-with-cypress";
 import {cypressTestDependencies} from "../../../../../testing/test-dependencies";
 import * as Perseus from "../../index";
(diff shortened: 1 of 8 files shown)


packages/perseus/src/widgets/grapher/grapher.cypress.ts
```diff
@@ -1,3 +1,8 @@
+import {
```
handeyeco, line 1: It will help to hide whitespace when reviewing this PR.

packages/perseus/src/widgets/grapher/grapher.tsx
```diff
@@ -250,10 +248,12 @@ class FunctionGrapher extends React.Component<FunctionGrapherProps> {
                             ) {
                                 const extraConstraint =
                                     this.props.model.extraCoordConstraint;
-                                // Calculat resulting coords and verify that
+                                // Calculate resulting coords and verify that
                                 // they're valid for this graph
-                                const proposedCoords = _.clone(this._coords());
-                                const oldCoord = _.clone(proposedCoords[i]);
+                                const proposedCoords = deepClone(
```
handeyeco, line 253: Worried about my switch to `deepClone` from `_.clone` because `_.clone` is shallow. However the tests didn't pass when I switched to `{...this._coords()}` and did pass when I used `deepClone`.

packages/perseus/src/widgets/grapher/grapher.tsx
```diff
@@ -346,10 +346,11 @@ class FunctionGrapher extends React.Component<FunctionGrapherProps> {
...
 
-type RenderProps = {
-    availableTypes: PerseusGrapherWidgetOptions["availableTypes"];
-    graph: PerseusGrapherWidgetOptions["graph"];
-    plot?: any;
+type RenderProps = Pick<
+    GrapherPublicWidgetOptions,
+    "availableTypes" | "graph"
+> & {
+    plot?: PerseusGrapherWidgetOptions["correct"];
```
handeyeco, line 353: The types are a little weird here. `correct` gets mapped to `plot` in `staticTransform` but not in `transform`.

packages/perseus/src/widgets/grapher/grapher.tsx
```diff
@@ -551,14 +551,14 @@ class Grapher extends React.Component<Props> implements Widget {
     render(): React.ReactNode {
         const type = this.props.plot.type;
         const coords = this.props.plot.coords;
-        const asymptote = this.props.plot.asymptote;
+        const asymptote = (this.props.plot as any).asymptote;
```
handeyeco, line 554: This is because of some weird union type stuff in TS. Some plot types have `asymptote` and some don't.

Myranae: You could instead do something like the below to keep some measure of typing, but it is more complex than just casting, so maybe it's not as easy to understand.
`const asymptote = "asymptote" in this.props.plot ? this.props.plot.asymptote : undefined;`

packages/perseus/src/widgets/grapher/grapher.tsx
Myranae: Would this benefit from the typing check that has been added to some widgets? When I test adding these in, there's an error with the first one, but I don't think it's related to your work 🤔 Relates to the `markings` prop being added as a top-level prop, but it's not added that way in PerseusGrapherWidgetOptions it seems.
```
0 as any as WidgetProps<PerseusGrapherWidgetOptions> satisfies PropsFor<
    typeof Grapher
>;

0 as any as WidgetProps<GrapherPublicWidgetOptions> satisfies PropsFor<
    typeof Grapher
>;
```

Myranae (approved):
Looks good :) Added some suggestions, but nothing blocking

packages/perseus/src/widgets/grapher/grapher.tsx
handeyeco: https://github.com/Khan/perseus/pull/2465

packages/perseus/src/widgets/grapher/grapher.testdata.ts
```diff
@@ -265,10 +265,10 @@ export const quadraticQuestion: PerseusRenderer = {
 
 export const sinusoidQuestion: PerseusRenderer = {
     content:
-        "###The answer\n\nWe found that the graph of $y=-4\\cos\\left(x\\right)+3$ has a minimum point at $(0,-1)$ and then intersects its midline at $\\left(\\dfrac{1}{2}\\pi,3\\right)$.\n\n[[☃ grapher 3]]\n  ",
+        "###The answer\n\nWe found that the graph of $y=-4\\cos\\left(x\\right)+3$ has a minimum point at $(0,-1)$ and then intersects its midline at $\\left(\\dfrac{1}{2}\\pi,3\\right)$.\n\n[[☃ grapher 1]]\n  ",
     images: {},
     widgets: {
-        "grapher 3": {
+        "grapher 1": {
```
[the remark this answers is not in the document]
handeyeco: I think it's fine to change; I get the impression it's a side-effect of using prod data rather than using real test data.

PR #2459, opened 2025-05-08
Use 'require' instead of 'import' in test script so it will work on more platforms

benchristel:
## Summary:
This test was failing on @handeyeco's machine with the following error:

```
import {parentPort} from "worker_threads";
^^^^^^

SyntaxError: Cannot use import statement outside a module
```

Using `require` is probably more cross-platform compatible.

The cause of the issue may be that some module detection features are not present in NodeJS  versions before 20.19:

> Module syntax detection (the --experimental-detect-module flag) is now enabled by default.
>
> Syntax detection attempts to run ambiguous files as CommonJS, and if the module fails to parse as CommonJS due to ES module syntax, Node.js tries again and runs the file as an ES module. Ambiguous files are those with a .js or no extension, where the nearest parent package.json has no "type" field (either "type": "module" or "type": "commonjs").
>
> —[the NodeJS blog](https://nodejs.org/en/blog/release/v20.19.0#module-syntax-detection-is-now-enabled-by-default)

Issue: none

## Test plan:

CI should pass.

Tests should pass on @handeyeco's dev machine.

### DIFF
--- a/.changeset/perfect-fans-listen.md
+++ b/.changeset/perfect-fans-listen.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-core": patch
+---
+
+Internal: use 'require' instead of 'import' in test script so it will work on more platforms


handeyeco (approved):
Works now, thanks!

PR #2465, opened 2025-05-09
Grapher: add type check, fix possible bug

handeyeco:
## Summary:
I think this might either be an ancient bug or unused code, but tests seem to pass and, with the change, type assertions pass too.

Issue: LEMS-2985

## Test plan:
🤷 Open to suggestions


Myranae (approved):
Looks good! Thanks for fixing this and adding the check :D 

Confirmed no TS errors, no test errors, no lint errors, still renders fine in storybook stories, and still works in storybook EditorPage (markings work just fine there). I suppose we could check rendering of widgets in a ZND, but that doesn't seem necessary for this. :)

packages/perseus/src/widgets/grapher/grapher.tsx
```diff
@@ -513,7 +509,7 @@ class Grapher extends React.Component<Props> implements Widget {
     };
 
     showHairlines: (arg1: Coord) => void = (point) => {
-        if (this.props.apiOptions.isMobile && this.props.markings !== "none") {
```
handeyeco, line 516: This isn't passed through the `transform` functions, so I think it will always be `undefined`.

PR #2466, opened 2025-05-09
Update bump command with new flags

Myranae:
## Summary:
Some recent updates have required us to use new flags in our bump Perseus command. This updates the command provided on each new PR to contain the new flags as shown in the [Shipping a Change to Perseus Confluence page](https://khanacademy.atlassian.net/wiki/spaces/LC/pages/2384887922/Shipping+a+Change+to+Perseus#Making-a-ZND-for-testing-and-feedback).

## Test plan:
- Confirm the command is correct in the next PR to be made
- Confirm all checks pass


.github/workflows/node-ci.yml
```diff
@@ -333,7 +333,7 @@ jobs:
 
                       If you are working in Khan Academy's webapp, you can run:
                       ```sh
-                      ./dev/tools/bump_perseus_version.js -t PR${{ github.event.pull_request.number }}
+                      ./dev/tools/bump_perseus_version.js -s y -t PR${{ github.event.pull_request.number }}
```
handeyeco, line 336: This is probably fine...I kinda didn't want to have this because I wanted users to be deliberate about what they're updating so they had to be reminded that they need to do both a FE and BE release.

handeyeco (approved):

.github/workflows/node-ci.yml
Myranae: Hmm, maybe we can update the description with the options instead of the copy / paste part? Or maybe we can add `-static [y/n]` with a reminder to pick one and don't forget to do FE/BE separately? Included an example for that below. I kind of assumed most bumps to Webapp would be for frontend, but I have a huge bias here since I mainly do frontend work.

```
If you are working in Khan Academy's webapp, you can run the following after updating the static flag to be "y" for frontend or "n" for backend (each requires a separate deploy):
                      ```sh
                      ./dev/tools/bump_perseus_version.js -static [y | n] -t PR${{ github.event.pull_request.number }}`
```

handeyeco: Within a couple of months, this shouldn't really matter (or will need to be changed) because the FE and BE will be in separate repos. I think it's fine to land this either way.

Myranae: Added info for what `-s` is and how to change it if the changes are for backend. Will use this for now and update later based on future changes. :)

PR #2468, opened 2025-05-09
Always displace the first card when shuffling Matcher options

benchristel:
## Summary:
This new approach to removing answer information from
`MatcherPublicWidgetOptions` solves for a few constraints:

1. CP needs the `getPublicWidgetOptions` functions to be
   deterministic — so calling `Math.random()` is not okay.
2. There is an existing product requirement to shuffle the
   cards on the clientside so not all learners see the cards in
   the same order.
3. We also don't want to show learners the correct answer in
   the initial state of the widget, since that could be confusing
   (it's not clear what they need to do to answer the question).
4. We do not want to reveal the correct answer in the public widget
   options, since that allows automated cheating.
5. We need to stay backward-compatible with the existing setup,
   where the client consumes the `PerseusMatcherWidgetOptions`
   (which contain the correct answer).

The solution involves a compromise between constraints 3 and 4.
In order to avoid showing the correct answer when we shuffle
clientside, we need to reveal _some_ information about the answer
to the client. This PR changes `getMatcherPublicWidgetOptions` so
it keeps the first card in each column in place. Then, when we
shuffle the cards on the frontend, we make sure that the card that
was originally first is no longer first after shuffling.

Issue: LEMS-2978

## Test plan:

Run through a matcher exercise several times.
You shouldn't see the cards in the same order every time.
The cards should never be presented in the correct order.

### DIFF
--- a/.changeset/wise-mails-peel.md
+++ b/.changeset/wise-mails-peel.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/perseus-core": minor
+"@khanacademy/perseus": patch
+---
+
+Make `getMatcherPublicWidgetOptions` deterministic, and ensure the correct answer is not shown to the learner in the initial widget state


packages/perseus-core/src/widgets/matcher/matcher-util.test.ts
```diff
@@ -1,51 +1,177 @@
...
+
+function range(min: number, max: number): number[] {
+    invariant(min <= max, "range: min must be <= max");
+    return [
+        ...(function* () {
+            for (let n = min; n <= max; n++) {
+                yield n;
+            }
+        })(),
+    ];
```
handeyeco, line 176: This is pretty clever, but I worry it would be confusing to some.

```suggestion
    const arr = []
        for (let i = min; i <= max; i++) {
            arr.push(i)
        }
    return arr
```

That's shorter and clearer IMO. There's also:

```suggestion
    return new Array(max-min+1)
        .fill()
        .map((_, i) => i + min)
```

But I like my first suggestion the most.

handeyeco (approved):

PR #2473, opened 2025-05-12
Add types and stories to ensure Matcher widgets work with answerless data.

benchristel:
Issue: LEMS-2978

## Test plan:

`pnpm storybook`

You should be able to answer the Matcher question that starts answerless.

### DIFF
--- a/packages/perseus/src/widgets/matcher/matcher.test.tsx
+++ b/packages/perseus/src/widgets/matcher/matcher.test.tsx
@@ -109,13 +109,9 @@ describe("matcher widget", () => {
         ["answerful", answerfulItem],
         ["answerless", answerlessItem],
     ])("given an %s item", (_, {question}) => {
-        it("can be answered correctly", async () => {
+        it("can be answered correctly", () => {
             // Arrange
-            const apiOptions: APIOptions = {
-                isMobile: false,
-            };
-            const {renderer} = renderQuestion(question, apiOptions);
-            await wait();
+            const {renderer} = renderQuestion(question);
 
             // Act
             const matcher: Matcher = renderer.findWidgets("matcher 1")[0];
@@ -140,13 +136,9 @@ describe("matcher widget", () => {
             expect(score).toHaveBeenAnsweredCorrectly();
         });
 
-        it("can be answered incorrectly", async () => {
+        it("can be answered incorrectly", () => {
             // Arrange
-            const apiOptions: APIOptions = {
-                isMobile: false,
-            };
-            const {renderer} = renderQuestion(question, apiOptions);
-            await wait();
+            const {renderer} = renderQuestion(question);
 
             // Act
             const matcher: Matcher = renderer.findWidgets("matcher 1")[0];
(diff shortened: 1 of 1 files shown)


packages/perseus/src/widgets/matcher/matcher.tsx
```diff
@@ -12,37 +12,58 @@ import Renderer from "../../renderer";
...
-    left: Props["left"];
-    right: Props["right"];
-    labels: Props["labels"];
-    orderMatters: Props["orderMatters"];
-    padding: Props["padding"];
-    problemNum: Props["problemNum"];
-    onChange: Props["onChange"];
-    linterContext: Props["linterContext"];
-};
+type RenderProps = Omit<ExternalProps, keyof UniversalWidgetProps>;
```
handeyeco, line 36: I personally find this way of deriving RenderProps from ExternalProps which is derived from the component props to be a little backwards conceptually. To me it adds to the confusion rather than mitigates it, which I feel like is compounded by us doing it for some widgets and not all of them.

I think we should stick to what we've been doing until there's a consensus about what we want and then do a PR that changes all of them at once. Maybe a small, inter-LEMS Perseus ADR? (Outside the normal ADR process?) Even if the goal is just to better explain the why, so that it's not adding to the confusion.

handeyeco (commented):
For most of this work we've been adding unit tests for answerless rendering, interaction, and scoring (through getUserInputMap). I don't think I saw it in the other PR. Do you think it would be worth adding them here?

packages/perseus/src/widgets/matcher/matcher.test.tsx
```diff
@@ -98,88 +102,96 @@ describe("matcher widget", () => {
...
+        ["answerful", answerfulItem],
+        ["answerless", answerlessItem],
+    ])("given an %s item", (_, {question}) => {
+        it("can be answered correctly", async () => {
+            // Arrange
+            const apiOptions: APIOptions = {
+                isMobile: false,
+            };
+            const {renderer} = renderQuestion(question, apiOptions);
+            await wait();
```
handeyeco, line 118: I wonder what this is for. Is it actually needed?

EDIT: I removed them locally and it seemed fine.

packages/perseus/src/widgets/matcher/matcher.test.tsx
```diff
@@ -98,88 +102,96 @@ describe("matcher widget", () => {
...
+    describe.each([
+        ["answerful", answerfulItem],
+        ["answerless", answerlessItem],
+    ])("given an %s item", (_, {question}) => {
+        it("can be answered correctly", async () => {
+            // Arrange
+            const apiOptions: APIOptions = {
+                isMobile: false,
+            };
+            const {renderer} = renderQuestion(question, apiOptions);
```
handeyeco, line 117: I think `isMobile` defaults to false.

```suggestion
            const {renderer} = renderQuestion(question);
```

packages/perseus/src/widgets/matcher/matcher.test.tsx
```diff
@@ -98,88 +102,96 @@ describe("matcher widget", () => {
...
 
-        // Act
-        const matcher: Matcher = renderer.findWidgets("matcher 1")[0];
+        it("can be answered incorrectly", async () => {
+            // Arrange
+            const apiOptions: APIOptions = {
+                isMobile: false,
+            };
+            const {renderer} = renderQuestion(question, apiOptions);
+            await wait();
```
handeyeco, line 149: If possible

```suggestion
            const {renderer} = renderQuestion(question);
```

handeyeco (approved):
Thanks!

PR #2474, opened 2025-05-12
Adds a warning message in the Issues Panel when a content creator uses a widget marked as inaccessible

tatianasnook:
## Summary:
Content creators should be warned when adding widgets marked as inaccessible (accessible: false) so they can make informed choices that support all learners.

We initially implemented checkAccessibilityAndWarn logic directly in editor.tsx, but later moved it into a custom Persues linter rule. This change allows for consistent, centralized warnings in the Issues Panel using the existing linter infrastructure.

Issue: LEMS-2972


.changeset/red-monkeys-ring.md
```diff
@@ -0,0 +1,9 @@
+---
+"@khanacademy/perseus-core": patch
+"@khanacademy/perseus-editor": patch
+---
+
+(LEMS-2972) This PR displays a warning in the Issues Panel when a content creator selects or pastes a widget that is marked as accessible: false.
```
jeremywiebe, line 6: We should not be referencing Jira tickets here as those are internal to Khan Academy. These changeset files become part of the changelog for this package when we ship the change. :) So I try to think how this would read if I were upgrading to this version of `perseus-editor` and wanting to know how it will impact my integration of it (ie. how will a consumer of `pereus-editor` be impacted by this change and what might they need to change?

packages/perseus-core/src/widgets/core-widget-registry.ts
```diff
@@ -71,6 +73,25 @@ export function getDefaultWidgetOptions(type: string) {
     return widgetLogic?.defaultWidgetOptions || {};
 }
 
+export function isAccessible(
```
jeremywiebe, line 76: Just a reminder that we should delete the same-named function in `@khanacademy/perseus` so we don't have two (we discussed this in today's sync, but wanted to note it here also). :)

packages/perseus-editor/src/editor.tsx
```diff
@@ -176,6 +176,7 @@ class Editor extends React.Component<Props, State> {
         textAreaValue: this.props.content,
     };
 
+    _prevWidgetIds: string[] = [];
```
jeremywiebe, line 179: I think this is unused now.

packages/perseus-editor/src/issue-details.tsx
```diff
@@ -25,6 +19,7 @@ const IssueDetails = ({issue}: IssueProps) => {
             animated={true}
             expanded={expanded}
             onToggle={toggleVisibility}
+            containerStyle={{backgroundColor: "#fff9eb"}}
```
jeremywiebe, line 22: Can we use a WonderBlocks color here?

packages/perseus-editor/src/issues-panel.tsx
```diff
@@ -46,17 +51,25 @@ const IssuesPanel = ({warnings = []}: IssuesPanelProps) => {
...
                     />
-                    <span>Issues Panel</span>
+                    <span>Issues</span>
                 </div>
-                {issuesCount}
                 <PhosphorIcon
                     icon={icon}
                     size="medium"
                     color={iconColor}
                     data-icon-type={icon}
```
jeremywiebe, line 69: Why do we need a `data-` attribute here? I don't see this documented in the Wonder-Blocks storybook [site](https://khan.github.io/wonder-blocks/?path=/docs/packages-icon-phosphoricon--docs).

packages/perseus-editor/src/item-editor.tsx
```diff
@@ -50,6 +57,48 @@ class ItemEditor extends React.Component<Props> {
...
+        warnings: [],
+    };
+
+    static getDerivedStateFromProps(props: Props): Partial<State> {
+        const parsed = PerseusMarkdown.parse(props.question?.content ?? "", {});
+        const linterContext = {
+            content: props.question?.content,
+            widgets: props.question?.widgets,
+            stack: [],
+        };
```
jeremywiebe, line 70: We might want to short-circuit this update logic here and return `null` if `content` and `widgets` haven't changed (those are the only things that affect the outcome of this function).  Otherwise we do quite a bit of work and will get the identical result.

packages/perseus-editor/src/item-editor.tsx
```diff
@@ -50,6 +57,48 @@ class ItemEditor extends React.Component<Props> {
...
+                // what we use directly from teh linter and the rest of the
+                // text/links is defined in the `WARNINGS` object.
+                return {
+                    id: linterWarning.rule,
+                    description: linterWarning.message,
+                    message: linterWarning.message,
+                    help: "Don't do that!",
+                    helpUrl: "https://www.example.com",
+                    impact: "HUGE! 🤯",
+                };
```
jeremywiebe, line 97: I just put a placeholder `Issue` here... we should clean this up before landing this PR.  This is the `Issue` we build for all other linter issues that we don't handle explicitly above.

packages/perseus-core/src/widgets/core-widget-registry.ts
```diff
@@ -71,6 +73,25 @@ export function getDefaultWidgetOptions(type: string) {
...
+    const {accessible} = widgetLogic;
+
+    if (accessible == null) {
+        return false;
+    }
+
+    return typeof accessible === "function"
+        ? accessible(widgetOptions)
+        : !!accessible;
+}
```
mark-fitzgerald, line 94: While I like the explicit logic here, I'm wondering if using optional chaining would be a valuable alternative:
```
const widgetAccessibility = widgets[type]?.accessible;
return typeof widgetAccessibility === "function"
    ? widgetAccessibility(widgetOptions)
    : !! widgetAccessibility;
```

packages/perseus-editor/src/item-editor.tsx
```diff
@@ -35,7 +38,11 @@ type Props = {
     itemId?: string;
 };
 
-class ItemEditor extends React.Component<Props> {
+type State = {
+    warnings: Issue[];
```
mark-fitzgerald, line 42: Should this be called `issues` instead of `warnings`?

.changeset/red-monkeys-ring.md
catandthemachines: I still see the reference to the Jira ticket. Friendly ping to make sure this note is removed <3

packages/perseus-core/src/widgets/core-widget-registry.ts
```diff
@@ -71,6 +73,25 @@ export function getDefaultWidgetOptions(type: string) {
...
+    widgetOptions: PerseusWidgetOptions,
+): boolean {
+    const widgetLogic = widgets[type];
+
+    if (!widgetLogic) {
+        return false;
+    }
+    const {accessible} = widgetLogic;
+
+    if (accessible == null) {
```
catandthemachines, line 87: I think this extra check can be removed, we added it during pairing time. But I don't think it's needed since the root of the original issue was the surfacing of `accessible` in `WidgetLogic`

packages/perseus-core/src/index.ts
```diff
@@ -25,6 +25,7 @@ export {
     getWidgetIdsFromContent,
     getWidgetIdsFromContentByType,
 } from "./utils/widget-id-utils";
+
```
catandthemachines, line 28: Nit: no need for this extra line.

packages/perseus-core/src/widgets/matcher/index.ts
catandthemachines, line 1: I'd remove this file's changes. This will happen in your other PR: https://github.com/Khan/perseus/pull/2485

packages/perseus-linter/src/proptypes.ts
```diff
@@ -1,19 +1,8 @@
-// Define the shape of the linter context object that is passed through the
-// tree with additional information about what we are checking.
-import PropTypes from "prop-types";
-
 import type {LinterContextProps} from "./types";
 
-export const linterContextProps = PropTypes.shape({
```
catandthemachines, line 7: Why remove this variable? 🤔

packages/perseus-editor/src/item-editor.tsx
catandthemachines: Agreed. I would add a generic message to message `message.ts` that would be a good stand-in. If you post in here or in the `#project-evolve-ich` what this message looks like or do a quick pair session I think we can find a good structure for these warning messages.

packages/perseus-editor/src/item-editor.tsx
```diff
@@ -98,7 +147,7 @@ class ItemEditor extends React.Component<Props> {
             <div className="perseus-editor-table">
                 <div className="perseus-editor-row perseus-question-container">
                     <div className="perseus-editor-left-cell">
-                        <IssuesPanel warnings={[]} />
+                        <IssuesPanel warnings={this.state.warnings} />
```
catandthemachines, line 150: Thought here. We probably want to send all the issues discovered by the linter through and sort in the `IssuesPanel` by what the severity is and apply the specific styling by the severity. Calling this out here as this functionality would need to be overhauled in it's current state. I think we can build this functionality pretty quick and light without much more effort but will save us time in the future.

When you have a moment let's pair on this!

packages/perseus-editor/src/issues-panel.tsx
tatianasnook: Thanks for the question! I added the data-icon-type attribute specifically to support testing (perseus-editor/src/_test_/issues-panel.test.tsx). Since <PhosphorIcon> renders as a <span> without any built-in test-friendly attributes (like data-testid or role), I needed a reliable way to differentiate icons in the DOM during tests.
Using data-icon-type allowed me to consistently query for specific icons (like "check-circle-fill.svg" or "warning-fill.svg") without relying on brittle DOM traversal or assumptions about markup structure.

tatianasnook: Used testId as the appropriate attribute for PhosphorIcon, based on the Wonder Blocks storybook documentation. Also refactored issues-panel.test.tsx to use these testIds for more reliable and readable testing.

packages/perseus-editor/src/issue-details.tsx
tatianasnook: Thank you for the suggestion! I’ve updated the background color to use var(--wb-semanticColor-status-warning-background) from Wonder Blocks, which corresponds to the original #fff9eb.

packages/perseus-editor/src/item-editor.tsx
tatianasnook: Thanks for the suggestion! I added a short-circuit at the top of getDerivedStateFromProps to check if props.question.content and props.question.widgets have changed. If not, we return null early to avoid unnecessary parsing and linting work.

packages/perseus-editor/src/issue-details.tsx
tatianasnook: Thanks for the suggestion! I’ve updated the background color to use color.fadedGold8 from Wonder Blocks, which corresponds to the original #fff9eb.

packages/perseus-core/src/widgets/core-widget-registry.ts
tatianasnook: Thanks for the note! You're right — that check was added during pairing but no longer seems necessary. I recently refactored this function based on Mark’s suggestion.

packages/perseus-editor/src/item-editor.tsx
tatianasnook: Thanks, Mark — that makes sense. Since these are rendered in the IssuesPanel and not all of them are strictly warnings, renaming warnings to issues makes things clearer and more consistent. I’ll go ahead and update that.

packages/perseus-editor/src/item-editor.tsx
catandthemachines: Per standup conversation let's worry about all the nitty gritty for the different linter severity in this follow-up task. But let's change the property name to `issues` instead of `warnings`.

https://khanacademy.atlassian.net/browse/LEMS-3109

packages/perseus-editor/src/item-editor.tsx
tatianasnook: Moved the linting warning message to messages.ts as suggested.

.changeset/red-monkeys-ring.md
tatianasnook: removed

packages/perseus-editor/src/messages.ts
```diff
@@ -0,0 +1,24 @@
...
+            "Selecting inaccessible widgets for a practice item will result in this exercise being hidden from users with 'Hide visually dependant content' setting set to true. Please select another widget or create an alternative practice item.",
+    }),
+
+    genericLinterWarning: (rule: string, message: string) => ({
+        id: rule,
+        description: message,
+        help: "Learn more about best practices for authoring items",
+        helpUrl:
+            "https://docs.google.com/document/d/1N13f4sY-7EXWDwQ04ivA9vJBVvPPd60qjBT73B4NHuM/edit?tab=t.0",
+        impact: "medium",
```
catandthemachines, line 21: nit: let's set this to low for the time being to differentiate it from the inaccessibleWidget message. We'll fix it in the future but it'll make it better in the short term.

packages/perseus-core/src/widgets/core-widget-registry.ts
```diff
@@ -31,20 +31,22 @@ import sorterWidgetLogic from "./sorter";
...
-const widgets = {};
+const widgets: Record<string, WidgetLogic> = {};
 
 function registerWidget(type: string, logic: WidgetLogic) {
     widgets[type] = logic;
 }
 
 export function isWidgetRegistered(type: string) {
     const widgetLogic = widgets[type];
+    // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
```
catandthemachines, line 49: nit: is this still needed?

packages/perseus-editor/src/__tests__/issues-panel.test.tsx
```diff
@@ -22,92 +23,63 @@ describe("IssuesPanel", () => {
...
-            // But <PhosphorIcon> renders as a <span> with no test-friendly props (e.g., no data-testid or role),
-            // and attempts like getByTestId, getByText().nextSibling, or adding a test ID weren’t reliable.
-            // So to verify the icon (e.g., data-icon-type="check-circle-fill.svg"), direct access was the most
-            // consistent option.
-            // eslint-disable-next-line testing-library/no-node-access
-            .parentElement?.querySelector(
-                '[data-icon-type="check-circle-fill.svg"]',
-            );
+        // Assert that the icon is passing icon
+        const icon = screen.getByTestId("issues-icon-check-circle-fill.svg");
```
catandthemachines, line 32: Nice! I love a cleaner test!

packages/perseus-editor/src/item-editor.tsx
```diff
@@ -45,11 +52,56 @@ class ItemEditor extends React.Component<Props> {
         question: {},
         answerArea: {},
     };
+    static prevContent: string | undefined;
+    static prevWidgets: any;
```
catandthemachines, line 56: I believe that `any` can be set to the type `PerseusWidgetsMap`. Can you try to add that? That will help us keep the code healthy to use a specific type ❤️

catandthemachines (approved):
Everything is looking nice and clean! I added a few comments for things to fix, once those are fixed with the linting errors are resolved you should be good to complete.

packages/perseus-editor/src/item-editor.tsx
jeremywiebe, line 56: Storing the previous rendered data here will be an issue. Because it's `static`, it will be shared across all instances of the `ItemEditor`. Granted, we probably never have more than one in a page, but I don't want to risk that. 

I think a short solution would be to simply store these two items in `state`. [`getDerivedStateFromProps`](https://react.dev/reference/react/Component#static-getderivedstatefromprops) can have a second parameter which is the current state so you'd be able to compare it against the incoming props then.

Ultimately, I'm feeling we should probably move this component to be a functional component soon as we'd be able to solve this better using the `React.useMemo()` hook on the `content` and `widgets` values. 

/cc @catandthemachines this is a larger post, but would you be in favour of scheduling the port of `ItemEditor` to be a functional component as part of Tatiana's work here?

jeremywiebe: And after saying all that, I think we could land this PR as it is if we plan to port this to a functional component in the very near future. :)

jeremywiebe (approved):
Nice work Tatiana. I have a long-ish comment about the use of `static` to hold the previous prop values. But I think we can address that by migrating the `ItemEditor` component to be a functional component.  So approving now with the understanding we'll port it as part of this project.

packages/perseus-core/src/widgets/core-widget-registry.ts
tatianasnook: refactor: remove eslint-disable by using Boolean() instead of !!

packages/perseus-editor/src/item-editor.tsx
tatianasnook: Updated prevWidgets to use PerseusWidgetsMap | undefined instead of any as suggested. Thanks for the catch!

packages/perseus-editor/src/__tests__/issues-panel.test.tsx
```diff
@@ -22,92 +23,63 @@ describe("IssuesPanel", () => {
...
         expect(icon).toBeInTheDocument();
```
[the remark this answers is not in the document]
tatianasnook: Updated to use toBeVisible() as suggested — confirms both presence and visibility.

PR #2477, opened 2025-05-12
Use ContentPreview for storybook EditorPage preview

jeremywiebe:
## Summary:

I couldn't figure out why lint violations weren't visible in our Editor Page preview (the one specifically for Storybook). I finally realized we weren't using the new, mostly unreleased `ContentPreview` component and just using the `Renderer` directly. This PR switches to use the `ContentPreview` component so we get a feww more bells and whistles in our Storybook EditorPage preview.

Eventually `ContentPreview` will become _the_ way to preview content. We can't switch _quite_ yet because current Perseus styling uses `@media` queries based on `screen-width`.  That requires we use an iframe to get accurate previews. Once we can use `container` queries, we can switch our styling to that and start using this _fast_ `ContentPreview` component everywhere. 

Issue: "none"

## Test plan:

`yarn start` -> go to http://localhost:6006/?path=/story/perseuseditor-editorpage--demo and play around with some content. In the screenshot, you can see examples of several different types of linter violations being highlighted.

image


.changeset/wicked-clocks-march.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
handeyeco, line 2: I would call this at least a patch, since it will have a user-facing change in [one of our targets](https://khan.github.io/perseus/). Not a blocker though.

handeyeco (approved):

catandthemachines (approved):
Oh my gosh thank you so much Jeremy! Excited to see this fixed in storybook 🥳

.changeset/wicked-clocks-march.md
jeremywiebe: @handeyeco I don't see any user-facing changes in production code (except adding comments to props - which I guess isn't user-facing).   The link you shared only goes to our Github-hosted storybook.  I feel I'm missing your point.

handeyeco: My point is just that Storybook _is_ one of our targets; it's a UI that people see. I think empty changesets should be used very rarely and for changes that are completely invisible to all UIs. Just a personal opinion.

jeremywiebe: I think we have differing views of what Changesets should cover. I see changesets as a way to communicate with folks who integrate Perseus into a project. Storybook is an internal tool and I see very little value in writing changesets to document UI changes to storybook (who would read them and what value would they provide over the PR and it's description?).

PR #2481, opened 2025-05-14
Add Horizontal Scrollbar for New Radio Widget to allow users to visually scroll a long radio option content

ivyolamit:
## Summary:
This PR is the start of implementing the Horizontal Scrollbar for the New Radio Widget that will allow users to visually scroll a long radio option content.

**Details**
- Add ScrollableView component, that can be reusable for other widget
- Add Horizontal Scrollbar for New Radio Widget to allow users to visually scroll a long radio option content
- Next two PRs will be for additional styling and accessibility/tab navigation

Issue: LEMS-3105

https://github.com/user-attachments/assets/e53f6014-22b9-4d37-8164-89f771d8bad4

## Test plan:
**General Test**
- Scrollbar visibility will respect user's system settings of 'Show scroll bars'
- When scrollbar is visible, show it at the maximally scrolled position
- Disable the button that controls the scroll direction when the widget cannot be scrolled in that direction any more
- This pattern should be implemented with the intent to extend it to other widgets.
- The choice icon should always be visible (regardless of scroll position) so that a learner can select their answer at any time.

**Storybook**
1. Navigate to RadioNew widget > Multi Select with Scroll
- [ ] Confirm that the Scrollbar buttons are visible
- [ ] Confirm that the left arrow button are initially disabled
- [ ] Confirm that the right arrow button are initially enabled (or whenever there's a content to scroll)
- [ ] Confirm the scrollbars are visible and respect the user's _Show scroll bars settings_
2. Click the right arrow button
- [ ] Confirm that the content are scrolled to the right
- [ ] Confirm that the left arrow button are now enabled
- [ ] Confirm that the choices are sticky while content scroll beneath it
- [ ] Confirm that the right arrow button will be disabled once no additional content to scroll to
3. Do a normal scroll above the radio widget
- [ ] Confirm that the scroll left and right works
- [ ] Confirm that the behavior of the left and right arrow buttons still works
4. Navigate to your laptop's settings Appearance > Show scroll bars
5. Select "Always"
6. Navigate back to the RadioNew widget
- [ ] Confirm that the scrollbars are always visible and respect the user settings

ZND Link: https://prod-znd-250513-ivy-01.khanacademy.org/math/early-math/cc-early-math-add-sub-basics/cc-early-math-add-sub-word-problem-within-10/e/addition-word-problems-within-10
ℹ️ Your KAID needs to be in the feature flag


testing/split-view.tsx
```diff
@@ -44,7 +44,6 @@ const SplitView = ({
 const styles = {
     sideBySide: {
         display: "flex",
-        flexWrap: "wrap",
```
SonicScrewdriver, line 47: This seems like a global change to our storybook testing component. I assume this was necessary to test the scroling?

packages/perseus/src/components/scrollable-view.tsx
```diff
@@ -0,0 +1,175 @@
...
+import IconButton from "@khanacademy/wonder-blocks-icon-button";
+import {spacing} from "@khanacademy/wonder-blocks-tokens";
+import {LabelSmall} from "@khanacademy/wonder-blocks-typography";
+import caretLeftIcon from "@phosphor-icons/core/regular/caret-left.svg";
+import caretRightIcon from "@phosphor-icons/core/regular/caret-right.svg";
+import {StyleSheet} from "aphrodite";
+import * as React from "react";
+import {useEffect, useRef} from "react";
+
+import type {CSSProperties, StyleDeclaration} from "aphrodite";
```
anakaren-rojas, line 11: We're moving away from using aphrodite I thought

packages/perseus/src/widgets/radio/base-radio.new.tsx
```diff
@@ -248,145 +249,155 @@ const BaseRadio = function ({
...
-                            <Element {...elementProps} ref={ref} />
-                        </li>
-                    );
-                })}
-            </ul>
+
+                        // HACK(abdulrahman): Preloads the selection-state
+                        // css because of a bug that causes iOS to lag
+                        // when selecting the button for the first time.
+                        aphroditeClassName(true);
```
SonicScrewdriver, line 325: I wonder if this is even necessary anymore. No need to change this as part of your PR though, just musing. :)

packages/perseus/src/components/scrollable-view.tsx
```diff
@@ -0,0 +1,175 @@
...
+        setIsScrollable(scrollWidth > clientWidth + 1); // 1px tolerance
+        setCanScrollLeft(scrollLeft > 0);
+        setCanScrollRight(scrollLeft + clientWidth < scrollWidth - 1);
+    };
+    const scroll = (direction: "left" | "right") => {
+        if (!containerRef.current) {
+            return;
+        }
+
+        const scrollAmount = 100; // Adjust this value as needed
```
anakaren-rojas, line 69: when would this need to be adjusted? would it be adjusted in the browser or manually by one of us?

packages/perseus/src/components/scrollable-view.tsx
```diff
@@ -0,0 +1,175 @@
...
+}: ScrollButtonsProps) {
+    return (
+        <View style={styles.scrollButtonsContainer}>
+            <IconButton
+                icon={caretLeftIcon}
+                actionType="neutral"
+                kind="secondary"
+                size="small"
+                onClick={onScrollLeft}
+                aria-label="scroll left"
```
anakaren-rojas, line 137: The aria labels will need to be translated, to start, these will need to be added `packages/perseus/src/strings.ts` file. In this file, you'll import the `strings` and call the respective label here. So it would look something like 
`strings.ScrollableAriaLabel` 
Here's an example: 
https://github.com/Khan/perseus/blob/d64277bc85b0b533ef85eec2e07d24a6c407f089/packages/perseus/src/widgets/interactive-graphs/graphs/angle.tsx#L192
https://github.com/Khan/perseus/blob/d64277bc85b0b533ef85eec2e07d24a6c407f089/packages/perseus/src/widgets/interactive-graphs/graphs/angle.tsx#L201

testing/split-view.tsx
ivyolamit: @SonicScrewdriver yes it's for the scrolling. I did test it with other widgets, didn't see any difference. But let me know if you see anything 🙏🏼

packages/perseus/src/components/scrollable-view.tsx
SonicScrewdriver: It might be good to pull this out above so that it's easier to find

packages/perseus/src/components/scrollable-view.tsx
```diff
@@ -0,0 +1,175 @@
...
+    },
+
+    scrollFadeRight: {
+        position: "absolute",
+        top: 0,
+        right: 0,
+        width: "max-content",
+        height: "100%",
+        backgroundColor: "linear-gradient( to left, white transparent)",
+        pointerEvents: "none",
```
SonicScrewdriver, line 169: When I test this (locally and on znds) I don't ever see the scrollFadeRight style getting applied to the div above.

One thing I'm noticing is that I think you might want `background` instead of `backgroundColor`. ([See linear-gradient()](https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/linear-gradient))

SonicScrewdriver (commented):
Great work!! SO happy to see this issue getting solved so nicely.

I noticed a couple minor issues while testing, however, that I wanted to check in with you about before approving:  

Screenshot 2025-05-14 at 9 12 00 AM
One thing I'm noticing is that I'm frequently seeing the scrollbars even when they're not necessary. 


https://github.com/user-attachments/assets/283d471b-a86f-464e-9389-ecc6e421efe3
I'm also noticing that sometimes the buttons aren't appearing on page load even though they should be there. They'll only appear after manually scrolling.

packages/perseus/src/components/scrollable-view.tsx
anakaren-rojas: also other files use the `usePerseusI18n()` function

> `    const {strings} = usePerseusI18n();`

packages/perseus/src/components/scrollable-view.tsx
ivyolamit: @anakaren-rojas the `scrollAmount` is the number of pixels to move the scroll position when the user clicks the scroll button (left or right). So if we want to shorten or increase the scroll position we would update this value.

I'll update the code
- better naming
- better comment
- move where it's defined as Third suggested

packages/perseus/src/components/scrollable-view.tsx
ivyolamit: @anakaren-rojas are you okay if I just update this file for now, and in my follow-up PR? i'll update the other files under radio removing the usage of aphrodite. Some of it is not straight forward change for the usage of `css` and classNames.

image

packages/perseus/src/components/scrollable-view.tsx
ivyolamit: That's a great catch @SonicScrewdriver! I'm missing the comma 🤦🏻‍♀️
i'll update that

packages/perseus/src/components/scrollable-view.tsx
```diff
@@ -0,0 +1,179 @@
...
+const SCROLL_DISTANCE = 100;
+
+function ScrollableView({
+    x,
+    y,
+    children,
+    showScrollButtons = false,
+    style,
+    role = "group",
+    ...rest
```
mark-fitzgerald, line 50: Can we name this something more obvious, like `additionalProps`, or `containerProps`, or something similar?

SonicScrewdriver (approved):
THANK YOU!!!!

packages/perseus/src/components/scrollable-view.tsx
```diff
@@ -0,0 +1,179 @@
...
+    React.HTMLAttributes<HTMLDivElement> &
+    ScrollableViewPropsBase;
+
+// This is the number of pixels to scroll when the left or right scroll buttons
+// are clicked. Adjust this value to change the scroll amount as needed.
+const SCROLL_DISTANCE = 100;
+
+function ScrollableView({
+    x,
+    y,
```
mark-fitzgerald, line 45: These `x` and `y` props feel non-descriptive to me. Can we name them something more meaningful? When I see them, I am thinking of positioning values (maybe because I spent time on the Interactive Graph project). Perhaps just sticking with `overflowX` and `overflowY`?

ivyolamit:
@anakaren-rojas @mark-fitzgerald i've update the PR to address your comments, please let me know if you have other feedback or questions 🙏🏼

packages/perseus/src/strings.ts
```diff
@@ -653,6 +656,9 @@ export const strings = {
             "This is a list of single-character labels that will appear in front of multiple-choice options. For instance, a multiple-choice question with three options would display (A) first option (B) second option (C) third option. There must be spaces between each of the different characters. The characters will show up next to options in the order that they are listed here. Most multiple choice questions have 5 or fewer options.",
         message: "A B C D E F G H I J K L M N O P Q R S T U V W X Y Z",
     },
+    scrollAnswers: "Scroll Answers",
+    scrollLeftArrow: "Scroll left arrow",
+    scrollRightArrow: "Scroll right arrow",
```
mark-fitzgerald, line 661: Since these are used for aria, is the word "arrow" needed?

packages/perseus/src/widgets/radio/base-radio.new.tsx
```diff
@@ -248,145 +249,155 @@ const BaseRadio = function ({
...
-                                // If the clicked element is inside of the
-                                // radio icon, then we want to trigger the
-                                // check by flipping the choice of the icon.
-                                if (elem.getAttribute("data-is-radio-icon")) {
-                                    updateChoice(i, {
-                                        checked: !choice.checked,
-                                        crossedOut: choice.crossedOut,
-                                    });
+            <ScrollableView
+                aria-label={"radio widget scrollable view"}
```
mark-fitzgerald, line 253: Is this needed? This seems too generic.

packages/perseus/src/components/scrollable-view.tsx
```diff
@@ -0,0 +1,179 @@
...
+    children,
+    showScrollButtons = false,
```
mark-fitzgerald, line 47: What is the use case for when the buttons shouldn't be shown?

packages/perseus/src/widgets/radio/choice.new.tsx
```diff
@@ -206,30 +194,48 @@ const Choice = function (props: ChoicePropsWithForwardRef): React.ReactElement {
...
+                                style={{
+                                    flex: "0 0 30px",
+                                    position: "sticky",
+                                    left: 0,
+                                    background:
+                                        "linear-gradient(to right, white 70%, rba(255,255,255, 0.8) 100%)",
+                                    zIndex: 1,
+                                    alignItems: "center",
+                                    justifyContent: "center",
+                                }}
```
mark-fitzgerald, line 216: Any reason why this can't be defined in the `styles` definition below?

packages/perseus/src/widgets/radio/choice.new.tsx
```diff
@@ -206,30 +194,48 @@ const Choice = function (props: ChoicePropsWithForwardRef): React.ReactElement {
...
-                                paddingTop: spacing.xSmall_8,
-                                paddingBottom: spacing.xSmall_8,
-                                paddingLeft: spacing.xSmall_8,
+                                paddingTop: sizing.size_080,
+                                paddingBottom: sizing.size_080,
+                                paddingLeft: sizing.size_080,
+                                alignItems: "flex-start",
+                                position: "relative",
+                                background: color.white64,
+                                minWidth: "100%",
```
mark-fitzgerald, line 203: I know this styling was already here prior to this PR, but I think moving it to the `styles` definition below would be helpful.

packages/perseus/src/widgets/radio/choice.new.tsx
```diff
@@ -206,30 +194,48 @@ const Choice = function (props: ChoicePropsWithForwardRef): React.ReactElement {
...
+                            </div>
                             <span
                                 style={{
-                                    paddingLeft: spacing.small_12,
+                                    paddingLeft: sizing.size_120,
                                     textAlign: "left",
                                     flex: 1,
-                                    paddingTop: 4,
+                                    paddingTop: sizing.size_040,
+                                    minWidth: 0,
```
mark-fitzgerald, line 238: Same here regarding where the styling is defined.

packages/perseus/src/components/scrollable-view.tsx
ivyolamit: Since this is a generic component, I added this option for widgets that might not need the buttons itself, but would still make the widget scrollable.

packages/perseus/src/widgets/radio/choice-icon/shared-styles.ts
```diff
@@ -8,6 +9,7 @@ const styles: StyleDeclaration = StyleSheet.create({
     iconWrapper: {
         display: "inline-block",
         position: "relative",
+        backgroundColor: color.white,
```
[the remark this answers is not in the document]
ivyolamit: This will solve make the content scrolled look visually behind the choice icon

packages/perseus/src/widgets/radio/choice.new.tsx
ivyolamit: @mark-fitzgerald are you okay if I address this in my follow-up PR? it will be together when addressing not using aphrodite. I already had the changes stashed when I was trying to address this comment [here](https://github.com/Khan/perseus/pull/2481#discussion_r2089920301)

packages/perseus/src/components/scrollable-view.tsx
mark-fitzgerald: I appreciate the effort to build for the future. Since this is something that is easy enough to add if we need it, and since having the buttons seem like an accessibility feature we would always want to keep, I recommend removing it for now.

anakaren-rojas (approved):
🎉

packages/perseus/src/widgets/radio/choice-icon/shared-styles.ts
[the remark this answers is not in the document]
mark-fitzgerald: I see. Since we can't guarantee that the background will always be white (think dark mode, or even other themes), I think we should take a different approach. How about making the content narrower, so that the left edge is just to the edge of the radio/checkbox buttons. There may need to be some margin involved to make the space where the buttons would go. Then, place the buttons in that space.

packages/perseus/src/widgets/radio/choice.new.tsx
mark-fitzgerald: My work to implement CSS Modules will handle the conversion of Aphrodite to regular CSS. If you place all the styling into the `styles` definition, it will make converting to CSS Modules much easier.

ivyolamit: ah ok will do that. Does it mean i can hold off my PR for now and wait for your changes on removing aphrodite as mentioned above? Probably makes it easier.

mark-fitzgerald: You probably won't need to do a follow-up PR to remove the Aphrodite styling. The other tickets in the queue will handle that. I'll let you know if that changes.

packages/perseus/src/widgets/radio/choice-icon/shared-styles.ts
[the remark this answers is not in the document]
ivyolamit: Would you be ok if we address this in a separate ticket? Currently looks like our widget doesn't support dark mode yet. I do have a solution in mind on this already. This will also involve the fade on the right side of the widget which is currently set to white background.

packages/perseus/src/widgets/radio/base-radio.new.tsx
ivyolamit: yes we would need it based on this

> A label that concisely describes the contents placed within the non-interactive element. The label can either be provided by:
> 1 An id/aria-labelledby association if it has a visible heading, or
> 2 An aria-label, if not.

https://www.a11yproject.com/posts/how-to-use-the-tabindex-attribute/#scrollable-overflow-containers

packages/perseus/src/widgets/radio/choice.new.tsx
ivyolamit: i ended up removing the Aphrodite in this file, because adding to the existing styles definition doesn't seem to work or apply the styling properly.

packages/perseus/src/widgets/radio/choice-icon/shared-styles.ts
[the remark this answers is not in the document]
ivyolamit: Please view details here:
https://khanacademy.atlassian.net/browse/LEMS-3108

ivyolamit:
@mark-fitzgerald i've updated the PR to address your additional comments, please let me know if you have other concerns or feedback 🙏🏼 And i'm happy to discuss regarding the white background color.

packages/perseus/src/widgets/radio/choice-icon/shared-styles.ts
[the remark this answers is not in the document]
ivyolamit: I also removed my changes in the shared-styles and instead isolated my styling changes to be applied only in new radio widget

packages/perseus/src/widgets/radio/choice-icon/choice-icon.new.tsx
```diff
@@ -12,8 +12,8 @@ import * as styleConstants from "../../../styles/constants";
 import FocusRing from "../focus-ring";
 import {getChoiceLetter} from "../util";
 
-import CrossOutLine from "./cross-out-line";
-import sharedStyles, {CHOICE_ICON_SIZE} from "./shared-styles";
+import choiceIconStyles, {CHOICE_ICON_SIZE} from "./choice-icon-styles";
+import CrossOutLine from "./cross-out-line.new";
```
mark-fitzgerald, line 16: We aren't supporting the cross-out in the revitalized version of this widget. Can we remove this?

packages/perseus/src/widgets/radio/choice.new.tsx
```diff
@@ -342,14 +302,49 @@ const Choice = function (props: ChoicePropsWithForwardRef): React.ReactElement {
...
+    choiceRow: {
+        display: "flex",
+        flexDirection: "row",
+        justifyContent: "flex-start",
+        alignItems: "center",
+        paddingTop: spacing.xSmall_8,
+        paddingBottom: spacing.xSmall_8,
+        paddingLeft: spacing.xSmall_8,
+        position: "relative",
+        background: color.white64,
```
mark-fitzgerald, line 322: Does this impede dark mode?

packages/perseus/src/components/scrollable-view.tsx
```diff
@@ -0,0 +1,182 @@
...
+            <IconButton
+                icon={caretRightIcon}
+                actionType="neutral"
+                kind="secondary"
+                size="small"
+                onClick={onScrollRight}
+                aria-label={strings.scrollRight}
+                disabled={!canScrollRight}
+            />
+            <LabelSmall>{strings.scrollAnswers}</LabelSmall>
```
mark-fitzgerald, line 156: I'm wondering if we should make this configurable. Right now "scroll answers" works for our purposes with the radio widget. Maybe allow for a component prop that sets this text, but defaults to "scroll answers"?

ivyolamit: That's a great idea! will update that

packages/perseus/src/widgets/radio/choice.new.tsx
ivyolamit: Yes it can, but this will be local to the new radio widget for now. Dark mode will be solve in this ticket: https://khanacademy.atlassian.net/browse/LEMS-3108

ivyolamit:
@mark-fitzgerald thanks for your patience here, have updated the PR to address your comments. When you have time please have another look 🙏🏼 or let me know for other concerns 🙏🏼

mark-fitzgerald (approved):
This is so helpful for users! Thank you!!

PR #2484, opened 2025-05-14
Ensure Sorter works with answerless data

benchristel:
This PR also changes how we remove answer information from Sorter.  The cards
are sorted instead of being shuffled. This was done for two reasons:

- the Content Platform team needs `splitPerseusItem` to be deterministic
- a deterministic shuffling algorithm would leak information about the initial
  order of the cards, which is unsuitable for cheating prevention.

When sorting the cards, we leave the first one in place. This allows us to
avoid showing the correct answer to learners when we shuffle the cards on the
frontend (at the cost of leaking some information about the correct answer).

See https://github.com/Khan/perseus/pull/2468 for more details about this approach.

Issue: LEMS-2984

## Test plan:

`pnpm storybook`

View the stories for Sorter and confirm that they are interactive and scored
correctly.

### DIFF
--- a/.changeset/four-rockets-hug.md
+++ b/.changeset/four-rockets-hug.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/perseus": patch
+"@khanacademy/perseus-core": patch
+---
+
+Remove answer information from the public widget options of Sorter by sorting the cards in `getSorterPublicWidgetOptions`.


packages/perseus-core/src/widgets/sorter/sorter-util.ts
```diff
@@ -20,23 +21,37 @@ type SorterPublicWidgetOptions = {
...
+    return [first, ...rest.sort()];
+}
+
+function shuffleDisplacingFirst<T>(array: readonly T[], rng: RNG): T[] {
+    function isFirstElementDisplaced(shuffled: readonly T[]) {
+        return shuffled[0] !== array[0];
+    }
+
+    return constrainedShuffle(array, rng, isFirstElementDisplaced);
+}
```
handeyeco, line 55: This looks like the same logic as Matcher. Could the two share logic?

packages/perseus-core/src/widgets/sorter/sorter-util.ts
```diff
@@ -20,23 +21,37 @@ type SorterPublicWidgetOptions = {
...
+export function shuffleSorter(props: {
+    correct: string[];
+    problemNum: number | null | undefined;
+}): string[] {
+    const {correct, problemNum} = props;
+    const rng = seededRNG(problemNum ?? 0);
+    // See getSorterPublicWidgetOptions for an explanation of why we need to
+    // displace the first card.
+    return shuffleDisplacingFirst(correct, rng);
+}
```
handeyeco, line 43: The reason this needs to be different than the Matcher sorting logic is because Matcher we sort two things?

packages/perseus-core/src/widgets/sorter/sorter-util.ts
```diff
@@ -20,23 +21,37 @@ type SorterPublicWidgetOptions = {
...
+        correct: sortAllButFirst(options.correct),
     };
 }
 
+export function shuffleSorter(props: {
+    correct: string[];
+    problemNum: number | null | undefined;
+}): string[] {
+    const {correct, problemNum} = props;
+    const rng = seededRNG(problemNum ?? 0);
```
handeyeco, line 39: I'm a little worried about depending on `problemNum`...actually on second thought it doesn't matter here. We're taking something that's obfuscated and making it more obfuscated.

handeyeco (approved):
Thanks!

packages/perseus-core/src/widgets/sorter/sorter-util.test.ts
```diff
@@ -15,9 +21,43 @@ describe("getSorterPublicWidgetOptions", () => {
...
 });
+
+function range(min: number, max: number): number[] {
+    invariant(min <= max, "range: min must be <= max");
+    const ret: number[] = [];
+    for (let i = min; i <= max; i++) {
+        ret.push(i);
+    }
+    return ret;
+}
```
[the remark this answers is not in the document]
handeyeco: `perseus-core/utils/range.ts`? `perseus-core/utils/test-utils.ts`?

PR #2485, opened 2025-05-14
Migrate accessibility logic from perseus to perseus-core

tatianasnook:
## Summary:
- Removed the accessibility field from widget definitions in perseus.
- Added an accessible flag(boolean or function) to each widget's widgetLogic in perseus-core.
- Added traverseChildWidgets flag to appropriate widgetLogic entries in perseus-core.
- Moved a11y.ts and traversal.ts from perseus to perseus-core.
- Moved the traverseChildWidgets function from perseus/src/widgets to perseus-core/src/widgets/core-widget-registry.

Issue: LEMS-3115

## Test plan:


.changeset/red-monkeys-ring.md
catandthemachines, line 1: Please remove this change set 😉

packages/perseus-core/src/widgets/group/index.ts
```diff
@@ -12,9 +12,15 @@ const defaultWidgetOptions: GroupDefaultWidgetOptions = {
     images: {},
 };
 
+const traverseChildWidgets = function (props: any, traverseRenderer: any): any {
```
catandthemachines, line 15: Why do some widgets need this `traverseChildWidgets` while other files do not need it?

packages/perseus-core/src/widgets/interactive-graph/index.ts
```diff
@@ -42,6 +42,7 @@ const interactiveGraphWidgetLogic: WidgetLogic = {
     name: "interactive-graph",
     defaultWidgetOptions,
     getPublicWidgetOptions: getInteractiveGraphPublicWidgetOptions,
+    accessible: true,
```
catandthemachines, line 45: Nit: could you add a TODO comment above that links to the following task:
https://khanacademy.atlassian.net/browse/LEMS-3102

You can see an example of that here: https://github.com/Khan/perseus/blob/d64277bc85b0b533ef85eec2e07d24a6c407f089/packages/perseus/src/widgets/expression/expression.tsx#L255

packages/perseus-core/src/widgets/label-image/index.ts
```diff
@@ -30,6 +30,7 @@ const labelImageWidgetLogic: WidgetLogic = {
     name: "label-image",
     defaultWidgetOptions,
     getPublicWidgetOptions: getLabelImagePublicWidgetOptions,
+    accessible: true,
```
catandthemachines, line 33: Nit: Same here, could you add a TODO comment above that links to the following task:
https://khanacademy.atlassian.net/browse/LEMS-3103

You can see an example of that here: https://github.com/Khan/perseus/blob/d64277bc85b0b533ef85eec2e07d24a6c407f089/packages/perseus/src/widgets/expression/expression.tsx#L255

packages/perseus-core/src/widgets/logic-export.types.ts
```diff
@@ -55,6 +55,8 @@ export type WidgetLogic = {
     defaultWidgetOptions?: any;
     supportedAlignments?: ReadonlyArray<Alignment>;
     defaultAlignment?: Alignment;
+    accessible?: boolean | ((options: PerseusWidgetOptions) => boolean);
+    traverseChildWidgets?: any;
```
catandthemachines, line 59: Nit: Can we add a type here?

packages/perseus-editor/src/editor.tsx
catandthemachines, line 1: I think I recall that this whole file needs to be reverted for this PR as this is content that's not in scope for the change.

packages/perseus-editor/src/issue-details.tsx
catandthemachines, line 1: This file also looks to be out of scope, please remove.

packages/perseus-editor/src/issues-panel.tsx
catandthemachines, line 1: Please remove, out of scope.

packages/perseus-editor/src/item-editor.tsx
catandthemachines, line 1: Please remove, out of scope.

packages/perseus-editor/src/messages.ts
catandthemachines, line 1: Please remove, out of scope.

packages/perseus-linter/src/rules/all-rules.ts
catandthemachines, line 1: Please remove, out of scope.

packages/perseus-linter/src/rules/inaccessible-widget.ts
catandthemachines, line 1: Please remove, out of scope.

packages/perseus-core/src/widgets/logic-export.types.ts
catandthemachines: At least `(props: any, traverseRenderer: any) => any` ❤️

catandthemachines (commented):
The changes look good, just need to remove all the content that's from the other PR. 😊

packages/perseus-editor/src/editor.tsx
```diff
@@ -739,6 +746,47 @@ class Editor extends React.Component<Props, State> {
         textarea.focus();
     };
 
+    // Tracks previous widget IDs to diff out removals and keep the Issues Panel in sync
```
jeremywiebe, line 749: I don't think we need this function anymore now that we do accessibility checks in the lint rule, right?

packages/perseus-editor/src/item-editor.tsx
```diff
@@ -90,6 +98,25 @@ class ItemEditor extends React.Component<Props> {
         };
     };
 
+    handleAddWarning = (warning: Issue) => {
```
jeremywiebe, line 101: This looks like the old code before we moved to the lint rule. 🤔

jeremywiebe (commented):
I think there is some code in this PR from the old way (before we moved the check to a lint rule). Let me know when you've cleaned up that old code and I'm happy to review again.

.changeset/red-monkeys-ring.md
tatianasnook: removed

packages/perseus-core/src/widgets/interactive-graph/index.ts
tatianasnook: Added the TODO comment with the task link

packages/perseus-core/src/widgets/label-image/index.ts
tatianasnook: Added the TODO comment with the task link

packages/perseus-core/src/widgets/logic-export.types.ts
tatianasnook: Updated traverseChildWidgets to use an explicit function type: (props: any, traverseRenderer: any) => any as suggested.

packages/perseus-core/src/widgets/group/index.ts
tatianasnook: Only certain widgets like group, graded-group, and graded-group-set define a traverseChildWidgets function because their widget options include nested renderers—such as sub-questions or embedded groups of content.
This field is necessary to recursively apply traversal logic (e.g., for accessibility validation or transformations) to those nested structures.
Most other widgets don’t embed other renderers in their options, so they don’t need this additional traversal step.

.changeset/red-monkeys-ring.md
```diff
@@ -0,0 +1,9 @@
+---
+"@khanacademy/perseus-core": patch
+"@khanacademy/perseus-editor": patch
+---
+
+(LEMS-2972) This PR displays a warning in the Issues Panel when a content creator selects or pastes a widget that is marked as accessible: false.
```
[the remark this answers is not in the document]
tatianasnook: removed

packages/perseus-editor/src/editor.tsx
tatianasnook: I've reverted this file to its previous state, as the changes are already being handled in PR #2474.

packages/perseus-editor/src/issue-details.tsx
tatianasnook: I've reverted this file to its previous state, as the changes are already being handled in PR #2474.

packages/perseus-core/src/utils/is-accessible.ts
```diff
@@ -0,0 +1,18 @@
...
+): boolean {
+    if (!widgetLogic) {
+        return false;
+    }
+    const {accessible} = widgetLogic;
+
+    return typeof accessible === "function"
+        ? accessible(widgetInfo.options)
+        : !!accessible;
+}
```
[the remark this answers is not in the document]
tatianasnook: Yep, you're right — this is duplicated logic. The core-widget-registry.ts version already does this lookup, so it makes sense to delete this one and treat the core registry as the single source of truth.

packages/perseus-core/src/widgets/core-widget-registry.ts
```diff
@@ -31,20 +36,24 @@ import sorterWidgetLogic from "./sorter";
...
 import videoWidgetLogic from "./video";
 
+import type {PerseusWidgetOptions} from "../data-schema";
 import type {
     PublicWidgetOptionsFunction,
     WidgetLogic,
 } from "./logic-export.types";
 import type {Alignment} from "../types";
+// eslint-disable-next-line @typescript-eslint/no-restricted-imports
+import type {PerseusWidget} from "@khanacademy/perseus-core";
```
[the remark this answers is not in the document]
tatianasnook: Thanks, good catch! I've updated the import to use a relative path as suggested

packages/perseus-editor/src/item-editor.tsx
tatianasnook: You're right — handleAddWarning did exist in an earlier version, but I’ve since updated item-editor.tsx to remove it.

packages/perseus-editor/src/editor.tsx
tatianasnook: You're right — we no longer need the checkAccessibilityAndWarn function. That logic has been moved to the custom lint rule.

packages/perseus-core/src/widgets/core-widget-registry.ts
```diff
@@ -37,22 +40,22 @@ import type {
     PublicWidgetOptionsFunction,
     WidgetLogic,
 } from "./logic-export.types";
-import type {PerseusWidgetOptions} from "../data-schema";
+import type {PerseusWidgetOptions, PerseusWidget} from "../data-schema";
 import type {Alignment} from "../types";
 
-const widgets = new Registry<WidgetLogic>("Core widget registry");
+const widgets: Record<string, WidgetLogic> = {};
```
jeremywiebe, line 46: We need to revert this. Matthew [recently introduced](https://github.com/Khan/perseus/pull/2483) the `Registry` class to protect against use before initialization.  We'll also want to revert the test changes you made in `packages/perseus-core/src/widgets/core-widget-registry.test.ts`

packages/perseus-core/src/widgets/core-widget-registry.ts
```diff
@@ -83,6 +86,37 @@ export function isAccessible(
...
+    // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
+    if (!widgetInfo || !widgetInfo.type || !widgets[widgetInfo.type]) {
+        return widgetInfo;
+    }
+
+    const widgetExports = widgets[widgetInfo.type];
+    const props = widgetInfo.options;
+
+    // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
+    if (widgetExports.traverseChildWidgets && props) {
```
jeremywiebe, line 110: **suggestion**: We can fix this lint warning like this: 

```suggestion
    if (widgetExports.traverseChildWidgets != null && props != null) {
```

packages/perseus-core/src/widgets/core-widget-registry.ts
```diff
@@ -83,6 +86,37 @@ export function isAccessible(
...
+        const newProps = widgetExports.traverseChildWidgets(
+            props,
+            traverseRenderer,
+        );
+        return _.extend({}, widgetInfo, {options: newProps});
```
jeremywiebe, line 115: We're working to remove usages of `underscore` (`_`) in our codebase.  In this case, this `extend()` function can be written in standard ES5 like this: 

```suggestion
        return {...widgetInfo, options: newProps};
```

packages/perseus-core/src/widgets/core-widget-registry.ts
```diff
@@ -99,7 +133,7 @@ export function isAccessible(
 export const getSupportedAlignments = (
     type: string,
 ): ReadonlyArray<Alignment> => {
-    const widgetLogic = widgets.get(type);
+    const widgetLogic = widgets[type];
```
jeremywiebe, line 136:
```suggestion
    const widgetLogic = widgets.get(type);
```

packages/perseus-core/src/widgets/core-widget-registry.ts
```diff
@@ -117,7 +151,7 @@ export const getSupportedAlignments = (
  * the exports of a widget's module.
  */
 export const getDefaultAlignment = (type: string): Alignment => {
-    const widgetLogic = widgets.get(type);
+    const widgetLogic = widgets[type];
```
jeremywiebe, line 154:
```suggestion
    const widgetLogic = widgets.get(type);
```

packages/perseus-core/src/widgets/image/index.ts
```diff
@@ -28,6 +31,14 @@ const imageWidgetLogic: WidgetLogic = {
     defaultWidgetOptions,
     supportedAlignments: ["block", "full-width"],
     defaultAlignment: "block",
+    // This widget's accessibility depends on its contents: if the image
```
jeremywiebe, line 34:
```suggestion
    // This widget's accessibility depends on its widget options: if the image
```

packages/perseus-core/src/widgets/image/index.ts
```diff
@@ -28,6 +31,14 @@ const imageWidgetLogic: WidgetLogic = {
...
+    // has a background but no alt text, it is not accessible
+    accessible: (widgetOptions: PerseusWidgetOptions): boolean => {
+        const imageOptions = widgetOptions as PerseusImageWidgetOptions;
+        const bgImage = imageOptions.backgroundImage;
+        // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
+        return !(bgImage && bgImage.url && !imageOptions.alt);
```
jeremywiebe, line 40: We can fix the lint warning instead of suppressing it. 

```suggestion
        return !(bgImage?.url != null && !imageOptions.alt);
```

jeremywiebe (approved):
There's a few things to clean up (see my comments), but then this is good to land! We should also remove the `accessible` field from the `export type WidgetExports<>` that lives in `packages/perseus/src/types.ts`.  That has also moved into `perseus-core` now. 

Feel free to tag me once more if you want another review.

packages/perseus-core/src/widgets/core-widget-registry.ts
tatianasnook: Restores use of the Registry class for widget logic management to preserve protection against access before initialization. Test cases have been updated to expect errors when accessing the registry prior to calling registerCoreWidgets().

packages/perseus-core/src/widgets/core-widget-registry.ts
tatianasnook: Thanks! Updated to use explicit != null checks as suggested — cleaner and satisfies the linter.

packages/perseus-core/src/widgets/core-widget-registry.ts
tatianasnook: Thank you for the suggestion! I've replaced _.extend with object spread syntax for consistency with modern JavaScript and to support the effort to remove underscore usage.

packages/perseus-core/src/widgets/core-widget-registry.ts
tatianasnook: I changed it back to .get() because widgets is an instance of the Registry class, so accessing entries requires the .get() method rather than direct property access.

packages/perseus-core/src/widgets/core-widget-registry.ts
tatianasnook: I changed it back to .get() because widgets is an instance of the Registry class, so accessing entries requires the .get() method rather than direct property access.

packages/perseus-core/src/widgets/image/index.ts
tatianasnook: Thanks for the suggestion! I updated the comment to clarify that the accessibility depends on the widget options rather than the contents.

packages/perseus-core/src/widgets/image/index.ts
tatianasnook: Thanks for the suggestion! I agree this is cleaner and avoids disabling lint rules.

packages/perseus-core/src/widgets/core-widget-registry.test.ts
```diff
@@ -1,39 +1,60 @@
...
+            "Core widget registry accessed before initialization!",
+        );
     });
 
-    test("getCurrentVersion returns default version before initialization", () => {
-        expect(CoreWidgetRegistry.getCurrentVersion("radio")).toEqual({
-            major: 0,
-            minor: 0,
-        });
+    test("isWidgetRegistered throws before initialization", () => {
```
jeremywiebe, line 19: I think all these individual "throws before initialization" tests are covered by the above `test.each()` test.  That tests basically iterates over each of the named functions and ensures that they throw if called before initialization.

packages/perseus/src/widgets/image/image.test.ts
```diff
@@ -50,25 +49,4 @@ describe.each([[true], [false]])("image widget - isMobile(%j)", (isMobile) => {
...
-            ...imageWidget,
-            options: {
-                ...options,
-                alt: "",
-            },
-        };
-
-        // Act and Assert
-        expect(isAccessible(inaccessibleWidgetInfo)).toBe(false);
-    });
```
jeremywiebe, line 73: Just want to double-check that these tests were ported to `perseus-core`'s `image` widget logic. We still want these unit tests, they just will live in `perseus-core` now. :)

packages/perseus-core/src/widgets/core-widget-registry.ts
```diff
@@ -78,11 +80,41 @@ export function isAccessible(
...
+
+    // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
+    if (!widgetInfo || !widgetInfo.type || !widgets[widgetInfo.type]) {
+        return widgetInfo;
+    }
+
+    const widgetExports = widgets[widgetInfo.type];
+    const props = widgetInfo.options;
+
+    if (widgetExports.traverseChildWidgets != null && props != null) {
```
jeremywiebe, line 108: This should be using the `.get()` function instead of an indexer(`[]`). Once you do this, you'll see that we need to adjust the check too slightly. 

```suggestion
    const widgetExports = widgets.get(widgetInfo.type);
    const props = widgetInfo.options;

    if (widgetExports?.traverseChildWidgets != null && props != null) {
```

packages/perseus/src/renderability.ts
```diff
@@ -65,12 +63,30 @@ const isRawWidgetInfoRenderableBy = function (
...
+            widgetInfo &&
+            (widgetInfo.type === "group" || widgetInfo.type === "sequence") &&
+            widgetInfo.options?.widgets
+        ) {
+            const subRenderable = isRendererContentRenderableBy(
+                widgetInfo.options,
+                rendererContentVersion,
+            );
+            isRenderable = isRenderable && subRenderable;
+        }
```
jeremywiebe, line 89: We can do this in a new PR, but we shouldn't be doing this manually. The `traverse()` function knows how to ask widgets to recurse into themselves via `traverseChildWidgets()` on the `WidgetLogic` type and so this manual work shouldn't be necessary.

jeremywiebe (approved):
There's a few remaining issue comments. I think we can create JIRA tickets for them and then land this PR (please update this PR with comments noting the JIRA tickets). Great work sorting out all this code moving.

packages/perseus/src/renderability.ts
catandthemachines: I second this, let's create a Jira ticket to clean this up later and add a TODO comment that links to the ticket ❤️

packages/perseus/src/widgets/image/image.test.ts
catandthemachines: I agree, lets make sure these tests are in perseus-core. Timebox the addition of these tests to an hour. If it takes any longer let's capture this with a Jira ticket and a TODO comment.

packages/perseus-core/src/widgets/core-widget-registry.test.ts
tatianasnook: Thanks, Jeremy! I’ve removed the redundant individual tests.

packages/perseus-editor/src/__tests__/traversal.test.ts
catandthemachines, line 1: This is interesting that this lived in the perseus-editor package when traversal.ts lived in the perseus package. Can we move this to the `perseus-core` package or leave it here. I would like to not remove tests that capture functionality. ❤️

packages/perseus/src/__tests__/a11y.test.ts
catandthemachines, line 1: Let's also move this over to `perseus-core` and not delete it. That way we don't loose tests ❤️

packages/perseus/src/widgets/group/group.test.tsx
```diff
@@ -525,54 +524,6 @@ describe("group widget", () => {
         ).toBeInTheDocument();
     });
 
-    describe("traverseChildWidgets", () => {
```
catandthemachines, line 528: Can this test be moved into `perseus-core` or can we create a jira ticket to move this test over? Don't want to loose our tests. 😉

packages/perseus-core/src/widgets/core-widget-registry.ts
tatianasnook: Thanks for pointing that out! I've updated it to use .get()

packages/perseus/src/widgets/image/image.test.ts
tatianasnook: Moved these tests to perseus-core under the image widget logic.

packages/perseus/src/renderability.ts
tatianasnook: I’ve created a Jira ticket for this cleanup: LEMS-3142, and added a TODO comment linking to it in the code.

packages/perseus-editor/src/__tests__/traversal.test.ts
tatianasnook: I’ve created a Jira ticket for this cleanup: LEMS-3142.

packages/perseus/src/__tests__/a11y.test.ts
tatianasnook: I’ve created a Jira ticket for this cleanup: LEMS-3142.

packages/perseus/src/widgets/group/group.test.tsx
tatianasnook: I’ve created a Jira ticket for this cleanup: LEMS-3142.

PR #2486, opened 2025-05-14
Update Free Response styles and accessibility

aag:
## Summary:
This commit updates the styles and layout of the Free Response Widget to
match the designs. It also updates some markup to improve the
accessibility of the widget, including using the `LabeledField` Wonder
Blocks component.

The designs can be found in Figma:
https://www.figma.com/design/h5nZthMLTd56M0GLYsrDry/ELA---Free-Response?node-id=10184-1509&t=7etutvzW6ikoM9cS-1

Issue: https://khanacademy.atlassian.net/browse/AX-963

## Test plan:
- View the widget variants in Storybook.
- Verify that the rendered widgets match the designs.

image


packages/perseus-editor/package.json
```diff
@@ -59,6 +59,7 @@
         "@khanacademy/wonder-blocks-form": "catalog:",
         "@khanacademy/wonder-blocks-icon": "catalog:",
         "@khanacademy/wonder-blocks-icon-button": "catalog:",
+        "@khanacademy/wonder-blocks-labeled-field": "catalog:",
```
jeremywiebe, line 62: I don't see any `perseus-editor` changes in this PR, so is adding this dependency here needed?

packages/perseus/package.json
```diff
@@ -65,6 +65,7 @@
         "@khanacademy/wonder-blocks-form": "catalog:",
         "@khanacademy/wonder-blocks-icon-button": "catalog:",
         "@khanacademy/wonder-blocks-icon": "catalog:",
+        "@khanacademy/wonder-blocks-labeled-field": "catalog:",
```
jeremywiebe, line 68: We should list this dependency in the `peerDependencies` of this package also. Some commentary that might help: https://khanacademy.atlassian.net/wiki/spaces/LC/pages/2473951425/Should+Perseus+dependencies+be+peer+dependencies

package.json
```diff
@@ -20,6 +20,7 @@
         "@khanacademy/wonder-blocks-button": "catalog:",
         "@khanacademy/wonder-blocks-core": "catalog:",
         "@khanacademy/wonder-blocks-icon": "catalog:",
+        "@khanacademy/wonder-blocks-labeled-field": "catalog:",
```
jeremywiebe, line 23: We also don't need this as a dependency here. I need to clean these up, but we should only have tooling and storybook-required dependencies listed here.  Each package defines the deps it needs in its own `package.json`. Sorry for the confusion.

jeremywiebe (approved):
Looks reasonable to me.  Just some deps cleanup to address and then it's good to go! :)

rgpass (approved):
LGTM! 👏

PR #2489, opened 2025-05-15
Convert LESS files to pre-built CSS files to simplify our styling framework

mark-fitzgerald:
## Summary: ##

Issue: LEMS-3088

## Test plan:


.storybook/main.ts
```diff
@@ -3,21 +3,28 @@ import {spacing} from "@khanacademy/wonder-blocks-tokens";
...
-        if (pathname.endsWith(".less")) {
-            const layerStatements =
-                "@layer reset, shared, legacy;\n@layer shared";
-            return {
-                code: `${layerStatements} { ${code} }`,
-                map: null,
-            };
+        if (pathname.endsWith(".css")) {
+            // Exclude the CSS files that are not part of the shared layer.
+            if (!excludedCssFiles.some((file) => pathname.endsWith(file))) {
```
mark-fitzgerald, line 17: I don't think we need this check in here, but since this is (relatively) temporary code, I decided to error on the side of caution.

The files that it is checking were excluded from the layering in Storybook, so I'm aiming to keep the same output.

packages/perseus/src/styles/perseus-renderer.css
```diff
@@ -0,0 +1,23 @@
...
+@import "widgets/matrix.css";
+@import "widgets/measurer.css";
+@import "widgets/numeric.css";
+@import "widgets/orderer.css";
+@import "widgets/passage.css";
+@import "widgets/plotter.css";
+@import "widgets/radio.css";
+@import "widgets/sortable.css";
+@import "widgets/table.css";
+@import "./perseus-renderer-part-2.css";
```
mark-fitzgerald, line 23: This file replicates the definition order of our styling when it was in LESS format. The styling in the original `perseus-renderer.less` had `import` statements interspersed among style definitions (which [you can't do in regular CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/@import#:~:text=An%20%40import%20rule%20must%20be%20defined%20at%20the%20top%20of%20the%20stylesheet%2C%20before%20any%20other%20at%2Drule%20(except%20%40charset%20and%20%40layer)%20and%20style%20declarations%2C%20or%20it%20will%20be%20ignored.)). So, I broke out the definitions into two separate CSS files (`-part-1.css` and `-part-2.css`) and structured the import order to mirror the original file.

packages/perseus/src/styles/zoom.css
```diff
@@ -1,70 +1,53 @@
...
     cursor: -moz-zoom-in;
 }
-
 .zoom-img {
-    // NOTE(charlie): We set the zoomed image's background color to white so
-    // that any images with transparent backgrounds (like SVGs and some PNGs)
-    // remain legible over the article content, as our backdrop is opaque.
     background-color: white;
     position: absolute;
-    z-index: @zoomOverlayZIndex + 1;
```
jeremywiebe, line 20: Can we use `calc()` here so we don't lose the intent of this value?

jeremywiebe (approved):

packages/math-input/styles/main.css
```diff
@@ -0,0 +1,84 @@
+@import "../node_modules/mathquill/build/mathquill.css";
```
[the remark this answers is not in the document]
mark-fitzgerald: I'm not sure. This is what was generated by the build out of the LESS compiler. I'm inclined to not mess with this since it is a CSS import statement, and not a JS import statement.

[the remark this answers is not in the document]
jeremywiebe: These CSS files are still processed by `postcss` through the `rollup-plugin-styles` so my suggestion should work. I wouldn't push if this was in a deprecated widget, but given this is in the `math-input`'s main styles, I think it's worth trying as this is  a file we need to actively maintain.  By using the variant I've suggested, we'll be shielded from breakage if this file is ever moved because the tooling resolves the `mathquill.css` file from the `mathquill` package for us.

.storybook/global.css
```diff
@@ -4,29 +4,22 @@
...
 }
-
 body {
-    font-family: @wonderBlocksFontFamily, "Helvetica", "Corbel", sans-serif;
+    font-family: "Lato", "Noto Sans", "Helvetica", "Corbel", sans-serif,
+        "Helvetica", "Corbel", sans-serif;
     font-size: 14px;
     margin: 0;
-    color: @offBlack;
+    color: #21242c;
```
ivyolamit, line 16: @mark-fitzgerald is there a reason why we are not using [WB](https://khan.github.io/wonder-blocks/?path=/docs/packages-tokens-color--docs) `color.offBlack` here?

note: Also some other usages in the PR using the explicit value instead of tokens.

packages/perseus/src/styles/perseus-renderer.css
mark-fitzgerald: Added.

.storybook/main.ts
```diff
@@ -3,21 +3,28 @@ import {spacing} from "@khanacademy/wonder-blocks-tokens";
 
 import type {StorybookConfig} from "@storybook/react-vite";
 
+const excludedCssFiles = ["lato.css", "protractor.css", "mafs-styles.css"];
 // This is a temporary plugin option to mimic what is in PROD in regard to cascade layers.
 // Perseus LESS files are wrapped in the 'shared' layer in Webapp.
 // To get the same ordering of precedence in Storybook, the imported LESS files need to be wrapped accordingly.
 // Once the LESS files have cascade layers included (LEMS-2801),
```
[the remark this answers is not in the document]
mark-fitzgerald: Done.

packages/math-input/styles/main.css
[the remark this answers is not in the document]
mark-fitzgerald: Done.

packages/perseus/src/styles/zoom.css
```diff
@@ -1,70 +1,53 @@
 /* Derived from the MIT-licensed zoom.js:
    https://github.com/fat/zoom.js/blob/fd4f3e43153da7596da0bade198e99f98b47791e/
 */
-
-@zoomOverlayZIndex: 9000;
```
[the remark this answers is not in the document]
mark-fitzgerald: Done.

packages/perseus/src/styles/zoom.css
mark-fitzgerald: Done.

.storybook/global.css
mark-fitzgerald: The LESS files didn't reference anything from Wonder Blocks because those tokens only existed as JavaScript objects, nothing in CSS. [Color tokens as CSS Variables](https://khan.github.io/wonder-blocks/?path=/docs/packages-tokens-semantic-color--docs#tokens) are just now available. While this PR is intended to only convert from LESS to CSS and not update styles to Wonder Blocks tokens, I did update this reference in the Storybook styling.

ivyolamit (approved):
@mark-fitzgerald PR looks logical to me and thanks for converting this, it will make our lives easier moving forward 💅🏼

PR #2493, opened 2025-05-15
Use LabeledField in FreeResponseEditor

aag:
## Summary:
This commit refactors the FreeResponseEditor component to use the
LabeledField component from Wonder Blocks. This makes the code more
consistent with the FreeResponse component and also eliminates a lint error.

Issue: https://khanacademy.atlassian.net/browse/AX-963

## Test plan:
- View the FreeResponseEditor Storybook.
- Verify that the component is displayed correctly.

image


packages/perseus-editor/src/widgets/free-response-editor.tsx
```diff
@@ -107,46 +108,61 @@ class FreeResponseEditor extends React.Component<Props> {
...
+                    label={<HeadingSmall>Question</HeadingSmall>}
+                    field={
+                        <textarea
+                            value={this.props.question}
+                            onChange={(e) =>
+                                this.props.onChange({question: e.target.value})
+                            }
+                        />
+                    }
+                    styles={{root: styles.labeledInputField}}
```
handeyeco, line 121: Just out of curiosity, what's the `root` thing? Pre-coffee on a Monday morning and I feel like I haven't seen that before.

handeyeco (approved):

packages/perseus-editor/src/widgets/free-response-editor.tsx
aag: Since the `LabeledField` component has multiple components and states, the `styles` prop accepts separate style objects for `root` (the top-level `LabeledField` element), `label` (the label element), and `error` (probably styles to apply to the `field` element when it's an an error state). As far as I know, `LabeledField` is the only Wonder Blocks component that does this.

PR #2498, opened 2025-05-16
Add Free Response scoring and validation functions

aag:
## Summary:
This commit adds scoring and validation functions for the Free Response
Widget and registers them.

Issue: https://khanacademy.atlassian.net/browse/AX-1005

## Test plan:
- Run `pnpm test`.

### DIFF
--- a/packages/perseus-score/src/widgets/free-response/score-free-response.ts
+++ b/packages/perseus-score/src/widgets/free-response/score-free-response.ts
@@ -1,0 +1,23 @@
+import type {
+    PerseusFreeResponseUserInput,
+    PerseusFreeResponseRubric,
+    PerseusScore,
+} from "@khanacademy/perseus-core";
+
+function scoreFreeResponse(
+    userInput: PerseusFreeResponseUserInput,
+    rubric: PerseusFreeResponseRubric,
+    locale: string,
+): PerseusScore {
+    // TODO: Implement support for external scoring. Since the user's input is
+    //       free text input, we can't easily do scoring inside the widget, so
+    //       we'll need to have a way to do it by some other method. For now,
+    //       we'll just always give full credit.
+    return {
+        type: "points",
+        earned: 1,
+        total: 1,
+    };
+}
+
+export default scoreFreeResponse;
--- a/packages/perseus-score/src/widgets/free-response/validate-free-response.test.ts
+++ b/packages/perseus-score/src/widgets/free-response/validate-free-response.test.ts
@@ -1,0 +1,13 @@
+import validateFreeResponse from "./validate-free-response";
+
+describe("free response validation", () => {
+    it("should return invalid for empty user input", () => {
+        const result = validateFreeResponse({currentValue: ""});
+        expect(result).toHaveInvalidInput();
+    });
+
+    it("should return null for non-empty user input", () => {
+        const result = validateFreeResponse({currentValue: "A short answer"});
+        expect(result).toBeNull();
+    });
+});
--- a/packages/perseus-score/src/widgets/free-response/validate-free-response.ts
+++ b/packages/perseus-score/src/widgets/free-response/validate-free-response.ts
@@ -1,0 +1,20 @@
+import type {
+    PerseusFreeResponseUserInput,
+    ValidationResult,
+} from "@khanacademy/perseus-core";
+
+/**
+ * Checks user input from the free response widget to see if it is scorable.
+ * Since the input is free text, we only check that it is not empty.
+ */
+function validateFreeResponse(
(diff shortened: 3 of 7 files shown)


packages/perseus-score/src/widgets/free-response/score-free-response.ts
```diff
@@ -0,0 +1,23 @@
...
+): PerseusScore {
+    // TODO: Implement support for external scoring. Since the user's input is
+    //       free text input, we can't easily do scoring inside the widget, so
+    //       we'll need to have a way to do it by some other method. For now,
+    //       we'll just always give full credit.
+    return {
+        type: "points",
+        earned: 1,
+        total: 1,
+    };
```
handeyeco, line 20: Do you mind using `scoreNoop`? It can do the same thing, it's just easier for us to find when we need to look into widgets that aren't actually doing real scoring. I know this is temporary, but just in case.

packages/perseus-score/src/widgets/free-response/validate-free-response.ts
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Checks user input from the free response widget to see if it is scorable.
+ * Since the input is free text, we only check that it is not empty.
+ */
+function validateFreeResponse(
+    userInput: PerseusFreeResponseUserInput,
+): ValidationResult {
+    if (userInput.currentValue.trim() === "") {
+        return {type: "invalid", message: "The answer cannot be empty."};
```
handeyeco, line 14: We can't use strings here. The problem is that the validation code will run on both the client and the server. Rather than having the server have access to all strings or trying to map `en` strings to other locales, we decided that the server will return error codes and those codes will get mapped to translated strings.

See `validateCategorizer`.

packages/perseus-score/src/widgets/free-response/validate-free-response.test.ts
```diff
@@ -0,0 +1,18 @@
+import validateFreeResponse from "./validate-free-response";
+
+describe("free response validation", () => {
+    it("should return invalid for empty user input", () => {
+        const result = validateFreeResponse({currentValue: ""});
+        expect(result).toHaveInvalidInput();
```
handeyeco, line 6: I would recommend adding an expected validation message (which is just an error code, like I mentioned in the other comment):

```suggestion
        expect(result).toHaveInvalidInput("USER_INPUT_EMPTY");
```

See `perseus/packages/perseus-score/src/widgets/categorizer/validate-categorizer.test.ts`

packages/perseus-score/src/widgets/free-response/score-free-response.ts
```diff
@@ -0,0 +1,23 @@
...
+    PerseusFreeResponseRubric,
+    PerseusScore,
+} from "@khanacademy/perseus-core";
+
+function scoreFreeResponse(
+    userInput: PerseusFreeResponseUserInput,
+    rubric: PerseusFreeResponseRubric,
+    locale: string,
+): PerseusScore {
+    // TODO: Implement support for external scoring. Since the user's input is
```
handeyeco, line 12: nit: it's nice when we can point to Jira tickets when adding TODOs.

handeyeco (changes requested):
Looking like a good start! Just a concern about possible learner-facing strings coming out of the server.

handeyeco:
Re TS errors: you're adding a potentially user-facing string. It'll need to go in `PerseusStrings`, `strings`, and `mockStrings`; it'll also need a translation request.

aag:
> Re TS errors: you're adding a potentially user-facing string. It'll need to go in `PerseusStrings`, `strings`, and `mockStrings`; it'll also need a translation request.

Should I see a "New i18n Strings for Translation" comment in the PR? Also, should I wait a day until the string is in Crowdin, or open the translation request right away?

packages/perseus/src/strings.ts
```diff
@@ -1472,8 +1475,8 @@ const errorToString: ErrorStringMap = {
     CHOOSE_CORRECT_NUM_ERROR: "chooseCorrectNum",
     NOT_NONE_ABOVE_ERROR: "notNoneOfTheAbove",
     FILL_ALL_CELLS_ERROR: "fillAllCells",
+    USER_INPUT_EMPTY: "USER_INPUT_EMPTY",
 };
-
```
handeyeco, line 1476: Would prefer to keep this.

handeyeco (approved):

packages/perseus/src/strings.ts
aag: Oops, not sure how that snuck in there.

PR #2504, opened 2025-05-19
Update dropdown labeling guidance to internal article

marcysutton:
## Summary:
I got feedback from the Content team that the w3c-WAI article linked from the editor was too technical for the audience. So I created an internal doc that we can maintain, that also links to the previous article for those who are curious.

Issue: FEI-6288

## Test plan:
1. Ensure the links work (the doc is on Confluence)
2. Ensure the link text fits ("this internal article" instead of "this article")

### DIFF
--- a/packages/perseus-editor/src/widgets/dropdown-editor.tsx
+++ b/packages/perseus-editor/src/widgets/dropdown-editor.tsx
@@ -147,11 +147,11 @@ class DropdownEditor extends React.Component<Props> {
                             exercise is accessible. For more information on
                             writing accessible labels, please see{" "}
                             <a
-                                href="https://www.w3.org/WAI/tips/designing/#ensure-that-form-elements-include-clearly-associated-labels"
+                                href="https://khanacademy.atlassian.net/wiki/spaces/A11Y/pages/3967615350/General+Guidance+for+Labeling+Dropdowns"
                                 target="_blank"
                                 rel="noreferrer"
                             >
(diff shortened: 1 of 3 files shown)


packages/perseus-editor/src/widgets/dropdown-editor.tsx
```diff
@@ -147,11 +147,11 @@ class DropdownEditor extends React.Component<Props> {
                             exercise is accessible. For more information on
                             writing accessible labels, please see{" "}
                             <a
-                                href="https://www.w3.org/WAI/tips/designing/#ensure-that-form-elements-include-clearly-associated-labels"
+                                href="https://khanacademy.atlassian.net/wiki/spaces/A11Y/pages/3967615350/General+Guidance+for+Labeling+Dropdowns"
                                 target="_blank"
                                 rel="noreferrer"
                             >
-                                this article.
+                                this internal article.
```
marcysutton, line 154: My one question is: can we link to an internal article in this project? Or am I not adequately considering external audiences since this is an OS project?

handeyeco: I personally think it would be preferable to avoid linking to internal links from the UI.

marcysutton: Thanks for the response! How do we give relevant advice to authors, in that case? Publish something external?

PR #2505, opened 2025-05-20
Fix Label Image so PreferredPopoverDirection works

Myranae:
## Summary:
This value was previously hardcoded in the Behavior component as "NONE". I was able to update the props to pass the value from the editor down to the Behaior component so the logic for popover direction has access to the information provided by the content creator.

Issue: LEMS-3110

## Test plan:
- Confirm all checks pass
- Test in Storybook and confirm changing PreferredPopoverDirection persists


Myranae:
Will be removing this functionality from the editor for now and will possibly spend time setting up the feature at a later date. Will reference this PR in the new ticket.

PR #2510, opened 2025-05-20
Add tab navigation to new radio widget

ivyolamit:
## Summary:
This PR will add the proper tab navigation to the new radio widget after the horizontal scrollbar was added. Removing the visible blue focus in the entire row instead of just the choice icon.

| Before | After |
| ------ | ------ |
|  |  |

**Note: Known Existing Issue**
When displaying an image as the content will cause an additional focus on the image.
| Before (Prod) | After |
| ------ | ------ |
|  |  |

Issue: LEMS-3105

## Test plan:
- [ ] Confirm that the tab navigation works for the New Radio widget storybook

### DIFF
--- a/.changeset/eighty-swans-love.md
+++ b/.changeset/eighty-swans-love.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Add tab navigation to new radio widget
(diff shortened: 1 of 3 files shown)


anakaren-rojas (approved):

mark-fitzgerald (approved):
Simple. Thank you!

PR #2511, opened 2025-05-21
Update TypeScript config to align with Khan Academy's minimum supported browser

jeremywiebe:
## Summary:

Aaaaactually, we're not gonna do this because the original function that I thought we needed polyfills for is supported in all major browsers that Khan Academy supports (Safari 15.6 most notably). So this PR now simply adjusts the TypeScript `target` to be `ES2021` (ES2021 is [fully supported](https://caniuse.com/?search=ES2021) by Safari 15.6 and Chrome 128 - Khan Academy's current minimum supported browsers). 

~~This PR updates our build setup to polyfill functions for browsers down to Khan Academy's lowest-supported browsers (and a little lower, for Safari). We've been seeing `p.at() is not a function` errors in our analytics dashboard. `Array.at` was added in Safari 15.4 (Khan Academy's minimum Safari browser is 15.6) and so it seems useful to fix this as we have some amount of users actively encountering issues with Perseus on Safari < 15.4 (reference: https://caniuse.com/?search=Array.at)~~

Issue: LEMS-3085

## Test plan:

`pnpm tsc` doesn't complain!

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,6 @@
         "@vitejs/plugin-react-swc": "^3.8.1",
         "ancesdir": "^3.1.0",
         "aphrodite": "catalog:",
-        "core-js": "3.42.0",
         "cross-env": "^5.2.0",
         "css-loader": "^6.8.1",
         "cypress": "^13.6.5",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -150,7 +150,6 @@
     "packageManager": "pnpm@10.8.0",
     "pnpm": {
         "onlyBuiltDependencies": [
-            "core-js",
```
jeremywiebe, line 153: We'd removed this dependency a while back so this was unused already.

ivyolamit (approved):

PR #2512, opened 2025-05-21
Remove preferred popover direction from Label Image editor

Myranae:
## Summary:
The option from the preferred pop-over direction dropdown was not persisting into the JSON for the exercise, and it did not show the chosen option to the content author. This removes the broken dropdown.

### Before
Label Editor with preferred popover direction

### After
Label Editor without preferred popover direction

There is potential future work here to re-implement this functionality in the editor and with Label Image if possible. See LEMS-3126.

Issue: LEMS-3110

## Test plan:
- Confirm all tests and checks pass
- Confirm the label image editor works as expected via Storybook and ZND
- Confirm the PreferredPopoverDirection dropdown is gone.


handeyeco (approved):
Thank you

PR #2517, opened 2025-05-22
Removing manual traversal.

catandthemachines:
## Summary:
Remove the manual traversal of subwidgets as the `traverse` function handles this logic.

Issue: LEMS-3142

## Test plan:
Run `pnpm test` all tests should pass.

### DIFF
--- a/.changeset/thick-weeks-kneel.md
+++ b/.changeset/thick-weeks-kneel.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Removed unnecessary logic.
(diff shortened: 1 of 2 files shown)


jeremywiebe (approved):
The [tests](https://github.com/Khan/perseus/blob/main/packages/perseus/src/__tests__/renderability.test.ts) continue to pass so I think we're good!

tatianasnook (approved):
After reviewing the implementation, I agree that we're already importing and using the traverse() utility from perseus-core, which internally handles recursion into nested widget structures using traverseChildWidgets().
This makes the manual checks for group and sequence widgets unnecessary and redundant.

PR #2518, opened 2025-05-22
docs(changeset): Updates to Radio Scoring to ensure non-specific multiselect questions are marked correctly.

SonicScrewdriver:
## Summary:
This PR is part of the Radio Widget Upgrade Project. 

This work ensures that non-specific multi-select Radios don't trigger CHOOSE_CORRECT_NUM_ERRORs, and are instead marked as "incorrect". 

Issue: LEMS-2996

## Test plan:
- New scoring test 
- Previous tests pass (after making them specific)

### DIFF
--- a/packages/perseus-core/src/validation.types.ts
+++ b/packages/perseus-core/src/validation.types.ts
@@ -256,6 +256,8 @@ export type PerseusPlotterUserInput = number[];
 export type PerseusRadioRubric = {
     // The choices provided to the user.
     choices: PerseusRadioChoice[];
+    // If true, the number of choices selected must match the number of correct choices.
+    countChoices?: boolean;
 };
 
 export type PerseusRadioUserInput = {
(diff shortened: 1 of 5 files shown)


packages/perseus-core/src/validation.types.ts
```diff
@@ -256,6 +256,8 @@ export type PerseusPlotterUserInput = number[];
 export type PerseusRadioRubric = {
     // The choices provided to the user.
     choices: PerseusRadioChoice[];
+    // If true, the number of choices selected must match the number of correct choices.
+    countChoices?: boolean;
```
ivyolamit, line 260: @SonicScrewdriver in which scenarios do we set this to true and false?
the check below indicates `numCorrect > 1 && numSelected !== numCorrect && rubric.countChoices`
from what I understand:
- numCorrect > 1, indicates this will be for a multi select
- numSelected !== numCorrect, check if selection matches to expected correct items
- rubric.countChoices, looks like an explicit boolean to indicate to count the choices, how does this differ from previous check?

ivyolamit (approved):

PR #2531, opened 2025-05-28
Fix getSaveWarnings in Free Response widget

aag:
## Summary:
This commit changes the getSaveWarnings() function definition in the
Free Response widget editor to be defined using public class field syntax of
the public method syntax. Previously, `this` wasn't being bound
properly, so the function was unable to read `this.props`. Additionally,
the Free Response widget is unhidden, so that it will appear in the widgets
dropdown in the exercise editor.

The commit also adds some more type annotations and changes all methods
on the widget editor to be defined in the same way.

Issue: https://khanacademy.atlassian.net/browse/AX-1005

## Test plan:
- Load the PR's Perseus bundle in local webapp.
- Go to the exercise editor in devadmin and verify that there is a "Free Response"
  option in the "Add a widget..." dropdown.
image

- Add a Free Response widget in the exercise editor and verify that it
  renders properly.
image

### DIFF
--- a/.changeset/dirty-singers-arrive.md
+++ b/.changeset/dirty-singers-arrive.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/perseus": minor
+"@khanacademy/perseus-editor": minor
+---
+
+Fix getSaveWarnings in Free Response widget and unhide it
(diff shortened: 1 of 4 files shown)


jeremywiebe (approved):
Sorry. This is old code that uses refs alot. I'm unclear why this was broken because we [call](https://github.com/Khan/perseus/blob/3389e06111b2c801fba6c34ed4e5bdc747ef1ed1/packages/perseus-editor/src/editor.tsx#L810) `getSaveWarnings` on the widget editor ref. Anyways, glad you figured it out.

PR #2532, opened 2025-05-28
Add isItemAccessible function and enhance types

jeremywiebe:
## Summary:

While integrating this into webapp, I noticed that the `violatingWidgets()` function is very poorly typed. This PR adds better typing so it'll be easier to use correctly.

Issue: LEMS-3053

## Test plan:

`yarn tsc`

### DIFF
--- a/packages/perseus-core/src/accessibility.test.ts
+++ b/packages/perseus-core/src/accessibility.test.ts
@@ -1,0 +1,184 @@
+import {isItemAccessible} from "./accessibility";
+import {registerCoreWidgets} from "./widgets/core-widget-registry";
+
+import type {PerseusItem} from "./data-schema";
+
+describe("isItemAccessible", () => {
+    beforeEach(() => {
+        registerCoreWidgets();
+    });
+
+    it("should return false if the item contains any inaccessible widgets", () => {
+        const itemData: PerseusItem = {
+            question: {
+                content:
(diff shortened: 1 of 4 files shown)


packages/perseus-core/src/accessibility.ts
```diff
@@ -8,27 +8,23 @@ import _ from "underscore";
...
 
-// Iterate over a single Perseus renderer, mutating `widgets` by appending
-// violating widget types discovered in this item.
-function traverseRenderer(itemData, widgets: Array<any>) {
-    traverse(itemData, null, function (info) {
-        if (info.type && !Widgets.isAccessible(info.type, info.options)) {
-            widgets.push(info.type);
-        }
-    });
-}
```
jeremywiebe, line 19: I inlined this function because:

* it didn't have a very good/descriptive name
* it had only one usage which actually felt clearer when inlined

packages/perseus-core/src/accessibility.test.ts
catandthemachines, line 1: I'm guessing this is the test file that got removed and we're adding back? ❤️

catandthemachines (approved):
Love this improvement. Thanks Jeremy!

packages/perseus-core/src/accessibility.test.ts
jeremywiebe: Oh! I forgot it was removed.  I actually just created this using AI.

tatianasnook: 😄Haha yes, I was planning to move that test back myself—so I appreciate you (and the AI!) saving me the trouble.

PR #2535, opened 2025-05-29
Make Sorter story a better demonstration of answer removal

benchristel:
## Summary:
Tamara noticed that the answerless story for Sorter had exactly the same data
as the answerful one. This was due to the peculiar way we remove answer
information from Sorters by sorting all cards but the first. Sometimes this
answer-removal procedure leaves the cards in their original order. This is by
design -- if we required the cards to be permuted by answer removal, we would
actually be leaking information about the answer!

This PR changes the story so it's a better illustration of how the sorter
options get reordered in the answerless case.

Issue: none

## Test plan:

```
pnpm storybook
```

View the stories for Sorter. Confirm that the JSON data differs between the
answerful and answerless stories.


Myranae (approved):
Thanks! This does make it more clear that there is something different between the answerless (right) and the answerful (left) data. 
image

PR #2540, opened 2025-05-30
Remove showRationalesForCurrentlySelectedChoices and replace with showSolutions where appropriate

SonicScrewdriver:
## History:
This PR started as part of the Radio Widget Project, and became a more holistic change to simplify Perseus. 

After much discussion across several teams, we've decided that we will now show all feedback/rationales/correctness information all at once, but only after a widget becomes "inactive" through one of these actions:
- User skips the question
- User answers correctly
- User requests hints (effectively skipping)
- Content Creator enables static mode

Previously, we were selectively showing rationals to users while they were still able to interact with the widgets — a feature which was causing several bugs across the widgets that supported it: 

- Radio Widget 
- Graded Group Widget 
- Label Image Widget 

This move allows us to:

- Solve numerous bugs/tickets in one fell swoop 
  - LEMS-3076, [LEMS-1243](https://khanacademy.atlassian.net/browse/LEMS-1243), [LEMS-279](https://khanacademy.atlassian.net/browse/LEMS-279), [LEMS-1222](https://khanacademy.atlassian.net/browse/LEMS-1222), [LEMS-256](https://khanacademy.atlassian.net/browse/LEMS-256), [LEMS-1987](https://khanacademy.atlassian.net/browse/LEMS-1987), [LEMS-3075](https://khanacademy.atlassian.net/browse/LEMS-3075)
- Provide a clearer and more consistent to our users across our Perseus widgets 
- Better align with upcoming SSS/Future Initiative goals 

## What's changed:

This PR accomplishes the following tasks:

1. Removes `showRationalesForCurrentlySelectedChoices` across Perseus, fully deprecating the feature. 
2. Fixes a bug in the Label Image Widget where the widget was still interactable when in "reviewMode".  Now the widget will correctly become non-interactive while auto-selecting the correct answers to provide the same holistic feedback as the other Perseus Widgets. 
3. Added `showSolutions` to Graded Group widgets so that the rationales can show when the user gets the correct answer. 
4. Adds several tests to help verify changes 

## Upstream requirements: 
A draft PR has been created for our platform to remove all calls to `showRationalesForCurrentlySelectedChoices` from the main repo as well. 

While the mobile repository does contain some old references to this feature, the entire parent folder (webview) is now unused and will be removed in a future release. 

Issue: LEMS-3077

## Test plan:
- Manual testing 
- Full QA Testing Review (Upcoming), which will include both the Perseus and upstream changes 
- New Tests

[LEMS-1243]: https://khanacademy.atlassian.net/browse/LEMS-1243?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LEMS-279]: https://khanacademy.atlassian.net/browse/LEMS-279?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LEMS-1222]: https://khanacademy.atlassian.net/browse/LEMS-1222?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LEMS-256]: https://khanacademy.atlassian.net/browse/LEMS-256?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/perseus/src/renderer.tsx
+++ b/packages/perseus/src/renderer.tsx
@@ -741,38 +741,6 @@ class Renderer
     };
 
     /**
-     * Tell each of the radio widgets to show rationales for each of the
-     * currently selected choices inside of them. If the widget is correct, it
-     * shows rationales for all of the choices. This also disables interaction
-     * with the choices that we show rationales for.
-     */
-    showRationalesForCurrentlySelectedChoices: () => void = () => {
-        Object.keys(this.props.widgets).forEach((widgetId) => {
-            const widget = this.getWidgetInstance(widgetId);
-            if (widget && widget.showRationalesForCurrentlySelectedChoices) {
-                widget.showRationalesForCurrentlySelectedChoices(
-                    this._getWidgetInfo(widgetId).options,
-                );
-            }
-        });
-    };
-
-    /**
-     * Tells each of the radio widgets to deselect any of the incorrect choices
-     * that are currently selected (leaving correct choices still selected).
-     */
-    deselectIncorrectSelectedChoices: () => void = () => {
-        // TODO(emily): this has the exact same structure as
-        // showRationalesForCurrentlySelectedChoices above. Maybe DRY this up.
-        Object.keys(this.props.widgets).forEach((widgetId) => {
-            const widget = this.getWidgetInstance(widgetId);
-            if (widget && widget.deselectIncorrectSelectedChoices) {
-                widget.deselectIncorrectSelectedChoices();
-            }
-        });
-    };
-
-    /**
      * Allows inter-widget communication.
      *
      * This function yields this Renderer's own internal widgets, and it's used
--- a/packages/perseus/src/types.ts
+++ b/packages/perseus/src/types.ts
@@ -27,6 +27,7 @@ import type {
     Relationship,
     LabelImageMarkerPublicData,
     PerseusLabelImageMarker,
+    ShowSolutions,
 } from "@khanacademy/perseus-core";
 import type {LinterContextProps} from "@khanacademy/perseus-linter";
 import type {Result} from "@khanacademy/wonder-blocks-data";
@@ -73,7 +74,6 @@ export interface Widget {
           }
         | boolean;
     getDOMNodeForPath?: (path: FocusPath) => Element | Text | null;
-    deselectIncorrectSelectedChoices?: () => void;
 
     /**
      * Returns widget state that can be passed back to `restoreSerializedState`
@@ -580,6 +580,7 @@ export type UniversalWidgetProps<
     onBlur: (blurPath: FocusPath) => void;
     findWidgets: (criterion: FilterCriterion) => ReadonlyArray<Widget>;
     reviewMode: boolean;
+    showSolutions?: ShowSolutions;
     onChange: ChangeHandler;
     isLastUsedWidget: boolean;
     // provided by widget-container.jsx#render()
--- a/packages/perseus/src/types.ts
+++ b/packages/perseus/src/types.ts
@@ -99,7 +99,6 @@ export interface Widget {
     getUserInputMap?: () => UserInputMap | undefined;
     getUserInput?: () => UserInputArray | UserInput | undefined;
 
-    showRationalesForCurrentlySelectedChoices?: (options?: any) => void;
     getPromptJSON?: () => WidgetPromptJSON;
 }
 
--- a/packages/perseus/src/widgets/graded-group-set/graded-group-set.stories.tsx
+++ b/packages/perseus/src/widgets/graded-group-set/graded-group-set.stories.tsx
@@ -1,6 +1,9 @@
 import {ArticleRendererWithDebugUI} from "../../../../../testing/article-renderer-with-debug-ui";
 
-import {article1, groupSetRadioQuestion} from "./graded-group-set.testdata";
+import {
+    article1,
+    groupSetRadioRationaleQuestion,
+} from "./graded-group-set.testdata";
 
 import type {Meta, StoryObj} from "@storybook/react-vite";
 
@@ -20,6 +23,6 @@ export const Article1: Story = {
 
 export const GroupSetRadioQuestion: Story = {
     args: {
-        json: groupSetRadioQuestion,
+        json: groupSetRadioRationaleQuestion,
     },
 };
--- a/packages/perseus/src/widgets/graded-group-set/graded-group-set.stories.tsx
+++ b/packages/perseus/src/widgets/graded-group-set/graded-group-set.stories.tsx
@@ -1,6 +1,6 @@
 import {ArticleRendererWithDebugUI} from "../../../../../testing/article-renderer-with-debug-ui";
 
-import {article1} from "./graded-group-set.testdata";
+import {article1, groupSetRadioQuestion} from "./graded-group-set.testdata";
 
 import type {Meta, StoryObj} from "@storybook/react-vite";
 
@@ -15,5 +15,11 @@ type Story = StoryObj<typeof ArticleRendererWithDebugUI>;
 export const Article1: Story = {
     args: {
         json: article1,
+    },
+};
+
+export const GroupSetRadioQuestion: Story = {
+    args: {
+        json: groupSetRadioQuestion,
     },
 };
--- a/packages/perseus/src/widgets/graded-group-set/graded-group-set.test.ts
+++ b/packages/perseus/src/widgets/graded-group-set/graded-group-set.test.ts
@@ -5,7 +5,10 @@ import {testDependencies} from "../../../../../testing/test-dependencies";
 import * as Dependencies from "../../dependencies";
 import {renderQuestion} from "../__testutils__/renderQuestion";
 
-import {article1, groupSetRadioQuestion} from "./graded-group-set.testdata";
+import {
+    article1,
+    groupSetRadioRationaleQuestion,
+} from "./graded-group-set.testdata";
 
 import type {PerseusRenderer} from "@khanacademy/perseus-core";
 import type {UserEvent} from "@testing-library/user-event";
@@ -315,7 +318,7 @@ describe("graded group widget", () => {
 
     it("should show rationales when answer is correct", async () => {
         // Arrange
-        renderQuestion(groupSetRadioQuestion);
+        renderQuestion(groupSetRadioRationaleQuestion);
 
         // Select the correct answer: "$8$" (index 2)
         await userEvent.click(screen.getAllByRole("radio")[2]);
@@ -331,7 +334,7 @@ describe("graded group widget", () => {
 
     it("should not show rationales when answer is incorrect", async () => {
         // Arrange
-        renderQuestion(groupSetRadioQuestion);
+        renderQuestion(groupSetRadioRationaleQuestion);
 
         // Select an incorrect answer: "$-8$" (index 1)
         await userEvent.click(screen.getAllByRole("radio")[1]);
--- a/packages/perseus/src/widgets/graded-group-set/graded-group-set.test.ts
+++ b/packages/perseus/src/widgets/graded-group-set/graded-group-set.test.ts
@@ -5,7 +5,7 @@ import {testDependencies} from "../../../../../testing/test-dependencies";
 import * as Dependencies from "../../dependencies";
 import {renderQuestion} from "../__testutils__/renderQuestion";
 
-import {article1} from "./graded-group-set.testdata";
+import {article1, groupSetRadioQuestion} from "./graded-group-set.testdata";
 
 import type {PerseusRenderer} from "@khanacademy/perseus-core";
 import type {UserEvent} from "@testing-library/user-event";
@@ -310,6 +310,40 @@ describe("graded group widget", () => {
         expect(screen.getByRole("button", {name: "Explain"})).toBeVisible();
         expect(
             screen.queryByRole("button", {name: "Hide explanation"}),
+        ).not.toBeInTheDocument();
+    });
+
+    it("should show rationales when answer is correct", async () => {
+        // Arrange
+        renderQuestion(groupSetRadioQuestion);
+
+        // Select the correct answer: "$8$" (index 2)
+        await userEvent.click(screen.getAllByRole("radio")[2]);
+
+        // Act
+        await userEvent.click(screen.getByRole("button", {name: "Check"}));
+
+        // Assert
+        expect(screen.getByRole("alert", {name: "Correct"})).toBeVisible();
+        // Verify the rationale for the correct answer is shown
+        expect(screen.getByText("This is the correct answer.")).toBeVisible();
+    });
+
+    it("should not show rationales when answer is incorrect", async () => {
+        // Arrange
+        renderQuestion(groupSetRadioQuestion);
+
+        // Select an incorrect answer: "$-8$" (index 1)
+        await userEvent.click(screen.getAllByRole("radio")[1]);
+
+        // Act
+        await userEvent.click(screen.getByRole("button", {name: "Check"}));
+
+        // Assert
+        expect(screen.getByRole("alert", {name: "Incorrect"})).toBeVisible();
+        // Verify that rationales are not shown
+        expect(
+            screen.queryByText("This is not the correct answer."),
         ).not.toBeInTheDocument();
     });
 });
--- a/packages/perseus/src/widgets/graded-group-set/graded-group-set.testdata.ts
+++ b/packages/perseus/src/widgets/graded-group-set/graded-group-set.testdata.ts
@@ -388,25 +388,25 @@ export const groupSetRadioQuestion: PerseusRenderer = {
                                         {
                                             content: "$-8$ and $8$",
                                             correct: false,
-                                            clue: "The square root operation ($\\sqrt{\\phantom{x}}$) calculates *only* the positive square root when performed on a number, so $x$ is equal to *only* $8$.",
+                                            clue: "This is not the correct answer.",
                                         },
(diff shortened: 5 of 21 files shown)


packages/perseus/src/renderer.tsx
```diff
@@ -713,38 +713,6 @@ class Renderer
...
-                );
-            }
-        });
-    };
-
-    /**
-     * Tells each of the radio widgets to deselect any of the incorrect choices
-     * that are currently selected (leaving correct choices still selected).
-     */
-    deselectIncorrectSelectedChoices: () => void = () => {
```
SonicScrewdriver, line 764: This method was specifically related to showRationalesForCurrentlySelectedChoices, and was only ever supported by the Radio Widget. I don't believe it's necessary to keep it around any longer!

packages/perseus/src/widgets/graded-group-set/graded-group-set.stories.tsx
```diff
@@ -17,3 +17,9 @@ export const Article1: Story = {
         json: article1,
     },
 };
+
+export const GroupSetRadioQuestion: Story = {
```
SonicScrewdriver, line 24: I wanted a story that would help us test rationales with a Graded Group

packages/perseus/src/widgets/graded-group/graded-group.tsx
```diff
@@ -304,6 +303,9 @@ export class GradedGroup
         const readOnly =
             apiOptions.readOnly || (apiOptions.isMobile && isCorrect);
 
+        // We only want to show the solutions and rationale if the answer is correct
```
SonicScrewdriver, line 306: There's no skip state for this widget, so we only show rationales after the question is answered correctly.

packages/perseus/src/renderer.tsx
SonicScrewdriver: This was also what was giving undue hints ;)

testing/server-item-renderer-with-debug-ui.tsx
```diff
@@ -158,7 +158,6 @@ export const ServerItemRendererWithDebugUI = ({
                             <Button
                                 onClick={() => {
                                     setAnswerless(false);
-                                    ref.current?.showRationalesForCurrentlySelectedChoices();
                                 }}
```
SonicScrewdriver: There's another PR that will improve this component to fully support the showSolutions approach. (https://github.com/Khan/perseus/pull/2542) I will fix any merge conflicts depending on which one lands first.

packages/perseus/src/renderer.tsx
ivyolamit: do we need to do the webapp update for this first? I saw this usages
- [exercise chrome](https://github.com/Khan/webapp/blob/747bb1ed11718a466bfc7c56de624011492dcc00/services/static/apps/khanacademy/src/content-library/ui/content-renderers/exercise/exercise-chrome.tsx#L511)
- [stateful-exercise manager](https://github.com/Khan/webapp/blob/747bb1ed11718a466bfc7c56de624011492dcc00/services/static/apps/khanacademy/src/exercises-package/stateful-exercise-manager.tsx#L417)

packages/perseus/src/widgets/graded-group-set/graded-group-set.test.ts
```diff
@@ -312,4 +312,38 @@ describe("graded group widget", () => {
             screen.queryByRole("button", {name: "Hide explanation"}),
         ).not.toBeInTheDocument();
     });
+
+    it("should show rationales when answer is correct", async () => {
```
jeremywiebe, line 319: We can go with this for now, but I feel this is a big regression in usefulness of the graded-group (and graded-group-set) widget.  These are used mostly in articles (and I think in hints).  So only showing rationales once the learner has gotten the correct answer feels like it doesn't help learning. 

This probably needs separate design work, but I'd think both of these group widgets need some affordance for the learner to "skip" and see the rationales.

packages/perseus/src/types.ts
```diff
@@ -100,7 +100,6 @@ export interface Widget {
     getUserInputMap?: () => UserInputMap | undefined;
     getUserInput?: () => UserInputArray | UserInput | undefined;
 
-    showRationalesForCurrentlySelectedChoices?: (options?: any) => void;
```
jeremywiebe, line 102: There's a `deselectIncorrectSelectedChoices` in this type also (line 77). Does that need to go away also?

packages/perseus/src/widgets/graded-group/graded-group.tsx
```diff
@@ -392,6 +395,7 @@ export class GradedGroup
                                 apiOptions={apiOptions}
                                 linterContext={this.props.linterContext}
                                 strings={this.context.strings}
+                                showSolutions={showSolutions}
```
jeremywiebe, line 398: Does `showSolutions` this apply to the hint? Or rather, do we see rationales showing up in hint content for graded-groups?

packages/perseus/src/widgets/group/group.tsx
```diff
@@ -28,12 +28,15 @@ import type {
...
     PerseusGroupWidgetOptions,
+    ShowSolutions,
     UserInputArray,
     UserInputMap,
 } from "@khanacademy/perseus-core";
 
 type RenderProps = PerseusGroupWidgetOptions; // exports has no 'transform'
-type Props = WidgetProps<RenderProps>;
+type Props = WidgetProps<RenderProps> & {
+    showSolutions?: ShowSolutions; // Provided by the renderer
```
jeremywiebe: Instead of selectively adding `showSolutions` to the widgets that use it, I think it'd be better to add it to `UniversalWidgetProps`. That's what is happening in reality (the Renderer [passes](https://github.com/Khan/perseus/blob/4ac22bdadcdad844ed9964485ef727de53c6dfa7/packages/perseus/src/renderer.tsx#L596) it to each and every widget it renders).

packages/perseus/src/types.ts
```diff
@@ -100,7 +100,6 @@ export interface Widget {
     getUserInputMap?: () => UserInputMap | undefined;
```
ivyolamit, line 99: Is there a reason why the type `deselectIncorrectSelectedChoices` in line 77 was not also removed?

packages/perseus/src/widgets/graded-group/graded-group.tsx
ivyolamit: is there a place we can document this properly? since this is the new expectation and the previous method was already removed.

packages/perseus/src/widgets/label-image/label-image.tsx
```diff
@@ -82,6 +83,7 @@ type Props = WidgetProps<Options> &
         // preferred placement for popover (preference, not MUST)
         // TODO: this is sus, probably never passed in
         preferredPopoverDirection?: PreferredPopoverDirection;
+        showSolutions?: ShowSolutions;
```
ivyolamit: This is a nitpick, just to be consistent with other usage
```suggestion
        showSolutions?: ShowSolutions; // Provided by the renderer
```

packages/perseus/src/widgets/radio/radio-component.tsx
```diff
@@ -256,58 +245,6 @@ class Radio extends React.Component<Props> implements Widget {
...
-                // @ts-expect-error - TS2345 - Argument of type 'null' is not assignable to parameter of type '(() => unknown) | undefined'.
-                null, // cb
-                true, // silent
-            );
-        }
-    };
-
     /**
      * Deselects any currently-selected choices that are not correct choices.
      */
```
ivyolamit: From the comment [above](https://github.com/Khan/perseus/pull/2540/files#r2121570728) are we also removing this?

packages/perseus/src/widgets/label-image/label-image.tsx
```diff
@@ -466,11 +434,43 @@ export class LabelImage
...
+     * selections for markers without answers.
+     *
+     * @param marker - The marker to update.
+     * @returns The updated marker state.
+     */
+    getUpdatedMarkerState(
+        marker: OptionalAnswersMarkerType,
+    ): OptionalAnswersMarkerType {
+        const shouldShowFeedback =
+            this.props.showSolutions === "all" || this.props.reviewMode;
```
jeremywiebe: Given such minimal use of `this` in this function, what do you think of making it static, or even moving it out of this component completely?  I think a pure function might make this function even clearer, plus easy to test.

packages/perseus/src/widgets/radio/radio.class.new.tsx
```diff
@@ -120,58 +118,6 @@ class Radio extends React.Component<Props> implements Widget {
         return _getPromptJSON(this.props, userInput);
     }
 
-    /**
-     * Turn on rationale display for the currently selected choices. Note that
-     * this leaves rationales on for choices that are already showing
-     * rationales.
-     * TODO: LEMS-3077 Remove deprecated class function
-     * @deprecated Internal only. Use `showSolutions` prop instead.
```
jeremywiebe, line 128: Woot, the day has come!!! Thanks!

packages/perseus/src/widgets/radio/radio.class.new.tsx
```diff
@@ -120,58 +118,6 @@ class Radio extends React.Component<Props> implements Widget {
...
-     */
-    showRationalesForCurrentlySelectedChoices: (
-        arg1: PerseusRadioWidgetOptions,
-    ) => void = (rubric) => {
-        const {choiceStates} = this.props;
-        if (choiceStates) {
-            const score = scoreRadio(this.getUserInput(), rubric);
```
ivyolamit, line 135: from what i remember the `scoreRadio` was only being used in this method, do we also want to cleanup the scoreRadio?

ivyolamit (commented):
@SonicScrewdriver Thanks for taking care of this, and cleaning it up. I just have a few questions in the PR.

packages/perseus/src/renderer.tsx
jeremywiebe: I believe there's still instances of this function leftover in this PR. Could you do a search through the code to make sure you've deleted them all?  :) 

```sh
$ git rev-parse --abbrev-ref HEAD
non-selective-rationality
$ rg deselectIncorrectSelectedChoices
packages/perseus/src/types.ts
77:    deselectIncorrectSelectedChoices?: () => void;

packages/perseus/src/widgets/radio/radio-component.tsx
251:    deselectIncorrectSelectedChoices: () => void = () => {

packages/perseus/src/widgets/radio/radio.class.new.tsx
124:    deselectIncorrectSelectedChoices: () => void = () => {
```

jeremywiebe (approved):
This is excellent. Thanks for cleaning up this crufty old API. I noted a few things that I think would be good to address before landing, but generally, this looks great!

packages/perseus/src/widgets/graded-group-set/graded-group-set.stories.tsx
handeyeco: If rationales are a motivation for the Story, would it be beneficial to include "rationales" in the name?

packages/perseus/src/widgets/graded-group-set/graded-group-set.testdata.ts
```diff
@@ -354,3 +354,216 @@ export const article1: PerseusRenderer = {
...
+                                    minor: 0,
+                                },
+                                static: false,
+                                type: "radio",
+                                options: {
+                                    displayCount: null,
+                                    onePerLine: false,
+                                    choices: [
+                                        {
+                                            content: "$-8$ and $8$",
```
handeyeco: I would like to advocate for Stories that use simpler content/data. It's kind of a PITA to test on widgets that require us to actually do math. I would love if the norm was:

1. Content explains the motivation for the test/story
2. Correct / incorrect answers were extremely obvious
3. If it's not possible to make the answers obvious (Radio and Dropdown are easy, IG and Expression are hard), then include the answer in the content.

Like this one test Renderer is over a hundred lines of code, but it's not clear why it exists or which parts of it are important. If we could make the data very explicit and only include the stuff required (Like does this radio need 4 options? Are there factory helpers we could use to make some of this data?), it makes things easier to maintain.

packages/perseus/src/widgets/radio/radio.class.new.tsx
handeyeco: `scoreRadio` is used by the scoring logic

packages/perseus/src/widgets/label-image/label-image.tsx
```diff
@@ -510,25 +510,51 @@ export class LabelImage
...
+                score = scoreLabelImageMarker(
+                    updatedMarkerState.selected,
+                    updatedMarkerState.answers,
+                );
             } else {
                 score = {
                     hasAnswers: false,
                     isCorrect: false,
                 };
+                // For markers without answers, mark as correct when showing feedback
```
handeyeco: This seems risky from a SSS perspective. We show answerless markers as correct? Because before we fetch answerful data, all markers will be answerless. I don't think it's a huge issue but it seems like a potential source of bugs.

packages/perseus/src/widgets/group/group.tsx
SonicScrewdriver: Perfect! I wasn't sure if I should add it to UniversalWidgetProps as it's not implemented in most of the widgets yet, but it's definitely a much cleaner solution. Thank you!

packages/perseus/src/widgets/graded-group-set/graded-group-set.stories.tsx
SonicScrewdriver: Makes sense to me!

packages/perseus/src/widgets/label-image/label-image.tsx
SonicScrewdriver: I'm only marking this as a no as I'm moving the property declaration to the `UniversalWidgetOptions`

packages/perseus/src/widgets/graded-group/graded-group.tsx
SonicScrewdriver: You know, that's a great question! 

I added it both places, but didn't consider it fully. I THINK it makes sense to keep it, as there could _possibly_ be a case where Content Creators want to show the individual rationales on a widget in a graded group in a hint ... but it does seem like a bit of a silly situation. Ideally they would use static mode, in that case. 

TBH, now that I think about it, I don't even know why we support graded groups in hints at all for that matter. I might try taking a look to see how many examples I can actually find about how we're using this. 

I would be okay with removing this, but I'm curious about your thoughts!

packages/perseus/src/widgets/graded-group-set/graded-group-set.testdata.ts
SonicScrewdriver: Some _**great**_ points! I've updated the content to be very obvious about the correct answers, rather than using math. I had copied over the radio widget from their stories, as I was being lazy, but we don't need to be verifying that so many sub features work over in the Graded Group. Thanks for the callout! 

Let me see what I can do about finding a pre-existing factory helper. If there isn't already one for the Radio widget, I will make us a ticket to make one as part of our Radio Widget project.

packages/perseus/src/widgets/label-image/label-image.tsx
SonicScrewdriver: Ohhhhh. I think I completely misunderstood here and thought the hasAnswers was related to possibly having markers on an image that content creators wanted left blank. I've removed this line entirely after further testing!

packages/perseus/src/widgets/graded-group-set/graded-group-set.testdata.ts
SonicScrewdriver: I really didn't see much by the way of helpers that could make the json shorter. If you know of any, please let me know! I'll make a ticket for getting a radio question builder.

packages/perseus/src/widgets/graded-group-set/graded-group-set.test.ts
SonicScrewdriver: I definitely agree! The user is still told if they're correct/incorrect with the graded-group's correctness indicator, but I also felt that we should add a skip button to the widget to allow students to quickly see the full rationales. 

It wouldn't be hard to add, but I figured this PR is already casting a pretty wide net. :) I can make a ticket for us to revisit the topic as a potential follow up

packages/perseus/src/types.ts
SonicScrewdriver: Yes! Thank you, I think I've cleared them all out now.

ivyolamit (approved):
@SonicScrewdriver sorry for the delay, changes here looks logical to me. 👍🏼 Maybe we can also ask @handeyeco if he is available for final check since it will have an impact with their work 🙏🏼

packages/perseus/src/widgets/graded-group-set/graded-group-set.testdata.ts
SonicScrewdriver: Just for history sake: I made a ticket for building a Radio Question Builder, which will be worked on this sprint!

PR #2541, opened 2025-05-30
Add typesafe parser for Grapher's user input type

Myranae:
## Summary:
Add a user input type parser and typetest for the Grapher widget

Issue: LEMS-3139

## Test plan:
- Confirm all checks pass
- Confirm widget still works as expected via Storybook


packages/perseus-core/src/parse-perseus-json/perseus-parsers/grapher-user-input.ts
```diff
@@ -0,0 +1,56 @@
...
+    pair,
+    union,
+} from "../general-purpose-parsers";
+
+const coord = pair(number, number);
+const coordPair = pair(coord, coord);
+
+export const parseGrapherUserInput = union(
+    object({
+        type: enumeration("absolute_value"),
```
benchristel: `enumeration` will work here, but in other places we use `constant` when there's only one allowed value.

```suggestion
        type: constant("absolute_value"),
```

packages/perseus-core/src/parse-perseus-json/perseus-parsers/grapher-user-input.ts
```diff
@@ -0,0 +1,56 @@
...
+    number,
+    object,
+    pair,
+    union,
+} from "../general-purpose-parsers";
+
+const coord = pair(number, number);
+const coordPair = pair(coord, coord);
+
+export const parseGrapherUserInput = union(
```
benchristel: Since the branches of this union are distinguished by one common field (`type`), we can use `discriminatedUnionOn("type")` instead of `union`. The advantage of `discriminatedUnionOn` is that it produces better error messages when parsing fails. See [the interaction widget parser](https://github.com/Khan/perseus/blob/558cc1cc6ad468ba6538648c57d47df5704e6858/packages/perseus-core/src/parse-perseus-json/perseus-parsers/interaction-widget.ts#L160-L168) for an example.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/grapher-user-input.ts
Myranae: Updated!

packages/perseus-core/src/parse-perseus-json/perseus-parsers/grapher-user-input.ts
Myranae: Updated!

PR #2543, opened 2025-05-30
Proposal for an improved EditorPageWithStorybookPreview to better mock actual implementation that uses iFrames

SonicScrewdriver:
## Summary:
[neverending-story-2] New Editor Page that properly implements iframe communication

NOTE: THIS IS A DRAFT PR and still needs polish / cleanup.

### DIFF
--- a/packages/perseus-editor/src/__stories__/editor-page-with-storybook-preview.tsx
+++ b/packages/perseus-editor/src/__stories__/editor-page-with-storybook-preview.tsx
@@ -1,40 +1,37 @@
-import {Renderer, type APIOptions, type DeviceType} from "@khanacademy/perseus";
+import {type APIOptions, type DeviceType} from "@khanacademy/perseus";
 import {
     type Hint,
     type PerseusAnswerArea,
     type PerseusRenderer,
 } from "@khanacademy/perseus-core";
-import Button from "@khanacademy/wonder-blocks-button";
 import {View} from "@khanacademy/wonder-blocks-core";
-import IconButton from "@khanacademy/wonder-blocks-icon-button";
-import {Strut} from "@khanacademy/wonder-blocks-layout";
-import {color, spacing} from "@khanacademy/wonder-blocks-tokens";
-import {LabelLarge} from "@khanacademy/wonder-blocks-typography";
-import xIcon from "@phosphor-icons/core/regular/x.svg";
-import {StyleSheet} from "aphrodite";
 import * as React from "react";
 import {action} from "storybook/actions";
 
-// eslint-disable-next-line import/no-relative-packages
-import {mockStrings} from "../../../perseus/src/strings";
-import ContentPreview from "../content-preview";
 import EditorPage from "../editor-page";
 
 type Props = {
     apiOptions?: APIOptions;
     question?: PerseusRenderer;
     hints?: ReadonlyArray<Hint>;
+    answerArea?: PerseusAnswerArea;
 };
+
+declare global {
+    interface Window {
+        iframeDataStore: Record<string, any>;
+    }
+}
 
 const onChangeAction = action("onChange");
 
 function EditorPageWithStorybookPreview(props: Props) {
     const [previewDevice, setPreviewDevice] =
-        React.useState<DeviceType>("phone");
+        React.useState<DeviceType>("desktop");
     const [jsonMode, setJsonMode] = React.useState<boolean | undefined>(false);
     const [answerArea, setAnswerArea] = React.useState<
         PerseusAnswerArea | undefined | null
-    >();
+    >(props.answerArea);
     const [question, setQuestion] = React.useState<PerseusRenderer | undefined>(
         props.question,
     );
@@ -42,11 +39,24 @@ function EditorPageWithStorybookPreview(props: Props) {
         props.hints,
     );
 
-    const [panelOpen, setPanelOpen] = React.useState<boolean>(true);
+    // Memoize apiOptions to fix dependency warnings
+    const apiOptions = React.useMemo(() => {
+        const isMobile =
+            previewDevice === "phone" || previewDevice === "tablet";
+        return {
+            ...props.apiOptions,
+            isMobile,
+            customKeypad: isMobile, // webapp pattern: customKeypad matches isMobile
+        };
+    }, [props.apiOptions, previewDevice]);
 
-    const apiOptions = props.apiOptions ?? {
-        isMobile: false,
-    };
+    // Set up iframe data store for Perseus frame communication
+    React.useEffect(() => {
+        // Check if iframeDataStore exists before setting it up
+        if (typeof window !== "undefined" && !(window as any).iframeDataStore) {
+            (window as any).iframeDataStore = {};
+        }
+    }, []);
 
     return (
         <View>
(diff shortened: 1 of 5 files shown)


packages/perseus-editor/src/__stories__/editor-page-with-storybook-preview.tsx
```diff
@@ -1,51 +1,57 @@
...
     );
     const [hints, setHints] = React.useState<ReadonlyArray<Hint> | undefined>(
         props.hints,
     );
 
-    const [panelOpen, setPanelOpen] = React.useState<boolean>(true);
-
-    const apiOptions = props.apiOptions ?? {
-        isMobile: false,
+    const isMobile = previewDevice === "phone" || previewDevice === "tablet";
```
ivyolamit, line 50: 🎉

SonicScrewdriver:
There's a failing linter here, but I want to confirm with the team about whether and/or where we should be declaring the globals before fixing it!

packages/perseus-editor/src/__stories__/editor-page-with-storybook-preview.tsx
```diff
@@ -1,51 +1,57 @@
...
     question?: PerseusRenderer;
     hints?: ReadonlyArray<Hint>;
+    answerArea?: PerseusAnswerArea;
 };
 
+declare global {
+    interface Window {
+        iframeDataStore: Record<string, any>;
+    }
+}
```
SonicScrewdriver, line 24: Temporary location while I determine the right location for this, if there is one.

jeremywiebe: I think the right spot is to get rid of this as a way to send data between the host page and the iframe. Ideally, this should pass the data along with the `postMessage()` call. Not asking you to do that now, but mid-term, that'd be how we fix this.

packages/perseus-editor/src/__stories__/editor-page-with-storybook-preview.tsx
```diff
@@ -61,9 +67,10 @@ function EditorPageWithStorybookPreview(props: Props) {
                 answerArea={answerArea}
                 question={question}
                 hints={hints}
-                frameSource="about:blank"
-                previewURL="about:blank"
-                itemId="1"
+                // These urls lead to the PerseusFrameComponent story
+                frameSource="/iframe.html?id=perseuseditor-perseus-frame--frame&viewMode=story"
+                previewURL="/iframe.html?id=perseuseditor-perseus-frame--frame&viewMode=story"
```
jeremywiebe, line 72: Does this leverage a Storybook iframe feature?  I'd be mildly concerned that this could break if someone re-names a story.  We could also just put a container page in the [Vite static asset](https://storybook.js.org/docs/configure/integration/images-and-assets#serving-static-files-via-storybook-configuration) directory if you think that's more resilient to changes.

packages/perseus-editor/src/__stories__/editor-page.stories.tsx
```diff
@@ -1,15 +1,151 @@
...
+        ],
+    },
+    parameters: {
+        docs: {
+            description: {
+                story: "Editor with dropdown widget to test preview frame communication and height adjustments.",
+            },
+        },
+    },
 };
```
jeremywiebe, line 151: I think you can also get documentation for a story by putting a JSDoc comment on the story itself. It's a little more compact and, I think, more natural to find for future maintenance. 

```suggestion
/** 
 * Editor with dropdown widget to test preview frame communication and 
 * height adjustments.
 */
export const EditorWithDropdownWidget: Story = {
    name: "Editor with Dropdown Widget",
    args: {
        question: {
            content:
                "Which of the following is a prime number?\n\n[[☃ dropdown 1]]",
            images: {},
            widgets: {
                "dropdown 1": {
                    type: "dropdown",
                    graded: true,
                    options: {
                        choices: [
                            {
                                content: "Choose an answer",
                                correct: false,
                            },
                            {
                                content: "4",
                                correct: false,
                            },
                            {
                                content: "6",
                                correct: false,
                            },
                            {
                                content: "7",
                                correct: true,
                            },
                            {
                                content: "8",
                                correct: false,
                            },
                        ],
                        placeholder: "Choose an answer",
                        static: false,
                    },
                    version: {
                        major: 0,
                        minor: 0,
                    },
                },
            },
        },
        hints: [
            {
                content:
                    "A prime number is a number greater than 1 that has no positive divisors other than 1 and itself.",
                images: {},
                widgets: {},
            },
        ],
    },
};
```

packages/perseus-editor/src/__stories__/editor-page.stories.tsx
```diff
@@ -1,15 +1,151 @@
...
 
-export default {
+const meta: Meta<typeof EditorPageWithStorybookPreview> = {
     title: "PerseusEditor/EditorPage",
+    component: EditorPageWithStorybookPreview,
+    parameters: {
+        docs: {
+            description: {
+                component:
+                    "Perseus EditorPage with built-in preview functionality using Storybook iframe. ",
```
jeremywiebe, line 16: Where does this description show up when it's done on the `Meta` object?

packages/perseus-editor/src/__stories__/perseus-frame-component.tsx
```diff
@@ -0,0 +1,184 @@
...
+      };
+
+/**
+ * PerseusFrameComponent is a component that renders Perseus content in an iframe.
+ * It is used to render the content of the EditorPage in order to provide previews
+ * of the questions and hints.
+ *
+ * @returns A PerseusFrameComponent that renders the content of the EditorPage.
+ */
+const PerseusFrameComponent = () => {
```
jeremywiebe, line 41: This is fantastic that you reverse-engineered this info. Ideally, I'd love to extract the same component out of webapp and move it into Perseus (so that we'd have both sides of the communication channel in Perseus) and then eventually refactor to use the same component in the stories as in production. Future work, but this definitely paves the way! Thanks.

jeremywiebe (commented):
Love that you've started into this!  I have some comments but generally am very happy to see you pushing into cleaning this up and enabling it in Storybook!

packages/perseus-editor/src/__stories__/editor-page-with-storybook-preview.tsx
SonicScrewdriver: I didn't need to leverage much of storybook at all! I tried to avoid too much storybook middleware as it seemed to diverge a little from how we would set it up in production. 

This currently links to the Perseus Frame Component's story in "story" mode (AKA the full page view). I think the vite static asset directory could make a lot of sense, as it's definitely possible for this to break if someone moves the Perseus Frame component story. While I don't think it's likely that someone will rename the story, I think the static asset directory is a good way to keep us safe. :) I'll look into getting that working!

packages/perseus-editor/src/__stories__/editor-page.stories.tsx
SonicScrewdriver: Oh weird. These were showing up locally for me on a "docs" page added for the relevant folders, but they no longer do after rebasing the newest changes from main into this branch. Did we turn off the docs feature when upgrading Storybook?

You can see the docs on the storybook publish for this PR: 
https://650db21c3f5d1b2f13c02952-ejozbvanyy.chromatic.com/?path=/docs/perseuseditor-perseus-frame--docs
https://650db21c3f5d1b2f13c02952-ejozbvanyy.chromatic.com/?path=/docs/perseuseditor-editorpage--docs

Here's one we had from before:
https://650db21c3f5d1b2f13c02952-ejozbvanyy.chromatic.com/?path=/docs/perseus-components-number-input--docs

jeremywiebe: We didn't turn it off as far as I know, but I did upgrade us to Storybook v9 and I wonder if that changed things!  I'll look into that.  Fix: https://github.com/Khan/perseus/pull/2565

However, I still think it'd be better to use JSDoc comments as suggested.. they show up the same way and (IMHO at least) are much more natural than figuring out how to add the right combination of `docs` `parameters` etc. :) 

See: https://storybook.js.org/docs/api/doc-blocks/doc-block-description#writing-descriptions

PR #2544, opened 2025-05-30
docs(changeset): Move traversal tests from perseus-editor and perseus into perseus-core

tatianasnook:
## Summary:
This change restores deleted traversal tests by migrating them into the perseus-core, including general traversal tests and the traverseChildWidgets group widget test, to maintain coverage and centralize test logic.

Issue: LEMS-3146

## Test plan:

### DIFF
--- a/.changeset/green-pumas-know.md
+++ b/.changeset/green-pumas-know.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-core": major
+---
+
+Move traversal tests from perseus-editor and perseus into perseus-core
--- a/packages/perseus-core/src/widgets/core-widget-registry.test.ts
+++ b/packages/perseus-core/src/widgets/core-widget-registry.test.ts
@@ -1,18 +1,75 @@
 import * as CoreWidgetRegistry from "./core-widget-registry";
+import {registerWidget, traverseChildWidgets} from "./core-widget-registry";
+
+const registryFnNames = [
+    "isWidgetRegistered",
+    "getCurrentVersion",
+    "getPublicWidgetOptionsFunction",
+    "getWidgetOptionsUpgrades",
+    "getDefaultWidgetOptions",
+    "getSupportedAlignments",
+    "getDefaultAlignment",
+];
+
+const mockWidgetType = "_test-mock-widget_";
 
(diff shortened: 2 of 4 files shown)


packages/perseus-core/src/widgets/core-widget-registry.test.ts
```diff
@@ -1,18 +1,103 @@
...
+        (fnName) => {
+            const fn = (CoreWidgetRegistry as any)[fnName];
+            expect(() => fn("radio")).toThrow(
+                "Core widget registry accessed before initialization!",
+            );
+        },
+    );
+});
+
+describe("traverseChildWidgets", () => {
```
catandthemachines: Move this under the `describe("core-widget-registry",() => {})` because `traverseChildWidgets` is a function within the `core-widget-registry`.

.changeset/green-pumas-know.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/perseus-core": major
```
jeremywiebe: I don't think this needs to be a major.  We aren't making any backwards-incompatible changes.  I'd mark this as a **patch** give that we aren't changing any production behaviour nor exporting any new functions/types from the `perseus-core` package.

jeremywiebe (approved):

.changeset/green-pumas-know.md
jeremywiebe: I didn't think to add this till after I submitted my review, but we follow [SemVer](https://semver.org) (SEMatic VERsioning) for our version numbering. That helps decide if a given PR is a major, minor, or patch changed.

packages/perseus-core/src/widgets/core-widget-registry.test.ts
tatianasnook: Moved traverseChildWidgets test block inside the core-widget-registry describe as suggested. Thanks!

.changeset/green-pumas-know.md
tatianasnook: Updated the changeset to use a patch release instead of major — agreed, this doesn’t introduce any breaking changes or new exports. Thanks for the clarification!

PR #2545, opened 2025-06-02
Add recursive typesafe parser for Group's user input type

Myranae:
## Summary:
Add a user input type parser and typetest for the Group widget. Following the model of `widget-map.ts`, in the course of adding the above parser, this adds functions for parsing a UserInputMap and its entries. Has a number of TODOs to finish once the remaining widget user input parsers are added.

Issue: LEMS-3140

## Test plan:
- Confirm all checks pass
- Confirm widget still acts as expected via Storybook


packages/perseus-core/src/parse-perseus-json/perseus-parsers/user-input-map.ts
```diff
@@ -0,0 +1,165 @@
...
+            return parseAndAssign(`expression ${n}`, parseExpressionUserInput);
+        case "free-response":
+            // TODO: Add free-response user input parser
+            return parseAndAssign(`free-response ${n}`, any);
+        case "grapher":
+            // TODO: Add grapher user input parser
+            return parseAndAssign(`grapher ${n}`, any);
+        case "group":
+            // Recursive case: group user input is itself a UserInputMap
+            return parseAndAssign(`group ${n}`, parseUserInputMap);
```
benchristel: Did you want to use `parseGroupUserInput` here?

packages/perseus-core/src/parse-perseus-json/perseus-parsers/widgets-map.ts
```diff
@@ -231,4 +245,17 @@ const parseStringToNonNegativeInt: Parser<number> = (rawValue, ctx) => {
...
+ * This parser takes the components of a widget ID (after splitting on spaces) and
+ * validates that it consists of exactly two parts: a widget type string and a
+ * non-negative integer.
+ *
+ * @type {Parser<[string, number]>} A parser that returns a tuple of [widgetType, widgetNumber]
+ */
+export const parseWidgetIdComponents = pair(
+    string,
+    parseStringToNonNegativeInt,
+);
```
benchristel: `parseWidgetIdComponents` and `parseStringToNonNegativeInt` are duplicated in `widgets-map.ts. We could move them to their own file, e.g. `perseus-parsers/widget-id-components.ts`; the widget ID format is a cohesive concept that's unlikely to be different between these two usages.

Myranae: Hmm, I'm not sure I follow. Originally, Cursor tried to duplicate those two, but I had it export them from `widgets-map.ts` and use the exported ones, so now they are only in that one file. I can totally move them to their own file for clarity though?

packages/perseus-core/src/parse-perseus-json/perseus-parsers/user-input-map.ts
Myranae: Yes! Thanks for catching that :) Updated!

packages/perseus-core/src/parse-perseus-json/perseus-parsers/widgets-map.ts
benchristel: You are right, I misread the diff.

I'd still prefer it if these functions were in their own file.

Myranae: Moved the functions to their own file and updated `parseUserInputMapEntry` with the missing parsers :)

packages/perseus-core/src/parse-perseus-json/perseus-parsers/user-input-map.ts
```diff
@@ -0,0 +1,169 @@
...
+ * widgets need to be tracked together.
+ *
+ * The parser handles recursive structures since Group widgets can contain other Group
+ * widgets, each with their own UserInputMap.
+ *
+ * @param rawValue - The raw value to parse, expected to be a plain object
+ * @param ctx - The parse context for error reporting and tracking
+ * @returns A ParseResult containing either the parsed UserInputMap or an error
+ */
+export const parseUserInputMap: Parser<UserInputMap> = (rawValue, ctx) => {
```
benchristel, line 43: I think it would be good to have a few tests to at least spot-check that this is working. Yes, we have types, but they'd be just as happy with this implementation:

```ts
export const parseUserInputMap: Parser<UserInputMap> = (rawValue, ctx) => {
    return ctx.success({});
}
```

I think it would be nice to have tests for these behaviors:

- invalid user input for a known widget type is rejected
- valid user input for a known widget type is accepted
- user input for unrecognized widget types is passed through
- invalid keys in the user input map, like `""` and `"radio -1"` are rejected.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/user-input-map.ts
Myranae: Actually, it looks like this change caused a circular dependency (Cursor pointed this out). The other parsers have a specific shape to meet, but here, `parseGroupUserInput` is just a wrapper to call `parseUserInputMap`. So I reverted the change and it's back to just calling `parseUserInputMap`.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/user-input-map.ts
Myranae: Added tests that follow the framework in `widgets-map.test.ts`. Added with help from Cursor.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/user-input-map.test.ts
```diff
@@ -0,0 +1,456 @@
...
+            expect(result).toEqual(
+                parseFailureWith({
+                    expected: ["UserInputMap"],
+                    badValue: "not an object",
+                }),
+            );
+        });
+    });
+
+    describe("invalid widget ID keys", () => {
```
benchristel, line 43: These are good cases to test, but I think we could move most of these detailed tests to a `widget-id-components.test.ts` file. Then we can just have one test here that says "it rejects an invalid widget ID".

packages/perseus-core/src/parse-perseus-json/perseus-parsers/user-input-map.test.ts
```diff
@@ -0,0 +1,456 @@
...
+
+            assertFailure(result);
+            expect(result.detail[0].path).toEqual([
+                "radio -1",
+                "(widget ID)",
+                1,
+            ]);
+            expect(result.detail[0].expected).toEqual([
+                "a string representing a non-negative integer",
+            ]);
```
benchristel: This could become:

```suggestion
            expect(result).toEqual(parseFailureWith({
                path: ["radio -1", "(widget ID)", 1],
                expected: "a string representing a non-negative integer",
            }));
```

FWIW, if we keep only one test for invalid widget IDs in this file, this should probably be the one.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/user-input-map.test.ts
```diff
@@ -0,0 +1,456 @@
...
+        });
+    });
+
+    describe("valid user input for known widget types", () => {
```
benchristel: Looks like Cursor left out some tests here... We have 20 widgets with parseable user input, but only 9 tests for valid input. I would rather go to one of two extremes:

- test only one widget type, banking on the supposition that the switch statement in `parseUserInputMap` is "too simple to break" and doesn't need detailed tests.
- test user inputs for all widget types, so e.g. if we break one of the user input parsers, we will get a test failure.

Since we have solid typetests for the widget parsers, it seems that they're unlikely to reject valid widgets. **So I'd prefer to test just one valid widget here.** I think it's more important to test the invalid case for every widget, since that guards against bugs that are easier to introduce (e.g. deleting a `case` from the switch statement).

Testing a smattering of widget types is confusing, because readers might think that all widgets are being tested, when in fact only some of them are.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/user-input-map.test.ts
```diff
@@ -0,0 +1,456 @@
...
+            const userInputMap = {
+                "dropdown 1": {value: "not a number"},
+            };
+            const result = parseUserInputMap(userInputMap, ctx());
+
+            assertFailure(result);
+            expect(result.detail[0].path).toEqual(["dropdown 1", "value"]);
+            expect(result.detail[0].expected).toEqual(["number"]);
+            expect(result.detail[0].badValue).toBe("not a number");
+        });
```
benchristel: These tests seem redundant, and it's weird that they're sandwiched between other tests for specific widget types. I'd just as soon remove them.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/user-input-map.test.ts
```diff
@@ -0,0 +1,456 @@
...
+                success({
+                    "dropdown 1": {value: 0},
+                    "radio 2": {choicesSelected: [false, true]},
+                    "numeric-input 3": {currentValue: "123"},
+                }),
+            );
+        });
+    });
+
+    describe("invalid user input for known widget types", () => {
```
benchristel: As with the valid cases, it seems like Cursor picked 4 widgets at random to test. I'd rather test all of the widgets here. That ensures that if we accidentally delete any `case` from the switch statement, or replace a user input parser with `any`, we'll get a test failure.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/user-input-map.test.ts
```diff
@@ -0,0 +1,456 @@
...
+            const result = parseUserInputMap(userInputMap, ctx());
+            expect(result).toEqual(
+                success({
+                    "dropdown 1": {value: 2},
+                    "unknown-widget 1": {customField: "custom value"},
+                    "radio 2": {choicesSelected: [false, true]},
+                    "another-mystery 3": [1, 2, 3],
+                }),
+            );
+        });
```
benchristel: I'm not really sure what the purpose of these two tests is... it's hard for me to imagine how we would change the code in a way that would cause them to fail. I.e. I'm not sure what bugs they protect us from.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/user-input-map.test.ts
```diff
@@ -0,0 +1,456 @@
...
+                                structure: [1, 2, 3],
+                                withBooleans: true,
+                            },
+                        },
+                        nullValue: null,
+                        arrayOfObjects: [{a: 1}, {b: 2}],
+                    },
+                }),
+            );
+        });
```
benchristel: This test seems redundant with the one above.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/user-input-map.test.ts
```diff
@@ -0,0 +1,456 @@
...
+            };
+            const result = parseUserInputMap(userInputMap, ctx());
+
+            assertFailure(result);
+            expect(result.detail[0].path).toEqual([
+                "group 1",
+                "dropdown 1",
+                "value",
+            ]);
+            expect(result.detail[0].expected).toEqual(["number"]);
```
benchristel: This could be

```suggestion
            expect(result).toEqual(parseFailureWith({
              path: ["group 1", "dropdown 1", "value"],
              expected: ["number"],
            }));
```

packages/perseus-core/src/parse-perseus-json/perseus-parsers/group-user-input.ts
```diff
@@ -0,0 +1,7 @@
+import {parseUserInputMap} from "./user-input-map";
+
+import type {PerseusGroupUserInput} from "../../validation.types";
+import type {Parser} from "../parser-types";
+
+export const parseGroupUserInput: Parser<PerseusGroupUserInput> =
+    parseUserInputMap;
```
benchristel: This gets around the circular dependency issue:

```suggestion
export const parseGroupUserInput: Parser<PerseusGroupUserInput> =
    (rawVal, ctx) => parseUserInputMap(rawVal, ctx);
```

Myranae: So adding this function wrapper allows us to call `parseGroupUserInput` from `parseUserInputMapEntry` in `user-input-map.ts`? Is this preferred because the case for "group" then follows the same pattern as the other widgets?

packages/perseus-core/src/parse-perseus-json/perseus-parsers/user-input-map.test.ts
Myranae: Updated~

packages/perseus-core/src/parse-perseus-json/perseus-parsers/group-user-input.ts
benchristel: Yes to both!

packages/perseus-core/src/parse-perseus-json/perseus-parsers/user-input-map.test.ts
Myranae: Looks like the first one is to protect against inconsistent handling between known and unknown widgets 
```
// These should BOTH work:
{
  "expression 1": "x^2 + 2x + 1",        // ✅ Known widget with primitive (string)
  "future-widget 1": "some string value" // ✅ Unknown widget with primitive (string)  
}

// The bug would be if unknown widgets couldn't handle primitives
// while known widgets (like expression) could
```

and the second one is to make sure things work correctly if unknown widget data is mixed with known widget data
```// User has mixed content:
{
    "dropdown 1": { value: 0 },
    "future-widget 1": "new data format",
    "radio 2": { choicesSelected: [true] }
}

// Potential bugs:
// 1. Dropdown validation gets skipped
// 2. Radio validation fails because of "contamination"
// 3. Future-widget gets incorrectly validated as dropdown
```
Totally cool removing them if you don't think they're worthwhile.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/user-input-map.test.ts
Myranae: Moved the tests :)

packages/perseus-core/src/parse-perseus-json/perseus-parsers/user-input-map.test.ts
Myranae: Kept this test and updated the expect section

packages/perseus-core/src/parse-perseus-json/perseus-parsers/user-input-map.test.ts
Myranae: Sounds good! Will just keep the first one unless you think a specific one would be better.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/user-input-map.test.ts
Myranae: Added tests so there is one for all 20 widgets with parseable user input.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/user-input-map.test.ts
Myranae: Removed

packages/perseus-core/src/parse-perseus-json/perseus-parsers/user-input-map.test.ts
Myranae: Removed

packages/perseus-core/src/parse-perseus-json/perseus-parsers/user-input-map.ts
Myranae: Fixed based on your other comment. Used the function wrapper

packages/perseus-core/src/parse-perseus-json/perseus-parsers/user-input-map.test.ts
Myranae: Decided to remove these two tests as they're basically duplicated tests.

PR #2547, opened 2025-06-02
Accept a range of versions in `peerDependencies`, but install specific versions in dev

benchristel:
Previously, we had a single workspace `catalog` that was used for all
our dependencies. This meant that our version specifiers for
`peerDependencies` and `devDependencies` were the same. The result was
that:

- either we would specify a single version of a package as our peer dep,
  which is bad because then clients get warnings if their package versions
  are even slightly different.
- or we would specify a version range as our dev dep. This usually resulted
  in a more recent version of the package being installed in dev, e.g. if
  we specified `^1.1.0`, the installed version might be `1.3.5`. This is
  bad because we might accidentally depend on features introduced in 1.3.x
  that aren't present in 1.1.0 — meaning our peer dep range is a lie.

This PR fixes things so that:

- We install a specific version of each package in dev, e.g. `1.1.0`.
- We accept any version compatible with that as a peer dependency,
  e.g. `^1.1.0`.
- We can run `utils/sync-dependencies.ts` to update our package catalog
  based on the webapp frontend's package.json.

Issue: none

## Test plan:

```
pnpm storybook
open http://localhost:6006
```

Review the stories. Everything should work.

Run `utils/sync-dependencies.ts ../webapp/services/static/package.json`.
There should be no changes to `pnpm-workspace.yaml`.


benchristel:
@catandthemachines I made some changes to future-proof this. I figured you might like to take another look.

.github/dependabot.yml
```diff
@@ -5,6 +5,7 @@ version: 2
 updates:
     - package-ecosystem: "npm"
       directory: "/"
+      versioning-strategy: increase-if-necessary
```
benchristel: I don't have much confidence that dependabot will magically do the right thing here. Dependabot will try to keep us on the very latest version of Wonder Blocks, but we might not actually want to update WB until Webapp has done so. I would still lean toward treating dependabot's PRs as "informational only" and using `utils/sync-dependencies.ts` to actually bump the versions.

jeremywiebe: Perhaps we should just give up on the Dependabot experiment. Or at least not use it for WonderBlocks. 

We could use `sync-dependencies` as the way to update WB as part of a period task that someone does (perhaps part of support rotation?)

packages/perseus-core/src/parse-perseus-json/regression-tests/item-data/orderer-missing-options.json
jeremywiebe, line 1: Why was this touched?

utils/sync-dependencies.ts
```diff
@@ -81,26 +96,47 @@ function main(argv: string[]) {
...
+            // any package version compatible with the one we install in dev.
+            workspace.catalogs.peerDeps[pkgName] = `^${minVersion}`;
         }
     }
 
+    const comment = dedent(`
+        # NOTE: The \`devDeps\` and \`peerDeps\` catalogs in this file are
+        # generated from webapp's package.json. To update them, run:
+        #
+        #     utils/sync-dependencies.ts ../webapp/services/static/package.json
```
benchristel, line 142: We'll need to update this path after the frontend moves to its own repo...

pnpm-workspace.yaml
```diff
@@ -1,42 +1,86 @@
+# NOTE: The `devDeps` and `peerDeps` catalogs in this file are
+# generated from webapp's package.json. To update them, run:
+#
+#     utils/sync-dependencies.ts ../webapp/services/static/package.json
+#
```
jeremywiebe, line 5: Could you expand this comment to explain how we manage dev and peer deps?  (basically much of what the PR description is).  That'll help folks managing this file to know how to update it.

jeremywiebe (approved):
Nice. Let's give this a try.  @anakaren-rojas @catandthemachines  I know you've been managing the Dependabot PRs, so FYI that this will change things a bit.

.github/dependabot.yml
benchristel: > We could use sync-dependencies as the way to update WB as part of a period task that someone does (perhaps part of support rotation?)

I'd be down to try it!

utils/sync-dependencies.ts
benchristel: I added a TODO comment.

packages/perseus-core/src/parse-perseus-json/regression-tests/item-data/orderer-missing-options.json
benchristel: Oops... this was left over from my debugging effort to support CP with parser failures in the publish pipeline. I'll remove it.

.github/dependabot.yml
jeremywiebe: I would be good if we deleted our dependabot config in this PR and documented that The Way to update WB dependencies was the sync-dependencies script.

benchristel: Documented here: https://khanacademy.atlassian.net/wiki/spaces/LC/pages/2384887922/Shipping+a+Change+to+Perseus#Updating-Dependencies.

I also set a weekly reminder for lems-support to run sync-dependencies.

PR #2548, opened 2025-06-02
Add typesafe parser for Iframe's user input type

Myranae:
## Summary:
Add a user input type parser and typetest for the Iframe widget

Issue: LEMS-3133

## Test plan:
- Confirm all checks pass
- Confirm the widget still works as expected via Storybook


Myranae:
@benchristel I think if I land this one _first_, then I can add this to the other PR, right? I'll make sure to update the Group switch case after I've finished the last remaining widgets :)

PR #2550, opened 2025-06-03
Add typesafe parser for CS Program's user input type

Myranae:
## Summary:
Add a user input type parser and typetest for the CS Program widget

Issue: LEMS-3137

## Test plan:
- Confirm all checks pass
- Check either on Webapp ZND or during next deploy to confirm CS Program works as expected (does not work in Storybook currently)


benchristel:
The CS program parser should be called by the userInputMap parser, so you'll have to decide which of https://github.com/Khan/perseus/pull/2545 and https://github.com/Khan/perseus/pull/2550 should land first, and then update the other. @Myranae

PR #2553, opened 2025-06-04
Adding partially inaccessible functions to InteractiveGraph and LabelImage.

catandthemachines:
## Summary:
Adding partially inaccessible functions to InteractiveGraph and LabelImage. These functions detect the state in which these functions are inaccessible. Those states are:

InteractiveGraph:
- interactive graphs with graphie backgrounds.
- interactive graphs with a protractor.

LabelImage:
- label image widgets with background images that don't have alt text

Issue: LEMS-3102

## Test plan:
Run `pnpm test`, all tests should pass
Go to storybook page [chromatic]/?path=/story/perseuseditor-editorpage--demo
- Add interactive graphs with inaccessible states (listed above) and confirm they are flagged as inaccessible.
- Add label image with inaccessible state and confirm it flags as inaccessible.


packages/perseus-core/src/widgets/interactive-graph/index.ts
```diff
@@ -44,8 +47,28 @@ const interactiveGraphWidgetLogic: WidgetLogic = {
...
+        // a protractor.
+        if (interactiveGraphOptions.showProtractor) {
+            return false;
+        }
+
+        // Return false (inaccessible) if the interactive graph contains
+        // a graphie background image.
+        if (interactiveGraphOptions.backgroundImage?.url?.includes("graphie")) {
+            return false;
+        }
```
jeremywiebe: We have a utility function to detect Graphie URLs: [`isLabeledSVG()`](https://github.com/Khan/perseus/blob/3389e06111b2c801fba6c34ed4e5bdc747ef1ed1/packages/perseus/src/util.graphie.ts#L19) (Perseus thinks of Graphie's as "labelled SVGs").  As with many things in this work we're doing, we'll need to move this function over to `perseus-core`, but it's small and self-contained so should be pretty easy to migrate.

jeremywiebe (commented):
Sorry, I know I'm requesting some more code moves, but I think we should use the `isLabeledSVG` function (which involves a bit of code moving).

packages/perseus-core/src/widgets/interactive-graph/index.ts
catandthemachines: Updated to use `isLabeledSVG()`. I also moved the utility into `perseus-core` as there's a lot of usage of this utility in perseus-core and perseus, which says to me it should live in `perseus-core`. But let me know what you think and we can go from there. 😉

packages/perseus/src/util.ts
```diff
@@ -618,12 +624,12 @@ const Util = {
...
-    isLabeledSVG: GraphieUtil.isLabeledSVG,
-    getBaseUrl: GraphieUtil.getBaseUrl,
-    getSvgUrl: GraphieUtil.getSvgUrl,
-    getDataUrl: GraphieUtil.getDataUrl,
+    getImageSizeModern: getImageSizeModern,
+    getRealImageUrl: getRealImageUrl,
+    isLabeledSVG: isLabeledSVG,
+    getBaseUrl: getBaseUrl,
+    getSvgUrl: getSvgUrl,
+    getDataUrl: getDataUrl,
```
jeremywiebe, line 632: Thanks.  Eventually we'll clean up this `util.ts` file so that users of it import just what they need.

PR #2555, opened 2025-06-04
WIP/POC: Dropdown: store user input in Renderer state

handeyeco:
## Summary:

Read the comments in [LEMS-3113](https://khanacademy.atlassian.net/browse/LEMS-3113) for context.

**tl;dr**

We think it would be best for SSS and Perseus as a whole if we hoisted user input state above widgets (for now, in Renderer) and to separate state from widget options. This is me experimenting with that idea for Dropdown.

**New Problem**

Right now most tests are passing ~except for serialization and restoring serialized state~ have a possible fix. As mentioned in the comments of the issue, we expected this to be a little bit of a problem and we also discovered that the serialization code is probably horribly broken. As far as we could tell, restoring serialized state doesn't go through the migration/upgrade logic so there's no way for us to handle old serialized state. This becomes even more problematic with the changes in this PR because I'm trying to split widget options from user input and historically we kind of just meshed the two of those together. Still investigating, but pushing to get thoughts/feels.

Issue: [LEMS-3185](https://khanacademy.atlassian.net/browse/LEMS-3185)

Part of [LEMS-3113](https://khanacademy.atlassian.net/browse/LEMS-3113)

### DIFF
--- a/packages/perseus/src/renderer.tsx
+++ b/packages/perseus/src/renderer.tsx
@@ -658,7 +658,7 @@ class Renderer
      */
     /**
      * @deprecated and likely very broken API
-     * do not trust serializedState/restoreSerializedState
+     * [LEMS-3185] do not trust serializedState/restoreSerializedState
      */
     getSerializedState: (widgetProps?: any) => {
         [id: string]: any;
@@ -681,7 +681,7 @@ class Renderer
 
     /**
      * @deprecated and likely very broken API
-     * do not trust serializedState/restoreSerializedState
+     * [LEMS-3185] do not trust serializedState/restoreSerializedState
      */
     restoreSerializedState: (
         serializedState: SerializedState,
--- a/packages/perseus/src/renderer.tsx
+++ b/packages/perseus/src/renderer.tsx
@@ -656,6 +656,10 @@ class Renderer
      * If an instance of widgetProps is passed in, it generates the serialized
      * state from that instead of the current widget props.
      */
+    /**
+     * @deprecated and likely very broken API
+     * do not trust serializedState/restoreSerializedState
+     */
     getSerializedState: (widgetProps?: any) => {
         [id: string]: any;
     } = (
@@ -675,6 +679,10 @@ class Renderer
         );
     };
 
+    /**
+     * @deprecated and likely very broken API
+     * do not trust serializedState/restoreSerializedState
+     */
     restoreSerializedState: (
         serializedState: SerializedState,
         callback?: () => void,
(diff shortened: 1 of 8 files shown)


packages/perseus/src/renderer.tsx
```diff
@@ -629,6 +633,15 @@ class Renderer
...
             },
+            handleUserInput: (newUserInput: UserInput) => {
+                this.setState({
+                    userInput: {
+                        ...this.state.userInput,
+                        [widgetId]: newUserInput,
+                    },
+                });
+            },
+            userInput: this.state.userInput[widgetId],
```
benchristel, line 644: Is the idea that `handleUserInput` is a more typesafe replacement for `onChange`?

packages/perseus/src/renderer.tsx
```diff
@@ -1718,13 +1731,18 @@ class Renderer
...
+            if (this.state.userInput[id]) {
+                // Get user input from Renderer state if possible
+                userInputMap[id] = this.state.userInput[id];
+            } else if (widget?.getUserInputMap) {
+                // Handle Groups, which have their own sets of widgets
                 userInputMap[id] = widget.getUserInputMap();
             } else if (widget?.getUserInput) {
+                // Legacy method of getting user input
                 userInputMap[id] = widget.getUserInput();
             }
```
benchristel, line 1762: I think we probably need to prefer calling `getUserInputMap` over `this.state.userInput[id]`. I.e.

```suggestion
            if (widget?.getUserInputMap) {
                // Handle Groups, which have their own sets of widgets
                userInputMap[id] = widget.getUserInputMap();
            } else if (this.state.userInput[id]) {
                // Get user input from Renderer state if possible
                userInputMap[id] = this.state.userInput[id];
            } else if (widget?.getUserInput) {
                // Legacy method of getting user input
                userInputMap[id] = widget.getUserInput();
            }
```

It would be better if `Group` widgets called `handleUserInput` when any of their widgets changed, instead of being special-cased in the renderer.

packages/perseus/src/renderer.tsx
handeyeco: Yeah, I guess so. `ChangeHandler` is a mess and for some reason I was thinking `onChange` handled stuff other than user input. If it's just for user input we could merge the two.

As far as types, ideally `WidgetProps` and `UniversalWidgetProps` could take some kind of "widget type" type argument and we could use that to determine specific types. For instance it would be nice to do something like `WidgetProps<"dropdown">` and have TS understand that we expect:

``` TS
handleUserInput: (newUserInput: PerseusDropdownUserInput) => void;
userInput: PerseusDropdownUserInput;
```

Even if we can't do that, we could just use `WidgetProps<PerseusDropdownUserInput>` now that I think about it...

**tl;dr**: I did it because I like state/props/callbacks to be focused and `ChangeHandler` is kind of a mess.

packages/perseus/src/renderer.tsx
handeyeco: Yeah, Group is tricky. It renders a Renderer and Renderers maintain their own state...

In this case I don't think the order has a practical effect. `this.state.userInput[id]` will be undefined for Groups until Group implements `handleUserInput` and once that happens `getUserInputMap` should be obsolete.

I agree that Group should call `handleUserInput` when one of its components changes. The problem is that Renderers render Groups which render Renderers and Renderers maintain state, if Group calls `handleUserInput` when its Renderer has a call to `handleUserInput` then we'll be storing state in two Renderers.

I'm probably not going to do this, but it makes me wonder if we need to split Renderer into a "smart component" that stores state and a "dumb component" that does everything else. Then we could have one smart component that maintains state and an unlimited number of dumb Renderers that just render the state that they receive. I feel like I'm the verge of reinventing Redux...

benchristel: Consulting my crystal ball... if we continue going down this path we might end up moving all state out of React components, and instead exporting a `const [state, setState] = usePerseusItemState(item)` hook that clients are supposed to call. That would also give clients a way to access the state without duplicating it, which is useful for e.g. user input validation and scoring.

handeyeco: Yeah, I've been advocating with Jeremy for a React Context in consumers. It'd be nice to store state there because:

1. The consumer could have direct access to data (like user input)
2. We could provide functions that might help get rid of other imperative APIs
3. We could use it to provide dependencies/API options
4. By using Context, we could do less prop drilling

Something like:

``` JSX
function WrappedChrome({ item })) {
  return (
    <PerseusProvider item={item} dependencies={perseusDependencies}>
        <Chrome />
    </PerseusProvider>
  )
}

function Chrome() {
  const {userInput} = useUserInput()

  function handleSubmit() {
    apiCall(userInput)
  }

  return (
    <>
      <ServerItemRenderer />
      <button onClick={handleSubmit}>Submit</button>
    </>
}

function Radio({ widgetId }) {
  const {options} = useWidgetOptions(widgetId)
  const {userInput, changeUserInput} = useUserInput(widgetId)
  const {isMobile} = useDependencies()

  render <>You get the point</>
}
```

packages/perseus/src/widgets/dropdown/dropdown.tsx
```diff
@@ -54,17 +63,43 @@ class Dropdown extends React.Component<Props> implements Widget {
...
+            placeholder: this.props.placeholder,
+            selected: this.props.userInput.value,
+        };
+    }
+
+    /**
+     * @deprecated and likely very broken API
+     * [LEMS-3185] do not trust serializedState/restoreSerializedState
+     */
+    getUserInputFromSerializedState(serialized: any): PerseusDropdownUserInput {
```
handeyeco, line 99: This is a pretty gnarly solution...basically the idea was:

1. Write tests to determine what serialized state the renderer was getting
2. Add `getSerializedState` to Dropdown to implement that serialized state
3. Add a new API called `getUserInputFromSerializedState` to split back out the user input
4. Add a test to make sure user input is restoring properly

This is extremely hacky, but it maintains backwards compatibility for an already problematic API (serialized state) so that I can move forward for SSS until we can circle back in LEMS-3185.

handeyeco: Turns out there's already kind of an API for this: `getUserInputFromProps`

I think maybe I'll repurpose that API for getting user input from serialized state and convert all non-serialized uses to using `userInput` directly.

handeyeco:
Closing in favor of:
- https://github.com/Khan/perseus/pull/2566
- https://github.com/Khan/perseus/pull/2568

PR #2556, opened 2025-06-04
Sync WB dependencies with webapp

benchristel:
In preparation for releasing a new Perseus version.

Issue: none

## Test plan:

```
pnpm test
pnpm storybook
open http://localhost:6006/
```


pnpm-workspace.yaml
```diff
@@ -76,29 +76,29 @@ catalogs:
...
+        "@khanacademy/wonder-blocks-switch": 3.2.9
         "@khanacademy/wonder-blocks-timing": 7.0.2
-        "@khanacademy/wonder-blocks-tokens": 10.3.0
-        "@khanacademy/wonder-blocks-toolbar": 5.1.15
-        "@khanacademy/wonder-blocks-tooltip": 4.1.18
-        "@khanacademy/wonder-blocks-typography": 4.0.0
+        "@khanacademy/wonder-blocks-tokens": 10.4.0
+        "@khanacademy/wonder-blocks-toolbar": 5.1.17
+        "@khanacademy/wonder-blocks-tooltip": 4.1.20
+        "@khanacademy/wonder-blocks-typography": 4.2.0
```
catandthemachines, line 101: Didn't we need a `^` on this package? I remember you mentioned this causing issues in webapp.

catandthemachines (approved):
Approved! But added a question on the devDep versions.

pnpm-workspace.yaml
benchristel: For dev deps, we want to install the minimum version allowed by webapp's version range. If we install any later version (which might happen if we used a range like `^4.2.0`) then we might end up depending on features of the package that don't exist in the version that's actually installed in webapp.

To ensure we get the exact version we want, we don't use version ranges for dev deps.

PR #2557, opened 2025-06-05
Add rationales to the getPromptJSON output of Radio widget and Label Image widget

marekweb:
## Summary:

Add rationales to the getPromptJSON output of Radio widget and Label Image widget.

This is a small addition to the information that getPromptJSON already returns, and only affects these specific widgets.

The intent is to have rationales accessible when we use getPromptJSON to get the state of widgets.

Issue: TUT-2087

## Test plan:
- Updated tests
- Run all tests

### DIFF
--- a/packages/perseus/src/widget-ai-utils/label-image/label-image-ai-utils.ts
+++ b/packages/perseus/src/widget-ai-utils/label-image/label-image-ai-utils.ts
@@ -8,7 +8,7 @@ type BaseMarker = {
 
 type UserInputMarker = {
     label: string;
-    rationale?: string[];
+    answers?: string[];
     selected?: ReadonlyArray<string>;
 };
 
@@ -34,7 +34,7 @@ export const getPromptJSON = (
             label: marker.label,
         };
         if (marker.answers?.length) {
-            userInputMarker.rationale = marker.answers;
+            userInputMarker.answers = marker.answers;
         }
         return userInputMarker;
     });
--- a/packages/perseus/src/widget-ai-utils/label-image/label-image-ai-utils.ts
+++ b/packages/perseus/src/widget-ai-utils/label-image/label-image-ai-utils.ts
@@ -8,6 +8,7 @@ type BaseMarker = {
 
 type UserInputMarker = {
     label: string;
+    rationale?: string[];
     selected?: ReadonlyArray<string>;
 };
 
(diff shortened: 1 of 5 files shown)


packages/perseus/src/widget-ai-utils/label-image/label-image-ai-utils.ts
```diff
@@ -8,6 +8,7 @@ type BaseMarker = {
 
 type UserInputMarker = {
     label: string;
+    rationale?: string[];
```
benchristel: A couple questions:

- Why is this called `rationale` and not something like `answers` or `correctAnswers`?

- The `answers` for each marker will soon be `undefined` on the frontend until the learner submits their first attempt on the question. To prevent cheating, we're removing answer information from the data that's initially sent to the frontend. Are you calling `getPromptJSON` on the frontend?

packages/perseus/src/widget-ai-utils/radio/radio-ai-utils.ts
```diff
@@ -26,9 +27,13 @@ export const getPromptJSON = (
     const choices = renderProps.choices || [];
 
     const options = choices.map((choice) => {
-        return {
+        const option: BasicOption = {
             value: choice.content,
         };
+        if (choice.clue) {
```
anakaren-rojas, line 33: As Jeremy mentioned in his comment, we're in the process of updating "clue" to "rationale" as well, that change will be coming some time next week

packages/perseus/src/widget-ai-utils/label-image/label-image-ai-utils.ts
marekweb: - I suppose it doesn't have to be called `rationale`, that's just the name that seems to have circulated to refer to these fields collectively, with the naming precedent coming from e.g. [widgetHasRationales](https://github.com/Khan/perseus/blob/cec57bbc6cf32cdce5a6b764b7399f69f87b4a89/packages/perseus-core/src/utils/item-has-rationales.ts#L18). It does seem that for this widget they're just the answers, so I can rename to `answers`.
- Yes, we're calling getPromptJSON on the frontend, and I'm aware that this will be changing. But we mostly want to provide the answers/rationales for tutoring **after** the user submits an attempt, so if that will still work, then that's still mostly pretty good for our use case.

packages/perseus/src/widget-ai-utils/radio/radio-ai-utils.ts
marekweb: Got it, is it fine to land this as-is with `clue` in the meantime before you do the rename?

packages/perseus/src/widget-ai-utils/label-image/label-image-ai-utils.ts
marekweb: I pushed a commit to rename it to `answers`

packages/perseus/src/widget-ai-utils/radio/radio-ai-utils.ts
anakaren-rojas: Yup! Sorry, it was more of an FYI in case you get tagged to review the PR that updates the field

anakaren-rojas (approved):
Approved on behalf of radio team

PR #2529, opened 2025-05-27
feat(LEMS-3069): Perseus Parser updates for randomization

anakaren-rojas:
## Summary:
Updates perseus parser to v4 to include choiceIds

Issue: LEMS-3069

## Test plan:
`pnpm test`

PR #2559, opened 2025-06-05
feat(LEMS-3070): add ids to radio choices

anakaren-rojas:
## Summary:
- Updates choices to include choice Ids 
- Includes "fallback" choice id implementation in radio component (old and new) in case choice doesn't have choice id 

Issue: LEMS-3070

## Test plan:


packages/perseus/src/widgets/radio/base-radio.new.tsx
```diff
@@ -26,6 +26,7 @@ const {captureScratchpadTouchStart} = Util;
 
 // exported for tests
 export type ChoiceType = {
+    id: string;
```
ivyolamit: In previous/base [PR](https://github.com/Khan/perseus/pull/2529/files#diff-17ad85d7f821dc080c7b004a2c0213fa3fad6c1e919a525d47d262202417b871R1388) the id is optional, but here it's not. Does my understanding align with you?
- it's assumed that this will always be provided because we generate a unique id for this in this line in the parse [here](https://github.com/Khan/perseus/pull/2529/files)

POV: just making sure the use of choice.id in line 343 will not cause any error 🙏🏼

ivyolamit: update: i just saw this [line](https://github.com/Khan/perseus/pull/2559/files#diff-be53029762fed6689372a5a9e2b07606aec659332e43fc25cc6d47a0dd8dd40fR211) where it ensures the choice id to be available.

with that, let's just test it for multiple radio widget 😉

packages/perseus/src/widgets/radio/base-radio.tsx
```diff
@@ -28,6 +28,7 @@ const {captureScratchpadTouchStart} = Util;
 
 // exported for tests
 export type ChoiceType = {
+    id: string;
```
mark-fitzgerald, line 31: Is this also needed in `base-radio.new.tsx`?

anakaren-rojas:
too stale - replacing with new PR

PR #2563, opened 2025-06-06
Register core widgets at top level in perseus-core/src/index.ts

benchristel:
## Summary:
This is one possible approach to reducing the number of magic setup functions
clients of Perseus have to call. Currently, clients must remember to call
`registerCoreWidgets()` before using any widget-specific perseus-core
functionality like answer removal. If they forget, they'll get an exception.

However, `registerCoreWidgets` only exists to prevent a const from being used
before it is bound (which happens because of an import cycle). By calling it in
the index file, we can defer its execution until all the modules in the package
have been evaluated and their bindings finalized, while NOT requiring Perseus
clients to call it.

I suspect there is a more elegant solution to this problem — maybe each core widget
could import the registry and register itself?

Issue: LEMS-3163

## Test plan:

`pnpm test`

### DIFF
--- a/.changeset/fair-rings-call.md
+++ b/.changeset/fair-rings-call.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/perseus-core": major
+"@khanacademy/perseus": patch
+---
+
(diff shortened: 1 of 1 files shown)


catandthemachines (approved):
This change looks good to me. I don't see any long term issues with this switch, but if something does come up we can always pivot. Thank you Ben!

PR #2564, opened 2025-06-06
DO NOT MERGE: POC new radio widget choice icon

ivyolamit:
## Summary:
DO NOT MERGE POC for new radio widget choice icon


packages/perseus/src/widgets/radio/choice-icon/choice-icon.new.css
```diff
@@ -0,0 +1,105 @@
+:root {
+    --choice-icon-size: 24px;
```
jandrade: **suggestion:** I'd recommend using `--wb-sizing-size_240`, that way:

a) more WB tokens yay!
b) more accessible, as these sizing tokens are using `rem` units, which are better for users that enable browser zooming (`rem` scales better than hard-coded `px` values :)).

WB Docs: https://khan.github.io/wonder-blocks/?path=/docs/packages-tokens-sizing--docs#tokens

packages/perseus/src/widgets/radio/choice-icon/choice-icon.new.css
```diff
@@ -0,0 +1,105 @@
...
+
+[dir="rtl"] .choice-icon-wrapper {
+    background: linear-gradient(to left, var(--wb-semantic-core-background-default) 10%, transparent);
+}
+
+/* Focus ring */
+.focus-ring {
+    margin: auto;
+    display: inline-block;
+    border-width: 2px;
```
jandrade: **suggestion:** you could try out using WB border tokens here too.

```suggestion
    border-width: var(--wb-border-width-medium);
```

**Why?** The final value won't change per se, but it's more future proof as it would align with the WB token in case we change that systematically in the future.

WB Docs: https://khan.github.io/wonder-blocks/?path=/docs/packages-tokens-border--docs

packages/perseus/src/widgets/radio/choice-icon/choice-icon.new.css
```diff
@@ -0,0 +1,105 @@
...
+    padding: 2px;
+    border-style: solid;
+    border-color: transparent;
+    border-radius: 50%;
```
jandrade: **suggestion:** Similar suggestion, it doesn't change the end result, but it aligns with the overall design language.

```suggestion
    border-radius: var(--wb-border-radius-radius_full);
```

WB Docs: https://khan.github.io/wonder-blocks/?path=/docs/packages-tokens-border--docs#radius

packages/perseus/src/widgets/radio/choice-icon/choice-icon.new.tsx
```diff
@@ -27,194 +28,96 @@ interface ChoiceIconProps {
...
+    // Button style based on circle or square shape
+    const buttonStyle = {
+        borderRadius: multipleSelect
+            ? border.radius.radius_040
+            : border.radius.radius_full,
+        width: sizing.size_240,
+        height: sizing.size_240,
+        padding: 0,
+        minWidth: "unset",
+        fontSize: "12px", // Smaller font size for the letter/symbol
```
jandrade: **suggestion:** You could use our WB typography/font tokens here:

```suggestion
        fontSize: font.body.size.xsmall, // Smaller font size for the letter/symbol
```

**why?** Another benefit is that fonts work in a similar way to semantic colors, as the font sizes (and related styles) might change based on the current theme. This would mean that you would be using different sizes depending on where the widget is being included (Classic or Classroom).

https://khan.github.io/wonder-blocks/?path=/docs/packages-tokens-typography--docs&globals=theme:default#size

packages/perseus/src/widgets/radio/choice-icon/choice-icon.new.tsx
```diff
@@ -27,194 +28,96 @@ interface ChoiceIconProps {
...
+        borderWidth: "2px", // Thicker border for better visibility
```
jandrade: **suggestion:** Similar to the CSS file, you can use the JS token here.

```suggestion
        borderWidth: border.width.medium, // Thicker border for better visibility
```

packages/perseus/src/widgets/radio/choice-icon/choice-icon.new.tsx
```diff
@@ -27,194 +28,96 @@ interface ChoiceIconProps {
...
+        boxShadow: "none",
+        fontWeight: "bold",
```
jandrade:
```suggestion
        fontWeight: font.weight.bold,
```

WB Docs: https://khan.github.io/wonder-blocks/?path=/docs/packages-tokens-typography--docs&globals=theme:default#size

jandrade (commented):
Left some suggestions around WB tokens usage, I hope that's helpful. Let me know if you have any questions :)

packages/perseus/src/widgets/radio/choice-icon/choice-icon.new.css
ivyolamit: yes 👍🏼 thanks for the recommendation.

packages/perseus/src/widgets/radio/choice-icon/choice-icon.new.css
ivyolamit: yes, these will be updated also 👍🏼

ivyolamit: For the padding is there a plan to support CSS variables for spacing?
I just noticed that it's not support yet [here](https://khan.github.io/wonder-blocks/?path=/docs/packages-tokens-spacing--docs)

But [sizing](https://khan.github.io/wonder-blocks/?path=/docs/packages-tokens-sizing--docs) has the CSS variable though
> All the available sizing values that can be used for margin, padding, width, height, border-width, font-size, line-height, etc.

jandrade: Not supporting CSS variables in `sizing` is an intentional choice, as we decided to make `sizing` the default category for anything related to size/space tokens. `spacing` is now considered deprecated and the goal is to migrate to `sizing` in places where it makes sense.

The reason why I mention this is because it makes sense to use `sizing` tokens is because we now default to `rem` units for better scalability/zoom, but there might be very specific cases where `rem` are not the best choice (e.g. `borders`, `outline`, etc). I'm going to update the docs to reflect this.

My recommendation in the short term is:

- For anything related to borders, please use `border` tokens. [^1]
- For `padding`, `margin`, `width`, `height`, use `sizing` tokens [^2].
- For `font-size`, `line-heigh`, etc, rely on `Heading` [^3] and `BodyText` [^4] components as much as you can. If that's not possible, fallback to `font` tokens [^5].

[^1]: https://khan.github.io/wonder-blocks/?path=/docs/packages-tokens-border--docs
[^2]: https://khan.github.io/wonder-blocks/?path=/docs/packages-tokens-sizing--docs
[^3]: https://khan.github.io/wonder-blocks/?path=/docs/packages-typography-heading-new--docs
[^4]: https://khan.github.io/wonder-blocks/?path=/docs/packages-typography-bodytext-new--docs
[^5]: https://khan.github.io/wonder-blocks/?path=/docs/packages-tokens-typography--docs

PR #2566, opened 2025-06-09
Renderer: separate user input state from render props state

handeyeco:
## Summary:

Issue: [LEMS-3208](https://khanacademy.atlassian.net/browse/LEMS-3208)

Edit see the ticket for full context.

**tl;dr: I need user input to be outside the chunk of state that Renderer stores arbitrary widget state in** (currently it's a mash of render props, user input, and whatever else widgets want to shove in there).

This ticket is the shared changes of this work - things that multiple widgets need. For the most part, I tried to have the other PRs _only_ affect things that are _specific_ to the individual widget. Anything outside of the immediate scope of the widget went into this PR. The main ideas:

- Renderer now has `userInput` state which is a `UserInputMap`
- `UniversalWidgetProps` now passes `userInput` and `handleUserInput` down to components for them to consume
- `WidgetExports` has some new optional APIs:
  - `getUserInputFromSerializedState` for restoring serialized state
  - `getStartUserInput` for initializing user input
  - `getCorrectUserInput` for static widgets
- Marked a bunch of things as deprecated

Note that besides some cleanup, no tests were harmed in the making of this PR. It's backwards compatible with our old way of doing things.

---

Reviewers don't need to go through all of these, but glancing at a couple might be helpful in understanding the impact of this PR.

Child PRs:
- [Categorizer](https://github.com/Khan/perseus/pull/2572)
- [CSProgram](https://github.com/Khan/perseus/pull/2584)
- [Dropdown](https://github.com/Khan/perseus/pull/2568)
- [Expression](https://github.com/Khan/perseus/pull/2569)
- [FreeResponse](https://github.com/Khan/perseus/pull/2585)
- [Grapher](https://github.com/Khan/perseus/pull/2596)
- [IFrame](https://github.com/Khan/perseus/pull/2582)
- [InputNumber](https://github.com/Khan/perseus/pull/2589)
- [InteractiveGraph](https://github.com/Khan/perseus/pull/2598)
- [LabelImage](https://github.com/Khan/perseus/pull/2577)
- [Matcher](https://github.com/Khan/perseus/pull/2575)
- [Matrix](https://github.com/Khan/perseus/pull/2581)
- [MockWidget](https://github.com/Khan/perseus/pull/2586)
- [NumberLine](https://github.com/Khan/perseus/pull/2573)
- [NumericInput](https://github.com/Khan/perseus/pull/2571)
- [Orderer](https://github.com/Khan/perseus/pull/2592)
- [Plotter](https://github.com/Khan/perseus/pull/2588)
- [Sorter](https://github.com/Khan/perseus/pull/2578)
- [Table](https://github.com/Khan/perseus/pull/2587)


packages/perseus/src/__tests__/server-item-renderer.test.tsx
```diff
@@ -15,56 +15,20 @@ import {
...
+
 import type {MockAssetLoadingWidget} from "../widgets/mock-widgets/mock-asset-loading-widget";
 import type {KeypadAPI} from "@khanacademy/math-input";
 import type {PerseusItem} from "@khanacademy/perseus-core";
-import type {PropsFor} from "@khanacademy/wonder-blocks-core";
 import type {UserEvent} from "@testing-library/user-event";
 
-// This looks alot like `widgets/__tests__/renderQuestion.jsx', except we use
-// the ServerItemRenderer instead of Renderer
-const renderQuestion = (
```
handeyeco, line 36: Just moved this to a separate file.

packages/perseus/src/__tests__/widgets.test.ts
```diff
@@ -102,171 +102,123 @@ describe("Widget API support", () => {
...
-        }
-    });
-
-    it("grapher widget getUserInputFromProps should return the correct user input", () => {
-        const Widget = Widgets.getWidget("grapher");
-
-        if (Widget && "getUserInputFromProps" in Widget) {
-            const props = {
-                plot: {
+    describe(`getUserInputFromSerializedState`, () => {
```
handeyeco, line 105: These tests were doing the same thing, so I consolidated them into `it.each`

packages/perseus/src/components/sortable.tsx
```diff
@@ -647,8 +647,9 @@ class Sortable extends React.Component<SortableProps, SortableState> {
         // @ts-expect-error - TS2551 - Property 'splice' does not exist on type 'readonly SortableItem[]'. Did you mean 'slice'?
         nextItems.splice(index, 0, item);
 
-        this.setState({items: nextItems});
-        this.props.onChange?.({});
+        this.setState({items: nextItems}, () => {
+            this.props.onChange?.({});
+        });
```
handeyeco: Sortable (it's a component, not a widget) stores state internally. Widgets that use Sortable had `getUserInput` methods that called into Sortable to get state. However now widgets should depend on `getUserInput` and the Renderer should always have an up-to-date set of user input.

What this does is every time we set state, we call `onChange`. This way widgets can call `handleUserInput` when Sorter changes. The problem with the original code is state was getting set after we called `onChange` because `setState` is async.

In the future we should probably just call `onChange` with the latest state. I was just trying to keep things mostly as-is since I have a ton of other changes I need to get through.

packages/perseus/src/mixins/widget-prop-denylist.ts
```diff
@@ -34,6 +37,29 @@ const denylist = [
...
     "keypadElement",
+    "linterContext",
+    "isLastUsedWidget",
+    "handleUserInput",
+    "analytics",
+    "showSolutions",
+    "alignment",
+    "reviewMode",
+    "reviewModeRubric",
+    "static",
```
handeyeco, line 48: This is all stuff that was getting serialized that shouldn't have been getting serialized. Overall the serialization APIs should be removed anyway (see: https://khanacademy.atlassian.net/browse/LEMS-3185) but in the meantime I'm just denying more stuff.

packages/perseus/src/renderer.tsx
```diff
@@ -168,6 +173,7 @@ type State = {
     widgetProps: Readonly<{
         [id: string]: any | null | undefined;
     }>;
+    userInput: UserInputMap;
```
handeyeco, line 165: This is the new way to store user input. It's typed! Yay!

packages/perseus/src/renderer.tsx
```diff
@@ -472,14 +478,35 @@ class Renderer
...
         };
     };
 
+    _getAllWidgetsStartUserInput(props: Props): UserInputMap {
+        const widgetMap = props.widgets;
+        const startUserInput: UserInputMap = {};
+        entries(widgetMap).forEach(([id, widgetInfo]) => {
+            const widgetExports = Widgets.getWidgetExport(widgetInfo.type);
+            if (widgetInfo.static && widgetExports?.getCorrectUserInput) {
+                startUserInput[id] = widgetExports.getCorrectUserInput(
```
handeyeco, line 484: This is how `staticTransform` was setting up user input state.

packages/perseus/src/renderer.tsx
```diff
@@ -472,14 +478,35 @@ class Renderer
...
+                    widgetInfo.options,
+                );
+            } else if (widgetExports?.getStartUserInput) {
+                startUserInput[id] = widgetExports.getStartUserInput(
```
handeyeco, line 488: This is how `transform` was setting up user input state.

packages/perseus/src/renderer.tsx
```diff
@@ -643,6 +676,10 @@ class Renderer
      * If an instance of widgetProps is passed in, it generates the serialized
      * state from that instead of the current widget props.
      */
+    /**
+     * @deprecated and likely a very broken API
```
handeyeco: I'm maintaining `getSerializedState` and `restoreSerializedState` for TX for now. Not fixing it, not removing it. https://khanacademy.atlassian.net/browse/LEMS-3185

At the start of each widget refactor I made a test file for serialization and then did what I needed to do to make those tests pass after the refactor.

packages/perseus/src/renderer.tsx
```diff
@@ -1718,10 +1799,14 @@ class Renderer
         const userInputMap = {};
         this.widgetIds.forEach((id: string) => {
             const widget = this.getWidgetInstance(id);
-            // Handle Groups, which have their own sets of widgets
-            if (widget?.getUserInputMap) {
+            if (this.state.userInput[id]) {
```
handeyeco, line 1783: Ideally at the end of this work, this will be the only case left; widgets won't have `getUserInput` or `getUserInputMap`.

packages/perseus/src/types.ts
```diff
@@ -477,13 +506,14 @@ type WidgetOptions = any;
...
 export type WidgetTransform = (
-    arg1: WidgetOptions,
+    widgetOptions: WidgetOptions,
     strings: PerseusStrings,
     problemNumber?: number,
 ) => any;
 
 export type WidgetExports<
     T extends React.ComponentType<any> & Widget = React.ComponentType<any>,
+    TUserInput = Empty,
```
handeyeco, line 515: This is handy because it semi-automatically populates types for widgets. Pass it in once and it's set for all the user input helpers. I think we need to use this pattern more, it helps narrow types for us.

packages/perseus/src/types.ts
```diff
@@ -582,6 +628,12 @@ export type UniversalWidgetProps<
     findWidgets: (criterion: FilterCriterion) => ReadonlyArray<Widget>;
     reviewMode: boolean;
     onChange: ChangeHandler;
+    handleUserInput: (
+        newUserInput: TUserInput,
+        cb?: any,
+        silent?: boolean,
+    ) => void;
+    userInput: TUserInput;
```
handeyeco, line 636: This is how widgets receive and update user input now: they'll receive `props.userInput` and `props.handleUserInput`.

I like this because it makes it very obvious what we're doing within the widgets.

packages/perseus/src/widgets/categorizer/categorizer.tsx
```diff
@@ -56,17 +56,13 @@ export class Categorizer
         uniqueId: _.uniqueId("perseus_radio_"),
     };
 
-    static getUserInputFromProps(props: Props): PerseusCategorizerUserInput {
```
handeyeco, line 59: `getUserInputFromProps` was mostly used for serialization (which is deprecated) and sometimes in `getUserInput` (which is deprecated). In the individual widget PRs, `getUserInput` now just returns `this.props.userInput`.

`getUserInputFromProps` > `getUserInputFromSerializedState` to make it more obvious.

packages/perseus/src/widgets/definition/definition.tsx
```diff
@@ -105,5 +105,4 @@ export default {
     name: "definition",
     displayName: "Definition",
     widget: Definition,
-    transform: (x: any) => x,
```
handeyeco, line 108: `transform` defaults to `_.identity` when a `transform` isn't provided.

packages/perseus/src/widgets/expression/expression.tsx
```diff
@@ -191,7 +187,7 @@ export class Expression
     };
 
     getUserInput(): PerseusExpressionUserInput {
-        return Expression.getUserInputFromProps(this.props);
+        return normalizeTex(this.props.value);
```
handeyeco, line 190: Expression and Radio are the two widgets that manipulate data in `getUserInput`. Expression is handled in its PR; Radio is still a todo.

testing/test-dependencies.tsx
```diff
@@ -125,7 +125,7 @@ export const storybookDependenciesV2: PerseusDependenciesV2 = {
     ...testDependenciesV2,
     analytics: {
         onAnalyticsEvent: async (event) => {
-            console.log("⚡️ Sending analytics event:", event);
```
handeyeco, line 128: This was just super annoying

packages/perseus/src/widgets/categorizer/categorizer.test.ts
```diff
@@ -12,7 +12,7 @@ import {
 } from "../../util/test-utils";
 import {renderQuestion} from "../__testutils__/renderQuestion";
 
-import {Categorizer} from "./categorizer";
+import CategorizerExport from "./categorizer";
```
handeyeco, line 15: I wouldn't think too much about the individual widgets in this PR. This is mostly cleanup to conform with some general changes to Renderer/types.

The real changes happen in the individual PRs, like for Categorizer: https://github.com/Khan/perseus/pull/2572

testing/test-dependencies.tsx
handeyeco: Maybe we could just call a SB action instead of putting it in the console

packages/perseus/src/components/sortable.tsx
```diff
@@ -647,8 +647,9 @@ class Sortable extends React.Component<SortableProps, SortableState> {
         // @ts-expect-error - TS2551 - Property 'splice' does not exist on type 'readonly SortableItem[]'. Did you mean 'slice'?
         nextItems.splice(index, 0, item);
 
-        this.setState({items: nextItems});
-        this.props.onChange?.({});
+        this.setState({items: nextItems}, () => {
+            this.props.onChange?.({});
```
handeyeco, line 654: **Note to self:** time box changing Sorter to pass state in `onChange`

packages/perseus/src/renderer.tsx
```diff
@@ -1618,16 +1682,32 @@ class Renderer
         );
     }
 
-    _setWidgetProps: SetWidgetPropsFn = (id, newProps, cb, silent) => {
+    _setWidgetProps(
+        id: string,
+        nextWidgetProps: any,
+        nextUserInput: UserInput | null,
```
handeyeco: **Note to self:** Try to fix this so these two aren't sharing a single setter

packages/perseus-core/src/types.ts
```diff
@@ -7,7 +7,15 @@ type State = any;
...
 export interface RendererInterface {
+    /**
+     * @deprecated and likely a very broken API
+     * [LEMS-3185] do not trust serializedState/restoreSerializedState
+     */
     getSerializedState(): State;
+    /**
+     * @deprecated and likely a very broken API
+     * [LEMS-3185] do not trust serializedState/restoreSerializedState
+     */
```
benchristel: I drafted some other inline comments, but I think what they all boil down to is that I'm getting hung up on how these deprecation notices are worded. If this is really `likely a very broken API` why don't we just finish breaking it and be done with it? (Answer: we can't break it, because coach reports are depending on it). If we can't break it yet, we should describe its behavior with characterization tests to protect it from damage as we remodel everything around it.

I also think that if we're going to deprecate an API, we should provide callers with an approved alternative. Otherwise, they're just going to continue to use the deprecated thing, because it's the only option. Maybe an `onUserInput` callback and an `initialUserInput` prop?

In summary, I would rather change these comments to something like

```suggestion
    // TODO(LEMS-3185): remove serializedState/restoreSerializedState
    /**
     * @deprecated - do not use in new code. Pass the `onUserInput` callback prop to `ServerItemRenderer` to receive user input state updates.
     */
    getSerializedState(): State;
    // TODO(LEMS-3185): remove serializedState/restoreSerializedState
    /**
     * @deprecated - do not use in new code. Pass the `initialUserInput` prop to `ServerItemRenderer` to set the initial user input state.
     */
```

But since those alternative APIs don't exist yet, maybe we should just put some TODO comments here and defer the API changes to a later date.

handeyeco: > If we can't break it yet, we should describe its behavior with characterization tests to protect it from damage as we remodel everything around it.

Before refactoring each widget I added tests for this stuff. [Here's Categorizer's](https://github.com/Khan/perseus/pull/2572/files#diff-417231ebfbffb86a32186085405717bd581fd5fd01becd777e43e5e6f2b91ee1). There were also already general serialization tests which didn't break during this refactor.

> I also think that if we're going to deprecate an API, we should provide callers with an approved alternative. 

TX is currently using this, but they're really not using it to its full ability - the alternative for this case would just be to use a PerseusItem and a ServerItemRenderer. I asked Sean DB to file a ticket for this work and to cross reference it with [LEMS-3185](https://khanacademy.atlassian.net/browse/LEMS-3185).

If the wording is bad though, I can change it. I just wanted to emphasis the danger of using the serialized state.

benchristel: Thanks for the context! I guess I'd prefer

```ts
// TODO(LEMS-3185): remove serializedState/restoreSerializedState
/**
 * @deprecated - do not use in new code.
 */
```

in that case.

packages/perseus/src/types.ts
```diff
@@ -582,6 +628,12 @@ export type UniversalWidgetProps<
     findWidgets: (criterion: FilterCriterion) => ReadonlyArray<Widget>;
     reviewMode: boolean;
     onChange: ChangeHandler;
+    handleUserInput: (
```
benchristel, line 631: Is this an event callback? If so, the name should conventionally be `onUserInput`.

```suggestion
    onUserInput: (
```

packages/perseus/src/types.ts
```diff
@@ -582,6 +628,12 @@ export type UniversalWidgetProps<
...
+        newUserInput: TUserInput,
+        cb?: any,
+        silent?: boolean,
```
benchristel: What are `cb` and `silent` for? Does `cb` really need to be type `any`?

It would be good to have doc comments explaining what these parameters do, for the benefit of future widget authors. Why would a widget want to set `cb` or `silent`?

packages/perseus/src/components/sortable.tsx
handeyeco: I tried for a little bit but my changes caused some jitter that I'm not sure how to handle. https://khanacademy.atlassian.net/browse/LEMS-3224

packages/perseus/src/mixins/widget-prop-denylist.ts
benchristel: I'm not sure I understand the impact of this change. Is it related to separating user input from widget options? Or just general cleanup?

packages/perseus/src/renderer.tsx
```diff
@@ -1637,18 +1693,17 @@ class Renderer
...
-                    widgetProps,
                 };
             },
             () => {
+                if (!silent) {
+                    this.props.onSerializedStateUpdated(
+                        this.getSerializedState(this.state.widgetProps),
+                    );
+                }
+
```
benchristel: How did you figure out that this call to `onSerializedStateUpdated` needed to move? Or is this more of a general React code hygiene change?

packages/perseus/src/renderer.tsx
```diff
@@ -629,6 +659,9 @@ class Renderer
             onChange: (newProps, cb, silent = false) => {
                 this._setWidgetProps(widgetId, newProps, cb, silent);
             },
+            handleUserInput: (newUserInput, cb, silent = false) => {
```
benchristel: GoLand is telling me `newUserInput` and `cb` implicitly have an `any` type. I'm not 100% sure that's true, but we have `noImplicitAny` turned off in our tsconfig, so it's possible. Can you add types here?

packages/perseus/src/renderer.tsx
```diff
@@ -1677,7 +1732,61 @@ class Renderer
                 }, 0);
             },
         );
-    };
+    }
+
+    _setUserInput(
+        id: string,
+        nextUserInput: UserInput | null,
```
benchristel: I don't see a way for `UserInput` to be null. If we made this type non-nullable, we could simplify some code (the ternary below).

packages/perseus/src/renderer.tsx
```diff
@@ -701,32 +744,48 @@ class Renderer
...
+            if (widget?.restoreSerializedState) {
+                // Note that we probably can't call
+                // `this.change()/this.props.onChange()` in this
+                // function, so we take the return value and use
+                // that as props if necessary so that
+                // `restoreSerializedState` in a widget can
+                // change the props as well as state.
+                // If a widget has no props to change, it can
+                // safely return null.
+                ++numCallbacks;
```
benchristel, line 756: If a widget has no `restoreSerializedState` method, but exports `getUserInputFromSerializedState`, we'll `setState` but won't call the callback. Is that a problem?

packages/perseus/src/renderer.tsx
```diff
@@ -629,6 +659,9 @@ class Renderer
             onChange: (newProps, cb, silent = false) => {
                 this._setWidgetProps(widgetId, newProps, cb, silent);
             },
```
benchristel, line 648: Do we still want `onChange`? It seems like any change to a widget is user input.

If not all changes are user input, maybe what we really want is to store widget _state_ separately from props/configuration, and compute user input from the state on demand.

packages/perseus/src/mixins/widget-prop-denylist.ts
handeyeco: It's related to the serialization tests. As I added tests I found a bunch of things we shouldn't have been serializing so I added them to the deny list.

packages/perseus/src/renderer.tsx
handeyeco: It was calling `onSerializedStateUpdated` with `getSerializedState` before the state was actually set, which was working because it had access to the latest computed state. But since I wanted to split up the state setters but wanted to keep the shared logic...shared...it needed access to the new widget state in a different way.

packages/perseus/src/renderer.tsx
handeyeco: Should be fine. `restoreSerializedState` seems to be a Group thing which probably won't be needed in the near future.

packages/perseus/src/renderer.tsx
handeyeco: It's still being used for some general UI state. My hunch is that we'll want to move that state into the widgets and then we won't need `onChange` anymore.

packages/perseus/src/renderer.tsx
handeyeco: That was left over from before I tweaked some stuff. Will fix!

packages/perseus/src/renderer.tsx
benchristel: If it's just internal UI state, why would we want to call `this._setWidgetProps()` in `onChange`? It seems like we might want to get rid of this now to avoid polluting the props data.

I'm probably missing something.

packages/perseus/src/renderer.tsx
benchristel: I dug a bit deeper on this... and I don't see any usages of `onSerializedStateUpdated` in the frontend repo. In fact, the only file that mentions it across all our codebases is this one, renderer.tsx.

Could we just delete it?

packages/perseus/src/types.ts
handeyeco: I'm honestly not sure what `cb` and `silent` are for. I think it's partially used to track interactions (and `silent` signifies we updated state but it wasn't a real interaction), but I'm not sure if that's all its for. This is just behavior I chose to move over from `onChange` because otherwise tests failed. I can put a note in LEMS Futures as a tech debt task, but this stuff didn't originate in this PR.

packages/perseus/src/renderer.tsx
handeyeco: > It seems like we might want to get rid of this now to avoid polluting the props data.

My plan is that I'll likely do that as a next step because I think I'll have to; I wanted to keep these PRs as focused as possible. I talk about this in the ticket.

packages/perseus/src/__tests__/test-utils.tsx
```diff
@@ -0,0 +1,49 @@
...
+    // eslint-disable-next-line import/no-relative-packages
+} from "../../../../testing/test-dependencies";
+import WrappedServerItemRenderer from "../server-item-renderer";
+
+import type {ServerItemRenderer} from "../server-item-renderer";
+import type {APIOptions} from "../types";
+import type {PerseusItem} from "@khanacademy/perseus-core";
+import type {PropsFor} from "@khanacademy/wonder-blocks-core";
+
+// This looks alot like `widgets/__tests__/renderQuestion.jsx', except we use
```
jeremywiebe: Nit that I noticed... 

```suggestion
// This looks alot like `widgets/__tests__/renderQuestion.tsx', except we use
```
```suggestion
// This looks alot like `widgets/__tests__/renderQuestion.jsx', except we use
```

packages/perseus/src/mixins/widget-prop-denylist.ts
jeremywiebe: We find some of these (or all? I haven't checked thoroughly) in our serialized widget data _in production_ when they really aren't part of widget options.  For example, you can find `linterContext` all over in our published assessment items. 🙃

packages/perseus/src/mixins/widget-prop-denylist.ts
```diff
@@ -34,6 +37,29 @@ const denylist = [
...
 ];
 
+export function excludeDenylistKeys(obj: Record<any, any>) {
+    if (!(obj as any)) {
```
jeremywiebe: Can you remove the cast if you type the function as taking an optional `obj`?

```suggestion
export function excludeDenylistKeys(obj: Record<string, any>) {
    if (obj == null) {
```

packages/perseus-core/src/validation.types.ts
```diff
@@ -289,7 +289,6 @@ export interface RubricRegistry {
     "graded-group": PerseusGradedGroupRubric;
     grapher: PerseusGrapherRubric;
     group: PerseusGroupRubric;
-    image: PerseusLabelImageRubric;
```
SonicScrewdriver, line 292: Oh interesting, we had this doubled for some reason?

packages/perseus/src/__tests__/test-utils.tsx
```diff
@@ -0,0 +1,49 @@
...
+// the ServerItemRenderer instead of Renderer
+export const renderQuestion = (
```
SonicScrewdriver, line 19: Love it, thank you. This is a much clearer location for this

packages/perseus/src/renderer.tsx
```diff
@@ -655,16 +683,22 @@ class Renderer
             (props, widgetId) => {
                 const widget = this.getWidgetInstance(widgetId);
                 if (widget && widget.getSerializedState) {
-                    return widget.getSerializedState();
+                    return excludeDenylistKeys(widget.getSerializedState());
```
SonicScrewdriver, line 686: Weird! So we had that deny list originally and we weren't even using it? 

Thanks for improving this!

packages/perseus/src/renderer.tsx
```diff
@@ -701,32 +735,48 @@ class Renderer
             }
         };
 
+        const restoredWidgetProps = {};
+        const restoredUserInput = {};
+        Object.entries(serializedState).forEach(([widgetId, props]) => {
```
SonicScrewdriver, line 740: This is much easier to read!

packages/perseus/src/widgets.ts
```diff
@@ -162,7 +162,12 @@ export const getVersionVector = (): {
 
 export const getPublicWidgets = (): Record<string, WidgetExports> => {
     return widgets.entries().reduce((acc, [key, value]) => {
-        if (!value.hidden) {
+        /**
+         * Even though we don't want content creators adding new "hidden" widgets,
+         * we still have to maintain editors for hidden widgets in order to support
+         * old content. So this lets us use hidden widgets in Storybook.
+         */
+        if (process.env.STORYBOOK || !value.hidden) {
```
SonicScrewdriver, line 170: Greatly appreciate this, and the commenting! It would be nice to have a way to swap the storybook view between "all" and "public" modes while testing the editors, but that's something we can worry about later if it comes up.

packages/perseus/src/widgets/numeric-input/numeric-input.class.tsx
```diff
@@ -65,9 +68,10 @@ type RenderProps = {
 // defaultProps into account, which is important because
 // PerseusNumericInputWidgetOptions has optional fields which receive defaults
 // via defaultProps.
-0 as any as WidgetProps<PerseusNumericInputWidgetOptions> satisfies PropsFor<
-    typeof NumericInput
->;
+0 as any as WidgetProps<
```
jeremywiebe, line 71: Unrelated to your changes here, but the comment above references the `matrix` widget options... guessing copy-pasta that we might as well fix while we're in the area. 

image

packages/perseus/src/widgets/explanation/explanation.tsx
```diff
@@ -16,7 +16,7 @@ import type {Widget, WidgetExports, WidgetProps} from "../../types";
 import type {ExplanationPromptJSON} from "../../widget-ai-utils/explanation/explanation-ai-utils";
 import type {PerseusExplanationWidgetOptions} from "@khanacademy/perseus-core";
 
-type RenderProps = PerseusExplanationWidgetOptions; // transform = _.identity
+type RenderProps = PerseusExplanationWidgetOptions;
 
 type Props = WidgetProps<RenderProps>;
```
jeremywiebe, line 21: I don't _think_ we need to tackle this in this PR, but I think we will have the opportunity to simplify our prop types somewhat after you get this all worked through. 

Previously, the `transform` was the reason we needed different types from `RenderProps` and `PerseusExplanationWidgetOptions` (transform was `PerseusExplanationWidgetOptions` => `RenderProps`). 

With `transform` and `staticTransform` going away, I _think_ we'll be at a place where our widgets accept `WidgetProps<PerseusExplanationWidgetOptions>` (always?).

benchristel:
> @handeyeco A thought: is this going to be compatible with dynamically-registered widgets that are designed to work with the current Renderer? Or are we pretty sure those don't actually exist?

I don't see any calls to `registerWidget` outside Perseus, so I think this is safe.

packages/perseus-core/src/validation.types.ts
handeyeco: I think this was just a mistake. We don't score Images so I don't know why its part of validation types.

packages/perseus/src/renderer.tsx
handeyeco: We were calling it for the editing experience back when we used serialization for that, but I guess not for this kind of serialization.

PR #2568, opened 2025-06-09
Dropdown: renderer user input state.

handeyeco:
## Summary:

---

DropdownEditor doesn't use Dropdown, so this shouldn't affect the editing experience

### DIFF
--- a/packages/perseus/src/widgets/dropdown/serialize-dropdown.test.ts
+++ b/packages/perseus/src/widgets/dropdown/serialize-dropdown.test.ts
@@ -72,14 +72,6 @@ describe("Dropdown serialization", () => {
         );
     });
 
-    afterEach(() => {
-        // The Renderer uses a timer to wait for widgets to complete rendering.
-        // If we don't spin the timers here, then the timer fires in the test
-        // _after_ and breaks it because we do setState() in the callback,
-        // and by that point the component has been unmounted.
-        act(() => jest.runOnlyPendingTimers());
-    });
-
     it("should serialize the current state", async () => {
         // Arrange
         const {renderer} = renderQuestion(generateBasicDropdown());


packages/perseus/src/widgets/dropdown/serialize-dropdown.test.ts
```diff
@@ -0,0 +1,134 @@
...
+            testDependencies,
+        );
+    });
+
+    afterEach(() => {
+        // The Renderer uses a timer to wait for widgets to complete rendering.
+        // If we don't spin the timers here, then the timer fires in the test
+        // _after_ and breaks it because we do setState() in the callback,
+        // and by that point the component has been unmounted.
+        act(() => jest.runOnlyPendingTimers());
```
benchristel: I'm curious — is there a reason not to `runAllTimers()` here? https://jestjs.io/docs/timer-mocks#run-all-timers

I would hope we don't have a timer loop.

packages/perseus/src/widgets/dropdown/serialize-dropdown.test.ts
```diff
@@ -0,0 +1,134 @@
...
+ * doesn't use WidgetOptions, but RenderProps; it's leveraging
+ * what is considered an internal implementation detail to support
+ * rehydrating previous state.
+ *
+ * The API is very fragile and likely broken. We have a ticket to remove it.
+ * However we don't have the bandwidth to implement an alternative right now,
+ * so I'm adding tests to make sure we're roughly still able to support
+ * what little we've been supporting so far.
+ *
+ * This API needs to be removed and these tests need to be removed with it.
```
jeremywiebe, line 30: Thanks for this very clear explanation!

packages/perseus/src/widgets/dropdown/serialize-dropdown.test.ts
```diff
@@ -0,0 +1,134 @@
...
+                },
+                hints: [],
+            }),
+        );
+
+        const userInput = renderer.getUserInput();
+
+        // Assert
+        // `value` would be 0 if we didn't properly restore serialized state
+        expect(userInput).toEqual({"dropdown 1": {value: 2}});
```
jeremywiebe, line 124: 🤔  Non-blocking, but I thought I'd mention this. Given that RTL is all about testing user-facing behaviour, would there be value in also asserting that the Dropdown has the restored option selected?

jeremywiebe (approved):

packages/perseus/src/widgets/dropdown/serialize-dropdown.test.ts
handeyeco: I gave this a shot but I think maybe WB `option` isn't actually a select/option element, so `.selected` doesn't exist on the option. I'm honestly not too worried about it though; I think if this were a critical API I'd agree that your solution is the correct way, but I think this might be good enough for a deprecated API.

PR #2569, opened 2025-06-09
Expression: renderer user input state 

handeyeco:
## Summary:

---

ExpressionEditor does use Expression, so this will affect the editing experience

### DIFF
--- a/packages/perseus-editor/src/widgets/expression-editor.tsx
+++ b/packages/perseus-editor/src/widgets/expression-editor.tsx
@@ -302,8 +302,8 @@ class ExpressionEditor extends React.Component<Props, State> {
                     userInput: ans.value,
                     // TODO: UniversalWidgetProps should have a generic type arg
                     // to help scope what user input a widget consumes
-                    handleUserInput: ((input: PerseusExpressionUserInput) =>
-                        this.changeExpressionWidget(index, input)) as any,
+                    handleUserInput: (input: PerseusExpressionUserInput) =>
+                        this.changeExpressionWidget(index, input),
                     trackInteraction: () => {},
                     widgetId: this.props.widgetId + "-" + ans.key,
                     visibleLabel: this.props.visibleLabel,


packages/perseus-editor/src/widgets/expression-editor.tsx
```diff
@@ -276,43 +277,44 @@ class ExpressionEditor extends React.Component<Props, State> {
...
+                    // answer area has the same settings for those
+                    times: this.props.times,
+                    functions: this.props.functions,
+                    buttonSets: this.props.buttonSets,
+                    buttonsVisible: "focused",
+                    userInput: ans.value,
+                    // TODO: UniversalWidgetProps should have a generic type arg
+                    // to help scope what user input a widget consumes
+                    handleUserInput: ((input: PerseusExpressionUserInput) =>
+                        this.changeExpressionWidget(index, input)) as any,
```
Myranae: Is there a way to not use the casting to any here and on line 317? I'm still not the best at knowing when this is the right decision, so just wanted to check. I just try to avoid casting in general. XD

packages/perseus/src/widgets/expression/expression.tsx
```diff
@@ -186,18 +184,18 @@ export class Expression
...
+        return normalizeTex(this.props.userInput);
     }
 
     getPromptJSON(): ExpressionPromptJSON {
         return _getPromptJSON(this.props, this.getUserInput());
     }
 
-    change: (...args: any) => any | undefined = (...args: any) => {
-        return Changeable.change.apply(this, args);
-    };
```
Myranae, line 199: Does this get removed because you are no longer using onChange?

packages/perseus-editor/src/widgets/expression-editor.tsx
```diff
@@ -276,43 +277,44 @@ class ExpressionEditor extends React.Component<Props, State> {
...
+                    visibleLabel: this.props.visibleLabel,
+                    ariaLabel: this.props.ariaLabel,
+                } as const;
 
                 return (
                     <AnswerOption
                         key={ans.key}
                         considered={ans.considered}
-                        expressionProps={expressionProps}
+                        expressionProps={expressionProps as any}
```
SonicScrewdriver, line 317: This any makes me sad. :(

packages/perseus/src/widgets/expression/expression.tsx
```diff
@@ -14,7 +14,6 @@ import _ from "underscore";
 import {PerseusI18nContext} from "../../components/i18n-context";
 import MathInput from "../../components/math-input";
 import {useDependencies} from "../../dependencies";
-import * as Changeable from "../../mixins/changeable";
```
SonicScrewdriver, line 17: Yesss, more deletion of this darned mixin

SonicScrewdriver (approved):
I think this all makes sense, considering we chatted separately about the Partial / casting requirements to help remove some over-zealous ts-overrides.

I'd love to avoid any wherever possible, but I also understand that this might be easier to tackle in future work.

packages/perseus/src/widgets/expression/expression.tsx
handeyeco: Right, for user input `onChange` has been replaced with `handleUserInput`. For widgets that were only using `onChange` for user input, they no longer need that function/mixin.

packages/perseus-editor/src/widgets/expression-editor.tsx
handeyeco: Oh, lol, I thought you were talking about line 306 which it turns out I didn't need to cast for.

I don't know what to do about the cast on 317. We're expecting props for an Expression widget but the editor doesn't provide all of those props. This happens in a couple of places in the editors because the widgets in the editors don't need all of the stuff they need in the learner experience. For the most part this worked because of poor type coverage or use of `ts-expect-error` which is much worse than type casting.

I think this is still an improvement than what was here before and I think further work on this is beyond the scope of my current work.

packages/perseus-editor/src/widgets/expression-editor.tsx
handeyeco: From my response to Tamara:

> I don't know what to do about the cast on 317. We're expecting props for an Expression widget but the editor doesn't provide all of those props. This happens in a couple of places in the editors because the widgets in the editors don't need all of the stuff they need in the learner experience. For the most part this worked because of poor type coverage or use of `ts-expect-error` which is much worse than type casting.
> 
> I think this is still an improvement than what was here before and I think further work on this is beyond the scope of my current work.

PR #2570, opened 2025-06-09
Update comments to refer to khan/frontend instead of webapp

benchristel:
Over the weekend, the frontend-infra team moved all of [webapp]'s frontend code to a new repo: https://github.com/khan/frontend. This PR updates Perseus to match.

[webapp]: https://github.com/khan/webapp

Issue: none

## Test plan:

no testing needed

### DIFF
--- a/.changeset/moody-birds-marry.md
+++ b/.changeset/moody-birds-marry.md
@@ -1,0 +1,8 @@
+---
+"@khanacademy/math-input": patch
+"@khanacademy/perseus": patch
+"@khanacademy/perseus-core": patch
+"@khanacademy/perseus-editor": patch
+---
+
+Internal: Update comments to refer to khan/frontend instead of webapp. Khan Academy's frontend code has moved.


packages/math-input/src/components/input/scroll-into-view.ts
```diff
@@ -6,12 +6,12 @@
...
  * TODO(charlie): Move this scroll logic out of our components and into a higher
- * level in the component tree--perhaps even into webapp, beyond Perseus.
+ * level in the component tree--perhaps even into khan/frontend, beyond Perseus.
  */
 
 // HACK(charlie): This should be injected by webapp somehow.
-// TODO(charlie): Add a link to the webapp location as soon as the footer
-// has settled down.
+// TODO(charlie): Add a link to the khan/frontend location as soon as the
+// footer has settled down.
```
benchristel, line 14: I suspect this TODO comment is not relevant anymore, but couldn't prove it within a few minutes.

packages/perseus/src/widgets/explanation/explanation.test.ts
```diff
@@ -198,7 +198,6 @@ describe("Explanation", function () {
     });
 
     it("does NOT use transitions when matchMedia is not available", async () => {
-        // e.g. during SSR in webapp
```
benchristel, line 201: We don't do server-side rendering anymore.

jeremywiebe (approved):
Thanks. I'm always a bit on the fence about having any references to a repo that is private from this public repo. But I don't think we're leaking any private/sensitive info so I think it's fine.

PR #2572, opened 2025-06-10
Categorizer: renderer user input state

handeyeco:
## Summary:
<Add your summary here>

Issue: LEMS-XXXX

## Test plan:

### DIFF
--- a/packages/perseus/src/widgets/categorizer/serialize-categorizer.test.ts
+++ b/packages/perseus/src/widgets/categorizer/serialize-categorizer.test.ts
@@ -1,0 +1,124 @@
+import {
+    generateTestPerseusItem,
+    generateTestPerseusRenderer,
+} from "@khanacademy/perseus-core";
+import {screen, act} from "@testing-library/react";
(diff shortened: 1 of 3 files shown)


packages/perseus/src/widgets/categorizer/serialize-categorizer.test.ts
```diff
@@ -0,0 +1,124 @@
...
+        const state = renderer.getSerializedState();
+
+        // Assert
+        expect(state).toEqual({
+            question: {
+                "categorizer 1": {
+                    categories: ["one", "two", "three"],
+                    items: ["uno", "dos", "tres"],
+                    randomizeItems: true,
+                    values: [undefined, 0],
```
mark-fitzgerald, line 97: The `values` item is somewhat unintuitive. Do we want a comment here to indicate that clicking on the items does, or what the shape of the array represents? For instance, not sure why there is an undefined entry in the array.

handeyeco: I agree it's unintuitive. I added some comments to the test.

PR #2573, opened 2025-06-10
NumberLine: renderer user input state

handeyeco:
## Summary:

---

NumberLineEditor doesn't use NumberLine, so this shouldn't affect the editing experience

### DIFF
--- a/.changeset/proud-cycles-kick.md
+++ b/.changeset/proud-cycles-kick.md
@@ -1,0 +1,7 @@
+---
+"@khanacademy/perseus": patch
+"@khanacademy/perseus-core": patch
+"@khanacademy/perseus-score": patch
+---
+
+NumberLine: use userInput/handleUserInput


handeyeco:
Added TUT/Mike since this touches AI tools

packages/perseus/src/widget-ai-utils/number-line/number-line-ai-utils.ts
```diff
@@ -23,12 +23,12 @@ export const getPromptJSON = (
         type: "number-line",
         options: {
             range: renderProps.range,
-            numDivisions: renderProps.numDivisions,
             snapDivisions: renderProps.snapDivisions,
         },
         userInput: {
             numLinePosition: userInput.numLinePosition,
             numDivisions: userInput.numDivisions,
+            rel: userInput.rel,
```
MikeKlemarewski, line 31: I'm not familiar with the terminology for number lines. Does this control whether the dot is inclusive or exclusive? Is `rel` short for something? I'm just wondering if an llm will understand what `rel` means in the context of a number line.

packages/perseus-core/src/validation.types.ts
```diff
@@ -205,14 +205,14 @@ export type PerseusNumberLineRubric = {
     range: number[];
     initialX: number | null | undefined;
     isInequality: boolean;
+    isTickCtrl?: boolean;
+    divisionRange: number[];
 };
```
SonicScrewdriver, line 210: This makes much more sense, I'm not sure why it was on the UserInput

packages/perseus-score/src/widgets/number-line/score-number-line.test.ts
```diff
@@ -6,13 +6,38 @@ import type {
 } from "@khanacademy/perseus-core";
 
 describe("scoreNumberLine", () => {
+    it("is invalid when outside allowed range", () => {
```
SonicScrewdriver, line 9: Great test addition! Thank you

packages/perseus-score/src/widgets/number-line/score-number-line.ts
```diff
@@ -10,6 +10,18 @@ function scoreNumberLine(
...
+        userInput.numDivisions > divisionRange[1] ||
+        userInput.numDivisions < divisionRange[0];
+
+    if (rubric.isTickCtrl && outsideAllowedRange) {
+        return {
+            type: "invalid",
+            message: "Number of divisions is outside the allowed range.",
+        };
+    }
+
```
SonicScrewdriver, line 24: This makes much more sense living over here, like the rest of our scoring / error logic, imo! 

I do wonder, though, about what our specific requirements are for using a separate validate file vs inline-score logic.

SonicScrewdriver (approved):
Looks great to me, thanks for cleaning so much of this up again!

packages/perseus/src/widget-ai-utils/number-line/number-line-ai-utils.ts
handeyeco: It stands for "Relationship" (equal, less than, greater than, less than or equal, greater than or equal): `export type Relationship = "eq" | "lt" | "gt" | "le" | "ge";`

I think this has always been part of the data, not sure why it wasn't part of the prompt...

PR #2574, opened 2025-06-10
Ensure that we properly parse any decimals in the userInput's according to the user's locale

SonicScrewdriver:
## Summary:
This PR fixes a bug where users would be erroneously marked correct if they entered `16,000` when the answer is `16`.

The issue was that we weren't properly handling locale-specific decimal separators in the Numeric Input and Input Number Widgets. While `scoreExpression` was already passing the locale to get the appropriate decimal separator (albeit with KAS), the other two widgets were missing this functionality.

Changes made:
1. Added locale parameter to `scoreInputNumber` and  `scoreNumericInput` functions
2. Added logic to pass the locale-specific decimal separator to the validators
3. Added tests to verify correct behavior with different locales

This ensures consistent behavior across all three scoring functions and prevents incorrect grading when users enter numbers with thousands separators or decimal separators that don't match their locale.

Issue: LEMS-3197

## Test plan:
- New tests pass

### DIFF
--- a/packages/perseus-score/src/util/answer-types.ts
+++ b/packages/perseus-score/src/util/answer-types.ts
@@ -223,10 +223,10 @@ const KhanAnswerTypes = {
                     const decimal = forms.decimal(text);
                     const rounded = forms.decimal(text, 1);
                     if (
-                        (decimal[0].value != null &&
-                            decimal[0].value === rounded[0].value) ||
-                        (decimal[1].value != null &&
-                            decimal[1].value === rounded[1].value)
+                        (decimal[0]?.value != null &&
+                            decimal[0].value === rounded[0]?.value) ||
+                        (decimal[1]?.value != null &&
+                            decimal[1].value === rounded[1]?.value)
                     ) {
                         return decimal;
                     }
@@ -540,10 +540,14 @@ const KhanAnswerTypes = {
                         return normal(text);
                     };
 
-                    return [
-                        {value: normal(text), exact: true},
-                        {value: commas(text), exact: true},
-                    ];
+                    const results = [{value: normal(text), exact: true}];
+
+                    // Only include the comma interpretation if the locale uses commas as decimal separators
+                    if (options.decimal_separator === ",") {
+                        results.push({value: commas(text), exact: true});
+                    }
+
+                    return results;
                 },
             } as const;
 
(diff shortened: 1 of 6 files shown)


packages/perseus-score/src/widgets/numeric-input/score-numeric-input.ts
```diff
@@ -95,6 +98,8 @@ function scoreNumericInput(
             inexact: true, // TODO(merlob) backfill / delete
             maxError: answer.maxError,
             forms: validatorForms,
+            // Pass locale-specific decimal separator like expression scorer does
+            ...(locale && {decimal_separator: getDecimalSeparator(locale)}),
```
SonicScrewdriver, line 103: We'd need to add this to expression and possibly input number as well.

SonicScrewdriver: Expression already had logic for this, as it uses KAS. I've added it to Input Number, however.

packages/perseus-score/src/widgets/input-number/score-input-number.test.ts
```diff
@@ -126,4 +126,61 @@ describe("scoreInputNumber", () => {
...
+
+        const userInput: PerseusInputNumberUserInput = {
+            currentValue: "16,5",
+        };
+
+        // Using French locale where comma is decimal separator
+        const score = scoreInputNumber(userInput, rubric, "fr");
+
+        expect(score).toHaveBeenAnsweredCorrectly();
+    });
```
nishasy, line 185: Would it make sense to add a test for this rejecting 16.500 in the French locale?

packages/perseus-score/src/util/answer-types.ts
```diff
@@ -223,10 +223,10 @@ const KhanAnswerTypes = {
                     const decimal = forms.decimal(text);
                     const rounded = forms.decimal(text, 1);
                     if (
-                        (decimal[0].value != null &&
-                            decimal[0].value === rounded[0].value) ||
-                        (decimal[1].value != null &&
-                            decimal[1].value === rounded[1].value)
+                        (decimal[0]?.value != null &&
+                            decimal[0].value === rounded[0]?.value) ||
+                        (decimal[1]?.value != null &&
```
nishasy, line 230: Can you add a comment for confused beans like me explaining that this would be undefined in the case that the locale doesn't use comma decimals?

SonicScrewdriver: That's a great call, thank you! I'll do that

PR #2577, opened 2025-06-11
LabelImage: renderer user input state

handeyeco:
## Summary:

---

LabelImageEditor doesn't use LabelImage, so this shouldn't affect the editing experience

### DIFF
--- a/packages/perseus/src/widgets/label-image/__tests__/serialize-label-image.test.ts
+++ b/packages/perseus/src/widgets/label-image/__tests__/serialize-label-image.test.ts
@@ -1,0 +1,226 @@
+import {
+    generateTestPerseusItem,
+    generateTestPerseusRenderer,
+} from "@khanacademy/perseus-core";
+import {screen, act} from "@testing-library/react";
+import {userEvent as userEventLib} from "@testing-library/user-event";
+
+import {testDependencies} from "../../../../../../testing/test-dependencies";
+import {renderQuestion} from "../../../__tests__/test-utils";
+import * as Dependencies from "../../../dependencies";
+import {registerAllWidgetsForTesting} from "../../../util/register-all-widgets-for-testing";
+
+import type {PerseusItem} from "@khanacademy/perseus-core";
+import type {UserEvent} from "@testing-library/user-event";
+
+/**
+ * [LEMS-3185] These are tests for the legacy Serialization API.
+ *
+ * This API is not built in a way that supports migrating data
+ * between versions of Perseus JSON. In fact serialization
+ * doesn't use WidgetOptions, but RenderProps; it's leveraging
+ * what is considered an internal implementation detail to support
+ * rehydrating previous state.
+ *
+ * The API is very fragile and likely broken. We have a ticket to remove it.
+ * However we don't have the bandwidth to implement an alternative right now,
+ * so I'm adding tests to make sure we're roughly still able to support
+ * what little we've been supporting so far.
+ *
+ * This API needs to be removed and these tests need to be removed with it.
+ */
+describe("LabelImage serialization", () => {
+    function generateBasicLabelImage(): PerseusItem {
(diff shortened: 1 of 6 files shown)


packages/perseus/src/widgets/label-image/__tests__/serialize-label-image.test.ts
```diff
@@ -0,0 +1,226 @@
...
+ * rehydrating previous state.
+ *
+ * The API is very fragile and likely broken. We have a ticket to remove it.
+ * However we don't have the bandwidth to implement an alternative right now,
+ * so I'm adding tests to make sure we're roughly still able to support
+ * what little we've been supporting so far.
+ *
+ * This API needs to be removed and these tests need to be removed with it.
+ */
+describe("LabelImage serialization", () => {
```
SonicScrewdriver, line 32: Thanks for the comments here, I was wondering why we were adding tests for a deprecated method

packages/perseus/src/widgets/label-image/label-image.tsx
```diff
@@ -356,36 +359,48 @@ export class LabelImage
...
+     * @deprecated get user input from Renderer state
+     */
     getUserInput(): PerseusLabelImageUserInput {
-        return {
-            markers: this.props.markers.map((marker) => ({
-                selected: marker.selected,
-                label: marker.label,
-            })),
-        };
+        return this.props.userInput;
```
SonicScrewdriver, line 367: ooo this is nice to see in action.

packages/perseus/src/widgets/label-image/label-image.tsx
```diff
@@ -356,36 +359,48 @@ export class LabelImage
...
+        const updatedUserInput = [
+            ...userInput.markers.slice(0, index),
+            {
+                label: marker.label,
+                selected: marker.selected,
+            },
+            ...userInput.markers.slice(index + 1),
+        ];
+
+        onChange({markers: updatedRenderProps});
```
SonicScrewdriver, line 402: I remember seeing [your comment in the base PR](https://github.com/Khan/perseus/pull/2566#discussion_r2152618860) that we're hoping to remove the onChange eventually, which would definitely be nice here.

packages/perseus/src/widgets/label-image/label-image.tsx
```diff
@@ -677,6 +697,21 @@ export class LabelImage
         );
     }
 
+    /**
+     * @deprecated and likely very broken API
+     * [LEMS-3185] do not trust serializedState/restoreSerializedState
+     */
+    getSerializedState(): any {
```
SonicScrewdriver, line 704: Oh fun, we didn't even have a `getSerializedState` here before?

packages/perseus/src/widgets/label-image/label-image.tsx
```diff
@@ -677,6 +697,21 @@ export class LabelImage
...
+        const {userInput, markers, ...rest} = this.props;
+        return {
+            ...rest,
+            markers: markers.map((m, i) => ({
+                ...m,
```
SonicScrewdriver: Totally ignorable nit: I've never been a fan of single letter variables, but this map is small enough that it probably doesn't matter.

packages/perseus/src/widgets/label-image/label-image.tsx
handeyeco: Right, originally Renderer would call `getSerializedState` or just use its stash of RenderProps for serializing. Since I'm splitting UserInput from RenderProps, we can't default to just using RenderProps anymore; so for some of these widgets I needed to add this function to merge RenderProps with UserInput for backwards compatibility.

packages/perseus/src/widgets/label-image/label-image.tsx
handeyeco: You're probably right, changed.

PR #2578, opened 2025-06-11
Sorter: renderer user input state

handeyeco:
## Summary:

---

SorterEditor doesn't use Sorter, so this shouldn't affect the editing experience

### DIFF
--- a/packages/perseus/src/components/__tests__/sorter.test.tsx
+++ b/packages/perseus/src/components/__tests__/sorter.test.tsx
@@ -14,6 +14,7 @@ import {question1} from "../__testdata__/sorter.testdata";
 
 import type {APIOptions} from "../../types";
 
+// TODO, this is in the wrong spot; it's in the "components" dir and not the "widgets" dir
 describe("sorter widget", () => {
     beforeEach(() => {
         /*
@@ -93,9 +94,10 @@ describe("sorter widget", () => {
             const {renderer} = renderQuestion(question);
 
             // Act
+            const userInput = renderer.getUserInputMap();
             const score = scorePerseusItemTesting(
                 answerfulItem.question,
-                renderer.getUserInputMap(),
+                userInput,
             );
 
             // Assert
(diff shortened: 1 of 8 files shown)


packages/perseus/src/components/__tests__/sorter.test.tsx
```diff
@@ -110,14 +112,15 @@ describe("sorter widget", () => {
             // Act
             // Put the options in the correct order
             ["Zeroth", "First", "Second", "Third", "Fourth"].forEach(
-                (option) => {
-                    act(() => sorter.moveOptionToIndex(option, 4));
+                (option, index) => {
+                    act(() => sorter.moveOptionToIndex(option, index));
```
nishasy, line 116: Is this doing anything? Wouldn't moving something from its starting index to the same index make it so that no movement is happening?

If everything is moved to the last index, then I think it results in the same thing by pushing everything to the end one by one, so I _think_ it makes sense in my brain to keep it the way it was?

Although the more I think about it, the more I'm wondering what the point of this test is, when the test directly below it tests correctness the way I'd expect.

nishasy (approved):
Looks good to me, but I don't have any experience with the sorter widget. Another set of eyes might be a good idea.

packages/perseus/src/widgets/sorter/sorter.tsx
```diff
@@ -57,28 +48,32 @@ class Sorter extends React.Component<Props, State> implements Widget {
...
-            // finished rendering.
-            // TODO(jeff, CP-3128): Use Wonder Blocks Timing API
-            // eslint-disable-next-line no-restricted-syntax
-            setTimeout(() => {
-                this.props.onChange(e as any);
-                this.props.trackInteraction();
-            }, 0);
+        this.props.handleUserInput({
+            options: this._getOptionsFromSortable(),
+            changed: true,
```
SonicScrewdriver, line 53: Nice to get rid of the setTimeout!

SonicScrewdriver (approved):
I think this all makes sense!

packages/perseus/src/components/__tests__/sorter.test.tsx
handeyeco: When we render, Sorter shuffles the cards. The array we're mapping over is the correct order, so I'm taking the shuffled card and putting it in the right position. The original behavior was doing the same thing but in a weirder way (rather than putting the card in its correct position, it was putting the cards in the correct order by putting each one at the end).

PR #2579, opened 2025-06-11
Fire onRendered callback at least once

jeremywiebe:
## Summary:

The ServerItemRenderer (and the `<LoadingContext>`) call an `onRendered(true)` prop when all assets are loaded (things that are loaded/rendered asynchronously). However, in cases where the rendered item does not have any assets, it wasn't ever calling `onRendered(true)`. 

This PR updates the `ServerItemRenderer` to check if any assets were registered in `componentDidMount`. If not, we can safely call `onRendered(true)` knowing that the item is ready for interaction and fully rendered.

Issue: LEMS-3200

## Test plan:

All tests pass. 

I will also take the snapshot PR and spin up a ZND to test the classroom link that was broken.

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -331,10 +331,14 @@ jobs:
                       steps.publish-snapshot.outputs.npm_snapshot_tag }}
                       ```
 
-                      If you are working in Khan Academy's webapp, you can run the below command. 
-                      If this is for backend code, change the static flag (-s) from y/yes to n/no.
+                      If you are working in Khan Academy's frontend, you can run the below command.
                       ```sh
-                      ./dev/tools/bump_perseus_version.js -s y -t PR${{ github.event.pull_request.number }}
+                      ./tools/bump_perseus_version.js -t PR${{ github.event.pull_request.number }}
+                      ```
+
+                      If you are working in Khan Academy's webapp, you can run the below command.
+                      ```sh
+                      ./dev/tools/bump_perseus_version.js -s n -t PR${{ github.event.pull_request.number }}
                       ```
 
             - name: Create or update npm snapshot comment - failure, snapshot publish failed
--- a/packages/perseus/src/__tests__/server-item-renderer.test.tsx
+++ b/packages/perseus/src/__tests__/server-item-renderer.test.tsx
@@ -3,6 +3,7 @@ import {within, render, screen, act} from "@testing-library/react";
 import {userEvent as userEventLib} from "@testing-library/user-event";
 import * as React from "react";
 
+import {LoadingContext} from "..";
 import {
     testDependencies,
     testDependenciesV2,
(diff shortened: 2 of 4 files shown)


packages/perseus/src/__tests__/server-item-renderer.test.tsx
```diff
@@ -271,6 +272,85 @@ describe("server item renderer", () => {
         expect(onRendered).toHaveBeenCalledWith(true);
     });
 
+    it("should call the onRendered callback when if no assets in content", () => {
```
handeyeco:
```suggestion
    it("should call the onRendered callback with no assets in content", () => {
```

.github/workflows/node-ci.yml
```diff
@@ -331,10 +331,14 @@ jobs:
...
-                      If this is for backend code, change the static flag (-s) from y/yes to n/no.
+                      If you are working in Khan Academy's frontend, you can run the below command.
                       ```sh
-                      ./dev/tools/bump_perseus_version.js -s y -t PR${{ github.event.pull_request.number }}
+                      ./tools/bump_perseus_version.js -t PR${{ github.event.pull_request.number }}
+                      ```
+
+                      If you are working in Khan Academy's webapp, you can run the below command.
+                      ```sh
+                      ./dev/tools/bump_perseus_version.js -s n -t PR${{ github.event.pull_request.number }}
```
handeyeco, line 341: We probably can get rid of the static flag in Webapp now.

packages/perseus/src/__tests__/server-item-renderer.test.tsx
```diff
@@ -271,6 +272,85 @@ describe("server item renderer", () => {
...
+                chi2Table: false,
+                financialCalculatorMonthlyPayment: false,
+                financialCalculatorTotalAmount: false,
+                financialCalculatorTimeToPayOff: false,
+                periodicTable: false,
+                periodicTableWithKey: false,
+                tTable: false,
+            },
+            hints: [],
+        };
```
handeyeco: Just because I'm saying this to everyone: I would love if we could make the simplest test objects possible when writing tests. This is long and it's unclear what's important for the same of the test.

```suggestion
const content: PerseusItem = {
    question: {
        content: "[[☃ radio 1]]",
        images: {},
        widgets: {
            "radio 1": {
                type: "radio",
                options: {
                    choices: [
                        {
                            content: "Correct",
                            correct: true,
                        },
                        {
                            content: "Incorrect",
                            correct: false,
                        },
                    ],
                },
            },
        },
    },
    answerArea: null,
    hints: [],
};
```

This is shorter and will have the same impact I think. Even this isn't particularly clear (we're not explicitly making an object without assets), but at least it's easier to see there's one widget, it has no assets, and there's nothing hiding in the content.

handeyeco (approved):
Nice!

packages/perseus/src/__tests__/server-item-renderer.test.tsx
jeremywiebe: Yeah, I was thinking about this after I pushed this PR.  I can actually completely remove the widget as we can render a simple item with `content` only!

PR #2580, opened 2025-06-11
Add template prop to EditorPage

daniellewhyte:
## Summary:
As part of the work to support a side-by-side layout in Interim Assessments, we are having content creators add a delimiter in the Perseus content to show where one column ends and the other begins. This is what the Perseus content would look like: 
Screenshot 2025-06-18 at 9 46 51 AM

To improve their user experience, we would like to have a template that they can use to add the delimiter.

This PR adds an optional prop `additionalTemplates` to the EditorPage component so that we can pass in a template for the side by side content.

Issue: AX-1006

## Test plan:


packages/perseus-editor/src/editor.tsx
```diff
@@ -112,6 +112,7 @@ const imageUrlsFromContent = function (content: string) {
 };
 
 type Props = Readonly<{
+    additionalTemplates?: Record<string, string>;
```
handeyeco: I think we could add this to default props and it would simplify the logic a bit.

```suggestion
    additionalTemplates: Record<string, string>;
```

``` TS
type DefaultProps = {
  additionalTemplates: Props["additionalTemplates"];
```

``` TS
  additionalTemplates: {},
```

``` TS
        } else if (
            templateType in this.props.additionalTemplates
        ) {
```

``` TS
                    {Object.entries(this.props.additionalTemplates).map(
                            ([key, value]) => (
                                <option value={key} key={key}>
                                    {key}
                                </option>
                            ),
                        )}
```

handeyeco (approved):

packages/perseus-editor/src/editor-page.tsx
```diff
@@ -25,6 +25,9 @@ import type {
 const {HUD} = components;
 
 type Props = {
+    // Additional templates that the host application would like to display
+    // within the Perseus Editor.
+    additionalTemplates?: Record<string, string>;
```
jeremywiebe: Perfect!  I'd just suggest switching this to be a JSDoc string as that surfaces the comments better in tools like VSCode. 
```suggestion
    /** 
     * Additional templates that the host application would like to display
     * within the Perseus Editor.
     */
    additionalTemplates?: Record<string, string>;
```

packages/perseus-editor/src/item-editor.tsx
```diff
@@ -24,6 +24,9 @@ import type {
 } from "@khanacademy/perseus-core";
 
 type Props = {
+    // Additional templates that the host application would like to display
+    // within the Perseus Editor.
```
jeremywiebe: Great.  Same request about switching this to a JSDoc comment.

jeremywiebe (approved):
Thanks for implementing this @daniellewhyte 

This is exactly what I had in mind! 👌

PR #2581, opened 2025-06-11
Matrix: renderer user input state

handeyeco:
## Summary:

---

MatrixEditor does use Matrix, so this will affect the editing experience

### DIFF
--- a/packages/perseus/src/widgets/matrix/matrix.tsx
+++ b/packages/perseus/src/widgets/matrix/matrix.tsx
@@ -83,8 +83,6 @@ const getRefForPath = function (path: FocusPath) {
     PerseusMatrixUserInput
 > satisfies PropsFor<typeof Matrix>;
 
-// type RenderProps = MatrixPublicWidgetOptions;
-
 type RenderProps = Pick<
     PerseusMatrixWidgetOptions,
     "matrixBoardSize" | "prefix" | "suffix"


packages/perseus-editor/src/widgets/matrix-editor.tsx
```diff
@@ -37,7 +38,7 @@ class MatrixEditor extends React.Component<Props> {
     static defaultProps: MatrixDefaultWidgetOptions =
         matrixLogic.defaultWidgetOptions;
 
-    change: (arg1: any, arg2: any, arg3: any) => any = (...args) => {
+    change: (arg1: any, arg2?: any, arg3?: any) => any = (...args) => {
```
SonicScrewdriver, line 41: Again, it'd be nice to remove changeable here too, but it's not required.

packages/perseus-editor/src/widgets/matrix-editor.tsx
```diff
@@ -67,15 +68,17 @@ class MatrixEditor extends React.Component<Props> {
     };
 
     render(): React.ReactNode {
-        const matrixProps = _.extend(
-            {
-                numericInput: true,
-                onBlur: () => {},
-                onFocus: () => {},
-                trackInteraction: () => {},
+        const matrixProps: Partial<PropsFor<typeof Matrix>> = {
```
SonicScrewdriver, line 71: I've already mentioned this separately in other PR's, but I wonder if we couldn't simplify how we're doing these types since we're making a partial and then casting it as the full props later.

packages/perseus/src/widgets/matrix/matrix.tsx
```diff
@@ -104,29 +83,32 @@ type ExternalProps = WidgetProps<
     PerseusMatrixUserInput
 > satisfies PropsFor<typeof Matrix>;
 
-type RenderProps = MatrixPublicWidgetOptions & {
-    emptyMatrix: ReadonlyArray<ReadonlyArray<string>>;
-};
+// type RenderProps = MatrixPublicWidgetOptions;
```
SonicScrewdriver: Should you clear this line out, given we have the proper declaration right after? If not, we might want to add a comment explaining why we're keeping this here.

packages/perseus/src/widgets/matrix/matrix.tsx
```diff
@@ -104,29 +83,32 @@ type ExternalProps = WidgetProps<
     PerseusMatrixUserInput
 > satisfies PropsFor<typeof Matrix>;
 
-type RenderProps = MatrixPublicWidgetOptions & {
-    emptyMatrix: ReadonlyArray<ReadonlyArray<string>>;
```
SonicScrewdriver, line 108: I see we're losing this emptyMatrix prop - was it unused?

packages/perseus/src/widgets/matrix/matrix.tsx
```diff
@@ -540,44 +536,46 @@ class Matrix extends React.Component<Props, State> implements Widget {
...
-    const rows = widgetOptions.matrixBoardSize[0];
-    const columns = widgetOptions.matrixBoardSize[1];
-    const blankInput = stringArrayOfSize2D({rows, columns});
-
     const {matrixBoardSize, prefix, suffix} = widgetOptions;
     return {
         matrixBoardSize,
         prefix,
         suffix,
-        emptyMatrix: blankInput,
```
SonicScrewdriver, line 554: Oh I see, here it is! Seems like a decent cleanup considering the tests are proving the inputs are properly empty. What a strange thing that doesn't seem to have been helping much at all.

SonicScrewdriver (approved):
Looks good to me. `emptyMatrix` gave me some pause, but I think it all makes sense.

packages/perseus/src/widgets/matrix/matrix.tsx
handeyeco: Good catch!

packages/perseus-editor/src/widgets/matrix-editor.tsx
handeyeco: From the other PR:

> I don't know what to do about the cast on 317. We're expecting props for an Expression widget but the editor doesn't provide all of those props. This happens in a couple of places in the editors because the widgets in the editors don't need all of the stuff they need in the learner experience. For the most part this worked because of poor type coverage or use of `ts-expect-error` which is much worse than type casting.
> 
> I think this is still an improvement than what was here before and I think further work on this is beyond the scope of my current work.

packages/perseus-editor/src/widgets/matrix-editor.tsx
handeyeco: I can't get rid of it everywhere right now because there are some uses that are beyond the scope of my current work. In this case the widget isn't wrapped in a Renderer so the editor is handling its own state. Maybe something that should change, but not related to my current work.

PR #2582, opened 2025-06-11
IFrame: renderer user input state

handeyeco:
## Summary:

---

IFrameEditor doesn't use IFrame, so this shouldn't affect the editing experience


SonicScrewdriver (approved):
LGTM

PR #2583, opened 2025-06-11
Replace custom icons with WB components for consistent styling in Perseus Editor

tatianasnook:
## Summary:
Replace custom icons with WB components for consistent styling in Perseus Editor

This PR updates the styling in the Perseus Editor to create a more unified and accessible experience across components.

- Replaced the anchor tag used for panel toggling with a Wonder Blocks IconButton.
- Applied consistent styles and toggle behavior across widget-editor.tsx and issues-panel.tsx.
- Partially implements styling and structure from [Mark Fitzgerald’s POC #2219](https://github.com/Khan/perseus/pull/2219) to align with design direction.

Screenshot 2025-06-12 at 9 51 37 AM

Issue: LEMS-3125

## Test plan:

### DIFF
--- a/packages/perseus-editor/src/components/widget-editor.tsx
+++ b/packages/perseus-editor/src/components/widget-editor.tsx
@@ -167,14 +167,19 @@ class WidgetEditor extends React.Component<
                 >
                     <div className="perseus-widget-editor-title-id">
                         <View
-                            style={{marginRight: 5, flexGrow: 0}}
+                            style={{
+                                display: "flex",
+                                flexDirection: "row",
+                                alignItems: "center",
+                                gap: "0.25em",
+                            }}
                             onClick={this._toggleWidget}
                         >
                             <ToggleableCaret
                                 isExpanded={this.state.showWidget}
                             />
+                            <span>{this.props.id}</span>
                         </View>
-                        <span>{this.props.id}</span>
                     </div>
 
                     {supportsStaticMode && (
--- a/packages/perseus-editor/src/components/widget-editor.tsx
+++ b/packages/perseus-editor/src/components/widget-editor.tsx
@@ -1,31 +1,24 @@
 /* eslint-disable @khanacademy/ts-no-error-suppressions */
-import {
-    components,
-    Widgets,
-    WIDGET_PROP_DENYLIST,
-    iconChevronDown,
-    iconTrash,
-} from "@khanacademy/perseus";
+import {Widgets, WIDGET_PROP_DENYLIST, iconTrash} from "@khanacademy/perseus";
 import {
     CoreWidgetRegistry,
     applyDefaultsToWidget,
 } from "@khanacademy/perseus-core";
+import IconButton from "@khanacademy/wonder-blocks-icon-button";
 import {Strut} from "@khanacademy/wonder-blocks-layout";
 import Switch from "@khanacademy/wonder-blocks-switch";
 import {spacing} from "@khanacademy/wonder-blocks-tokens";
+import caretDown from "@phosphor-icons/core/regular/caret-down.svg";
+import caretRight from "@phosphor-icons/core/regular/caret-right.svg";
 import * as React from "react";
 import {useId} from "react";
 import _ from "underscore";
-
-import {iconChevronRight} from "../styles/icon-paths";
 
 import SectionControlButton from "./section-control-button";
 
 import type Editor from "../editor";
 import type {APIOptions} from "@khanacademy/perseus";
 import type {Alignment, PerseusWidget} from "@khanacademy/perseus-core";
-
-const {InlineIcon} = components;
 
 type WidgetEditorProps = {
     // Unserialized props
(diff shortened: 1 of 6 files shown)


jeremywiebe (changes requested):
I remembered that I'd built a re-usable rotatable caret icon for exactly this situation. 

Can you use/adapt it to these two use cases?

https://github.com/Khan/perseus/blob/8534a9f017a9a5404201c2459fabe91a83d6709f/packages/perseus-editor/src/components/toggleable-caret.tsx#L14

You can see it in use here: https://github.com/Khan/perseus/blob/3cdabf1a39b4ef3e1f6b8a76f53669f64818e407/packages/perseus-editor/src/tex-error-view.tsx#L34


https://github.com/user-attachments/assets/b9f6d2d8-694c-4078-b1fd-2380b125b259

packages/perseus-editor/src/issues-panel.tsx
```diff
@@ -43,26 +41,16 @@ const IssuesPanel = ({issues = []}: IssuesPanelProps) => {
...
                     >
-                        {showPanel ? (
-                            <InlineIcon {...iconChevronDown} />
-                        ) : (
-                            <InlineIcon {...iconChevronRight} />
-                        )}
-                    </button>
+                        <ToggleableCaret isExpanded={showPanel} />
+                    </View>
                     <span>Issues</span>
```
jeremywiebe: I think this changes the behaviour of the panel (and doesn't stay true to what the tests document).  The tests state "opens the panel when the heading is clicked".  The way this is structured means only clicking the caret will toggle.  I _think_ if you nest the `<span>Issues</span>` _inside_ the `<View>` you will get the behaviour you want (being able to click anywhere in the header). 

Plus, you should be able to get rid of the `testId="issues-toggle-caret"` because you can simply click on `screen.getByText("Issues")`. (Although React Testing Library does support finding elements by test ids, it's meant to be a [last resort](https://testing-library.com/docs/queries/bytestid/):

> In the spirit of [the guiding principles](https://testing-library.com/docs/guiding-principles), it is recommended to use this only after the other queries don't work for your use case. Using data-testid attributes do not resemble how your software is used and should be avoided if possible. That said, they are way better than querying based on DOM structure or styling css class names. Learn more about data-testids from the blog post ["Making your UI tests resilient to change"](https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change)

packages/perseus-editor/src/components/widget-editor.tsx
```diff
@@ -173,19 +165,17 @@ class WidgetEditor extends React.Component<
...
+                        <View
+                            style={{marginRight: 5, flexGrow: 0}}
+                            onClick={this._toggleWidget}
+                        >
+                            <ToggleableCaret
+                                isExpanded={this.state.showWidget}
+                            />
+                        </View>
+                        <span>{this.props.id}</span>
+                    </div>
```
jeremywiebe: As with my comment lower down, the `<span>` should be inside the clickable element, otherwise you can only click on the caret icon to toggle expansion.

packages/perseus-editor/src/issues-panel.tsx
tatianasnook: I updated it so the entire header is clickable by nesting the <span> inside the <View>. Thanks for the feedback!

packages/perseus-editor/src/issues-panel.tsx
```diff
@@ -26,15 +24,15 @@ type IssuesPanelProps = {
...
     const [showPanel, setShowPanel] = useState(false);
-    const icon = hasWarnings ? iconWarning : iconPass;
-    const iconColor = hasWarnings ? wbColor.gold : wbColor.green;
+
+    const hasWarnings = issues.length > 0;
     const issuesCount = `${issues.length} issue${
         issues.length === 1 ? "" : "s"
     }`;
 
-    const editorClasses = `perseus-widget-editor${showPanel ? " perseus-widget-editor-open" : ""}`;
```
catandthemachines, line 37: I didn't catch this in earlier PRs, but I'm glad it's going away in favor of a better less fragile implementation. Nice @tatianasnook !

packages/perseus-editor/src/components/widget-editor.tsx
tatianasnook: Done

jeremywiebe (approved):
Fantastic! Thanks for adjusting this to use the `ToggleableCaret`. 

I have one request before landing it.  Can you add:

```css
    cursor: pointer;
```

To the `.perseus-widget-editor .perseus-widget-editor-title` rule in packages/perseus-editor/src/styles/perseus-editor.css?

It's around line 119.  

That'll cause the mouse cursor to indicate that you can click on the header more clearly.

https://github.com/user-attachments/assets/3b4ab1ea-dd77-4077-951f-e1103d184c8d

Then you're good to land this!

PR #2585, opened 2025-06-12
FreeResponse: renderer user input state

handeyeco:
## Summary:
<Add your summary here>

Issue: LEMS-XXXX

## Test plan:

### DIFF
--- a/packages/perseus/src/widgets/free-response/serialize-free-response.test.ts
+++ b/packages/perseus/src/widgets/free-response/serialize-free-response.test.ts
@@ -1,0 +1,151 @@
+import {
+    generateTestPerseusItem,
+    generateTestPerseusRenderer,
+} from "@khanacademy/perseus-core";
+import {screen, act} from "@testing-library/react";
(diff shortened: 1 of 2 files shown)


packages/perseus/src/widgets/free-response/serialize-free-response.test.ts
```diff
@@ -0,0 +1,151 @@
...
+        );
+
+        // Act
+        const state = renderer.getSerializedState();
+        const userInput = renderer.getUserInput();
+
+        // Assert
+        // FreeResponse doesn't seem to include the user's answer
+        // in the serialized state, this assertion just affirms
+        // that there _is_ a user input, it's just not serialized
```
aag, line 96: Is it bad that the user's input isn't included in the object returned by `getSerializedState()`? That could easily be added.

benchristel: `getSerializedState` is deprecated. If no one is using the data it outputs for free response widgets, we'd prefer not to encourage them.

PR #2587, opened 2025-06-12
Table: renderer user input state

handeyeco:
## Summary:
<Add your summary here>

Issue: LEMS-XXXX

## Test plan:

### DIFF
--- a/packages/perseus/src/widgets/table/table.tsx
+++ b/packages/perseus/src/widgets/table/table.tsx
@@ -23,9 +23,9 @@ type EditorProps = {
     Editor: any;
 };
 
-type RenderProps = PerseusTableWidgetOptions & EditorProps;
+type RenderProps = Omit<PerseusTableWidgetOptions, "answers">;
 
-type Props = WidgetProps<PerseusTableWidgetOptions, PerseusTableUserInput> &
+type Props = WidgetProps<RenderProps, PerseusTableUserInput> &
     ChangeableProps &
     EditorProps;
 
@@ -35,7 +35,6 @@ type DefaultProps = {
     editableHeaders: Props["editableHeaders"];
     rows: Props["rows"];
     columns: Props["columns"];
-    answers: Props["answers"];
     linterContext: Props["linterContext"];
 };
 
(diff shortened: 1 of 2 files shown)


packages/perseus/src/widgets/table/table.tsx
```diff
@@ -23,9 +23,9 @@ type EditorProps = {
     Editor: any;
 };
 
-type RenderProps = PerseusTableWidgetOptions & EditorProps;
+type RenderProps = Omit<PerseusTableWidgetOptions, "answers">;
 
-type Props = WidgetProps<PerseusTableWidgetOptions, PerseusTableUserInput> &
+type Props = WidgetProps<RenderProps, PerseusTableUserInput> &
     ChangeableProps &
```
SonicScrewdriver, line 29: It'd be lovely to get rid of changeable again here, but I understand if you don't want to stuff too much work into this already large task. :)

packages/perseus/src/widgets/table/table.tsx
```diff
@@ -310,22 +305,27 @@ class Table extends React.Component<Props> implements Widget {
...
-    const blankAnswers = Util.stringArrayOfSize2D({rows, columns});
-    return {
-        ...options,
-        answers: blankAnswers,
-    };
+    return Util.stringArrayOfSize2D({rows, columns});
+}
+
+function getUserInputFromSerializedState(
+    serializedState: any,
```
SonicScrewdriver, line 322: I'm not too fussed as I know we want to remove the serialized state eventually, but I do think avoiding "any" wherever possible is probably going to help us. 

I also think this method should have the deprecated comment you've added to the other serialized states across your PRs

packages/perseus/src/widgets/table/table.tsx
handeyeco: I tried to remove uses that were only related to user input. In my next leg of work I'll being looking at UI state which should remove more uses of Changeable. Then the only thing that will be left are uses for the editor.

packages/perseus/src/widgets/table/table.tsx
handeyeco: `any` is actually kind of correct here. We don't have a type for this - it's just a blob of data. That's a big part of why I want to remove the serialization stuff.

Comment added!

PR #2588, opened 2025-06-12
Plotter: renderer user input state

handeyeco:
## Summary:

---

PlotterEditor does use Plotter, so this will affect the editing experience

### DIFF
--- a/packages/perseus-editor/src/widgets/plotter-editor.tsx
+++ b/packages/perseus-editor/src/widgets/plotter-editor.tsx
@@ -14,6 +14,7 @@ import type {
     PerseusPlotterWidgetOptions,
     PlotterDefaultWidgetOptions,
 } from "@khanacademy/perseus-core";
+import type {PropsFor} from "@khanacademy/wonder-blocks-core";
 
 const {InfoTip, NumberInput, RangeInput, TextListEditor} = components;
 const Plotter = PlotterWidget.widget;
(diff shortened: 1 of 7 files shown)


packages/perseus-editor/src/widgets/plotter-editor.tsx
```diff
@@ -285,10 +286,16 @@ class PlotterEditor extends React.Component<Props, State> {
             this.props.type,
         );
         const canChangeSnaps = !_.contains(["pic", "dotplot"], this.props.type);
-        const props = {
-            trackInteraction: () => {},
+        const plotterProps: any = {
```
SonicScrewdriver, line 289: Could we not set the type here instead of using any and then casting in the jsx?

packages/perseus/src/widgets/plotter/plotter.tsx
```diff
@@ -1080,18 +1073,30 @@ export class Plotter extends React.Component<Props, State> implements Widget {
...
     };
 
+    /**
+     * Plotter uses Graphie and Graphie is inaccessible,
+     * this helps us populate user input in tests
+     * (not great, but not a lot of options)
+     *
+     * @internal
+     */
+    _testInsertUserInput(userInput: PerseusPlotterUserInput) {
```
SonicScrewdriver, line 1096: Ooo interesting, I was wondering where this was coming from. 

I think this makes sense for now, but I wonder if we shouldn't consider making a ticket to find a more holistic / standardized / better way to handle this. It seems like a possible idiosyncrasy that we'd wonder about in the future. 

Maybe it's simply not worthwhile if our goal is to move away from Graphie anyway.

SonicScrewdriver (approved):
Happy to approve! I only really had one question about the plotterProps

packages/perseus-editor/src/widgets/plotter-editor.tsx
handeyeco: I honestly can't figure out any other way to avoid an error besides `ts-expect-error`, which I removed because it's worse than `any`. Open to ideas though.

packages/perseus/src/widgets/plotter/plotter.tsx
handeyeco: Plotter is a hidden widget, so I think it's fine to ignore the problem until it can be replaced with IG.

PR #2590, opened 2025-06-12
Remove renderer's deprecated getUserInput in favor of the preferred getUserInputMap

Myranae:
## Summary:
Renderer currently has two methods for getting user input: the deprecated method getUserInput which returns an array of type UserInputArray and the preferred getUserInputMap which returns an object of type UserInputMap. The object type is the preferred format going forward and is currently used in scoring and will be used in scoring via the new Perseus service.

To remove the deprecated method, we replace all uses of it in Perseus with the preferred method and then delete tests and comments related to the deprecated method. This change will require some coordination with other teams that rely on this functionality outside of Perseus and most likely updates to where user input is consumed. Testing will take place to determine the necessary places that require updates.

Issue: LEMS-2523

## Test plan:
- Confirm all checks/tests pass
- Add tests to Webapp around UserInputArray usage so we can locate places that need to be updated
--- Use the tests to update all usages until all tests pass
- Possibly play tests or some manual testing of all places known to be affected


.changeset/little-turtles-train.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/perseus": major
+---
+
+Replace deprecated array format of user input with preferred object format to consolidate the user input types.
```
Myranae, line 5: I am pretty sure this is a major change, but not sure if I need to add more information to the changelog. I add more details in the PR summary, so not sure.

Myranae:
Not moving forward with this

PR #2592, opened 2025-06-12
Orderer: renderer user input state

handeyeco:
## Summary:

---

OrdererEditor doesn't use Orderer, so this shouldn't affect the editing experience

### DIFF
--- a/packages/perseus/src/widgets/orderer/orderer.tsx
+++ b/packages/perseus/src/widgets/orderer/orderer.tsx
@@ -294,19 +294,16 @@ class Card extends React.Component<CardProps, CardState> {
     }
 }
 
-type RenderProps = OrdererPublicWidgetOptions & {
-    current: any;
-};
+type RenderProps = OrdererPublicWidgetOptions;
 
 type OrdererProps = WidgetProps<RenderProps, PerseusOrdererUserInput>;
 
 type OrdererDefaultProps = Pick<
     OrdererProps,
-    "current" | "options" | "height" | "layout" | "linterContext"
+    "options" | "height" | "layout" | "linterContext" | "userInput"
 >;
 
 type OrdererState = {
-    current: ReadonlyArray<any>;
     dragging: boolean;
     placeholderIndex: number | null | undefined;
     dragKey: string | null | undefined;
@@ -336,15 +333,16 @@ class Orderer
     implements Widget
 {
     static defaultProps: OrdererDefaultProps = {
-        current: [],
         options: [],
         height: "normal",
         layout: "horizontal",
         linterContext: linterContextDefault,
+        userInput: {
+            current: [],
+        },
     };
 
     state: OrdererState = {
-        current: [],
         dragging: false,
         placeholderIndex: null,
         dragKey: null,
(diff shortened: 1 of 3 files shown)


packages/perseus/src/widgets/orderer/orderer.tsx
```diff
@@ -357,11 +355,11 @@ class Orderer
...
-    UNSAFE_componentWillReceiveProps(nextProps: OrdererProps) {
-        if (!_.isEqual(this.props.current, nextProps.current)) {
-            this.setState({current: nextProps.current});
-        }
-    }
+    // UNSAFE_componentWillReceiveProps(nextProps: OrdererProps) {
+    //     if (!_.isEqual(this.props.current, nextProps.current)) {
+    //         this.setState({current: nextProps.current});
+    //     }
+    // }
```
SonicScrewdriver: Why only comment out this method rather than delete it?

packages/perseus/src/widgets/orderer/orderer.tsx
```diff
@@ -371,7 +369,7 @@ class Orderer
     ) => {
         // @ts-expect-error - TS2769 - No overload matches this call.
         const $draggable = $(ReactDOM.findDOMNode(draggable));
```
SonicScrewdriver, line 365: Ohhh another findDomNode to get rid of. (More of a self note to help me remember)

packages/perseus/src/widgets/orderer/orderer.tsx
```diff
@@ -597,30 +596,30 @@ class Orderer
...
-            return {content: value};
-        });
-        this.props.onChange({
-            // @ts-expect-error - TS2345 - Argument of type '{ current: { content: string; }[]; }' is not assignable to parameter of type '{ hints?: readonly Hint[] | undefined; replace?: boolean | undefined; content?: string | undefined; widgets?: PerseusWidgetsMap | undefined; images?: ImageDict | undefined; ... 13 more ...; plot?: any; }'.
-            current: list,
-        });
-
-        this.setState({current: list});
+    setListValues: (values: string[]) => void = (values) => {
+        this.props.handleUserInput({current: values});
```
SonicScrewdriver, line 594: This code is just so much cleaner. I'm so happy to see it in every single PR

packages/perseus/src/widgets/orderer/orderer.tsx
```diff
@@ -656,15 +655,14 @@ class Orderer
...
             return (
                 <Card
                     key={`sortableCard${i}`}
                     ref={"sortable" + i}
                     fakeRef={"sortable" + i}
                     floating={false}
-                    content={opt.content}
-                    width={opt.width}
+                    content={opt}
                     linterContext={this.props.linterContext}
```
SonicScrewdriver, line 660: Do we not need the width anymore?

SonicScrewdriver (approved):
Happy to approve this, but I would recommend double checking that removing the width property isn't causing any regressions!

packages/perseus/src/widgets/orderer/orderer.tsx
handeyeco: Whoops, just forgot!

packages/perseus/src/widgets/orderer/orderer.tsx
handeyeco: I can't for the life of me remember why I did this...I compared the old/new behavior in Storybook and they seem to be identical. It doesn't seem like we needed it?

PR #2596, opened 2025-06-12
Grapher: renderer user input state

handeyeco:
## Summary:
<Add your summary here>

Issue: LEMS-XXXX

## Test plan:

### DIFF
--- a/packages/perseus-core/src/validation.types.ts
+++ b/packages/perseus-core/src/validation.types.ts
@@ -127,7 +127,7 @@ export type PerseusGrapherRubric = {
     correct: GrapherAnswerTypes;
 };
 
-export type PerseusGrapherUserInput = PerseusGrapherRubric["correct"];
+export type PerseusGrapherUserInput = GrapherAnswerTypes;
 
 export type PerseusIFrameUserInput = {
     status: UserInputStatus;
--- a/packages/perseus/src/widgets/grapher/grapher.stories.tsx
+++ b/packages/perseus/src/widgets/grapher/grapher.stories.tsx
@@ -10,6 +10,7 @@ import {
     logarithmQuestion,
     quadraticQuestion,
     sinusoidQuestion,
+    simpleQuestion,
 } from "./grapher.testdata";
 
 import type {Meta, StoryObj} from "@storybook/react-vite";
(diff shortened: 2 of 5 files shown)


packages/perseus-editor/src/widgets/grapher-editor.tsx
```diff
@@ -67,35 +71,34 @@ class GrapherEditor extends React.Component<Props> {
...
-                // @ts-expect-error - TS2769 - No overload matches this call.
-                <Grapher
-                    {...graphProps}
-                    apiOptions={this.props.apiOptions}
-                    containerSizeClass={sizeClass}
-                />
+                <Grapher {...(graphProps as PropsFor<typeof Grapher>)} />
+            );
+            equationString = getEquationString(
+                graphProps.userInput as GrapherAnswerTypes,
```
nishasy, line 100: Can you use `satisfies` instead of `as` here?

packages/perseus/src/widgets/grapher/util.tsx
```diff
@@ -26,11 +29,14 @@ const DEFAULT_BACKGROUND_IMAGE = {
...
-    const plot = props.plot;
+export const getEquationString = (plot: GrapherAnswerTypes): string => {
     if (plot.type && plot.coords) {
         const handler = GrapherUtil.functionForType(plot.type);
-        const result = handler.getEquationString(plot.coords, plot.asymptote);
+        const result = handler.getEquationString(
+            plot.coords,
+            // some graph types don't have this
+            (plot as any).asymptote,
+        );
```
nishasy, line 39: I'm not a fan of how this gets the equation string from the handler. (But! Still not as gross as interactive graph using strings to build the function names 😬) Just a qualm with the original code, not with your changes.

You don't have to do anything, I'm just making a note to myself that this needs to be updated whenever we try to merge grapher into interactive graph.

nishasy (approved):
Looks good to me! And an great reminder that I do not like the grapher widget.

packages/perseus-editor/src/widgets/grapher-editor.tsx
handeyeco: That didn't seem to work 🤔

PR #2597, opened 2025-06-12
[Interactive Graph] Update autogen labels to correctly reflect locked line vs segment vs ray

nishasy:
## Summary:
Request from content authors regarding autogenerated locked figures' aria labels:

> We are noticing that the screen reader tree seems to default to "from this point to that point" whether it's talking about a line, a segment, or a ray.  We can certainly ensure that we write accurate alt text regardless, but could we have the tree use:
> from this point through that point for RAY
> from this point to that point for SEGMENT
> through this point and that point for LINE

Updating the labels in the locked line settings.

Issue: https://khanacademy.atlassian.net/browse/LEMS-3203

## Test plan:
`pnpm jest packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-line-settings.test.tsx`

Storybook
- http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-with-aria-label
- Add a locked line
- Change the type to "line"
- Press the auto-generate label button
- Confirm that it says "through point1 and point2"
- Change the type to "segment"
- Press the auto-generate label button
- Confirm that it says "from point1 to point2"
- Change the type to "ray"
- Press the auto-generate label button
- Confirm that it says "from point1 through point2"


packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-line-settings.tsx
```diff
@@ -96,7 +97,29 @@ const LockedLineSettings = (props: Props) => {
...
+                break;
+            case "segment":
+                connector1 = "from";
+                connector2 = "to";
+                break;
+            default:
+                throw new UnreachableCaseError(kind, "Unknown line kind");
+        }
+
+        let str = `${capitalizeKind}${visiblelabel} ${connector1} point${point1VisibleLabel} at ${spokenPoint1X} comma ${spokenPoint1Y} ${connector2} point${point2VisibleLabel} at ${spokenPoint2X} comma ${spokenPoint2Y}`;
```
jeremywiebe: TBH, this feels like it's getting overly clever. I wonder if it would easier to read and maintain if each line type built it's spoken string as this one here. Something like this? 

```suggestion
        const spokenPoint1 = `${await generateSpokenMathDetails(
            `$${point1.coord[0]}$`,
        )} comma ${await generateSpokenMathDetails(
            `$${point1.coord[1]}$`,
        )}`;
        const spokenPoint2 = `${await generateSpokenMathDetails(
            `$${point2.coord[0]}$`,
        )} comma ${await generateSpokenMathDetails(
            `$${point2.coord[1]}$`,
        )};

         let str = "";
         switch (kind) {
             case "line":
                 str = `${capitalizeKind}${visiblelabel} through point${point1VisibleLabel} at ${spokenPoint1} and point${point2VisibleLabel} at ${spokenPoint2}`
                 break;
             case "ray":
                 str = `${capitalizeKind}${visiblelabel} from point${point1VisibleLabel} at ${spokenPoint1} through point${point2VisibleLabel} at ${spokenPoint2}`
                 break;
             case "segment":
                 str = `${capitalizeKind}${visiblelabel} from point${point1VisibleLabel} at ${spokenPoint1} to point${point2VisibleLabel} at ${spokenPoint2}`
                 break;
             default:
                 throw new UnreachableCaseError(kind, "Unknown line kind");
         }
```

jeremywiebe (commented):
Crashing through with my opinions... 😂 

I'll leave it for folks more closely-related to the project to approve.

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-line-settings.tsx
nishasy: That's so funny 😂 That's what I had originally, but then decided to change it because I couldn't figure out which one was harder to read.

jeremywiebe: Not a deal-breaker, but I thought less logic made for easier reading of what the final strings would be (note that I also proposed building each point's text fully instead of having individual `x` and `y` strings.  🤷‍♂️

mark-fitzgerald (approved):
Lovely. Great to see how easy it is to accommodate requests like this.

PR #2599, opened 2025-06-12
Linter: Break each rule's tests into their own file!

jeremywiebe:
## Summary:

I was doing some research and work in the Linter (researching the origin and use of `BULK_WARNING` severity.  

The way the rules tests were grouped in one file felt difficult to find (it's not obvious a rule has tests because it's in `../__tests__/rules.test.ts` and not close to the implementation file at all. 

So I just pulled them into their own files beside the rule. Hopefully this will encourage better maintenance of the rule test. 

No production changes!

Issue: "none"

## Test plan:

`pnpm test`


catandthemachines (commented):
Already beat me too it but, approved! 😉

PR #2601, opened 2025-06-13
docs(changeset): Add test for protecting against edgecase regressions regarding undefined correct value in Radio Widget.

SonicScrewdriver:
## Summary:
This micro-pr is part of the Radio Widget Project. 

All it does is add a test to help prevent regressions. :) 

Issue: LEMS-2931

## Test plan:
- new test passes. 
-- (Also confirmed that the test does NOT pass if I revert the fix from the original PR https://github.com/Khan/perseus/pull/2307/files)

### DIFF
--- a/packages/perseus/src/widgets/radio/__tests__/radio.test.ts
+++ b/packages/perseus/src/widgets/radio/__tests__/radio.test.ts
@@ -913,7 +913,7 @@ describe("Radio Widget", () => {
          * use the reviewChoice to determine correctness, but the choices were already shuffled
          * while the reviewChoice order remained the same.
          *
-         * The fix is to explicitly set choice.correct to a boolean value in radio.ts:
+         * The fix was to explicitly set choice.correct to a boolean value in radio.ts:
          * correct: Boolean(choice.correct)
          */
         it("handles undefined choice.correct properly when multipleSelect and randomize are enabled", async () => {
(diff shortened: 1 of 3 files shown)


packages/perseus/src/widgets/radio/__tests__/radio.test.ts
```diff
@@ -904,5 +905,45 @@ describe("Radio Widget", () => {
...
+         * The fix was to explicitly set choice.correct to a boolean value in radio.ts:
+         * correct: Boolean(choice.correct)
+         */
+        it("handles undefined choice.correct properly when multipleSelect and randomize are enabled", async () => {
+            // Arrange
+            renderQuestion(
+                questionWithUndefinedCorrect,
+                {},
+                {
+                    reviewMode: true,
```
SonicScrewdriver, line 925: By using review mode, I can easily recreate the issue in the radio.test file without having to test the individual user interaction. I verified that the test fails if the original fix is removed from radio.ts.

packages/perseus/src/widgets/radio/__tests__/radio.testdata.ts
```diff
@@ -580,3 +580,41 @@ export const shuffledNoneQuestion: PerseusRenderer = {
...
+            type: "radio",
+            options: {
+                choices: [
+                    {
+                        content: "Choice A",
+                        correct: undefined,
+                    },
+                    {
+                        content: "Choice B",
+                        correct: true,
```
anakaren-rojas, line 605: What is the behavior when all of them are undefined?

SonicScrewdriver: That should technically be impossible as content creators aren't allowed to save with absolutely no answers! It just seems that historically we allowed the "wrong" answers to be undefined rather than "false" - unfortunately. 

If no questions are marked correct, I think it's safe to consider that a content problem rather than a widget problem.

anakaren-rojas (approved):

PR #2602, opened 2025-06-13
Remove dead code in Group

handeyeco:
## Summary:
During a 1/1 Jeremy noticed some code that's not used anymore.


catandthemachines (approved):
Nice catch!

PR #2608, opened 2025-06-16
Don't remove answers from static widgets

handeyeco:
## Summary:
<Add your summary here>

Issue: LEMS-3194

## Test plan:

### DIFF
--- a/packages/perseus-core/src/utils/split-perseus-item.test.ts
+++ b/packages/perseus-core/src/utils/split-perseus-item.test.ts
@@ -72,6 +72,92 @@ describe("splitPerseusItem", () => {
         expect(rv.question).toEqual(question);
     });
 
+    it("doesn't strip static widgets", () => {
+        // Arrange
+        const question: PerseusRenderer = {
+            content: "[[☃ radio 1]]",
+            widgets: {
+                "radio 1": {
+                    ...getFullRadio(),
+                    static: true,
+                },
+                "radio 2": {
+                    ...getFullRadio(),
+                    static: false,
+                },
+            },
+            images: {},
+        };
+
+        // Act
+        const rv = splitPerseusItem(generateTestPerseusItem({question}));
+
+        // Assert
+        // `radio 1` is static, so we don't want to remove answerful data
+        expect(rv.question.widgets["radio 1"].options.choices[0].correct).toBe(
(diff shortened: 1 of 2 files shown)


packages/perseus-core/src/utils/split-perseus-item.test.ts
```diff
@@ -72,6 +72,92 @@ describe("splitPerseusItem", () => {
         expect(rv.question).toEqual(question);
     });
 
+    it("doesn't strip static widgets", () => {
+        // Arrange
+        const question: PerseusRenderer = {
+            content: "[[☃ radio 1]]",
```
Myranae: I'm probably missing something, but should radio 2 be in content as well?

packages/perseus-core/src/utils/split-perseus-renderer.ts
```diff
@@ -22,13 +22,24 @@ export default function splitPerseusRenderer(
...
-            ...widget,
-            options: publicWidgetOptionsFun(widget.options as any),
-        };
+        /**
+         * Static widgets are presentational, used to represent concepts
+         * rather than collect answers from learners. For instance we might
+         * want a static InteractiveGraph that shows learners a chart before
+         * we give them a Radio widget to check their understanding of the chart.
+         * We need to keep the answers in this case because it's pre-scoring information.
+         */
```
Myranae, line 31: Love the comment!

packages/perseus-core/src/utils/split-perseus-renderer.ts
```diff
@@ -22,13 +22,24 @@ export default function splitPerseusRenderer(
...
+        if (widget.static) {
+            splitWidgets[id] = widget;
+        } else {
```
Myranae, line 34: I'm curious, were you able to see an issue in Storybook or in some version of prod that this fixes? I'm all for it, but I'd love to know if it is actually working via a manual test. It seems to have the same outcome when I tested in prod storybook versus local storybook for this branch, so wondering how to manual test to see it work 🤔

handeyeco: I think in this case since it's operating on data (not a UI change) it should be enough to assert the data is in the right shape.

PR #2611, opened 2025-06-16
[Radio] | (DX) | Use the new question builder in testdata file

nishasy:
## Summary:
Now that we have a new question builder, we can use it to replace the
existing testdata json blobs to make them more readable.

Issue: https://khanacademy.atlassian.net/browse/LEMS-3187

## Test plan:
`pnpm jest packages/perseus/src/widgets/radio/__tests__/radio.test.ts`

Look at all the radio stories in the RadioNew docs page:
http://localhost:6006/?path=/docs/perseus-widgets-radionew--docs

### DIFF
--- a/packages/perseus/src/widgets/radio/__tests__/radio.testdata.ts
+++ b/packages/perseus/src/widgets/radio/__tests__/radio.testdata.ts
@@ -1,59 +1,34 @@
+import {radioQuestionBuilder} from "../radio-question-builder";
+
 import type {
     PerseusRenderer,
     RadioWidget,
     PassageWidget,
 } from "@khanacademy/perseus-core";
 
-export const question: PerseusRenderer = {
-    content:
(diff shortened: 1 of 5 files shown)


SonicScrewdriver (approved):
GLORIOUS!

packages/perseus/src/widgets/radio/__tests__/radio.testdata.ts
```diff
@@ -127,6 +103,7 @@ export const questionWithPassage: PerseusRenderer = {
     },
 };
 
+// Can't use radioQuestionBuilder here because it also includes a passage widget.
```
SonicScrewdriver, line 106: Makes sense! I've had some chats and it sounds like this feature would be a "nice to have" to go away in the future as the passage widget is largely deprecated.

PR #2613, opened 2025-06-16
Convert Aphrodite to CSS Modules in scrollable-view component

mark-fitzgerald:
## Summary:
Simple move of styling into a separate CSS file.

Issue: LEMS-3184

## Test plan:
1. Open Storybook with any of the scrollable stories (locally, or [from this PR](https://650db21c3f5d1b2f13c02952-ieiguhvhqn.chromatic.com/?path=/story/perseus-widgets-radionew--single-select-with-scroll)).
1. Open Storybook with the same scrollable story from the previous step ([PROD](https://khan.github.io/perseus/?path=/story/perseus-widgets-radionew--single-select-with-scroll)).
1. Compare the two visually and functionally.
   * There should be no difference between the two.


SonicScrewdriver (approved):
Looks GREAT! Thank you for cleaning this up.

anakaren-rojas (approved):

packages/perseus/src/components/scrollable-view.tsx
```diff
@@ -156,27 +156,18 @@ function ScrollableView({
...
-                            ? styles.scrollFadeLeft
-                            : styles.scrollFadeRight),
-                    }}
+                    className={`${styles.scrollFade} ${isRtl ? styles.scrollFadeLeft : styles.scrollFadeRight}`}
                 />
             )}
             <div
                 {...additionalProps}
                 role={role}
+                className={className}
```
nishasy, line 170: Checking my understanding: Are you adding this `className` prop so people can override the styles later if needed?

nishasy (approved):
Looks good to me!

packages/perseus/src/components/scrollable-view.tsx
mark-fitzgerald: Yes. Override or add to.

PR #2617, opened 2025-06-17
rough out regression test

handeyeco:
## Summary:
<Add your summary here>

Issue: LEMS-XXXX

## Test plan:


packages/perseus/src/widgets/radio/__tests__/broken-cypress.test.tsx
benchristel, line 1: You can create a set of regression tests by putting the item JSON that caused the parse failure in `packages/perseus-core/src/parse-perseus-json/regression-tests/item-data/`. No need for a whole new .test.tsx file!

PR #2618, opened 2025-06-17
[Radio] | (CX) | Resize question box height when a word is too long in the editor

nishasy:
## Summary:
If a word (or image link, specifically) is too long in the Radio editor, it overflows off the right of the question box.

We can quickly fix this by using the `overflow-wrap: anywhere` on the underlay (which determines the width) to break up the word/link.

Issue: https://khanacademy.atlassian.net/browse/LEMS-3226

## Test plan:
- Go to http://localhost:6006/?path=/story/perseuseditor-editorpage--demo
- Add a radio widget
- Paste a long image link
  - Example: A grass plant with green leaves and yellow stalks at the top.
- Confirm that it does not overflow off the right edge of the text area

### DIFF
--- a/.changeset/ten-taxis-decide.md
+++ b/.changeset/ten-taxis-decide.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": patch
+---
+
+[Radio] | (CX) | Resize question box height when a word is too long in the editor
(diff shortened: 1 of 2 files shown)


SonicScrewdriver (approved):
Overflow anywhere all the time.

anakaren-rojas (approved):

PR #2619, opened 2025-06-17
[Radio] | (CX) | Convert editor buttons to use Wonder Blocks equivalents

nishasy:
## Summary:
Make the Radio editor look slightly nicer using wonder blocks.

- Update checkboxes
- Updated style link
- Update the 'delete this choice' button
- Update 'add a choice' and 'none of the above' buttons
- Update WB link dependency in perseus-editor package

Issue: https://khanacademy.atlassian.net/browse/LEMS-3036

| Before | After | 
| --- | --- |
| Screenshot 2025-06-17 at 2 19 55 PM | image |

## Test plan:
Storybook
- Add a Radio widget
- Confirm that the links and buttons all use wonder blocks
- Confirm that the links and buttons still all work as expected
-


packages/perseus-editor/src/components/labeled-switch.tsx
nishasy, line 1: Moved this into the common components directory. It was previously in the interactive graph directory.

packages/perseus-editor/src/widgets/radio/editor.tsx
```diff
@@ -301,50 +295,42 @@ class RadioEditor extends React.Component<RadioEditorProps> {
...
+                    />
+                    <LabeledSwitch
+                        label="Multiple selections"
+                        checked={this.props.multipleSelect}
+                        onChange={(value) => {
+                            this.onMultipleSelectChange({
+                                multipleSelect: value,
+                            });
+                        }}
+                        style={{marginBottom: spacing.xxSmall_6}}
```
SonicScrewdriver: Iirc, I think we want to use sizing wherever possible, as spacing is going away eventually.

packages/perseus-editor/src/widgets/radio/editor.tsx
```diff
@@ -391,25 +377,23 @@ class RadioEditor extends React.Component<RadioEditorProps> {
...
-                            href="#"
-                            // eslint-disable-next-line react/jsx-no-bind
-                            onClick={this.addChoice.bind(this, true)}
-                        >
-                            <InlineIcon {...iconPlus} /> None of the above{" "}
-                        </a>
-                    )}
+                        Add a choice
+                    </Button>
+                    <Strut size={spacing.large_24} />
```
SonicScrewdriver, line 388: We're stuck with spacing here though, as strut requires it and does not support sizing.

SonicScrewdriver (approved):
Looks great! The trash icon has been driving me crazy for a long time. 

I only have one small comment about using sizing tokens wherever possible, but I also don't know that I consider it a major issue either.

packages/perseus-editor/src/widgets/radio/editor.tsx
```diff
@@ -301,50 +295,42 @@ class RadioEditor extends React.Component<RadioEditorProps> {
...
-                        <Checkbox
-                            label="Multiple selections"
-                            checked={this.props.multipleSelect}
+                    <LabeledSwitch
+                        label="Randomize order"
+                        checked={this.props.randomize}
+                        onChange={(value) => {
+                            this.props.onChange({randomize: value});
+                        }}
+                        style={{marginBottom: spacing.xxSmall_6}}
```
mark-fitzgerald: Even though this is in an editing screen, I think we should get into the habit of using some newer techniques. Instead of `marginBottom` or `marginLeft`, we should be using `marginBlockEnd` and `marginInlineStart` (respectively) to aid in right-to-left (RTL) languages. Also, for measurements, we should start using `rem` units, so this style could be `marginBlockEnd: "0.6rem"`.

packages/perseus-editor/src/widgets/radio/editor.tsx
mark-fitzgerald: My preference is to not add a `<div>` just to create a space. Is there a reason that we can't just add some margin between the two?

packages/perseus-editor/src/widgets/radio/editor.tsx
nishasy: I updated this to use the `sizing` tokens, which use rem, so that should be good now.

I'm a little surprised the `marginBlockEnd` is still recommended for RTL when "bottom" is still underneath for both RTL and LTR languages. I thought using `marginStart`/`marginEnd` was only needed to replace `marginLeft` and `marginRight`?

mark-fitzgerald (approved):
This looks and feels so much better!

Just a couple questions regarding styling options.

packages/perseus-editor/src/widgets/radio/editor.tsx
mark-fitzgerald: They (`marginStart/End`) are. I only brought them into the conversation for reference.

PR #2620, opened 2025-06-17
docs(changeset): Update accessibility status for Perseus Widgets previously inaccessible

tatianasnook:
## Summary:
Updated the accessibility field for the following Perseus widgets to align with our Widget Fundamentals dicumentation:
- Graded Group
- Graded Group Set
- Python Program
- Video

Issue: LEMS-3196

## Test plan:

### DIFF
--- a/.changeset/young-pandas-build.md
+++ b/.changeset/young-pandas-build.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-core": major
+---
+
+Update accessibility status for Perseus Widgets previously inaccessible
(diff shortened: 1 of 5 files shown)


jeremywiebe (approved):

.changeset/young-pandas-build.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/perseus-core": major
```
catandthemachines: Let's make this a minor or a patch update, (personally I would do a patch since this is a small bug). We don't need to be bumping major versions for this change. 😉

catandthemachines (approved):
Approved! But please adjust the changeset severity before landing.

PR #2621, opened 2025-06-17
fix: revert LEMS-3061

anakaren-rojas:
## Summary:
Reverts changes introduced by: https://github.com/Khan/perseus/pull/2435

Issue: LEMS-XXXX

## Test plan:


packages/perseus/src/widgets/radio/radio-component.tsx
```diff
@@ -243,6 +256,84 @@ class Radio extends React.Component<Props> implements Widget {
         return _getPromptJSON(this.props, userInput);
     }
 
+    /**
```
SonicScrewdriver: We definitely don't want to bring back `showRationalesForCurrentlySelectedChoices` or `deselectIncorrectSelectedChoices`! I'm not sure how that got mixed up in your pr work.

packages/perseus/src/widgets/radio/radio-component.tsx
```diff
@@ -110,6 +114,15 @@ class Radio extends React.Component<Props> implements Widget {
         };
     }
 
+    componentDidUpdate(prevProps: Props) {
```
SonicScrewdriver: We'll want to delete this entire componentDidUpdate as well

packages/perseus/src/widgets/radio/radio.new.tsx
```diff
@@ -29,6 +29,7 @@ export interface RenderProps {
     deselectEnabled?: boolean;
     choices: ReadonlyArray<RadioChoiceWithMetadata>;
     selectedChoices: ReadonlyArray<PerseusRadioChoice["correct"]>;
+    showSolutions?: ShowSolutions;
```
SonicScrewdriver, line 32: Huh strange, I wonder why this was removed as I don't see anything calling it here being added, and we're not getting any linter errors. Either way, it makes sense for this line to stay to me as we're definitely using showSolutions further down the radio hierarchy

SonicScrewdriver (changes requested):
Hey! I think we can fix this up pretty fast. I left some comments about things that should be removed. I'm not sure how they snuck in as part of your PR but they're also the only things I'm seeing that need adjustment.

packages/perseus/src/widgets/radio/radio.new.tsx
SonicScrewdriver: I was actually getting linter errors on main! I have no idea how neither of our PRs flagged errors in the linter for this. I've checked this PR in the upstream repo with the snapshot, however, and it works properly. I'm going to consider this solved lol

PR #2622, opened 2025-06-18
Parse ELA radio widgets

benchristel:
In [ELA exercises] on Khan Academy, widgets are parsed by Go code and
re-serialized as JSON before being served to the frontend. The Go parser fills
in zero values for any missing fields, e.g. `nil` (which serializes to JSON
`null`). Previously our Radio widget parser didn't accept `null` values for
`onePerLine` and `noneOfTheAbove`; it wanted `undefined`. As a result, we saw
parse errors in ELA exercises when we tried to deploy the recent Radio v3
upgrade.

This PR fixes the parser so it can handle Radio widgets that have had `null`
filled in for `onePerLine` and `noneOfTheAbove`.

[ELA exercises]: https://www.khanacademy.org/ela/new-6th-grade-reading-and-vocabulary/x8ddea1200317e822:the-rules-we-live-by-mixed-skills-lpassage-practice/x8ddea1200317e822:reading-argumentative-texts/e/a-seriously-play-ful-proposal-a-text-about-recess

Issue: none

## Test plan:

`pnpm test`

### DIFF
--- a/packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
+++ b/packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
@@ -32,6 +32,12 @@ function getDefaultOptions() {
     };
 }
 
+const parseOnePerLine = defaulted(optional(boolean), () => undefined);
+const parseNoneOfTheAbove = defaulted(
+    optional(constant(false)),
+    () => undefined,
+);
+
 const version3 = optional(object({major: constant(3), minor: number}));
 const parseRadioWidgetV3 = parseWidgetWithVersion(
     version3,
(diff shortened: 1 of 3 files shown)


packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
```diff
@@ -32,6 +32,12 @@ function getDefaultOptions() {
     };
 }
 
+const parseOnePerLine = defaulted(optional(boolean), () => undefined);
+const parseNoneOfTheAbove = defaulted(
+    optional(constant(false)),
```
handeyeco, line 37: Could we just make `optional` handle `null` too? My concern is that while this would solve the problem for Anakaren's work, the problem would resurface next time we make a similar schema change.

benchristel: It might cause bugs if we change the output type of the parsers though. Ideally, nothing should be treating `null` and `undefined` differently, but it's somewhat hard to confirm that.

benchristel: I created a separate ticket for treating undefined and null the same: https://khanacademy.atlassian.net/browse/LEMS-3232

That's going to be a bigger chunk of work.

jeremywiebe (approved):

benchristel:
Closed in favor of https://github.com/Khan/perseus/pull/2640

PR #2629, opened 2025-06-20
Restructuring and Renaming New Radio Files (widget and component only) 

SonicScrewdriver:
## Summary:
This PR is part of the Radio Widget Improvement project. 

Please see the Proposed New Hierarchy documents to understand the history of this work. However, here's a quick overview:

1. Merged `radio.new.tsx` and `radio-class.new.tsx` into `multiple-choice-widget.tsx`, which will handle the widget and state logic for the Radio widget. 
2. Moved a very small amount of logic from base-radio into multiple-choice-widget.tsx to keep it focused on UI/Rendering 
3. Renamed base-radio to `multiple-choice-component.tsx`, and reorganized code. 
4. Removed Aphrodite in favour of CSS Modules, and massively cleaned up the class structure 
5. Investigated the `highlighted` property, and marked it for future removal
6. Created new test files for verifying new widget is free from regressions

Issue: LEMS-3235

## Test plan:
- New tests pass 
- Manual testing 

## Video comparison
https://github.com/user-attachments/assets/17e78c58-d72c-4561-9758-8dd2366f8659


packages/perseus/src/widgets/radio/multiple-choice.module.css
```diff
@@ -0,0 +1,114 @@
...
+    min-width: max-content;
+    width: 100%;
+    border-bottom: var(--wb-border-width-thin) solid
+        var(--wb-semanticColor-core-border-neutral-subtle);
+    border-top: var(--wb-border-width-thin) solid
+        var(--wb-semanticColor-core-border-neutral-subtle);
+    scrollbar-width: thin;
+}
+
+@media screen and (max-width: 681px) {
```
SonicScrewdriver, line 28: I want to check with the team about whether I should merge this with the media query above, or keep it co-located for clarity.

packages/perseus/src/widgets/radio/multiple-choice-component.new.tsx
```diff
@@ -126,69 +122,27 @@ const BaseRadio = ({
...
-    }, [apiOptions, choices, isLastUsedWidget, reviewModeRubric]);
-
-    // When a particular choice's `onChange` handler is called, indicating a
-    // change in a single choice's values, we need to call our `onChange`
-    // handler in order to notify our parent. However, our API with our parent
-    // is that we always provide *all* values for *all* choices, even if just
-    // one choice's values changed. (This is because sometimes an interaction
-    // with one choice can affect many choices, like how checking a new answer
-    // will usually cause the old answer to become unchecked.)
-    //
```
SonicScrewdriver, line 138: This is no longer true as we've cleaned up the state management so that we can now achieve this by simply passing up the choiceIndex and new value.

packages/perseus/src/widgets/radio/multiple-choice-component.new.tsx
```diff
@@ -126,69 +122,27 @@ const BaseRadio = ({
...
+        firstChoiceHighlighted &&
+            isMobile &&
+            styles.radioContainerFirstHighlighted,
+        lastChoiceHighlighted &&
+            isMobile &&
+            styles.radioContainerLastHighlighted,
     );
 
     const instructionsClassName = classNames(
-        "instructions",
```
SonicScrewdriver, line 190: We were overriding all of the styles in this class anyway, so it makes sense to simply remove it to keep things clean.

packages/perseus/src/widgets/radio/multiple-choice-component.new.tsx
```diff
@@ -126,69 +121,17 @@ const BaseRadio = ({
...
-
-        onChange({
-            checked: newCheckedList,
-        });
-    }
-
-    // some commonly used shorthands
-    const isMobile = apiOptions.isMobile;
-
-    const firstChoiceHighlighted = choices[0].highlighted;
```
SonicScrewdriver, line 171: It turns out highlighted was never set to true, anywhere in our code! I was able to remove all of this silly logic and useless class styles without any regressions in functionality in either mobile or desktop.

packages/perseus/src/widgets/radio/multiple-choice-component.new.tsx
```diff
@@ -126,69 +121,17 @@ const BaseRadio = ({
...
-    const lastChoiceHighlighted = choices[choices.length - 1].highlighted;
-
-    const className = classNames(
+    const choiceListClassName = classNames(
         "perseus-widget-radio",
-        !editMode && "perseus-rendered-radio",
-        css(
-            styles.radio,
-            styles.responsiveRadioContainer,
```
SonicScrewdriver, line 179: I consolidated radio, responsiveRadioContainer, and fieldSet from the aphrodite styles into a single class as this was quite silly.

packages/perseus/src/widgets/radio/multiple-choice-component.new.tsx
```diff
@@ -126,69 +121,17 @@ const BaseRadio = ({
...
-            firstChoiceHighlighted &&
-                isMobile &&
-                styles.radioContainerFirstHighlighted,
-            lastChoiceHighlighted &&
-                isMobile &&
-                styles.radioContainerLastHighlighted,
-        ),
+        !editMode && "perseus-rendered-radio", // Styles to be applied in the Editor
```
SonicScrewdriver: We might not even need this `editMode` check anymore as we're no longer (weirdly) directly accessing base-radio in the Radio Editor. I'll leave it for now until we are able to fully implement the new editor experience, however.

packages/perseus/src/widgets/radio/__tests__/multiple-choice.test.ts
```diff
@@ -0,0 +1,893 @@
+import {describe, beforeEach, it} from "@jest/globals";
```
SonicScrewdriver, line 1: I figured it made sense to have a completely separate test file, as it's likely that we may have different test conditions with the upcoming new designs. Plus, it's easier to remember that we need to clean it up at the end of the project. ;)

packages/perseus/src/widgets/radio/multiple-choice-widget.new.tsx
```diff
@@ -0,0 +1,400 @@
...
+                choices={choicesProps}
+                onChoiceChange={this.handleChoiceChange}
+            />
+        );
+    }
+}
+
+// Export as Radio for backwards compatibility until we can
+// perform Content Backfills to officially rename the Radio Widget
+const Radio = MultipleChoiceWidget;
```
SonicScrewdriver, line 352: This is completely unnecessary, but I liked having a clear way to call this out. 🤷‍♀️

packages/perseus/src/widgets/radio/multiple-choice.module.css
```diff
@@ -0,0 +1,78 @@
...
+.item {
+    margin-left: var(--wb-sizing-size_200);
+}
+
+.responsiveItem {
+    margin-left: 0;
+    padding: 0;
+}
+
+.responsiveItem:not(:last-child) {
```
SonicScrewdriver, line 44: Tbh, I'm wondering if we really want the "responsive" prefix for these.

packages/perseus/src/widgets/radio/multiple-choice-widget.new.tsx
```diff
@@ -0,0 +1,401 @@
...
+        | "linterContext"
+        | "showSolutions"
+    >
+>;
+
+/**
+ * MultipleChoiceWidget implements the Widget interface for multiple choice questions.
+ *
+ * It handles Perseus-specific logic and state management while delegating
+ * UI rendering to the MultipleChoiceComponent.
```
mark-fitzgerald, line 72: All these comments are fabulous!!! Our future (and current) selves thank you!

packages/perseus/src/widgets/radio/multiple-choice-widget.new.tsx
```diff
@@ -0,0 +1,401 @@
...
+     */
+    static getUserInputFromProps(
+        props: Props,
+        unshuffle: boolean = true,
+    ): PerseusRadioUserInput {
+        // Return checked inputs in the form {choicesSelected: [bool]}. (Dear
+        // future timeline implementers: this used to be {value: i} before
+        // multiple select was added)
+        if (props.choiceStates) {
+            const choiceStates = props.choiceStates;
```
mark-fitzgerald: Nit: For consistency with line 117 (`values`), should this be moved above the `if()` statement?

packages/perseus/src/widgets/radio/multiple-choice.module.css
```diff
@@ -0,0 +1,78 @@
+/* CSS module for MultipleChoiceComponent */
+
+.instructions {
+    display: block;
+    color: var(--wb-semanticColor-core-foreground-instructive-strong);
+    font-size: 14px;
```
mark-fitzgerald: We should be using `rem` instead of `px`. That would make this:
```
font-size: 1.4rem;
```

packages/perseus/src/widgets/radio/multiple-choice.module.css
```diff
@@ -0,0 +1,78 @@
...
+    font-style: normal;
+    font-weight: bold;
+    margin-bottom: var(--wb-sizing-size_160);
+}
+
+.instructionsMobile {
+    font-size: 18px;
+}
+
+@media screen and (max-width: 681px) {
```
mark-fitzgerald: RE: use of `rem`
Media queries don't have access to the root font size, so we still use `px` when defining them. So, leave this one (and other media queries) as is. However, the `font-size` setting within the media query (i.e. line 20) should be converted to `rem`.

packages/perseus/src/widgets/radio/multiple-choice.module.css
mark-fitzgerald: Co-located is nice for simple changes. However, if there are a lot then I like to keep all the media query changes together in one.

packages/perseus/src/widgets/radio/multiple-choice.module.css
mark-fitzgerald: Wondering if we should be using `@media` queries, or `@container` queries, or nothing at all. Why is a screen width important for these font sizes? Is this a mobile setting? Why would we care what the screen/container size is when defining the font size?

packages/perseus/src/widgets/radio/multiple-choice.module.css
```diff
@@ -0,0 +1,78 @@
...
+        var(--wb-semanticColor-core-border-neutral-subtle);
+    border-top: var(--wb-border-width-thin) solid
+        var(--wb-semanticColor-core-border-neutral-subtle);
+    scrollbar-width: thin;
+}
+
+@media screen and (max-width: 681px) {
+    .choiceList {
+        margin-left: calc(-1 * var(--wb-sizing-size_160));
+        margin-right: calc(-1 * var(--wb-sizing-size_160));
```
mark-fitzgerald: We should be using `margin-inline-start/end` instead of `margin-left` and `margin-right`. Enables RTL support.

packages/perseus/src/widgets/radio/multiple-choice.module.css
```diff
@@ -0,0 +1,78 @@
...
+}
+
+.selectedItem {
+    background: var(--wb-semanticColor-core-foreground-inverse-strong);
+}
+
+.responsiveContainer {
+    overflow: auto;
+    margin-left: calc(-1 * var(--wb-sizing-size_160));
+    padding-left: var(--wb-sizing-size_160);
```
mark-fitzgerald: Padding should be `padding-inline-start/end`, as well.

packages/perseus/src/widgets/radio/multiple-choice.module.css
SonicScrewdriver: Yeah I thought this was strange as well! The font size IS bigger on mobile, but it seems incredibly arbitrary to change the font size for _just_ a single line of text. 

The default size of 14px is also much smaller than the rest of our text in our content. I'd say it makes sense to just increase this to the rem equivalent of 18px for consistency, which should solve any readability issues on mobile. I can't find the holistic design mockups right now that show the initial question text, but I suspect this is probably inline with our upcoming designs as well. 

Screenshot example of the current mobile style: 
Screenshot 2025-06-25 at 3 14 07 PM

SonicScrewdriver: Here's an example showing the flat 1.8rem size. 

Screenshot 2025-06-25 at 3 21 21 PM

packages/perseus/src/widgets/radio/multiple-choice-widget.new.tsx
SonicScrewdriver: I like that! Seems much clearer, thank you

packages/perseus/src/widgets/radio/multiple-choice-component.new.tsx
nishasy: I don't like that it was written this way in the first place 😭

packages/perseus/src/widgets/radio/multiple-choice-widget.new.tsx
```diff
@@ -0,0 +1,401 @@
...
+
+/**
+ * RadioChoiceWithMetadata is used for server-side scoring
+ */
+export interface RadioChoiceWithMetadata extends PerseusRadioChoice {
+    originalIndex: number;
+    correct?: boolean;
+}
+
+type Props = WidgetProps<RenderProps, PerseusRadioRubric>;
```
anakaren-rojas: the props going into the component should be converted to interfaces, but it doesn't seem like we can do a straight conversion using `WidgetProps` since its a type
This is what cursor suggests: 
```suggestion
interface Props extends RenderProps, UniversalWidgetProps<PerseusRadioRubric>{};
```

packages/perseus/src/widgets/radio/multiple-choice-widget.new.tsx
```diff
@@ -0,0 +1,401 @@
...
+     * 3. Renders content and rationales with nested widget support
+     * 4. Determines correctness display based on choice data or review rubric
+     *
+     * @param choiceStates - The current state of each choice
+     * @returns An array of formatted choice props ready for the MultipleChoiceComponent
+     */
+    buildChoiceProps = (
+        choiceStates: ReadonlyArray<ChoiceState>,
+    ): ReadonlyArray<ChoiceType> => {
+        const {
```
anakaren-rojas, line 236: these can be destructured in the param

anakaren-rojas (commented):
my initial nit-pick pass for the coding styles

packages/perseus/src/widgets/radio/multiple-choice-widget.new.tsx
SonicScrewdriver: I'm not sure we want to be changing WidgetProps yet as it's the standard across our widgets. Iirc we're not barred from using types, we just want to default to using interfaces unless we have a reason to do so. 🤔

anakaren-rojas: I see, I was under the impression we wanted to change all of the Props in the new files to interfaces

packages/perseus/src/widgets/radio/multiple-choice-widget.new.tsx
SonicScrewdriver: Hmmm I think I'm not quite grepping this! Do you mind helping me understand which item you mean here? I'd be happy to make any adjustments. :) 

Do you mean the choiceStates? We're destructuring it in the map due to the array, which I think makes sense?

packages/perseus/src/widgets/radio/multiple-choice-widget.new.tsx
SonicScrewdriver: Hmm it might be worth discussing this afternoon as I could have misunderstood!

SonicScrewdriver: After discussion with the team, we'll focus on doing this holistically later on.

packages/perseus/src/types.ts
```diff
@@ -110,6 +110,8 @@ export type EditorMode = "edit" | "preview" | "json";
 
 export type ChoiceState = {
     selected: boolean;
+    // Deprecated; never implemented
+    // TODO(third): Remove this field when we remove the old Radio files (LEMS-2994)
```
nishasy, line 138: (Optional) Nit: Could you update this to say the following? And then update the [LEMS-2994 ticket](https://khanacademy.atlassian.net/browse/LEMS-2994) to say "Look for `TODO(LEMS-2994)`"

```suggestion
    // TODO(LEMS-2994): Remove this field when we remove the old Radio files
```

packages/perseus/src/widgets/radio/utils/general-utils.test.ts
```diff
@@ -21,8 +21,9 @@ describe("getChoiceStates", () => {
 
     const defaultState: ChoiceState = {
         selected: false,
-        readOnly: false,
+        // TODO(third): Remove this field when we remove the old Radio files (LEMS-2994)
```
nishasy, line 24: (Option) Nit: Same here

packages/perseus/src/widgets/radio/multiple-choice-component.new.tsx
nishasy: For posterity, Third and I figured out what the highlight style actually was if manually set to true, and it was totally unhelpful. It's good that it's not being used, and we should definitely get rid of it.

packages/perseus/src/widgets/radio/multiple-choice-component.new.tsx
nishasy: Could you add a comment
```
TODO(LEMS-3229): Remove this line after we stop using BaseRadio in editor
```

packages/perseus/src/widgets/radio/multiple-choice.module.css
nishasy: We may want to leave it for now just to have an obvious 1:1 with the old style names, but we could revisit the naming later.

packages/perseus/src/widgets/radio/__tests__/multiple-choice.test.ts
```diff
@@ -0,0 +1,893 @@
...
+    const options = screen.getAllByRole("radio");
+
+    // element that is null/undefined (ie. if the index is invalid) so we
+    // manually check and throw here to protect future me, and others :)
+    if (index > options.length) {
+        throw new Error("Invalid array index for radio");
+    }
+
+    await userEvent.click(options[index]);
+};
```
nishasy, line 47:
```suggestion
};

```

packages/perseus/src/widgets/radio/__tests__/multiple-choice.test.ts
```diff
@@ -0,0 +1,893 @@
...
+            // Arrange
+            renderQuestion(question, apiOptions);
+
+            // Act
+            await userEvent.tab();
+            await userEvent.keyboard(" ");
+
+            // Assert
+            expect(screen.getAllByRole("radio")[0]).toBeChecked();
+        });
```
nishasy, line 253: Can you add another "should be able to select an option by keyboard" test for the Enter key?

packages/perseus/src/widgets/radio/__tests__/multiple-choice.test.ts
```diff
@@ -0,0 +1,893 @@
...
+                multipleCorrectChoicesQuestion,
+                renderer.getUserInputMap(),
+            );
+
+            // Assert
+            expect(score).toHaveInvalidInput("CHOOSE_CORRECT_NUM_ERROR");
+        });
+
+        it.each(incorrect)(
+            "should reject an incorrect answer - test #%#",
```
nishasy, line 707: What does "test #%#" mean?

packages/perseus/src/widgets/radio/multiple-choice-component.new.tsx
```diff
@@ -35,65 +33,62 @@ export interface ChoiceType {
...
     previouslyAnswered: boolean;
     revealNoneOfTheAbove: boolean;
     disabled: boolean;
 }
 
-interface BaseRadioProps {
+/**
+ * Props for the MultipleChoiceComponent
+ */
+interface MultipleChoiceComponentProps {
```
nishasy, line 44: The `reviewMode` comment is super helpful! Would it be possible to add comments explaining what all the other props are for too?

packages/perseus/src/widgets/radio/multiple-choice-component.new.tsx
```diff
@@ -249,34 +187,13 @@ const BaseRadio = ({
...
-                                    styles.nextHighlighted,
-                            );
-                        };
-
-                        // HACK(abdulrahman): Preloads the selection-state
-                        // css because of a bug that causes iOS to lag
-                        // when selecting the button for the first time.
-                        aphroditeClassName(true);
+                        // Whether or not to show correctness borders
+                        // for this choice and the next choice.
```
nishasy: Just for my own understanding, what is a correctness border?

packages/perseus/src/widgets/radio/multiple-choice-widget.new.tsx
```diff
@@ -0,0 +1,401 @@
...
+export type RenderProps = {
+    numCorrect: number;
+    hasNoneOfTheAbove?: boolean;
+    multipleSelect?: boolean;
+    countChoices?: boolean;
+    deselectEnabled?: boolean;
+    choices: ReadonlyArray<RadioChoiceWithMetadata>;
+    selectedChoices: ReadonlyArray<PerseusRadioChoice["correct"]>;
+    choiceStates?: ReadonlyArray<ChoiceState>;
+    // Depreciated; support for legacy way of handling changes
```
nishasy:
```suggestion
    // Deprecated; support for legacy way of handling changes
```

nishasy (approved):
This is an amazing improvement! I left a couple of small nits, nothing blocking.

packages/perseus/src/widgets/radio/__tests__/multiple-choice.test.ts
SonicScrewdriver: I didn't make these tests, I just copied them over to match the radio widget. But sure, I can add another one.

packages/perseus/src/widgets/radio/__tests__/multiple-choice.test.ts
SonicScrewdriver: No idea, I just copied over these tests from the original radio tests. I'm guessing it's simply testing symbols that are obviously an incorrect answer.

packages/perseus/src/widgets/radio/multiple-choice-widget.new.tsx
SonicScrewdriver: Sure we might as well fix this up

packages/perseus/src/widgets/radio/multiple-choice-component.new.tsx
SonicScrewdriver: Hmm this might be worth a larger discussion about where we should be documenting our props, considering we're declaring them both here and in types. I added a comment for edit mode though since that was the other one that felt a little unclear for me.

packages/perseus/src/widgets/radio/multiple-choice-component.new.tsx
SonicScrewdriver: No idea, this comment is an artifact from before. Perhaps related to highlighted. I'll remove it.

packages/perseus/src/widgets/radio/multiple-choice-component.new.tsx
nishasy: Oh! I tooootally missed that they're already in the types. You can ignore my comment then.

ivyolamit (approved):
I like the new name. 👍🏼 
Changes looks logical to me.

PR #2630, opened 2025-06-20
docs(changeset): Fixed bugs related to numCorrect not updating properly for the Radio widget

SonicScrewdriver:
## Summary:
We had 2 bugs reported related to the numCorrect value not correctly updating for the Radio Widget. 

This PR fixes two of these bugs by ensuring that we're recalculating the numCorrect value when we delete choices, or switch in and out of multiple selection mode. 

Unfortunately, I have not been able to recreate the third bug in the ticket. 

Issue: LEMS-3230

## Test plan:
- New tests pass

### DIFF
--- a/packages/perseus-core/src/widgets/radio/derive-num-correct.ts
+++ b/packages/perseus-core/src/widgets/radio/derive-num-correct.ts
@@ -1,7 +1,7 @@
 import type {PerseusRadioWidgetOptions} from "../../data-schema";
 
 export function deriveNumCorrect(options: PerseusRadioWidgetOptions) {
-    const {choices, numCorrect} = options;
+    const {choices} = options;
 
-    return numCorrect ?? choices.filter((c) => c.correct).length;
+    return choices.filter((c) => c.correct).length;
 }
--- a/packages/perseus-core/src/widgets/radio/derive-num-correct.ts
+++ b/packages/perseus-core/src/widgets/radio/derive-num-correct.ts
@@ -1,7 +1,7 @@
 import type {PerseusRadioWidgetOptions} from "../../data-schema";
 
-export function deriveNumCorrect(options: PerseusRadioWidgetOptions) {
-    const {choices} = options;
-
+export function deriveNumCorrect(
+    choices: PerseusRadioWidgetOptions["choices"],
+) {
     return choices.filter((c) => c.correct).length;
 }
--- a/packages/perseus-editor/src/widgets/__tests__/radio-editor.test.tsx
+++ b/packages/perseus-editor/src/widgets/__tests__/radio-editor.test.tsx
@@ -23,6 +23,19 @@ function renderRadioEditor(onChangeMock = () => undefined) {
 
 describe("radio-editor", () => {
     let userEvent: UserEvent;
+    function getCorrectChoice(): PerseusRadioChoice {
+        return {
+            content: "",
+            correct: true,
+        };
+    }
+
+    function getIncorrectChoice(): PerseusRadioChoice {
+        const choice = getCorrectChoice();
+        choice.correct = false;
+        return choice;
+    }
+
     beforeEach(() => {
         userEvent = userEventLib.setup({
             advanceTimers: jest.advanceTimersByTime,
@@ -147,19 +160,6 @@ describe("radio-editor", () => {
     it("derives num correct when serializing", () => {
         const editorRef = React.createRef<RadioEditor>();
 
-        function getCorrectChoice(): PerseusRadioChoice {
-            return {
-                content: "",
-                correct: true,
-            };
-        }
-
-        function getIncorrectChoice(): PerseusRadioChoice {
-            const choice = getCorrectChoice();
-            choice.correct = false;
-            return choice;
-        }
-
         render(
             <RadioEditor
                 ref={editorRef}
@@ -183,19 +183,6 @@ describe("radio-editor", () => {
 
     it("derives num correct when calling onChange", async () => {
         const onChangeMock = jest.fn();
-
-        function getCorrectChoice(): PerseusRadioChoice {
-            return {
-                content: "",
-                correct: true,
-            };
-        }
-
-        function getIncorrectChoice(): PerseusRadioChoice {
-            const choice = getCorrectChoice();
-            choice.correct = false;
-            return choice;
-        }
 
         render(
             <RadioEditor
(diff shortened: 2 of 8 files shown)


packages/perseus-editor/src/widgets/radio/editor.tsx
```diff
@@ -116,39 +116,55 @@ class RadioEditor extends React.Component<RadioEditorProps> {
...
-                multipleSelect: allowMultiple,
-            });
+        const multipleSelect = allowMultiple.multipleSelect;
+
+        // When switching to single-select mode, if multiple options are selected,
+        // we need to deselect all options
+        let choices = this.props.choices;
+        if (!multipleSelect) {
+            const numCorrect = _.reduce(
+                this.props.choices,
```
mark-fitzgerald: Should this be the local `choices` object defined on line 123?

packages/perseus-editor/src/widgets/radio/editor.tsx
```diff
@@ -116,39 +116,55 @@ class RadioEditor extends React.Component<RadioEditorProps> {
...
-            this.props.onChange({
-                multipleSelect: allowMultiple,
-            });
+        const multipleSelect = allowMultiple.multipleSelect;
+
+        // When switching to single-select mode, if multiple options are selected,
+        // we need to deselect all options
+        let choices = this.props.choices;
+        if (!multipleSelect) {
+            const numCorrect = _.reduce(
```
mark-fitzgerald: Would this be a good time to remove underscore, or would that be a big distraction from the purpose of this PR?

packages/perseus-editor/src/widgets/radio/editor.tsx
```diff
@@ -116,39 +116,55 @@ class RadioEditor extends React.Component<RadioEditorProps> {
...
-            });
-            this.props.onChange({
-                multipleSelect: allowMultiple,
-                choices: choices,
-            });
-        } else {
-            this.props.onChange({
-                multipleSelect: allowMultiple,
-            });
+        const multipleSelect = allowMultiple.multipleSelect;
```
mark-fitzgerald: Do we have a naming convention for boolean values like this? For instance, `isMultipleSelect`?

packages/perseus-editor/src/widgets/radio/editor.tsx
SonicScrewdriver: Let's do it! I'm down for any excuse to remove it

packages/perseus-editor/src/widgets/radio/editor.tsx
SonicScrewdriver: Tbh I don't know! It just bugged me that we were extracting multipleSelect out of allowMultiple and then calling it allowMultiple. 😅 I'm happy to make it isMultipleSelect though!

packages/perseus-editor/src/widgets/__tests__/radio-editor.test.tsx
```diff
@@ -235,4 +240,154 @@ describe("radio-editor", () => {
...
+        await userEvent.click(
+            screen.getByRole("switch", {
+                name: "Multiple selections",
+            }),
+        );
+
+        // Check that multipleSelect is updated and numCorrect is calculated
+        expect(onChangeMock).toHaveBeenCalledWith(
+            expect.objectContaining({
+                numCorrect: expect.any(Number),
```
mark-fitzgerald: How does this show that `numCorrect` has been calculated?

packages/perseus-editor/src/widgets/__tests__/radio-editor.test.tsx
```diff
@@ -235,4 +240,154 @@ describe("radio-editor", () => {
             }),
         );
     });
+
+    it("updates numCorrect when deleting an option", async () => {
+        const onChangeMock = jest.fn();
+
+        function getCorrectChoice(): PerseusRadioChoice {
```
mark-fitzgerald: Can this function and its "incorrect" counterpart be hoisted to the `describe` parent instead of rebuilt for each test?

packages/perseus-editor/src/widgets/__tests__/radio-editor.test.tsx
SonicScrewdriver: It doesn't! Apparently my tired Friday brain didn't quite finish up this PR. I've fixed this up quite a bit!

packages/perseus-editor/src/widgets/radio/editor.tsx
```diff
@@ -116,39 +116,44 @@ class RadioEditor extends React.Component<RadioEditorProps> {
     };
 
     onMultipleSelectChange: (arg1: any) => any = (allowMultiple) => {
-        allowMultiple = allowMultiple.multipleSelect;
-
-        const numCorrect = _.reduce(
```
SonicScrewdriver, line 117: No idea why were were doing this manually every time. We already have a nifty little method to handle this (deriveNumCorrect)!

packages/perseus-editor/src/widgets/radio/editor.tsx
```diff
@@ -116,39 +116,39 @@ class RadioEditor extends React.Component<RadioEditorProps> {
...
     };
 
     onCountChoicesChange: (arg1: any) => void = (count) => {
-        count = count.countChoices;
-        this.props.onChange({countChoices: count});
+        const countChoices = count.countChoices;
+        this.props.onChange({
+            countChoices,
+        });
     };
```
SonicScrewdriver, line 145: This was silly naming, imo.

packages/perseus-core/src/widgets/radio/derive-num-correct.ts
```diff
@@ -1,7 +1,7 @@
 import type {PerseusRadioWidgetOptions} from "../../data-schema";
 
 export function deriveNumCorrect(options: PerseusRadioWidgetOptions) {
-    const {choices, numCorrect} = options;
+    const {choices} = options;
 
-    return numCorrect ?? choices.filter((c) => c.correct).length;
+    return choices.filter((c) => c.correct).length;
 }
```
SonicScrewdriver, line 7: This could be a spicy move, but I really don't understand why we were ever returning `numCorrect` in the first place, as everything I saw indicates that we always want to recalculate this value. 

This is only used in the radio `editor.tsx` file with `undefined` for `numCorrect`, and in our parser for converting a Radio widget from v1 to a v2 as v1 did not have this option.

packages/perseus-core/src/widgets/radio/derive-num-correct.ts
```diff
@@ -1,7 +1,7 @@
 import type {PerseusRadioWidgetOptions} from "../../data-schema";
 
-export function deriveNumCorrect(options: PerseusRadioWidgetOptions) {
-    const {choices, numCorrect} = options;
-
-    return numCorrect ?? choices.filter((c) => c.correct).length;
+export function deriveNumCorrect(
+    choices: PerseusRadioWidgetOptions["choices"],
```
anakaren-rojas, line 4: should this be destructured? 
```suggestion
    {choices} : PerseusRadioWidgetOptions,
```

SonicScrewdriver: Oooo it could, but I think we've generally used the first format  🤔.

PR #2633, opened 2025-06-23
feat(LEMS-3069): Add Ids to radio choices within perseus parser

anakaren-rojas:
## Summary:
- Uses perseus parser to add ids to radio choices
- Updates parser test and mock data include v3 parser object with choice ids 
- Updates widget map test to use latest version of parser

Issue: LEMS-3069

## Test plan:
znd: https://prod-znd-250623-769-ddbda54.khanacademy.org/


packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.test.ts
```diff
@@ -120,6 +128,10 @@ describe("migration functions", () => {
     });
 
     it("migrates v2 to v3", () => {
-        expect(migrateV2toV3(v2Widget)).toEqual(v3Widget);
+        expect(migrateV2ToV3(v2Widget)).toEqual(v3Widget);
```
nishasy, line 131: I'm very confused about why it's saying this line is different 🤔

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
```diff
@@ -161,7 +187,35 @@ const parseRadioWidgetV0 = parseWidgetWithVersion(
...
+        version: {major: 4, minor: 0},
+        options: {
+            numCorrect: options.numCorrect,
+            hasNoneOfTheAbove: options.hasNoneOfTheAbove,
+            countChoices: options.countChoices,
+            randomize: options.randomize,
+            multipleSelect: options.multipleSelect,
+            deselectEnabled: options.deselectEnabled,
+            choices: options.choices.map((choice, index) => {
+                const choiceId = `choice-${index + 1}`;
```
nishasy, line 200: Is this going to cause issues if there are multiple radio widgets on the same page, like in an article? Choices may have repeated IDs, which could pose a problem depending on what these IDs are used for.

nishasy (commented):
I just left a question about using the index for the ID.

Note: You may also want to add the new ID field to the [radio builder choices](https://github.com/Khan/perseus/blob/main/packages/perseus/src/widgets/radio/radio-question-builder.ts#L69-L73) as default values.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
anakaren-rojas: Oh! I forgot to add the note to this PR. This is just the initial implementation, in the next PR I update this to be based on the content and index
Here it is for your reference: https://github.com/Khan/perseus/pull/2638/files#diff-67dcfd25b1d1e4eaca771b5dd24d6e359bc86b9b1a701492eab0a4477b8400d5R206

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.test.ts
anakaren-rojas: i capitalized the `t` 😆

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
nishasy: Thank you!

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
```diff
@@ -38,6 +38,32 @@ const parseNoneOfTheAbove = defaulted(
...
+    defaulted(
+        object({
+            numCorrect: optional(number),
+            choices: array(
+                object({
+                    content: defaulted(string, () => ""),
+                    rationale: optional(string),
+                    correct: optional(boolean),
+                    isNoneOfTheAbove: optional(boolean),
+                    id: optional(string),
```
benchristel, line 54: I think `id` should be required, right?

```suggestion
                    id: string,
```

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
```diff
@@ -161,7 +187,35 @@ const parseRadioWidgetV0 = parseWidgetWithVersion(
...
+                const choiceId = `choice-${index + 1}`;
+                return {
+                    content: choice.content,
+                    rationale: choice.rationale,
+                    correct: choice.correct,
+                    isNoneOfTheAbove: choice.isNoneOfTheAbove,
+                    id: choiceId,
+                };
+            }),
+        },
```
benchristel: Could this be

```suggestion
        options: {
            ...options,
            choices: options.choices.map((choice, index) => {
                return {
                    ...choice,
                    id: `choice-${index + 1}`,
                };
            }),
        },
```

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
handeyeco: It looks like this is a stack of PRs so I'm guessing you're going to land these all at once and do a single release? I think that would be ideal.

In case it's helpful: [Stacked Pull-Requests](https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/3588325403/Stacked+Pull-Requests)

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
anakaren-rojas: It doesn't seem like I can make this required without affecting previous migration functions - namely  `migrateV1ToV2` because it calls `deriveNumCorrect`. 
Here's the sequence: 
If I make this required, then I would have to update the `PerseusRadioWidgetOptions` to also require `id` 
Since the props for `deriveNumCorrect` are `PerseusRadioWidgetOptions`, the migration from v1 to v2 would require an ID

jeremywiebe: In our schema, `PerseusRadioWidgetOptions` is always the "current" schema version.  The parser is responsible for moving us from any previous widget option version to the current.  So I think I'd tend to agree with Ben.  `id` in `PerseusRadioChoice` should be non-optional, I think.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
```diff
@@ -55,8 +81,8 @@ const parseRadioWidgetV3 = parseWidgetWithVersion(
         hasNoneOfTheAbove: optional(boolean),
         countChoices: optional(boolean),
         randomize: optional(boolean),
-        multipleSelect: optional(boolean),
         deselectEnabled: optional(boolean),
+        multipleSelect: optional(boolean),
```
jeremywiebe, line 85: Is there a reason this needs to move?  IMHO we should be extremely cautious with touching our prior-version parsers.

packages/perseus-core/src/data-schema.ts
```diff
@@ -1386,6 +1386,7 @@ export type PerseusRadioChoice = {
     // If this is none of the above, override the content with "None of the above"
     // NOTE: perseus_data.go says this is required even though it isn't necessary.
     isNoneOfTheAbove?: boolean;
+    id?: string;
```
jeremywiebe, line 1389: If our v4 schema will always have an ID, then I think we can remove the `?` and make this required.

PR #2634, opened 2025-06-23
[Build] Use postcss-url to copy font files to output bundle

jeremywiebe:
## Summary:

When we landed #2431 (specifically when we moved from `rollup-plugin-styles` to `rollup-plugin-postcss` [here](https://github.com/Khan/perseus/pull/2431/files#diff-45f213140a165c743b42e38a00d97984fcec3f841195e8587d03a861d402ca17R151)), we lost some behaviour that rewrote `url()`'s found in CSS so they were relative to the `dist/` output folder and copied files they pointed to also. 

This PR fixes the build to use the `postcss-url` plugin to restore this behaviour. 

Issue: "none"

## Test plan:

`pnpm build` -> check out the `dist` folder for `packages/math-input` and ensure that there's an `assets/` folder with Symbola font files in it (also that the process CSS in the dist folder points to the font files).

### DIFF
--- a/config/build/rollup.config.js
+++ b/config/build/rollup.config.js
@@ -168,8 +168,8 @@ const createConfig = (
                     // `patches/postcss-url.patch`. If you ever upgrade to a
                     // newer version of this plugin, please double-check that:
                     //   a) the font file copying still works (ie. the fonts are
-                    //      copied to the `math-input/dist/assets` folder and/or
-                    //   b) if the patch is even needed anymore)
+                    //      copied to the `math-input/dist/assets` folder) and/or
+                    //   b) if the patch is even needed anymore
                     postcssUrl({
                         url: "copy",
                         // The postcssUrl() plugin doesn't know anything about


config/build/rollup.config.js
```diff
@@ -152,7 +154,38 @@ const createConfig = (
...
+                    // This is mostly for the Symbola fonts referenced by
+                    // Mathquill CSS.
+                    //
+                    // CAUTION: This PostCSS plugin is patched (using `pnpm
+                    // patch`). You can see the patch in this repo at
+                    // `patches/postcss-url.patch`. If you ever upgrade to a
+                    // newer version of this plugin, please double-check that:
+                    //   a) the font file copying still works (ie. the fonts are
+                    //      copied to the `math-input/dist/assets` folder and/or
+                    //   b) if the patch is even needed anymore)
```
handeyeco: Was this meant to be this?
```suggestion
                    //   a) the font file copying still works (ie. the fonts are
                    //      copied to the `math-input/dist/assets` folder) and/or
                    //   b) if the patch is even needed anymore
```

patches/postcss-url.patch
```diff
@@ -0,0 +1,25 @@
+diff --git a/.editorconfig b/.editorconfig
```
handeyeco: Shesh, what a pain this must have been!

handeyeco (approved):
Thanks Jeremy! This is some big-brain stuff 🧠

patches/postcss-url.patch
jeremywiebe: It was. At first I thought I'd have to fork the whole repo, but pnpm provides a nice way to patch a dependency without having to fork the whole deal. I think this should be fine as this npm package hasn't been updated in 4 years. It works, just is unmaintained (but there are no real other alternatives).

PR #2638, opened 2025-06-24
feat(LEMS-3070): Consume choice ids within the radio widget

anakaren-rojas:
## Summary:
- Initializes choice ids within radio widget
- Consumes choice ids
- Updates test data to include choice ids 

### Low level summaries 
**In radio-widget.ts:** 
- Updates choice id to include a hash of the content

**In choice-id-utils.ts** 
Adds two utility functions: 
- convert strings to hash values using djb2 hash algo
- generateChoiceId combines the content and an index and uses convertStringToHash to generate an id
Reference: https://mojoauth.com/hashing/bernsteins-hash-djb2-in-typescript/

### Implementation Notes
- `data-testid` is added temporarily to test that choice id is being passed correctly in `Frontend`. 

Issue: LEMS-3070

## Test plan:
[ZND](https://prod-znd-250709-1611-7e1ba25.khanacademy.org/)

## Screen recording
https://github.com/user-attachments/assets/4bf875d4-9b2a-433f-b089-48affcca90bb


packages/perseus/src/widgets/radio/base-radio.new.tsx
```diff
@@ -344,6 +346,7 @@ const BaseRadio = ({
                                         ? undefined
                                         : captureScratchpadTouchStart
                                 }
+                                data-testid={choice.id}
```
anakaren-rojas: data test ids added temporarily for testing

packages/perseus-core/src/widgets/radio/choice-id-utils.ts
```diff
@@ -0,0 +1,43 @@
+/**
+ * Hash function to create a short hash from a string
+ * Based on djb2 hash algo for better distribution
+ */
+export function convertStringToHash(input: string): string {
+    let hash = 5381; // Use djb2 hash algorithm constant
+    if (input.length === 0) {
+        return "empty";
+    }
```
benchristel: This is pretty weird. Why return `empty` instead of letting the magic constant get base36'd?

packages/perseus-core/src/widgets/radio/choice-id-utils.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    for (let i = 0; i < input.length; i++) {
+        hash = (hash * 33) ^ input.charCodeAt(i);
+    }
+
+    // Base36 for shorter string, 8 chars for consistent, short IDs
```
benchristel: `substring` will return strings shorter than 8 chars if the input is short, so it's not "consistent". I don't think string formatting really belongs in this function; a hash function should just return an integer and let the caller decide how best to format it for their needs. E.g. the length and compatibility tradeoffs between bases 10, 16, 36, and 64 are application-dependent.

packages/perseus-core/src/widgets/radio/choice-id-utils.ts
```diff
@@ -0,0 +1,43 @@
+/**
+ * Hash function to create a short hash from a string
+ * Based on djb2 hash algo for better distribution
+ */
+export function convertStringToHash(input: string): string {
```
benchristel: I could have sworn we had a hash function already, but now I can't find it.

Since hashing is something we're likely to reuse, I'd prefer to put it somewhere easily findable, like `perseus-core/src/utils/hash.ts`.

packages/perseus-core/src/widgets/radio/choice-id-utils.ts
```diff
@@ -0,0 +1,43 @@
...
+    if (input.length === 0) {
+        return "empty";
+    }
+
+    for (let i = 0; i < input.length; i++) {
+        hash = (hash * 33) ^ input.charCodeAt(i);
+    }
+
+    // Base36 for shorter string, 8 chars for consistent, short IDs
+    return Math.abs(hash).toString(36).substring(0, 8);
```
benchristel: At first I couldn't figure out how `hash` could become negative — then I realized that the conversion to a 32-bit _signed_ int via the XOR operator (^) could produce a negative.

I _think_ the possibility of a negative could be avoided if we use the [original algorithm](https://theartincode.stanis.me/008-djb2/) which uses a shift and add instead of multiplying. However, I am not sure about that. Note that the code comment in that blog post does mention `hash * 33 + c` as an alternative, but I think that only works if your language has unsigned int arithmetic, which JS doesn't.

I think the `Math.abs` at least deserves a comment. However, I think that whichever way we go, the hashes produced by this function will be inconsistent with the ones from the true djb2 hash. That might not matter for the quality of the hash — or it might. I'm not good enough at math to be able to tell.

My takeaway from this is that we probably shouldn't be writing our own hash functions, and we definitely shouldn't be letting AI generate weird versions of well-known algorithms 😉 

Is there something from `window.crypto` that we can use? E.g. an `md5` implementation?

packages/perseus-core/src/widgets/radio/choice-id-utils.ts
```diff
@@ -0,0 +1,43 @@
...
+ * Based on djb2 hash algo for better distribution
+ */
+export function convertStringToHash(input: string): string {
+    let hash = 5381; // Use djb2 hash algorithm constant
+    if (input.length === 0) {
+        return "empty";
+    }
+
+    for (let i = 0; i < input.length; i++) {
+        hash = (hash * 33) ^ input.charCodeAt(i);
```
benchristel, line 10: The [original function](https://theartincode.stanis.me/008-djb2/) adds the char code, rather than xoring it.

I'm not 100% sure char code is what we want either — it may be greater than 255. It would be closer to the original algorithm to convert the string to a byte array first and hash the bytes.

packages/perseus-core/src/widgets/radio/choice-id-utils.ts
benchristel: It looks like [imurmurhash] is a dependency of many of our dev dependencies, so we could consider using that. We'd need an ADR to make it a production dependency, though.

I was tipped off to the existence of imurmurhash by a mention of a murmur-based algorithm in this doc: [Radio output changes].

> Stable choiceIds have recently been introduced using murmurhash-based logic, enabling ID-based tracking of selections. 

Were hash-based choice IDs implemented at one point and then removed? 🤔

[Radio output changes]: https://khanacademy.atlassian.net/wiki/spaces/LC/pages/4163404040/Radio+output+changes
[imurmurhash]: https://www.npmjs.com/package/imurmurhash

.changeset/happy-kids-grab.md
```diff
@@ -0,0 +1,6 @@
+---
+"@khanacademy/perseus": minor
+"@khanacademy/perseus-core": minor
+---
+
+Initializes and consumes choice ids
```
handeyeco: For changesets that focus on one widget, it's helpful to specify the widget since the changeset will be displayed to consumers without context.

packages/perseus-core/src/widgets/radio/choice-id-utils.test.ts
```diff
@@ -0,0 +1,225 @@
...
+describe("convertStringToHash", () => {
+    it("returns 'empty' for empty string", () => {
+        expect(convertStringToHash("")).toBe("empty");
+    });
+
+    it("returns hash for whitespace-only string", () => {
+        const hash = convertStringToHash("   ");
+        expect(hash).toMatch(/^[a-z0-9]+$/);
+        expect(hash).not.toBe("empty");
+    });
```
handeyeco: I wonder if it would be helpful long-term to share the regex declaration

```suggestion
describe("convertStringToHash", () => {
    const hashRegex = /^[a-z0-9]+$/;

    it("returns 'empty' for empty string", () => {
        expect(convertStringToHash("")).toBe("empty");
    });

    it("returns hash for whitespace-only string", () => {
        const hash = convertStringToHash("   ");
        expect(hash).toMatch(hashRegex);
        expect(hash).not.toBe("empty");
    });
```

packages/perseus-core/src/widgets/radio/choice-id-utils.test.ts
```diff
@@ -0,0 +1,225 @@
...
+    it("handles special characters in content", () => {
+        const id = generateChoiceId("Test@#$%^&*()", 0);
+        expect(id).toMatch(/^[a-z0-9]+0$/);
+        expect(id.length).toBeLessThanOrEqual(9); // hash + index
+    });
+
+    it("handles unicode characters in content", () => {
+        const id = generateChoiceId("Test 🚀 Emoji", 0);
+        expect(id).toMatch(/^[a-z0-9]+0$/);
+        expect(id.length).toBeLessThanOrEqual(9); // hash + index
```
handeyeco: I wonder if it'd be better or worse to combine these kinds of checks:

```suggestion
        expect(id).toMatch(/^[a-z0-9]{8}0$/);
```

or

```TS
const choiceIdRegex = /^[a-z0-9]{8}\d+$/
expect(id).toMatch(choiceIdRegex);
expect(id.endsWith("0")).toBeTrue();
```

packages/perseus-core/src/widgets/radio/choice-id-utils.test.ts
```diff
@@ -0,0 +1,225 @@
...
+        const ids = [
+            generateChoiceId("Content 1", 0),
+            generateChoiceId("Content 2", 1),
+            generateChoiceId("Content 3", 2),
+            generateChoiceId("Different Content", 3),
+        ];
+
+        // All IDs should be unique
+        const uniqueIds = new Set(ids);
+        expect(uniqueIds.size).toBe(ids.length);
```
handeyeco: I love this kind of inner-test assertion to make sure we're not accidentally breaking tests.

packages/perseus-core/src/widgets/radio/choice-id-utils.ts
handeyeco: Ben's comment makes me think: why are we doing the algorithm ourselves? Is there a tiny npm package we could use instead? I don't feel strongly about it, but we might benefit from someone else's work.

packages/perseus-core/src/widgets/radio/choice-id-utils.ts
```diff
@@ -0,0 +1,43 @@
+/**
+ * Hash function to create a short hash from a string
+ * Based on djb2 hash algo for better distribution
+ */
+export function convertStringToHash(input: string): string {
+    let hash = 5381; // Use djb2 hash algorithm constant
```
handeyeco: Just a thought

```suggestion
    const DJB2_CONSTANT = 5381;
    let hash = DJB2_CONSTANT; 
```

packages/perseus-core/src/widgets/radio/choice-id-utils.ts
handeyeco: > substring will return strings shorter than 8 chars if the input is short

Seems like this would be a good test case

packages/perseus-core/src/widgets/radio/choice-id-utils.ts
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Create a normalized version of content for hashing
+ * This removes extra whitespace and normalizes the text to ensure consistent hashing
+ */
+export function normalizeContent(content: string): string {
+    if (!content) {
+        return "";
+    }
+    return content.trim().replace(/\s+/g, " ").toLowerCase();
+}
```
handeyeco: I don't really understand the point of normalizing the text, especially as it regards whitespace/case. Maybe there are known reasons in the hashing community, but to me it seems like we're reducing our chances for unique IDs (although I recognize this chance is offset by appending the index to the hash).

packages/perseus/src/widgets/radio/radio-component.tsx
```diff
@@ -313,6 +314,10 @@ class Radio extends React.Component<Props> implements Widget {
 
                 const reviewChoice = this.props.reviewModeRubric?.choices[i];
 
+                const choiceId =
+                    choice.id ||
+                    generateChoiceId(choice.content, choice.originalIndex);
```
handeyeco, line 285: Why are we doing this here if we're doing it in the parser? I think if there are cases where Radio isn't getting the ID, we should make those cases use the parser rather than duplicate logic in multiple places.

packages/perseus/src/widgets/radio/radio.new.tsx
```diff
@@ -187,6 +188,9 @@ const RadioComponent = ({
 
             // Get the reviewChoice from the rubric, if it exists.
             const reviewChoice = reviewModeRubric?.choices[i];
+            const choiceId =
+                choice.id ||
+                generateChoiceId(choice.content, choice.originalIndex);
```
handeyeco: Same here

packages/perseus/src/widgets/radio/radio.ts
```diff
@@ -73,10 +78,12 @@ const _choiceTransform = (
     // Add meta-information to choices
     const choices: ReadonlyArray<RadioChoiceWithMetadata> =
         widgetOptions.choices.map((choice, i): RadioChoiceWithMetadata => {
+            const choiceId = choice.id || generateChoiceId(choice.content, i);
```
handeyeco, line 86: Same here. I'm guessing you ran into this because tests were rendering `Radio` directly. Likely those tests need to be refactored to use the parser (and likely something like `renderQuestion`)

packages/perseus-core/src/index.ts
```diff
@@ -57,7 +57,11 @@ export {Errors} from "./error/errors";
 export {PerseusError} from "./error/perseus-error";
 
 export * from "./data-schema";
-
+export {
+    convertStringToHash,
+    generateChoiceId,
+    normalizeContent,
+} from "./widgets/radio/choice-id-utils";
```
handeyeco: I don't think these need to be exported. You seem to only use `generateChoiceId` and I don't think you should be (it should only be used in the parser)...hm...except I guess in the editing experience. Have we checked the editing experience to make sure they're making Choices with IDs?

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.mockData.ts
```diff
@@ -192,29 +192,29 @@ export const v4Widget = {
...
                 rationale: "I am some rationale for a choice",
-                id: "choice-2",
+                id: "nglt5v1",
             },
             {
                 content: "I am an incorrect choice",
                 correct: false,
                 isNoneOfTheAbove: false,
-                id: "choice-3",
+                id: "nglt5v2",
```
handeyeco: I'm a little confused here, not because of you but just around the motivation for the hash. We have two hashes here: `nglt5v1` and `nglt5v2`. The only thing that differentiates them is the index, but if the index is enough to differentiate two hashes...do we need the hash at all? Could we just use the original index as the ID?

I think:
1. It's _very_ unlikely two choices will have the same content. However I think the chances of having the same hash increase since we normalize the content and we cap the length of the hash. For example if the question is about grammar: "A. She went to paris. B. She went to Paris. C. she went to Paris. D. she went to paris" will all be the same hash because of normalizing case.
2. It _feels_ better to have a hash over just an index. I don't know why though and that doesn't seem like enough to add code complexity.

So I think maybe some thoughts on why we need the hash at all could be helpful.

---

Just a random aside, I wonder if we should be appending the index to the final hash rather than prepending the index to the content before hashing. I think that would increase our chances at unique, short hashes that come from normalized content. It would also give away less information; this isn't a concern now but I could just see someone in the future asking us to shuffle on the server before sending to the client.

handeyeco:
> Question: how important is it for the choice IDs to be unique within the assessment item?

@benchristel It shouldn't matter at all I wouldn't think. Both widget options and user input are stored by their widget's ID, there should be no cross-talk between two Radio widgets.

packages/perseus-core/src/widgets/radio/choice-id-utils.test.ts
```diff
@@ -0,0 +1,225 @@
+import {
+    convertStringToHash,
+    normalizeContent,
+    generateChoiceId,
+} from "./choice-id-utils";
+
+describe("convertStringToHash", () => {
```
handeyeco, line 5: Really appreciate the ample test coverage in this PR! Love to see it!

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.mockData.ts
benchristel: > It feels better to have a hash over just an index. I don't know why though

A possible reason: hashing obscures the index and prevents consumers of Perseus data from writing code that assumes the ID will _always_ be an index. We probably don't want to make any guarantees about format.

I don't think that's necessarily a strong enough reason to prefer hashing, though.

handeyeco: Yeah but that's part of my question. In this PR we're appending the original index to the hash as part of the final ID, so the index is not obfuscated.

packages/perseus-core/src/index.ts
anakaren-rojas: I've only been able to check in storybook that the IDs are being generated, when I try to make an FE ZND everything in other files breaks 😢 
Will try again after this though

packages/perseus-core/src/widgets/radio/choice-id-utils.ts
anakaren-rojas: tbh, I'm a bit hesitant of going through the process of adding a package as a peer dep in frontend

packages/perseus-core/src/widgets/radio/choice-id-utils.ts
anakaren-rojas: Hehe that's my doc about this!

packages/perseus-core/src/widgets/radio/choice-id-utils.ts
anakaren-rojas: it was a step toward making the ids deterministic, so `option A`, `Option a` and `Option A` all return the same id since its basically the same content, but if it doesn't seem necessary, I'm happy to remove it

packages/perseus/src/widgets/radio/radio-component.tsx
anakaren-rojas: the main reason is because choice.id isn't required, so it can be undefined 
I can't make it required in v4 unless I add it to earlier migration versions in the parser 
the second, maybe less probable reason is as a fall back in case the data were to make it here without having an id for whatever reason

packages/perseus-core/src/widgets/radio/choice-id-utils.ts
anakaren-rojas: i thought it was more clear to have it return "empty" than to have it return the hash constant, but if it seems redundant, I can take this out

packages/perseus-core/src/widgets/radio/choice-id-utils.ts
anakaren-rojas: the [source](https://mojoauth.com/hashing/bernsteins-hash-djb2-in-typescript/) i used for this algo has the last step as 

> `return hash >>> 0; // Ensure a positive integer`

I can update it to this, but I am also not as familiar with this style of coding and wouldn't be comfortable answering questions about it 😆 
In terms of switching to a imurmurhash dependency, I would be onboard, but as a separate task to this. 
An ADR in addition to completing this work might be out of scope for the original task

packages/perseus-core/src/index.ts
anakaren-rojas: Was able to confirm that the ID is being added in the editor 🥳 
image

packages/perseus-core/src/widgets/radio/choice-id-utils.ts
handeyeco: You wouldn't need to, just a new dependency in Perseus. Especially if it were really small/focused you could probably get by with a one-day ADR.

packages/perseus-core/src/widgets/radio/choice-id-utils.ts
handeyeco: It'd still be deterministic without the normalization and would give us more chances for unique hashes, so I think it'd be good to remove.

packages/perseus/src/widgets/radio/radio-component.tsx
handeyeco: I think you could make it required. Then make the editor create the ID and use the parser to generate IDs when migrating from older versions of content.

anakaren-rojas:
> > Question: how important is it for the choice IDs to be unique within the assessment item?
> 
> @benchristel It shouldn't matter at all I wouldn't think. Both widget options and user input are stored by their widget's ID, there should be no cross-talk between two Radio widgets.

@benchristel  @handeyeco 
A weird 🌟 feature 🌟 that I noticed in test everything is that when randomizing, if I turn one radio's randomization off and on, all of the choices on all of the radio widgets who have randomize on are re-randomized. 
This isn't behavior that I would expect, which leads me to think that unique ids might matter more than expected

https://github.com/user-attachments/assets/50f282d3-71f6-4c91-a174-59fd12dd69a7

In the recording, I would only expect the top single select radio to randomize when I click the button off and on

packages/perseus-core/src/widgets/radio/choice-id-utils.ts
```diff
@@ -0,0 +1,30 @@
+/**
+ * DJB2 hash algo to create a hash from a string
+ * Reference: https://mojoauth.com/hashing/bernsteins-hash-djb2-in-typescript/
+ */
+export function convertStringToHash(input: string): number {
```
anakaren-rojas, line 5: this function would be replaced with a murmurhash implementation when ADR is approved
https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/4199907561/ADR+851+New+Dependency+Proposal+-+Add+imurmurhash+to+Perseus

PR #2640, opened 2025-06-24
Remove deprecated fields from Radio, rename `clue` to `rationale`

benchristel:
## Summary:
This PR does two related things:
- It creates a v3 of the Radio widget, with deprecated fields removed and
  `clue` renamed to `rationale`. Effectively, this re-does #2435, which was
  reverted by #2621.
- It updates the Radio widget parser to accept null values for `onePerLine` and
  `noneOfTheAbove` and convert them to `undefined`. This is needed to avoid
  breaking ELA content which munges the widgets in Go code. See
  [this slack thread] for context.

[this slack thread]: https://khanacademy.slack.com/archives/C01AZ9H8TTQ/p1750165141757919

Issue: LEMS-3061

## Test plan:

`pnpm test`

### DIFF
--- a/packages/perseus/src/widgets/explanation/__snapshots__/explanation.test.ts.snap
+++ b/packages/perseus/src/widgets/explanation/__snapshots__/explanation.test.ts.snap
@@ -21,7 +21,7 @@ exports[`Explanation should snapshot when expanded: expanded 1`] = `
             aria-controls=":r1:"
             aria-disabled="false"
             aria-expanded="true"
-            class="button_vr44p2-o_O-reset_1ck4g21-o_O-shared_14wd1lh-o_O-default_ejemin-o_O-small_1hll8b3-o_O-inlineStyles_1s8anjv"
+            class="button_vr44p2-o_O-reset_1ck4g21-o_O-shared_1jpsfmg-o_O-default_n0abw6-o_O-small_tjsyec-o_O-inlineStyles_1s8anjv"
             role="button"
             type="button"
           >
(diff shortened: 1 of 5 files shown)


packages/perseus/src/widgets/radio/radio-component.tsx
```diff
@@ -28,7 +27,6 @@ export type RenderProps = {
     deselectEnabled?: boolean;
     choices: RadioChoiceWithMetadata[];
     selectedChoices: PerseusRadioChoice["correct"][];
-    showSolutions?: ShowSolutions;
```
anakaren-rojas: we weren't able to figure out with Sarah why these were removed before, but they should be kept in
Lines 17 & 31

packages/perseus/src/widgets/radio/radio.new.tsx
```diff
@@ -29,7 +29,6 @@ export interface RenderProps {
     deselectEnabled?: boolean;
     choices: ReadonlyArray<RadioChoiceWithMetadata>;
     selectedChoices: ReadonlyArray<PerseusRadioChoice["correct"]>;
-    showSolutions?: ShowSolutions;
```
anakaren-rojas: same here, this should be kept in

packages/perseus-core/src/utils/item-has-rationales.test.ts
```diff
@@ -132,7 +133,8 @@ describe("itemHasRationales", () => {
                                             {
                                                 content: "Choice 1",
                                                 correct: false,
-                                                clue: "This is a clue",
+                                                rationale:
+                                                    "This is a rationale",
```
anakaren-rojas:
```suggestion
                                                    "This is some rationale",
```

packages/perseus/src/widgets/radio/radio-component.tsx
benchristel: Thanks! It's kind of scary that typechecking passed even though these were removed. I really want to fix how our widget prop types work...

anakaren-rojas (approved):
🥳

PR #2647, opened 2025-06-25
[WIP!!!! - Do not review] Nicer image editing in radio

nishasy:
## Summary:

TODO

Issue: https://khanacademy.atlassian.net/browse/LEMS-3028

## Test plan:
TODO


packages/perseus-editor/src/widgets/__docs__/radio-editor.stories.tsx
```diff
@@ -31,9 +31,13 @@ export const Default: Story = {
 };
 
 export const SingleChoice = (): React.ReactElement => (
-    <EditorPageWithStorybookPreview question={singleSelectQuestion} />
+    <div style={{width: 338}}>
```
nishasy: Added a manual 338px width so it looks more accurate to prod rather than stretching all the way.

packages/perseus-editor/src/widgets/radio/radio-image-editor.tsx
nishasy, line 1: I initially had the "add image tile" and "editor accordion" as different files with these "Image URL" and "Image alt text" fields duplicated, but I didn't like the duplication.

So instead I broke out the "inner" text fields into this file, and then consolidated those other two editors one file. This might seem like overkill with the flexible props, but at least the text fields are consistent between the two editing areas.

I also tried removing the "Add image" tile in favor of always using the accordions in order to reduce code, but it was a noticeably worse experience.

.changeset/good-mails-argue.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": minor
+---
+
+[Radio] | (CX) | Implment UI for adding images in Radio editor
```
ivyolamit: looks like changeset was generated twice, please delete second one.

packages/perseus-editor/src/widgets/__docs__/radio-editor.stories.tsx
ivyolamit: magic number, since the width value is used twice, please add a styling or const value for this so it can be used in both places.

packages/perseus-editor/src/widgets/radio/radio-image-editor.tsx
```diff
@@ -0,0 +1,133 @@
...
+import xIcon from "@phosphor-icons/core/regular/x.svg";
+import * as React from "react";
+
+import {AutoResizingTextArea} from "./auto-resizing-text-area";
+import styles from "./radio-editor.module.css";
+
+// Flexible props to work for both the "add image" tile and
+// the "edit image" accordion cases.
+interface RadioImageEditorProps {
+    initialImageUrl?: string;
```
ivyolamit: do we want this as required props? Since this is a radio image editor, I would imagine we want to require the url all the time. This will also lessen additional checks in multiple places.
```suggestion
    initialImageUrl: string;
```

packages/perseus-editor/src/widgets/radio/radio-image-editor.tsx
```diff
@@ -0,0 +1,133 @@
...
+        <div className={containerClassName}>
+            {/* Close button */}
+            {onClose && (
+                <IconButton
+                    aria-label="Close"
+                    icon={xIcon}
+                    size="small"
+                    kind="tertiary"
+                    onClick={handleClose}
+                    style={{position: "absolute", top: 4, right: 4}}
```
ivyolamit, line 66: not sure what our guidance here adding styles in css vs inline styling.
suggestion adding all the explicit styles in `radio-image-editor` file in css file. And please use tokens consistently as well like `--wb-sizing-size_040`

packages/perseus-editor/src/widgets/radio/radio-image-editor.tsx
```diff
@@ -0,0 +1,133 @@
...
+import {AutoResizingTextArea} from "./auto-resizing-text-area";
+import styles from "./radio-editor.module.css";
+
+// Flexible props to work for both the "add image" tile and
+// the "edit image" accordion cases.
+interface RadioImageEditorProps {
+    initialImageUrl?: string;
+    initialImageAltText?: string;
+    containerClassName?: string;
+    primaryAction?: React.ReactNode;
```
ivyolamit: would this always an optional props?
_same with other props in this group._

packages/perseus-editor/src/widgets/radio/radio-image-editor.tsx
```diff
@@ -0,0 +1,133 @@
...
+                tag="label"
+                htmlFor={imageAltTextTextAreaId}
+                style={{marginBlockEnd: sizing.size_040}}
+            >
+                Image Alt Text
+            </HeadingXSmall>
+            <AutoResizingTextArea
+                id={imageAltTextTextAreaId}
+                value={imageAltText}
+                placeholder="The Moon appears as a bright gray circle in black space..."
```
ivyolamit: Suggestion: add a useful information placeholder here, so it will be easy first time content editor to use.

packages/perseus-editor/src/widgets/radio/radio-image-editor.tsx
```diff
@@ -0,0 +1,133 @@
...
+                onChange={(value) => {
+                    setImageAltText(value);
+                }}
+            />
+
+            {/* Save and delete buttons */}
```
ivyolamit, line 105: Please add aria-labels to these buttons below.

packages/perseus-editor/src/widgets/radio/radio-option-content-and-image-editor.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+            {/* Image editor accordions */}
+            {images?.map((image, imageIndex) => (
+                <PerseusEditorAccordion
+                    key={image.url}
+                    header={`Image ${imageIndex + 1}`}
+                    expanded={true}
+                    containerStyle={{
+                        // White to contrast with the blue choice tile.
+                        backgroundColor: semanticColor.surface.primary,
```
ivyolamit, line 192: Similar to my suggestion above on adding the styling in css file instead of inline styling.

packages/perseus-editor/src/widgets/radio/radio-option-settings.tsx
```diff
@@ -96,30 +96,31 @@ export function RadioOptionSettings({
...
+                choiceIndex={index}
+                isNoneOfTheAbove={isNoneOfTheAbove ?? false}
+                onContentChange={onContentChange}
+            />
+
+            <HeadingXSmall
+                tag="label"
+                htmlFor={rationaleTextAreaId}
+                style={{
+                    marginBlockStart: sizing.size_040,
```
ivyolamit, line 110: Similar to my suggestion above on adding the styling in css file instead of inline styling.

.changeset/good-mails-argue.md
nishasy: I'll delete it when we're done with the ZND. The ZND wasn't generating with the first one for some reason.

packages/perseus-editor/src/widgets/radio/radio-image-editor.tsx
nishasy: Sure, but then I'd have to pass in `""` in the adding image tile. I can do that, but I figured it's a matter of preference.

packages/perseus-editor/src/widgets/radio/radio-image-editor.tsx
nishasy: As far as I know, you can't use CSS inside wonder blocks. @mark-fitzgerald Is that still correct?

packages/perseus-editor/src/widgets/radio/radio-image-editor.tsx
nishasy: oops, that wasn't even supposed to be there. It doesn't get used.

packages/perseus-editor/src/widgets/radio/radio-image-editor.tsx
nishasy: 🤔 Placeholders are supposed to have examples for what kind of stuff is supposed to go in that box. I used an example alt text that was already in an article in Khan Academy.

What would you suggest I put here instead?

packages/perseus-editor/src/widgets/radio/radio-image-editor.tsx
nishasy: I don't think the buttons need aria labels since they have actual text names already ("Save image" and "Delete this image")

packages/perseus-editor/src/widgets/radio/radio-option-content-and-image-editor.tsx
nishasy: This is also Wonder Blocks.

packages/perseus-editor/src/widgets/radio/radio-option-settings.tsx
nishasy: This is also Wonder Blocks.

packages/perseus-editor/src/widgets/radio/radio-image-editor.tsx
nishasy: Updating

ivyolamit:
Smoke test comment: The row where the buttons and toggle are part of the Edit tab. Currently, it doesn't have visual separation with the content below.

Suggestion: Add a row container that will have a bottom border and a slightly different background color that will contain all the buttons and toggle so visually content editors will know the top section is for the controls. And ideally having them in a container the buttons and toggle will adjust accordingly when screen is adjusted.

image

Note: it can be in a different PR since it's a different behavior handling.

ivyolamit:
Smoke test comment: Have you thought about following the usual add image or drag and drop visual?

image

Suggestion:
image

🔗 https://css-tricks.com/drag-and-drop-file-uploading/
Note: it can be in a different PR to iterate on the work.

ivyolamit:
Smoke testing comments:
1. When an image was added the container doesn't have a consistent top & bottom spacing.
~~2. I also noticed all added image will **all** have the header "Image 1" was this intentional?~~ Updated: it's expected behavior

| Initial | After image is added |
| ------ | -------------------- |
| Screenshot 2025-08-05 at 1 59 05 PM | Screenshot 2025-08-05 at 1 50 48 PM |

packages/perseus-editor/src/widgets/radio/radio-image-editor.tsx
ivyolamit: How about?
```suggestion
                placeholder="Add alternate image text"
```

nishasy: I think that counts as instructions, but [placeholders](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/placeholder) are not supposed to be used for instructions, but rather for an example or hint. If instructions are needed, they should be in the label description.

I can update this to be something like `placeholder="Example: Graph of a linear function..."` so it might be less confusing?

nishasy: You can see how I did the placeholder for the image URL by text area by setting it to `cdn.kastatic.org...` as an example to show that it's supposed to be an image URL from the Khan Academy CDN.

packages/perseus-editor/src/widgets/radio/radio-image-editor.tsx
mark-fitzgerald: Most [WB components accept a `className` attribute](https://khan.github.io/wonder-blocks/?path=/docs/packages-iconbutton-iconbutton--docs). Unfortunately, since WB is unlayered and Perseus is in the "shared" layer, it means that any styling done in Perseus to affect something in WB (temporarily) requires the use of `!important` on each styling property. Once we get WB in the "shared" layer, and Perseus can have mixed layers, then we can remove these `!important` modifiers.

ivyolamit (approved):
Thanks for the update. LGTM 👍🏼

packages/perseus-editor/src/widgets/radio/radio-image-editor.tsx
nishasy: Would it be better to move these into a CSS module with !important everywhere? Or should I leave these inline for now?

packages/perseus-editor/src/components/perseus-editor-accordion.tsx
```diff
@@ -39,7 +41,7 @@ const PerseusEditorAccordion = (props: Props) => {
                 expanded={expanded}
                 onToggle={onToggle}
                 style={[styles.container, containerStyle]}
-                headerStyle={styles.accordionHeader}
+                headerStyle={[styles.accordionHeader, headerStyle]}
```
mark-fitzgerald, line 44: Does this render `undefined` if no `headerStyle` is passed in?

packages/perseus-editor/src/widgets/radio/radio-editor.module.css
```diff
@@ -17,3 +17,22 @@
...
+    align-items: center;
+    margin-block-start: 8px;
+}
+
+.image-editor-container {
+    display: flex;
+    flex-direction: column;
+    position: relative;
+    padding: 8px;
+    padding-top: 24px;
```
mark-fitzgerald: Nit: As a matter of good practice (meaning, even though it's "just the editor"), using `padding-block-start` helps us to keep in practice with proper language orientation techniques.

packages/perseus-editor/src/components/perseus-editor-accordion.tsx
nishasy: I think so? It will just have the base `styles.accordionHeader` of a `headerStyle` isn't passed in. This is a [common practice across wonder blocks](https://github.com/Khan/wonder-blocks/blob/main/packages/wonder-blocks-button/src/components/button-core.tsx#L158).

packages/perseus-editor/src/widgets/radio/radio-editor.module.css
nishasy: Oops. I used `-block-start` for the others and missed this one.

packages/perseus-editor/src/widgets/radio/radio-image-editor.tsx
mark-fitzgerald: Both options are valid. I lean towards adding them to the CSS file and including `!important` along with a note to remove them once Cascade Layering has been corrected. However, I leave it to the author to decide since both require follow-up at a later date.

packages/perseus-editor/src/widgets/radio/radio-option-content-and-image-editor.tsx
```diff
@@ -0,0 +1,222 @@
...
+    // States for adding an image
+    const [addingImage, setAddingImage] = React.useState(false);
+
+    React.useEffect(() => {
+        const [niceContent, images] = setNiceContentAndImages(content ?? "");
+        setNiceContent(niceContent);
+        setImages(images);
+    }, [content]);
+
+    // Add the image markdown where the caret currently is in the textarea.
```
mark-fitzgerald: Doesn't this function append the image, not insert it?

nishasy: Oh!! Yes. I forgot to change the comment after I changed the implementation.

Inserting it at the caret was causing a lot of issues.

packages/perseus-editor/src/widgets/radio/radio-image-editor.tsx
nishasy: Okay, I think I might leave it like this for now, since I'll have to change it in a lot of spots. Maybe I'll switch to CSS in a separate PR.

packages/perseus-editor/src/components/perseus-editor-accordion.tsx
mark-fitzgerald: As long as it doesn't render `undefined` in the `class` attribute of the DOM element.

anakaren-rojas:
I was having a bit of fun with this so I might've gone overboard with the testing, but here are the things I tested: 

- [x] Able to add image
- [x] Able to delete image
- [x] Able to delete choice with image 
- [ ] Able to move choice with image > Image stays in the same position instead of moving with the rest of the choice
https://github.com/user-attachments/assets/142ed0f3-a5bc-4c2f-b481-5420c0d623ca
- [x] Unable to drag images saved locally
- [ ] Save is disabled when image and alt text are empty > Can save
- [ ] Save is disabled with image is empty but alt text is populated > Can save
Screenshot 2025-08-06 at 14 29 26

### A few questions: 

- Should there be internal image configuration in case a user pastes a link to an imagine that's larger than certain dimensions? 
Screenshot 2025-08-06 at 14 16 43

- Should alt text be required? My guess is `yes` but you might have more background on situations where it would be ok to leave it blank 

- (This might be a follow up PR): If there is text in the answer choice already, can the image be added to its own line? Right now its on the same line and looks a little funky

packages/perseus-editor/src/widgets/radio/radio-option-content-and-image-editor.tsx
```diff
@@ -0,0 +1,222 @@
...
+    const handleDeleteImageConfirmation = (imageIndex: number) => {
+        if (
+            // eslint-disable-next-line no-alert
+            window.confirm("Are you sure you want to delete this image?")
+        ) {
+            handleDeleteImage(imageIndex);
+        }
+    };
+
+    if (isNoneOfTheAbove) {
```
mark-fitzgerald, line 118: ~Why are we rendering an uneditable image dialog for this option? Can we instead just not include the "Add image" button for that option?~

mark-fitzgerald: Sorry. I misunderstood what was happening here.

packages/perseus-editor/src/widgets/radio/radio-option-content-and-image-editor.tsx
```diff
@@ -0,0 +1,222 @@
...
+                Content
+            </HeadingXSmall>
+            <AutoResizingTextArea
+                id={contentTextAreaId}
+                value={niceContent}
+                placeholder="Type a choice here..."
+                onChange={(value) => {
+                    handleContentChange(choiceIndex, value);
+                }}
+                onPaste={handlePaste}
```
mark-fitzgerald, line 151: I like that you don't force people to use the image editor if they already have the content copied.

packages/perseus-editor/src/widgets/radio/utils.ts
```diff
@@ -71,3 +73,62 @@ export function getMovedChoices(
 
     return newChoices;
 }
+
+// Take the hard-to-read image markdown, and replace it with a nice placeholder.
+// ex. "![abc](https://...) -> "![Image 1]"
+export function setNiceContentAndImages(
```
mark-fitzgerald: Nit: "Nice" makes me think of something different than what is intended here. Just a suggestion (not required to be changed): "Proxy" as in `setContentWithImageProxy`.

mark-fitzgerald (approved):
This is soooo helpful!! Looks fabulous!

nishasy:
@anakaren-rojas 

> Should there be internal image configuration in case a user pastes a link to an imagine that's larger than certain dimensions?

I don't think this is possible with our current markdown parser, but it could be worth looking into in the future.

> Should alt text be required? My guess is yes but you might have more background on situations where it would be ok to leave it blank

Yes, it should be required. This isn't surfaced inside the issue panel yet, but it is surfaced by the Red Dot of Doom, thanks to CP.

image

 
> (This might be a follow up PR): If there is text in the answer choice already, can the image be added to its own line? Right now its on the same line and looks a little funky

I believe this is a known issue with the editor. I think @mark-fitzgerald started a [conversation about it in Slack](https://khanacademy.slack.com/archives/C08FXG7TUR5/p1753311973628049).

packages/perseus-editor/src/components/perseus-editor-accordion.tsx
nishasy: I don't see it rendering as an undefined class anywhere. We'll probably have to change this syntax once we switch to CSS modules though. But for now, I think `style` can handle it.

image

PR #2649, opened 2025-06-25
[Interactive Graph] [Locked Figures] Add `weight` option to locked polygon and locked polygon settings

nishasy:
## Summary:

To make locked figures easier to see over axis lines, and to give content authors a bit more freedom
in how the locked figures look, we're adding a `weight` option to locked figures.

Adding this `weight` select to the locked polygon editor and hooking it up to the rendered locked
polygon here.

## Test plan:
`pnpm jest packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-polygon-settings.test.tsx`

Storybook
- Confirm that it updates the polygon in the preview

### DIFF
--- a/packages/perseus-core/src/data-schema.ts
+++ b/packages/perseus-core/src/data-schema.ts
@@ -873,6 +873,7 @@ export type LockedPolygonType = {
     showVertices: boolean;
     fillStyle: LockedFigureFillType;
     strokeStyle: LockedLineStyle;
+    weight?: "thin" | "medium" | "thick";
     labels: LockedLabelType[];
     ariaLabel?: string;
 };
--- a/packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
+++ b/packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
@@ -137,6 +137,8 @@ const parseLockedFigureFillType = enumeration(
 
 const parseLockedLineStyle = enumeration("solid", "dashed");
 
+const parseStrokeWeight = optional(enumeration("medium", "thin", "thick"));
+
 const parseLockedLabelType = object({
     type: constant("label"),
     coord: pairOfNumbers,
@@ -193,6 +195,7 @@ const parseLockedPolygonType = object({
     showVertices: boolean,
     fillStyle: parseLockedFigureFillType,
     strokeStyle: parseLockedLineStyle,
+    weight: parseStrokeWeight,
     labels: defaulted(array(parseLockedLabelType), () => []),
     ariaLabel: optional(string),
 });
--- a/packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/line-weight-select.tsx
+++ b/packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/line-weight-select.tsx
@@ -2,27 +2,34 @@ import {OptionItem, SingleSelect} from "@khanacademy/wonder-blocks-dropdown";
 import {Strut} from "@khanacademy/wonder-blocks-layout";
 import {spacing} from "@khanacademy/wonder-blocks-tokens";
 import {LabelMedium} from "@khanacademy/wonder-blocks-typography";
-import {StyleSheet} from "aphrodite";
 import * as React from "react";
 
 import type {StyleType} from "@khanacademy/wonder-blocks-core";
 
-type StyleOptions = "solid" | "dashed";
 type Props = {
-    selectedValue: StyleOptions;
-    onChange: (newValue: StyleOptions) => void;
+    selectedValue: "thin" | "medium" | "thick";
+    onChange: (newValue: "thin" | "medium" | "thick") => void;
     containerStyle?: StyleType;
 };
 
-const LineStrokeSelect = (props: Props) => {
+const LineWeightSelect = (props: Props) => {
     const {selectedValue, containerStyle, onChange} = props;
 
     return (
         <LabelMedium
             tag="label"
-            style={[styles.lineStrokeSelect, containerStyle]}
+            style={[
+                {
+                    display: "flex",
+                    flexDirection: "row",
+                    alignItems: "center",
+                    // Allow truncation, stop bleeding over the edge.
+                    minWidth: 0,
+                },
+                containerStyle,
+            ]}
         >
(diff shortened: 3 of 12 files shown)


packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
```diff
@@ -137,6 +137,8 @@ const parseLockedFigureFillType = enumeration(
 
 const parseLockedLineStyle = enumeration("solid", "dashed");
 
+const parseStrokeWeight = optional(enumeration("medium", "thin", "thick"));
```
benchristel: Should we default to `medium` here? That ensures that existing graphs continue to have medium stroke weight if we ever change the default in the component.

```suggestion
const parseStrokeWeight = defaulted(enumeration("medium", "thin", "thick"), () => "medium" as const);
```

packages/perseus-core/src/data-schema.ts
```diff
@@ -805,6 +805,14 @@ export const lockedFigureColors: Record<LockedFigureColor, string> = {
     orange: "#946700",
 } as const;
 
+export type LockedFigureWeight = "thin" | "medium" | "thick";
+
+export const lockedFigureWeights: Record<LockedFigureWeight, number> = {
+    thin: 1,
+    medium: 2,
+    thick: 4,
+} as const;
```
benchristel: It feels like these weights are a display concern, not a data concern, and might belong better in `locked-polygon.tsx`. Nothing else in the codebase should have to know what the specific pixel measurements are.

packages/perseus/src/widgets/interactive-graphs/interactive-graph.test.tsx
```diff
@@ -931,6 +932,47 @@ describe("Interactive Graph", function () {
             });
         });
 
+        it.each([
+            {weight: undefined, expectedStrokeWidth: 2},
+            {weight: "thin", expectedStrokeWidth: 1},
+            {weight: "medium", expectedStrokeWidth: 2},
+            {weight: "thick", expectedStrokeWidth: 4},
```
benchristel: This made me think of a couple things:

- Since this is a visual change, what if we had Storybook stories / visual snapshot tests for it?
- Is this change accessible? I wouldn't rule out the possibility of content creators using multiple weights in an exercise and assigning them different meanings.

packages/perseus-core/src/data-schema.ts
nishasy: This is going to be in a bunch of locked figures coming up. I figured it would be better to follow the pattern we have for colors, but I can move it to a util file instead.

packages/perseus/src/widgets/interactive-graphs/interactive-graph.test.tsx
nishasy: I can add some cases in the interactive graph regression stories and add this onto the autogen aria labels.

packages/perseus-core/src/data-schema.ts
```diff
@@ -805,6 +805,8 @@ export const lockedFigureColors: Record<LockedFigureColor, string> = {
     orange: "#946700",
 } as const;
 
+export type LockedFigureWeight = "thin" | "medium" | "thick";
```
benchristel: Should the type name be `StrokeWeight` to agree with the parser? I think I like StrokeWeight better than LockedFigureWeight because it's a little more precise.

```suggestion
export type StrokeWeight = "thin" | "medium" | "thick";
```

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/line-weight-select.tsx
```diff
@@ -0,0 +1,48 @@
...
+import {Strut} from "@khanacademy/wonder-blocks-layout";
+import {spacing} from "@khanacademy/wonder-blocks-tokens";
+import {LabelMedium} from "@khanacademy/wonder-blocks-typography";
+import * as React from "react";
+
+import type {StyleType} from "@khanacademy/wonder-blocks-core";
+
+type Props = {
+    selectedValue: "thin" | "medium" | "thick";
+    onChange: (newValue: "thin" | "medium" | "thick") => void;
```
benchristel: Should these props use the type from data-schema?

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/line-weight-select.tsx
```diff
@@ -0,0 +1,48 @@
...
+                    minWidth: 0,
+                },
+                containerStyle,
+            ]}
+        >
+            weight
+            <Strut size={spacing.xxxSmall_4} />
+            <SingleSelect
+                selectedValue={selectedValue}
+                onChange={onChange as any}
```
benchristel: Idea: rather than cast to `any` here, we could parse the input:

```ts
onChange={(selected) => {
  const parsed = parseStrokeWeight(selected);
  if (isSuccess(parsed)) {
    onChange(parsed.value);
  }
  // Do nothing if we get an unrecognized value - should never happen.
}}
```

I'm not sure how big a difference this makes to the developer experience. I think it _will_ prevent the type of the callback prop from getting out of sync with data-schema, but using the data-schema type in the props would also accomplish that.

packages/perseus/src/widgets/interactive-graphs/interactive-graph-question-builder.ts
```diff
@@ -431,6 +431,7 @@ class InteractiveGraphQuestionBuilder {
             showVertices?: boolean;
             fillStyle?: LockedFigureFillType;
             strokeStyle?: "solid" | "dashed";
+            weight?: "thin" | "medium" | "thick";
```
benchristel: Suggestion: we could use the type from data-schema here.

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/line-weight-select.tsx
nishasy: Oh huh. I didn't really pay attention to the `onChange as any` since it was copied over from some of the other editor select elements. If I update it here, I should be able to update it in the color select and the stroke select as well.

PR #2651, opened 2025-06-26
Bugfix: Plotter dot graph crashing

handeyeco:
## Summary:
[LEMS-3251/plotter-dot] fix plotter dot problem

Issue: LEMS-3251

https://github.com/user-attachments/assets/2826bcbb-b29e-4ec0-a1cc-5c3434d81faa


packages/perseus/src/widgets/plotter/plotter.tsx
```diff
@@ -142,7 +142,7 @@ export class Plotter extends React.Component<Props, State> implements Widget {
         return 2;
     };
 
-    setupGraphie: (arg1: any) => void = (prevState) => {
```
handeyeco, line 145: Because of the `any` TS didn't catch that `State` no longer included `values`

PR #2652, opened 2025-06-26
Bugfix: static switch not working in editor

handeyeco:
## Summary:
I got too aggressive adding things to the deny list.

[LEMS-3252/static-update] fix static filtering

Issue: LEMS-3252

Before:

https://github.com/user-attachments/assets/3eab734f-9ea2-4b93-b7be-028138be187a

After:

https://github.com/user-attachments/assets/82c2c438-0cd8-42d3-bf39-e0e692418bd2


packages/perseus/src/mixins/widget-prop-denylist.ts
```diff
@@ -42,10 +42,8 @@ const denylist = [
     "handleUserInput",
     "analytics",
     "showSolutions",
-    "alignment",
     "reviewMode",
     "reviewModeRubric",
-    "static",
```
handeyeco, line 48: This was the fix, I had starting filtering this in the feature branch

packages/perseus/src/mixins/widget-prop-denylist.ts
```diff
@@ -42,10 +42,8 @@ const denylist = [
     "handleUserInput",
     "analytics",
     "showSolutions",
-    "alignment",
```
handeyeco, line 45: I decided to keep this too just in case

packages/perseus/src/__tests__/server-item-renderer.test.tsx
```diff
@@ -467,20 +467,22 @@ describe("server item renderer", () => {
...
+  "question": {
+    "mock-widget 1": {
+      "alignment": "default",
+      "currentValue": "-42",
+      "static": undefined,
+      "value": "3",
+    },
+  },
+}
+`);
```
benchristel, line 485: Does the whitespace have to change here? I think this was easier to read when it was indented.

handeyeco: My IDE must have done this. I don't know what's up.

handeyeco: Also to answer your other question...

> I'm a bit confused about how the filtering caused the bug, though. Does this mean that getSerializedState is still load-bearing in the editors?

`WIDGET_PROP_DENYLIST` is used by Changeable which handles `onChange` calls which is how the editor sends updates to the parent I s'pose.

handeyeco: > My IDE must have done this. I don't know what's up.

Actually when I change it the test fails and when I run `pnpm test -u` it changes the file to remove the indent. So it's something with Jest.

PR #2657, opened 2025-06-26
Accept any wonder-blocks-link version compatible with 9.1.9 as a peer dep of perseus-editor

benchristel:
## Summary:
Kevin B. ran into this error trying to run the Perseus version bump script in webapp.

```
MISMATCH: @khanacademy/wonder-blocks-link (webapp: 9.1.10, expected: 9.1.9)
```

The issue is that Perseus was requiring an exact version (9.1.9) as a peer dep,
and we want to accept a range, (^9.1.9) so webapp is free to do minor and patch
upgrades. It looks like the cause was that we mistakenly used the `devDeps`
catalog (which hardcodes exact versions) instead of the `peerDeps` catalog.

Issue: none

## Test plan:

```
pnpm test
```

### DIFF
--- a/.changeset/wise-moose-camp.md
+++ b/.changeset/wise-moose-camp.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": patch
+---
+
+Accept any wonder-blocks-link version compatible with 9.1.9 as a peer dep of perseus-editor


mark-fitzgerald (approved):
I can't think of a reason why we would need `devDeps`. Thanks for fixing this.

nishasy:
Thank you for fixing this! Sorry, this went out because of one of my PRs. I'm pretty sure this was output from Cursor - remember to double check everything it does!

benchristel:
Huh, that's pretty surprising! I'd expect Cursor to maintain the pattern in the surrounding code and use `peerDeps`. 👍 for double-checking.

PR #2662, opened 2025-06-27
remove simple transforms

handeyeco:
## Summary:
<Add your summary here>

Issue: LEMS-XXXX

## Test plan:

### DIFF
--- a/packages/perseus/src/types.ts
+++ b/packages/perseus/src/types.ts
@@ -525,6 +525,14 @@ export type WidgetExports<
     tracking?: Tracking;
 
     /**
+     * Originally we were looking to see if a widget has `staticTransform`
+     * to determine whether or not it could be static; however we'd like
+     * to move away from transforms (especially transforms that aren't
+     * doing anything interesting), so we have a flag for that now.
+     */
+    canBeStatic?: boolean;
+
+    /**
      * Transforms the widget options to the props used to render the widget.
      *
      * @deprecated see LEMS-3199
--- a/packages/perseus/src/widgets.ts
+++ b/packages/perseus/src/widgets.ts
@@ -217,8 +217,7 @@ export const getRendererPropsForWidgetInfo = (
  */
 export const supportsStaticMode = (type: string): boolean | undefined => {
     const widgetInfo = widgets.get(type);
-    // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
-    return widgetInfo && widgetInfo.staticTransform != null;
+    return widgetInfo?.canBeStatic === true;
 };
 
 /**
--- a/packages/perseus/src/widgets/categorizer/categorizer.tsx
+++ b/packages/perseus/src/widgets/categorizer/categorizer.tsx
@@ -22,7 +22,10 @@ import type {
     CategorizerPublicWidgetOptions,
 } from "@khanacademy/perseus-core";
 
-type Props = WidgetProps<RenderProps, PerseusCategorizerUserInput>;
+type Props = WidgetProps<
+    CategorizerPublicWidgetOptions,
+    PerseusCategorizerUserInput
+>;
 
 type DefaultProps = {
     items: Props["items"];
@@ -293,12 +296,6 @@ const styles = StyleSheet.create({
     },
 });
 
-type RenderProps = {
-    items: PerseusCategorizerWidgetOptions["items"];
-    categories: PerseusCategorizerWidgetOptions["categories"];
-    randomizeItems: PerseusCategorizerWidgetOptions["randomizeItems"];
-};
-
 /**
  * @deprecated and likely a very broken API
  * [LEMS-3185] do not trust serializedState/restoreSerializedState
@@ -320,21 +317,12 @@ function getCorrectUserInput(
     return {values: options.values};
 }
 
-function transform(widgetOptions: CategorizerPublicWidgetOptions): RenderProps {
-    return {
-        items: widgetOptions.items,
-        categories: widgetOptions.categories,
-        randomizeItems: widgetOptions.randomizeItems,
-    };
-}
-
 export default {
     name: "categorizer",
     displayName: "Categorizer",
     hidden: true,
     widget: Categorizer,
-    transform,
-    staticTransform: transform,
+    canBeStatic: true,
     getUserInputFromSerializedState,
     getCorrectUserInput,
     isLintable: true,
--- a/packages/perseus/src/widgets/grapher/grapher.tsx
+++ b/packages/perseus/src/widgets/grapher/grapher.tsx
@@ -661,6 +661,7 @@ export default {
     widget: Grapher,
     transform,
     staticTransform,
+    canBeStatic: true,
     getUserInputFromSerializedState,
     getStartUserInput,
 } satisfies WidgetExports<typeof Grapher>;
--- a/packages/perseus/src/widgets/interactive-graph.tsx
+++ b/packages/perseus/src/widgets/interactive-graph.tsx
@@ -862,7 +862,7 @@ export default {
     name: "interactive-graph",
     displayName: "Interactive graph",
     widget: InteractiveGraph,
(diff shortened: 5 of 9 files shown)


packages/perseus/src/widgets/input-number/input-number.tsx
```diff
@@ -185,10 +184,19 @@ class InputNumber extends React.Component<Props> implements Widget {
      * [LEMS-3185] do not trust serializedState/restoreSerializedState
      */
     getSerializedState(): any {
-        const {userInput, ...rest} = this.props;
+        const selectedProps = _.pick(
```
SonicScrewdriver, line 187: Oh no, underscore is sneaking back in! :P

benchristel: We could write our own `pick` instead of using underscore, I suppose.

packages/perseus/src/widgets.ts
```diff
@@ -217,8 +217,7 @@ export const getRendererPropsForWidgetInfo = (
  */
 export const supportsStaticMode = (type: string): boolean | undefined => {
```
benchristel, line 218: The return type can just be `boolean` now:

```suggestion
export const supportsStaticMode = (type: string): boolean => {
```

packages/perseus/src/widgets/input-number/input-number.tsx
handeyeco: [I wrote a little about this here](https://khanacademy.atlassian.net/wiki/spaces/LC/blog/2025/01/31/3692757119/The+Elements+of+Style+After+a+Refactor#Underscore-is-not-intrinsically-bad).

The basic idea is that we shouldn't use Underscore when there are equivalent alternatives (ie `_.each` vs `Array.forEach`). I don't think it's intrinsically bad to use for things that don't have equivalents.

In this case it's easy to argue we have an alternative:

``` TS
const myCoolObject = {one: 1, two: 2, three: 3}

const example1 = _.pick(myCoolObject, "one", "two")

const {one, two} = myCoolObject
const example2 = {one, two}
```

For big objects, the native method is just a little too unruly IMO plus you have to litter the scope with temporary variables. On top of this, we still get TS support:

``` TS
const myCoolObject = {one: 1, two: 2, three: 3}

const example1 = _.pick(myCoolObject, "one", "two")

console.log(example1.one)     // fine
console.log(example1.three)   // TS error
```

So, I dunno, that's why I used it.

packages/perseus/src/types.ts
```diff
@@ -524,6 +524,14 @@ export type WidgetExports<
     isLintable?: boolean;
     tracking?: Tracking;
 
+    /**
+     * Originally we were looking to see if a widget has `staticTransform`
+     * to determine whether or not it could be static; however we'd like
+     * to move away from transforms (especially transforms that aren't
+     * doing anything interesting), so we have a flag for that now.
```
benchristel, line 531: I'd rather not include historical information like this in a doc comment. It's likely to confuse future readers. The doc comment just needs to say what the thing does _right now_.

packages/perseus/src/types.ts
```diff
@@ -524,6 +524,14 @@ export type WidgetExports<
...
+     */
+    canBeStatic?: boolean;
```
benchristel, line 533: Could we name this `supportsStaticMode` for consistency with the function? I don't see a strong reason to have two names for the same idea.

```suggestion
    supportsStaticMode?: boolean;
```

packages/perseus/src/types.ts
```diff
@@ -524,6 +524,14 @@ export type WidgetExports<
...
+
```
jeremywiebe, line 534: I can't comment on the lines below, but if this is about removing `transform` and `staticTransform` can we delete those functions off this type? (lines 535 - 547)

packages/perseus/src/widget-ai-utils/dropdown/dropdown-ai-utils.test.ts
```diff
@@ -55,7 +55,12 @@ describe("Dropdown AI utils", () => {
 
     it("it returns JSON with the expected format and fields", () => {
         const renderProps: any = {
-            choices: ["Pickles", "Tomato", "Onion", "Lettuce"],
+            choices: [
+                {content: "Pickles"},
+                {content: "Tomato"},
+                {content: "Onion"},
+                {content: "Lettuce"},
```
jeremywiebe, line 62: Why did this need to change?  Could we remove the typing of `renderProps` here so that TypeScript can infer based on usage and help us keep the types correct?

packages/perseus/src/widgets/definition/definition.tsx
```diff
@@ -15,9 +15,7 @@ import type {
     PerseusDefinitionWidgetOptions,
 } from "@khanacademy/perseus-core";
 
-type RenderProps = PerseusDefinitionWidgetOptions;
-
-type DefinitionProps = WidgetProps<RenderProps> & {
+type DefinitionProps = WidgetProps<PerseusDefinitionWidgetOptions> & {
     widgets: PerseusRenderer["widgets"];
```
jeremywiebe, line 19: Not exactly related, but I think we can remove the `widgets` here.  The [editor](https://github.com/Khan/perseus/blob/10086652408464bb863cf82f861686163d030e5c/packages/perseus-editor/src/widgets/definition-editor.tsx#L62) disables widgets for the the content this widget renders. I don't see the `Definition` widget's component being used directly anywhere so there's no way that widgets would even be passed in. 

Future PR or a ticket would also be fine. :)

packages/perseus/src/widgets/expression/expression.tsx
```diff
@@ -293,14 +286,34 @@ export class Expression
...
+        );
         return {
-            ...rest,
+            ...selectedProps,
             value: this.props.userInput,
+            keypadConfiguration: {
+                keypadType: "EXPRESSION",
+                times: this.props.times,
+            },
         };
```
jeremywiebe, line 307: This function nicely illustrates how problematic these serialized state functions are... mixing props and user input for the _current_ (as of serialization time) version of the widget. Glad they're on their way out!

packages/perseus/src/widgets/graded-group-set/graded-group-set.tsx
```diff
@@ -88,9 +88,7 @@ class Indicators extends React.Component<IndicatorsProps> {
     }
 }
 
-type RenderProps = PerseusGradedGroupSetWidgetOptions; // no transform
-
-type Props = WidgetProps<RenderProps> & {
+type Props = WidgetProps<PerseusGradedGroupSetWidgetOptions> & {
     trackInteraction: () => void;
```
jeremywiebe, line 92: Pretty sure this is unused and can be removed. Thinking further, these widget props that have extra keys outside of `WidgetProps` and it's generic type parameter (for widget options) feels like it's almost always wrong/unused.

jeremywiebe (commented):
Didn't do a full, thorough review, but have some fly-by comments. :) 

Love that you're continuing to push through this Matthew! Thank-you!

packages/perseus/src/widgets/graded-group-set/graded-group-set.tsx
handeyeco: > Thinking further, these widget props that have extra keys outside of WidgetProps and it's generic type parameter (for widget options) feels like it's almost always wrong/unused.

Yes! It's almost always the wrong way to do things! Some widgets stuff extra state in Renderer's stash of RenderProps (Matrix's `cursorPosition` is a good example) and those I think could probably/usually be internal to the widget. Other uses seem to be manipulating WidgetOptions (which could be computed values within the widget, which would be much easier in functional components) or tweaking types (misuse of generics or converting something into a non-nullable type).

`WidgetProps<> & {` is a Perseus anti-pattern IMO.

handeyeco:
This got too stale, I think it's better to start over when we're ready too address this.

PR #2663, opened 2025-06-27
Parse answerless data

benchristel:
## Summary:
Previously, the Perseus parsers returned errors for answerless data. This meant
that any app relying on serverside scoring could not upgrade its learner-facing
data to the latest schema – which meant any schema changes we released would
break them!

This PR fixes the parsers to accept answerless data and provide defaults for
the fields that the answerless data doesn't have. This is potentially confusing
because it means that widgets will receive data with incorrect answer
information and are expected not to use the answers.  However, the alternative
would be to overhaul our data-schema.ts, splitting the schema into answerful
and answerless variants. We decided early in the project not to attempt that,
and I still think that's probably the right call.

Issue: https://khanacademy.atlassian.net/browse/LEMS-3237

## Test plan:

`pnpm test`

### DIFF
--- a/packages/perseus-core/src/parse-perseus-json/regression-tests/__snapshots__/parse-perseus-json-regression.test.ts.snap
+++ b/packages/perseus-core/src/parse-perseus-json/regression-tests/__snapshots__/parse-perseus-json-regression.test.ts.snap
@@ -491,63 +491,6 @@ exports[`parseAndMigratePerseusItem given categorizer-missing-randomizeItems.jso
 }
 `;
 
-exports[`parseAndMigratePerseusItem given categorizer-missing-randomizeItems.json returns the same result as before with answer information is removed 1`] = `
-{
-  "answerArea": {
-    "calculator": false,
-    "chi2Table": false,
-    "financialCalculatorMonthlyPayment": false,
-    "financialCalculatorTimeToPayOff": false,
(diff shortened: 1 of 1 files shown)


benchristel:
I realized I forgot to commit the regression tests I added, and they're still failing. I need to do more work on this.

packages/perseus/src/widgets/interactive-graph.tsx
```diff
@@ -156,9 +156,7 @@ type RenderProps = {
      */
     graph: PerseusGraphType;
     /**
-     * The correct answer for this widget. Will be undefined if the graph is
-     * being provided answerless data (e.g. because the learner has not yet
-     * submitted their guess).
+     * The correct answer for this widget.
      */
     // TODO(LEMS-2344): make the type of `correct` more specific
     correct?: PerseusGraphType;
```
benchristel, line 162: Maybe `correct` should be required now. The reason I didn't make it required is that there's a type assertion requiring that the answerless data be assignable to the props:

```ts
0 as any as WidgetProps<
    InteractiveGraphPublicWidgetOptions,
    PerseusInteractiveGraphRubric
> satisfies PropsFor<typeof InteractiveGraph>;
```

handeyeco (approved):
Cool!

PR #2664, opened 2025-06-30
Proof-of-concept: Use the object path as the ID of radio choices

benchristel:
Issue: none

## Test plan:

TODO


packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
```diff
@@ -75,6 +75,7 @@ const parseRadioWidgetV2 = parseWidgetWithVersion(
                     isNoneOfTheAbove: optional(boolean),
                     // deprecated
                     widgets: parseWidgetsMapOrUndefined,
+                    id: defaulted(string, (_, ctx) => ctx.getPath().join(".")),
```
benchristel: If we want to be totally sure the path is unique, we could use `formatPath` from `object-path.ts` here.

benchristel:
Closing this PR because I don't think we're going to go this route for assigning IDs to choices.

benchristel (commented):
@anakaren-rojas this commit has the relevant change: https://github.com/Khan/perseus/pull/2664/commits/7daa1aeb768a5d5be5db4c2779d4acdbaae6a88a

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
```diff
@@ -175,11 +177,12 @@ export function migrateV2toV3(
             randomize: options.randomize,
             multipleSelect: options.multipleSelect,
             deselectEnabled: options.deselectEnabled,
-            choices: options.choices.map((choice) => ({
+            choices: options.choices.map((choice, i) => ({
                 content: choice.content,
                 rationale: choice.clue,
                 correct: choice.correct,
                 isNoneOfTheAbove: choice.isNoneOfTheAbove,
+                id: [...ctx.getPath(), "options", "choices", i].join("."),
```
anakaren-rojas, line 185: oh interesting, I tried getting access to the path here but got stuck on the implementation

PR #2666, opened 2025-06-30
Remove unused stats tables from item extras

benchristel:
## Summary:
In the very distant past, Khan Academy's exercise chrome may have used the
`chi2table`, `tTable`, and `zTable` settings to show helpful references to
learners doing statistics exercises. However, these settings have not been used
for years, and I just recently removed all references to them from webapp code
(https://github.com/Khan/frontend/pull/748). This PR removes them from Perseus
as well.

Issue: LEMS-3236

## Test plan:

Review the Item Extras Editor stories in Storybook. There should be no
checkboxes for "t table", "z table", or "chi-squared table".

Screenshot 2025-06-30 at 4 02 33 PM

### DIFF
--- a/.changeset/nine-vans-buy.md
+++ b/.changeset/nine-vans-buy.md
@@ -1,0 +1,7 @@
+---
+"@khanacademy/perseus-core": major
+"@khanacademy/perseus-editor": minor
+"@khanacademy/perseus": patch
+---
+
+Remove `"chi2Table"`, `"tTable"`, and `"zTable"` from the `ItemExtras` type exported from `@khanacademy/perseus-core`. These properties weren't used. This is a breaking change because consumers might see type errors if they set chi2Table, tTable, or zTable properties on the `answerArea` object of a `PerseusItem`. The fix is to avoid setting those properties.


packages/perseus-core/src/parse-perseus-json/perseus-parsers/perseus-answer-area.test.ts
```diff
@@ -7,15 +7,12 @@ import type {PerseusAnswerArea} from "../../data-schema";
 
 describe("parsePerseusAnswerArea", () => {
     const allFalse: PerseusAnswerArea = {
-        zTable: false,
```
handeyeco, line 10: Seeing all this in tests makes me wonder if we should be leveraging `getDefaultAnswerArea` (or something like it) more. No action needed; not related to your PR.

benchristel: It might be nice to have an item builder that could create any type of item with whatever widgets we want. That would be a lot of code, though. Maybe the parsers could create blank "default" items for us. I have an idea about how that could work.

handeyeco: > Maybe the parsers could create blank "default" items for us.

I worded my last comment weirdly. `getDefaultAnswerArea` exists already, we're just not using it everywhere.

PR #2667, opened 2025-07-01
Update sync-dependencies.ts to handle 'catalog:' versions

benchristel:
## Summary:
The frontend repo has moved some package versions to a `pnpm-workspace.yaml`
file — notably React. Our sync-dependencies script was failing because it tried
to parse `catalog:` as a semantic version.

This PR fixes the script to resolve package versions from the catalog.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2903

## Test plan:

Run `utils/sync-dependencies.ts ../frontend/package.json`. The package versions
in pnpm-workspace.yaml should be updated.  None of the versions should be
`catalog:`.

### DIFF
--- a/utils/sync-dependencies.ts
+++ b/utils/sync-dependencies.ts
@@ -104,16 +104,16 @@ function main(argv: string[]) {
         "pnpm-workspace.yaml",
     );
 
-    const ourWorkspace = yaml.parse(
-        fs.readFileSync("pnpm-workspace.yaml", "utf-8"),
-    );
-
     const clientWorkspace = yaml.parse(
         fs.readFileSync(clientWorkspaceYamlPath, "utf-8"),
     );
 
     const clientPackageJson = JSON.parse(
         fs.readFileSync(clientPackageJsonPath, "utf-8"),
+    );
+
+    const ourWorkspace = yaml.parse(
+        fs.readFileSync("pnpm-workspace.yaml", "utf-8"),
     );
 
     const packageNamesInRepo = unique(


utils/sync-dependencies.ts
```diff
@@ -114,14 +128,21 @@ function main(argv: string[]) {
 
     // Dependency ranges used by the consumer of Perseus (like khan/frontend)
     const clientVersionRanges = filterUnusableTargetVersions(
-        JSON.parse(fs.readFileSync(clientPackageJson).toString()).dependencies,
+        clientPackageJson.dependencies,
         packageNamesInRepo,
     );
 
+    function getClientVersionRange(pkgName: string) {
```
handeyeco, line 135: I don't know how to describe this, but there's something weird about the naming here. There's `clientVersionRanges` and `getClientVersionRange`, but we don't use `getClientVersionRange` to set `clientVersionRanges`; rather we use `clientVersionRanges` as an argument for `getClientVersionRange`.

I think maybe `clientVersionRanges` is no longer descriptive of what it is? It's part of the story but not the whole story.

utils/sync-dependencies.ts
```diff
@@ -97,13 +98,26 @@ function main(argv: string[]) {
...
         fs.readFileSync("pnpm-workspace.yaml", "utf-8"),
     );
+
+    const clientWorkspace = yaml.parse(
+        fs.readFileSync(clientWorkspaceYamlPath, "utf-8"),
+    );
+
+    const clientPackageJson = JSON.parse(
+        fs.readFileSync(clientPackageJsonPath, "utf-8"),
+    );
```
handeyeco, line 113: This is a micro-nit, but I feel like the `ourWorkspace` declaration should come after `clientPackageJson` declaration so that it was something like:

```
const clientPackageJsonPath = ...
const clientPackageJson = ...

const clientWorkspaceYamlPath = ...
const clientWorkspace = ...

const ourWorkspace = ...
```

handeyeco (approved):

utils/sync-dependencies.ts
benchristel: I agree. I think the problem is that it isn't a cohesive concept. The other name I thought of is `filteredClientPackageJsonDependencyVersions`. Even that isn't quite right, though, because the values aren't actually _versions_ thanks to NPM's overloading of the field.

I can imagine refactoring this into a better shape, but I think that should be a separate PR.

benchristel: I refactored the code in this PR: https://github.com/Khan/perseus/pull/2673/files

Instead of renaming `clientVersionRanges`, I ended up changing the code around it so the current name makes sense 😂

PR #2668, opened 2025-07-01
[Interactive Graph] [Locked Figures] Add `weight` option to locked vector and locked vector settings

nishasy:
## Summary:
To make locked figures easier to see over axis lines, and to give content authors a bit more freedom in how the locked figures look, we're adding a weight option to locked figures.

Adding this weight select to the locked vector editor and hooking it up to the rendered locked vector here.

Note: The schema has already been updated in the backend in widgets.go, so we should be okay to release this whenever.

## Demo:

https://github.com/user-attachments/assets/3fc3fe87-4f8d-4860-b40c-6a9de8853dd0

## Test plan:
`pnpm jest packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-vector-settings.test.tsx`

Storybook
- Open the locked vector settings
- Confirm that it updates the vector in the preview


packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-vector-settings.tsx
```diff
@@ -152,13 +163,20 @@ const LockedVectorSettings = (props: Props) => {
                 </View>
             }
         >
-            <View style={[styles.row, styles.spaceUnder]}>
-                {/* Line color settings */}
-                <ColorSelect
-                    selectedValue={lineColor}
-                    onChange={handleColorChange}
-                />
-            </View>
```
benchristel, line 161: Just curious — why don't we need the wrapping `View` anymore?

nishasy: 1. It was being used for the `row` style, but the `ColorSelect` was the only thing in the row, so that was unnecessary.
2. It had `styles.spaceUnder` which does not exist in this file, so it wasn't doing anything. I think it was copied from somewhere else and not double checked.

Since both styles were unnecessary, the whole View was unnecessary.

benchristel: > `styles.spaceUnder` which does not exist in this file

Whoa, I'd have expected typechecking to fail on that. Though I guess I understand why it doesn't; the type of `styles` is probably something like `Record<string, CSSProperties>`. That's a good gotcha to know about!

PR #2669, opened 2025-07-01
[Interactive Graph] [Locked Figures] Add weight option to locked ellipse and locked ellipse settings

nishasy:
## Summary:
To make locked figures easier to see over axis lines, and to give content authors a bit more freedom in how the locked figures look, we're adding a weight option to locked figures.

Adding this weight select to the locked ellipse editor and hooking it up to the rendered locked ellipse here.

Note: The schema has already been updated in the backend in widgets.go, so we should be okay to release this whenever.

Demo:

https://github.com/user-attachments/assets/d5c2bbdc-d057-45ec-b3dd-2128673dd2cd

## Test plan:
`pnpm jest packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-ellipse-settings.test.tsx`

Storybook
- Open the locked ellipse settings
- Confirm that it updates the ellipse in the preview


.changeset/strange-sloths-prove.md
```diff
@@ -0,0 +1,7 @@
+---
+"@khanacademy/perseus": patch
+"@khanacademy/perseus-core": patch
+"@khanacademy/perseus-editor": patch
```
benchristel, line 4: I missed this when reviewing earlier PRs, but I think the version bumps should be:

```suggestion
"@khanacademy/perseus": minor
"@khanacademy/perseus-core": major
"@khanacademy/perseus-editor": minor
```

The change to `perseus-core` is breaking because clients that are manually creating `PerseusItem`s or `PerseusArticle`s with locked ellipses will need to update their code to include `weight`.

The changes to the other packages are backward-compatible feature additions, hence the minor version bump.

That said, I honestly don't know how much we should care about following semver. We're the ones fixing things up in webapp and frontend when we make a breaking change, so it's not like these version numbers really help us communicate outside the team. They might help us communicate inside the team, though.

packages/perseus/src/widgets/interactive-graphs/interactive-graph-question-builder.ts
```diff
@@ -413,6 +414,7 @@ class InteractiveGraphQuestionBuilder {
             color: "grayH",
             fillStyle: "none",
             strokeStyle: "solid",
+            weight: options?.weight ?? "medium",
```
benchristel, line 417: I'm really glad we have this builder so we don't have to individually update hundreds of test graphs in different files :)

.changeset/strange-sloths-prove.md
nishasy: Ooo yeah! I'm so used to marking everything as `patch` from making incremental tiny changes.

nishasy: I'm thinking I'll put the version bumps on the last PR of the stack, since it would be weird to have three major version bumps for this.

benchristel: The Changesets CLI collapses the version bumps in each release, so the major version wouldn't go up by 3.

> The core concept that changesets follows is that contributors to a repository should be able to declare an intent to release, and that multiple intents should be able to be combined sensibly. Sensibly here refers to if there is one intent to release button as a 'minor' and another to release button as a 'patch', only one release will be made, at the higher of the two versions.
>
> —https://www.npmjs.com/package/@changesets/cli

...and if 3 changeset files request a major version bump, the major version will go up by 1 the next time the package is released.

nishasy: Would that have the same result as only putting the major version bump on the last PR?

I could put it on all of them for the sake of correctness, but I'd have to update, push, and merge all the stacked PRs.

PR #2675, opened 2025-07-02
Run `pnpm install` at the end of `utils/sync-dependencies.ts`

benchristel:
## Summary:
It's easy to forget to `pnpm install` after syncing dependencies. This PR makes
it automatic.

Issue: LEMS-3269

## Test plan:

`utils/sync-dependencies.ts ../frontend/package.json`. Output from `pnpm
install` should be printed, e.g.

```
> pnpm install
Scope: all 17 workspace projects
Lockfile is up to date, resolution step is skipped
Already up to date
Done in 669ms using pnpm v10.8.0
```


utils/sync-dependencies.ts
```diff
@@ -198,6 +199,9 @@ function main(argv: string[]) {
             encoding: "utf-8",
         },
     );
+
+    process.stderr.write("> pnpm install\n");
```
handeyeco, line 203:
```suggestion
    process.stdout.write("> pnpm install\n");
```

Right?

benchristel: My reasoning is that `stdout` is supposed to be pipeable to other programs. `stderr` is for meta-informational messages. That's probably a purely philosophical point as far as this script is concerned, though, and I should just `console.log`.

PR #2676, opened 2025-07-02
WIP: move choice state out of Renderer

handeyeco:
## Summary:
[rui-radio] split user input from UI state

Issue: LEMS-XXXX

## Test plan:

### DIFF
--- a/packages/perseus/src/widget-ai-utils/radio/radio-ai-utils.test.ts
+++ b/packages/perseus/src/widget-ai-utils/radio/radio-ai-utils.test.ts
@@ -70,6 +70,7 @@ describe("Radio AI utils", () => {
         );
     });
 
+    // why are these tests named the same?
     it("should get prompt json which matches the state of the UI", () => {
         const renderProps: any = {
             numCorrect: 1,
@@ -120,6 +121,7 @@ describe("Radio AI utils", () => {
         });
     });
 
+    // why are these tests named the same?
     it("should get prompt json which matches the state of the UI", async () => {
         const indexToSelect = 1;
         const {renderer} = renderQuestion(shuffledQuestion);
--- a/packages/perseus/src/widgets/group/group.test.tsx
+++ b/packages/perseus/src/widgets/group/group.test.tsx
@@ -182,33 +182,32 @@ describe("group widget", () => {
         await userEvent.type(screen.getAllByRole("textbox")[0], "99");
 
         // Act
-        const userInput = renderer.getUserInput();
+        const userInput = renderer.getUserInputMap();
 
         // Assert
         expect(userInput).toMatchInlineSnapshot(`
-            [
-              [
-                {
-                  "choicesSelected": [
-                    false,
-                    false,
-                    false,
-                    false,
-                    false,
-                  ],
-                },
-              ],
-              [
-                {
-                  "currentValue": "99",
-                },
-                {
-                  "currentValue": "",
-                },
-                undefined,
-              ],
-            ]
-        `);
+{
+  "group 1": {
+    "radio 1": {
+      "choicesSelected": [
+        false,
+        false,
+        false,
+        false,
+        false,
+      ],
+    },
+  },
+  "group 2": {
+    "numeric-input 1": {
+      "currentValue": "99",
+    },
+    "numeric-input 2": {
+      "currentValue": "",
+    },
+  },
+}
+`);
     });
 
     it("should return contained renderer's getSerializedState", async () => {
@@ -224,6 +223,49 @@ describe("group widget", () => {
 {
   "group 1": {
     "radio 1": {
+      "alignment": "default",
+      "choiceStates": [
+        {
+          "correctnessShown": false,
+          "highlighted": false,
+          "previouslyAnswered": false,
+          "rationaleShown": false,
+          "readOnly": false,
+          "selected": false,
+        },
+        {
+          "correctnessShown": false,
+          "highlighted": false,
+          "previouslyAnswered": false,
+          "rationaleShown": false,
+          "readOnly": false,
+          "selected": false,
+        },
+        {
+          "correctnessShown": false,
+          "highlighted": false,
+          "previouslyAnswered": false,
+          "rationaleShown": false,
+          "readOnly": false,
+          "selected": false,
+        },
+        {
+          "correctnessShown": false,
+          "highlighted": false,
+          "previouslyAnswered": false,
+          "rationaleShown": false,
+          "readOnly": false,
+          "selected": false,
+        },
+        {
+          "correctnessShown": false,
+          "highlighted": false,
+          "previouslyAnswered": false,
+          "rationaleShown": false,
+          "readOnly": false,
+          "selected": false,
+        },
+      ],
       "choices": [
         {
           "content": "$45$",
@@ -264,6 +306,7 @@ describe("group widget", () => {
         false,
         true,
       ],
+      "static": false,
     },
   },
   "group 2": {
@@ -431,42 +474,41 @@ describe("group widget", () => {
 
         const guess = renderer.getUserInputMap();
         const score = scorePerseusItem(question1, guess, "en");
-        const guessAndScore = [renderer.getUserInput(), score];
+        const guessAndScore = [renderer.getUserInputMap(), score];
 
         // Assert
         expect(score).toHaveBeenAnsweredCorrectly();
         expect(guessAndScore).toMatchInlineSnapshot(`
-            [
-              [
-                [
-                  {
-                    "choicesSelected": [
-                      false,
-                      false,
-                      false,
-                      false,
-                      true,
-                    ],
-                  },
-                ],
-                [
-                  {
-                    "currentValue": "230",
-                  },
-                  {
-                    "currentValue": "200",
-                  },
-                  undefined,
-                ],
-              ],
-              {
-                "earned": 3,
-                "message": null,
-                "total": 3,
-                "type": "points",
-              },
-            ]
-        `);
+[
+  {
+    "group 1": {
+      "radio 1": {
+        "choicesSelected": [
+          false,
+          false,
+          false,
+          false,
+          true,
+        ],
+      },
+    },
+    "group 2": {
+      "numeric-input 1": {
+        "currentValue": "230",
+      },
+      "numeric-input 2": {
+        "currentValue": "200",
+      },
+    },
+  },
+  {
+    "earned": 3,
+    "message": null,
+    "total": 3,
+    "type": "points",
+  },
+]
+`);
     });
 
     it("should return input paths from contained Renderer", () => {
--- a/packages/perseus/src/widgets/radio/__tests__/radio.test.ts
+++ b/packages/perseus/src/widgets/radio/__tests__/radio.test.ts
@@ -749,7 +749,7 @@ describe("Radio Widget", () => {
     describe("scoring", () => {
         /**
          * (LEMS-2435) We want to be sure that we're able to score shuffled
-         * Radio widgets outside of the component which means `getUserInput`
+         * Radio widgets outside of the component which means `getUserInputMap`
          * should return the same order that the scoring data provides
          */
         it("can be scored correctly when shuffled", async () => {
@@ -761,8 +761,9 @@ describe("Radio Widget", () => {
                 screen.getByRole("radio", {name: /Correct Choice$/}),
             );
 
-            const userInput =
-                renderer.getUserInput()[0] as PerseusRadioUserInput;
+            const userInput = renderer.getUserInputMap()[
+                "radio 1"
+            ] as PerseusRadioUserInput;
             const rubric = shuffledQuestion.widgets["radio 1"].options;
             const widgetScore = scoreRadio(userInput, rubric);
             const rendererScore = scorePerseusItemTesting(
@@ -777,7 +778,7 @@ describe("Radio Widget", () => {
 
         /**
          * (LEMS-2435) We want to be sure that we're able to score shuffled
-         * Radio widgets outside of the component which means `getUserInput`
+         * Radio widgets outside of the component which means `getUserInputMap`
          * should return the same order that the scoring data provides
          */
         it("can be scored incorrectly when shuffled", async () => {
@@ -789,8 +790,9 @@ describe("Radio Widget", () => {
                 screen.getByRole("radio", {name: /Incorrect Choice 1$/}),
             );
 
-            const userInput =
-                renderer.getUserInput()[0] as PerseusRadioUserInput;
+            const userInput = renderer.getUserInputMap()[
+                "radio 1"
+            ] as PerseusRadioUserInput;
             const rubric = shuffledQuestion.widgets["radio 1"].options;
             const widgetScore = scoreRadio(userInput, rubric);
             const rendererScore = scorePerseusItemTesting(
@@ -805,7 +807,7 @@ describe("Radio Widget", () => {
 
         /**
          * (LEMS-2435) We want to be sure that we're able to score shuffled
-         * Radio widgets outside of the component which means `getUserInput`
+         * Radio widgets outside of the component which means `getUserInputMap`
          * should return the same order that the scoring data provides
          */
         it("can be scored correctly when shuffled with none of the above", async () => {
@@ -817,8 +819,9 @@ describe("Radio Widget", () => {
                 screen.getByRole("radio", {name: /None of the above$/}),
             );
 
-            const userInput =
-                renderer.getUserInput()[0] as PerseusRadioUserInput;
+            const userInput = renderer.getUserInputMap()[
+                "radio 1"
+            ] as PerseusRadioUserInput;
             const rubric = shuffledNoneQuestion.widgets["radio 1"].options;
             const widgetScore = scoreRadio(userInput, rubric);
             const rendererScore = scorePerseusItemTesting(
@@ -833,7 +836,7 @@ describe("Radio Widget", () => {
 
         /**
          * (LEMS-2435) We want to be sure that we're able to score shuffled
-         * Radio widgets outside of the component which means `getUserInput`
+         * Radio widgets outside of the component which means `getUserInputMap`
          * should return the same order that the scoring data provides
          */
         it("can be scored incorrectly when shuffled with none of the above", async () => {
@@ -845,8 +848,9 @@ describe("Radio Widget", () => {
                 screen.getByRole("radio", {name: /Incorrect Choice 1$/}),
             );
 
-            const userInput =
-                renderer.getUserInput()[0] as PerseusRadioUserInput;
+            const userInput = renderer.getUserInputMap()[
+                "radio 1"
+            ] as PerseusRadioUserInput;
             const rubric = shuffledNoneQuestion.widgets["radio 1"].options;
             const widgetScore = scoreRadio(userInput, rubric);
             const rendererScore = scorePerseusItemTesting(
--- a/packages/perseus/src/widgets/radio/radio-component.tsx
+++ b/packages/perseus/src/widgets/radio/radio-component.tsx
@@ -75,48 +75,6 @@ class Radio extends React.Component<Props> implements Widget {
         showSolutions: "none",
     };
 
-    static getUserInputFromProps(
-        props: Props,
-        unshuffle: boolean = true,
-    ): PerseusRadioUserInput {
-        // Return checked inputs in the form {choicesSelected: [bool]}. (Dear
-        // future timeline implementers: this used to be {value: i} before
-        // multiple select was added)
-        if (props.choiceStates) {
-            const choiceStates = props.choiceStates;
-            const choicesSelected = choiceStates.map(() => false);
-
-            for (let i = 0; i < choicesSelected.length; i++) {
-                const index = unshuffle ? props.choices[i].originalIndex : i;
-
-                choicesSelected[index] = choiceStates[i].selected;
-            }
-
-            return {
-                choicesSelected,
-            };
-            // Support legacy choiceState implementation
-        }
-        /* c8 ignore if - props.values is deprecated */
-        const {values} = props;
-        if (values) {
-            const choicesSelected = [...values];
-            const valuesLength = values.length;
-
-            for (let i = 0; i < valuesLength; i++) {
-                const index = unshuffle ? props.choices[i].originalIndex : i;
-                choicesSelected[index] = values[i];
-            }
-            return {
-                choicesSelected,
-            };
-        }
-        // Nothing checked
-        return {
-            choicesSelected: props.choices.map(() => false),
-        };
-    }
-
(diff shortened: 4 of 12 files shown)


packages/perseus/src/widgets/group/group.test.tsx
```diff
@@ -224,6 +224,49 @@ describe("group widget", () => {
 {
   "group 1": {
     "radio 1": {
+      "alignment": "default",
+      "choiceStates": [
```
handeyeco, line 228: This is because I initialize `choiceStates` in the widget constructor rather than initializing the state on interaction. I personally think this is better but maybe I'm wrong.

packages/perseus/src/widgets/radio/__tests__/radio.test.ts
```diff
@@ -761,8 +761,9 @@ describe("Radio Widget", () => {
                 screen.getByRole("radio", {name: /Correct Choice$/}),
             );
 
-            const userInput =
-                renderer.getUserInput()[0] as PerseusRadioUserInput;
+            const userInput = renderer.getUserInputMap()[
```
handeyeco: `getUserInput` is deprecated

packages/perseus/src/widgets/radio/radio-component.tsx
```diff
@@ -240,13 +199,16 @@ class Radio extends React.Component<Props> implements Widget {
         this.props.trackInteraction();
     };
 
+    /**
+     * TODO: remove this when everything is pulling from Renderer state
+     * @deprecated get user input from Renderer state
+     */
     getUserInput(): PerseusRadioUserInput {
-        return Radio.getUserInputFromProps(this.props);
+        return this.props.userInput;
```
handeyeco, line 207: We don't have to unshuffle here anymore because we always keep UserInput in a scorable state now.

packages/perseus/src/widgets/radio/radio.ff.tsx
```diff
@@ -29,9 +36,114 @@ class Radio extends RadioOld {
...
+    /**
+     * @deprecated and likely very broken API
+     * [LEMS-3185] do not trust serializedState/restoreSerializedState
+     */
+    getSerializedState() {
+        const {userInput: _, ...rest} = this._mergePropsAndState();
+        return {
+            ...rest,
+        };
+    }
```
handeyeco, line 68: This is just to try and keep things consistent for our deprecated API (which hopefully we can remove soon). I wrote serialization tests before making changes and then I use the serialization APIs to make the tests pass after my changes.

packages/perseus/src/widgets/radio/radio.ff.tsx
```diff
@@ -29,9 +36,114 @@ class Radio extends RadioOld {
...
+             *
+             * This code maintains the original Radio props (ChoiceState)
+             * while allowing Renderer to have UserInput in the shape it needs
+             * and for this component to take over managing UI state
+             * (ChoiceStateWithoutSelected). To do that we convert ChoiceState
+             * into those two chunks of data.
+             */
+            this.setState(
+                {
+                    choiceStates: choiceStates.map((choiceState) => {
```
handeyeco: UI state, now out of Renderer

packages/perseus/src/widgets/radio/radio.ff.tsx
```diff
@@ -29,9 +36,114 @@ class Radio extends RadioOld {
...
+                                ...choiceState,
+                                selected: choiceStates[index].selected,
+                            };
+                        },
+                    );
+                    // Use getUserInputFromSerializedState to get
+                    // unshuffled user input so that we can score with it
+                    const unshuffledUserInput =
+                        getUserInputFromSerializedState(props);
+                    this.props.handleUserInput(unshuffledUserInput);
```
handeyeco, line 119: UserInput state, now in a designated chunk of state in Renderer

packages/perseus/src/widgets/radio/radio.ff.tsx
```diff
@@ -44,13 +156,15 @@ class Radio extends RadioOld {
     }
 
     render(): React.ReactNode {
+        const props = this._mergePropsAndState();
```
handeyeco, line 166: By doing this I'm hoping to avoid being disruptive to the rest of the Radio team's work. Y'all can consume the data like you always have but I can continue with my work.

packages/perseus/src/widgets/radio/radio.ts
```diff
@@ -115,22 +120,32 @@ const transform = (
         countChoices,
         deselectEnabled,
         choices,
+        // doesn't seem used? choiceStates includes selected...
         selectedChoices: _.pluck(choices, "correct"),
     };
 };
 
+function getStartUserInput(
```
handeyeco, line 128: Used to initialize user input state

packages/perseus-editor/src/widgets/__tests__/radio-editor.test.tsx
```diff
@@ -5,7 +5,7 @@ import {userEvent as userEventLib} from "@testing-library/user-event";
 import * as React from "react";
 
 import {testDependencies} from "../../../../../testing/test-dependencies";
-import RadioEditor from "../radio/editor";
+import RadioEditor from "../radio/radio-editor";
```
handeyeco: It was annoying that this is the only editor just called "editor" so I changed it to be less annoying.

packages/perseus/src/widgets/group/group.test.tsx
benchristel: I don't know what the impact of this change will be on consumers, so I'd prefer it if we didn't have to change this test. I could be convinced otherwise if you know why it won't break anything.

packages/perseus/src/widgets/radio/__tests__/radio.test.ts
```diff
@@ -817,8 +819,9 @@ describe("Radio Widget", () => {
                 screen.getByRole("radio", {name: /None of the above$/}),
             );
 
-            const userInput =
-                renderer.getUserInput()[0] as PerseusRadioUserInput;
+            const userInput = renderer.getUserInputMap()[
+                "radio 1"
+            ] as PerseusRadioUserInput;
```
benchristel: Wow, I hate Prettier's choice of whitespace here. No action needed.

Actually, the cast to `PerseusRadioUserInput` doesn't seem to be needed! Maybe removing it will solve two problems:

```suggestion
            const userInput = renderer.getUserInputMap()["radio 1"];
```

packages/perseus/src/widgets/radio/__tests__/radio.test.ts
```diff
@@ -789,8 +790,9 @@ describe("Radio Widget", () => {
                 screen.getByRole("radio", {name: /Incorrect Choice 1$/}),
             );
 
-            const userInput =
-                renderer.getUserInput()[0] as PerseusRadioUserInput;
+            const userInput = renderer.getUserInputMap()[
+                "radio 1"
+            ] as PerseusRadioUserInput;
```
benchristel:
```suggestion
            const userInput = renderer.getUserInputMap()["radio 1"];
```

packages/perseus/src/widgets/radio/__tests__/radio.test.ts
```diff
@@ -761,8 +761,9 @@ describe("Radio Widget", () => {
                 screen.getByRole("radio", {name: /Correct Choice$/}),
             );
 
-            const userInput =
-                renderer.getUserInput()[0] as PerseusRadioUserInput;
+            const userInput = renderer.getUserInputMap()[
+                "radio 1"
+            ] as PerseusRadioUserInput;
```
benchristel:
```suggestion
            const userInput = renderer.getUserInputMap()["radio 1"];
```

packages/perseus/src/widgets/radio/__tests__/radio.test.ts
```diff
@@ -845,8 +848,9 @@ describe("Radio Widget", () => {
                 screen.getByRole("radio", {name: /Incorrect Choice 1$/}),
             );
 
-            const userInput =
-                renderer.getUserInput()[0] as PerseusRadioUserInput;
+            const userInput = renderer.getUserInputMap()[
+                "radio 1"
+            ] as PerseusRadioUserInput;
```
benchristel:
```suggestion
            const userInput = renderer.getUserInputMap()["radio 1"];
```

packages/perseus/src/widgets/radio/util.ts
```diff
@@ -17,3 +19,45 @@ export function getChoiceLetter(pos: number, strings: PerseusStrings): string {
...
     return " ";
 }
+
+export function getUserInputFromSerializedState(
+    props: Props,
+    unshuffle: boolean = true,
+): PerseusRadioUserInput {
+    // Return checked inputs in the form {choicesSelected: [bool]}. (Dear
+    // future timeline implementers: this used to be {value: i} before
+    // multiple select was added)
```
benchristel: I feel like this comment could be removed. The historical note is long out of date, and we don't need to describe the return type because we have actual types.

packages/perseus/src/widgets/radio/radio.ff.tsx
```diff
@@ -29,9 +36,114 @@ class Radio extends RadioOld {
...
+                        return {
+                            ...rest,
+                        };
+                    }),
+                },
+                () => {
+                    // Restructure the data in a format that
+                    // getUserInputFromSerializedState will understand
+                    const props = this._mergePropsAndState();
+                    props.choiceStates = props.choiceStates.map(
```
benchristel: I don't think it's necessarily safe to mutate `props` like this. What if `_mergePropsAndState` evolves to return `this.props` in some cases?

packages/perseus/src/widgets/radio/radio.ff.tsx
```diff
@@ -29,9 +36,114 @@ class Radio extends RadioOld {
...
+     */
+    getSerializedState() {
+        const {userInput: _, ...rest} = this._mergePropsAndState();
+        return {
+            ...rest,
+        };
+    }
+
+    _handleChange(arg: {choiceStates?: ChoiceState[]}) {
+        const choiceStates = arg.choiceStates;
```
benchristel: There are too many different things called `choiceStates`. 😅 I think this one could be called `newChoiceStates`.

```suggestion
        const newChoiceStates = arg.choiceStates;
```

packages/perseus/src/widgets/radio/util.ts
```diff
@@ -17,3 +19,45 @@ export function getChoiceLetter(pos: number, strings: PerseusStrings): string {
...
+
+        for (let i = 0; i < choicesSelected.length; i++) {
+            const index = unshuffle ? props.choices[i].originalIndex : i;
+
+            choicesSelected[index] = choiceStates[i].selected;
+        }
+
+        return {
+            choicesSelected,
+        };
```
benchristel, line 39: I think we could extract this chunk of logic into a function like `unshuffleChoices`, and then call it in the component instead of calling `getUserInputFromSerializedState`. AFAICT the rest of the logic in `getUserInputFromSerializedState` isn't needed for the change you're making in this PR.

packages/perseus/src/widgets/group/group.test.tsx
handeyeco: This is adding stuff to serialization and it's not _new_ stuff per se; serialization added all of this after initial interaction and now we're just adding it on initial render.

So it's not really new stuff and it's for a deprecated API. On top of that, I think this is a side-effect from a better initial state - like I said, now we always have this state for a widget whereas before we only created it on interaction which IMO is a little confusing.

Edit: also if it was trivial to change I would change it, but I think this code is cleaner and a fix might be hacky.

packages/perseus/src/widgets/radio/util.ts
handeyeco: This is function was copied as-is. I'm trying not to step on the Radio teams' toes too much, but I can make this change if you feel like it's a blocker.

packages/perseus/src/widgets/radio/util.ts
handeyeco: This is function was copied as-is. I'm trying not to step on the Radio teams' toes too much, but I can make this change if you feel like it's a blocker.

packages/perseus-editor/src/widgets/__tests__/radio-editor.test.tsx
handeyeco: I changed it back to avoid merge conflicts

packages/perseus/src/widgets/radio/radio.ff.tsx
handeyeco: ~I made a change to clone props/state inside of `_mergePropsAndState` so that we can't accidentally mutate either.~

Actually I couldn't just clone props because our `deepClone` helper doesn't handle functions so I lost all the callbacks. I added a different way to make it safer against accidental mutation.

packages/perseus/src/widgets/group/group.test.tsx
benchristel: > serialization added all of this after initial interaction and now we're just adding it on initial render.

That _could_ be a crucial difference. We track whether the learner has ever interacted with a widget in some places, so the fact that we weren't setting this data until the initial user interaction could be significant.

I don't know, maybe I'm just being paranoid. 🤷

packages/perseus/src/widgets/radio/util.ts
benchristel: I don't feel that strongly either way, but:

- we know that removing a comment won't break anything
- removing the comment would remove a distraction for future readers of the code
- any concurrent changes to this function would cause merge conflicts anyway, since the whole function moved.

Not a blocker, just a suggestion.

packages/perseus/src/widgets/radio/radio.ff.tsx
benchristel, line 1: Huh, I thought I left a comment on this file in my last review, but it seems it got lost. Anyway:

It looks like the original purpose of this component was to render the correct version of the Radio widget based on a feature flag. I.e. the component is temporary scaffolding. By adding more logic to it, we're obscuring its purpose and making it harder to remove.

I'd rather add another wrapper component around this one that contains the UI state.

packages/perseus/src/widgets/radio/util.ts
benchristel: Not a blocker. I had some trouble following the code in this area, but once I realized that we were only using _part_ of this function, I felt like I understood things a bit better.

packages/perseus/src/widget-ai-utils/radio/radio-ai-utils.test.ts
```diff
@@ -120,6 +121,7 @@ describe("Radio AI utils", () => {
         });
     });
 
+    // why are these tests named the same?
```
SonicScrewdriver, line 124: Great question

packages/perseus/src/widgets/radio/__tests__/radio.test.ts
```diff
@@ -761,8 +758,7 @@ describe("Radio Widget", () => {
                 screen.getByRole("radio", {name: /Correct Choice$/}),
             );
 
-            const userInput =
-                renderer.getUserInput()[0] as PerseusRadioUserInput;
+            const userInput = renderer.getUserInputMap()["radio 1"];
```
SonicScrewdriver, line 761: This already looks cleaner!

packages/perseus/src/widgets/radio/radio.ff.tsx
SonicScrewdriver: This is a good point! Radio.ff is definitely intended to be fully deleted once we're done with the Radio project. 

However, I'm not totally against making a ticket for myself to integrate this logic into the new [multiple-choice-widget](https://github.com/Khan/perseus/pull/2629/files#diff-c2a433ec93d784f710ad14ec29aa60320faaa322b5b6818a5025696f96b679d8) file **_when we're getting ready to get rid of the feature flag._** I've set that file up to hold all of the Perseus / Widget logic so that seems like a logical place. 

Less relevant: I was considering splitting out the logic from radio.ts into this file at that point as well (move the widget registration into multiple-choice-widget, and split the functions into testable exports in a utils file). Thought I'd highlight this consideration in case it affects anything related to your work  - and also to indicate that we'll have a few clean-up things to do when we're removing the radio.ff file anyway. :)

packages/perseus/src/widgets/radio/radio.ff.tsx
```diff
@@ -1,21 +1,30 @@
+import {
+    deepClone,
+    type PerseusRadioRubric,
+    type PerseusRadioUserInput,
+} from "@khanacademy/perseus-core";
 import * as React from "react";
+import _ from "underscore";
```
SonicScrewdriver: Is underscore even used in this file anymore? Maybe I'm missing it, but I don't see it in any of your new logic

packages/perseus/src/widgets/radio/util.ts
SonicScrewdriver: I'm happy either way! We DO have logic to support "value" when we're building the props for base-radio (soon to be multiple-choice-component), but I've been wondering if we could just handle that over in the parser so that nothing else has to support this old approach.

packages/perseus/src/widgets/radio/util.ts
SonicScrewdriver: Again I'm happy either way! I don't feel our toes are stepped on too much with the user input stuff. I'd be happy to take a look at refactoring this as I know y'all have tons of other timely concerns. 

I'm noticing _some_ similarities between this and the cleaned up getChoiceStates I created over in [general-utils.ts](https://github.com/Khan/perseus/blob/main/packages/perseus/src/widgets/radio/utils/general-utils.ts).

packages/perseus/src/widgets/radio/serialize-radio.test.ts
```diff
@@ -0,0 +1,208 @@
...
+            originalIndex: 0,
+        },
+        {
+            content: "Content 3",
+            correct: false,
+            originalIndex: 2,
+        },
+    ],
+    // no idea what this is, it doesn't seem to change...
+    selectedChoices: [false, false, true, false],
```
SonicScrewdriver, line 47: I suspect this is an archaic remnant

packages/perseus/src/widgets/radio/radio.class.new.tsx
```diff
@@ -21,6 +21,7 @@ export type RenderProps = {
     countChoices?: boolean;
```
SonicScrewdriver, line 21: I've deleted this file in my restructure PR, but I'll happily pull this work over into that PR before landing. (I'll let you land first). I'll send you a message once I've integrated your changes, just so that you can review to verify that I didn't miss anything. :)

packages/perseus/src/widgets/radio/radio-component.tsx
```diff
@@ -27,6 +27,7 @@ export type RenderProps = {
     countChoices?: boolean;
```
SonicScrewdriver, line 27: Same note here as radio.class.new, so I'll copy the comment: I've deleted this file in my restructure PR, but I'll happily pull this work over into that PR before landing. (I'll let you land first). I'll send you a message once I've integrated your changes, just so that you can review to verify that I didn't miss anything. :)

SonicScrewdriver (approved):
I left a few comments here about the conflicts with the Radio Widget Project, but I'm happy to approve so that you're unblocked. 

Y'all have tons of time-sensitive work on your ends, so I'm happy to create myself some tickets to help cleanup / do any refactoring of these changes. It's probably also easier for us to find these opportunities as we have a better idea for what's changing / moving.

packages/perseus/src/widgets/radio/radio.class.new.tsx
```diff
@@ -21,6 +21,7 @@ export type RenderProps = {
...
     deselectEnabled?: boolean;
     choices: ReadonlyArray<RadioChoiceWithMetadata>;
+    // doesn't seem used? choiceStates includes selected...
```
ivyolamit, line 24: This is interesting, you're right it's not used. Looks like it's a remnant from the past since the last known history it was from "[Copy over perseus-all-package and perseus-merged-editor-package](https://github.com/Khan/perseus/commit/afeec158b6d3cdd5d62a538f32480e83211b3594)". Maybe @jeremywiebe has context on this, and we can safely remove it on another PR. 😉

packages/perseus/src/widgets/radio/radio.ff.tsx
ivyolamit: it's weird that lint error didn't catch this 🤔 it usually complains about unused imports. Good catch @SonicScrewdriver

handeyeco: `_` is ignored by the linter to allow for:

``` TS
// remove `something` from `myObj`
const {something: _, ...rest} = myObj;
```

I'll fix it, good catch.

packages/perseus/src/widgets/group/group.test.tsx
handeyeco: I feel okay about taking the risk since it's a risk for a deprecated and known problematic API (and we've signaled to consumers that it's prone to breakage) and in exchange for taking the risk I think we have cleaner, more intuitive code.

packages/perseus/src/widgets/radio/util.ts
handeyeco: Removed!

packages/perseus/src/widgets/radio/util.ts
handeyeco: I think it's a good suggestion, but I'm going to leave this up to the Radio team to pursue if they want to.

packages/perseus/src/widgets/radio/radio.ff.tsx
handeyeco: > However, I'm not totally against making a ticket for myself to integrate this logic into the new [multiple-choice-widget](https://github.com/Khan/perseus/pull/2629/files#diff-c2a433ec93d784f710ad14ec29aa60320faaa322b5b6818a5025696f96b679d8) file when we're getting ready to get rid of the feature flag.

Okay, I think that sounds good. 👍

packages/perseus/src/widgets/radio/util.ts
```diff
@@ -17,3 +19,42 @@ export function getChoiceLetter(pos: number, strings: PerseusStrings): string {
...
+export function getUserInputFromSerializedState(
+    props: Props,
+    unshuffle: boolean = true,
+): PerseusRadioUserInput {
+    if (props.choiceStates) {
+        const choiceStates = props.choiceStates;
+        const choicesSelected = choiceStates.map(() => false);
+
+        for (let i = 0; i < choicesSelected.length; i++) {
+            const index = unshuffle ? props.choices[i].originalIndex : i;
```
benchristel, line 32: I'm late commenting on this, but: do we still need `unshuffle`? I think it's always set to true, and if we always have the `originalIndex`, we should just use it anyway.

PR #2678, opened 2025-07-03
UserInputManager: a central state for user input

handeyeco:
## Summary:

Integration with FE: https://github.com/Khan/frontend/pull/1702

In my previous PRs, I went widget-by-widget refactoring user input state to be an isolated chunk of state separate from other UI state and stored in Renderer (see https://github.com/Khan/perseus/pull/2566 for more context). The only widgets I didn't touch were the "group" widgets: Group, GradedGroup, and GradedGroupSet. The problem was that the group widgets use Renderers themselves so we had multiple layers of user input:

```
- ServerItemRenderer
  - Renderer (has user input)
    - Group
      - Renderer (has user input)
```

My goal however was to store the full user input in one place. In this PR I add UserInputManager, a stateful wrapper, and I make Renderer consume user input state.

```
- ServerItemRenderer
  - UserInputManager (has user input)
    - Renderer 
      - Group
        - Renderer
```

This had a lot of cause/effect that made this PR get a little bloated. An overview of changes:

- Add UserInputManager and tests
- Since Renderer is no longer managing user input state itself, I needed to add some props to it:
  - `userInput` the map of widget user input
  - `handleUserInput` callback to change user input for a widget
  - `initializeUserInput` used to set/reset user input to an initial state
  - `restoreUserInputFromSerializedState` used to restore user input from serialized state
- Removed `getUserInput` and `getUserInputMap` from widgets since parent components should already have user input
- Updated `getPromptJSON` to get user input from props since all widgets should now receive `props.userInput` and I removed `getUserInput` from widgets
- Added some `getStartUserInput` implementations to widgets to init user input state
- Update test utils to use `UserInputManager`
- The `setInputValue` API broke resulting in test failures. [We decided to remove the API.](https://khanacademy.slack.com/archives/C01AZ9H8TTQ/p1751991376195009)

Issue: [LEMS-3208](https://khanacademy.atlassian.net/browse/LEMS-3208)

## Test plan:

This is meant to be an under-the-hood kind of change. Nothing should really visibly change, just an infrastructure refactor.


packages/perseus/src/__tests__/renderer-api.test.tsx
```diff
@@ -40,77 +38,6 @@ describe("Perseus API", function () {
         registerWidget("mock-widget", MockWidget);
     });
 
-    describe("setInputValue", function () {
```
handeyeco, line 43: `setInputValue` API wasn't being used. [See link.](https://khanacademy.slack.com/archives/C01AZ9H8TTQ/p1751991376195009)

packages/perseus/src/__tests__/renderer.test.tsx
```diff
@@ -1322,7 +1322,7 @@ describe("renderer", () => {
             rerender(question1, extraProps);
 
             // How do I tell it rerendered?
-            expect(onRender).toHaveBeenCalledTimes(2);
+            expect(onRender).toHaveBeenCalledTimes(3);
```
handeyeco: I think this is just because Renderer is wrapped in `UserInputManager` now which initializes state, triggering a rerender. I question how useful this test is TBH.

packages/perseus/src/__tests__/renderer.test.tsx
```diff
@@ -1643,58 +1643,6 @@ describe("renderer", () => {
         });
     });
 
-    describe("setInputValue", () => {
```
handeyeco, line 1646: `setInputValue` API wasn't being used. [See link.](https://khanacademy.slack.com/archives/C01AZ9H8TTQ/p1751991376195009)

packages/perseus/src/article-renderer.tsx
```diff
@@ -216,43 +217,73 @@ class ArticleRenderer
...
-                </div>
-            );
-        });
+        const sections = this._sections().map(
+            (section: PerseusRenderer, sectionIndex: number) => {
+                return (
+                    <div key={sectionIndex} className="clearfix">
+                        <UserInputManager
+                            widgets={section.widgets}
+                            problemNum={0}
```
handeyeco: Articles don't have problem numbers I don't believe.

packages/perseus/src/server-item-renderer.tsx
```diff
@@ -438,30 +439,58 @@ export class ServerItemRenderer
...
-                />
+                    problemNum={this.props.problemNum ?? 0}
+                >
+                    {({
+                        userInput,
+                        handleUserInput,
+                        initializeUserInput,
+                        restoreUserInputFromSerializedState,
+                    }) => {
+                        this.userInput = userInput;
```
handeyeco, line 456: This is kind of gross. In a perfect world the consumer of Perseus would wrap `ServerItemRenderer` with `UserInputManager` which would give them access to user input without the imperative `getUserInput` API. What would make this easier to do is to combine context/hooks: the user input could be in a context provider and widgets can consume their user input directly from the context using a hook.

I don't think this is the right time for that though since I've already spent a ton of time refactoring. So my middle ground was to have a single user input stash and to access it via the existing `getUserInput` API; `this.userInput = userInput;` lets me access that state in `getUserInput`.

packages/perseus/src/types.ts
```diff
@@ -101,28 +98,6 @@ export interface Widget {
...
-    /**
-     * TODO: remove this when everything is pulling from Renderer state
-     * @deprecated get user input from Renderer state
-     */
-    getUserInputMap?: () => UserInputMap | undefined;
-    /**
-     * TODO: remove this when everything is pulling from Renderer state
-     * @deprecated get user input from Renderer state
-     */
-    getUserInput?: () => UserInputArray | UserInput | undefined;
```
handeyeco, line 124: All dead APIs now

packages/perseus/src/types.ts
```diff
@@ -548,7 +523,10 @@ export type WidgetExports<
     /**
      * @deprecated - do not use in new code.
      */
-    getUserInputFromSerializedState?: (props: any) => TUserInput;
+    getUserInputFromSerializedState?: (
+        props: any,
+        widgetOptions?: any, // <= for groups
```
handeyeco: SerializedState has widget IDs but not widget types. We don't want to determine widget type from widget IDs, but we need the widget type to get the correct `getUserInputFromSerializedState` function for the widget. In ServerItemRenderer/Renderer we already have access to the WidgetOptions but Groups didn't.

packages/perseus/src/user-input-manager.tsx
```diff
@@ -0,0 +1,167 @@
...
+};
+
+/**
+ * Initialize the starting UserInput state:
+ * - for static widgets, that's the correct UserInput
+ * - some widgets just like having something to start with
+ * - some widgets use initial UserInput state as the "shuffled state"
+ *   (which is why we need problemNum since it's the seed)
+ */
+export function sharedInitializeUserInput(
```
handeyeco, line 48: These functions are exported because we need them in the group widgets.

packages/perseus/src/widget-ai-utils/categorizer/categorizer-ai-utils.test.ts
```diff
@@ -53,17 +53,18 @@ describe("Categorizer AI utils", () => {
...
             values: [1, 0, 1, 1],
+            userInput,
         };
 
-        const userInput: PerseusCategorizerUserInput = {
-            values: [1, 0, 0, 1],
-        };
-
-        const resultJSON = getPromptJSON(renderProps, userInput);
+        const resultJSON = getPromptJSON(renderProps);
```
handeyeco, line 67: All widgets receive user input from props now, so I refactored these functions to only take props rather than props + user input.

packages/perseus/src/widget-ai-utils/matcher/matcher-ai-utils.test.tsx
```diff
@@ -128,7 +127,13 @@ describe("Matcher AI utils", () => {
                             "Left - Four",
                             "Left - Five",
                         ],
-                        right: shuffledRightItems,
+                        right: [
```
handeyeco, line 125: I think this was a mistake with the test itself, it seemed to use user input twice whereas I think it was wanting the correct order of things. I'll CC Mike.

packages/perseus/src/widgets/categorizer/categorizer.tsx
```diff
@@ -53,14 +53,6 @@ export class Categorizer
         uniqueId: _.uniqueId("perseus_radio_"),
     };
 
-    /**
-     * TODO: remove this when everything is pulling from Renderer state
-     * @deprecated get user input from Renderer state
-     */
-    getUserInput(): PerseusCategorizerUserInput {
-        return this.props.userInput;
-    }
```
handeyeco, line 62: User input is now stored in the parent and we no longer access it through widgets' `getUserInput`

packages/perseus/src/widgets/expression/expression.test.tsx
```diff
@@ -453,30 +453,6 @@ describe("Expression Widget", function () {
...
         });
 
-        it("sets input value directly", () => {
-            // arrange
-            const {renderer} = renderQuestion(expressionItem2.question);
-            act(() => jest.runOnlyPendingTimers());
-
-            // act
-            act(() =>
-                renderer.setInputValue(["expression 1"], "123-x", () => {}),
```
handeyeco, line 491: Removed API

packages/perseus/src/widgets/graded-group-set/graded-group-set.test.ts
```diff
@@ -189,24 +189,6 @@ describe("graded group widget", () => {
...
     });
 
-    it("should be able to set input value on child widget", () => {
-        // Arrange
-        const {renderer} = renderQuestion(article1);
-        const cb = jest.fn();
-
-        // Act
-        act(() =>
-            renderer.setInputValue(
```
handeyeco, line 199: Removed API

packages/perseus/src/widgets/graded-group/graded-group-answer-bar.tsx
```diff
@@ -12,18 +12,22 @@ import {phoneMargin, negativePhoneMargin} from "../../styles/constants";
 
 import type {APIOptions} from "../../types";
 
-export type ANSWER_BAR_STATES = // Initial state before the question is answerable.  The user must complete
+export type ANSWER_BAR_STATES =
+    // Initial state before the question is answerable.  The user must complete
```
handeyeco, line 16: These comments were all messed up

packages/perseus/src/widgets/graded-group/graded-group.tsx
```diff
@@ -386,18 +410,45 @@ export class GradedGroup
...
-                            <Renderer
-                                {...this.props.hint}
-                                ref={this.hintRendererRef}
-                                apiOptions={apiOptions}
-                                linterContext={this.props.linterContext}
-                                strings={this.context.strings}
-                                showSolutions={showSolutions}
-                            />
+
+                            <UserInputManager
```
handeyeco, line 392: This is for hints...which I think can be interactive? But we don't want that input to be in the global store?

packages/perseus/src/widgets/group/group.test.tsx
```diff
@@ -186,29 +191,25 @@ describe("group widget", () => {
...
-                  "currentValue": "99",
-                },
-                {
-                  "currentValue": "",
-                },
-                undefined,
-              ],
-            ]
-        `);
+[
```
handeyeco: Jest is doing this. I don't fight with automated tools.

packages/perseus/src/widgets/group/group.test.tsx
```diff
@@ -535,23 +535,6 @@ describe("group widget", () => {
...
     });
 
-    it("should set value and call callback for input path", async () => {
-        // Arrange
-        const {renderer} = renderQuestion(question1);
-        const cb = jest.fn();
-
-        // Act
-        act(() =>
-            renderer.setInputValue(["group 2", "numeric-input 2"], "2021", cb),
```
handeyeco, line 545: Removed API

packages/perseus/src/widgets/interaction/__snapshots__/interaction.test.ts.snap
```diff
@@ -647,7 +647,7 @@ exports[`interaction widget renders movable point elements with blank constraint
                 style="position: absolute; left: 0px; top: 0px;"
               >
                 <div
-                  style="position: absolute; width: 26px; height: 26px; left: 23.36363636363636px; top: 350.6363636363636px; transform: translateX(0px) translateY(0px) translateZ(0);"
+                  style="position: absolute; width: 26px; height: 26px; left: 23.36363636363636px; top: 350.6363636363636px; transform: translateX(0px) translateY(-72.72727272727275px) translateZ(0);"
```
handeyeco, line 650: I think this change has to do with the fact that UserInputManager is causing an additional rerender. I made a story and compared it between `main` and this PR - no changes. I think this is a good example of why we don't want to test implementation details; I believe we're experiencing the same user-facing behavior but we're asserting against code.

packages/perseus/src/widgets/matcher/matcher.tsx
```diff
@@ -24,7 +24,7 @@ const HACKY_CSS_CLASSNAME = "perseus-widget-matcher";
 
 type RenderProps = Pick<
     PerseusMatcherWidgetOptions,
-    "labels" | "orderMatters" | "padding"
+    "labels" | "orderMatters" | "padding" | "left" | "right"
```
handeyeco, line 27: I think `getPromptJson` wants this? Again, I'll CC Mike.

packages/perseus-core/src/validation.types.ts
```diff
@@ -120,9 +120,12 @@ export type PerseusGroupValidationData = {widgets: ValidationDataMap};
 export type PerseusGroupUserInput = UserInputMap;
 
 export type PerseusGradedGroupRubric = PerseusGradedGroupWidgetOptions;
+export type PerseusGradedGroupUserInput = UserInputMap;
 
 export type PerseusGradedGroupSetRubric = PerseusGradedGroupSetWidgetOptions;
 
+export type PerseusGradedGroupSetUserInput = UserInputMap[];
+
```
benchristel: Why do GradedGroup and GradedGroupSet need user input types? Can't we keep scoring internal to those widgets and let the externally-visible user input be `null` or whatever?

handeyeco:
@MikeKlemarewski I CC'd you since I changed a bunch of AI utils

packages/perseus/src/article-renderer.tsx
```diff
@@ -219,37 +220,59 @@ class ArticleRenderer
...
+                                        newFocusPath,
+                                        oldFocusPath,
+                                    ) => {
+                                        // Prefix the paths with the relevant section index,
+                                        // so as to allow us to distinguish between
+                                        // equivalently-named inputs across Renderers.
+                                        this._handleFocusChange(
+                                            newFocusPath &&
+                                                [sectionIndex].concat(
+                                                    newFocusPath as any,
```
benchristel, line 254: The `as any` casts weren't in the original code. Are they needed?

packages/perseus/src/renderer.tsx
```diff
@@ -117,6 +121,10 @@ type WidgetState = {
 };
 
 type Props = Partial<React.ContextType<typeof DependenciesContext>> & {
+    userInput?: UserInputMap;
+    handleUserInput?: HandleUserInputCallback;
+    initializeUserInput?: InitializeUserInputCallback;
+    restoreUserInputFromSerializedState?: RestoreUserInputFromSerializedStateCallback;
```
benchristel, line 127: Could these props be required?

packages/perseus/src/renderer.tsx
```diff
@@ -203,26 +210,33 @@ type DefaultProps = Required<
...
         };
         const answerless = splitPerseusItem(answerful);
         return JSON.stringify(answerless);
     }
 
-    return answerlessStringified(prev) !== answerlessStringified(next);
+    return (
+        propsA.problemNum !== propsB.problemNum ||
+        answerlessStringified(makeRenderer(propsA.content, propsA.widgets)) !==
+            answerlessStringified(makeRenderer(propsB.content, propsB.widgets))
```
benchristel: Just noticed this — comparing JSON strings is a brittle form of deep comparison because the property order can vary. E.g.

```json
{"foo":1,"bar":2}
```

and

```json
{"bar":2,"foo":1}
```

represent equivalent objects but will be treated as different by this check.

~~Here's an alternative: https://github.com/benchristel/taste/blob/main/src/predicates.js#L7-L40~~ EDIT: we should use `_.isEqual` instead of writing our own.

packages/perseus/src/renderer.tsx
```diff
@@ -305,6 +319,11 @@ class Renderer
         this.handleRender({});
         this._currentFocus = null;
 
+        this.props.initializeUserInput?.(
+            this.props.widgets,
+            this.props.problemNum ?? 0,
+        );
```
benchristel, line 322: Why do we need a callback for this? The parent component already has this information (since it just passed it in).

packages/perseus/src/renderer.tsx
```diff
@@ -417,7 +440,7 @@ class Renderer
 
         if (
             this.props.serializedState &&
-            !_.isEqual(this.props.serializedState, this.getSerializedState())
+            !_.isEqual(this.props.serializedState, prevProps.serializedState)
```
benchristel, line 440: I think this changes behavior, but it probably doesn't matter much?

packages/perseus/src/renderer.tsx
```diff
@@ -1741,57 +1712,36 @@ class Renderer
...
-                userInputMap[id] = widget.getUserInput();
-            }
-        });
-        return userInputMap;
+        const userInput = this.props.userInput;
+        if (!userInput) {
+            throw new Error(
+                `getUserInputMap called without providing userInput to Renderer`,
+            );
+        }
```
benchristel, line 1740: If we made `userInput` a required prop, we wouldn't need this check.

packages/perseus/src/renderer.tsx
```diff
@@ -1828,10 +1778,16 @@ class Renderer
      * @deprecated use scorePerseusItem
      */
     score(): PerseusScore {
+        if (!this.props.userInput) {
+            throw new Error(
+                `score called without providing userInput to Renderer`,
+            );
+        }
```
benchristel, line 1782: If we made `userInput` a required prop, we wouldn't need this check.

packages/perseus/src/renderer.tsx
```diff
@@ -1741,57 +1712,36 @@ class Renderer
...
      * TODO: can we remove this?
      * @deprecated use getUserInputMap
      */
     getUserInput(): UserInputArray {
+        const userInput = this.props.userInput;
+        if (!userInput) {
+            throw new Error(
+                `getUserInput called without providing userInput to Renderer`,
+            );
+        }
```
benchristel, line 1724: If we made `userInput` a required prop, we wouldn't need this check.

packages/perseus/src/server-item-renderer.tsx
benchristel: Agreed about the ideal state and intermediate compromise. 👍

packages/perseus/src/types.ts
benchristel, line 1: The changes in this file require a major version bump, right?

packages/perseus/src/user-input-manager.test.tsx
```diff
@@ -0,0 +1,330 @@
...
+} from "./user-input-manager";
+import type {PerseusWidgetsMap} from "@khanacademy/perseus-core";
+import type {UserEvent} from "@testing-library/user-event";
+
+function generateNumberLineMap(): PerseusWidgetsMap {
+    return {
+        "number-line 1": {
+            type: "number-line",
+            options: {
+                isInequality: true, // <= important
```
benchristel: I don't think this is the only important property here. I think `range` is important too, and maybe `tickStep` and `numDivisions`.

Ideally, I think the individual tests would specify the properties that are important to them.

packages/perseus/src/widget-ai-utils/matcher/matcher-ai-utils.test.tsx
```diff
@@ -64,24 +64,23 @@ describe("Matcher AI utils", () => {
...
-                left: ["1", "2", "3"],
-                right: ["a", "b", "c"],
-            },
-            orderMatters: false,
-        };
-
-        const userInput: PerseusMatcherUserInput = {
-            left: ["3", "1", "2"],
+            left: ["1", "2", "3"],
             right: ["a", "b", "c"],
```
benchristel, line 76: The `left` and `right` properties were added to `renderProps`. Is that intentional? It would be great if we didn't have to use an `any` type here.

packages/perseus/src/widgets/expression/expression-mobile.test.tsx
```diff
@@ -82,7 +82,6 @@ function ConnectedRenderer({item = expressionItem2}) {
 describe("expression mobile", () => {
     beforeAll(() => {
         // TODO(LEMS-2656): remove TS suppression
```
benchristel: The TODO comment can be deleted too! 🎉

packages/perseus/src/widgets/numeric-input/numeric-input.class.tsx
```diff
@@ -213,7 +183,6 @@ export default {
     transform: propsTransform,
     isLintable: true,
     // TODO(LEMS-2656): remove TS suppression
```
benchristel: We can delete this TODO now! 🎉

packages/perseus/src/widget-ai-utils/graded-group/graded-group-ai-utils.test.ts
```diff
@@ -58,7 +58,7 @@ const question: PerseusRenderer = {
                             static: true,
                             graded: true,
                             options: {
-                                static: false,
+                                static: true,
```
benchristel: Why did this change?

packages/perseus/src/user-input-manager.tsx
```diff
@@ -0,0 +1,167 @@
...
```
benchristel, line 48: What does `shared` mean in these function names?

packages/perseus/src/user-input-manager.tsx
```diff
@@ -0,0 +1,167 @@
...
+ * Restore user input from serialized state. It's tricky
+ * because there is no definite type for serialized state,
+ * so widgets need to handle it themselves and overall the process
+ * is very fragile.
+ *
+ * @deprecated - do not use in new code.
+ */
+export function sharedRestoreUserInputFromSerializedState(
+    serializedState: unknown,
+    widgetOptions: PerseusWidgetsMap,
```
benchristel: Why is this called `widgetOptions`? Seems like it should be `widgets` or `widgetsMap`?

packages/perseus/src/widgets/group/group.tsx
```diff
@@ -198,10 +190,29 @@ class Group extends React.Component<Props> implements Widget {
...
+function getStartUserInput(
+    options: PerseusRenderer,
+    problemNum: number,
+): PerseusGroupUserInput {
+    return sharedInitializeUserInput(options.widgets, problemNum);
+}
+
+function getUserInputFromSerializedState(
+    serializedState: unknown,
+    widgetOptions: any,
```
benchristel: Could this type be `PerseusWidgetsMap` instead of `any`?

packages/perseus/src/user-input-manager.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+export type RestoreUserInputFromSerializedStateCallback = (
+    serializedState: unknown,
+    widgetOptions: PerseusWidgetsMap,
+) => void;
+
+export type HandleUserInputCallback = (
+    id: string,
+    userInput: UserInputMap[keyof UserInputMap],
+) => void;
```
benchristel: I think it's going to be hard to implement this callback in a typesafe way, because the type of `id` isn't correlated with the type of `userInput`. What about:

<details>
<summary>EDIT: my first suggestion didn't work</summary>

```ts
export type HandleUserInputCallback<ID extends keyof UserInputMap> = (
    id: ID,
    userInput: UserInputMap[ID],
) => void;
```

</details>

or

```ts
export type HandleUserInputCallback = (
    userInput: UserInputMap,
) => void;
```

packages/perseus/src/user-input-manager.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+type WrapperPayload = {
```
benchristel: React calls the "wrapper" a ["render prop"]. Other sources refer to it as a ["function as children" or "render callback"](https://medium.com/@abubakarmemon/pass-function-as-children-in-react-a1bac3e11d58). These are all pretty terrible names, but at least they're somewhat widespread.

["render prop"]: https://react.dev/reference/react/Children#calling-a-render-prop-to-customize-rendering

```suggestion
type RenderPropParams = {
```

packages/perseus/src/__tests__/renderer.test.tsx
ivyolamit: looking at the code, probably okay to remove this, since other test on checking what's displayed would satisfy this scenario.

packages/perseus/src/article-renderer.tsx
handeyeco: It's because I typed some of the surrounding code. I tried to figure out what the original intent was but it wasn't immediately clear. This is still a net positive because more of the code is typed now, but the tradeoff was marking these as `any`. I'll try again to see what's going on.

packages/perseus/src/renderer.tsx
handeyeco: `Renderer` is used all over the place but very few uses need user input.

packages/perseus/src/user-input-manager.tsx
handeyeco: I feel like `RenderProps` has too much of a loaded meaning in Perseus. I changed it to `UserInputManagerPayload`, does that seem better?

ivyolamit (commented):
@handeyeco I'm not able to comment on the entire PR due to my limited context, but for the changes in Radio looks good to me and still working as expected. I did try to check other changes and observed:
- pattern of the changes in the widgets are consistent, remaining usage I saw was in the interactive graph which is still valid usage
- other code clean up looks valid to me and make sense
- looks like it's not a breaking change in webapp front-end base on [existing usages](https://github.com/search?q=repo%3AKhan%2Ffrontend%20.getUserInput(&type=code)

packages/perseus/src/user-input-manager.tsx
handeyeco: Hmm, not sure. This _is_ typed at the widget level, basically:

```
type UniversalWidgetProps<
    TUserInput = Empty,
> = {
    userInput: TUserInput;
    handleUserInput: (
        newUserInput: TUserInput,
    ) => void;
}

export type WidgetExports<
    TUserInput = Empty,
> = Readonly<{
    getUserInputFromSerializedState?: (
        serializedState: unknown,
        widgetOptions?: PerseusWidgetsMap,
    ) => TUserInput;

    getCorrectUserInput?: (widgetOptions: WidgetOptions) => TUserInput;

    getStartUserInput?: (
        widgetOptions: WidgetOptions,
        problemNum: number,
    ) => TUserInput;
}
```

The Renderer/UserInputManager just pass that data up.

packages/perseus/src/renderer.tsx
handeyeco: So just to expand on this after a conversation with Ben:

We kind of over-rely on `Renderer`. We use it as a high-level component that's tasked with managing several things (including user input before this PR) but we also use it as a low-level generic markdown renderer.

For instance, we use it in each choice within a Radio and each option within a Dropdown. These uses can never and should never have a fully interactive scorable widget. We're just using `Renderer` because it handles markdown (and likely TeX) for us.

Jeremy and I have talked at one point of splitting `Renderer` into two components: one for managing widgets (to be used by ServerItemRenderer, ArticleRenderer, HintsRenderer, and the group widgets) and one for just rendering static content (for things like Radio and Dropdown). But I think that's beyond the scope of this PR and not urgently needed.

packages/perseus/src/__tests__/renderer.test.tsx
jeremywiebe: Many of these tests were reverse engineered based on the existing behaviour. We wanted Renderers to have close to 100% code coverage, especially this one (`Renderer`).  

I suspect some of these tests cover behaviour we no longer use ("force re-rendering" feels like unuseful and antithetical to the way React works). Perhaps we can just remove this behaviour (in a future PR)? Is it actually used?

packages/perseus/src/renderer.tsx
```diff
@@ -1741,57 +1712,36 @@ class Renderer
...
-        });
-        return userInputMap;
+        const userInput = this.props.userInput;
+        if (!userInput) {
+            throw new Error(
+                `getUserInputMap called without providing userInput to Renderer`,
+            );
+        }
+        return userInput;
     }
```
jeremywiebe, line 1742: Theoretically, these two getUserInput() functions could go away soon, right?  I mean, the `UserInputManager` now owns the user input data and so we should ask _it_ for the user input instead?

packages/perseus/src/server-item-renderer.tsx
```diff
@@ -324,10 +323,12 @@ export class ServerItemRenderer
             questionHighlightedWidgets: withRemoved,
         });
 
-        // Call the interactionCallback, if it exists, with the current user input data
-        this.props.apiOptions?.interactionCallback?.(
-            this.questionRenderer.getUserInputMap(),
-        );
+        setTimeout(() => {
```
jeremywiebe: Can you explain in a comment why this `setTimeout()` is needed?  My guess is something to do with consumers of this callback and a race... but I'd just be guessing.

packages/perseus/src/server-item-renderer.tsx
jeremywiebe: Also agree.  Would it make sense to document the idea state here in a TODO? Or maybe on the `UserInputManager` component?

packages/perseus/src/user-input-manager.test.tsx
```diff
@@ -0,0 +1,330 @@
...
+                                {userInput["number-line 1"].numLinePosition}
+                            </p>
+                        </>
+                    );
+                }}
+            </UserInputManager>,
+        );
+
+        // Start with nothing
+        expect(screen.getByText("User input is: -4")).toBeInTheDocument();
```
jeremywiebe, line 221: Took me a bit, but `-4` is the lowest point (initial state) because we have 10 divisions and the initial position is at the bottom?  So we go from -4 to +5?  Could you note that because the `-4` appears very magical at first glance. :)

packages/perseus/src/user-input-manager.test.tsx
```diff
@@ -0,0 +1,330 @@
...
+        act(() =>
+            restoreCallback?.(
+                generateExpressionSerializeState(),
+                generateExpressionWidgetsMap(),
+            ),
+        );
+        expect(screen.getByText(/Hello world/)).toBeInTheDocument();
+    });
+
+    // TODO
```
jeremywiebe: For a future PR?

packages/perseus/src/article-renderer.tsx
```diff
@@ -219,37 +220,59 @@ class ArticleRenderer
...
-                                        [sectionIndex].concat(newFocusPath),
-                                    oldFocusPath &&
-                                        [sectionIndex].concat(oldFocusPath),
-                                );
-                            },
-                        }}
-                        linterContext={PerseusLinter.pushContextStack(
-                            this.props.linterContext,
-                            "article",
+                    <UserInputManager widgets={section.widgets} problemNum={0}>
```
jeremywiebe, line 223: I guess this is simpler than a single `UserInputManager` for all sections?

packages/perseus/src/user-input-manager.tsx
```diff
@@ -0,0 +1,165 @@
...
+    problemNum: number,
+) => void;
+
+export type RestoreUserInputFromSerializedStateCallback = (
+    serializedState: unknown,
+    widgetOptions: PerseusWidgetsMap,
+) => void;
+
+export type HandleUserInputCallback = (
+    id: string,
```
jeremywiebe: **nit**: Could we name this `widgetId` to be extra clear what kind of ID we're talking about?

packages/perseus/src/user-input-manager.tsx
```diff
@@ -0,0 +1,165 @@
+import {entries} from "@khanacademy/wonder-stuff-core";
```
jeremywiebe: We don't need to use `entries` from WS here.  We can just use the native `Object.entries()`.

packages/perseus/src/user-input-manager.tsx
```diff
@@ -0,0 +1,165 @@
...
+    const [initialized, setInitialized] = useState<boolean>(false);
+    const [userInput, setUserInput] = useState<UserInputMap>({});
+    const prevProblemNum = useRef<number>(props.problemNum);
+
+    useEffect(() => {
+        if (prevProblemNum.current !== props.problemNum) {
+            prevProblemNum.current = props.problemNum;
+            setInitialized(false);
+        }
+    }, [props.problemNum]);
```
jeremywiebe: Mostly my ignorance, but if you just had the `props.problemNum` as a dependency here, doesn't that guarantee that the `useEffect()` will fire only when it's changed? Ie. is the `if` block needed?

packages/perseus/src/widgets/free-response/free-response.tsx
```diff
@@ -43,6 +43,10 @@ export class FreeResponse extends React.Component<Props> implements Widget {
     static contextType = PerseusI18nContext;
     declare context: React.ContextType<typeof PerseusI18nContext>;
 
+    // this just helps with TS weak typing when a Widget
+    // doesn't implement any Widget methods
+    isWidget = true as const;
```
jeremywiebe, line 48: Eventually I'd love to have an interface for base widgets and another for interactive widgets.  Non-interactive widgets may not need any interface as they just render (ie. they're a basic React component). 🤔  No action required here!

packages/perseus/src/widgets/graded-group/graded-group-answer-bar.tsx
jeremywiebe: I think some versions of `prettier` sometimes freaked out on these types of comments.  

Thanks for cleaning them up!

packages/perseus/src/widgets/graded-group/graded-group.test.ts
```diff
@@ -282,13 +282,8 @@ describe("graded-group", () => {
...
             act(() => jest.runOnlyPendingTimers());
 
-            await checkAnswer(userEvent);
-
             // Act
-            await userEvent.click(
-                screen.getAllByRole("button", {name: "False"})[2],
-            );
-            act(() => jest.runOnlyPendingTimers());
+            await checkAnswer(userEvent);
```
jeremywiebe, line 288: The removal of the second round of answering here makes this test name incorrect. The point was to ensure that you could hit `Check Answer`, get it wrong, fix it up, and check again (and potentially get it right).  Or does the button switch to "Try again" now without needing to interact with any widgets?

packages/perseus/src/widgets/graded-group/graded-group.tsx
```diff
@@ -321,10 +317,16 @@ export class GradedGroup
...
+                    handleUserInput={(widgetId, userInput) => {
+                        this.props.handleUserInput({
+                            ...this.props.userInput,
+                            [widgetId]: userInput,
+                        });
+                    }}
                     ref={this.rendererRef}
                     apiOptions={{...apiOptions, readOnly}}
                     showSolutions={showSolutions}
-                    onInteractWithWidget={this._onInteractWithWidget}
```
jeremywiebe, line 327: So with all these changes... are we moving to a place where `onInteractWithWidget` could be implemented fully in the `UserInputManager`?

packages/perseus/src/widgets/graded-group/graded-group.tsx
jeremywiebe: I think that feels right, yeah.

packages/perseus/src/widgets/graded-group/graded-group.tsx
```diff
@@ -386,18 +388,45 @@ export class GradedGroup
...
+                                        <Renderer
+                                            {...this.props.hint}
+                                            userInput={userInput}
+                                            handleUserInput={handleUserInput}
+                                            initializeUserInput={
+                                                initializeUserInput
+                                            }
+                                            restoreUserInputFromSerializedState={
+                                                restoreUserInputFromSerializedState
+                                            }
```
jeremywiebe: I don't know that we'd even care to support serializing and restoring state... but I suspect there's no harm in just passing these down because you've isolated hint state here.

packages/perseus/src/widgets/graded-group/graded-group.tsx
```diff
@@ -386,18 +388,45 @@ export class GradedGroup
...
+                                }) => {
+                                    {
+                                        /**
+                                         * We're passing a couple of props to Renderer that it doesn't
+                                         * require as part of {...this.props.hint}.
+                                         */
+                                    }
+                                    return (
+                                        <Renderer
+                                            {...this.props.hint}
```
jeremywiebe: I wonder if we could pass the parts of `hint` to `Renderer` explicitly instead of spreading here? That'd also allow cleaning up the vagueness of the comment above us here. Not required for this PR, but could be a nice cleanup.

packages/perseus/src/widgets/group/group.test.tsx
jeremywiebe: You've mentioned your preference for using `.toEqual` over `.toMatchInlineSnapshot`.  Maybe a good move here?

packages/perseus/src/widgets/interaction/__snapshots__/interaction.test.ts.snap
jeremywiebe: These snapshot tests (especially on legacy/unmaintained widgets) were mostly a very course way of ensuring we didn't break things. We also may have been _missing_ a render when we captured this snapshot. 🤔   I don't think there's much value in spending time on these widget's tests unless we can verify they're broken.

packages/perseus/src/widgets/group/group.tsx
```diff
@@ -181,6 +166,13 @@ class Group extends React.Component<Props> implements Widget {
...
                 {problemNumComponent}
                 <Renderer
+                    userInput={this.props.userInput}
+                    handleUserInput={(widgetId, userInput) => {
+                        this.props.handleUserInput({
+                            ...this.props.userInput,
+                            [widgetId]: userInput,
+                        });
+                    }}
                     content={this.props.content}
```
jeremywiebe, line 176: So the `group`'s user input is a map of `widgetId` to `widgetUserInput` (per the widgets _in_ the group)?  And then at a higher level, that map of user input will be nested within _this_ groups userInput object.  Just thinking through the way user input is managed.

handeyeco: Ideally what I'm going for is if there's this hierarchy:

```
radio 1
group 1
    number-line 1
```

User input should look like:

```
{
    "radio 1": {},
    "group 1": {
        "number-line 1": {}
    }
}
```

packages/perseus/src/widgets/graded-group/graded-group.tsx
handeyeco: I don't think I'd want UserInputManager calling `onInteractWithWidget` but maybe something more generic like it could have its own `onUserInput` and the parent of UserInputManager could call `onInteractWithWidget` in the `onUserInput` callback. I like that UserInputManager has a very specific purpose and I wouldn't want it to become just another Renderer layer.

packages/perseus/src/widgets/graded-group/graded-group.test.ts
handeyeco: I'm having a hard time remembering but, looking at the code again, I think this was a test banking on stale state.

The original test was:

1. Render question
2. Click two buttons in the Categorizer to get an incorrect answer
3. Check answer
4. Click another button on the Categorizer
5. Assert that the "Try again" button was visible

But that's incorrect. By changing the user input "Try again" should be replaced with "Check". This is from prod:

https://github.com/user-attachments/assets/00be339d-4470-495a-9ee7-1ea22e962a07

So I'm going to change the test:

1. Render question
2. Click two buttons in the Categorizer to get an incorrect answer
3. Check answer
4. Assert that the "Try again" button is visible
5. Click another button on the Categorizer
6. Assert that the "Check" button is visible

packages/perseus/src/widget-ai-utils/matcher/matcher-ai-utils.test.tsx
handeyeco: I don't love changing these tests because we told TUT that they would be owned by them. But the test did seem a little wrong anyway...so I changed it...

packages/perseus/src/user-input-manager.tsx
benchristel: I think the problem is that the callback implementation will need to do something like this:

```ts
const [userInputMap, setUserInputMap] = React.useState({});
// ...

function handleUserInput(id: string, userInput: UserInputMap[keyof UserInputMap]) {
  setUserInputMap({
    ...userInputMap,
    [id]: userInput,
    // ^ type error: 'string' is not a valid key of 'UserInputMap'
  });
}
```

And if you change the type of ID to `keyof UserInputMap`, the error will be something like "user input of widget A is not assignable to user input of widget B" because there's nothing correlating the type of the ID with the type of the user input.

I actually tried this in the code and saw the errors.

All that said, I think this is very much a two-way door.

benchristel: The first fix I suggested didn't work either, so I think it would have to be:

```
function handleUserInput(updates: UserInputMap): void`
```

packages/perseus/src/widget-ai-utils/graded-group/graded-group-ai-utils.test.ts
handeyeco: I don't remember why I did that. Changed it back.

packages/perseus/src/server-item-renderer.tsx
handeyeco: Tweaked and commented

packages/perseus/src/user-input-manager.test.tsx
handeyeco: Nope, just forgot to do it for this one

packages/perseus-core/src/validation.types.ts
handeyeco: So an ArticleRenderer wouldn't have UserInputManager, the individual GradedGroup/GradedGroupSet would have it? I guess not even GradedGroupSet, just GradedGroup since GradedGroupSet is just a wrapper around GradedGroup. I'll give it a try.

packages/perseus/src/article-renderer.tsx
handeyeco: I think this is related to Ben's comment: https://github.com/Khan/perseus/pull/2678#discussion_r2193459717

I'll see if I can mess with it.

packages/perseus/src/user-input-manager.tsx
benchristel: Good point about RenderProps. TBH, I think we should change that name (or get rid of the whole transform concept that necessitates it) but that's too much to consider at this point.

I'd prefer `FunctionChildParams` or `UserInputManagerFunctionChildParams` — using the unofficial (but still widespread) term.

handeyeco: We should be able to get really close to removing the concept of RenderProps from widgets with this: https://khanacademy.atlassian.net/browse/LEMS-3199

packages/perseus/src/article-renderer.tsx
handeyeco: Now GradedGroup manages its own user input

packages/perseus-core/src/validation.types.ts
handeyeco: Removed

packages/perseus/src/renderer.tsx
handeyeco: So I guess I have 3-ish options:

1. We pass in everything we need to detect a new problem directly into UserInputManager and UserInputManager watches for changes to the data to determine when to init user input. I don't really want to do this because I want UserInputManager to be really clean and focused. I already messed this by passing `widgets` and `problemNum` into UserInputManager.
2. I can have ServerItemRenderer watch for a problem change. One way of doing this is SIR detects a problem change and calls an imperative API on UserInputManager, but I didn't want to add an imperative API. I could also do some kind of edge detection: a `initializeUserInput: boolean` prop to UIM. Okay maybe.
3. Do this in Renderer where we were already watching for problem changes. It's the conductor - taking in data and callbacks and coordinating. This is what I did.

In my mind this is the direction we would want to go in longer-term. Because I think it makes sense to hoist user input state up, eventually we'll want this logic in the Chrome. The Chrome has way more awareness about when we change problems, so it can just call `initializeUserInput` when it's time to reset user input.

packages/perseus/src/renderer.tsx
handeyeco: Yeah they could. I didn't want to add an imperative API to UserInputManager and Renderer already has one so it didn't seem like a loss. Ideally when we can hoist user input state, Chrome won't need to ask Perseus at all for user input it'll just have it.

SonicScrewdriver:
This looks good to me from a radio project standpoint!

packages/perseus/src/__tests__/renderer.test.tsx
```diff
@@ -1756,3 +1687,140 @@ describe("renderer", () => {
...
+                        ],
+                    },
+                },
+            },
+            problemNum: 0,
+        };
+        expect(isDifferentQuestion(answerful, answerless as any)).toBe(false);
+    });
+
+    it("considers answerful/answerless to be the same", () => {
```
jeremywiebe: This test and the previous test seem to have their names mixed up.  This one doesn't seem to be comparing answerful/answerless data.

packages/perseus/src/__tests__/server-item-renderer.test.tsx
```diff
@@ -1,3 +1,4 @@
+import {type PerseusItem} from "@khanacademy/perseus-core";
```
jeremywiebe: So weird. Our ESLint setup should be putting these down with the `type` imports and so I'm surprised at this change.

packages/perseus/src/widgets/graded-group/graded-group.test.ts
```diff
@@ -329,4 +335,82 @@ describe("graded-group", () => {
...
+
+            // select an option
+            const dropdown = screen.getByRole("combobox");
+            await userEvent.click(dropdown);
+            await userEvent.click(screen.getByText("Right answer"));
+
+            // make sure the option was selected properly
+            expect(dropdown).toHaveTextContent("Right answer");
+
+            await userEvent.click(screen.getByRole("button", {name: "Check"}));
```
jeremywiebe: FWIW: not sure this adds clarity, or not, but I think you could use the `checkAnswer()` function here too.

jeremywiebe (approved):
I think we're there!! 😁   I had a few small comments but nothing blocking.

PR #2679, opened 2025-07-03
I think this is where the bug comes from

jeanettehead:
## Summary:
<Add your summary here>

Issue: LEMS-XXXX

## Test plan:

### DIFF
--- a/packages/perseus-editor/src/editor.tsx
+++ b/packages/perseus-editor/src/editor.tsx
@@ -200,13 +200,13 @@ class Editor extends React.Component<Props, State> {
             .on("paste", this._maybePasteWidgets);
     }
 
-    // TODO(arun): This is a deprecated method, use the appropriate replacement
-    // eslint-disable-next-line react/no-unsafe
-    UNSAFE_componentWillReceiveProps(nextProps: Props) {
-        if (this.props.content !== nextProps.content) {
-            this.setState({textAreaValue: nextProps.content});
-        }
-    }
+    // // TODO(arun): This is a deprecated method, use the appropriate replacement
+    // // eslint-disable-next-line react/no-unsafe
+    // UNSAFE_componentWillReceiveProps(nextProps: Props) {
+    //     if (this.props.content !== nextProps.content) {
+    //         this.setState({textAreaValue: nextProps.content});
+    //     }
+    // }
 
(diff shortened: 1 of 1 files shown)


packages/perseus-editor/src/editor.tsx
```diff
@@ -200,14 +200,6 @@ class Editor extends React.Component<Props, State> {
...
     }
 
-    // TODO(arun): This is a deprecated method, use the appropriate replacement
-    // eslint-disable-next-line react/no-unsafe
-    UNSAFE_componentWillReceiveProps(nextProps: Props) {
-        if (this.props.content !== nextProps.content) {
-            this.setState({textAreaValue: nextProps.content});
-        }
-    }
-
```
jeanettehead, line 210: Here's the bug - a debounce was implemented for performance reasons. 

Here's the debounce code which is on line 400-something of this file

```
        this.deferredChange = setTimeout(() => {
            if (this.state.textAreaValue !== this.props.content) {
                this.props.onChange({content: this.state.textAreaValue});
            }
        }, this.props.apiOptions.editorChangeDelay);
    };
```

So there's a race condition

1. Edit 1 is made
2. Edit 1 is queued up for a save
3. Edit 2 quickly follows
4. Edit 1 saves
5. The content prop updates and state is updated with edit 1's content
6. Edit 2's save is called, but the state has been updated with 1's content  
7. Edit 2 is lost, the textarea doesn't know where its cursor belongs anymore, puts it at the end.

I fixed this by removing the state update in the willReceiveProps function.  I believe that this widget only needs to know its starting value from its parent, and the it can manage its own changes from then on out. It's not an uncontrolled component, but similar behavior to one. I verified in testing that things like adding a widget, image, etc still update as expected.

Other ways to fix this
1. store the value outside of state in the debounced function so that it is never updated externally
2. only update state if there isn't a timeout in flux

I picked this approach because it has the added bonus of removing a deprecated react lifecycle method.

jeanettehead:
Here's a frontend znd: https://prod-znd-250707-1527-89e618c.khanacademy.org/

Myranae (approved):
Did not experience the jump in the ZND while I did in the prod page I tested! The only thing I'm not sure of is if there's any other aspect of editor usage that should be tested, but the jump is gone. :) Thanks!

PR #2681, opened 2025-07-04
[Interactive Graph] [Polygon] Fix concave polygon vertex labels jumping between interior and exterior angles

nishasy:
## Summary:
There is currently an issue where a concave polygon will have its concave vertex jump between
its internal and external angle, only if the other vertices around it line up in the same
x and y values as each other. This is happening because of the raytracing method being
used to determine the angles, which has the wrong value when the rays intersect.

To fix this, I'm replacing the raytracing method with the cross product method to determine
if a particular vertex in the polygon is convex or concave.

Issue: https://khanacademy.atlassian.net/browse/LEMS-3262

| Before | After | 
| --- | --- |
| image | image |

## Test plan:
`pnpm jest packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.test.ts`
`pnpm jest packages/perseus/src/widgets/interactive-graphs/graphs/polygon.test.tsx`

Storybook
- http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-polygon
- The starting coordinates are already set to the case that was having problems
- Confirm that the angles are all < 180 degrees except the convex vertex
- The convex vertex should be > 180 degrees
- Use the starting coordinate to reverse the coords (counter-clockwise instead of clockwise)
- Confirm the angles are the same as before

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.test.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.test.ts
@@ -121,18 +121,18 @@ describe("shouldDrawArcOutsidePolygon", () => {
 
     // Test each point in this chevron shaped polygon with clockwise points.
     it.each([
-        {index: 0, expected: false}, // inside
-        {index: 1, expected: false}, // inside
-        {index: 2, expected: false}, // inside
-        {index: 3, expected: false}, // inside
-        {index: 4, expected: false}, // inside
-        {index: 5, expected: true}, // outside
+        {index: 0, isOutside: false},
+        {index: 1, isOutside: false},
+        {index: 2, isOutside: false},
+        {index: 3, isOutside: false},
+        {index: 4, isOutside: false},
+        {index: 5, isOutside: true},
     ] satisfies {
         index: number;
-        expected: boolean;
+        isOutside: boolean;
     }[])(
-        "should return $expected for concave clockwise polygon vertex $index",
-        ({index, expected}) => {
+        "should return $isOutside for concave clockwise polygon vertex $index",
+        ({index, isOutside}) => {
             // Get the previous and next vertices.
             const previousIndex =
                 (index - 1 + clockwiseConcaveCoords.length) %
@@ -147,7 +147,7 @@ describe("shouldDrawArcOutsidePolygon", () => {
             ] satisfies [vec.Vector2, vec.Vector2];
 
             expect(shouldDrawArcOutsidePolygon(vertex, endPoints)).toBe(
-                expected,
+                isOutside,
             );
         },
     );
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.test.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.test.ts
@@ -1,62 +1,99 @@
 import {angles} from "@khanacademy/kmath";
 
-import {shouldDrawArcOutside} from "./angle-indicators";
+import {
+    shouldDrawArcOutside,
+    shouldDrawArcOutsidePolygon,
+} from "./angle-indicators";
 
-import type {Coord} from "@khanacademy/perseus-core";
-import type {vec} from "mafs";
+import type {Coord, CollinearTuple} from "@khanacademy/perseus-core";
+import type {vec, Interval} from "mafs";
 
 const {getClockwiseAngle} = angles;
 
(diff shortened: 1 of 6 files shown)


packages/perseus/src/widgets/interactive-graphs/interactive-graph.testdata.ts
```diff
@@ -185,15 +185,20 @@ export const unlimitedPolygonQuestion: PerseusRenderer =
...
 export const polygonWithStartingCoordsQuestion: PerseusRenderer =
     interactiveGraphQuestionBuilder()
+        .withMarkings("grid")
         .withPolygon("grid", {
             startCoords: [
-                [6, 6],
-                [8, 6],
-                [8, 8],
-                [6, 8],
+                [-7, 5],
```
catandthemachines, line 194: Why changes the coordinates of this test? Do we want a test for clockwise and one for counter clockwise?

catandthemachines (approved):
Had a question on some of the tests. But everything else looks good to me.

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.test.ts
```diff
@@ -105,3 +108,47 @@ describe("shouldDrawArcOutside", () => {
...
+    it.each([
+        {index: 0, expected: false}, // inside
+        {index: 1, expected: false}, // inside
+        {index: 2, expected: false}, // inside
+        {index: 3, expected: false}, // inside
+        {index: 4, expected: false}, // inside
+        {index: 5, expected: true}, // outside
+    ] satisfies {
+        index: number;
+        expected: boolean;
```
jeremywiebe: Could you just name this field `isOutside` instead to avoid the need for all the `inside`/`outside` comments? :)

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.tsx
```diff
@@ -20,25 +20,22 @@ const {getAngleFromVertex} = angles;
 interface PolygonAngleProps {
     centerPoint: vec.Vector2;
     endPoints: [vec.Vector2, vec.Vector2];
-    polygonLines: readonly CollinearTuple[];
-    range: [Interval, Interval];
     showAngles: boolean;
     snapTo: SnapTo;
+    areClockwise: boolean;
```
jeremywiebe: **nit**: I'd place this below `endPoints` since it is closely related to them. I'd also consider naming this `areEndpointsClockwise` to more clearly denote the relationship.  Without it, `areClockwise` feels very vague.

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.tsx
```diff
@@ -80,13 +77,8 @@ export const PolygonAngle = ({
...
 
-    // Midpoint betwen ends of arc
-    const isOutside = shouldDrawArcOutside(
-        [x3, y3],
-        centerPoint,
-        range,
-        polygonLines,
-    );
+    // Note: Need to pass in endpoints in a clockwise order for the cross product.
+    const isOutside = shouldDrawArcOutsidePolygon(centerPoint, endPoints);
```
jeremywiebe, line 81: Just want to double-check if other uses of `shouldDrawArcOutside` should use this new method as well or if the ray-tracing method is still best for them?

jeremywiebe (approved):

packages/perseus/src/widgets/interactive-graphs/interactive-graph.testdata.ts
nishasy: Oh this is just the testdata used in storybook. I just wanted to be able to see the scenario in question, and there wasn't really any significance to the previous starting coords.

packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.tsx
nishasy: I did try using this for Angle, which is the only other place it's used. It broke the entire angle, so I reverted it back to what it was before 😅

nishasy: Here's the screenshot of the angle being broken

image

jeremywiebe: Ok, so we need both methods. Thanks for checking at least.

PR #2685, opened 2025-07-07
[BUGFIX] Fix Perseus templates

daniellewhyte:
## Summary:
Right now Perseus templates are broken! You cannot add any templates in the Perseus editor because of incorrect logic when the `additionalTemplates` prop was introduced to the EditorPage component. 
See: #2580

Issue: XXX-XXXX

## Test plan:
Verify that templates can be added in the EditorPage story:
http://localhost:6006/?path=/story/perseuseditor-editorpage--demo

### DIFF
--- a/packages/perseus-editor/src/editor.tsx
+++ b/packages/perseus-editor/src/editor.tsx
@@ -995,21 +995,24 @@ class Editor extends React.Component<Props, State> {
 
             const insertTemplateString = "Insert template\u2026";
             templatesDropDown = (
-                <select onChange={this.addTemplate}>
+                <select
+                    onChange={this.addTemplate}
+                    data-testid="editor__template-select"
+                >
                     <option value="">{insertTemplateString}</option>
                     <option disabled>--</option>
                     <option value="table">Table</option>
                     <option value="titledTable">Titled table</option>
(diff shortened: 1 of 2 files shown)


packages/perseus-editor/src/editor.tsx
```diff
@@ -1003,12 +1002,14 @@ class Editor extends React.Component<Props, State> {
...
                     {Object.entries(this.props.additionalTemplates).map(
                         ([key]) => (
-                            <option value={key} key={key}>
-                                {key}
-                            </option>
+                            <>
+                                <option disabled>--</option>
+                                <option value={key} key={key}>
+                                    {key}
+                                </option>
```
handeyeco: This doesn't seem right. For every option we also want a disabled option?

.changeset/purple-dryers-fly.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": patch
+---
+
+Fix templates in the ItemEditor
```
handeyeco, line 5: If the logic wasn't working it makes me think we don't have proper test coverage for the feature. Could we add that?

PR #2687, opened 2025-07-08
[Radio] Refactor multiple choice indicator (logic and UI updates)

mark-fitzgerald:
## Summary:

Issue: LEMS-3173

## Test plan:


config/test/test.config.js
```diff
@@ -54,6 +54,7 @@ module.exports = {
         // Compile .svg files using a custom transformer that returns the
         // basename of the file being transformed.
         "^.+.svg$": "<rootDir>/config/test/svg.transform.js",
+        "^.+\\.module\\.css$": "jest-css-modules-transform",
```
mark-fitzgerald, line 59: This is needed to pull in the class names from CSS Modules files. Otherwise, they just show up as `undefined`, and prevents unit testing.

config/test/test.config.js
```diff
@@ -75,7 +76,7 @@ module.exports = {
         ...pkgMap,
         ...vendorMap,
         // Load a .js file with no exports whenever a .css file is requested.
-        "\\.(css)$": "<rootDir>/config/test/style-mock.js",
+        "(?<!\\.module)\\.css$": "<rootDir>/config/test/style-mock.js",
```
mark-fitzgerald, line 81: I adjusted this entry to only be applied on regular CSS files, but to not apply to CSS Module files (otherwise, it would conflict with the transform instructions on line 57).

packages/perseus/src/widgets/radio/__tests__/__snapshots__/multiple-choice.test.ts.snap
mark-fitzgerald, line 1: Snapshots that used components that referenced CSS Module files were updated due to the Jest configuration change.

packages/perseus/src/widgets/radio/__stories__/indicator.stories.tsx
```diff
@@ -0,0 +1,153 @@
...
+                marginBlockEnd: "25px",
+            }}
+        >
+            {props.children}
+        </div>
+    );
+};
+
+export default {
+    title: "Perseus/Widgets/Radio/Indcator",
```
ivyolamit: No need to change in this PR. Just a heads up on our proposal for the new documentation any internal components to the widget will be under
```suggestion
    title: "Perseus/Widgets/Radio/Widget Internal Components/Indicator",
```

packages/perseus/src/widgets/radio/choice-indicator.module.css
```diff
@@ -0,0 +1,104 @@
+.base {
+    appearance: none;
+    background-color: var(--wb-semanticColor-core-transparent);
+    border: solid 0.2rem var(--wb-semanticColor-core-border-neutral-strong);
```
ivyolamit: it's unfortunate that the border values in WB is still not very detailed e.g. like we can use the `--wb-border-width-medium` but the other option we have based on the docs is using the sizing as mentioned in the [docs](https://khan.github.io/wonder-blocks/?path=/docs/packages-tokens-sizing--docs) it can be used for border width.
```suggestion
    border: solid var(--wb-sizing-size_020) var(--wb-semanticColor-core-border-neutral-strong);
```

packages/perseus/src/widgets/radio/choice-indicator.module.css
```diff
@@ -0,0 +1,104 @@
...
+    color: var(--wb-semanticColor-core-foreground-neutral-default);
+    cursor: pointer;
+    font-family: var(--wb-font-family-sans);
```
ivyolamit, line 9: is this something we want to define here? and not in global?

packages/perseus/src/widgets/radio/choice-indicator.module.css
```diff
@@ -0,0 +1,104 @@
...
+    font-size: var(--wb-font-heading-size-small);
+    height: var(--wb-c-button-root-sizing-height-small);
+    line-height: var(--wb-font-heading-lineHeight-small);
+    outline: 0 solid var(--wb-semanticColor-core-background-instructive-default);
```
ivyolamit, line 13: not sure what our guidance on values with `0` but WB has the token `--wb-sizing-size_0`, do we want to use it here?

packages/perseus/src/widgets/radio/choice-indicator.module.css
```diff
@@ -0,0 +1,104 @@
...
+    outline-offset: 0;
+    padding: 0;
```
ivyolamit, line 15:
```suggestion
    padding: var(--wb-sizing-size_0);
```

packages/perseus/src/widgets/radio/choice-indicator.module.css
```diff
@@ -0,0 +1,104 @@
...
+}
+
+@media (prefers-reduced-motion: no-preference) {
+    .base {
+        transition: all 0.2s;
+    }
+}
+
+.base:hover,
+.base:focus {
```
ivyolamit: do we want to use the WB tokens here?
```
    outline-offset: var(--wb-sizing-size_020);
    outline-width: var(--wb-sizing-size_020);
```

packages/perseus/src/widgets/radio/choice-indicator.module.css
```diff
@@ -0,0 +1,104 @@
...
+    outline-offset: 0.2rem;
+    outline-width: 0.2rem;
+}
+
+/* button is checked */
+.base[aria-pressed="true"] {
+    background-color: var(
+        --wb-semanticColor-core-background-instructive-default
+    );
+    border-width: 0;
```
ivyolamit, line 47: same comment with `0` above

packages/perseus/src/widgets/radio/choice-indicator.module.css
```diff
@@ -0,0 +1,104 @@
...
+    color: var(--wb-semanticColor-core-foreground-inverse-strong);
+}
+
+/* icon inside the button when showing correctness */
+.base [role="img"] {
+    margin-inline-end: 0.4rem;
```
ivyolamit:
```suggestion
    margin-inline-end: var(--wb-sizing-size_040);
```

packages/perseus/src/widgets/radio/choice-indicator.module.css
```diff
@@ -0,0 +1,104 @@
...
+           in order to make the measurement relative to the size of the button.
+       In the case where the single-select indicator has multiple elements in its content (like a checkmark),
+           the shape changes from a circle to a gel-capsule.
+       This property setting automatically handles that case without the need for JS logic.
+    */
+    border-radius: 100vh;
+}
+
+.square-shape {
+    border-radius: 0.4rem;
```
ivyolamit:
```suggestion
    border-radius: var(--wb-sizing-size_040);
```

ivyolamit (commented):
@mark-fitzgerald here's my first round of review comments. I'll do a second round shortly after my meetings 🙏🏼

packages/perseus/src/widgets/radio/choice-indicator.module.css
ivyolamit: Update: Based on discussion with WB team this morning for border we'll use px and border token:
https://khan.github.io/wonder-blocks/?path=/docs/packages-tokens-border--docs

packages/perseus/src/widgets/radio/choice-indicator.new.tsx
```diff
@@ -0,0 +1,57 @@
...
+import minusIcon from "@phosphor-icons/core/bold/minus-circle-bold.svg";
+import * as React from "react";
+import {useState} from "react";
+
+import styles from "./choice-indicator.module.css";
+
+export type IndicatorProps = {
+    checked: boolean;
+    content: string;
+    shape: "circle" | "square";
```
ivyolamit, line 13: In naming, do we want to associate the circle to be single select and square for multiple select?

packages/perseus/src/widgets/radio/choice-indicator.new.tsx
```diff
@@ -0,0 +1,57 @@
...
+            ? checkIcon
+            : isChecked && showCorrectness === "wrong"
+              ? minusIcon
+              : undefined;
+    const icon = iconImage ? (
+        <PhosphorIcon
+            aria-hidden={true}
+            icon={iconImage}
+            role="img"
+            style={{width: "1.4rem", height: "1.4rem"}}
```
ivyolamit: what's our guidance on inline styles? this is something I've seen in our codebase that's not consistent some code uses inline styling while some uses variables.

This is only personal preference, either use variables or css class, it's easy to get out of hand on inline styling that make the code long which makes it hard to read. But whatever our team guidance on this, I can follow moving forward.

packages/perseus/src/widgets/radio/choice-indicator.module.css
mark-fitzgerald: Preferably, this would be set somewhere higher up. I'm not sure that we can rely upon a proper ancestor setting, though. Probably something that I need to check with the team.

packages/perseus/src/widgets/radio/choice-indicator.module.css
mark-fitzgerald: I think that this is an artifact from the auto-generated variables from the JS tokens. I don't see any value in having a variable that always returns a zero.

packages/perseus/src/widgets/radio/choice-indicator.module.css
mark-fitzgerald: Changing to 1/3 of an `em`. The space between the icon and the choice letter should scale visually with the `font-size`. Even though the checkbox is an icon, it is acting like text (i.e. similar to unicode icons). The space between the icon and the letter should be proportional to their size. By setting the margin to an `em` measurement, we don't have to worry about  adjusting this space if we change the font size of the letter, and it will scale with any user-based font-size changes.

packages/perseus/src/widgets/radio/choice-indicator.new.tsx
mark-fitzgerald: The indicator isn't doing any logic based upon whether the widget is single or multiple select. Its only job is to display an indicator in a designated shape. I suppose an argument could be made that it should "display as single-select" or "display as multi-select", and so naming should follow. Let me know if anyone has strong feelings on this matter.

config/test/test.config.js
SonicScrewdriver: I wonder if this would be a good comment to add for our future selves

packages/perseus/src/widgets/radio/choice-indicator.module.css
ivyolamit: what do you think of adding todo comment here for now so we won't forget that it would be preferable to have this on higher level instead of here?

packages/perseus/src/widgets/radio/choice-indicator.module.css
SonicScrewdriver: I'm honestly not sure there's a lot of value in a size 0 either as it doesn't seem like it would scale or change at all between themes. If anything, it makes it harder to tell that this is zero'd out - so I kind of prefer avoiding the tokens here.

packages/perseus/src/widgets/radio/choice-indicator.module.css
SonicScrewdriver: It makes sense to declare this higher up, but it is probably easier to do a full css pass as we integrate the new design, and after we have the individual components all hooked up. We have a lot of very specific css rules floating around in Perseus that it might be simpler to do it holistically

packages/perseus/src/widgets/radio/__tests__/multiple-choice-indicator.test.tsx
```diff
@@ -0,0 +1,170 @@
+import {describe, it} from "@jest/globals";
```
ivyolamit: are we renaming this file to `choice-indicator.test.tsx` ?

packages/perseus/src/widgets/radio/choice-indicator.module.css
```diff
@@ -0,0 +1,129 @@
...
+        --wb-semanticColor-core-background-instructive-default
+    );
+    border-width: 0;
+    color: var(--wb-semanticColor-core-foreground-inverse-strong);
+}
+
+.is-correct[aria-pressed="true"] {
+    background-color: var(--wb-semanticColor-core-background-success-strong);
+    color: var(
+        --wb-semanticColor-action-primary-progressive-default-foreground
```
SonicScrewdriver, line 89: One thing that I'm not the _biggest_ fan of is how long these tokens can get.

packages/perseus/src/widgets/radio/choice-indicator.module.css
```diff
@@ -0,0 +1,129 @@
...
+    );
+    width: calc(
+        1.5em + var(--wb-c-button-root-sizing-height-small)
+    ); /* adding 1.5em accounts for the checkmark icon */
```
SonicScrewdriver, line 93: Greatly appreciate the explanation here!

packages/perseus/src/widgets/radio/choice-indicator.module.css
```diff
@@ -0,0 +1,129 @@
...
+    appearance: none;
+    background-color: var(--wb-semanticColor-core-transparent);
+    border: solid var(--wb-border-width-medium)
+        var(--wb-semanticColor-core-border-neutral-strong);
+    color: var(--wb-semanticColor-core-foreground-neutral-default);
+    cursor: pointer;
+    flex-shrink: 0;
+    font-family: var(--wb-font-family-sans);
+    font-size: var(--wb-font-heading-size-small);
+    height: var(--wb-c-button-root-sizing-height-small);
```
ivyolamit: i can't seem to find this in the docs, is this something available for us to use?

SonicScrewdriver (approved):
This looks great to me! Thanks for the amazing comments and explanations for the unique CSS applications

packages/perseus/src/widgets/radio/choice-indicator.module.css
```diff
@@ -0,0 +1,129 @@
...
+
+/* icon inside the button when showing correctness */
+.icon {
+    /* The size of the icon, and the space between the icon and the choice letter
+           should scale visually with the font-size.
+       Even though the checkbox is an icon, it is acting like text (i.e. similar to unicode icons).
+       If we change the font size of the letter, we want the icon to scale proportionally.
+       The space between the icon and the letter should also be proportional to their size.
+       By setting the width, height, and margin to an em unit,
+           we don't have to worry about adjusting these if we change the font size of the letter,
```
ivyolamit, line 59: this is just a nitpick, please properly aligning any multi-line comments

ivyolamit (approved):
@mark-fitzgerald overall PR looks good. I have few comments, but i'll leave it up to you.
And also i think one of the file was missed in renaming `multiple-choice-indicator.test.tsx` --> `choice-indicator.test.tsx`

packages/perseus/src/widgets/radio/choice-indicator.module.css
mark-fitzgerald: It's not in the docs. In fact, they cautioned against using it. However, they also didn't provide an alternative to our needs. I suppose setting our own size here would be more appropriate, since the size of this button isn't tied to anything in Wonder Blocks. I'll update accordingly.

ivyolamit: @mark-fitzgerald just saw this in 
```
node_modules/.pnpm/@khanacademy+wonder-blocks-button@10.2.11_@phosphor-icons+core@2.0.2_aphrodite@1.2.5_re_f6b95c8f1f4f35b85908aaae53936269/node_modules/@khanacademy/wonder-blocks-button/dist/css/vars.css
```

has `--wb-c-button-root-sizing-height-small: var(--wb-sizing-size_320);` ?

if we can't use `--wb-sizing-size_320` please add a comment on it so no-one will copy it

PR #2688, opened 2025-07-08
Add more documentation to parse-perseus-json/README.md

benchristel:
Issue: none

## Test plan:

Read the README. It should make sense.

### DIFF
--- a/packages/perseus-core/src/parse-perseus-json/README.md
+++ b/packages/perseus-core/src/parse-perseus-json/README.md
@@ -149,7 +149,8 @@ parser change might look like:
 
 Re-run the regression tests. The snapshots will fail, reflecting the fact that
 you added a new widget option. Update the snapshots (`pnpm test -u`), make sure
-they reflect the desired output, and commit them.
+they reflect the desired output (in this case, the default value of `average`
+should be set for `brewStrength`), and commit the changes.
 
 </details>
 


packages/perseus-core/src/parse-perseus-json/README.md
```diff
@@ -87,14 +87,66 @@ typetests to keep the parsers in sync with data-schema.
...
 were generated from a snapshot of Khan Academy content taken in November 2024.
 
+The regression tests are data-driven: we have a bunch of real Perseus items and
+articles taken from production, and we run those through the parser and make
+sure they all parse correctly. The data files live in
+`regression-tests/item-data` and `regression-tests/article-data`.
+
+When you add new widget fields or versions, please also add regression test
+data for the _current_ (i.e. soon-to-be-outdated) version of the widget. This
+will help ensure that future developers don't accidentally break your parser.
```
Myranae, line 104: I could see this as a good spot for adding an example. Maybe something showing a pretend field being added and what the regression test data would look like? I think the goal would be to show what the data would look like for a specific field addition (especially since you already point out where data files are). Possibly also to highlight what you emphasized twice (how it should be for the current/almost old version). Totally optional, but just felt an example would help for this bit.

packages/perseus-core/src/parse-perseus-json/README.md
```diff
@@ -87,14 +87,66 @@ typetests to keep the parsers in sync with data-schema.
...
+});
+```
+
+Sometimes, though, the migration needs to be more involved. For example, if we
+want to rename a field, we need some way to get the data from the old field and
+copy it to the new one. The tool for doing that is the `versionedWidgetOptions`
+function. For usage instructions, see the source code,
+`versioned-widget-options.ts`.
+
+## Changing existing parsers
```
Myranae, line 182: Really appreciate this section! Glad to know what's okay to change and why.

Myranae (approved):
I haven't read through the whole file yet, but reading through what you added makes sense! Thanks for adding more!

PR #2689, opened 2025-07-08
Mark `deprecated-standin` widget as accessible

jeremywiebe:
## Summary:

This PR adds a missing widget registration to `perseus-core` for the `deprecated-standin` widget. Technically, this widget does nothing, but our default for widgets that don't have an explicit registration is to be _inaccessible_. This means that we would consider assessment items having a deprecated widget as inaccessible when, in fact, they are (because the `deprecated-standin` has no interactive elements). 

This PR has some snapshot updates caused by use having an entry in the core registry for the deprecated-standin widget now (specifically because when we apply defaults for a widget, we leave the widget options unchanged [if no widget is registered](https://github.com/Khan/perseus/blob/843d66257ceb32b7c3d3eea4dea47dfda58e7945/packages/perseus-core/src/widgets/apply-defaults.ts#L33), but when we find a registration, we proceed to set some defaults like `static` and `version`).

Issue: LEMS-3096

## Test plan:

  * ✅ Run tests: `pnpm test`

### DIFF
--- a/.changeset/loud-cups-ring.md
+++ b/.changeset/loud-cups-ring.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-core": minor
+---
+
+Change deprecated-standin widget to be considered accessible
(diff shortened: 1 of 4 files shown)


jeremywiebe:
Added @handeyeco and @benchristel just because the snapshot updates relate to splitting and parsing.

handeyeco (approved):

catandthemachines (approved):
Thanks for fixing this gap. ❤️

PR #2690, opened 2025-07-08
[Simple Markdown] Remove deprecated APIs and features (html output)

jeremywiebe:
## Summary:

This PR removes long-deprecated features and APIs from the `@khanacademy/simple-markdown` project. We haven't used HTML output for a long, long time (everything is rendered to React) and so that is being removed. 

There are also a few APIs that are marked as deprecated that I've verified as unused, which we can delete.

This work is being done because I need to make another change and thought it would be handy to clean up the library a bit before doing any new work.

Note that the `output => html` is still used in tests (which is unrelated to this deprecated removal) to validate rendering to React, but we take the React and render it to HTML so we can assert compared to an HTML string. Eventually we could migrate that to comparing the React elements to JSX. I didn't do that in this PR to avoid too much churn. 

Issue: 

## Test plan:

`pnpm tsc`
`pnpm test`

Install the snapshot version of the package in the frontend repo and type check everything there.

### DIFF
--- a/packages/simple-markdown/src/__tests__/simple-markdown.test.tsx
+++ b/packages/simple-markdown/src/__tests__/simple-markdown.test.tsx
@@ -12,7 +12,6 @@ var inlineParse = SimpleMarkdown.defaultInlineParse;
 var blockParse = SimpleMarkdown.defaultBlockParse;
 var implicitParse = SimpleMarkdown.defaultImplicitParse;
 var defaultReactOutput = SimpleMarkdown.defaultReactOutput;
-var defaultHtmlOutput = SimpleMarkdown.defaultHtmlOutput;
 
 /**
  * Asset that two ast parse trees are equal
@@ -57,29 +56,10 @@ var htmlFromReactMarkdown = function (source) {
 
 /**
  * @param {string} source
- * @returns {string}
- */
-var htmlFromMarkdown = function (source) {
-    var html = defaultHtmlOutput(implicitParse(source));
-    var simplifiedHtml = html.replace(/\s+/g, " ");
-    return simplifiedHtml;
-};
-
-/**
- * @param {string} source
  * @param {string} html
  */
 var assertParsesToReact = function (source: string, html: string) {
     var actualHtml = htmlFromReactMarkdown(source);
-    assert.strictEqual(actualHtml, html);
-};
-
-/**
- * @param {string} source
- * @param {string} html
- */
-var assertParsesToHtml = function (source: string, html) {
-    var actualHtml = htmlFromMarkdown(source);
     assert.strictEqual(actualHtml, html);
 };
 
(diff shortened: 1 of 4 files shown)


packages/simple-markdown/src/__tests__/simple-markdown.test.tsx
```diff
@@ -3163,28 +3148,22 @@ describe("simple markdown", function () {
...
              */
             var validateAligns = function (
                 tableSrc,
-                expectedAligns: Array<null> | Array<string>,
+                expectedAligns: Array<TableAlignment>,
             ) {
                 var parsed = blockParse(tableSrc + "\n");
                 assert.strictEqual(parsed[0].type, "table");
                 var actualAligns = parsed[0].align;
-                validateParse(actualAligns, expectedAligns);
```
jeremywiebe, line 3156: `validateParse()` is really a glorified wrapper around an equality assertion, we were fighting the types here so I've just reverted to use... `expect()` here.

packages/simple-markdown/src/__tests__/simple-markdown.test.tsx
```diff
@@ -4005,45 +3985,44 @@ describe("simple markdown", function () {
...
-                                '<span class="' +
-                                (state.spanClass || "default") +
-                                '">' +
-                                /*SimpleMarkdown.sanitizeText*/ node.content +
-                                "</span>"
-                            );
+                            return SimpleMarkdown.reactElement("span", null, {
+                                className: state.spanClass || "default",
+                                children: node.content,
+                            });
```
handeyeco: Could these just be JSX? Since it's in a TSX file?


```suggestion
                            return <span className={state.spanClass || "default"}>
                                {node.content}
                            </span>
```

packages/simple-markdown/src/__tests__/simple-markdown.test.tsx
```diff
@@ -4005,45 +3985,44 @@ describe("simple markdown", function () {
...
                 },
-                "html",
+                "react",
             );
 
             var parsed1 = SimpleMarkdown.defaultBlockParse("hi there!");
             var result1 = output(parsed1, {spanClass: "special"});
             assert.strictEqual(
-                result1,
+                reactToHtml(result1),
```
handeyeco, line 4016: Is there not a way to do this kind of assertion without `reactToHtml`? Would this be a good use of snapshot testing?

packages/simple-markdown/src/__tests__/simple-markdown.test.tsx
```diff
@@ -4062,23 +4041,22 @@ describe("simple markdown", function () {
                     ) {
                         return {content: parse(capture[1])};
                     },
-                    html: function (
+                    react: function (
                         /** @type {SimpleMarkdown.SingleASTNode} */ node,
-                        /** @type {SimpleMarkdown.HtmlOutput} */ output,
+                        /** @type {SimpleMarkdown.ReactOutput} */ output,
```
handeyeco, line 4045: Is this Flow? Should it be converted to TS?

packages/simple-markdown/src/__tests__/simple-markdown.test.tsx
```diff
@@ -4430,301 +4408,6 @@ describe("simple markdown", function () {
         });
     });
 
-    describe("html output", function () {
```
handeyeco, line 4433: I was worried about losing tests, but it seems like we already had these copied between HTML and React 🎊 Yay!

packages/simple-markdown/src/index.ts
```diff
@@ -1736,8 +1519,8 @@ var ruleOutput = function <Rule>(
     // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
     if (!property && typeof console !== "undefined") {
         console.warn(
-            "simple-markdown ruleOutput should take 'react' or " +
-                "'html' as the second argument.",
+            "simple-markdown ruleOutput should take 'react' as the " +
+                "second argument.",
```
handeyeco, line 1523: If "react" is the only option, should we remove the argument altogether?

handeyeco (approved):
Love to see it!

packages/simple-markdown/src/__tests__/simple-markdown.test.tsx
jeremywiebe: There is. But this "html" is unrelated to the feature I'm removing and I didn't want to balloon this PR.  The way we resolve this is by switching to an assertion that is something like this: 

```
   expect(result1).toEqual(<p>...</p>); // use JSX
```

packages/simple-markdown/src/__tests__/simple-markdown.test.tsx
jeremywiebe: This is TypeScript in JSDoc comments.  Again, I don't want to touch things outside the focus of this PR, which is deleting the `html` output feature.  We _could_ change this to regular TypeScript types though.

packages/simple-markdown/src/__tests__/simple-markdown.test.tsx
jeremywiebe: Good idea, I tried that and it works!

packages/simple-markdown/src/index.ts
jeremywiebe: I like the idea, but I think I'll avoid removing it for now. It causes API changes which touch external uses and I want to minimize that for now. I'll leave a TODO that this could be removed.

Myranae (approved):
Super love the cleanup! Thanks so much!

PR #2692, opened 2025-07-09
[PEICH] Adding check to whether perseus markdown is accessible.

catandthemachines:
## Summary:
Updated our `isItemAccessible` function to include a check to confirm our markdown is accessible. Which ammounts to checking if markdown images include alt text. If it does not the item is considered inaccessible.

Issue: LEMS-3106

## Test plan:
Run `pnpm test` and all tests should return as passing.


packages/perseus-core/src/accessibility.test.ts
```diff
@@ -8,177 +8,342 @@ describe("isItemAccessible", () => {
         registerCoreWidgets();
     });
 
-    it("should return false if the item contains any inaccessible widgets", () => {
```
catandthemachines, line 11: Note: The diff here is being weird. All that was added is the widget specific tests were wrapped in a `describe` statement and I added a new section of tests for the markdown.

packages/perseus-core/src/accessibility.ts
```diff
@@ -3,6 +3,9 @@
  * or screen, based on the widgets it contains.
  */
 
+// eslint-disable-next-line @typescript-eslint/no-restricted-imports
+import SimpleMarkdown from "@khanacademy/simple-markdown";
```
catandthemachines, line 5: Need to look into how to navigate this restricted import. 🤔

packages/perseus-core/src/accessibility.test.ts
jeremywiebe: I've found in situations like this that the Hide Whitespace option helps alot! :) 

image

packages/perseus-core/src/accessibility.ts
```diff
@@ -42,5 +43,19 @@ export function violatingWidgets(itemData: PerseusItem): Array<string> {
  * any widgets that violate our accessibility requirements).
  */
 export function isItemAccessible(itemData: PerseusItem): boolean {
+    // Traverse the item question and check if markdown images have alt text.
+    // If it does note then the item is not accessible and we return false.
+    const nodes = SimpleMarkdown.defaultInlineParse(itemData.question.content);
+    for (const node of nodes) {
```
jeremywiebe, line 49: This won't catch all images because they might be nested in non-top-level nodes. Let's land this as it is and I'll fix it in the PR that I'm working on based on this branch. :)

catandthemachines: I actually found the opposite problem, the defaultBlockParse function created a lot of nested nodes and I found this to better flatten the nodes. But maybe there's an example I'm not thinking of. 🤔

jeremywiebe (approved):
Good stuff. Glad we're able to start doing much of these PEICH changes with the tools Perseus already has. 😁

tatianasnook (approved):
Thanks for the clear summary and update!
It’s great that this adds a check for image alt text in markdown to improve accessibility.

PR #2693, opened 2025-07-09
Delete the blocked-widget and blockquoted-math content linter rules

jeanettehead:
## Summary:
From Amanda Leff (content creator): When we can't indent definition widget it causes pedagogical issues for ELA and social studies - For example, means we can't appropriately scaffold text that needs to be indented, can't format some article content in the way that would be most accessible to students, and can't use some authentic texts that we plan to use in the courses b/c those texts use block quotes of other texts.

Issue: LEMS-3288

## Test plan:
1. in the content editor add a definition widget inside a blockquote - `> [☃️ definition 1]` (or reference the test everything [definition article](https://www.khanacademy.org/internal-courses/test-everything/test-everything-1/definition/a/definition-article))
2. observe that there is no linter warning about the widget being inside quoted text 
3. observe that the line does not break before and after the definition widget


jeremywiebe (approved):

catandthemachines:
Thank you for this fix!

PR #2694, opened 2025-07-10
[PEICH] Fix isAccessible() check to ignore unused widgets

jeremywiebe:
## Summary:

Until now, our `isAccessible` check considered _all_ widgets in the `PerseusItem`'s `widgets` dictionary. This caused many `PerseusItem`'s to be marked as inaccessible even though the violating widgets would never be seen/used (this is a long-standing issue with our content editors - they leave unused widgets in place when saving).

Issue: LEMS-3096

## Test plan:

`pnpm test`


packages/perseus-core/.eslintrc.js
```diff
@@ -16,7 +16,6 @@ module.exports = {
             "@khanacademy/perseus-editor",
             "@khanacademy/perseus-linter",
             "@khanacademy/perseus-score",
-            "@khanacademy/pure-markdown",
```
jeremywiebe, line 19: pure-markdown doesn't depend on any of our perseus packages so this is a safe exception (ie. it doesn't cause a circular dep).

packages/perseus-core/package.json
```diff
@@ -27,7 +27,7 @@
     "dependencies": {
         "@khanacademy/kas": "workspace:*",
         "@khanacademy/perseus-utils": "workspace:*",
-        "@khanacademy/simple-markdown": "workspace:*",
```
jeremywiebe, line 30: I switched from simple to pure-markdown because pure-markdown is where we have Perseus-specific rules (such as `math` and `widgets`).  These aren't strictly necessary for what we're doing _now_ but I thought it was good to use the right set of parsing rules now so we don't create a hard-to-figure-out problem in the future.

packages/perseus-core/src/accessibility.test.ts
```diff
@@ -182,14 +182,12 @@ describe("isItemAccessible", () => {
...
-    });
 
-    describe("markdown", () => {
-        it("should return false if the item markdown contains any inaccessible images", () => {
+        it("should ignore widgets that are not used in the markdown content", () => {
             const itemData: PerseusItem = {
                 question: {
                     content:
-                        "Here's an image: ![](https://example.com/image.png)",
+                        "Here's an explanation: [[☃ explanation 1]]\n\nAnd a graph: [[☃ interactive-graph 1]]\\But matcher 1 config is unused!!",
```
jeremywiebe, line 190: The two widgets in use here are accessible, but `matcher` isn't. It's in the `widgets` dictionary, but is not used in `content` so will never be shown to learners or used in any way.

packages/perseus-core/src/accessibility.test.ts
```diff
@@ -217,6 +237,23 @@ describe("isItemAccessible", () => {
...
+        });
+    });
+
+    describe("markdown", () => {
+        it("should return false if the item markdown contains any inaccessible images", () => {
+            const itemData: PerseusItem = {
+                question: {
+                    content:
+                        "Here's an image: ![](https://example.com/image.png)",
+                    widgets: {},
```
jeremywiebe, line 250: Because of the changes in this PR, it doesn't matter if we do, or do not, have `widgets` in this dictionary now. They're all ignored because they aren't used in the `content` string!

packages/perseus-core/src/accessibility.test.ts
```diff
@@ -259,39 +276,8 @@ describe("isItemAccessible", () => {
             const itemData: PerseusItem = {
                 question: {
                     content:
-                        "Here's an image: ![alt text](https://example.com/image.png)",
+                        "Here's an inaccessible widget: [[☃ matcher 1]]\n\nHere's an image: ![alt text](https://example.com/image.png)",
```
jeremywiebe, line 279: This is an important change for this test. We _must_ include the inaccessible widget in the `content` otherwise its (correctly) ignore entirely.  Other widgets unused in the `content` have been removed in this test because that's not what we're testing in this test specifically.

packages/pure-markdown/src/traversal.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * Traverse all of the nodes in the Perseus Markdown AST. The callback is
+ * called for each node in the AST.
+ */
+export const traverseContent = (ast: any, cb: (node: any) => void) => {
```
jeremywiebe, line 5: I've moved this from `@khanacademy/perseus` as it has nothing to do with UI/Rendering. `@khanacademy/pure-markdown` is the place where our Perseus-specific markdown logic goes so `traverseContent` makes sense here.

packages/perseus/src/perseus-markdown.tsx
```diff
@@ -303,31 +307,6 @@ const inlineParser = (source: string, state: any): any => {
     });
 };
 
-/**
- * Traverse all of the nodes in the Perseus Markdown AST. The callback is
- * called for each node in the AST.
- */
-const traverseContent = (ast: any, cb: (arg1: any) => unknown) => {
```
jeremywiebe, line 310: Moved unchanged into the `@khanacademy/pure-markdown` package.

packages/pure-markdown/src/index.test.ts
```diff
@@ -1,6 +1,6 @@
 import {describe, it} from "@jest/globals";
 
-import {parse} from "../index";
+import {parse} from "./index";
```
jeremywiebe, line 3: I restructured the source tree slight by moving tests out of the `__tests__` folder (there's only two test files and two production files). Hopefully makes it easier to navigate and see that we _have_ tests.

packages/pure-markdown/src/index.ts
```diff
@@ -294,11 +296,24 @@ export const pureMarkdownRules = {
...
+ * Markdown including things like math (`$...$`), tables, and widgets (`[[☃
+ * ...]]`).
+ * @param source The Perseus Markdown string to parse.
+ * @param state The state object to pass to the parser.
+ * @returns The Abstract Syntax Tree (AST) of the parsed Markdown.
+ *
+ * @todo The return type should be Array<SingleASTNode> but that breaks the
+ * perseus-linter's types, so leaving it as `any` for now.
+ */
+export function parse(source: string, state?: any): any {
```
jeremywiebe, line 312: A small change here, I added JSDoc comments, but also changed it from a `const parse = () => {}` arrow function to a plain old JS `function`.

packages/pure-markdown/src/traversal.test.ts
```diff
@@ -0,0 +1,445 @@
+import {describe, it, expect} from "@jest/globals";
+
+import {traverseContent} from "./traversal";
+
+import {parse} from "./index";
+
+describe("traverseContent", () => {
```
jeremywiebe, line 7: Me and 🧞‍♂️AladdIn* created these tests. :) 

* Ie. the AI

packages/perseus-core/src/accessibility.ts
```diff
@@ -43,19 +44,44 @@ export function violatingWidgets(itemData: PerseusItem): Array<string> {
...
+
+    if (hasInaccessibleImage) {
+        return false;
     }
 
-    // Finally, if the markdown is accessible. Check if any widgets are
-    // inaccessible.
-    return violatingWidgets(itemData).length === 0;
+    // Finally, check if any widgets are inaccessible.
+    const itemDataWithOnlyActiveWidgets: PerseusItem = {
```
catandthemachines, line 72: Why change this section of the function. What's the purpose of updating this? 🤔

packages/perseus-core/src/accessibility.test.ts
catandthemachines: Was that erroring or causing issues before? I didn't notice that when adding these tests. 🤔

catandthemachines (approved):
Looks good to me, just a few questions on why some lines were updated or changed.

tatianasnook (approved):
The fix makes a lot of sense — filtering out unused widgets ensures the accessibility check reflects only what's actually shown to learners.

packages/perseus-core/src/accessibility.test.ts
jeremywiebe: They weren't causing issues, but they were adding noise to the test (they were not part of what we were trying to verify here).  So I trimmed the input data to the core of what we're verifying:

packages/perseus-core/src/accessibility.ts
jeremywiebe: This is the core of the PR.  It's building a PerseusItem that excludes any unused widgets so that we correctly decide if it is accessible. 

If we pass in the `itemData` value unchanged, it could have widgets in `itemData.question.widgets` that are unused and inaccessible and cause us to think the item is inaccessible: exactly what we don't want.

packages/perseus-core/src/accessibility.ts
```diff
@@ -43,19 +44,44 @@ export function violatingWidgets(itemData: PerseusItem): Array<string> {
...
+    const ast = parse(itemData.question.content);
+    const widgetIdsInUse = new Set<string>();
+    let hasInaccessibleImage = false;
+
+    traverseContent(ast, (node) => {
+        // Markdown images without alt text are inaccessible!
+        if (node.type === "image" && (node.alt == null || node.alt === "")) {
+            hasInaccessibleImage = true;
+            return;
+        }
```
jeremywiebe, line 58: I had to change this slightly. Before we were looping over the `nodes` and returning `false` as soon as we found an image without alt text. 

Now that this logic was inside the `traverseContent()` node callback, we can't exit out of the `isAccessible()` function using a `return`. So we have to track things with a variable and then return after `traverseContent()` is finished.

PR #2699, opened 2025-07-11
Improve Perseus documentation

ivyolamit:
## Summary:
Improve Perseus documentation

This PR will serve as the base improvement of Perseus documentation to make it easier for users including navigation and consistency:
1. Follows the new sidebar hierarchy
- Introduction
- Renderers
- Components
- Editors
- Widgets
2. The Editors group will contain existing high level perseus editors
3. Specific widget editors will now be located in docs together with the actual widget
4. Additional improvements will be done iteratively

image

https://github.com/user-attachments/assets/a540024c-8af3-49db-91de-9cf02d717b61

Issue: LEMS-3038

## Test plan:

### DIFF
--- a/packages/perseus/src/__docs__/renderers-overview.mdx
+++ b/packages/perseus/src/__docs__/renderers-overview.mdx
@@ -26,36 +26,42 @@ There are 4 main Perseus renderers:
 Perseus renderers implement several sophisticated features to enhance content rendering and user interaction:
 
 **1. Math Rendering**
+
 -   All renderers use MathJax for rendering mathematical expressions
 -   Support for inline and display-mode TeX expressions
 -   Custom preprocessing via `preprocessTex` for enhanced rendering
 -   Zoomable math expressions via `ZoomableTeX` component
 
 **2. Widget System**
+
 -   Extensible widget architecture for interactive elements
 -   Widget state management through `WidgetContainer`
 -   Error boundaries to prevent widget failures from affecting the entire content
 -   Support for widget-specific transformations and upgrades
 
 **3. Accessibility**
+
 -   Comprehensive focus management for keyboard navigation
 -   ARIA attributes for interactive elements
 -   Screen reader compatibility
 -   Support for alternative input methods
 
 **4. Internationalization**
+
 -   Integration with translation systems (JIPT)
 -   Bidirectional text support
 -   Translation linting via `TranslationLinter`
 -   Localized content rendering
 
 **5. Mobile Support**
+
 -   Responsive layouts for different screen sizes
 -   Touch-friendly interactions
 -   On-screen keypad integration for math inputs
 -   Mobile-specific styling controlled via apiOptions
 
 **6. Performance Optimization**
+
 -   Selective re-rendering of widgets
 -   Asset preloading for images
 -   Caching of rendered markdown content
@@ -96,9 +102,9 @@ When rendering a Perseus item (exercise):
 **1. A `PerseusItem` object is passed to `ServerItemRenderer`**
 
 **2. `ServerItemRenderer` renders two main components:**
-- The question content using `Renderer` (with the item's question content and widgets)
-- Hints content using `HintsRenderer` (which uses `HintRenderer` for each hint)
 
+-   The question content using `Renderer` (with the item's question content and widgets)
+-   Hints content using `HintsRenderer` (which uses `HintRenderer` for each hint)
 
 ### Article Rendering
 
@@ -112,16 +118,19 @@ When rendering a Perseus article:
 The `Renderer` component handles the core functionality of rendering widgets through several key steps:
 
 **1. During initialization (`_getInitialWidgetState`), the Renderer:**
+
 -   Processes the widgets configuration from props
 -   Applies default values to widget options using `applyDefaultsToWidgets`
 -   Creates initial widget props and state
 
 **2. For rendering content with widgets, the Renderer:**
+
 -   Parses markdown content using `PerseusMarkdown`
 -   Identifies widget references in the content
 -   Manages widget containers that will host each widget instance
 
 **3. For each widget in the content, the `Renderer`:**
+
 -   Runs the widget's options through upgrade transforms
 -   Creates widget props by combining:
     -   The widget's specific configuration
@@ -133,6 +142,7 @@ The `Renderer` component handles the core functionality of rendering widgets thr
     -   Applies accessibility attributes
 
 **4. The Renderer also manages widget lifecycle by:**
+
 -   Tracking interactions with `InteractionTracker`
 -   Handling focus changes between widgets
 -   Coordinating with the parent renderer for scoring and state management
@@ -142,28 +152,33 @@ The `Renderer` component handles the core functionality of rendering widgets thr
 Perseus renderers implement sophisticated focus management to handle user interactions across multiple widgets:
 
 **1. Focus Tracking**
+
 -   Renderers maintain a `_currentFocus` property that stores the current focus path
 -   The focus path is an array that identifies which widget and which part of the widget is focused
 -   This tracking enables proper keyboard navigation and accessibility features
 
 **2. Focus Path System**
+
 -   Focus paths use a hierarchical structure: `[widgetId, ...internalPath]`
 -   For complex widgets, the internal path can identify specific sub-components
 -   The `isIdPathPrefix` helper function determines if one path is a prefix of another
 
 **3. Focus Event Handling**
+
 -   Focus changes trigger callbacks between components:
     -   `_handleFocusChange` processes focus transitions
     -   `_setCurrentFocus` updates the current focus state
     -   `_onRendererBlur` handles when focus leaves a renderer
 
 **4. Mobile Support**
+
 -   On mobile devices, focus events trigger keypad visibility:
     -   When a math input is focused, the on-screen keypad appears
     -   When focus moves away, the keypad is hidden
 -   The `ServerItemRenderer` coordinates with `keypadElement` to manage this behavior
 
 **5. Focus Delegation**
+
 -   Parent renderers delegate focus to child components:
     -   `ServerItemRenderer.focus()` delegates to `questionRenderer.focus()`
     -   `ServerItemRenderer.focusPath()` allows focusing specific widgets by path


testing/server-item-renderer-with-debug-ui.tsx
```diff
@@ -28,7 +28,7 @@ type Props = {
 };
 
 /**
- * A component that renders a Perseus item with debug UI controls
+ * ServerItemRendererWithDebugUI is a component that renders a Perseus item with debug UI controls
```
ivyolamit, line 31: Since this is considered as documentation, I updated this to be more specific so it will show nicely in storybook docs.
**Previous:**
image

**Updates**
image

.storybook/preview.tsx
```diff
@@ -89,6 +114,8 @@ const preview: Preview = {
         },
         docs: {
             toc: true,
+            // This hides stories from autodocs by default, set to true in specific stories to show them
+            disable: true,
```
ivyolamit: we are intentionally setting this to true for now since we do not want to show all autodocs by default and we want to be intentional on what we are showing so the improved documentation will be more focused and shows what's only relevant to users. Without this stories like internal components will be shown by default.

.storybook/preview.tsx
```diff
@@ -61,7 +61,32 @@ const preview: Preview = {
 
         options: {
             storySort: {
-                order: ["Perseus", "PerseusEditor", "Math-Input", "*"],
+                method: "alphabetical",
```
ivyolamit: Something I just read recently, I'm considering using deeper sort in next PR: https://storybook.js.org/addons/storybook-deeper-sort

Current issue: by default storybook will only follow sort order up to 2 levels which cause the Widget subfolder not ordered as intended. To mitigate for now in base PR, alphabetical method is used.

anakaren-rojas (approved):
This is incredible!

ivyolamit:
latest change includes deep sort of the sidebar navigation

| Before | After |
|------- | ----- |
|image | image |

.storybook/main.ts
```diff
@@ -1,7 +1,6 @@
 import {mergeConfig} from "vite";
-import {spacing} from "@khanacademy/wonder-blocks-tokens";
-
-import type {StorybookConfig} from "@storybook/react-vite";
+import {configureSort} from "storybook-multilevel-sort";
+import {StorybookConfig} from "@storybook/react-vite";
```
jeremywiebe: **nit**: This should still be a `type` import. Although in this instance it matters little, I find it helpful to be able to see when we're importing a type vs. actual behaviour.

.storybook/main.ts
```diff
@@ -29,31 +28,66 @@ const cssWrapper = {
...
-        "../packages/*/src/**/*@(.stories|.fixturestories).@(ts|tsx)",
+        // This will be used for the main documentation pages
+        "../__docs__/**/*.@(stories.ts|stories.tsx|mdx)",
+
+        // Docs for Perseus editor
+        "../packages/perseus-editor/src/**/__docs__/**/*.@(stories.ts|stories.tsx|mdx)",
+
+        // Docs for Perseus widgets, components, and renderers
+        "../packages/perseus/src/**/__docs__/**/*.@(stories.ts|stories.tsx|mdx)",
+        "../packages/perseus/src/widgets/**/*.@(stories.ts|stories.tsx|mdx)",
```
jeremywiebe, line 74: Why did we need to adjust these globs?  These are now fairly specific and if we add stories anywhere else, they won't show up, right?  Perhaps that's fine and will guide us. 🤷‍♂️

__docs__/introduction.mdx
```diff
@@ -0,0 +1,92 @@
+import {Meta} from "@storybook/addon-docs/blocks";
```
jeremywiebe: With the introduction of this mdx file, could we eliminate the `architecture.md` file entirely?  Or pull out the pieces that you've added in this file?  If it doesn't all fit here, maybe we could add an `Architecture` node in storybook to hold the other info.

__docs__/introduction.mdx
```diff
@@ -0,0 +1,92 @@
...
+
+-   Standard Markdown formatting (headings, lists, links, etc.)
+-   Math expressions enclosed in dollar signs: `$y = mx + b$`
+-   Widgets using the syntax: `[[☃ widget-type 1]]`, where each widget has a unique ID
+
+When a learner interacts with widgets (such as answering questions), Perseus can:
+
+-   Validate answers against expected solutions
+-   Provide feedback and hints
+-   Track progress through exercises
```
jeremywiebe: I'm not sure this is true. The progress tracking through exercises is a part of the chrome that hosts Perseus and is not inherently part of Perseus.

__docs__/introduction.mdx
```diff
@@ -0,0 +1,92 @@
...
+-   Support a variety of input methods
+
+## Core Components
+
+Perseus consists of four main components:
```
jeremywiebe: Most of these components have multiple implementations... Perhaps we should even say they're "concepts" rather than "components".

```suggestion
Perseus consists of four main types of components:
```

__docs__/introduction.mdx
```diff
@@ -0,0 +1,92 @@
...
+### 2. Widgets
+
+Widgets are self-contained React components that provide interactive functionality beyond standard Markdown. They can:
+
+-   Appear within articles and exercises (in questions and hints)
+-   Be nested within other widgets (e.g., the "graded-group" widget)
+-   Handle various input types (multiple-choice, graphing, numeric input, etc.)
+-   Validate and score user responses
+
+See our widgets in action in the [Test Everything](https://www.khanacademy.org/internal-courses/test-everything) course.
```
jeremywiebe: FWIW, I don't think we should publish references to our internal courses like this. This is an open source repo and this storybook info is published publicly on Github.

__docs__/introduction.mdx
```diff
@@ -0,0 +1,92 @@
...
+-   Preview content as it will appear to learners
+
+Each widget type has its own specialized editor component.
+
+#### Preview Frames
+
+As part of the editing experience, preview frames allow content creators to see how their content will look when published:
+
+-   **Side-by-side view**: Shows both the content source and rendered output
+-   **Full preview**: Displays only the rendered content as it will appear to users
```
jeremywiebe: Although true, much of the preview functionality lives inside our internal webapp repository. As a result, Perseus doesn't _really_ ship with much of a preview system by itself. We aspire to have something, but I'm not sure this section is worth keeping/noting right now.

__docs__/introduction.mdx
jeremywiebe: Also, we only have 3 numbered items in this section... so should this say "three main components"? 😄

__docs__/introduction.mdx
```diff
@@ -0,0 +1,92 @@
...
+
+## Data Flow
+
+Perseus follows a straightforward data flow:
+
+1. Content is created and stored as structured data (PerseusItem or PerseusRenderer)
+2. When displayed, this data is passed to the appropriate renderer
+3. The renderer processes the content, instantiates widgets, and renders the final output
```
jeremywiebe:
```suggestion
3. The renderer parses the content, instantiates widgets, and renders the final output
```

.storybook/main.ts
```diff
@@ -29,31 +28,66 @@ const cssWrapper = {
...
+// See https://storybook.js.org/addons/storybook-multilevel-sort
+configureSort({
+    storyOrder: {
+        introduction: null,
+        widgetGallery: null,
+        theming: null,
+        renderers: {
+            overview: null,
+        },
+        components: null,
```
jeremywiebe: Could we place `components` lower down in this list?  At least below `widgets`.  Renderers, Editors, and Widgets are more important and higher-level concepts IMHO.  The `components` are really mostly internal pieces we use to build up the widgets and editors.

packages/math-input/src/components/keypad/keypad-button.stories.tsx
```diff
@@ -22,6 +22,7 @@ export default {
             options: {...keyConfigs},
         },
     },
+    tags: ["!dev"],
```
jeremywiebe, line 25: **question**: Why is this needed? or what does it do?

packages/perseus-editor/src/widgets/__docs__/categorizer-editor.stories.tsx
```diff
@@ -11,7 +11,13 @@ type Story = {
...
 
 export default {
-    title: "PerseusEditor/Widgets/Categorizer Editor",
+    title: "Widgets/Categorizer/Editor Demo",
+    component: CategorizerEditor,
+    tags: ["!dev"],
+    parameters: {
+        componentSubtitle:
+            "An editor for adding a categorizer widget that allow users to sort items into categories.",
+    },
```
jeremywiebe: For a future PR if you're continuing to work in this area, you could consider adjusting to a more modern Story approach: https://github.com/Khan/perseus/pull/1802 

I wonder if that'd also allow you to put this documentation as a JSDoc comment directly on the editor component. 🤔

jeremywiebe (approved):
LOVE IT! Thanks for this huge improvement to the Storybook setup.

.storybook/main.ts
ivyolamit: This is just temporary that it's too specific for now, easier for the transition. I'm planning to simplify this at the end of the improvements.

.storybook/main.ts
ivyolamit: good catch! original one actually has the type 😅 I'll update it.

.storybook/main.ts
ivyolamit: @jeremywiebe do you happen to know why we have a separate "Math-Input" section? just below PerseusEditor in the current [docs](https://khan.github.io/perseus/?path=/docs/math-input-full-keypad--docs) looks like it can be [merged](https://khan.github.io/perseus/?path=/docs/perseus-components-math-input--docs) together with `Components/Math-Input`, what do you think?

image

packages/math-input/src/components/keypad/keypad-button.stories.tsx
ivyolamit: The `dev` tag will itemize/show each stories in the sidebar navigation, which can cause too much clutter. Instead we can leverage the right TOC since it's already enabled in our configuration. See image below:

| Before | After |
| ------ | ----- |
| image | image |


image

.storybook/main.ts
jeremywiebe: I think it's because these come from a different package `@khanacademy/math-input`.  There are more than one components in our ecosystem named "math input"... this section is the separate package.  I think for now it's useful to keep it separate for that reason.

packages/math-input/src/components/keypad/keypad-button.stories.tsx
jeremywiebe: Oh Cool.... thanks!

packages/perseus-editor/src/widgets/__docs__/categorizer-editor.stories.tsx
ivyolamit: Yes we can definitely do that. I actually prefer it that way directly have the doc in the editor component. The only reason I had the  `componentSubtitle` is that we are not consistent on using the `component` in our docs for the most part in editor it uses the correct equivalent editor component. A few uses `EditorPageWithStorybookPreview` and in our widgets mostly uses `ServerItemRendererWithDebugUI` which can a bit mess up with the info 😅 it would look like this:

image


👉🏼 But I'll find a way moving forward on using the JSDoc instead of using `componentSubtitle`

jeremywiebe: Definitely not blocking. I'm aware of `ServerItemRendererWithDebugUI` usage and yeah, that makes sourcing the JSDoc on the widgets difficult.  :) I'd say if there's components that are used directly in stories, start with those.  Thanks again... this is all looking amazing!

__docs__/introduction.mdx
ivyolamit: after checking the files i can pull out the missing pieces from `architecture.md` and delete it afterwards. Currently some of the renderer specific information from `architecture.md` are found in `renderers-overview.mdx` and the high-level information are in introduction.

I'll update this shortly.

PR #2701, opened 2025-07-14
Remove the expression widget validation tooltip

jeanettehead:
## Summary:
this is an anti-pattern, no other widget (to my knowledge) has this type of realtime feedback - feedback is generally provided after submit - including for this invalid use case. It is causing a styling issue with assessment and a pattern we want to move away from, so removing this code.

Issue: LEMS-3292

## Test plan:
1. go to an expression widget
2. enter an invalid input such as just a single hyphen
3. observe that there is no immediate feedback that your answer is invalid
(upon submitting, user should be warned their answer is invalid and not scored incorrectly, maintaining the existing behavior)

### DIFF
--- a/packages/perseus/src/widgets/expression/expression.test.tsx
+++ b/packages/perseus/src/widgets/expression/expression.test.tsx
@@ -4,7 +4,7 @@ import {
     generateTestPerseusItem,
 } from "@khanacademy/perseus-core";
 import {scorePerseusItem} from "@khanacademy/perseus-score";
-import {act, screen, waitFor} from "@testing-library/react";
+import {act, screen} from "@testing-library/react";
 import {userEvent as userEventLib} from "@testing-library/user-event";
 
 import {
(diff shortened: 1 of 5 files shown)


jeanettehead:
BEFORE
Screenshot 2025-07-14 at 2 28 56 PM

AFTER
Screenshot 2025-07-14 at 2 28 50 PM

packages/perseus/src/widgets/expression/expression.tsx
```diff
@@ -1,9 +1,7 @@
-import * as KAS from "@khanacademy/kas";
```
handeyeco, line 1: It's kind of amazing that we can remove `kas` from Expression.

packages/perseus/src/widgets/expression/expression.tsx
```diff
@@ -353,58 +282,26 @@ export class Expression
...
-                            extraKeys={
-                                this.props.keypadConfiguration?.extraKeys
-                            }
-                            onAnalyticsEvent={
-                                this.props.analytics?.onAnalyticsEvent ??
-                                (async () => {})
-                            }
-                        />
-                    </Tooltip>
+                <div className={className}>
```
handeyeco: Could this be simplified now?

```suggestion
                <div className="perseus-widget-expression">
```

packages/perseus/src/widgets/expression/expression.test.tsx
```diff
@@ -501,54 +501,6 @@ describe("Expression Widget", function () {
...
-        it("shows error text in tooltip", async () => {
-            // Arrange
-            const {renderer} = renderQuestion(expressionItem2.question);
-            const expression = renderer.findWidgets("expression 1")[0];
-
-            // Act
-            // Note(jeremy): You might think you could collapse all of these calls
-            // inside a single act() block, but that didn't work. The only way this
-            // test passes is with each statement in its own act() call. :/
-            act(() => expression.insert("x&&&&&^1"));
```
handeyeco, line 521: Wonder if it's worth adding ""x&&&&&^1" to the `scoreExpression` tests or something so we're not losing a check.

packages/perseus/src/widgets/expression/expression.test.tsx
```diff
@@ -501,54 +501,6 @@ describe("Expression Widget", function () {
...
-            );
-        });
-
-        it("does not show error text when the sen() function is used (Portuguese for sin())", async () => {
-            // Arrange
-            const {renderer} = renderQuestion(expressionItem2.question);
-            const expression = renderer.findWidgets("expression 1")[0];
-
-            // Act
-            act(() => expression.insert("sen(x)"));
```
handeyeco, line 540: Wonder if it's worth adding "sen(x)" to the `scoreExpression` tests or something so we're not losing a check.

handeyeco (approved):
Nice! Thanks for doing this.

Might be worth letting TUT know that this is happening so we don't get bug reports.

packages/perseus/src/widgets/expression/expression.test.tsx
jeanettehead: good eye ty

jeanettehead:
> @jeanettehead changes looks logical to me and suited with the new needs. 👍🏼 Looks like the validation tooltip was intentionally added in the past as mentioned in this [PR](https://github.com/Khan/perseus/pull/711) probably it made sense back then and with the newer requirement and styling in assessment this makes sense to me.
> 
> Do we know if there are other instances of this anti-pattern? or something we need to share to the team since it's a pattern we want moving forward?

@ivyolamit I saw that too! It looks like there used to be a ⚠️ icon that would show up inside the input when it was invalid, and then that was inaccessible and got refactored into the tooltip. So this is just removing the functionality altogether, no icon no nothing.

I don't know other instances of this anti-pattern. As far as sharing with the team, I aligned with Caitlyn on this, and we believe we have some widget validation work coming up where we will start to have stronger opinions about this kind of behavior. But I'll do a demo next sprint planning to share out!

PR #2704, opened 2025-07-15
Update utils/sync-dependencies.ts to operate on pnpm-workspace.yaml

benchristel:
## Summary:
`utils/sync-dependencies.ts` was broken because frontend's root package.json
no longer lists production dependencies, only dev dependencies. Each app and lib
gets its dependency versions from the `pnpm-workspace.yaml` catalog.

This PR updates `utils/sync-dependencies.ts` to account for this change.

NOTE TO REVIEWERS: This PR will probably be easiest to review one commit at a
time.

Issue: none

## Test plan:

Run `utils/sync-dependencies.ts ../frontend/pnpm-workspace.yaml`. You should
see some package versions update in Perseus's pnpm-workspace.yaml and
pnpm-lock.yaml.


utils/sync-dependencies.ts
```diff
@@ -97,71 +69,51 @@ function main(argv: string[]) {
...
 
-    for (const pkgName of packageNamesInRepo) {
-        if (!(pkgName in clientVersionRanges)) {
+    // In our peer dependencies, declare that Perseus will work with any
+    // package version compatible with the one we install in dev.
+    for (const pkgName of ourPeerDeps) {
+        if (!clientCatalog.has(pkgName)) {
+            // TODO(benchristel): throw an error here instead of ignoring the
+            // package. If the client doesn't provide a package that we need as
+            // a peer dep, it's not safe to deploy Perseus!
```
benchristel, line 93: Ugh, actually, the situation is even more complicated than this comment implies.

- The frontend pnpm-workspace.yaml might list a package that's a peer dep of Perseus, but that package might not actually be installed in an app that uses Perseus.
- If the app doesn't install the peer dep, that might actually be okay if it doesn't use the specific Perseus package (e.g. `@khanacademy/perseus-editor`) that requires it. I.e. not every peer dep listed in our catalog is needed by every Perseus package.

benchristel: On third thought, though, I guess it's not the job of this script to ensure that peer deps are actually provided, just to make sure we're developing against versions compatible with, and no newer than, the ones webapp uses. So throwing an error is probably the right call.

utils/sync-dependencies.ts
```diff
@@ -97,71 +69,51 @@ function main(argv: string[]) {
...
-                    "We don't know what dev dependency to install!",
-            );
+        ourWorkspace.catalogs.peerDeps[pkgName] =
+            `^${clientCatalog.minimumVersionOf(pkgName)}`;
+    }
+
+    // In development, install the minimum version of each package
+    // required by the client application. This ensures we don't
+    // accidentally depend on features of the package added after that
+    // version.
```
jeremywiebe, line 103: Agreed.  That's good to note here!

jeremywiebe (approved):
Thanks for fixing this.  I think this is the right fix for now. In the near future, as Khan/frontend isolates apps more, we might need to re-think how we deal with Perseus peer dep compatibility.

PR #2706, opened 2025-07-15
[Editor] Remove unused frameSource prop from EditorPage component

jeremywiebe:
## Summary:

I kept noticing that the `frameSource` prop is a required prop but we don't do anything with it, so this PR finally removes it. It looks like at one point it was related to preview, but now we use the `previewURL` instead.

Issue: "none"

## Test plan:

`pnpm tsc`

### DIFF
--- a/.changeset/tasty-jars-invent.md
+++ b/.changeset/tasty-jars-invent.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": major
+---
+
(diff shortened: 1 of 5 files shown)


nishasy (approved):
Makes sense to me

nishasy:
Wow, the changeset for this is from 3 years ago

Screenshot 2025-07-17 at 12 13 52 PM

PR #2708, opened 2025-07-16
test a fix

jeanettehead:
## Summary:
<Add your summary here>

Issue: LEMS-XXXX

## Test plan:

### DIFF
--- a/packages/perseus-editor/src/editor.tsx
+++ b/packages/perseus-editor/src/editor.tsx
@@ -448,10 +448,9 @@ class Editor extends React.Component<Props, State> {
         this.setState({textAreaValue: e.currentTarget.value});
         // TODO(jeff, CP-3128): Use Wonder Blocks Timing API.
         // eslint-disable-next-line no-restricted-syntax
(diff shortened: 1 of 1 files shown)


packages/perseus-editor/src/editor.tsx
```diff
@@ -200,6 +200,22 @@ class Editor extends React.Component<Props, State> {
...
     }
 
+    // TODO(arun): This is a deprecated method, use the appropriate replacement
+    // eslint-disable-next-line react/no-unsafe
+    UNSAFE_componentWillReceiveProps(nextProps: Props) {
+        // this function matters for adding images to the editor
+
+        // when images are added, the editor is never focused (the image button is)
+        const textAreaHasFocus =
+            this.textarea.current === document.activeElement;
```
benchristel, line 210: What if `this.textarea.current` and `document.activeElement` are both null?

jeanettehead: ....then we get an error and crash the editor.

AKA ty for pointing that out <3

PR #2710, opened 2025-07-16
[Interactive Graph] Stop alongEdge labels from overflowing out of graph container

nishasy:
## Summary:
The alongEdge graph label overflows out of the container on web, and is completely cut off on mobile.

This only happens for very specific cases where the y-axis tick labels are rendered outside the graph. This means when the x range is greater than 0, 0, or relatively close to 0.

To fix this, I added extra padding by the exact offset needed for these edge cases.

Issue: https://khanacademy.atlassian.net/browse/LEMS-3282

## Screenshots and demo

| Before | After |
| --- | --- |
| Screenshot 2025-07-16 at 12 52 11 PM | image |

Observe how the y-axis label stays in the same spot no matter what the range of the graph is or where the tick labels are.

https://github.com/user-attachments/assets/b44cc732-20c1-40c6-8083-d25f9d2df574

## Test plan:
Storybook
- http://localhost:6006/?path=/story/perseuseditor-widgets-interactive-graph--interactive-graph-linear
- Set the label location to "along edge"
- Set the x Range start to 0
- Confirm that the y label is still visible in the preview
- Try all 0.1 interavals between -1 and 0 for the x range start
- Confirm that the y label is sitll visible in all the cases

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
@@ -24,7 +24,11 @@ import AxisLabels from "./backgrounds/axis-labels";
 import {AxisTicks} from "./backgrounds/axis-ticks";
 import {Grid} from "./backgrounds/grid";
 import {LegacyGrid} from "./backgrounds/legacy-grid";
-import {getLabelPosition} from "./backgrounds/utils";
+import {
+    fontSize,
+    fontSizeYAxisLabelMultiplier,
+    getLabelPosition,
+} from "./backgrounds/utils";
 import GraphLockedLabelsLayer from "./graph-locked-labels-layer";
 import GraphLockedLayer from "./graph-locked-layer";
 import {renderAngleGraph} from "./graphs/angle";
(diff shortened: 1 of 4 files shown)


packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -157,6 +159,33 @@ export const MafsGraph = (props: MafsGraphProps) => {
...
+        height,
+    };
+
+    const [xAxisLabelLocation, yAxisLabelLocation] = getLabelPosition(
+        graphInfo,
+        labelLocation,
+        tickStep,
+    );
+
+    // -17.5 is the standard label offset on the graphs with alongEdge labels.
```
mark-fitzgerald: 17.5 feels like a magic number. Is there a way that you can expand upon why it "is the standard label offset"? Can you point to something like line-height, font-size, or anything that demonstrates this value?

mark-fitzgerald (approved):

SonicScrewdriver (approved):
This is a great improvement!

PR #2711, opened 2025-07-16
Detect schema in Github Action

jeremywiebe:
## Summary:

This PR sets up a GitHub Action that "builds" our `data-schema.ts` file for the current PR and the PR's base branch and compares them. The build should include any files that `data-schema.ts` imports (although the number of imports is, and should remain, pretty small).

The goal is to detect schema changes to give ourselves a heads-up if this happens. 

Issue: LEMS-3304

## Test plan:

- Add a schema change in this PR and push 
- Wait for actions to finish - the change should be posted in the "Schema Changes" comment
- Revert the schema change
- Wait for actions to finish - the "Schema Changes" should be updated to say "No schema changes"

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -2,6 +2,7 @@
 
 root = true
 
+[*]
 indent_style = space
 indent_size = 4
 
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -273,11 +273,11 @@ jobs:
             - name: Check for Schema Change
               id: check_for_schema_change
               run: |
-                  set -x
-                  set +e
+                  # We generate a random DELIMeter so that the existence of the
+                  # delimeter in the diff output doesn't cause problems.
                   DELIM=$(uuidgen)
                   echo "SCHEMA_DIFF<<$DELIM" >> "$GITHUB_OUTPUT"
-                  diff ${{ runner.temp }}/schema-pr.d.ts ${{ runner.temp }}/schema-base.d.ts >> "$GITHUB_OUTPUT"
+                  diff -C 3 ${{ runner.temp }}/schema-pr.d.ts ${{ runner.temp }}/schema-base.d.ts >> "$GITHUB_OUTPUT"
                   DIFF_EXIT_CODE=$? # Capture the exit code of diff
                   echo "$DELIM" >> "$GITHUB_OUTPUT"
 
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -273,7 +273,7 @@ jobs:
             - name: Check for Schema Change
               id: check_for_schema_change
               run: |
-                  # So a non-zero exit code from 'diff' doesn't bail this step
+                  # So a non-zero exit code from 'diff' doesn't bail this step.
                   set +e
                   # We generate a random DELIMeter so that the existence of the
                   # delimeter in the diff output doesn't cause problems.
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -273,7 +273,8 @@ jobs:
             - name: Check for Schema Change
               id: check_for_schema_change
               run: |
-                  set +e # So non-zero exit code from 'diff' doesn't bail this step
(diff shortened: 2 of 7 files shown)


.editorconfig
```diff
@@ -2,6 +2,7 @@
 
 root = true
 
+[*]
```
jeremywiebe, line 5: Without this, many tools ignore the `.editorconfig` because it's malformed.

.github/workflows/node-ci.yml
```diff
@@ -209,9 +209,10 @@ jobs:
               with:
                   node-version: ${{ matrix.node-version }}
 
-            # Make sure our packages aren't growing unexpectedly
-            # This must come last as it builds the old code last and so leaves the
-            # wrong code in place for the next job; in other words, it leaves the repo on a base branch.
+            # Make sure our packages aren't growing unexpectedly This must come
```
nishasy:
```suggestion
            # Make sure our packages aren't growing unexpectedly. This must come
```

nishasy (commented):
Sorry I'm not confident enough in my knowledge to approve this, but I did find a missing `.`.

.github/workflows/node-ci.yml
```diff
@@ -228,6 +229,111 @@ jobs:
...
+                    packages/perseus-core/src/data-schema.ts
+
+                  pnpm prettier --write ${{ runner.temp }}/schema-base.d.ts
+
+            - name: Check for Schema Change
+              id: check_for_schema_change
+              run: |
+                  # So a non-zero exit code from 'diff' doesn't bail this step.
+                  set +e
+                  # We generate a random DELIMeter so that the existence of the
```
handeyeco: Reading this top-to-bottom, this was confusing. I thought you were making some kind of deli joke.

```suggestion
                  # We generate a random delimiter (DELIM) so that the existence of the
```

.github/workflows/node-ci.yml
```diff
@@ -228,6 +229,111 @@ jobs:
...
+                  PR_NUMBER="${{ github.event.pull_request.number }}"
+                  gh pr edit "$PR_NUMBER" --remove-label "schema-change" || true
+
+            - name: Find Schema Change comment
+              uses: peter-evans/find-comment@3eae4d37986fb5a8592848f6a574fdf654e61f9e
+              id: find-schema-change-comment
+              with:
+                  issue-number: ${{ github.event.pull_request.number }}
+                  comment-author: "github-actions[bot]"
+                  body-includes: "Schema Change"
```
handeyeco: No action needed, but this seems like it could be a little fragile. I noticed some of our automation does a markdown hack to add a comment to...a comment...I wonder if we could use that to make sure we're editing the right thing.

I don't actually care that much, so if you think this is good enough for now it's good enough for me.

handeyeco (approved):
I don't actually understand how this will detect answerless schema changes. It's an internal tool though, so I feel like we can try it out and see how it goes.

.github/workflows/node-ci.yml
jeremywiebe: Good suggestion. It's not a DELI Meter.  It's a delimiter. Thanks. :)

.github/workflows/node-ci.yml
jeremywiebe: This is a frustrating part of these find/add/edit GitHub Actions we use. They don't give you any definitive way to find an existing comment. Once you have the ID your good, but before that you have to find it by text matching.  Perhaps we could put a token at the end of the comment that wouldn't ever be added by a human (like a GUID). But we have this pattern in place in other flows and as far as I know it hasn't failed.

jeremywiebe:
> I don't actually understand how this will detect answerless schema changes. It's an internal tool though, so I feel like we can try it out and see how it goes.

@handeyeco the reference to answerless in `node-ci.yml` is a bit of a red herring. The detection of answerless schema changes is implicit in the detection of changes to `data-schema.ts`. I've removed references to answerful/answerless in the `.yml` file as I'm just focused initially on helping engineers understand that changes to `data-schema.ts` are significant.

PR #2713, opened 2025-07-16
Bugfix: Radio in editor crashing when removing an option

handeyeco:
## Summary:

Talking with Nisha:

> It's because we're storing two chunks of state in two places.
>
> 1. The choices (set in the editor): this is the list of choices that are represented in widget options
> 2. The UI state (set in the widget): this is additional information that has to do with the state of user interaction with the widget
>
> The original problem was that these two things get out of sync when we delete an option: suddenly the widgets options have 3 options but the UI state (in the widget) have 4 options. When the widget tries to merge the choices from widget options with the choices from state, it's trying to access a choice that no longer exists.
> 
> The Renderer / Preview really should be doing a hard reset anyway because it doesn't make sense to use old state with a new PerseusItem (and every change to the data is technically a new PerseusItem).

Issue: LEMS-3299

## Test plan:

### DIFF
--- a/.changeset/empty-ghosts-exist.md
+++ b/.changeset/empty-ghosts-exist.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Bugfix: Radio in editor crashing when removing an option


packages/perseus/src/widgets/radio/radio.ff.tsx
```diff
@@ -24,6 +25,16 @@ type State = {
     choiceStates: ChoiceStateWithoutSelected[];
 };
 
+function initChoiceStates(choices: Props["choices"]) {
+    return choices.map(() => ({
+        highlighted: false,
```
nishasy, line 30: I thought `highlighted` was removed recently? @SonicScrewdriver Was that something you were working on?

nishasy: Looks like it hasn't been removed yet, so it makes sense to keep it here for now

SonicScrewdriver: We can't remove it yet as it's part of the ChoiceState iirc, which is used by the old radio widget. 

However, it's some kind of zombie property that doesn't seem to have been used in a LONG time, and we are never setting it to true.

nishasy (approved):
Thanks for explaining the reset logic to me.

PR #2714, opened 2025-07-16
[Interactive Graph][Point] Allow points on Point graphs to sit on edge of graph

nishasy:
## Summary:
We received a request from content authors to allow users to move a point to
the edge of the graph, specifically for the case that a graph start at 0 in
both the x and y direction (like in physics) and a point needs to sit at (0, 0).

To allow this, I'm making it so that the bounds are not inset and the overflow
is not hidden ONLY FOR POINT GRAPHS.

I'll explore other graph types if specifically requested, but this is not possible
for most graph types because of their overflow.

Issue: https://khanacademy.atlassian.net/browse/LEMS-3213

| Before | After |
| --- | --- |
| Screenshot 2025-07-16 at 3 30 45 PM | Screenshot 2025-07-16 at 3 30 52 PM |

## Test plan:
`pnpm jest packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.test.ts`

Storybook
- Move a point to the corner of the graph
- Confirm that it can sit at the corner of the graph, fully visible
- Confirm it can't go past the bounds of the graph

### DIFF
--- a/.changeset/strange-snakes-lay.md
+++ b/.changeset/strange-snakes-lay.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+[Interactive Graph][point] Allow points on Point graphs to sit on edge of graph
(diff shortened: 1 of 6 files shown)


SonicScrewdriver (approved):
Makes sense to me! I tried breaking it in storybook and it all seemed to work perfectly regardless of the crazy values I gave it. :) 

It does feel like all the "boundAndSnap" functions should probably be moved to utils to clean up the reducer, but that could be a future ticket.

PR #2716, opened 2025-07-17
Fix the way we import OrdererPublicWidgetOptions

handeyeco:
## Summary:

See this thread: https://khanacademy.slack.com/archives/C07K2C44LDB/p1752773142981419

Just a small tweak so we're exporting/importing correctly.


jeremywiebe (approved):
Thank-you!

PR #2719, opened 2025-07-17
Detect item splitting function changes in Github Action

jeremywiebe:
## Summary:

This PR builds on the work in #2711 by adding another Github Action job to check for changes to our item splitting (for answerless data) logic. It does this by:

  * adding a new export to `@khanacademy/perseus-core` named `item-splitting` (so you'd import it as `import {splitPerseusItem} from "@khanacademy/perseus-core/item-splitting";`).  This export contains _all_ of the item splitting code in a single file so we can be sure if it changes, we know the logic has changed.
  * adding a Github Action job in `node-ci.yml` to build the current PR's code and its base branch code and compare the two item-splitting bundles.
  * refactors the schema change logic into a new, composite Github action and uses it for both schema change checks and item splitting logic checks

Issue: LEMS-3305

## Test plan:

This PR cannot fully test the split function change detection because it introduces an `"exports"` to the `perseus-core` package. This export causes _all_ of the split logic to be bundled into the new export and we use that as the basis for checking for changes. 

So, my plan is to review this PR and land it.  Then start a new PR and test/verify the change detection. I can also tweak the action and diff rendering in that new PR because by that time the base branch will be building the new `"exports"` bundle. 

Finally, the two "change detection" jobs are failing in the "post" step (see screenshot). I am fairly sure that it's because it's trying to use the composite action from the base branch to cleanup but it doesn't exist there so it fails. My hunch is that this error will also resolve once I land this PR and begin testing in a new one. 

image

### DIFF
--- a/.changeset/fair-pumpkins-move.md
+++ b/.changeset/fair-pumpkins-move.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/perseus": minor
+"@khanacademy/perseus-score": minor
+---
+
+Export public widget options so they can be used in Perseus
--- a/.changeset/rude-schools-collect.md
+++ b/.changeset/rude-schools-collect.md
@@ -1,0 +1,14 @@
+---
+"@khanacademy/kas": minor
+"@khanacademy/keypad-context": minor
+"@khanacademy/kmath": minor
+"@khanacademy/perseus": minor
+"@khanacademy/perseus-core": minor
+"@khanacademy/perseus-linter": minor
+"@khanacademy/perseus-score": minor
+"@khanacademy/perseus-utils": minor
+"@khanacademy/pure-markdown": minor
+"@khanacademy/simple-markdown": minor
+---
+
+Update package to use modern 'exports' in package.json
--- a/.changeset/rude-schools-collect.md
+++ b/.changeset/rude-schools-collect.md
@@ -1,14 +1,5 @@
 ---
-"@khanacademy/kas": minor
-"@khanacademy/keypad-context": minor
-"@khanacademy/kmath": minor
-"@khanacademy/perseus": minor
 "@khanacademy/perseus-core": minor
-"@khanacademy/perseus-linter": minor
-"@khanacademy/perseus-score": minor
-"@khanacademy/perseus-utils": minor
-"@khanacademy/pure-markdown": minor
-"@khanacademy/simple-markdown": minor
 ---
 
-Update package to use modern 'exports' in package.json
+Add modern `"exports"` entry in package.json and export `item-splitting` separately
--- a/.github/actions/branch-compare/action.yml
+++ b/.github/actions/branch-compare/action.yml
@@ -18,6 +18,10 @@ outputs:
         description: "Path to the artifact generated for the PR branch"
     base-artifact-path:
         description: "Path to the artifact generated for the base branch"
+    diff-exit-code:
+        description: "Exit code of the diff command (0 if no changes, non-zero otherwise)"
+    diff-content:
+        description: "The diff content between the PR and base branch artifacts"
 
 runs:
     using: "composite"
@@ -66,3 +70,15 @@ runs:
               mkdir -p "$BASE_ARTIFACT_DIR"
               cp ${{ inputs.path-to-artifact }} "$BASE_ARTIFACT_DIR/"
               echo "base-artifact-path=$BASE_ARTIFACT_DIR" >> $GITHUB_OUTPUT
+
+        - name: Compare Artifacts
+          shell: bash
+          run: |
+              set +e
+              DELIM=$(uuidgen)
+              echo "DIFF_CONTENT<<$DELIM" >> "$GITHUB_OUTPUT"
+              diff --unified "$BASE_ARTIFACT_DIR/$(basename ${{ inputs.path-to-artifact }})" \
+                        "$PR_ARTIFACT_DIR/$(basename ${{ inputs.path-to-artifact }})" >> "$GITHUB_OUTPUT"
+              DIFF_EXIT_CODE=$?
+              echo "$DELIM" >> "$GITHUB_OUTPUT"
+              echo "diff-exit-code=$DIFF_EXIT_CODE" >> "$GITHUB_OUTPUT"
--- a/.github/actions/branch-compare/action.yml
+++ b/.github/actions/branch-compare/action.yml
@@ -75,7 +75,7 @@ runs:
 
         - uses: actions/upload-artifact@v4
           with:
-              name: pr-artifacts
+              name: base-artifacts
               path: ${{ steps.capture-base-artifacts.outputs.artifact-path }}
 
         - name: Compare Artifacts
--- a/.github/actions/branch-compare/action.yml
+++ b/.github/actions/branch-compare/action.yml
@@ -64,7 +64,14 @@ runs:
           id: compare-artifacts
           shell: bash
           run: |
+              # Don't stop execution if an error occurs, we'll capture the
+              # return value from `diff` and don't want it to blow up this
+              # Github action in cases where it finds differences (non-zero
+              # exit code).
               set +e
+              # We generate a delimeter as a UUID to avoid the chance of our
+              # delimeter showing up in the diff output. A new UUID from this
+              # tool is guaranteed to be unique over both space and time.
               DELIM=$(uuidgen)
               echo "diff-content<<$DELIM" >> "$GITHUB_OUTPUT"
               diff --unified "${{ steps.capture-base-artifacts.outputs.artifact-path }}" \
--- a/.github/actions/branch-compare/action.yml
+++ b/.github/actions/branch-compare/action.yml
@@ -12,6 +12,9 @@ inputs:
             `build-command` to `${{ runner.temp }}`/... to pass files in a temp
             directory.
         required: true
+    label-name:
+        description: "The name of the label to add or remove based on the diff result"
+        required: true
 
 outputs:
     pr-artifact-path:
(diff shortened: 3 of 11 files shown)


.changeset/fair-pumpkins-move.md
benchristel, line 1: These changeset files don't seem to be related to your changes. Are they from a different branch?

.changeset/rude-schools-collect.md
benchristel, line 1: These changeset files don't seem to be related to your changes. Are they from a different branch?

packages/perseus-score/src/index.ts
```diff
@@ -14,6 +14,7 @@ export {default as scoreInteractiveGraph} from "./widgets/interactive-graph/scor
 export {
     default as scoreLabelImage,
     scoreLabelImageMarker,
+    type InteractiveMarkerScore,
```
benchristel, line 17: Is this import used?

.github/workflows/node-ci.yml
```diff
@@ -245,102 +245,46 @@ jobs:
...
-
-                      <!---
-                      Comment ID: schema-change
-                      -->
+                  node-version: ${{ matrix.node-version }}
 
-            - name: Create or update Schema Change comment - failure
-              if: steps.check_for_schema_change.outputs.diff_exit_code != '0'
-              uses: peter-evans/create-or-update-comment@71345be0265236311c031f5c7866368bd1eff043
+            - name: Use branch-compare action to generate item splitting artifacts
```
benchristel, line 278: This step isn't generating artifacts, is it? Just checking for changes in the `split-perseus-item.js` artifact?

.github/actions/branch-compare/action.yml
```diff
@@ -0,0 +1,147 @@
...
+    label-name:
+        description: "The name of the label to add or remove based on the diff result"
+        required: true
+    comment-title-term:
+        description: >-
+            The term to include in the title of comments made by this action.
+    comment-title-emoji:
+        description: >-
+            The emoji to prefix comment titles with. Makes these comments
+            easily identifiable and differentiated.
```
benchristel: The `comment-title-emoji` is always used together with the `comment-title-term`. Why have a separate input for the emoji? Can't the `comment-title-term` contain emoji?

.github/actions/branch-compare/action.yml
```diff
@@ -0,0 +1,147 @@
...
+        - name: Check out base branch
+          uses: actions/checkout@v4
+          with:
+              # This checks out the base branch of the PR (the branch the PR
+              # is targeting)
+              ref: ${{ github.event.pull_request.base.ref }}
+
+        - name: Re-install node_modules for base branch
+          shell: bash
+          run: pnpm install
```
benchristel: The `build-command` is abstracted out, but here we hardcode a dependency on `pnpm`, which kind of limits the generality of this action. Why not have the caller of this action specify `pnpm install` as part of the build command?

```yaml
build-command: "pnpm install && pnpm build"
```

.github/actions/branch-compare/action.yml
```diff
@@ -0,0 +1,147 @@
...
+              issue-number: ${{ github.event.pull_request.number }}
+              comment-author: "github-actions[bot]"
+              body-includes: "Comment ID: ${{ inputs.comment-title-term }}"
+
+        - name: Create or update comment - success
+          if: steps.compare-artifacts.outputs.diff-exit-code == '0'
+          uses: peter-evans/create-or-update-comment@71345be0265236311c031f5c7866368bd1eff043
+          with:
+              issue-number: ${{ github.event.pull_request.number }}
+              comment-id: ${{ steps.find-comment.outputs.comment-id}}
```
benchristel, line 109: I suppose you've tested this, but... does this work if the comment is not found (i.e. doesn't exist yet?)

.github/actions/branch-compare/action.yml
```diff
@@ -0,0 +1,147 @@
...
+              mkdir -p "$BASE_ARTIFACT_DIR"
+              if [ -e "${{ inputs.path-to-artifact }}" ]; then
+                cp -R ${{ inputs.path-to-artifact }} "$BASE_ARTIFACT_DIR/"
+              fi
+              echo "artifact-path=$BASE_ARTIFACT_DIR" >> $GITHUB_OUTPUT
+
+        - uses: actions/upload-artifact@v4
+          with:
+              name: base-artifacts-${{ inputs.label-name }}
+              path: ${{ steps.capture-base-artifacts.outputs.artifact-path }}
```
benchristel: Why do we need to upload the artifact?

packages/perseus-score/src/index.ts
jeremywiebe: Yes. If you look at the changes in `packages/perseus/src/widgets/label-image/label-image.tsx` (below) we were doing a "source" import of this type. We need to export `InteractiveMarkerScore` from `perseus` in order to use it in `perseus-score`.  But... that gets me thinking.  We definitely shouldn't have a dependency on `perseus` from `perseus-score`, the whole point of `perseus-score` was to have a library that didn't depend on `react`!  

I'll see if I can move the type entirely (probably to `perseus-core`).

.changeset/fair-pumpkins-move.md
jeremywiebe: It is, but I'll update the description to be clearer. We need to export a type from `perseus` that is used in `perseus-score`. However, we shouldn't have a dependency in that direction so I'll re-work it and  update this changeset entry.

.github/actions/branch-compare/action.yml
jeremywiebe: Yes. In that case it creates a new comment.

.github/actions/branch-compare/action.yml
jeremywiebe: We don't, it was for debugging. I forgot to clean this up. Thanks for catching it.

packages/perseus-score/src/index.ts
benchristel: oh, I see that this is an export, not an import 🤦

.changeset/rude-schools-collect.md
jeremywiebe: Sorry, I had made changes to these packages as I was building out this PR and didn't back out this change.  There _is_ an `exports` change to `perseus-core`, but not the rest.

.github/actions/branch-compare/action.yml
benchristel: I'd suggest combining the two inputs into `comment-title`.

.github/actions/branch-compare/action.yml
jeremywiebe: I was on the fence as this repo is fully TypeScript/JS so doing `pnpm install` when changing branches feels necessary and implicit. I'm good with making it explicit in the places we use it though. Updated.

.github/workflows/node-ci.yml
jeremywiebe: Leftover debugging code. :(  Removed

packages/perseus-score/src/index.ts
jeremywiebe: EDIT: I was looking at this wrong. `perseus` is fine depending on `perseus-score`.  Ok, no change needed. :)

packages/perseus-core/package.json
```diff
@@ -18,6 +18,20 @@
...
+            "types": "./dist/index.d.ts",
+            "source": "./src/index.ts"
+        },
+        "./item-splitting": {
+            "import": "./dist/es/index.item-splitting.js",
+            "require": "./dist/index.item-splitting.js",
+            "types": "./dist/index.item-splitting.d.ts",
+            "source": "./src/index.item-splitting.ts"
+        }
+    },
```
jeremywiebe, line 34: `"exports"` is the modern (ahem, since Node v10) way to define entry points for an npm package ([docs](https://nodejs.org/api/packages.html#package-entry-points)). 

Note from the linked documentation (especially last sentence in **bold**): 

> The ["exports"](https://nodejs.org/api/packages.html#exports) provides a modern alternative to ["main"](https://nodejs.org/api/packages.html#main) allowing multiple entry points to be defined, conditional entry resolution support between environments, and preventing any other entry points besides those defined in ["exports"](https://nodejs.org/api/packages.html#exports). This encapsulation allows module authors to clearly define the public interface for their package.
> 
> For new packages targeting the currently supported versions of Node.js, the ["exports"](https://nodejs.org/api/packages.html#exports) field is recommended. For packages supporting Node.js 10 and below, the ["main"](https://nodejs.org/api/packages.html#main) field is required. **If both ["exports"](https://nodejs.org/api/packages.html#exports) and ["main"](https://nodejs.org/api/packages.html#main) are defined, the ["exports"](https://nodejs.org/api/packages.html#exports) field takes precedence over ["main"](https://nodejs.org/api/packages.html#main) in supported versions of Node.js.**

.github/actions/branch-compare/action.yml
```diff
@@ -0,0 +1,136 @@
...
+          with:
+              issue-number: ${{ github.event.pull_request.number }}
+              comment-id: ${{ steps.find-comment.outputs.comment-id}}
+              edit-mode: replace
+              body: |
+                  # ${{ inputs.comment-title }}: No Changes ✅
+
+                  <!---
+                  Comment ID: ${{ inputs.comment-title }}
+                  -->
```
jeremywiebe, line 116: This trick allows us to embed some "structured" information into the PR comment that is not visible in the Github UI. This makes it possible to find this comment in the `Find comment` step and be sure we found the right Github comment (ie. avoids false positives). 

Thanks to @handeyeco for the idea!

PR #2720, opened 2025-07-18
will convert any relative url to always use khanacademy legacy as the base but won't convert absolutes

marnikostman:
## Summary:
We want perseus content to route to legacy domain (rather than classroom) subdomain. Was able to confirm outputting as expected in storybook but unable to get local dev to use this version of perseus to confirm on local build of webapp. Will ask about how to do this Monday but headed out for evening. Tried to build using this: [this](https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/2072412231/Testing+Perseus+locally+in+Mobile) but didn't seem to take as even logs weren't outputting. 

Issue: [TUT-2621](https://khanacademy.atlassian.net/browse/TUT-2621)

## Test plan:

[TUT-2621]: https://khanacademy.atlassian.net/browse/TUT-2621?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/perseus/src/renderer.tsx
+++ b/packages/perseus/src/renderer.tsx
@@ -1142,6 +1142,16 @@ class Renderer
         const apiOptions = this.getApiOptions();
         const imagePlaceholder = apiOptions.imagePlaceholder;
 
+        if (node.type === "link") {
+            // allow this to reroute on either .dev or .org
+            const legacyOrigin = new URL(document.baseURI).origin.replace(
+                "classroom.",
+                "",
+            );
+            node.target = new URL(node.target, legacyOrigin).href;
+            console.log(node.target);
+        }
+
         if (node.type === "widget") {
             const widgetPlaceholder = apiOptions.widgetPlaceholder;
 
--- a/packages/perseus/src/renderer.tsx
+++ b/packages/perseus/src/renderer.tsx
@@ -1149,7 +1149,6 @@ class Renderer
                 "",
             );
             node.target = new URL(node.target, legacyOrigin).href;
-            console.log(node.target);
         }
 
         if (node.type === "widget") {


jeremywiebe (changes requested):
Unfortunately, we don't want to embed knowledge specific to Khan Academy's domain/url structures into Perseus as there are other places that Perseus is hosted. 

That said, I think we can do this type of work by customizing the Markdown parsing rule for `link` in the classroom app. Tony is familiar with customizing Markdown parsing but I'd be happy to help you identify where and how this would look if you want a hand.

marnikostman:
> Unfortunately, we don't want to embed knowledge specific to Khan Academy's domain/url structures into Perseus as there are other places that Perseus is hosted.
> 
> That said, I think we can do this type of work by customizing the Markdown parsing rule for `link` in the classroom app. Tony is familiar with customizing Markdown parsing but I'd be happy to help you identify where and how this would look if you want a hand.

Thanks Jeremy, I'll ask Tony.

tony-dinh:
Thanks Marni for picking up this ticket!

>  I think we can do this type of work by customizing the Markdown parsing rule for link in the classroom app.

I don't have deep knowledge on the Perseus renderers. Is it the case that the markdown parsing happens within the renderers @jeremywiebe? If we wanted to solve this at a markdown level, would that not need to happen in the `Perseus.ArticleRenderer` from within the perseus package? 

That being said, I don't know if that is the only path forward we have. I don't have any fully baked thoughts on a solution but I wonder if it'd make sense to expose an optional callback to override link behaviors in Perseus article renderers or something to keep perseus internals platform agnostic

jeremywiebe:
You're right @tony-dinh.  We can't customize the parsing (that's baked into the Markdown parsing setup that is internal to Perseus. 

But, I think we can customize the `Link` on `apiOptions.baseElements` to adjust the URL however we want. 

Something like this (also DM'd with @marnikostman and shared this there):

```ts
const apiOptions = {
    ... // whaatever exists here already
    baseElements: {
        Link: (props) => {
            const legacyOrigin = new URL(
                document.baseURI,
            ).origin.replace("classroom.", "");
            const href = new URL(props.href, legacyOrigin).href;

            return <a {...props} href={href} />;
        },
    },
};

return <ServerItemRenderer 
    ...
    apiOptions={apiOptions}
/>
```

marnikostman:
> ```ts
> baseElements: {
> ```

@jeremywiebe These can't be passed in from frontend, correct? We'd just export another component type from Perseus? Would we need do this for all 3 renderers? (Server/Article/Hints)

jeremywiebe:
> > ```ts
> > baseElements: {
> > ```
> 
> @jeremywiebe These can't be passed in from frontend, correct? We'd just export another component type from Perseus? Would we need do this for all 3 renderers? (Server/Article/Hints)

I think we discussed this, but these would be changes to `Khan/frontend`, not Perseus.

PR #2726, opened 2025-07-22
Schema change detection - change diff format

jeremywiebe:
## Summary:

A small PR to tweak the diff format used for schema change notices. Switching to `--unified` diffs makes for much clearer output.

Issue: <none>

## Test plan:

Compare output. Note that the change of `range` from `number[]` to `string[]` is split up in the **Before** but is together in the **After** (seems much more readable (to me at least)). 

| Before | After | 
| -- | -- |
| image | image |

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -279,7 +279,7 @@ jobs:
                   # delimeter in the diff output doesn't cause problems.
                   DELIM=$(uuidgen)
                   echo "SCHEMA_DIFF<<$DELIM" >> "$GITHUB_OUTPUT"
-                  diff -u --context 3 ${{ runner.temp }}/schema-base.d.ts ${{ runner.temp }}/schema-pr.d.ts >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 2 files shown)


.github/workflows/node-ci.yml
```diff
@@ -279,7 +279,7 @@ jobs:
                   # delimeter in the diff output doesn't cause problems.
                   DELIM=$(uuidgen)
                   echo "SCHEMA_DIFF<<$DELIM" >> "$GITHUB_OUTPUT"
-                  diff -C 3 ${{ runner.temp }}/schema-base.d.ts ${{ runner.temp }}/schema-pr.d.ts >> "$GITHUB_OUTPUT"
+                  diff --unified ${{ runner.temp }}/schema-base.d.ts ${{ runner.temp }}/schema-pr.d.ts >> "$GITHUB_OUTPUT"
```
jeremywiebe, line 282: This gives us a much nicer diff to look at and `--unified`, by default, gives us 3 lines of context.

mark-fitzgerald (approved):
Totally agree - much easier to read. Thank you!

PR #2659, opened 2025-06-26
[Radio] | (CX) | Stop showing NOTA button if NOTA choice exists in the editor

nishasy:
## Summary:
Right now, the editor allows the ability to add multiple "none of the above" choices.

In this PR, I'm hiding the "Add none of the above" button if there is already
a "none of the above" answer choice.

Issue: https://khanacademy.atlassian.net/browse/LEMS-3229

## Test plan:
`pnpm jest packages/perseus-editor/src/widgets/__tests__/radio-editor.test.tsx`

Storybook
- http://localhost:6006/?path=/story/perseuseditor-editorpage-widgets-radio--multi-choice

PR #2727, opened 2025-07-22
[Radio] | (CX) | Stop showing NOTA button if NOTA choice exists in the editor

nishasy:
## Summary:
Redo of https://github.com/Khan/perseus/pull/2659 because it got messed up in an accidental merge.

Small update to new editor UI.

Right now, the editor allows the ability to add multiple "none of the above" choices.

In this PR, I'm hiding the "Add none of the above" button if there is already
a "none of the above" answer choice.

Issue: https://khanacademy.atlassian.net/browse/LEMS-3229

## Test plan:
`pnpm jest packages/perseus-editor/src/widgets/__tests__/radio-editor.test.tsx`

Storybook
- http://localhost:6006/?path=/story/perseuseditor-editorpage-widgets-radio--multi-choice

### DIFF
--- a/packages/perseus-editor/src/widgets/__tests__/radio-editor.test.tsx
+++ b/packages/perseus-editor/src/widgets/__tests__/radio-editor.test.tsx
@@ -236,6 +236,28 @@ describe("radio-editor", () => {
         expect(onChangeMock).not.toBeCalled();
     });
 
+    it("shows the 'None of the above' button when there is no 'None of the above' choice", async () => {
+        renderRadioEditor();
+        expect(
+            screen.getByRole("button", {name: "None of the above"}),
+        ).toBeInTheDocument();
+    });
+
+    it("hides the 'None of the above' button when there is a 'None of the above' choice", async () => {
+        renderRadioEditor(() => {}, {
+            choices: [
+                {content: "Choice 1", isNoneOfTheAbove: false},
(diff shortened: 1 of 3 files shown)


ivyolamit (approved):
LGTM 👍🏼

packages/perseus-editor/src/widgets/__tests__/radio-editor.test.tsx
```diff
@@ -236,6 +236,28 @@ describe("radio-editor", () => {
...
+                {content: "Choice 3", isNoneOfTheAbove: false},
+                {content: "None of the above", isNoneOfTheAbove: true},
+            ],
+            hasNoneOfTheAbove: true,
+        });
+        expect(
+            screen.queryByRole("button", {name: "None of the above"}),
+        ).not.toBeInTheDocument();
+    });
+
```
anakaren-rojas, line 260: Can we do one more test that covers the re-appearance of the button after the NOTA choice is removed?

nishasy: I don't think that's possible with our current jest infrastructure, because it would just re-render the component without the NOTA option in the new props, which is already covered by the first test.

mark-fitzgerald (approved):

PR #2731, opened 2025-07-24
Bugfix: fix hints in GradedGroups

handeyeco:
## Summary:

So this was a little weird. I went through the repro steps, got the JSON, copy/pasted it, and gave it a `PerseusArticle` type. According to the type, the editor was making invalid JSON:

``` JSON
"hint": {
  "content": "((Hint))"
},
```

A hint is supposed to be a `PerseusRenderer` and that requires `widgets` and `images`. I think this was a long-time bug that just never showed itself:

``` TS
// hint-editor.tsx
// note it's a full PerseusRenderer 
const hints = this.props.hints.concat([
  {content: "", images: {}, widgets: {}},
]);

// graded-group-editor.tsx
// note it's an incomplete PerseusRenderer
const hint = {content: ""} as const;
```

Anyway, the reason this bug manifested now is: `PerseusRenderer` is supposed to have `widgets` and my new code was expecting `PerseusRenderer` to have `widgets`. So I also made my code more resilient.

Issue: LEMS-3321

## Test plan:

1. Navigate to [Khan Academy](https://www.khanacademy.org/devadmin/content/articles/dsat--practice-test--10-m-2/x90fb23c3f0cae50d#) 
2. Scroll down to Question 2
3. Add a hint to the Graded Group widget 
4. Type some text in the hint input (with no widgets)
5. In the preview, click Explain

Before: we would throw
After: we should not throw

I also added unit tests.

### DIFF
--- a/packages/perseus-editor/src/hint-editor.tsx
+++ b/packages/perseus-editor/src/hint-editor.tsx
@@ -25,7 +25,11 @@ import type {
     DeviceType,
     ImageUploader,
 } from "@khanacademy/perseus";
-import type {Hint, PerseusWidgetsMap} from "@khanacademy/perseus-core";
+import type {
+    Hint,
+    PerseusRenderer,
+    PerseusWidgetsMap,
+} from "@khanacademy/perseus-core";
 
 const {InfoTip, InlineIcon} = components;
 
@@ -366,9 +370,8 @@ class CombinedHintsEditor extends React.Component<CombinedHintsEditorProps> {
     };
 
     addHint: () => void = () => {
-        const hints = this.props.hints.concat([
-            {content: "", images: {}, widgets: {}},
-        ]);
+        const hint: PerseusRenderer = {content: "", images: {}, widgets: {}};
+        const hints = [...this.props.hints, hint];
         this.props.onChange({hints: hints}, () => {
             const i = hints.length - 1;
             // eslint-disable-next-line react/no-string-refs
--- a/packages/perseus-editor/src/widgets/__tests__/graded-group-editor.test.tsx
+++ b/packages/perseus-editor/src/widgets/__tests__/graded-group-editor.test.tsx
@@ -1,0 +1,53 @@
+import {Dependencies} from "@khanacademy/perseus";
+import {render, screen} from "@testing-library/react";
+import {userEvent as userEventLib} from "@testing-library/user-event";
+import * as React from "react";
+
+import {testDependencies} from "../../../../../testing/test-dependencies";
+import GradedGroupEditor from "../graded-group-editor";
+
+import type {PerseusRenderer} from "@khanacademy/perseus-core";
+import type {UserEvent} from "@testing-library/user-event";
+
+describe("GradedGroupEditor", () => {
+    let userEvent: UserEvent;
+    beforeEach(() => {
+        userEvent = userEventLib.setup({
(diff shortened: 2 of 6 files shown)


packages/perseus-editor/src/hint-editor.tsx
```diff
@@ -366,9 +370,8 @@ class CombinedHintsEditor extends React.Component<CombinedHintsEditorProps> {
     };
 
     addHint: () => void = () => {
-        const hints = this.props.hints.concat([
-            {content: "", images: {}, widgets: {}},
-        ]);
+        const hint: PerseusRenderer = {content: "", images: {}, widgets: {}};
+        const hints = [...this.props.hints, hint];
```
jeremywiebe, line 374: I _think_ that this worked historically because the `Renderer` defaults `widgets` and `images` (it still does).  So, it seems that some place between the Graded Group Editor and the Renderer we now look at the hint data and then fail.

packages/perseus-editor/src/widgets/__tests__/graded-group-editor.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+    it("should render", async () => {
+        render(<GradedGroupEditor onChange={() => undefined} />);
+
+        expect(
+            await screen.findByText("Graded Groups should contain a prompt"),
+        ).toBeInTheDocument();
+    });
+
+    /**
+     * Regression LEMS-3321: make sure a hint is a full PerseusRenderer
```
jeremywiebe, line 33: **nit**: I don't _think_ the regression is on the editor here. Rather, some other part of our system assumes that the `hint` is a full-blown `PerseusRenderer`.  Given this is a bug inside the editing system I guess we can't solve this in the parser either. 🤔

packages/perseus/src/user-input-manager.tsx
```diff
@@ -46,10 +46,15 @@ type Props = {
...
 export function sharedInitializeUserInput(
-    widgetOptions: PerseusWidgetsMap,
+    widgetOptions: PerseusWidgetsMap | undefined,
     problemNum: number,
 ): UserInputMap {
     const startUserInput: UserInputMap = {};
+
+    if (!widgetOptions) {
+        return startUserInput;
+    }
```
jeremywiebe, line 56: I _think_ this is the core of the regression.

packages/perseus-editor/src/widgets/__tests__/graded-group-editor.test.tsx
handeyeco: > Rather, some other part of our system assumes that the hint is a full-blown PerseusRenderer

It's in the type:

```
export type PerseusGradedGroupWidgetOptions = {
    // ...
    hint?: PerseusRenderer | null | undefined;
    // ...
};

export type PerseusRenderer = {
    content: string;
    widgets: PerseusWidgetsMap;
    images: {
        [imageUrl: string]: PerseusImageDetail;
    };
};
```

So either the type is wrong or the existing content is malformed.

jeremywiebe: I'd say the type is wrong. We've been trying to retrofit types onto untyped JS for probably 5 or more years and still keep finding areas where the code behaves differently than our types say.

PR #2734, opened 2025-07-24
[Radio] | (CX) | Add dynamically resizing text areas to Radio editor UI

nishasy:
## Summary:
We got a request from content authors to dynamically resize the text area as they type
so that they don't have to manually resize the text area all the time.

There is actually already a ticket for this in Wonder Blocks for Wonder Blocks TextArea,
but I have this quick solution in the meantime. I put a TODO to make sure that we use WB
whenever it's ready.

Issue: https://khanacademy.atlassian.net/browse/LEMS-3327

## Before:

https://github.com/user-attachments/assets/ad67e558-e0db-474d-9f3a-38d51ba4faa4

## After:

https://github.com/user-attachments/assets/897470e2-e051-48df-8347-ab9a53c1c75e

## Test plan:
Unfortunately there isn't really a way to unit test styles...?

Storybook
- http://localhost:6006/?path=/docs/widgets-radio-editor-demo--docs#single-choice-2
- Type a bunch of stuff in `content` and `rationale` text areas
- Confirm that they dynamically resize with new lines
- select all and delete everything
- Confirm that the text areas dynamically resize back down to one row

### DIFF
--- a/packages/perseus-editor/src/widgets/radio/auto-resizing-text-area.tsx
+++ b/packages/perseus-editor/src/widgets/radio/auto-resizing-text-area.tsx
@@ -1,0 +1,40 @@
+import {TextArea} from "@khanacademy/wonder-blocks-form";
+import * as React from "react";
+
+import type {PropsFor} from "@khanacademy/wonder-blocks-core";
+
+type TextAreaProps = Omit<PropsFor<typeof TextArea>, "rows">;
+
+// TODO(3326): Remove this whole component once WB TextArea supports
+// dynamic resizing.
+export const AutoResizingTextArea = (props: TextAreaProps) => {
+    const textAreaRef = React.useRef<HTMLTextAreaElement>(null);
+
+    React.useEffect(() => {
+        function adjustHeight() {
+            const textArea = textAreaRef.current;
+            if (textArea) {
+                // Reset the height to get the correct scrollHeight.
+                // 42px is the default height of the Wonder Blocks TextArea
+                // when it has one row.
+                textArea.style.height = "42px";
+                // Set height to scrollHeight to fit all content.
+                textArea.style.height = `${textArea.scrollHeight}px`;
+            }
+        }
+
+        adjustHeight();
+    }, [props.value]);
+
+    return (
+        <TextArea
+            {...props}
+            ref={textAreaRef}
+            resizeType="none"
+            style={{
+                overflow: "hidden",
+                ...props.style,
+            }}
+        />
+    );
+};
--- a/packages/perseus-editor/src/widgets/radio/auto-resizing-text-area.tsx
+++ b/packages/perseus-editor/src/widgets/radio/auto-resizing-text-area.tsx
@@ -11,19 +11,15 @@ export const AutoResizingTextArea = (props: TextAreaProps) => {
     const textAreaRef = React.useRef<HTMLTextAreaElement>(null);
 
     React.useEffect(() => {
-        function adjustHeight() {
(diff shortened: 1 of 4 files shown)


packages/perseus-editor/src/widgets/radio/radio-option-settings.tsx
```diff
@@ -37,6 +37,9 @@ export function RadioOptionSettings({
     onMove,
 }: RadioOptionSettingsProps) {
     const {content, rationale, correct, isNoneOfTheAbove} = choice;
+    const uniqueId = React.useId();
+    const contentTextAreaId = `${uniqueId}-content-textarea`;
+    const rationaleTextAreaId = `${uniqueId}-rationale-textarea`;
```
nishasy, line 42: For some reason it forced me to use IDs with the label once I stopped using a Wonder Blocks TextArea 🤷🏽‍♀️

packages/perseus-editor/src/widgets/radio/auto-resizing-text-area.tsx
```diff
@@ -0,0 +1,40 @@
...
+    const textAreaRef = React.useRef<HTMLTextAreaElement>(null);
+
+    React.useEffect(() => {
+        function adjustHeight() {
+            const textArea = textAreaRef.current;
+            if (textArea) {
+                // Reset the height to get the correct scrollHeight.
+                // 42px is the default height of the Wonder Blocks TextArea
+                // when it has one row.
+                textArea.style.height = "42px";
```
mark-fitzgerald: Why is this reset needed if the height is set on line 22?

packages/perseus-editor/src/widgets/radio/auto-resizing-text-area.tsx
```diff
@@ -0,0 +1,40 @@
...
+                // Set height to scrollHeight to fit all content.
+                textArea.style.height = `${textArea.scrollHeight}px`;
+            }
+        }
+
+        adjustHeight();
```
mark-fitzgerald: Why is the function built within the `useEffect` and then executed? I'm missing the purpose of this technique.

packages/perseus-editor/src/widgets/radio/radio-option-settings.tsx
mark-fitzgerald: The content text area was moved outside of the WB heading, so maybe that's the reason? Did it also complain about the rationale text area (which stayed inside its label)?

nishasy: Oh it complained about it while the new text area was within the label. It didn't make sense to keep it in there after adding IDs, so I moved it out after.

packages/perseus-editor/src/widgets/radio/auto-resizing-text-area.tsx
nishasy: It needs to reset the height so it can figure out what the _real_ scroll height is, which is what's used on line 22. Otherwise you get super wonky behavior like this.

https://github.com/user-attachments/assets/e8c8a744-bc9d-4997-bab9-b75a92161f93

packages/perseus-editor/src/widgets/radio/auto-resizing-text-area.tsx
nishasy: I had it outside and then moved it inside when there was a lint error for it. Updating.

ivyolamit (approved):
@nishasy changes here looks logical to me.

PR #2735, opened 2025-07-24
Sync dependencies

SonicScrewdriver:
## Summary:
<Add your summary here>

Issue: LEMS-XXXX

## Test plan:


pnpm-workspace.yaml
```diff
@@ -1,7 +1,7 @@
 # NOTE: The `devDeps` and `peerDeps` catalogs in this file are
-# generated from khan/frontend's package.json. To update them, run:
+# generated from khan/frontend's pnpm-workspace.yaml. To update them, run:
 #
-#     utils/sync-dependencies.ts ../frontend/package.json
+#     utils/sync-dependencies.ts ../frontend/pnpm-workspace.yaml
```
SonicScrewdriver, line 4: This is the new location we need to be referencing - I've just updated the documentation for this so that no one else should get confused by script failures!

PR #2740, opened 2025-07-25
Remove JSON mode checkbox from EditorPage

benchristel:
While investigating how we might parse/migrate PerseusItem data before emitting
it from the editor, I noticed this weird `jsonMode` param in the `onChange`
callback. Nothing in webapp seems to be using it, so I'd just as soon remove
it. It's a footgun.

Issue: none

## Test plan:

```
pnpm storybook
open http://localhost:6006
```

View the EditorPage stories.

### DIFF
--- a/.changeset/green-spiders-jam.md
+++ b/.changeset/green-spiders-jam.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": patch
+---
+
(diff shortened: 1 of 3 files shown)


jeremywiebe (approved):
I was going to ask for changes because I thought we still needed to render the JSON switch/checkbox, but I see that's been migrated up into the new exercise chrome! 

Good to go! Thanks for removing this bit of now-unused code!

PR #2743, opened 2025-07-25
Return EXTRA_SYMBOLS_ERROR if expression widget input can't be parsed

benchristel:
Requested in this Slack thread: https://khanacademy.slack.com/archives/C07K2C44LDB/p1753467767519219

Issue: LEMS-3333

## Test plan:

On a ZND, type some random symbols, e.g. backticks, into an expression widget
in an exercise.  Click the "check answer" button. You should get a validation
error message "We could not understand your answer. Please check your answer
for extra text or symbols." You should be able to change your answer and
resubmit.

### DIFF
--- a/packages/perseus-score/src/widgets/expression/score-expression.ts
+++ b/packages/perseus-score/src/widgets/expression/score-expression.ts
@@ -45,8 +45,7 @@ function scoreExpression(
         decimal_separator: getDecimalSeparator(locale),
     });
 
-    const userInputExpression = KAS.parse(userInput, options);
-    if (!userInputExpression.parsed) {
+    if (!KAS.parse(userInput, options).parsed) {
         return {
             type: "invalid",
             message: ErrorCodes.EXTRA_SYMBOLS_ERROR,


packages/perseus-score/src/util/answer-types.ts
```diff
@@ -783,11 +783,10 @@ const KhanAnswerTypes = {
                     // is that `message` only will be shown if the guess is
                     // graded as empty for every solution.
                     ungraded: false,
-                } as const;
+                };
 
                 // Don't bother parsing an empty input
                 if (!guess) {
-                    // @ts-expect-error - TS2540 - Cannot assign to 'empty' because it is a read-only property.
```
benchristel, line 790: I don't know why we added `as const` to `score` and then added all these `ts-expect-error`s to make up for it. 🤷

Myranae: Seems like it's better to have it with the type like you added since part gets modified. Or maybe it would be better to mark the fields that stay as readonly as readonly and leave the one that gets modified without? Pretty sure score gets modified through user interaction, so it probably doesn't need the readonly bits. Yay for removing ts-expect-errors!

packages/perseus-score/src/util/answer-types.ts
```diff
@@ -860,14 +853,11 @@ const KhanAnswerTypes = {
...
                         );
                         if (resultX.equal) {
-                            // @ts-expect-error - TS2540 - Cannot assign to 'ungraded' because it is a read-only property.
                             score.ungraded = true;
-                            // @ts-expect-error - TS2540 - Cannot assign to 'message' because it is a read-only property.
                             score.message =
                                 ErrorCodes.MULTIPLICATION_SIGN_ERROR;
                         } else if (resultX.message) {
                             // TODO(aasmund): I18nize `score.message`
-                            // @ts-expect-error - TS2540 - Cannot assign to 'message' because it is a read-only property.
```
Myranae, line 870: Wow, that one change cleared a lot of these ts-expect-errors! 🎉

packages/perseus-score/src/widgets/expression/score-expression.ts
```diff
@@ -44,6 +45,13 @@ function scoreExpression(
         decimal_separator: getDecimalSeparator(locale),
     });
 
+    if (!KAS.parse(userInput, options).parsed) {
+        return {
+            type: "invalid",
+            message: ErrorCodes.EXTRA_SYMBOLS_ERROR,
+        };
+    }
```
Myranae, line 53: Is there any reason other than extra symbols that the input wouldn't parse?

benchristel: We could try to get more detailed with the error message, but:

- We want to reuse the existing error code since the message for it is already translated.
- Making the errors more detailed would probably require changes to `KAS.parse`, which is pretty arcane and hasn't been touched for a decade or more.
- `EXTRA_SYMBOLS_ERROR` is pretty general already. The message is: "We could not understand your answer. Please check your answer for extra text or symbols." That pretty much covers all the bases, since if the learner removes all "text and symbols" from the input, what's left will either be a number (which is always valid), or empty (which has a different error code).

All that's to say, I don't think we should try to add more detailed error handling in this PR.

Myranae (approved):
Image

Image

Was able to see the error message in Perseus Storybook and also in local dev on a testing branch in frontend. Looks like it's working!

packages/perseus-score/src/widgets/expression/score-expression.ts
handeyeco: What about looking at it from a different angle: if we don't want to get specific with different types of errors, could we make this error more generic? Rather than `EXTRA_SYMBOLS_ERROR`, maybe `PARSING_ERROR`?

PR #2745, opened 2025-07-29
[Radio] Refactor multiple choice component (logic and UI updates)

mark-fitzgerald:
## Summary:
This PR refactors the logic, HTML, and CSS for the top-level UI choice component. This component was simplified to just handle the display of the choice group content and its styling. Accessibility features will be added after the component is refactored so that it can be applied in a holistic manner across the various files.

Issue: LEMS-3300

## Test plan:
1. Launch Storybook locally
1. Navigate to Widgets => RadioNew => Docs => [Single Select](http://localhost:6006/?path=/docs/widgets-radionew--docs#single-select)
   - Visually, the radio button group now matches the new specs.
   - After checking the answer, the review mode state of the options matches the new specs.
   - Functionality of the widget remains the same as the old component.


packages/perseus/src/widgets/radio/__tests__/choice-indicator.test.tsx
```diff
@@ -139,15 +139,6 @@ describe("Multiple choice indicator", () => {
...
             expect(mockClickHandler).toHaveBeenCalledTimes(1);
-            expect(mockClickHandler).toHaveBeenCalledWith(true);
-
-            // Click again to toggle back
-            mockClickHandler.mockReset();
-            act(() => {
-                screen.getByRole("button").click();
-            });
-            expect(mockClickHandler).toHaveBeenCalledTimes(1);
-            expect(mockClickHandler).toHaveBeenCalledWith(false);
```
mark-fitzgerald, line 150: This extra step in the test is no longer needed since the logic that interacts with the click handler no longer tracks an internal "checked" state (it's now handled by the component).

packages/perseus/src/widgets/radio/__tests__/multiple-choice-component.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+    hasRationale: false,
+    showRationale: false,
+    showCorrectness: false,
+    correct: false,
+    isNoneOfTheAbove: false,
+    previouslyAnswered: false,
+    revealNoneOfTheAbove: false,
+    disabled: false,
+};
+
```
mark-fitzgerald, line 30: This test suite is incomplete and will be updated soon.

packages/perseus/src/widgets/radio/__tests__/multiple-choice.test.ts
```diff
@@ -26,7 +26,7 @@ import type {UserEvent} from "@testing-library/user-event";
 const createApiOptions = (options: Partial<APIOptions> = {}): APIOptions => ({
     ...options,
     flags: {
-        "new-radio-widget": true,
+        "new-radio-widget": false,
```
mark-fitzgerald, line 29: These tests need to be evaluated in context of the new UI & logic files. Resetting back to original UI so that these tests aren't failing.

packages/perseus/src/widgets/radio/choice-option.temp.new.tsx
```diff
@@ -0,0 +1,273 @@
+/* eslint-disable react/no-unsafe */
```
mark-fitzgerald: This file will be removed. Do not review.

packages/perseus/src/perseus-api.tsx
```diff
@@ -158,13 +158,16 @@ export const ClassNames = {
     RESPONSIVE_RENDERER: "perseus-renderer-responsive",
     INPUT: "perseus-input",
     FOCUSED: "perseus-focused",
+    // TODO: (START) Remove these entries, and corresponding styles in perseus/src/styles/widgets/radio.css,
```
ivyolamit: 🧹 suggestion: add a ticket number here so it will be easier to search during cleanup.

packages/perseus/src/widgets/radio/__tests__/multiple-choice.test.ts
ivyolamit: in line 25 it has the comment
```
// Create API options with the new-radio-widget flag enabled
```

do we want to do the reset in a afterEach or after block instead?

packages/perseus/src/widgets/radio/choice-indicator.module.css
```diff
@@ -58,6 +62,9 @@
        By setting the width, height, and margin to an em unit,
            we don't have to worry about adjusting these if we change the font size of the letter,
            and it will scale with any user-based font-size changes.
+       The height and width of the icon is calculated based upon the revised specs,
+           (https://www.figma.com/design/BokB9NfT0ntq5sMfOAyhJQ/Radio-Widget?node-id=152-1017&t=fJjnnrNse62NOL4g-1)
```
ivyolamit: we probably don't want to add a private link here since this is a public code.

packages/perseus/src/widgets/radio/choice-indicator.new.tsx
```diff
@@ -2,7 +2,6 @@ import {PhosphorIcon} from "@khanacademy/wonder-blocks-icon";
 import checkIcon from "@phosphor-icons/core/bold/check-bold.svg";
```
ivyolamit, line 2: this doesn't need to be in this PR, just code organization suggestion: radio widget is different from other widgets since it has multiple internal components available, what do you think of adding all new internal components under `widgets/radio/components`?

packages/perseus/src/widgets/radio/choice.module.css
```diff
@@ -1,73 +1,161 @@
+:global(.framework-perseus) {
+    /* OVERRIDES
+       These styles should be removed when either layering control within Perseus is available,
```
ivyolamit, line 3: let's a TODO ticket number here

packages/perseus/src/widgets/radio/choice.module.css
```diff
@@ -1,73 +1,161 @@
...
-    &:hover,
-    &:has(+ .choice:hover) {
-        border-bottom: var(--wb-border-width-medium) solid
-            var(--wb-semanticColor-core-border-neutral-default);
-        padding-bottom: calc(
-            var(--perseus-multiple-choice-spacing) - var(--wb-sizing-size_010)
-        ); /* Account for increased border width */
-    }
+/*****************
+ BORDER MANAGEMENT
```
ivyolamit, line 37: ❤️👌

packages/perseus/src/widgets/radio/multiple-choice-component.new.tsx
```diff
@@ -72,208 +37,102 @@ interface MultipleChoiceComponentProps {
  * Created as part of the Radio Revitalization Project (LEMS-2933).
  */
 const MultipleChoiceComponent = ({
-    apiOptions,
```
ivyolamit, line 75: Just checking, since it's not clear to me yet, these props are no longer necessary? or where are they handled now?

ivyolamit (commented):
@mark-fitzgerald great work here! it's all coming together now 👏🏼 🎉 
Here's my first round of review, i'll do my second pass later this afternoon, after my meetings 🙏🏼

packages/perseus/src/perseus-api.tsx
mark-fitzgerald: Added.

packages/perseus/src/widgets/radio/choice-indicator.new.tsx
mark-fitzgerald: I think that once we fully implement these changes, the file structure will be greatly simplified, and it will be much easier to navigate among the files. I think we will end up with just 4 tsx files, and 3 css files. That seems manageable to me.

packages/perseus/src/widgets/radio/choice.module.css
mark-fitzgerald: I don't have a ticket for this. However, I do plan on revisiting this soon after this project is complete, and this will be my test case.

packages/perseus/src/widgets/radio/multiple-choice-component.new.tsx
mark-fitzgerald: Since we've simplified a lot of the work that each part of the component does, many of the properties in `apiOptions` are no longer needed. Originally, we had to pass down a lot of the information in `apiOptions` in order for the sub-components to figure out how to work. That's all been lifted up higher in the component structure, now. The other properties that have been removed were used for functionality that is no longer relevant. For instance, handling scratchpad interaction had to be managed in the widget, but now it click, drag, hover events are all handled by the scratchpad and no longer accidentally impact this widget.

packages/perseus/src/widgets/radio/__tests__/multiple-choice-component.test.tsx
```diff
@@ -0,0 +1,260 @@
...
+                    choices={[baseChoiceValues]}
+                    countChoices={false}
+                    numCorrect={1}
+                    onChoiceChange={() => {}}
+                    reviewMode={false}
+                />,
+            );
+            const classes = Array.from(screen.getByRole("list").classList);
+            expect(classes).toHaveLength(1);
+            expect(classes).toContain(`choice-list`);
```
ivyolamit, line 64: @mark-fitzgerald what do you think of a snapshot for styling tests? Since styling has specific values or more of visual while checking for classes here has some value but it doesn't check if someone accidentally changes some padding or other values.

packages/perseus/src/widgets/radio/__tests__/multiple-choice-component.test.tsx
```diff
@@ -0,0 +1,260 @@
...
+describe("Multiple choice component", () => {
+    let choiceSpy: jest.SpyInstance;
+
+    beforeEach(() => {
+        choiceSpy = jest.spyOn(Choice, "default");
+    });
+
+    describe("styling", () => {
+        it("applies expected styling by default", () => {
+            render(
```
ivyolamit: what do you think of simplifying/DRY-ing the tests?

suggestion
```
    // line 28
    const baseMultipleChoiceProps = {
        choices: [baseChoiceValues],
        countChoices: false,
        numCorrect: 1,
        onChoiceChange: () => {},
        reviewMode: false,
    };
    function setup({
        choices,
        countChoices,
        numCorrect,
        onChoiceChange,
        reviewMode,
    }) {
        return render(
            <MultipleChoiceComponent
                choices={choices}
                countChoices={countChoices}
                numCorrect={numCorrect}
                onChoiceChange={onChoiceChange}
                reviewMode={reviewMode}
            />,
        );
    }
```

then in the test it would look like below, instead of repeating the rendering code in each test
```
setup(baseMultipleChoiceProps);
```

for some cases it would look like:
```
            const rationaleContent = "foo-bar-zot";
            const choiceValues = {
                ...baseChoiceValues,
                rationale: rationaleContent,
                hasRationale: true,
            };
            setup({
                ...baseMultipleChoiceProps,
                choices: [choiceValues],
                reviewMode: true,
            });
```

packages/perseus/src/widgets/radio/__tests__/multiple-choice-component.test.tsx
```diff
@@ -0,0 +1,260 @@
...
+                />,
+            );
+            expect(
+                screen.queryByText(rationaleContent),
+            ).not.toBeInTheDocument();
+        });
+    });
+
+    describe("functionality", () => {
+        it("calls the 'onChoiceChange' callback when a choice is activated", () => {
```
ivyolamit, line 216: suggestion (when DRY-ing the test)
```
        it("calls the 'onChoiceChange' callback when a choice is activated", () => {
            const onChoiceChangeFn = jest.fn();
            setup({
                ...baseMultipleChoiceProps,
                onChoiceChange: onChoiceChangeFn,
            });

            act(() => {
                screen.getByRole("listitem").click();
            });
            expect(onChoiceChangeFn).toHaveBeenCalledTimes(1);
            expect(onChoiceChangeFn.mock.calls[0][0]).toBe(0);
            expect(typeof onChoiceChangeFn.mock.calls[0][1]).toBe("boolean");
        });
```

packages/perseus/src/styles/widgets/radio.css
```diff
@@ -1,3 +1,5 @@
+/* TODO: (LEMS-3301) Remove this file once the flag is removed */
```
SonicScrewdriver, line 1: Greatly appreciate these comments!

packages/perseus/src/widgets/radio/__tests__/multiple-choice-component.test.tsx
SonicScrewdriver: That's a good callout! 

Chromatic has also been very useful for finding/confirming minor styling changes with interactive graph.

packages/perseus/src/widgets/radio/__tests__/multiple-choice-component.test.tsx
```diff
@@ -0,0 +1,260 @@
...
+                    reviewMode={true}
+                />,
+            );
+            const classes = Array.from(
+                screen.getByText(rationaleContent).classList,
+            );
+            expect(classes).toEqual(expect.arrayContaining(["rationale"]));
+        });
+
+        it("applies EXTRA stying to rationale for correct choices", () => {
```
SonicScrewdriver: total nit: I think there's a typo with the word "styling" in the names of a few of these tests

packages/perseus/src/widgets/radio/__tests__/multiple-choice.test.ts
SonicScrewdriver: Makes sense, I think. Would it be better to simply comment out the file, or does that anger Jest? 

We already duplicated these tests from the original UI, so it might be weird to have them all running twice against the same set of code.

packages/perseus/src/widgets/radio/choice-indicator.module.css
```diff
@@ -1,19 +1,23 @@
 .base {
-    appearance: none;
+    appearance: none; /* Remove user agent formatting in favor of our own */
```
SonicScrewdriver, line 2: Love a good comment

packages/perseus/src/widgets/radio/choice.module.css
SonicScrewdriver: I do really like how you've called this out, as it provides a lot of clarity for why these are here, and when we can remove them

packages/perseus/src/widgets/radio/choice.module.css
SonicScrewdriver: ++

packages/perseus/src/widgets/radio/choice.module.css
```diff
@@ -1,73 +1,171 @@
...
+        (
+                var(--perseus-multiple-choice-divider-space) -
+                    (
+                        (
+                                var(--perseus-multiple-choice-divider-space) -
+                                    var(--wb-border-width-thin)
+                            ) / 2
+                    )
+            ) * -1
+    );
```
SonicScrewdriver, line 66: It's weird how this was expanded, likely due to the prettier standards. I find it a little harder to read, but such is life.

packages/perseus/src/widgets/radio/choice.module.css
```diff
@@ -1,73 +1,171 @@
...
-    padding-top: calc(
-        var(--perseus-multiple-choice-spacing) - var(--wb-sizing-size_010)
-    ); /* Account for increased border width */
+/* Enhance (thicker and darker) the divider between choices when:
+      - the choice indicator has focus (not in review mode) OR
+      - the choice is hovered (not in review mode) OR
+      - the choice is in review mode AND
+         - the choice is incorrect AND
+         - the choice indicator has focus
+ */
```
SonicScrewdriver, line 101: Great comment!

packages/perseus/src/widgets/radio/__tests__/multiple-choice.test.ts
```diff
@@ -26,7 +26,7 @@ import type {UserEvent} from "@testing-library/user-event";
 const createApiOptions = (options: Partial<APIOptions> = {}): APIOptions => ({
```
ivyolamit, line 26: if we are only setting the APIOption with the same values, we can simplify the test by using the `beforeEach` block:

```suggestion
let apiOptions: APIOptions;
beforeEach(() => {
    // Reset the API options before each test
    apiOptions = createApiOptions();
});
// Create API options with the new-radio-widget flag enabled
const createApiOptions = (options: Partial<APIOptions> = {}): APIOptions => ({
    ...options,
    flags: {
        "new-radio-widget": false,
        ...(options.flags || {}),
    },
});
```

then in each test we can remove this line:
const apiOptions = createApiOptions();

packages/perseus/src/widgets/radio/choice.new.tsx
```diff
@@ -1,273 +1,50 @@
...
-        paddingTop: 0,
-        marginInlineStart: 54,
-        color: color.offBlack64,
-        [mediaQueries.smOrSmaller]: {
-            padding: intermediateCheckboxPaddingPhone,
-            paddingTop: 0,
-        },
-    },
-} as const;
+const Choice = (props: IndicatorProps) => {
```
SonicScrewdriver, line 15: step-brothers-room-for-activities

Look at how much more svelte this component is! Great work

packages/perseus/src/widgets/radio/choice.new.tsx
```diff
@@ -1,273 +1,50 @@
...
+        .join(" ");
+    return (
+        /*
+            A click handler is added to the <li> element to accommodate mouse users.
+            Keyboard users will still be able to use the <button> element inside
+                the <Indicator>.
+            Therefore, WCAG 2.1.1 is still satisfied since functionality is
+                aligned with the input method.
+         */
+        // eslint-disable-next-line jsx-a11y/click-events-have-key-events,jsx-a11y/no-noninteractive-element-interactions
```
SonicScrewdriver, line 35: This is another great comment that will save us future confusion, while also allowing us to quickly reassess as WCAG requirements evolve in the future

packages/perseus/src/widgets/radio/focus-ring.new.tsx
```diff
@@ -1,61 +0,0 @@
-/**
```
SonicScrewdriver, line 1: get-out

packages/perseus/src/widgets/radio/__tests__/multiple-choice.test.ts
ivyolamit: Do you know why we need to update this to false from true?
Based on previous comment this is for 
> // Create API options with the new-radio-widget flag enabled

Also my understanding is that component is for the new radio widget, am I missing something?

packages/perseus/src/widgets/radio/multiple-choice-component.new.tsx
```diff
@@ -72,208 +37,102 @@ interface MultipleChoiceComponentProps {
...
     const {strings} = usePerseusI18n();
 
-    const choiceRefs = useRef<Array<React.RefObject<HTMLButtonElement>>>([]);
-    // Keep track of the previous review mode rubric to avoid unnecessary
-    // scrolling when switching between review mode and edit mode.
-    const prevReviewModeRubric = useRef<
-        PerseusRadioWidgetOptions | undefined | null
-    >();
-
-    useEffect(() => {
```
SonicScrewdriver, line 96: I'm so glad we were able to get rid of this useEffect, as it just took up a huge portion of this file

SonicScrewdriver (changes requested):
LOVE this - The transition of the checkbox is gorgeously smooth. I left one comment about the test file that was pointed back to the old widget, but I did find a weird bug with the single selection radio mode. I think it's possibly related to the ServerItemRendererWithDebugUI rather than your work, so I'll investigate further.

packages/perseus/src/widgets/radio/choice-indicator.new.tsx
```diff
@@ -40,13 +38,12 @@ const Indicator = (props: IndicatorProps) => {
         ? undefined // Don't register anything when showing answers
         : (event: React.MouseEvent<HTMLButtonElement>) => {
               event.stopPropagation();
-              setIsChecked(!isChecked);
-              props.updateChecked(!isChecked);
+              props.updateChecked(!checked);
           };
 
     return (
         <button
```
anakaren-rojas, line 45: Are there any advantages to having the button isolated within the Indicator vs having it encompass the entire choice content?
```
<li onClick={clickHandler}>  // Click forwarding layer
    <Indicator>
        <button>             // Button inside Indicator component
            {icon} {letter}
        </button>
    </Indicator>
    {choiceContent}          // Content outside button
</li>
```
vs 
```
<li>
    <button>                 // Button wraps everything at Choice level
        <Indicator>          // Pure visual component
            {icon} {letter}
        </Indicator>
        {choiceContent}      // Content inside button
    </button>
</li>
```

ivyolamit:
Missing fade on the right when scrolling LTR. Also similar styling will also apply for RTL where the fade will be on the left when scrolling.

| LTR | RTL |
| --- | ----- |
| Screenshot 2025-07-30 at 10 54 12 AM | Screenshot 2025-07-30 at 10 56 07 AM |

packages/perseus/src/widgets/radio/__tests__/multiple-choice-component.test.tsx
mark-fitzgerald: Good catch! Drying code...

packages/perseus/src/widgets/radio/__tests__/multiple-choice-component.test.tsx
mark-fitzgerald: I plan on adding Chromatic visual snapshots (for regression tests) in a separate ticket. I'm testing for a CSS class here primarily for test coverage. I'm verifying the base case here since subsequent tests check for changes.

packages/perseus/src/widgets/radio/__tests__/multiple-choice-component.test.tsx
mark-fitzgerald: Done.

packages/perseus/src/widgets/radio/__tests__/multiple-choice.test.ts
mark-fitzgerald: This file will be going away once we switch to the new widget and their associated tests.

packages/perseus/src/widgets/radio/__tests__/multiple-choice.test.ts
mark-fitzgerald: I don't remember why this file was created. I think it was to test the new widget parts. However, since those parts have their own tests now, this file is probably no longer needed. It will definitely go away once we switch over to the new widget parts.

packages/perseus/src/widgets/radio/choice.module.css
mark-fitzgerald: I tried fighting Prettier on this. Guess who won.

packages/perseus/src/widgets/radio/choice-indicator.new.tsx
mark-fitzgerald: Both are valid approaches. One benefit that I see in keeping the `<button>` with the indicator is simplicity of the DOM and the styling. When the button is in the indicator, then the `<button>` is the wrapping element. If it is moved up the hierarchy, then another element is needed to wrap the indicator. That's not a big deal, just an example of the difference. That said, when we do the accessibility pass on this widget, if adjusting the DOM would help then I'm fine changing it.

packages/perseus/src/widgets/radio/__tests__/multiple-choice-component.test.tsx
nishasy: @mark-fitzgerald can you add a comment to the styling tests explaining what you just explained

jest-css-modules-transform-config.js
```diff
@@ -0,0 +1,7 @@
+/* eslint-disable @typescript-eslint/no-require-imports */
```
nishasy: I think you can remove this `no-require-imports` lint disable comment. The `no-commonjs` should be sufficient?

jest-css-modules-transform-config.js
```diff
@@ -0,0 +1,7 @@
...
+/* eslint-disable import/no-commonjs */
+module.exports = {
+    cssLoaderConfig: {
+        exportLocalsStyle: "camelCase",
+    },
+};
```
nishasy, line 6: Question for my own understanding: How were our styles loading correctly without this config existing before?

packages/perseus/src/widgets/radio/__stories__/multiple-choice-option.stories.tsx
```diff
@@ -1,6 +1,6 @@
 import * as React from "react";
 
-import Choice from "../multiple-choice-option.new";
+import Choice from "../choice.new";
```
nishasy, line 3: Is the [Widgets/RadioNew/Widget Internal Components/Choice](https://650db21c3f5d1b2f13c02952-knpmiiicyz.chromatic.com/?path=/docs/widgets-radionew-widget-internal-components-choice--docs) story supposed to look like this?

Screenshot 2025-08-04 at 2 09 12 PM

packages/perseus/src/widgets/radio/__tests__/multiple-choice-component.test.tsx
```diff
@@ -0,0 +1,225 @@
...
+            expect(classes).toEqual(
+                expect.arrayContaining(["rationale", `is-correct`]),
+            );
+        });
+    });
+
+    describe("choice option configuration", () => {
+        it.each`
+            reviewMode | isCorrect | expectedCorrectness | when
+            ${false}   | ${true}   | ${undefined}        | ${"NOT in review mode"}
```
nishasy: I recognize it doesn't matter too much, but could you add a `reviewMode=false` and `isCorrect=false` case as well just to cover all our bases?

packages/perseus/src/widgets/radio/multiple-choice-component.new.tsx
```diff
@@ -72,208 +37,102 @@ interface MultipleChoiceComponentProps {
...
         multipleSelect,
         countChoices,
         numCorrect,
         strings,
     });
 
-    const fieldset = (
+    const choiceListClasses = [styles.choiceList]
+        .concat(reviewMode ? [styles.reviewAnswers] : [])
+        .join(" ");
```
nishasy: I'm finding this a little hard to parse? Could this just be
```suggestion
    const choiceListClasses = reviewMode
        ? `${styles.choiceList} ${styles.reviewAnswers}`
        : styles.choiceList;
```
?

nishasy (approved):
Some small comments, but looks good overall!

jest-css-modules-transform-config.js
mark-fitzgerald: This file is just for tests. Previously, jest would resolve the class names in the `styles` object (imported CSS Modules files) as `undefined`. After adding this library, the referenced class names would show up in the rendered DOM in the jest tests. However, only the kabob-cased versions of the class names were rendered. With this config in place, it will now also recognize the camelCased versions of the class names and will output the converted (to kabob-case) class name.

packages/perseus/src/widgets/radio/__stories__/multiple-choice-option.stories.tsx
mark-fitzgerald: Nope. That's gross! I changed the styling in this PR to be more symbiotic among the component parts. That particular story shows a component part in isolation, and therefore is missing key styling information. Also, there are aspects of it that are no longer valid. I have added a [ticket to revise the Storybook entries](https://khanacademy.atlassian.net/browse/LEMS-3352).

packages/perseus/src/widgets/radio/__tests__/multiple-choice-component.test.tsx
mark-fitzgerald: Done.

packages/perseus/src/widgets/radio/__tests__/multiple-choice-component.test.tsx
mark-fitzgerald: Done.

packages/perseus/src/widgets/radio/multiple-choice-component.new.tsx
mark-fitzgerald: Good call out. Done.

PR #2747, opened 2025-07-30
Force certain widgets to always be LTR, even in RTL langs

mahtabsabet:
## Summary:
I am continuing on @anniegallagher's work during the 2024 RTL Hackathon, and making certain widgets stay LTR even in RTL languages. That's because even for RTL languages, math stays LTR!

Numeric input before:

image

Numeric input after:

image

Keypad+tabbar before:

image

Keypad+tabbar after:

image

Issue: CP-XXXX

## Test plan:
`pnpm storybook` and set `dir="rtl"` when testing each component to visually assess

### DIFF
--- a/packages/perseus/src/widgets/numeric-input/numeric-input.tsx
+++ b/packages/perseus/src/widgets/numeric-input/numeric-input.tsx
@@ -73,6 +73,8 @@ export const NumericInputComponent = forwardRef<Focusable, NumericInputProps>(
                 padding: isFocused ? "4px" : "4px 5px",
                 textAlign: props.rightAlign ? "right" : "left",
                 width: props.size === "small" ? 40 : 80,
+                // Even in RTL languages, math is LTR.
+                // So we force this component to always render LTR
                 direction: "ltr",
             },
         });
--- a/packages/math-input/src/components/keypad/keypad.tsx
+++ b/packages/math-input/src/components/keypad/keypad.tsx
@@ -219,6 +219,8 @@ const styles = StyleSheet.create({
         display: "flex",
         flexDirection: "row",
         backgroundColor: "#DBDCDD",
+        // Even in RTL languages, math is LTR.
+        // So we force this component to always render LTR
         direction: "ltr",
     },
     keypadGrid: {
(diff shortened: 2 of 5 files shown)


SonicScrewdriver (approved):
Wonderful! I'm glad to hear that the keypad can remain ltr. Thank you for getting these set up

packages/perseus/src/widgets/numeric-input/numeric-input.tsx
```diff
@@ -73,6 +73,7 @@ export const NumericInputComponent = forwardRef<Focusable, NumericInputProps>(
                 padding: isFocused ? "4px" : "4px 5px",
                 textAlign: props.rightAlign ? "right" : "left",
                 width: props.size === "small" ? 40 : 80,
+                direction: "ltr",
```
mark-fitzgerald, line 78: I'm not sure how important it is in this PR, but we also have the Input Number widget (legacy). There is a fair amount of content that uses that widget (precursor to this Numeric Input). However, we will be removing it once we are able to conduct a data "backfill" and convert instances of Input Number to Numeric Input. So, just calling this out in case it's important enough for the interim to have that LTR, as well.

SonicScrewdriver, line 78: It could possibly be worthwhile to have a comment here explaining to our future selves why we're setting this specifically - but I don't consider it a blocker or something absolutely required.

mark-fitzgerald (approved):
Looks good. Perhaps adding a simple comment to each entry indicating reason that we want to force LTR for these instances. That way, we don't revert them in the future thinking that we are correcting a misapplied property (we have lots of misapplied properties in Perseus).

ivyolamit (approved):
LGTM 👍🏼

packages/perseus/src/widgets/numeric-input/numeric-input.tsx
mahtabsabet: Good point! Added for Input Number as well

mahtabsabet: Added comments for each spot I added this exception :)

PR #2750, opened 2025-07-30
fix(LEMS-3339): Display "check" button for articles

anakaren-rojas:
## Summary:
"Check" button in articles is not displayed when users are on mobile.
This change checks whether the learner is on an article and optionally displays the "GradedGroupAnswerBar" if they are on mobile but not in an article. 

Issue: LEMS-3339

## Test plan:

### DIFF
--- a/packages/perseus/src/widgets/graded-group/graded-group.tsx
+++ b/packages/perseus/src/widgets/graded-group/graded-group.tsx
@@ -331,7 +331,7 @@ export class GradedGroup
                     )}
                 </UserInputManager>
 
-                {!apiOptions.isMobile && (
+                {(!apiOptions.isMobile || apiOptions.isArticle) && (
                     <>
                         {icon != null && (
                             <div className="group-icon">{icon}</div>
@@ -440,14 +440,16 @@ export class GradedGroup
                             {this.context.strings.explain}
                         </button>
                     ))}
-                {apiOptions.isMobile && answerBarState !== "HIDDEN" && (
-                    <GradedGroupAnswerBar
-                        apiOptions={apiOptions}
-                        answerBarState={answerBarState}
-                        onCheckAnswer={this._checkAnswer}
-                        onNextQuestion={this.props.onNextQuestion}
-                    />
-                )}
+                {apiOptions.isMobile &&
+                    !apiOptions.isArticle &&
+                    answerBarState !== "HIDDEN" && (
+                        <GradedGroupAnswerBar
+                            apiOptions={apiOptions}
+                            answerBarState={answerBarState}
+                            onCheckAnswer={this._checkAnswer}
+                            onNextQuestion={this.props.onNextQuestion}
+                        />
+                    )}
             </div>
         );
     }


mark-fitzgerald (approved):
I'm not a fan of using `setTimeout`, but I also don't like the alternatives when in a React class component. This looks like it does the trick, though.

packages/perseus/src/widgets/graded-group/graded-group.tsx
```diff
@@ -157,15 +157,30 @@ export class GradedGroup
...
+                // Show answer bar if there's any user input (not all widgets empty)
+                const allWidgets = this.rendererRef.current.getWidgetIds();
+                const hasAnyInput = emptyWidgets.length < allWidgets.length;
+
+                let nextState = answerBarState;
+                if (answerBarState === "HIDDEN" && hasAnyInput) {
+                    nextState = answerable ? "ACTIVE" : "INACTIVE";
+                } else {
+                    nextState = getNextState(answerBarState, answerable);
+                }
```
benchristel: Why did this logic change? It seems like the new logic should be added to `getNextState` if it's needed.

packages/perseus/src/widgets/graded-group/graded-group.tsx
```diff
@@ -157,15 +157,30 @@ export class GradedGroup
...
-            const answerable = emptyWidgets.length === 0;
-            const answerBarState = this.state.answerBarState;
-            const nextState = getNextState(answerBarState, answerable);
-            this.setState({
-                answerBarState: nextState,
-            });
-        }
+        // Use setTimeout to ensure widget states have been updated before checking
+        // if the question is answerable.
+        setTimeout(() => {
```
benchristel: I don't think `setTimeout` is a reliable way of scheduling an event relative to a React update. React can debounce and batch state updates, so the `setTimeout` callback might run before React updates the state in some cases.

> State updates in React are asynchronous; when an update is requested, there is no guarantee that the updates will be made immediately. The updater functions queue changes to the component state, but React may delay the changes, updating several components in a single pass.
>
> —https://blog.logrocket.com/why-react-doesnt-update-state-immediately/

I think the correct place to fix this is in the UserInputManager. The problem is that the UserInputManager is calling its `handleUserInput` callback before the renderer has actually updated. @handeyeco created UserInputManager pretty recently, and would probably know the most about why this bug might be happening.

benchristel: I think the bug might be happening because `Renderer.emptyWidgets()` recently changed to use `props.userInput`, rather than userInput from the Renderer's own state. This means that `emptyWidgets()` won't have the correct value until the Renderer component re-renders.

```ts
emptyWidgets(): ReadonlyArray<string> {
    if (!this.props.userInput) {
        throw new Error(
            `emptyWidgets called without providing userInput to Renderer`,
        );
    }

    return emptyWidgetsFunctional(
        this.state.widgetInfo,
        this.widgetIds,
        this.props.userInput,
        this.context.locale,
    );
}
```

I wonder if we should take this bug as an opportunity to stop calling the imperative `emptyWidgets` method (which is a React antipattern) and instead switch to having the GradedGroup or the UserInputManager call `emptyWidgetsFunctional`. I can think of a couple ways to structure the code:

- if `GradedGroup` has all the information it needs to pass to `emptyWidgetsFunctional`, it can call it directly.
  - `UserInputManager` passes the `UserInputMap` to `handleUserInput`, so GradedGroup can get the UserInputMap from there.
- if that won't work, `UserInputManager` could call `emptyWidgetsFunctional` when it receives user input, and pass the return value to its `handleUserInput` callback so GradedGroup can use it.

anakaren-rojas: To avoid a randomization repeat 😆 Would either of you be available for a pairing to align on strategy @benchristel @handeyeco? I'm open to making the change, but want to be sure we're all on the same page before starting

packages/perseus-score/src/validate.ts
```diff
@@ -31,5 +31,8 @@ export function emptyWidgetsFunctional(
         if (score) {
             return scoreIsEmpty(score);
         }
+
+        // If validator returned null, the widget is valid/filled, so not empty
```
anakaren-rojas, line 35: explicit return instead of implicit `undefined`

packages/perseus/src/renderer.tsx
```diff
@@ -646,7 +646,23 @@ class Renderer
                 this._setWidgetProps(widgetId, newProps, cb, silent);
             },
             handleUserInput: (newUserInput: UserInput) => {
-                this.props.handleUserInput?.(widgetId, newUserInput);
+                // Calculate widgetsEmpty using the updated user input
+                const updatedUserInput = {
+                    ...this.props.userInput,
+                    [widgetId]: newUserInput,
```
anakaren-rojas, line 652: `this.emptyWidgets()` checks `this.props.userInput` which still has the old/stale user input because React hasn't updated the props yet. Using `newUserInput` values instead to calculate empty widgets based on user's input

packages/perseus/src/widgets/graded-group/graded-group.tsx
```diff
@@ -135,7 +136,7 @@ export class GradedGroup
         status: GRADING_STATUSES.ungraded,
         showHint: false,
         message: "",
-        answerBarState: "HIDDEN",
+        answerBarState: "INACTIVE",
```
anakaren-rojas, line 137: Per request from Jeanette and Susan - making answerBar always visible instead of visible only after a user has interacted with the widget

benchristel: Does that mean we can delete the `HIDDEN` state from the `ANSWER_BAR_STATES` type (and the `switch` statement that handles it in `getNextState`) now?

packages/perseus/src/renderer.tsx
benchristel: Very cool!

packages/perseus/src/widgets/graded-group/__snapshots__/graded-group.test.ts.snap
```diff
@@ -744,6 +744,23 @@ exports[`graded-group should snapshot: initial render (mobile: true) 1`] = `
...
+                role="button"
+                type="button"
+              >
+                <span
+                  class="text_f1191h-o_O-LabelLarge_16mkhxv-o_O-text_1kowsup"
+                >
+                  Check
+                </span>
+              </button>
+            </div>
```
benchristel, line 763: This is because the answer bar is now displayed by default, right?

packages/perseus/src/widgets/graded-group/graded-group.test.ts
```diff
@@ -334,6 +335,22 @@ describe("graded-group", () => {
                 screen.queryByText(/Some bacteria synthesize their own fuel./),
             ).not.toBeInTheDocument();
         });
+
+        it("should enable Check button when all widgets are completed", async () => {
+            // Arrange
+            renderQuestion(groupedRadioRationaleQuestion, apiOptions);
+
+            // Act - Fill all radio buttons
+            await userEvent.click(screen.getByRole("radio", {name: /Correct/}));
```
benchristel: Tiny nit — to guard against this test erroneously passing in the future, I think it would be good to check that the button is _not_ enabled before clicking the Radio option.

Also, I think the comment on `// Act` was meant for a categorizer widget. My suggested code updates it for Radio.

```suggestion
            renderQuestion(groupedRadioRationaleQuestion, apiOptions);

            const checkButton = screen.getByRole("button", {name: "Check"});
            expect(checkButton).not.toBeEnabled();

            // Act - Click a radio option
            await userEvent.click(screen.getByRole("radio", {name: /Correct/}));
```

packages/perseus/src/widgets/graded-group/graded-group.test.ts
```diff
@@ -334,6 +335,22 @@ describe("graded-group", () => {
...
+
+            // Wait for setTimeout to complete
+            act(() => jest.runOnlyPendingTimers());
```
benchristel: Now that the `setTimeout` is gone, do we still need this?

```suggestion
```

packages/perseus/src/widgets/graded-group/__snapshots__/graded-group.test.ts.snap
anakaren-rojas: yes!

packages/perseus/src/widgets/graded-group/graded-group.tsx
anakaren-rojas: it sure does

PR #2751, opened 2025-07-31
Enable react/display-name rule for better React dev tools support

jeremywiebe:
## Summary:

React dev tools can do things like displaying a render tree of React components. This debugging view depends on React components having a display name. React derives this display name in a few ways:
 
  * By looking for a `.displayName` property attached to the component
  * By using the functional components function name (ie `function MyComponent() { ... }`). 

This PR enables the ESLint rule that ensures all React components have a derivable display name and fixes the issues that came up after enabling it. 

Issue: "none"

## Test plan:

`pnpm lint`

### DIFF
--- a/packages/perseus/src/server-item-renderer.tsx
+++ b/packages/perseus/src/server-item-renderer.tsx
@@ -541,14 +541,19 @@ const styles = StyleSheet.create({
     },
 });
 
-const ref = React.forwardRef<
+export default React.forwardRef<
     ServerItemRenderer,
     Omit<PropsFor<typeof ServerItemRenderer>, "onRendered">
->((props, ref) => (
-    <LoadingContext.Consumer>
-        {({onRendered}) => (
-            <ServerItemRenderer {...props} onRendered={onRendered} ref={ref} />
-        )}
-    </LoadingContext.Consumer>
-));
-export default ref;
+>(function ServerItemRendererWithRef(props, ref) {
+    return (
+        <LoadingContext.Consumer>
+            {({onRendered}) => (
+                <ServerItemRenderer
+                    {...props}
+                    onRendered={onRendered}
+                    ref={ref}
+                />
+            )}
+        </LoadingContext.Consumer>
+    );
+});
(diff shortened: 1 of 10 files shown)


packages/perseus/src/server-item-renderer.tsx
```diff
@@ -541,14 +541,19 @@ const styles = StyleSheet.create({
...
     Omit<PropsFor<typeof ServerItemRenderer>, "onRendered">
->((props, ref) => (
-    <LoadingContext.Consumer>
-        {({onRendered}) => (
-            <ServerItemRenderer {...props} onRendered={onRendered} ref={ref} />
-        )}
-    </LoadingContext.Consumer>
-));
-export default ref;
+>(function ServerItemRendererWithRef(props, ref) {
```
handeyeco, line 547: Some part of my OCD brain wants the function to live outside of this when using the keyword `function`.

Like instead of:

``` TS
export default React.forwardRef<
  ServerItemRenderer,
  PropsFor<typeof ServerItemRenderer>
>(function ServerItemRendererWithRef(props, ref) {
  // ...whatever...
})
```

Do this:

``` TS
function ServerItemRendererWithRef(props, ref) {
  // ...whatever...
}

export default React.forwardRef<
  ServerItemRenderer,
  PropsFor<typeof ServerItemRenderer>
>(ServerItemRendererWithRef)
```

But maybe that's not even possible with `ref` 🤔 

Just a nit/thought, disregard.

handeyeco (approved):

packages/perseus/src/server-item-renderer.tsx
jeremywiebe: I'll try, but I think we'll have an easier time with types if we leave it inline.  If it's outside we will likely have to type `ServerItemRendererWithRef` function fully so that it fits in the `forwardRef()`. If we leave it inline, TypeScript helps us because it can infer the `ServerItemRendererWithRef` function's parameter types because it's inside the `forwardRef`.

jeremywiebe: Ok.  We _can_ extract the function, but without additional typing, both `props` and `ref` are `any`. 

I can do this, but this feels uglier than the current syntax:

image

I'm going to land this as-is. Feel free to adjust in a new PR if you have ideas for a better approach.

PR #2752, opened 2025-07-31
Score widgets with missing user input as 'invalid'

benchristel:
This fixes a bug observed while testing the Perseus service with production
traffic.

Issue: https://khanacademy.atlassian.net/browse/LEMS-3341

## Test plan:

`pnpm test`

### DIFF
--- a/packages/perseus-score/src/widgets/expression/score-expression.test.ts
+++ b/packages/perseus-score/src/widgets/expression/score-expression.test.ts
@@ -7,7 +7,7 @@ import type {
 } from "@khanacademy/perseus-core";
 
 describe("scoreExpression", () => {
-    it("should score undefined user input as invalid", function() {
+    it("should score undefined user input as invalid", function () {
         const err = scoreExpression(undefined, expressionItem3Options, "en");
         expect(err).toHaveInvalidInput();
     });
--- a/packages/perseus-score/src/widgets/free-response/validate-free-response.test.ts
+++ b/packages/perseus-score/src/widgets/free-response/validate-free-response.test.ts
@@ -12,10 +12,7 @@ const MOCK_STANDARD_OPTIONS: PerseusFreeResponseWidgetOptions = {
 
 describe("free response validation", () => {
     it("should return invalid for undefined user input", () => {
-        const result = validateFreeResponse(
-            undefined,
-            MOCK_STANDARD_OPTIONS,
-        );
+        const result = validateFreeResponse(undefined, MOCK_STANDARD_OPTIONS);
         expect(result).toHaveInvalidInput("USER_INPUT_EMPTY");
     });
 
--- a/packages/perseus-score/src/widgets/grapher/score-grapher.test.ts
+++ b/packages/perseus-score/src/widgets/grapher/score-grapher.test.ts
@@ -31,7 +31,7 @@ describe("scoreGrapher", () => {
 
         // Assert
         expect(result).toHaveInvalidInput();
-    })
+    });
 
     it("is incorrect when user input type doesn't match scoring data type", () => {
         const asymptote: [Coord, Coord] = [
--- a/packages/perseus-score/src/widgets/group/score-group.test.ts
+++ b/packages/perseus-score/src/widgets/group/score-group.test.ts
@@ -1,5 +1,6 @@
 import scoreGroup from "./score-group";
-import {PerseusGroupRubric} from "@khanacademy/perseus-core";
+
+import type {PerseusGroupRubric} from "@khanacademy/perseus-core";
 
 describe("scoreGroup", () => {
     it("returns a score of 'invalid' when the user input is undefined", () => {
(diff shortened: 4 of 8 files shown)


anakaren-rojas:
Would we eventually want to have default values?

benchristel:
@handeyeco I tried the alternative of removing the scoring functions for the widgets that use scoreNoop. The problem I ran into is that `measurer` and `deprecated-standin` need to award 1/1 points instead of 0/0, and also need to accept `undefined` as user input. I don't see a way to accomplish this without either a) making `scoreWidgetsFunctional` aware of logic for individual widgets, or b) moving handling of `undefined` to the scoring functions for specific widgets as I've done here.

benchristel:
> Would we eventually want to have default values?

It would be nice if widgets always set the userInput so we didn't have to handle undefined. But we might still want to handle undefined for type safety reasons.

Types can't rule out the possibility of `undefined` when you access a property of a map-like object. `{}` is a perfectly valid value of type `UserInputMap`, and if we receive that value, all widgets will have undefined user input. Frustratingly, TypeScript doesn't acknowledge the possibility of getting `undefined` when you access a map-like object, so we have to explicitly write `undefined` into the type.

packages/perseus-score/src/widgets/cs-program/score-cs-program.test.ts
```diff
@@ -3,6 +3,18 @@ import scoreCSProgram from "./score-cs-program";
...
+    it("is 'invalid' when the state is undefined", () => {
+        // Arrange
+        const state = undefined;
+
+        // Act
+        const result = scoreCSProgram(state);
+
+        // Assert
+
+        expect(result).toHaveInvalidInput();
```
handeyeco, line 15:
```suggestion
        // Assert
        expect(result).toHaveInvalidInput();
```

packages/perseus-score/src/widgets/categorizer/score-categorizer.ts
```diff
@@ -5,9 +5,14 @@ import type {
...
 function scoreCategorizer(
-    userInput: PerseusCategorizerUserInput,
+    // NOTE(benchristel): userInput can be undefined if the widget has never
+    // been interacted with.
+    userInput: PerseusCategorizerUserInput | undefined,
     rubric: PerseusCategorizerRubric,
 ): PerseusScore {
+    if (userInput == null) {
+        return {type: "invalid", message: null};
+    }
```
handeyeco, line 15: I guess I don't see any harm in doing these checks (better safe than sorry I s'pose) but they feel a little repetitive and out of place in the scorer.

- I still think we need to get rid of the `noop` scoring functions ([LEMS-2543](https://khanacademy.atlassian.net/browse/LEMS-2543)) which I know is more complicated than it should be because some return 0 points and some return 1 point. If we could get rid of those, the logic calling the scoring functions can do a null check first so we don't have to do one at the start of each scoring function.
- This feels like it should be more of a responsibility for the validation functions rather than scoring functions. The logic using these should call the validator before the scorer anyway. Plus we'd get the added benefit of client-side validation. At one point we called the validation function at the start of each scoring function; maybe we could do that again if we don't trust the consumer of the scorers to remember to call the validator first.

I don't see any harm in merging this though except that it obfuscates original intent because ideally:

- User input should be initialized for scorable widgets `getStartUserInput`
- `emptyWidgetsFunctional` should stop people from submitting empty widgets
- Before calling the scoring function, we should be calling the validation function on the server
- So the scoring function should never get called with null/undefined user input

handeyeco (approved):

packages/perseus-score/src/widgets/categorizer/score-categorizer.ts
benchristel: > Before calling the scoring function, we should be calling the validation function on the server

@handeyeco is there any reason we shouldn't call `emptyWidgetsFunctional` in `scoreWidgetsFunctional`? I'd rather do that then force the caller to know that scoring might throw if they haven't validated. That kind of [connascence of execution] makes code harder to reason about IME.

[connascence of execution]: https://connascence.io/execution.html

handeyeco: `emptyWidgetsFunctional` is mostly just a wrapper around the `validator`s. `scoreWidgetsFunctional` also calls the validators. I'd be a little worried about calling the validators twice (especially if any of them are doing TeX/expression parsing) but I wonder if there are any small tweaks to the logic that could get us to the behavior you want without calling them twice?

benchristel: We do validate during scoring; apparently that's not enough to handle all the cases, though:

```ts
// We do validation (empty checks) first and then scoring. If
// validation fails, it's result is itself a PerseusScore.
const score =
    validator?.(userInput, widget.options, locale) ??
    scorer?.(userInput, widget.options, locale);
```

PR #2753, opened 2025-08-01
Removes the width override for the cs-program widget

MikeKlemarewski:
## Summary:
Remove width override for cs-program widget. This width was causing some overflow issues in KA Classroom, and after testing this in Classroom and Classic, I don't think it's needed.

### Classic

https://github.com/user-attachments/assets/ac3f350a-68a7-4ee1-92c8-10f6c891dcfa

### Classroom

https://github.com/user-attachments/assets/cbe67463-c9dd-43f4-b688-83accfbf3e0e

Issue: TUT-2719

## Test plan:
- Bump the perseus version to this PR in the frontend
- Test content with the widget and ensure it renders correctly

### DIFF
--- a/packages/perseus/src/widgets/cs-program/__snapshots__/cs-program.test.ts.snap
+++ b/packages/perseus/src/widgets/cs-program/__snapshots__/cs-program.test.ts.snap
@@ -16,7 +16,7 @@ exports[`cs-program widget should snapshot on mobile: first mobile render 1`] = 
           class="perseus-widget-container widget-nohighlight widget-block"
         >
           <div
-            class="widthOverride_1vdzrqd-o_O-container_1v2yujs"
+            class="container_1v2yujs"
           >
             <iframe
               allowfullscreen=""
@@ -49,7 +49,7 @@ exports[`cs-program widget should snapshot: first render 1`] = `
           class="perseus-widget-container widget-nohighlight widget-block"
         >
           <div
-            class="widthOverride_1vdzrqd-o_O-container_1v2yujs"
+            class="container_1v2yujs"
           >
             <iframe
               allowfullscreen=""


handeyeco:
Just for context: are you saying there was no visual impact for KA Classic but it helped fix a bug in Classroom? That's nice!

My main concern then, if that's the case, is the impact on mobile...but I'm not sure how much we support CSProgram on mobile. 🤔 Might be worth taking a look though.

MikeKlemarewski:
> Just for context: are you saying there was no visual impact for KA Classic but it helped fix a bug in Classroom? That's nice!
> 
> My main concern then, if that's the case, is the impact on mobile...but I'm not sure how much we support CSProgram on mobile. 🤔 Might be worth taking a look though.

Luke found a bit more context for the original override which was added almost 7 years ago. It was because some buttons were rendering off screen. [I added a screen recording and some extra suggestions here in Slack](https://khanacademy.slack.com/archives/C04UM6J0955/p1754413725580309?thread_ts=1753976144.891329&cid=C04UM6J0955).

packages/perseus/src/widgets/cs-program/cs-program.tsx
```diff
@@ -174,11 +174,6 @@ class CSProgram extends React.Component<Props> implements Widget {
 }
 
 const styles = StyleSheet.create({
-    // Override the inherited width from the perseus paragraph class
```
lsmith, line 177: Since the removal of lines doesn't leave blame traces, can you include a code comment noting that removing a width override to 820px exposed a layout issue with rendering CS programs in articles, but this was preferable to leaving the override which introduced a layout issue in the new Classroom view.

lsmith (approved):
This change introduces a regression for classic CS programs embedded in articles, but that is an acceptable tradeoff to improve the new Classroom experience.

PR #2758, opened 2025-08-04
Categorizer UserInput parsing issue (null / undefined values)

handeyeco:
## Summary:

Categorizer user input was typed as `number[]`. _Correct_ answers can only be `number[]` but incomplete user input could be `Array<number | undefined>`. To complicate things, `JSON.stringify` converts `[undefined, 2]` to `[null,2]` so the backend also needs to support `null`.

So fix the parser to handle number, undefined, or  null.

Issue: LEMS-3344

### DIFF
--- a/.changeset/weak-dots-wave.md
+++ b/.changeset/weak-dots-wave.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/perseus": patch
+"@khanacademy/perseus-core": patch
+---
+
+Bugfix: Categorizer user input parser needs to handle undefined and null, not just numbers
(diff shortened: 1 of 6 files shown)


handeyeco:
@benchristel I explained the source of the nulls in the description and the ticket. I believe this solution is correct but let me know if you want to huddle and think through an alternative.

benchristel:
@handeyeco ah, got it. I think I was mixing up Categorizer and Matcher. For the curious, this is a categorizer:

Screenshot 2025-08-04 at 3 05 23 PM

PR #2760, opened 2025-08-04
Fix orderer bug that was adding null/undefined values to user input

benchristel:
## Summary:
When we briefly turned on serverside scoring for Khan Academy, we saw errors like
this coming from the Perseus service:

```
Failed to parse user input map: At (root)["orderer 1"].current[0] -- expected string, but got null
```

These nulls were erroneously getting into the user input because of a bug in
the UI. This PR fixes the bug. The Orderer is not accessible and not
recommended for use, so I am applying the minimal fix.

Issue: https://khanacademy.atlassian.net/browse/LEMS-3343

## Test plan:

In Storybook, drag a few cards into an Orderer. Try to reorder the cards.  A
broken-looking "ghost card" will appear. Try dragging and dropping the ghost
into the orderer. It should disappear.

### DIFF
--- a/packages/perseus/src/widgets/orderer/orderer.tsx
+++ b/packages/perseus/src/widgets/orderer/orderer.tsx
@@ -413,7 +413,7 @@ class Orderer
             if (!inCardBank) {
                 const cardContent: string | null | undefined =
                     this.state.dragContent;
-                if (this.isValidCard(cardContent)) {
+                if (typeof cardContent === "string") {
                     // Insert the new card into the position
                     const newCard = {
                         content: cardContent,
@@ -590,13 +590,6 @@ class Orderer
             bankOffset.left + bankWidth
         );
     };
-
-    private isValidCard(content: string | null | undefined): content is string {
-        return (
-            typeof content === "string" &&
-            this.props.options.some((opt) => opt.content === content)
-        );
-    }
 
     // This component makes use of a lot of DOM manipulation
     // For testing and direct manipulation of values there's a function


packages/perseus/src/widgets/orderer/orderer.tsx
```diff
@@ -587,6 +591,13 @@ class Orderer
         );
     };
 
+    private isValidCard(content: string | null | undefined): content is string {
+        return (
+            typeof content === "string" &&
+            this.props.options.some((opt) => opt.content === content)
```
benchristel: I guess we don't really need to check if the card is present in options. Just checking the type prevents the bug. 🤔

packages/perseus/src/widgets/orderer/orderer.tsx
```diff
@@ -411,14 +411,18 @@ class Orderer
...
-                const newCard = {
-                    content: this.state.dragContent,
-                    key: _.uniqueId("perseus_draggable_card_"),
-                    width: this.state.dragWidth,
-                } as const;
-
-                list.splice(index, 0, newCard.content);
+                const cardContent: string | null | undefined =
+                    this.state.dragContent;
+                if (typeof cardContent === "string") {
```
handeyeco, line 416: I'm guessing there's a reason for not doing this:

```suggestion
                if (cardContent != null) {
```

Not a blocker, just curiosity.

.changeset/purple-bulldogs-thank.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Fix a bug in Orderer that allowed learners to add cards with null content.
```
handeyeco, line 5: Can we add a regression test? Or is that impossible due to Draggable?

packages/perseus/src/widgets/orderer/orderer.tsx
benchristel: I figured it was safer/clearer to make a positive assertion and remove all doubt about what the type is. TS only "knows" that `cardContent` can be `null | undefined` because we told it, after all. It's not really keeping us safe here.

.changeset/purple-bulldogs-thank.md
benchristel: Yeah, I don't think there's an easy way to test this specific interaction. I'm not too concerned, because orderer is not recommended for use and if we wanted to make it usable I think we would just rewrite it.

PR #2762, opened 2025-08-05
feat(LEMS-3070): Add IDs to radio choices

anakaren-rojas:
## Summary:
Adds IDs as a required field for radio choices
- Legacy content's IDs are formatted `radio-choice-{index}`
- New content's IDs are UUIDs

Issue: LEMS-3070

## Test plan:
~ ZND Incoming ~

## Screenshots

https://github.com/user-attachments/assets/9c745f0a-9efc-4e7d-9180-6e33e2065d5a


packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
```diff
@@ -189,12 +197,19 @@ export function migrateV1ToV2(
     widget: ParsedValue<typeof parseRadioWidgetV1>,
 ): ParsedValue<typeof parseRadioWidgetV2> {
     const {options} = widget;
+
+    const choicesForCalculation = options.choices.map((choice, index) => ({
+        ...choice,
+        id: `radio-choice-${index}`, // <<what should go here??>>
```
anakaren-rojas: Attempted to do a path based ID, but didn't have access to the path in the migration function, defaulting to a simpler implementation - can be tweaked in the future

packages/perseus-editor/package.json
```diff
@@ -37,16 +37,17 @@
...
         "@khanacademy/perseus-core": "workspace:*",
+        "@khanacademy/perseus-linter": "workspace:*",
         "@khanacademy/perseus-score": "workspace:*",
         "@khanacademy/perseus-utils": "workspace:*",
         "@khanacademy/pure-markdown": "workspace:*",
         "katex": "0.11.1",
         "mafs": "^0.19.0",
-        "tiny-invariant": "catalog:prodDeps"
+        "tiny-invariant": "catalog:prodDeps",
+        "uuid": "^10.0.0"
```
handeyeco: Should this be a workspace or catalog dependency?

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
handeyeco: I think it would be better to add a TODO (ideally with a ticket explaining the need for future work) rather than having a comment be a question to future devs/AI.

packages/perseus-editor/src/widgets/__tests__/radio-editor.mockData.ts
```diff
@@ -0,0 +1,165 @@
...
+    {id: "f47ac10b-58cc-4372-a567-0e02b2c3d479", content: "Choice 2"},
+    {id: "6ba7b810-9dad-11d1-80b4-00c04fd430c8", content: "Choice 3"},
+    {id: "3e4d5c6b-7a89-4012-b345-6789cdef0123", content: "Choice 4"},
+];
+
+// id: "a1b2c3d4-e5f6-4789-a012-345678901234",
+// id: "f47ac10b-58cc-4372-a567-0e02b2c3d479",
+// id: "6ba7b810-9dad-11d1-80b4-00c04fd430c8",
+// id: "3e4d5c6b-7a89-4012-b345-6789cdef0123",
+// id: "9c8b7a65-4321-4fed-9876-543210fedcba"
```
handeyeco: I'm guessing this needs to get removed?

packages/perseus-editor/src/widgets/__tests__/radio-editor.test.tsx
```diff
@@ -1,15 +1,24 @@
+// Mock uuid module
+jest.mock("uuid", () => ({
+    v4: jest.fn(),
+}));
```
handeyeco: Feels like this needs to happen after imports

packages/perseus-editor/src/widgets/__tests__/radio-editor.test.tsx
```diff
@@ -1,15 +1,24 @@
...
 import {RenderStateRoot} from "@khanacademy/wonder-blocks-core";
 import {render, screen} from "@testing-library/react";
 import {userEvent as userEventLib} from "@testing-library/user-event";
 import * as React from "react";
+import {v4 as uuid} from "uuid";
 
 import {testDependencies} from "../../../../../testing/test-dependencies";
 import RadioEditor from "../radio/editor";
 
+import * as mockData from "./radio-editor.mockData";
```
handeyeco: I believe it's easier to automated tools like Knip to find dead code if we explicitly import what we need rather than `*` imports.

packages/perseus-editor/src/widgets/__tests__/radio-editor.test.tsx
```diff
@@ -142,12 +161,7 @@ describe("radio-editor", () => {
         const onChangeMock = jest.fn();
 
         renderRadioEditor(onChangeMock, {
-            choices: [
-                {content: "Choice 1"},
-                {content: "Choice 2", correct: true},
-                {content: "Choice 3"},
-                {content: "None of the above", isNoneOfTheAbove: true},
-            ],
+            choices: mockData.fourChoicesWithNotaAndCorrect,
```
handeyeco: Not a big deal, but it threw me for a loop at first trying to understand `Nota`

packages/perseus-editor/src/widgets/__tests__/radio-editor.mockData.ts
SonicScrewdriver: I think these commented out ids snuck in here

packages/perseus/src/__testdata__/extract-perseus-data.testdata.ts
```diff
@@ -18,18 +18,22 @@ export const PerseusItemWithRadioWidget = generateTestPerseusItem({
                     randomize: true,
                     choices: [
                         {
+                            id: "a1b2c3d4-e5f6-4789-a012-345678901234",
```
handeyeco: Going through all of these changes makes me think we need some kind of generator function for this kind of test data...although I'm not sure that would have helped with `uuid` unless there was some way to seed it.

packages/perseus/src/widgets/radio/radio-question-builder.test.ts
```diff
@@ -1,3 +1,7 @@
+jest.mock("uuid", () => ({
+    v4: jest.fn(() => "a1b2c3d4-e5f6-4789-a012-345678901234"),
+}));
```
handeyeco: Again, unless absolutely necessary, I would like to avoid having logic above imports.

handeyeco (approved):
This seems like an innocuous, low-risk change and I don't want to block you so I'll approve. I have some notes and I know you mentioned wanting Ben's review - I think it would be good to have him look at this in part because he's an insightful reviewer but also is the most knowledgeable about the parser.

Thanks for working on this! I'm excited for the potential of switching to IDs for Radio.

packages/perseus/src/widgets/radio/radio-question-builder.test.ts
```diff
@@ -1,3 +1,7 @@
+jest.mock("uuid", () => ({
```
SonicScrewdriver: This should be below the imports!

.changeset/ten-adults-count.md
```diff
@@ -0,0 +1,8 @@
+---
+"@khanacademy/perseus": minor
+"@khanacademy/perseus-core": minor
+"@khanacademy/perseus-editor": minor
+"@khanacademy/perseus-score": minor
+---
+
+adds choice id as a required field
```
SonicScrewdriver: Ignorable nit: It would be good to have some specificity here as I believe this is used for the version packages. (ex: "Adds Choice id as a required field to the Radio Widget")

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
SonicScrewdriver: Makes sense! I'd strongly recommend to provide some clearer context to the code comment, ideally with a specific ticket id for fixing this in the future, just so that the context doesn't get lost to time.

packages/perseus/src/__testdata__/extract-perseus-data.testdata.ts
SonicScrewdriver: I noticed this too! We do have the new radio question builder - I'll make us a ticket to explore updating all our testdata to make sure it's using it wherever possible.

packages/perseus-editor/src/widgets/__tests__/radio-editor.test.tsx
SonicScrewdriver: Agreed, this naming could possibly be clearer

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
benchristel: I actually think index-based IDs might be okay as a long-term solution! But @anakaren-rojas if you know of a reason we need to base the ID on the full path, we should figure that out _before_ this PR lands. Once this change is deployed, any existing radio widgets that are updated by content creators will have these index-based choice IDs written to datastore, and then we'll be stuck with them.

I'd be down to pair on implementing path-based IDs. I think we'd just have to add a `getPath()` method to `ParseContext`, and then pass the `ParseContext` to the callbacks that need the path. See this proof-of-concept: https://github.com/Khan/perseus/pull/2664

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
```diff
@@ -50,6 +55,7 @@ const parseRadioWidgetV3 = parseWidgetWithVersion(
                 rationale: optional(string),
                 correct: optional(boolean),
                 isNoneOfTheAbove: optional(boolean),
+                id: defaulted(string, () => ""),
```
benchristel: This blank ID will be used for existing V3 Radio widgets, which is probably not what we want. Each choice is supposed to have a unique ID, right?

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
```diff
@@ -73,6 +79,7 @@ const parseRadioWidgetV2 = parseWidgetWithVersion(
                     clue: optional(string),
                     correct: optional(boolean),
                     isNoneOfTheAbove: optional(boolean),
+                    id: defaulted(string, () => ""),
```
benchristel: Same here — this default will cause all existing V2 Radio widgets to have blank IDs.

packages/perseus-editor/package.json
anakaren-rojas: great question, I have no idea how we determine that

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
anakaren-rojas: It was mainly for more uniqueness in the ids for the situation where we have multiple radio widgets on one page. The full path isn't necessary, but having the widget id would be nice

packages/perseus-core/src/parse-perseus-json/general-purpose-parsers/array.ts
```diff
@@ -14,6 +14,27 @@ export function array<T>(elementParser: Parser<T>): Parser<T[]> {
...
+
+        const elementResults = rawValue.map((elem, index) => {
+            const elementParser = elementParserFactory(index);
+            return elementParser(elem, ctx.forSubtree(index));
+        });
+
+        return Result.all(elementResults, concat);
+    };
+}
+
```
benchristel, line 37: Cool! Can you add tests for this parser please? (examples in array.test.ts)

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
```diff
@@ -28,7 +29,12 @@ const parseWidgetsMapOrUndefined = defaulted(
 function getDefaultOptions() {
     // See parse-perseus-json/README.md for why we want these defaults here.
     return {
-        choices: [{content: ""}, {content: ""}, {content: ""}, {content: ""}],
+        choices: [
+            {content: "", id: ""},
+            {content: "", id: ""},
+            {content: "", id: ""},
+            {content: "", id: ""},
```
benchristel: Should these IDs be filled in with default values?

```suggestion
            {content: "", id: "radio-choice-0"},
            {content: "", id: "radio-choice-1"},
            {content: "", id: "radio-choice-2"},
            {content: "", id: "radio-choice-3"},
```

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
```diff
@@ -189,12 +198,19 @@ export function migrateV1ToV2(
     widget: ParsedValue<typeof parseRadioWidgetV1>,
 ): ParsedValue<typeof parseRadioWidgetV2> {
     const {options} = widget;
+
+    const choicesForCalculation = options.choices.map((choice, index) => ({
```
benchristel: What does `ForCalculation` mean here? I'm slightly confused by this name. Would `choicesWithIds` work?

```suggestion
    const choicesWithIds = options.choices.map((choice, index) => ({
```

Alternatively, this calculation could be inlined into the `options` object — the IDs aren't needed for `deriveNumCorrect`.

```ts
    return {
        ...widget,
        version: {major: 2, minor: 0},
        options: {
            ...options,
            numCorrect: deriveNumCorrect(options.choices),
            choices: options.choices.map((choice, index) => ({
                ...choice,
                id: `radio-choice-${index}`,
            })),
        },
    };
```

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
```diff
@@ -67,12 +74,13 @@ const parseRadioWidgetV2 = parseWidgetWithVersion(
...
         object({
             numCorrect: optional(number),
-            choices: array(
+            choices: arrayWithIndex((index) =>
                 object({
                     content: defaulted(string, () => ""),
                     clue: optional(string),
                     correct: optional(boolean),
                     isNoneOfTheAbove: optional(boolean),
+                    id: defaulted(string, () => `radio-choice-${index}`),
```
benchristel: This `` `radio-choice-${index}` `` template is used in a bunch of places. Maybe it's worth extracting a function?

```ts
function choiceIdFor(index: number): string {
  return `radio-choice-${index}`;
}
```

packages/perseus-editor/package.json
benchristel: I think this is fine as-is!

`workspace:` dependencies are for when you're importing a package from within the same monorepo. So `workspace:` wouldn't work here. See: https://pnpm.io/workspaces#workspace-protocol-workspace

`catalog:` dependencies are useful when you want the same version of a third-party package to be used by all your packages. This helps keep bundle size low by ensuring that only one version of the third-party package is included in the bundle. https://pnpm.io/catalogs#advantages

If we have multiple Perseus packages that depend on `uuid` in the future, it might be worth moving `uuid` to the `prodDeps` catalog. Or we could just move it now. 🤷

packages/perseus-editor/src/widgets/__tests__/radio-editor.mockData.ts
```diff
@@ -0,0 +1,159 @@
+export const mockIds = [
+    "a1b2c3d4-e5f6-4789-a012-345678901234",
+    "f47ac10b-58cc-4372-a567-0e02b2c3d479",
+    "6ba7b810-9dad-11d1-80b4-00c04fd430c8",
+    "3e4d5c6b-7a89-4012-b345-6789cdef0123",
+    "9c8b7a65-4321-4fed-9876-543210fedcba",
+];
```
benchristel: Most of the elements in this `mockIds` array aren't used in this file. Could the array just be removed?

packages/perseus-editor/src/widgets/__tests__/radio-editor.mockData.ts
```diff
@@ -0,0 +1,159 @@
...
+
+export const fourChoicesWithOneRationale = [
+    {
+        id: "a1b2c3d4-e5f6-4789-a012-345678901234",
+        content: "Choice 1",
+        rationale: "A",
+    },
+    {id: "f47ac10b-58cc-4372-a567-0e02b2c3d479", content: "Choice 2"},
+    {id: "6ba7b810-9dad-11d1-80b4-00c04fd430c8", content: "Choice 3"},
+    {id: "3e4d5c6b-7a89-4012-b345-6789cdef0123", content: "Choice 4"},
```
benchristel: Nit: Is there an advantage to having these be realistic UUIDs? They are kind of long, which means more to write and read in tests.

I'd just as soon use one or two characters and leave it at that. Maybe `f4`, `6b`, `3e`, `a1` would be enough to convey that the IDs are hexadecimal and unordered?

packages/perseus-core/src/data-schema.ts
```diff
@@ -1378,6 +1378,7 @@ export type PerseusRadioWidgetOptions = {
 export type PerseusRadioChoice = {
     // Translatable Markdown; The label for this choice
     content: string;
+    id: string;
```
benchristel, line 1385: Suggestion: It might be nice to include a doc comment for `id`, to document its format and scope of uniqueness. Something like

```suggestion
    /**
     * An opaque string that uniquely identifies this choice within
     * the radio widget. The format of this ID is subject to change.
     */
    id: string;
```

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
benchristel, line 1: Nit: I think the changes to this file could be simplified slightly if we added the ID in version 3 instead of version 2. Then only the version 3 parser would have to change, instead of the parsers for both 2 and 3. I don't think this matters too much, but I just thought I'd point it out.

packages/perseus-editor/package.json
benchristel: Hmm, on second thought, `uuid` is not currently in our prod dependencies for webapp, so... **I think this new dependency would need to go through the ADR process** 😞 .

I notice that the frontend repo has its own in-house UUID implementation, so maybe there's some reason we can't use the `uuid` package from NPM.

packages/perseus-score/src/widgets/radio/score-radio.test.ts
```diff
@@ -113,10 +194,26 @@ describe("scoreRadio", () => {
...
+                    content: "Choice 1",
+                    correct: true,
+                },
+                {
+                    id: "f47ac10b-58cc-4372-a567-0e02b2c3d479",
+                    content: "Choice 2",
+                    correct: false,
+                },
+                {
+                    id: "6ba7b810-9dad-11d1-80b4-00c04fd430c8",
```
benchristel: The UUIDs that LLMs generate are not random. 😂 This one is famous: https://stackoverflow.com/questions/10867405/generating-v5-uuid-what-is-name-and-namespace

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
anakaren-rojas: For calculating the numCorrect, I'll think of a clearer name 
With this change making IDs required in V2, even though it wont use the IDs, `deriveNumCorrect` still expects IDs

benchristel: Oh interesting. The way I handled that [in my proof-of-concept](
https://github.com/Khan/perseus/pull/2664/files#diff-dc8bbc5174c1707b8ed11f39b8c77a9d3bf5a62320d2b2b552042bba39183d80) was by making `deriveNumCorrect` require only the fields it needs (the [interface segregation principle](https://en.wikipedia.org/wiki/Interface_segregation_principle)):

```ts
interface Choice {
    correct?: boolean;
}

export function deriveNumCorrect(choices: Choice[]) {
    return choices.filter((c) => c.correct).length;
}
```

This works because TypeScript uses structural typing. Any object with a `correct: boolean | undefined` property is assignable to this `Choice`, even if it has additional properties.

Not saying you have to do it that way — just an FYI in case it simplifies things.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
anakaren-rojas: the reason I did v2 vs v3 is because with the schema requiring ID, derive num correct function requires ID (even though it doesn't use it) ~ originally I had empty strings as fill ins in v2 and created a v4 that added the id, but decided it would be better to just have v2 do the changes instead

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
anakaren-rojas: I like this idea, also a good reminder to refresh myself on SOLID

anakaren-rojas:
Instead of uuids, I swapped to using crypto.randomUUID @benchristel 
Hopefully that keeps us from having to do another ADR 😆

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
```diff
@@ -67,7 +78,7 @@ const parseRadioWidgetV2 = parseWidgetWithVersion(
     defaulted(
         object({
             numCorrect: optional(number),
-            choices: array(
+            choices: arrayWithIndex((index) =>
```
benchristel: We're not using the index here, so I don't think this line has to change.

```suggestion
            choices: array(
```

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.ts
```diff
@@ -28,7 +29,12 @@ const parseWidgetsMapOrUndefined = defaulted(
 function getDefaultOptions() {
     // See parse-perseus-json/README.md for why we want these defaults here.
     return {
-        choices: [{content: ""}, {content: ""}, {content: ""}, {content: ""}],
+        choices: [
+            {content: "", id: "radio-choice-0"},
+            {content: "", id: "radio-choice-1"},
+            {content: "", id: "radio-choice-2"},
+            {content: "", id: "radio-choice-3"},
```
benchristel: We could use the `generateChoiceId` function here to avoid duplicating the ID format string.

```suggestion
            {content: "", id: generateChoiceId(0)},
            {content: "", id: generateChoiceId(1)},
            {content: "", id: generateChoiceId(2)},
            {content: "", id: generateChoiceId(3)},
```

config/test/crypto-polyfill.js
```diff
@@ -0,0 +1,11 @@
...
+ * Polyfill crypto.randomUUID for Jest test environment
+ * This runs before all test files to ensure crypto is available during module imports
+ */
+Object.defineProperty(global, "crypto", {
+    value: {
+        randomUUID: () => "0-0-0-0-0",
+    },
+    writable: true,
+    configurable: true,
+});
```
benchristel, line 11: This will overwrite the entire `global.crypto` object, which already exists in Jest environments and contains other functions. I think we might want to do something like the below instead, using Node.js's crypto library (webapp does something similar)

```ts
import crypto from "node:crypto";
/**
 * Polyfill crypto.randomUUID for Jest test environment
 * This runs before all test files to ensure crypto is available during module imports
 */
Object.defineProperty(global, "crypto", {
    value: crypto.webcrypto,
    writable: true,
    configurable: true,
});
```

Note that this will cause the UUIDs to be real instead of always `0-0-0-0-0`. I think that's probably a good thing. If individual tests want a deterministic generator instead of a real one, they should probably mock it themselves.

config/test/test.config.js
```diff
@@ -49,6 +49,7 @@ swcrc.jsc.experimental.plugins.push(["swc_mut_cjs_exports", {}]);
 /** @type {import('jest').Config} */
 module.exports = {
     rootDir: path.join(__dirname, "../../"),
+    setupFiles: ["<rootDir>/config/test/crypto-polyfill.js"],
```
benchristel: For some reason, all the other test setup files are listed in `setupFilesAfterEnv`. I don't think it really matters where this one goes, except for readability — maybe there's value in having all the setup files in `setupFilesAfterEnv` so people only have to look in one place.

packages/perseus-core/src/parse-perseus-json/general-purpose-parsers/array.test.ts
```diff
@@ -11,11 +13,11 @@ describe("array", () => {
         expect(arrayOfStrings([], ctx())).toEqual(success([]));
     });
 
-    it("accepts a array of one element", () => {
+    it("accepts an array of one element", () => {
```
benchristel, line 16: Thanks for fixing these test descriptions!

config/test/crypto-polyfill.js
anakaren-rojas: I tried having the individual test files mock randomUUID, but for components that were consuming components that created an id (like radio builder), it wouldn't mock, this ended up being the only thing that worked 😞

packages/perseus-editor/src/widgets/radio/editor.tsx
```diff
@@ -174,6 +174,7 @@ class RadioEditor extends React.Component<RadioEditorProps> {
         const newChoice: PerseusRadioChoice = {
             isNoneOfTheAbove: noneOfTheAbove,
             content: "",
+            id: crypto.randomUUID(),
```
jeremywiebe, line 177: I don't suspect this will cause problems, but wanted to call out that this API is only available in HTTPS (secure) environments. 

https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID

Do we know what the failure mode is when this API isn't available?  

I don't mean to cause re-working of this feature, but there is a crypto API that is available in insecure contexts that might also be workable for this: https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues

PR #2765, opened 2025-08-05
Update snapshots

nishasy:
## Summary:
I noticed some snapshots were updating, unrelated to my changes somewhere else.

I realized they were on main, so I'm opening up this PR to update these.

Issue: none

## Test plan:
`pnpm jest`


SonicScrewdriver (approved):
Happy to approve the update of the snapshots! We might want to check if we should sync deps with wonderblocks again, just in case it changes these snapshots again.

PR #2767, opened 2025-08-05
Add Perseus logo in Storybook

ivyolamit:
## Summary:
Add perseus logo in storybook

| Before | After |
| ------ | ------ |
| image | image |


.storybook/manager.ts
```diff
@@ -6,4 +6,51 @@ addons.setConfig({
...
+    theme: {
+        base: "light",
+
+        // Brand
+        brandTitle: "Perseus",
+        brandUrl: "/",
+        brandImage: "./perseus-logo.svg",
+        brandTarget: "_self",
+
+        // Colors
```
ivyolamit: needed to add the following since they're required props for `theme`.

catandthemachines: Could we use wonderblocks colors for this area so they don't get out of sync? Or could you explain where these colors come from?

ivyolamit: For now they're just temporary colors to satisfy the `theme` required props. The next PR will do is for theming, and that will follow the WonderBlocks colors. I didn't add it in this PR because it would require me to also add the WB theming package.

ivyolamit: Update: I found a way so I don't have to provide all the required props and only set what we need. Also used semantic colors as suggested 😉

PR #2768, opened 2025-08-06
docs(changeset): [Interactive Graph] | (DX) | Shorten interactive-graph-editor.tsx to avoid 1000+ line lint error

nishasy:
## Summary:
I wanted to add a couple lines of code in a different branch, and I got a lint error saying
that this file (interactive-graph-editor.tsx) is too long.

Shortening it here by breaking out the individual graph specifications into a separate file.

Issue: none

## Test plan:
`pnpm jest`

Storybook
http://localhost:6006/?path=/docs/widgets-interactive-graph-editor-demo--docs
- Confirm that all the graph types' specific settings still work right

### DIFF
--- a/packages/perseus-editor/src/widgets/interactive-graph-editor/components/graph-type-answer-specifications.tsx
+++ b/packages/perseus-editor/src/widgets/interactive-graph-editor/components/graph-type-answer-specifications.tsx
@@ -1,0 +1,337 @@
+import {components} from "@khanacademy/perseus";
+import {View} from "@khanacademy/wonder-blocks-core";
+import {OptionItem, SingleSelect} from "@khanacademy/wonder-blocks-dropdown";
+import {Checkbox} from "@khanacademy/wonder-blocks-form";
+import {spacing} from "@khanacademy/wonder-blocks-tokens";
+import {LabelSmall} from "@khanacademy/wonder-blocks-typography";
+import {StyleSheet} from "aphrodite";
+import * as React from "react";
+import invariant from "tiny-invariant";
+import _ from "underscore";
+
+import {parsePointCount} from "../../../util/points";
+import LabeledRow from "../locked-figures/labeled-row";
+
+import GraphPointsCountSelector from "./graph-points-count-selector";
+import SegmentCountSelector from "./segment-count-selector";
+
+import type {
+    PerseusGraphType,
+    PerseusGraphTypePolygon,
+} from "@khanacademy/perseus-core";
+
+const {InfoTip} = components;
+
+const POLYGON_SIDES = _.map(_.range(3, 13), function (value) {
+    return (
+        <OptionItem
+            key={`polygon-sides-${value}`}
+            value={`${value}`}
+            label={`${value} sides`}
+        />
+    );
+});
+
+interface Props {
+    correct: PerseusGraphType;
+    graph: PerseusGraphType | undefined;
+    onChange: (props: Partial<Props>) => void;
+}
+
+export const GraphTypeAnswerSpecifications = ({
+    correct,
+    graph,
+    onChange,
+}: Props) => {
+    return (
+        <>
+            {correct?.type === "point" && (
+                <LabeledRow label="Number of Points:">
+                    <GraphPointsCountSelector
+                        numPoints={correct?.numPoints}
+                        onChange={(points) => {
+                            onChange({
+                                correct: {
+                                    type: "point",
+                                    numPoints: points,
+                                },
+                                graph: {
+                                    type: "point",
+                                    numPoints: points,
+                                },
+                            });
+                        }}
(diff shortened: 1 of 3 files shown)


packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
nishasy, line 1: Question: Starting on line 425, there are a few more specific settings. I'm not sure why they're in a different spot than the other ones? Should I consolidate them to the same place in the new file?

nishasy: Like these are both in polygon?

| Top | Bottom |
| --- | --- |
| Screenshot 2025-08-05 at 5 31 33 PM | Screenshot 2025-08-05 at 5 31 55 PM |

The bottom one isn't even a common setting, so I feel like it shouldn't be under common settings?

nishasy: I'm asking content authors to be sure.

packages/perseus-editor/src/widgets/interactive-graph-editor/components/graph-type-answer-specifications.tsx
```diff
@@ -0,0 +1,337 @@
...
+                            )}
+                            {correct?.numSides !== "unlimited" && (
+                                <OptionItem
+                                    value="sides"
+                                    label="side measures"
+                                />
+                            )}
+                        </SingleSelect>
+                        <InfoTip>
+                            <p>
```
ivyolamit: Do we need translations on these?

packages/perseus-editor/src/widgets/interactive-graph-editor/components/graph-type-answer-specifications.tsx
```diff
@@ -0,0 +1,337 @@
...
+}
+
+export const GraphTypeAnswerSpecifications = ({
+    correct,
+    graph,
+    onChange,
+}: Props) => {
+    return (
+        <>
+            {correct?.type === "point" && (
```
ivyolamit: For code readability/maintenance suggestion:
```
{correct?.type === "point" && (<GraphPointsCounter numPoints={correct?.numPoints} onChange={onChange} />)
{correct?.type === "angle" && (<AngleMeasures graphType={graph?.type} onChange={onChange} />)
{correct?.type === "polygon" && (<PolygonSidesCounter ... />)
{correct?.type === "segment" && ( <SegmentCounter ... />)
```
Smaller chunks of code is easier to read and eventually easier to maintain/debug. Also internal components name are just samples, you can name it appropriately.

ivyolamit (approved):
@nishasy thanks for refactoring this, interactive graph editor file looks much better 🫶🏻. I have one non-blocking suggestion to further improve the code to make it easier to read and maintain. Additionally by doing smaller internal components makes it easier to add test for each one.

packages/perseus-editor/src/widgets/interactive-graph-editor/components/graph-type-answer-specifications.tsx
benchristel: +1 — I think with Ivy's suggestion the graph-type-specific components could each live in their own file. The `GraphTypeAnswerSpecifications` component could be inlined back into `InteractiveGraphEditor`. 

I think the benefit of this approach is that readers of the code can focus on the specific graph subtype they're interested in.

packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
nishasy: Content authors said it makes sense for everything to be in the same spot, so I'll move the bottom ones up to the top ones.

packages/perseus-editor/src/widgets/interactive-graph-editor/components/graph-type-answer-specifications.tsx
nishasy: No, we don't need translations in the editors.

packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
nishasy, line 1: It's not loading this file by default because it's too big, but this is where all the individual sections got broken out of into the new/modified files.

nishasy:
@ivyolamit @benchristel Just letting you know that I re-requested a review on this! It's still showing up as approved, so I don't know if it let you know.

ivyolamit (approved):
Thanks for the update, changes still look good to me 👍🏼

PR #2769, opened 2025-08-06
docs(changeset): Remove unused code and usages of findDOMNode

SonicScrewdriver:
## Summary:
[cleaning-the-perseus-finddom] SSSSOMEBODY STOP ME
[cleaning-the-perseus-finddom] Initial cleanup

Issue: LEMS-XXXX

## Test plan:


packages/perseus/src/renderer.tsx
```diff
@@ -1421,29 +1416,6 @@ class Renderer
         );
     };
 
-    handleRender: (prevProps: Props) => void = (prevProps: Props) => {
```
SonicScrewdriver, line 1424: Tested this earlier with Jeremy and we're fairly certain nothing is calling this. I'm going to be testing this upstream to see if I can notice any regressions.

packages/perseus/src/widgets/matrix/matrix.tsx
```diff
@@ -462,7 +461,6 @@ class Matrix extends React.Component<Props, State> implements Widget {
                                             <SimpleKeypadInput
                                                 {...inputProps}
                                                 style={style}
-                                                scrollable={true}
```
SonicScrewdriver, line 465: This is a nonsense property that doesn't exist, so I've removed it! We simply didn't catch it before as SimpleKeypadInput wasn't very specific about props

packages/perseus/src/widgets/number-line/number-line.tsx
```diff
@@ -247,6 +246,8 @@ class NumberLine extends React.Component<Props, State> implements Widget {
     static contextType = PerseusI18nContext;
     declare context: React.ContextType<typeof PerseusI18nContext>;
 
+    tickControlRef: HTMLInputElement | null = null;
```
SonicScrewdriver, line 248: There's only ever one input in NumberLine, so I figured I'd simplify this accordingly.

packages/perseus/src/widgets/number-line/number-line.tsx
```diff
@@ -346,41 +347,39 @@ class NumberLine extends React.Component<Props, State> implements Widget {
...
-            // eslint-disable-next-line react/no-string-refs
-            // @ts-expect-error - TS2339 - Property 'blur' does not exist on type 'ReactInstance'.
-            this.refs[path[0]].blur();
+            this.tickControlRef?.blur();
         }
     };
 
+    // There's only one input path for the tick control, but the renderer
+    // expects this method to be implemented.
     getInputPaths: () => ReadonlyArray<ReadonlyArray<string>> = () => {
```
SonicScrewdriver, line 369: Since these methods are part of our our main renderer logic, I didn't want to mess with them too much. 

I think improving this will require a more holistic and modern approach to our focus handling across Perseus (perhaps a context?).

packages/perseus/src/widgets/matrix/matrix.tsx
```diff
@@ -201,8 +199,8 @@ class Matrix extends React.Component<Props, State> implements Widget {
         const maxCol = this.props.matrixBoardSize[1];
         let enterTheMatrix = null;
 
-        // eslint-disable-next-line react/no-string-refs
-        const curInput = this.refs[getRefForPath(getInputPath(row, col))];
+        const inputID = getRefForPath(getInputPath(row, col));
+        const curInput = this.answerRefs[inputID];
         // @ts-expect-error - TS2339 - Property 'getStringValue' does not exist on type 'ReactInstance'.
         const curValueString = curInput.getStringValue();
```
SonicScrewdriver, line 199: I can't remove these ts-expect errors as the SimpleKeypadInput never supported these methods, although our TextInput does. (getStringValue, getSelectionStart, getSelectionEnd). 

This means that none of this keyboard logic was ever working for our mobile users. However, they were also very unlikely to hit this as it requires hitting specific keys on their mobile device's keyboard.

packages/perseus/src/widgets/table/table.tsx
```diff
@@ -138,31 +138,22 @@ class Table extends React.Component<Props> implements Widget {
     focusInputPath(path: FocusPath): void {
         const inputID = getRefForPath(path as Path);
         const inputComponent = this.answerRefs[inputID];
-        if (this.props.apiOptions.customKeypad) {
-            inputComponent.focus();
-        } else {
-            // @ts-expect-error - TS2531 - Object is possibly 'null'. | TS2339 - Property 'focus' does not exist on type 'Element | Text'.
-            ReactDOM.findDOMNode(inputComponent).focus();
-        }
```
SonicScrewdriver, line 146: This is the section of code that was able to be cleaned up by converting SimpleKeypadInput into the forward ref functional component

packages/perseus/src/widgets/number-line/number-line.tsx
```diff
@@ -679,8 +678,9 @@ class NumberLine extends React.Component<Props, State> implements Widget {
                 <label>
                     {strings.numDivisions}{" "}
                     <Input
-                        // eslint-disable-next-line react/no-string-refs
-                        ref="tick-ctrl"
+                        ref={(ref) => {
+                            this.tickControlRef = ref;
+                        }}
```
SonicScrewdriver, line 675: Ooo I can simplify this

packages/perseus/src/renderer.tsx
```diff
@@ -1533,10 +1511,6 @@ class Renderer
             if (widget?.getDOMNodeForPath) {
                 return widget.getDOMNodeForPath(interWidgetPath);
             }
-            if (interWidgetPath.length === 0) {
-                // @ts-expect-error - TS2345 - Argument of type 'Widget | null | undefined' is not assignable to parameter of type 'ReactInstance | null | undefined'.
-                return ReactDOM.findDOMNode(widget);
-            }
```
SonicScrewdriver, line 1539: We only ever seem to hit line 1536 when we are in a Graded Group, but the length is never zero. I've been unable to find any cases where we ever actually need this fallback. I tested every widget in storybook in both desktop/mobile views. The only place we ever satisfy this condition is when we're explicitly forcing the situation in tests.

SonicScrewdriver: Hmmm after further testing, I can't really seem to see any regressions from removing this entirely. All tests pass in both Perseus and in frontend, and I haven't noticed anything while testing every single widget. I'll make sure to get a full QA pass, but it'd be nice to greatly simplify all of this.

packages/perseus/src/widgets/sorter/sorter.testdata.ts
```diff
@@ -22,3 +22,26 @@ export const question1: PerseusRenderer = {
...
+            graded: true,
+            options: {
+                padding: true,
+                layout: "horizontal",
+                correct: [
+                    "![alttext](https://cdn.kastatic.org/ka-content-images/9cb2cf618c16501d01abf97036deb355d9393949.png)",
+                    "![alttext](https://cdn.kastatic.org/ka-content-images/9cb2cf618c16501d01abf97036deb355d9393949.png)",
+                    "![alttext](https://cdn.kastatic.org/ka-content-images/9cb2cf618c16501d01abf97036deb355d9393949.png)",
+                    "![alttext](https://cdn.kastatic.org/ka-content-images/9cb2cf618c16501d01abf97036deb355d9393949.png)![alttext](https://cdn.kastatic.org/ka-content-images/9cb2cf618c16501d01abf97036deb355d9393949.png)",
+                ],
```
SonicScrewdriver, line 43: I wanted a story with images in a sorter, to see if there were any regressions. It seems to work identically!

PR #2771, opened 2025-08-06
[Interactive Graph] | (CX) | Use dynamically resizing textarea for Description field

nishasy:
## Summary:
The Interactive Graph "Description" field takes up a ton of vertical space by default.

Instead of having fixed rows, I'm going to use the dynamically resizing textarea
that was previously only used by radio.

- Move AutoResizingTextArea to the shared editor components folder
- Use it inside the interactive graph editor

Issue: none

| Before | After | 
| --- | --- |
| Screenshot 2025-08-06 at 2 57 46 PM |  Screenshot 2025-08-06 at 2 57 52 PM |

Demo

https://github.com/user-attachments/assets/c3def134-9736-4c35-93c0-f4eb71086ce5

## Test plan:
Storybook
- http://localhost:6006/?path=/story/editors-editorpage--demo
- Add an interactive graph widget
- Check that the Description field is only one line by default now
  when it's empty.
- Type in it so it takes up multiple lines and confirm that it
  resizes dynamically


packages/perseus-editor/src/widgets/interactive-graph-editor/components/interactive-graph-description.tsx
```diff
@@ -52,23 +56,21 @@ export default function InteractiveGraphDescription(props: Props) {
...
-                            onChange={(newValue) =>
-                                onChange({
-                                    fullGraphAriaDescription:
-                                        // Save as undefined if the new value
-                                        // is an empty string.
-                                        newValue || undefined,
-                                })
-                            }
-                            style={styles.spaceAbove}
-                        />
```
nishasy, line 70: It won't let me use the textarea as the child of a label anymore after changing it to a non-wonder blocks component.

mark-fitzgerald (approved):
Niiiiice!

PR #2773, opened 2025-08-06
Add Theme button in Perseus storybook to allow theme switching

ivyolamit:
## Summary:
Add Theme button in Perseus storybook to allow theme switching

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -1,6 +1,7 @@
 import * as React from "react";
 import {color} from "@khanacademy/wonder-blocks-tokens";
 import {RenderStateRoot} from "@khanacademy/wonder-blocks-core";
+import {ThemeSwitcher} from "@khanacademy/wonder-blocks-theming";
 
 import {
     setDependencies,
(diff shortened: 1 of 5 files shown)


pnpm-workspace.yaml
```diff
@@ -21,6 +21,12 @@
 # The sync-dependencies.ts script ensures that peer deps are always
 # specified as a range, and the dev deps are always pinned to the
 # version at the bottom of that range.
+#
+# To add a new package:
+# 1. Add the new package both in peerDeps and devDeps
+# 2. Then add it to whichever Perseus package you want
+#    e.g. "@khanacademy/wonder-blocks-link": "catalog:devDeps", in dev/package.json
+# 3. Run pnpm install in root
```
nishasy, line 31: Thank you for adding these instructions!

.storybook/preview.tsx
```diff
@@ -17,35 +18,70 @@ import "./reset.css";
...
+                    from prod.
+                */}
+                <div className="framework-perseus box-sizing-border-box-reset">
+                    {/* The ThemeSwitcher component applies the selected theme to all children.*/}
+                    <ThemeSwitcher theme={theme}>
+                        <Story />
+                    </ThemeSwitcher>
+                </div>
+            </DependenciesContext.Provider>
+        </RenderStateRoot>
```
jeremywiebe: **nit**: I wonder if we could split this into two decorators: one for theme switching (lines 43-45) and another for Perseus hosting (the outer parts).  

As it is now, the Dependencies and `div` with `framework-perseus` have little, or nothing, to do with theming.

pnpm-workspace.yaml
jeremywiebe: Item 2 should call for adding the package to both `devDependencies` and `peerDependencies` in the `package.json` of choice. This is because `devDependencies` dictate the package and version we'll use when running tests and doing local dev, `peerDependencies` is the package's declaration of what versions it is safe to use with in hosting applications.

.storybook/preview.tsx
```diff
@@ -17,35 +18,75 @@ import "./reset.css";
...
-                        the global styles from prod.
-                    */}
-                    <div className="framework-perseus box-sizing-border-box-reset">
-                        <Story />
-                    </div>
-                </DependenciesContext.Provider>
-            </RenderStateRoot>
-        ),
-    ],
+    decorators: [withPerseusDecorator, withThemeSwitcher],
```
jeremywiebe, line 84: Thanks for splitting this!

PR #2775, opened 2025-08-07
Fix parsing errors by fixing Matrix user input type

Myranae:
## Summary:
Matrix user input has always been received as array of string arrays. The type incorrectly defined it as numbers instead of strings, which caused parsing errors. This updates PerseusMatrixUserInput to reflect the actual type of user input, updates test files to show user input as strings,

Issue: LEMS-3342

## Test plan:
- Confirm all checks pass
- Confirm Matrix still works as expected via Storybook
- After deploying, confirm related parsing errors decrease


packages/perseus-score/src/widgets/matrix/score-matrix.ts
```diff
@@ -20,7 +20,9 @@ function scoreMatrix(
     }
 
     const solution = rubric.answers;
-    const supplied = userInput.answers;
+    const supplied = userInput.answers.map((row) =>
+        row.map((str) => Number(str)),
+    );
```
Myranae: `getMatrixSize` says it requires array of array of numbers, but then the logic inside the function converts each number to a string. Maybe I should just update that function to take an array of arrays of strings and remove the .toString() call instead of converting here and then converting back there 🤔  Except we also call `getMatrixSize` with the solution, which is an array of arrays of numbers.

packages/perseus/src/widgets/matrix/matrix.tsx
```diff
@@ -537,7 +542,7 @@ function getStartUserInput(): PerseusMatrixUserInput {
 
 function getCorrectUserInput(
     options: PerseusMatrixWidgetOptions,
-): PerseusMatrixUserInput {
+): PerseusMatrixRubric {
```
Myranae: Not sure about this change. Going to double check how this function is used.

handeyeco: The return value has to be `PerseusMatrixUserInput`. `getCorrectUserInput` does what it says on the tin: it takes answerful widget options and returns the user input in the correct state.

The problem you're running into is that the type PerseusMatrixWidgetOptions stores the correct answer in (numbers) is different than the way PerseusMatrixUserInput stores user input (strings).

A quick fix: convert the numbers to strings in `getCorrectUserInput`. A more holistic (out of scope) fix is to make PerseusMatrixWidgetOptions and PerseusMatrixUserInput use the same type for user input and correct answer.

packages/perseus-score/src/widgets/matrix/score-matrix.ts
handeyeco: Maybe we can make it take either `string[][] | number[][]` and convert it to `string[][]` in `getMatrixSize`.

Myranae: Updated as you suggested :)

packages/perseus/src/widgets/matrix/matrix.tsx
Myranae: Reverted return type and converted the numbers inside the function.

handeyeco (approved):
Thanks!

PR #2779, opened 2025-08-08
Remove ScrollableView left and right fade

ivyolamit:
## Summary:
Remove ScrollableView left and right fade

Issue: LEMS-3108

## Test plan:

### DIFF
--- a/packages/perseus/src/components/scrollable-view.module.css
+++ b/packages/perseus/src/components/scrollable-view.module.css
@@ -5,31 +5,3 @@
     gap: 0.8rem;
     padding: 1.2rem;
 }
-
-.scroll-fade {
-    position: absolute;
-    top: 0;
-    height: 100%;
-    width: max-content;
-    pointer-events: none;
-    z-index: 1;
-    transition: opacity 0.3s ease;
-}
-
-.scroll-fade-left {
-    left: 0;
-    background: linear-gradient(
-        to right,
-        rgba(255, 255, 255, 1),
-        rgba(255, 255, 255, 0)
-    );
-}
-
-.scroll-fade-right {
-    right: 0;
-    background: linear-gradient(
-        to left,
-        rgba(255, 255, 255, 1),
-        rgba(255, 255, 255, 0)
-    );
-}
--- a/packages/perseus/src/widgets/radio/multiple-choice.module.css
+++ b/packages/perseus/src/widgets/radio/multiple-choice.module.css
@@ -65,11 +65,11 @@
 
 .choice-list:dir(ltr):before,
 .choice-list:dir(rtl):after {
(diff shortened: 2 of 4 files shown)


packages/perseus/src/components/scrollable-view.module.css
```diff
@@ -5,31 +5,3 @@
     gap: 0.8rem;
     padding: 1.2rem;
 }
-
```
ivyolamit, line 8: These styling is no longer necessary and redundant, fade is already done in multiple-choice.

packages/perseus/src/widgets/radio/multiple-choice.module.css
```diff
@@ -65,11 +65,11 @@
 
 .choice-list:dir(ltr):before,
 .choice-list:dir(rtl):after {
-    /* TODO: The following gradient should be updated to use appropriate background color to automatically handle when user has chosen dark mode. */
     background: linear-gradient(
         to left,
         transparent,
-        white calc(var(--perseus-multiple-choice-spacing) / 2)
+        var(--wb-semanticColor-core-foreground-inverse-strong)
```
ivyolamit: use semantic color here equivalent to white so that when WB supports the dark theme mode, this will be the inverse color.

SonicScrewdriver (approved):
Love it, this is a great improvement. Thank you!!

packages/perseus/src/widgets/radio/multiple-choice.module.css
SonicScrewdriver: You could possibly add a comment here explaining what this styling is doing -- it might help new engineers understand the purpose of the styling. :)

ivyolamit: existing comment in line 40 and 85 indicates
```
/* Fade bars for scrolling (Start) */
```
and
```
/* Fade bars for scrolling (End) */
```

Is there any specific additional info that would be helpful?

mark-fitzgerald: @ivyolamit - Could you check in with WB team to find out if they have a better semantic name for something like this? Using a `foreground` variable for a background color seems odd to me. I'm curious if they plan on adding a semantic background color that reflects the theme and/or light/dark mode color that is in place.

ivyolamit: Good point! Ok i'll check with them 👍🏼

ivyolamit: as chatted in slack...

existing comment in line 40 and 85 indicates
/* Fade bars for scrolling (Start) */
and
/* Fade bars for scrolling (End) */

PR #2784, opened 2025-08-11
docs(changeset): Create a PR snapshot that updates all packages for Bump Script Tests

SonicScrewdriver:
## Summary:
<Add your summary here>

Issue: LEMS-3362

## Test plan:

### DIFF
--- a/.changeset/pink-countries-camp.md
+++ b/.changeset/pink-countries-camp.md
@@ -1,0 +1,20 @@
+---
+"perseus-build-settings": minor
+"@khanacademy/perseus-dev-ui": minor
+"@khanacademy/kas": minor
+"@khanacademy/keypad-context": minor
+"@khanacademy/kmath": minor
+"@khanacademy/math-input": minor
+"@khanacademy/perseus": minor
+"@khanacademy/perseus-core": minor
+"@khanacademy/perseus-editor": minor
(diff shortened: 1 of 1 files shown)


.changeset/pink-countries-camp.md
```diff
@@ -0,0 +1,20 @@
+---
```
SonicScrewdriver, line 1: Hmmm I set it to default update all the packages available in the script, but I wonder if I should manually add some. I know we have perseus-score that seems like it should be in here?

.changeset/pink-countries-camp.md
```diff
@@ -0,0 +1,20 @@
...
+"@khanacademy/perseus": minor
+"@khanacademy/perseus-core": minor
+"@khanacademy/perseus-editor": minor
+"@khanacademy/perseus-linter": minor
+"@khanacademy/perseus-score": minor
+"@khanacademy/perseus-utils": minor
+"@khanacademy/pure-markdown": minor
+"@khanacademy/simple-markdown": minor
+"jsdiff": minor
+"raphael": minor
```
SonicScrewdriver, line 17: Hmm I suspect jsdiff and raphael are not necessary / useful for our testing purposes, but I guess they're not harmful for our testing purposes.

PR #2788, opened 2025-08-12
Fix strong styling issue not properly applying to rendered text

ivyolamit:
## Summary:
Fix strong styling issue not properly applying to rendered text in Storybook.

| Before | After |
| ------ | ------ |
| image | image |

## Test plan:
1. Navigate to the editor page docs
2. Input the `**hello** world` text in the question field
- [ ] Confirm that the strong styling is applied


jeremywiebe:
I'm concerned that this is not the correct way to fix this. We've had this `reset.css` in place for about 4 months and it's exactly what we have in prod.  By modifying this file we now diverge from prod _and_ potentially fix this in a way that's specific to `<strong>` elements but leave more subtle styling breakage in place. Have you bisected back to check when the bold styling broke?

ivyolamit:
> I'm concerned that this is not the correct way to fix this. We've had this `reset.css` in place for about 4 months and it's exactly what we have in prod. By modifying this file we now diverge from prod _and_ potentially fix this in a way that's specific to `\<strong\>` elements but leave more subtle styling breakage in place. Have you bisected back to check when the bold styling broke?

Good point, yes i did bisect earlier and went all the way before I made storybook changes it was the same from current. Let me check a different approach.

.storybook/global.css
```diff
@@ -1,25 +0,0 @@
-/**
```
ivyolamit, line 1: renamed file to `./styles/shared.css` reorganizing files and styling

.storybook/preview.tsx
```diff
@@ -12,11 +12,7 @@ import {
     storybookDependenciesV2,
 } from "../testing/test-dependencies";
 
-import "./global.css";
-import "./reset.css";
```
ivyolamit, line 16: these are now available in `./styles/shared.css`

.storybook/styles/reset.css
```diff
@@ -91,7 +91,6 @@
         font: inherit;
         vertical-align: baseline;
     }
-
```
ivyolamit, line 94: Prettier did this 🤷🏻‍♀️

.storybook/styles/base-element-styles.css
```diff
@@ -0,0 +1,114 @@
+@layer shared {
```
ivyolamit, line 1: the remaining css files are based on `frontend/libs/app-shell/src/styles`

.storybook/styles/shared.css
```diff
@@ -0,0 +1,26 @@
+/**
+ * This file contains styles that are used in prod that ultimately
+ * get passed onto Perseus components. This file is imported in the
+ * .storybook/preview.tsx file so that the components can have the same
+ * styles as prod when viewed within Storybook.
+ */
+
```
ivyolamit, line 7: This used to be `global.css` i like how our frontend organize imports it's cleaner and readable so I applied it here.

ivyolamit:
Recent changes undid previous solution and updated styling to reflect latest changes from prod. Also organizes styling files for easy maintainability.

| Before | After |
| ------ | ------ |
| image | image |

.storybook/preview.tsx
```diff
@@ -12,11 +12,7 @@ import {
...
-
-// Import the Wonder Blocks CSS variables
-import "@khanacademy/wonder-blocks-tokens/styles.css";
+import "./styles/shared.css";
```
mark-fitzgerald, line 17: Perhaps add a note here that this is bringing in the shared styles from PROD.

mark-fitzgerald (approved):
Thanks for sorting this out!

Probably not necessary for this PR, but we may want to consider how the [ThemeSwitcher works in our implementation](https://github.com/Khan/perseus/blob/main/.storybook/preview.tsx#L51). This seems to be working fine in our regular stories, but isn't causing the editor preview section to inherit the SYL theme when switched. Maybe a ticket for another time?

ivyolamit:
> Thanks for sorting this out!
> 
> Probably not necessary for this PR, but we may want to consider how the [ThemeSwitcher works in our implementation](https://github.com/Khan/perseus/blob/main/.storybook/preview.tsx#L51). This seems to be working fine in our regular stories, but isn't causing the editor preview section to inherit the SYL theme when switched. Maybe a ticket for another time?

Thanks Mark for the review and pointing out the editor part. I'll address theme part in a different followup PR, it might be related to missing styles I thought it wasn't necessary to bring in. Also was thinking based on feedback that it's in your face having the editor preview show in a modal I'm considering it having it beside in the editor instead.

PR #2791, opened 2025-08-12
Remove transform from Dropdown

handeyeco:
## Summary:
<Add your summary here>

Issue: LEMS-3199

## Test plan:

### DIFF
--- a/packages/perseus/src/widgets/dropdown/dropdown.tsx
+++ b/packages/perseus/src/widgets/dropdown/dropdown.tsx
@@ -16,14 +16,10 @@ import type {
     PerseusDropdownUserInput,
 } from "@khanacademy/perseus-core";
 
-type RenderProps = {
-    placeholder: PerseusDropdownWidgetOptions["placeholder"];
-    visibleLabel: PerseusDropdownWidgetOptions["visibleLabel"];
-    ariaLabel: PerseusDropdownWidgetOptions["ariaLabel"];
-    choices: PerseusDropdownWidgetOptions["choices"];
-};
-
-type Props = WidgetProps<RenderProps, PerseusDropdownUserInput>;
+type Props = WidgetProps<
+    PerseusDropdownWidgetOptions,
+    PerseusDropdownUserInput
+>;
 
 type DefaultProps = {
     choices: Props["choices"];


packages/perseus/src/__tests__/renderer.test.tsx
```diff
@@ -219,19 +219,18 @@ describe("renderer", () => {
             expect(renderer.state.lastUsedWidgetId).toBeNull();
 
             expect(renderer.state.widgetInfo).toStrictEqual(question1.widgets);
-            expect(renderer.state.widgetProps).toMatchInlineSnapshot(`
```
handeyeco, line 222: I changed this test because it's testing against internal Renderer state which is changing to temporarily store Dropdown's WidgetOptions (vs RenderProps). Once the transforms are gone, it won't need to store this in state at all.

packages/perseus/src/widget-ai-utils/dropdown/dropdown-ai-utils.test.ts
```diff
@@ -59,7 +59,12 @@ describe("Dropdown AI utils", () => {
         };
 
         const renderProps: any = {
-            choices: ["Pickles", "Tomato", "Onion", "Lettuce"],
+            choices: [
+                {content: "Pickles"},
+                {content: "Tomato"},
+                {content: "Onion"},
+                {content: "Lettuce"},
```
handeyeco, line 66: AI utils basically take the props the widget receives and returns data for a prompt. Since the props are changing from WidgetOptions to RenderProps, the test data also needed to change.

packages/perseus/src/widget-ai-utils/dropdown/dropdown-ai-utils.ts
```diff
@@ -17,7 +17,7 @@ export const getPromptJSON = (
     return {
         type: "dropdown",
         options: {
-            items: renderProps.choices,
+            items: renderProps.choices.map((choice) => choice.content),
```
handeyeco, line 20: Using WidgetOptions and not RenderProps now.

packages/perseus/src/widgets/dropdown/dropdown.tsx
```diff
@@ -73,13 +72,25 @@ class Dropdown extends React.Component<Props> implements Widget {
...
+            choices: choices.map((choice) => choice.content),
             selected: userInput.value,
         };
     }
 
+    /**
+     * @deprecated and likely very broken API
+     * [LEMS-3185] do not trust serializedState/restoreSerializedState
+     */
+    restoreSerializedState(state: any) {
```
handeyeco, line 83: This is to keep the existing serialized state tests happy.

packages/perseus/src/widgets/dropdown/dropdown.tsx
```diff
@@ -21,7 +20,7 @@ type RenderProps = {
     placeholder: PerseusDropdownWidgetOptions["placeholder"];
     visibleLabel: PerseusDropdownWidgetOptions["visibleLabel"];
     ariaLabel: PerseusDropdownWidgetOptions["ariaLabel"];
-    choices: ReadonlyArray<string>;
+    choices: PerseusDropdownWidgetOptions["choices"];
```
benchristel: Do we still need `RenderProps`? Or can we do

```ts
type Props = WidgetProps<PerseusDropdownWidgetOptions, PerseusDropdownUserInput>
```

?

handeyeco: Good catch!

PR #2793, opened 2025-08-12
Remove transform from InputNumber

handeyeco:
## Summary:
InputNumber's transform was just filtering stuff that got through. We don't care, pass it all through.

Issue: LEMS-3199

### DIFF
--- a/packages/perseus/src/widgets/input-number/input-number.tsx
+++ b/packages/perseus/src/widgets/input-number/input-number.tsx
@@ -50,12 +50,10 @@ const formExamples: Record<string, FormExampleFunction> = {
     },
 } as const;
 
-type RenderProps = Pick<
+type ExternalProps = WidgetProps<
     PerseusInputNumberWidgetOptions,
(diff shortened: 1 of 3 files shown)


packages/perseus/src/widgets/input-number/input-number.tsx
```diff
@@ -162,7 +160,13 @@ class InputNumber extends React.Component<Props> implements Widget {
...
      */
     getSerializedState(): any {
-        const {userInput, ...rest} = this.props;
+        const {
+            userInput,
+            inexact: _,
+            maxError: __,
+            value: ___,
+            ...rest
+        } = this.props;
```
handeyeco: Just to make the serialization test happy

benchristel: Suggestion: I think it would be the same number of lines to list the props that _are_ included, rather than those that aren't. That might be more readable 🤷

PR #2795, opened 2025-08-12
Minor transform changes

handeyeco:
## Summary:
Transform changes that weren't big enough to justify their own PRs.

Issue: LEMS-3199

### DIFF
--- a/.changeset/tiny-crews-behave.md
+++ b/.changeset/tiny-crews-behave.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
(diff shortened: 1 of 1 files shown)


packages/perseus/src/widgets/interactive-graphs/interactive-graph.tsx
```diff
@@ -885,7 +883,6 @@ export default {
     name: "interactive-graph",
     displayName: "Interactive graph",
     widget: InteractiveGraph,
-    staticTransform: _.identity,
```
handeyeco, line 888: We default to `_.identity` anyway when calling.

PR #2797, opened 2025-08-13
Remove NumericInput transform

handeyeco:
## Summary:
<Add your summary here>

Issue: LEMS-3199

## Test plan:

### DIFF
--- a/packages/perseus/src/widgets/numeric-input/numeric-input.class.tsx
+++ b/packages/perseus/src/widgets/numeric-input/numeric-input.class.tsx
@@ -6,7 +6,7 @@ import {ApiOptions} from "../../perseus-api";
 import {getPromptJSON as _getPromptJSON} from "../../widget-ai-utils/numeric-input/prompt-utils";
 
 import {NumericInputComponent} from "./numeric-input";
-import {processAnswerForms} from "./utils";
+import {normalizeCorrectAnswerForms} from "./utils";
 
 import type {Focusable, Widget, WidgetExports, WidgetProps} from "../../types";
 import type {NumericInputPromptJSON} from "../../widget-ai-utils/numeric-input/prompt-utils";
@@ -128,7 +128,7 @@ export class NumericInput
         return (
             <NumericInputComponent
                 {...this.props}
-                answerForms={processAnswerForms(this.props.answers)}
+                answerForms={normalizeCorrectAnswerForms(this.props.answers)}
                 ref={this.inputRef}
             />
         );
--- a/packages/perseus/src/widgets/numeric-input/numeric-input.class.tsx
+++ b/packages/perseus/src/widgets/numeric-input/numeric-input.class.tsx
@@ -6,7 +6,7 @@ import {ApiOptions} from "../../perseus-api";
 import {getPromptJSON as _getPromptJSON} from "../../widget-ai-utils/numeric-input/prompt-utils";
 
 import {NumericInputComponent} from "./numeric-input";
-import {unionAnswerForms} from "./utils";
+import {processAnswerForms} from "./utils";
 
 import type {Focusable, Widget, WidgetExports, WidgetProps} from "../../types";
 import type {NumericInputPromptJSON} from "../../widget-ai-utils/numeric-input/prompt-utils";
@@ -45,15 +45,6 @@ type DefaultProps = Pick<
     | "linterContext"
     | "userInput"
 >;
-
-type RenderProps = {
-    answerForms: ReadonlyArray<PerseusNumericInputAnswerForm>;
-    labelText?: string;
-    size: string;
-    coefficient: boolean;
-    rightAlign?: boolean;
-    static: boolean;
-};
 
 // Assert that the PerseusNumericInputWidgetOptions parsed from JSON can be passed
 // as props to this component. This ensures that the PerseusNumericInputWidgetOptions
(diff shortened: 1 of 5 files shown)


packages/perseus/src/widgets/numeric-input/numeric-input.class.tsx
```diff
@@ -126,43 +117,23 @@ export class NumericInput
...
 }
-// Transforms the widget options to the props used to render the widget.
-const propsTransform = function (
-    widgetOptions: PerseusNumericInputWidgetOptions,
-): RenderProps {
-    // Omit the answers from the widget options since they are
-    // not needed for rendering the widget.
-    const {answers: _, ...rendererProps} = {
-        ...widgetOptions,
-        answerForms: unionAnswerForms(
```
handeyeco, line 148: Instead of doing this as part of the transform, doing this as part of the render.

packages/perseus/src/widgets/numeric-input/numeric-input.test.ts
```diff
@@ -446,81 +445,6 @@ describe("Numeric input widget", () => {
     });
 });
 
-describe("transform", () => {
-    it("removes the answers and extracts the answer forms", () => {
```
handeyeco, line 450: Moved and refactored these to work outside of `transform`

packages/perseus/src/widgets/numeric-input/utils.ts
```diff
@@ -119,3 +122,28 @@ export const unionAnswerForms: (
...
+/**
+ * Filter out the correct answers and map them to the answer forms
+ * so that we can generate the examples for the widget.
+ */
+export function processAnswerForms(
+    answers: PerseusNumericInputWidgetOptions["answers"],
+) {
+    if (answers == null) {
+        return [];
+    }
```
handeyeco, line 138: This has to do with serialization. Serialization doesn't seem to serialize the answer, so restoring the serialized state restores without the answer too. A null check seemed better than trying to fix a deprecated API.

packages/perseus/src/widgets/numeric-input/numeric-input.class.tsx
```diff
@@ -126,43 +117,23 @@ export class NumericInput
      * [LEMS-3185] do not trust serializedState/restoreSerializedState
      */
     getSerializedState() {
-        const {userInput, ...rest} = this.props;
+        const {userInput, answers: _, ...rest} = this.props;
```
handeyeco, line 120: Making the serialization test happy

packages/perseus/src/widgets/numeric-input/utils.ts
benchristel: Sounds good. Could you add a TODO comment here so some future person can clean this up?

packages/perseus/src/widgets/numeric-input/utils.ts
```diff
@@ -119,3 +122,28 @@ export const unionAnswerForms: (
         );
     });
 };
+
+/**
+ * Filter out the correct answers and map them to the answer forms
+ * so that we can generate the examples for the widget.
+ */
+export function processAnswerForms(
```
benchristel: `process` is a vague word. How about `normalizeCorrectAnswerForms` or `collateCorrectAnswerForms`?

Might also be worth adding a return type `PerseusNumericInputAnswerForm[]`, just for documentation purposes.

packages/perseus/src/widgets/numeric-input/utils.test.ts
```diff
@@ -191,3 +196,54 @@ describe("unionAnswerForms", () => {
...
+
+    it("only uses answer forms from correct answers", () => {
+        const result = processAnswerForms([
+            {
+                status: "correct",
+                maxError: null,
+                strict: true,
+                value: 0.5,
+                simplify: "required",
+                answerForms: ["proper"],
```
benchristel, line 268: What about the case where `answerForms` contains more (or less) than one element?

packages/perseus/src/widgets/numeric-input/utils.test.ts
```diff
@@ -191,3 +196,54 @@ describe("unionAnswerForms", () => {
         expect(result).toEqual(expected);
     });
 });
+
+describe("processAnswerForms", () => {
+    it("removes the answers and extracts the answer forms", () => {
```
benchristel: What does `removes the answers` refer to? I guess it's excluding `value` from the output, but it also excludes a bunch of other stuff. I'd just as soon change this test name to something like

```suggestion
    it("pairs answer forms with `simplify` values", () => {
```

packages/perseus/src/widgets/numeric-input/utils.test.ts
```diff
@@ -191,3 +196,54 @@ describe("unionAnswerForms", () => {
...
+            },
+        ]);
+        expect(result).toEqual([
+            {
+                simplify: "required",
+                name: "proper",
+            },
+        ]);
+    });
+});
```
benchristel, line 288: It seems like these tests don't cover the behavior of `unionAnswerForms` — i.e. sorting the answer forms and removing duplicates. Should they?

handeyeco: That has its own test. None of this logic is new to the PR, just moving stuff around. But I think I covered most of your feedback.

PR #2798, opened 2025-08-13
Remove NumberLine transforms

handeyeco:
## Summary:
Remove transforms from NumberLine. This one was a little weird. It seemed like the main point of the transforms was to convert certain `null` value props to `undefined` just so default props would kick in (default props only works with undefined, not null).

This would be a lot easier in a functional component where everything is in the same scope. For the class component I just decided to add some getter methods as an alternative.

Issue: LEMS-3199


packages/perseus/src/widgets/number-line/number-line.tsx
```diff
@@ -625,11 +630,21 @@ class NumberLine extends React.Component<Props, State> implements Widget {
      * [LEMS-3185] do not trust serializedState/restoreSerializedState
      */
     getSerializedState() {
-        const {userInput, ...rest} = this.props;
+        const props = this.props;
         return {
-            ...rest,
-            numDivisions: userInput.numDivisions,
-            numLinePosition: userInput.numLinePosition,
+            alignment: props.alignment,
```
handeyeco, line 635: This is just to filter some stuff the serialization function isn't expecting.

packages/perseus/src/widgets/number-line/number-line.tsx
```diff
@@ -727,57 +742,6 @@ class NumberLine extends React.Component<Props, State> implements Widget {
...
-
-        "isTickCtrl",
-        "isInequality",
-
-        "showTooltips",
-    ]);
-
-    return {
-        ...props,
-        isTickCtrl: editorProps.isTickCtrl ?? undefined,
```
handeyeco, line 749: Yeah, idk, this whole thing seems pretty weird to me. It's a very indirect way to set a value: convert null to undefined just to let default props set the value... 🤷

catandthemachines (approved):
LGTM!

PR #2799, opened 2025-08-13
feat(LEMS-3191): Set up Radio scoring to choice Ids

anakaren-rojas:
## Summary:
- Creates new type `UserSelectedChoices` made up of an object with an id and selected field
- Converts `choicesSelected` type from `boolean[]` to `UserSelectedChoices[]` 
- Creates element parser to consume new `UserSelectedChoices[]`

Issue: LEMS-3191

## Test plan:

### DIFF
--- a/packages/perseus-score/src/widgets/radio/score-radio.ts
+++ b/packages/perseus-score/src/widgets/radio/score-radio.ts
@@ -17,8 +17,8 @@ function scoreRadio(
         return {type: "invalid", message: null};
     }
 
-    const numSelected = userInput.choicesSelected.reduce((sum, selected) => {
-        return sum + (selected ? 1 : 0);
+    const numSelected = userInput.choicesSelected.reduce((sum, choice) => {
+        return sum + (choice.selected ? 1 : 0);
     }, 0);
 
     const numCorrect: number = rubric.choices.reduce((sum, currentChoice) => {
@@ -35,9 +35,10 @@ function scoreRadio(
         };
     }
 
+
     const noneOfTheAboveSelected = rubric.choices.some(
-        (choice, index) =>
-            choice.isNoneOfTheAbove && userInput.choicesSelected[index],
+        (choice) =>
+            choice.isNoneOfTheAbove && userInput.choicesSelected.some(userChoice => userChoice.id === choice.id && userChoice.selected),
     );
 
     if (noneOfTheAboveSelected && numSelected > 1) {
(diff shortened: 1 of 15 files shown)


packages/perseus/src/__tests__/extract-perseus-data.test.ts
```diff
@@ -1195,4 +1196,124 @@ describe("ExtractPerseusData", () => {
             ]);
         });
     });
+
```
anakaren-rojas, line 1199: is this function still needed??? I didn't see it being used anywhere, but wrote some tests just in case it was 🤔

packages/perseus/src/widgets/radio/radio.ts
```diff
@@ -114,19 +113,19 @@ const transform = (
     };
 };
 
-function getStartUserInput(
-    options: RadioPublicWidgetOptions,
-): PerseusRadioUserInput {
+function getStartUserInput(): PerseusRadioUserInput {
```
anakaren-rojas, line 116: Same question as above: is this function still needed?? I don't see it used anywhere

packages/perseus-score/src/widgets/radio/score-radio.ts
```diff
@@ -47,16 +54,10 @@ function scoreRadio(
...
-                return i === j || !choice.correct;
-            });
-        } else {
-            isCorrect = !!rubric.choices[i].correct;
-        }
-        return isCorrect === selected;
+    const correct = rubric.choices.every((choice) => {
+        const isSelected = userInput.selectedChoiceIds.includes(choice.id);
+        const isCorrect = !!choice.correct;
+        return isCorrect === isSelected;
```
mark-fitzgerald, line 60: Much easier to read and reason about. Thank you!!

packages/perseus/src/widgets/radio/radio.ts
ivyolamit: Looks like this is used on a higher level, did a quick search and saw this https://github.com/Khan/perseus/blob/c6171cd82ed4dc9555f34431375a09c316be402a/packages/perseus/src/user-input-manager.tsx#L66

packages/perseus-core/src/parse-perseus-json/parse-perseus-json.test.ts
```diff
@@ -185,21 +185,21 @@ describe("parseAndMigratePerseusArticle", () => {
 describe("parseAndMigrateUserInputMap", () => {
     it("parses a UserInputMap from a JSON string", () => {
         const result = parseAndMigrateUserInputMap(
-            `{"radio 1": {"choicesSelected": [true, false]}}`,
+            `{"radio 1": {"selectedChoiceIds": [ "0-0-0-0-0"]}}`,
```
ivyolamit, line 188: Just a nit, non-blocking comment: Can we have a sample choice id that's closer on what it normally look like? I was a bit confused earlier reading them 😅

packages/perseus/src/widgets/radio/radio.ts
anakaren-rojas: oh nice! I hadn't seen it in my search, but good to know its _is_ being used, now we have tests

ivyolamit (approved):

packages/perseus-core/src/parse-perseus-json/parse-perseus-json.test.ts
anakaren-rojas: I was specifically asked in my last PR to _not_ use realistic ids because they were difficult to read 😆 
I could do something like `1-2-3-4-5` if that makes it easier? 
A realistic one looks like this: `36b8f84d-df4e-4d49-b662-bcde71a8764f`

PR #2800, opened 2025-08-13
Prework cleanup

handeyeco:
## Summary:
<Add your summary here>

Issue: LEMS-3199

## Test plan:

### DIFF
--- a/packages/perseus/src/widgets/radio/radio.ts
+++ b/packages/perseus/src/widgets/radio/radio.ts
@@ -1,8 +1,13 @@
-import {radioLogic, randomizeArrayElements} from "@khanacademy/perseus-core";
+import {radioLogic} from "@khanacademy/perseus-core";
 import _ from "underscore";
 
 import Radio from "./radio.ff";
-import {getUserInputFromSerializedState} from "./util";
+import {
+    addNoneOfAbove,
+    enforceOrdering,
+    getUserInputFromSerializedState,
+    maybeRandomize,
+} from "./util";
 
 import type {RenderProps, RadioChoiceWithMetadata} from "./radio-component";
 import type {PerseusStrings} from "../../strings";
@@ -18,53 +23,6 @@ const _choiceTransform = (
     widgetOptions: PerseusRadioWidgetOptions,
     strings: PerseusStrings,
 ) => {
-    const _maybeRandomize = function (
-        array: ReadonlyArray<RadioChoiceWithMetadata>,
-    ) {
-        return widgetOptions.randomize ? randomizeArrayElements(array) : array;
-    };
-
-    const _addNoneOfAbove = function (
-        choices: ReadonlyArray<RadioChoiceWithMetadata>,
-    ) {
-        let noneOfTheAbove = null;
-
-        const newChoices = choices.filter((choice, index) => {
-            if (choice.isNoneOfTheAbove) {
-                // @ts-expect-error - TS2322 - Type 'RadioChoiceWithMetadata' is not assignable to type 'null'.
-                noneOfTheAbove = choice;
-                return false;
-            }
-            return true;
-        });
-
-        // Place the "None of the above" options last
-        // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
-        if (noneOfTheAbove) {
-            newChoices.push(noneOfTheAbove);
-        }
-
-        return newChoices;
-    };
-
-    const enforceOrdering = (
-        choices: ReadonlyArray<RadioChoiceWithMetadata>,
-    ) => {
-        // Represents choices that we automatically re-order if encountered.
-        // Note: these are in the reversed (incorrect) order that we will swap, if
-        // found.
-        // Note 2: these are internationalized when compared later on.
-        const ReversedChoices: ReadonlyArray<[string, string]> = [
-            [strings.false, strings.true],
-            [strings.no, strings.yes],
-        ];
-        const content = choices.map((c) => c.content);
-        if (ReversedChoices.some((reversed) => _.isEqual(content, reversed))) {
-            return [choices[1], choices[0]];
-        }
-        return choices;
-    };
-
     // Add meta-information to choices
     const choices: ReadonlyArray<RadioChoiceWithMetadata> =
         widgetOptions.choices.map((choice, i): RadioChoiceWithMetadata => {
@@ -78,12 +36,13 @@ const _choiceTransform = (
     // Apply all the transforms. Note that the order we call these is
     // important!
     // 3) finally add "None of the above" to the bottom
-    return _addNoneOfAbove(
+    return addNoneOfAbove(
         // 2) then (potentially) enforce ordering (eg. False, True becomes
         //    True, False)
         enforceOrdering(
             // 1) we randomize the order first
-            _maybeRandomize(choices),
+            maybeRandomize(choices, widgetOptions.randomize),
+            strings,
         ),
     );
 };
@@ -108,7 +67,7 @@ const transform = (
         multipleSelect,
         countChoices,
         deselectEnabled,
-        choices,
+        choices: [...choices],
         // doesn't seem used? choiceStates includes selected...
         selectedChoices: _.pluck(choices, "correct"),
     };
--- a/packages/perseus/src/widgets/radio/util.test.ts
+++ b/packages/perseus/src/widgets/radio/util.test.ts
@@ -1,6 +1,8 @@
 import {mockStrings} from "../../strings";
 
-import {getChoiceLetter} from "./util";
+import {addNoneOfAbove, enforceOrdering, getChoiceLetter} from "./util";
+
+import type {RadioChoiceWithMetadata} from "./radio-component";
 
 describe("getChoiceLetter (in English)", () => {
     it("returns the first 5 letters for most questions", () => {
@@ -16,5 +18,196 @@ describe("getChoiceLetter (in English)", () => {
         for (let i = 26; i < 100; i++) {
             expect(getChoiceLetter(i, mockStrings)).toEqual(" ");
         }
+    });
+});
+
+describe("addNoneOfAbove", () => {
+    function generateTestChoices(): RadioChoiceWithMetadata[] {
+        return [
+            {
+                content: "Option 1",
+                id: "option-1",
+                originalIndex: 0,
+            },
+            {
+                content: "Option 2",
+                id: "option-2",
+                originalIndex: 1,
+            },
+            {
+                content: "Option 3",
+                id: "option-3",
+                originalIndex: 2,
+            },
+        ];
+    }
+
+    it("moves isNoneOfTheAbove to the end", () => {
+        const input: RadioChoiceWithMetadata[] = generateTestChoices();
+        input[1].isNoneOfTheAbove = true;
+
+        // make sure we don't accidentally break the test
+        expect(input[2].isNoneOfTheAbove).toBeUndefined();
+
+        const rv = addNoneOfAbove(input);
+
+        expect(rv[2]).toEqual({
+            content: "Option 2",
+            id: "option-2",
+            originalIndex: 1,
+            isNoneOfTheAbove: true,
+        });
+    });
+
+    it("keeps isNoneOfTheAbove at the end", () => {
+        const input: RadioChoiceWithMetadata[] = generateTestChoices();
+        input[2].isNoneOfTheAbove = true;
+
+        // make sure we don't accidentally break the test
+        expect(input[2].isNoneOfTheAbove).not.toBeUndefined();
+
+        const rv = addNoneOfAbove(input);
+
+        expect(rv[2]).toEqual({
+            content: "Option 3",
+            id: "option-3",
+            originalIndex: 2,
+            isNoneOfTheAbove: true,
+        });
+    });
+
+    it("handles choices without isNoneOfTheAbove", () => {
+        const input: RadioChoiceWithMetadata[] = generateTestChoices();
+
+        // make sure we don't accidentally break the test
+        input.forEach((choice) => {
+            expect(choice.isNoneOfTheAbove).toBeUndefined();
+        });
+
+        const rv = addNoneOfAbove(input);
+
+        expect(rv).toEqual(input);
+    });
+});
+
+describe("enforceOrdering", () => {
+    it("maintains true/false order", () => {
+        const input: RadioChoiceWithMetadata[] = [
+            {
+                content: "True",
+                id: "option-1",
+                originalIndex: 0,
+            },
+            {
+                content: "False",
+                id: "option-2",
+                originalIndex: 1,
+            },
+        ];
+
+        // make sure we don't accidentally break the test
+        expect(input[0].content).toBe("True");
+        expect(input[1].content).toBe("False");
+
+        const rv = enforceOrdering(input, mockStrings);
+
+        expect(rv[0].content).toBe("True");
+        expect(rv[1].content).toBe("False");
+    });
+
+    it("fixes false/true order", () => {
+        const input: RadioChoiceWithMetadata[] = [
+            {
+                content: "False",
+                id: "option-1",
+                originalIndex: 0,
+            },
+            {
+                content: "True",
+                id: "option-2",
+                originalIndex: 1,
+            },
+        ];
+
+        // make sure we don't accidentally break the test
+        expect(input[0].content).toBe("False");
+        expect(input[1].content).toBe("True");
+
+        const rv = enforceOrdering(input, mockStrings);
+
+        expect(rv[0].content).toBe("True");
+        expect(rv[1].content).toBe("False");
+    });
+
+    it("maintains yes/no order", () => {
+        const input: RadioChoiceWithMetadata[] = [
+            {
+                content: "Yes",
+                id: "option-1",
+                originalIndex: 0,
+            },
+            {
+                content: "No",
+                id: "option-2",
+                originalIndex: 1,
+            },
+        ];
+
+        // make sure we don't accidentally break the test
+        expect(input[0].content).toBe("Yes");
+        expect(input[1].content).toBe("No");
+
+        const rv = enforceOrdering(input, mockStrings);
+
+        expect(rv[0].content).toBe("Yes");
+        expect(rv[1].content).toBe("No");
+    });
+
+    it("fixes no/yes order", () => {
+        const input: RadioChoiceWithMetadata[] = [
+            {
+                content: "No",
+                id: "option-1",
+                originalIndex: 0,
+            },
+            {
+                content: "Yes",
+                id: "option-2",
+                originalIndex: 1,
+            },
+        ];
+
+        // make sure we don't accidentally break the test
+        expect(input[0].content).toBe("No");
+        expect(input[1].content).toBe("Yes");
+
+        const rv = enforceOrdering(input, mockStrings);
+
+        expect(rv[0].content).toBe("Yes");
+        expect(rv[1].content).toBe("No");
+    });
+
+    it("doesn't affect unordered strings", () => {
+        const input: RadioChoiceWithMetadata[] = [
+            {
+                content: "Hello",
+                id: "option-1",
+                originalIndex: 0,
+            },
+            {
+                content: "World",
+                id: "option-2",
+                originalIndex: 1,
+            },
+        ];
+
+        // make sure we don't accidentally break the test
+        expect(input[0].content).toBe("Hello");
+        expect(input[1].content).toBe("World");
+
+        const rv = enforceOrdering(input, mockStrings);
+
+        expect(rv[0].content).toBe("Hello");
+        expect(rv[1].content).toBe("World");
     });
 });
--- a/packages/perseus/src/widgets/radio/util.ts
+++ b/packages/perseus/src/widgets/radio/util.ts
@@ -1,6 +1,11 @@
-import type {Props} from "./radio-component";
+import {
+    randomizeArrayElements,
+    type PerseusRadioUserInput,
+} from "@khanacademy/perseus-core";
+import _ from "underscore";
+
+import type {Props, RadioChoiceWithMetadata} from "./radio-component";
 import type {PerseusStrings} from "../../strings";
-import type {PerseusRadioUserInput} from "@khanacademy/perseus-core";
 
 /**
  * Given a choice's position in the radio widget, return the corresponding
@@ -57,4 +62,53 @@ export function getUserInputFromSerializedState(
     return {
         choicesSelected: props.choices.map(() => false),
     };
+}
+
+export function addNoneOfAbove(
+    choices: ReadonlyArray<RadioChoiceWithMetadata>,
+): ReadonlyArray<RadioChoiceWithMetadata> {
+    let noneOfTheAbove = null;
+
+    const newChoices = choices.filter((choice, index) => {
+        if (choice.isNoneOfTheAbove) {
+            // @ts-expect-error - TS2322 - Type 'RadioChoiceWithMetadata' is not assignable to type 'null'.
+            noneOfTheAbove = choice;
+            return false;
+        }
+        return true;
+    });
+
+    // Place the "None of the above" options last
+    // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
+    if (noneOfTheAbove) {
+        newChoices.push(noneOfTheAbove);
+    }
+
+    return newChoices;
+}
+
+export function enforceOrdering(
+    choices: ReadonlyArray<RadioChoiceWithMetadata>,
+    strings: PerseusStrings,
+): ReadonlyArray<RadioChoiceWithMetadata> {
+    // Represents choices that we automatically re-order if encountered.
+    // Note: these are in the reversed (incorrect) order that we will swap, if
+    // found.
+    // Note 2: these are internationalized when compared later on.
+    const ReversedChoices: ReadonlyArray<[string, string]> = [
+        [strings.false, strings.true],
+        [strings.no, strings.yes],
+    ];
+    const content = choices.map((c) => c.content);
+    if (ReversedChoices.some((reversed) => _.isEqual(content, reversed))) {
+        return [choices[1], choices[0]];
+    }
+    return choices;
+}
+
+export function maybeRandomize(
+    array: ReadonlyArray<RadioChoiceWithMetadata>,
+    randomize?: boolean,
+): ReadonlyArray<RadioChoiceWithMetadata> {
+    return randomize ? randomizeArrayElements(array) : array;
 }


packages/perseus/src/widgets/radio/multiple-choice-widget.new.tsx
```diff
@@ -45,13 +45,7 @@ export type RenderProps = {
     countChoices?: boolean;
     deselectEnabled?: boolean;
     choices: ReadonlyArray<RadioChoiceWithMetadata>;
-    // (LEMS-3278) - Remove all references to selectedChoices, as it's not used anywhere.
-    // We're handling selected in the choiceStates array.
-    selectedChoices: ReadonlyArray<PerseusRadioChoice["correct"]>;
     choiceStates?: ReadonlyArray<ChoiceState>;
-    // Deprecated; support for legacy way of handling changes
-    // Adds proptype for prop that is used but was lacking type
-    values?: ReadonlyArray<boolean>;
```
handeyeco, line 54: `selectedChoices` and `values` weren't being used (or passed through the transform so that data could never make it to the widget).

packages/perseus/src/widgets/group/group.test.tsx
```diff
@@ -330,46 +329,6 @@ describe("group widget", () => {
                     static: false,
                 },
             },
-            "radio 1": {
```
handeyeco, line 333: This widget wasn't in the `content`. No reason to add dead test data IMO.

packages/perseus/src/widgets/group/group.testdata.ts
```diff
@@ -152,43 +152,6 @@ export const question1: PerseusRenderer = {
             type: "group",
             version: {major: 0, minor: 0},
         },
-        "radio 1": {
```
handeyeco, line 155: This widget wasn't in the `content`. No reason to add dead test data IMO.

packages/perseus/src/__testdata__/renderer.testdata.ts
```diff
@@ -233,27 +233,25 @@ export const mockedShuffledRadioProps: {[key in string]: RenderProps} = {
                 correct: false,
                 originalIndex: 3,
             },
-            {
-                id: "1-1-1-1-1",
```
handeyeco, line 237: These changes are due to a change in how we shuffle.

packages/perseus/src/__tests__/renderer.test.tsx
```diff
@@ -75,21 +74,6 @@ describe("renderer", () => {
                 ok: true,
             }),
         ) as jest.Mock;
-
-        // Mock Math.random to return a deterministic sequence for consistent test results
-        mathRandomSpy = jest.spyOn(Math, "random");
-        let callCount = 0;
-        mathRandomSpy.mockImplementation(() => {
```
handeyeco, line 80: New shuffle doesn't use `Math.random` so there's no reason to mock it.

packages/perseus/src/__tests__/widgets.test.ts
```diff
@@ -25,20 +24,11 @@ describe("Widget API support", () => {
             expect(Widgets.getWidget("coolbeans")).toBe(null);
         });
 
-        it("returns a transform function when widgets provide one", () => {
```
handeyeco, line 28: Radio transform removed and this is the last widget to have a transform.

packages/perseus/src/widget-ai-utils/radio/radio-ai-utils.test.ts
```diff
@@ -74,23 +73,8 @@ describe("Radio AI utils", () => {
         jest.spyOn(Dependencies, "getDependencies").mockReturnValue(
             testDependencies,
         );
-
-        // Mock Math.random to return a deterministic sequence for consistent test results
-        mathRandomSpy = jest.spyOn(Math, "random");
```
handeyeco, line 79: New shuffle doesn't use Math.random so there's no reason to mock it.

packages/perseus/src/widget-ai-utils/radio/radio-ai-utils.test.ts
```diff
@@ -136,47 +120,61 @@ describe("Radio AI utils", () => {
...
             userInput: {
-                selectedOptions: [true, false, false, false],
+                // shuffled
+                selectedOptions: [false, false, true, false],
             },
         });
     });
 
-    // why are these tests named the same?
-    it("should get prompt json which matches the state of the UI", async () => {
```
handeyeco, line 145: I tried to make this test more robust because it was passing when it shouldn't have been passing due to random chance. Now the test runs against each choice in the widget options.

packages/perseus/src/widgets/group/group.test.tsx
```diff
@@ -330,50 +329,10 @@ describe("group widget", () => {
...
-                hasNoneOfTheAbove: false,
-                multipleSelect: false,
-                numCorrect: 1,
-                selectedChoices: [false, false, false, false, true],
-            },
         });
     });
 
-    it("should be able to restore serialized state", async () => {
+    it.skip("should be able to restore serialized state", async () => {
```
handeyeco, line 335: This PR breaks our serialized state APIs. An upcoming PR will remove them. This is with Jeremy and Jeanette's blessing and I'll be talking with the last consumer about this.

packages/perseus/src/widgets/radio/__tests__/radio.test.ts
```diff
@@ -940,4 +940,72 @@ describe("Radio Widget", () => {
             expect(correctChoiceContent).toContain("Choice B");
         });
     });
+
+    describe("shuffling", () => {
+        // regression LEMS-297
+        it("shuffles differently for multiple radios in the same exercise", () => {
```
handeyeco, line 946: I was told this was an important requirement of shuffling but there wasn't a test for it.

packages/perseus/src/widgets/radio/multiple-choice-widget.new.tsx
```diff
@@ -96,6 +88,9 @@ type DefaultProps = Required<
  * Created as part of the Radio Revitalization Project (LEMS-2933).
  */
 class MultipleChoiceWidget extends React.Component<Props> implements Widget {
+    static contextType = PerseusI18nContext;
+    declare context: React.ContextType<typeof PerseusI18nContext>;
```
handeyeco, line 92: Fixed the way we access strings

packages/perseus/src/widgets/radio/radio-component.tsx
```diff
@@ -17,23 +17,19 @@ import type {
     ShowSolutions,
     PerseusRadioRubric,
     PerseusRadioUserInput,
+    PerseusRadioWidgetOptions,
 } from "@khanacademy/perseus-core";
 
 // RenderProps is the return type for radio.jsx#transform
-export type RenderProps = {
+export type RenderProps = PerseusRadioWidgetOptions & {
```
handeyeco, line 24: In the future `RenderProps` should be removed in favor of `PerseusRadioWidgetOptions`.

packages/perseus/src/widgets/radio/radio-question-builder.test.ts
```diff
@@ -2,9 +2,6 @@ import {radioQuestionBuilder} from "./radio-question-builder";
 
 import type {PerseusRenderer} from "@khanacademy/perseus-core";
 
-// Mock crypto.randomUUID to return a predictable value for testing
-crypto.randomUUID = jest.fn(() => "0-0-0-0-0");
```
handeyeco, line 6: This was causing **A LOT** of noise in tests; IDs are used as React keys and React complains when multiple elements have the same key. So I fixed this in `radio-question-builder.ts`

packages/perseus/src/widgets/radio/radio.ff.tsx
```diff
@@ -77,7 +78,11 @@ class Radio extends RadioOld {
      * [LEMS-3185] do not trust serializedState/restoreSerializedState
      */
     getSerializedState() {
-        const {userInput: _, ...rest} = this._mergePropsAndState();
+        const {
+            userInput: _,
+            randomize: __,
```
handeyeco, line 83: Just to keep the serialization test happy

packages/perseus/src/widgets/radio/radio.ff.tsx
```diff
@@ -154,10 +166,20 @@ class Radio extends RadioOld {
...
          * legacy code will understand.
          */
+        const choices = [
+            ...choiceTransform(
+                this.props.choices,
+                this.props.randomize,
+                this.context.strings,
+                this.props.problemNum ?? 0,
+            ),
+        ];
```
handeyeco: This is the main change. The main thing `transform` was doing was transforming choices. The `transform` API is being removed, so I needed a new way to do things. This PR makes shuffling deterministic so it can be called over-and-over again with the same results. Now I run the choice transform as part of the render process so there's no need to have Renderer stash widget state.

packages/perseus/src/widgets/radio/radio.ff.tsx
```diff
@@ -178,16 +200,20 @@ class Radio extends RadioOld {
         return false;
     }
 
+    getPromptJSON(): RadioPromptJSON {
+        return this.radioRef.current!.getPromptJSON();
+    }
```
handeyeco, line 210: This was a source of pain, related to `class Radio extends RadioOld`. Component inheritance is considered an anti-pattern in React (see [here](https://legacy.reactjs.org/docs/composition-vs-inheritance.html#so-what-about-inheritance) and [here](https://khanacademy.slack.com/archives/C0918TZ5G/p1755271883587439)) because it introduces really weird and hard to debug issues!

The tl;dr: `getPromptJSON` in `radio-component.tsx` was getting a different set of props than `render` in `radio-component.tsx` because of some inheritance weirdness (the binding of the methods were binding to different parent components; React likely auto-binds lifecycle methods but not custom methods added to the component).

packages/perseus/src/widgets/radio/radio.ts
```diff
@@ -1,119 +1,16 @@
...
-import type {PerseusStrings} from "../../strings";
 import type {WidgetExports} from "../../types";
 import type {
     PerseusRadioUserInput,
     PerseusRadioWidgetOptions,
     RadioPublicWidgetOptions,
 } from "@khanacademy/perseus-core";
 
-// Transforms the choices for display.
-const _choiceTransform = (
```
handeyeco, line 17: I broke these functions apart, moved them to `util`, and added a test suite around them.

packages/perseus/src/widgets/radio/serialize-radio.test.ts
```diff
@@ -23,10 +23,10 @@ const expectedSerializedRadio = {
     deselectEnabled: false,
     choices: [
         {
-            id: "5-5-5-5-5",
```
handeyeco: These changes are due to a change in how we shuffle.

packages/perseus/src/widgets/radio/serialize-radio.test.ts
```diff
@@ -156,20 +153,6 @@ describe("Radio serialization", () => {
         jest.spyOn(Dependencies, "getDependencies").mockReturnValue(
             testDependencies,
         );
-
-        // Mock Math.random to return a deterministic sequence for consistent test results
-        mathRandomSpy = jest.spyOn(Math, "random");
```
handeyeco, line 161: Don't need to mock `Math.random` anymore because we're not using it anymore.

packages/perseus/src/widgets/radio/util.test.ts
```diff
@@ -18,3 +32,480 @@ describe("getChoiceLetter (in English)", () => {
         }
     });
 });
+
+describe("addNoneOfAbove", () => {
```
handeyeco: New tests!

packages/perseus/src/widgets/radio/util.ts
```diff
@@ -20,16 +27,22 @@ export function getChoiceLetter(pos: number, strings: PerseusStrings): string {
     return " ";
 }
 
+/**
+ * @deprecated and likely a very broken API
+ * [LEMS-3185] do not trust serializedState/restoreSerializedState
+ */
 export function getUserInputFromSerializedState(
-    props: Props,
+    serializedState: any,
```
handeyeco, line 35: Serialized state isn't technically props: it's the history of every type of RenderProps we've ever had. That's why it's so dangerous.

packages/perseus/src/widgets/radio/utils/general-utils.test.ts
```diff
@@ -90,26 +90,7 @@ describe("getChoiceStates", () => {
         });
     });
 
-    describe("Case 3: Legacy user selection without submission", () => {
```
handeyeco, line 93: `values` is no longer a thing.

packages/perseus/src/__tests__/renderer.test.tsx
```diff
@@ -340,7 +324,7 @@ describe("renderer", () => {
             );
         });
 
-        it("should call the onWidgetStartProps callback if provided in apiOptions", () => {
+        it.skip("should call the onWidgetStartProps callback if provided in apiOptions", () => {
```
handeyeco: This API is set for removal: https://khanacademy.slack.com/archives/C06NBBNCV8B/p1755185651787759

packages/perseus/src/widgets/radio/radio.ff.tsx
```diff
@@ -130,8 +135,15 @@ class Radio extends RadioOld {
                     };
                     // Use getUserInputFromSerializedState to get
                     // unshuffled user input so that we can score with it
-                    const unshuffledUserInput =
-                        getUserInputFromSerializedState(mergedProps);
```
handeyeco, line 134: I didn't like that I had used `getUserInputFromSerializedState` for unshuffling, so I made a specific helper for it with tests.

packages/perseus/src/widgets/radio/util.ts
```diff
@@ -39,22 +52,148 @@ export function getUserInputFromSerializedState(
...
+    return choices;
+}
+
+export function maybeRandomize(
+    array: ReadonlyArray<RadioChoiceWithMetadata>,
+    seed: number,
+    randomize?: boolean,
+): ReadonlyArray<RadioChoiceWithMetadata> {
+    return randomize ? shuffle(array, seed) : array;
+}
```
benchristel: I don't think this `maybeRandomize` function is pulling its weight. We pass it a boolean just so it can use a ternary on it? I'd inline it.

packages/perseus/src/widgets/radio/util.ts
```diff
@@ -39,22 +52,148 @@ export function getUserInputFromSerializedState(
...
+ * generate hash from string (for generating a seed)
+ */
+function generateHash(string: string): number {
+    let hash = 0;
+    for (const char of string) {
+        hash = (hash << 5) - hash + char.charCodeAt(0);
+        hash |= 0; // Constrain to 32bit integer
+    }
+    return hash;
+}
```
benchristel: This is the third djb2 variant I've seen! Where did you get it?

I'd really like to avoid implementing our own hash algorithm. They're hard to get right and tricky to test. See the discussion on [ADR 851].

In particular, this djb2 variant starts the hash at 0, which means any sequence of null/0 bytes will have the same hash, no matter how many 0s there are. I don't think that's _particularly_ relevant for this case where we're hashing text, but the problem is, if someone else needs a hash algorithm in the future, they might copy-paste this one, and it might be a pretty bad fit for their needs.

Another difference from the [original djb2] is that we're multiplying by 31 instead of 33. I don't know what difference that makes! It seems like it shouldn't make much of a difference, but why take the risk?

If we really need to hash here, I think this might need a quick ADR explaining why. We might want to consider (again) adding `imurmurhash` as a dependency, to avoid implementing our own algorithm.

As an alternative to hashing, we could get a different sequence of random numbers in each widget by seeding the generator with the same seed but throwing away some of the generated numbers before shuffling.

E.g. in the first radio widget, we'd just use the RNG as-is. In the second widget, we'd skip the first 10 generated numbers before shuffling. In the third, we'd skip the first 20...

[ADR 851]: https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/4199907561/ABANDONED+ADR+851+New+Dependency+Proposal+-+Add+imurmurhash+to+Perseus
[original djb2]: https://stackoverflow.com/questions/1579721/why-are-5381-and-33-so-important-in-the-djb2-algorithm

packages/perseus/src/widgets/radio/util.ts
```diff
@@ -39,22 +52,148 @@ export function getUserInputFromSerializedState(
...
-        for (let i = 0; i < valuesLength; i++) {
-            const index = unshuffle ? props.choices[i].originalIndex : i;
-            choicesSelected[index] = values[i];
+    // Nothing checked
+    return {
+        choicesSelected: serializedState.choices.map(() => false),
+    };
+}
+
+export function addNoneOfAbove(
```
benchristel: I know you didn't name this function, but I think a name like `moveNoneOfTheAboveToEnd` would be clearer. This function doesn't add a "none of the above" option, it just reorders existing options.

packages/perseus/src/widgets/radio/serialize-radio.test.ts
```diff
@@ -225,7 +208,7 @@ describe("Radio serialization", () => {
                 // note we unshuffle!
                 // in expectedSerializedRadio.choiceStates the first element
                 // is selected; here the last element is selected
-                choicesSelected: [false, false, false, true],
+                choicesSelected: [false, false, true, false],
```
benchristel: The comment `here the last element is selected` is wrong now. Could you update it?

packages/perseus/src/__tests__/renderer.test.tsx
benchristel: Could we just delete the test? Or add a TODO comment? I worry that this skipped test might stick around for a long time.

packages/perseus/src/__tests__/widgets.test.ts
```diff
@@ -25,20 +24,11 @@ describe("Widget API support", () => {
...
-                widgetOptions,
-            );
-        });
-
         it("returns an identity function when widgets don't provide a transform", () => {
             const widgetOptions = {cool: "beans"};
             // Group does not provide a `transform` function
             const transform = Widgets.getTransform("group");
-            expect(transform?.(widgetOptions, mockStrings)).toEqual(
+            expect(transform?.(widgetOptions, mockStrings, 0)).toEqual(
```
benchristel, line 31: Could we remove `Widgets.getTransform` now? It looks like it's only used in tests.

packages/perseus/src/types.ts
```diff
@@ -470,7 +470,7 @@ type WidgetOptions = any;
 export type WidgetTransform = (
     widgetOptions: WidgetOptions,
     strings: PerseusStrings,
-    problemNumber?: number,
+    problemNumber: number,
```
benchristel, line 463: The whole `WidgetTransform` type can go away now, right?

packages/perseus/src/widget-ai-utils/radio/radio-ai-utils.test.ts
```diff
@@ -136,47 +120,61 @@ describe("Radio AI utils", () => {
...
+                const nodeText = listItem.textContent;
+
+                if (!nodeText) {
+                    throw new Error("Not able to get text from UI element");
+                }
+
+                // make sure the text from the prompt is in
+                // the text for the UI at a given index
+                expect(nodeText.includes(promptJSONItemText)).toBe(true);
+            });
```
benchristel: I think this is more or less equivalent to:

```suggestion
expect(listItems.map((li) => li.textContent)).toEqual(
    widgetJSON.options.map((opt) => 
        expect.stringContaining(opt.value),
    ),
);
```

packages/perseus/src/widgets/radio/__tests__/radio.test.ts
```diff
@@ -940,4 +940,72 @@ describe("Radio Widget", () => {
...
+            lis1.forEach((li1, idx) => {
+                const li2 = lis2[idx];
+                const li1Text = li1.textContent;
+                const li2Text = li2.textContent;
+                if (li1Text !== li2Text) {
+                    sameOrder = false;
+                }
+            });
+
+            expect(sameOrder).toBe(false);
```
benchristel: I think this could be

```suggestion
            const getText = (element) => element.textContent;
            expect(lis1.map(getText)).not.toEqual(lis2.map(getText));
```

packages/perseus/src/widgets/radio/__tests__/radio.test.ts
```diff
@@ -940,4 +940,72 @@ describe("Radio Widget", () => {
             expect(correctChoiceContent).toContain("Choice B");
         });
     });
+
+    describe("shuffling", () => {
+        // regression LEMS-297
+        it("shuffles differently for multiple radios in the same exercise", () => {
+            let counter: number = 0;
+            const generateId = () =>
+                `${counter++}-${counter++}-${counter++}-${counter++}-${counter++}`;
```
benchristel: This seems overcomplicated. Why not just

```suggestion
            const generateId = () => `${counter++}`;
```

packages/perseus/src/widgets/radio/multiple-choice-widget.new.tsx
benchristel: Is this React magic? If you set `contextType` the context automatically gets set as `component.context`?

packages/perseus/src/widgets/radio/util.ts
```diff
@@ -39,22 +52,148 @@ export function getUserInputFromSerializedState(
...
+            strings,
+        ),
+    );
+}
+
+/**
+ * converts an unshuffled user input to a shuffled user input
+ * based on the original index of the choices
+ *
+ * not used for shuffling, used for syncing shuffled/unshuffled
```
benchristel: I didn't understand this comment on first read. `shuffleUserInput` isn't used for shuffling? I might name this function differently. Something like `putUserInputInShuffledChoiceOrder`.

Instead of saying that it's `used for syncing shuffled/unshuffled`, I might say it's `used for syncing user input with the shuffled choice order`.

The need for this function is a pretty good argument for why we need an ID-based user input format.

packages/perseus/src/widgets/radio/util.ts
```diff
@@ -39,22 +52,148 @@ export function getUserInputFromSerializedState(
...
+        });
+
+    /**
+     * problemNum helps us to have a deterministic yet random order per-exercise
+     * but we still want to have a random order per widget in an exercise;
+     * so we use a combination of problemNum (random per learner) and
+     * a hash derived from IDs (random per widget)
+     */
+    const randomSeed =
+        generateHash(choices.map((c) => c.id).join()) + problemNum;
```
benchristel: I would prefer to do

```
generateHash(`${widgetId}:${problemNum}`);
```

I don't think it's a guarantee that choice IDs are different between widgets. The widgetId is guaranteed to be unique within the Renderer, though.

Adding the `problemNum` after hashing is problematic, too. The `problemNum` might be small (e.g. in the case where it is actually the _problem number_) and so not affect many bits. That's why I'd prefer to concatenate the `problemNum` onto the string being hashed.

packages/perseus/src/widgets/radio/util.ts
SonicScrewdriver: I fully support renaming any of these silly functions so that they better describe what they're doing.

packages/perseus/src/widgets/radio/multiple-choice-widget.new.tsx
handeyeco: Yeah, seems to just be React magic based on existing code.

packages/perseus/src/widgets/radio/util.ts
handeyeco: Rather than hashing, I figured out how to pass widget index down so now the seed can be `problemNum + widgetIndex` and we don't need to hash.

packages/perseus/src/__tests__/widgets.test.ts
handeyeco: I would prefer to remove the API in its own PR.

packages/perseus/src/types.ts
handeyeco: I would prefer to remove the API in its own PR.

packages/perseus/src/widgets/radio/util.ts
handeyeco: Aged out: no longer doing a hash

packages/perseus/src/widgets/radio/radio.ff.tsx
mark-fitzgerald: Thanks for accommodating this weirdness. This file (`radio.ff.tsx`) is temporary. Once we fully switch to the new widget code, this will be removed in favor of `multiple-choice-widget.new.tsx`

packages/perseus/src/renderer.tsx
```diff
@@ -546,6 +546,24 @@ class Renderer
         );
     };
 
+    _getWidgetIndexById(id: string) {
+        const widgetIndex = Object.keys(this.props.widgets).indexOf(id);
```
benchristel, line 547: Depending on key order makes me kind of nervous, but I think it's fine for this use case. Object properties are iterated (and therefore returned by `Object.keys`) in the order in which they were added to the object.

It might be worth adding a comment here to call this behavior out and emphasize that the index of a widget, according to this function, has nothing to do with the order in which it appears in the content (e.g. the order we use for `UserInputArray`)

packages/perseus/src/widgets/radio/__tests__/radio.test.ts
```diff
@@ -991,21 +990,12 @@ describe("Radio Widget", () => {
...
-            let sameOrder = true;
-            lis1.forEach((li1, idx) => {
-                const li2 = lis2[idx];
-                const li1Text = li1.textContent;
-                const li2Text = li2.textContent;
-                if (li1Text !== li2Text) {
-                    sameOrder = false;
-                }
-            });
+            const lis2 = Array.from(radios[1].querySelectorAll("li"));
```
benchristel, line 995: I don't see `Array.from` used much! I think a `[...]` spread would be more typical (and shorter):

```suggestion
            const lis2 = [...radios[1].querySelectorAll("li")];
```

packages/perseus/src/widgets/radio/__tests__/radio.test.ts
anakaren-rojas: were you able to find a solution for the regression?

anakaren-rojas:
This also seems to resolve https://khanacademy.atlassian.net/browse/LEMS-3192 🥳

handeyeco:
Closing in favor of: https://github.com/Khan/perseus/pull/2857

PR #2801, opened 2025-08-13
Allow autoplay for videos rendered in the video widget

MikeKlemarewski:
## Summary:
Our embedded videos are not working in KAC. Clicking play doesn't actually play the video. Enabling autoplay seems to fix this. Supposedly we can do this by adding the correct `Permissions-policy` header. I played around with that, but didn't have any luck. I think adding the attribute directly on the iframe for our video widget is a safe alternative.

### Demo:
Here's a video of the frontend using this version of Perseus from this PR.

https://github.com/user-attachments/assets/350dc1a7-fc82-45e6-b53a-cff8e5d09d61

## Test plan:
- Ensure tests pass
- In the frontend repo, point the perseus version to this PR and ensure videos in articles play

### DIFF
--- a/packages/perseus/src/widgets/video/video.test.ts
+++ b/packages/perseus/src/widgets/video/video.test.ts
@@ -60,11 +60,10 @@ describe("video widget", () => {
 
         // Assert
         // eslint-disable-next-line testing-library/no-node-access
-        expect(
-            document
-                .getElementsByTagName("iframe")[0]
-                .attributes.getNamedItem("allow")?.value,
-        ).toContain("autoplay");
+        expect(document.getElementsByTagName("iframe")[0]).toHaveAttribute(
+            "allow",
+            "autoplay",
+        );
     });
 
     it("vimeo widget should contain dnt param", () => {


packages/perseus/src/widgets/video/video.test.ts
```diff
@@ -49,6 +49,24 @@ describe("video widget", () => {
...
+        // Act
+        renderQuestion(question1, apiOptions);
+
+        // Assert
+        // eslint-disable-next-line testing-library/no-node-access
+        expect(
+            document
+                .getElementsByTagName("iframe")[0]
+                .attributes.getNamedItem("allow")?.value,
+        ).toContain("autoplay");
```
jeremywiebe: I agree we need to resort to direct-DOM APIs to find the iframe, but can we use `toHaveAttribute()` instead to do the assertion? 
```suggestion
        expect(
            document.getElementsByTagName("iframe")[0]
        ).toHaveAttribute("allow", "autoplay");
```

jeremywiebe (approved):

packages/perseus/src/widgets/video/video.test.ts
MikeKlemarewski: Good call!

PR #2803, opened 2025-08-13
[Radio] Enhance content for screen reader access

mark-fitzgerald:
## Summary:
Screen readers can access all the information in the radio widget, but some enhancements are needed to ensure clarity of what is being read. Specifically:
- The word "Choice" is prefixed to the choice letter
- The full choice content is connected to the choice button in order to present a complete readout of the choice
- The rationale is referenced via `aria-describedby` as additional material, leaving it to the screen reader to determine how to surface that information for the choice

Additionally, the choice button is marked with `aria-disabled` when in review mode to indicate that the button is no longer able to be activated.

Issue: LEMS-3364

## Test plan:
1. Launch Storybook locally
2. Navigate to Widgets => RadioNew => [Docs](http://localhost:6006/?path=/docs/widgets-radionew--docs)
3. Using a screen reader, traverse any of the examples using a keyboard (i.e. tab navigation)
   - Note how the screen reader announces the information for the widget as a whole, and for each choice within it
4. Using a screen reader, traverse any of the examples using the screen reader's key commands
   - Note how the screen reader announces each part of widget (more options than just using the tab key)
5. The announcements from the screen reader should make sense and reflect what is shown visually in the widget

### DIFF
--- a/packages/perseus/src/widgets/radio/__tests__/choice-indicator.test.tsx
+++ b/packages/perseus/src/widgets/radio/__tests__/choice-indicator.test.tsx
@@ -5,6 +5,13 @@ import * as React from "react";
 
 import Indicator from "../choice-indicator.new";
 
+import type {IndicatorContent} from "../choice-indicator.new";
+
+const indicatorContent: IndicatorContent = {
+    visible: "A",
+    screenReader: "Choice A",
+};
+
 describe("Multiple choice indicator", () => {
     let iconMock: jest.SpyInstance;
     let mockClickHandler: jest.Mock;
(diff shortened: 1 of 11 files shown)


packages/perseus/src/widgets/radio/__tests__/choice-indicator.test.tsx
```diff
@@ -106,6 +118,68 @@ describe("Multiple choice indicator", () => {
...
+                    checked={true}
+                    shape="circle"
+                    content={contentWithLabel}
+                    showCorrectness="wrong"
+                    updateChecked={mockClickHandler}
+                />,
+            );
+            const indicator = screen.getByRole("button");
+            expect(indicator).toHaveAttribute("aria-label", "Choice A");
+            expect(indicator).toHaveAttribute(
```
anakaren-rojas, line 156: These two can be on the element, but the screen reader will ignore the aria label, right?

mark-fitzgerald: Not exactly. `aria-labelledby` takes priority over `aria-label`. However, since the indicator references itself in the `aria-labelledby`, the readout will include the text in the `aria-label` attribute.

ivyolamit (approved):
@mark-fitzgerald thanks for looking into this, changes looks logical to me.

PR #2807, opened 2025-08-15
fix(LEMS-3400): Set default choice ids in Radio index.ts

anakaren-rojas:
## Summary:
- Sets default values for content and choice ids in Radio's index.ts file
- Updates radio test data to include choice ids

Issue: LEMS-3400

## Test plan:
Issue doesn't seem to exist in prod story book, only in local storybook 
- Validated issue is no longer present in local story book 

### Prod story book

https://github.com/user-attachments/assets/a5a33748-2616-4e77-8028-b6988c325301

### Main 

https://github.com/user-attachments/assets/88d3e642-77d6-43e0-88ca-30a5a99727aa

### Fix

https://github.com/user-attachments/assets/9f9fcb06-365a-4cbf-aae3-3ba7ab860cff

### DIFF
--- a/packages/perseus/src/widgets/radio/__tests__/radio.testdata.ts
+++ b/packages/perseus/src/widgets/radio/__tests__/radio.testdata.ts
@@ -124,7 +124,7 @@ export const choicesWithImages: PerseusRenderer = {
             options: {
                 choices: [
                     {
-                        id: "0-0-0-0-0",
+                        id: "4-4-4-4-4",
                         content:
                             "Same \nLine\n![2 micron diameter cell](https://ka-perseus-images.s3.amazonaws.com/b17cfb6a3270c6f41f66099462e495c841cf6ca9.png)\nSame\nLine",
                         correct: false,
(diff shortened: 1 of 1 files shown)


packages/perseus/src/widgets/radio/__tests__/radio.testdata.ts
nishasy, line 1: Can you make the IDs unique in this testdata file instead of having them restart from 0 for each question data? Sometimes we render multiple of these on the same page like in storybook, and that can cause issues, especially now since I'm using the choice ID as React keys.

anakaren-rojas: done!

nishasy (approved):
Thanks for making that change. Looks good to me :)

packages/perseus/src/widgets/radio/__tests__/radio.testdata.ts
```diff
@@ -53,24 +53,28 @@ export const questionWithPassage: PerseusRenderer = {
             options: {
                 choices: [
                     {
+                        id: "0-0-0-0-0",
```
SonicScrewdriver, line 56: It might be nice to make a helper function that can generate these ids for us, but that doesn't need to be part of this PR. :) 

I could see us adding a new set of testdata and missing that these are numbered uniquely, etc.

nishasy:
To @SonicScrewdriver's point:

I think this isn't using the radio builder because that's in a different repo, but I think it makes sense to copy that builder over to `perseus-editor` and use in these testdata files.

SonicScrewdriver:
Ohhhhh my bad, I missed realizing that the question builder would be in that core repo. That's a good point, I wonder if we should have a core place for all the question builders to live. 🤔 (Definitely a separate ticket)

PR #2809, opened 2025-08-15
[Image] (DX) Add an image generator + use it in testdata

nishasy:
## Summary:
To make it easier to write tests for the Image widget, we can use an
Image widget question builder, similar to the ones we already have for
Interactive Graph and Radio.

- Add a new question generator in perseus-core
- Add new Image question generator
- Add tests for Image question generator
- Use builder in image.testdata.ts

Issue: https://khanacademy.atlassian.net/browse/LEMS-3380

### Before/After check that nothing changed

https://github.com/user-attachments/assets/aa47c8f2-7756-475f-ac01-662460405b3a

## Test plan:
`pnpm jest packages/perseus/src/widgets/image/image-question-builder.test.ts`

Storybook
- Go to http://localhost:6006/?path=/docs/widgets-image--docs
- Also open the existing docs in a new tab - https://khan.github.io/perseus/?path=/docs/widgets-image--docs#question-2
- Compare the two and confirm that they're identical


nishasy:
Converting to draft until I can figure out where the builder should go.

packages/perseus-core/src/utils/generators/perseus-question-generator.ts
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    let content = "";
+    const widgetTypes = Object.keys(widgets);
+    for (let i = 0; i < widgetTypes.length; i++) {
+        const widgetType = widgetTypes[i];
+        content += `[[☃ ${widgetType}]]\n`;
+    }
+    return content;
+}
```
nishasy: I don't really like this `generateDefaultContent` function, especially since you can't be completely certain what the number is going to be when you have to manually put like `"image 1": generateImageWidget(...)` later.

Should I just remove this and make people manually write their snowmen in the content? Is there a different approach we'd prefer?

I added this because I wrote a bunch of questions without populating the `content` manually and then got confused when nothing showed up.

nishasy: I mean. I guess it's in order, so it's not _that_ bad, but it's not intuitive or accurate to how the real content works.

packages/perseus-core/src/utils/generators/image-widget-generator.ts
```diff
@@ -0,0 +1,28 @@
...
+    return {
+        type: "image",
+        graded: true,
+        version: {major: 0, minor: 0},
+        static: false,
+        alignment: "default",
+        options: generateImageOptions({}), // default options
+        ...imageWidgetProperties,
+    };
+}
```
handeyeco: In the future, it would be neat if we could figure out how to make our `WidgetOptions` type a little smarter. That way we wouldn't need `ImageWidget`, `OrdererWidget`, etc; they would all be `WidgetOptions` and TS would just understand that a `WidgetOptions` with `widgetOptions.type === "image"` has `PerseusLabelImageWidgetOptions` for its options.

Then instead of `generateImageWidget` we could have a reusable `generateWidgetOptions("image", generateImageOptions())`. Future goals though.

packages/perseus-core/src/utils/generators/perseus-question-generator.test.ts
```diff
@@ -0,0 +1,62 @@
+import {
+    generateImageOptions,
+    generateImageWidget,
+} from "./image-widget-generator";
+import {generateQuestion} from "./perseus-question-generator";
+
+import type {PerseusRenderer} from "../../data-schema";
+
+describe("generateQuestion", () => {
```
handeyeco: We have `generateTestPerseusRenderer` and `generateTestPerseusItem` already. I'm not saying you have to use those, but I imagine it'd be better if we just had one helper for generating a PerseusRender and one for a PerseusItem.

I'm open to talking about this if you want, I don't think the helpers I made are perfect and I'd be open to a more holistic approach.

packages/perseus-core/src/utils/generators/perseus-question-generator.ts
```diff
@@ -0,0 +1,31 @@
...
+    if (!widgets) {
+        return "";
+    }
+
+    let content = "";
+    const widgetTypes = Object.keys(widgets);
+    for (let i = 0; i < widgetTypes.length; i++) {
+        const widgetType = widgetTypes[i];
+        content += `[[☃ ${widgetType}]]\n`;
+    }
```
handeyeco: An alternative that will get you pretty close:
```suggestion
    let content = Object.keys(widgets)
        .map(widgetId => `[[☃ ${widgetId}]]`)
        .join("\n");
```

packages/perseus-core/src/utils/generators/perseus-question-generator.ts
handeyeco: > Should I just remove this and make people manually write their snowmen in the content? Is there a different approach we'd prefer?

I don't really mind manually writing content/snowmen. 🤷 I'm a little worried that `generateQuestion` is trying to do too much and that we'll get weird conflicts between passed `content` and generated `content`.

Maybe it could just be a separate helper?

```TS
function addWidget(renderer, widgetType, widgetOptions) {
    const rendererClone = deepClone(renderer);
    const optionsClone = deepClone(widgetOptions);

    const widgetId = uuid();
    rendererClone.widgets[widgetId] = generateWidgetOptions(widgetType, optionsClone);
    rendererClone.content += `\n[[☃ ${widgetId}]]`;

    return renderer;
}

const imageOptions = generateImageOptions({
    alt: "Hello alt",
});

let renderer = generateRenderer({
    content: "Hello world",
});

renderer = addWidget(renderer, "image", imageOptions);

console.log(renderer);

/*
{
    content: "Hello world\n[[☃ 3hj3-73hj]]",
    widgets: {
        "3hj3-73hj": {
            type: "image",
            options: {
                alt: "Hello alt",
                ...etc...
            },
            ...etc...
        },
    },
    images: {},
}
*/
```

But I've been handwriting content for awhile now and it hasn't been a big deal.

packages/perseus-core/src/utils/generators/image-question-generator.test.ts
```diff
@@ -0,0 +1,95 @@
+import {
+    generateImageOptions,
+    generateImageWidget,
+} from "./image-widget-generator";
+
+import type {ImageWidget, PerseusImageWidgetOptions} from "../../data-schema";
+
+describe("generateImageOptions", () => {
+    test("builds a default image options", () => {
+        const options: PerseusImageWidgetOptions = generateImageOptions({});
```
handeyeco: You could make options optional and then you'd be able to do `generateImageOptions()`; just a thought

packages/perseus-core/src/utils/generators/perseus-question-generator.test.ts
```diff
@@ -0,0 +1,62 @@
...
+                }),
+                "image 2": generateImageWidget({
+                    options: generateImageOptions({
+                        alt: "image 2",
+                    }),
+                }),
+            },
+        });
+
+        expect(question.content).toBe("[[☃ image 1]]\n[[☃ image 2]]\n");
```
handeyeco: In my other comment I mentioned worrying about `generateQuestion` doing too much with `content` generation, but this is pretty cool to see in practice so IDK.

packages/perseus-core/src/utils/generators/image-widget-generator.ts
nishasy: I'm running into some issues trying to make this generic, so I might try to come back to that another time.

nishasy:
@handeyeco I removed `generateQuestion` in favor of `generateTestPerseusRenderer` since they were doing the same thing, like you said. As a side effect, I also removed the auto content generator. I suppose people can figure out their own snowmen.

What do think of this approach instead?

packages/perseus-core/src/utils/generators/image-widget-generator.ts
```diff
@@ -0,0 +1,28 @@
...
+export function generateImageWidget(
+    imageWidgetProperties?: Partial<Omit<ImageWidget, "type">>,
+): ImageWidget {
+    return {
+        type: "image",
+        graded: true,
+        version: {major: 0, minor: 0},
+        static: false,
+        alignment: "default",
+        options: generateImageOptions({}), // default options
```
handeyeco, line 25: 🎉
```suggestion
        options: generateImageOptions(), // default options
```

PR #2812, opened 2025-08-19
docs(changeset): New Storybook Docs for Radio Widget (New)

SonicScrewdriver:
## Summary:
This PR reorganizes our Storybook stories for the Radio (New) Widget, with the following objectives:

1. Creation of new simple Regression Test stories to be used with Chromatic.
2. New Doc page that outlines all widget options (and key subtypes) along with the UserInput format. 
3. Turn on story subpages for the Widget Demos, in order to improve the ability of devs to test/debug the widget. 
4. Add missing options to the RadioQuestionBuilder (numCorrect and static) 

Check out the new docs here, in [this PR snapshot for Storybook](https://650db21c3f5d1b2f13c02952-bjivnxuqih.chromatic.com/?path=/docs/widgets-radionew--docs).

We discussed the Component/States gallery, but there's some issues with those stories right now so it seems best to wait until they are resolved before we turn on Chromatic for them. Rather than duplicating the component states pages, we'll just set Chromatic up to use them directly. 

Issue: LEMS-3352

## Test plan:
- tests pass 
- new chromatic tests!

### DIFF
--- a/packages/perseus-editor/src/widgets/__docs__/radio-editor.stories.tsx
+++ b/packages/perseus-editor/src/widgets/__docs__/radio-editor.stories.tsx
@@ -19,6 +19,7 @@ registerAllWidgetsAndEditorsForTesting();
 const meta: Meta = {
     title: "Widgets/RadioNew/Editor Demo",
     component: RadioEditor,
+    tags: ["!autodocs"],
 } satisfies Meta<typeof RadioEditor>;
 export default meta;
 


packages/perseus/src/widgets/radio/__docs__/multiple-choice-demo.new.stories.tsx
```diff
@@ -36,9 +36,9 @@ type StoryArgs = {
  */
 
 export default {
```
SonicScrewdriver, line 39: The stories in here will now be under "Widget Demo" (I'm open for pitches for better names).

I've turned on the sub-pages for this file (each story will have a page) so that we can more easily test individual states again. This seemed like a decently clean solution that allows us to have a clean sidebar and easily testable stories.

packages/perseus/src/widgets/radio/__docs__/multiple-choice.new.doc.stories.tsx
```diff
@@ -0,0 +1,227 @@
+import * as React from "react";
```
SonicScrewdriver, line 1: This is the new 'Doc' entry for the Radio (New) widget in the Storybook sidebar. The primary objective was to create clear, comprehensive documentation for the Widget Options, Choice properties, and User Input structures.

I've added a brief introduction that highlights the widget's significance and capabilities, though we may want to enhance this content further.

Ideally, this documentation would be generated automatically to prevent drift as the widget evolves. However, there are technical challenges: directly rendering the widget requires complex type parameters, while using Renderer or ServerItemRenderer introduces excessive props that cloud the documentation. It just seems to be another case of Perseus Khanplexity. 

I've tried to aim for a more balanced solution - simple, focused stories with explicit documentation of the widget's data structures. This provides clear visibility into the API while keeping the examples straightforward and maintainable.

However, I'm also very happy to pivot or adjust if we feel there's value in a different approach.

packages/perseus/src/widgets/radio/__docs__/multiple-choice-regression.stories.tsx
```diff
@@ -0,0 +1,364 @@
...
+ */
+
+/* States */
+
+// Basic Single Select
+export const SingleSelect: Story = {
+    args: {
+        item: generateTestPerseusItem({
+            question: radioQuestionBuilder()
+                .addChoice("Choice 1", {correct: true})
```
ivyolamit, line 40: I noticed the indicator is currently it's not representing the correct visuals, Is there anything we need to add here to surface the letter and check icon like below?

image

Actual:
image

packages/perseus/src/widgets/radio/__docs__/multiple-choice-regression.stories.tsx
```diff
@@ -0,0 +1,364 @@
...
+                component:
+                    "Regression tests for the radio widget, which will be used with Chromatic.",
+            },
+        },
+        chromatic: {disableSnapshot: false},
+    },
+};
+
+/**
+ * Single Selection Mode
```
ivyolamit: Do we need this?
image

packages/perseus/src/widgets/radio/__docs__/multiple-choice-regression.stories.tsx
```diff
@@ -0,0 +1,364 @@
...
+                .addChoice(
+                    "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Mauris fringilla lorem metus, eleifend eleifend ante sodales non. Sed nec tempus lacus. Suspendisse finibus sit amet enim quis facilisis. Sed eu nisl in ligula mollis dapibus. Maecenas faucibus interdum est, nec scelerisque ligula. Nulla vestibulum tristique vulputate. Etiam hendrerit tellus sit amet lacus suscipit, in auctor nisi hendrerit. Mauris mi turpis, scelerisque eget diam a, rhoncus sodales dui. Fusce sit amet magna nisl. Donec mi nisi, accumsan nec libero consequat, ultrices ornare mauris. Sed leo dolor, facilisis sed urna eu, feugiat dapibus metus. Nunc accumsan lacus ac massa molestie, placerat malesuada massa placerat. Proin ac dolor sit amet metus semper cursus. Suspendisse nec ipsum non nunc commodo posuere. Duis quis porttitor quam. Nulla facilisi.",
+                )
+                .build(),
+        }),
+    },
+};
+
+/**
+ * Multi Selection Mode
```
ivyolamit: Do we need this?

image

packages/perseus/src/widgets/radio/__docs__/multiple-choice.new.doc.stories.tsx
```diff
@@ -0,0 +1,226 @@
...
+        question: singleSelectQuestion,
+    },
+};
+
+export const MultipleSelect: Story = {
+    args: {
+        question: multipleSelectQuestion,
+    },
+};
+
```
ivyolamit, line 201: Please also add the RTL scenario here, so we can visually spot check it.

packages/perseus/src/widgets/radio/__stories__/choice-indicator.stories.tsx
```diff
@@ -27,7 +27,7 @@ const Container = (props: {children: React.ReactNode}): React.ReactElement => {
 };
 
 export default {
-    title: "Widgets/Radio/Widget Internal Components/Indicator",
+    title: "Widgets/RadioNew/Widget Internal Components/Indicator",
```
ivyolamit, line 30: Probably something to address later or in a different PR, we probably need the high-level css styling so that the choice indicator will be displayed properly and not like this:
image

ivyolamit (commented):
@SonicScrewdriver here's my initial comment. I like how this is setup now which makes me think do we also want to apply the same thing with Editor Demo?

packages/perseus/src/widgets/radio/__docs__/multiple-choice-regression.stories.tsx
mark-fitzgerald: I think it's because the FF isn't being set.

packages/perseus/src/widgets/radio/__docs__/multiple-choice.new.doc.stories.tsx
```diff
@@ -0,0 +1,226 @@
...
+                "An opaque string that uniquely identifies this choice within the radio widget",
+            table: {
+                category: "Perseus Radio Choice",
+                type: {summary: "string"},
+            },
+        },
+        rationale: {
+            control: false,
+            description:
+                "Translatable Markdown; Rationale to give the user when they get it wrong",
```
mark-fitzgerald: Perhaps expand the explanation to be:
```
Translatable Markdown; Rationale to give the user as to why an answer is correct or not
```

packages/perseus/src/widgets/radio/__stories__/choice-indicator.stories.tsx
mark-fitzgerald: Agreed - this should be a separate PR. Please write a ticket for it.

packages/perseus/src/widgets/radio/__docs__/multiple-choice-regression.stories.tsx
SonicScrewdriver: Oh I didn't realize storybook was surfacing those, I just added comments to help myself find key sections while adding new tests. I've removed them down and I just have a simpler comment (that doesn't seem to surface) in the code now.

packages/perseus/src/widgets/radio/__stories__/choice-indicator.stories.tsx
SonicScrewdriver: Yeah I called this out in the main PR text as I think there's something wrong here. I don't want to turn Chromatic on for something that's in a broken state.

packages/perseus/src/widgets/radio/__docs__/multiple-choice-regression.stories.tsx
SonicScrewdriver: Whoops my flags disappeared when I restructured the files! Fixed :D

mark-fitzgerald (approved):
Super helpful organization! Thank you!!

ivyolamit (approved):
this looks better thanks!

PR #2813, opened 2025-08-20
Allow iframe user inputs to have an undefined 'message' property

benchristel:
## Summary:
This fixes an issue found in side-by-side testing where iframes were scored
correct on the client but invalid on the server (because the user input failed
to parse).

Issue: https://khanacademy.atlassian.net/browse/LEMS-3414

## Test plan:

Deploy. The side-by-side errors should stop happening.

### DIFF
--- a/.changeset/dull-seas-attend.md
+++ b/.changeset/dull-seas-attend.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-core": minor
+---
+
+Allow the `message` property of `PerseusIFrameUserInput` to be `undefined`.


packages/perseus-core/src/parse-perseus-json/perseus-parsers/iframe-user-input.ts
```diff
@@ -2,10 +2,11 @@ import {
...
     nullable,
     object,
+    optional,
     string,
 } from "../general-purpose-parsers";
 
 export const parseIFrameUserInput = object({
     status: enumeration("correct", "incorrect", "incomplete"),
-    message: nullable(string),
+    message: optional(nullable(string)),
```
Myranae, line 11: This is interesting! So even though it was nullable, that would mean that message would still need to be in the object, but just empty? So now this change shows that it's okay to just completely leave the field out of the object?

Myranae (approved):
Left a comment with thoughts. Thanks for fixing!

packages/perseus-core/src/parse-perseus-json/perseus-parsers/iframe-user-input.ts
benchristel: Correct!

- `nullable()` only allows `null`, not `undefined`.
- `optional()` allows the property to be `undefined` or missing entirely.

PR #2815, opened 2025-08-20
[Radio] Adjust the transition time of the indicator so that it doesn't appear out of sync with the "Check" button

mark-fitzgerald:
## Summary
The transition of the indicator color from a background of white to blue/purple is long enough to make it appear that there is a delay of registering the selection when compared to the color change of the "Check" button. This CSS change speeds up the transition time in an effort to reduce the apparent lag between the two.

Issue: LEMS-3354


SonicScrewdriver (approved):
Wonderful, thank you!

PR #2816, opened 2025-08-20
[Radio] Adjust horizontal scroll fade bar color to use background color of container

mark-fitzgerald:
## Summary
The Graded Group Set has a background color that is not the page color.
When a radio widget is included in this widget, the fade bars used for horizontal scrolling show as white against the gray background.
The radio widget doesn't know that it is being displayed in the graded group set.
This bugfix adds a CSS variable to the graded group set container element that indicates the background color.
The fade bars in the radio widget now reference this new background color variable, but falls back to the semantic color for backgrounds if a container color doesn't exist.

Issue: LEMS-3405

## Test plan:
1. Open Storybook (either for this PR, or locally)
1. Navigate to the new story: Widges => RadioNew => Docs => Scrolling in Graded Group Set
1. Note that when scrolling horizontally, the fade bars match the background color

### DIFF
--- a/packages/perseus/src/widgets/graded-group-set/graded-group-set.tsx
+++ b/packages/perseus/src/widgets/graded-group-set/graded-group-set.tsx
@@ -308,6 +308,10 @@ const styles = StyleSheet.create({
         borderTop: `1px solid ${gray76}`,
         borderBottom: `1px solid ${gray76}`,
         backgroundColor: tableBackgroundAccent,
+        // The following CSS variable is put into place to let child widgets know that there is a background color in effect.
+        // For example, this is helpful for the radio widget to properly color its horizontal scroll fade bars.
+        //  @ts-expect-error TS2353: Object literal may only specify known properties
+        "--perseus-widget-background-color": tableBackgroundAccent,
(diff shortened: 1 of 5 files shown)


packages/perseus/src/widgets/graded-group-set/graded-group-set.tsx
```diff
@@ -308,6 +308,10 @@ const styles = StyleSheet.create({
         borderTop: `1px solid ${gray76}`,
         borderBottom: `1px solid ${gray76}`,
         backgroundColor: tableBackgroundAccent,
+        // The following CSS variable is put into place to let child widgets know that there is a background color in effect.
+        // For example, this is helpful for the radio widget to properly color its horizontal scroll fade bars.
+        //  @ts-expect-error TS2353: Object literal may only specify known properties
+        "--perseus-widget-background-color": tableBackgroundAccent,
```
SonicScrewdriver, line 314: Ooooo good to know this still works! I'm fine with this as it seems like we can clean up all the styles/aphrodite when we get a chance to overhaul this widget.

PR #2819, opened 2025-08-21
[Image] | (DX) | Add regression stories for Image widget

nishasy:
## Summary:
Add .regression.stories.tsx file so that Chromatic can catch any unintended regressions to the widget.

Issue: https://khanacademy.atlassian.net/browse/LEMS-3385

image

## Test plan:
Check for `UI Tests` below and confirm all the stories added in this PR are shown as new stories in Chromatic.


packages/perseus/src/widgets/image/__docs__/image-regression.stories.tsx
```diff
@@ -0,0 +1,172 @@
...
+    component: ImageQuestionRenderer,
+    tags: ["!dev"],
+    parameters: {
+        chromatic: {disableSnapshot: false},
+    },
+};
+export default meta;
+
+// Avoid using `args` from the first story in order to remove
+// the control panel from the storybook UI.
```
ivyolamit: TIL 👍🏼

ivyolamit: 👆🏼 @SonicScrewdriver something we can apply in RadioNew 😉

packages/perseus/src/widgets/image/__docs__/image-regression.stories.tsx
```diff
@@ -0,0 +1,172 @@
...
+    args: {
+        question: generateTestPerseusRenderer({
+            content: "[[☃ image 1]]",
+            widgets: {
+                "image 1": generateImageWidget({
+                    options: generateImageOptions({
+                        backgroundImage: {
+                            url: earthMoonImageUrl,
+                        },
+                        title: "Earth and Moon",
```
ivyolamit: Will the title always aligned this way?
image

ivyolamit (approved):
@nishasy changes looks good to me, just one question that is non-blocking.

packages/perseus/src/widgets/image/__docs__/image-regression.stories.tsx
nishasy: That's a really good point. The image widget is supposed to be center aligned, so I'm not sure why it's left aligned here. 🤔 I'll look into that.

nishasy: It's center aligned when I use the dev tools make the renderer preview big 🤔 🤔 🤔 

Maybe something is missing in how it's rendered within the regression stories.

Screenshot 2025-08-22 at 11 13 12 AM

mark-fitzgerald: There might be styling that determines alignment based upon whether it's in an article or exercise. So, that could affect how this is rendered in Storybook (with its hit-or-miss inclusion of frontend context).

packages/perseus/src/widgets/image/__docs__/image-regression.stories.tsx
nishasy: Just an FYI that I'm not sure this works in 100% of the cases. I couldn't get this to work for the editor stories for some reason 🤔

packages/perseus/src/widgets/image/__docs__/image-regression.stories.tsx
nishasy: I'm going to create a new ticket to fix this, because I want these regression stories to be out so we can work on more stuff with a bit more security.

packages/perseus/src/widgets/image/__docs__/image-regression.stories.tsx
```diff
@@ -0,0 +1,158 @@
...
+    args: {
+        backgroundImage: {
+            url: frescoImageUrl,
+            width: 1698,
+            height: 955,
+        },
+    },
+};
+
+export const RightToLeftImage: Story = {
```
ivyolamit, line 118: non-blocking comment, just a question about the widget:
For RTL, was it intentional that the caption vertical line will remain on the left instead on the right?

Screenshot 2025-08-25 at 8 51 23 AM

ivyolamit (approved):
👍🏼 LGTM

packages/perseus/src/widgets/image/__docs__/image-regression.stories.tsx
nishasy: Yeah, it's just like that. There's already a [ticket to fix this](https://khanacademy.atlassian.net/browse/LEMS-3425).

nishasy: I wanted the regression available for this so we can see the change when we do finally fix it 😁

PR #2820, opened 2025-08-21
docs(changeset): Remove the isLastUsedWidget property from Perseus, as it is not used.

SonicScrewdriver:
## Summary:
Remove isLastWidgetUsed from Perseus as it was unused, and really only ever used for the Radio Widget. 

Issue: LEMS-3363

## Test plan:
- Tests pass

### DIFF
--- a/packages/perseus/src/renderer.tsx
+++ b/packages/perseus/src/renderer.tsx
@@ -157,7 +157,6 @@ type State = {
     translationLintErrors: ReadonlyArray<string>;
     widgetInfo: Readonly<PerseusWidgetsMap>;
     jiptContent: any;
-    lastUsedWidgetId: string | null | undefined;
 };
 
 type FullLinterContext = LinterContextProps & {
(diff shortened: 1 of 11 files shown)


packages/perseus/src/renderer.tsx
```diff
@@ -1649,49 +1642,34 @@ class Renderer
...
-                        // TODO(jack): Figure out why this is happening and fix it
-                        // As far as I can tell, this is only an issue in the
-                        // editor-page, so doing this shouldn't break clients
-                        // hopefully
-                        this._setCurrentFocus([id]);
-                    }
-                }, 0);
-            },
-        );
+        this.setState({}, () => {
```
SonicScrewdriver: We likely don't need this set state at all anymore

SonicScrewdriver: Hm maybe it's better to keep it, given that the other comments are specific about our timing here. I'll try testing their called out example (pasting orderer) over in frontend.

SonicScrewdriver:
@ivyolamit yes! I saw that yesterday - I actually meant to speak with Jeremy about that file.

PR #2821, opened 2025-08-21
Fix mobile new radio widget scroll styling removing extra spaces.

ivyolamit:
## Summary:
Fix mobile new radio widget scroll styling removing extra spaces.

Issue: LEMS-3369

## Test plan:

### DIFF
--- a/packages/perseus/src/components/scrollable-view.module.css
+++ b/packages/perseus/src/components/scrollable-view.module.css
@@ -4,4 +4,7 @@
     flex-direction: row;
     gap: 0.8rem;
     padding: 1.2rem;
+    /* Ensure scroll buttons stay above fade effects and remain clickable */
+    position: relative;
+    z-index: 3;
 }
--- a/packages/perseus/src/components/scrollable-view.tsx
+++ b/packages/perseus/src/components/scrollable-view.tsx
@@ -45,6 +45,17 @@ function ScrollableView({
     const [canScrollStart, setCanScrollStart] = useState(false);
     const [canScrollEnd, setCanScrollEnd] = useState(false);
     const [isRtl, setIsRtl] = useState(false);
+    const [isScrolling, setIsScrolling] = useState(false);
+
+    // Calculate the scrollable threshold once to avoid performance issues during scroll events
+    const scrollableThreshold = React.useMemo(() => {
+        const isMobile = window.innerWidth <= 767;
+        if (!isMobile) {
+            return 5; // Default for desktop
+        }
+
+        return 8; // Reduced threshold for optimized mobile spacing
+    }, []); // Empty dependency array - calculate only once
 
     /**
      * Updates scroll state variables based on current scroll position.
@@ -79,7 +90,7 @@ function ScrollableView({
 
         // Only consider content scrollable if there's a meaningful amount to scroll
         // Using a slightly higher threshold to prevent micro-scrolling issues
-        const scrollableThreshold = 5; // 5px threshold
+        // Using the pre-calculated threshold to avoid performance issues during scroll events
         setIsScrollable(scrollWidth > clientWidth + scrollableThreshold);
 
         // In RTL mode, scrollLeft values work differently (can be negative)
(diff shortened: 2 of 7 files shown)


packages/perseus/src/components/scrollable-view.module.css
```diff
@@ -4,4 +4,7 @@
     flex-direction: row;
     gap: 0.8rem;
     padding: 1.2rem;
+    /* Ensure scroll buttons stay above fade effects and remain clickable */
+    position: relative;
+    z-index: 3;
```
mark-fitzgerald: I recommend using `z-index: 10;` to allow for other elements to fit between the fade and the buttons.

packages/perseus/src/components/scrollable-view.tsx
```diff
@@ -45,6 +45,17 @@ function ScrollableView({
     const [canScrollStart, setCanScrollStart] = useState(false);
     const [canScrollEnd, setCanScrollEnd] = useState(false);
     const [isRtl, setIsRtl] = useState(false);
+    const [isScrolling, setIsScrolling] = useState(false);
+
+    // Calculate the scrollable threshold once to avoid performance issues during scroll events
+    const scrollableThreshold = React.useMemo(() => {
+        const isMobile = window.innerWidth <= 767;
```
mark-fitzgerald: Does this threshold get triggered on desktop when either the window is smaller than 767, or when zoomed in makes it effectively less than this value? If so, then I recommend calling this something else, or just rewriting to something like:
```
return window.innerWidth <= 767 ? 8 : 5;
```

mark-fitzgerald: Also, explaining the relevance for 767 would be helpful. Perhaps a comment explaining why this breakpoint is used?

packages/perseus/src/styles/styles.css
```diff
@@ -543,7 +543,7 @@
         border-left: 4px solid #d8d8d8;
     }
 }
-@media (min-width: 768px) and (max-width: 1199px) {
+@media (min-width: 767px) and (max-width: 1199px) {
```
mark-fitzgerald, line 546: Why the change of 1px?

packages/perseus/src/widgets/radio/multiple-choice.module.css
```diff
@@ -33,6 +33,7 @@
 .choice-list {
     padding: 0 !important;
     list-style: none !important;
+    margin: 0 !important; /* TODO(LEMS-3301): When old radio widget is deleted, remove this !important. */
```
mark-fitzgerald: This has more to do with the styling in `perseus/src/styles/perseus-renderer-part-1.css`. I recommend updating the comment to reference that file so that there isn't a regression due to the other CSS file not being updated.

packages/perseus/src/widgets/radio/choice-indicator.module.css
```diff
@@ -18,6 +18,8 @@
      */
     position: fixed;
     width: var(--perseus-multiple-choice-indicator-size);
+    /* Ensure indicator stays above fade effects and other elements */
+    z-index: 2;
```
mark-fitzgerald: Which elements were on top of the indicator? The scroll fade bars weren't on top of the indicator - were they in the mobile view?

packages/perseus/src/styles/styles.css
ivyolamit: This is for consistency in our code, we use `767px`

packages/perseus/src/components/scrollable-view.tsx
ivyolamit: Great suggestion! i'll update that.

ivyolamit:
Updated [ZND](https://prod-znd-250822-3116-98718c0.khanacademy.org/math/early-math/cc-early-math-add-sub-basics/cc-early-math[%E2%80%A6]b-word-problem-within-10/e/addition-word-problems-within-10) after the review comment changes: 


https://github.com/user-attachments/assets/c78230d4-ea03-4a11-a310-443ee1ba183e

anakaren-rojas (approved):
Tested in chromatic, looks good!
Screenshot 2025-08-25 at 14 19 31

PR #2823, opened 2025-08-21
[Image] | (DX) | Update the Storybook experience for Image Widget

nishasy:
## Summary:
- Update image.stories.tsx
  - Move images.stories.tsx under `__docs__` folder
  - Update the questions so they make sense and are not redundant
- Update image-editor.stories.tsx
  - Add argTypes so that unused props are not editable
  - Add a "WithinEditorPage" story

Issue: https://khanacademy.atlassian.net/browse/LEMS-3384

## Test plan:
- http://localhost:6006/?path=/docs/widgets-image--docs
- http://localhost:6006/?path=/docs/widgets-image-editor-demo--docs


packages/perseus/src/widgets/image/__docs__/image.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+    },
+};
+
+/**
+ * An image in a narrow container - tap it to zoom.
+ */
+export const ImageWithZoom: Story = {
+    decorators: [
+        (Story) => (
+            <div style={{width: "50%", margin: "0 auto"}}>
```
jeremywiebe: FWIW, I wonder if we should just use a specific pixel width instead of a percentage.  Otherwise the story may, or may not, do the tap-to-zoom thing if the device is wide enough (or will it always... I can't remember).

packages/perseus/src/widgets/image/__docs__/image.stories.tsx
```diff
@@ -0,0 +1,51 @@
+import {generateTestPerseusItem} from "@khanacademy/perseus-core";
+import * as React from "react";
+
+import {ServerItemRendererWithDebugUI} from "../../../../../../testing/server-item-renderer-with-debug-ui";
```
jeremywiebe, line 9: **off-topic**: What's a few (hundred!) dots... 😅 

I've sometimes thought about making that `testing` folder an actual private npm package and then we could reference it a bit more easily.

jeremywiebe (commented):
I'll leave it for Ivy or Cat to approve, but had two quick thoughts. :) (also, thanks for continuing to improve Storybook!!)

packages/perseus/src/widgets/image/__docs__/image.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+import {question, questionWithZoom} from "../image.testdata";
+
+import type {Meta, StoryObj} from "@storybook/react-vite";
+
+const meta: Meta = {
+    title: "Widgets/Image",
+    component: ServerItemRendererWithDebugUI,
+    tags: ["!dev"],
```
ivyolamit, line 20: This is not a blocker in your PR, bu something that was discussed in another [PR](https://github.com/Khan/perseus/pull/2812), the New Radio widget is trying something different see it [here](https://650db21c3f5d1b2f13c02952-glqabmwzkr.chromatic.com/?path=/story/widgets-radionew-widget-demo--answerless-multi-select)
- Docs is barebones in terms of information but it does have the correct props table now
- The Widget Demo has the `tags: ["!autodocs"]` making all stories interactive and per page

Advantage: info is close to our widget props and widget demo is interactive
Cons: overhead maintenance, @SonicScrewdriver had to create manual table for the props 😢 

I would be interested in what your thoughts about it 🙏🏼

ivyolamit (approved):
LGTM 👍🏼

packages/perseus/src/widgets/image/__docs__/image.stories.tsx
nishasy: I _hate_ that the table shows the props for ServerItemRendererWithDebugUI instead of the actual image widget. I tried for soooo loooong to figure out how to get the right args without using an argtypes table but ultimately gave up.

I do prefer the stories being interactive and per page because I find those easier to use for manual testing.

I'll look into it further to see if I can take inspiration from Radio.

jeremywiebe: What if...   we made a Storybook decorator that did the ServerItemRendererWithDebugUI wrapping.  Then our `Meta` for each widget's stories would be for the actual Widget's React component. The table for props would be correct because it would be derived from the widget's component props.  And the decorator could take those props and wrap them in a `PerseusItem`.

packages/perseus/src/widgets/image/__docs__/image.stories.tsx
nishasy: We were talking about that for the builders/generators too!

I ended up putting the image generator inside `perseus-core`, but I'd be in favor of a testing-specific package. (In a separate PR though.)

PR #2824, opened 2025-08-21
[Image] | (DX) | Modernize the Image widget code

nishasy:
## Summary:
Convert the image widget component so it's a functional component.
The top level is still a class component.

Taking inspiration from the Numeric Input widget.

Issue: https://khanacademy.atlassian.net/browse/LEMS-3382

## Test plan:
`pnpm jest`
`pnpm typecheck`
`pnpm lint`

Check the regression stories in this PR

`/?path=/docs/widgets-image-visual-regression-tests--docs`


packages/perseus/src/widgets/image/image.class.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+const defaultBoxSize = 400;
+const defaultRange: Range = [0, 10];
+const defaultBackgroundImage = {
+    url: null,
+    width: 0,
+    height: 0,
+} as const;
+
+type ExternalProps = WidgetProps<PerseusImageWidgetOptions>;
```
ivyolamit, line 22: I like how this is named as `ExternalProps` it's more descriptive and intentional, normally in our code base we just use `Props` it would be nice to have this as a standard naming in our widgets since this is a normal thing we do - for consistency.

packages/perseus/src/widgets/image/image.class.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+export type ImageProps = ExternalProps & {
```
ivyolamit: Base on team discussion earlier this year `interface` is recommended over `type` and we are aiming switching over to it.

```
export interface ImageProps extends ExternalProps {
    alignment: NonNullable<ExternalProps["alignment"]>;
    title: NonNullable<ExternalProps["title"]>;
    range: NonNullable<ExternalProps["range"]>;
    box: NonNullable<ExternalProps["box"]>;
    backgroundImage: NonNullable<ExternalProps["backgroundImage"]>;
    labels: NonNullable<ExternalProps["labels"]>;
    alt: NonNullable<ExternalProps["alt"]>;
    caption: NonNullable<ExternalProps["caption"]>;
    linterContext: NonNullable<ExternalProps["linterContext"]>;
}
```

packages/perseus/src/widgets/image/image.class.tsx
```diff
@@ -0,0 +1,82 @@
...
+    range: NonNullable<ExternalProps["range"]>;
+    box: NonNullable<ExternalProps["box"]>;
+    backgroundImage: NonNullable<ExternalProps["backgroundImage"]>;
+    labels: NonNullable<ExternalProps["labels"]>;
+    alt: NonNullable<ExternalProps["alt"]>;
+    caption: NonNullable<ExternalProps["caption"]>;
+    linterContext: NonNullable<ExternalProps["linterContext"]>;
+};
+
+type DefaultProps = {
```
ivyolamit:
```suggestion
interface DefaultProps extends Pick<
    ImageProps,
    "alignment" |
    "title" |
    "range" |
    "box" |
    "backgroundImage" |
    "labels" |
    "alt" |
    "caption" |
    "linterContext"
> {}
```
Using pick here will help avoid repetition and keeps types consistent.

packages/perseus/src/widgets/image/image.tsx
```diff
@@ -1,161 +1,85 @@
...
+import type {ImageProps} from "./image.class";
+
+export const ImageComponent = ({
+    apiOptions,
+    alt,
+    backgroundImage,
+    box, // automatically calculated by the editor
+    caption,
+    linterContext,
+    labels, // deprecated
```
ivyolamit: Suggestion: Add the `@deprecated` tag in the interface JSDoc, in this way lint will mark the props as deprecated and when we are able to pull in the interface in our storybook, it will be visible.

packages/perseus/src/widgets/image/image.tsx
```diff
@@ -1,161 +1,85 @@
...
+                    />
+                </div>
+            )}
+            {backgroundImage.url && (
                 <AssetContext.Consumer>
                     {({setAssetStatus}) => (
                         <SvgImage
-                            src={url}
-                            alt={alt}
+                            src={backgroundImage.url ?? undefined}
```
ivyolamit: What happens if the source is undefined?
previous implementation made user the url is existing.

packages/perseus/src/widgets/image/image.tsx
```diff
@@ -1,161 +1,85 @@
...
-                        strings={this.context.strings}
-                    />
-                </div>
-            );
-        }
-
-        if (this.props.caption) {
-            caption = (
+            )}
+            {caption && (
```
ivyolamit, line 71: I'm on the fence if it's too early to suggest code refactor here like:
```
    return (
        <figure
            className="perseus-image-widget"
            style={{
                maxWidth: backgroundImage.width,
            }}
        >
            <Title ... />
            <BackgroundImage ... />
            <Caption ... />
        </figure>
```

Note `...` above are the contents of the existing logic
Having the code like this it's more readable that the image will have the 3 parts Title, Background Image, and the Caption. Let me know what you think?

ivyolamit (commented):
@nishasy Added few suggestion and comments, I will be happy to discuss it with you and see what you think.

packages/perseus/src/widgets/image/image.class.tsx
nishasy: Would this be the same as using `interface DefaultProps extends Partial<ImageProps> {...}`?

packages/perseus/src/widgets/image/image.tsx
nishasy: It can't be undefined because of the `{backgroundImage.url &&` on line 43. I added this because it was giving me a type error saying that `string | undefined | null` is not compatible with `string | undefined`.

I can add a comment to explain that.

packages/perseus/src/widgets/image/image.tsx
nishasy: I'm down to do that if the code here gets longer, but the code is so short that that feels like overkill to me as is.

I can add some comments to make it more readable in the meantime.

packages/perseus/src/widgets/image/image.tsx
nishasy: Oh, I can use `backgroundImage.url!` to tell it that I'm sure it exists.

nishasy:
Why do comments count as schema changes?

packages/perseus/src/widgets/image/image.class.tsx
```diff
@@ -0,0 +1,82 @@
...
+    range: ImageProps["range"];
+    box: ImageProps["box"];
+    backgroundImage: ImageProps["backgroundImage"];
+    labels: ImageProps["labels"];
+    alt: ImageProps["alt"];
+    caption: ImageProps["caption"];
+    linterContext: ImageProps["linterContext"];
+};
+
+class ImageWidget extends React.Component<ImageProps> implements Widget {
```
jeremywiebe: Perhaps controversial, but I really like seeing simply `Props` as the type for a React component (except in the rare-ish cases where we have more than one component in the file).  We rarely export the prop type (in fact, we likely don't need un most cases to given that we have Wonder-Block's `PropsFor<>` helper type).

packages/perseus/src/widgets/image/image.class.tsx
jeremywiebe: In Perseus, the convention is that `ExternalProps` is the sum of `RenderProps` wrapped in `WidgetOptions`.  The final type for the component itself should, imho, be simply `Props`.

https://khan.github.io/perseus/?path=/docs/introduction--docs#react-types

nishasy: @jeremywiebe I was thinking the same thing initially, but it looks like only Expression, Input Number, and Numeric Input have ExternalProps now, outside of Image. Does it still count as a convention?

packages/perseus-core/src/data-schema.ts
```diff
@@ -693,12 +693,15 @@ export type PerseusImageWidgetOptions = {
     static?: boolean;
     // A list of labels to display on the image
     // NOTE: perseus_data.go says this is required even though it isn't necessary.
+    /** @deprecated - labels were removed from the image widget in 2017 */
```
catandthemachines, line 696: The labels property is coming back in the upcoming project right? Or am I thinking of the `title` property? 🤔

jeremywiebe:
> Why do comments count as schema changes?

The schema change detection is simply detecting if _any_ changes were made to the `data-schema.ts` file. It's meant to signal you to be a bit more careful with the PR changes, not block you. Comment changes can safely be ignored and no changes needed.

packages/perseus/src/widgets/image/image.class.tsx
jeremywiebe: Yeah, looks like it's not as much of a convention anymore. I think with Matthew's work we're getting closer to the raw widget options that are defined in `data-schema.ts` being what's passed into each widget. So I'm fine if we're migrating away. I _would_ like to see us being consistent though. Perseus has many well-intentioned changes that were left 80% done and it makes things even more complicated to maintain.

catandthemachines (approved):
This looks like a good step in modernizing the image widget code. Had a few questions, and ++ to a few comments, but looks overall good to me.

packages/perseus/src/widgets/image/image.class.tsx
nishasy: Discussed separately. `Partial` is fine.

packages/perseus-core/src/data-schema.ts
nishasy: `title` is coming back! Labels are deprecated.

PR #2822, opened 2025-08-21
[Radio] Fix missing styling when showing solution in graded group

mark-fitzgerald:
## Summary:
The Graded Group widget would signal to its child widgets when to show answers by using `showSolutions` instead of `reviewMode`. This PR updates the radio component to also check for that property when telling its components when to show answer information.

Issue: LEMS-3428

## Test plan:
1. Open Storybook (either locally or linked in this PR)
1. Navigate to Widgets => RadioNew => Docs => Graded Group
1. Choose the "correct" answer from the options shown
1. Click the "Check" button to show the answers
   - Note that the indicator now shows correctness (icon and color change)
   - Note that the choice is now outlined in green
   - Note that all rationale are showing, as well

PR #2826, opened 2025-08-22
[Radio] Fix how Graphie text is placed when inside a choice

mark-fitzgerald:
## Summary
Styling exists in the old radio widget that excludes graphie images. This PR adds that exclusion to the new CSS styling. This fixes a bug where the text in the Graphie was shifted down.

### Before
Graphie in choice - Before

### After
Graphie in choice - After

Issue: LEMS-3353

## Test plan:
1. Open Storybook (either locally or linked in this PR)
1. Navigate to Widgets => RadioNew => Docs => Select With Graphie
   - Note that the placement of text in the choice images match the example image in the question text


packages/perseus/src/widgets/radio/__docs__/multiple-choice-demo.new.stories.tsx
```diff
@@ -114,6 +115,14 @@ export const SingleSelect = {
     },
 };
 
+export const SelectWithGraphie = {
```
ivyolamit, line 119: This will be valuable item in Visual Regression Tests, let's add it there, makes it easier for spot checking.

ivyolamit (approved):
@mark-fitzgerald changes here looks logical to me. I have one suggestion below.

anakaren-rojas:
Would it be feasible to add unit tests that capture the fix? Even just a snapshot test that captures the correct structure would work

packages/perseus/src/widgets/radio/__docs__/multiple-choice-demo.new.stories.tsx
mark-fitzgerald: I'll move it once I land the other PR (#2822).

mark-fitzgerald:
> Would it be feasible to add unit tests that capture the fix? Even just a snapshot test that captures the correct structure would work

A DOM snapshot would not have caught this issue. It is purely visual (aka CSS). Moving the Storybook file to the visual regression test suite will catch such issues going forward.

PR #2828, opened 2025-08-22
Fix to allow users to scroll content after multiple choice is checked

ivyolamit:
## Summary:
Fix to allow users to scroll content after multiple choice is checked

Issue: LEMS-3406

## Test plan:


ivyolamit:
Refer to this updated PR instead: https://github.com/Khan/perseus/pull/2831

PR #2829, opened 2025-08-25
Export deriveUserInputFromSerializedState

handeyeco:
## Summary:

See: https://github.com/Khan/frontend/pull/3107

I'm using this in a hack for TX to restore user input from serialized state.

### DIFF
--- a/packages/perseus/src/index.ts
+++ b/packages/perseus/src/index.ts
@@ -156,7 +156,7 @@ export {
 } from "./widgets/interactive-graphs/reducer/initialize-graph-state";
 export {
     default as UserInputManager,
-    sharedRestoreUserInputFromSerializedState,
+    deriveUserInputFromSerializedState,
 } from "./user-input-manager";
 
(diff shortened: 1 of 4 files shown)


packages/perseus/src/index.ts
```diff
@@ -154,7 +154,10 @@ export {
     getQuadraticCoords,
     getAngleCoords,
 } from "./widgets/interactive-graphs/reducer/initialize-graph-state";
-export {default as UserInputManager} from "./user-input-manager";
+export {
+    default as UserInputManager,
+    sharedRestoreUserInputFromSerializedState,
```
benchristel: What does `shared` mean here? Could we just call this function `restoreUserInputFromSerializedState`?

handeyeco: There was already a `restoreUserInputFromSerializedState` so I renamed `sharedRestoreUserInputFromSerializedState` to `deriveUserInputFromSerializedState`.

PR #2831, opened 2025-08-25
[Radio] Fix to allow users to scroll content after multiple choice is checked

ivyolamit:
## Summary:
Fix to allow users to still scroll the scrollable content when the fieldset is disabled and staticOverlayStyles is applied after the user checked the answer

**Before**
https://github.com/user-attachments/assets/c63567f9-771e-4b60-bb7c-7d0dd4e7ff20

**After**

Issue: LEMS-3406

## Test plan:
1. Navigate to the [ZND link](https://prod-znd-250826-3249-4e4c9ff.khanacademy.org/math/early-math/cc-early-math-add-sub-basics/cc-early-math[%E2%80%A6]b-word-problem-within-10/e/addition-word-problems-within-10)
2. In a question with a scrollable view, select the correct answer
3. After the question is marked correct
- [ ] Confirm that the scroll buttons still works left and right

### DIFF
--- a/packages/perseus/src/components/scrollable-view.test.tsx
+++ b/packages/perseus/src/components/scrollable-view.test.tsx
@@ -44,11 +44,9 @@ const setupScrollableMock = (isScrollable: boolean) => {
         this: HTMLElement,
         options?: ScrollToOptions | undefined,
     ) {
-        const element = screen.getByRole("group");
         if (options !== undefined && options !== null) {
-            const currentScrollLeft = (element as any).scrollLeft || 0;
-            (element as any).scrollLeft =
-                currentScrollLeft + (options.left || 0);
+            const currentScrollLeft = (this as any).scrollLeft || 0;
+            (this as any).scrollLeft = currentScrollLeft + (options.left || 0);
         }
     }) as any;
 };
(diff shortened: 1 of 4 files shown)


packages/perseus/src/widgets/radio/multiple-choice-component.new.tsx
```diff
@@ -71,30 +71,38 @@ const MultipleChoiceComponent = ({
...
-            </ScrollableView>
-        </fieldset>
+                <legend
+                    id={legendId}
+                    aria-hidden="true"
+                    className={styles.instructions}
+                >
+                    {instructions}
+                </legend>
+                <ScrollableView id={scrollId} overflowX="auto">
```
ivyolamit, line 89: ScrollableView was refactored so that the left and right buttons will be outside the fieldset block which would add the `staticOverlayStyles` css once the multiple choice is checked and this will cause the area not intractable.

packages/perseus/src/components/scrollable-view.tsx
```diff
@@ -59,6 +82,43 @@ function ScrollableView({
         return window.innerWidth <= 767 ? 8 : 5;
     }, []);
 
+    const scroll = React.useCallback(
```
ivyolamit, line 85: Logic is still the same from [original](./dev/tools/bump_perseus_version.ts -t PR2831) except for adding the `React.useCallback` here

packages/perseus/src/components/scrollable-view.test.tsx
```diff
@@ -44,11 +44,9 @@ const setupScrollableMock = (isScrollable: boolean) => {
         this: HTMLElement,
         options?: ScrollToOptions | undefined,
     ) {
-        const element = screen.getByRole("group");
         if (options !== undefined && options !== null) {
-            const currentScrollLeft = (element as any).scrollLeft || 0;
-            (element as any).scrollLeft =
-                currentScrollLeft + (options.left || 0);
+            const currentScrollLeft = (this as any).scrollLeft || 0;
+            (this as any).scrollLeft = currentScrollLeft + (options.left || 0);
```
SonicScrewdriver: I think we can avoid the type casting here:
```suggestion
             const currentScrollLeft = this.scrollLeft ?? 0;
            this.scrollLeft = currentScrollLeft + (options.left || 0);
```

packages/perseus/src/components/scrollable-view.tsx
SonicScrewdriver: Thanks! That's very helpful for reviewing given the larger diff :)

SonicScrewdriver (changes requested):
Great work - thank you! I just have one quick change request for fixing up the any casting, and then I think this is good to go. :)

SonicScrewdriver (approved):
Thank you for getting this working! :)

packages/perseus/src/widgets/radio/multiple-choice-component.new.tsx
mark-fitzgerald: I think this comment would be very helpful to have in the `scrollable-view.tsx` file to explain the reasoning behind the split in component parts.

I'm curious as to why/how the `staticOverlayStyles` class is restricted to the `<fieldset>` element.

mark-fitzgerald (approved):
Another reason why the static designation needs to be revised. That overlay is so intrusive! Thanks for solving this tough issue.

packages/perseus/src/components/scrollable-view.tsx
```diff
@@ -173,81 +221,110 @@ function ScrollableView({
...
+
+    const description = overrideDescription || scrollState.scrollDescription;
 
     return (
-        <div className={styles.scrollButtonsContainer}>
+        <div
+            className={styles.scrollButtonsContainer}
+            aria-live="polite"
+            role="group"
+            aria-label={description}
```
mark-fitzgerald, line 282: I think an explanation as to how this `aria-live` region is supposed to work would be helpful. When should we expect a readout? How often will it update?

Also, just checking that all of the descriptions being passed to this component have been translated?

ivyolamit: Normally for the `aria-label={description}` when content changes, screen readers will announce the new text. The `aria-live` `polite` will be helpful here since it would wait for the user to finish what they're doing before announcing, in this case scrolling to left and right.

ivyolamit: > Also, just checking that all of the descriptions being passed to this component have been translated?
Yes they are, this was added before in `packages/perseus/src/strings.ts`

PR #2832, opened 2025-08-26
Add utility function to extract widget IDs from renderer

Myranae:
## Summary:
 Adds a new utility function `extractWidgetIds()` that extracts widget IDs from a Perseus Renderer instance in the order they appear in the content, with deduplication logic that matches the renderer's existing behavior. Written with Claude Code.

Issue: LEMS-3432

## Test plan:
- Confirm all checks pass


packages/perseus/src/util/extract-widget-ids.ts
```diff
@@ -0,0 +1,41 @@
...
+    function collectWidgetIds(ast: any): void {
+        if (Array.isArray(ast)) {
+            ast.forEach(collectWidgetIds);
+        } else if (ast?.type === "widget") {
+            // Only add if not already present (deduplication)
+            if (!widgetIds.includes(ast.id)) {
+                widgetIds.push(ast.id);
+            }
+        } else if (ast?.content) {
+            collectWidgetIds(ast.content);
```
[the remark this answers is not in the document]
Myranae: Added a type

packages/perseus/src/util/extract-widget-ids.ts
```diff
@@ -0,0 +1,41 @@
...
+              isJipt: renderer.translationIndex != null,
+          });
+
+    // Walk the AST and collect widget IDs
+    function collectWidgetIds(ast: any): void {
+        if (Array.isArray(ast)) {
+            ast.forEach(collectWidgetIds);
+        } else if (ast?.type === "widget") {
+            // Only add if not already present (deduplication)
+            if (!widgetIds.includes(ast.id)) {
```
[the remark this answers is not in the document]
Myranae: I wrote code for this update, but not sure if it is valuable here since I don't think we'll ever have lists that are very long. Does this seem like a good addition?

packages/perseus/src/util/extract-widget-ids.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+export function extractWidgetIds(renderer: Renderer): ReadonlyArray<string> {
+    const widgetIds: string[] = [];
+
+    // Get content from the renderer's props
+    const content = renderer.props.content;
+    const inline = renderer.props.inline;
+
+    // Parse the markdown content using the same logic as the renderer
+    const parsedMarkdown = inline
```
jeremywiebe, line 20: Thank-you for using the AST instead of regexes to do this. :)

packages/perseus/src/util/extract-widget-ids.ts
```diff
@@ -0,0 +1,42 @@
+import PerseusMarkdown from "../perseus-markdown";
+
+import type Renderer from "../renderer";
+import type {SingleASTNode} from "@khanacademy/simple-markdown";
+
+/**
+ * Extracts widget IDs from a Perseus Renderer in the order they appear in the content.
+ * Handles deduplication by only including the first occurrence of each widget ID.
+ */
+export function extractWidgetIds(renderer: Renderer): ReadonlyArray<string> {
```
handeyeco: This seems similar to `getWidgetIdsFromContent`, is there a reason we couldn't use that?

packages/perseus/src/util/extract-widget-ids.ts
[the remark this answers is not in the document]
benchristel: You are right that it's probably not worth using a Set for de-duplication here. Iterating over an array is very fast per-index; `Set.has()` is comparatively slow. The advantage of a Set is that `Set.has()` doesn't get any _slower_ as the set gets large. An interesting side project for you might be to determine where the break-even point is; i.e. how many widgets would there need to be before a Set is worth it? I recently learned about [`benchmark`](https://www.npmjs.com/package/benchmark), which makes it pretty easy to measure relative performance.

packages/perseus/src/util/extract-widget-ids.ts
benchristel: `getWidgetIdsFromContent` doesn't actually parse the markdown; it just searches for the snowman placeholder using a regex. Is that the approach we want to take? It can produce false positives, e.g. a code block in the markdown like

````markdown
```markdown
[[☃ radio 1]]
```
````

will be identified as a widget by the regex but not by the markdown parser.

Comment (handeyeco) [reply packages/perseus/src/util/extract-widget-ids.ts]: I don't have a strong opinion on how it should work, I just think if `getWidgetIdsFromContent` isn't doing what we want it should be refactored or replaced.

Comment (Myranae) [reply packages/perseus/src/util/extract-widget-ids.ts]: I can make a ticket to update usages of `getWidgetIdsFromContent` if that is something we need. Should I move this function into that same file at least? Might clutter up the `widget-id-utils.test.ts` file if I move all the relevant tests there though 🤔

File: packages/perseus/src/util/extract-widget-ids.ts [hunk 5]
```diff
@@ -0,0 +1,45 @@
...
+        } else if (ast?.type === "widget") {
+            // Only add if not already seen (deduplication with O(1) lookup)
+            if (!seenIds.has(ast.id)) {
+                seenIds.add(ast.id);
+                widgetIds.push(ast.id);
+            }
+        } else if (ast?.content) {
+            collectWidgetIds(ast.content);
+        }
+    }
```
benchristel: I think there is already a function for traversing a Markdown AST and processing every node, so you don't need to write these recursive calls yourself.

See this example (thanks to Jeremy for digging it up!) https://github.com/Khan/frontend/blob/cc7ed6299176632d581e72fe307c9fd6ca7a0bdb/libs/component/perseus/src/perseus-content-contains-image.ts#L8

packages/perseus/src/util/extract-widget-ids.ts
```diff
@@ -0,0 +1,45 @@
...
+import type {PerseusRenderer} from "@khanacademy/perseus-core";
+import type {SingleASTNode} from "@khanacademy/simple-markdown";
+
+/**
+ * Extracts widget IDs from a PerseusRenderer in the order they appear in the content.
+ * Handles deduplication by only including the first occurrence of each widget ID.
+ */
+export function extractWidgetIds(
+    perseusRenderer: PerseusRenderer,
+    options?: {inline?: boolean; isJipt?: boolean},
```
benchristel: What does the `isJipt` option do? There aren't any tests for the case where it's `true`. I'd just as soon remove this option if we don't need it. We can always add it later.

packages/perseus/src/util/extract-widget-ids.test.ts
```diff
@@ -0,0 +1,308 @@
...
+                    type: "radio",
+                    graded: true,
+                    options: {
+                        choices: [
+                            {content: "Option A", correct: true, id: "a"},
+                            {content: "Option B", correct: false, id: "b"},
+                        ],
+                        randomize: false,
+                    },
+                },
```
benchristel: The actual configuration of these widgets doesn't matter, right? I think the tests could be abbreviated if the widget data were extracted to a const:

```ts
const question: PerseusRenderer = {
  content,
  widgets: {
    "radio 1": radio,
    "numeric-input 2": numericInput,
    "expression 3": expression,
  },
}
```

packages/perseus/src/util/extract-widget-ids-integration.test.ts
```diff
@@ -0,0 +1,169 @@
+import {renderQuestion} from "../widgets/__testutils__/renderQuestion";
+
+import {extractWidgetIds} from "./extract-widget-ids";
+
+import type {PerseusRenderer} from "@khanacademy/perseus-core";
+
+// Integration test to compare our extracted function against the actual renderer behavior
```
benchristel: This is a cool idea! I wouldn't call it an "integration" test exactly, though. I think it is more like a contract test.

In contract testing, you have a specification of behavior (the contract) that you want multiple implementers to adhere to. In this case, the implementers of our contract are the Renderer component and the `extractWidgetIds` function.

One way to do contract testing is to have a function that takes an implementer and generates a suite of tests for that implementer. E.g.

```ts
function testWidgetIdExtraction(
  name: string,
  runExtractor: (r: PerseusRenderer) => string[],
) {
  describe(`extracting widget IDs with ${name}`, () => {
    it("returns the ID of a single widget", () => {
      const question: PerseusRenderer = {
        content: "Here's a widget: [[☃ radio 1]]",
        widgets: {
          "radio 1": { /* ... */ },
        },
      };

      const widgetIds = runExtractor(question);
      expect(widgetIds).toEqual(["radio 1"]);
    });
  });
}
```

Then you pass each implementation to the function in your test file:

```ts
testWidgetIdExtraction(
  "the Renderer component",
  (question: PerseusRenderer) => {
    return renderer.getWidgetIds();
  },
);

testWidgetIdExtraction(
  "the extractWidgetIds function",
  (question: PerseusRenderer) => {
    return extractWidgetIds(question);
  },
);
```

This makes it really easy to see that both implementations pass exactly the same tests.

packages/perseus/src/util/extract-widget-ids.ts
Myranae: Updated to use the traverseContent function. Can you double check I implemented it correctly?

packages/perseus/src/util/extract-widget-ids.ts
Myranae: Removed

packages/perseus/src/util/extract-widget-ids.test.ts
Myranae: Extracted data to a const for each widget

packages/perseus/src/util/extract-widget-ids-integration.test.ts
Myranae: Formatted tests to be more like a contract test and added a comment

Ooo, so do we only need the contract test file then? Hmm, maybe I just need to move the 4 tests that are not duplicated into the contract and then I can get rid of that file.

packages/perseus/src/util/extract-widget-ids.ts
```diff
@@ -0,0 +1,38 @@
...
+        PerseusMarkdown.traverseContent(ast, (node) => {
+            if (node.type === "widget") {
+                if (!widgetIds.includes(node.id)) {
+                    widgetIds.push(node.id);
+                }
+            }
+        });
+    }
+
+    collectWidgetIds(parsedMarkdown);
```
jeremywiebe: Non-blocking, but I can't see why we need an internal function to wrap `traverseContent(). 

Couldn't we just do this? 

```suggestion
    PerseusMarkdown.traverseContent(ast, (node) => {
        if (node.type === "widget") {
             if (!widgetIds.includes(node.id)) {
                 widgetIds.push(node.id);
             }
        }
    });
```

packages/perseus/src/util/extract-widget-ids.test.ts
```diff
@@ -0,0 +1,186 @@
...
+    it("handles content with no widgets", () => {
+        const content = "This is just plain text with no widgets.";
+        const question = {
+            content,
+            widgets: {},
+            images: {},
+        };
+        const result = extractWidgetIds(question);
+        expect(result).toEqual([]);
+    });
```
benchristel: These tests are duplicated in the contract file, so they can be removed now :)

packages/perseus/src/util/extract-widget-ids.test.ts
```diff
@@ -0,0 +1,186 @@
...
+                "radio 1": radio,
+                "numeric-input 2": numericInput,
+                "expression 3": expression,
+            },
+            images: {},
+        };
+        const result = extractWidgetIds(question);
+        // Should maintain order of first occurrence
+        expect(result).toEqual(["radio 1", "numeric-input 2", "expression 3"]);
+    });
```
benchristel: Is there a reason these tests aren't in the contract? This seems like behavior that both implementations should support.

packages/perseus/src/util/extract-widget-ids.ts
Myranae: Got rid of the wrapper and passed in parsedMarkdown to traverseContent.

packages/perseus/src/util/extract-widget-ids.test.ts
Myranae: Removed

packages/perseus/src/util/extract-widget-ids.test.ts
Myranae: Moved to the contract :)

PR #2833, opened 2025-08-26
fix(LEMS-3424): change css to fix orderer layout

anakaren-rojas:
## Summary:
Orderer has started to wrap. These CSS changes make it so it doesn't wrap
- Added css fixes so cards don't wrap to new row (restore to previous behavior)
- Added story for visual regression testing
- Added additional test data to support story

Issue: LEMS-3424

## Test plan:
1. Navigate to Orderer widget storybook 
2. Add _lots_ of cards to the orderer 
3. Fixed behavior: cards don't wrap into new row, allow you to put cards at the end of a single row

### Screen recording

https://github.com/user-attachments/assets/fe7f8794-414e-4be7-bd53-a6abbe398f97

### DIFF
--- a/packages/perseus/src/styles/widgets/orderer.css
+++ b/packages/perseus/src/styles/widgets/orderer.css
@@ -1,22 +1,24 @@
 .orderer {
     position: relative;
-    /* Fix for wrapping issue caused by CSS reset removal */
+    /* Start fix for wrapping issue caused by CSS reset removal */
     box-sizing: border-box;
     margin: 0;
     padding: 0;
     width: 100%;
+    /* End Fix */
 }
 .orderer.layout-horizontal .draggable-box {
     margin-left: 0;
     margin-top: 30px;
     padding: 13px;
-    /* Fix for wrapping issue caused by CSS reset removal */
+    /* Start fix for wrapping issue caused by CSS reset removal */
     box-sizing: border-box;
     width: 100%;
     display: flex;
     flex-wrap: nowrap;
     overflow-x: auto;
     white-space: nowrap;
+    /* End Fix */
 }
 .orderer .card {
     padding: 0 10px;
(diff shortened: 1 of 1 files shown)


Myranae (commented):
How does the scrolling work on mobile? I tried testing with inspect and looking at it in phone mode, but had a hard time finding the right part of the box to click to scroll. Could it have not scrolled before on purpose for the mobile experience?

ivyolamit:
> How does the scrolling work on mobile? I tried testing with inspect and looking at it in phone mode, but had a hard time finding the right part of the box to click to scroll. Could it have not scrolled before on purpose for the mobile experience?

This might be a good candidate to use the ScrollableView component like what we have in radio, for future UX improvement.
http://localhost:6006/?path=/story/widgets-radionew-widget-demo--single-select-with-scroll

image

anakaren-rojas:
> How does the scrolling work on mobile? I tried testing with inspect and looking at it in phone mode, but had a hard time finding the right part of the box to click to scroll. Could it have not scrolled before on purpose for the mobile experience?

Its seems to @Myranae 

https://github.com/user-attachments/assets/445ccf84-ba4a-4b45-96fa-99660f1d70d8

packages/perseus/src/styles/widgets/orderer.css
```diff
@@ -1,10 +1,22 @@
 .orderer {
     position: relative;
+    /* Fix for wrapping issue caused by CSS reset removal */
+    box-sizing: border-box;
+    margin: 0;
+    padding: 0;
+    width: 100%;
```
handeyeco: It's hard to know what's for the fix and what isn't. Would it be useful to mark the end?

```suggestion
    /* Start fix for wrapping issue caused by CSS reset removal */
    box-sizing: border-box;
    margin: 0;
    padding: 0;
    width: 100%;
    /* End fix */
```

packages/perseus/src/styles/widgets/orderer.css
```diff
@@ -33,18 +45,28 @@
...
+    padding: 0;
 }
 .orderer.layout-horizontal .card-wrap {
-    float: left;
-}
-.orderer.layout-horizontal .card-wrap:not(:first-child) {
-    margin-left: 8px;
+    flex-shrink: 0;
+    display: inline-block;
+    margin: 4px; /* Uniform spacing around each card */
```
handeyeco, line 73: This kind of seems like a regression. Originally the first element wouldn't have left-side padding and now it will have 4px right?

handeyeco (approved):

packages/perseus/src/styles/widgets/orderer.css
anakaren-rojas: Some of the other css changes removed the spacing around the cards, so without this, they're up in each other's business 
Without:
Screenshot 2025-09-04 at 11 09 24
With:
image

PR #2834, opened 2025-08-26
[functionally-equivalent] Updating (New) Radio widget to be fully functional. 

SonicScrewdriver:
## Summary:
This quick PR converts our new Multiple Choice Widget to be a fully functional component with a `useImperative` handle for returning `getPromptJSON`. 

When we delete the old radio files + feature flag, we will probably need to move other functions into the `useImperative` handle (such as `getSerializedState` or `focus`). However, both of these methods are deprecated, so we should look into it when we're cleaning up the project. 

Issue: LEMS-3446

## Test plan:
- Tests pass in Perseus
- Tests pass in our Frontend 
- Manual testing in Storybook and in our Frontend

### DIFF
--- a/packages/perseus/src/widgets/radio/multiple-choice-widget.new.tsx
+++ b/packages/perseus/src/widgets/radio/multiple-choice-widget.new.tsx
@@ -1,5 +1,5 @@
-import {linterContextDefault} from "@khanacademy/perseus-linter";
 import * as React from "react";
+import {forwardRef, useImperativeHandle} from "react";
 
 import Renderer from "../../renderer";
 import {mockStrings} from "../../strings";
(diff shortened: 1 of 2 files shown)


packages/perseus/src/widgets/radio/multiple-choice-widget.new.tsx
```diff
@@ -71,278 +71,258 @@ type Props = WidgetProps<
...
 
-        // Extract props that need to be passed to the component
-        const {apiOptions, multipleSelect, countChoices, numCorrect} =
-            this.props;
-        const reviewMode =
-            this.props.reviewMode || this.props.showSolutions === "all";
+        // This is strange, but currently we're showing the same view for both reviewMode and
+        // showSolutions === "all". We may wish to differentiate between the two in the future,
+        // depending on the Perseus GUTC project.
+        const isReviewMode = reviewMode || showSolutions === "all";
```
SonicScrewdriver: I wanted to call out the strange way we're handling this now, in case we find ourselves confused again in the future. That being said, neither of these variables are particularly clear about how they _should_ be treated differently. I opted not to dig into this too much now, as we'll probably have larger discussions after some progress on GUTC.

packages/perseus/src/widgets/radio/multiple-choice-widget.new.tsx
```diff
@@ -71,278 +71,258 @@ type Props = WidgetProps<
...
  *
  * Created as part of the Radio Revitalization Project (LEMS-2933).
  */
-class MultipleChoiceWidget extends React.Component<Props> implements Widget {
-    static defaultProps: DefaultProps = {
-        choices: [],
-        multipleSelect: false,
-        countChoices: false,
-        deselectEnabled: false,
-        linterContext: linterContextDefault,
```
SonicScrewdriver, line 104: We were setting a default for linterContext...even though we're hardcoding the only place we're using it. (Line 162) Given that, I've removed the default. 

We also aren't using `deselectEnabled` anywhere, so I've removed the default setting for that too. We'll want to _fully_ delete `deselectEnabled` in a separate PR as that results in changes to the schema. My content searches indicate that we still have a bunch of content that sets this value to false, even though we're not using it. :P

packages/perseus/src/widgets/radio/multiple-choice-widget.new.tsx
```diff
@@ -71,278 +71,258 @@ type Props = WidgetProps<
...
- * It handles Perseus-specific logic and state management while delegating
+ * It handles Perseus-specific logic and user state management while delegating
  * UI rendering to the MultipleChoiceComponent.
  *
- * This class is exported as "Radio" for backwards compatibility with existing content,
+ * This component is exported as "Radio" for backwards compatibility with existing content,
  * though it supports both radio-button (single select) and checkbox (multiple select) modes.
  * Eventually, when Content Backfills are set up, we will officially rename the Radio Widget to MultipleChoiceWidget.
  *
  * Created as part of the Radio Revitalization Project (LEMS-2933).
```
ivyolamit, line 84: Do we still need this comment?

packages/perseus/src/widgets/radio/multiple-choice-widget.new.tsx
```diff
@@ -71,278 +71,258 @@ type Props = WidgetProps<
...
-                previouslyAnswered,
+        /**
+         * Renders content that may contain nested widgets (currently only passage-refs).
+         * Parses the content, extracts any widgets, and returns a Renderer component
+         * configured with the appropriate context.
+         *
+         * @param content - The content to render (defaults to empty string)
+         * @returns A React element with the rendered content
+         */
+        const renderContent = (content = ""): React.ReactNode => {
```
ivyolamit, line 135: these newly converted const in this PR, do we want some of these to be in util functions? renderContent, buildChoiceProps, prepareChoicesProps

ivyolamit (approved):
@SonicScrewdriver great work here. Changes looks logical to me. Smoke test looks good as well. Just few non-blocking questions.

packages/perseus/src/widgets/radio/multiple-choice-widget.new.tsx
SonicScrewdriver: We do! We're stuck with this until we can do a backfill to replace the widget name... and I expect this will be the LOWEST priority back fill. :P 

We COULD do a half step where we register this widget under both names, but I've seen enough half-steps to figure we should just wait until we can do it in one fell swoop. :)

packages/perseus/src/widgets/radio/multiple-choice-widget.new.tsx
SonicScrewdriver: We could maybe move the `buildChoiceProps` / `prepareChoicesProps`, but I wasn't sure it was really worth it any more after vastly simplifying the file down. 

- Both those functions require quite a few properties from `this.props` and are essentially managing the local state, and the functions aren't useful anywhere else. 
- `prepareChoicesProps` is already just a wrapper around a pre-existing util function.
- `renderContent` requires being in a JSX file.

I do think we'll have more abilities to simplify some of this logic after we delete the old Radio Widget, though, as then we can simplify the ChoiceState.

packages/perseus/src/widgets/radio/multiple-choice-widget.new.tsx
```diff
@@ -71,280 +71,259 @@ type Props = WidgetProps<
...
+        /**
+         * Handles a user's selection of a choice.
+         *
+         * Updates choice states based on the selection, handles single/multiple
+         * selection logic, and notifies Perseus of the interaction.
+         *
+         * @param choiceIndex - The index of the choice that changed
+         * @param newCheckedState - Whether the choice is now selected
+         */
+        const handleChoiceChange = (
```
anakaren-rojas, line 177: Maybe the same answer as Ivy's question above, but can some of these be moved out into hooks to simplify the component? something like 
```
function useChoiceManagement(props) {
    const handleChoiceChange = useCallback((choiceIndex, newCheckedState) => {
        // ... existing logic
    }, [props.choices, props.multipleSelect, props.onChange, props.trackInteraction]);

    const prepareChoicesProps = useCallback(() => {
        // ... existing logic  
    }, [props.choices, props.showSolutions, props.choiceStates, props.values]);

    return { handleChoiceChange, prepareChoicesProps };
}
```

anakaren-rojas: Or maybe memoizing so they're not recreated on every render?

packages/perseus/src/widgets/radio/multiple-choice-widget.new.tsx
```diff
@@ -71,280 +71,259 @@ type Props = WidgetProps<
...
+            // We pass in a key here so that we avoid a semi-spurious
+            // react warning when the ChoiceNoneAbove renders a
+            // different renderer in the same place. Note this destroys
+            // state, but since all we're doing is outputting
+            // "None of the above", that is okay. Widgets inside this Renderer
+            // are not discoverable through the parent Renderer's `findWidgets` function.
+            // alwaysUpdate={true} so that passage-refs findWidgets
+            // get called when the outer passage updates the renderer
+            const linterContext: LinterContextProps = {
+                contentType: "radio",
```
anakaren-rojas, line 149: to preserve radio as a literal type vs as a string
```suggestion
                contentType: "radio" as const,
```

anakaren-rojas (approved):
smh idk why i didn't approve earlier, my questions were non-blocking

packages/perseus/src/widgets/radio/multiple-choice-widget.new.tsx
SonicScrewdriver: Ooo not a bad plan either. I'll make a new ticket and look into it!

PR #2836, opened 2025-08-26
Add documentation warning for using Graphie component

ivyolamit:
## Summary:
Add documentation warning for using Graphie component
| Before | After |
| ------- | ---- |
| image | image |

### DIFF
--- a/packages/perseus/src/components/__docs__/graphie.stories.tsx
+++ b/packages/perseus/src/components/__docs__/graphie.stories.tsx
@@ -4,6 +4,8 @@ import {ServerItemRendererWithDebugUI} from "../../../../../testing/server-item-
 import {itemWithPieChart} from "../../__testdata__/graphie.testdata";
 import Graphie from "../graphie";
 
+import GraphieDocsPage from "./graphie.mdx";
+
 import type {StoryObj, Meta} from "@storybook/react-vite";
 
 type Story = StoryObj<typeof Graphie>;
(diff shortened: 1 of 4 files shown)


types/assets.d.ts
```diff
@@ -7,6 +7,7 @@ declare module "*.jpg";
 declare module "*.png";
 declare module "*.svg";
 declare module "*.css";
+declare module "*.mdx";
```
ivyolamit, line 10: currently looks like TS doesn't know how to handle .mdx imports, this will fix the TS error when importing `import GraphieDocsPage from "./graphie.mdx";`
```
TS2307: Cannot find module ./graphie.mdx or its corresponding type declarations.
```

packages/perseus/src/components/__docs__/graphie.stories.tsx
```diff
@@ -13,20 +15,31 @@ const size = 200;
...
  * A demonstration of a Graphie rendered using the Perseus `Renderer` complete
  * with overlaid labels and an image caption below.
  */
-export const PieChartGraphieLabels = () => {
-    return <ServerItemRendererWithDebugUI item={itemWithPieChart} />;
+export const PieChartGraphieLabels: Story = {
+    args: {
+        box: [size, size],
+        setup: () => {},
+        setDrawingAreaAvailable: () => {},
```
Myranae, line 34: Is having the args in the story better practice than having them in the meta? Just curious why they were brought down to the stories instead of kept in the meta object

jeremywiebe: I think we can do this as @Myranae suggests, moving these "default" args up to the `meta`. 

Thanks for converting these to CSF v3 format though!

PR #2838, opened 2025-08-26
[Radio] Bugfix - Correctness indicator was missing in static widget

mark-fitzgerald:
## Summary:
When the widget is in "static" mode, it should show the correctness of each choice, just like in "reviewMode".

Issue: LEMS-3356

### Before
Before

### After
After

## Test plan:
1. Open Storybook (either locally or linked in this PR)
1. Navigate to the static examples:
   - Widgets => RadioNew => Visual Regression Tests => Static => Single Select Static
   - Widgets => RadioNew => Visual Regression Tests => Static => Multi Select Static
1. Note that correctness is shown for the options

### DIFF
--- a/.changeset/gorgeous-eyes-cheat.md
+++ b/.changeset/gorgeous-eyes-cheat.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+[Radio] Bugfix - Correctness indicator was missing in static widget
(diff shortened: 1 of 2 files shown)


SonicScrewdriver (approved):
Makes sense - thank you very much!

I wonder if we should chat with design about these 3 confusing properties, so that we can make a proposal for how to clarify them when we DO work on GUTC.

PR #2840, opened 2025-08-27
[Radio] Fix mobile hover issue when use is dragging the screen left or right

ivyolamit:
## Summary:
[LEMS-3406/fix-mobile-hover] Fix mobile hover issue when use is dragging the screen left or right

Issue: LEMS-3406

## Test plan:


ivyolamit:
Will recreate PR with the correct JIRA ticket number

PR #2841, opened 2025-08-27
Remove some serialized state APIs

handeyeco:
## Summary:
This removes the `serializedState` prop on Renderer and the `restoreSerializedState` API.

- The `serializedState` prop was used by TX. That usage was removed in [Frontend #3107](https://github.com/Khan/frontend/pull/3107) ([see this Slack thread](https://khanacademy.slack.com/archives/C3X1EG1J5/p1755643606843629))
- `restoreSerializedState` was used by TUT for an internal cheating tool. That usage was removed in [Frontend #3272](https://github.com/Khan/frontend/pull/3272) ([see this Slack thread](https://khanacademy.slack.com/archives/C06NBBNCV8B/p1756237089280909))

The ability to restore serialized state banks on the existence of RenderProps, which I'm actively trying to remove (see [LEMS-3199](https://khanacademy.atlassian.net/browse/LEMS-3199)). These APIs were already deprecated as they're considered dangerous, but my hand was forced while working on Answerless->Answerful because RenderProps essentially caches widget state/props which kept resulting in stale state when transitioning between Answerless/Answerful.

Part of: [LEMS-3185](https://khanacademy.atlassian.net/browse/LEMS-3185)

[LEMS-3199]: https://khanacademy.atlassian.net/browse/LEMS-3199?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/perseus/src/widgets/interactive-graphs/interactive-graph.test.tsx
```diff
@@ -115,7 +115,7 @@ describe("Interactive Graph", function () {
             correct: ReadonlyArray<Coord>,
             incorrect: ReadonlyArray<Coord>,
         ) => {
-            it("Should accept the right answer", async () => {
+            it.skip("Should accept the right answer", async () => {
```
handeyeco: IG tests are using serialized state as a way to inject user input, but this needs to change. Either:

1. (Preferred) we should use RTL to imitate the behavior of a user interacting with IG
2. We can add a way to pass down user input (this likely exists already if these tests are using Renderer)

benchristel: I think the problem might be that the StatefulMafsGraph has its own internal user input state. That's my fault — I didn't know that the widgets were supposed to store state by handing it off to the Renderer.

I agree that it would be nice to use either of the approaches you mentioned.

packages/perseus/src/user-input-manager.test.tsx
```diff
@@ -301,61 +268,109 @@ describe("UserInputManager", () => {
...
+                            </p>
+                        </>
+                    );
+                }}
+            </UserInputManager>,
+        );
+
+        expect(
+            screen.getByText("User input is: 0-0-0-0-0"),
+        ).toBeInTheDocument();
```
benchristel: Rather than use a `screen` query, we could pass a spy function to `UserInputManager`:

```ts
const spy = jest.fn()
render(
  <UserInputManager widgets={widgets} problemNum={0}>
    {({userInput}) => spy(userInput)}
  </UserInputManager>,
);
expect(spy).toHaveBeenCalledWith({/* ... */});
```

Or even use a variable:

```ts
let renderedUserInput;
render(
  <UserInputManager widgets={widgets} problemNum={0}>
    {({userInput}) => {renderedUserInput = userInput}}
  </UserInputManager>,
);
expect(renderedUserInput).toEqual({/* ... */});
```

packages/perseus/src/user-input-manager.tsx
```diff
@@ -135,26 +127,16 @@ export default function UserInputManager(props: Props) {
...
-
-    /**
-     * Restore UserInput from SerializedState (please do not add new uses of this)
-     */
-    function restoreUserInputFromSerializedState(
-        serializedState: unknown,
-        widgetOptions: PerseusWidgetsMap,
     ) {
         setUserInput(
-            deriveUserInputFromSerializedState(serializedState, widgetOptions),
```
benchristel, line 150: Can `deriveUserInputFromSerializedState` be deleted now?

packages/perseus/src/user-input-manager.test.tsx
```diff
@@ -301,61 +268,109 @@ describe("UserInputManager", () => {
         expect(screen.getByText("User input is: -4")).toBeInTheDocument();
     });
 
-    it("restores serialized State", async () => {
-        let restoreCallback:
-            | RestoreUserInputFromSerializedStateCallback
-            | undefined;
+    it("initializes static user input", async () => {
```
benchristel, line 271: I'm confused by this test. It looks like the `UserInputManager` is arbitrarily picking the first choice to be selected. Why do we want that?

packages/perseus/src/widgets/__testutils__/renderQuestion.tsx
```diff
@@ -94,30 +94,28 @@ const RendererWrapper = React.forwardRef<
     }
 >(function RendererWithDependencies(props, ref) {
     const dependencies = useDependencies();
+    const {userInput: initialUserInput, ...restExtraProps} =
```
benchristel: This seems weird: `extraProps` is `PropsFor<typeof Perseus.Renderer>`, yet we're passing its `userInput` to the `UserInputManager` as a prop, not to the `Renderer`. It seems like maybe the type of `extraProps` should list the actual props that are expected. Maybe `extraProps` shouldn't be its own object at all, and everything should just be a top-level field on `props`.

packages/perseus/src/user-input-manager.tsx
handeyeco: Nah, we're using it as a (hopefully) temp hack as a replacement for `restoreSerializedState`

packages/perseus/src/user-input-manager.test.tsx
handeyeco: It's picking the correct answer. A static widget is a read-only widget with the user input set to the correct answer.

PR #2846, opened 2025-08-27
Add regression tests for user input parsers

benchristel:
We want to make sure the user input format used by widgets does not change
unless we also update the user input parsers to migrate old formats to the
latest. This is because AX is going to pin their version of Perseus for
their upcoming pilot, so if we change the user input format, AX will still
be sending the old format to the Perseus service for scoring.

In other words, we need to keep our user input parsers backward-compatible
with the current format.

To that end, this PR adds data-driven regression tests for the user input
parsers, similar to the ones we have for PerseusItem and PerseusArticle.
The regression tests feed the current user input format to the parsers
and fail if the format can’t be parsed.

Issue: https://khanacademy.atlassian.net/browse/LEMS-3448

## Test plan:

`pnpm test`

### DIFF
--- a/.changeset/large-jeans-dress.md
+++ b/.changeset/large-jeans-dress.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-core": patch
+---
+
+Internal: add data-driven regression tests for user input parsers


packages/perseus-core/src/parse-perseus-json/regression-tests/parse-perseus-json-regression.test.ts
```diff
@@ -144,6 +145,47 @@ describe("parseAndMigratePerseusArticle", () => {
...
+            // failure easier to read, since otherwise the entire `invalidObject`
+            // from the ParseFailureDetail would be printed.
+            const resultWithMessage = mapFailure(getMessage)(result);
+
+            expect(resultWithMessage).toEqual(anySuccess);
+        });
+
+        it("returns the same result as before", () => {
+            assertSuccess(result);
+            expect(result.value).toMatchSnapshot();
```
handeyeco, line 168: We talked about this in person, but just for the paper trail: I personally think it would benefit us to move away from these kinds of raw JSON / snapshot tests. It's hard to immediately understand the test (inputs/outputs/expectations) and they're really easy to auto-update without understanding the implications of the changes.

I think Ben landed on keeping these as-is while considering a refactor away from these kinds of tests. I'm also fine to have these exist as-is, so I'll approve regardless.

handeyeco (approved):

packages/perseus-core/src/parse-perseus-json/regression-tests/parse-perseus-json-regression.test.ts
benchristel: Here is the ticket for replacing: https://khanacademy.atlassian.net/browse/LEMS-3473

PR #2847, opened 2025-08-27
[Image] | (CX) | Add the Title field back in the Image editor

nishasy:
## Summary:
Adding the `title` field back in the Image editor.

It got removed in 2017 without any reason as far as I can tell, but
the `title` prop is still used in a number of images.

It seems like something that could be valuable, so we decided to bring it back.

Issue: https://khanacademy.atlassian.net/browse/LEMS-3427

## Test plan:

Storybook
`/?path=/docs/widgets-image-editor-demo--docs#populated-within-editor-page`

### DIFF
--- a/packages/perseus-editor/src/widgets/image-editor/image-settings.tsx
+++ b/packages/perseus-editor/src/widgets/image-editor/image-settings.tsx
@@ -15,10 +15,12 @@ export default function ImageSettings({
     alt,
     backgroundImage,
     caption,
+    title,
     onChange,
 }: Props) {
     const uniqueId = React.useId();
     const altId = `${uniqueId}-alt`;
+    const titleId = `${uniqueId}-title`;
     const captionId = `${uniqueId}-caption`;
 
     if (!backgroundImage.url) {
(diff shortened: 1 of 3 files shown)


packages/perseus-editor/src/widgets/image-editor/image-settings.tsx
```diff
@@ -15,10 +15,12 @@ export default function ImageSettings({
     alt,
     backgroundImage,
     caption,
+    title,
```
ivyolamit, line 18: Question: Are we planning to fix this comment in this project?
```
```

link: packages/perseus-core/src/data-schema.ts > PerseusImageWidgetOptions

packages/perseus-editor/src/widgets/image-editor/image-settings.tsx
```diff
@@ -110,13 +121,15 @@ export default function ImageSettings({
...
-                    marginBlockStart: sizing.size_040,
-                    marginBlockEnd: sizing.size_080,
-                }}
+                style={textAreaStyle}
             />
         </>
     );
 }
+
+// TODO: Use CSS modules after Wonder Blocks styles
```
ivyolamit, line 130: 👍🏼

packages/perseus-editor/src/widgets/image-editor/image-settings.tsx
nishasy: I'm actually not even sure why that's there or what it means. I'm going to ask around.

nishasy: Even if we can remove those comments, I'm probably not going to do that in this PR, since those comments are on a ton of different fields that are actually being used across many widgets, even `caption` and `alt` within Image widget.

PR #2852, opened 2025-08-28
[Image] | (DX) | Revert to old way of saving empty string data for the sake of data consistency

nishasy:
## Summary:
It seems like saving empty strings as undefined might introduce inconsistency in our data.
Since these have been saved as `""` for years, I'm going back to doing it that way so
it's easier to understand our data in the future.

[Discussion](https://khanacademy.slack.com/archives/C01AZ9H8TTQ/p1756321954129759)

Issue: none

## Test plan:

### DIFF
--- a/packages/perseus-editor/src/widgets/__tests__/image-editor.test.tsx
+++ b/packages/perseus-editor/src/widgets/__tests__/image-editor.test.tsx
@@ -176,8 +176,12 @@ describe("image editor", () => {
 
         // Assert
         expect(onChangeMock).toHaveBeenCalledWith({
-            backgroundImage: {},
-            box: undefined,
+            backgroundImage: {
+                url: "",
+                width: 0,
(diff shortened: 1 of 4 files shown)


ivyolamit (approved):

packages/perseus-editor/src/widgets/__tests__/image-editor.test.tsx
```diff
@@ -176,8 +176,12 @@ describe("image editor", () => {
 
         // Assert
         expect(onChangeMock).toHaveBeenCalledWith({
-            backgroundImage: {},
-            box: undefined,
+            backgroundImage: {
+                url: "",
```
ivyolamit, line 180: base on our type

```
    // The URL of the image
    url?: string | null;
```
can we please add a scenario when url is null

PR #2853, opened 2025-08-28
[Image] | (DX) | Regression stories that include mobile and "within article" cases

nishasy:
## Summary:
The mobile stories and "within article" stories will help us see the image widget
with more contexts.

Screenshot 2025-08-28 at 4 09 13 PM
Screenshot 2025-08-28 at 4 09 10 PM
Screenshot 2025-08-28 at 4 09 17 PM
Screenshot 2025-08-28 at 4 09 19 PM

Issue: none

## Test plan:
Check the UI tests below


packages/perseus/src/widgets/image/__docs__/image-regression.stories.tsx
```diff
@@ -46,6 +50,12 @@ const rendererDecorator = (_, {args}) => {
     );
 };
 
+const mobileDecorator = (Story) => (
+    <div className="framework-perseus perseus-mobile">
```
ivyolamit, line 54: just a thought right, do we also need to put the normal case inside the `framework-perseus` className?
```
<div className="framework-perseus">
```

from what I remember we have some stylings that would check for `framework-perseus`

ivyolamit (approved):
@nishasy changes looks good to me. I just have one non-blocking question.

packages/perseus/src/widgets/image/__docs__/image-regression.stories.tsx
nishasy: Oops. Turns out the stories already have `framework-perseus`, and I just re-added it for mobile. I'll try to figure out how to put the `perseus-mobile` style in the right place.

image

nishasy: Ahhh it's part of the Storybook `preview.tsx`. I think that means that works for the non-mobile cases, but I think it's fine to have it be there twice for the mobile ones (especially since I can't find a way to override the global decorator anyway). I was just worried that was there as part of the widget and not storybook.

PR #2857, opened 2025-09-02
Remove Radio transforms (take 2)

handeyeco:
## Summary:
This is a modified copy of this PR: https://github.com/Khan/perseus/pull/2800

I had #2800 ready to go, but there was an urgent need to land #2799 quickly. Unfortunately this caused a lot of merge conflicts but also some shifts around the logic I was working on. I just decided to redo #2800 with a version that worked with the new user input format.

Issue: LEMS-3199

## Test plan:
Nothing should change from the learner's perspective. Unit tests added to make sure there were no major regressions we're trying to support.

---

From the original PR:

> Why this PR exists
>
> We're removing transform and staticTransform to support Answerless/Answerful rendering for the Server Side Scoring project. Now widgets will receive all of their widget options as props.
How this affects Radio
>
> For the most part, Radio's transform was mostly just filtering/passing props. However it also called choiceTransform which does a couple of things:
>
>    It makes sure "None of the above" options were at the end of the choices
>    It enforces correct order of "Yes"/"No" and "True"/"False" questions
>    It shuffles (if randomize is true)
>
> 1 & 2 are deterministic: we can call them over and over again and the output will always be the same. Shuffling was using Math.random though which requires stashing the output of transform because repeat calls would result in different shuffle orders. I believe that we do not want true randomness though:
>
>    In the Teacher Experience we want to be able to show the exact order the learner saw
>    In the future we'll need to be able to sync Khanmigo with the UI and I think we need to do that between the FE and BE
>
> ANYWAY, I made shuffling deterministic. Anakaren told me about the requirement of [LEMS-297](https://khanacademy.atlassian.net/browse/LEMS-297) which means we couldn't just use problemNum as the seed, so I wrote a regression test and made sure to continue support the [LEMS-297](https://khanacademy.atlassian.net/browse/LEMS-297) requirement.

[LEMS-297]: https://khanacademy.atlassian.net/browse/LEMS-297?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/perseus/src/renderer.tsx
+++ b/packages/perseus/src/renderer.tsx
@@ -594,6 +594,24 @@ class Renderer
         return null;
     };
 
+    _getWidgetIndexById(id: string) {
+        const widgetIndex = Object.keys(this.props.widgets).indexOf(id);
+        if (widgetIndex < 0) {
+            Log.error(
+                "Unable to get widget index in _getWidgetIndexById",
+                Errors.Internal,
+                {
+                    loggedMetadata: {
+                        widgets: JSON.stringify(this.props.widgets),
+                        widgetId: JSON.stringify(id),
+                    },
+                },
+            );
+            return 0;
+        }
+        return widgetIndex;
+    }
+
     getWidgetProps(
         widgetId: string,
     ): WidgetProps<any, any, PerseusWidgetOptions> {
@@ -627,6 +645,7 @@ class Renderer
             ...widgetProps,
             userInput: this.props.userInput?.[widgetId],
             widgetId: widgetId,
+            widgetIndex: this._getWidgetIndexById(widgetId),
             alignment: widgetInfo && widgetInfo.alignment,
             static: widgetInfo?.static,
             problemNum: this.props.problemNum,
--- a/packages/perseus/src/renderer.tsx
+++ b/packages/perseus/src/renderer.tsx
@@ -595,7 +595,7 @@ class Renderer
     };
 
     _getWidgetIndexById(id: string) {
-        const widgetIndex = Object.keys(this.props.widgets).indexOf(id);
+        const widgetIndex = this.widgetIds.indexOf(id);
         if (widgetIndex < 0) {
             Log.error(
                 "Unable to get widget index in _getWidgetIndexById",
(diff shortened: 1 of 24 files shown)


packages/perseus/src/widget-ai-utils/radio/radio-ai-utils.test.ts
```diff
@@ -145,44 +129,41 @@ describe("Radio AI utils", () => {
...
+                expect(uiText.includes(promptText)).toBe(true);
+            });
+
+            // Ensure the user input contains the correct ID
+            expect(widgetJSON.userInput.selectedOptions).toEqual([choice.id]);
+        },
+    );
 
     // regression (TUT-2738) TODO: we shouldn't have to handle this
     // (user input should be initialized already)
```
ivyolamit, line 169: Please remove `selectedChoices` in line 216/197

packages/perseus/src/widgets/passage/__tests__/passage.test.tsx
```diff
@@ -51,10 +51,12 @@ function renderPassage(
         static: true,
         trackInteraction: () => {},
         widgetId: "passage",
+        widgetIndex: 0,
         reviewMode: false,
     } as const;
 
     const extended = {...base, ...overwrite} as const;
+    // TODO: use a Renderer wrapper rather than rendering this directly
```
ivyolamit, line 59: will this be worked on follow-up PR or some future work?

handeyeco: I consider it beyond the scope of my current work but something I noticed while working.

packages/perseus/src/renderer.tsx
```diff
@@ -594,6 +594,24 @@ class Renderer
         return null;
     };
 
+    _getWidgetIndexById(id: string) {
+        const widgetIndex = Object.keys(this.props.widgets).indexOf(id);
```
benchristel: The order of the keys returned by `Object.keys` is the order in which the keys were originally inserted into the object. Is that actually what you want here? I wouldn't want to depend on the key order being the same across renders.

I'd much rather use the order in which the widgets appear in the markdown.

benchristel (approved):
I left one comment inline about `Object.keys` not necessarily being a safe way to derive the widget index. I don't think that needs to be blocking if you know otherwise, though. Everything else LGTM!

packages/perseus/src/renderer.tsx
handeyeco: I switched to using `this.widgetIds` which seems to be the used IDs from the parsing process.

PR #2859, opened 2025-09-03
[Radio] Fix iOS mobile hover issue when use is dragging the screen left or right

ivyolamit:
## Summary:
Fix iOS mobile hover issue when use is dragging the screen left or right

Issue: LEMS-3467

## Test plan:


packages/perseus/src/widgets/radio/choice.module.css
```diff
@@ -107,14 +114,44 @@
...
+        --perseus-multiple-choice-divider-color: var(
+            --wb-semanticColor-core-border-neutral-default
+        );
+    }
+}
+
+/*********************
+ ACTIVELY SCROLLING
+ *********************/
+
```
ivyolamit: Styling below is special handling so we can have better visual feedback in mobile. Mobile do not have a hover gesture but current issue is iOS shows the hover state and Android doesn't. The code considers:
- when user is actively scrolling we don't show the dividers
- active state to properly apply the correct styling with dividers, since mobile do not have the hover state

ivyolamit:
> Looks good. Nice simple trigger for devices that support hover.
> 
> I wonder if it might be better to consolidate the `@media (hover: hover)` definitions into a single section in each file, and add a comment to both explaining that these are broken out in order to provide a more native experience for touch-only devices.

I was debating to do that or now 😅 but decided to put it where it's close to its related styling. But I think adding the comment like you suggested make sense to me and much more cleaner approach. I'll update that.

update: on the second thought, after re-reading the code again, grouping it where the other styling is make sense so it will be easier for reading the code/debugging.

PR #2862, opened 2025-09-04
Support new division symbol (:) used by several locales

SonicScrewdriver:
## Summary:
This PR adds support for locales (particularly Ukraine) that use `:` instead of `/` for denoting division in math equations. We're only focusing on the Expression Widget as Numeric Input / Input Number do not support the division symbol. 

We're currently checking with the language advocates about how fractions are written in Ukrainian, but initial research is indicating that the bar `/` is still used in those cases. 

This PR does not update the actual icon on the keypad, as we're still waiting for design. We will update the keypad button (in the appropriate locales) in a future PR, after LEMS-3535 has been completed. 

## Video Example: 
https://github.com/user-attachments/assets/dd358910-ec21-4e76-8cc1-4405f59b5002

Issue: LEMS-3435

## Test plan:
- New tests pass 
- Manual testing using my [draft locale dropdown feature](https://github.com/Khan/perseus/pull/2861) 
- No regressions noticed in existing tests

### DIFF
--- a/packages/perseus-core/src/utils/get-divide-symbol.ts
+++ b/packages/perseus-core/src/utils/get-divide-symbol.ts
@@ -24,7 +24,7 @@ const localesThatUseColon = [
 ];
 
 /**
- *  Get the character used for dividing numbers.
+ *  Get the character used for dividing numbers depending on the locale.
  */
 export const getDivideSymbol = (locale: string): string => {
     // If the locale uses a colon for division, return a colon
@@ -36,9 +36,12 @@ export const getDivideSymbol = (locale: string): string => {
     return "/";
 };
 
+/**
+ * Get the TeX command used for dividing numbers depending on the locale.
+ */
 export const getDivideSymbolForTex = (locale: string): string => {
     const divideSymbol = getDivideSymbol(locale);
 
-    // If the locale uses the forward slash, return the appropriate TeX command
+    // If the locale uses the forward slash, return the TeX command for ÷
     return divideSymbol === "/" ? "\\div" : divideSymbol;
 };
--- a/packages/perseus-score/src/widgets/expression/score-expression.test.ts
+++ b/packages/perseus-score/src/widgets/expression/score-expression.test.ts
@@ -67,12 +67,22 @@ describe("scoreExpression", () => {
         expect(result).toHaveInvalidInput();
     });
 
-    it("should handle correct answers with colon divide symbol", function () {
+    it("should handle corrent answers in en with forward slash divide symbol", function () {
+        const result = scoreExpression("z/3", expressionItem3Options, "en");
+        expect(result).toHaveBeenAnsweredCorrectly();
+    });
+
+    it("should handle corrent answers in en with TeX divide symbol", function () {
+        const result = scoreExpression("z\\div3", expressionItem3Options, "en");
+        expect(result).toHaveBeenAnsweredCorrectly();
+    });
+
+    it("should handle correct answers in uk with colon divide symbol", function () {
         const result = scoreExpression("z:3", expressionItem3Options, "uk");
         expect(result).toHaveBeenAnsweredCorrectly();
     });
 
-    it("should handle invalid answers with colon divide symbol", function () {
+    it("should handle invalid answers in en with colon divide symbol", function () {
         const result = scoreExpression("z:3", expressionItem3Options, "en");
         expect(result).toHaveInvalidInput();
     });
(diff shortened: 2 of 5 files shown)


testing/debug-check-answer-footer.tsx
```diff
@@ -160,7 +160,7 @@ const styles = {
         left: 0,
         right: 0,
         padding: `${sizing.size_120} ${sizing.size_160}`,
-        backgroundColor: semanticColor.core.background.base.default,
+        backgroundColor: "#FFFFFF",
```
SonicScrewdriver, line 165: This token was updated by wonderblocks, but this is breaking most of our storybook docs even after running pnpm install. I've just changed this to white for now as this is a testing component and I'd rather not have to maintain all the semantic colour tokens if they're still fluctuating so frequently.

packages/perseus-core/src/utils/get-divide-symbol.ts
```diff
@@ -0,0 +1,39 @@
...
+    "nl",
+    "pl",
+    "pt-pt",
+    "ro",
+    "sv",
+    "uk",
+    "vi",
+    "ro",
+    "ru",
+];
```
SonicScrewdriver, line 24: I determined these using an internal spreadsheet that breaks down the differences across locales. If you'd like the reference material, please let me know and I'll share it! My searches have indicated that there's no automated way to detect this, so we're stuck checking against a list of known locales.

packages/perseus-core/src/utils/get-divide-symbol.ts
```diff
@@ -0,0 +1,44 @@
...
+    // If the locale uses a colon for division, return a colon
+    if (localesThatUseColon.includes(locale)) {
+        return ":";
+    }
+
+    // Default to using a forward slash
+    return "/";
+};
+
+export const getDivideSymbolForTex = (locale: string): string => {
```
SonicScrewdriver, line 42: The keypad enters TeX commands into our Math Inputs in order to properly display certain math formats/symbols. Unfortunately, the division symbol is one such symbol that requires TeX. (Something that wasn't required for our decimal separator) 

However, our scoring logic requires that we return a forward slash. In order to support both features, I've opted to make a separate function for returning the TeX command, that uses the same logic as the simpler getDivideSymbol. I'm happy to take any alternative suggestions on this, however!

packages/perseus-core/src/utils/get-divide-symbol.ts
```diff
@@ -0,0 +1,44 @@
...
+    const divideSymbol = getDivideSymbol(locale);
+
+    // If the locale uses the forward slash, return the appropriate TeX command
```
nishasy:
```suggestion
    // If the locale uses the forward slash, return the TeX command for ÷
```

packages/perseus-core/src/utils/get-divide-symbol.ts
nishasy: Makes sense to me. It could be helpful to put this explanation in a doc comment for `getDivideSymbolForTex`

packages/perseus-score/src/widgets/expression/score-expression.test.ts
```diff
@@ -67,6 +67,16 @@ describe("scoreExpression", () => {
...
+    it("should handle correct answers with colon divide symbol", function () {
+        const result = scoreExpression("z:3", expressionItem3Options, "uk");
+        expect(result).toHaveBeenAnsweredCorrectly();
+    });
+
+    it("should handle invalid answers with colon divide symbol", function () {
+        const result = scoreExpression("z:3", expressionItem3Options, "en");
+        expect(result).toHaveInvalidInput();
+    });
+
```
nishasy, line 89: It doesn't look like there are any tests for the en divide sign in this file? Could it be worth adding one? I know it's already written as `z/3` in the test data, but I think it could be worth confirming that it still works in the `z/3` case.

packages/perseus/src/widgets/expression/__docs__/expression.stories.tsx
```diff
@@ -71,6 +75,12 @@ export const ExpressionItem3: Story = {
     },
 };
 
+export const ExpressionItem4: Story = {
+    args: {
+        item: expressionItem4,
+    },
+};
+
```
nishasy, line 86: Can you add a doc comment above this story explaining how it can be used to test division? The comment should show up above the story in Storybook as a description.

testing/debug-check-answer-footer.tsx
nishasy: Can you add that as a comment above this line so it's obvious when we come back to it in the future?

nishasy (approved):
Mostly suggestions about comments, nothing blocking. Looks good to me!

PR #2863, opened 2025-09-04
[Image] | (a11y) | Add `longDescription` to Image widget ("Explore image" button + modal)

nishasy:
## Summary:
Add `longDescription` to image widget.
- Add `longDescription` to the schema
- Add `longDescription` to the parser
- Add the "explore image" button when a long description is present
- Add the "explore image" _icon_ button when a long description is present with a caption
- Add a modal that includes the title, image, and description.

NOTE: Backend has landed, so this should be good to land/deploy. HOWEVER, we should NOT update the editor to include long descriptions until we're cleared to do so.

Issue: https://khanacademy.atlassian.net/browse/LEMS-3377

## Test plan:
`pnpm jest packages/perseus/src/widgets/image/image.test.ts`

Storybook
`/?path=/docs/widgets-image-widget-internal-components-imageexplorationmodal--docs`
`/?path=/docs/widgets-image-visual-regression-tests--docs`

## Screenshots:
| case | legacy | classroom |
| --- | --- | --- |
| longdesc + caption | Screenshot 2025-09-10 at 1 01 49 PM | Screenshot 2025-09-10 at 1 02 03 PM |
| longdesc only | Screenshot 2025-09-10 at 1 01 54 PM | Screenshot 2025-09-10 at 1 02 07 PM |

| big picture modal | small picture modal |
| --- | --- |
| Screenshot 2025-09-10 at 1 02 40 PM | Screenshot 2025-09-10 at 1 02 17 PM |

### DIFF
--- a/packages/perseus/src/widgets/image/image-widget.module.css
+++ b/packages/perseus/src/widgets/image/image-widget.module.css
@@ -1,7 +1,6 @@
 :root {
-    /* Undo the default padding for the top of the modal. We need to use
-       px to make sure it's the exact same value as the original padding. */
-    --modal-padding-offset: -32px;
+    /* Undo the default padding for the top of the modal. */
+    --modal-padding-offset: calc(-1 * var(--wb-sizing-size_320));
     /* Fixed px values from the design. */
     --modal-panel-height: 568px;
     --modal-description-container-width: 420px;


packages/perseus/src/widgets/image/image-description-and-caption.tsx
```diff
@@ -48,3 +61,35 @@ export const ImageDescriptionAndCaption = ({
...
+
+function ExploreImageButton({
+    hasCaption,
+    onClick,
+}: {
+    hasCaption: boolean;
+    onClick: () => void;
+}) {
+    const context = React.useContext(PerseusI18nContext);
+    if (hasCaption) {
```
ivyolamit: This is probably preference, it would be good to see this part as the short circuit for the less usual case for code readability:

```suggestion
    const context = React.useContext(PerseusI18nContext);

    if (!hasCaption) {
        return (
            <Button kind="secondary" startIcon={infoIcon} onClick={onClick}>
                {context.strings.imageExploreButton}
            </Button>
        );
    }
    
    return (
            <IconButton
                aria-label={context.strings.imageExploreButton}
                icon={infoIcon}
                kind="secondary"
                onClick={onClick}
                style={{
                    // Stop the button from getting squished by the caption text.
                    // TODO: Use CSS modules after Wonder Blocks styles
                    // are moved to a different layer.
                    minWidth: "40px",
                }}
            />
        );
```

packages/perseus/src/widgets/image/image-description-and-caption.tsx
```diff
@@ -48,3 +61,35 @@ export const ImageDescriptionAndCaption = ({
...
+            <IconButton
+                aria-label={context.strings.imageExploreButton}
+                icon={infoIcon}
+                kind="secondary"
+                onClick={onClick}
+                style={{
+                    // Stop the button from getting squished by the caption text.
+                    // TODO: Use CSS modules after Wonder Blocks styles
+                    // are moved to a different layer.
+                    minWidth: "40px",
```
ivyolamit: is this necessary? doesn't WB handle this?

Also probably consider mobile small screens.

nishasy: Huh... I'm not sure what changed, but it looks like I don't need this anymore. It's not getting squished now.

nishasy: ^ it was the div that fixed it. The squishing is back in the next PR, but Mark showed me how to fix it again.

packages/perseus/src/widgets/image/image-description-and-caption.tsx
```diff
@@ -1,31 +1,47 @@
...
 import {PerseusI18nContext} from "../../components/i18n-context";
 import Renderer from "../../renderer";
 
+import {ImageExplorationModal} from "./image-exploration-modal";
 import styles from "./image-widget.module.css";
 
-import type {APIOptions} from "../../types";
-import type {PerseusImageBackground} from "@khanacademy/perseus-core";
-import type {LinterContextProps} from "@khanacademy/perseus-linter";
+import type {Props} from "./image.class";
```
ivyolamit: This is just a nit, but would be more helpful to have the `Props` called `ImageProps` in that way the type is more clear to read rather than the generic `Props` name.

packages/perseus/src/widgets/image/image-description-and-caption.tsx
```diff
@@ -1,31 +1,47 @@
...
 
-interface Props {
-    caption: string;
-    backgroundImage: PerseusImageBackground;
-    apiOptions: APIOptions;
-    linterContext: LinterContextProps;
-}
+export const ImageDescriptionAndCaption = (props: Props) => {
```
ivyolamit:
```suggestion
export const ImageDescriptionAndCaption = ({
    apiOptions,
    backgroundImage,
    caption,
    linterContext,
    longDescription,
}) => {
```

then we can delete line 23

packages/perseus/src/widgets/image/image-description-and-caption.tsx
```diff
@@ -1,31 +1,47 @@
...
-    linterContext,
-}: Props) => {
     const context = React.useContext(PerseusI18nContext);
+    const imageUpgradeFF = isFeatureOn({apiOptions}, "image-widget-upgrade");
+
     return (
         <div className={styles.descriptionAndCaptionContainer}>
-            {/* Description code will go here */}
+            {/* Description */}
+            {longDescription && imageUpgradeFF && (
```
ivyolamit:
```suggestion
            {imageUpgradeFF && longDescription && (
```
easier to short circuit to check the feature flag first

packages/perseus/src/widgets/image/image-description-and-caption.tsx
nishasy: I was about to leave a comment about our naming convention being `Props` for the main props of the file.

However, I actually think I'm going to create a new interface and _not_ use all the props from `image.class.tsx` because we only need a few specific ones in this file.

packages/perseus/src/widgets/image/image-description-and-caption.tsx
```diff
@@ -1,31 +1,47 @@
+import {isFeatureOn} from "@khanacademy/perseus-core";
```
ivyolamit: what do you think of adding internal components in a folder like this?
```
packages/perseus/src/widgets/image/components/image-description-and-caption.tsx
```

packages/perseus/src/widgets/image/image-description-and-caption.tsx
```diff
@@ -48,3 +64,29 @@ export const ImageDescriptionAndCaption = ({
         </div>
     );
 };
+
+function ExploreImageButton({
```
ivyolamit: can we add this in `packages/perseus/src/widgets/image/components/explore-image-button.tsx`?

packages/perseus/src/widgets/image/image-description-and-caption.tsx
nishasy: I actually need it to stay `props: Props` because I use the destructured `...props` to pass into the modal.

packages/perseus/src/widgets/image/image-exploration-modal.tsx
```diff
@@ -0,0 +1,166 @@
...
+import type {APIOptions} from "../../types";
+import type {
+    Interval,
+    PerseusImageBackground,
+    PerseusImageLabel,
+    Size,
+} from "@khanacademy/perseus-core";
+import type {LinterContextProps} from "@khanacademy/perseus-linter";
+
+const MODAL_HEIGHT = 568;
```
ivyolamit: I'm curious how this will impact mobile or smaller viewport that is below 568?

packages/perseus/src/widgets/image/image-exploration-modal.tsx
```diff
@@ -0,0 +1,166 @@
...
+    const context = React.useContext(PerseusI18nContext);
+    return (
+        <div
+            // We need to manually add the `framework-perseus` class so that
+            // the modal can have the correct styling, even when the portal
+            // makes it render outside the normal `framework-perseus` container.
+            className={`framework-perseus ${styles.modalContainer}`}
+        >
+            <FlexibleDialog
+                title={
```
ivyolamit: I'm not a hug fan of injecting things or using the ternary operator as multiline, it makes it less readable long term especially if more items are added. What do you think of...

```
    const titleText = props.title || context.strings.imageAlternativeTitle;
    const title = (
        <h1
            className={`perseus-image-modal-title ${styles.modalTitleContainer}`}
        >
            {/* Use Renderer so that the title can support markdown and TeX. */}
            <Renderer
                content={titleText}
                apiOptions={props.apiOptions}
                linterContext={props.linterContext}
                strings={context.strings}
            />
        </h1>
    );
```

then the diaglog can now be simple as

```suggestion
            <FlexibleDialog
                title={title}
                content={<ImageExplorationModalContent {...props} />}
                styles={{
                    root: wbStyles.root,
                }}
            />
```

packages/perseus/src/widgets/image/image-exploration-modal.tsx
```diff
@@ -0,0 +1,166 @@
...
+                content={<ImageExplorationModalContent {...props} />}
+                styles={{
+                    root: wbStyles.root,
+                }}
+            />
+        </div>
+    );
+};
+
+const ImageExplorationModalContent = ({
```
ivyolamit: Let's move this out, as another internal component so this file will not be so big.
For code readability, it's easier to have smaller context at a time and more specific.

ivyolamit: This is a good read 😉 
https://www.epicreact.dev/soul-crushing-components

packages/perseus/src/widgets/image/image-exploration-modal.tsx
```diff
@@ -0,0 +1,166 @@
...
+    alt: string;
+    longDescription: string;
+    box: Size;
+    labels: Array<PerseusImageLabel>;
+    range: [Interval, Interval];
+    linterContext: LinterContextProps;
+    apiOptions: APIOptions;
+}
+
+export const ImageExplorationModal = (props: Props) => {
```
ivyolamit: suggestion `ExploreImageModal` so it will be consistent naming with `ExploreImageButton`. also it would be easier to navigate in the code when they're next to each other.

packages/perseus/src/widgets/image/image-exploration-modal.tsx
```diff
@@ -0,0 +1,166 @@
...
+                    apiOptions={apiOptions}
+                    linterContext={linterContext}
+                    strings={context.strings}
+                />
+            </div>
+        </div>
+    );
+};
+
+const wbStyles = {
```
ivyolamit: is there a reason we can't put these styles in .css file?

nishasy: Yeah, wonder blocks components in Perseus can't use CSS because they're in the same layer. I'll add a comment about that.

packages/perseus/src/widgets/image/image-exploration-modal.tsx
nishasy: I'll make that change.

I spent a while going back and forth on whether I really wanted to move it out, because it's so tightly coupled with the modal itself. Ideally, I would have put it in the modal component similar to the title, but it got too bulky and awkard.

packages/perseus/src/widgets/image/image-exploration-modal.tsx
nishasy: This is mostly to stop images from getting too big. That's not a problem on smaller screen sizes. It should automatically get smaller. I think this is necessary for the desktop modal, but I'm totally down to test/change/address in the mobile PR to make sure. I can add a conditional there if necessary.

packages/perseus/src/widgets/image/__docs__/image-exploration-modal.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+import {ApiOptions} from "../../../perseus-api";
+import {ExploreImageModal} from "../components/explore-image-modal";
+import {earthMoonImage, frescoImage} from "../utils";
+
+import styles from "./image-stories.module.css";
+
+import type {Meta, StoryObj} from "@storybook/react-vite";
+
+const meta: Meta<typeof ExploreImageModal> = {
+    title: "Widgets/Image/Widget Internal Components/ImageExplorationModal",
```
ivyolamit:
```suggestion
    title: "Widgets/Image/Widget Internal Components/Image Exploration Modal",
```

packages/perseus/src/widgets/image/__docs__/image-stories.module.css
```diff
@@ -0,0 +1,8 @@
+.container {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    justify-content: center;
+    padding: 20px;
```
ivyolamit:
```suggestion
    padding: var(--wb-sizing-size_200);
```

packages/perseus/src/widgets/image/components/explore-image-button.tsx
```diff
@@ -0,0 +1,32 @@
...
+import IconButton from "@khanacademy/wonder-blocks-icon-button";
+import infoIconBold from "@phosphor-icons/core/bold/info-bold.svg";
+import * as React from "react";
+
+import {PerseusI18nContext} from "../../../components/i18n-context";
+
+export default function ExploreImageButton({
+    hasCaption,
+    onClick,
+}: {
```
ivyolamit:
```suggestion
}: ExploreImageButtonProps
```

packages/perseus/src/widgets/image/components/explore-image-modal-content.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+import AssetContext from "../../../asset-context";
+import {SvgImage} from "../../../components";
+import {PerseusI18nContext} from "../../../components/i18n-context";
+import Renderer from "../../../renderer";
+import styles from "../image-widget.module.css";
+
+import type {Props as ImageDescriptionProps} from "./image-description-and-caption";
+
+type Props = ImageDescriptionProps;
```
ivyolamit: is there a reason why this line is added? I think we can remove this, and in line 26 we can use `ImageDescriptionProps` directly.
```
}: ImageDescriptionProps) {
```

packages/perseus/src/widgets/image/components/explore-image-modal.test.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+describe("ImageExplorationModal", () => {
+    beforeEach(() => {
+        jest.spyOn(Dependencies, "getDependencies").mockReturnValue(
+            testDependencies,
+        );
+    });
+
+    it("should render the modal", () => {
+        // Arrange
```
ivyolamit, line 46: since it has nothing for this scenario we can remove this
```suggestion
```
This is also applicable to other scenarios in this file.

packages/perseus/src/widgets/image/components/explore-image-modal.tsx
```diff
@@ -0,0 +1,58 @@
...
+import Renderer from "../../../renderer";
+import styles from "../image-widget.module.css";
+
+import ExploreImageModalContent from "./explore-image-modal-content";
+
+import type {Props as ImageDescriptionProps} from "./image-description-and-caption";
+
+type Props = ImageDescriptionProps;
+
+export const ExploreImageModal = (props: Props) => {
```
ivyolamit:
```suggestion
export const ExploreImageModal = (props: ImageDescriptionProps) => {
```

adn we can remove line 13

packages/perseus/src/widgets/image/components/image-description-and-caption.tsx
```diff
@@ -0,0 +1,84 @@
...
+import type {APIOptions} from "../../../types";
+import type {
+    Interval,
+    PerseusImageBackground,
+    PerseusImageLabel,
+    Size,
+} from "@khanacademy/perseus-core";
+import type {LinterContextProps} from "@khanacademy/perseus-linter";
+
+export interface Props {
```
ivyolamit: suggestion: making the props more descriptive than just generic Props, this makes it easier for maintenance/debugging.
```suggestion
export interface ImageDescriptionAndCaption {
```

packages/perseus/src/widgets/image/image-widget.module.css
```diff
@@ -1,3 +1,12 @@
+:root {
+    /* Undo the default padding for the top of the modal. We need to use
+       px to make sure it's the exact same value as the original padding. */
+    --modal-padding-offset: -32px;
```
ivyolamit:
```suggestion
    --modal-padding-offset: -var(--wb-sizing-size_320);
```

packages/perseus/src/widgets/image/image.tsx
```diff
@@ -5,25 +5,31 @@ import {PerseusI18nContext} from "../../components/i18n-context";
 import SvgImage from "../../components/svg-image";
 import Renderer from "../../renderer";
 
-import {ImageDescriptionAndCaption} from "./image-description-and-caption";
+import {ImageDescriptionAndCaption} from "./components/image-description-and-caption";
 import styles from "./image-widget.module.css";
 
 import type {Props} from "./image.class";
```
ivyolamit:
```suggestion
import type {Props as ImageDescriptionProps} from "./image-description-and-caption";
```

packages/perseus/src/widgets/image/utils.ts
ivyolamit, line 1: aha! using your github tip!
What do you think if the contents of this file will be moved in `__docs__` since this is more useful in our stories?

packages/perseus/src/widgets/image/components/explore-image-modal-content.tsx
nishasy: I added this line because we have a convention to name the main props in the file just `Props`. I did it this way to make it clear that the props are coming from a different file and are remaining the same as those.

I'd prefer to either keep it like this, or do this and still have it named `Props`:

```
import type {Props} from "./image-description-and-caption";
```

packages/perseus/src/widgets/image/components/explore-image-modal.test.tsx
nishasy: We're supposed to keep the `// Arrange` even if it's empty. All the tests are supposed to have `// Arrange`, `// Act`, and `// Assert`.

Some people put `// Arrage, Act` so it's one line, but frontend ifra prefers them to be separate if I recall correctly.

packages/perseus/src/widgets/image/components/explore-image-modal.tsx
nishasy: I think this is the same as the other comment I left. I'd prefer to keep the name as `Props` because that's our longstanding convention and makes it clear that it's the main props for that file.

packages/perseus/src/widgets/image/__docs__/image-exploration-modal.stories.tsx
nishasy: I'm going to change this to `ExploreImageModal`. I don't think I want to add spaces though, because the component name doesn't have spaces.

packages/perseus/src/widgets/image/components/image-description-and-caption.tsx
nishasy: Okay I think I'm coming around to changing these in this case since this particular interface is exported. I'll make that change.

packages/perseus/src/widgets/image/image.tsx
nishasy: I'm using props from the image class here because it has `trackInteraction` in it.

packages/perseus/src/widgets/image/image-widget.module.css
nishasy: I was about to comment to point out the my code has a comment saying we have to use px to make sure it's the exact same as wonder blocks, so we can't use wb sizing because it's in rems.

BUT I just checked the actual wb code, and it's using wb sizing, so I'll make that change.

packages/perseus/src/widgets/image/utils.ts
nishasy: It's also used for the tests in `components/explore-image-modal.test.tsx` and `image.test.ts`, so I think it might make more sense to keep it where it is.

packages/perseus/src/widgets/image/image-exploration-modal.tsx
ivyolamit: i think i keep forgetting this 😓

packages/perseus/src/widgets/image/__docs__/image-exploration-modal.stories.tsx
ivyolamit: Good point, Probably i would need to revisit other widgets for consistency, i added spaces in the improvement pr id dit while back

image

nishasy: I think the widget _names_ are fine to have spaces in, since they're sort of bigger concepts? But internal components are like... component components? I'm not sure if that makes sense, but I think of widgets as sort of different than regular components.

PR #2864, opened 2025-09-04
Remove RenderProps & transform & staticTransform

handeyeco:
## Summary:

The reason I'm looking at RenderProps is because of the need for SSS to transition from an "answerless" to an "answerful" state - in order to provide hints, rationales, and answers to learners post-attempt. The problem is that Renderer stashes a copy of RenderProps (essentially the transformed output of passing widget options through either `transform` or `staticTransform`). Originally this problem manifested by resetting user input - when we went from answerless to answerful, Renderer considered it a new exercise and completely reset state. Once that was fixed the problem manifested in a different way - when going from answerless to answerful Renderer kept a copy of RenderProps which meant we never received the answerful data in the widgets.

> There are only two hard things in Computer Science: cache invalidation, naming things, and off-by-1 errors.

So what if there were no cache? That's the aim of this PR.

In previous PRs we:
- split UserInput into its own chunk of state
- removed all uses of `transform` and `staticTransform`
- internalized widget state within the widget itself (rather than in RenderProps)
- removed all "restore" serialized state uses / APIs (which depended on RenderProps)

This PR is mostly about cleaning up code. Since UserInput is stored separately, widgets store state internally, and there are no more transforms, RenderProps is just a copy of WidgetOptions now. With this PR we cut out the middle-man and pass WidgetOptions directly to the widgets removing the need for RenderProps.

Issues:
- [LEMS-3113](https://khanacademy.atlassian.net/browse/LEMS-3113) - Answerless Rendering trial run in Khan Classic
- [LEMS-3199](https://khanacademy.atlassian.net/browse/LEMS-3199) - WidgetOptions should be read only in the learner UX, deprecate transform/staticTransform
- [LEMS-3185](https://khanacademy.atlassian.net/browse/LEMS-3185) - Deprecate `getSerializedState` and `restoreSerializedState`

## Test plan:
- There should be no UX changes
- Unit tests should pass
- Doing a playtest before release

[LEMS-3199]: https://khanacademy.atlassian.net/browse/LEMS-3199?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/perseus/src/__testdata__/renderer.testdata.ts
```diff
@@ -218,74 +217,3 @@ export const mockedRandomItem: PerseusRenderer = {
         },
     },
 } as PerseusRenderer;
-
-export const mockedShuffledRadioProps: {[key in string]: RenderProps} = {
```
handeyeco, line 222: Wasn't used

packages/perseus/src/__tests__/renderer.test.tsx
```diff
@@ -201,18 +201,6 @@ describe("renderer", () => {
             expect(renderer.state.lastUsedWidgetId).toBeNull();
 
             expect(renderer.state.widgetInfo).toStrictEqual(question1.widgets);
-            expect(renderer.state.widgetProps).toEqual({
```
handeyeco, line 204: `renderer.state.widgetProps` is essentially Renderer's stash of RenderProps which this PR removes.

packages/perseus/src/renderer.tsx
```diff
@@ -249,9 +245,6 @@ class Renderer
 
     _widgetContainers: Map<string, WidgetContainer> = new Map();
 
-    lastRenderedMarkdown: React.ReactNode;
-    // @ts-expect-error - TS2564 - Property 'reuseMarkdown' has no initializer and is not definitely assigned in the constructor.
-    reuseMarkdown: boolean;
```
handeyeco, line 254: So this was basically cashing the result of the render for reuse. Super weird. Maybe we want to separate out parsing/rendering and that's what this was for? To prevent reparsing Markdown? But if so this isn't the way we should do it and I'm not sure I agree that we should do it at all.

packages/perseus/src/renderer.tsx
```diff
@@ -458,33 +398,13 @@ class Renderer
...
-        widgetProps: State["widgetProps"];
     } = (props: Props) => {
         const allWidgetInfo = applyDefaultsToWidgets(props.widgets);
         return {
             widgetInfo: allWidgetInfo,
-            widgetProps: this._getAllWidgetsStartProps(allWidgetInfo, props),
         };
     };
 
-    _getAllWidgetsStartProps: (
```
handeyeco, line 470: This generated RenderProps, which we've removed.

packages/perseus/src/renderer.tsx
```diff
@@ -556,7 +476,7 @@ class Renderer
                         }
                     }}
                     type={type}
-                    initialProps={this.getWidgetProps(id)}
+                    widgetProps={this.getWidgetProps(id)}
```
handeyeco, line 479: I'll talk about this change in WidgetContainer

packages/perseus/src/widget-container.tsx
```diff
@@ -30,7 +30,6 @@ type DefaultProps = {
 
 type State = {
     sizeClass: "small" | "medium" | "large" | "xlarge";
-    widgetProps: WidgetProps<any, PerseusWidgetOptions>;
```
handeyeco, line 33: So it seems like it wasn't just Renderer that was keeping a stash of RenderProps, but WidgetContainer too. Instead, with this PR we're now just passing through WidgetProps and WidgetContainer is now a little dumber (_"it's a good thing"_). I got Jeremy's blessing, but lmk if you see a problem.

packages/perseus/src/widgets/categorizer/categorizer.tsx
```diff
@@ -44,7 +49,7 @@ export class Categorizer
     static defaultProps: DefaultProps = {
         items: [],
         categories: [],
-        linterContext: linterContextDefault,
+        linterContext: linterContextDefault as any,
```
handeyeco: I couldn't figure this out. I was doing the exact same thing that NumericInput was doing, but I couldn't get rid of the error. Open to suggestions.

packages/perseus/src/widgets/cs-program/cs-program.tsx
```diff
@@ -95,6 +96,18 @@ class CSProgram extends React.Component<Props> implements Widget {
         return _getPromptJSON();
     }
 
+    /**
+     * @deprecated and likely very broken API
+     * [LEMS-3185] do not trust serializedState
+     */
+    getSerializedState(): any {
```
handeyeco, line 103: I added/updated `getSerializedState` in a couple of places to keep the serialized state shape consistent.

packages/perseus/src/widgets/explanation/explanation.test.ts
```diff
@@ -244,23 +242,4 @@ describe("Explanation", function () {
             "transitionCollapsed",
         );
     });
-
-    it("communicates changes to its parent by using the provided 'onChange' callback", () => {
-        // Arrange
-        // @ts-expect-error // Argument of type {onChange: () => void;} is not assignable to parameter of type Props | Readonly<Props>
-        const widget = new ExplanationWidgetExports.widget({
-            onChange: () => {},
```
handeyeco, line 252: `onChange` in widgets was our way of updating RenderProps.

packages/perseus/src/widgets/group/group.tsx
```diff
@@ -130,17 +123,6 @@ class Group extends React.Component<Props> implements Widget {
             this.props.widgetId,
         );
 
-        // This is a little strange because the id of the widget that actually
-        // changed is going to be lost in favor of the group widget's id. The
-        // widgets prop also wasn't actually changed, and this only serves to
-        // alert our renderer (our parent) of the fact that some interaction
-        // has occurred.
-        const onInteractWithWidget = (id) => {
```
handeyeco, line 138: This should be handled by `handleUserInput` now

packages/perseus/src/widgets/label-image/label-image.tsx
```diff
@@ -367,19 +367,7 @@ export class LabelImage
...
-        const updatedRenderProps = [
-            ...markers.slice(0, index),
-            {
-                ...markers[index],
-                // Do not show correctness state if user changes answer
-                // selection for marker.
-                showCorrectness: undefined,
-            },
-            ...markers.slice(index + 1),
-        ];
```
handeyeco, line 382: I manually tested and the data here didn't seem to be used.

packages/perseus/src/widgets/matrix/matrix.tsx
```diff
@@ -83,30 +83,23 @@ const getRefForPath = function (path: FocusPath) {
     PerseusMatrixUserInput
 > satisfies PropsFor<typeof Matrix>;
 
-type Props = WidgetProps<MatrixPublicWidgetOptions, PerseusMatrixUserInput> & {
-    // The coordinate location of the cursor position at start. default: [0, 0]
-    cursorPosition: ReadonlyArray<number>;
```
handeyeco, line 88: Moved this to be internal state rather than RenderProps

packages/perseus/src/widgets/passage/passage.tsx
```diff
@@ -64,26 +62,23 @@ const styles = StyleSheet.create({
...
+type PassageProps = WidgetProps<PerseusPassageWidgetOptions> & {
+    findWidgets: (arg1: FindWidgetsCallback) => ReadonlyArray<Passage>;
+};
 
 type DefaultPassageProps = {
     passageTitle: PassageProps["passageTitle"];
     passageText: PassageProps["passageText"];
     footnotes: PassageProps["footnotes"];
     showLineNumbers: PassageProps["showLineNumbers"];
-    highlights: PassageProps["highlights"];
```
handeyeco, line 79: Moved this to be internal state

packages/perseus/src/widgets/plotter/plotter.tsx
```diff
@@ -46,6 +47,8 @@ type State = {
     categoryHeights: Record<string, number>;
 };
 
+const plotterDefaults = CoreWidgetRegistry.getDefaultWidgetOptions("plotter");
```
handeyeco, line 50: So we have several ways to default data:

- The parser defaults data
- We have `getDefaultWidgetOptions`
- We have `DefaultProps`

For some reason Plotter had multiple values for the same defaults. I'm talking with Ben on how we can consolidate all of these defaults.

packages/perseus/src/renderer.tsx
```diff
@@ -590,7 +510,7 @@ class Renderer
         widgetId: string,
     ): WidgetProps<any, any, PerseusWidgetOptions> {
         const apiOptions = this.getApiOptions();
-        const widgetProps = this.state.widgetProps[widgetId] || {};
+        const widgetProps = this.props.widgets[widgetId].options;
```
handeyeco, line 513: Now instead of using the cache of RenderProps, we pass WidgetOptions straight through.

packages/perseus/src/widgets.ts
```diff
@@ -178,59 +164,20 @@ export const getAllWidgetTypes = (): ReadonlyArray<string> => {
...
-
 /**
  * Handling for static mode for widgets that support it.
  */
 
 /**
  * Returns true iff the widget supports static mode.
  * A widget implicitly supports static mode if it exports a
- * staticTransform function.
+ * getCorrectUserInput function.
```
benchristel, line 174: Thanks for fixing this comment!

packages/perseus/src/widgets/categorizer/categorizer.tsx
benchristel: It's because `PerseusLinterContext` has `paths` and `stack` as read-write arrays, but `LinterContextProps` has readonly arrays. Based on the TS error message, it seems like we `&` these two types together at some point:

```
Type 'LinterContextProps' is not assignable to type 'PerseusLinterContext & LinterContextProps'.
  Type 'LinterContextProps' is not assignable to type 'PerseusLinterContext'.
    Types of property 'paths' are incompatible.
      The type 'readonly string[]' is 'readonly' and cannot be assigned to the mutable type 'string[]'.
```

I think the `&` operator is nondeterministic if you give it object types with overlapping sets of properties that have different types.

Could one of the types be eliminated? Why do we need both `LinterContextProps` and `PerseusLinterContext`?

EDIT: oh, I think `PerseusLinterContext` exists because `data-schema.ts` isn't supposed to import types from elsewhere.

I tried changing `PerseusLinterContext` to use ReadonlyArray:


```ts
export type PerseusLinterContext = {
    contentType: string;
    paths: ReadonlyArray<string>;
    stack: ReadonlyArray<string>;
};
```

but that gave me a different type error from the parser typetests:

```ts
The type 'readonly string[]' is 'readonly' and cannot be assigned to the mutable type 'string[]'.

   summon<CategorizerWidget>() satisfies Parsed;
```

We removed readonly types from the data schema in https://github.com/Khan/perseus/pull/2398. I think the simplest course of action would be to remove `readonly` from `LinterContextProps`:

```ts
export type LinterContextProps = {
    contentType: string;
    highlightLint: boolean;
    paths: string[];
    stack: string[];
}
```

handeyeco: What about: https://github.com/Khan/perseus/pull/2864/commits/c731529ad4ea9c9ec7d3e0506d641c614c722649

I think the problem is the schema is trying to type something that's not really a widget option, it's part of the universal widget options.

benchristel: So you're saying it got included in the JSON by mistake? That sounds very plausible to me. I think removing it from the widget options is probably the right call.

packages/perseus/src/widgets/radio/radio.ff.tsx
```diff
@@ -85,6 +83,7 @@ class Radio extends RadioOld {
         } = this._mergePropsAndState();
         return {
             ...rest,
+            hasNoneOfTheAbove: rest.hasNoneOfTheAbove ?? false,
```
benchristel, line 86: Why is this line needed?

packages/perseus/src/widgets/categorizer/categorizer.tsx
handeyeco: Right. I think we have content that includes it (because the data schema was made by extracting properties from existing content), but I think it was by accident. We used to make content by getting serialized state which used the deny list to exclude things we didn't want saved; the deny list includes `linterContext` which signals that we don't want to save it, but it's possible that was added to the deny list after content was already created with it.

packages/perseus/src/widgets/radio/radio.ff.tsx
handeyeco: I added this to make these tests pass:

- Radio serialization › should serialize the current state
- group widget › should return contained renderer's getSerializedState

`hasNoneOfTheAbove` was getting defaulted to false when undefined as part of `applyDefaultsToWidgets`. Like we talked about, I think we're pretty covered with defaults: from the parser and from the widget's default props. I just don't think we run test data through the parser so it wasn't getting updated. This was a small fix for a deprecated API, so I didn't think much about it. Maybe I need to do something else?

PR #2867, opened 2025-09-04
fix(LEMS-3482): add non-empty string validation in radio parser

anakaren-rojas:
## Summary:
- Creates defaulted function which handles empty strings as invalid
- Consumes new function in radio parser, sets `generateChoiceIds` function as fallback when existing ID is an empty string
- Adds tests that validate choice ids will not be empty strings and are properly updated when id is not present 

Issue: LEMS-3482

## Test plan:

https://github.com/user-attachments/assets/9c66ea80-0f3c-4978-af08-b632ef1f3282

### DIFF
--- a/packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.test.ts
+++ b/packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.test.ts
@@ -141,7 +141,6 @@ describe("parseRadioWidget", () => {
 
         expect(parse(widget, parseRadioWidget)).toEqual(anySuccess);
     });
-
 });
 
 describe("migration functions", () => {
--- a/packages/perseus-core/src/parse-perseus-json/general-purpose-parsers/defaulted.test.ts
+++ b/packages/perseus-core/src/parse-perseus-json/general-purpose-parsers/defaulted.test.ts
@@ -47,12 +47,12 @@ describe("defaultedNonEmptyString()", () => {
     const stringWithDefault = defaultedNonEmptyString(() => "default-id");
 
     it.each`
-        input        | expected           | description
-        ${undefined} | ${"default-id"}    | ${"undefined"}
-        ${null}      | ${"default-id"}    | ${"null"}
-        ${""}        | ${"default-id"}    | ${"empty string"}
-        ${"   "}     | ${"default-id"}    | ${"whitespace-only string"}
-        ${"my-id"}   | ${"my-id"}         | ${"valid non-empty string"}
+        input        | expected        | description
+        ${undefined} | ${"default-id"} | ${"undefined"}
+        ${null}      | ${"default-id"} | ${"null"}
+        ${""}        | ${"default-id"} | ${"empty string"}
+        ${"   "}     | ${"default-id"} | ${"whitespace-only string"}
+        ${"my-id"}   | ${"my-id"}      | ${"valid non-empty string"}
     `(
         "should return $expected when given $description",
         ({input, expected}) => {


packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.test.ts
```diff
@@ -108,6 +108,160 @@ describe("parseRadioWidget", () => {
...
+        it("generates default ID when choice ID is missing", () => {
+            const widget = {
+                type: "radio",
+                version: {major: 3, minor: 0},
+                graded: true,
+                options: {
+                    choices: [
+                        {
+                            content: "Choice without ID",
+                            // id property is missing
```
mark-fitzgerald: This is testing for `undefined`, correct? Should there be an explicit test for `null`?

mark-fitzgerald (approved):
Quick fix. Thank you!

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.test.ts
anakaren-rojas: Added!

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.test.ts
```diff
@@ -108,6 +108,193 @@ describe("parseRadioWidget", () => {
...
+                                content: "Choice with empty ID",
+                                id: "radio-choice-0", // Should be generated
+                            },
+                        ],
+                    },
+                }),
+            );
+        });
+
+        it("generates default ID when choice ID is whitespace-only", () => {
```
benchristel: Do we actually need to treat whitespace-only IDs as invalid? It doesn't seem like a case we're likely to ever encounter.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/radio-widget.test.ts
```diff
@@ -108,6 +108,193 @@ describe("parseRadioWidget", () => {
...
+                            {content: "First choice", id: "radio-choice-0"},
+                            {content: "Second choice", id: "custom-id"}, // preserved
+                            {content: "Third choice", id: "radio-choice-2"},
+                            {content: "Fourth choice", id: "radio-choice-3"},
+                        ],
+                    },
+                }),
+            );
+        });
+    });
```
benchristel: These tests cover all the cases, but they're a little verbose, and they're really testing `defaultedNonEmptyString`. I'd prefer to have comprehensive tests in `defaulted.test.ts` that call `defaultedNonEmptyString` directly, and then we'd only need to test a couple cases here (maybe the empty string case and the nonempty string case).

If there's published content that has blank choice IDs, we should also put an example of that in a regression test data file in `regression-tests/item-data`.

.changeset/funny-seahorses-reflect.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/perseus-core": patch
+---
+
+Radio fix that disallows choice ids that are empty strings
```
benchristel: Can we say something like "defaults" here instead of "disallows"?

```suggestion
Radio fix that defaults choice ids that are empty strings
```

benchristel (approved):
I left a few suggestions inline, but I don't think any of them are blocking. Thanks for the fix!

packages/perseus-core/src/parse-perseus-json/general-purpose-parsers/defaulted.ts
```diff
@@ -13,3 +13,33 @@ export function defaulted<T>(
...
+            // If empty or whitespace-only, use fallback
+            if (rawValue.trim() === "") {
+                return ctx.success(fallback());
+            }
+            // Valid non-empty string, use as-is
+            return ctx.success(rawValue);
+        }
+
+        // Invalid type, expect string
+        return ctx.failure("string", rawValue);
```
benchristel, line 43: I don't think this "invalid type" case is covered by a test. Does anything fail if you replace this line with `return ctx.success("")`?

PR #2868, opened 2025-09-05
fix: update radio editor to normalize choice ids for existing content

anakaren-rojas:
## Summary:
For existing content
- Adds logic into radio editor to normalize ids when they are empty strings, only whitespace, null, or undefined.
- Adds tests to validate that choice ids are being normalized

NOTES:
- This fix does not cover new content, choice ids are added to new content in the `add choice` method 
- For existing content, the ID normalization will follow the same pattern as the parser - `radio-choice-<index>`, new content gets a UUID

Issue: LEMS-XXXX

## Test plan:
[ZND](https://prod-znd-250909-3109-b3007b0.khanacademy.org/devadmin/content/exercises/survey-copy/x1055272e2e055025/x2a37cc6b9716baea)
1. Navigate to "survey copy"
2. Open dev tools and navigate to `Elements` tab
3. Inspect an existing choice
4. Add a new choice
5. Delete an existing choice 

### Expected behavior
- Existing content will have `data-choice-id` in the format of `radio-choice-<index>`
- New content will have `data-choice-id` in the format of a UUID (x-x-x-x-x)

Validated that choice ids are being added to choices on load.
https://github.com/user-attachments/assets/50ced07f-ab59-47e1-80a0-e021a5c6d991


packages/perseus-editor/src/widgets/radio/radio-option-settings.tsx
```diff
@@ -43,7 +43,7 @@ export function RadioOptionSettings({
     const rationaleTextAreaId = `${uniqueId}-rationale-textarea`;
 
     return (
-        <div className={styles.tile}>
+        <div className={styles.tile} data-choice-id={choice.id}>
```
anakaren-rojas: temp for testing

ivyolamit (approved):
@anakaren-rojas changes looks logical to me, smoke test works as expected.

Since this is an editor change, is this something we need to worry with the pinned version of AX?

anakaren-rojas:
> @anakaren-rojas changes looks logical to me, smoke test works as expected.
> 
> Since this is an editor change, is this something we need to worry with the pinned version of AX?

Confirmed that we don't need to worry about it since there aren't schema changes

PR #2869, opened 2025-09-05
[Image] | (UX) | Fix extra spacing that got released with image redesign

nishasy:
## Summary:
Turns out that Renderers (i.e. Image caption and title) *inside articles*
have some extra spacing.

- Add CSS to fix extra spacing
- Add regression stories to account for the article context
- Make `image.css` file easier to read

Issue: https://khanacademy.atlassian.net/browse/LEMS-3484

| Before | After |
| --- | --- |
| image | Screenshot 2025-09-05 at 2 10 08 PM |

## Test plan:
storybook
`?path=/docs/widgets-image-visual-regression-tests--docs&globals=theme:thunderblocks#within-article-desktop-caption-and-title`

Use this PR to make a ZND and confirm the extra spacing is gone.

### DIFF
--- a/.changeset/hip-wasps-tell.md
+++ b/.changeset/hip-wasps-tell.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+[Image] | (UX) | Fix extra spacing that got released with image redesign


packages/perseus/src/widgets/image/__docs__/image-regression.stories.tsx
```diff
@@ -56,6 +56,18 @@ const mobileDecorator = (Story) => (
     </div>
 );
 
+const articleDecorator = (Story) => (
```
ivyolamit, line 59: For these decorators, what do you think of moving these up on a higher level so it can be reused?

like for example the mobile decorator i'm already seeing it in two places
```
        <div className="framework-perseus perseus-mobile">
            <Story />
        </div>
```

And i think there's a value to make these available for other widgets to reuse.

nishasy: That makes sense to me. Would you be okay with me doing it in a separate PR?

ivyolamit: yes, that's sounds good to me.

ivyolamit (approved):
Thanks for the quick fix @nishasy changes LGTM 👍🏼

PR #2873, opened 2025-09-08
Add optional generateUrl function to dependencies

MikeKlemarewski:
## Summary:
[In this thread](https://khanacademy.slack.com/archives/C01AZ9H8TTQ/p1753388936490089) we chatted about creating one single dependency function that the client could use to manipulate urls if it so wished. This would allow the client to decide the fully qualified url that should be loaded when faced with a relative url, or maybe it decides to load an embedded video from a separate domain than where the renderer is currently running.

Here I add a `generateUrl` dependency function. Instead of calling `staticUrl` or `absoluteUrl` we can use `generateUrl` and move any url specific logic into the client. [Here is a PR](https://github.com/Khan/frontend/pull/3785) that shows how webapp would use this new dependency function.

## Test plan:
- Check out [the related webapp PR](https://github.com/Khan/frontend/pull/3785) and make sure the app behaves as expected
- Ensure tests pass

### DIFF
--- a/.changeset/silly-geckos-cover.md
+++ b/.changeset/silly-geckos-cover.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": patch
+---
+
+Updated tests as a result of the new generateUrl dependency function


packages/perseus/src/widgets/python-program/python-program.tsx
```diff
@@ -39,7 +40,13 @@ class PythonProgram extends React.Component<Props> implements Widget {
     }
 
     render(): React.ReactNode {
-        const url = getUrlFromProgramID(this.props.programID);
+        let url = getUrlFromProgramID(this.props.programID);
```
MikeKlemarewski, line 45: I wonder if we even want this logic built into perseus anymore, or if we'd make the client specify it inside of the `generateUrl` function.

packages/perseus/src/widgets/video/video.tsx
```diff
@@ -77,6 +78,12 @@ class Video extends React.Component<Props> implements Widget {
             url = url.replace("{host}", embedHostname);
         }
```
MikeKlemarewski, line 85: Same question here, we have some very specific logic about where to load embedded videos from. Should this move to the client?

packages/perseus/src/widgets/python-program/python-program.tsx
jeremywiebe: I'd agree.  It seems like this is very Khan Academy-specific (well, I'd argue that both `cs-program` and `python-program` are _very- KA-specific.  

I think for now I'd like to avoid making major changes to this and deal with URL generation once we have a built URL (instead of dealing with mapping of program ID to URL as part of this work).  But you make a good point!

packages/perseus/src/types.ts
```diff
@@ -160,6 +161,11 @@ type TrackInteractionArgs = {
 } & Partial<TrackingGradedGroupExtraArguments> &
     Partial<TrackingSequenceExtraArguments>;
 
+type GenerateUrlArgs = {
+    url: string;
+    widget: keyof PerseusWidgetTypes;
```
jeremywiebe: I was thinking about how this feature would be used.  I'm tempted to make this slightly more generic than being directly tied to widget. 

I know that there are places in Perseus where we use the [`staticUrl()`](https://github.com/Khan/perseus/blob/5c7da8d1a2270bf73d60fcfc9823414db167af45/packages/perseus/src/types.ts#L378) dependency and it's not within a widget (but rather inside helper components/functions). 

So, perhaps we could just call this `context`? This would give us flexibility to define whatever "context" string makes sense for each usage. We can use the widget type if that's all that's needed, or if there's multiple usages in the widget, we can qualify it (like `<widget>:<use>`).  

Something like this:

```typescript 
        url =
            this.props.apiOptions.generateUrl?.({
                url,
                context: "video",
            }) ?? url;
```

Or... 

```typescript 
        url =
            this.props.apiOptions.generateUrl?.({
                url,
                context: "python_program:program_url",
            }) ?? url;
```

jeremywiebe (commented):
This looks like what we discussed.  I have a few requests before we finalize this PR if you can make time for them:

  1. Could you mark both [`toAbsoluteUrl()`](https://github.com/Khan/perseus/blob/5c7da8d1a2270bf73d60fcfc9823414db167af45/packages/perseus/src/util/url-utils.ts#L35) and [`staticUrl`](https://github.com/Khan/perseus/blob/5c7da8d1a2270bf73d60fcfc9823414db167af45/packages/perseus/src/types.ts#L378) as `@deprecated` with a note to use this new function instead? 

  2. Could you take a stab at adjusting at least one of `toAbsoluteUrl` and one of `staticUrl` to use this function as a proof that this function will be able to replace these two functions fully? We can make a ticket to do the full replacement, but I think using one site for each would be a good proof of this new function. 


Thanks Mike!

packages/perseus/src/util/use-generate-url.ts
```diff
@@ -0,0 +1,42 @@
...
+            let updatedUrl = args.url;
+
+            switch (args.context) {
+                case "image_loader":
+                    updatedUrl = staticUrl(args.url);
+                    break;
+                case "python_program:program_url":
+                    updatedUrl = toAbsoluteUrl(args.url);
+                    break;
+            }
```
MikeKlemarewski: @jeremywiebe I'm not sure if this is what you had in mind or not. We could isolate all calls to `staticUrl` and `toAbsoluteUrl` inside this function call, and deprecate those functions.

MikeKlemarewski: With this change, I removed the call to `toAbsolutePath` [from the python-program widget](https://github.com/Khan/perseus/pull/2873/files#diff-bcd5e521843f02e5313d638ecc6eafb573330770cf2567e501b1a930b97df763L17) and removed the staticUrl call [from the image-loader](https://github.com/Khan/perseus/pull/2873/files#diff-df413e315e2189412ecf7f9140fc76b5c462258fff24582ee81ce332ed585550L143)

packages/perseus/src/types.ts
```diff
@@ -283,6 +295,13 @@ export type APIOptions = Readonly<{
      * Define the feature flag name in packages/perseus-core/src/feature-flags.ts
      */
     flags?: Record<(typeof PerseusFeatureFlags)[number], boolean>;
+    /**
+     * A function that takes a URL or partial url and may modify it to return
+     * the full URL. This may be used to request a resource from a different
+     * app to where the widget is rendered, like when embedding a video from
+     * khanacademy.org inside KAC.
```
jeremywiebe: Since Perseus is open source, I try not to talk too much about the concrete implementations within Khan Academy unless absolutely necessary.  I think we can just talk about embedding videos cross-domain or something like that.

packages/perseus/src/util/use-generate-url.ts
jeremywiebe: I was thinking that this logic that switches on context would be lifted all the way out of Perseus and live in the `generateUrl` function passed to the dependencies.  The whole point is to give the host application the control to transform the input URL/path however it wants. 

So in my mind, `staticUrl` and `toAbsoluteUrl` would be removed eventually and all we'd have is `generateUrl()` which the consuming app fully controls. At that point the `context` passed along to `generateUrl` would be used to decide if we need to make the url a static url (ie. `kastatic.org/...`) or do something else.

packages/perseus/src/widgets/python-program/python-program.tsx
```diff
@@ -39,7 +38,13 @@ class PythonProgram extends React.Component<Props> implements Widget {
     }
 
     render(): React.ReactNode {
-        const url = getUrlFromProgramID(this.props.programID);
+        const path = getPathFromProgramID(this.props.programID);
+
+        const url = this.props.generateUrl({
```
jeremywiebe: Yeah, this is what I'm thinking.  Then the `generateUrl()` function passed into Perseus will be able to use the context to know how to build the fully-qualified URL for a python program.

packages/perseus/src/article-renderer.tsx
```diff
@@ -276,7 +277,11 @@ class ArticleRenderer
         return (
             <div className={classes}>
                 <DependenciesContext.Provider value={this.props.dependencies}>
-                    {sections}
+                    <GenerateUrlContext.Provider
```
jeremywiebe: Sorry, I completely didn't think of this when I reviewed the first time, but I think this is a bit over-engineered for what we need.  

We already have an `ApiOptions` for passing in things like this (that's how we pass in `staticUrl`). Whether it's the best pattern, I think it's better to stick to that and add this new `generateUrl` function to `ApiOptions`.  That automatically gets passed to all widgets. If there are helper components that need the function we can pass it in manually at those sites (it looks minimal). 

Then we don't need a new context and a new hook.

packages/perseus/src/widgets/video/video.tsx
jeremywiebe: Yes, I'd say so.  That's where the `context` parameter will help us. This logic is specific to Khan Academy and its content and not something that Purseus should be concerned with.

.changeset/sweet-seas-matter.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/perseus": minor
```
MikeKlemarewski: I think I'll need to change this to a major version bump

jeremywiebe: Yep, agreed!

packages/perseus/src/types.ts
```diff
@@ -159,6 +159,20 @@ type TrackInteractionArgs = {
 } & Partial<TrackingGradedGroupExtraArguments> &
     Partial<TrackingSequenceExtraArguments>;
 
+// Convert this to an enum, so webapp can swtich on context and make sure it handles all the cases.
```
jeremywiebe: I think this _is_ an enum now, right? :)

packages/perseus/src/types.ts
```diff
@@ -159,6 +159,20 @@ type TrackInteractionArgs = {
...
+export enum GenerateUrlContext {
+    IMAGE_LOADER_IMAGE_URL = "image_loader:image_url",
+    PYTHON_PROGRAM_PROGRAM_URL = "python_program:program_url",
+    VIDEO_VIDEO_URL = "video:video_url",
+}
```
jeremywiebe: Sorry, I was loose with my words. I said, "enum", but we typically don't use TypeScript enums in Perseus code ([here's why](https://khanacademy.slack.com/archives/C01AZ9H8TTQ/p1683815239266449)). 

So, we can simplify this to just a string union.
```suggestion
export type GenerateUrlContext =
    | "image_loader:image_url"
    | "python_program:program_url"
    | "video:video_url";
```

As we started learning TypeScript we found that enums don't actually give us much but have a few shortcomings, so unless we _really_ need them, we just use string unions. This has a benefit of the usage being the literal string rather than the enum name abstraction.

packages/perseus/src/widgets/python-program/python-program.tsx
```diff
@@ -39,7 +46,13 @@ class PythonProgram extends React.Component<Props> implements Widget {
     }
 
     render(): React.ReactNode {
-        const url = getUrlFromProgramID(this.props.programID);
+        let url = getUrlFromProgramID(this.props.programID);
+        url =
+            this.props.dependencies.generateUrl?.({
```
jeremywiebe: Is the question mark necessary here? It looks like the `generateUrl` key in `PerseusDependenciesV2` isn't optional, so I don't _think_ it can be optional here.

MikeKlemarewski: You're right. I think I was just tabbing through and letting cursor make some updates and missed this. Good catch!

packages/perseus/src/widgets/__testutils__/renderQuestion.tsx
```diff
@@ -30,6 +30,7 @@ export const renderQuestion = (
     apiOptions: APIOptions = Object.freeze({}),
     extraProps?: ExtraProps,
     initialUserInput?: UserInputMap,
+    dependencies?: PerseusDependenciesV2,
```
jeremywiebe: Two quick things:

1) I think we can use a default value here to provide the default instead of having to provide it on line 47. Then we don't have to mark it as optional also. 

```suggestion
    dependencies: PerseusDependenciesV2 = testDependenciesV2,
```

2) You mentioned to me in a DM about possibly converting these arguments to a single object parameter. I'm in favour of that but I think we'd want to tackle that in a separate PR as it'll cause alot of churn at all the call sites. If you're interested in doing this, I'd suggest starting the PR with the change here in this file and updating a small sampling of sites and getting review from the team. Then when the details are sussed out moving through all the usages (as I think there'll be quite a few). :)

packages/perseus/src/components/__tests__/svg-image.test.tsx
```diff
@@ -119,4 +126,25 @@ describe("SvgImage", () => {
         // Assert
         expect(container).toMatchSnapshot();
     });
+
+    it("should call the generateUrl dependency to set the img src", () => {
```
jeremywiebe, line 130: I think this is fine for testing integration here. Thanks!

packages/perseus/src/widgets/python-program/python-program.test.ts
```diff
@@ -29,5 +32,22 @@ describe("python-program widget", () => {
         expect(container).toMatchSnapshot("first mobile render");
     });
 
+    it("python program widget should call the generateUrl dependency to set the iframe src", () => {
```
jeremywiebe, line 35: Nice. Thanks for adding a test here also!

packages/perseus/src/widgets/python-program/python-program.tsx
```diff
@@ -39,7 +45,13 @@ class PythonProgram extends React.Component<Props> implements Widget {
     }
 
     render(): React.ReactNode {
-        const url = getUrlFromProgramID(this.props.programID);
+        let url = getUrlFromProgramID(this.props.programID);
+        url =
+            this.props.dependencies.generateUrl({
+                url,
+                context: "python_program:program_url",
+            }) ?? url;
```
jeremywiebe: I'd expect `generateUrl()` to return a non-optional parameter ... so do we need the `?? url` default here?

packages/perseus/src/widgets/python-program/python-program.tsx
```diff
@@ -39,7 +45,12 @@ class PythonProgram extends React.Component<Props> implements Widget {
     }
 
     render(): React.ReactNode {
-        const url = getUrlFromProgramID(this.props.programID);
+        let url = getUrlFromProgramID(this.props.programID);
```
MikeKlemarewski, line 45: We can replace `toAbsoluteUrl` within this function.

PR #2876, opened 2025-09-09
Add dated warning for schema changes

benchristel:
We can't make schema changes until November 1st because of AX.

Issue: none

## Test plan:

CI should add a comment with the new title to this PR.


handeyeco (commented):
I don't think we should do this PR.

- We were very open with AX that we were unwilling to continue doing code freezes past Pilot 2.
- We warned them months ago that it would be problematic to freeze the learner experience without freezing the content experience and advised them against it.
- In response, they promised they wouldn't edit content for active assessments.

I don't think the solution is for us to freeze development (even partially), I think it's up to them to make sure they're doing what they need to do with the tools we've provided them.

benchristel:
@handeyeco fair enough.

PR #2877, opened 2025-09-09
Sync dependencies with Frontend

handeyeco:
## Summary:

Just running `perseus/utils/sync-dependencies.ts` so I don't get the error from the Frontend bump script.


utils/sync-dependencies.ts
```diff
@@ -133,6 +133,14 @@ function main(argv: string[]) {
         # The sync-dependencies.ts script ensures that peer deps are always
         # specified as a range, and the dev deps are always pinned to the
         # version at the bottom of that range.
+        #
+        # To add a new package:
```
handeyeco, line 137: My previous commit removed this from the `pnpm-workspace.yaml` since the comment is generated from this script and someone added additional comments to the file itself.

PR #2879, opened 2025-09-09
[Interactive graph] Fix polygon angles' shape and size

nishasy:
## Summary:
There were some issues with polygon angles. Fixing them here:
- Scale the arc radius so that the angles markers are visible with large ranges.
- Use `vectorEffect="non-scaling-stroke"` so the stroke doesn't warp with the range.
- Rename `shouldDrawArcOutsidePolygon` to `isConcavePolygonVertex` for clarity.
- Update the arc path so that it doesn't incorrectly draw angles outside.

Issue: https://khanacademy.atlassian.net/browse/LEMS-2907

## Test plan:
`pnpm jest packages/perseus/src/widgets/interactive-graphs/graphs/utils.test.ts`

Storybook
`/?path=/docs/widgets-interactive-graph-editor-demo--docs#interactive-graph-polygon`

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.tsx
@@ -17,6 +17,24 @@ import type {Interval} from "mafs";
 const {clockwise} = geometry;
 const {getAngleFromVertex} = angles;
 
+/**
+ * Calculate an appropriate radius for angle arcs based on the graph range.
+ * The radius scales with the range so it's visible at all zoom levels.
+ */
+function calculateScaledRadius(range: [Interval, Interval]): number {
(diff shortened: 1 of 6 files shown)


packages/perseus/src/widgets/interactive-graphs/graphs/components/angle-indicators.tsx
```diff
@@ -73,21 +77,21 @@ export const PolygonAngle = ({
...
                 vertex={[x2, y2]}
                 end={[x3, y3]}
+                nonScalingStroke={true}
             />
         ) : null;
     }
 
     // Note: Need to pass in endpoints in a clockwise order for the cross product.
-    const isOutside = shouldDrawArcOutsidePolygon(centerPoint, endPoints);
+    const isConcave = isConcavePolygonVertex(centerPoint, endPoints);
```
SonicScrewdriver, line 86: Muuuch better name.

PR #2882, opened 2025-09-10
[Radio] Fix ScrollableView when in RTL the scroll buttons is not working properly

ivyolamit:
## Summary:
Fix ScrollableView when in RTL the scroll buttons are not working properly

Issue: LEMS-3480

https://github.com/user-attachments/assets/03794eec-bc48-46c0-8ac7-bf2b7cd0b8b0

## Test plan:
1. Navigate to RadioNew > Widget Demo > Single Select With Scroll in Perseus documentation
2. Confirm that the scrolls works as expected when in LTR
3. Confirm that the scrolls works as expected when in RTL

### DIFF
--- a/packages/perseus/src/components/scrollable-view.tsx
+++ b/packages/perseus/src/components/scrollable-view.tsx
@@ -91,12 +91,12 @@ function ScrollableArea({
             setIsScrolling(true);
 
             // Get current RTL state dynamically to ensure it's up to date
-            const currentIsRtl =
+            const contentIsRtl =
                 getComputedStyle(containerRef.current).direction === "rtl";
 
             const scrollNegative =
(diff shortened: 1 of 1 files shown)


packages/perseus/src/components/scrollable-view.tsx
```diff
@@ -91,9 +90,13 @@ function ScrollableArea({
             // Set scrolling state to prevent rapid clicks
             setIsScrolling(true);
 
+            // Get current RTL state dynamically to ensure it's up to date
+            const currentIsRtl =
```
mark-fitzgerald: Nit: Maybe name it `contentIsRtl` or `containerIsRtl`?

SonicScrewdriver (approved):
Works well, thanks for the fix!

PR #2884, opened 2025-09-10
[Image] Add image-widget-upgrade feature flag

ivyolamit:
## Summary:
Add image-widget-upgrade feature flag so image widget team can work behind the flag without affecting prod.

Issue: LEMS-3519

## Test plan:

### DIFF
--- a/packages/perseus-core/src/feature-flags.test.ts
+++ b/packages/perseus-core/src/feature-flags.test.ts
@@ -1,0 +1,28 @@
+import {isFeatureOn} from "./feature-flags";
+
+describe("isFeatureOn", () => {
+    it("returns true when the flag is enabled", () => {
+        const props = {apiOptions: {flags: {"new-radio-widget": true}}};
+        expect(isFeatureOn(props, "new-radio-widget")).toBe(true);
+    });
+
+    it("returns false when the flag is disabled", () => {
+        const props = {apiOptions: {flags: {"new-radio-widget": false}}};
+        expect(isFeatureOn(props, "new-radio-widget")).toBe(false);
+    });
+
+    it("returns false when the flag is missing", () => {
+        const props = {apiOptions: {flags: {}}};
+        expect(isFeatureOn(props, "non-existent-flag")).toBe(false);
+    });
+
+    it("returns false when flags is undefined", () => {
+        const props = {apiOptions: {}};
+        expect(isFeatureOn(props, "new-radio-widget")).toBe(false);
+    });
+
+    it("returns false when apiOptions is undefined", () => {
+        const props = {};
+        expect(isFeatureOn(props, "new-radio-widget")).toBe(false);
+    });
+});
--- a/testing/feature-flags-util.test.ts
+++ b/testing/feature-flags-util.test.ts
@@ -1,0 +1,24 @@
+import {getFeatureFlags} from "./feature-flags-util";
+
+describe("getFeatureFlags", () => {
+    it("returns default flags when no overrides are provided", () => {
+        const flags = getFeatureFlags();
+        expect(flags["new-radio-widget"]).toBe(false);
+        expect(flags["image-widget-upgrade"]).toBe(false);
+    });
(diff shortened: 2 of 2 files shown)


packages/perseus/src/widgets/radio/__docs__/multiple-choice-demo.new.stories.tsx
```diff
@@ -88,9 +89,7 @@ const applyStoryArgs = (args: StoryArgs): PerseusItem => {
             widgets: {},
         },
         apiOptions: {
-            flags: {
```
ivyolamit, line 91: Had to update these tests related to radio since it's causing a lint error. To avoid similar issue in the future when adding a new flag, i added the `getFeatureFlags` util function.

packages/perseus-core/src/feature-flags.ts
```diff
@@ -7,7 +7,21 @@
  * 2. Do cleanup unused feature flags
  * 3. When no active feature flag is available value will be:
  *    export const PerseusFeatureFlags = [];
+ * 4. Also update the testing/feature-flags-util.ts for testing purpose
  */
-const PerseusFeatureFlags = ["new-radio-widget"] as const;
+const PerseusFeatureFlags = [
+    "new-radio-widget", // TODO(LEMS-2994): cleanup feature flag
+    "image-widget-upgrade", // TODO(LEMS-3521): cleanup feature flag
```
ivyolamit, line 14: @nishasy here's the feature flag for the `image-widget-upgrade`
you can use:
```
getFeatureFlags(props, "image-widget-upgrade")
```

let me know if i missed anything?

testing/feature-flags-util.ts
```diff
@@ -0,0 +1,15 @@
...
+
+/** Utility to get feature flags with optional overrides for testing.
+ *  sample usage:
+ *  getFeatureFlags({"image-widget-upgrade": true})
+ */
+export function getFeatureFlags(
+    overrides: Partial<typeof DEFAULT_FEATURE_FLAGS> = {},
+) {
+    return {...DEFAULT_FEATURE_FLAGS, ...overrides};
+}
```
nishasy, line 15: Ooo I really like this!

packages/perseus-core/src/feature-flags.ts
nishasy: That makes sense to me!

packages/perseus-core/src/feature-flags.ts
```diff
@@ -7,7 +7,21 @@
...
+
+/**
+ * Checks if a feature flag is enabled in the given props.
+ */
+export function isFeatureOn(
+    props: {apiOptions?: {flags?: Record<string, boolean>}},
+    flag: string,
+): boolean {
+    return props.apiOptions?.flags?.[flag] ?? false;
+}
```
nishasy, line 27: What do you think about moving the `getFeatureFlags` function to this file?

And then you can import it from `@khanacademy/perseus-core` rather than doing
```
import {getFeatureFlags} from "../../../../../../testing/feature-flags-util";
```

nishasy: Suggestion: Can you test `isFeatureOn` and `getFeatureFlags`? You could make a `feature-flag.test.ts` file in the same folder as `feature-flag.ts`.

ivyolamit: > What do you think about moving the `getFeatureFlags` function to this file?
> 
> And then you can import it from `@khanacademy/perseus-core` rather than doing
> 
> ```
> import {getFeatureFlags} from "../../../../../../testing/feature-flags-util";
> ```

I initially considered it, but the reason I retained it in this file is for easy code maintainability so it will be close to the feature flag definitions, since adding a feature flag doesn't happen all the time, it will be easier for the engineer adding a new feature flag to see we have that function available. Also the things in testing folders looks like are used for testing or storybook.

PR #2885, opened 2025-09-10
[Image] Add missing export in perseus-core

ivyolamit:
## Summary:
Add missing export in perseus-core

Issue: LEMS-3519

## Test plan:


nishasy (approved):
Thank you!

PR #2887, opened 2025-09-11
Add test for answerless to answerful transition

handeyeco:
## Summary:
We need to be able to seamlessly transition from "answerless" PerseusItems (pre-attempt) to "answerful" PerseusItems (post-attempt). Unfortunately this wasn't as easy as it sounds. However it seems to be working now so I added a test to make sure we don't regress.

Issue: [LEMS-3113](https://khanacademy.atlassian.net/browse/LEMS-3113)

## Test plan:
- This is the test plan

### DIFF
--- a/packages/perseus/src/__tests__/server-item-renderer.test.tsx
+++ b/packages/perseus/src/__tests__/server-item-renderer.test.tsx
@@ -646,10 +646,8 @@ describe("server item renderer", () => {
             ).toBeInTheDocument();
 
             // make sure we're showing rationales
-            for (let i = 0; i < 4; i++) {
-                expect(
-                    screen.getByText(`Rationale ${i + 1}`),
-                ).toBeInTheDocument();
+            for (let i = 1; i <= 4; i++) {
+                expect(screen.getByText(`Rationale ${i}`)).toBeInTheDocument();
             }
         });
     });


packages/perseus/src/__tests__/test-utils.tsx
```diff
@@ -45,5 +46,30 @@ export const renderQuestion = (
     if (!renderer) {
         throw new Error(`Failed to render!`);
     }
-    return {container, renderer};
+    const renderAgain = (
```
handeyeco, line 49: We need to consolidate our two `renderQuestions`...this one didn't have the ability to rerender.

packages/perseus/src/__tests__/server-item-renderer.test.tsx
```diff
@@ -518,4 +524,133 @@ describe("server item renderer", () => {
...
+                            ],
+                        },
+                    },
+                },
+            });
+
+            return generateTestPerseusItem({question});
+        }
+
+        it("can transition between answerless and answerful data correctly", async () => {
```
handeyeco, line 576: This is admittedly a big test. I personally don't mind that. I suppose I could collapse some of this down using loops, but I kind of like how clear what I'm doing is by using minimal abstraction. Open to push-back though.

packages/perseus/src/__tests__/server-item-renderer.test.tsx
```diff
@@ -518,4 +524,133 @@ describe("server item renderer", () => {
...
+                }),
+            ).toBeInTheDocument();
+            expect(
+                screen.getByRole("checkbox", {
+                    name: "(Choice D, Checked, Correct) Right 2",
+                }),
+            ).toBeInTheDocument();
+
+            // make sure we're showing rationales
+            for (let i = 0; i < 4; i++) {
```
jeremywiebe: **nit**: We could just loop over the range of values we actually need. :) 
```suggestion
            for (let i = 1; i <= 4; i++) {
```

packages/perseus/src/__tests__/server-item-renderer.test.tsx
```diff
@@ -518,4 +524,133 @@ describe("server item renderer", () => {
...
+                                    correct: true,
+                                    id: "13-13-13-13-13",
+                                },
+                            ],
+                        },
+                    },
+                },
+            });
+
+            return generateTestPerseusItem({question});
```
jeremywiebe, line 573: FWIW, couldn't we just use the `renderQuestion` for `Renderer` instead? We aren't exercising hints in this test so we don't need `ServerItemRenderer`.  🤷‍♂️

handeyeco: Part of what this is testing is how we handle user input. Now that Renderer doesn't handle user input, I feel like the test would be most useful against ServerItemRenderer (which wraps UserInputManager which provides user input to Renderer).

PR #2892, opened 2025-09-11
[Image] Add helpful comment to lint the front-end change related to adding a new feature flag

ivyolamit:
## Summary:
Add helpful comment to lint the front-end change related to adding a new feature flag

Issue: LEMS-3519

## Test plan:

### DIFF
--- a/packages/perseus-core/src/feature-flags.ts
+++ b/packages/perseus-core/src/feature-flags.ts
@@ -4,10 +4,11 @@
  *    This will be the definition of the active feature flags in Perseus
  *    that will hide features/changes under development. This will be used
  *    by flags of APIOptions type.
- * 2. Do cleanup unused feature flags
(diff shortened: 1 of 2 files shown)


packages/perseus-core/src/feature-flags.ts
```diff
@@ -4,10 +4,11 @@
  *    This will be the definition of the active feature flags in Perseus
  *    that will hide features/changes under development. This will be used
  *    by flags of APIOptions type.
- * 2. Do cleanup unused feature flags
- * 3. When no active feature flag is available value will be:
+ * 2. Also add the new flag in front-end repo's perseus-feature-flags.ts file
```
jeremywiebe, line 7: Thought: when we originally introduced feature flags in Perseus, I thought we'd organized the flags so that you couldn't have missing flags when integrating perseus. Is that no longer the case?

PR #2895, opened 2025-09-12
docs(changeset): Fixing several iOS app bugs related to Interactive Graph

SonicScrewdriver:
## Summary:
This PR fixes two key bugs regarding our Interactive Graph Widget in our iOS Mobile Application:

**1. iOS Safari Hover Styling Issue** 
This is resolved with a simple CSS property to turn off iOS' native highlighting of links/clickables. 
If you want some fun history, the fix was actually found in [a 15 year old stackoverflow](https://stackoverflow.com/questions/1549228/mobile-safari-svg-rendering-issues-with-raphaeljs) from someone specifically asking about making an Interactive Graph with Raphael. _What a coincidence!_ ;)
I'll update our documentation so that this little fact doesn't disappear again 

**2. iOS WebView Phantom Click Issue**
This is resolved by adding a bit of a debouncer to the click handler on our Unlimited graphs when we're dragging a point. The reason this is necessary is because of how Safari WebViews handle touch-to-mouse emulation. It seems that the Safari browser has better gesture recognition to suppress phantom clicks, which is compounded by the fact that the UIWebView has a longer delay before single taps activate links/buttons (350ms) in order to detect if the user is doing a double tap gesture. 

If you'd like more information on either of these issues, take a look at the comments on the Jira ticket! 

Note: I've also added Wonderblocks Timing, as I didn't want to add new "setTimeout" tech debt. 

Issue: LEMS-2873

## Test plan:
- Manually tested in Mobile App on iOS device to confirm that both issues are resolved.

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/point.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/point.tsx
@@ -1,6 +1,5 @@
 import {useTimeout} from "@khanacademy/wonder-blocks-timing";
 import * as React from "react";
-import {useState} from "react";
 
 import {actions} from "../reducer/interactive-graph-action";
 import useGraphConfig from "../reducer/use-graph-config";
@@ -103,7 +102,7 @@ function UnlimitedPointGraph(statefulProps: StatefulProps) {
     // When users drag a point on iOS Safari, the browser fires a click event after the mouseup
     // at the original click location, which would add an unwanted new point. We track drag
     // state and delay clearing it to block these phantom clicks (LEMS-2873)
-    const [isCurrentlyDragging, setIsCurrentlyDragging] = useState(false);
+    const [isCurrentlyDragging, setIsCurrentlyDragging] = React.useState(false);
     const dragEndCallbackTimer = useTimeout(
         () => setIsCurrentlyDragging(false),
         400, // Safari Webkit has up to a 350ms delay before a click event is fired


packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.tsx
```diff
@@ -26,6 +26,7 @@ type Props = {
     sequenceNumber?: number;
     onBlur?: (event: React.FocusEvent) => unknown;
     onClick?: () => unknown;
+    onDragEnd?: () => unknown;
```
SonicScrewdriver, line 29: I'm not the biggest fan of adding a new callback, but it seemed like the best way to handle this situation. I'm open to suggestions!

packages/perseus/src/widgets/interactive-graphs/graphs/point.tsx
```diff
@@ -1,4 +1,6 @@
+import {useTimeout} from "@khanacademy/wonder-blocks-timing";
 import * as React from "react";
+import {useState} from "react";
```
nishasy: Nit: We're already importing React. I would remove this extra import, and use `React.useState()` below.

nishasy (commented):
Hello! Do you have a ZND for this PR for testing in prod within different environments?

packages/perseus/src/widgets/interactive-graphs/graphs/point.tsx
SonicScrewdriver: Great point! I think I just auto imported without checking 😅

packages/perseus/src/widgets/interactive-graphs/graphs/point.tsx
```diff
@@ -98,6 +99,15 @@ function UnlimitedPointGraph(statefulProps: StatefulProps) {
     const {dispatch, graphConfig, pointsRef, top, left} = statefulProps;
     const {coords} = statefulProps.graphState;
 
+    // When users drag a point on iOS Safari, the browser fires a click event after the mouseup
+    // at the original click location, which would add an unwanted new point. We track drag
+    // state and delay clearing it to block these phantom clicks (LEMS-2873)
```
nishasy, line 104: Thanks for the comments

nishasy (approved):
Thank you for this fix! I confirmed the experience works as it should now in the ZND.

PR #2899, opened 2025-09-15
[Color - NAW] Refactor Definiition widget to use semantic colors

Myranae:
## Summary:
To allow theming with Perseus, we are updating a number of key widgets to use WonderBlocks semantic colors. This work updates the Definition widget with semantic color tokens.

Issue: LEMS-3492

## Test plan:
- Manually check changes using Storybook theme button
- Add regression tests and confirm update is as intended
- Confirm all checks pass
- Will be QE tested

### DIFF
--- a/.storybook/modes.ts
+++ b/.storybook/modes.ts
@@ -1,0 +1,36 @@
+export const allModes = {
+    // Responsive modes
+    small: {
+        viewport: "small",
+    },
+    medium: {
+        viewport: "medium",
+    },
+    large: {
+        viewport: "large",
+    },
+    chromebook: {
+        viewport: "chromebook",
+    },
+    // Theming
+    themeDefault: {
+        theme: "default",
+    },
+    themeThunderBlocks: {
+        theme: "thunderblocks",
+    },
+    // NOTE: This will go away when we fully remove the light variants.
+    dark: {
+        background: "darkBlue",
+    },
+    // Accessibility
+    "themeDefault rtl": {
+        theme: "default",
+        direction: "rtl",
+    },
+};
+
+export const themeModes = {
+    default: allModes.themeDefault,
+    thunderblocks: allModes.themeThunderBlocks,
+};
--- a/packages/perseus/src/widgets/definition/__docs__/definition-interaction-regression.stories.tsx
+++ b/packages/perseus/src/widgets/definition/__docs__/definition-interaction-regression.stories.tsx
@@ -1,9 +1,10 @@
-import type {PerseusItem} from "@khanacademy/perseus-core";
 import {generateTestPerseusItem} from "@khanacademy/perseus-core";
 import * as React from "react";
 
 import {ServerItemRendererWithDebugUI} from "../../../../../../testing/server-item-renderer-with-debug-ui";
 import {question} from "../definition.testdata";
+
+import type {PerseusItem} from "@khanacademy/perseus-core";
 import type {Meta} from "@storybook/react-vite";
 
 type StoryArgs = {
--- a/packages/perseus/src/widgets/definition/__docs__/definition-interaction-regression.stories.tsx
+++ b/packages/perseus/src/widgets/definition/__docs__/definition-interaction-regression.stories.tsx
@@ -1,0 +1,96 @@
+import type {PerseusItem} from "@khanacademy/perseus-core";
+import {generateTestPerseusItem} from "@khanacademy/perseus-core";
+import * as React from "react";
+
+import {ServerItemRendererWithDebugUI} from "../../../../../../testing/server-item-renderer-with-debug-ui";
+import {question} from "../definition.testdata";
+import type {Meta} from "@storybook/react-vite";
+
+type StoryArgs = {
+    // Story Option
+    item: PerseusItem;
+    // Definition Options
+    static: boolean;
+    // Testing Options
+    startAnswerless: boolean;
+} & Pick<
+    React.ComponentProps<typeof ServerItemRendererWithDebugUI>,
+    "reviewMode" | "showSolutions"
+>;
+
+export default {
+    title: "Widgets/Definition/Visual Regression Tests/Interactive",
+    component: ServerItemRendererWithDebugUI,
+    tags: ["!autodocs"],
+    parameters: {
+        docs: {
+            description: {
+                component:
+                    "Regression tests for the definition widget that DO need some sort of interaction to test, which will be used with Chromatic. Stories are displayed on their own page.",
+            },
+        },
+        chromatic: {disableSnapshot: false},
+    },
+    args: {
+        static: false,
+        startAnswerless: false,
+        reviewMode: false,
+        showSolutions: "none",
+        item: generateTestPerseusItem({
+            question,
+        }),
+    } satisfies StoryArgs,
+    argTypes: {
+        showSolutions: {
+            options: ["none", "all", "selected"],
+            control: {
+                type: "select",
+            },
+        },
+    },
+    render: (args: StoryArgs) => (
+        <ServerItemRendererWithDebugUI
+            item={applyStoryArgs(args)}
+            reviewMode={args.reviewMode}
+            showSolutions={args.showSolutions}
+            startAnswerless={args.startAnswerless}
+        />
+    ),
+} satisfies Meta<StoryArgs>;
+
+const applyStoryArgs = (args: StoryArgs): PerseusItem => {
+    const storyItem = {
+        ...args.item,
+        question: {
+            ...args.item.question,
+            widgets: {},
+        },
+    };
+    for (const [widgetId, widget] of Object.entries(
+        args.item.question.widgets,
+    )) {
+        storyItem.question.widgets[widgetId] = {...widget, static: args.static};
+    }
+
+    return storyItem;
+};
+
+export const FocusedState = {
+    play: async ({canvas}) => {
+        // eslint-disable-next-line testing-library/prefer-screen-queries
+        const definitionTrigger = canvas.getByRole("button", {
+            name: "the Pequots",
+        });
+        definitionTrigger.focus();
+    },
+};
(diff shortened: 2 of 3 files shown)


packages/perseus/src/widgets/definition/definition.tsx
```diff
@@ -92,10 +93,10 @@ class Definition extends React.Component<DefinitionProps> implements Widget {
 
 const styles = {
     tooltipBody: {
-        color: color.offBlack,
-        fontSize: 20,
-        fontWeight: 500,
-        lineHeight: "30px",
+        color: semanticColor.core.foreground.neutral.strong,
+        fontSize: "2rem",
```
Myranae: Still need to discuss this with our designer as this is one attribute that does have semantic tokens, just not ones that provide this value. Might need to add new token, change the design, or just go with this.

packages/perseus/src/widgets/definition/definition.tsx
```diff
@@ -92,10 +93,10 @@ class Definition extends React.Component<DefinitionProps> implements Widget {
...
+        fontWeight: font.weight.medium,
+        lineHeight: 1.5,
```
Myranae: This also may need a new token, though this value may work as it is. Will also confirm with designer.

packages/perseus/src/widgets/definition/definition.tsx
Myranae: Not adding new tokens at the moment, so decision was to normalize it to the largest body token available: `font.body.size.medium`

packages/perseus/src/widgets/definition/definition.tsx
Myranae: Confirmed correct values with designer

.storybook/modes.ts
```diff
@@ -0,0 +1,36 @@
+export const allModes = {
```
Myranae, line 1: The file content here was copied from [WonderBlocks](https://github.com/Khan/wonder-blocks/blob/main/.storybook/modes.ts)

packages/perseus/src/widgets/definition/__docs__/definition-interactions-regression.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+    component: ServerItemRendererWithDebugUI,
+    tags: ["!autodocs"],
+    parameters: {
+        docs: {
+            description: {
+                component:
+                    "Regression tests for the definition widget that DO need some sort of interaction to test, which will be used with Chromatic. Stories are displayed on their own page.",
+            },
+        },
+        chromatic: {disableSnapshot: false, modes: themeModes},
```
Myranae, line 20: modes: themeModes is what allows Chromatic to do a snapshot for the story in both the Classic/OG theme and the Thunderblocks/Share Your Learning theme.

packages/perseus/src/widgets/definition/definition.tsx
```diff
@@ -92,10 +93,10 @@ class Definition extends React.Component<DefinitionProps> implements Widget {
...
 const styles = {
     tooltipBody: {
-        color: color.offBlack,
-        fontSize: 20,
-        fontWeight: 500,
-        lineHeight: "30px",
+        color: semanticColor.core.foreground.neutral.strong,
+        fontSize: font.body.size.medium,
+        fontWeight: font.weight.medium,
+        lineHeight: font.body.lineHeight.large,
```
Myranae, line 99: `font.body.lineHeight.large` is 22px. This change (from 30px to 22px) was discussed with our designer and approved as the best course of action for now since we are not adding new tokens. 

The `fontSize` is also different; 16px instead of 20px. This makes the text smaller than the body font, but is a more future-proof decision. Smaller text size makes the popover smaller and therefore easier to fit in a smaller view as well.

packages/perseus/src/widgets/definition/definition.tsx
```diff
@@ -72,10 +72,11 @@ class Definition extends React.Component<DefinitionProps> implements Widget {
                             {({hovered, focused, pressed}) => (
                                 <span
                                     style={{
-                                        color: color.blue,
+                                        color: semanticColor.core.foreground
+                                            .instructive.default,
                                         borderBottom:
                                             hovered || focused || pressed
-                                                ? `2px solid ${color.blue}`
+                                                ? `2px solid ${semanticColor.core.border.instructive.default}`
```
beaesguerra, line 79: If you're up for using more WB tokens, there are also tokens for border width! https://khan.github.io/wonder-blocks/?path=/docs/packages-tokens-border--docs#border-width

.storybook/modes.ts
```diff
@@ -0,0 +1,36 @@
...
+    // Theming
+    themeDefault: {
+        theme: "default",
+    },
+    themeThunderBlocks: {
+        theme: "thunderblocks",
+    },
+    // NOTE: This will go away when we fully remove the light variants.
+    dark: {
+        background: "darkBlue",
```
beaesguerra: In WB, we recently changed this value to `neutralStrong` so that it would work for any theme, and we're moving away from using the `darkBlue` background!

`neutralStrong` is mapped to semantic colors in the [Storybook backgrounds config](https://github.com/Khan/wonder-blocks/blob/0995834b05dac0df51085754efaeee2f5ee9efb0/.storybook/preview.tsx#L80-L96)

I see in Perseus, the primitive background colors for the classic theme is used as background options in Storybook: 
https://github.com/Khan/perseus/blob/37eeac97844467b249910e1f9b0c87c3404f3053/.storybook/preview.tsx#L123-L129

packages/perseus/src/widgets/definition/__docs__/definition-interactions-regression.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+        }),
+    },
+    play: async ({canvas}) => {
+        // eslint-disable-next-line testing-library/prefer-screen-queries
+        const definitionTrigger = canvas.getByRole("button", {
+            name: "the Pequots",
+        });
+        definitionTrigger.focus();
+    },
+};
```
beaesguerra, line 39: We use a [Storybook add on for pseudo states](https://storybook.js.org/addons/storybook-addon-pseudo-states) so that we can easily trigger certain states of interactive elements like focus, hover, active etc! Here is an [example of how we can configure the pseudo states](https://github.com/Khan/wonder-blocks/blob/0995834b05dac0df51085754efaeee2f5ee9efb0/__docs__/wonder-blocks-styles/action-styles-variants.stories.tsx#L245-L265) in stories. This is also what we use for [StateSheet stories](https://khan.github.io/wonder-blocks/?path=/story/packages-button-testing-snapshots-button--state-sheet-story) that cover all the different variants and states.

One limitation to this though is it will only trigger CSS pseudo-element styles, so styles relying on JS won't be triggered by this add on! So styling applied using `:hover` would be shown, however, styling applied by keeping track of `onMouseEnter`/`onMouseLeave` events won't

beaesguerra (approved):
Nice work, the semantic color token updates look good to me! 

I left a few comments around things I've found helpful in WB, in case it could be helpful in Perseus too (no changes necessary though)!

packages/perseus/src/widgets/definition/__docs__/definition-interactions-regression.stories.tsx
mark-fitzgerald: That's a nice add-on! While this project is primarily for swapping colors and font-sizes, we will (hopefully in the coming months) be switching to standard CSS files. At that time, we would then be able to fully utilize this add-on. I've added it to the [ticket that would be able to investigate using it](https://khanacademy.atlassian.net/browse/LEMS-3525).

packages/perseus/src/widgets/definition/__docs__/definition-initial-state-regression.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+    args: {
+        item: generateTestPerseusItem({question}),
+    },
+};
+
+export const Article = (): React.ReactNode => {
+    return (
+        <ArticleRenderer
+            json={article}
+            useNewStyles
```
mark-fitzgerald: What is this for? I don't see it used in the Article Renderer anywhere.

mark-fitzgerald (approved):
Yay! First migration to semantic tokens!!

packages/perseus/src/widgets/definition/definition.tsx
Myranae: Thanks for pointing this out! I think after we do colors and fonts, we'll review the remaining aspects of Perseus that need to be switched to WB, including borders, sizing, and some of the CSS not associated with specific widgets. :)

packages/perseus/src/widgets/definition/__docs__/definition-initial-state-regression.stories.tsx
Myranae: Looks like it is part of the props for article renderer, but it is deprecated. 
```
        // Whether to use the new Bibliotron styles for articles
        /**
         * @deprecated Does nothing
         */
        useNewStyles: boolean;
```
I'll go ahead and remove it from both the definition stories page (where I found it) and this story. Thanks!

packages/perseus/src/widgets/definition/__docs__/definition-interactions-regression.stories.tsx
Myranae: Thanks for sharing this info, Bea, and thanks for adding it to a ticket for future review, Mark!

.storybook/modes.ts
Myranae: Thanks for pointing this out! I updated the background value here from "darkBlue" to "neutralStrong." I'll take a look at the preview component file and see about either switching to semanticColor instead of primitive color or just copying what WB has for values.

PR #2905, opened 2025-09-16
Downgrade our JS language support level from ES2021 to ES2020

jeremywiebe:
## Summary:

A while ago we (I?) upgraded our JS langugage support level to a more modern version (ES2021) based on Khan Academy's officially supported browser list. Later, code was merged that took advantage of one of these newly supported functions ([`String.replaceAll()`](https://caniuse.com/?search=replaceAll)). 

However, now we are seeing errors in production from learners who's browser is just _slightly_ out of date. 

Instead of cutting those learners off or giving them a broken experience, we can downgrade our JS language level from ES2021 to ES2020. This is a minor inconvenience for Perseus engineers but allows us to not break older browsers as quickly. 

Issue: "none"

## Test plan:

`pnpm typecheck` - passes!

### DIFF
--- a/tsconfig-common.json
+++ b/tsconfig-common.json
@@ -3,6 +3,11 @@
 {
     "compilerOptions": {
         /* Language and Environment */
+        /* This is not quite the latest we _could_ support based on Khan
+         * Academy's supported browser list. We intentionally choose to use a
+         * slightly older JS version to allow the JS to run, unbroken in
+         * recently unsupported browsers.
+         */
         "target": "ES2020",
         "jsx": "preserve",
         /* Modules */


.changeset/chatty-years-applaud.md
```diff
@@ -0,0 +1,2 @@
+---
```
jeremywiebe, line 1: Changeset is empty because this is a tooling-only change. No production code will change because even though we're changing the target language level, that doesn't even affect the code that's generated. So this is completely invisible to end users of the Perseus libraries.

tsconfig-common.json
```diff
@@ -3,7 +3,7 @@
 {
     "compilerOptions": {
         /* Language and Environment */
-        "target": "ES2021",
+        "target": "ES2020",
```
ivyolamit, line 11: can we add an info here about we are being intentional on downgrading this to support our learners? 🙏🏼

jeremywiebe: Good idea. I'll add a comment!

ivyolamit (approved):
Thanks for the update, I've landed the other PR, that should fix the lint error you have 😉

PR #2910, opened 2025-09-17
docs(changeset): Fixing several issues related to User Input in our Explanation and Number Line Widgets

SonicScrewdriver:
## Summary:
[resolving-the-unexplainable] The goal of this PR is to solve the several issues that are causing LEMS-3474. 

Unfortunately, our Explanation Widget allows scorable widgets to be nested inside its content. This is not a very well known feature of the widget, and we missed adding the User Input Manager to it as a result. 

This PR adds in the User Input Manager to the Explanation widget, while also fixing several small issues related to how the Number Line Widget accesses the User Input. 

Issue: LEMS-3474

## Test plan:
- Manual testing 
- Existing tests pass

### DIFF
--- a/packages/perseus-editor/src/widgets/number-line-editor.tsx
+++ b/packages/perseus-editor/src/widgets/number-line-editor.tsx
@@ -359,7 +359,7 @@ class NumberLineEditor extends React.Component<Props> {
                         <div className="perseus-widget-left-col">
                             <Checkbox
                                 label="Show tick controller"
-                                checked={this.props.isTickCtrl}
+                                checked={!!this.props.isTickCtrl}
                                 onChange={(value) => {
                                     this.props.onChange({isTickCtrl: value});
                                 }}
--- a/packages/perseus/src/widgets/number-line/number-line.tsx
+++ b/packages/perseus/src/widgets/number-line/number-line.tsx
@@ -261,7 +261,7 @@ class NumberLine extends React.Component<Props, State> implements Widget {
 
     isValid: () => boolean = () => {
         const range = this.props.range;
-        let initialX = this.props.userInput.numLinePosition;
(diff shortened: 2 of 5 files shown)


packages/perseus/src/widgets/number-line/number-line.tsx
```diff
@@ -261,7 +261,7 @@ class NumberLine extends React.Component<Props, State> implements Widget {
 
     isValid: () => boolean = () => {
         const range = this.props.range;
-        let initialX = this.props.userInput.numLinePosition;
+        let initialX = this.props.userInput?.numLinePosition;
```
SonicScrewdriver, line 264: I used optional chaining here, but I'm wondering if we need to be setting defaults. 🤔

packages/perseus-editor/src/widgets/number-line-editor.tsx
```diff
@@ -359,7 +359,7 @@ class NumberLineEditor extends React.Component<Props> {
                         <div className="perseus-widget-left-col">
                             <Checkbox
                                 label="Show tick controller"
-                                checked={this.props.isTickCtrl}
+                                checked={!!this.props.isTickCtrl}
```
SonicScrewdriver, line 362: This was throwing a bunch of errors in the editor due to the value initially being undefined, and seemed like an easy fix. Now the "Show Tick Control" checkbox doesn't load in an indeterminate state, which is better UX.

packages/perseus/src/widgets/number-line/number-line.tsx
handeyeco: Maybe we need to handle it better in `getStartUserInput` for NumberLine; or is this something that we're seeing in the editing experience?

handeyeco (approved):
Nice, thank you!

PR #2915, opened 2025-09-23
[Image][Radio] Fix editor image preview for Graphie images

nishasy:
## Summary:
In Image widget, the preview was intentionally removed for Graphie images.
In the Radio widget image editor, the preview was broken for Graphie images.

Presumably, it was originally implemented this way because the simple `` tag
can't load graphies. However, we can remedy this by using the `SvgImage` component
from Perseus within our editor instead.

Issue: https://khanacademy.atlassian.net/browse/LEMS-3565

| Widget | Before | After |
| --- | --- | --- |
| Image | Screenshot 2025-09-23 at 1 36 26 PM | Screenshot 2025-09-23 at 1 36 32 PM |
| Radio | Screenshot 2025-09-23 at 1 37 16 PM | Screenshot 2025-09-23 at 1 37 22 PM |

## Test plan:
Storybook
- Go to editor page demo (`/?path=/story/editors-editorpage--demo`)
- Add an image widget
- Load in a graphie (ex: `web+graphie://ka-perseus-graphie.s3.amazonaws.com/c906d5a2bbdc3c8eddb3022f297b3b6daf9fb2f2`)
- Confirm the preview works and the size is right
- Load in a normal image (ex: `https://cdn.kastatic.org/ka-content-images/61831c1329dbc32036d7dd0d03e06e7e2c622718.jpg`)
- Confirm the preview works and the size is right
- Repeat the above for radio widget options

### DIFF
--- a/packages/perseus-editor/src/widgets/image-editor/image-settings.tsx
+++ b/packages/perseus-editor/src/widgets/image-editor/image-settings.tsx
@@ -59,6 +59,7 @@ export default function ImageSettings({
                 width={backgroundImage.width}
                 height={backgroundImage.height}
             />
+
             {/* Dimensions */}
             <div className={styles.dimensionsContainer}>
                 <HeadingXSmall
@@ -74,6 +75,7 @@ export default function ImageSettings({
                 </HeadingXSmall>
                 {dimensionString}
             </div>
+
             {/* Alt text */}
             <div className={styles.labelWithInfoTip}>
                 <HeadingXSmall tag="label" htmlFor={altId}>
@@ -91,6 +93,7 @@ export default function ImageSettings({
                 onChange={(value) => onChange({alt: value})}
                 style={textAreaStyle}
             />
+
             {imageUpgradeFF && (
                 <>
                     {/* Long Description */}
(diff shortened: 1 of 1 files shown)


nishasy:
NOTE: This might be easier to review with "Hide whitespace" turned on

image

packages/perseus-editor/src/widgets/radio/radio-option-content-and-image-editor.tsx
```diff
@@ -199,10 +202,9 @@ export const RadioOptionContentAndImageEditor = (props: Props) => {
                         paddingBlockEnd: sizing.size_120,
                     }}
                 >
-                    <img
+                    <SvgImage
                         src={image.url}
-                        alt={image.altText}
-                        style={{marginBlockEnd: sizing.size_080}}
+                        alt={`Preview: ${image.altText ?? "No alt text"}`}
```
mark-fitzgerald, line 207: Does this not need the `height` and `width` properties specified, like it is in the image editor?

nishasy: The image editor does weird things where it automatically calculates the height and width, and it makes the image too big without it.

But for the radio editor, it looks like these are not defined and get calculated automatically within the `SvgImage` component.

Although now that I'm typing this out, I think it might have something to do with the containers that these previews are rendered within, within the editors 🤔

PR #2918, opened 2025-09-23
[Image] Add decorative props in Image Widget

ivyolamit:
## Summary:
Add decorative props in Image Widget

Issue: LEMS-3450

## Test plan:

### DIFF
--- a/packages/perseus/src/widgets/image/image.tsx
+++ b/packages/perseus/src/widgets/image/image.tsx
@@ -33,6 +33,45 @@ export const ImageComponent = (props: ImageWidgetProps) => {
         return null;
     }
 
+    const svgImage = (
+        <AssetContext.Consumer>
+            {({setAssetStatus}) => (
+                <SvgImage
+                    src={backgroundImage.url!}
+                    width={backgroundImage.width}
+                    height={backgroundImage.height}
+                    preloader={apiOptions.imagePreloader}
+                    extraGraphie={{
+                        box: box,
+                        range: range,
+                        labels: labels,
+                    }}
+                    trackInteraction={trackInteraction}
+                    zoomToFullSizeOnMobile={apiOptions.isMobile}
+                    constrainHeight={apiOptions.isMobile}
+                    allowFullBleed={apiOptions.isMobile}
+                    renderSpacer={false}
+                    alt={decorative || caption === alt ? "" : alt}
+                    setAssetStatus={setAssetStatus}
+                />
+            )}
+        </AssetContext.Consumer>
+    );
+
+    // Early return for decorative images
+    if (decorative) {
+        return (
+            <figure
+                className="perseus-image-widget"
+                style={{
+                    maxWidth: backgroundImage.width,
+                }}
+            >
+                {svgImage}
+            </figure>
+        );
+    }
+
     return (
         <figure
             className="perseus-image-widget"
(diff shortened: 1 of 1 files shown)


packages/perseus/src/widgets/image/__docs__/image-demo.stories.tsx
```diff
@@ -0,0 +1,117 @@
...
+const ImageWidget = getWidget("image")!;
+
+type Story = StoryObj<typeof ImageWidget>;
+
+const earthMoonImageCaption =
+    "The Moon above Earth's horizon, captured by the International Space Station, [NASA](https://images.nasa.gov/details/iss071e515452)";
+const articleContent = `But in other cases, an object may experience a centripetal force for an extended time and complete *repeated* revolutions. An example of this type of motion is an astronomical object in **orbit**.\n\n[[☃ image 1]]\n\nLet's explore some of the language and relationships involved in orbital motion.`;
+
+const meta: Meta<typeof ImageWidget> = {
+    title: "Widgets/Image/Widget Demo",
```
ivyolamit, line 21: added this to follow our storybook pattern of having the widget demo

packages/perseus/src/widgets/__testutils__/image-renderer-decorator.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+import {getFeatureFlags} from "../../../../../testing/feature-flags-util";
+import {ApiOptions} from "../../perseus-api";
+import Renderer from "../../renderer";
+import {mockStrings} from "@khanacademy/perseus/strings";
+import UserInputManager from "../../user-input-manager";
+
+import type {PerseusRenderer} from "@khanacademy/perseus-core";
+
+export const imageRendererDecorator = (_, {args, parameters}) => {
```
ivyolamit, line 19: move this to make it reusable to stories

packages/perseus/src/widgets/image/__docs__/image-regression.stories.tsx
```diff
@@ -63,42 +34,17 @@ const meta: Meta<typeof ImageWidget> = {
...
         backgroundImage: earthMoonImage,
         alt: "Earth and Moon",
-        longDescription:
-            "This is a *very* long description of the earth and moon.",
         title: "Earth and Moon",
         caption: earthMoonImageCaption,
     },
 };
 
-export const ImageWithAlt: Story = {
```
ivyolamit, line 84: Simplified our stories since there's too many of them. The idea is:
- have all components displayed
- have partial display for special cases

packages/perseus/src/widgets/__testutils__/image-renderer-decorator.tsx
nishasy: Nice!

packages/perseus/src/widgets/image/image.tsx
nishasy, line 1: Proposal:

You can short circuit this so that if decorative is true, it just loads the image - maybe we can extract out the SvgImage component with all the props so that it can be reused in the decorative check and the rest of the non-decorative code.

That way, we can easily put this behind a feature flag. I think all the new props we're adding should be behind the image upgrade feature flag, including `decorative`.


```
// maybe this could be a function that can take alt so you can make it "" for decorative images
const svgImage = <AssetContext.Consumer>
  {({setAssetStatus}) => (<SvgImage... />)}
</AssetContext.Consumer>

if (imageUpgradeFF && decorative) {
  return svgImage;
}

return ... // everything

```

catandthemachines:
Will this PR also include updating the `imageWidgetLogic` and specifically the `accessible` function to encorporate the new `decorative` property in determining whether the image widget is accessible?

https://github.com/Khan/perseus/blob/c2bcd7b0db198b0217c8e9dc9cb68e865a2630a4/packages/perseus-core/src/widgets/image/index.ts#L36

packages/perseus/src/widgets/image/image.tsx
ivyolamit: good call.

nishasy (approved):
Nice! Looks good to me 🎉

PR #2919, opened 2025-09-23
[Radio] Add screen reader announcements to help clarify state to the user

mark-fitzgerald:
## Summary:
Since the multiple choice widget (aka Radio widget) is now a group of toggle buttons, we need a bit more information communicated to the screen reader user in order to ensure clarity of state when interacting with it. Specifically:
- Single Select
   - When toggling "on" a choice, screen readers already announce that the button is selected (or something equivalent).
   - When toggling "off" a choice, some screen readers (looking at you VoiceOver) don't announce the new unselected state.
      - This PR adds an announcement of "not selected".
- Multiple Select
   - The "on"/"off" state of an individual choice isn't as important as the total number of choices that are selected at any given time.
      - Instead of announcing whether a choice is selected or not (users can explore the choices to determine which ones are selected), this PR adds an announcement of "X choice(s) selected".

Issue: LEMS-3436

## Test plan:
1. Open Storybook for this PR (found at the bottom of the checks - Storybook Publish)
2. Navigate to any of the "Widget Demo" examples
   - Single Select: Widgets => RadioNew => Widget Demo => Single Select
   - Multiple Select: Widgets => RadioNew => Widget Demo => Multi Select
3. Open the demo in its own window (avoids a VoiceOver bug involving iFrames) by clicking the "Open canvas in new tab" icon in the top right corner of the page.
Open canvas in new tab

4. Start VoiceOver
5. Interact with the widget by toggling choices on and off
6. Announcements should include what is outlined in the summary of this PR


SonicScrewdriver (approved):
Love it!!! This works great!

My only question is whether we should be confirming to users when they've successfully deselected an option. Perhaps it's unnecessary, but I could see myself, personally, wanting stronger confirmation than catching the simple absence of "selected" when it rereads the option out. 🤔

mark-fitzgerald:
> My only question is whether we should be confirming to users when they've successfully deselected an option. Perhaps it's unnecessary, but I could see myself, personally, wanting stronger confirmation than catching the simple absence of "selected" when it rereads the option out. 🤔

When tabbing/navigating to different choices to review their state (selected or not), VoiceOver choses not to announce anything when the toggle button is NOT selected. However, NVDA choses to announce both states: "pressed" and "not pressed". Since adding something to account for Safari's lack of an announcement would add duplication to NVDA, we decided to not do anything for it and let the default behavior prevail (users will be used to how their screen reader announces things). It is a fine line between not doing it for that use case (reviewing choice states) and doing it for the de-select action. We decided to do it for this use case (deselecting/untoggling) since the absence of radio button semantics could cause confusion for the screen reader user. With a Single Select instance, the number of choices selected isn't important, whereas whether the current choice is selected or not is important.

ivyolamit (approved):
@mark-fitzgerald changes looks good to me 👍🏼

PR #2920, opened 2025-09-24
[Image] | (CX) | Add a warning message about alt text length in editor

nishasy:
## Summary:
We need a way to indicate to content authors that long description should be used
if alt text needs to exceed 150 characters.

In this PR:
- Update warning messages within Image editor - alt text too short, alt text too long
  - "Too short" updated on blur
  - "Too long" updated on change
- Add a description to alt text field to inform content authors that alt text should
  only be up to 150 characters, based on design feedback.
- Update Image URL tooltip to be description instead, based on design feedback
- Removed ":" from field labels as they were unnecessary
- Updated styling

Issue: https://khanacademy.atlassian.net/browse/LEMS-3379

## Test plan:
`pnpm jest packages/perseus-editor/src/__tests__/editor.test.tsx`

Storybook
- `?path=/docs/widgets-image-editor-demo--docs#populated-within-editor-page`

https://github.com/user-attachments/assets/f2645885-ad3c-4071-8228-d3e2dc4a3504

### DIFF
--- a/packages/perseus-editor/src/widgets/image-editor/image-settings.tsx
+++ b/packages/perseus-editor/src/widgets/image-editor/image-settings.tsx
@@ -50,7 +50,7 @@ export default function ImageSettings({
         if (value.length === 0) {
             // If the user clears the alt text, clear the error
             setAltFieldError(null);
-        } else if (value.length > MAX_ALT_TEXT_LENGTH) {
+        } else if (imageUpgradeFF && value.length > MAX_ALT_TEXT_LENGTH) {
             setAltFieldError(altTextTooLongError);
         } else if (value.length >= MIN_ALT_TEXT_LENGTH) {
             setAltFieldError(null);


packages/perseus-editor/src/widgets/image-editor/image-settings.tsx
```diff
@@ -34,15 +44,38 @@ export default function ImageSettings({
...
         <>
             {/* Preview */}
             <LabeledField
-                label="Preview:"
+                label="Preview"
                 field={
                     <SvgImage
                         src={backgroundImage.url}
-                        alt={`Preview: ${alt ?? "No alt text"}`}
+                        alt={`Preview: ${alt || "No alt text"}`}
```
nishasy, line 78: Realized this wasn't using the "No alt text" message when the alt text was ""

packages/perseus-editor/src/widgets/__tests__/image-editor.test.tsx
```diff
@@ -30,16 +37,48 @@ describe("image editor", () => {
         jest.spyOn(Dependencies, "getDependencies").mockReturnValue(
             testDependencies,
         );
+
+        originalImage = window.Image;
+        // The editor preview uses SvgImage, which uses ImageLoader.
+        // We need to mock the image loading in ImageLoader for it to render.
+        // Mock HTML Image so we can trigger onLoad callbacks and see full
+        // image rendering.
+        // @ts-expect-error - TS2322 - Type 'Mock<Record<string, any>, [], any>' is not assignable to type 'new (width?: number | undefined, height?: number | undefined) => HTMLImageElement'.
```
ivyolamit, line 46: we can probably remove this TS error by doing
```
        window.Image = jest.fn(() => {
            const img: Record<string, any> = {};
            images.push(img);
            return img;
        }) as any;
```

nishasy: Is `as any` better than having the eslint error?

I think I might keep this as is, since it's duplicated in a few different test files. I think we should go back later and update all of them to have the right types.

jeremywiebe: Overall, we should prefer using `as any` over `@ts-expect-error` because `@ts-expect-error` completely suppresses any TypeScript type checking on the affected lines. We've had instances where lines that had `@ts-expect-error` was hiding legitimate type check failures unrelated to what we were suppressing. 

Ideally, instead of `as` we use `satisfies` if possible.

nishasy: I remember trying to resolve this issue before but removing the lint error and avoiding `any`, and it was pretty bad. I'll have to take another look soon.

PR #2925, opened 2025-09-25
Wrap article editor content in dependency context provider

MikeKlemarewski:
## Summary:
The ArticleEditor was not wrapping it's content in the dependency context provider. This caused issues because ImageLoader expects to be able to look up dependencies on context in order to use the new `generateUrl` function in place of the older `staticUrl` dependency.

## Test plan:

### DIFF
--- a/packages/perseus-editor/src/__docs__/editor-page-with-storybook-preview.tsx
+++ b/packages/perseus-editor/src/__docs__/editor-page-with-storybook-preview.tsx
@@ -1,4 +1,9 @@
-import {Renderer, type APIOptions, type DeviceType} from "@khanacademy/perseus";
+import {
+    type PerseusDependenciesV2,
+    Renderer,
+    type APIOptions,
+    type DeviceType,
+} from "@khanacademy/perseus";
 import {
     type Hint,
     type PerseusAnswerArea,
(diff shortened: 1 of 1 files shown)


handeyeco (approved):
I think this looks fine. Since this is a regression, I think there's an opportunity to add some regression tests. Ideally as part of this PR but, if we're strapped for time, could we file a ticket to come back and add some regression tests?

packages/perseus-editor/src/__docs__/editor-page-with-storybook-preview.tsx
```diff
@@ -26,6 +31,23 @@ type Props = {
     hints?: ReadonlyArray<Hint>;
 };
 
+const testDependenciesV2: PerseusDependenciesV2 = {
```
jeremywiebe: Can you import the one from [here](https://github.com/Khan/perseus/blob/5d2a6b07ef29fdbd6b3458b8abe4dbad480ce2b5/testing/test-dependencies.tsx#L104) instead of creating a new object here?

jeremywiebe (approved):

packages/perseus-editor/src/__docs__/editor-page-with-storybook-preview.tsx
MikeKlemarewski: I updated it to import the shared testDependencies, and I now remember why I re-defined these here. I was running into [this typescript error](https://github.com/Khan/perseus/actions/runs/18050375627/job/51370929353?pr=2925) and I'm not sure how to fix it. I tried excluding this storybook file from the build in `perseus-editor/tsconfig-build.json` as well, but that didn't seem to help either.

PR #2926, opened 2025-09-25
[Image] | (CX) | Make image dimensions editable in Image widget editor

nishasy:
## Summary:
We received a request from content to make image resizable.

We already save the dimensions of the image inside the widget's `backgroundImage` prop,
so we can just make that value updatable.

Added a couple text fields for width and height to make those updatable.

Issue: https://khanacademy.atlassian.net/browse/LEMS-3576

## Test plan:
`pnpm jest packages/perseus-editor/src/widgets/radio/utils.test.ts`

Storybook
- `/?path=/story/editors-editorpage--demo`
- `http://localhost:6006/?path=/docs/widgets-image-editor-demo--docs#populated-within-editor-page`

https://github.com/user-attachments/assets/9581f33c-e86f-4fa8-9f7f-8083f53431ac

### DIFF
--- a/packages/perseus-editor/src/widgets/image-editor/image-settings.tsx
+++ b/packages/perseus-editor/src/widgets/image-editor/image-settings.tsx
@@ -2,10 +2,11 @@ import {components} from "@khanacademy/perseus";
 import {isFeatureOn} from "@khanacademy/perseus-core";
 import {LabeledField} from "@khanacademy/wonder-blocks-labeled-field";
 import {sizing} from "@khanacademy/wonder-blocks-tokens";
-import {HeadingXSmall} from "@khanacademy/wonder-blocks-typography";
 import * as React from "react";
 
 import {AutoResizingTextArea} from "../../components/auto-resizing-text-area";
+import ScrolllessNumberTextField from "../../components/scrollless-number-text-field";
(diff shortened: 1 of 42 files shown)


packages/perseus-editor/src/widgets/image-editor/image-settings.tsx
```diff
@@ -50,18 +89,27 @@ export default function ImageSettings({
...
-                        // are moved to a different layer.
-                        padding: 0, // reset default padding
-                        marginInlineEnd: sizing.size_080,
-                        color: "var(--wb-semanticColor-core-foreground-neutral-strong)",
-                    }}
-                >
-                    Dimensions:
-                </HeadingXSmall>
-                {dimensionString}
+                <LabeledField
```
ivyolamit: can we add this to its own internal component in `packages/perseus-editor/src/widgets/image-editor/components` advantage of doing this are:
- logic specific to dimension change will be isolated in its own component include the two onchange handler
- makes it easier to create an isolated test for it
- makes the image-settings code readable, as you can imagine the more we add things in this file the bigger this gets
- image settings test will also be much more simple

ivyolamit (commented):
@nishasy PR looks pretty straight forward. I just have one comment, please let me know what you think about it.

ivyolamit (approved):
sweet! thanks @nishasy for the update.

PR #2927, opened 2025-09-25
[Image] | (CX) | Remove "math in caption" linter error

nishasy:
## Summary:

There's a linter error that shows up if there is TeX in an image caption:

> No math in image captions: Don't include math expressions in image captions.

This was likely added before we had adequate i18n support for math expressions, but this is no longer the case. It should be safe to remove this linter warning now.

Issue: none

## Test plan:
`pnpm jest packages/perseus-linter/src/rules/image-widget.test.ts`

Storybook
`/?path=/docs/widgets-image-editor-demo--docs#populated-within-editor-page`

| Before | After |
| ----- | ----- |
| Screenshot 2025-09-29 at 1 50 24 PM | Screenshot 2025-09-29 at 1 50 33 PM |

### DIFF
--- a/.changeset/great-houses-wonder.md
+++ b/.changeset/great-houses-wonder.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-linter": patch
+---
+
+[Image] | (CX) | Remove "math in caption" linter error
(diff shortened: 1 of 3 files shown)


catandthemachines (approved):
Short and sweat. Approved!

PR #2928, opened 2025-09-26
[Radio] Bugfix - Ensure fade bars match the background color of container

mark-fitzgerald:
## Summary:
Widgets can be placed inside containers that have a background color that isn't communicated to the widget. In order for the radio widget to have scrolling fade bars match the background color of the container, it needs a way to get the background color. This PR adds a utility function that traverses the DOM ancestry of a given element until it finds a valid background color. It then uses it for the fade bar color. If no background color is found, then it uses the default background color for the theme.

Issue: LEMS-3545

## Test plan:
1. Open Storybook for this PR (found at the bottom of the checks - Storybook Publish)
2. Navigate to the Graded Group Set with Scroll regression story
   - Widgets => RadioNew => Visual Regression Tests => Static => Graded Group Set With Scroll
3. The fade area on the sides of the widget should match the background of the graded group container
   - Note: The first story on the page is not the story to inspect (it's merely a copy of the first story that Storybook automatically displays). The duplicate (first) story displayed will show white fade bars due to how Storybook implements that duplicate. The actual (second) story displayed will consistently show the fade bars with the same background as the container.


packages/perseus/src/widgets/graded-group-set/graded-group-set.tsx
```diff
@@ -308,10 +308,6 @@ const styles = StyleSheet.create({
         borderTop: `1px solid ${gray76}`,
         borderBottom: `1px solid ${gray76}`,
         backgroundColor: tableBackgroundAccent,
-        // The following CSS variable is put into place to let child widgets know that there is a background color in effect.
-        // For example, this is helpful for the radio widget to properly color its horizontal scroll fade bars.
-        //  @ts-expect-error TS2353: Object literal may only specify known properties
-        "--perseus-widget-background-color": tableBackgroundAccent,
```
mark-fitzgerald, line 313: The line above was originally just addressing the Graded Group Set container. This PR adds the utility that handles all situations with a background color, making this line no longer needed.

packages/perseus/src/widgets/radio/multiple-choice-component.new.tsx
```diff
@@ -78,6 +95,8 @@ const MultipleChoiceComponent = ({
             <fieldset
                 className={styles.container}
                 data-feature-flag="feature flag is ON"
+                ref={containerRef}
+                style={cssVariableDeclaration}
```
ivyolamit, line 112: I might be missing the intention of this PR, but probably this is not where we want to add the styling. In our code we set the fade in two places:
- https://github.com/Khan/perseus/blob/main/packages/perseus/src/widgets/radio/choice-icon/choice-icon.new.tsx#L111
- https://github.com/Khan/perseus/blob/main/packages/perseus/src/widgets/radio/multiple-choice.module.css#L60

image

mark-fitzgerald: The reference to the choice-icon.new.tsx is not part of the actual "new" widget files (it has old code in it, like the focus ring).

The styling for the fade area is applied in `multiple-choice.module.css`, as you pointed out. It is referencing a color variable `--perseus-multiple-choice-fade-color`. That [variable is set on the container](https://github.com/Khan/perseus/blob/main/packages/perseus/src/widgets/radio/multiple-choice.module.css#L20-L23) and uses the `--perseus-widget-background-color` variable, if set. This PR sets that background variable on the same container element, but since this is determined via JavaScript, it has to be passed along to CSS via the `style` attribute.

SonicScrewdriver (approved):
Thank you for getting us a nice solution for this!

PR #2929, opened 2025-09-26
Wrap necessary image editor tests in dependency context provider

MikeKlemarewski:
## Summary:
A couple tests for the image editor end up rendering SvgImages and need to be wrapped in the dependency context provider.

## Test plan:
- ensure tests pass

### DIFF
--- a/packages/perseus-editor/src/widgets/__tests__/image-editor.test.tsx
+++ b/packages/perseus-editor/src/widgets/__tests__/image-editor.test.tsx
@@ -4,7 +4,11 @@ import {userEvent as userEventLib} from "@testing-library/user-event";
 import * as React from "react";
 
 import {getFeatureFlags} from "../../../../../testing/feature-flags-util";
-import {testDependencies} from "../../../../../testing/test-dependencies";
+import {
+    testDependencies,
+    testDependenciesV2,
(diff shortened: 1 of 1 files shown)


packages/perseus-editor/src/widgets/__tests__/image-editor.test.tsx
```diff
@@ -213,12 +217,14 @@ describe("image editor", () => {
 
         // Act
         render(
-            <ImageEditor
-                apiOptions={apiOptions}
-                backgroundImage={earthMoonImage}
-                alt="Earth and moon alt"
-                onChange={() => {}}
-            />,
+            <DependenciesContext.Provider value={testDependenciesV2}>
```
mark-fitzgerald, line 220: I'm not familiar with what's in V2. What's the reason for wrapping this editor in a context provider instead of replacing the mock return value on line 42?

MikeKlemarewski: It's because the v2 dependencies are provided by a context provider. The <SvgImage> component will use a new dependency function called `generateUrl` which gets provided by the v2 dependencies, so this needs to be wrapped in the context provider.

PR #2930, opened 2025-09-26
[Image] Add decorative toggle for Image widget editor and hide it behind a feature flag

ivyolamit:
## Summary:
Add decorative toggle for Image widget editor and hide it behind a feature flag

image

Issue: LEMS-3451

## Test plan:
1. Navigate to the image editor
2. Fill up the fields
3. Toggle the decorative image
- [ ] Confirm when decorative is toggled the confirmation window is displayed
- [ ] Confirm that the fields (title, caption, long description, alt) are empty and disabled
- [ ] Confirm when the decorative is toggled again it will enable the fields
- [ ] Confirm on initial load, the decorative image toggle is false

### DIFF
--- a/packages/perseus-editor/src/widgets/image-editor/components/image-settings.tsx
+++ b/packages/perseus-editor/src/widgets/image-editor/components/image-settings.tsx
@@ -6,8 +6,8 @@ import * as React from "react";
 import {AutoResizingTextArea} from "../../../components/auto-resizing-text-area";
 import {wbFieldStyles, wbFieldStylesWithDescription} from "../utils";
 
-import ImageDimensionsInput from "./image-dimensions-input";
 import DecorativeToggle from "./decorative-toggle";
+import ImageDimensionsInput from "./image-dimensions-input";
 
 import type {Props} from "../image-editor";
 


packages/perseus-editor/src/widgets/__tests__/image-editor.test.tsx
nishasy, line 1: Can you add a "renders decorative toggle as checked when prop is true" here too? I see it's tested in the other file, but I also like knowing that it's hooked up with the onChange handler in the full editor experience. It can be just the one test, not a repeat of all the tests below.

ivyolamit:
Here's a demo of the decorative toggle in the image widget editor

https://github.com/user-attachments/assets/e165858a-23b6-4dfe-a9cf-de105f82aa25

packages/perseus-editor/src/widgets/image-editor/image-settings.tsx
```diff
@@ -97,6 +99,18 @@ export default function ImageSettings({
...
 
+            {/* Decorative */}
+            {imageUpgradeFF && (
+                <DecorativeToggle
+                    decorative={decorative}
+                    alt={alt}
+                    caption={caption}
+                    title={title}
+                    longDescription={longDescription}
+                    onChange={onChange}
```
nishasy, line 98: Proposal

```suggestion
                <DecorativeToggle
                    decorative={decorative}
                    hasPopulatedFields={alt || caption || title || longDescription}
                    onChange={onChange}
```

packages/perseus-editor/src/widgets/image-editor/components/decorative-toggle.tsx
```diff
@@ -0,0 +1,83 @@
...
+    title?: string;
+    longDescription?: string;
+    onChange: (changes: {
+        decorative?: boolean;
+        alt?: string;
+        caption?: string;
+        title?: string;
+        longDescription?: string;
+    }) => void;
+};
```
nishasy: Proposal

```suggestion
interface Props {
    decorative?: boolean;
    hasPopulatedFields?: boolean;
    onChange: ImageEditorProps["onChange"]
};
```

nishasy (commented):
I did a super quick review while it was still a draft. I'll do another pass soon.

packages/perseus-editor/src/widgets/__tests__/image-editor.test.tsx
```diff
@@ -623,4 +623,85 @@ describe("image editor", () => {
             screen.queryByText(altTextTooShortError),
         ).not.toBeInTheDocument();
     });
+
+    describe("when decorative toggle", () => {
```
ivyolamit: added this group so it will read nicely when running the tests

image

nishasy: Nit: Can you update the test titles so they read more naturally?

I think you could change "when decorative toggle" to just "decorative toggle" and then:

- "should render decorative toggle when feature flag is enabled" ==> "renders when feature flag is enabled"
- "should not render decorative toggle when feature flag is disabled" ==> "should not render when feature flag is disabled"
- "should render decorative toggle checked when decorative is true" ==> "should render as checked when decorative is true"
- "should call onChange when decorative toggle is clicked" ==> "should call onChange when clicked"

so then in the terminal, these would read as "image editor decorative toggle should call onChange when clicked"

packages/perseus-editor/src/widgets/image-editor/components/decorative-toggle.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+        jest.spyOn(Dependencies, "getDependencies").mockReturnValue(
+            testDependencies,
+        );
+    });
+
+    afterEach(() => {
+        jest.restoreAllMocks();
+    });
+
+    test("renders with decorative toggle unchecked by default", () => {
```
nishasy: Personally, I prefer using `test`, but I think we decided to use `it` for consistency across tests (😞)

packages/perseus-editor/src/widgets/image-editor/components/decorative-toggle.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+        // Arrange & Act
```
nishasy: Nit: I think we usually do 

```
// Arrange

// Act
```

But I've also seen

```
// Arrange, Act
```

Can you use one of these for consistency?

packages/perseus-editor/src/widgets/image-editor/components/decorative-toggle.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+            wrapper: RenderStateRoot,
+        });
+
+        // Assert
+        const toggle = screen.getByRole("switch", {name: "Decorative"});
+        expect(toggle).toBeInTheDocument();
+        expect(toggle).not.toBeChecked();
+
+        // Check for InfoTip icon presence (tooltip content requires interaction)
+        // Since InfoTip doesn't expose a specific role, we'll verify the component structure
```
nishasy: Is this comment correct?

I think `expect(screen.getByLabelText("Decorative")).toBeInTheDocument();` is just getting the toggle again, not the tooltip?

tbh I think we may want to remove the tooltip and add the guidance as a description on this labeled toggle. Caitlyn expressed that it's better to avoid tooltips if possible.

packages/perseus-editor/src/widgets/image-editor/components/decorative-toggle.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+        // Assert
+        expect(mockConfirm).toHaveBeenCalledWith(
+            "Setting this image as decorative will automatically reset all other fields (title, caption, alt text, and long description). Do you want to continue?",
+        );
+        expect(mockOnChange).toHaveBeenCalledWith({
+            decorative: true,
+            alt: undefined,
+            caption: undefined,
+            title: undefined,
+            longDescription: undefined,
```
nishasy, line 128: nice

packages/perseus-editor/src/widgets/image-editor/components/decorative-toggle.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+        // Act
+        const toggle = screen.getByRole("switch", {name: "Decorative"});
+        await userEvent.click(toggle);
+
+        // Assert
+        expect(mockConfirm).toHaveBeenCalledWith(
+            "Setting this image as decorative will automatically reset all other fields (title, caption, alt text, and long description). Do you want to continue?",
+        );
+        expect(mockOnChange).not.toHaveBeenCalled();
```
nishasy, line 153: very nice

packages/perseus-editor/src/widgets/image-editor/components/decorative-toggle.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+    });
+
+    test("does not show warning dialog when fields are not populated", async () => {
```
nishasy: Is this the same as the test "calls onChange with decorative: true when toggle is clicked and no fields are populated"?

packages/perseus-editor/src/widgets/image-editor/components/decorative-toggle.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+        // Assert - Verify component renders with expected structure
+        // The InfoTip should be present alongside the toggle
+        expect(
+            screen.getByRole("switch", {name: "Decorative"}),
+        ).toBeInTheDocument();
+
+        // Since InfoTip is implemented as a tooltip, we verify the component structure is complete
+        // The presence of the switch indicates the component has rendered successfully with InfoTip
+        expect(screen.getByLabelText("Decorative")).toBeInTheDocument();
```
nishasy: same comment as before. I also don't think you need this test if you're already checking for this in the first test?

packages/perseus-editor/src/widgets/image-editor/components/decorative-toggle.tsx
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    return (
+        <div className={styles.decorativeToggleContainer}>
+            <div className={styles.flexRow}>
+                <LabeledSwitch
+                    label="Decorative"
+                    checked={decorative ?? false}
+                    onChange={handleDecorativeToggle}
+                />
```
nishasy, line 58: I think I wrote the `LabeledSwitch` component, but now I'm thinking we may want to update it to use the new Wonder Blocks `LabeledField` Then we can add a description instead of using the tooltip... I think.

I say maybe give it a try, and if it's not working out, then we can keep the tooltip.

packages/perseus-editor/src/widgets/image-editor/image-settings.tsx
```diff
@@ -97,6 +99,17 @@ export default function ImageSettings({
                 {dimensionString}
             </div>
 
+            {/* Decorative */}
+            {imageUpgradeFF && (
+                <DecorativeToggle
+                    decorative={decorative}
+                    hasPopulatedFields={Boolean(
+                        alt || caption || title || longDescription,
+                    )}
```
nishasy: Optional: What do you think about breaking out `Boolean(alt || caption || title || longDescription)` into a variable above the return? And then you can do `hasPopulatedFields={hasPopulatedFields}`.

packages/perseus-editor/src/widgets/image-editor/components/decorative-toggle.test.tsx
ivyolamit: Good catch! i'll have to add this in claude.md

packages/perseus-editor/src/widgets/image-editor/components/decorative-toggle.tsx
ivyolamit: Am i missing something?
Based on the docs, [LabeledField](https://khan.github.io/wonder-blocks/?path=/docs/packages-labeledfield--docs) doesn't have a switch 🤔

nishasy: Oh, LabeledField is used to wrap some other field. So in this case, it would be 

```
<LabeledField
  label="Decorative"
  field={<Switch ... />
/>
```

But now that I'm looking at [the docs](https://khan.github.io/wonder-blocks/?path=/docs/packages-labeledfield--docs), it doesn't seem like something that would work for switch, since switches would need the label on the side rather than on the top.

So you can just ignore my suggestion.

PR #2935, opened 2025-09-30
Measure the accessibility of our content

benchristel:
This PR adds a new `perseus-a11y-metrics` package. It includes a script that
downloads a snapshot of Khan Academy's content and evaluates the accessibility
of each exercise.

Issue: LEMS-3261

## Test plan:

```
packages/perseus-a11y-metrics/src/main.ts
```

should output a report that looks like this:

```
EXERCISE ACCESSIBILITY STATS:
-----------------------------
full:         7424 exercises
limited:      1706 exercises
inaccessible: 2469 exercises
-----------------------------
total:        11599 exercises
```

It will also upload a file with more detailed data, suitable for ingestion
into BigQuery, to Google Cloud.

Note: you need `gcloud` and `jq` installed.


packages/perseus-a11y-metrics/src/parse-snapshot.ts
```diff
@@ -0,0 +1,61 @@
...
+    curationPages: unknown(),
+    curriculums: unknown(),
+    domains: unknown(),
+    interactives: unknown(),
+    lessons: unknown(),
+    projects: unknown(),
+    talkthroughs: unknown(),
+    units: unknown(),
+    version: unknown(),
+    video: unknown(),
```
benchristel: I should delete all these fields. There's no reason to list them in the schema.

packages/perseus-a11y-metrics/src/parse-snapshot.ts
```diff
@@ -0,0 +1,61 @@
...
+    // NOTE: there are more fields in the ItemSchema object. This schema just
+    // lists the fields we currently care about.
+    id: string(),
+});
+
+const ProblemTypeSchema = object({
+    // NOTE: there are more fields in the ProblemTypeSchema object. This schema just
+    // lists the fields we currently care about.
+    items: array(ItemSchema),
+});
```
benchristel, line 24: I don't think we actually need to parse `problemTypes` or items.

packages/perseus-a11y-metrics/src/content-repository.ts
```diff
@@ -0,0 +1,142 @@
...
+            });
+            return this.readLocalSnapshotJsonWithJqFiltering();
+        }
+    }
+
+    private async readLocalSnapshotJsonWithJqFiltering(): Promise<string> {
+        // The snapshot data is too large (600 MB) to fit into a NodeJS string.
+        // The maximum size of a string is 512 MB. So we use `jq` to filter the
+        // data to just the exercises.
+        // TODO: de-duplicate path
```
benchristel: I should fix this TODO.

packages/perseus-a11y-metrics/src/main.ts
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env -S node -r @swc-node/register
+
+import {isItemAccessible} from "@khanacademy/perseus-core";
+
+import {ContentRepository} from "./content-repository";
+import {getPublishedContentVersion} from "./content-version";
+
+async function main() {
```
benchristel, line 13: We could print a nice error message here if `gcloud` or `jq` isn't installed.

config/cypress/cypress.config.ts
```diff
@@ -17,6 +17,9 @@ fs.readdirSync(path.join(__dirname, "../../packages")).forEach((name) => {
     }
     const pkgPath = path.join("../../packages", name, "package.json");
     const pkgJson = require(pkgPath);
+    if (pkgJson.source == null) {
+        return;
```
catandthemachines: What's the purpose of this line? And what cases would we hit it?

packages/perseus-a11y-metrics/src/main.ts
catandthemachines, line 1: I know this file is under construction and still being build. But just to confirm this script will output data that we could join with `assessmentItemIds` or other content metrics right? It won't just output the totals for all our content right?

catandthemachines (commented):
Added a few questions and comments on the early version of this script. 😊

packages/perseus-a11y-metrics/src/main.ts
benchristel: Outputting the totals was just for the proof of concept. My plan is to have the script upload a JSON file to GCS where it can be read by BigQuery.

packages/perseus-a11y-metrics/src/platform/command.ts
```diff
@@ -0,0 +1,88 @@
...
+}
+
+class VoidCommand implements Command<void> {
+    constructor(
+        private program: string,
+        private args: string[] = [],
+    ) {}
+
+    /**
+     * Runs the command and returns the exit code.
```
benchristel: This comment is wrong.

config/cypress/cypress.config.ts
benchristel: This was a hack to get the build working. Trying to use the `pkgJson.source` was failing because the new `perseus-a11y-metrics` package doesn't export anything and `source` was undefined. But I think there's a better way to do this.

packages/perseus-a11y-metrics/README.md
catandthemachines, line 1: Yay!! Thank you for writing this!

packages/perseus-a11y-metrics/src/main.ts
```diff
@@ -0,0 +1,65 @@
...
+        formatExerciseAccessibilityData(a11yStats),
+        "utf-8",
+    );
+
+    // TODO(benchristel): there is probably a better place for the data to
+    // live. I'm not sure what the khanflow-prod-bq-archive-unused-table-expiration
+    // bucket is supposed to be for.
+    // TODO(benchristel): should the extension for the data file be .ndjson?
+    // See: https://github.com/ndjson/ndjson-spec?tab=readme-ov-file#33-mediatype-and-file-extensions
+    await gcloudStorage.cp(
```
catandthemachines, line 46: Adding a note here, this script works when run once on a new object, but when you attempt to overwrite an existing object you will experience `404` errors.

Added a bug here detailing this phenomenon, will also add a TODO here in case anyone wants to run this script before it is fixed.
https://khanacademy.atlassian.net/browse/LEMS-3674

benchristel: This might be a permissions issue. From [the Google Cloud docs](https://docs.cloud.google.com/storage/docs/access-control/iam-permissions):

> Note: In order to replace existing objects, both storage.objects.create and storage.objects.delete permissions are required.

@catandthemachines can you confirm that you have the `storage.objects.delete` permission on this bucket?

PR #2936, opened 2025-10-01
Test that the ItemDiff component will render an image diff

MikeKlemarewski:
## Summary:
Adds a test to make sure the ItemDiff component will correctly render and load an Image.
A change to the dependencies API caused a regression last week, and this test would have caught it.
Unfortunately in order to ensure it catches the regression, some tricky image loading mocking has to be done.

## Test plan:
- Ensure the tests pass

### DIFF
--- a/packages/perseus-editor/src/diffs/__tests__/item-diff.test.tsx
+++ b/packages/perseus-editor/src/diffs/__tests__/item-diff.test.tsx
@@ -1,0 +1,111 @@
+import {Dependencies} from "@khanacademy/perseus";
+import {
+    generateImageOptions,
+    generateImageWidget,
+    generateTestPerseusItem,
+    generateTestPerseusRenderer,
+    getDefaultAnswerArea,
+    type PerseusItem,
+} from "@khanacademy/perseus-core";
+import {act, render, screen} from "@testing-library/react";
+import * as React from "react";
+
+import {mockImageLoading} from "../../../../../testing/image-loader-utils";
+import {
+    testDependencies,
+    testDependenciesV2,
+} from "../../../../../testing/test-dependencies";
+import {registerAllWidgetsAndEditorsForTesting} from "../../util/register-all-widgets-and-editors-for-testing";
+import ItemDiff from "../item-diff";
+
+describe("ItemDiff", () => {
+    let unmockImageLoading: () => void;
+
+    beforeAll(() => {
+        registerAllWidgetsAndEditorsForTesting();
+    });
+
+    beforeEach(() => {
+        jest.spyOn(Dependencies, "getDependencies").mockReturnValue(
+            testDependencies,
+        );
+        unmockImageLoading = mockImageLoading();
+    });
+
+    afterEach(() => {
+        jest.restoreAllMocks();
+        unmockImageLoading();
+    });
+
+    it("renders an image widget in the diff view", () => {
+        // Arrange
+
+        const beforeItem: PerseusItem = generateTestPerseusItem({
+            question: generateTestPerseusRenderer({
+                content: "[[☃ image 1]]",
+                widgets: {
+                    "image 1": generateImageWidget({
+                        options: generateImageOptions({
+                            title: "Original Image",
+                            caption: "This is the original image",
+                            alt: "Original image description",
+                            backgroundImage: {
+                                url: "https://example.com/original-image.jpg",
+                                width: 400,
+                                height: 300,
+                            },
+                        }),
+                    }),
+                },
+            }),
+            answerArea: getDefaultAnswerArea(),
+            hints: [],
+        });
+
+        const afterItem: PerseusItem = generateTestPerseusItem({
+            question: generateTestPerseusRenderer({
+                content: "[[☃ image 1]]",
+                widgets: {
+                    "image 1": generateImageWidget({
+                        options: generateImageOptions({
+                            title: "Updated Image",
+                            caption: "This is the updated image",
+                            alt: "Updated image description",
+                            backgroundImage: {
+                                url: "https://example.com/updated-image.jpg",
+                                width: 500,
+                                height: 400,
+                            },
+                        }),
+                    }),
+                },
+            }),
+            answerArea: getDefaultAnswerArea(),
+            hints: [],
+        });
+
+        // Act
+        render(
+            <ItemDiff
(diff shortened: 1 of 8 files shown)


packages/perseus-editor/src/diffs/__tests__/item-diff.test.tsx
```diff
@@ -0,0 +1,111 @@
...
+            <ItemDiff
+                before={beforeItem as any}
+                after={afterItem as any}
+                dependencies={testDependenciesV2}
+            />,
+        );
+
+        // Wait for the next tick to allow setTimeout to execute
+        act(() => {
+            jest.runAllTimers();
```
MikeKlemarewski, line 98: Unfortunately we need this in order for the test to properly catch the regression. The diff has to render, and then the fake "onload" image event needs to fire and then get handled. I'm not sure if there's a better way to ensure these kinds of things fail in tests without running the jest timers.

packages/perseus-editor/src/diffs/__tests__/item-diff.test.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+    beforeAll(() => {
+        registerAllWidgetsAndEditorsForTesting();
+    });
+
+    beforeEach(() => {
+        jest.spyOn(Dependencies, "getDependencies").mockReturnValue(
+            testDependencies,
+        );
+        unmockImageLoading = mockImageLoading();
```
MikeKlemarewski, line 32: LMK what you think about this pattern. It's a bit funky I think.

packages/perseus-editor/src/diffs/__tests__/item-diff.test.tsx
jeremywiebe: :/ yeah, there's places in Perseus that re-render that we have to handle like this. A necessary evil. Maybe one day we'll work Suspense through everything (although I'm not sure if that'd solve this problem).

testing/image-loader-utils.ts
```diff
@@ -0,0 +1,37 @@
+import {act} from "@testing-library/react";
+
+/**
+ * Mocks image loading done by the ImageLoader component to immediately trigger
+ * the load event.
+ * @returns A function to unmock the image loading
+ */
+export const mockImageLoading = () => {
```
jeremywiebe, line 8: Cool. I know we have at least one other place we do this in https://github.com/Khan/perseus/blob/e11c3f8071d2b0946351ff700c846c6c48fbebfa/packages/perseus/src/__tests__/renderer.test.tsx#L290

jeremywiebe: Not asking for changes.  But could you leave a TODO comment there for future so if we touch that code we can refactor it to use this mock function instead?

jeremywiebe (approved):

packages/perseus-editor/src/diffs/__tests__/item-diff.test.tsx
jeremywiebe: I've seen this pattern of "unregistering" something in `afterEach` before. I suspect we can't call `afterEach` _inside_ the `mockImageLoading` function to register the cleanup automatically, right?

testing/image-loader-utils.ts
nishasy: We have this in a few places!

- renderer.test.tsx
- svg-image.test.tsx
- image.test.ts
- editor.test.tsx

Looks like either this PR could cover these, or I can repurpose the the following cleanup ticket to do it myself later: https://khanacademy.atlassian.net/browse/LEMS-3579

packages/perseus/src/__tests__/renderer.test.tsx
```diff
@@ -286,41 +287,16 @@ describe("renderer", () => {
...
-            originalImage = window.Image;
-            // Mock HTML Image so we can trigger onLoad callbacks and see full
-            // image rendering.
-            // @ts-expect-error - TS2322 - Type 'Mock<Record<string, any>, [], any>' is not assignable to type 'new (width?: number | undefined, height?: number | undefined) => HTMLImageElement'.
-            window.Image = jest.fn(() => {
-                const img: Record<string, any> = {};
-                images.push(img);
-                return img;
-            });
+            unmockImageLoading = mockImageLoading();
```
nishasy, line 292: Sorry, I'm a little be confused about this line. Why did you do `unmockImageloading = mockImageLoading()`? Maybe it's just the naming that's confusing me.

Just to clarify, this is equivalent to the following?

```
unmockImageLoading = () => {
        window.Image = originalImage;
    };
```

nishasy (commented):
The image loader util is super helpful! Would it be possible to use it in the other files that mock the image loading as well?

packages/perseus/src/__tests__/renderer.test.tsx
MikeKlemarewski: Yeah, the pattern is a bit odd. Rather than have the `mockImageLoading` function return the original image and require the user to reset it using the code you wrote, I thought it might be handy to just return a function that un-does the mocking.

Is there a way I could make it more clear?

nishasy: Hmmm... maybe you could just comment that in the function implementation where the return happens? 

Or maybe you could rename it to be `mockImageLoadingAndReturnUnmock()`? It's a mouthful, but it may be more clear?

MikeKlemarewski: Screenshot 2025-10-02 at 4 24 48 PM

Do you not think the JSDoc is sufficient? It mentions the return type

nishasy: Sounds good to me!

nishasy:
I don't understand why these tests are failing. I pulled your branch, and the tests all pass for me locally.

nishasy (approved):
Looks good to me! Thank you for doing this!

packages/perseus/src/__tests__/renderer.test.tsx
jeremywiebe: It's definitely a pattern I've seen before (calling a registration function returns one that cleans things up). 

I think the updated JSDocs are just perfect, thanks. :)

PR #2937, opened 2025-10-01
Add new SVG for keypad in Ukraine

SonicScrewdriver:
## Summary:
We only _semi_ implemented support for the colon division symbol in our Expression widget, as we were waiting for a new SVG asset from design. This PR adds that new SVG and the logic required to provide the correct symbol according to the local. 

Issue: LEMS-3535

## Video:
https://github.com/user-attachments/assets/9ad1b718-a190-40d3-a049-b5fcacd79793

## Test plan:
- Manual testing using my (unreleased) locale feature

### DIFF
--- a/packages/math-input/src/components/keypad/button-assets.tsx
+++ b/packages/math-input/src/components/keypad/button-assets.tsx
@@ -299,7 +299,7 @@ export default function ButtonAsset({id}: Props): React.ReactNode {
             );
         case "DIVIDE":
             // Different locales use different division symbols
-            if (id === "DIVIDE" && getDivideSymbol(locale) === ":") {
+            if (getDivideSymbol(locale) === ":") {
                 // colon divide symbol (:)
                 return (
                     <svg


packages/math-input/src/components/keypad/button-assets.tsx
```diff
@@ -298,6 +298,25 @@ export default function ButtonAsset({id}: Props): React.ReactNode {
                 </svg>
             );
         case "DIVIDE":
+            // Different locales use different division symbols
+            if (id === "DIVIDE" && getDivideSymbol(locale) === ":") {
```
handeyeco: The switch is on `id` already, so the fact that we hit `case "DIVIDE"` means we already know `id` is `DIVIDE`.

```
if (getDivideSymbol(locale) === ":") {
```

packages/math-input/src/components/keypad/button-assets.tsx
```diff
@@ -298,6 +298,25 @@ export default function ButtonAsset({id}: Props): React.ReactNode {
...
...4 23.0401 21.176 23.1921C21.328 23.3441 21.448 23.5241 21.536 23.7321C21.624 23.9321 21.668 24.1481 21.668 24.3801C21.668 24.6201 21.624 24.8401 21.536 25.0401C21.448 25.2401 21.328 25.4161 21.176 25.5681C21.024 25.7201 20.844 25.8361 20.636 25.9161C20.436 26.0041 20.22 26.0481 19.988 26.0481C19.756 26.0481 19.536 26.0041 19.328 25.9161C19.128 25.8361 18.952 25.7201 18.8 25.5681C18.656 25.4161 18.54 25.2401 18.452 25.0401C18.372 24.8401 18.332 24.6201 18.332 24.3801ZM18.332 15.6321C18.332 15.4001 18.372 15.1841 18.452 14.9841C18.54 14.7761 18.656 14.5961 18.8 14.4441C18.952 14.2921 19.128 14.1721 19.328 14.0841C19.536 13.9961 19.756 13.9521 19.988 13.9521C20.22 13.9521 20.436 13.9961 20.636 14.0841C20.844 14.1721 21.024 14.2921 21.176 14.4441C21.328 14.5961 21.448 14.7761 21.536 14.9841C21.624 15.1841 21.668 15.4001 21.668 15.6321C21.668 15.8721 21.624 16.0921 21.536 16.2921C21.448 16.4921 21.328 16.6681 21.176 16.8201C21.024 16.9721 20.844 17.0921 20.636 17.1801C20.436 17.2601 20.22 17.3001 19.988 17.3001C19.756 17.3001 19.536 17.2601 19.328 17.1801C19.128 17.0921 18.952 16.9721 18.8 16.8201C18.656 16.6681 18.54 16.4921 18.452 16.2921C18.372 16.0921 18.332 15.8721 18.332 15.6321Z"
```
handeyeco, line 313: I realize this had more to do with the image generation tool, but this seems pretty overkill for what could essentially be two `<circle>` elements. No action needed.

packages/math-input/src/components/keypad/button-assets.tsx
SonicScrewdriver: Yeah I wondered about that. I just copied the decimal symbol implementation, but I can simplify it

packages/math-input/src/components/keypad/button-assets.tsx
SonicScrewdriver: Yeah, agreed. :(  Unfortunately, I don't think anyone really knows how the previous designer was exporting the images and I don't have the license to be able to export from Figma myself.

PR #2939, opened 2025-10-01
Use pure markdown to parse instead of regex

nishasy:
## Summary:
Created a new Perseus linter rule so that the editor has a warning when markdown
images are used.

Bonus - added a new story for the Image widget editor that shows this lint error.

Issue: https://khanacademy.atlassian.net/browse/LEMS-3568

## Test plan:
`pnpm jest packages/perseus-linter/src/rules/image-markdown.test.ts`

Storybook
`/?path=/story/widgets-image-editor-demo--with-markdown-image-linter-warning`

### DIFF
--- a/packages/perseus-linter/src/rules/image-markdown.ts
+++ b/packages/perseus-linter/src/rules/image-markdown.ts
@@ -5,6 +5,12 @@ export default Rule.makeRule({
     severity: Rule.Severity.WARNING,
     selector: "image",
     lint: function (state, content, nodes, match, context) {
+        console.log("state", state);
+        console.log("content", content);
+        console.log("nodes", nodes);
+        console.log("match", match);
+        console.log("context", context);
+
         // Discourage using markdown images
         //
         // Regex for ![alt](image url):
@@ -14,6 +20,10 @@ export default Rule.makeRule({
         //   \( start of image url
         //   [^\)]* any characters except )
         //   \) end of link
+        //
+        // By confirming the markdown is inside the content paragraph, we can
+        // avoid flagging images within other widgets (e.g. Radio) that still
+        // have to use markdown.
         //
         // NOTE: Can't use PerseusMarkdown.parse() to identify the markdown
         // image because it would try to access 'allWidgets' before
@@ -25,3 +35,5 @@ Please use the Image widget instead.`;
         }
     },
 }) as Rule;
+
+// STOP!!!!! THIS DOESN'T STOP FLAGGING STUFF WITHIN RADIO WIDGETS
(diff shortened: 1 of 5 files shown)


.changeset/empty-rice-relate.md
```diff
@@ -0,0 +1,6 @@
+---
+"@khanacademy/perseus-editor": minor
+"@khanacademy/perseus-linter": minor
+---
+
+[Image] | (CX) | Add a warning in editor if markdown images are detected
```
jeremywiebe, line 6: What does "CX" stand for here?

packages/perseus-linter/src/rules/image-markdown.ts
```diff
@@ -0,0 +1,33 @@
+import {parse, traverseContent} from "@khanacademy/pure-markdown";
+
+import Rule from "../rule";
+
+export default Rule.makeRule({
```
jeremywiebe, line 12: Although it's not common practice in our current rules, I'd love to see our linter rules better-documented to explain the reasoning for them. 

Would you mind adding a JSDoc comment on this export explaining the detail of this rule and why we added it? Something like this... 

```suggestion
/**
 * This rule warns when it sees Markdown images used in the 
 * top-level content. We are migrating to use the `image` 
 * widget exclusively for images and this helps us to alert
 * content authors of this policy change. This rule ignores
 * markdown images found within other widgets for now. 
 */
export default Rule.makeRule({
```

packages/perseus-linter/src/rules/image-markdown.ts
```diff
@@ -0,0 +1,33 @@
...
+        traverseContent(parsedMarkdown, (node: any) => {
+            if (node.type === "image") {
+                hasInlineImageNode = true;
+            }
+        });
+
+        if (hasInlineImageNode) {
+            return `No inline markdown images:
+Markdown images (![alt](url) format) are not recommended.
+Please use the Image widget instead.`;
```
jeremywiebe: I find this wording challenging and I think content authors might be a bit confused also. I think most rules use the message format something like this `[human-readable rule name]: [explanation]`. 

What do you think of this? `Inline markdown image: ...`

.changeset/empty-rice-relate.md
nishasy: content experience

packages/perseus-linter/src/rules/image-markdown.ts
nishasy: I really like how you worded this. Thanks for writing up a description I can use!

packages/perseus-linter/src/rules/image-markdown.ts
nishasy: How about this?

image

jeremywiebe: Yeah, I think that reads more clearly! thanks.

PR #2941, opened 2025-10-02
Add `getCorrectUserInput` for dropdown, expression, input-number, label-image, and numeric-input

jaredly:
## Summary:
This is to enable the 'show answers' feature of the chatgpt embedded experience

Issue: FEI-XXXX

## Test plan:

### DIFF
--- a/packages/perseus/src/widgets/expression/__docs__/expression.stories.tsx
+++ b/packages/perseus/src/widgets/expression/__docs__/expression.stories.tsx
@@ -39,7 +39,6 @@ export const DesktopKitchenSink = (args: Story["args"]): React.ReactElement => {
                 alignment={null}
                 visibleLabel=""
                 ariaLabel=""
-                isLastUsedWidget
                 containerSizeClass="small"
                 findWidgets={(callback) => []}
                 problemNum={1}


packages/perseus/src/widgets/numeric-input/numeric-input.class.tsx
```diff
@@ -151,11 +151,38 @@ function getUserInputFromSerializedState(
...
+            }
+        }
+    }
+    return value.toString();
+}
+
+function getCorrectUserInput(
+    options: PerseusNumericInputWidgetOptions,
+): PerseusNumericInputUserInput {
+    for (const answer of options.answers) {
```
jeremywiebe, line 171: We should be filtering on answers that have a `status: "correct"`.  Content authors can add answers here that are not the correct answer (such as commonly guessed incorrect answers along with a reason why it's wrong). 

https://github.com/Khan/perseus/blob/97271406f2e929ff00306cbfdeb6ca950512a7af/packages/perseus-core/src/data-schema.ts#L1216

packages/perseus/src/widgets/numeric-input/numeric-input.class.tsx
```diff
@@ -151,11 +151,38 @@ function getUserInputFromSerializedState(
...
+        if (answer.value != null) {
+            return {currentValue: findCommonFractions(answer.value)};
```
jeremywiebe: Not all answers for numeric-input are fractions. I think we need to look at the `answerForms` to make sure `proper` and/or `improper` fractions are allowed. 

https://github.com/Khan/perseus/blob/97271406f2e929ff00306cbfdeb6ca950512a7af/packages/perseus-core/src/data-schema.ts#L1218

packages/perseus/src/widgets/expression/expression.tsx
```diff
@@ -338,6 +338,12 @@ function getOneCorrectAnswerFromRubric(
     return correctAnswers[0].value;
 }
 
+function getCorrectUserInput(
+    options: PerseusExpressionWidgetOptions,
+): PerseusExpressionUserInput {
+    return options.answerForms[0].value;
```
jeremywiebe: Some answer forms are considered incorrect or ungraded... so this should be something like `return options.answerForms.find(f => f.considered === "correct")`

PR #2943, opened 2025-10-02
[Image] | (UX) | Use large image size for zoom if available

nishasy:
## Summary:
We recently added the ability to resize an image in the editor - the new size
gets saved as part of the `backgroundImage.height` and `backgroundImage.width`
properties.

In the current implementation, this saved size is used for both rendering the
image at that size and for determining if it's a large enough resolution for
zooming. (The SVGImage only has the zoom feature if the image is larger than
its parent width.)

In this PR:
- Use the saved size for rendering the image within its item container
- Beteween the saved size and the original/natural image size, use the
  bigger one to determine zoom and "explore image" modal's image size.

This means that images can be rendered at a smaller size, even though they're
downloaded at full resolution, and this full resolution size can be used to
zoom into the image within the Image widget.

Issue: https://khanacademy.atlassian.net/browse/LEMS-3389

## Test plan:
Storybook
`/?path=/story/widgets-image-widget-demo--image-with-different-sizes`

https://github.com/user-attachments/assets/5ff93eed-f8a9-4f90-a692-b727cf3229ea

### DIFF
--- a/packages/perseus/src/widgets/image/components/image-description-and-caption.tsx
+++ b/packages/perseus/src/widgets/image/components/image-description-and-caption.tsx
@@ -30,7 +30,9 @@ export interface ImageDescriptionAndCaptionProps {
     linterContext: LinterContextProps;
     apiOptions: APIOptions;
     /**
-     * zoomSize is whichever one is larger between the natural size of the
+     * zoomSize represents the larger of the image’s natural size (calculated on load)
+     * and the saved backgroundImage size (specified when the content is written). This
+     * ensures that zooming is enabled only if the image is sufficiently large.
      * image (calculated on load) and the saved backgroundImage size (specified
      * when the content is written). This larger image size is used to
      * determine if the image is large enough to allow zooming.


packages/perseus/src/widgets/image/image.tsx
```diff
@@ -29,6 +30,21 @@ export const ImageComponent = (props: ImageWidgetProps) => {
...
+    React.useEffect(() => {
+        // Wait to figure out what the original size of the image is.
+        // Use whichever is larger between the original image size and the
+        // saved background image size for zooming.
+        Util.getImageSizeModern(backgroundImage.url!).then((size) => {
+            if (size[0] > (backgroundImage.width || 0)) {
+                setLargerImageSize(size);
+            }
+        });
+    }, [backgroundImage]);
```
nishasy, line 51: @jeremywiebe Do you know if this implementation has any issues?

I'm wondering if there's a reason we decide to save the image size in the widget data rather than use `Util.getImageSizeModern` on load. It helped me implement image resizing, but now I'd like to have both values (original size and preferred size). It would be nice to be able to use the util function to get the original size on load rather than have to save both sizes in the data.

jeremywiebe: I recall Nicole working on that Util.getImageSizeModern function but I don’t know the history of why we captured the size.

My guess is that it allows us to present a placeholder of the right size while we load the image… and then when we render the image, there’s no page jank because the placeholder consumed the exact same dimensions as the image.

nishasy: Okay, cool. I think that means the implementation in this PR should be fine. If it causes any loading or timing issues in the future (which I don't anticipate), we can revert this.

ivyolamit:
@nishasy looks like something broke in the [docs](https://650db21c3f5d1b2f13c02952-dwyjriqykv.chromatic.com/?path=/docs/widgets-image-widget-internal-components-exploreimagemodal--docs)

image

packages/perseus/src/widgets/image/components/image-description-and-caption.tsx
```diff
@@ -64,7 +60,7 @@ export const ImageDescriptionAndCaption = (
                 <figcaption
                     className="perseus-image-caption"
                     style={{
-                        maxWidth: backgroundImage.width,
+                        maxWidth: imageSize[0],
```
ivyolamit: magic number. Can we name the indexes of the imageSize, so it will be clear what index 0 or 1 is for to make it easy to read and debug.

packages/perseus/src/widgets/image/components/image-description-and-caption.tsx
```diff
@@ -29,18 +29,14 @@ export interface ImageDescriptionAndCaptionProps {
     range: [Interval, Interval];
     linterContext: LinterContextProps;
     apiOptions: APIOptions;
+    imageSize: [number, number];
```
ivyolamit: would this be a required field? please double check previous stories.

packages/perseus/src/widgets/image/components/image-description-and-caption.tsx
nishasy: I don't think we usually consider the 0/1 for height/width or x/y magic numbers, but I'll make a constant for that for clarity. Or I can update this to an object with named fields instead of an array of two numbers.

packages/perseus/src/widgets/image/components/image-description-and-caption.tsx
nishasy: I think it's supposed to be required. At the top level, it's set to the regular backgroundImage dimensions by default, or the new dimensions from the Util function if it's successful and it's bigger.

I just forgot to update the args in the `explore-image-modal.stories.tsx` 🤦🏽‍♀️

packages/perseus/src/widgets/image/components/explore-image-modal.test.tsx
```diff
@@ -20,6 +20,7 @@ const defaultProps = {
     caption: "",
     alt: "",
     longDescription: "",
+    imageSize: {width: 0, height: 0},
```
catandthemachines: nit: could we set this to a different height and width. Feels like we should not have our tests assume an image has this height and widget. 😉

nishasy: Hm... The default props don't have a backgroundImage either, so I'm not sure what would be appropriate to put here.

I think tests/stories are exceptions since they can render this modal standalone, but it wouldn't actually be possible for this modal to render without a backgroundImage/imageSize when used within the actual Image widget.

What do you suggest?

nishasy: I could make the image size optional, and then return null if it's undefined, or if the width or height are 0?

ivyolamit: please don't make it optional, just thinking about it when using an image widget, the required field will be the url and the imageSize. When a url is provided, the imageSize will be the image width and height. So we can add here `imageSize: {}` similar to the `backgroundImage` props which represents the url.

ivyolamit: just realizing this now, was there a reason why we are not using the backgroundImage props here?

```
export type PerseusImageBackground = {
    // The URL of the image
    url?: string | null;
    // The width of the image
    width?: number;
    // The height of the image
    height?: number;
    // The top offset of the image
    top?: number;
    // The left offset of the image
    left?: number;
    // The scale of the image
    scale?: number;
    // The bottom offset of the image
    bottom?: number;
};
```

nishasy: Talked about this offline - the backgroundImage is already being used, the new imageSize is specifically for zooming and it's not a new prop on the image widget.

Updated the name and added a doc comment for clarity.

packages/perseus/src/widgets/image/components/image-description-and-caption.tsx
jeremywiebe: This relates to Ivy's [comment](https://github.com/Khan/perseus/pull/2943#discussion_r2407202175) but also to this comment.  

There is a `Size` type in Perseus that you could use. This type names the two items in the tuple as `width` and `height`. I do agree that having named constants for the indexes is **not** something I'd want to do right now because we have this convention everywhere (its always `[width, height]`).

https://github.com/Khan/perseus/blob/ea232094e764e77d61067509cc137c5ed53f8f0e/packages/perseus-core/src/data-schema.ts#L41

packages/perseus/src/widgets/image/image.tsx
```diff
@@ -5,12 +5,15 @@ import AssetContext from "../../asset-context";
...
 import SvgImage from "../../components/svg-image";
 import Renderer from "../../renderer";
+import Util from "../../util";
 
 import {ImageDescriptionAndCaption} from "./components/image-description-and-caption";
 import styles from "./image-widget.module.css";
 
 import type {ImageWidgetProps} from "./image.class";
 
+export type ImageSize = {width: number; height: number};
```
jeremywiebe: Is there a reason you aren't using the [`Size`](https://github.com/Khan/perseus/blob/ea232094e764e77d61067509cc137c5ed53f8f0e/packages/perseus-core/src/data-schema.ts#L41) type here?

nishasy: You're right, it's better for me to switch to Size. It's a bummer that that's `[number, number]` though, so I'll have to use `[0]` and `[1]` instead of `.width` and `.height`.

packages/perseus/src/widgets/image/components/image-description-and-caption.tsx
```diff
@@ -29,18 +29,22 @@ export interface ImageDescriptionAndCaptionProps {
     range: [Interval, Interval];
     linterContext: LinterContextProps;
     apiOptions: APIOptions;
+    /**
+     * zoomSize is whichever one is larger between the natural size of the
```
ivyolamit:
```suggestion
     * zoomSize represents the larger of the image’s natural size (calculated on load)
     * and the saved backgroundImage size (specified when the content is written). This
     * ensures that zooming is enabled only if the image is sufficiently large.
```

packages/perseus/src/widgets/image/components/explore-image-modal.test.tsx
```diff
@@ -20,6 +20,7 @@ const defaultProps = {
     caption: "",
     alt: "",
     longDescription: "",
+    zoomSize: [0, 0] satisfies Size,
```
ivyolamit, line 23: does typescript not like without `satisfies Size,`?

nishasy: no, it says it's an array and it can be too big 🤦🏽‍♀️

PR #2945, opened 2025-10-02
[Radio] Bugfix - Borders next to correct choices should not show

mark-fitzgerald:
## Summary:
The choice borders next to a correct choice (when in review mode) should not be visible as it is distracting. This PR adjusts the styling to account for these situations.

Issue: LEMS-3456

## Test plan:
1. Open Storybook (use the link in this PR)
2. View the regression stories that "show solutions"
   - Single Select: Widgets => RadioNew => Visual Regression Tests => Initial State => Single Select Show Solutions
   - Multiple Select: Widgets => RadioNew => Visual Regression Tests => Initial State => Multiple Select Show Solutions
3. Notice that the horizontal borders between choices is not visible above and below choices that are marked correct

### Before
Screenshot 2025-10-02 at 3 10 51 PM

### After
Screenshot 2025-10-02 at 3 12 25 PM

### DIFF
--- a/.changeset/shy-cats-cheat.md
+++ b/.changeset/shy-cats-cheat.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+[Radio] Bugfix - Borders next to correct choices should not show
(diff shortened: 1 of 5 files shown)


nishasy (approved):
Looks much cleaner!

PR #2946, opened 2025-10-02
docs(changeset): Update the Radio Widget current version to 3.0

SonicScrewdriver:
## Summary:
[three-is-the-best-number] Increment the Radio Widget Version

Issue: LEMS-3604

## Test plan:

### DIFF
--- a/.changeset/angry-boxes-listen.md
+++ b/.changeset/angry-boxes-listen.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-core": minor
+---
+
+Update the Radio Widget current version to 3.0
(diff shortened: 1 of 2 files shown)


packages/perseus/src/__tests__/widgets.test.ts
```diff
@@ -6,17 +6,6 @@ describe("Widget API support", () => {
         registerAllWidgetsForTesting();
     });
 
-    describe("replaceWidget", () => {
```
SonicScrewdriver, line 9: I had to move this test under `getVersionVector` as replacing the transformer widget with the Radio widget was creating a false situation for the next test that confirms the version numbers. Since the Transformer widget has been removed and replaced with the Deprecated Stand-in widget, the Major version should be 0 — not 2 or 3.

mark-fitzgerald (approved):
Great piece of detective work!

benchristel (commented):
The code LGTM!

Can you add regression test data to `regression-tests/item-data` that includes a problematic Radio widget (v2 with rationale)? That will make sure we don't break this case in the future!

benchristel (approved):
LGTM!

PR #2950, opened 2025-10-03
[Radio] Bugfix - Selected choices don't turn blue when in articles

mark-fitzgerald:
## Summary:
Articles have extra styling that is overly broad, which prevents content in a multiple choice content from turning blue when selected. This PR adds a style rule with higher specificity to override it.

Issue: LEMS-3485

## Test plan:


SonicScrewdriver (approved):
Oooo great work, thanks for the new interaction story too!

PR #2951, opened 2025-10-03
[Image] | (CX) | Add a button to allow auto-conversion of markdown images to image widgets

nishasy:
## Summary:
Putting this up as a draft PR so I don't lose my work over the weekend.

Issue: https://khanacademy.atlassian.net/browse/LEMS-3569

## Test plan:
todo

### DIFF
--- a/packages/perseus-editor/src/util/issue-ctas-utils.test.ts
+++ b/packages/perseus-editor/src/util/issue-ctas-utils.test.ts
@@ -1,0 +1,61 @@
+import {
+    generateImageOptions,
+    generateImageWidget,
+    generateTestPerseusRenderer,
+} from "@khanacademy/perseus-core";
+
+import {convertImageMarkdownToImageWidget} from "./issue-ctas-utils";
+
+describe("convertImageMarkdownToImageWidget", () => {
+    it("should convert image markdown to image widget", () => {
+        const question = {
+            content: "![some alt text](https://example.com/image.png)",
+            widgets: {},
+            images: {},
+        };
+        const onEditorChange = jest.fn();
+        convertImageMarkdownToImageWidget(question, onEditorChange);
+        expect(onEditorChange).toHaveBeenCalledWith({
+            content: "image",
+        });
+    });
+
+    it("should convert image markdown to image widget with surrounding text", () => {
+        const question = {
+            content:
+                "Hello ![some alt text](https://example.com/image.png) World",
+            widgets: {},
+            images: {},
+        };
+        const onEditorChange = jest.fn();
+        convertImageMarkdownToImageWidget(question, onEditorChange);
+        expect(onEditorChange).toHaveBeenCalledWith({
+            content: "Hello image World",
+        });
+    });
+
+    it("should convert the image markdown and keep surrounding widgets", () => {
+        const question = generateTestPerseusRenderer({
+            content:
+                "Hello ![some alt text](https://example.com/image.png) World [[☃ image 1]]",
+            widgets: {
+                "image 1": generateImageWidget({
+                    options: generateImageOptions({
+                        backgroundImage: {
+                            url: "https://example.com/image.png",
+                            width: 100,
+                            height: 100,
+                        },
+                        alt: "some alt text",
+                    }),
+                }),
+            },
+        });
+
+        const onEditorChange = jest.fn();
+        convertImageMarkdownToImageWidget(question, onEditorChange);
+        expect(onEditorChange).toHaveBeenCalledWith({
+            content: "Hello image World [[☃ image 1]]",
+        });
+    });
+});
--- a/packages/perseus-editor/src/util/issue-ctas-utils.test.ts
+++ b/packages/perseus-editor/src/util/issue-ctas-utils.test.ts
@@ -15,7 +15,8 @@ describe("convertImageMarkdownToImageWidget", () => {
 
     beforeEach(() => {
         unmockImageLoading = mockImageLoading();
-        jest.spyOn(Util, "getImageSizeModern").mockResolvedValue([400, 255]);
+        // Size of earthMoonImage
+        jest.spyOn(Util, "getImageSizeModern").mockResolvedValue([400, 225]);
     });
 
     afterEach(() => {
@@ -23,7 +24,7 @@ describe("convertImageMarkdownToImageWidget", () => {
         jest.restoreAllMocks();
     });
 
-    it("should convert image markdown to image widget", async () => {
+    it("converts image markdown to image widget", async () => {
         const question = {
             content: `![some alt text](${earthMoonImage.url})`,
             widgets: {},
@@ -36,11 +37,7 @@ describe("convertImageMarkdownToImageWidget", () => {
             widgets: {
                 "image 1": generateImageWidget({
                     options: generateImageOptions({
-                        backgroundImage: {
-                            url: earthMoonImage.url,
-                            width: 400,
-                            height: 255,
-                        },
+                        backgroundImage: earthMoonImage,
                         alt: "some alt text",
                     }),
                 }),
@@ -48,7 +45,7 @@ describe("convertImageMarkdownToImageWidget", () => {
         });
     });
 
-    it("should convert image markdown to image widget with surrounding text", async () => {
+    it("converts image markdown to image widget with surrounding text", async () => {
         const question = {
             content: `Hello ![some alt text](${earthMoonImage.url}) World`,
             widgets: {},
@@ -61,11 +58,7 @@ describe("convertImageMarkdownToImageWidget", () => {
             widgets: {
                 "image 1": generateImageWidget({
                     options: generateImageOptions({
-                        backgroundImage: {
-                            url: earthMoonImage.url,
-                            width: 400,
-                            height: 255,
-                        },
+                        backgroundImage: earthMoonImage,
                         alt: "some alt text",
                     }),
                 }),
@@ -73,17 +66,13 @@ describe("convertImageMarkdownToImageWidget", () => {
         });
     });
 
-    it("should convert the image markdown and keep surrounding widgets", async () => {
+    it("converts the image markdown and keep surrounding widgets", async () => {
         const question = generateTestPerseusRenderer({
             content: `Hello ![some alt text](${earthMoonImage.url}) World [[☃ image 1]]`,
             widgets: {
                 "image 1": generateImageWidget({
                     options: generateImageOptions({
-                        backgroundImage: {
-                            url: earthMoonImage.url,
-                            width: 400,
-                            height: 255,
-                        },
+                        backgroundImage: earthMoonImage,
                         alt: "some alt text",
                     }),
                 }),
@@ -97,21 +86,13 @@ describe("convertImageMarkdownToImageWidget", () => {
             widgets: {
                 "image 1": generateImageWidget({
                     options: generateImageOptions({
-                        backgroundImage: {
-                            url: earthMoonImage.url,
-                            width: 400,
-                            height: 255,
-                        },
+                        backgroundImage: earthMoonImage,
                         alt: "some alt text",
                     }),
                 }),
                 "image 2": generateImageWidget({
                     options: generateImageOptions({
-                        backgroundImage: {
-                            url: earthMoonImage.url,
-                            width: 400,
-                            height: 255,
-                        },
+                        backgroundImage: earthMoonImage,
                         alt: "some alt text",
                     }),
                 }),
@@ -119,7 +100,7 @@ describe("convertImageMarkdownToImageWidget", () => {
         });
     });
 
-    it("should conver multiple images", async () => {
+    it("converts multiple images", async () => {
         const question = generateTestPerseusRenderer({
             content: `![alt 1](${earthMoonImage.url}) ![alt 2](${earthMoonImage.url})`,
             widgets: {},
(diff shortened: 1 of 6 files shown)


packages/perseus-editor/src/util/issue-ctas-utils.test.ts
nishasy, line 1: Don't forget to review this. It's hidden by default because it's so big.

packages/perseus-editor/src/components/issue-cta.tsx
```diff
@@ -0,0 +1,43 @@
...
+            style={styles.button}
+        >
+            {cta.label}
+        </Button>
+    );
+};
+
+export default IssueCta;
+
+// TODO: Use CSS modules after Wonder Blocks styles
```
catandthemachines, line 36: Is there any bug or issue tracking that move? If so could we add the issue id here? 😊

packages/perseus-editor/src/issue-details.tsx
```diff
@@ -48,6 +62,13 @@ const IssueDetails = ({issue}: IssueProps) => {
                 Issue:
             </LabelSmall>
             <span>{issue.message}</span>
+            {imageUpgradeFF && (
+                <IssueCta
+                    issue={issue}
+                    question={question}
```
catandthemachines: Side idea (non blocking at all but want to surface), as we update more of our editor code to use more modern React functionality, I think `question` and other properties would be great candidates to build a context. Thus removing the need to pass it down so many sub components.

Do you know if that is something you could try? Again, non blocking but I'm curious if that is something that we can add here, or file a bug for future improvement.

packages/perseus-editor/src/issue-cta.tsx
```diff
@@ -0,0 +1,43 @@
...
+import * as React from "react";
+
+import {getCtaForIssueId} from "./util/issue-ctas-utils";
+
+import type {Issue} from "./issues-panel";
+import type {PerseusRenderer} from "@khanacademy/perseus-core";
+
+interface Props {
+    issue: Issue;
+    question?: PerseusRenderer;
```
ivyolamit: will there be a case that the question is not provided, in this context here?
I think it's safe to assume when we do show the CTA the item being edited has an equivalent question, this will save us additional check on `if (!question)` and `if (!cta)`.

packages/perseus-editor/src/components/issue-cta.tsx
ivyolamit, line 1: what do you think of moving these `issue-*` related components under `packages/perseus-editor/src/components`? since they're just sub-components of packages/perseus-editor/src/item-editor.tsx

packages/perseus-editor/src/issue-cta.tsx
```diff
@@ -0,0 +1,43 @@
...
+    onEditorChange: (newProps: any) => void;
+}
+
+const IssueCta = ({issue, question, onEditorChange}: Props) => {
```
ivyolamit: Do you know what other scenarios we can use the IssueCta?

packages/perseus-editor/src/components/issue-cta.tsx
```diff
@@ -0,0 +1,43 @@
...
+// are moved to a different layer.
+const styles = {
+    button: {
+        marginTop: sizing.size_080,
```
ivyolamit, line 40: we can also add a marginBottom here so the button will not be too close to the bottom of the panel.
```suggestion
        marginBottom: sizing.size_080,
```

packages/perseus-editor/src/components/issue-details.tsx
```diff
@@ -1,20 +1,34 @@
...
 import * as React from "react";
 
 import PerseusEditorAccordion from "./components/perseus-editor-accordion";
+import IssueCta from "./issue-cta";
 
 import type {Issue} from "./issues-panel";
+import type {APIOptions} from "@khanacademy/perseus";
 
 type IssueProps = {
+    apiOptions?: APIOptions;
```
ivyolamit, line 14: just an observation no need to address this in your PR: this is painful to watch to prop drill these props 2-3 levels deep 😢 it would be ideal to have the editor context available so we can avoid prop drilling.

packages/perseus-editor/src/components/issue-details.tsx
```diff
@@ -1,20 +1,34 @@
...
+const IssueDetails = ({
+    apiOptions,
+    issue,
+    question,
+    onEditorChange,
+}: IssueProps) => {
     const [expanded, setExpanded] = React.useState(false);
     const toggleVisibility = () => setExpanded(!expanded);
 
+    const imageUpgradeFF = isFeatureOn({apiOptions}, "image-widget-upgrade");
```
ivyolamit, line 24:
```suggestion
    // TODO(LEMS-3520): Feature flag cleanup
    const imageUpgradeFF = isFeatureOn({apiOptions}, "image-widget-upgrade");
```

packages/perseus-editor/src/components/issues-panel.tsx
```diff
@@ -20,10 +23,18 @@ export type Issue = {
 };
 
 type IssuesPanelProps = {
+    apiOptions?: APIOptions;
```
ivyolamit, line 27: do you think we would need this in the future? currently we are only using it for the feature flag.
```suggestion
    // TODO(LEMS-3520)
    apiOptions?: APIOptions;
```

packages/perseus-editor/src/util/issue-ctas-utils.ts
```diff
@@ -0,0 +1,112 @@
...
+        const imageUrl = imageNode[0].content[0].target;
+        const imageAlt = imageNode[0].content[0].alt;
+
+        const imageSize = await Util.getImageSizeModern(imageUrl);
+
+        newWidgets[`image ${imageIndex}`] = generateImageWidget({
+            options: generateImageOptions({
+                backgroundImage: {
+                    url: imageUrl,
+                    width: imageSize[0],
```
ivyolamit: do you remember what was the conclusion of our discussion about using `0` and `1` for the array of image size?

packages/perseus-editor/src/util/issue-ctas-utils.ts
```diff
@@ -0,0 +1,112 @@
...
+        newContent = newContent.replace(original, replacement);
+    }
+
+    onEditorChange({
+        content: newContent,
+        widgets: newWidgets,
+    });
+}
+
+export function getCtaForIssueId(
```
ivyolamit, line 122: please add a description/code comment about this function so it will be easier for anyone using this in the future to add.

packages/perseus-editor/src/util/issue-ctas-utils.ts
```diff
@@ -0,0 +1,112 @@
...
+};
+
+// Exported for testing
+export function getFirstAvailableWidgetIndex(
+    widgetType: string,
+    widgets: PerseusWidgetsMap,
+): number {
+    const widgetNames = Object.keys(widgets);
+
+    for (let i = 1; i <= widgetNames.length + 1; i++) {
```
ivyolamit, line 35: Why start at 1 here instead of 0?
and related question in line 30 why is it returning 1 instead of 0 the first item in the array?

ivyolamit (commented):
@nishasy here's my first pass review comments, i'll be reviewing the big test file you mentioned next and smoke testing.

packages/perseus-editor/src/components/issue-cta.tsx
nishasy: I'll create a new ticket, and I think I'll go through the codebase in a separate PR later to add the ticket reference on all the comments for this.

packages/perseus-editor/src/issue-cta.tsx
nishasy: That's the weird thing. I don't _think_ there's a case where there isn't a question, but question is optional inside the props for the item editor. BUT it does have default props setting question to `{}`, so I think you're right.

packages/perseus-editor/src/issue-cta.tsx
nishasy: We don't have any yet, but we can add them as they come up in the future. Maybe @catandthemachines could have some ideas since she did more work with the issues panel and linter in general?

packages/perseus-editor/src/components/issue-details.tsx
nishasy: Yeah, Cat said the same thing. I'll look into adding a context.

packages/perseus-editor/src/components/issues-panel.tsx
nishasy: I don't think we'll need it in the future. I figured we'd remove it along with the feature flag cleanup.

packages/perseus-editor/src/util/issue-ctas-utils.ts
nishasy: Oh, I think I was destructuring everywhere but I might have written this before we came to that conclusion. I'll update it.

packages/perseus-editor/src/util/issue-ctas-utils.ts
ivyolamit: This might be related it has an additional panel that doesn't open
image

nishasy: All the widget names start with 1. If you add a new image, its name gets saved as "image 1" and so on, so I'm just staying with that pattern.

I can switch the for loop to start at 0, but then I'd have to return i + 1 instead, so I think it doesn't really matter in the end.

nishasy: I'll add a comment to explain this better.

ivyolamit: for the panel that opens, i don't know how i got that state 🙈

packages/perseus-editor/src/issue-cta.tsx
nishasy: I think we still need the `!cta` check because that means that there's no CTA associated with the issue ID, so there's no need to render a CTA button.

packages/perseus-editor/src/util/item-editor-context.ts
catandthemachines, line 1: Yay thank you!

packages/perseus-editor/src/util/issue-ctas-utils.ts
```diff
@@ -0,0 +1,138 @@
...
+    const matches = Array.from(content.matchAll(IMAGE_MARKDOWN_REGEX));
+
+    if (matches.length === 0) {
+        return; // No image markdown to convert
+    }
+
+    // Process each match sequentially to ensure unique widget indices
+    const replacements: Array<{original: string; replacement: string}> = [];
+    for (const match of matches) {
+        const imageIndex = getFirstAvailableWidgetIndex("image", newWidgets);
```
catandthemachines: I see every loop you ask "what is the first avaliable image widget index"? if you have multiple markdown images, don't you need to only ask this before the loop, and shouldn't you increment it each time you loop through the matches?

Example:
Content with one image widget in it but 2 markdown images, you find the first image widget at 1, and add the first markdown at "image 1", the second time around it would do it again and add an image with "image 1". (I haven't run this locally so let me know if I'm off or there's a piece of logic I'm missing here).

packages/perseus-editor/src/util/issue-ctas-utils.test.ts
```diff
@@ -0,0 +1,455 @@
...
+                    options: generateImageOptions({
+                        backgroundImage: earthMoonImage,
+                        alt: "some alt text",
+                    }),
+                }),
+            },
+        });
+    });
+
+    it("converts multiple images", async () => {
```
catandthemachines, line 241: Can we add a case of an existing image widget with 1 or 2 markdown images and see how it does?

catandthemachines: Ah looks like there is a test `converts multiple images with skipped indices` that does that never mind 😉

packages/perseus-editor/src/__docs__/editor-page-with-storybook-preview.tsx
```diff
@@ -64,6 +64,7 @@ function EditorPageWithStorybookPreview(props: Props) {
     );
 
     const apiOptions = props.apiOptions ?? {
+        flags: getFeatureFlags({"image-widget-upgrade": true}),
```
ivyolamit: Please remove this before landing the PR

packages/perseus-editor/src/util/item-editor-context.ts
ivyolamit: ❤️

ivyolamit (commented):
Thanks @nishasy great work here!
Changes looks good to me.
 
| Warning | Button Clicked |
| -------- | -------------- |
| image | image |

packages/perseus-editor/src/util/issue-ctas-utils.ts
```diff
@@ -0,0 +1,138 @@
...
+ */
+export async function convertImageMarkdownToImageWidget(
+    question: PerseusRenderer,
+    onEditorChange: (newProps: any) => void,
+) {
+    const {content, widgets} = question;
+    const newWidgets = {...widgets};
+
+    // Find all image markdown matches
+    const matches = Array.from(content.matchAll(IMAGE_MARKDOWN_REGEX));
```
jeremywiebe: We should not be inspecting the `content` using regexes.  Instead we should use Perseus' Markdown-parsing library and search that way.  See the [FAQ](https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/4300800034/Perseus+FAQ) for details.

PR #2952, opened 2025-10-03
[Keypad] Update Keypad to use WonderBlocks Tabs component

ivyolamit:
## Summary:
Refactor Keypad to use WB Tabs instead of Tabbar:
1. Add the WB Tabs component in Perseus
2. Code Refactor to use WB Tabs in Keypad component
3. Delete unused Tabbar component

Issue: LEMS-3361

## Test plan:


packages/math-input/src/components/keypad/__tests__/keypad.test.tsx
ivyolamit, line 1: Dismiss button was removed from the keypad and uses the pop-over close button on the parent component.

packages/math-input/src/components/keypad/utils/get-available-tabs.tsx
ivyolamit, line 1: This used to be inside the Keypad component. Moved it to a utility function to dry the Keypad code. Intention is still the same as the original but the logic is now follows the props required for the [WB Tabs component](https://khan.github.io/wonder-blocks/?path=/docs/packages-tabs-tabs--docs&globals=theme:default).

packages/math-input/src/full-mobile-input.stories.tsx
```diff
@@ -53,7 +53,7 @@ const Basic = ({keypadElement, setKeypadElement}) => {
     }, [keypadElement, expression, times]);
 
     return (
-        <div style={{padding: "1rem 2rem"}}>
+        <div style={{padding: "1rem 2rem", minHeight: "400px"}}>
```
ivyolamit, line 56: added minHeight here so that the keypad will be visible, current height doesn't show the entire keypad

Image

packages/math-input/src/components/keypad/utils/get-cursor-context-config.ts
```diff
@@ -1,7 +1,7 @@
-import Keys from "../../data/key-configs";
-import {CursorContext} from "../input/cursor-contexts";
+import Keys from "../../../data/key-configs";
+import {CursorContext} from "../../input/cursor-contexts";
 
-import type {MathInputStrings} from "../../strings";
+import type {MathInputStrings} from "@khanacademy/math-input/strings";
```
handeyeco: Packages shouldn't import from themselves. I was thinking we had a lint rule for this 🤔 

`math-input` shouldn't import from `@khanacademy/math-input`

packages/math-input/src/components/keypad/utils/index.ts
```diff
@@ -0,0 +1,5 @@
+export {getAvailableTabs} from "./get-available-tabs";
```
handeyeco: I don't really get the point of the `index.ts` file here. Now there are two ways to import these:

```
import {getAvailableTabs} from "./utils"
import {getAvailableTabs} from "./utils/get-available-tabs"
```
IMO, for the sake of consistency, it would be better to only have one way.

packages/perseus/src/components/math-input.tsx
```diff
@@ -1,5 +1,3 @@
-/* eslint-disable @khanacademy/ts-no-error-suppressions */
```
handeyeco, line 1: Why add a folder for `math-input` if it only contains the `math-input.tsx` file? Doesn't seem like the folder is needed but, if it is, maybe it would make sense for this file to be `index.ts` so that we don't have to update imports.

packages/perseus/src/components/math-input.tsx
```diff
@@ -13,7 +11,7 @@ import {
 import Clickable from "@khanacademy/wonder-blocks-clickable";
 import {View} from "@khanacademy/wonder-blocks-core";
 import {Popover, PopoverContentCore} from "@khanacademy/wonder-blocks-popover";
-import {color, spacing} from "@khanacademy/wonder-blocks-tokens";
```
handeyeco, line 16: (comment on file/dir, not this line of code)

I don't think we need to add a `math-input` folder if the only file in it is just `math-input.tsx`. If for some reason we do need the `math-input` folder, maybe it would make sense for this file to be `index.tsx` so we don't need to update imports.

packages/perseus/src/components/math-input/math-input.tsx
```diff
@@ -13,7 +11,7 @@ import {
 import Clickable from "@khanacademy/wonder-blocks-clickable";
 import {View} from "@khanacademy/wonder-blocks-core";
 import {Popover, PopoverContentCore} from "@khanacademy/wonder-blocks-popover";
-import {color, spacing} from "@khanacademy/wonder-blocks-tokens";
+import {color, semanticColor, sizing} from "@khanacademy/wonder-blocks-tokens";
 import {HeadingMedium} from "@khanacademy/wonder-blocks-typography";
 import {StyleSheet} from "aphrodite";
```
handeyeco: (comment on file/dir, not this line of code)

I don't think we need to add a `math-input` folder if the only file in it is just `math-input.tsx`. If for some reason we do need the `math-input` folder maybe it would make sense if this file was `index.tsx` so that we don't need to update imports.

packages/math-input/src/components/keypad/keypad.tsx
```diff
@@ -130,108 +80,73 @@ export default function Keypad({extraKeys = [], ...props}: Props) {
...
+                )}
+                <Tabs
+                    aria-label="Keypad"
+                    tabs={availableTabs}
+                    selectedTabId={selectedPage}
+                    onTabSelected={(newSelectedPage: string) => {
+                        setSelectedPage(newSelectedPage as KeypadPageType);
+                    }}
+                    styles={{
+                        tab: {
```
handeyeco: Can we move these styles out of the JSX rendering? If we move them out of the component it will make the JSX easier to read and we'll get an added bonus in that we won't have to redeclare them every time we rerender.

packages/perseus/src/components/math-input.tsx
```diff
@@ -281,6 +278,12 @@ class InnerMathInput extends React.Component<InnerProps, State> {
     blur: () => void = () => this.setState({focused: false});
 
     handleKeypadPress: (key: KeypadKey, e: any) => void = (key, e) => {
+        // Handle dismiss key to close the keypad
+        if (key === "DISMISS") {
```
handeyeco, line 283: Do we still need a "DISMISS" key? Seems like that's obsolete now if we're using WB.

packages/math-input/src/components/keypad/keypad-fractions-only.tsx
```diff
@@ -0,0 +1,54 @@
...
+import * as React from "react";
+
+import FractionsPage from "./keypad-pages/fractions-page";
+import styles from "./keypad.module.css";
+import {RenderKeyPadPanel} from "./render-keypad-panel";
+
+import type {KeypadProps} from "./keypad";
+import type {KeypadPageType} from "../../types";
+
+export type KeypadFractionsOnlyProps = Pick<
```
handeyeco, line 12: It's not clear to me why we need both KeypadFractionsOnly and FractionsPage. Couldn't KeypadFractionsOnly just be part of Keypad?

packages/math-input/src/components/keypad/keypad.tsx
```diff
@@ -1,25 +1,20 @@
...
+import {KeypadFractionsOnly} from "./keypad-fractions-only";
+import styles from "./keypad.module.css";
+import {getAvailableTabs} from "./utils";
 
 import type {ClickKeyCallback, KeypadPageType} from "../../types";
 import type {CursorContext} from "../input/cursor-contexts";
 import type {PerseusAnalyticsEvent, KeypadKey} from "@khanacademy/perseus-core";
 
-type Props = {
+export interface KeypadProps {
```
handeyeco, line 17: Just curious: is there a reason for switching from `type` to `interface`?

handeyeco (commented):
It's a huge win when we can use someone else's code instead of maintaining our own version of things! Thanks for digging into this. 🌈 

In the future it might be helpful to try stacked PRs. This PR accomplishes a lot but not everything is related so it's can be hard to follow.

packages/perseus/src/components/math-input.tsx
ivyolamit: Unfortunately yes,
- Initial approach was using the WB Tabs for the x button and had it styled to be positioned on the right most away from the other tabs, that didn't looked so good
- Second approach is using Popover x button but had to remove that solution because other usages of the keypad doesn't have a popover
- So i added back in the DISMISS key and hooked it in IconButton and positioned it at the right of the Tabs component

ivyolamit:
ZND Link: https://prod-znd-251117-5339-ce5d6e6.khanacademy.org/

packages/math-input/src/components/keypad/keypad.tsx
ivyolamit: Our team had a [discussion](https://khanacademy.slack.com/archives/CJDRXTGQ7/p1744230806279719?thread_ts=1744223490.274449&cid=CJDRXTGQ7) about this preference using interface.

packages/perseus/src/components/math-input.tsx
ivyolamit: Good catch! this was a left over of my initial refactor attempt. I'll update that.

packages/math-input/src/components/keypad/keypad-fractions-only.tsx
ivyolamit: During the refactor i've separated the logic
1. Displaying of the fractions only without any tabs
2. Displaying of fractions together with other tabs

Both of this uses the FractionsPage, the first one is rendered directly from the KeyPad while the second one is is rendered via the util getAvailableTabs which will form the props that will be provided to the WB Tabs component.

Initially with #1 I added it in the Keypad component, but since it has a number of lines of code i decided to make it a different component so it will be easier to read in Keypad component like this (making the code dry):
```
// We don't want to show any available tabs on the fraction keypad
    if (fractionsOnly) {
        return <KeypadFractionsOnly {...props} />;
    }
```

packages/perseus/src/components/math-input.tsx
ivyolamit: @handeyeco we do have potential code that we can remove though related to DIMISS. I'm just not confident if it's safe to remove now that we use the WB IconButton, you might have more context to this:
- https://github.com/Khan/perseus/blob/main/packages/math-input/src/components/keypad/button-assets.tsx#L388
- packages/math-input/src/components/keypad/tab-icon-label.tsx (this one will cause a ts error)

packages/math-input/src/components/keypad/__tests__/keypad.test.tsx
ivyolamit: Using popover close button was removed since not all use cases for the keypad uses a Popover like the mobile.

packages/math-input/src/components/keypad/keypad.tsx
```diff
@@ -42,69 +37,42 @@ type Props = {
             {type: "math-input:keypad-opened" | "math-input:keypad-closed"}
         >,
     ) => void;
-};
-
-function getAvailableTabs(props: Props): ReadonlyArray<KeypadPageType> {
```
ivyolamit, line 47: This logic is now in https://github.com/Khan/perseus/pull/2952/files#diff-d477508c258a48e849093b97a9ccc4bce701586040fdc70183deb9ab85454e41

packages/math-input/src/components/keypad/keypad.tsx
```diff
@@ -130,108 +98,54 @@ export default function Keypad({extraKeys = [], ...props}: Props) {
...
+                        setSelectedPage(newSelectedPage as KeypadPageType);
+                    }}
+                    styles={tabsStyles}
+                />
             </View>
         </View>
     );
 }
-
-const styles = StyleSheet.create({
```
ivyolamit, line 204: Moved styling in [keypad.module.css](https://github.com/Khan/perseus/pull/2952/files#diff-305139668d735c1cb3eff30183ebd4be51d7bfd6eee7896721946482d84ccf1f)

ivyolamit:
> It's a huge win when we can use someone else's code instead of maintaining our own version of things! Thanks for digging into this. 🌈
> 
> In the future it might be helpful to try stacked PRs. This PR accomplishes a lot but not everything is related so it's can be hard to follow.

Thanks @handeyeco for the review and i'll take note on your suggestion for stacking the PRs in the future 🙏🏼 
I've updated the PR to address your comments and this is the latest [ZND](https://prod-znd-251117-5339-9de50e9.khanacademy.org/internal-courses/test-everything/test-everything-1/expression/e/expression-exercise). When you have time please have another look 🙏🏼

packages/math-input/src/components/keypad/keypad.tsx
```diff
@@ -130,108 +98,54 @@ export default function Keypad({extraKeys = [], ...props}: Props) {
         };
     }, [onAnalyticsEvent, isMounted]);
 
+    // We don't want to show any available tabs on the fraction keypad
+    if (fractionsOnly) {
+        return <KeypadFractionsOnly {...props} />;
```
handeyeco, line 103: I'm not a huge fan of the name. What about `FractionKeypad`? If we're splitting out FractionKeypad, I wonder if it would make sense to also split out a ExpressionKeypad too. Then `Keypad`s return could be something like

```TSX
return fractionsOnly ? <FractionKeypad {...props} /> : <ExpressionKeypad {...props} />
```

Although I'm not a fan of spreading props either. For instance you're still passing `advancedRelations` to `FractionKeypad` even though it's not used (and the type system doesn't allow using it). I think it makes more sense to explicitly pass props except in places where you really want to pass all props (like in `forwardRef`).

ivyolamit: Thanks @handeyeco i'll look into it and do a follow-up PR.

ivyolamit:
👆🏼 Rebased from latest master which includes the adding for the packages, to minimize the merge conflict later as package files are often updated.
See: https://github.com/Khan/perseus/pull/3054

PR #3189, opened 2026-01-20
Delete `ServerItemRenderer.getUserInputLegacy()`

benchristel:
## Summary:
This method was deprecated, and is no longer used by any callers within Khan
Academy.

Issue: LEMS-2523

## Test plan:

CI should be green.

### DIFF
--- a/testing/item-renderer-hooks.tsx
+++ b/testing/item-renderer-hooks.tsx
@@ -100,11 +100,11 @@ export const useItemRenderer = (
             "en",
         );
 
+        const widgetIds = renderer.getWidgetIds();
+        const userInputArray = widgetIds.map((id) => userInput[id]);
+
         // Continue to include an empty guess for the now defunct answer area.
-        const maxCompatGuess: [UserInput[], []] = [
-            renderer.getUserInputLegacy(),
-            [],
-        ];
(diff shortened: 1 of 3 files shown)


testing/item-renderer-hooks.tsx
```diff
@@ -100,11 +100,11 @@ export const useItemRenderer = (
...
 
+        const widgetIds = renderer.getWidgetIds();
+        const userInputArray = widgetIds.map((id) => userInput[id]);
+
         // Continue to include an empty guess for the now defunct answer area.
-        const maxCompatGuess: [UserInput[], []] = [
-            renderer.getUserInputLegacy(),
-            [],
-        ];
+        const maxCompatGuess: [UserInput[], []] = [userInputArray, []];
```
handeyeco, line 107: I wonder if we could just get rid of this stuff. Really this isn't part of Perseus anymore, it's legacy code that lives in mostly in TUT now.

handeyeco (approved):

testing/item-renderer-hooks.tsx
benchristel: I plan to! That will be a near-future PR.

PR #3212, opened 2026-01-28
Remove sloc and pnpm as package dependencies - unused

jeremywiebe:
## Summary:

I was checking something in Perseus and noticed neither of these dependencies are needed. I suspect we haven't used the `sloc` tool (used to count lines of code) in a long time (and if we want to, we can install it manually run it, and move on). 

Issue: --none--

## Test plan:


package.json
```diff
@@ -98,7 +98,6 @@
         "jest-specific-snapshot": "^5.0.0",
         "knip": "^5.33.3",
         "nyc": "^15.1.0",
-        "pnpm": "^9.15.4",
```
jeremywiebe: We don't need to "depend" on `pnpm` this way... plus, we use pnpm v10 not v9 today.

benchristel (approved):
I've used `sloc`, actually. But I can install it globally. I agree there's not much point in having it in `scripts`.

LGTM!

PR #3235, opened 2026-02-06
[wip] Add prompt library docs and update CLAUDE.md

Myranae:
## Summary:
Shorten our CLAUDE.md by referencing sub-documents on specific topics. The goal is to use progressive disclosure to keep the main file short so context stays as relevant as possible. I will be testing these out on my next couple tickets to see if it works well, but would love feedback!


.claude/architecture-research.md
Myranae, line 1: This document is what was used to find updated information for all these docs. It would not be part of the final merge. Will delete this after discussions. Kept for reference.

.claude/widget-patterns-research.md
Myranae, line 1: This document is what was used to find updated information for all these docs. It would not be part of the final merge. Will delete this after discussions. Kept for reference.

CLAUDE.md
Myranae, line 1: It may be possible to trim this file even more, but we do want some amount of info here.

.claude/prompts/component-best-practices.md
```diff
@@ -0,0 +1,324 @@
...
+    return <div>{props.content}</div>;
+}
+ComponentName.displayName = "ComponentName"; // Explicit for debugging
+
+// ❌ Avoid - Arrow function requires displayName
+const ComponentName = (props: Props) => {
+    return <div>{props.content}</div>;
+};
+ComponentName.displayName = "ComponentName"; // Extra step
+```
```
mark-fitzgerald, line 18: I'm having trouble understanding the difference between these two examples. I understand that one is a named function, and the other is an arrow function, but I don't see the "Extra step" that is called out. I also don't know why the named function is the preferred form.

.claude/prompts/component-best-practices.md
```diff
@@ -0,0 +1,324 @@
...
+### Type Definitions
+```typescript
+// Place types at the top of the file
+type Props = {
+    content: string;
+    onAction: () => void;
+    optional?: boolean;
+};
+
+// For complex props, use intersection types
```
mark-fitzgerald, line 29: Did we decide that we prefer interfaces when we export types? I don't recall if that impacts how we define complex props.

.claude/prompts/component-best-practices.md
```diff
@@ -0,0 +1,324 @@
...
+
+```typescript
+import {View} from "@khanacademy/wonder-blocks-core";
+import {Button} from "@khanacademy/wonder-blocks-button";
+import {TextField} from "@khanacademy/wonder-blocks-form";
+import {LabelMedium} from "@khanacademy/wonder-blocks-typography";
+
+function MyComponent() {
+    return (
+        <View>
```
mark-fitzgerald, line 47: Perhaps this is more appropriate for a separate discussion: I would like to get away from using the `<View>` component as a standard "element". That component has some core styling associated with it (namely `flexbox`) that isn't always needed. Generally speaking, if we want a Perseus component to output elements that "need" a container with specific styling, then a `<div>` element (or other more semantically relevant element) with styling attached to it is my preference (simpler styling rather than trying to override `<View>`). However, if the component is contained within some other component that manages its allotted area, then my preference is to return the component content surrounded by React fragments (`<> ... </>`). Otherwise, if we promote using `<View>`, we end up with div-soup, which is cumbersome and inefficient. The `<View>` component has its place, but I think we should be judicious with it.

.claude/prompts/component-best-practices.md
```diff
@@ -0,0 +1,324 @@
...
+### Component State (UI only)
+```typescript
+// Local state for UI concerns only
+function Component() {
+    const [isExpanded, setIsExpanded] = React.useState(false);
+    const [hoveredIndex, setHoveredIndex] = React.useState<number | null>(null);
+
+    // User answer state goes through handleUserInput, not local state!
+}
+```
```
mark-fitzgerald, line 73: Could you expand on why this is helpful for the AI? Doesn't it already know about state management?

.claude/prompts/component-best-practices.md
```diff
@@ -0,0 +1,324 @@
...
+
+### Derived State
+```typescript
+// Calculate values instead of storing them
+function Component({items, filter}: Props) {
+    // ✅ Good - Derived from props
+    const filteredItems = React.useMemo(
+        () => items.filter(item => item.matches(filter)),
+        [items, filter]
+    );
```
mark-fitzgerald, line 83: We may want to check in with Jeff Yates on the matter of `useMemo`. I believe he has some well-founded opinions on the matter of when (if ever) to use this feature.

.claude/prompts/component-best-practices.md
```diff
@@ -0,0 +1,324 @@
...
+```
+
+### Avoid Inline Objects/Functions
+```typescript
+// ❌ Bad - Creates new object every render
+<Component style={{margin: 10}} />
+
+// ✅ Good - Stable reference
+const style = {margin: 10};
+<Component style={style} />
```
mark-fitzgerald, line 118: This instruction is combining two concepts: inline object definition and how to style. I would like to keep them separate. Is there a different inline object definition that could be used as an example? I don't want to give the AI any idea that it can provide styling inline.

.claude/prompts/component-best-practices.md
```diff
@@ -0,0 +1,324 @@
...
+## Accessibility
+
+### ARIA Labels
+```typescript
+<input
+    aria-label="Answer"
+    aria-describedby="hint-text"
+    aria-invalid={hasError}
+/>
+```
```
mark-fitzgerald, line 133: What is this telling the AI to do? If it is trying to enforce the use of labels/names on element, then I think this needs to be expanded significantly. My experience has been that the AI doesn't know when it is appropriate to add `aria` and when it isn't, nor how to properly generate a label.

.claude/prompts/component-best-practices.md
```diff
@@ -0,0 +1,324 @@
...
+    // Auto-focus on mount when appropriate
+    React.useEffect(() => {
+        if (shouldAutoFocus) {
+            inputRef.current?.focus();
+        }
+    }, [shouldAutoFocus]);
+
+    return <input ref={inputRef} />;
+}
+```
```
mark-fitzgerald, line 149: Is focus management something that we do in Perseus?

.claude/prompts/component-best-practices.md
```diff
@@ -0,0 +1,324 @@
...
+        case "Escape":
+            clearInput();
+            break;
+        case "ArrowUp":
+            selectPrevious();
+            e.preventDefault(); // Prevent scroll
+            break;
+    }
+}
+```
```
mark-fitzgerald, line 167: Does having this in the document encourage the AI to implement keyboard handling over native functionality?

.claude/prompts/component-best-practices.md
```diff
@@ -0,0 +1,324 @@
...
+function Component() {
+    const isMobile = useIsMobile();
+
+    return (
+        <View flexDirection={isMobile ? "column" : "row"}>
+            {/* Content adapts to screen size */}
+        </View>
+    );
+}
+```
```
mark-fitzgerald, line 214: I like this concept. However, I would prefer that it is handled via CSS, since that would work better in a wider variety of situations (i.e. zoomed viewport and increased font sizes).

.claude/prompts/component-best-practices.md
```diff
@@ -0,0 +1,324 @@
...
+});
+```
+
+### Snapshot Testing (use sparingly)
+```typescript
+it("renders complex structure", () => {
+    const {container} = render(<ComplexComponent {...props} />);
+    expect(container).toMatchSnapshot();
+});
+```
```
mark-fitzgerald, line 289: It would be cool if we could add guidance on when and how to generate visual regression tests, as well.

.claude/prompts/component-best-practices.md
```diff
@@ -0,0 +1,324 @@
...
+// Simple conditions
+{showHint && <HintComponent />}
+
+// Multiple conditions
+{(() => {
+    if (loading) return <Spinner />;
+    if (error) return <ErrorMessage />;
+    return <Content />;
+})()}
+```
```
mark-fitzgerald, line 304: I would like to expand on this concept. I believe that we should be transitioning our hidden components instead of conditionally rendering them. In order to transition them (fade-in/fade-out, expand/collapse), we need the component rendered at all times, and CSS would determine how to show/hide them. This is also helpful for elements that control the visibility of other components (like an accordion). They need an `aria-controls` attribute that references an ID of an element that exists on the page. If that element is conditionally rendered, then it's a validation error that fails WCAG.

.claude/prompts/codebase-map.md
```diff
@@ -0,0 +1,95 @@
+# Perseus Codebase Map
+
+## Package Structure Overview
+
+```
+packages/
+├── perseus/ (v74.0.2)          # Core rendering engine
```
ivyolamit, line 7: suggestion add a section on our versioning instead of adding it here, since our version changes almost all the time

.claude/prompts/codebase-map.md
```diff
@@ -0,0 +1,95 @@
...
+│   ├── src/
+│   │   ├── widgets/            # Interactive components
+│   │   ├── components/         # Reusable UI components
+│   │   ├── renderers/          # Content renderers
+│   │   ├── util/               # Utility functions
+│   │   └── __docs__/           # Storybook documentation
```
ivyolamit, line 13: currently widget documentation is under the specific widget, not sure how we want to represent that here.

.claude/prompts/codebase-map.md
```diff
@@ -0,0 +1,95 @@
...
+│   └── __testdata__/           # Test fixtures
```
ivyolamit, line 14: we also have the `__tests__` folder that contains the test

.claude/prompts/widget-development.md
```diff
@@ -0,0 +1,286 @@
...
+
+React.useImperativeHandle(props.widgetRef, () => ({
+    focus: () => inputRef.current?.focus(),
+    blur: () => inputRef.current?.blur(),
+}));
+```
+
+## Migration Notes
+
+- Radio widget is transitioning (TODO(LEMS-2994))
```
ivyolamit, line 274: This information will be outdated soon

.claude/prompts/file-organization.md
```diff
@@ -0,0 +1,271 @@
...
+```
+widgets/widget-name/
+├── index.ts                          # Public exports only
+├── widget-name.tsx                   # Main component
+├── widget-name.test.ts               # Component tests
+├── widget-name.testdata.ts           # Test data using generators
+├── types.ts                          # Widget-specific types (if complex)
+├── utils.ts                          # Widget-specific utilities (if needed)
+└── __docs__/
+    ├── widget-name.stories.tsx       # Storybook stories
```
mark-fitzgerald, line 28: Can we add in Visual Regression stories here, as well? (see Radio widget for examples)

.claude/prompts/file-organization.md
```diff
@@ -0,0 +1,271 @@
...
+├── sub-component.tsx                 # Private sub-components
+├── utils.ts                          # Component-specific utils
+└── types.ts                          # Component-specific types
+```
+
+## File Naming Rules
+
+### TypeScript/JavaScript Files
+```typescript
+// Components - PascalCase file, named export
```
mark-fitzgerald, line 58: Shouldn't this be "kebab-case file"?

.claude/prompts/file-organization.md
```diff
@@ -0,0 +1,271 @@
...
+packages/perseus/src/widgets/interactive_graph/
+```
+
+## Widget File Structure
+
+Each widget should follow this structure:
+```
+widgets/widget-name/
+├── index.ts                          # Public exports only
+├── widget-name.tsx                   # Main component
```
mark-fitzgerald, line 22: Should we also indicate CSS files in this structure?

.claude/prompts/file-organization.md
```diff
@@ -0,0 +1,271 @@
...
+    title: "Perseus/Widgets/WidgetName",  // Hierarchical organization
+    component: WidgetName,
+};
+
+// Stories in logical order
+export const Default = {};
+export const WithHint = {};
+export const Mobile = {};
+export const ErrorState = {};
+```
```
mark-fitzgerald, line 199: Is this another place where we want to call out visual regression stories?

.claude/prompts/file-organization.md
```diff
@@ -0,0 +1,271 @@
...
+│   └── widget-name/
+│       └── assets/      # Widget-specific assets
+```
+
+### Styles
+```
+packages/perseus/src/
+├── styles/
+│   ├── global.less      # Global styles
+│   └── variables.less   # Shared variables
```
mark-fitzgerald, line 218: We don't use LESS anymore.

.claude/prompts/file-organization.md
```diff
@@ -0,0 +1,271 @@
...
+├── widgets/
+│   └── widget-name/
+│       └── widget-name.less  # Widget-specific styles
```
mark-fitzgerald, line 221: Widgets should be using CSS Modules: `widget-name.module.css`

.claude/prompts/iteration-and-feedback.md
```diff
@@ -0,0 +1,421 @@
...
+function Widget() {
+    return <input />;
+}
+
+// Iteration 2: Add user input handling (autonomous)
+function Widget({userInput, handleUserInput}) {
+    return (
+        <input
+            value={userInput.value}
+            onChange={(e) => handleUserInput({value: e.target.value})}
```
mark-fitzgerald, line 132: Should this reflect the `useCallback` optimization that was presented in the best-practices file?

.claude/prompts/iteration-and-feedback.md
```diff
@@ -0,0 +1,421 @@
...
+### Autonomous Optimizations:
+```typescript
+// Clear performance win - optimize without asking
+// Before: O(n²)
+const hasDuplicate = items.some(
+    (item, i) => items.slice(i + 1).includes(item)
+);
+
+// After: O(n)
+const hasDuplicate = items.length !== new Set(items).size;
```
mark-fitzgerald, line 153: This is a good example!

.claude/prompts/iteration-and-feedback.md
```diff
@@ -0,0 +1,421 @@
...
+```
+
+### Storybook Development & Debugging
+Use Storybook for isolated development:
+
+1. **Test different prop combinations** - Create stories for edge cases
+2. **Verify accessibility** - Use the a11y addon to catch issues
+3. **Check mobile layouts** - Use device frame addon
+4. **Debug interactions** - Use actions addon to log callbacks
+5. **Visual debugging** - Compare component states side-by-side
```
mark-fitzgerald, line 296: How does comparing "side-by-side" work here?

.claude/prompts/iteration-and-feedback.md
```diff
@@ -0,0 +1,421 @@
...
+
+# 4. Format code
+pnpm prettier . --write
+
+# 5. Build packages to ensure no build errors
+pnpm build
+
+# 6. Test in Storybook
+pnpm storybook
+```
```
mark-fitzgerald, line 340: Wasn't this called out in an earlier document?

.claude/prompts/iteration-and-feedback.md
```diff
@@ -0,0 +1,421 @@
...
+pnpm test --coverage
+```
+
+### Post-Submit Monitoring
+
+After PR is merged:
+1. Monitor for any reverted changes
+2. Check for related bug reports
+3. Watch performance metrics
+4. Respond to code review feedback
```
mark-fitzgerald, line 411: Are we expecting the AI to be doing these?

.claude/prompts/widget-development.md
```diff
@@ -0,0 +1,286 @@
+# Widget Development Guide
+
+## Three-Layer Type Architecture
+
+Perseus widgets use a three-layer type system:
```
ivyolamit, line 5: Do we need to reference this in here?
https://github.com/Khan/perseus/blob/main/__docs__/introduction.mdx

.claude/prompts/widget-development.md
```diff
@@ -0,0 +1,286 @@
...
+3. **Rubric** - Full scoring data: `ValidationData & WidgetOptions` (in validation.types.ts)
+
+## Creating a New Widget
+
+### 1. Directory Structure
+```
+packages/perseus/src/widgets/[widget-name]/
+├── index.ts                           # Exports
+├── [widget-name].tsx                  # Main component
+├── [widget-name].test.ts              # Tests
```
ivyolamit, line 18: we have our tests under `__tests__` folder

.claude/prompts/widget-development.md
```diff
@@ -0,0 +1,286 @@
...
+
+### 2. Modern Widget Implementation
+
+```typescript
+// [widget-name].tsx
+import React from "react";
+import type {PerseusWidgetNameUserInput} from "@khanacademy/perseus-core";
+
+// Props are just the widget options spread directly, plus system props
+type Props = {
```
ivyolamit, line 33: Based on previous discussion we would use an interface instead of the Props type and we would give it proper name for readability instead of just Props.

see sample here: https://github.com/Khan/perseus/blob/main/packages/perseus/src/widgets/radio/multiple-choice-component.tsx#L20

.claude/prompts/widget-development.md
```diff
@@ -0,0 +1,286 @@
...
+        </div>
+    );
+}
+
+// IMPORTANT: Set displayName for debugging
+WidgetName.displayName = "WidgetName";
+
+// Export with WidgetExports interface
+export default {
+    name: "widget-name" as const,
```
ivyolamit, line 71: not sure if this would matter but widget names are in kebab case and the widget itself will be in upper camel case

.claude/prompts/component-best-practices.md
```diff
@@ -0,0 +1,324 @@
...
+const ComponentName = (props: Props) => {
+    return <div>{props.content}</div>;
+};
+ComponentName.displayName = "ComponentName"; // Extra step
+```
+
+### Type Definitions
+```typescript
+// Place types at the top of the file
+type Props = {
```
ivyolamit, line 23: Base on discussion we would prefer interface than types see slack discussion [here](https://khanacademy.slack.com/archives/CJDRXTGQ7/p1744230806279719?thread_ts=1744223490.274449&cid=CJDRXTGQ7)

.claude/prompts/component-best-practices.md
```diff
@@ -0,0 +1,324 @@
...
+
+// For complex props, use intersection types
+type Props = BaseProps & {
+    specificProp: string;
+};
+```
+
+## Wonder Blocks Integration
+
+Perseus uses Khan Academy's Wonder Blocks design system:
```
ivyolamit, line 37: What do you think of adding a context where wonder blocks is found `https://khan.github.io/wonder-blocks` by explicitly saying this here it will only search what available components based on what are installed in our packages, but there might be more to it available in the documentation.

CLAUDE.md
```diff
@@ -30,176 +48,68 @@ pnpm prettier . --write    # Auto-format code
...
+- onChange used instead of handleUserInput
+- Local state for user answers (should use UserInputManager)
+- Unescaped backslashes in TeX strings
+
+## Migration & Deprecation
+
+### Deprecated Patterns
+- `onChange` prop - Use `handleUserInput` instead (LEMS-3245, LEMS-3542)
+- Local widget state for answers - Use UserInputManager
+- Class components for new widgets - Use functional components
```
ivyolamit, line 107: These here is a bit confusing to me. We can probably simplify these statements here, like:

More descriptive: "New widgets must use functional components, not classes"
Action-oriented: "Use functional components for new widgets"

ivyolamit (commented):
Hi @Myranae here's' my first round of review. I'll add in my final review comments 🔜 

I'm also interested to see if this will be lesser use of tokens or not 🤔

.claude/prompts/widget-development.md
ivyolamit: I think some of my comments were not posted :(
Just a thought, if we can reference existing documentation links here for context or sample it will be better. It will be less maintenance and having information outdated.

.claude/prompts/math-and-content.md
```diff
@@ -0,0 +1,270 @@
...
+Test math in different contexts:
+1. **Articles** - Both inline and display math
+2. **Exercise questions** - Main problem statement
+3. **Hints** - Step-by-step solutions
+4. **Answer choices** - In radio/multiple choice widgets
+5. **Feedback messages** - Correct/incorrect explanations
+
+### MathJax Configuration
+
+Perseus uses MathJax for rendering. Key points:
```
mark-fitzgerald, line 57: Should this be something more like "Perseus uses MathJax for rendering math"? The implied context might be missed and I wouldn't want to see MathJax used in ways that it shouldn't.

.claude/prompts/math-and-content.md
```diff
@@ -0,0 +1,270 @@
...
+                }
+            }
+        }
+    },
+    "hints": [
+        {"content": "Think about counting on your fingers."},
+        {"content": "The answer is $4$."}
+    ]
+}
+```
```
mark-fitzgerald, line 106: Should this be labelled as an example and not a template? Or, should there be a template here, instead?

.claude/prompts/math-and-content.md
```diff
@@ -0,0 +1,270 @@
...
+// ❌ Bad - Generic or missing alt text
+![Graph](graph.png)
+![](graph.png)
+```
+
+### Widget Labels
+```typescript
+// Ensure widgets have proper labels
+<TextField
+    aria-label="Enter the x-coordinate"
```
mark-fitzgerald, line 184: The preference would be to have a visible label for the text field that is associated with the input so that no `aria-label` is needed. I think this example should be removed in favor of a more expansive accessibility-focused document.

.claude/prompts/math-and-content.md
```diff
@@ -0,0 +1,270 @@
...
+
+## Performance Considerations
+
+### Math Rendering Performance
+- Initial render can be slow for many expressions
+- Use `React.memo` for components with static math
+- Consider pagination for content with 50+ expressions
+
+### Content Loading
+- Large exercises should lazy-load hints
```
mark-fitzgerald, line 197: What constitutes as "large"?

.claude/prompts/math-and-content.md
```diff
@@ -0,0 +1,270 @@
...
+- Snapshot complex math layouts
+- Test responsive behavior
+
+### Manual Testing Checklist
+- [ ] Math renders correctly
+- [ ] Widgets display properly
+- [ ] Images load and align correctly
+- [ ] Content is responsive on mobile
+- [ ] Screen reader announces math properly
+- [ ] Print view works correctly
```
mark-fitzgerald, line 226: What is this referencing? Are we supposed to consider printing when we style widgets?

.claude/prompts/testing-best-practices.md
```diff
@@ -0,0 +1,181 @@
...
+        const props = {...};
+
+        // Act
+        const result = doSomething(props);
+
+        // Assert
+        expect(result).toBe(expected);
+    });
+
+    it("handles user interaction", async () => {
```
mark-fitzgerald, line 19: Maybe a different test title? I would expect that a "user interaction" would have an "Act" section.

.claude/prompts/testing-best-practices.md
```diff
@@ -0,0 +1,181 @@
...
+render(<LegacyWidget {...props} onChange={onChange} />);
+```
+
+## Testing Priorities
+
+1. **User interactions** - How users actually use the widget
+2. **Accessibility** - Keyboard navigation, screen readers
+3. **Edge cases** - Empty states, invalid input, boundaries
+4. **Scoring logic** - Correct/incorrect answers
+5. **Mobile behavior** - Touch interactions
```
mark-fitzgerald, line 85: Maybe this is a side discussion, but I am confused as to what "Touch interactions" mean. I would expect that the standard `onClick` event would handle touches. What are we testing in a mobile environment?

.claude/prompts/testing-best-practices.md
```diff
@@ -0,0 +1,181 @@
...
+```
+
+## Test Utilities
+
+### Custom Matchers
+```typescript
+// In test files
+expect(result).toHaveBeenAnsweredCorrectly();
+expect(result).toHaveInvalidInput("message");
+```
```
mark-fitzgerald, line 126: Perhaps it's because I'm not familiar with the custom matchers, but is it possible to have the result both "answered correctly" and "have invalid input"?

.claude/prompts/testing-best-practices.md
```diff
@@ -0,0 +1,181 @@
...
+});
+```
+
+### Testing Accessibility
+```typescript
+it("has accessible labels", () => {
+    render(<Widget />);
+    expect(screen.getByLabelText("Answer")).toBeInTheDocument();
+});
+```
```
mark-fitzgerald, line 153: I recommend removing this example in favor of a more detailed accessibility-focused document.

.claude/prompts/testing-best-practices.md
```diff
@@ -0,0 +1,181 @@
...
+
+### Testing Mobile Interactions
+```typescript
+it("handles touch events", async () => {
+    const user = userEvent.setup();
+    render(<Widget />);
+
+    await user.pointer({keys: "[TouchA>]", target: element});
+    // Assert touch-specific behavior
```
mark-fitzgerald, line 162: Is this a model that we are following - touch-specific behavior? I thought we were wanting to get away from mobile-specific anything.

.claude/prompts/widget-development.md
```diff
@@ -0,0 +1,286 @@
...
+```
+packages/perseus/src/widgets/[widget-name]/
+├── index.ts                           # Exports
+├── [widget-name].tsx                  # Main component
+├── [widget-name].test.ts              # Tests
+├── [widget-name].testdata.ts          # Test data using generators
+└── __docs__/
+    ├── [widget-name].stories.tsx      # Storybook stories
+    └── a11y.mdx                       # Accessibility docs
+```
```
mark-fitzgerald, line 23: Do we need this here, again? It seems like it would be a pain to have to update this information in multiple locations whenever we make a change to directory structure.

.claude/prompts/widget-development.md
```diff
@@ -0,0 +1,286 @@
...
+type Props = {
+    // Widget-specific options (from WidgetOptions)
+    correctAnswer?: string;
+    placeholder?: string;
+    // System props
+    userInput: PerseusWidgetNameUserInput;
+    handleUserInput: (input: PerseusWidgetNameUserInput) => void;
+    trackInteraction: () => void;
+    onFocus: () => void;
+    onBlur: () => void;
```
mark-fitzgerald, line 42: Do we want to encourage outside containers telling the widget how to handle these kinds of events?

.claude/prompts/widget-development.md
```diff
@@ -0,0 +1,286 @@
...
+
+### Focus Management
+```typescript
+const inputRef = React.useRef<HTMLInputElement>(null);
+
+React.useImperativeHandle(props.widgetRef, () => ({
+    focus: () => inputRef.current?.focus(),
+    blur: () => inputRef.current?.blur(),
+}));
+```
```
mark-fitzgerald, line 270: Same concern here about focus management.

CLAUDE.md
ivyolamit, line 1: I'm just watching the session from last week, I wonder if we want to do this structure of claude.md per package, so it will only load the needed context when needed and will not overload claude context on the outset.

Image

Myranae:
@claude can you review the comments and make appropriate updates?

Myranae:
Closing in favor of https://github.com/Khan/perseus/pull/3699

PR #3285, opened 2026-02-25
Give `nodeWarnings` a real type instead of `any[]`

benchristel:
## Summary:
I told Claude to fix a TODO of its choice. It tried to get rid of an `any` type
in the linter... and bungled it.  As I was trying to fix up the result, I
discovered this lower-hanging fruit.

Issue: none

## Test plan:

Lint errors should still appear in the editor.

### DIFF
--- a/packages/perseus-linter/src/rule.ts
+++ b/packages/perseus-linter/src/rule.ts
@@ -152,7 +152,7 @@ export type LinterWarning = {
     start: number;
     end: number;
     target?: string;
-    severity?: number;
+    severity: number;
     metadata?: Record<string, any>;
 };
 
@@ -346,6 +346,7 @@ export default class Rule {
             // a rule was failing.
             return {
                 rule: "lint-rule-failure",
+                severity: this.severity,
                 message: `Exception in rule ${this.name}: ${e.message}
 Stack trace:
 ${e.stack}`,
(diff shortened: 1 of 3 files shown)


packages/perseus-linter/src/run-linter.ts
```diff
@@ -211,8 +211,8 @@ export function runLinter(
                 const content = node.content; // Text nodes have content
                 const warning = nodeWarnings[0]; // There is only one warning.
                 // These are the lint boundaries within the content
-                const start = warning.start || 0;
-                const end = warning.end || content.length;
+                const start = warning.start;
+                const end = warning.end;
```
benchristel, line 215: This is the only risky part of this change. According to the types, `warning.start` and `warning.end` can't be null or undefined. I can't prove that's really true, though.

packages/perseus-linter/src/rule.ts
```diff
@@ -346,6 +346,7 @@ export default class Rule {
             // a rule was failing.
             return {
                 rule: "lint-rule-failure",
+                severity: this.severity,
```
jeremywiebe, line 349: I wonder if this should be an extreme severity or if it's fine to be whatever severity the rule is set to.  Thoughts?

jeremywiebe (approved):

packages/perseus-linter/src/rule.ts
benchristel: I guess if it's extreme severity we're more likely to get a bug report sooner... but by using the lint rule's severity, we signal to the content creator "this is how severe the problem would be if we could report it."

I don't know what makes more sense, but I'm inclined not to overthink it.

jeremywiebe: Agreed (even though overthinking is my specialty!).  Let's leave it as you have it. :)

PR #3291, opened 2026-02-26
Make return types of functions for dealing with polar coordinates more specific.

benchristel:
## Summary:
This fixes a couple of TODO comments.

Issue: none

## Test plan:

`pnpm tsc`


benchristel:
On second thought... I'm going to close this PR. The polar functions in question are not used anywhere. Not in Perseus, not in Frontend. We should just delete them.

PR #3293, opened 2026-02-28
Remove unused default value for polygon points

benchristel:
## Summary:
This fixes a TODO comment. `coords` can't be null or undefined here.

Issue: none

## Test plan:

CI checks should pass.


catandthemachines (approved):
The code looks good, but I would like to request that when you roll this out to double check our error logs on the interactive graph for a day or two after. There were some weird edge case for polygons that could produce odd states that can break the polygon, and I can't remember if this was to prevent one. 🤔

PR #3327, opened 2026-03-11
Remove orphaned widgets from PerseusRenderers when parsing

benchristel:
Perseus renderers sometimes contain entries in the `widgets` map that
are not referenced in the `content` string. I'm not sure if this
happens due to a bug in the editors, or because people are manually
editing the JSON.

In any case, these unused widgets are confusing to humans and LLMs
alike. This PR removes them during parsing.

Issue: LEMS-3983

## Test plan:

Verify there are no regressions in the editor related to:

- cut/copy and paste of content including widgets
- undoing (with ctrl+Z) a change that deletes or adds a widget

### DIFF
--- a/.changeset/fast-guests-hope.md
+++ b/.changeset/fast-guests-hope.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-core": minor
+---
+
+Widgets that are "orphaned" (i.e. not referenced in the content string of their parent Renderer) are now removed during parsing.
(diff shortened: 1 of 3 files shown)


benchristel:
@jeremywiebe I found a potential problem with this. I think the widget snowman syntax can be replaced by CrowdIn placeholder strings for just-in-place translation.

This means that when we parse Perseus content in the translation editors, all the widgets will be removed.

We might have to make orphaned widget removal optional. We can do that by passing an option down via the `ParseContext`. That would enable us to only remove orphaned widgets during publish.

jeremywiebe (approved):

jeremywiebe:
Doh, I had two tabs open and submitted my review without realizing that it was on the one where you had this Crowdin question.  Feel free to re-request, if needed, once you settle that question. Sorry for jumping the gun here. :)

benchristel:
We can't safely remove orphaned widgets in the parsers. Translatable content has the widget snowman syntax replaced by CrowdIn placeholders for just-in-place translation. When this content is parsed (e.g. in the translation editors) all the widgets get removed.

Closing this PR. We'll have to take a different approach. For example, we could remove orphaned widgets in the content editor just before saving.

PR #3442, opened 2026-04-01
[ Interactive Graph | Vector Graph ] PR4: Scoring

SonicScrewdriver:
## Summary

_This PR was created with the help of AI, albeit with heavy oversight and review._

This is part of a series of PRs implementing the vector graph type for the Interactive Graph widget:

PR1 – type definitions and schema
PR2 – state management
PR3 – rendering & accessibility
▶️ PR4 – scoring (this PR)
PR5 – editor support

**Issue:** [LEMS-3971](https://khanacademy.atlassian.net/browse/LEMS-3971)

*   Added scoring logic for the vector graph type. The graph is now gradeable.

### Changes:

*   score-interactive-graph.ts — Adds vector scoring branch. Both tail (`coords[0]`) and tip (`coords[1]`) must match exactly via `approximateDeepEqual`. Unlike ray (which uses `collinear` for the second point), vector scoring is order-sensitive — swapping tail and tip is incorrect.
*   score-interactive-graph.test.ts — 6 new tests: undefined guess (invalid), missing coords (invalid), correct answer, wrong tail (incorrect), wrong tip (incorrect), swapped tail and tip (incorrect).

### Test plan:

*    `pnpm tsc` — no new type errors
*    `pnpm test packages/perseus-score/src/widgets/interactive-graph/` — 44 tests pass, including 6 new vector scoring tests
*    No existing scoring tests regress

[LEMS-3971]: https://khanacademy.atlassian.net/browse/LEMS-3971?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/perseus-score/src/widgets/interactive-graph/score-interactive-graph.ts
+++ b/packages/perseus-score/src/widgets/interactive-graph/score-interactive-graph.ts
@@ -431,18 +431,30 @@ function scoreInteractiveGraph(
             userInput.coords != null &&
             rubric.correct.coords != null
         ) {
-            // Vector scoring: both tail and tip must match exactly.
-            // Order matters — coords[0] is tail, coords[1] is tip.
-            if (
-                approximateDeepEqual(
-                    userInput.coords[0],
-                    rubric.correct.coords[0],
-                ) &&
-                approximateDeepEqual(
-                    userInput.coords[1],
-                    rubric.correct.coords[1],
-                )
-            ) {
+            const guess = userInput.coords;
+            const correct = rubric.correct.coords;
+
+            let match: boolean;
+            if (rubric.correct.match === "congruent") {
+                // Congruent: same direction and magnitude, any position.
+                // Compare the component form ⟨dx, dy⟩ of each vector.
+                const guessDelta: Coord = [
+                    guess[1][0] - guess[0][0],
+                    guess[1][1] - guess[0][1],
+                ];
+                const correctDelta: Coord = [
+                    correct[1][0] - correct[0][0],
+                    correct[1][1] - correct[0][1],
+                ];
+                match = approximateDeepEqual(guessDelta, correctDelta);
+            } else {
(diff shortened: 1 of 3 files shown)


packages/perseus-score/src/widgets/interactive-graph/score-interactive-graph.ts
```diff
@@ -417,6 +417,43 @@ function scoreInteractiveGraph(
                     collinear(correct[1], correct[2], guess[2]);
             }
 
+            if (match) {
```
ivyolamit, line 420: is this logic intentional?

packages/perseus-score/src/widgets/interactive-graph/score-interactive-graph.ts
```diff
@@ -417,6 +417,43 @@ function scoreInteractiveGraph(
...
+                    earned: 1,
+                    total: 1,
+                    message: null,
+                };
+            }
+        } else if (
+            userInput.type === "vector" &&
+            rubric.correct.type === "vector" &&
+            userInput.coords != null &&
+            rubric.correct.coords != null
```
ivyolamit, line 432: this is actually good point, do we want this also applied to other new graphs?

ivyolamit (approved):
@SonicScrewdriver overall looking good 👍🏼 I just have 1 non-blocking question.

packages/perseus-score/src/widgets/interactive-graph/score-interactive-graph.ts
SonicScrewdriver: Yeah! Github (or possibly Claude) is just weird about how it updates the code lines.

You'll notice that the same ` if (match) {` logic continues below the Vector graph. Since this end logic is the same for every graph, sometimes it seems to think we've updated the previous graph.

PR #3443, opened 2026-04-01
[ Interactive Graph | Vector Graph ] PR5: Editor Support

SonicScrewdriver:
## Summary

_This PR was created with the help of AI, albeit with heavy oversight and review._

This is part of a series of PRs implementing the vector graph type for the Interactive Graph widget:

PR1 – type definitions and schema
PR2 – state management
PR3 – rendering & accessibility
PR4 – scoring
▶️ PR5 – editor support (this PR)

**Issue:** [LEMS-3971](https://khanacademy.atlassian.net/browse/LEMS-3971)

*   Surfaces the vector graph type in the content-creator UI. Once landed and the feature flag is enabled, content creators can author vector graph exercises.

### Changes:

*   graph-type-selector.tsx — Adds "Vector" option to the graph type dropdown, gated behind `interactive-graph-vector` feature flag.
*   start-coords/types.ts — Adds `{type: "vector"}` to `GraphTypesThatHaveStartCoords` union.
*   start-coords/util.ts — Imports `getVectorCoords` and adds `case "vector"` in `getDefaultGraphStartCoords`.
*   start-coords/start-coords-settings.tsx — Adds `case "vector"` dispatching to `<StartCoordsLine>` (reuses the existing component since vector has the same `CollinearTuple` shape as ray/linear).
*   interactive-graph-question-builder.ts — Adds `withVector()` builder method and `VectorGraphConfig` class to the editor's test data builder.
*   __testdata__/interactive-graph.testdata.ts — Adds `vectorMinimalQuestion` test fixture for the editor story.
*   interactive-graph-editor.stories.tsx — Adds `InteractiveGraphVector` editor Storybook story.

### Design decisions:

*   **No editor validation for overlapping start coords** — Consistent with ray, linear, and segment which also don't validate for overlapping points in the editor. The reducer already rejects moves that would cause overlap during interaction.
*   **Reuses `StartCoordsLine`** — Vector's `CollinearTuple` start coords shape matches ray/linear, so no custom start coords component is needed (unlike exponential which has an asymptote field).

### Test plan:

*    `pnpm tsc` — no new type errors
*    `pnpm lint` — no lint errors
*    Verified in Storybook: editor correctly shows vector graph type when feature flag is enabled, start coords UI works
*    No existing tests regress

[LEMS-3971]: https://khanacademy.atlassian.net/browse/LEMS-3971?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/perseus-editor/src/widgets/__docs__/interactive-graph-editor.stories.tsx
+++ b/packages/perseus-editor/src/widgets/__docs__/interactive-graph-editor.stories.tsx
@@ -27,6 +27,7 @@ import {
     segmentWithLockedFigures,
     segmentWithStartingCoordsQuestion,
     exponentialMinimalQuestion,
+    vectorMinimalQuestion,
     sinusoidMinimalQuestion,
     sinusoidWithStartingCoordsAndPiTicksQuestion,
     unlimitedPolygonWithCorrectAnswerQuestion,
@@ -119,6 +120,10 @@ export const InteractiveGraphExponential = (): React.ReactElement => {
     return (
         <EditorPageWithStorybookPreview question={exponentialMinimalQuestion} />
     );
+};
+
+export const InteractiveGraphVector = (): React.ReactElement => {
+    return <EditorPageWithStorybookPreview question={vectorMinimalQuestion} />;
 };
 
 export const InteractiveGraphSinusoid = (): React.ReactElement => {
--- a/packages/perseus-editor/src/widgets/interactive-graph-editor/components/graph-type-selector.tsx
+++ b/packages/perseus-editor/src/widgets/interactive-graph-editor/components/graph-type-selector.tsx
@@ -68,9 +68,7 @@ const GraphTypeSelector = (props: GraphTypeSelectorProps) => {
             <OptionItem value="polygon" label="Polygon" />
             <OptionItem value="segment" label="Line Segment(s)" />
             <OptionItem value="ray" label="Ray" />
-            {showVector && (
-                <OptionItem value="vector" label="Vector" />
-            )}
+            {showVector && <OptionItem value="vector" label="Vector" />}
             <OptionItem value="angle" label="Angle" />
         </SingleSelect>
     );
--- a/packages/perseus-editor/src/widgets/interactive-graph-editor/components/graph-type-selector.tsx
+++ b/packages/perseus-editor/src/widgets/interactive-graph-editor/components/graph-type-selector.tsx
@@ -34,6 +34,11 @@ const GraphTypeSelector = (props: GraphTypeSelectorProps) => {
         "interactive-graph-logarithm",
     );
 
+    const showVector = isFeatureOn(
+        {apiOptions: props.apiOptions},
+        "interactive-graph-vector",
+    );
+
     return (
         <SingleSelect
             selectedValue={props.graphType}
@@ -63,6 +68,9 @@ const GraphTypeSelector = (props: GraphTypeSelectorProps) => {
             <OptionItem value="polygon" label="Polygon" />
             <OptionItem value="segment" label="Line Segment(s)" />
             <OptionItem value="ray" label="Ray" />
+            {showVector && (
+                <OptionItem value="vector" label="Vector" />
+            )}
             <OptionItem value="angle" label="Angle" />
         </SingleSelect>
     );
--- a/packages/perseus-editor/src/widgets/interactive-graph-editor/components/vector-answer-options.tsx
+++ b/packages/perseus-editor/src/widgets/interactive-graph-editor/components/vector-answer-options.tsx
@@ -1,0 +1,53 @@
+import {components} from "@khanacademy/perseus";
+import {OptionItem, SingleSelect} from "@khanacademy/wonder-blocks-dropdown";
+import * as React from "react";
+import invariant from "tiny-invariant";
+
+import styles from "../interactive-graph-editor.module.css";
+import LabeledRow from "../locked-figures/labeled-row";
+
+import type {Props as EditorProps} from "../interactive-graph-editor";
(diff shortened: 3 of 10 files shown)


packages/perseus-editor/src/widgets/__docs__/interactive-graph-editor.stories.tsx
ivyolamit, line 1: ooopppss I think not all new graph types has editor stories 🙈
let's create a follow-up tasks for it

packages/perseus-editor/src/widgets/interactive-graph-editor/components/graph-type-selector.tsx
```diff
@@ -63,6 +68,7 @@ const GraphTypeSelector = (props: GraphTypeSelectorProps) => {
             <OptionItem value="polygon" label="Polygon" />
             <OptionItem value="segment" label="Line Segment(s)" />
             <OptionItem value="ray" label="Ray" />
+            {showVector && <OptionItem value="vector" label="Vector" />}
             <OptionItem value="angle" label="Angle" />
```
ivyolamit, line 72: Not sure if there was a reason adding the Angle at the last, but can we rearrange this, looks weird Angle is at the bottom 😓 

Image

packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
```diff
@@ -457,6 +458,12 @@ class InteractiveGraphEditor extends React.Component<Props> {
                                 onChange={this.props.onChange}
                             />
                         )}
+                        {this.props.correct?.type === "vector" && (
+                            <VectorAnswerOptions
+                                correct={this.props.correct}
+                                onChange={this.props.onChange}
+                            />
+                        )}
```
ivyolamit, line 466: Just an observation, probably it's the intended behavior of Correct Answer section the coordinates will not show up at initial render. I noticed this in Absolute Value. But it would be nice to see the initial coordinates like other graphs. (See `(5.000, 5.000)` on the right.)

| Initial | After Moving |
| ----- | ------------- |
| Image | Image |

packages/perseus-editor/src/widgets/interactive-graph-editor/components/vector-answer-options.tsx
```diff
@@ -0,0 +1,53 @@
...
+const {InfoTip} = components;
+
+interface Props {
+    correct: PerseusGraphTypeVector;
+    onChange: (props: Partial<EditorProps>) => void;
+}
+
+export default function VectorAnswerOptions({correct, onChange}: Props) {
+    return (
+        <LabeledRow label="Student answer must">
```
ivyolamit, line 21: This is something I noticed when doing the [POC](https://github.com/Khan/perseus/pull/3466#discussion_r3082935022) yesterday in Matthew's interactive but not scored PR, the font of a label beside a dropdown is not consistent with the rest of the editor. I'm not sure if it was intentional, but in the POC i did made it consistent. I'll leave that up to you if you'll update that here, it's a trivial thing and not a blocker.

packages/perseus-editor/src/widgets/interactive-graph-editor/components/vector-answer-options.tsx
```diff
@@ -0,0 +1,53 @@
...
+                    });
+                }}
+                // Never uses placeholder, always has value
+                placeholder=""
+                className={styles.singleSelectShort}
+            >
+                <OptionItem value="exact" label="match exactly" />
+                <OptionItem value="congruent" label="be congruent" />
+            </SingleSelect>
+            <InfoTip>
```
ivyolamit, line 43: The font in the info tip here looks smaller and not consistent styling with the screen reader tree.

| Congruent | Screen reader tree |
| ----------- | ------------------ |
| Image | Image |

I can't pin point where it is defined in screen reader tree code, it looks the same how you implemented here `<InfoTip>` 🤔

ivyolamit (approved):
@SonicScrewdriver overall PR looks good, i just have few non-blocking comments. Great work here! Editor looks good 👌🏼

packages/perseus-editor/src/widgets/interactive-graph-editor/components/graph-type-selector.tsx
SonicScrewdriver: I agree! I wonder if we should have this be a separate ticket to discuss with content. My inclination is just to make everything alphabetical, but perhaps they would prefer a different order according to the graph types frequency of use?

packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
SonicScrewdriver: Oh! Yeah I fully agree, I'll update this so that it's always visible.

packages/perseus-editor/src/widgets/interactive-graph-editor/components/vector-answer-options.tsx
SonicScrewdriver: I'm less inclined to worry about this if it's a deep dive, but I'll give it a quick look to see if I can notice anything different about my implementation! :)

packages/perseus-editor/src/widgets/interactive-graph-editor/components/vector-answer-options.tsx
SonicScrewdriver: Oooo I'll look into this to make sure I've implemented it how we've set up the feature previously. I was pretty sure I copied this logic exactly from the Polygon Graph, but I could have missed something!

PR #3463, opened 2026-04-08
[Interactive Graph] Planning document: interactive, unscored IG

handeyeco:
## Summary:

I took the research from the ticket and used Claude to make a planning document for upcoming work.

Issue: LEMS-3958


packages/perseus/src/widgets/interactive-graphs/__docs__/notes/not-graded.md
```diff
@@ -0,0 +1,142 @@
...
+
+In `packages/perseus-editor/src/components/widget-editor.tsx`:
+
+- Add a `_setGraded` handler method, mirroring `_setStatic`.
+- In `render()`, call `Widgets.supportsGradedToggle(widgetInfo.type)` (analogous to the `supportsStaticMode` check on line 155).
+- Render a `LabeledSwitch` labeled "Graded" when that returns true. The switch should default to `checked={widgetInfo.graded !== false}` (treats `undefined` and `true` as graded).
+
+---
+
+### 6. Add "Not graded" indicator in `WidgetContainer`
```
handeyeco: I think this might be the wrong approach since we have existing content where `graded: false` and things like Explanation and Image don't need this kind of indication. I'll probably update this so that:

- [Same] Add a toggle to the editor
- [Same] Pass the flag to the widget
- [Different] Don't add a visual update to WidgetContainer (because that would affect existing content)
- [Different] IG updates the UI to indicate that it's not going to be scored

handeyeco: Updated

packages/perseus/src/widgets/interactive-graphs/__docs__/notes/not-graded.md
```diff
@@ -0,0 +1,141 @@
...
+### Why not add a flag to `PerseusInteractiveGraphWidgetOptions`?
+
+Putting `graded` on the per-widget options type would require every consumer to handle it individually. `WidgetOptions.graded` is already the right home — it's shared across all widget types — so we're building on that.
+
+---
+
+## Objectives
+
+1. **Editor toggle** — Content creators get a "Graded" switch in the IG widget editor, defaulting to on (graded). The switch is only shown for widgets that opt in.
+2. **Hide answer fields when ungraded** — When `graded` is off, the answer-configuration sections in the IG editor are hidden. The stored answer data is preserved so toggling back on restores it.
```
ivyolamit, line 22: Would this include the Start coordinates?  i imagine this is only referring to the "Correct answer" section. I wonder if we would still allow Locked figures here?

packages/perseus/src/widgets/interactive-graphs/__docs__/notes/not-graded.md
```diff
@@ -0,0 +1,141 @@
...
+
+This ensures a `graded: false` IG is never replaced by the correct answer, even if it's in static mode. The widget stays in whatever state the learner left it.
+
+`interactive-graph.tsx` will need to receive `graded` from `UniversalWidgetProps` (step 2) and forward it to `StatefulMafsGraph`.
+
+---
+
+### 9. Tests and stories
+
+Each change above needs coverage:
```
ivyolamit, line 118: we would also need the regression story for this especially it will look different from normal IG

ivyolamit (approved):
@handeyeco overall looking good, i just have 2 non-blocking comments.

PR #3466, opened 2026-04-08
[InteractiveGraph] Interactive but unscored (implementation)

handeyeco:
## Summary:
InteractiveGraph: interactive but unscored. The goal is to give learners the ability to visualize _math things_ while answering other parts of an exercise (like a NumericInput or a Radio). By leveraging the `graded` field, we're able to take the IG out of the normal flow of scoring, but still leave it interactive (unlike `static`).

The main bits of this were:

- Update the editor so that we could switch `graded`. Even though `graded` existed for a long time, it didn't seem to be used for anything except _maybe_ IFrame; so we're using it for IG now.
- Display a message to learners that the IG is not part of scoring.
- Add tests/stories

For the most part `graded` was already kind of wired up: it existed in the data schema and the scoring function ignored widgets where `graded === false`. So this PR is more about adding some UI elements.

Issue: LEMS-3970

## Test plan:

Editor experience:

https://github.com/user-attachments/assets/262ea357-2977-4572-94a9-0462954aa0f7

Learner experience:

https://github.com/user-attachments/assets/b93cff92-69d0-40f4-8526-897f7037982e

In the ZND:

https://github.com/user-attachments/assets/ff2fb8c3-50d4-40dd-917c-75c3d269e3d0


packages/perseus-core/src/data-schema.ts
```diff
@@ -376,8 +376,14 @@ export type WidgetOptions<
      */
     static?: boolean;
     /**
-     * Whether a widget is scored. Usually true except for IFrame widgets (deprecated).
+     * Whether a widget is scored.
      * Default: true
+     *
+     * When false, the widget acts as an unscored "sketchpad" that learners can
```
handeyeco: This comment is too specific to the IG use case.

packages/perseus/src/mixins/widget-prop-denylist.ts
```diff
@@ -42,6 +42,8 @@ const denylist = [
     "showSolutions",
     "reviewMode",
     "widgetIndex",
+    // added by src/renderer.jsx (widget-level metadata, not widget state)
+    "graded",
```
handeyeco: I don't think we want to add this to the deny list because I think that would prevent it from being serialized?

packages/perseus/src/widgets/interactive-graphs/interactive-graph.test.tsx
```diff
@@ -1933,3 +1933,42 @@ describe("Interactive Graph", function () {
...
+            testDependencies,
+        );
+    });
+
+    it("renders a 'not graded' message when graded is false", () => {
+        // Arrange, Act
+        const question = interactiveGraphQuestionBuilder()
+            .withSegments({numSegments: 1})
+            .build();
+        question.widgets["interactive-graph 1"].graded = false;
```
handeyeco: Seems like maybe this should be a part of the object builder.

packages/perseus/src/widgets/interactive-graphs/interactive-graph.tsx
```diff
@@ -301,16 +301,24 @@ class InteractiveGraph extends React.Component<Props, State> {
...
-                snapStep={snapStep}
-                box={box}
-                showTooltips={!!this.props.showTooltips}
-                readOnly={this.props.apiOptions?.readOnly}
-                widgetId={this.props.widgetId}
-            />
+            <>
+                {this.props.graded === false && (
+                    <p>
+                        This graph is for your use only and will not be graded.
```
handeyeco: This text needs to be translated text. Also it needs to be different.

packages/perseus/src/widgets/interactive-graphs/interactive-graph.tsx
```diff
@@ -938,4 +946,5 @@ export default {
     getStartUserInput,
     getCorrectUserInput,
     getUserInputFromSerializedState,
+    supportsGradedToggle: true,
```
handeyeco: I wonder if this should be a part of `perseus-core` data vs `perseus` data.

packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.test.tsx
```diff
@@ -142,4 +142,46 @@ describe("StatefulMafsGraph", () => {
...
+        // is an angle (3 movable points). When graded, static mode should show
+        // the correct answer.
+        const correctAngle = {type: "angle" as const};
+
+        // Act
+        render(
+            <StatefulMafsGraph
+                {...getBaseStatefulMafsGraphProps()}
+                graph={{type: "segment"}}
+                static={true}
```
handeyeco: These tests shouldn't be using `static: true`. Also "static: true" and "graded: false" are kind of mutually exclusive for IG 🤔

packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.test.tsx
```diff
@@ -142,4 +142,46 @@ describe("StatefulMafsGraph", () => {
...
+                graded={true}
+                correct={correctAngle}
+            />,
+        );
+
+        // Assert: angle has 3 movable points (vertex + 2 endpoints)
+        expect(screen.getAllByTestId("movable-point").length).toBe(3);
```
handeyeco: This test is pointless.

packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.test.tsx
```diff
@@ -142,4 +142,46 @@ describe("StatefulMafsGraph", () => {
...
+        // is an angle (3 movable points). When ungraded, static mode should NOT
+        // replace the learner's state with the correct answer.
+        const correctAngle = {type: "angle" as const};
+
+        // Act
+        render(
+            <StatefulMafsGraph
+                {...getBaseStatefulMafsGraphProps()}
+                graph={{type: "segment"}}
+                static={true}
```
handeyeco: Same here

packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.test.tsx
```diff
@@ -142,4 +142,46 @@ describe("StatefulMafsGraph", () => {
...
+                graded={false}
+                correct={correctAngle}
+            />,
+        );
+
+        // Assert: segment has 2 movable points (the learner's state is preserved)
+        expect(screen.getAllByTestId("movable-point").length).toBe(2);
```
handeyeco: Same here

packages/perseus/src/mixins/widget-prop-denylist.ts
handeyeco: Confirmed: this needs to serialized or the feature doesn't work. Claude was trying to skirt around failing tests. This shows that the feature isn't properly tested because tests are passing while the feature is broken.

packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
```diff
@@ -410,54 +416,58 @@ class InteractiveGraphEditor extends React.Component<Props> {
...
+                                {this.props.correct?.type === "segment" && (
+                                    <SegmentCountSelector
+                                        correct={this.props.correct}
+                                        graph={this.props.graph}
+                                        onChange={this.props.onChange}
+                                    />
+                                )}
+
+                                {this.props.graph?.type &&
+                                    shouldShowStartCoordsUI(
```
handeyeco: I don't think we want this to be hidden when `graded: false`, start position still seems valid to me.

packages/perseus-editor/src/components/widget-editor.tsx
```diff
@@ -106,6 +106,14 @@ class WidgetEditor extends React.Component<
         this.props.onChange(newWidgetInfo);
     };
 
+    _setGraded = (value: boolean) => {
```
ivyolamit, line 111: Would it be too much if we convert this to function component?

packages/perseus-editor/src/components/widget-editor-settings.tsx
ivyolamit, line 1: @handeyeco added this component for the editor settings. This now contains all the toggle and drop down from the widget header to improve the UX. See slack [discussion](https://khanacademy.slack.com/archives/C0AHH9H21H7/p1776204333299159) for details.

Image

packages/perseus-editor/src/components/widget-editor.tsx
handeyeco: It would be good if we could, but I don't think it should be a part of this PR. It would make it difficult to see what changes happened as part of this specific chunk of work.

packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.test.tsx
handeyeco: Cleaned up pointless tests, but I think we should still:

- handle the static/ungraded mutual exclusivity
- add new tests

EDIT: done

packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
handeyeco: In fact we can't really hide "Correct answer" without a larger refactor. See: https://khanacademy.atlassian.net/browse/LEMS-3958?focusedCommentId=474872

handeyeco:
@claude review once

packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
```diff
@@ -156,6 +162,10 @@ export type Props = {
  */
 class InteractiveGraphEditor extends React.Component<Props> {
     static widgetName = "interactive-graph";
+    static bestPractices = {
+        url: "https://www.khanacademy.org/internal-courses/content-creation-best-practices/xe46daa512cd9c644:question-writing/xe46daa512cd9c644:multiple-choice/a/stems",
+        label: "Interactive Graph best practices",
+    };
```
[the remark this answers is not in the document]
handeyeco: @ivyolamit

handeyeco:
I made a dummy Frontend PR to spin up a ZND: https://github.com/Khan/frontend/pull/10557

packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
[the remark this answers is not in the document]
ivyolamit: Currently the interactive graph best practices doesn't exist yet like the radio widget. But for now we can use this link for our widget fundamentals: `https://khanacademy.atlassian.net/wiki/spaces/LC/pages/3295281289/Interactive+Graph+Widget#Adding-a-new-Interactive-Graph-Widget`

Do you mind adding also a TODO comment to update the link once content team created it, see chat [here](https://khanacademy.slack.com/archives/C6NS8DWJU/p1776200887990069)

packages/perseus-editor/src/components/alignment-select.tsx
```diff
@@ -42,17 +86,12 @@ export function AlignmentSelect({
...
-                ))}
-            </select>
-        </>
+        </View>
     );
 }
+
+const styles = StyleSheet.create({
+    singleSelectShort: {
+        height: 26,
```
ivyolamit:
```suggestion
        height: sizing.size_260,
```

packages/perseus-editor/src/components/alignment-select.tsx
```diff
@@ -13,24 +20,61 @@ interface Props {
...
+                    flexDirection: "row",
+                    alignItems: "center",
+                },
+                style,
+            ]}
+        >
+            <BodyText id={labelId} tag="span">
+                Alignment
+            </BodyText>
+            <Strut size={spacing.xSmall_8} />
```
ivyolamit: `spacing` is [deprecated](https://khan.github.io/wonder-blocks/?path=/docs/packages-tokens-deprecated-spacing-deprecated--docs) please use sizing instead
see: https://khan.github.io/wonder-blocks/?path=/docs/packages-tokens-sizing--docs#sizing-2

packages/perseus-editor/src/components/alignment-select.tsx
```diff
@@ -13,24 +20,61 @@ interface Props {
...
+            >
+                {supportedAlignments.map((alignment) => (
+                    <OptionItem
+                        key={alignment}
+                        value={alignment}
+                        label={alignment}
+                    />
+                ))}
+            </SingleSelect>
+            <Strut size={spacing.xxSmall_6} />
```
ivyolamit: See comment above about spacing is deprecated

packages/perseus-editor/src/components/widget-editor-settings.tsx
```diff
@@ -0,0 +1,105 @@
...
+        supportedAlignments.length > 1;
+
+    if (!bestPractices && !hasControls) {
+        return null;
+    }
+
+    return (
+        <View
+            style={{
+                paddingInline: spacing.small_12,
```
ivyolamit: Please see spacing comment above.

ivyolamit: Ideally, the styles will be in a .css file so we can use `classNames` here. This is something Mark has been working in our code base since last year. But in the editor to my knowledge it's not fully migrated yet so we would need the !important tag.
I would recommend moving the styles in its css file, including other styles in this PR.

packages/perseus-editor/src/components/widget-editor-settings.tsx
```diff
@@ -0,0 +1,105 @@
...
+import LabeledSwitch from "./labeled-switch";
+
+import type {Alignment, PerseusWidget} from "@khanacademy/perseus-core";
+
+type BestPractices = {
+    url: string;
+    label: string;
+};
+
+type WidgetEditorSettingsProps = {
```
ivyolamit: I think this qualify as `interface`
something the team has discussed before on using type vs interface. And we would be in favor with using interface.

packages/perseus-editor/src/components/widget-editor.tsx
```diff
@@ -102,6 +98,20 @@ class WidgetEditor extends React.Component<
         const newWidgetInfo = {
             ...this.state.widgetInfo,
             static: value,
+            // if it's "interactive but ungraded" (ungraded)
+            // we don't also want it to be "non-interactive" (static)
```
ivyolamit, line 102: Can you please expand this a little, this would be helpful context in the future. 🙏🏼 
And does content team knows about this already?

packages/perseus-editor/src/components/widget-editor.tsx
ivyolamit: @handeyeco that's fair, do you mind adding a ticket for this 🙏🏼 might be a good cleanup work at the end of the project.

packages/perseus-editor/src/styles/perseus-editor.css
```diff
@@ -138,8 +136,7 @@ code {
 }
 .perseus-widget-editor .perseus-widget-editor-content {
     border-radius: 0 0 3px 3px;
-    border-top: 1px solid #ddd;
-    padding: 10px;
+    padding: 0 10px 10px;
```
ivyolamit:
```suggestion
    padding: 0 var(--wb-sizing-size_100) var(--wb-sizing-size_100);
```

packages/perseus-score/src/score.test.ts
```diff
@@ -407,4 +408,64 @@ describe("scorePerseusItem", () => {
...
+            },
+            "dropdown 2": {
+                value: 2, // correct
+            },
+        };
+
+        const score = scorePerseusItem(item, userInputMap, "en");
+
+        expect(score).toHaveBeenAnsweredCorrectly({
+            shouldHavePoints: true,
```
ivyolamit, line 468: Initially I got confused on this expectation when the scenario is `ignores scoring when graded is false`
but upon rereading the input it has 2 dropdown.

Suggestion only have one item for ungraded this is to simplify the scenario.
```
            widgets: {
                "dropdown 1": ungraded,
            },
```

packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.tsx
```diff
@@ -133,9 +134,11 @@ export const StatefulMafsGraph = React.forwardRef<
         allowReflexAngles,
     ]);
 
-    // If the graph is static, it always displays the correct answer. This is
-    // standard behavior for Perseus widgets (e.g. compare the Radio widget).
-    if (props.static && props.correct) {
+    // If the graph is static and graded, it always displays the correct answer.
+    // This is standard behavior for Perseus widgets (e.g. compare the Radio
+    // widget). When graded is false the widget is a sketchpad and should never
+    // reveal the correct answer, even in static/review mode.
```
ivyolamit: Looks like this is conflicting comment to previous one above 🤔 
>             // if it's "interactive but ungraded" (ungraded)
>             // we don't also want it to be "non-interactive" (static)

packages/perseus/src/strings.ts
```diff
@@ -799,6 +800,8 @@ export const strings = {
     addPoint: "Add Point",
     removePoint: "Remove Point",
     graphKeyboardPrompt: "Press Shift + Enter to interact with the graph",
+    ungradedInteractiveGraph:
+        "Use this graph to check your thinking, but it does not count as your answer.",
```
ivyolamit, line 804: Please don't forget to file the translations when deploying the frontend, it should add a comment in the PR on how to file it in a jira ticket.

ivyolamit (changes requested):
@handeyeco I added my review comments for the most part of the change request is changing the usage of `spacing` which is deprecated to `sizing` as documented [here in WB](https://khan.github.io/wonder-blocks/?path=/docs/packages-tokens-sizing--docs#sizing-2)

Also a note let's make sure that content team is aware of the expected behavior and this is align with what they're thinking. And lastly, when this PR will be deployed in frontend please double check that the failed test (if any) is not related to the changes, especially this PR is added the new props.

Please let me know if you have questions about my comments. 🙏🏼

packages/perseus-editor/src/components/alignment-select.tsx
handeyeco: See: https://github.com/Khan/wonder-blocks/pull/3029

IMO this shouldn't be a blocker, it's copying established patterns in other files.

packages/perseus-editor/src/components/alignment-select.tsx
handeyeco: See: https://github.com/Khan/wonder-blocks/pull/3029

packages/perseus-editor/src/components/widget-editor-settings.tsx
handeyeco: It says it was added by you, do you mind submitting a fix?

packages/perseus-editor/src/components/widget-editor-settings.tsx
handeyeco: It says it was added by you, do you mind submitting a fix?

packages/perseus-editor/src/components/alignment-select.tsx
jandrade: I would recommend to go with `gap: sizing.size_080` (or `size_060`) in the `View.style` prop defined in line 38.

packages/perseus-editor/src/components/widget-editor.tsx
handeyeco: It's built into the UI so that it's not possible to do both `static` and `ungraded`; that's what this code is doing. But we can explain it more in the playtest/release.

packages/perseus-editor/src/components/alignment-select.tsx
jandrade: re: established patterns

I've been doing a lot of migrations in both `perseus` and `frontend` to update our UI code to switch to newer patterns so we can use more modern CSS features instead of relying on previous patterns that were useful before. I'll take a note on migrating us from `Strut/Spring` to use CSS properties directly.

packages/perseus-editor/src/styles/perseus-editor.css
handeyeco: It says it was added by you, do you mind submitting a fix?

packages/perseus-score/src/score.test.ts
handeyeco: It was intentional to use two to prove that regular scoring works as expected even when there's a non-graded widget that's correct/incorrect. It also matches expected use, which is to combine non-graded with graded widgets.

packages/perseus-editor/src/components/widget-editor-settings.tsx
jandrade: From the WB perspective, it is fine to keep the styles in an object for now. I'm about to start the CSS modules migration in Wonder Blocks, and one of my main goals is to continue supporting the `style` prop (instead of using `className`) so the migration can be done incrementally and it will be backwards compatible with our existing styling (cc @mark-fitzgerald for you to be aware of this).

packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.tsx
handeyeco: Good catch

packages/perseus-editor/src/components/widget-editor.tsx
handeyeco: I added an additional comment

packages/perseus-editor/src/components/alignment-select.tsx
handeyeco: @ivyolamit maybe it would be best to replace all uses of Strut in Perseus in a single (separate) PR.

jandrade: If moving to `gap` is not possible at this point, I'd recommend at least changing any of the `Strut` uses to have the same spacing size, that way there won't be flagged as a visual regression when this is converted to `gap` (which makes a bit harder to review in Chromatic).

handeyeco: I just changed it to use `gap` (removing `Strut`)

ivyolamit:
Latest commit uses the sizing and interface in `packages/perseus-editor/src/components/widget-editor-settings.tsx` 

image

ivyolamit (approved):
👍🏼

handeyeco:
TODO: Ivy and I decided to add the toggle switch in the editor behind a feature flag.

PR #3479, opened 2026-04-10
Handle download.cypress.io domain, switch to ephemeral runners.

jeresig:
## Summary:
Looks like we sometimes download Cypress from download.cypress.io during the pnpm install process, so we need to handle that. Also we're still seeing flakiness on the github hosted runners so I want try switching over to our self-hsoted runners instead. I was able to use the self-hosted runners for all workflows except for the publish-related ones as those have to be run on the Github hosted runners in order for the trusted publishing  to work.

Issue: FEI-XXXX

## Test plan:


.github/workflows/node-ci.yml
```diff
@@ -271,28 +268,31 @@ jobs:
...
     check_schema:
         name: Check for schema changes
-        runs-on: ${{ matrix.os }}
+        runs-on: ${{ vars.USE_GITHUB_RUNNERS == 'true' && '${{ matrix.os }}' || 'ephemeral-runner' }}
         strategy:
             matrix:
                 os: [ubuntu-latest]
                 node-version: [24.x]
         steps:
+            - name: Install uuidgen
```
jeresig, line 283: uuidgen isn't included in our self-hosted runners so we have to explicitly install it.

.github/npm-extra-domains.conf
```diff
@@ -0,0 +1,6 @@
+# npm package registry for pnpm install
```
jeresig, line 1: We now have 3 domains that we need for every pnpm install so I split it out into a separate config file to simplify things.

.github/workflows/chromatic.yml
```diff
@@ -2,8 +2,8 @@ name: Publish Storybook to Chromatic
 
 # See: https://www.chromatic.com/docs/setup-turbosnap/#github-pull_request-triggers
 on:
-    workflow_dispatch:
-    # Other Khan Academy repositories use this trigger for the chromaui/action
+    workflow_dispatch: # Other Khan Academy repositories use this trigger for the chromaui/action
```
jeresig: This is coming from oxfmt running against the workflow files (for better or worse). It's part of the validate-workflows workflow to ensure consistency in our workflow formatting.

.claude/settings.json
```diff
@@ -76,7 +76,9 @@
                 "github.com",
                 "api.github.com",
                 "registry.npmjs.org",
-                "npm.pkg.github.com"
+                "npm.pkg.github.com",
+                "download.cypress.io",
+                "cdn.cypress.io"
```
jeremywiebe, line 81: Yep.  We cache the Cypress binaries, but I think in cases where the cache is invalidated, we need to get the binaries directly from Cypress. Thanks for figuring this out!

.github/workflows/chromatic.yml
jeremywiebe: Could we invert the order of the `on` items here so that the formatter doesn't break this comment up like this?

.github/workflows/gh-pages.yml
```diff
@@ -7,11 +7,11 @@ on:
     # Allows us to run this workflow manually from the Actions tab
     workflow_dispatch:
 
-# This workflow will deploy the Storybook static files to GitHub Pages.
+        # This workflow will deploy the Storybook static files to GitHub Pages.
```
jeremywiebe, line 10: Oxfmt makes some weird formatting choices!

.github/workflows/validate-workflows.yml
```diff
@@ -31,10 +31,10 @@ jobs:
               uses: Khan/actions@secure-network-v1
               timeout-minutes: 5
               with:
-                  extra-domains: registry.npmjs.org
+                  conf-files: .github/npm-extra-domains.conf
 
             - name: Fix checkout-followed-by-setup
-              uses: Khan/actions@fix-workflows-v2
+              uses: Khan/actions@fix-workflows-v3
```
jeremywiebe, line 37: Not needed for as part of this PR, but should we be pinning our own "Khan/actions" actions to a commit SHA also?

.github/npm-extra-domains.conf
```diff
@@ -0,0 +1,6 @@
...
+registry.npmjs.org
+# Cypress binary download during pnpm install
+download.cypress.io
+# Cypress binary CDN for pnpm install
```
jeremywiebe: **nit**: Not sure it matters, but this is not really during `pnpm install` is it? 

FWIW: I think it's when we run [`pnpm exec cypress install`](https://github.com/Khan/perseus/blob/4454fc25104e42c0938e01901143582f73098e38/.github/workflows/node-ci.yml#L220) that the Cypress npm package downloads the Cypress binary.

jeremywiebe (approved):

.claude/settings.json
jeresig: Yeah - I think that was exactly it!

.github/workflows/validate-workflows.yml
jeresig: I'm not sure! I feel that it's safe to not have to with our own internal actions (and it makes updating them a lot simpler, since we'd avoid having to put new commits up to every repo that uses them, for bug fixes). But I'm open to discussing this!

.github/npm-extra-domains.conf
jeresig: Ahh you're probably right - it was happening during the node_modules install step, but that's probably what was actually happening.

.github/workflows/gh-pages.yml
jeresig: Agreed - let me see if I can adjust some things...

PR #3480, opened 2026-04-10
Hide exports that aren't used in frontend from the documentation.

benchristel:
## Summary:
We want the documentation to be tidy so other teams can use it. This PR removes
clutter from unused exports.

Issue: LEMS-4014

## Test plan:

View the generated docs site.

### DIFF
--- a/packages/perseus-core/src/index.ts
+++ b/packages/perseus-core/src/index.ts
@@ -1,4 +1,5 @@
 export type {PerseusAnalyticsEvent, AnalyticsEventHandlerFn} from "./analytics";
+/** @hidden */
 export type {
     RendererInterface,
     MarkerType,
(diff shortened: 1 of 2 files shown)


packages/perseus-core/src/index.ts
```diff
@@ -7,191 +8,313 @@ export type {
...
 export {makeSafeUrl} from "./utils/make-safe-url";
 export {removeOrphanedWidgetsFromPerseusItem} from "./utils/remove-orphaned-widgets";
 
 export {
     parseAndMigratePerseusItem,
     parseAndMigratePerseusArticle,
+} from "./parse-perseus-json";
+/** @hidden */
+export {
     parseAndMigrateUserInputMap,
```
jeremywiebe, line 61: Is this not used by the service?

jeremywiebe (approved):
Makes sense.  It's pretty easy to reverse any of these changes to make them visible in the API docs.  Thanks for more cleanup!

packages/perseus-core/src/index.ts
benchristel: It is!

PR #3481, opened 2026-04-10
Add `showAngles` option to polygon locked figure

mojadem:
## Summary:
Extend the polygon locked figure to include a `showAngles` option similar to the polygon interactive graph.

I'm contributing upstream to Perseus becuase this feature would enable the Visual Output team to lean on Perseus for computing and displaying angles rather than trusting the LLM to do the math. I plan on following up with a change to allow these angle measure labels to be configurable per-vertex.

Feel free to nit-pick style / conventions, and let me know if I'm missing any tests!

## Test plan:
- `pnpm lint && pnpm test`
- Test that angle measure switch calls `onChangeProps` when toggled in locked polygon settings (mirrors existing test cases)
- Test that angle measures are rendered correctly (mirrors `packages/perseus/src/widgets/interactive-graphs/graphs/polygon.test.tsx`


mojadem:
I'll be following through with the CI failures next week!

packages/perseus-editor/src/__testdata__/interactive-graph-question-builder.ts
```diff
@@ -493,6 +494,7 @@ class InteractiveGraphQuestionBuilder {
             points: points,
             color: "grayH",
             showVertices: false,
+            showAngles: false,
```
SonicScrewdriver, line 497: We should pipe in the `options.showAngles` value here so that we can use the builder in tests/storybook. :)

packages/perseus/src/widgets/interactive-graphs/locked-figures/locked-polygon.test.tsx
```diff
@@ -0,0 +1,157 @@
+import {render, screen} from "@testing-library/react";
```
SonicScrewdriver, line 1: It seems like we were previously just using the global set of tests in `locked-function.test.tsx`, but I like the coverage and organization here, so I'm all far it. :)

packages/perseus/src/widgets/interactive-graphs/locked-figures/locked-polygon.tsx
```diff
@@ -51,6 +63,24 @@ const LockedPolygon = (props: LockedPolygonType) => {
...
+                    return null;
+                }
+                return (
+                    <PolygonAngle
+                        key={`locked-polygon-angle-${i}`}
+                        centerPoint={point}
+                        endPoints={[pt1, pt2]}
+                        areEndPointsClockwise={clockwise(points)}
+                        showAngles={showAngles}
+                        snapTo="grid"
```
SonicScrewdriver, line 80: I'm noticing that you've hard-coded this to grid! 

That's fine if that's all you will ever need for locked figures, but you may want to consider setting up the full angle snapping setting for locked polygons. If you don't need the full angle snapping logic, I'd love a comment here with a brief explanation for the hard coded value for our future selves. :)

packages/perseus/src/widgets/interactive-graphs/interactive-graph-question-builder.ts
```diff
@@ -522,6 +523,7 @@ class InteractiveGraphQuestionBuilder {
             points: points,
             color: "grayH",
             showVertices: false,
+            showAngles: false,
```
SonicScrewdriver, line 526: We should pipe in the options.showAngles value here so that we can use the builder in tests/storybook. :) (For both builders)

SonicScrewdriver (approved):
This honestly looks great to me, thank you so much for getting this feature built! 

I'm happy to approve as I only have a couple small change requests regarding piping through the showAngles value in the question builders, and the hard coded angle snapping. 

I'll chat with you elsewhere about when to land/deploy this, as we'll need to coordinate with the platform team due to the schema changes — otherwise we could bring down publishing if we attempt to use the new value. ;)

catandthemachines (approved):

PR #3482, opened 2026-04-10
[Interactive Graph] Reverted Asymptote/Drag Handle Auto-Blur on Drag End

ivyolamit:
## Summary:
- Reverted the auto-blur behavior in `MovableAsymptote` that was added in commit `607c6bc4ed`
- The drag handle now retains focus after dragging, consistent with movable point behavior

## Details

During the logarithm work, a fix was added to `MovableAsymptote` (`movable-asymptote.tsx`) that automatically
blurred the asymptote drag handle when a mouse drag ended. The intent was to prevent the focus ring from
persisting after the user released the mouse and moved away from the line.

**Why it was reverted:** The auto-blur made the drag handle behave differently from movable points in the graph.
Points retain focus after dragging — focus only clears when the user clicks away or navigates away via keyboard.
The asymptote drag handle should follow the same pattern for consistency.

**Current behavior:** After a mouse drag ends, the drag handle retains focus (focus ring remains visible).
Focus is only cleared when the user clicks elsewhere or navigates away via keyboard — matching how movable
points behave across all interactive graph types.

Issue: LEMS-4016

## Test plan:
1. Navigate to the Exponential Graph
2. Drag the asymptote and notice that the drag handle has the focused state
3. Hover away from the asymptote and notice it still had the focused state
4. Click another point or anywhere notice it will loose its focused state
5. Also confirm the same behavior when navigating the keyboard
6. Also confirm the same behavior for Logarithm Graph

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/__docs__/interactive-graph-regression.stories.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/__docs__/interactive-graph-regression.stories.tsx
@@ -530,6 +530,36 @@ export const ShowNoArrows: Story = {
     },
 };
 
+/* Asymptote drag handle regression tests */
+
+export const ExponentialAsymptoteFocused: Story = {
+    args: {
+        question: interactiveGraphQuestionBuilder()
+            .withExponential({asymptote: 3})
+            .build(),
+    },
+    play: async ({canvas}) => {
+        const asymptote = canvas.getByRole("button", {
+            name: /^Horizontal asymptote/,
+        });
+        asymptote.focus();
+    },
+};
+
+export const LogarithmAsymptoteFocused: Story = {
+    args: {
+        question: interactiveGraphQuestionBuilder()
+            .withLogarithm({asymptote: -3})
+            .build(),
+    },
+    play: async ({canvas}) => {
+        const asymptote = canvas.getByRole("button", {
+            name: /^Vertical asymptote/,
+        });
+        asymptote.focus();
+    },
+};
+
 function MafsQuestionRenderer(props: {question: PerseusRenderer}) {
     const {question} = props;
     return (
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-asymptote.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-asymptote.tsx
@@ -56,16 +56,6 @@ export function MovableAsymptote(props: Props) {
         constrainKeyboardMovement: constrainKeyboardMovement ?? ((p) => p),
     });
 
-    // When a mouse drag ends, blur the element so the focus ring doesn't
-    // persist after the user releases the mouse and moves away.
-    const wasDragging = React.useRef(false);
-    React.useEffect(() => {
-        if (wasDragging.current && !dragging) {
-            groupRef.current?.blur();
-        }
-        wasDragging.current = dragging;
-    }, [dragging]);
-
     return (
         <g
             ref={groupRef}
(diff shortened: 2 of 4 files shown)


ivyolamit:
@claude review once

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-asymptote.tsx
ivyolamit, line 1: During the logarithm implementation this was added to address the reported bug during the play test with the movable asymptote, but with the recent play test this behavior makes it inconsistent with the focus state with the graph points. See more info [here](https://khanacademy.atlassian.net/browse/LEMS-4016?focusedCommentId=470824)

packages/perseus/src/widgets/interactive-graphs/__docs__/interactive-graph-regression.stories.tsx
```diff
@@ -530,6 +530,36 @@ export const ShowNoArrows: Story = {
     },
 };
 
+/* Asymptote drag handle regression tests */
```
SonicScrewdriver, line 533: I _**love**_ the interactive test addition, but I have some small change requests before approving:

- I think these stories might need to be reviewed to make sure they're showing the desired state, as I'm not seeing the asymptote focused for any of them currently. 
- We created a separate interactive regression file/story for Radio (`multiple-choice-interactions-regression.stories.tsx`), which I would recommend that we do here as well to maintain consistency.

SonicScrewdriver (changes requested):
Great work — thank you!! The functionality looks like it's great to me, I just have some change requests for the new interactive stories before approving.

packages/perseus/src/widgets/interactive-graphs/__docs__/interactive-graph-regression.stories.tsx
ivyolamit: > I think these stories might need to be reviewed to make sure they're showing the desired state, as I'm not seeing the asymptote focused for any of them currently.

@SonicScrewdriver were you looking at storybook? I also saw that initially, but when I check the chromatic it shows up [there](https://www.chromatic.com/build?appId=650db21c3f5d1b2f13c02952&number=12576)

image

> We created a separate interactive regression file/story for Radio (multiple-choice-interactions-regression.stories.tsx), which I would recommend that we do here as well to maintain consistency.

Looking at [radio](https://khan.github.io/perseus/?path=/story/widgets-radio-visual-regression-tests-interactions--choice-text-color-in-multiple-select), i get what you mean now. Would you be okay if I add this state visual regression in the next PR i'm currently doing? since it also updates the same visual regression [here](https://github.com/Khan/perseus/pull/3484/changes#diff-efc3017e05c210922bd1b774fcb27f08eb417db54138bf8f84ea697ec405aaaeR535)

SonicScrewdriver: Yeah! I was checking storybook, and the stories do not show the handle focused. 

Image

Image

Hmm, if we're merging these two PRs at the same time (looks like they're stacked) I can approve this one for now if you'd like to make the updates in the second PR. We're currently missing the benefits of the interactive story, and I suspect that following that pattern will resolve the storybook focus-state issue as well. :)

SonicScrewdriver (approved):
Approved with the understanding that the interactive stories will be fixed up in [this PR](https://github.com/Khan/perseus/pull/3484)! :)

packages/perseus/src/widgets/interactive-graphs/__docs__/interactive-graph-regression.stories.tsx
ivyolamit: This is now addressed in this [PR](https://github.com/Khan/perseus/pull/3489/changes#diff-640466d76ddf5df9f983be3e8f60be89bf98b90aab12830a7571ad74aced2a75R1)

PR #3487, opened 2026-04-13
Fix release workflow network config for sigstore domains

jeremywiebe:
## Summary:

The last fix for the release failed because I used a comma to separate the extra domains instead of a space. This broke the action and caused the release to fail. 

This PR just removes the comma and leave the space. 

Issue: <none>

## Test plan:

 - After merging, trigger a release (the push-to-main trigger should handle it, or manually via `gh workflow run publish.yml --ref main -f run_type=release`)
  - Confirm the publish succeeds and packages show up on npm with provenance


nishasy (approved):
Looks good to me!

jeresig (approved):

jeresig:
Sorry about that - should've caught that in the previous review!

jeremywiebe:
> Sorry about that - should've caught that in the previous review!

All good. I should have been more careful!

PR #3488, opened 2026-04-13
[Free Response] Add Visual Regression Tests

mark-fitzgerald:
## Summary:
Reviewed Free Response widget for color and font tokens. This widget leverages Wonder Blocks and has no custom styling that uses color or fonts. Adjusted how the widget handles the informational/error text below the text area. Therefore, just adding visual regression tests for the widget for future reference with any changes.

Issue: LEMS-3502

## Test plan:
1. Launch Storybook
2. Navigate to the Free Response widget
3. Notice that a folder for Visual Regression Tests exists
   - One for initial state of the widget
   - One that involves interactions


packages/perseus/src/widgets/__testutils__/free-response-renderer-decorator.tsx
```diff
@@ -0,0 +1,55 @@
...
+                        }),
+                    }),
+                },
+            })}
+            apiOptions={parameters?.apiOptions}
+        />
+    );
+};
+
+function FreeResponseQuestionRenderer(props: {
```
nishasy: I think you can remove this `FreeResponseQuestionRenderer` function, and just replace its usage with `QuestionRendererForStories` like in `image-renderer-decorator.tsx`.

packages/perseus/src/widgets/free-response/__docs__/free-response-initial-state-regression.stories.tsx
```diff
@@ -0,0 +1,63 @@
...
+import {freeResponseRendererDecorator} from "../../__testutils__/free-response-renderer-decorator";
+import {rtlDecorator} from "../../__testutils__/story-decorators";
+
+import type {Meta, StoryObj} from "@storybook/react-vite";
+
+const FreeResponseWidget = getWidget("free-response")!;
+
+const meta: Meta<typeof FreeResponseWidget> = {
+    title: "Widgets/Free Response/Visual Regression Tests/Initial State",
+    component: FreeResponseWidget,
```
nishasy, line 12: Yay!

packages/perseus/src/widgets/free-response/__docs__/free-response-initial-state-regression.stories.tsx
```diff
@@ -0,0 +1,63 @@
...
+    tags: ["!dev"],
```
nishasy: Can we use `tags: ["!autodocs"]` here instead? I personally prefer having the stories in standalone pages, and we don't need the `docs` page that appears when you remove all tags.

packages/perseus/src/widgets/free-response/__docs__/free-response-interactions-regression.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+export const OverCharacterLimit = {
+    decorators: [freeResponseRendererDecorator],
+    args: {
+        allowUnlimitedCharacters: false,
+        characterLimit: 10,
+        placeholder: "Enter your answer here",
+        question: "What is the theme of the essay?",
+    },
+    play: async ({canvasElement, userEvent}) => {
+        const canvas = within(canvasElement);
```
nishasy: We don't need this line anymore, now that the linter error that yells at you about `screen` usage is ignored in regression stories.

```suggestion
    play: async ({canvas, userEvent}) => {
```

packages/perseus/src/widgets/free-response/__docs__/free-response-interactions-regression.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+export const RightToLeftOverCharacterLimit = {
+    decorators: [freeResponseRendererDecorator, rtlDecorator],
+    args: {
+        allowUnlimitedCharacters: false,
+        characterLimit: 10,
+        placeholder: "Enter your answer here",
+        question: "What is the theme of the essay?",
+    },
+    play: async ({canvasElement, userEvent}) => {
+        const canvas = within(canvasElement);
```
nishasy: Same thing here

```suggestion
    play: async ({canvas, userEvent}) => {
```

packages/perseus/src/widgets/free-response/free-response.tsx
```diff
@@ -111,15 +82,18 @@ export class FreeResponse extends React.Component<Props> implements Widget {
...
                             onChange={this._handleUserInput}
                             placeholder={this.props.placeholder}
                             style={styles.textarea}
                             value={this.props.userInput.currentValue}
                         />
                     }
+                    additionalHelperMessage={
+                        isOverLimit ? undefined : characterCountText
+                    }
+                    errorMessage={isOverLimit ? characterCountText : undefined}
```
nishasy: This is so much nicer!

packages/perseus/src/widgets/__testutils__/free-response-renderer-decorator.tsx
mark-fitzgerald: Done.

packages/perseus/src/widgets/free-response/__docs__/free-response-initial-state-regression.stories.tsx
mark-fitzgerald: Done.

packages/perseus/src/widgets/free-response/__docs__/free-response-interactions-regression.stories.tsx
mark-fitzgerald: Done.

packages/perseus/src/widgets/free-response/__docs__/free-response-interactions-regression.stories.tsx
mark-fitzgerald: Done.

packages/perseus/src/widgets/free-response/free-response.tsx
Myranae: I was reviewing with Claude, and it pointed out a couple things with this change that I wanted to check on. I think this is probably the better way, but I'm still learning a11y coding :) 

  ▎ The old character count used `role="status"` (a polite live region), so screen readers announced the
  ▎  count on every keystroke. The new `additionalHelperMessage` appears to be a static element — 
  ▎ meaning updates while typing wouldn't be announced, and users wouldn't hear they're approaching   
  ▎ the limit. Is this intentional, or does `LabeledField` handle live updates internally? Also       
  ▎ confirming: once over the limit, the `errorMessage` / `aria-live="assertive"` kicks in — is that
  ▎right?

packages/perseus/src/widgets/free-response/__docs__/free-response-interactions-regression.stories.tsx
Myranae: We may want to update our instructions doc if we don't want to use `within` anymore as I think it is explicitly called out to use it. It sounds like keeping `within` could help scope the query to just the story's canvas, but not sure how necessary that is.

Update: It might just be called out in my local files actually, so we might just want to check :)

packages/perseus/src/widgets/free-response/free-response.tsx
mark-fitzgerald: Great call out! I will check with Adam on this. My initial thought is that announcing keystroke information would be quite verbose and annoying. However, there might be a SRX requirement for this by design.

packages/perseus/src/widgets/free-response/__docs__/free-response-interactions-regression.stories.tsx
nishasy: Ohhh, I forgot to remove `within` from `regression-stories.md`.

I can update and re-upload.

nishasy (approved):
Looks good to me!

packages/perseus/src/widgets/free-response/free-response.tsx
```diff
@@ -47,6 +48,15 @@ export class FreeResponse extends React.Component<Props> implements Widget {
         },
     };
 
+    announceCharacterCount = (messageRaw: string, isOverLimit: boolean) => {
+        const level = isOverLimit ? "assertive" : "polite";
+        const message =
+            (isOverLimit ? "Error: " : "") +
+            messageRaw.replace("/", "out of") +
+            " used.";
```
aag: How does this interact with localization? It seems like this wouldn't work with non-English languages.

mark-fitzgerald: Great catch!! I'll revert this back to just use the original character count text.

PR #3492, opened 2026-04-13
[Preview NG] Add functions to sanitize objects for sending to Preview

jeremywiebe:
## Summary:

_This PR is part of a series building a typed, hook-based preview system for the Perseus editor. The new system replaces the untyped window.iframeDataStore + raw postMessage(string) communication with structured, validated message passing via usePreviewHost and usePreviewClient hooks. The new system is being built alongside the old one — no existing behaviour changes until the final PR in the series flips the switch._

* https://github.com/Khan/perseus/pull/3464
* https://github.com/Khan/perseus/pull/3467 
* https://github.com/Khan/perseus/pull/3474

---

Adds `sanitizeApiOptions()` and `sanitizePreviewData()` which strip non-serializable values (function callbacks, React components) from preview data before it's sent via `postMessage`. The structured clone algorithm used by `postMessage` can't handle functions, so they need to be removed before sending.                                                               
                                                                                                                            
`sanitizeApiOptions()` destructures `APIOptions` to remove known non-serializable fields (`onFocusChange`, `trackInteraction`, `imagePreloader`, `nativeKeypadProxy`, placeholder components, etc.) and returns only the serializable remainder.                                                                                                                  
                                                                      
`sanitizePreviewData()` wraps this for all preview content types — question, hint, article, and article-all. The article-all case currently sanitizes per-section since each `ArticlePreviewData` carries its own `apiOptions`; this will be simplified when we restructure the preview data types in a later PR.
                                                                      
Purely additive — no existing behaviour is changed.                                                                          
                                                                                                                            
**Depends on:** #3474

Issue: LEMS-3741
                 
## Test plan:                           
                                                                                                                            
`pnpm test`

### DIFF
--- a/packages/perseus-editor/src/preview/preview-data-sanitizer.test.ts
+++ b/packages/perseus-editor/src/preview/preview-data-sanitizer.test.ts
@@ -305,13 +305,9 @@ describe("sanitizePreviewData", () => {
             if (result.type === "article-all") {
                 expect(result.data).toHaveLength(2);
                 expect(result.data[0].apiOptions.readOnly).toBe(true);
-                expect(
-                    result.data[0].apiOptions.onFocusChange,
-                ).toBeUndefined();
+                expect(result.data[0].apiOptions.onFocusChange).toBeUndefined();
                 expect(result.data[1].apiOptions.readOnly).toBe(false);
-                expect(
-                    result.data[1].apiOptions.onFocusChange,
-                ).toBeUndefined();
+                expect(result.data[1].apiOptions.onFocusChange).toBeUndefined();
             }
         });
 
@@ -385,9 +381,7 @@ describe("sanitizePreviewData", () => {
             expect(result.type).toBe("article-all");
             if (result.type === "article-all") {
                 expect(result.data).toHaveLength(2);
-                expect(
-                    result.data[0].apiOptions.onFocusChange,
-                ).toBeUndefined();
+                expect(result.data[0].apiOptions.onFocusChange).toBeUndefined();
                 // Section with null apiOptions should be unchanged
                 expect(result.data[1].apiOptions).toBeNull();
             }
--- a/packages/perseus-editor/src/preview/preview-data-sanitizer.test.ts
+++ b/packages/perseus-editor/src/preview/preview-data-sanitizer.test.ts
@@ -1,4 +1,5 @@
 import {getDefaultAnswerArea} from "@khanacademy/perseus-core";
+import invariant from "tiny-invariant";
 
 import {sanitizePreviewData} from "./preview-data-sanitizer";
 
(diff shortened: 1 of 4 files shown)


packages/perseus-editor/src/preview/preview-data-sanitizer.ts
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Sanitizes preview content by removing non-serializable functions and React
+ * components from apiOptions before sending via postMessage.
+ *
+ * NOTE: The article-all case currently sanitizes apiOptions on each section
+ * individually because the current type is ArticlePreviewData[] (each section
+ * carries its own apiOptions). This will be simplified to a single apiOptions
+ * when we restructure the preview data types.
+ */
+export function sanitizePreviewData(
```
jeremywiebe, line 14: The preview system needs to send API options into the preview panel because some API Options can affect behaviour/rendering. But, some API Options are not serializable and deal with host integration (for example, callbacks) and so we need to sanitize the options before sending them over the "bridge".

packages/perseus-editor/src/preview/sanitize-api-options.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    const {
+        onFocusChange: _,
+        answerableCallback: __,
+        getAnotherHint: ___,
+        interactionCallback: ____,
+        imagePreloader: _____,
+        trackInteraction: ______,
+        setDrawingAreaAvailable: ________,
+        baseElements: _________,
```
jeremywiebe, line 30: This is one key that can affect rendering. It is used in some contexts in the frontend application, but since this contains React elements, we can't send it across the bridge.  In reality, host applications only set the `Link` element within this key and generally use it to override the behaviour of _handling_ the clicked link vs. actually changing how it renders.

packages/perseus-editor/src/preview/preview-data-sanitizer.test.ts
```diff
@@ -0,0 +1,496 @@
...
+
+            const previewContent: PreviewContent = {
+                type: "question",
+                data: questionData,
+            };
+
+            const result = sanitizePreviewData(previewContent);
+
+            expect(result.type).toBe("question");
+            if (result.type === "question") {
```
mark-fitzgerald: Is this conditional needed? If `result.type` does not equal "question", then the assertion before it will fail and cause the test to exit, correct? Is this here for some TypeScript reason?

packages/perseus-editor/src/preview/preview-data-sanitizer.test.ts
```diff
@@ -0,0 +1,496 @@
...
+            };
+
+            const result = sanitizePreviewData(previewContent);
+
+            expect(result.type).toBe("question");
+            // Should return unchanged when apiOptions is null
+            expect(result).toEqual(previewContent);
+        });
+
+        it("does not mutate original question data", () => {
```
mark-fitzgerald, line 109: Great callout!!

mark-fitzgerald (approved):

packages/perseus-editor/src/preview/preview-data-sanitizer.test.ts
jeremywiebe: This is needed because the previous assertion doesn't ["narrow"](https://www.typescriptlang.org/docs/handbook/2/narrowing.html) the type. Without the `if` block, line 63+ don't know that we're dealing with a result that is a `question` type. I'll see if I can change how we do that. I know we use the `tinyinvariant` library in places and I think that narrows properly.

PR #3493, opened 2026-04-14
docs(changeset): Implementing bug fix for jumping MovableLines in the Correct Answer graph in the editor

SonicScrewdriver:
## Summary:
[deltas-creep] Resolve bug related to the MovableLine component in the "Correct Answer" graph in the editor.

This is just a draft as I'm pitching that we resolve this now, due to the effect it has on the upcoming Vector Graph. 

Issue: LEMS-4044

## Test plan:
- Tests pass

### DIFF
--- a/.changeset/olive-planes-raise.md
+++ b/.changeset/olive-planes-raise.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Implementing bug fix for jumping MovableLines in the Correct Answer graph in the editor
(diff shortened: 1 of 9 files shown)


ivyolamit (approved):
Changes looks logical to me. Thanks for this fix Third, do you mind adding a test item for this for our play test this Thursday just to ensure that it has no regression that we missed. And also update this [QA ticket](https://khanacademy.atlassian.net/browse/LEMS-4024) to capture the scenario 🙏🏼

PR #3495, opened 2026-04-15
[Preview NG] Introduce hooks to manage preview communication

jeremywiebe:
## Summary:

Previous PRs in this series:
- #3464
- #3467
- #3474
- #3492

---

Adds the two core hooks that implement the new preview communication protocol.

`usePreviewController(iframeRef)` is used on the editor/parent side. It sends sanitized preview data to an iframe via typed `postMessage`, listens for height updates back, and filters incoming messages by `event.source` to ensure they come from the correct iframe. Returns `{sendData, height}`.

`usePreviewPresenter()` is used inside the preview iframe. It listens for content data from the parent, sends back height reports and data requests via the typed message protocol. Throws if used outside an iframe. Returns `{data, reportHeight}`.

Both hooks use the message types and validators from #3474 and the data sanitizer from #3492. Purely additive — no existing behaviour is changed.

Issue: LEMS-3741

## Test plan:

`pnpm test`

### DIFF
--- a/packages/perseus-editor/src/preview/message-types.ts
+++ b/packages/perseus-editor/src/preview/message-types.ts
@@ -84,9 +84,9 @@ export type ParentToIframeMessage = PreviewDataMessage;
 // ---- Iframe → Parent messages ----
 
 /**
- * Message from iframe requesting data from parent
+ * Message from iframe to parent telling it the iframe is ready
  */
-interface PreviewDataRequestMessage extends PreviewMessageBase {
+interface PreviewIframeReadyMessage extends PreviewMessageBase {
     type: "iframe-ready";
 }
 
@@ -102,5 +102,12 @@ interface PreviewHeightUpdateMessage extends PreviewMessageBase {
  * Union of all messages sent from iframe to parent
  */
 export type IframeToParentMessage =
-    | PreviewDataRequestMessage
+    | PreviewIframeReadyMessage
     | PreviewHeightUpdateMessage;
+
+export function createPreviewIframeReadyMessage(): PreviewIframeReadyMessage {
+    return {
+        source: PREVIEW_MESSAGE_SOURCE,
+        type: "iframe-ready",
+    };
+}
--- a/packages/perseus-editor/src/preview/message-types.ts
+++ b/packages/perseus-editor/src/preview/message-types.ts
@@ -71,10 +71,10 @@ export type PreviewContent =
 /**
  * Message from parent sending content data to iframe
  */
-type PreviewDataMessage = PreviewMessageBase & {
+interface PreviewDataMessage extends PreviewMessageBase {
     type: "content-data";
     content: PreviewContent;
-};
+}
 
 /**
  * Union of all messages sent from parent to iframe
@@ -86,17 +86,17 @@ export type ParentToIframeMessage = PreviewDataMessage;
 /**
  * Message from iframe requesting data from parent
  */
-type PreviewDataRequestMessage = PreviewMessageBase & {
+interface PreviewDataRequestMessage extends PreviewMessageBase {
     type: "request-data";
-};
+}
 
 /**
  * Message from iframe reporting its content height
  */
-type PreviewHeightUpdateMessage = PreviewMessageBase & {
+interface PreviewHeightUpdateMessage extends PreviewMessageBase {
     type: "height-update";
     height: number;
-};
+}
 
 /**
  * Union of all messages sent from iframe to parent
--- a/packages/perseus-editor/src/preview/message-types.ts
+++ b/packages/perseus-editor/src/preview/message-types.ts
@@ -24,17 +24,6 @@ interface PreviewMessageBase {
 }
 
 /**
- * Base type for messages with iframe ID
- *
- * Note: The ID is primarily used for debugging/logging purposes.
- * Message routing is handled by event.source filtering in the hooks,
- * not by comparing ID strings.
- */
-interface PreviewMessageWithId extends PreviewMessageBase {
-    id: string;
-}
-
-/**
  * Data for question preview (full item with question, answer area, and hints)
  */
 export type QuestionPreviewData = {
@@ -82,7 +71,7 @@ export type PreviewContent =
 /**
  * Message from parent sending content data to iframe
  */
-type PreviewDataMessage = PreviewMessageWithId & {
+type PreviewDataMessage = PreviewMessageBase & {
     type: "content-data";
     content: PreviewContent;
 };
@@ -97,14 +86,14 @@ export type ParentToIframeMessage = PreviewDataMessage;
 /**
  * Message from iframe requesting data from parent
  */
-type PreviewDataRequestMessage = PreviewMessageWithId & {
+type PreviewDataRequestMessage = PreviewMessageBase & {
     type: "request-data";
 };
 
 /**
  * Message from iframe reporting its content height
  */
-type PreviewHeightUpdateMessage = PreviewMessageWithId & {
+type PreviewHeightUpdateMessage = PreviewMessageBase & {
     type: "height-update";
     height: number;
 };
--- a/packages/perseus-editor/src/preview/message-types.ts
+++ b/packages/perseus-editor/src/preview/message-types.ts
@@ -87,7 +87,7 @@ export type ParentToIframeMessage = PreviewDataMessage;
  * Message from iframe requesting data from parent
  */
 interface PreviewDataRequestMessage extends PreviewMessageBase {
-    type: "request-data";
+    type: "iframe-ready";
 }
 
 /**
--- a/packages/perseus-editor/src/preview/message-types.ts
+++ b/packages/perseus-editor/src/preview/message-types.ts
@@ -110,17 +110,8 @@ type PreviewHeightUpdateMessage = PreviewMessageWithId & {
 };
 
 /**
- * Message from iframe reporting lint warnings
- */
-type PreviewLintReportMessage = PreviewMessageWithId & {
-    type: "lint-report";
-    lintWarnings: ReadonlyArray<any>;
-};
-
-/**
  * Union of all messages sent from iframe to parent
  */
 export type IframeToParentMessage =
     | PreviewDataRequestMessage
-    | PreviewHeightUpdateMessage
-    | PreviewLintReportMessage;
+    | PreviewHeightUpdateMessage;
--- a/packages/perseus-editor/src/preview/use-preview-controller.ts
+++ b/packages/perseus-editor/src/preview/use-preview-controller.ts
@@ -114,9 +114,7 @@ export function usePreviewController(
         return () => {
             window.removeEventListener("message", handleMessage);
         };
-        // iframeRef is intentionally excluded - it's a stable ref that shouldn't trigger re-runs
-        // eslint-disable-next-line react-hooks/exhaustive-deps
-    }, []);
+    }, [iframeRef]);
 
     // Memoized function to send data to iframe
     const sendData = React.useCallback(
@@ -146,9 +144,7 @@ export function usePreviewController(
 
             contentWindow.postMessage(message, "/");
         },
-        // iframeRef is intentionally excluded - it's a stable ref that shouldn't trigger re-runs
-        // eslint-disable-next-line react-hooks/exhaustive-deps
-        [],
+        [iframeRef],
     );
 
     return {
--- a/packages/perseus-editor/src/preview/use-preview-controller.ts
+++ b/packages/perseus-editor/src/preview/use-preview-controller.ts
@@ -58,7 +58,10 @@ export function usePreviewController(
     React.useEffect(() => {
         const handleMessage = (event: MessageEvent) => {
             // Filter by source window - only messages from OUR iframe
-            if (event.source !== iframeRef.current?.contentWindow) {
+            if (
+                iframeRef.current?.contentWindow == null ||
+                event.source !== iframeRef.current?.contentWindow
+            ) {
                 return;
             }
 
--- a/packages/perseus-editor/src/preview/use-preview-controller.ts
+++ b/packages/perseus-editor/src/preview/use-preview-controller.ts
@@ -77,19 +77,17 @@ export function usePreviewController(
                 case "iframe-ready": {
                     // Send the pending message (if any)
                     if (pendingDataRef.current) {
-                        const contentWindow = iframeRef.current?.contentWindow;
-                        if (contentWindow) {
-                            const sanitizedData = sanitizePreviewData(
-                                pendingDataRef.current,
-                            );
+                        const sanitizedData = sanitizePreviewData(
+                            pendingDataRef.current,
+                        );
 
-                            const msg: ParentToIframeMessage = {
-                                source: PREVIEW_MESSAGE_SOURCE,
-                                type: "content-data",
-                                content: sanitizedData,
-                            };
-                            contentWindow.postMessage(msg, "/");
-                        }
+                        const msg: ParentToIframeMessage = {
+                            source: PREVIEW_MESSAGE_SOURCE,
+                            type: "content-data",
+                            content: sanitizedData,
+                        };
+                        iframeRef.current.contentWindow.postMessage(msg, "/");
+
                         // Clear the pending data
                         pendingDataRef.current = null;
                     }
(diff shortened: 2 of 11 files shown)


packages/perseus-editor/src/preview/use-preview-controller.ts
```diff
@@ -0,0 +1,159 @@
+import {Log} from "@khanacademy/perseus";
+import * as React from "react";
+
+import {PREVIEW_MESSAGE_SOURCE} from "./message-types";
+import {isIframeToParentMessage} from "./message-validators";
+import {sanitizePreviewData} from "./preview-data-sanitizer";
+
+import type {ParentToIframeMessage, PreviewContent} from "./message-types";
+
+type UsePreviewHostResult = {
```
benchristel: Should this be

```suggestion
type UsePreviewControllerResult = {
```

?

packages/perseus-editor/src/preview/use-preview-controller.ts
```diff
@@ -0,0 +1,159 @@
...
+            }
+        };
+
+        window.addEventListener("message", handleMessage);
+
+        return () => {
+            window.removeEventListener("message", handleMessage);
+        };
+        // iframeRef is intentionally excluded - it's a stable ref that shouldn't trigger re-runs
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
benchristel: > iframeRef is intentionally excluded - it's a stable ref that shouldn't trigger re-runs

This hook can't guarantee that `iframeRef` is actually a stable ref, since the caller can pass anything with a `current: HTMLIFrameElement` property. I'd prefer to add `iframeRef` to the deps — if it is a stable reference, it won't trigger re-renders, and if it's not, we get predictable behavior (and no linter suppression).

packages/perseus-editor/src/preview/use-preview-controller.ts
```diff
@@ -0,0 +1,159 @@
...
+                }
+
+                case "height-update":
+                    setHeight(message.height);
+                    break;
+
+                case "lint-report":
+                    Log.log("LINTER REPORT", {
+                        lintWarnings: JSON.stringify(message.lintWarnings),
+                    });
```
benchristel: It seems weird to handle this event just to log it. Why couldn't we log within the iframe? I assume this is preserving existing behavior?

packages/perseus-editor/src/preview/use-preview-controller.ts
```diff
@@ -0,0 +1,159 @@
...
+
+            // Handle the message
+            switch (message.type) {
+                case "request-data": {
+                    // Store the iframe ID (used for debugging/logging, not routing)
+                    iframeIdRef.current = String(message.id);
+
+                    // Send the pending message (if any)
+                    if (pendingDataRef.current) {
+                        const iframe = iframeRef.current;
```
benchristel: `iframe` is only used once — could be inlined.

packages/perseus-editor/src/preview/use-preview-controller.ts
```diff
@@ -0,0 +1,159 @@
...
+                                pendingDataRef.current,
+                            );
+
+                            const msg: ParentToIframeMessage = {
+                                source: PREVIEW_MESSAGE_SOURCE,
+                                type: "content-data",
+                                id: iframeIdRef.current,
+                                content: sanitizedData,
+                            };
+                            contentWindow.postMessage(msg, "*");
```
benchristel: It's *probably* okay to use `"*"` as the target origin here, because this component won't be used with non-KA frames, and we aren't sending any sensitive data.

Still, it might be good to use `"/"` instead, if we can, to prevent cross-origin communication. Alternatively, the caller of this hook could explicitly pass in the target origin.

See: https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage

packages/perseus-editor/src/preview/use-preview-controller.ts
```diff
@@ -0,0 +1,159 @@
...
+            const sanitizedData = sanitizePreviewData(data);
+
+            const message: ParentToIframeMessage = {
+                source: PREVIEW_MESSAGE_SOURCE,
+                type: "content-data",
+                id: iframeIdRef.current,
+                content: sanitizedData,
+            };
+
+            contentWindow.postMessage(message, "*");
```
benchristel: This block of code is duplicated in `handleMessage` above.

packages/perseus-editor/src/preview/use-preview-controller.ts
```diff
@@ -0,0 +1,159 @@
...
+            window.removeEventListener("message", handleMessage);
+        };
+        // iframeRef is intentionally excluded - it's a stable ref that shouldn't trigger re-runs
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, []);
+
+    // Memoized function to send data to iframe
+    const sendData = React.useCallback(
+        (data: PreviewContent) => {
+            const iframe = iframeRef.current;
```
benchristel: `iframe` is only used once — could be inlined.

packages/perseus-editor/src/preview/use-preview-presenter.ts
```diff
@@ -0,0 +1,179 @@
...
+import type {
+    IframeToParentMessage,
+    ParentToIframeMessage,
+    PreviewContent,
+} from "./message-types";
+
+type UsePreviewClientResult = {
+    /**
+     * The iframe's unique identifier (from data-id attribute). Use for
+     * debugging/logging, but not for message routing.
```
benchristel: Why is it not okay to use the id for message routing?

packages/perseus-editor/src/preview/use-preview-presenter.ts
```diff
@@ -0,0 +1,179 @@
...
+import {PREVIEW_MESSAGE_SOURCE} from "./message-types";
+import {isParentToIframeMessage} from "./message-validators";
+
+import type {
+    IframeToParentMessage,
+    ParentToIframeMessage,
+    PreviewContent,
+} from "./message-types";
+
+type UsePreviewClientResult = {
```
benchristel:
```suggestion
type UsePreviewPresenterResult = {
```

packages/perseus-editor/src/preview/use-preview-presenter.ts
```diff
@@ -0,0 +1,179 @@
...
+ * - Requests initial data on mount
+ * - Provides functions to report height and lint warnings back to parent
+ *
+ * @example
+ * ```tsx
+ * function PreviewPage() {
+ *   const { data, isMobile, reportHeight } = usePreviewPresenter();
+ *
+ *   React.useEffect(() => {
+ *     if (containerRef.current) {
```
benchristel, line 50: `containerRef` is not defined in this example.

packages/perseus-editor/src/preview/use-preview-presenter.ts
```diff
@@ -0,0 +1,179 @@
...
+        if (id == null) {
+            throw new Error(
+                "usePreviewPresenter could not identify its id from the hosting iframe",
+            );
+        }
+
+        setIframeId(id);
+        setIsMobile(mobile);
+        setHasLintGutter(lintGutter);
+    }, []);
```
benchristel: Does this need to be a `useEffect`? `window.frameElement` should always exist if this code is running in an iframe — we don't need to wait for any React rendering to happen before getting the dataset attributes.

I think we could get rid of the `useState` calls above and just return the values from this hook:

```ts
const iframe = window.frameElement as HTMLIFrameElement | null;

// ...

return {
    data,
    isMobile: iframe.dataset.mobile === "true",
    hasLintGutter: iframe.dataset.lintGutter === "true",
    id: iframe.dataset.id,
    reportHeight,
    reportLintWarnings,
};
```

packages/perseus-editor/src/preview/use-preview-presenter.ts
```diff
@@ -0,0 +1,179 @@
...
+
+            const message = event.data;
+
+            // Ignore messages that aren't from Perseus preview system
+            if (!isParentToIframeMessage(message)) {
+                return;
+            }
+
+            // Handle content data
+            // Note: ID check is for extra validation/debugging; actual routing is by event.source
```
benchristel: > Note: ID check is for extra validation/debugging

Do we need the ID check, then? Routing by `event.source` is sufficient and correct. Also, there's no debugging code here, so it's not clear what that part of the comment refers to.

packages/perseus-editor/src/preview/use-preview-presenter.ts
benchristel: On second thought — I think the client/server metaphor is better (this hook doesn't actually "present" or render anything; rather, it manages communication) but the iframe is the server, not the client. The main window is the client because it drives the content in this iframe.

benchristel (commented):
Thanks for doing this! Our iframe communication code has been neglected for a long time.

I still need to do a thorough review of the tests. My overall thoughts at this point are that we're trying to use AI to redesign a big chunk of existing code, and I don't think AI is good at this type of work. Hopefully my inline comments demonstrate why.

It seems like a lot of things could be simplified if we rewrote this by hand with a similar architecture to what's proposed here. Happy to pair on it if you want!

packages/perseus-editor/src/preview/use-preview-controller.ts
jeremywiebe: Thanks for that nudge Bed.  I'll update all the `postMessage()` calls in this new code to avoid using `"*"` for the `targetOrigin`!

packages/perseus-editor/src/preview/use-preview-controller.ts
jeremywiebe: Yes!  Thanks for catching that!

packages/perseus-editor/src/preview/use-preview-controller.ts
jeremywiebe: Yes, that's correct (see [here](https://github.com/Khan/perseus/blob/7fcef1a49d15bc9ba524a39d00b51f88e5e49bec/packages/perseus-editor/src/iframe-content-renderer.tsx#L41)).  

However, looking at the frontend codebase, it doesn't seem like we ever dispatch this message (in the original, it's a message that has a `lintWarnings` property on the message object. 

So, perhaps we should just drop it. 🤔   I'll remove it and we can always add the message and its handler back in later.

packages/perseus-editor/src/preview/use-preview-presenter.ts
jeremywiebe: I'm going to fix it to match the hook for now. But I'd like to pair with you this week to look at this entire PR in more detail. We can talk naming then. :)

packages/perseus-editor/src/preview/use-preview-presenter.ts
jeremywiebe: Actually, thinking about this again, I don't think that we need an `id` at all.  

The `usePreviewController` hook accepts an iframe ref. That ref is what's used to send messages down into the iframe. Messages that are posted back to the parent are filtered against the hook's instance of the iframe ref (`if (event.source !== iframeRef.current?.contentWindow) { ... }`) so we already have filtering in both directions.

I also checked and we don't log this ID anywhere.

packages/perseus-editor/src/preview/message-types.ts
```diff
@@ -23,17 +23,6 @@ interface PreviewMessageBase {
...
 
-/**
- * Base type for messages with iframe ID
- *
- * Note: The ID is primarily used for debugging/logging purposes.
- * Message routing is handled by event.source filtering in the hooks,
- * not by comparing ID strings.
- */
-interface PreviewMessageWithId extends PreviewMessageBase {
-    id: string;
```
jeremywiebe, line 34: Based on Ben's question about why the ID shouldn't be used for routing, I realized that the "gating" by message source that we do in the message handlers is sufficient. The `id` is just an extra value that doesn't add any further safety and so is just extra work for no benefit.

packages/perseus-editor/src/preview/use-preview-presenter.ts
jeremywiebe: No, I've now removed IDs from this PR entirely!  Thanks for the nudge!

packages/perseus-editor/src/preview/use-preview-controller.ts
jeremywiebe: Deduped.

packages/perseus-editor/src/preview/use-preview-controller.ts
jeremywiebe: That's a good point.  In some ways it even feels weird passing a ref into the hook. However, I can't think of way to allow the caller to define where the `<iframe>` is placed while giving the hook control.  

I've added the ref to the dependency list, which addresses your immediate comment.

packages/perseus-editor/src/preview/message-types.ts
```diff
@@ -97,30 +86,21 @@ export type ParentToIframeMessage = PreviewDataMessage;
...
+type PreviewHeightUpdateMessage = PreviewMessageBase & {
     type: "height-update";
     height: number;
 };
 
-/**
- * Message from iframe reporting lint warnings
- */
-type PreviewLintReportMessage = PreviewMessageWithId & {
-    type: "lint-report";
```
jeremywiebe, line 116: This feature exists today, but it just logs the lint report to the JS console, so effectively, it does nothing!  Also, the frontend doesn't actually send this data so we're waiting for a message that'll never come!

We'll bring it back if/when we actually want to implement it.

packages/perseus-editor/src/preview/message-types.ts
```diff
@@ -97,30 +86,21 @@ export type ParentToIframeMessage = PreviewDataMessage;
...
-type PreviewDataRequestMessage = PreviewMessageWithId & {
+type PreviewDataRequestMessage = PreviewMessageBase & {
     type: "request-data";
 };
 
 /**
  * Message from iframe reporting its content height
  */
-type PreviewHeightUpdateMessage = PreviewMessageWithId & {
+type PreviewHeightUpdateMessage = PreviewMessageBase & {
```
benchristel: Could we turn all of these `&`s into `extends`?

Use of `&` with object types can cause typechecking [performance] and correctness issues.

[performance]: https://github.com/microsoft/TypeScript/wiki/Performance#preferring-interfaces-over-intersections

On second thought, it seems like `PreviewMessageBase` has only a single property that always has the same value (`"perseus-preview"`). Could we get rid of it entirely? Perhaps we could replace `isIframeToParentMessage` and `isParentToIframeMessage` with typesafe parsers? Then we might not even need to write the message types out, because the parsers would generate them...

packages/perseus-editor/src/preview/use-preview-controller.ts
```diff
@@ -0,0 +1,146 @@
...
+): UsePreviewControllerResult {
+    const [height, setHeight] = React.useState<number | null>(null);
+    const [isIframeReady, setIsIframeReady] = React.useState(false);
+    const pendingDataRef = React.useRef<PreviewContent | null>(null);
+
+    // Listen for messages from iframe
+    React.useEffect(() => {
+        const handleMessage = (event: MessageEvent) => {
+            // Filter by source window - only messages from OUR iframe
+            if (event.source !== iframeRef.current?.contentWindow) {
```
jeremywiebe: We should protect against `iframeRef.current` being `undefined`.

packages/perseus-editor/src/preview/use-preview-controller.ts
```diff
@@ -0,0 +1,146 @@
...
+            const message = event.data;
+
+            // Check if it's a Perseus preview message
+            if (!isIframeToParentMessage(message)) {
+                return;
+            }
+
+            // Handle the message
+            switch (message.type) {
+                case "request-data": {
```
jeremywiebe: Rename to `iframe-ready`?

packages/perseus-editor/src/preview/use-preview-controller.ts
```diff
@@ -0,0 +1,146 @@
...
+
+            if (!contentWindow) {
+                return;
+            }
+
+            // If iframe hasn't sent request-data yet, store the data
+            if (!isIframeReady) {
+                pendingDataRef.current = data;
+                return;
+            }
```
jeremywiebe, line 121: Move this check before the check for `contentWindow` so we don't lose pending data when `contentWindow` isn't yet set.

jeremywiebe (commented):
Review feedback from pairing with Ben!

packages/perseus-editor/src/preview/message-types.ts
jeremywiebe: I like both ideas.  I hadn't thought about making real parsers for these messages, but I love that idea.  I think we already converted to interfaces as we paired and I'll plan to introduce a parser in the next PR instead of making this one any bigger.

jeremywiebe: Oh, I forgot to reply about the need for the `"perseus-preview"` constant in all messages.  This helps to filter out messages that might arrive in the parent or iframe child documents from non-Perseus sources. In my testing, I've found that Storybook as well as some browser extensions use `postMessage`. If we don't have a common key that helps to differentiate messages, it becomes tricky to easily ignore messages we aren't interested in.

catandthemachines (approved):
Looks good to me 😊

PR #3496, opened 2026-04-15
[A11y | Numeric Input ] Adding a fallback label for Numeric Input and a Linter Warning for the Editor if the label is missing. 

SonicScrewdriver:
## Summary:
[back-up-plan-lisa-needs-basis] This PR seeks to improve the a11y of the Numeric Input widget by achieving two things:

1. Adding a fallback label for Numeric Inputs so that screen reader users are given slightly more direction for what the input is. 
2. Adding a new linter warning that warns Content Creators if they're saving a Numeric Input widget without setting the aria-label. 

Issue: LEMS-3888

## Test plan:
- Tests pass (including new ones) 
- Manual testing


packages/perseus/src/widgets/numeric-input/numeric-input.test.ts
```diff
@@ -467,6 +467,54 @@ describe("Numeric input widget", () => {
...
+                }),
+            },
+        };
+
+        // Act
+        renderQuestion(item);
+
+        // Assert
+        expect(
+            screen.getByRole("textbox", {hidden: true}),
```
nishasy, line 491: Why does the textbox have `{hidden: true}`?

packages/perseus/src/widgets/numeric-input/numeric-input.tsx
```diff
@@ -117,7 +117,7 @@ export const NumericInputComponent = forwardRef<Focusable, NumericInputProps>(
                 ref={inputRef}
                 value={props.userInput.currentValue}
                 onChange={handleChange}
-                labelText={props.labelText}
+                labelText={props.labelText || context.strings.yourAnswerLabel}
```
nishasy, line 120: Oh, it's nice that we already had this string in `strings.ts`.

nishasy (approved):
Looks good to me! Thanks for adding the warning too.

One thing that may be worth considering - I think it's not super obvious what the aria label is for in the numeric input editor. I think it could be good to rename the field "answer field aria label" or something? And maybe add an example in the associated tooltip? What do you think?

Image

nishasy:
Was this supposed to change? This doesn't look like it should be related to your changes?

| Before | After |
| --- | --- |
| image | image |
| image | image |

PR #3502, opened 2026-04-15
Replaced older underscore utility calls with Native JS

marnikostman:
## Summary:
- `_.flatten(_.invoke(...))` → `.flatMap()`                                                                                      
- `_.any()` → `.some()`                                                                                                          
- `_.reject()` → `.filter()`                                                                                                     
- `_.flatten(_.pluck(..., "terms"), true)` → `.flatMap(s => s.terms)`
- `_.invoke(this.terms, "print")` → `.map(t => t.print())`                                                                       
- `_.zip.apply(_, pairs)[1]` → `pairs.map(p => p[1])`
                                                                                                                                   
  Also hoisted `has(Func)` / `has(Unit)` out of the 12-iteration loop in
  `Expr.compare()` — these traverse the full expression tree but were being                                                        
  recomputed on every iteration.

### DIFF
--- a/.changeset/angry-carrots-clap.md
+++ b/.changeset/angry-carrots-clap.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/kas": patch
+---
+
+Patch implementation to remove underscore utilities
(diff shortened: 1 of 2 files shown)


marnikostman:
Came across these and there was guidance to remove: https://khanacademy.slack.com/archives/C01AZ9H8TTQ/p1731612438191179

marnikostman:
@claude review

PR #3503, opened 2026-04-16
Make expression comparison more thorough

benchristel:
## Summary:
Content creators reported a bug in an expression widget where the correct
answer was `-x / sqrt(1-x^2)`. You could put anything in the numerator and your
answer would (usually) be marked correct.

The bug happened because of the nondeterministic (and not very thorough) way
that we compared expressions. Previously, we selected 12 random variable
bindings and checked that the two expressions evaluated the same for all of
them. Some effort was made to cover different orders of magnitude, but simple
values like `-1`, `0`, and `1` were not always checked.

This meant that for expressions with a restricted domain, like `-x / sqrt(1 -
x^2)` (which is only defined in the interval `[-1, 1]`), we might check *only*
variable bindings for which the expression was not defined.

Expressions are now compared more thoroughly. Now we always check that the
expressions evaluate the same with all variables bound to -1, 0, and 1. We also
check more randomly-chosen values: 28 instead of 12.

## Summary:

Issue: LEMS-4042

## Test plan:

The expression in this article should be scored correctly:
https://www.khanacademy.org/devadmin/content/articles/arc-length-of-function-graphs-examples/x615e6443

### DIFF
--- a/packages/kas/src/__tests__/comparing.test.ts
+++ b/packages/kas/src/__tests__/comparing.test.ts
@@ -136,7 +136,8 @@ describe("comparing", () => {
             expect(forms2[0]).toEqualExpr(form);
         });
 
-        expect("x").toEqualExpr("xy/y");
+        expect("x").toEqualExpr("x(y^2 + 1)/(y^2 + 1)");
+        expect("x").not.toEqualExpr("x(y^2 + 1)/(y^2 + 2)");
         expect("e^x").toEqualExpr("e^x");
         expect("e^x").not.toEqualExpr("e^x + 1");
 
@@ -268,8 +269,8 @@ describe("comparing", () => {
     test("simplify can't yet handle these", () => {
         expect("sin(x + 2pi)").toEqualExpr("sin(x)");
         expect("y = sin(x + 2pi)").toEqualExpr("y = sin(x)");
-        expect("sin^2(x)+cos^2(x)").toEqualExpr("x/x");
-        expect("y = sin^2(x)+cos^2(x)").toEqualExpr("y = x/x");
+        expect("sin^2(x)+cos^2(x)").toEqualExpr("1");
+        expect("y = sin^2(x)+cos^2(x)").toEqualExpr("y = 1");
     });
 
     test("partially evaluating functions", () => {
--- a/packages/kas/src/nodes.ts
+++ b/packages/kas/src/nodes.ts
@@ -436,7 +436,6 @@ abstract class Expr {
             return false;
         }
 
-
         // To compare expressions, we evaluate them for a set of randomly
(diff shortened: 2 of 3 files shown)


packages/kas/src/__tests__/comparing.test.ts
```diff
@@ -268,8 +269,8 @@ describe("comparing", () => {
     test("simplify can't yet handle these", () => {
         expect("sin(x + 2pi)").toEqualExpr("sin(x)");
         expect("y = sin(x + 2pi)").toEqualExpr("y = sin(x)");
-        expect("sin^2(x)+cos^2(x)").toEqualExpr("x/x");
-        expect("y = sin^2(x)+cos^2(x)").toEqualExpr("y = x/x");
+        expect("sin^2(x)+cos^2(x)").toEqualExpr("1");
+        expect("y = sin^2(x)+cos^2(x)").toEqualExpr("y = 1");
```
SonicScrewdriver, line 273: This seems much clearer!

packages/kas/src/nodes.ts
```diff
@@ -437,40 +436,61 @@ abstract class Expr {
...
+            intGenerator(0, 1000),
+            intGenerator(-1000, 0),
+            floatGenerator(0, 1000),
+            floatGenerator(-1000, 0),
+        ];
+
+        const variableValueGenerators = [
+            // Always check x = -1, x = 0, and x = 1.
+            // Requested by content creators:
+            // https://khanacademy.slack.com/archives/CJDRXTGQ7/p1776180198083239?thread_ts=1776096483.309839&cid=CJDRXTGQ7
```
SonicScrewdriver, line 479: This was honestly very useful, but I wonder if we want to permanently include our slack link in our open source repo. Maybe we could add the ticket ID and a comment on the ticket with the slack link? 

This is truly just a nit / musing, so feel free to disregard at your discretion.

packages/kas/src/nodes.ts
```diff
@@ -3666,6 +3686,18 @@ export class Unit extends Sym {
...
+function constantGenerator(value: number): () => number {
+    return () => value;
+}
+
+function floatGenerator(min: number, max: number): () => number {
+    return () => randomFloat(min, max);
+}
+
+function intGenerator(min: number, max: number): () => number {
+    return () => _.random(min, max);
```
SonicScrewdriver, line 3698: I'd love it if we could clear out some of the underscore while we're in here, but that juice might not be worth the squeeze right now.

SonicScrewdriver (approved):
From a code standpoint, I think this looks like a great improvement and all the tests pass. Thank you for including such detailed comments! 

It wouldn't be a bad plan to get the Math Content team to double check that everything looks right, but I'm happy to approve myself!

PR #3504, opened 2026-04-16
[Image] | (a11y) | Add aria-describedby to Explore Image modal

nishasy:
## Summary:
The internal audit pointed out that we're missing aria-describedby on the Explore Image modal.
In order to follow best practices, I'm adding that here.

As per the suggestion in the ticket, I'm adding both caption and description IDs if the
caption is present. If there is no caption, then it's just the description ID.

NOTE: This change is not possible to test in VoiceOver since it has a behavior discrepancy -
it will be the same as before adding the aria-describedby. We should be able to see a
difference when using NVDA though. 

Update: I confirmed that it reads the describedby content before the close button when using
NVDA on my personal Windows machine!

Issue: https://khanacademy.atlassian.net/browse/LEMS-4046

## Test plan:
`pnpm jest packages/perseus/src/widgets/image/components/explore-image-modal.test.tsx`

Storybook
- Go to `/?path=/story/widgets-image-widget-demo--image`
- Open the Elements tab in dev tools
- Confirm that the `role="dialog"` element has `aria-describedby` that points to the
  caption and description IDs.

| Before | After |
| --- | --- |
| Screenshot 2026-04-16 at 2 10 32 PM | Screenshot 2026-04-16 at 2 10 23 PM |

### DIFF
--- a/packages/perseus/src/widgets/image/components/explore-image-modal-content.tsx
+++ b/packages/perseus/src/widgets/image/components/explore-image-modal-content.tsx
@@ -16,7 +16,12 @@ import type {CommonImageProps, ZoomProps, GifProps} from "./image-info-area";
 
 const MODAL_HEIGHT = 568;
 
-type Props = CommonImageProps & ZoomProps & GifProps;
+type Props = CommonImageProps &
+    ZoomProps &
+    GifProps & {
+        captionId: string;
+        longDescId: string;
+    };
 
 export default function ExploreImageModalContent({
     backgroundImage,
@@ -30,6 +35,8 @@ export default function ExploreImageModalContent({
     labels,
     range,
     zoomSize,
+    captionId,
+    longDescId,
 }: Props) {
     const [isGifPlaying, setIsGifPlaying] = React.useState(false);
     const context = React.useContext(PerseusI18nContext);
@@ -146,7 +153,10 @@ export default function ExploreImageModalContent({
                 )}
 
                 {caption && (
-                    <div className={styles.modalCaptionContainer}>
+                    <div
+                        id={captionId}
+                        className={styles.modalCaptionContainer}
+                    >
                         {/* Use Renderer so that the caption can support markdown and TeX. */}
                         <Renderer
                             content={caption}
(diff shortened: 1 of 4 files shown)


packages/perseus/src/widgets/image/components/explore-image-modal-content.tsx
```diff
@@ -165,12 +175,14 @@ export default function ExploreImageModalContent({
                     {context.strings.imageDescriptionLabel}
                 </Heading>
                 {/* Use Renderer so that the description can support markdown and TeX. */}
-                <Renderer
-                    content={longDescription}
-                    apiOptions={apiOptions}
-                    linterContext={linterContext}
-                    strings={context.strings}
-                />
+                <div id={longDescId}>
```
catandthemachines, line 178: Does Renderer not have an `id` property? That's annoying! If you have enough time it think it would be worth while to add it. But non-blocking.

catandthemachines (approved):
Looks good. Added one comment, but non blocking.

mark-fitzgerald (approved):

packages/perseus/src/widgets/image/components/explore-image-modal-content.tsx
jeremywiebe: I think this might be the first time I've seen a Renderer linked to an `aria-describedby`. Although we could drill the id down to the `<div>` that the Renderer renders, I'm not sure it feels warranted for only one use. Perhaps we could wait and see if this comes up again?

nishasy: Hypothetically, one could argue this is another example of a Renderer wrapped in something with an ID: https://github.com/Khan/perseus/blob/main/packages/perseus/src/widgets/numeric-input/input-with-examples.tsx#L143-L154

It looks like it also needs that wrapper for the classname, so I'm not sure if that's a valid example for this. I'm inclined to leave the code as is for now, since I don't currenly have a way to test that any changes here would be valid (I'd need a windows machine with NVDA).

PR #3505, opened 2026-04-16
[Image] | (CX) | Add a warning for large images

nishasy:
## Summary:
If an image is too large, it can cause slowdowns in both the editor and for our learners.

Add a warning here to let the content authors know if that's the case.

Issue: https://khanacademy.atlassian.net/browse/LEMS-4050

## Test plan:

Storybook
- Go to `/?path=/story/editors-editorpage--with-all-flags`
- Add an image widget
- Use a large image (example: `https://cdn.kastatic.org/ka-perseus-images/01f44d5b73290da6bec97c75a5316fb05ab61f12.jpg`)
- Confirm that a warning shows up about the image being too large
- Use a small image (example: `https://cdn.kastatic.org/ka-content-images/61831c1329dbc32036d7dd0d03e06e7e2c622718.jpg`)
- Confirm that the warning does not show up

| Before | After |
| --- | --- |
| Screenshot 2026-04-16 at 2 37 47 PM | Screenshot 2026-04-16 at 2 37 52 PM |

### DIFF
--- a/.changeset/itchy-students-repair.md
+++ b/.changeset/itchy-students-repair.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": patch
+---
+
+[Image] | (CX) | Add warning for large images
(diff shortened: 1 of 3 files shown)


catandthemachines (approved):
Implementation looks good to me. One thing to note is that this feature is behind the scaling update. As long as that is expected I see no issues.

PR #3506, opened 2026-04-16
[Image] | (DX) | Recheck height in Graphie to stop font position flakiness

nishasy:
## Summary:
The `ZoomClickedWithGraphieImage` regression story keeps being reported by Chromatic as
having a ~2px difference from base.

In this PR, we're rechecking the height after the line height is corrected to make sure we're using the appropriate height for the Graphie labels. 

Explanation from Claude:
> Summary of the fix (graphie.ts:1767-1772): when the line-height correction inside      
  setLabelMargins updates height, it now also writes the corrected value back into
  originalLabelSize. This means the ResizeObserver-driven _recalculateLabels path will   
  use the same corrected height on every subsequent call — eliminating the
  non-deterministic 2.5 px flip between -16px and -18.5px.

Issue: https://khanacademy.atlassian.net/browse/LEMS-4053

## Test plan:
Wait and see if anything gets reported in this PR. Continue monitoring after.


mark-fitzgerald (approved):
Fingers crossed.

nishasy:
Nope 😭 Back to the drawing board

image

nishasy:
Trying something new here. This is what Claude came up with after I started a new session and mentioned that the explore image modal story wasn't flakey like the zoom view one was.

> ★ Insight ─────────────────────────────────────                                        
  Why only ZoomClickedWithGraphieImage and not                                         
  LongDescriptionClickedStateWithGraphieImage? In the zoom modal the image fills the full
   dialog width — so as the WonderBlocks FlexibleDialog animates open, the graphie     
  container changes size, triggering the ResizeObserver. In the long description modal,  
  the image sits in a fixed two-column panel; its container width is already stable by
  the time MathJax finishes, so _recalculateLabels never fires post-correction and the
  stale originalLabelSize is never read.
  ─────────────────────────────────────────────────

 > Summary of the fix (graphie.ts:1767-1772): when the line-height correction inside      
  setLabelMargins updates height, it now also writes the corrected value back into
  originalLabelSize. This means the ResizeObserver-driven _recalculateLabels path will   
  use the same corrected height on every subsequent call — eliminating the
  non-deterministic 2.5 px flip between -16px and -18.5px.
  
  Edit: That didn't work

packages/perseus/src/util/graphie.ts
```diff
@@ -1760,12 +1757,17 @@ const setLabelMargins = function (span: HTMLElement, size: Coord): void {
         // Inherited line-height values can really mess up placement.
         $container.css("line-height", "normal");
 
-        // If the change in line-height affected the height of the element,
-        //     then the height used for calculations should be updated.
-        // This can happen when the first label in the container calls this method,
-        //     and the line-height was different when the height measurement was originally referenced.
-        if (currentHeightMatchesProps && span.scrollHeight !== height) {
```
mark-fitzgerald, line 1767: I'm having trouble seeing how this situation (line-height affecting the height) is handled by the new conditional involving just the "originnalLabelSize" data attribute, below. Can you verify the timing of this?

nishasy: It looks like in the main code, the height gets set to the scroll height only if a change in the scroll height is detected. In the new code, the height gets set to the scroll height always on the first run. The timing on the two should be the same.

mark-fitzgerald (approved):
Graphie is always tricky. Thanks for diving into these anomalies.

PR #3508, opened 2026-04-20
[Image] | (CX) | Give explore modal launcher a label saying it has description

nishasy:
## Summary:
Right now, the label for the explore image modal launcher button is "Explore image,"
but that doesn't make it clear to screen reader users that the explore image modal
will also contain the long description for the image.

To make this more clear, use the aria label "Explore image and description"

Issue: https://khanacademy.atlassian.net/browse/LEMS-4045

## Test plan:
`pnpm jest packages/perseus/src/widgets/image/image.test.ts`

Storybook
- Open `/?path=/story/widgets-image-widget-demo--image`
- Turn on a screen reader
- Navigate to the explore image launcher button
- Confirm it says "Explore image and description"

| Before | After |
| --- | --- |
| Screenshot 2026-04-20 at 4 03 41 PM | Screenshot 2026-04-20 at 3 51 10 PM |

### DIFF
--- a/packages/perseus/src/strings.ts
+++ b/packages/perseus/src/strings.ts
@@ -597,6 +597,7 @@ export type PerseusStrings = {
         point2Y: string;
     }) => string;
     imageExploreButton: string;
+    imageExploreButtonAriaLabel: string;
     imageAlternativeTitle: string;
     imageDescriptionLabel: string;
     imageZoomAriaLabel: string;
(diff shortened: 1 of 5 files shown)


packages/perseus/src/strings.ts
```diff
@@ -1320,6 +1321,7 @@ export const strings = {
             "Tangent graph with inflection point at %(point1X)s comma %(point1Y)s and control point at %(point2X)s comma %(point2Y)s.",
     },
     imageExploreButton: "Explore image",
+    imageExploreButtonAriaLabel: "Explore image and description",
     imageAlternativeTitle: "Explore image and description",
```
nishasy, line 1325: I know there's an identical string to the one I'm adding, but they're for totally different things, so I thought it would be better to add a separate string to make it less confusing in context.

catandthemachines (approved):
Looks great. Thanks Nisha!

PR #3509, opened 2026-04-21
Add docs folder to .eslintignore

benchristel:
## Summary:
`pnpm lint` failed because of linter errors in generated code in `docs/`. The
fix is to tell eslint to ignore the docs folder.

Issue: none

## Test plan:

`pnpm lint` should show no errors.


SonicScrewdriver (approved):
Makes sense to me! :)

PR #3510, opened 2026-04-21
Add fixc, linc, and tesc scripts to package.json, mirroring the frontend repo

benchristel:
Issue: none

## Test plan:

- Introduce a lint error
- Run `pnpm fixc`
- The lint error should be gone in a few seconds.


SonicScrewdriver (approved):
Looks great to me! Seems like you just need a changeset and maybe to exclude changed-files.sh from knip, and then you're good to go :) 

Thanks for getting this synced up!

PR #3512, opened 2026-04-21
Allow requests from the Claude sandbox to `cdn.jsdelivr.net`

benchristel:
Without this configured, the following test always fails when run by Claude:

```
packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/util.test.ts
```

because the Speech Rule Engine used to render accessible math descriptions
cannot fetch its configuration file.

See: https://khanacademy.slack.com/archives/C01AZ9H8TTQ/p1775760948570359?thread_ts=1775760655.549589&cid=C01AZ9H8TTQ

Issue: none

## Test plan:

Tell Claude to run:

```
pnpm test packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/util.test.ts
```

The test should pass.


.claude/settings.json
```diff
@@ -78,7 +78,8 @@
                 "registry.npmjs.org",
                 "npm.pkg.github.com",
                 "download.cypress.io",
-                "cdn.cypress.io"
+                "cdn.cypress.io",
+                "cdn.jsdelivr.net"
```
jeresig, line 39: In the future you're going to need to make this change in `~/khan/agent-settings` so that it doesn't get overwritten the next time this goes out. I've just handled this for you so it should be fine going forward: https://github.com/Khan/agent-settings/commit/5c9fddc11a1e4d76bef1a978b97e0648e02a34ad

jeresig (approved):

jeresig:
Looks good! Left a comment for future changes here.

benchristel:
We switched to a different approach: [using hardcoded data for the speech rule engine in tests, so it doesn't need to make a network request](https://github.com/Khan/perseus/pull/3513).

Closing this PR.

PR #3513, opened 2026-04-21
Use local data for the MathJax Speech Rule Engine in tests

benchristel:
## Summary:
MathJax's Speech Rule Engine generates accessible descriptions of math.
To localize the descriptions, it is configured via JSON files for each locale.
By default, it fetches these JSON files from `cdn.jsdelivr.net`. These network
requests meant that:

- You couldn't run the Perseus test suite without an internet connection.
- Claude couldn't run the test suite in the sandbox, because there is a [bug]
  which makes the `allowedDomains` list not affect NodeJS processes.

The fix implemented in this PR is to use local JSON files so the Speech Rule
Engine doesn't need to make any network requests.

This also means we can remove some of the convoluted mocking of the SRE that we
were previously doing.

[bug]: https://github.com/anthropics/claude-code/issues/37782

Issue: LEMS-2616

## Test plan:

- Tests should pass with no internet connection.
- Claude should be able to run `pnpm test` with no failures.


config/test/mathjax-sre/base.json
benchristel, line 1: Copied from https://cdn.jsdelivr.net/npm/speech-rule-engine@4.0.7/lib/mathmaps/base.json

config/test/mathjax-sre/en.json
benchristel, line 1: Copied from https://cdn.jsdelivr.net/npm/speech-rule-engine@4.0.7/lib/mathmaps/en.json

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-ellipse-settings.test.tsx
```diff
@@ -21,15 +17,6 @@ const defaultProps = {
...
 const defaultLabel = getDefaultFigureForType("label");
 
-// Mock the async functions
-jest.mock("./util", () => ({
-    ...jest.requireActual("./util"),
-    generateSpokenMathDetails: (input) =>
-        mockedGenerateSpokenMathDetailsForTests(input),
-    joinLabelsAsSpokenMath: (input) =>
-        mockedJoinLabelsAsSpokenMathForTests(input),
-}));
```
nishasy, line 31: Yayy

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-ellipse-settings.test.tsx
```diff
@@ -409,10 +396,12 @@ describe("LockedEllipseSettings", () => {
...
 
             // Assert
-            expect(onChangeProps).toHaveBeenCalledWith({
-                ariaLabel:
-                    "Circle with radius 2, centered at spoken $0$ comma spoken $0$. Appearance solid gray border, with no fill.",
-            });
+            await waitFor(() =>
+                expect(onChangeProps).toHaveBeenCalledWith({
+                    ariaLabel:
+                        "Circle with radius 2, centered at 0 comma 0. Appearance solid gray border, with no fill.",
```
nishasy, line 402: Yayyyy

nishasy (approved):
This looks awesome!

One question: Is there a way to confirm/double check that the tests are using the global mock and not making real network calls?

benchristel:
> One question: Is there a way to confirm/double check that the tests are using the global mock and not making real network calls?

I checked this by turning off my wifi and running the tests. They all pass!

PR #3514, opened 2026-04-22
docs(changeset): Allowing all appropriate graph types to have points placed on the graph edge.

SonicScrewdriver:
## Summary:
Our content team would love the ability to place points on the edge of the graph, particularly for creating graphs for physics content where a point needs to be at 0 and the visibility of a -1 does not make sense in that context. Currently, this ability is only allowed for `point` graphs due to it causing rendering issues with other graph types.

This PR enables the placement of points on the graph edge for all other graphs type _except_ `Linear`, `Linear System`, and `Angle` due to issues related to the point covering up the arrows, and the fact that there's little value in allowing edge-placement for these types. 

Essentially, I've shuffled up the "Mafs Sandwich" a little by creating a new `GraphBoundsSvg` component that we can reuse wherever necessary. This also allowed me to clean up the logic in mafs-graph a little. 

**Note 1:** Another side effect from this work is that I've added bounding to the protractor to ensure that it cannot get lost "off screen". While this was technically a pre-existing bug, it seemed related enough to what I was doing that I fixed it up here. 

**Note 2:** While this started as a proof of concept, it ended up working far better than I thought it would, so I'm opening up the PR for review.

**Issue:** LEMS-4056

## Test plan:
- Extensive Manual testing
- Sharing of Prototype 
- Tests pass


packages/perseus/src/widgets/interactive-graphs/__docs__/interactive-graph.stories.tsx
```diff
@@ -130,6 +131,12 @@ export const Tangent: Story = {
     },
 };
 
+export const AbsoluteValue: Story = {
```
SonicScrewdriver, line 143: We were missing a story

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -302,6 +325,27 @@ function doMovePointInFigure(
                 coords: newCoords,
             };
         }
+        case "ray": {
```
SonicScrewdriver: I needed to split `ray` case apart, as we only want to allow the "tail" point on the graph edge.

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -801,6 +848,11 @@ function doMoveCenter(
                 const possibleNewX = radX - xJumpDist;
                 if (possibleNewX >= xMin && possibleNewX <= xMax) {
                     newRadiusPoint[X] = possibleNewX;
+                } else {
```
SonicScrewdriver, line 770: I'm open to discussion on this. There's a specific glitch state you can get into, which requires the following steps: 

1. Use the point to make the Circle to fill the graph
2. Drag the Circle's drag handle to move the circle over to the left or right
3. Move the Circle "size" point to the graph edge again 
4. Attempt to drag the Circle's drag handle back to the center of the graph 

Essentially, this was resulting in the point extending past the graph bounds — aka just floating in white space. There were two ways to resolve this: either resizing the circle when in this state, or blocking a drag handle movement that forces the point off the graph. I opted for the latter, but I'm happy to adjust according to feedback!

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -399,16 +399,15 @@ export const MafsGraph = (props: MafsGraphProps) => {
...
                                     {...nestedSVGAttributes}
                                     style={{
-                                        // We want to allow points to be directly
-                                        // on the edge of the graph, so we need to
-                                        // set overflow to visible. For other
-                                        // graphs, we want to hide the overflow
-                                        // so that the graph can't go way off
-                                        // the edge.
-                                        overflow:
-                                            type === "point"
```
SonicScrewdriver, line 428: ~~Instead of extending this check, I've opted to just show the overflow for all graphs. I've tested each graph type and this does not appear to have created any regressions as I have added specific solutions for each of the affected graph types. `Linear`, `Linear System`, and `Angle` were all already bound to the graph — so they worked "out of the box".~~

Actually, I've updated this so that we're ONLY nesting the protractor. :) After running through all of our test stories, I was able to confirm that we didn't need the wrapper around the graph anymore. However, we do need to make sure the protractor is bound to the graph limits still.

packages/perseus/src/widgets/interactive-graphs/graphs/components/graph-bounds-svg.tsx
```diff
@@ -0,0 +1,33 @@
+import * as React from "react";
+
+import useGraphConfig from "../../reducer/use-graph-config";
+import {calculateNestedSVGCoords} from "../../utils";
+
+// A nested <svg> sized to the graph's visible bounds. Its default
```
SonicScrewdriver: This is exactly the same thing we are doing with the Mafs Sandwich, just at the graph type level. :)

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -120,23 +117,6 @@ export const MafsGraph = (props: MafsGraphProps) => {
     const graphRef = React.useRef<HTMLElement>(null);
     const {analytics} = useDependencies();
 
-    // Set up the SVG attributes for the nested SVGs that help lock
```
SonicScrewdriver, line 128: It's kind of nice to get to clean all this up a bit!

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
nishasy: Oh nice, this is the section of the code I was referring to when I mentioned "Point-specific logic"

nishasy (commented):
This seems to be working perfectly in storybook! How exciting 😄

packages/perseus/src/widgets/interactive-graphs/graphs/absolute-value.tsx
```diff
@@ -69,13 +70,15 @@ function AbsoluteValueGraph(props: AbsoluteValueGraphProps) {
...
     return (
         <g aria-label={srAbsoluteValueGraph} aria-describedby={descriptionId}>
-            <Plot.OfX
-                y={(x) => m * Math.abs(x - h) + v}
-                color={interactiveColor}
-                svgPathProps={{
-                    "aria-hidden": true,
-                }}
-            />
+            <GraphBoundsSvg>
```
handeyeco: Maybe this is going to be answered later, but how do future humans/AIs know what to put inside GraphBoundsSvg vs outside it? Would a different name for the wrapper make it easier to understand what's meant to go in it?

packages/perseus/src/widgets/interactive-graphs/graphs/exponential.test.tsx
```diff
@@ -280,16 +277,15 @@ describe("getExponentialKeyboardConstraint", () => {
             range,
         );
 
-        // Assert — no valid right move, falls back to original position
-        expect(constraint.right).toEqual([8, 3]);
+        // Assert — walks past x=9 and lands at x=10 (graph edge)
+        expect(constraint.right).toEqual([10, 3]);
```
handeyeco, line 332: This "fixes" the test in the sense that it's passing now, but it kind of feels like it's changing the intent of the test. This is taking a test that makes sure the point doesn't move to an invalid point by skipping and turns it into a test that makes sure we can move to the edge. I think both are valid but we're losing the former to create the later, right?

So I would just generally double check that AI isn't cheating by changing the intent of tests to make them pass.

packages/perseus/src/widgets/interactive-graphs/graphs/exponential.tsx
```diff
@@ -180,14 +183,11 @@ export const getExponentialKeyboardConstraint = (
...
-            // The reducer clamps the destination via boundAndSnapToGrid
+            // The reducer clamps the destination via boundToEdgeAndSnapToGrid
             // before applying its own collision checks. We must predict
             // the clamped position to avoid accepting coords that the
             // reducer will silently reject.
-            const clamped = snap(
-                snapStep,
-                bound({snapStep, range, point: coord}),
-            );
+            const clamped = snap(snapStep, boundToEdge({range, point: coord}));
```
handeyeco: Shouldn't bounding still have access to snap step?

For instance if a grid is 7x7, we start at 0,0, and we have a snap of 2: we actually don't want it to get to the edge, right?

packages/perseus/src/widgets/interactive-graphs/graphs/logarithm.test.tsx
```diff
@@ -280,15 +280,15 @@ describe("getLogarithmKeyboardConstraint", () => {
...
 
-    it("stays put when all rightward positions would cause a clamped collision", () => {
-        // Arrange — asymptote at x=8, points at [7,3] and [4,1].
-        // Moving point 0 right: x=8 is the asymptote, x=9 crosses the
-        // asymptote so reflectedX = 2*8-4 = 12 which clamps to 9
-        // (same as clamped coord), and x>=10 all clamp to 9 as well.
-        // No valid rightward position exists, so the point stays put.
+    it("stays put when all rightward positions would cause a y-collision", () => {
+        // Arrange — asymptote at x=-5, points at [5,3] and [6,3].
+        // Moving point 0 right: x=6 matches otherPoint.x (reject).
```
handeyeco: Not you, but I don't have the faintest idea what this comment means 😅

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
```diff
@@ -344,8 +324,8 @@ export const MafsGraph = (props: MafsGraphProps) => {
                             >
                                 {/* Svg definitions to render only once */}
                                 <SvgDefs />
-                                {/* Cartesian grid nested in an SVG to lock to graph bounds */}
-                                <svg {...nestedSVGAttributes}>
+                                {/* Cartesian grid clipped to graph bounds */}
+                                <GraphBoundsSvg>
```
handeyeco: So this will get rendered twice? Once for MafsGraph and once for the inner graph (ex AbsoluteValue)?

handeyeco (commented):
I have to go to an appointment so I didn't have time to do a full review.

My main concerns are:

1. It feels like the meaning of tests shifted in a way that makes me think an LLM was trying to skirt around failed tests, but I could be wrong.
2. Some of this is just hard to grok. The comments before your PR are too technical IMO and I'm not sure how to properly use `GraphBoundsSvg`.

packages/perseus/src/widgets/interactive-graphs/graphs/absolute-value.tsx
SonicScrewdriver: Great question! I'm absolutely down to improve the name, I just wasn't sure of a good one myself before. I think `ClipToGraphBounds` might be clearer? I can change it to that. :)

packages/perseus/src/widgets/interactive-graphs/mafs-graph.tsx
SonicScrewdriver: Sort of! I'm basically just replacing the existing usages of the "Mafs Sandwich" svg with our new component here. The old `svg` with the `nestedSVGAttributes` was doing the same thing that my new `GraphBoundsSVG` component does. 

The main difference in Mafs Graph is that the "Interactive Layer" (line 385) is no longer nested within the clipped SVG at this level. We handle that at the graph level so that we can allow the points to render outside the graph bounds, while keeping the actual graph _lines_ confined. Otherwise half the points were cut off if they were placed at the graphs edge. :)

packages/perseus/src/widgets/interactive-graphs/graphs/components/clip-to-graph-bounds.tsx
```diff
@@ -0,0 +1,37 @@
...
+// clipping). Wrap any rendering that could extend past the graph if
+// left unclipped — continuous curves (Plot.OfX), shapes that can grow
+// beyond their control points (the circle's ellipse), the protractor
+// image, locked figures, etc. Don't wrap interactive handles that are
+// expected to remain visible when sitting on the graph boundary.
+//
+// React context from Mafs and the --mafs-view-transform CSS variable
+// both survive the nested SVG, so Mafs primitives continue to render
+// correctly inside it.
+export function ClipToGraphBounds({children}: {children: React.ReactNode}) {
```
ivyolamit, line 17: Do you have a before and after screenshot?

ivyolamit: And can we add a specific story or visual regression test for this

packages/perseus/src/widgets/interactive-graphs/graphs/logarithm.test.tsx
SonicScrewdriver: I tried to make them a little clearer for us 😅

ivyolamit (commented):
@SonicScrewdriver I did a first pass on this PR, it looks logical to me. I'll give it a second pass later this afternoon and smoke test.

packages/perseus/src/widgets/interactive-graphs/graphs/components/clip-to-graph-bounds.tsx
SonicScrewdriver: I can add a story that has the points on the edge, but that'd really be the only new thing to test, I think. Our existing regression tests / chromatic already cover everything else

ivyolamit (approved):
LGTM, let's have this go through QA since it will affect existing behavior live in prod.

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -315,36 +318,13 @@ function doMovePointInFigure(
...
-            // point (index 1) stays one snap step inside so its handle
-            // doesn't visually cover the arrow glyph, which always
-            // renders just inside the edge.
-            const bounded =
-                action.pointIndex === 0
-                    ? boundToEdgeAndSnapToGrid(action.destination, state)
-                    : boundAndSnapToGrid(action.destination, state);
+        case "linear":
+        case "ray":
+        case "vector": {
```
SonicScrewdriver, line 323: After these changes, I was able to merge all of these switch cases again

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -1448,15 +1406,22 @@ export function calculateSideSnap(
...
-    const topRight = bound({snapStep, range, point: [Infinity, Infinity]});
+    const topRight = boundToEdgeAndSnapToGrid([Infinity, Infinity], {
+        snapStep,
+        range,
+    });
     return vec.sub(topRight, point);
 }
 
-// Returns the vector from the given point to the bottom-left corner of the graph when snapped to the grid
+// Mirror of maxMove for the bottom-left corner.
```
SonicScrewdriver: Updated this comment to:

```
// Returns the vector from the given point to the snap-aligned bottom-left
// corner of the graph.
```

In the next commit, as this note was temporary and snuck through.

packages/perseus/src/widgets/interactive-graphs/utils.ts
```diff
@@ -82,13 +82,21 @@ export function bound({
...
 export function boundToEdgeAndSnapToGrid(
     point: vec.Vector2,
     {snapStep, range}: {snapStep: vec.Vector2; range: [Interval, Interval]},
 ): vec.Vector2 {
-    const snapped = snap(snapStep, point);
-    const [[xMin, xMax], [yMin, yMax]] = range;
     const [snapX, snapY] = snapStep;
+    const [[xMin, xMax], [yMin, yMax]] = range;
+
+    if (snapX === 0 && snapY === 0) {
```
SonicScrewdriver, line 95: This minor early return addition is to support an edge case regarding polygon's snapping logic attempting to divide by zero.

PR #3515, opened 2026-04-22
Make `pnpm fixc` filter out unrecognized file types

benchristel:
## Summary:
Previously, the `pnpm linc` and `pnpm fixc` scripts would fail if non-JS files
had changed.  `.eslintignore` filtered out some of these files, but AFAICT
there's no way to exclude files with no extension. Trying to ignorelist all the
files we don't want seemed like a futile game of whack-a-mole.

This PR introduces utils/fix.sh and utils/lint.sh that fix/lint files provided
on stdin, ignoring any filetypes we don't recognize.

Issue: none

## Test plan:

- Mess up the formatting in a JS file. `pnpm fixc` should fix it.
- `echo hi > foo`. `pnpm fixc` should ignore it.


utils/lint.sh
```diff
@@ -0,0 +1,18 @@
...
+# ensures that xargs won't misinterpret filenames with spaces as multiple
+# files.
+echo "$files" \
+    | grep -E "\.(js|jsx|cjs|mjs|ts|tsx|cts|mts)$" \
+    | tr '\n' '\0' \
+    | xargs -r -0 pnpm eslint
+
+echo "$files" \
+    | tr '\n' '\0' \
+    | xargs -r -0 pnpm prettier --ignore-unknown --check
```
benchristel: The `fix.sh` and `lint.sh` scripts are awfully similar... maybe they should be combined.

utils/changed-files.sh
```diff
@@ -1,9 +1,8 @@
 #!/bin/sh
 set -e
 
-# Prints a null-separated list of files changed since the commit referred to by
-# $UPSTREAM, or since the current branch's upstream if $UPSTREAM is empty or
-# not set. Untracked files are included.
+# Prints a newline-separated list of files changed since the commit referred
+# to by $UPSTREAM, which defaults to the current branch's upstream.
```
benchristel, line 5: I changed `changed-files.sh` to output newline-separated filenames instead of null-separated filenames for the benefit of `grep`. In `lint.sh`, we do:

```
grep -E "\.(js|jsx|cjs|mjs|ts|tsx|cts|mts)$" \
```

However, grep can take null-separated data too: `grep -z`. So maybe I should revert the changes to this file. Then I could remove a bunch of `tr` calls.

The counterargument is that newlines are more interoperable.

SonicScrewdriver (approved):
I think this looks good!

SonicScrewdriver (changes requested):
I may have found a minor issue here! I tried creating an extensionless file and found that running `pnpm fixc` just exits with code 1, instead of silently ignoring the file.

I think it's due to the combination of `grep` exiting with an error code when it finds zero matches and `set -eo pipefail`. As a result, `fixc` or `linc` will fail if _none_ of the changed files match the filter.  We might just want to add a fallback to the `grep` command.

benchristel:
> I may have found a minor issue here! I tried creating an extensionless file and found that running `pnpm fixc` just exits with code 1, instead of silently ignoring the file.
> 
> I think it's due to the combination of `grep` exiting with an error code when it finds zero matches and `set -eo pipefail`. As a result, `fixc` or `linc` will fail if _none_ of the changed files match the filter. We might just want to add a fallback to the `grep` command.

Thank you! Fixed.

SonicScrewdriver (approved):

PR #3517, opened 2026-04-22
Check hints for inaccessible widgets in `isItemAccessible`

benchristel:
This fixes some TODO comments.

I removed `violatingWidgets` since it was deprecated and not used in frontend.

Issue: none

## Test plan:

CI checks should pass.


catandthemachines (approved):
Looks great! Thanks for cleaning this up @benchristel !

PR #3519, opened 2026-04-23
Silence console output when running pnpm tesc

benchristel:
## Summary:
LLMs seem to get overwhelmed by all the `console.error` noise when they run the
tests.  To mitigate this issue, this PR silences the output in `pnpm tesc`.

Issue: LEMS-XXXX

## Test plan:

- Run `pnpm tesc`. If no tests run (e.g. you are on main) set `UPSTREAM=<commit>`
- You should see a `PASS` line as usual for each test suite, and the summary
  should be printed, but there should be no React warnings etc.


SonicScrewdriver (approved):
Love it! I honestly prefer it this way as well.

PR #3520, opened 2026-04-23
Increase the timeout for waitFor calls in locked-vector-settings tests

benchristel:
## Summary:
These tests appear to be flaky now that we no longer mock the Speech Rule
Engine calls. The default timeout for `waitFor` is one second, which is too
short to be reliable.

We might have to go back to mocking, but this PR is an attempt to salvage the
no-mock approach.

Issue: LEMS-2616

## Test plan:

CI should pass.


SonicScrewdriver (approved):
Oooo thank you, this was failing on my PRs. The need for timeouts does seem to make me think we might need to think of a better long term solution here, but I also don't want these tests blocking anyone. 

I'll make a note to discuss a longer term solution with the interactive graphs team on Monday

PR #3521, opened 2026-04-24
docs(changeset): Allow the asymptote to be placed between points. Also allow for points to be on asymptote line so long as they don't overlap with the drag handle.

SonicScrewdriver:
## Summary:
The old Grapher Tangent / Exponential graphs used to allow users to place the asymptote in-between the points (an “invalid” location). When we implemented these graphs in Interactive Graph, the team actually thought that this was a “glitch state” that needed to be resolved. 

This PR creates the logic to mimic the functionality from Grapher in order to allow users to place points between or on the asymptote line so long as it does not overlap with the drag handle. 

Issue: LEMS-4055

## Test plan:
- tests pass

### DIFF
--- a/.changeset/light-teachers-compare.md
+++ b/.changeset/light-teachers-compare.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Allow the asymptote to be placed between points. Also allow for points to be on asymptote line so long as they don't overlap with the drag handle.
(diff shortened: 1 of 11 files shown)


ivyolamit (approved):
Changes looks logical to me and smoke testing 👌🏼 
Nonblocking ask, can you update the md file to reflect the new expected behavior of the asymptote in https://github.com/Khan/perseus/blob/main/packages/perseus/src/widgets/interactive-graphs/__docs__/notes/logarithm.md

Currently the logarithm.md file was updated so that it becomes the technical specification instead of just a research notes. This is something I brought up in [slack](https://khanacademy.slack.com/archives/C0AHH9H21H7/p1776116147304789?thread_ts=1773856464.073039&cid=C0AHH9H21H7) but didn't get any response from the team yet. I'll leave it up to you if you're going to update the exponential notes or not, but at least update the logarithm technical notes. 🙏🏼

PR #3522, opened 2026-04-24
Improve labels for Absolute Value Graph Editor points

SonicScrewdriver:
## Summary:
This PR creates the StartCoordsAbsoluteValue component to help differentiate between the Vertex and the Arm for our Content Creators in the Interactive Graph Editor. 

Issue: LEMS-4019

## Test plan:
- Tests pass 
- Manual testing

### DIFF
--- a/packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-absolute-value.tsx
+++ b/packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-absolute-value.tsx
@@ -1,0 +1,62 @@
+import {View} from "@khanacademy/wonder-blocks-core";
+import {Strut} from "@khanacademy/wonder-blocks-layout";
+import {semanticColor, spacing} from "@khanacademy/wonder-blocks-tokens";
+import {BodyText} from "@khanacademy/wonder-blocks-typography";
+import {StyleSheet} from "aphrodite";
+import * as React from "react";
+
+import CoordinatePairInput from "../../../components/coordinate-pair-input";
+
+import type {Coord} from "@khanacademy/perseus";
+
+type AbsoluteValueCoords = [Coord, Coord];
+
+type Props = {
+    startCoords: AbsoluteValueCoords;
+    onChange: (startCoords: AbsoluteValueCoords) => void;
+};
+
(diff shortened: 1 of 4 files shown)


packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-absolute-value.tsx
```diff
@@ -0,0 +1,62 @@
...
+                    coord={arm}
+                    labels={["x", "y"]}
+                    onChange={(value) => onChange([vertex, value])}
+                />
+            </View>
+        </>
+    );
+};
+
+const styles = StyleSheet.create({
```
handeyeco, line 51: This is also declared in:
- start-coords-angle
- start-coords-line
- start-coords-point
- start-coords-quadratic
- start-coords-sinuoid

They all define something like:

``` JSX
<View style={styles.tile}>
    <BodyText size="medium" weight="bold" tag="span">
        Point 3:
    </BodyText>
    <Strut size={spacing.small_12} />
    <CoordinatePairInput
        coord={startCoords[2]}
        labels={["x", "y"]}
        onChange={(value) =>
            onChange([startCoords[0], startCoords[1], value])
        }
    />
</View>
```

I wonder if it would be worth consolidating these into one abstraction? Could even just be `Tile`:

``` JS
function Tile({ children }) {
    return (
         <View style={styles.tile}>
            { children }
        </View>
    )
}

    tile: {
        backgroundColor: semanticColor.core.background.instructive.subtle,
        marginTop: spacing.xSmall_8,
        padding: spacing.small_12,
        borderRadius: spacing.xSmall_8,
        flexDirection: "row",
        alignItems: "center",
    },
});
```

But maybe it's worth having it all?

Not a blocker.

handeyeco (approved):

packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-absolute-value.tsx
SonicScrewdriver: That's a great point! I think it makes sense to make those changes in a separate cleanup PR, but I'll make a note to get a ticket for it

PR #3524, opened 2026-04-24
docs(changeset): Ensure that Ray and Linear points cannot overlap

SonicScrewdriver:
## Summary:
This PR simply ensures that the points in the Ray or Linear graphs cannot overlap. 

Note: This will have a conflict with another PR I have, as I had to split apart the reducer logic for Ray and Linear. I'll resolve that manually when I land these tickets. 

Issue: LEMS-4020

## Test plan:
- tests pass 
- manual testing


ivyolamit (approved):
changes looks logical to me 👍🏼

PR #3525, opened 2026-04-28
[Color Sync] Convert color values to semantic tokens in Matrix

nishasy:
## Summary:

- Updated matrix.tsx with the two semantic token replacements (#f3f3f3 → neutral.subtle, #fff → base.default) + updated snapshots.
- Added Chromatic visual regression stories (matrix-initial-state-regression.stories.tsx and matrix-interactions-regression.stories.tsx) to establish a baseline before conversion and capture diffs after
- Added generator for Matrix to use within the stories decorator.
- Removed the "Enter the Matrix" Easter egg.

Issue: https://khanacademy.atlassian.net/browse/LEMS-3499

## Test plan:
 * Check Chromatic diffs for color changes.
 * Verify storybook play function: all interaction stories complete without errors, interactions tab shows green checkmarks
 * Visual states confirmed in Storybook
 * `pnpm jest` shows no errors, including snapshot tests


packages/perseus-core/src/utils/generators/matrix-widget-generator.ts
nishasy, line 1: Added this to be consistent with other widgets. Matrix was missing a generator function.

packages/perseus/src/widgets/matrix/matrix.tsx
```diff
@@ -442,9 +427,9 @@ class Matrix extends React.Component<Props, State> implements Widget {
                                             // Ensure that any borders are included in
                                             // the provided width.
                                             boxSizing: "border-box",
-                                            backgroundColor: outside
-                                                ? "#f3f3f3"
-                                                : "#fff",
+                                            borderWidth: outside
+                                                ? border.width.thin
+                                                : border.width.medium,
```
Myranae, line 441: I might have missed something, but why is the background color being replaced with borderWidth here and below? Is this related to removing the Easter egg?

Myranae: Here's a better comment that Claude helped me with:

> In matrix.tsx, the two inline backgroundColor properties were replaced with borderWidth rather than a background color token. At the same time, the .outside CSS rule was removed from matrix.css without a replacement. The DefaultEmptyMatrix story describes the expected result as "only cell (0,0) is white; all other cells are gray" — but I don't see what produces that gray now.
> Was the borderWidth change intentional, or is this a mistake? If the plan is for CSS to handle the inside/outside distinction instead of inline styles, the .outside CSS rule needs a token-based replacement. The stories and implementation should describe the same behavior.

packages/perseus/src/styles/widgets/matrix.css
```diff
@@ -31,57 +31,30 @@
...
-}
 .perseus-matrix input:focus,
 .perseus-matrix .number-input:focus {
     border: none;
     outline: none;
 }
-.static-mode.perseus-matrix input,
-.static-mode.perseus-matrix .number-input {
-    background: #f5f5f5;
-}
```
Myranae, line 66: Claude pointed out that static is still in use, so this might to be re-evaluated, unless we are removing static. Here's a better worded comment:

> The .static-mode.perseus-matrix input CSS rule was removed, but in matrix.tsx the static-mode class is still applied when this.props.static is true. Is static mode being intentionally removed? If not, this needs a token replacement alongside the other background color conversions.

packages/perseus/src/widgets/matrix/__docs__/matrix-initial-state-regression.stories.tsx
```diff
@@ -0,0 +1,99 @@
...
+        chromatic: {disableSnapshot: false, modes: themeModes},
+    },
+};
+export default meta;
+
+type Story = StoryObj<typeof MatrixWidget>;
+
+// Verifies the default empty state: only cell (0,0) is white (#fff); all
+// other cells in the 3×3 board are gray (#f3f3f3) because they fall outside
+// the current matrix size.
```
Myranae: We may want to reference the semantic color tokens now that we are phasing out hex and base color token uses.

packages/perseus/src/widgets/matrix/matrix.tsx
nishasy: Here is the new [spec for Matrix on Figma](https://www.figma.com/design/dpEbtoeF5obxMN4VDifUna/Project-Color-Sync?node-id=5-175&p=f&t=lUHYn4wB180BQnPd-0)!

and the [specific values from Slack](https://khanacademy.slack.com/archives/C8Z9DSKC7/p1777422136338059?thread_ts=1777416421.251549&cid=C8Z9DSKC7)

packages/perseus/src/styles/widgets/matrix.css
nishasy: It looks like this was removed with the change in design, since the colors are different. I'll see what I can do, but I might have to ask for designs.

packages/perseus/src/widgets/matrix/__docs__/matrix-initial-state-regression.stories.tsx
nishasy: This comment isn't even correct anymore 😬 Thanks for catching that

packages/perseus/src/styles/widgets/matrix.css
nishasy: Weirdly enough, it looks like this background color didn't actually make a difference. You can see what happens when I turn on `static` in prod.

https://github.com/user-attachments/assets/7a8d64d6-7641-4d03-968e-ce16e3774323

nishasy: Okay I wanted to check my understanding, so I asked Claude, and it gave me a very robust answer for this.

tl;dr the static styling hasn't been working for a long time now, so I think it's okay to remove it here, and someone can come back to this and use WB `disabled` whenever we get to matrix remediation.

> ⏺ I have a definitive answer: No, the .static-mode rule on main is doing nothing visible —
>    and your manual test correctly confirms it.
>                                                                                           
>   Here's the chain of reasoning, and it's a fun one:                                      
>                                                                                           
>   The selector itself is correct. matrix.tsx:343-347 produces <div class="perseus-matrix  
>   static-mode"> when props.static is true, and the CSS selector
>   .static-mode.perseus-matrix input (specificity 0,2,1) properly targets descendant       
>   <input> elements. That part of the cascade is fine.
> 
>   But the <input> is no longer a plain HTML input — it's rendered via Wonder Blocks. Trace
>    at matrix.tsx:478 → custom wrapper at
>   packages/perseus/src/components/text-input.tsx:114 → Wonder Blocks <TextField>. In the  
>   Wonder Blocks bundle (@khanacademy/wonder-blocks-form@7.5.7), the actual <input> is
>   styled with Aphrodite via StyledInput = addStyle("input"), and the relevant
>   StyleSheet.create(...) block sets:
> 
>   default: { background: semanticColor.input.default.background, ... }
>   disabled: { background: semanticColor.input.disabled.background, ... }                  
>   readOnly: { background: semanticColor.input.readOnly.background, ... }
>                                                                                           
>   Aphrodite (v1.2.5, the default entry point — not aphrodite/no-important) compiles every 
>   declaration with !important. I confirmed Wonder Blocks doesn't import
>   aphrodite/no-important anywhere in the bundle. So the input's actual background is set  
>   with !important, which beats any selector specificity in matrix.css. The static-mode
>   rule lands on the element but is silently outranked.
> 
>   ★ Insight ─────────────────────────────────────
> 
>   - !important defeats specificity, full stop. This is a place where intuition often fails
>    — many CSS authors expect "more specific selector wins," but !important is a separate
>   axis of the cascade that outranks specificity entirely. A selector like                 
>   (specificity 0,1,0) under normal rules. But .input_<hash> { background: X !important }
>   beats both, regardless of how many classes the other selector chains together. The only
>   ways to fight back are: (a) also use !important, (b) use an inline style="..." attribute
>    with !important, or (c) get the styling out of the Aphrodite layer entirely.           
>   - Aphrodite's !important default is a deliberate library choice. The reasoning is that
>   component authors using Aphrodite want their styles to be self-contained and predictable
>    — they don't want a stray global stylesheet in some app to break their components.     
>   There's an opt-out (aphrodite/no-important), but Wonder Blocks chose not to use it. So
>   all Wonder Blocks components have this property: their internal styles are immune to    
>   override from external CSS.                                                         
>   - This is a silent migration regression, not a planned removal. When the matrix was
>   originally written, the <input> elements were plain HTML, and the .static-mode and 
>   .outside rules worked as intended. At some point the inputs were swapped to Wonder      
>   Blocks TextField, and both rules silently became dead — the rules still appear to apply
>   in DevTools (you'll see them listed), but they're crossed out as overridden by the      
>   Aphrodite !important rule. There's no error, no warning, no test failure — just two CSS
>   rules that quietly stopped doing anything.                                             
>   - This also explains why the [matrix-colors] commit removed both rules. The design
>   team's color refresh probably observed that removing them produced no visual change
>   (because they were dead anyway). What looked like an intentional design decision was    
>   likely just acknowledgment of the existing reality.                                 
>   - DevTools verification is easy: open a static-mode matrix in browser DevTools, click on
>    an input, and check the Styles pane. You'll see the matrix .static-mode rule listed but
>    with strikethrough — it's matched but overridden by the Aphrodite-generated input      
>   style.                                                                            
>   ─────────────────────────────────────────                                      
>                                                    
>   The same is true for .perseus-matrix input.outside { background: #f3f3f3; } — that rule
>   (which was meant to give a slightly different shade to cells outside the user's selected
>    matrix dimensions) is also dead for the same reason. The outside class is still being
>   applied to inputs, but the background override is overridden by Aphrodite's !important. 
>                   
>   What this means for your earlier question: the static-mode styles haven't been          
>   functionally part of the matrix's appearance for a while. If you want static mode to
>   visually differ from interactive mode, the right path is the one we already started on  
>   the matrix-colors branch — pass disabled={this.props.static} to the <TextField>, which
>   makes WB apply its own disabled styling (the semanticColor.input.disabled.background
>   from the same StyleSheet.create(...) block above). That's the only way to change the
>   input background in a Wonder Blocks world: go through the component's own props, not
>   through external CSS.

Myranae: Thanks for the video! Should we also remove the static-mode code from `matrix.tsx` if we're removing the styling that was related to it? Do you know if anything else depends on that code?

mark-fitzgerald: At the minimum, we should at least add a comment in the code stating that it was used for styling, but that the styling wasn't "working" (or something to that effect) so that when this widget is revitalized, project members have this information.

nishasy: I'm inclined to leave the static logic in and add a comment like Mark said. I'm still trying to figure out the line to update the colors without crossing into too much behavior/logic territory.

packages/perseus/src/styles/widgets/matrix.css
```diff
@@ -31,57 +32,20 @@
...
-.perseus-matrix input:focus,
-.perseus-matrix .number-input:focus {
-    border: none;
-    outline: none;
-}
-.static-mode.perseus-matrix input,
-.static-mode.perseus-matrix .number-input {
-    background: #f5f5f5;
-}
```
nishasy, line 66: Removed all these styles because they were not being applied, since Wonder Blocks styles from the underlying TextFields override them. The new specs match WB styles, so we can just rely on WB to provide the styling, rather than hardcoding them in the CSS.

packages/perseus/src/styles/widgets/matrix.css
```diff
@@ -10,8 +10,9 @@
 .perseus-matrix div.paragraph {
     margin: 0;
 }
+/* The fill behind the matrix input cells */
```
Myranae, line 13: Thank you!! ❤️ Super appreciate these comments. 😄

packages/perseus/src/widgets/matrix/__docs__/matrix-initial-state-regression.stories.tsx
```diff
@@ -0,0 +1,97 @@
...
+    args: {
+        matrixBoardSize: [3, 3],
+    } satisfies Partial<MatrixPublicWidgetOptions>,
+};
+
+// Verifies the partially-filled state: the top-left 2×2 block inputs
+// have medium borders; the remaining cells (outside the active area)
+// have a thin border.
+export const PartiallyFilledMatrix: Story = {
+    decorators: [matrixRendererDecorator],
```
Myranae, line 39: Apparently this can be pulled into the meta area (useful for both you regression story files)! Jeremy just pointed this out on my PR. Not sure how to handle when some have multiple decorators. I just left those stories with both listed, but maybe you can just put the shared one in the meta and adding this line to a story with a unique decorator just adds the unique decorator? Not sure, so will probably just list the shared decorator twice in my stories that have unique decorators 😅

packages/perseus/src/widgets/__testutils__/matrix-renderer-decorator.tsx
```diff
@@ -0,0 +1,43 @@
...
+    generateTestPerseusRenderer,
+} from "@khanacademy/perseus-core";
+import * as React from "react";
+
+import QuestionRendererForStories from "./question-renderer-for-stories";
+
+import type {APIOptions} from "../../types";
+import type {UserInputMap} from "@khanacademy/perseus-core";
+
+export const matrixRendererDecorator = (
```
Myranae:
```suggestion
import type {Decorator} from "@storybook/react-vite";

export const matrixRendererDecorator: Decorator = (
```
Jeremy just pointed this out in mine, but I guess we can type this!

packages/perseus/src/widgets/matrix/__docs__/matrix-initial-state-regression.stories.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+import type {MatrixPublicWidgetOptions} from "@khanacademy/perseus-core";
+import type {Meta, StoryObj} from "@storybook/react-vite";
+
+const MatrixWidget = getWidget("matrix")!;
+
+const meta: Meta<typeof MatrixWidget> = {
+    title: "Widgets/Matrix/Visual Regression Tests/Initial State",
+    component: MatrixWidget,
+    tags: ["!autodocs"],
```
Myranae:
```suggestion
    tags: ["!autodocs", "!manifest"],
```

Claude mentioned adding `!manifest` here based on instructions in the CLAUDE.md:

<details>
<summary>CLAUDE.md instructions on !manifest</summary>
`!manifest` keeps the story out of the component manifest — the index Storybook uses for component discovery and documentation browsing. Regression stories are Chromatic-only snapshots, not component documentation, so they shouldn't appear alongside the real widget stories in the manifest. The CLAUDE.md explicitly lists *-regression.stories.tsx as one of the file types that should carry the tag.   
</details>

So I added that to my stories. If that's not what we want to do though, I can update mine as well :) Would be relevant to your other regression story file too.

packages/perseus/src/widgets/matrix/matrix.tsx
```diff
@@ -460,15 +454,18 @@ class Matrix extends React.Component<Props, State> implements Widget {
                                     } else {
                                         const updatedProps = {
                                             ...inputProps,
+                                            // TODO(LEMS-3686): Use CSS modules after
+                                            // Wonder Blocks supports it instead of
+                                            // inline styles.
```
Myranae, line 459: Thank you! 🤩

packages/perseus/src/widgets/matrix/__docs__/matrix-initial-state-regression.stories.tsx
nishasy: That's a good idea! I think you do have to put all of them if there are multiple, because I think it fully overrides the `decorators` value from the meta. So your approach sounds right to me.

mark-fitzgerald (approved):

PR #3527, opened 2026-04-28
Fix racing Prettier invocations in VSCode that corrupt the file

jeremywiebe:
## Summary:

Both Nisha and I noticed that recently (maybe that last month or two) that when saving files, sometimes the file ends up corrupted (one, or more, lines of code are modifed during the save in a way that leaves them with missing characters). When this happens, we noticed that you can do a single Undo and it will fix the file to a clean, formatted state. 

I did some digging today and with Claude's help, I think I figured it out.  Sorry, long story... 

A while back, the VSCode Pretty extension added itself as a `source.fixall` entry (`source.fixall.prettier`). The ESLint extension has always been registered as one (`source.fixall.eslint`). These `source.fixall` are used in the `codeActionsOnSave` setting to define what is run ... on save. 

Before this PR, we registered _all_ `source.fixall` actions to run on save _and_ because the Prettier extension didnt' register itself as a `source.fixall` item, just ESLint ran on save. 

Now that prettier is a `source.fixall` item, we get ESLint running Prettier itself (because we register the `prettier/prettier` ESLint plugin) _and_ the Prettier extension running on save.

These two actions race. ESLint does not reformat the entire file, but instead applies "range edits" (edits to specific pieces of the file). By the time this happens, the file has been reformatted entirely by Prettier and the ranges are stale. This breaks the file.

Here's a video showing it before (breaking):

https://github.com/user-attachments/assets/2334ad03-f1bc-4ae7-ba33-d7aec4836877

Issue: LEMS-4065 

## Test plan:

Open the project in VSCode. Add a space within `{}` of an import statement and save. The file should be re-formatted back to its original state, but not corrupted.


.vscode/settings.json
```diff
@@ -1,8 +1,9 @@
 {
     "editor.renderWhitespace": "boundary",
     "editor.codeActionsOnSave": {
-        "source.fixAll": "explicit"
+        "source.fixAll.eslint": "explicit"
     },
+    "editor.formatOnSave": false,
```
jeremywiebe, line 6: This overrides any personal VSCode setting. If this is on, it adds a third formatter (which is a _second_ Prettier formatter). We don't want that!

.vscode/settings.json
```diff
@@ -1,8 +1,9 @@
 {
     "editor.renderWhitespace": "boundary",
     "editor.codeActionsOnSave": {
-        "source.fixAll": "explicit"
+        "source.fixAll.eslint": "explicit"
```
jeremywiebe, line 4: This changes things so only the ESLint extension runs on save. But we still get formatting because our ESLint config runs Prettier itself.

nishasy (approved):
Omg thank you so much!!

PR #3534, opened 2026-04-29
chore(kmath): remove underscore dependency and replace with native JS

marnikostman:
## Summary:
Replaces all underscore usages in kmath with native Array methods and
built-in JS equivalents, removing the package from devDependencies and peerDependencies.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>
                                                                                                                                       
  Changes                                                                                                                              
                                                                                                                                      
  - Removed underscore from devDependencies and peerDependencies in package.json                                                       
  - Replaced all _ usages with native JS equivalents:
    - _.map / _.reduce / _.all → Array.map / Array.reduce / Array.every                                                                
    - _.zip → index-based mapping (avoids the [T | undefined, T | undefined] typing issues _.zip introduced)                           
    - _.isNumber / _.isNaN → typeof x === "number" / Number.isNaN                                                                      
    - _.contains → Array.includes                                                                                                      
  - Fixed several pre-existing @ts-expect-error suppressions that were artifacts of underscore's weak types — these are now clean      
                                                                                                                                       
  Testing                                                
                                                                                                                                       
  Existing test suite covers the changed functions. No behavioral changes intended.

### DIFF
--- a/packages/kmath/package.json
+++ b/packages/kmath/package.json
@@ -33,6 +33,6 @@
     },
     "keywords": [],
     "khan": {
-        "catalogHash": "d7663955f3310f2b"
+        "catalogHash": "e3b0c44298fc1c14"
     }
 }
--- a/packages/kmath/package.json
+++ b/packages/kmath/package.json
@@ -30,12 +30,10 @@
     },
     "devDependencies": {
         "jquery": "catalog:devDeps",
-        "perseus-build-settings": "workspace:*",
-        "underscore": "catalog:devDeps"
+        "perseus-build-settings": "workspace:*"
     },
     "peerDependencies": {
-        "jquery": "catalog:peerDeps",
-        "underscore": "catalog:peerDeps"
+        "jquery": "catalog:peerDeps"
     },
     "keywords": [],
     "khan": {
--- a/packages/kmath/package.json
+++ b/packages/kmath/package.json
@@ -29,11 +29,7 @@
         "@khanacademy/perseus-utils": "workspace:*"
     },
     "devDependencies": {
-        "jquery": "catalog:devDeps",
         "perseus-build-settings": "workspace:*"
-    },
-    "peerDependencies": {
-        "jquery": "catalog:peerDeps"
     },
     "keywords": [],
     "khan": {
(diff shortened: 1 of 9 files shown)


marnikostman:
Worked on this during Agentic Coding 201 workshop :)

marnikostman:
@Khan/lems Any chance someone has time for review?

packages/kmath/src/math.ts
```diff
@@ -256,7 +252,10 @@ const KhanMath = {
             }
         }
 
-        if (_(["proper", "improper", "mixed", "fraction"]).contains(format)) {
+        if (
+            format != null &&
+            ["proper", "improper", "mixed", "fraction"].includes(format)
```
jeremywiebe, line 257: FWIW, I think we could allow `null` to flow into the `includes()` call here.  `['a', 'b', 'c'].includes(null)` is false just like any other string that isn't in the list. 
```suggestion
            ["proper", "improper", "mixed", "fraction"].includes(format)
```

packages/kmath/package.json
```diff
@@ -29,16 +29,10 @@
...
     "devDependencies": {
-        "jquery": "catalog:devDeps",
-        "perseus-build-settings": "workspace:*",
-        "underscore": "catalog:devDeps"
-    },
-    "peerDependencies": {
-        "jquery": "catalog:peerDeps",
-        "underscore": "catalog:peerDeps"
+        "perseus-build-settings": "workspace:*"
     },
```
jeremywiebe, line 39: Really happy to drop these peer deps!

jeremywiebe (approved):
Cool!  Thanks for doing this cleanup work!

packages/kmath/src/math.ts
marnikostman: Tried to apply this and got a TS error: error TS2345: Argument of type 'string | undefined' is not assignable to parameter of type 'string'. I think I would have had to do an 'as string' cast here or lose the literal types which doesn't seem ideal.

jeremywiebe: Ah!  That makes sense.

PR #3537, opened 2026-04-29
Stop minifying CSS in rollup build to drop browserslist transitive dep

jeremywiebe:
## Summary:

Stop minifying CSS in the rollup build. This removes our transitive dependency on `cssnano` → `caniuse-lite`/`browserslist`, which was the source of the periodic "browsers data is N months old" warnings during pnpm build. We had a Github Action that ran periodically, but it meant we had a weekly PR we needed to review and land. With the age of rampant supply-chain attacks, it seems easier to just not do this. We don't really need to minify because downstream consumers can (and do) minify all CSS during bundling!

  - config/build/rollup.config.js: set `minimize: false` in `rollup-plugin-postcss`
  - `package.json`: drop the now-unused `update-browserslist-db` `devDependency` (its only purpose was refreshing `caniuse-lite` when the warnings appeared)

**Background**

Two chains pull browserslist into node_modules:
  1. `@cypress/code-coverage` → `@cypress/webpack-preprocessor` → `@babel/preset-env` — never executes (Cypress here is configured to use the Vite bundler, not webpack), so this chain doesn't fire the staleness warning. Dead weight, but harmless. Left alone for this PR.
  2. `rollup-plugin-postcss` → `cssnano` → `caniuse-api` → `caniuse-lite` — runs on every pnpm build because `minimize: true` was set. `cssnano` queries `caniuse-lite` to decide which CSS optimizations are safe, and that lookup is what prints the staleness warning. This is the chain this PR addresses. Since Perseus publishes a library and consuming applications re-minify CSS in their own bundlers, pre-minifying here was redundant work.

**Verification**

In order to make sure removing magnification isn't break anything, I built the packages that produce CSS (`perseus`, `math-input`, and `perseus-editor`) on this branch (unminified) and on `main` (minified). I then minified `main`'s CSS manually using `cssnano` and compared SHA-1 hashes against main's output, ignoring the trailing /*# 
  sourceMappingURL=... */ comment. All three files matched byte-for-byte:                                                     
                                                                                
| Package | SHA-1 |
| ------- | ----- |
| @khanacademy/perseus        | 28970cf925a7aa7f01df37e6d9710feedfd4cee4 |
| @khanacademy/math-input     | a799b415360205ceea185ba12534604cd042ee08 |                                                  
| @khanacademy/perseus-editor | 15a21f98ed99984e59f34deb26a7c4c4c5dcb23d |                                                  

**Size impact**

CSS files in the npm tarballs grow because they ship unminified now:

|           Package           | Before (gzip) | After (gzip) |
|-----------------------------|---------------|--------------|
| @khanacademy/perseus        | 10.1 KB       | 18.6 KB      |
| @khanacademy/math-input     | 2.9 KB        | 3.6 KB       |
| @khanacademy/perseus-editor | 4.8 KB        | 6.0 KB       |
                                                                       
End-user payload should be unchanged once consumers re-minify and gzip. If any consumer ships these CSS files straight to the browser without their own bundler step, they'll see a real size regression — please flag if you know of such a consumer.

Issue: 

## Test plan:                                                                                                           
            
  - `pnpm clean; pnpm install` -- clean                                                                                   
  - `pnpm build` — completes without "browserslist" staleness errors
  - `pnpm test` 
  - `pnpm tsc`
  - Verify dist/index.css is produced for perseus, math-input, and perseus-editor (just unminified now)


jeremywiebe:
Hrm... I'm not loving this approach.  It still doesn't remove `browserslist` from the lock file.

PR #3538, opened 2026-04-30
Add Storybook MCP addon for AI-assisted UI development

ivyolamit:
## Summary:
- Adds `@storybook/addon-mcp` (v0.6.0) to enable AI coding assistants (like Claude Code) to connect directly to our running Storybook instance
- Adds `!manifest` tags to 27 non-component stories (regression tests, documentation pages, a11y docs, dev utilities) to keep the component manifest clean
- Adds a developer guide at `.storybook/storybook-mcp-guide.md` explaining what the addon does, how to set it up, and how it works
- Updates `CLAUDE.md` with Storybook MCP workflow and `!manifest` tagging rules

## What is this?

`@storybook/addon-mcp` runs an MCP (Model Context Protocol) server inside the Storybook dev server. This lets AI assistants query Storybook for project-specific story authoring conventions and generate live preview URLs to verify their work visually — instead of guessing at patterns from source files alone.

**Available tools:**
| Tool | Purpose |
|------|---------|
| `get-storybook-story-instructions` | Returns story authoring guidelines specific to our setup |
| `preview-stories` | Returns live Storybook URLs for specific stories |

## Why?

Perseus has specific story patterns (widget generators, `__docs__/` directories, strict versioned data schema) that are easy to get wrong without context. This addon gives AI assistants direct access to those conventions so they follow them correctly on the first try.

### New files
- `.storybook/storybook-mcp-guide.md` — Developer guide covering setup, usage, architecture, and `!manifest` tagging

### Configuration
- `package.json` — Added `@storybook/addon-mcp: ^0.6.0` to devDependencies
- `.storybook/main.ts` — Registered `@storybook/addon-mcp` in addons array
- `CLAUDE.md` — Added Storybook MCP section with workflow and Perseus-specific rules

### `!manifest` tags (27 files)
Added `!manifest` to exclude non-component entries from the component manifest:

- **Root documentation MDX** (4): introduction, getting-started, api, renderers-overview
- **Accessibility docs MDX** (7): a11y.mdx for definition, dropdown, explanation, expression, interactive-graph, numeric-input, radio
- **Visual regression stories** (15): all `*-regression.stories.tsx` across definition, dropdown, explanation, free-response, image, interactive-graph, label-image, numeric-input, radio, article-editor, editor-page
- **Dev utility** (1): `preview.stories.tsx`

Actual widget/component stories and internal sub-component stories were intentionally left without `!manifest` — they belong in the manifest.

## Setup (per developer, one-time)

```bash
pnpm storybook
claude mcp add storybook-mcp --transport http http://localhost:6006/mcp --scope project
```
Co-authored by Claude Code

Issue: LEMS-XXXX

## Test plan

- [ ] Run `pnpm storybook` and verify Storybook starts without errors
- [ ] Open `http://localhost:6006/mcp` in a browser and verify the MCP server status page loads
- [ ] Run `claude mcp list` and verify `storybook-mcp` shows as connected
- [ ] Verify regression stories and documentation pages still render correctly in Storybook (the `!manifest` tag doesn't affect rendering)
- [ ] Run `pnpm tsc` — type-check passes

### DIFF
--- a/.storybook/storybook-mcp-guide.md
+++ b/.storybook/storybook-mcp-guide.md
@@ -56,7 +56,7 @@ When writing or modifying UI components, Claude Code now has two tools it can us
 
 - **Storybook must be running** on port 6006 for the MCP tools to work. If it's not running, Claude Code will still function normally — it just won't have access to the Storybook tools.
 - **This is local only.** The MCP server registration is per-developer and doesn't affect CI, builds, or other developers.
-- **It doesn't replace testing.** Continue using `pnpm test` for unit tests, Cypress for interaction tests, and the a11y addon for accessibility checks. The MCP tools help the AI write *better* code and stories — they don't run tests.
+- **It doesn't replace testing.** AI assistants should write tests for any feature they implement, not just stories. Continue using `pnpm test` for unit tests, Cypress for interaction tests, and the a11y addon for accessibility checks. The MCP tools help the AI write *better* code and stories — they don't run tests.
 
 ---
 
--- a/.storybook/storybook-mcp-guide.md
+++ b/.storybook/storybook-mcp-guide.md
@@ -107,8 +107,8 @@ MCP (Model Context Protocol) is a standard way for AI assistants to connect to e
 **The flow:**
 1. You give Claude Code a prompt (e.g., "Create stories for the dropdown widget")
 2. Claude Code calls the MCP tools on the running Storybook server
-3. Storybook returns project-specific conventions and story metadata
-4. Claude Code writes the code and gives you a preview URL
+3. Storybook returns story authoring guidelines and live preview URLs
+4. Claude Code writes the code following those guidelines and shares the preview URL
 5. You click the link to verify the result in your browser
 
 ### What happens without MCP vs. with MCP
@@ -167,7 +167,7 @@ In `.mdx` files, add the `tags` prop to the `<Meta>` component:
 
 Perseus has specific patterns that are easy to get wrong:
 - **Widget generators** in `packages/perseus-core/src/utils/generators` should be used for test data — not hand-rolled props
-- **Stories live in `__docs__/` directories** within each widget/component folder, not in a central location
+- **New stories go in `__docs__/` directories** within each widget/component folder. Some older stories are colocated directly (e.g., `matrix/matrix.stories.tsx`) but haven't been migrated yet.
 - **The widget data schema is versioned and strict** — inventing props that don't exist will cause type errors or runtime failures
 - There are **~60+ existing story files** across widgets, editors, components, and math-input with established conventions
 
--- a/.storybook/storybook-mcp-guide.md
+++ b/.storybook/storybook-mcp-guide.md
@@ -1,0 +1,183 @@
+# Using Storybook MCP in Perseus
+
+## What is it?
+
+`@storybook/addon-mcp` connects Claude Code (or other AI coding assistants) directly to your running Storybook instance. Instead of the AI guessing at component APIs and story patterns, it can query your actual Storybook for real documentation and generate live preview URLs to verify its work visually.
+
+## What's already done
+
+The addon is installed and configured in the repo — no code changes needed on your end.
+
+## One-time setup (per developer)
+
+### 1. Start Storybook
+
+```bash
+pnpm storybook
+```
+
+### 2. Register the MCP server with Claude Code
+
+```bash
+claude mcp add storybook-mcp --transport http http://localhost:6006/mcp --scope project
+```
+
+This only needs to be done once, and creates the file `.mcp.json`. It saves the connection to your local Claude Code project settings (not committed to git).
+
+### 3. Verify it's connected
+
+```bash
+claude mcp list
+```
+
+You should see:
+
+```
+storybook-mcp: http://localhost:6006/mcp (HTTP) - Connected
+```
+
+## How it helps day-to-day
+
+When writing or modifying UI components, Claude Code now has two tools it can use:
+
+| Tool                               | What it does                                                       | When it's useful                                   |
+| ---------------------------------- | ------------------------------------------------------------------ | -------------------------------------------------- |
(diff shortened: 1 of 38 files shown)


.storybook/storybook-mcp-guide.md
ivyolamit, line 1: I'm not sure where to put this yet, but for now i'll add it here as we are experimenting the use of this mcp.

.storybook/storybook-mcp-guide.md
```diff
@@ -0,0 +1,183 @@
...
+- "Create stories for the new widget I just built"
+- "Add a story showing the error state for numeric-input"
+- "Update the expression widget stories to cover the mobile keypad"
+- "Show me what the radio widget looks like in its selected state"
+
+## Important notes
+
+- **Storybook must be running** on port 6006 for the MCP tools to work. If it's not running, Claude Code will still function normally — it just won't have access to the Storybook tools.
+- **This is local only.** The MCP server registration is per-developer and doesn't affect CI, builds, or other developers.
+- **It doesn't replace testing.** Continue using `pnpm test` for unit tests, Cypress for interaction tests, and the a11y addon for accessibility checks. The MCP tools help the AI write *better* code and stories — they don't run tests.
```
handeyeco: I would maybe add that we also expect LLMs to add tests for features it's writing. It's not enough to just write a story and _run_ the tests, it also need to _write_ tests.

.storybook/storybook-mcp-guide.md
```diff
@@ -0,0 +1,183 @@
...
+│         views in browser    │ Components  │                    │
+│                             │ (Perseus)   │                    │
+│                             └─────────────┘                    │
+└────────────────────────────────────────────────────────────────┘
+```
+
+**The flow:**
+1. You give Claude Code a prompt (e.g., "Create stories for the dropdown widget")
+2. Claude Code calls the MCP tools on the running Storybook server
+3. Storybook returns project-specific conventions and story metadata
```
handeyeco: Isn't this already in the code?

.storybook/storybook-mcp-guide.md
```diff
@@ -0,0 +1,183 @@
...
+In `.mdx` files, add the `tags` prop to the `<Meta>` component:
+```mdx
+<Meta title="Widgets/Definition/Accessibility" tags={["!manifest"]} />
+```
+
+### Why does Perseus benefit from this?
+
+Perseus has specific patterns that are easy to get wrong:
+- **Widget generators** in `packages/perseus-core/src/utils/generators` should be used for test data — not hand-rolled props
+- **Stories live in `__docs__/` directories** within each widget/component folder, not in a central location
```
handeyeco: That's not strictly true and I'd argue maybe it shouldn't be the norm. If you look at `perseus/packages/perseus/src/widgets/matrix` it's:

```
+ matrix
    - index.ts
    - matrix.ts
    - matrix.test.ts
    - matrix.stories.ts
```

Which would be my preference rather than using `__docs__` for stories.

.storybook/storybook-mcp-guide.md
```diff
@@ -0,0 +1,183 @@
...
+- There are **~60+ existing story files** across widgets, editors, components, and math-input with established conventions
+
+The MCP connection gives Claude Code direct access to these conventions so it follows them correctly on the first try, rather than requiring you to review and correct its output.
+
+### What you don't need to change about how you work
+
+- You still write and review code the same way
+- You still run `pnpm test`, lint, and type-check before submitting PRs
+- You still use Storybook in your browser for visual development
+- The MCP addon is invisible if you're not using an AI assistant — it adds no overhead to Storybook's normal behavior
```
handeyeco, line 181: Do we know if this uses more tokens? Is there a cost ramification at all?

package.json
```diff
@@ -34,6 +34,7 @@
         "@storybook/addon-a11y": "10.3.1",
         "@storybook/addon-docs": "10.3.1",
         "@storybook/addon-links": "10.3.1",
+        "@storybook/addon-mcp": "^0.6.0",
```
handeyeco:
```suggestion
        "@storybook/addon-mcp": "0.6.0",
```

I think with all the supply chain attacks we should be locking to a specific version.

EDIT: IDK if that's true or not 🤷

handeyeco (approved):
I don't really understand the benefit still after reading this, but I'm down to try it.

.storybook/storybook-mcp-guide.md
anakaren-rojas: This is also what I was thinking 🤔

package.json
ivyolamit: good catch!

.storybook/storybook-mcp-guide.md
ivyolamit: This is great question! Since this runs locally mostly it doesn't use our tokens, but it only uses tokens when it call the command like `get-storybook-story-instructions` and other commands when called multiple times. I've updated the claude.md file to improve how it interacts, and this will significantly improve token usage because reading the story conventions via MCP will be cheaper than navigating through source directories.

.storybook/storybook-mcp-guide.md
ivyolamit: Actually, after the documentation improvement last year our guide indicates that we put our docs in `__docs__` this was posted in one of the slack thread before see guide [here](https://khanacademy.atlassian.net/wiki/spaces/LC/pages/4236673137/Perseus+Documentation+Guide+Storybook#Code-Structure)
The previous discussion is align on how Jeremy mentioned that test will be under `__tests__` folder, this is guidance is something we applied during radio widget project. But looks like it's not known to all members 😢

.storybook/storybook-mcp-guide.md
ivyolamit: good call out, i've updated the flow

ivyolamit:
> I don't really understand the benefit still after reading this, but I'm down to try it.

@handeyeco thanks!
I think the huge benefit for this is, in theory, is that it will be cheaper in the long run than having claude to read the source code. This is something we need to keep an eye on since we don't have a data for it atm. 😓

PR #3540, opened 2026-04-30
Fix TypeScript suppressions tagged with LEMS-2656

benchristel:
## Summary:
We turned on `strictFunctionTypes` in TS back in 2024, which revealed a bunch
of type errors. This PR fixes the ones that still remain.

Issue: LEMS-2656

## Test plan:

CI checks should pass.

### DIFF
--- a/packages/math-input/src/components/keypad/keypad-mathquill.stories.tsx
+++ b/packages/math-input/src/components/keypad/keypad-mathquill.stories.tsx
@@ -10,7 +10,6 @@ import {getKeyTranslator} from "../key-handlers/key-translator";
 
 import type {MathFieldInterface} from "../input/mathquill-types";
 import type {KeypadKey} from "@khanacademy/perseus-core";
-import type MathQuill from "mathquill";
 
 import Keypad from "./index";
 
--- a/packages/math-input/src/components/keypad/keypad-mathquill.stories.tsx
+++ b/packages/math-input/src/components/keypad/keypad-mathquill.stories.tsx
@@ -12,6 +12,7 @@ import type {MathFieldInterface} from "../input/mathquill-types";
 import type {KeypadKey} from "@khanacademy/perseus-core";
 
 import Keypad from "./index";
+import type MathQuill from "mathquill";
 
 export default {
     title: "Math Input/Components/v2 Keypad With Mathquill",
@@ -32,13 +33,15 @@ export function V2KeypadWithMathquill() {
                 mathFieldWrapperRef.current,
                 "en",
                 mockStrings,
-                // TODO(LEMS-2656): remove TS suppression
-                // @ts-expect-error: Type 'EditableMathQuill' is not assignable to type 'MathFieldInterface'.
                 (baseConfig) => ({
                     ...baseConfig,
                     handlers: {
-                        edit: (_mathField: MathFieldInterface) => {
-                            setCursorContext(getCursorContext(_mathField));
+                        edit: (_mathField) => {
+                            // Cast to MathFieldInterface to get the types for
+                            // the cursor() and controller() methods, which
+                            // exist but aren't part of the published
+                            // MathQuill API.
+                            setCursorContext(getCursorContext(_mathField as MathFieldInterface));
                         },
                     },
                 }),
--- a/packages/math-input/src/components/keypad/keypad-mathquill.stories.tsx
+++ b/packages/math-input/src/components/keypad/keypad-mathquill.stories.tsx
@@ -10,9 +10,9 @@ import {getKeyTranslator} from "../key-handlers/key-translator";
 
 import type {MathFieldInterface} from "../input/mathquill-types";
 import type {KeypadKey} from "@khanacademy/perseus-core";
+import type MathQuill from "mathquill";
 
 import Keypad from "./index";
-import type MathQuill from "mathquill";
 
 export default {
     title: "Math Input/Components/v2 Keypad With Mathquill",
@@ -41,7 +41,11 @@ export function V2KeypadWithMathquill() {
                             // the cursor() and controller() methods, which
                             // exist but aren't part of the published
                             // MathQuill API.
-                            setCursorContext(getCursorContext(_mathField as MathFieldInterface));
+                            setCursorContext(
+                                getCursorContext(
+                                    _mathField as MathFieldInterface,
+                                ),
+                            );
                         },
                     },
                 }),
--- a/packages/math-input/src/components/keypad/keypad-mathquill.stories.tsx
+++ b/packages/math-input/src/components/keypad/keypad-mathquill.stories.tsx
@@ -36,12 +36,13 @@ export function V2KeypadWithMathquill() {
                     ...baseConfig,
                     handlers: {
                         edit: (_mathField) => {
-                            // Cast to MathFieldInterface to get the types for
-                            // the cursor() and controller() methods, which
-                            // exist but aren't part of the published
-                            // MathQuill API.
                             setCursorContext(
                                 getCursorContext(
+                                    // Cast to MathFieldInterface to get the types for
+                                    // the cursor() and controller() methods, which
+                                    // exist but aren't part of the published
+                                    // MathQuill API.
+                                    // eslint-disable-next-line no-restricted-syntax
                                     _mathField as MathFieldInterface,
                                 ),
                             );
--- a/packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
+++ b/packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
@@ -209,9 +209,11 @@ const parseLockedPointType = object({
     ariaLabel: optional(string),
 });
 
+export const parseLockedLineKind = enumeration("line", "ray", "segment");
+
 const parseLockedLineType = object({
     type: constant("line"),
-    kind: enumeration("line", "ray", "segment"),
+    kind: parseLockedLineKind,
     points: pair(parseLockedPointType, parseLockedPointType),
     color: parseLockedFigureColor,
     lineStyle: parseLockedLineStyle,
--- a/packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
+++ b/packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
@@ -176,7 +176,7 @@ export const parsePerseusGraphType = discriminatedUnionOn("type")
 
 export const parseLockedFigureColor = enumeration(...lockedFigureColorNames);
 
-const parseLockedFigureFillType = enumeration(
+export const parseLockedFigureFillType = enumeration(
     "none",
     "white",
     "translucent",
--- a/packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
+++ b/packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
@@ -176,7 +176,7 @@ export const parsePerseusGraphType = discriminatedUnionOn("type")
 
 export const parseLockedFigureColor = enumeration(...lockedFigureColorNames);
 
-export const parseLockedFigureFillType = enumeration(
+export const parseLockedFigureFillStyle = enumeration(
     "none",
     "white",
     "translucent",
@@ -239,7 +239,7 @@ const parseLockedEllipseType = object({
     radius: pairOfNumbers,
     angle: number,
     color: parseLockedFigureColor,
-    fillStyle: parseLockedFigureFillType,
+    fillStyle: parseLockedFigureFillStyle,
     strokeStyle: parseLockedLineStyle,
     weight: parseStrokeWeight,
     labels: defaulted(array(parseLockedLabelType), () => []),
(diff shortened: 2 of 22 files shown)


packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
```diff
@@ -348,3 +352,25 @@ export const parseInteractiveGraphWidget = parseWidget(
         fullGraphAriaDescription: optional(string),
     }),
 );
+
+/**
+ * A parser that accepts the type field of an interactive graph.
+ */
+export const interactiveGraphTypeParser = enumeration(
```
benchristel: I guess this should be named `parseInteractiveGraphType` for consistency.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.typetest.ts
```diff
@@ -14,3 +18,13 @@ describe("the InteractiveGraphWidget parser", () => {
...
+
+describe("interactiveGraphTypeParser", () => {
+    it("should return the same type as the type field of an interactive graph", () => {
+        const result = interactiveGraphTypeParser("", ctx());
+        assertSuccess(result);
+        expect(result.value).type.toBe<
+            InteractiveGraphWidget["options"]["graph"]["type"]
+        >();
+    });
+});
```
benchristel, line 28: This type test ensures the parser doesn't get out of sync with the data schema as we add more graph types.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.typetest.ts
```diff
@@ -14,3 +18,13 @@ describe("the InteractiveGraphWidget parser", () => {
         >();
     });
 });
+
+describe("interactiveGraphTypeParser", () => {
+    it("should return the same type as the type field of an interactive graph", () => {
+        const result = interactiveGraphTypeParser("", ctx());
+        assertSuccess(result);
+        expect(result.value).type.toBe<
+            InteractiveGraphWidget["options"]["graph"]["type"]
```
benchristel:
```suggestion
            PerseusGraphType["type"]
```

packages/math-input/src/components/keypad/keypad-mathquill.stories.tsx
```diff
@@ -32,13 +32,19 @@ export function V2KeypadWithMathquill() {
...
                 (baseConfig) => ({
                     ...baseConfig,
                     handlers: {
-                        edit: (_mathField: MathFieldInterface) => {
-                            setCursorContext(getCursorContext(_mathField));
+                        edit: (_mathField) => {
+                            // Cast to MathFieldInterface to get the types for
+                            // the cursor() and controller() methods, which
+                            // exist but aren't part of the published
+                            // MathQuill API.
```
handeyeco: Don't we maintain our own fork of MathQuill. Should we file a ticket to change it upstream?

packages/perseus-editor/src/widgets/interactive-graph-editor/components/graph-type-selector.tsx
```diff
@@ -42,7 +48,13 @@ const GraphTypeSelector = (props: GraphTypeSelectorProps) => {
...
         <SingleSelect
             selectedValue={props.graphType}
-            onChange={props.onChange}
+            onChange={(value) => {
+                const parsedType = parse(value, parseInteractiveGraphType);
+                if (isFailure(parsedType)) {
+                    throw new Error(parsedType.detail);
+                }
+                props.onChange(parsedType.value);
+            }}
```
handeyeco: This seems a little heavy handed to me. If we want the additional type safety to prevent us from adding options that don't conform, couldn't we do something like:

```
const options: {type: GraphType, display: string, show?: boolean}[] = [
  type: "absolute-value", display: "Absolute value",
  type:  "none", display: "None",
  type: "exponential", display: "Exponential function", show: !!showExponential,
]

<SingleSelect>
    { options.map(o => o.show ? <OptionItem value={o.type} label={o.display} /> : null}
</SingleSelect>
```

I guess my main concern is pulling out these fragments of parser (ie `parseInteractiveGraphType`) for doing things that isn't parsing a PerseusItem. I'm sure it would work short-term, but I worry that it might start pulling unrelated logic into the parser which until now has kind of been its own little island.

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/color-select.tsx
```diff
@@ -30,8 +35,16 @@ const ColorSelect = (props: Props) => {
                 <Strut size={spacing.xxSmall_6} />
                 <SingleSelect
                     selectedValue={selectedValue}
-                    // TODO(LEMS-2656): remove TS suppression
-                    onChange={onChange as any}
+                    onChange={(value) => {
+                        const parsedColor = parse(
```
handeyeco, line 40: Yeah, same here. We control the available options that go into the select, so I don't know why we need to validate the selection.

packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
```diff
@@ -356,13 +352,8 @@ class InteractiveGraphEditor extends React.Component<Props> {
                 static: this.props.apiOptions?.editingDisabled ?? false,
                 trackInteraction: function () {},
                 userInput: correct,
-                handleUserInput: (
-                    newGraph: InteractiveGraphProps["userInput"],
-                ) => {
+                handleUserInput: (newGraph: PerseusGraphType) => {
```
handeyeco, line 365: This is probably fine. I was looking at a ticket yesterday though ([LEMS-2344](https://khanacademy.atlassian.net/browse/LEMS-2344)) that got me thinking that maybe we use PerseusGraphType too liberally; long-term it probably needs to be split into more focused types.

packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
```diff
@@ -379,10 +370,20 @@ class InteractiveGraphEditor extends React.Component<Props> {
             graph = (
                 // There are a bunch of props that renderer.jsx passes to widgets via
                 // getWidgetProps() and widget-container.jsx that the editors don't
-                // bother passing.
-                // @ts-expect-error - TS2769 - No overload matches this call.
+                // bother passing. We pass `undefined` for these props and
+                // cast to `any`.
                 <InteractiveGraph
                     {...graphProps}
+                    graph={undefined as any}
```
handeyeco, line 388: Oof...

If it's a valid use to pass `undefined`, should these props just be optional? Or maybe we could have a component in the editor that wraps `InteractiveGraph` and provides defaults as needed.

packages/perseus/src/components/graphie-classes.ts
```diff
@@ -128,7 +128,6 @@ const createSimpleClass = function (addFunction: any): any {
         },
 
         remove: function () {
-            // @ts-expect-error - TS2554 - Expected 3 arguments, but got 2.
```
handeyeco, line 131: It's great seeing all these `@ts-expect-error`s disappear.

packages/perseus/src/components/math-input.tsx
```diff
@@ -251,12 +250,15 @@ class InnerMathInput extends React.Component<InnerProps, State> {
...
+                        upOutOf: (mathField) => {
                             // This handler is called when the user presses the up
                             // arrow key, but there is nowhere in the expression to go
                             // up to (no numerator or exponent). For ease of use,
                             // interpret this as an attempt to create an exponent.
-                            mathField.typedText("^");
+
+                            // BaseMathQuill doesn't have a typedText method,
+                            // so we have to cast here.
+                            (mathField as MathFieldInterface).typedText("^");
```
handeyeco, line 263: I personally don't mind using `as`, but I know @nishasy was talking about adding a lint rule against it?

packages/perseus/src/widgets/input-number/input-number.tsx
```diff
@@ -131,17 +131,13 @@ class InputNumber extends React.Component<Props> implements Widget {
         return true;
     };
 
-    // TODO(LEMS-2656): remove TS suppression
-    // @ts-expect-error: Type 'FocusPath' is not assignable to type 'Path'.
-    focusInputPath: (arg1: Path) => void = (inputPath) => {
+    focusInputPath: () => void = () => {
```
handeyeco, line 134: We just need to eventually get rid of these methods. I don't think we even use them anymore?

handeyeco (commented):
Thanks for looking at this!

This PR got me thinking: we're all trying to make AI work. I'm guessing this is at least AI-assisted? And it makes sweeping changes across the code base. I know @SonicScrewdriver also has a bunch of AI-assisted PRs in the works too that make pretty large changes. Both of these are in IG and doesn't include Cat, Ivy, and myself who are all making big changes to IG with LLMs.

I wonder if we need to be more deliberate about coordinating when work happens so we're not stepping on each other's toes. I'd be curious to hear from you and Third about landing your respective PRs.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
```diff
@@ -348,3 +352,25 @@ export const parseInteractiveGraphWidget = parseWidget(
         fullGraphAriaDescription: optional(string),
     }),
 );
+
+/**
+ * A parser that accepts the type field of an interactive graph.
+ */
+export const parseInteractiveGraphType = enumeration(
```
anakaren-rojas, line 362: We have this list in at least one other [place](https://github.com/Khan/perseus/blob/fb2d5a529e3e069bb164ed8db390113e413bb2c0/packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts#L159-L175) 
I'm wondering if a centralized list would work? Something like
```
export const PERSEUS_GRAPH_TYPES = [
    "angle",
    "circle",
    "linear",
...
]
```

packages/perseus-editor/src/widgets/interactive-graph-editor/components/graph-type-selector.tsx
benchristel: The problem is that the `SingleSelect`'s `onChange` callback receives a `string`, but our `GraphTypeSelector`'s `onChange` takes a `PerseusGraphType["type"]`. To ensure that, we either have to parse at runtime, or we have to use an `as` cast and inspect the code every time we change it to make sure it's consistent. I figured we might as well parse since we'll get better error messages if we screw up. However, I'd be okay changing this to use `as`, since the risk seems pretty low.

packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
benchristel: Part of the problem is that we're rendering a widget without a Renderer wrapping it. The Renderer would provide some of these props. I agree that we need a better solution here. I'll add a TODO to come back to this.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
benchristel: Unfortunately, I don't think that other usage could be easily generated from a list. We'd need at least a map of type names to graph parsers, or something like that.

Other things equal, I'd rather have the duplication (which is kept in sync by the type system) than more complicated code.

packages/math-input/src/components/keypad/keypad-mathquill.stories.tsx
benchristel: I'm not sure whether we should make the upstream changes or not (do all `MathQuill.v3.EditableMathQuill` actually have controller and cursor methods?) but I've added a ticket ([LEMS-4089]) to investigate.

[LEMS-4089]: https://khanacademy.atlassian.net/browse/LEMS-4089

benchristel:
> Thanks for looking at this!
> 
> This PR got me thinking: we're all trying to make AI work. I'm guessing this is at least AI-assisted? And it makes sweeping changes across the code base. I know @SonicScrewdriver also has a bunch of AI-assisted PRs in the works too that make pretty large changes. Both of these are in IG and doesn't include Cat, Ivy, and myself who are all making big changes to IG with LLMs.
> 
> I wonder if we need to be more deliberate about coordinating when work happens so we're not stepping on each other's toes. I'd be curious to hear from you and Third about landing your respective PRs.

It was not AI-assisted, actually! Fixing type suppressions is one thing I do not trust LLMs to do, since they don't have a concept of type safety and can only tell whether typechecking passes or not. My goal was to avoid invasive changes and make targeted local fixes. I am also curious to see whether any merge conflicts will result from this.

Thanks for the reminder to coordinate with IG folks. I'm happy to hold off on landing this, since it's not particularly urgent.

packages/perseus-editor/src/widgets/interactive-graph-editor/components/graph-type-selector.tsx
benchristel: @handeyeco I took your suggestion and built a `TypedSingleSelect` component around it: https://github.com/Khan/perseus/pull/3540/changes/fdde527ca8085bdd0d106e0021a6e690eea736a8

PR #3541, opened 2026-04-30
Fix bugs in lint.sh (`pnpm fixc`)

benchristel:
## Summary:
Previously, `pnpm fixc` would never run `eslint` - it would only run prettier.
The `grep || true` in the eslint pipeline broke it in half because `||` has
lower precedence than `|`. This PR fixes that issue by running grep in a
subshell.

Issue: none

## Test plan:

Change a TS file and run `pnpm fixc`. Both prettier and eslint should run.


SonicScrewdriver (approved):
Thank you!I :)

PR #3546, opened 2026-05-01
Add lint rule to avoid using `as` for type safety

nishasy:
## Summary:

We want to avoid using `as`, as it ignores type safety. If we need to, we should use `satisfies`.

Adding a lint rule here to discourage `as` and inform people they can use `satisfies`.

Issue: none

## Test plan:
`pnpm lint`


jeremywiebe (approved):
Thank-you for adding this!!

PR #3566, opened 2026-05-01
[ColorSync] Convert matcher font and colors to semantic tokens

Myranae:
## Summary
Part of the ColorSync migration (LEMS-3487) to align Perseus widget colors with the design system's semantic token layer. Adds initial-state regression stories for Chromatic visual coverage, establishing a baseline for future Matcher and Sortable migrations. No interactions stories here as interactions will be covered by Sortable if possible to test. Also adds test data generators for Matcher since there weren't any before and other widgets use generators.

Issue: LEMS-3501

## Test Plan
- [x] Chromatic diffs show only intentional color changes against the baseline
- [x] All checks pass


Myranae:
@claude review once

.claude/artifacts/implementation/instructions/audit-widget.md
```diff
@@ -0,0 +1,35 @@
...
+grep -r "color\." packages/perseus/src/widgets/[widget-name]/ --include="*.tsx" --include="*.ts" --include="*.css"
+
+# Find hardcoded hex values
+grep -r "#[0-9a-fA-F]\{3,6\}" packages/perseus/src/widgets/[widget-name]/
+
+# Find hardcoded rgb(a) values
+grep -r "rgba?\([^)]+\)" packages/perseus/src/widgets/[widget-name]/
+
+# Check 4 font attributes
+grep -rE "fontSize|fontWeight|lineHeight|fontFamily" packages/perseus/src/widgets/free-response/
```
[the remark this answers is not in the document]
Myranae: Working on the first thing. The second is based on my own doc structure and I forgot to copy oover the font token mapping reference file.

packages/perseus/src/widgets/matcher/__docs__/matcher-initial-state-regression.stories.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+import type {PerseusMatcherWidgetOptions} from "@khanacademy/perseus-core";
+import type {Meta, StoryObj} from "@storybook/react-vite";
+
+const MatcherWidget = getWidget("matcher")!;
+
+const meta: Meta<typeof MatcherWidget> = {
+    title: "Widgets/Matcher/Visual Regression Tests/Initial State",
+    component: MatcherWidget,
+    tags: ["!autodocs"],
```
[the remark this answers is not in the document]
Myranae: @nishasy @mark-fitzgerald Is this a thing? It does seem like other files have this, so adding it.

packages/perseus/src/widgets/matcher/__docs__/matcher-initial-state-regression.stories.tsx
```diff
@@ -0,0 +1,96 @@
...
+// Verifies the default state with column labels: both the vertical column
+// divider (borderLeft) and the header row underline (borderBottom) use #444,
+// and fontWeight: "inherit" is applied to the label cells.
+export const DefaultWithLabels: Story = {
+    decorators: [matcherRendererDecorator],
+    args: sharedArgs,
+};
+
+// Verifies the no-labels state: header row is hidden so only the vertical
+// column divider (borderLeft using #444) is visible.
```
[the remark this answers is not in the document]
Myranae: Fixed

packages/perseus-core/src/index.ts
```diff
@@ -370,6 +370,11 @@ export {
     generateLabelImageOptions,
     generateLabelImageWidget,
 } from "./utils/generators/label-image-widget-generator";
+/** @hidden */
```
Myranae, line 373: Claude mentioned this is here to keep these out of documentation, but I'm not exactly sure when to use this and when not to (and it wasn't sure either). I think this relates to our JSdoc documentation method? Anyone know?

Sounds like all generators except these three are hidden, so maybe these are the outliers:
  - generateInteractiveGraphWidget (and all its graph type helpers)                                   
  - generateNumericInputWidget                                                                        
  - generateRadioWidget

packages/perseus/src/widgets/matcher/__docs__/matcher-initial-state-regression.stories.tsx
```diff
@@ -0,0 +1,96 @@
...
+            "Average global temperatures will rise",
+        ],
+        right: [
+            "Medium-sized stars exist for ~10 billion years",
+            "The current trajectory of tectonic plate movement",
+            "Rapid escalation of greenhouse gas emissions",
+        ],
+        orderMatters: false,
+        padding: true,
+    } satisfies Partial<PerseusMatcherWidgetOptions>,
```
mark-fitzgerald: Could this leverage the `sharedArgs` object, and be something like:
```
export const WithoutLabels: Story = {
    decorators: [matcherRendererDecorator],
    args: {
        ...sharedArgs,
        labels: ["", ""],
    }
}
```

packages/perseus/src/widgets/matcher/matcher.tsx
```diff
@@ -286,7 +287,7 @@ export default {
 } satisfies WidgetExports<typeof WrappedMatcher>;
 
 const padding = 5;
-const border = "1px solid #444";
+const border = `1px solid ${semanticColor.core.border.neutral.strong}`;
```
mark-fitzgerald: Since we are generating a style as a string instead of an object, perhaps we should just reference the CSS variable directly? I'm thinking this will simplify the conversion from Aphrodite in a later phase.
```
const border = `var(--wb-border-width-thin) solid var(--wb-semanticColor-core-border-neutral-strong}`;
```

mark-fitzgerald (approved):

packages/perseus/src/widgets/matcher/__docs__/matcher-initial-state-regression.stories.tsx
Myranae: Done!

packages/perseus/src/widgets/matcher/matcher.tsx
Myranae: Done! Updating docs to have this happen for the last couple ones too.

packages/perseus/src/widgets/matcher/__docs__/matcher-initial-state-regression.stories.tsx
```diff
@@ -0,0 +1,85 @@
...
+// are deterministic and don't depend on async TeX rendering timing.
+export const WithTexLabels: Story = {
+    decorators: [matcherRendererDecorator],
+    args: {
+        labels: ["$f(x)$", "$f'(x)$"],
+        left: ["f(x) = 1/x", "f(x) = 1/x²", "f(x) = 1/x³"],
+        right: ["f'(x) = -1/x²", "f'(x) = -2/x³", "f'(x) = -3/x⁴"],
+        orderMatters: false,
+        padding: true,
+    } satisfies Partial<PerseusMatcherWidgetOptions>,
```
Myranae: @mark-fitzgerald I'm thinking I could just replace this with the sharedArgs too. Though the content in shared args isn't relevant, we only care about the labels in this story anyway. What do you think? Last improvement and I'll move onto sorter 😂

mark-fitzgerald: I like how this example shows the use of TeX throughout, not just in the labels. I realize that the styling targets the header, but it would probably be helpful to have TeX in the body to make sure that future styling changes aren't too broad. I recommend keeping this as is. Perhaps change the title to just "WithTeX", and add a comment in the code that says this is primarily for the header.

Myranae: There is no TeX in the content. It is just text. The TeX in the cards was makes the story flakey as the spacing is not stable. It changes when it shouldn't change. Claude explained it was because of timing with the snapshots and how the spacing is calculated. "The flakiness is about when Chromatic takes the screenshot relative to the onMeasure height measurement cycle." The below screenshot shows the TeX story spacing (with actual TeX) changing even though nothing in this story was changed (text in another story was changed to cause the snapshots to recapture).

image

So, I think I'll probably keep the TeX out of the content for now, I can either keep the card content as TeX like text or use the content in the shared args. Or, up for other suggestions as well :) ... Actually, I'll take out the fake TeX. I think it's confusing and unhelpful. More so than if it was just not TeX in the content.

Myranae: Added the TeX back with a delay and all is well. Thanks!

PR #3568, opened 2026-05-01
Remove unused paths field from LinterContextProps and contentPaths prop from editors

jeremywiebe:
## Summary:

While working with Ben today on some preview-related refactoring, we started asking what some of the LinterContext values were for. Well, turns out `paths` is completely unused. A few places set it, but no code looks at it, so we can remove it. 
This propogates out to the `contentPaths` props on the `ArticleEditor`, `EditorPage`, and `HintEditor` components. `contentPaths` is used to set up `paths` on linter contexts, but no client passes it down, and because `paths` is dead, there's no point leaving the prop here.

Issue: LEMS-4083 

## Test plan:

`pnpm test`
`pnpm typecheck`

Run storybook and make sure the previews for EditorPage work!

### DIFF
--- a/.changeset/quiet-paths-vanish.md
+++ b/.changeset/quiet-paths-vanish.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/perseus-linter": major
+"@khanacademy/perseus-editor": major
+---
+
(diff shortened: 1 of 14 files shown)


benchristel (approved):
LGTM!

PR #3569, opened 2026-05-02
Narrow preview-bridge linterContext to only used values

jeremywiebe:
## Summary:

Following after #3568 which removed some dead code around preview and linter context, I found more things we can trim. 

The preview-bridge message types in `packages/perseus-editor/src/preview/message-types.ts` had linterContext typed as the full `LinterContextProps`, which includes `contentType`, `highlightLint`, and `stack`. In practice, the editors only ever set `contentType` and `highlightLint`; `stack` is built up entirely on the receiver side by `pushContextStack()` calls as the renderer tree descends. The type was wider than the data being sent and makes it confusing to understand.

This PR narrows `linterContext` on the three `*PreviewData` shapes to a local `PreviewLinterContext` which is built using the `LinterContextProps` type to reflect reality. There is no runtime change: the editors already only populated those two fields, and `pushContextStack` already tolerates a missing stack. 

Issue: LEMS-4083

## Test plan:

`pnpm typecheck`
`pnpm test`

Run storybook and check through the various editor pages to ensure the preview still loads.  Add a 1st-level heading (which is a linter violation) to the **Question** area (`# Hello`) and then toggle the "Style Warnings" button (bottom-right corner of the page with the eyeball icon). You should see the lint warning disappear.

### DIFF
--- a/packages/perseus-editor/src/preview/message-types.ts
+++ b/packages/perseus-editor/src/preview/message-types.ts
@@ -9,17 +9,17 @@ import type {
     PerseusArticle,
     PerseusItem,
 } from "@khanacademy/perseus-core";
-import type {LinterContextProps} from "@khanacademy/perseus-linter";
 
 /**
  * The subset of `LinterContextProps` that's meaningful to send across the
  * preview bridge. The receiver wraps this in `pushContextStack(...)` before
  * passing it to the renderer, which fills in the `stack`.
  */
-type PreviewLinterContext = Pick<
-    LinterContextProps,
-    "contentType" | "highlightLint"
->;
+// TODO(jeremy): Remove this type and inline it into the messages that use it.
+type PreviewLinterContext = {
+    contentType: string;
+    highlightLint: boolean;
+};
 
 /**
  * Constant identifier for all Perseus preview messages
--- a/packages/perseus-editor/src/preview/message-types.ts
+++ b/packages/perseus-editor/src/preview/message-types.ts
@@ -12,6 +12,16 @@ import type {
 import type {LinterContextProps} from "@khanacademy/perseus-linter";
 
 /**
+ * The subset of `LinterContextProps` that's meaningful to send across the
+ * preview bridge. The receiver wraps this in `pushContextStack(...)` before
+ * passing it to the renderer, which fills in the `stack`.
+ */
+type PreviewLinterContext = Pick<
+    LinterContextProps,
+    "contentType" | "highlightLint"
+>;
+
+/**
  * Constant identifier for all Perseus preview messages
  */
 export const PREVIEW_MESSAGE_SOURCE = "perseus-preview" as const;
@@ -31,7 +41,7 @@ export type QuestionPreviewData = {
     apiOptions: APIOptions;
     initialHintsVisible: number;
     device: DeviceType;
-    linterContext: LinterContextProps;
+    linterContext: PreviewLinterContext;
     reviewMode?: boolean;
     legacyPerseusLint?: ReadonlyArray<string>;
     problemNum?: number;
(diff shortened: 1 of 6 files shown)


packages/perseus-editor/src/preview/preview-data-sanitizer.test.ts
```diff
@@ -46,7 +46,6 @@ describe("sanitizePreviewData", () => {
                 linterContext: {
                     contentType: "exercise",
                     highlightLint: false,
-                    stack: [],
```
jeremywiebe, line 49: Stack is only used when we actually call the linter.  This happens _inside_ the `Renderer` _inside_ the preview iframe.  It starts as an empty array when we call the linter, so there's no point in passing this from the Editor into the preview iframe.

packages/perseus-editor/src/preview/message-types.ts
```diff
@@ -11,6 +11,16 @@ import type {
...
 
+/**
+ * The subset of `LinterContextProps` that's meaningful to send across the
+ * preview bridge. The receiver wraps this in `pushContextStack(...)` before
+ * passing it to the renderer, which fills in the `stack`.
+ */
+type PreviewLinterContext = Pick<
+    LinterContextProps,
+    "contentType" | "highlightLint"
+>;
```
benchristel: Could we write the specific types inline here? I'd rather not have to jump between multiple files to see what type `PreviewLinterContext` actually is. The duplication is not harmful because the type system ensures the types stay consistent.

```suggestion
type PreviewLinterContext = {
    contentType: string;
    highlightLint: boolean;
};
```

I might also question whether this type should have `Linter` in the name at all. Since it isn't a whole `LinterContextProps`, it can't be directly passed to a Renderer. Perhaps these two fields should be inlined into the `*PreviewData` types.

benchristel (approved):
I left one suggestion inline, but I'd also be fine with this as-is. Thanks for the simplifications!

packages/perseus-editor/src/preview/message-types.ts
jeremywiebe: I like this. I think we can take this further in the future by having a "preview options" type that houses all the things that are out of band of the actual content we want to preview.   I'll start by inlining these fields as you suggest.

jeremywiebe: 🛑  Actually, I just remembered. This is one reason why I'm working in this preview system to begin with. We can't make this structural change (inlining these values) because there is code in our consuming apps that expects this structure in the iframe child page. 

The Preview NG work moves both sides of the bridge into Perseus so we can restructure iframe messages confidently, but until then we shouldn't re-work the message structures.

packages/perseus-editor/src/preview/message-types.ts
```diff
@@ -9,7 +9,17 @@ import type {
...
     PerseusItem,
 } from "@khanacademy/perseus-core";
-import type {LinterContextProps} from "@khanacademy/perseus-linter";
+
+/**
+ * The subset of `LinterContextProps` that's meaningful to send across the
+ * preview bridge. The receiver wraps this in `pushContextStack(...)` before
+ * passing it to the renderer, which fills in the `stack`.
+ */
+// TODO(jeremy): Remove this type and inline it into the messages that use it.
```
jeremywiebe, line 18: The reason I'm not doing it in _this_ PR is that we still haven't fully moved to the new preview system that moves all communication to components within this repo. As a result, we can't (easily) change the shape of the messages that cross the iframe bridge and changing them here would be risky. 

Once the Preview NG ([this](https://github.com/Khan/perseus/pull/3581) and related PRs) work lands, we can do this TODO!

PR #3570, opened 2026-05-04
Disallow NaN and Infinity in number parser

benchristel:
## Summary:
Content creators reported that the article editor was crashing because an image
widget had `scale: null`. This surprised us because the Perseus parsers don't
allow `null` for that field.  We think this null value got into the data via
division by zero -> NaN -> null, since `JSON.stringify()` converts NaN to null.

To make sure this issue doesn't reoccur, this PR disallows NaN and Infinity in
the `number` parser, and creates a new `notANumber` parser for situations where
we need to opt into NaN.

See the slack thread for historical context:

https://khanacademy.slack.com/archives/C099VM99YF4/p1777905116428849?thread_ts=1777660853.554009&cid=C099VM99YF4

Issue: none

## Test plan:

CI checks should pass.


packages/perseus-core/src/parse-perseus-json/general-purpose-parsers/not-a-number.ts
```diff
@@ -0,0 +1,14 @@
+import type {Parser} from "../parser-types";
+
+/**
+ * Returns success given NaN, and failure for any other value. This parser is
+ * needed because NaN !== NaN, and therefore `constant(NaN)` will not work as
```
benchristel: On second thought... `constant` should use `Object.is`, not `===`. Then we could compare NaNs properly. Maybe I should change `constant` and delete this `notANumber` parser.

nishasy (approved):
Thanks for doing this

PR #3571, opened 2026-05-04
Mark exported editor component's `serialize()` functions as deprecated

jeremywiebe:
## Summary:

We've been working for a while to get rid of the `.serialize()` functions on the exported editor components. This is a legacy function that returns the current state of the edited item. However, a more React way is to provide that via a callback prop. We already have that prop in `onChange`. 

As of today (May 2026) no clients that use Perseus editors are using serialize anymore, so we can deprecate it and shortly delete it!

Issue: LEMS-4073

## Test plan:

:eyes:

### DIFF
--- a/packages/perseus-editor/src/editor.tsx
+++ b/packages/perseus-editor/src/editor.tsx
@@ -854,6 +854,11 @@ class Editor extends React.Component<Props, State> {
         }
     };
 
+    /**
+     * Returns the current version of the edited {@see {@link PerseusRenderer}}.
+     *
+     * @deprecated Use the {@see {@link Props.onChange}} prop instead.
+     */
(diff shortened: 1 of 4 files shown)


packages/perseus-editor/src/editor.tsx
```diff
@@ -133,6 +133,9 @@ type Props = Readonly<{
     warnNoWidgets: boolean;
     widgetIsOpen?: boolean;
     imageUploader?: ImageUploader;
+    /**
+     * asdf
```
handeyeco: This comment isn't very clear

handeyeco (approved):
Approving on the condition that you change the `asdf` comment

PR #3573, opened 2026-05-04
[ColorSync] Add Sorter regression stories and widget generator

Myranae:
## Summary
Establishes Chromatic baseline coverage for the Sorter widget before the Sortable component color migration. No font or color changes were needed in the widget itself — all styling lives in the Sortable component, which will be migrated in a follow-up PR. Also adds a `generateSorterWidget`/`generateSorterOptions` generator to `@khanacademy/perseus-core` for consistency with the rest of the widget test infrastructure.

Issue: LEMS-3500

## Test Plan
- [x] Chromatic snapshots for Sorter stories are approved
- [x] All checks pass


packages/perseus/src/widgets/__testutils__/sorter-renderer-decorator.tsx
```diff
@@ -0,0 +1,36 @@
...
+    generateTestPerseusRenderer,
+} from "@khanacademy/perseus-core";
+import * as React from "react";
+
+import QuestionRendererForStories from "./question-renderer-for-stories";
+
+import type {APIOptions} from "../../types";
+import type {PerseusSorterWidgetOptions} from "@khanacademy/perseus-core";
+
+export const sorterRendererDecorator = (
```
jeremywiebe: I'd recommend typing this decorator just for safety sake. 

```suggestion

import type {Decorator} from "@storybook/react-vite";

export const sorterRendererDecorator: Decorator = (
```

packages/perseus/src/widgets/sorter/__docs__/sorter-initial-state-regression.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+const horizontalItems = {
+    correct: ["20 km", "150 km", "500 km", "1200 km"],
+    padding: true,
+    layout: "horizontal",
+} satisfies Partial<PerseusSorterWidgetOptions>;
+
+// Verifies the default horizontal layout: cards arranged in a row with padding
+export const DefaultHorizontal: Story = {
+    decorators: [sorterRendererDecorator],
```
jeremywiebe: I _think_ you can move the `sorterRendererDecorator` up into the `meta` if you want.  Then you don't have to re-state it for every story.

packages/perseus/src/widgets/sorter/__docs__/sorter-initial-state-regression.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+};
+
+// Verifies horizontal layout with padding disabled: cards rendered without inner spacing
+export const NoPadding: Story = {
+    decorators: [sorterRendererDecorator],
+    args: {
+        correct: ["Item one", "Item two", "Item three"],
+        padding: false,
+        layout: "horizontal",
+    } satisfies Partial<PerseusSorterWidgetOptions>,
```
jeremywiebe, line 71: Why is this `satisfies` needed here?  The `Story` type should define it, shouldn't it?

packages/perseus/src/widgets/sorter/__docs__/sorter-initial-state-regression.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+            },
+        },
+        chromatic: {disableSnapshot: false, modes: themeModes},
+    },
+};
+export default meta;
+
+type Story = StoryObj<typeof SorterWidget>;
+
+const horizontalItems = {
```
jeremywiebe, line 31: Can we inline this? It looks like it's only used once and the other stories all define their `args` inline.

jeremywiebe (commented):
Just some drive-by comments. :) I leave it for someone from the ColorSync team to approve.

mark-fitzgerald (approved):
Looks good, assuming that Jeremy's comments are addressed.

packages/perseus/src/widgets/sorter/__docs__/sorter-initial-state-regression.stories.tsx
Myranae: It is used twice! :)

packages/perseus/src/widgets/sorter/__docs__/sorter-initial-state-regression.stories.tsx
Myranae: It looks like `SorterWidget` resolves to `React.ComponentType<any>`, so it's not actually working here to check types? 🤔 

It looks like other regression stories might also have this issue and just compensate by using `satisfies`. One option Claude suggests is:

<details>
<summary>
Way to remove satisfies in stories
</summary>
  Option 2 — Type meta directly and derive Story from it (Storybook-idiomatic):
  
 const meta: Meta<**PerseusSorterWidgetOptions**> = { component: SorterWidget, ... };
  type Story = StoryObj<typeof meta>;
  StoryObj<typeof meta> extracts the type parameter from Meta<T>, so args is again
  Partial<PerseusSorterWidgetOptions>.
</details>

But I would prefer something that fixes the root of the problem, which sounds like we would have to make `getWidget` generic maybe, but that will touch a lot of other files and is a larger change. 

So I can either keep this with the `satisfies` to match other stories, update the `meta` type directly, or... well, I guess I would want to do a separate PR for updating `getWidget`, so the first two options plus a ticket about `getWidget` maybe 😅

Myranae: Going to ping @mark-fitzgerald and @nishasy here because we'd want to make sure we update our stories based on this going forward (and probably do some fixing of already made stories).

packages/perseus/src/widgets/sorter/__docs__/sorter-initial-state-regression.stories.tsx
jeremywiebe: Doh! I was sure I looked! 😂

packages/perseus/src/widgets/sorter/__docs__/sorter-initial-state-regression.stories.tsx
```diff
@@ -0,0 +1,72 @@
+import {themeModes} from "../../../../../../.storybook/modes";
+import {getWidget} from "../../../widgets";
+import {sorterRendererDecorator} from "../../__testutils__/sorter-renderer-decorator";
+import {mobileDecorator} from "../../__testutils__/story-decorators";
+
+import type {PerseusSorterWidgetOptions} from "@khanacademy/perseus-core";
+import type {Meta, StoryObj} from "@storybook/react-vite";
+
+const SorterWidget = getWidget("sorter")!;
```
jeremywiebe, line 9: I don't think it's worth getting the widget this way.  It triggers going through a bunch of code when we know which widget we want (the Sorter! 😁 )

We _might_ need to export it from the implementation file, but I think that's actually preferable to using `getWidget` here... because I think I'd push back on making `getWidget` generic.

packages/perseus/src/widgets/sorter/__docs__/sorter-initial-state-regression.stories.tsx
Myranae: See comment about this a little further down, but will be making changes related to this for all regression stories in a separate PR.

packages/perseus/src/widgets/sorter/__docs__/sorter-initial-state-regression.stories.tsx
Myranae: I'm going to open a separate PR with my proposed solution to this situation, but it will be separate from this color/font conversion PR. :) Going to move forward with this as is and remove the satisfies part in the next PR.

jeremywiebe: That sounds like a good plan to me.  Thanks!

packages/perseus/src/widgets/matcher/__docs__/matcher-initial-state-regression.stories.tsx
```diff
@@ -66,7 +66,7 @@ export const WithoutLabels: Story = {
 export const WithTeX: Story = {
     decorators: [matcherRendererDecorator],
     parameters: {
-        chromatic: {delay: 500},
+        chromatic: {delay: 1000},
```
Myranae, line 69: Looks like the Chromatic snapshots are still flakey here. Going to be landing an actual fix for this soon.

PR #3574, opened 2026-05-05
Make `angleOffsetDeg` nullable

handeyeco:
## Summary:
We received `angleOffsetDeg: null` which made the parser made. This PR allows the parser to accept null for `angleOffsetDeg`.

See: https://khanacademy.slack.com/archives/CJDRXTGQ7/p1777927558183819

I don't really understand `angleOffsetDeg`. It doesn't get set directly in the editor and it seems to default to 0 everywhere that it's used: `The angle offset in degrees if an "angle" type. default: 0`. Somehow content was created where it was `null` but that's considered invalid by the parser, which is why we got this report from CP.

The proposed solution is to let `angleOffsetDeg` be `null` in the parser.

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/reducer/initialize-graph-state.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/reducer/initialize-graph-state.ts
@@ -127,7 +127,7 @@ export function initializeGraphState(
                 type: graph.type,
                 showAngles: Boolean(graph.showAngles),
                 coords: getAngleCoords({graph, range, step}),
-                angleOffsetDeg: Number(graph.angleOffsetDeg),
+                angleOffsetDeg: graph.angleOffsetDeg,
                 allowReflexAngles: Boolean(graph.allowReflexAngles),
                 snapDegrees: Number(graph.snapDegrees),
             };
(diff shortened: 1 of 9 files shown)


benchristel (approved):
Thanks for fixing this! Could you also update usages of `angleOffsetDeg` to remove handling of `undefined`? E.g.

```ts
angleOffsetDeg: Number(graph.angleOffsetDeg),
```

would become

```ts
angleOffsetDeg: graph.angleOffsetDeg,
```

Other than that, LGTM!

packages/perseus/src/widgets/interactive-graphs/reducer/initialize-graph-state.ts
```diff
@@ -127,7 +127,7 @@ export function initializeGraphState(
                 type: graph.type,
                 showAngles: Boolean(graph.showAngles),
                 coords: getAngleCoords({graph, range, step}),
-                angleOffsetDeg: Number(graph.angleOffsetDeg),
+                angleOffsetDeg: graph.angleOffsetDeg,
```
benchristel: I think this PR could be simplified by defaulting to zero here:

```suggestion
                angleOffsetDeg: graph.angleOffsetDeg ?? 0,
```

Then none of the interactive-graph business logic would need to change (the reducer, ai-utils, and AngleGraphState). The change would be limited to the data schema and parser, and this one initializer.

benchristel (approved):
LGTM! I left a suggestion for simplification inline.

PR #3578, opened 2026-05-05
[Preview NG] Add PreviewWithIframe component

jeremywiebe:
## Summary:

Previous PRs in this series:
- #3464
- #3467
- #3474
- #3492
- #3495
- #3499

---

Adds `PreviewWithIframe`, a new editor-side component that wraps a content preview iframe and communicates with it through `usePreviewController` and the typed postMessage protocol introduced earlier in this series (#3474, #3492, #3495, #3499). It's a component that the editor can use without needing to manage the iframe and controller hook directly. 

This PR is purely additive. It introduces a "spiritual successor" to the legacy `IframeContentRenderer`. No editor or preview page yet uses `PreviewWithIframe`; that flip happens in the next branch in the series, where the storybook preview page and all four editors switch over together. 

I chose to fork rather than rewrite `IframeContentRenderer` in place because `khan/frontend`'s devadmin imports the legacy component directly — rewriting in place would have required coordinating each Perseus PR with a frontend version bump. The fork lets the early Perseus PRs in this series land independently, and the legacy component will be deleted in a later cleanup pass once nothing imports it.

The long-term goal is to drop the iframe entirely once browser support for CSS `@container` queries reaches our floor; the `With*` suffix leaves room for a sibling `PreviewWithDiv` without renaming this one.

Issue: LEMS-3741

## Test plan:

- [x] `pnpm tsc`  - clean
- [x] `pnpm lint` - clean for the new files
- [x] `pnpm test`

### DIFF
--- a/packages/perseus-editor/src/preview-with-iframe.test.tsx
+++ b/packages/perseus-editor/src/preview-with-iframe.test.tsx
@@ -92,7 +92,6 @@ describe("PreviewWithIframe", () => {
             data: {
                 item: {
                     question: {content: "Q", widgets: {}, images: {}},
-                    answerArea: {calculator: false} as any,
                     hints: [],
                 },
                 apiOptions: {},
@@ -124,7 +123,6 @@ describe("PreviewWithIframe", () => {
             data: {
                 item: {
                     question: {content: "Q", widgets: {}, images: {}},
-                    answerArea: {calculator: false} as any,
                     hints: [],
                 },
                 apiOptions: {},
--- a/packages/perseus-editor/src/preview-with-iframe.test.tsx
+++ b/packages/perseus-editor/src/preview-with-iframe.test.tsx
@@ -36,58 +36,37 @@ describe("PreviewWithIframe", () => {
         expect(iframe).toBeInTheDocument();
         expect(iframe).toHaveAttribute("src", "/preview");
     });
+    it.each([true, false])(
+        "sets data-mobile attribute isMobile is set",
+        (isMobile: boolean) => {
+            render(
+                <PreviewWithIframe
+                    url="/preview"
+                    isMobile={isMobile}
+                    seamless={false}
+                />,
+            );
 
-    it("sets data-mobile to 'true' when isMobile is true", () => {
-        render(
-            <PreviewWithIframe
-                url="/preview"
-                isMobile={true}
-                seamless={false}
-            />,
-        );
+            const iframe = screen.getByTitle(/perseus-preview/);
+            expect(iframe.dataset.mobile).toBe(isMobile.toString());
+        },
+    );
 
-        const iframe = screen.getByTitle(/perseus-preview/);
-        expect(iframe.dataset.mobile).toBe("true");
-    });
+    it.each([true, false])(
+        "sets data-lint-gutter attribute when seamless is provided",
+        (seamless: boolean) => {
+            render(
+                <PreviewWithIframe
+                    url="/preview"
+                    isMobile={false}
+                    seamless={seamless}
+                />,
+            );
 
-    it("sets data-mobile to 'false' when isMobile is false", () => {
-        render(
-            <PreviewWithIframe
-                url="/preview"
-                isMobile={false}
-                seamless={false}
-            />,
-        );
-
-        const iframe = screen.getByTitle(/perseus-preview/);
-        expect(iframe.dataset.mobile).toBe("false");
-    });
-
-    it("sets data-lint-gutter to 'true' when seamless is true", () => {
-        render(
-            <PreviewWithIframe
-                url="/preview"
-                isMobile={false}
-                seamless={true}
-            />,
-        );
-
-        const iframe = screen.getByTitle(/perseus-preview/);
-        expect(iframe.dataset.lintGutter).toBe("true");
-    });
-
-    it("sets data-lint-gutter to 'false' when seamless is false", () => {
-        render(
-            <PreviewWithIframe
-                url="/preview"
-                isMobile={false}
-                seamless={false}
-            />,
-        );
-
-        const iframe = screen.getByTitle(/perseus-preview/);
-        expect(iframe.dataset.lintGutter).toBe("false");
-    });
+            const iframe = screen.getByTitle(/perseus-preview/);
+            expect(iframe.dataset.lintGutter).toBe(seamless.toString());
+        },
+    );
 
     it("delegates sendNewData to usePreviewController's sendData via ref", () => {
         const ref = React.createRef<PreviewWithIframeRef>();
@@ -115,7 +94,6 @@ describe("PreviewWithIframe", () => {
                 linterContext: {
                     contentType: "exercise",
                     highlightLint: false,
-                    paths: [],
                     stack: [],
                 },
             },
--- a/packages/perseus-editor/src/preview-with-iframe.test.tsx
+++ b/packages/perseus-editor/src/preview-with-iframe.test.tsx
@@ -153,31 +153,4 @@ describe("PreviewWithIframe", () => {
         const div = screen.getByTestId("preview-with-iframe-container");
         expect(div.style.height).toBe("500px");
     });
-
-    it("assigns unique iframe IDs across instances", () => {
-        const {unmount} = render(
-            <PreviewWithIframe
-                url="/preview"
-                isMobile={false}
-                seamless={false}
-            />,
-        );
-
-        const iframe1 = screen.getByTitle(/perseus-preview/);
-        const id1 = iframe1.dataset.id;
-        unmount();
-
-        render(
-            <PreviewWithIframe
-                url="/preview"
-                isMobile={false}
-                seamless={false}
-            />,
-        );
-
-        const iframe2 = screen.getByTitle(/perseus-preview/);
-        const id2 = iframe2.dataset.id;
-
-        expect(id1).not.toBe(id2);
-    });
 });
--- a/packages/perseus-editor/src/preview-with-iframe.test.tsx
+++ b/packages/perseus-editor/src/preview-with-iframe.test.tsx
@@ -86,7 +86,7 @@ describe("PreviewWithIframe", () => {
         },
     );
 
-    it("sends content using usePreviewController's sendData", () => {
+    it("sends data again when the content prop changes", () => {
         const data: PreviewContent = {
             type: "question",
             data: {
--- a/packages/perseus-editor/src/preview-with-iframe.test.tsx
+++ b/packages/perseus-editor/src/preview-with-iframe.test.tsx
@@ -1,10 +1,11 @@
+import {ApiOptions} from "@khanacademy/perseus";
 import {render, screen} from "@testing-library/react";
 import * as React from "react";
 
 import PreviewWithIframe from "./preview-with-iframe";
+import {clone} from "./testing/object-utils";
 
 import type {PreviewContent} from "./preview/message-types";
-import type {PreviewWithIframeRef} from "./preview-with-iframe";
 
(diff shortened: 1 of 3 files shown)


packages/perseus-editor/src/preview-with-iframe.tsx
```diff
@@ -0,0 +1,88 @@
...
+            [sendData],
+        );
+
+        // Update container height based on iframe content height
+        let containerHeight = "100%";
+        if (!props.seamless) {
+            containerHeight = "100%";
+        } else if (height !== null) {
+            containerHeight = `${height}px`;
+        }
```
benchristel:
```suggestion
        let containerHeight = 100%;
        if (props.seamless && height != null) {
            containerHeight = `${height}px`;
        }
```

packages/perseus-editor/src/preview-with-iframe.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+        render(
+            <PreviewWithIframe
+                url="/preview"
+                isMobile={false}
+                seamless={false}
+            />,
+        );
+
+        const div = screen.getByTestId("preview-with-iframe-container");
+        expect(div.style.height).toBe("100%");
```
benchristel: It would be nice if we could avoid using test IDs... could we assert that the root element returned from `render` has `height: 100%`?

packages/perseus-editor/src/preview-with-iframe.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+                initialHintsVisible: 0,
+                linterContext: {
+                    contentType: "exercise",
+                    highlightLint: false,
+                    stack: [],
+                },
+            },
+        };
+
+        ref.current?.sendNewData(data);
```
benchristel: We probably want:

```suggestion
        invariant(ref.current);
        ref.current.sendNewData(data);
```

otherwise, we'll never call `sendNewData` if the ref isn't set, and the test failure message won't be helpful.

packages/perseus-editor/src/preview-with-iframe.tsx
```diff
@@ -0,0 +1,88 @@
...
+        const containerRef = React.useRef<HTMLDivElement>(null);
+        const iframeRef = React.useRef<HTMLIFrameElement>(null);
+
+        const {sendData, height} = usePreviewController(iframeRef);
+
+        // Expose sendNewData method via ref
+        React.useImperativeHandle(
+            ref,
+            () => ({
+                sendNewData: (data: PreviewContent) => {
```
benchristel: Are we aiming for compatibility with the old `IframeContentRenderer` component? It seems like maybe not since the props are different. If not, could we avoid the imperative interface, and use a prop to pass the data into this component and a useEffect to send it to the preview controller?

Note that the exercise editor in frontend calls sendNewData on `UNSAFE_componentWillReceiveProps`. So I don't see any reason not to make the data a prop here.

packages/perseus-editor/src/preview-with-iframe.tsx
```diff
@@ -0,0 +1,88 @@
...
+        return (
+            <div
+                ref={containerRef}
+                data-testid="preview-with-iframe-container"
+                style={{width: "100%", height: containerHeight}}
+            >
+                <iframe
+                    ref={iframeRef}
+                    title={`perseus-preview`}
+                    data-mobile={props.isMobile ? "true" : "false"}
```
benchristel: Could be simplified to

```suggestion
                    data-mobile={String(props.isMobile)}
```

packages/perseus-editor/src/preview-with-iframe.tsx
```diff
@@ -0,0 +1,88 @@
...
+                    // The seamless prop is the same as the "nochrome" prop that
+                    // gets passed to DeviceFramer. If it is set, then we're going
+                    // to be displaying editor previews and want to leave some room
+                    // for lint indicators in the right margin.
```
benchristel: It feels like this logic is too coupled to the current callers of this component. Sizing the iframe to fit its content is a different concern from showing a lint gutter. Maybe they should be independently controllable.

packages/perseus-editor/src/preview-with-iframe.tsx
```diff
@@ -0,0 +1,88 @@
...
+                    data-lint-gutter={props.seamless ? "true" : "false"}
```
benchristel:
```suggestion
                    data-lint-gutter={String(props.seamless)}
```

packages/perseus-editor/src/preview-with-iframe.tsx
```diff
@@ -0,0 +1,88 @@
...
+    seamless: boolean;
+};
+
+export type PreviewWithIframeRef = {
+    sendNewData: (data: PreviewContent) => void;
+};
+
+const PreviewWithIframe = React.forwardRef<PreviewWithIframeRef, Props>(
+    (props, ref) => {
+        const containerRef = React.useRef<HTMLDivElement>(null);
```
benchristel: `containerRef` is not used.


```suggestion
```

packages/perseus-editor/src/preview-with-iframe.tsx
```diff
@@ -0,0 +1,88 @@
...
+        } else if (height !== null) {
+            containerHeight = `${height}px`;
+        }
+
+        return (
+            <div
+                ref={containerRef}
+                data-testid="preview-with-iframe-container"
+                style={{width: "100%", height: containerHeight}}
+            >
```
benchristel: Do we need this wrapper div? Could we just render the iframe?

packages/perseus-editor/src/preview-with-iframe.tsx
```diff
@@ -0,0 +1,88 @@
...
+import type {PreviewContent} from "./preview/message-types";
+
+type Props = {
+    // The URL that the iframe should load
+    url: string;
+    // Whether the iframe should be displaying in mobile mode
+    isMobile: boolean;
+    // Whether to make the iframe's height match its content's height,
+    // used to prevent scrolling inside the iframe.
+    seamless: boolean;
```
benchristel: These could be JSDoc comments for better editor integration.


```suggestion
    /** The URL that the iframe should load */
    url: string;
    /** Whether the iframe should be displaying in mobile mode */
    isMobile: boolean;
    /**
     * Whether to make the iframe's height match its content's height,
     * used to prevent scrolling inside the iframe.
     */
    seamless: boolean;
```

packages/perseus-editor/src/preview-with-iframe.tsx
```diff
@@ -0,0 +1,88 @@
...
+ * mobile breakpoints based on media queries work (because the body of the
+ * document is not the body of the editor. To make this work, this component
+ * renders an iframe and communicates with it via the `usePreviewController`
+ * hook and its use of postMessage.
+ *
+ * This is the new preview iframe component, replaces `IframeContentRenderer`
+ * once all editors have migrated. It exists alongside the legacy component so
+ * that consumers can keep using `IframeContentRenderer` until they are all
+ * ported over.
+ */
```
benchristel: Maybe this comment should be on the `PreviewWithIframe` component? Both for tooltip discoverability, and because Goland collapses the imports and everything above them by default.

packages/perseus-editor/src/preview-with-iframe.test.tsx
benchristel: If we get rid of the container div and just have an iframe, we could also find the element by title.

packages/perseus-editor/src/preview-with-iframe.tsx
jeremywiebe: Oh, I like this.  In fact, I think this can just become a ternary with a `const`.

packages/perseus-editor/src/preview-with-iframe.test.tsx
jeremywiebe: Good idea!

packages/perseus-editor/src/preview-with-iframe.tsx
jeremywiebe: Nope, it can be removed! Thanks for noticing!

packages/perseus-editor/src/preview-with-iframe.tsx
jeremywiebe: Right!  I'll do the same change on `props.seamless`!

packages/perseus-editor/src/preview-with-iframe.tsx
jeremywiebe: Agreed.  I noticed in our codebase there's alot of components that are documented at the top of the file, which, like you point out, means they aren't in any Goland/VSCode intelligence tooltips.  I've been moving them when I see them... thanks for catching this one!

packages/perseus-editor/src/preview-with-iframe.tsx
jeremywiebe: 😅  I was able to refactor away the ref entirely! Thanks!

packages/perseus-editor/src/preview-with-iframe.tsx
jeremywiebe: I agree. I'll push fixing this up to a new PR so I don't trigger too many new changes in this PR.

jeremywiebe:
@benchristel addressed your comments (thank-you).  Ready for a final review, i think!

packages/perseus-editor/src/preview-with-iframe.test.tsx
```diff
@@ -67,24 +86,43 @@ describe("PreviewWithIframe", () => {
...
                 seamless={false}
+                content={data}
             />,
         );
 
-        const data: PreviewContent = {
+        expect(mockSendData).toHaveBeenCalledWith(data);
+    });
+
+    it("sends content using usePreviewController's sendData", () => {
```
benchristel, line 118: This test has the same title as the one above. It could be `sends data again when the content prop changes`.

packages/perseus-editor/src/preview-with-iframe.tsx
```diff
@@ -1,87 +1,75 @@
...
-            () => ({
-                sendNewData: (data: PreviewContent) => {
-                    sendData(data);
-                },
-            }),
-            [sendData],
-        );
+    React.useEffect(() => {
+        sendData(props.content);
+    }, [sendData, props.content]);
```
benchristel, line 56: This is going to call `sendData` every time the identity of `props.content` changes, which might be every render. I don't think that's necessarily a problem, because the old iframe preview system does something similar, updating the iframe every time new props are received. Just wanted to point it out.

packages/perseus-editor/src/preview-with-iframe.test.tsx
jeremywiebe: Fixed!

packages/perseus-editor/src/preview-with-iframe.tsx
jeremywiebe: This is good to note. This could actually be problematic in two ways:

  1. we "over send" updates due to this being an identity comparison 
  2. we miss sending updates if the `props.content` object is mutated in-place

I think there are two things we could do: 

  1. we could do a deep-object comparison (maybe using `JSON.stringify()`) or
  2. we could debounce sending updates (this solution feels a bit hacky though)

I'm going to add a note above this to this effect (no pun intended 😂) and defer doing anything about it right now.

jeremywiebe: I guess we could wrap this in `React.memo()` 🤔

benchristel (approved):
LGTM!

PR #3582, opened 2026-05-06
[Image][GIF] Fixing overflow issue.

catandthemachines:
## Summary:
Fixing overflow issue with GIF images, the root cause being a result of the canvas styles not configured to properly respond to the parent FixedToResponsive component.

Before:
image

After:
image

Issue: LEMS-4075

## Test plan:
Run `pnpm test`

Repo: /?path=/story/widgets-image-widget-demo--gif-image

### DIFF
--- a/packages/perseus/src/components/gif-image.tsx
+++ b/packages/perseus/src/components/gif-image.tsx
@@ -298,6 +298,8 @@ const GifImage = (props: Props) => {
     // aspect-ratio derives it so the canvas scales proportionally.
     const displayHeight = !width && height ? height * scale : undefined;
 
+    const aspectRatio = width && height ? `${width} / ${height}` : undefined;
+
     return (
         <>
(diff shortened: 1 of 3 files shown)


packages/perseus/src/components/gif-image.tsx
```diff
@@ -314,7 +318,10 @@ const GifImage = (props: Props) => {
                 data-testid="gif-canvas"
                 style={{
                     width: width ? width * scale : undefined,
-                    height: height ? height * scale : undefined,
+                    maxWidth: "100%",
+                    aspectRatio:
+                        width && height ? `${width} / ${height}` : undefined,
```
nishasy: Nit: I have a personal preference for this line to be broken out into a new variable, maybe just below where you define `displayHeight`. But not necessary.

nishasy (approved):
Makes sense to me! I was able to confirm it works in storybook too.

PR #3583, opened 2026-05-06
[Image] | (CX) | Handle null width/height in image editor

nishasy:
## Summary:
There was a crash that happened due to a divide by zero error when the
width or height was zero.

Adding a check for that.

I also disabled the scale fields for images with invalid size, along
with a warning encouraging pressing the recalculate button.

Issue: https://khanacademy.atlassian.net/browse/LEMS-4080

## Test plan:
`pnpm jest packages/perseus-editor/src/widgets/image-editor/components/image-scale-input.tsx`

Storybook
- `/?path=/story/widgets-image-editor-demo--image-with-empty-size-with-scale-flag`

## Screenshots & demo

| Before | After |
| --- | --- |
| Screenshot 2026-05-05 at 6 37 16 PM | Screenshot 2026-05-05 at 6 40 07 PM |

https://github.com/user-attachments/assets/21553a20-82cd-4de8-8d36-2fecf9ade3dd

### DIFF
--- a/.changeset/thick-donuts-fly.md
+++ b/.changeset/thick-donuts-fly.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": patch
+---
+
+[Image] | (CX) | Handle null width/height in image editor
--- a/packages/perseus-editor/src/widgets/__docs__/image-editor.stories.tsx
+++ b/packages/perseus-editor/src/widgets/__docs__/image-editor.stories.tsx
@@ -129,7 +129,7 @@ export const GraphieImageWithScaleFlag: Story = {
 };
 
 /**
- * This Image widget editor has a zero-sized image.
+ * This Image widget editor has an image with missing size.
  */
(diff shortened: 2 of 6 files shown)


benchristel (approved):
LGTM! I left one question inline.

EDIT: okay, my inline comment apparently didn't post, because `"An internal error occurred, please try again.."` I was wondering if we should also validate that `width` and `height` are positive and real, i.e. not `NaN` or `Infinity`.

catandthemachines:
> LGTM! I left one question inline.
> 
> EDIT: okay, my inline comment apparently didn't post, because `"An internal error occurred, please try again.."` I was wondering if we should also validate that `width` and `height` are positive and real, i.e. not `NaN` or `Infinity`.

If you're able to add this @nishasy , I think it's worth while to future proof it.

catandthemachines (approved):
Looks good to me!

benchristel (approved):
LGTM!

catandthemachines (approved):

PR #3584, opened 2026-05-06
[Image][GIF] Adding aria-label to button in long description modal.

catandthemachines:
## Summary:
Our pause/play button in the long description modal does not have an aria-label, this is typically fine as screen readers will read the content of the button as the name. However, since we are dynamically changing the name of the button, that information isn't reliably communicated to assistive tech. To fix this it is best to apply an aria-label, so when it is changed it will more reliably get picked up by the tooling.

See this indepth article for reference: https://sarahmhigley.com/writing/playing-with-state/

Issue: LEMS-4077

## Test plan:
Run `pnpm test`

Visual test at: ?path=/story/widgets-image-widget-demo--gif-image


nishasy (approved):
Good catch

PR #3585, opened 2026-05-06
[Interactive Graph] Fix mobile Interactive Graph's asymptote drag not blurring previously focused point

ivyolamit:
## Summary:
- Programmatically focus the asymptote group when a drag starts so focus follows the element the user is interacting with, including on mobile.
- Touch input doesn't shift DOM focus the way a click does, so before this fix, dragging the asymptote on mobile left the previously focused movable point still showing its focus ring.
- Mirrors the existing pattern in useControlPoint (added in #2282) — the asymptote was the only draggable that didn't have it.

Issue: LEMS-4058

## Test plan:

### DIFF
--- a/.changeset/loud-rules-explode.md
+++ b/.changeset/loud-rules-explode.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+Fix mobile Interactive Graph's asymptote drag not blurring previously focused point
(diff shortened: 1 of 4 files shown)


ivyolamit:
| Before | After |
| ------ | ----- |
| <video src="https://github.com/user-attachments/assets/f5094c4a-b775-450c-aea2-208effaaa7f5"></video> | <video src="https://github.com/user-attachments/assets/f09ab377-47e2-4f00-b901-3bf48278781d"></video> |

ZND Link: https://prod-znd-260506-10953-1c5b5f5.khanacademy.org/internal-courses/test-everything/test-everything-1/te-interactive-graph/e/logarithm-graph

handeyeco (approved):

PR #3581, opened 2026-05-05
[Preview NG] Migrate Perseus to the new preview system

jeremywiebe:
## Summary:

*This PR is part of a series building a typed, hook-based preview system for the Perseus editor. The new system replaces the untyped `window.iframeDataStore` + raw `postMessage(string)` communication with structured, validated message passing via `usePreviewController` and `usePreviewPresenter` hooks. The new system is being built alongside the old one — no existing behavior changes until the final PR in the series flips the switch.*

Previous PRs in this series:
- #3464
- #3467
- #3474
- #3492
- #3495
- #3499
- #3578 

---

This is the Preview NG changeover PR! After this lands, Perseus will be exclusively using the new Preview communication system (no more legacy `iframeDataStore` + string-postMessage preview protocol). This means that all Storybook preview code now uses the new `usePreviewPresenter` hook, and all four editors (`EditorPage`, `ArticleEditor`, `HintEditor`, `ItemEditor`) switch from `IframeContentRenderer` to `PreviewWithIframe` (introduced in #3578).

The legacy `IframeContentRenderer` remains because some client's still use it directly. When this change is released, I will switch clients over as part of the Perseus release. The legacy component will be deleted in a later cleanup pass once that's done.

https://github.com/user-attachments/assets/a801bb4a-2cbd-4154-940f-09a2bd4037c5

NOTE: the issue of the keypad being clipped by the preview iframe is a longstanding issue and is unrelated to the changes in this PR. 

Issue: LEMS-3741

## Test plan:

- [x] `pnpm tsc` clean
- [x] `pnpm lint` clean for touched files
- [x] `pnpm test`
- [ ] Reviewer: open Storybook, navigate to the editor stories, and confirm each editor's preview renders content (question, hint, article single section, article-all). The behaviour should be identical to `main`.

PR #3588, opened 2026-05-06
[Preview NG] Replace deprecated ChangeHandler with typed onChange signatures

jeremywiebe:
## Summary:

Previous PRs in this series:
- #3464
- #3467
- #3474
- #3492
- #3495
- #3499
- #3578
- #3581 

--- 

`Editor` and `ArticleEditor` were using `ChangeHandler` (and `Changeable.ChangeableProps`), a deprecated signature whose `callback` and `silent` parameters were dead code throughout the codebase. This replaces both with concrete typed signatures so they are accurate and don't include data that the editor doesn't actually dispatch. 

**Note**: This change ripples in a few places worth flagging in review. Inside `ArticleEditor`, removing `Changeable.ChangeableProps` lets us drop the local `RendererProps` shim type that existed only to widen the parameter; the affected helpers (`_updateIssues`, `_sections`, `_handleEditorChange`, `_apiOptionsForSection`, `getSaveWarnings`) now use `PerseusRenderer` / `PerseusArticle` directly. In `editor.tsx`, two `props.onChange` call sites previously used the trailing `callback` parameter to position the cursor after a synchronous text replacement; that side-effect is dropped here because no caller actually wired it through, and the cursor positioning code wasn't reachable in practice.

Issue: LEMS-3741

## Test plan:

- [x] `pnpm tsc`
- [x] `pnpm test`
- [x] `pnpm lint`

### DIFF
--- a/packages/perseus-editor/src/__tests__/editor.test.tsx
+++ b/packages/perseus-editor/src/__tests__/editor.test.tsx
@@ -138,21 +138,17 @@ describe("Editor", () => {
         await userEvent.tab(); // blurring the input triggers onChange to be called
 
         // Assert
-        expect(changeFn).toHaveBeenCalledWith(
-            {
-                widgets: {
-                    "image 1": expect.objectContaining({
-                        type: "image",
-                        graded: true,
-                        options: expect.objectContaining({
-                            caption: "kittens",
-                        }),
+        expect(changeFn).toHaveBeenCalledWith({
+            widgets: {
+                "image 1": expect.objectContaining({
+                    type: "image",
+                    graded: true,
+                    options: expect.objectContaining({
+                        caption: "kittens",
                     }),
-                },
+                }),
             },
-            undefined,
-            undefined,
-        );
+        });
     });
 
     it("should not log a warning given a widget with an undefined key", () => {
--- a/packages/perseus-editor/src/article-editor.tsx
+++ b/packages/perseus-editor/src/article-editor.tsx
@@ -454,7 +454,7 @@ export default class ArticleEditor extends React.Component<Props, State> {
     /**
      * Returns the current version of the edited {@link PerseusArticle}.
      *
-     * @deprecated Use the {@link Props.onChange} prop instead.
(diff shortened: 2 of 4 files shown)


packages/perseus-editor/src/article-editor.tsx
```diff
@@ -65,7 +58,8 @@ type Props = DefaultProps & {
     previewURL: string;
     /** @deprecated `issues` has no effect. */
     issues?: Issue[];
-} & Changeable.ChangeableProps;
```
jeremywiebe, line 68: `ChangeableProps` is a really kludgey type that is the union of _all_ `onChange` props across the editors. Meaning we have no idea what a given editor's `onChange` will _really_ raise. Migrating to explicitly defined `onChange` handlers per-editor is much preferred.

packages/perseus-editor/src/article-editor.tsx
```diff
@@ -108,7 +102,7 @@ export default class ArticleEditor extends React.Component<Props, State> {
      */
     _updateIssues() {
         // Get sections array
-        const sections: ReadonlyArray<RendererProps> =
+        const sections: PerseusArticle =
```
jeremywiebe, line 105: Someday... I wonder if we could make `PerseusArticle` _just_ an array of `PerseusRenderer`s... instead of `PerseusRenderer | Array<PerseusRenderer>`.

packages/perseus-editor/src/__tests__/editor.test.tsx
```diff
@@ -138,21 +138,17 @@ describe("Editor", () => {
...
+                    type: "image",
+                    graded: true,
+                    options: expect.objectContaining({
+                        caption: "kittens",
                     }),
-                },
+                }),
             },
-            undefined,
-            undefined,
```
jeremywiebe, line 154: `onChange` no longer has two (ignored) parameters!

ivyolamit (approved):

mark-fitzgerald (approved):
So clean!

PR #3590, opened 2026-05-06
[Preview NG] Rename preview data field names and unify article-all apiOptions

jeremywiebe:
## Summary:

Previous PRs in this series:
- #3464
- #3467
- #3474
- #3492
- #3495
- #3499
- #3578
- #3588
- #3581 
- #3590 

-- 

This PR is a message-field cleanup — it restructures the preview data shapes now that the editor migration is in place (#3581).

  * The data-shape changes: `ArticlePreviewData.json` is renamed to `article` (and is now a single `PerseusRenderer` for the one section being edited, not an array)
  * `QuestionPreviewData.initialHintsVisible` is dropped (it was always `0` in practice)
  * `article-all` previews now use a new `ArticleAllPreviewData` (`{article: ReadonlyArray<PerseusRenderer>, apiOptions}`) with a single shared `apiOptions` instead of the previous per-section `ArticlePreviewData[]`
  * The presenter hook's returned `data` is also now named `content`, so consumers read `content` instead of the awkward `message.content.data`.
  * The preview message types now type their `apiOptions` as a new `SerializableApiOptions` defined next to the stripping logic in `sanitize-api-options.ts`) rather than the full `APIOptions`. This makes the types honest about what actually survives the `postMessage` structured-clone boundary.

Issue: LEMS-3741

## Test plan:

- [x] `pnpm tsc`
- [x] `pnpm lint`
- [x] `pnpm test`
- [ ] Reviewer: open Storybook, navigate to the editor previews, and confirm:
  - Question preview still renders (no visible hints; behavior matches main)
  - Article single-section preview renders content
  - Article-all preview renders all sections concatenated with shared apiOptions
  - Hint preview unchanged

### DIFF
--- a/packages/perseus-editor/src/preview/message-types.ts
+++ b/packages/perseus-editor/src/preview/message-types.ts
@@ -6,8 +6,8 @@
 import type {APIOptions, DeviceType} from "@khanacademy/perseus";
 import type {
     Hint,
-    PerseusArticle,
     PerseusItem,
+    PerseusRenderer,
 } from "@khanacademy/perseus-core";
 
 /**
@@ -39,7 +39,6 @@ interface PreviewMessageBase {
 export type QuestionPreviewData = {
     item: PerseusItem;
     apiOptions: APIOptions;
-    initialHintsVisible: number;
     device: DeviceType;
     linterContext: PreviewLinterContext;
     reviewMode?: boolean;
@@ -58,13 +57,21 @@ export type HintPreviewData = {
 };
 
 /**
- * Data for article section preview
+ * Data for a single article section preview (used in edit mode for one section)
  */
 export type ArticlePreviewData = {
+    article: PerseusRenderer;
     apiOptions: APIOptions;
-    json: PerseusArticle;
     linterContext: PreviewLinterContext;
     legacyPerseusLint?: ReadonlyArray<string>;
+};
+
+/**
+ * Data for article-all preview (used in preview mode to render all sections at once)
+ */
+export type ArticleAllPreviewData = {
+    article: ReadonlyArray<PerseusRenderer>;
+    apiOptions: APIOptions;
 };
 
 /**
@@ -74,7 +81,7 @@ export type PreviewContent =
     | {type: "question"; data: QuestionPreviewData}
     | {type: "hint"; data: HintPreviewData}
     | {type: "article"; data: ArticlePreviewData}
-    | {type: "article-all"; data: ArticlePreviewData[]};
+    | {type: "article-all"; data: ArticleAllPreviewData};
 
 // ---- Parent → Iframe messages ----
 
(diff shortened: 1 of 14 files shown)


packages/perseus-editor/src/preview/preview-data-sanitizer.test.ts
```diff
@@ -63,15 +63,14 @@ describe("sanitizePreviewData", () => {
             expect(result.data.apiOptions.customKeypad).toBe(true);
 
             // Non-serializable functions should be removed
-            expect(result.data.apiOptions.onFocusChange).toBeUndefined();
-            expect(result.data.apiOptions.answerableCallback).toBeUndefined();
-            expect(result.data.apiOptions.getAnotherHint).toBeUndefined();
-            expect(result.data.apiOptions.interactionCallback).toBeUndefined();
-            expect(result.data.apiOptions.trackInteraction).toBeUndefined();
+            expect("onFocusChange" in result.data.apiOptions).toBe(false);
```
jeremywiebe, line 60: Needed to change these assertions as these keys aren't on the `.apiOptions` type at all now, so it was either this or a bunch of `// @ts-expect` suppressions.

packages/perseus-editor/src/preview/message-types.ts
```diff
@@ -53,28 +52,36 @@ export type QuestionPreviewData = {
...
+    apiOptions: SerializableApiOptions;
     linterContext: PreviewLinterContext;
     legacyPerseusLint?: ReadonlyArray<string>;
 };
 
+/**
+ * Data for article-all preview (used in preview mode to render all sections at once)
+ */
+export type ArticleAllPreviewData = {
+    article: ReadonlyArray<PerseusRenderer>;
```
mark-fitzgerald, line 73: Should this be plural: `articles`?

mark-fitzgerald (approved):
Thanks for all the comments - super helpful!

packages/perseus-editor/src/preview/message-types.ts
jeremywiebe: No.  An Article in Perseus is (annoyingly) defined as `PerseusRenderer` _OR_ `PerseusRenderer[]`.  

https://github.com/Khan/perseus/blob/7bf9c46454af953213ac8bdceba0bbda9e31e70d/packages/perseus-core/src/data-schema.ts#L237-L241

So "article-all" is a way to preview all sections in an article.  Related to your question though, I'm wondering if the `ArticlePreviewData`'s field should be `section` instead.

https://github.com/Khan/perseus/blob/f4f4f54b7194f47858b29189894853febaba99ef/packages/perseus-editor/src/preview/message-types.ts#L66-L71

jeremywiebe: Yeah, I think I'm going to rename it to `article-section` instead of just `article`... the article-editor is definitive proof:

https://github.com/Khan/perseus/blob/d604b5d3ca3eab35e6f7bc80293af10284c710ec/packages/perseus-editor/src/article-editor.tsx#L284-L290

PR #3592, opened 2026-05-06
Change Item Splitting and Schema Change checks into real Github Checks

jeremywiebe:
## Summary

I've long been unimpressed with how cluttered our Perseus PRs have become. Each new thing we want to check becomes a comment, and that makes reviewing PRs difficult (for both author and reviewer).

This PR moves two of our checks (Item Splitting and Schema Change) into true GitHub Checks. They no longer post PR comments — they surface in the Checks area at the bottom of the PR alongside type checking, unit tests, etc.

When either check trips (i.e. the relevant artifact actually differs from the base branch), the job concludes as a `failure` and requires the author to apply an acknowledgement label (`schema-change-ack` or `item-splitting-change-ack` - see screenshot below) confirming the change has been reviewed. Adding the label re-runs only the affected check and flips it back to `success` ("Changes Acknowledged"); removing the label flips it back to failure. If a later commit reverts the change so the diff goes empty, the ack label is auto-removed too — so a future re-introduction of a diff will require a fresh ack rather than silently riding on a stale label.

image

Each check's pass/fail is the job's natural conclusion (no separate Checks-API check is created). The rich change summary — action items, ack call-to-action, collapsible diff — is rendered as a step summary on the workflow run page, reached via the check's "Details" link.

image

The two checks live in their own workflow (`pr-comparison-checks.yml`), separate from `node-ci.yml`. That's because they need to react to PR metadata events (`labeled` / `unlabeled`) — and if those events lived on `node-ci.yml`, every label toggle would also retrigger the rest of CI (lint, Cypress, build matrix, etc.). Splitting keeps the label-driven retriggers scoped.

## Issue

none

## Test plan

Verified end-to-end by introducing temporary probes (since reverted) on this branch:

- [x] Both checks render ✅ on a PR with no relevant diff (`PR Comparison Checks / Schema change`, `PR Comparison Checks / Item splitting`).
- [x] Adding a schema diff fails the `Schema change` job; clicking "Details" shows the rich step summary with the Content Platform action items, ack call-to-action, and collapsible diff.
- [x] Adding the `schema-change-ack` label re-runs *only* the schema job (not item-splitting, not the rest of CI) and flips the check back to ✅ (step summary heading: "Changes Acknowledged").
- [x] Reverting the schema diff restores ✅ "No Changes".
- [x] Removing the `-ack` label while a diff is still present fails the job again.
- [x] An item-splitting diff fails the `Item splitting` job.
- [x] Adding `item-splitting-change-ack` flips it back to ✅.
- [x] When a diff is reverted, the corresponding `-ack` label is auto-removed (and a future re-added diff requires a fresh ack).

### DIFF
--- a/.github/actions/branch-compare/action.yml
+++ b/.github/actions/branch-compare/action.yml
@@ -109,11 +109,18 @@ runs:
           if: always()
           env:
               GITHUB_TOKEN: ${{ inputs.github-token }}
+              ACK_LABEL_NAME: ${{ inputs.ack-label-name }}
           shell: bash
           run: |
               PR_NUMBER="${{ github.event.pull_request.number }}"
               if [ "${{ steps.compare-artifacts.outputs.diff-exit-code }}" == "0" ]; then
                   gh pr edit "$PR_NUMBER" --remove-label "${{ inputs.label-name }}" || true
+                  # Drop the ack label too: nothing to acknowledge anymore. If
+                  # a future commit reintroduces a diff, the author will need
+                  # to re-apply the label, which is the desired behaviour.
+                  if [ -n "$ACK_LABEL_NAME" ]; then
+                      gh pr edit "$PR_NUMBER" --remove-label "$ACK_LABEL_NAME" || true
+                  fi
               else
                   gh pr edit "$PR_NUMBER" --add-label "${{ inputs.label-name }}"
               fi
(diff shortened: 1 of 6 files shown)


mark-fitzgerald (approved):
I like having these as checks instead of comments. Thank you for making this change! The comments are super helpful, as well.

PR #3593, opened 2026-05-06
docs(changeset): Fix blue line on vector when graph becomes interactive after correct answer

SonicScrewdriver:
## Summary:
[grey-hair] Fix blue line on vector when graph becomes interactive after correct answer

Issue: LEMS-XXXX

## Test plan:


ivyolamit (approved):
This logic look logical to me  based on existing usage of interactiveColor.

PR #3597, opened 2026-05-07
Add workaround for `graded` during serialization

handeyeco:
## Summary:

The problem: in #3466 we added a switch for `graded`. Everything was just dandy except that we broke publishes. That's because we started passing `graded` down to widgets (which we needed so IG could change its behavior based on the `graded` flag) but then every widgets started getting `graded`. Because of the janky way we serialize, that started adding `graded` in the WidgetOptions of some widgets ([ticket](https://khanacademy.atlassian.net/browse/LEMS-4105) and [Slack convo](https://khanacademy.slack.com/archives/C01AZ9H8TTQ/p1778089642003609)). Rather than discard unneeded fields, the publish pipeline just explodes when it sees a field it doesn't expect (which is totally cool and awesome) and it wasn't expecting `graded`.

The best solution is to make data serialization opt-in vs opt-out (see [LEMS-4108](https://khanacademy.atlassian.net/browse/LEMS-4108)). However that's probably a medium risk change and we have a playtest today, so this change just skips around the denylist for `graded`.

**Why this should be more safe than the last PR:** in #3466 there were some red flags...namely that I had to add `graded` to a bunch of tests. That was reverted as part of the bugfix (#3589) and the fact that I didn't need to add those changes back in this PR to pass checks indicated that this doesn't have the same problem as the last attempt.

[LEMS-4108]: https://khanacademy.atlassian.net/browse/LEMS-4108?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/perseus-editor/src/components/widget-editor.tsx
+++ b/packages/perseus-editor/src/components/widget-editor.tsx
@@ -16,7 +16,8 @@ import type Editor from "../editor";
 import type {APIOptions} from "@khanacademy/perseus";
 import type {Alignment, PerseusWidget} from "@khanacademy/perseus-core";
 
-type WidgetEditorProps = {
+// exported for tests
+export type WidgetEditorProps = {
     // Unserialized props
     id: string;
     onChange: (
@@ -34,7 +35,8 @@ type WidgetEditorState = {
     widgetInfo: PerseusWidget;
 };
 
-const _upgradeWidgetInfo = (props: WidgetEditorProps): PerseusWidget => {
+// exported for tests
+export const _upgradeWidgetInfo = (props: WidgetEditorProps): PerseusWidget => {
     // We can't call serialize here because this.refs.widget
     // doesn't exist before this component is mounted.
     const filteredProps = excludeDenylistKeys(props);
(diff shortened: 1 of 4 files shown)


packages/perseus-editor/src/components/widget-editor.tsx
```diff
@@ -38,6 +38,15 @@ const _upgradeWidgetInfo = (props: WidgetEditorProps): PerseusWidget => {
...
     // doesn't exist before this component is mounted.
     const filteredProps = excludeDenylistKeys(props);
+
+    // This is circumventing an issue with excludeDenylistKeys;
+    // it removes `graded` from WidgetOptions.options (good)
+    // but it's also recursive so it removes `graded`
+    // from the higher-up WidgetOptions (bad)
+    // See: LEMS-4108 and https://khanacademy.slack.com/archives/C01AZ9H8TTQ/p1778089642003609
+    // eslint-disable-next-line no-restricted-syntax
+    (filteredProps as any).graded = props.graded;
```
ivyolamit, line 50: can we add a test coverage for this.

ivyolamit (approved):
This change looks logical, since it's specific removal of graded, it will less likely have other side effects.

PR #3598, opened 2026-05-07
Add function to convert input number options to numeric input options

benchristel:
## Summary:
The plan is to use this to convert input number widgets to numeric inputs at
runtime, for both display and scoring.

Based on [this code analysis](https://khanacademy.atlassian.net/wiki/spaces/LC/pages/edit-v2/4746543421).

Issue: LEMS-4082

## Test plan:

CI checks should pass.


SonicScrewdriver (approved):
Looks great!

PR #3599, opened 2026-05-07
[Preview NG] Support dual-protocol editor preview (support legacy mode)

jeremywiebe:
## Summary:

Previous PRs in this series:
- #3464
- #3467
- #3474
- #3492
- #3495
- #3499
- #3578
- #3588
- #3581 

---

This PR adds a parallel legacy preview controller hook to `PreviewWithIframe` so the component can drive both new typed-protocol preview frames and legacy frames (running an older Perseus that predates `usePreviewPresenter`) without code changes at the call site.

The new `useLegacyPreviewController` mirrors the API shape of `usePreviewController` (`{sendData, height, ...}`) but speaks the legacy wire protocol: parent assigns each iframe an id via `data-id`, the iframe posts that id back as a raw string when ready, parent writes the latest preview content into `window.iframeDataStore[id]` and posts the id back, and the iframe reports height as `{id, height}` postMessages (vs. the typed protocol's `{source: "perseus-preview", type: "height-update", height}`).

`PreviewWithIframe` now invokes both controllers unconditionally and fans `sendNewData` out to both `sendData` calls. The two protocols are mutually exclusive for any given iframe — the typed listener filters by `source: "perseus-preview"` and ignores raw strings / `{id, ...}` objects; the legacy listener filters by the iframe's id and ignores typed messages — so they don't interfere. Heights from either protocol surface as `typedHeight ?? legacyHeight`.

The legacy support is transitional. Once all downstream consumers have upgraded to a Perseus version that uses usePreviewPresenter, both this hook and the legacy `IframeContentRenderer` class component can be deleted. The hook's JSDoc and the component's file header both note the deletion trigger; this is also why the deferred Branch 9 (message-field rework) stays deferred — its renames would break the data shape the legacy receivers consume.

Issue: LEMS-3741

Test plan:

  * `pnpm test`
  * `pnpm tsc`
  * `pnpm lint`

### DIFF
--- a/.changeset/quiet-bridges-stand.md
+++ b/.changeset/quiet-bridges-stand.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus-editor": minor
+---
+
+`PreviewWithIframe` now also speaks the legacy `IframeContentRenderer` wire protocol (raw-string handshake + `window.iframeDataStore`) alongside the typed `usePreviewController` protocol, so editor consumers can drive preview frames from older perseus versions without code changes. This is transitional — will be removed once all clients can upgrade to a version of Perseus that supports the modernized preview system (aka `usePreviewPresenter`).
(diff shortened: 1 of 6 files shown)


jeremywiebe:
Closing for now. We thought one consumer needed the old preview protocol, but that consumer hides the EditorPage preview and so the point is moot.

PR #3600, opened 2026-05-07
Add `scorePerseusItemWithInputNumberAsNumericInput` for side-by-side testing

benchristel:
## Summary:
The plan is to call this function in the Perseus service when scoring,
side-by-side with the current scoring logic, and log any differences. If
converting input-number widgets to numeric-input results in a different score,
it means we messed up the conversion logic somewhere.

Issue: LEMS-4091

## Test plan:

CI checks should pass.

### DIFF
--- a/packages/perseus-score/src/score.ts
+++ b/packages/perseus-score/src/score.ts
@@ -1,3 +1,5 @@
+import {convertInputNumberOptionsToNumericInput} from "@khanacademy/perseus-core";
+
 import flattenScores from "./util/flatten-scores";
 import getScoreableWidgets from "./util/get-scoreable-widgets";
 import isWidgetScoreable from "./util/is-widget-scoreable";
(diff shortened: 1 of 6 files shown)


packages/perseus-score/src/score.ts
```diff
@@ -72,3 +95,54 @@ export function scoreWidgetsFunctional(
 
     return widgetScores;
 }
+
+// TODO(LEMS-4085): remove this function once the Input Number to Numeric
+//  Input project is complete.
```
benchristel, line 100: I took the approach of duplicating `scoreWidgetsFunctional` wholesale, instead of adding a parameter to the existing function. This makes the temporary code easier to remove later, but introduces duplication which might get out of sync. I'm not sure I made the right choice.

packages/perseus-score/src/score.ts
```diff
@@ -72,3 +95,54 @@ export function scoreWidgetsFunctional(
...
+        }
+
+        const validator = getWidgetValidator(widgetType);
+        const scorer = getWidgetScorer(widgetType);
+
+        // We do validation (empty checks) first and then scoring. If
+        // validation fails, it's result is itself a PerseusScore.
+        const score =
+            validator?.(userInput, widgetOptions, locale) ??
+            scorer?.(userInput, widgetOptions, locale);
```
benchristel, line 141: Conveniently, the `userInput` has the same format for input-number and numeric-input, so we don't have to convert it.

packages/perseus-score/src/score.ts
SonicScrewdriver: I think the approach you took makes sense! We should be the only ones affecting this score function right now, so I think we can be careful to make sure the two functions get out of sync while we're completing the project. Cleanup seems easier with the two separate functions.

SonicScrewdriver (approved):
Looks good to me! Knip is upset as it sees this as dead code, but once we fix that up we should be good to go!

PR #3602, opened 2026-05-07
Fix regression story arg types by replacing getWidget with typed Meta

Myranae:
## Summary:

- Replaces `getWidget("widget-name")!` (which returns `ComponentType<any>`) with `Meta<PerseusXxxWidgetOptions>` across all regression story files, giving story `args` proper TypeScript type checking
- Removes all `satisfies Partial<PerseusXxxWidgetOptions>` workarounds from story args — now redundant because `StoryObj<typeof meta>` types them correctly
- For shared args constants defined outside stories, replaces `satisfies` with a plain type annotation to prevent string literal widening

Issue: LEMS-4113

## Test plan:

- `pnpm tsc` passes with no errors
- `pnpm lint` passes with no errors
- Verified no story behavior changes — all rendering continues to go through existing decorators unchanged

### DIFF
--- a/.changeset/cool-dodos-dance.md
+++ b/.changeset/cool-dodos-dance.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/perseus": patch
+"@khanacademy/perseus-core": patch
+---
+
+Fix regression story typing
--- a/.changeset/cool-dodos-dance.md
+++ b/.changeset/cool-dodos-dance.md
@@ -1,6 +1,5 @@
 ---
 "@khanacademy/perseus": patch
-"@khanacademy/perseus-core": patch
 ---
 
 Fix regression story typing
--- a/packages/perseus/src/widgets/free-response/__docs__/free-response-initial-state-regression.stories.tsx
+++ b/packages/perseus/src/widgets/free-response/__docs__/free-response-initial-state-regression.stories.tsx
@@ -1,15 +1,12 @@
 import {themeModes} from "../../../../../../.storybook/modes";
-import {getWidget} from "../../../widgets";
 import {freeResponseRendererDecorator} from "../../__testutils__/free-response-renderer-decorator";
 import {rtlDecorator} from "../../__testutils__/story-decorators";
 
+import type {PerseusFreeResponseWidgetOptions} from "@khanacademy/perseus-core";
 import type {Meta, StoryObj} from "@storybook/react-vite";
 
-const FreeResponseWidget = getWidget("free-response")!;
-
-const meta: Meta<typeof FreeResponseWidget> = {
+const meta: Meta<PerseusFreeResponseWidgetOptions> = {
     title: "Widgets/Free Response/Visual Regression Tests/Initial State",
-    component: FreeResponseWidget,
     tags: ["!autodocs", "!manifest"],
     parameters: {
         docs: {
@@ -25,7 +22,7 @@ const meta: Meta<typeof FreeResponseWidget> = {
 };
 export default meta;
 
-type Story = StoryObj<typeof FreeResponseWidget>;
+type Story = StoryObj<typeof meta>;
 
 // Verifies the default state with a character count displayed using
 // semanticColor.core.foreground.neutral.default and font.size.small.
--- a/packages/perseus/src/widgets/image/__docs__/image-initial-state-regression.stories.tsx
+++ b/packages/perseus/src/widgets/image/__docs__/image-initial-state-regression.stories.tsx
@@ -3,7 +3,6 @@ import * as React from "react";
 import {themeModes} from "../../../../../../.storybook/modes";
 import {ApiOptions} from "../../../perseus-api";
 import {getFeatureFlags} from "../../../testing/feature-flags-util";
-import {getWidget} from "../../../widgets";
 import {imageRendererDecorator} from "../../__testutils__/image-renderer-decorator";
 import {
     mobileDecorator,
@@ -28,11 +27,8 @@ import {
     scienceImageCaption,
 } from "../utils";
 
+import type {PerseusImageWidgetOptions} from "@khanacademy/perseus-core";
 import type {Meta, StoryObj} from "@storybook/react-vite";
-
-const ImageWidget = getWidget("image")!;
-
-type Story = StoryObj<typeof ImageWidget>;
 
 const earthMoonImageCaption =
     "The Moon above Earth's horizon, captured by the International Space Station, [NASA](https://images.nasa.gov/details/iss071e515452)";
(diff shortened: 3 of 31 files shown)


Myranae:
@claude review once

packages/perseus/src/widgets/free-response/__docs__/free-response-initial-state-regression.stories.tsx
```diff
@@ -1,15 +1,12 @@
...
 
+import type {PerseusFreeResponseWidgetOptions} from "@khanacademy/perseus-core";
 import type {Meta, StoryObj} from "@storybook/react-vite";
 
-const FreeResponseWidget = getWidget("free-response")!;
-
-const meta: Meta<typeof FreeResponseWidget> = {
+const meta: Meta<PerseusFreeResponseWidgetOptions> = {
     title: "Widgets/Free Response/Visual Regression Tests/Initial State",
-    component: FreeResponseWidget,
```
Myranae, line 12: Looks like this property was used to help with storybook docs and argtype generation, but now, argtype generation is handled by the Meta type field. Since we are handling that differently, we don't need to import the widgets! (Based on Claude explanation so please let me know if that sounds wrong) Looks like this could actually be an improvement!

Documentation is not necessary for the visual regression stories because they all have tags: ["!autodocs", "!manifest"], which means automatic docs page generation is explicitly disabled.

<details>
<summary> Claude's Insight</summary>

- Storybook has two ways to know what goes in the controls panel: the `component` field (infer from the component's prop types) and the `Meta<T>` type parameter (use T directly). They serve the same purpose.
- Before the refactor: `component: getWidget("sorter")!` gave Storybook a `ComponentType<any>` — so no controls were generated from it. The controls panel was empty or useless.
- After the refactor: `Meta<PerseusSorterWidgetOptions>` tells Storybook exactly which fields exist (`correct`, `layout`, `padding`, etc.), their types, and optionally their JSDoc descriptions from the type definition. So the Storybook controls panel now has properly typed knobs for all the widget options.
- `StoryObj<typeof meta>` flows that same type down to each story's `args`, which is why TypeScript now enforces the args instead of accepting `any`.
</details.

packages/perseus/src/widgets/image/__docs__/image-initial-state-regression.stories.tsx
```diff
@@ -253,7 +250,6 @@ export const FloatLeftImageWithinArticle: Story = {
         backgroundImage: scienceImage,
         alt: scienceImageAlt,
         caption: scienceImageCaption,
-        alignment: "wrap-left",
```
Myranae, line 256: This information was already being passed to the stories via the decorator. Also, alignment is a top level widget field and not in the options, so the info from this property here was being lost before.

packages/perseus/src/widgets/sorter/__docs__/sorter-initial-state-regression.stories.tsx
```diff
@@ -50,12 +46,12 @@ export const DefaultVertical: Story = {
         ],
         padding: true,
         layout: "vertical",
-    } satisfies Partial<PerseusSorterWidgetOptions>,
+    },
 };
 
 // Verifies the horizontal layout on mobile: narrower margin between cards (8px vs 5px)
 export const MobileHorizontal: Story = {
-    decorators: [sorterRendererDecorator, mobileDecorator],
```
Myranae, line 58: These were previously pulled into the meta object but this was left before as I wasn't sure if it needed to be included here with mobileDecorator being added. Looks like it gets added to whatever is in meta if I understood correctly.

.changeset/cool-dodos-dance.md
```diff
@@ -0,0 +1,6 @@
+---
+"@khanacademy/perseus": patch
+"@khanacademy/perseus-core": patch
+---
+
+Fix regression story typing
```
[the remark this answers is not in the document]
Myranae: I think this is because this is a stacked PR! I will have to update the changeset once I land the base PR and rebase this one.

packages/perseus/src/widgets/image/__docs__/image-interactions-regression.stories.tsx
```diff
@@ -10,18 +9,16 @@ import {
     svgImage,
 } from "../utils";
 
+import type {PerseusImageWidgetOptions} from "@khanacademy/perseus-core";
 import type {Meta, StoryObj} from "@storybook/react-vite";
 
-const ImageWidget = getWidget("image")!;
-
-type Story = StoryObj<typeof ImageWidget>;
+type Story = StoryObj<typeof meta>;
```
nishasy: super duper teeny tiny nit: this `Story` type is after the Meta declaration in the other stories files

packages/perseus/src/widgets/label-image/__docs__/label-image-initial-state-regression.stories.tsx
```diff
@@ -65,7 +61,7 @@ export const DefaultUnanswered: Story = {
         ],
         multipleAnswers: false,
         hideChoicesFromInstructions: true,
-    } satisfies Partial<PerseusLabelImageWidgetOptions>,
```
nishasy, line 68: nice

packages/perseus/src/widgets/sorter/__docs__/sorter-initial-state-regression.stories.tsx
nishasy: Oh interesting! That's good to know. I think there might be some other stories where this could be helpful later.

nishasy (approved):
Looks good to me!

packages/perseus/src/widgets/free-response/__docs__/free-response-initial-state-regression.stories.tsx
jeremywiebe: I was going to push back, but I think Claude's reasoning works here... especially since we use a decorator to map the `args` for each story to an actual rendered UI.  So I'm +1 with this.  This feels like a nice way to provide the right options in the control panel without a lot of gymnastics!

jeremywiebe (commented):
Super.. thanks for re-working this Tamara!  

I wonder if this could become guidance for agents as part of a new Storybook skill?   Doesn't need to be part of this PR... just something to consider for future so we can guide the AI better when making new widget-level stories.

mark-fitzgerald (approved):

packages/perseus/src/widgets/image/__docs__/image-interactions-regression.stories.tsx
Myranae: Fixed!

.changeset/cool-dodos-dance.md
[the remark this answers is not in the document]
Myranae: Updated the changeset file to be only the perseus package!

PR #3603, opened 2026-05-07
[Radio] Remove feature flags

mark-fitzgerald:
## Summary:
The radio revitalization project is complete. This PR removes references to the feature flag that was used during development. It is no longer needed - only one version of the widget is now in place.

Issue: LEMS-2994

## Test plan:
1. All unit tests
1. Stories function normally

### DIFF
--- a/packages/perseus-core/src/feature-flags.test.ts
+++ b/packages/perseus-core/src/feature-flags.test.ts
@@ -2,13 +2,13 @@ import {isFeatureOn} from "./feature-flags";
 
 describe("isFeatureOn", () => {
     it("returns true when the flag is enabled", () => {
-        const props = {apiOptions: {flags: {"new-radio-widget": true}}};
-        expect(isFeatureOn(props, "new-radio-widget")).toBe(true);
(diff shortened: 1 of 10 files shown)


packages/perseus-core/src/feature-flags.test.ts
```diff
@@ -2,13 +2,13 @@ import {isFeatureOn} from "./feature-flags";
 
 describe("isFeatureOn", () => {
     it("returns true when the flag is enabled", () => {
-        const props = {apiOptions: {flags: {"new-radio-widget": true}}};
-        expect(isFeatureOn(props, "new-radio-widget")).toBe(true);
+        const props = {apiOptions: {flags: {"widget-feature-flag": true}}};
```
mark-fitzgerald, line 5: I changed this to a generic named flag so that we don't accidentally remove it in the future.

SonicScrewdriver: That's probably a good plan!

SonicScrewdriver (approved):
I love the amount of cleanup this let us do :)! Great work

PR #3604, opened 2026-05-07
fix(LEMS-4079): add condition to only trigger answerable callback if there is userinput

anakaren-rojas:
## Summary:
There is a second renderer nested in the popover that opens for the definition widget. This second renderer is setting an unanswerable state for the exercise when opened because definition isn't an answerable widget. 

This change adds a condition to check whether there is user input before firing the answerable callback. 

Issue: LEMS-4079

## Test plan:
ZND:
https://prod-znd-260507-10947-accbae9.khanacademy.org
1. Navigate to exercise with definition widget ([example](https://prod-znd-260507-10947-accbae9.khanacademy.org/ela/9th-grade-reading-vocabulary-indiana/x068d7167a2598a90:bridging-the-gap-9-indiana/x068d7167a2598a90:determining-central-ideas-9-indiana/e/central-ideas-9-indiana))
2. Answer question 
3. Click definition widget 

### Expected behavior 
`Check` button is not disabled at any point when opening the definition popover 

Behavior has been validated by SarahP (see [thread](https://khanacademy.slack.com/archives/C09V0V3PHN0/p1778536867802909?thread_ts=1778257476.527869&cid=C09V0V3PHN0))

### Screen recording
https://github.com/user-attachments/assets/718d4096-3d4c-4099-8011-d6fab9da1b86

### DIFF
--- a/packages/perseus/src/renderer.tsx
+++ b/packages/perseus/src/renderer.tsx
@@ -305,7 +305,9 @@ class Renderer
             );
         }
 
-        this.props.apiOptions?.answerableCallback?.(this._isAnswerable());
+        if (this.props.userInput) {
+            this.props.apiOptions?.answerableCallback?.(this._isAnswerable());
+        }
     }
 
     UNSAFE_componentWillReceiveProps(nextProps: Props) {
--- a/.changeset/real-donuts-return.md
+++ b/.changeset/real-donuts-return.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+sets condition to check for user input before setting answerable


packages/perseus/src/renderer.tsx
```diff
@@ -305,7 +305,9 @@ class Renderer
             );
         }
 
-        this.props.apiOptions?.answerableCallback?.(this._isAnswerable());
+        if (this.props.userInput) {
+            this.props.apiOptions?.answerableCallback?.(this._isAnswerable());
```
handeyeco, line 309: I don't see why this would be a problem 🤷 

Another approach though, remove `answerableCallback` when using Renderer in Definition. Something like:

``` TSX
<Renderer
    apiOptions={{
        ...this.props.apiOptions,
        answerableCallback: () => {}
    }}
    content={this.props.definition}
    widgets={this.props.widgets}
    strings={this.context.strings}
/>
```

This kind of makes me think we're being a little too careless with how we're passing down props. The Renderer in Definition probably doesn't need all of the stuff in APIOptions (as we see with `answerableCallback`).

anakaren-rojas: Maybe we could add a larger `cleanup API Options` ticket for the summer squash 🤔

catandthemachines (approved):
Dang what a wild find. Thanks for the fix, and +1 to a follow-up Jira ticket to clean up API options.

PR #3605, opened 2026-05-08
[Interactive Graph] Add pointLabels props for interactive graph types

ivyolamit:
## Summary:
Add pointLabels props for interactive graph related types this will be used for custom point labels.

- Adds optional `pointLabels?: string[]` to all 15 interactive `PerseusGraphType<X>` union members in `packages/perseus-core/src/data-schema.ts`
- Teaches the JSON parser at `packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts` to accept the new field
- **Pure data-shape change. No behavior, no rendering, no editor surface yet — those land in succeeding PRs 2–7.**

> **i18n note:** an earlier draft of this PR widened `srPointAtCoordinates`'s `num` parameter from `number` to `number | string`. Reverted — `num` is reserved file-wide for plural selection (see `strings.ts:8-9`, `:640-641`). PR 2 will introduce a separate string key (e.g., `srPointAtCoordinatesNamed`) for the named-point announcement instead of overloading `num`.

## Background — the bug overview

**Bug:** Interactive coordinate-plane questions ask the learner to plot a specific named point — e.g., *"Plot point T to complete the rectangle"* — but JAWS announces the point as *"Point 1 at zero comma zero"* instead of *"Point T at zero comma zero."* The screen-reader label is hard-coded to a numeric index across every interactive graph type, so it never matches the prompt and confuses screen-reader users.

**Solution:** Add an optional `pointLabels?: string[]` field to each interactive graph type's data schema. Content authors can supply a label per point (e.g., `["T"]`), and the graph uses it to build the screen-reader announcement instead of the default index. Existing content keeps working unchanged — authors opt in only when they want letters that match the prompt. This PR (PR 1 of 7) only adds the data shape; rendering, state, and editor surface land in PRs 2–7.

The full fix ships across **7 sequential PRs**:

- 👉🏼 **PR 1** — [Schema — no behavior. Only data-shape stop.](https://github.com/Khan/perseus/pull/3605)

## Schema-first rationale

1. **Decouples reviewer concerns.** This PR is reviewed for "does the type shape make sense across the union?" — no behavior to litigate.
2. **Web-app integration is one-shot.** Schema-only changes are additive type widenings. The web app updates its mirror types once.
3. **Avoids per-PR schema drift.** No window where some graph types have `pointLabels` and others don't.

The "ships dead code" concern is real but small — `pointLabels?: string[]` is observably a no-op until PR 2 reads it. Mitigated by landing PR 2 close behind.

## i18n notes

- This PR makes **no** changes to `strings.ts`. The named-point announcement string will be introduced in PR 2 as a separate key, so existing `srPointAtCoordinates` translations stay valid and the `num`-is-for-plurality convention is preserved.
- `pointLabels` itself is content data, not a translation string. If an English item ships with `pointLabels: ["T"]`, translators need to update both the prompt AND `pointLabels` if their locale uses different letter conventions. Will be called out in the rollout / release notes.

Issue: LEMS-3995

## Test plan

- [x] `pnpm tsc` — passes
- [x] `pnpm lint packages/perseus packages/perseus-core` — passes
- [x] `pnpm --filter perseus-core test` — 1047 tests pass (including `interactive-graph-widget.test.ts` and `parse-perseus-json-regression.test.ts`)
- [x] `pnpm test packages/perseus/src/widgets/interactive-graphs` — 1094 tests pass, 28 pre-existing skips
- [ ] Reviewer sanity-check: schema additions match the documented JSDoc verbatim across all 15 types

### DIFF
--- a/packages/perseus-core/src/data-schema.ts
+++ b/packages/perseus-core/src/data-schema.ts
@@ -1146,6 +1146,17 @@ export type PerseusGraphTypeAngle = {
      * the vertex, while rays BA and BC form the angle.
      */
     startCoords?: [Coord, Coord, Coord];
+    /**
+     * Custom screen-reader labels for each interactive point.
+     * If provided, pointLabels[i] replaces the default numeric "Point {i+1}"
+     * announcement. Type is string so authors can use letters (e.g., "T") to
+     * match question text, but the editor defaults to the numeric "1", "2", …
+     * — authors override only when they want a letter.
+     *
+     * Example: pointLabels: ["T"] → "Point T at 0 comma 0"
+     *          (instead of the default "Point 1 at 0 comma 0")
+     */
+    pointLabels?: string[];
 };
 
 export type PerseusGraphTypeCircle = {
@@ -1157,6 +1168,17 @@ export type PerseusGraphTypeCircle = {
         center: Coord;
         radius: number;
     };
+    /**
+     * Custom screen-reader labels for each interactive point.
+     * If provided, pointLabels[i] replaces the default numeric "Point {i+1}"
+     * announcement. Type is string so authors can use letters (e.g., "T") to
+     * match question text, but the editor defaults to the numeric "1", "2", …
+     * — authors override only when they want a letter.
+     *
+     * Example: pointLabels: ["T"] → "Point T at 0 comma 0"
+     *          (instead of the default "Point 1 at 0 comma 0")
+     */
+    pointLabels?: string[];
 };
 
 export type PerseusGraphTypeLinear = {
@@ -1165,6 +1187,17 @@ export type PerseusGraphTypeLinear = {
     coords?: CollinearTuple | null;
     /** The initial coordinates the graph renders with. */
     startCoords?: CollinearTuple;
+    /**
+     * Custom screen-reader labels for each interactive point.
+     * If provided, pointLabels[i] replaces the default numeric "Point {i+1}"
+     * announcement. Type is string so authors can use letters (e.g., "T") to
+     * match question text, but the editor defaults to the numeric "1", "2", …
+     * — authors override only when they want a letter.
+     *
+     * Example: pointLabels: ["T"] → "Point T at 0 comma 0"
+     *          (instead of the default "Point 1 at 0 comma 0")
+     */
+    pointLabels?: string[];
 };
 
 export type PerseusGraphTypeLinearSystem = {
@@ -1173,6 +1206,17 @@ export type PerseusGraphTypeLinearSystem = {
     coords?: CollinearTuple[] | null;
     /** The initial coordinates the graph renders with. */
     startCoords?: CollinearTuple[];
+    /**
+     * Custom screen-reader labels for each interactive point.
+     * If provided, pointLabels[i] replaces the default numeric "Point {i+1}"
+     * announcement. Type is string so authors can use letters (e.g., "T") to
+     * match question text, but the editor defaults to the numeric "1", "2", …
+     * — authors override only when they want a letter.
+     *
+     * Example: pointLabels: ["T"] → "Point T at 0 comma 0"
+     *          (instead of the default "Point 1 at 0 comma 0")
+     */
+    pointLabels?: string[];
 };
 
 export type PerseusGraphTypeNone = {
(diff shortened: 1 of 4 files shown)


ivyolamit:
@claude review once

packages/perseus/src/strings.ts
```diff
@@ -135,7 +135,7 @@ export type PerseusStrings = {
         x,
         y,
     }: {
-        num: number;
+        num: number | string;
```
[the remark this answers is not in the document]
ivyolamit: Good catch — both observations check out (verified num-as-plurality convention at strings.ts:8-9 and 640-641, and the stale context at :831). Pushing back on the rename suggestion though: changing the message-template variable from %(num)s to %(label)s would invalidate every existing locale's translation memory for this key, which is steep for a forward-looking nit.                      
   
  Reverting the type widening from this PR instead. PR 2 will introduce a separate string key (e.g., srPointAtCoordinatesNamed: "Point %(label)s at %(x)s comma %(y)s.") when it actually needs it — so existing translations stay valid, the num-is-for-plurality convention is preserved, and the new key gets its own translator context that explicitly names the author-supplied-letter case.

packages/perseus-core/src/data-schema.ts
```diff
@@ -1157,6 +1168,17 @@ export type PerseusGraphTypeCircle = {
...
+    /**
+     * Custom screen-reader labels for each interactive point.
+     * If provided, pointLabels[i] replaces the default numeric "Point {i+1}"
+     * announcement. Type is string so authors can use letters (e.g., "T") to
+     * match question text, but the editor defaults to the numeric "1", "2", …
+     * — authors override only when they want a letter.
+     *
+     * Example: pointLabels: ["T"] → "Point T at 0 comma 0"
+     *          (instead of the default "Point 1 at 0 comma 0")
+     */
```
SonicScrewdriver: Hmmm this is adding significant verbosity to our data schema file, I wonder if we could simplify the comments (maybe a note to check the more verbose explanation elsehwere) or maybe we could make a PointLabel type so that we only need one set of comments about it.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
```diff
@@ -35,6 +35,7 @@ const parsePerseusGraphTypeAngle = object({
     match: optional(constant("congruent")),
     coords: optional(trio(pairOfNumbers, pairOfNumbers, pairOfNumbers)),
     startCoords: optional(trio(pairOfNumbers, pairOfNumbers, pairOfNumbers)),
+    pointLabels: optional(array(string)),
```
handeyeco: Should the type generally match `coords` and `startCoords`?

Like in this example, should it be:


```suggestion
    pointLabels: optional(trio(string, string, string)),
```

packages/perseus-core/src/data-schema.ts
```diff
@@ -1146,6 +1146,17 @@ export type PerseusGraphTypeAngle = {
...
```
handeyeco: This comment seems more noisy than helpful, especially repeated over-and-over. Should it be a one-liner? Maybe linking to a resource if more information is needed?

SonicScrewdriver (approved):
Seems good, but I'd like to see if we could clean up the comments as they're adding a lot of repetitious noise to our data-schema.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
SonicScrewdriver: Good point!

ivyolamit: Excellent point! updated them.

packages/perseus-core/src/data-schema.ts
ivyolamit: Thanks for pointing that out, I simplified the comments.

catandthemachines (approved):
Looks great, thanks @ivyolamit

ivyolamit:
@handeyeco I've addressed your comments above, please let me know if you have any other concerns on this PR 🙏🏼

PR #3607, opened 2026-05-08
Move `getEquationString` to its own file

handeyeco:
## Summary:

The goal was to convert the `static` methods in `InteractiveGraph` into helper functions. Turns out that all the `static` functions basically existed for `getEquationString`, so I moved them all to `get-equation-string.ts` and only exported `getEquationString`.

Issue: LEMS-4071

## Test plan:

**THIS PR ONLY MOVES CODE** and I think that's best. Combining larger refactors with big moves makes PR reviews a pain. For that reason, I'd prefer not to make any logic changes and just focus on moving logic for now.


packages/perseus/src/widgets/interactive-graphs/get-equation-string.test.ts
```diff
@@ -0,0 +1,71 @@
...
+            const props = makeProps(
+                [
+                    [3, 2],
+                    [5, 4],
+                ],
+                0,
+            );
+
+            // Act
+            const equation = getEquationString(props);
```
handeyeco, line 29: These tests existing in `interactive-graph.test.tsx` but used `getLogarithmEquationString`. Since that would mean exporting that function just for tests, I decided it was best just to use `getEquationString` directly.

packages/perseus/src/widgets/interactive-graphs/get-equation-string.ts
```diff
@@ -0,0 +1,673 @@
...
+        Errors.NotAllowed,
+        {metadata: {graphType}},
+    );
+};
+
+/**
+ * @param {object} graph Like props.graph or props.correct
+ * @param {object} props of an InteractiveGraph instance
+ */
+function getLineCoords(graph: PerseusGraphType, props: Props): Coord[] {
```
handeyeco, line 62: This is all just pulled from `interactive-graph.tsx`. I replaced things like `static getLineCoords` with `function getLineCoords` and things like `InteractiveGraph.getLineCoords` with `getLineCoords`. Other than that, there shouldn't be any changes to the code.

packages/perseus/src/widgets/interactive-graphs/interactive-graph.test.tsx
```diff
@@ -2036,76 +2035,6 @@ describe("Interactive Graph", function () {
         );
     });
 
-    describe("getLogarithmEquationString", () => {
```
handeyeco, line 2039: Moved

packages/perseus/src/widgets/interactive-graphs/interactive-graph.tsx
```diff
@@ -190,7 +139,7 @@ type InteractiveGraphProps = {
     fullGraphAriaDescription?: string;
 };
 
-type Props = WidgetProps<
+export type Props = WidgetProps<
```
handeyeco, line 142: I needed to export this because `get-equation-string.ts` needed it.

packages/perseus/src/widgets/interactive-graphs/interactive-graph.tsx
```diff
@@ -257,6 +206,10 @@ class InteractiveGraph extends React.Component<Props, State> {
         },
     };
 
+    static getEquationString(props: Props) {
```
handeyeco, line 209: This makes it available to the editor which uses it. I'd prefer to keep this as-is right now because I think a bigger refactor would make this PR needlessly messy.

packages/perseus/src/widgets/interactive-graphs/interactive-graph.tsx
```diff
@@ -1,11 +1,4 @@
-/* eslint-disable @typescript-eslint/no-invalid-this, react/no-unsafe, react/sort-comp, max-lines */
```
handeyeco, line 1: Yay

packages/perseus/src/widgets/interactive-graphs/get-equation-string.test.ts
```diff
@@ -0,0 +1,71 @@
+import {getEquationString} from "./get-equation-string";
+
+import type {Coord} from "../../interactive2/types";
+
+describe("getEquationString", () => {
+    describe("logarithm", () => {
+        function makeProps(coords: [Coord, Coord], asymptote: number) {
```
jeremywiebe, line 7: I think the call sites would read more clearly if this were an object and you destructured it here:


```suggestion
        function makeProps({coords, aymptote}: {coords: [Coord, Coord], asymptote: number}) {
```

Then call sites would look like: 

```
    const props = makeProps({
        coords: [
            [3, 2],
            [5, 4],
        ],
        asymptote: 0,
    });
```

jeremywiebe (commented):
I'll leave final approval for IG2 folks.  But I'm very happy about this!  Thanks.

ivyolamit (approved):

packages/perseus/src/widgets/interactive-graphs/get-equation-string.test.ts
handeyeco: I mentioned this in the description, but I'm trying as much as possible to keep this PR about moving code without changing code to keep the PR easy to review while being minimally disruptive.

PR #3608, opened 2026-05-08
Scoring: Introduce per-widget scoring info

jeremywiebe:
## Summary:

Perseus scores items by asking each widget to score the learner's input against its rubric. Originally, our scoring system "flattened" the per-widget scores into a single `PerseusScore` (which is modelled as a `total` available points and `earned` points - so when `earned == total`, the answer is correct). 

With this PR we modify the result value so that it also provides the per-widget scores - allowing consumers to understand what parts of the question the learner got right and which ones they got wrong. 

Issue: LEMS-4114

## Test plan:

  * `pnpm lint`
  * `pnpm tsc`
  * `pnpm test`

### DIFF
--- a/config/test/custom-matchers.ts
+++ b/config/test/custom-matchers.ts
@@ -14,7 +14,6 @@ declare global {
             }): R;
             toHaveInvalidInput(message?: string | null): R;
             toHaveBeenAnsweredIncorrectly(): R;
-            // TODO: Delete this unused custom matcher and associated implementation
             toBeHighlighted(): R;
         }
     }
--- a/config/test/custom-matchers.ts
+++ b/config/test/custom-matchers.ts
@@ -14,6 +14,7 @@ declare global {
             }): R;
             toHaveInvalidInput(message?: string | null): R;
             toHaveBeenAnsweredIncorrectly(): R;
+            // TODO: Delete this unused custom matcher and associated implementation
             toBeHighlighted(): R;
         }
     }
--- a/packages/perseus-core/src/validation.types.ts
+++ b/packages/perseus-core/src/validation.types.ts
@@ -129,13 +129,10 @@ export type ValidationResult = Extract<PerseusScore, {type: "invalid"}> | null;
 /**
  * The result of scoring a Perseus item.
  *
- * `score` is the aggregate result across all scoreable widgets in the item
- * (this is the same value that earlier versions of `scorePerseusItem`
- * returned directly).
+ * `score` is the aggregate result across all scoreable widgets.
  *
  * `widgetScores` is the per-widget breakdown, keyed by widget ID, that was
- * combined to produce `score`. This lets consumers display feedback for
- * individual widgets in addition to the overall result.
+ * combined to produce `score`.
  */
 export type PerseusItemScore = {
     score: PerseusScore;
--- a/packages/perseus-core/src/validation.types.ts
+++ b/packages/perseus-core/src/validation.types.ts
@@ -127,47 +127,6 @@ export type PerseusScore =
 export type ValidationResult = Extract<PerseusScore, {type: "invalid"}> | null;
 
 /**
- * The result of scoring a Perseus item.
- */
-export type PerseusItemScore =
-    | {
-          /** Indicates the input is not ready to score. */
-          type: "invalid";
-          /**
-           * An error code describing why the item cannot be scored yet, or
-           * null for a generic invalid state.
-           */
-          message?: ErrorCode | null;
-          /** When true, suppresses the "almost there" hint. */
-          suppressAlmostThere?: boolean | null | undefined;
-          // NOTE: We don't reveal widget scores when the overall score is
-          // {type: invalid} so that we don't leak scoring info. If this info
-          // surfaced to the learner's browser, they could take advantage of the
-          // info by leaving the one widget that is invalid in that state and
-          // making multiple attempts on other widgets ({type: invalid} doesn't
-          // count as an attempt, so they'd get free attempts).
-      }
-    | {
-          /**
-           * Indicates the item has been scored and points awarded. The score
-           * is considered "correct" when `earned` >= `total`.
-           */
-          type: "points";
-          /** The number of points the learner earned. */
-          earned: number;
-          /** The total possible points for this item. */
-          total: number;
-          /** An optional feedback message to display alongside the score. */
-          message?: string | null | undefined;
-          /**
-           * The per-widget scores, keyed by widget ID, that were combined to
-           * produce the item's overall score (see the `earned` and `total`
-           * fields).
-           */
-          widgetScores: {[widgetId: string]: PerseusScore};
-      };
-
-/**
  * The outcome status reported by a self-grading widget (CS Program or IFrame)
  * via postMessage.
  */
--- a/packages/perseus-core/src/validation.types.ts
+++ b/packages/perseus-core/src/validation.types.ts
@@ -127,6 +127,22 @@ export type PerseusScore =
 export type ValidationResult = Extract<PerseusScore, {type: "invalid"}> | null;
 
 /**
+ * The result of scoring a Perseus item.
+ *
+ * `score` is the aggregate result across all scoreable widgets in the item
+ * (this is the same value that earlier versions of `scorePerseusItem`
+ * returned directly).
+ *
+ * `widgetScores` is the per-widget breakdown, keyed by widget ID, that was
+ * combined to produce `score`. This lets consumers display feedback for
+ * individual widgets in addition to the overall result.
+ */
+export type PerseusItemScore = {
+    score: PerseusScore;
+    widgetScores: {[widgetId: string]: PerseusScore};
+};
+
+/**
  * The outcome status reported by a self-grading widget (CS Program or IFrame)
  * via postMessage.
  */
--- a/packages/perseus-core/src/validation.types.ts
+++ b/packages/perseus-core/src/validation.types.ts
@@ -134,10 +134,21 @@ export type ValidationResult = Extract<PerseusScore, {type: "invalid"}> | null;
  * `widgetScores` is the per-widget breakdown, keyed by widget ID, that was
  * combined to produce `score`.
  */
-export type PerseusItemScore = {
-    score: PerseusScore;
-    widgetScores: {[widgetId: string]: PerseusScore};
-};
+export type PerseusItemScore =
+    | {
+          score: Extract<PerseusScore, {type: "invalid"}>;
+          // NOTE: We don't reveal widget scores when the overall score is
+          // {type: invalid} so that we don't leak scoring info. If this info
+          // surfaced to the learner's browser, they could take advantage of the
+          // info by leaving the one widget that is invalid in that state and
+          // making multiple attempts on other widgets ({type: invalid} doesn't
+          // count as an attempt, so they'd get free attempts).
+          widgetScores?: never;
+      }
+    | {
+          score: Extract<PerseusScore, {type: "points"}>;
+          widgetScores: {[widgetId: string]: PerseusScore};
+      };
 
 /**
  * The outcome status reported by a self-grading widget (CS Program or IFrame)
--- a/packages/perseus-core/src/validation.types.ts
+++ b/packages/perseus-core/src/validation.types.ts
@@ -128,25 +128,42 @@ export type ValidationResult = Extract<PerseusScore, {type: "invalid"}> | null;
 
 /**
  * The result of scoring a Perseus item.
- *
- * `score` is the aggregate result across all scoreable widgets.
- *
- * `widgetScores` is the per-widget breakdown, keyed by widget ID, that was
- * combined to produce `score`.
  */
 export type PerseusItemScore =
     | {
-          score: Extract<PerseusScore, {type: "invalid"}>;
+          /** Indicates the input is not ready to score. */
+          type: "invalid";
+          /**
+           * An error code describing why the item cannot be scored yet, or
+           * null for a generic invalid state.
+           */
+          message?: ErrorCode | null;
+          /** When true, suppresses the "almost there" hint. */
+          suppressAlmostThere?: boolean | null | undefined;
           // NOTE: We don't reveal widget scores when the overall score is
           // {type: invalid} so that we don't leak scoring info. If this info
           // surfaced to the learner's browser, they could take advantage of the
           // info by leaving the one widget that is invalid in that state and
           // making multiple attempts on other widgets ({type: invalid} doesn't
           // count as an attempt, so they'd get free attempts).
-          widgetScores?: never;
       }
     | {
-          score: Extract<PerseusScore, {type: "points"}>;
+          /**
+           * Indicates the item has been scored and points awarded. The score
+           * is considered "correct" when `earned` >= `total`.
+           */
+          type: "points";
+          /** The number of points the learner earned. */
+          earned: number;
+          /** The total possible points for this item. */
+          total: number;
+          /** An optional feedback message to display alongside the score. */
+          message?: string | null | undefined;
+          /**
+           * The per-widget scores, keyed by widget ID, that were combined to
+           * produce the item's overall score (see the `earned` and `total`
+           * fields).
+           */
           widgetScores: {[widgetId: string]: PerseusScore};
       };
 
--- a/packages/perseus-score/src/score.test.ts
+++ b/packages/perseus-score/src/score.test.ts
@@ -8,6 +8,7 @@ import {
     generateInputNumberWidget,
     generateInputNumberOptions,
 } from "@khanacademy/perseus-core";
+import invariant from "tiny-invariant";
 
 import {
     scorePerseusItem,
@@ -314,7 +315,7 @@ describe("scorePerseusItem", () => {
         const userInputMap = {};
 
         // Act:
-        const {score} = scorePerseusItem(item, userInputMap, "en");
+        const score = scorePerseusItem(item, userInputMap, "en");
 
         // Assert:
         expect(score).toHaveInvalidInput();
@@ -323,7 +324,7 @@ describe("scorePerseusItem", () => {
 
     it("should return empty if any validator returns empty", () => {
         // Act
-        const {score} = scorePerseusItem(
+        const score = scorePerseusItem(
             {
                 content: "[[☃ dropdown 1]] [[☃ dropdown 2]]",
                 widgets: {
@@ -348,7 +349,7 @@ describe("scorePerseusItem", () => {
         // Arrange
 
         // Act
-        const {score} = scorePerseusItem(
+        const score = scorePerseusItem(
             {
                 content: "[[☃ dropdown 1]] [[☃ dropdown 2]]",
                 widgets: {
@@ -371,6 +372,20 @@ describe("scorePerseusItem", () => {
             total: 2,
             earned: 2,
             message: null,
+            widgetScores: {
+                "dropdown 1": {
+                    earned: 1,
+                    message: null,
+                    total: 1,
+                    type: "points",
+                },
+                "dropdown 2": {
+                    earned: 1,
+                    message: null,
+                    total: 1,
+                    type: "points",
+                },
+            },
         });
     });
 
@@ -383,11 +398,7 @@ describe("scorePerseusItem", () => {
             images: {},
         };
         json.widgets["dropdown 1"].static = true;
-        const {score} = scorePerseusItem(
-            json,
-            {"dropdown 1": {value: 2}},
-            "en",
-        );
+        const score = scorePerseusItem(json, {"dropdown 1": {value: 2}}, "en");
 
         expect(score).toHaveBeenAnsweredCorrectly({
             shouldHavePoints: false,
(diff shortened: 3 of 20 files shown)


packages/perseus/src/util/test-utils.ts
```diff
@@ -14,13 +14,15 @@ import type {
...
+ * Thin wrapper around scorePerseusItem for internal testing. Returns the
+ * aggregate score; tests that need the per-widget breakdown should call
+ * `scorePerseusItem` directly.
  */
 export function scorePerseusItemTesting(
     perseusRenderData: PerseusRenderer,
     userInputMap: UserInputMap,
 ): PerseusScore {
-    return scorePerseusItem(perseusRenderData, userInputMap, "en");
+    return scorePerseusItem(perseusRenderData, userInputMap, "en").score;
```
handeyeco: No action needed, but `scorePerseusItemTesting` was an abstraction that I think I made. Back then it was more tricky to score because we had to pipe strings in. Now that we don't, I wonder if `scorePerseusItemTesting` is a needless abstraction.

packages/perseus-core/src/validation.types.ts
```diff
@@ -126,6 +126,30 @@ export type PerseusScore =
...
 
+/**
+ * The result of scoring a Perseus item.
+ *
+ * `score` is the aggregate result across all scoreable widgets.
+ *
+ * `widgetScores` is the per-widget breakdown, keyed by widget ID, that was
+ * combined to produce `score`.
+ */
+export type PerseusItemScore =
```
benchristel: Why create a new type? If we're making a breaking change to `scorePerseusItem`, could we change the existing `PerseusScore` type at the same time?

handeyeco (approved):
I think this is fine. Some thoughts:
- I get a bad feeling when things splinter off like `PerseusItemScore` and `PerseusScore`. I think it'll be confusing long-term to have both and `Item` is such a generic term as-is. I don't know what the solution to that is though.
- "With a single scoring function, consumers can just ignore what they don't want" rhetorical question...will they though? Do we care about increase network traffic and DB storage if people start sending/storing everything when they only need the flattened score? I think the answer is probably "we don't care and even if we did care, this is good data to have anyway"

packages/perseus-core/src/validation.types.ts
```diff
@@ -126,6 +126,30 @@ export type PerseusScore =
...
+    | {
+          score: Extract<PerseusScore, {type: "invalid"}>;
+          // NOTE: We don't reveal widget scores when the overall score is
+          // {type: invalid} so that we don't leak scoring info. If this info
+          // surfaced to the learner's browser, they could take advantage of the
+          // info by leaving the one widget that is invalid in that state and
+          // making multiple attempts on other widgets ({type: invalid} doesn't
+          // count as an attempt, so they'd get free attempts).
+          widgetScores?: never;
```
benchristel: Why have a `widgetScores` property if it must never be set? Is this the field that callers are expected to use to discriminate branches of the union?

packages/perseus-core/src/validation.types.ts
benchristel: On second thought, I'm not sure why this needs to be a breaking change. It seems like we could add a `widgets` field to the existing `PerseusScore` type and it would be backward compatible (unless callers are creating `PerseusScore` objects themselves... this is where the guidelines in [semver-ts](https://www.semver-ts.org/) become relevant. If we *really* want to be strict about semver, we should "brand" non-user-constructible types so only Perseus can create instances).

packages/perseus-core/src/validation.types.ts
```diff
@@ -126,6 +126,30 @@ export type PerseusScore =
...
+ * The result of scoring a Perseus item.
+ *
+ * `score` is the aggregate result across all scoreable widgets.
+ *
+ * `widgetScores` is the per-widget breakdown, keyed by widget ID, that was
+ * combined to produce `score`.
+ */
+export type PerseusItemScore =
+    | {
+          score: Extract<PerseusScore, {type: "invalid"}>;
```
benchristel: Does this way of structuring the data allow TS to narrow the type of a `PerseusItemScore` when a conditional checks `score.score.type`?

I'd prefer to keep the data flat. That's equivalent to my proposal above (adding a widgetScores field to the existing `PerseusScore` type instead of creating a new type).

benchristel (commented):
I think we can make this change without breaking compatibility, and it would simplify the code too! I left details inline.

packages/perseus-core/src/validation.types.ts
jeremywiebe: The core reason is that `PerseusScore` is used throughout Perseus internally. If we add the `widgetScores` to this type, all widget scoring functions will be ignoring it and I feel it will be confusing for this field to be ignored everywhere _inside_ Perseus except the top-level `scorePerseusItem` function. 

As I type this, I think this points to the fact that `PerseusScore` is really doing the duty of two types (the internal, per-widget scoring) and the final item score result.  

I'd be in favour of branding both `PerseusScore` and `PerseusItemScore` as non-user-constructable.  In fact, pretty much all of our types are that.

packages/perseus-core/src/validation.types.ts
jeremywiebe: I want to make it explicit that `widgetScores` is _never_ provided when `type: "invalid"`.  If this isn't here, then there are no type errors for structures like this: 


```typescript
const score = {
    type: "invalid",
    message?: "Please provide a numerator",
    widgetScores: {},
}
```

By marking the `widgetScores` as optional but it's value as `never`, it's impossible to include `widgetScores` in a `type: "invalid"` score (which I thought was useful to callers so that they don't get confused why the aren't seeing per-widget scoring when `type: "invalid"`).

packages/perseus-core/src/validation.types.ts
jeremywiebe: As I'm working through review feedback.   One thing we could do without touching the `PerseusScore` type itself, is extending it for just `scorePerseusItem()` like this: 

```typescript
const scorePerseusItem(...): PerseusScore & { widgetScores: {[widgetId: string}: PerseusScore} => {...}
```

🤔

benchristel: I like that idea! One problem is that then `invalid` scores would also have `widgetScores`.

I'd be in favor of writing the type out explicitly:

```ts
type PerseusItemScore =
    | {
          type: "invalid";
          message?: ErrorCode | null;
          suppressAlmostThere?: boolean | null | undefined;
      }
    | {
          type: "points";
          earned: number;
          total: number;
          message?: string | null | undefined;
          widgetScores: {[widgetId: string]: PerseusScore};
      };
```

packages/perseus-core/src/validation.types.ts
benchristel: `tsc` gives you a type error if an object literal with a type annotation has extra properties, so we already have protection against score objects getting created with extra fields. As for communicating to callers that `widgetScores` is not present when the type is `"invalid"`, the absence of a `widgetScores` field on the `invalid` branch would also do that.

The `?: never` idiom is useful when the union type has no discriminant, e.g. https://github.com/Khan/frontend/blob/7673c5440189dd7853b4b81ede5c473242bea2f9/libs/component/scrollable-view/src/scrollable-view.tsx#L13-L16 I don't think that applies here, since callers should be able to discriminate on the type field. Maybe I'm missing something, though?

packages/perseus-core/src/validation.types.ts
benchristel: I agree with your first two paragraphs, but I don't think we should start branding types at this point. We've got bigger fish to fry if we want Perseus to follow semver strictly (which might not even be a good goal at this point).

packages/perseus-core/src/validation.types.ts
jeremywiebe: I think I'm trying to achieve something that isn't doable. I think TypeScript probably needs the discriminator at the top-level, but the way I've modelled this, it's in the `score` sub-object. In fact, when I model this in the [TypeScript playground](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAygxgewE7QLxQFBSgHygb1EgC4oAiMBASwDtgBnMgbm22AWAEMAbUmgVwC2AIwhIWrCJyQ0IAEz5DR4gL6ZsefKyhEIpMrQBuPKnObbBEevU4BzCAH5S9YElq2mKphgy6oASWAIQXhkNAIsVnpEFFIAUQAPV044YAAeUJQAGgJdfUpaBjIVAD5vbQB3U3tgTKtSfABtKrka-wUoFzcaWwBdUjqvDDU8CO1osPikpBT0upzCcD1yIxMzUu8VHz9A4LqAJStXKlT5KHR8SOwJ2KhE5NSMmIgFvPICukYNq6gWmrr6A1mtUIMB2s5jj1+rBnkMRthLuNnlMHnNnq8lvpVtxTMUyj8-qCAU4oLJDGJNj5EDQXFAYihUqRdiFnucxlFkblMe9qJ8yDl2FxeFAAEw5KQyeSkEUqHIEkG1Z6A9nabBkGZyKgIKAARjIDTeFF5RQFHB4pB14uksg6Otl6lYsuG3gw1Np2NMTKCLLCbMRHMmXJIKxoxhxZhylmsdmWZAA1hAIGAdEgQO5inLKgqAQ0ftp1ZxNdq9QbuUbCoxTUKLVbJbb7T8nUNXQgacAoB65Icuicgh1mQcjm5TnI-T8bstFsGDKG1vyoFGbPZ9Amkym0z0Mw7sITFWFlf7VeQNVrdfqg7GPiadGbhZaoBKbRaG9omy6gA), it looks like the `widgetScores` becomes the discriminator!




So, I'm leaning to just always having `widgetScores` and documenting that we won't provide them if the `type: "invalid"`.  It's not _perfect_ typing, but if the object is empty, it's not technically incorrect. 🤔  I feel I'm overthinking this a bit.

packages/perseus-core/src/validation.types.ts
jeremywiebe: I think you got it Ben!  Related to my other [comment](https://github.com/Khan/perseus/pull/3608#discussion_r3222841985), we want the discriminator at the top-level of the object, then I think everything will work.

packages/perseus-core/src/validation.types.ts
```diff
@@ -126,6 +126,47 @@ export type PerseusScore =
...
+           */
+          message?: ErrorCode | null;
+          /** When true, suppresses the "almost there" hint. */
+          suppressAlmostThere?: boolean | null | undefined;
+          // NOTE: We don't reveal widget scores when the overall score is
+          // {type: invalid} so that we don't leak scoring info. If this info
+          // surfaced to the learner's browser, they could take advantage of the
+          // info by leaving the one widget that is invalid in that state and
+          // making multiple attempts on other widgets ({type: invalid} doesn't
+          // count as an attempt, so they'd get free attempts).
```
benchristel: The field associated with this comment is gone now.


```suggestion
```

As I'm re-reading it, it occurs to me that it might be useful to surface info about which widgets are invalid. I feel like that feature has been requested at some point. Perhaps we could do something like, if any widget is invalid, return only the invalid widget scores.

packages/perseus-core/src/validation.types.ts
benchristel, line 105: Maybe we should mark `suppressAlmostThere` deprecated on `PerseusScore`, since it's not used anymore.

packages/perseus/src/util/test-utils.ts
```diff
@@ -14,7 +14,9 @@ import type {
 } from "@khanacademy/perseus-core";
 
 /**
- * Thin wrapper around scorePerseusItem for internal testing
+ * Thin wrapper around scorePerseusItem for internal testing. Returns the
+ * aggregate score; tests that need the per-widget breakdown should call
+ * `scorePerseusItem` directly.
```
benchristel, line 19: Is this comment still accurate? I see the return type is `PerseusScore`, but I think it could be `PerseusItemScore` now.

The only salient thing about this function is that it hardcodes the locale to `"en"`, so I feel like the doc comment should mention that.

packages/perseus-score/src/score.ts
```diff
@@ -27,36 +26,21 @@ export function scorePerseusItem(
...
-        scoreableWidgetIds,
-        userInputMap,
-        locale,
-    );
-    return flattenScores(scores);
+    if (score.type === "invalid") {
+        return score;
+    } else {
+        return {...score, widgetScores};
+    }
```
benchristel: No action needed here... but it seems like we might want to review and refactor this area of the code holistically. The way responsibilities are divided up between functions seems kind of arbitrary. E.g. `getScoreableWidgets` is called here instead of in `scoreWidgetsFunctional`; `flattenScores` returns the aggregate score object but then we add `widgetScores` to it. Maybe those functions should be inlined. It's hard to say without more context.

packages/perseus-core/src/validation.types.ts
```diff
@@ -126,6 +126,47 @@ export type PerseusScore =
...
+      }
+    | {
+          /**
+           * Indicates the item has been scored and points awarded. The score
+           * is considered "correct" when `earned` >= `total`.
```
benchristel: > The score is considered "correct" when `earned` >= `total`.

The need to document this logic far from where it's actually implemented makes me think that `Score` should have been a class with an `isCorrect` method. No action needed.

packages/perseus-core/src/validation.types.ts
```diff
@@ -126,6 +126,47 @@ export type PerseusScore =
...
+          /** The total possible points for this item. */
+          total: number;
+          /** An optional feedback message to display alongside the score. */
+          message?: string | null | undefined;
+          /**
+           * The per-widget scores, keyed by widget ID, that were combined to
+           * produce the item's overall score (see the `earned` and `total`
+           * fields).
+           */
+          widgetScores: {[widgetId: string]: PerseusScore};
```
handeyeco: I asked Jeremy via DM about the motivation for duplicating the `PerseusScore` type when creating the `PerseusItemScore` type - alternatively we could have derived one from the other or had smaller, shared types that could be composed into more complete types - but Jeremy said that the duplication was intentional. His argument was that while they looked the same, they are semantically different: one is for PerseusItems and one is for Widgets.

My original concern might have been founded on how KEScore and PerseusScore are used and the confusion they caused during the SSS refactor. This is slightly different though because they are two types that represent a single concept (PerseusItem score).

I still don't really buy that we need these two separate types, but the argument is somewhat convincing and the stakes are low enough that I think it's worth trying.

packages/perseus-core/src/validation.types.ts
```diff
@@ -126,6 +126,47 @@ export type PerseusScore =
  */
 export type ValidationResult = Extract<PerseusScore, {type: "invalid"}> | null;
 
+/**
+ * The result of scoring a Perseus item.
+ */
+export type PerseusItemScore =
```
handeyeco: I think there's a somewhat implicit conceptual shift: `PerseusScore` used to represent both a widget's score and a PerseusItem's score, but now `PerseusItemScore` represents a PerseusItem score and `PerseusScore` represents a widget's score.

I think we need to make sure the code is clear about that distinction and is up-to-date with that usage:

- I think the names could be clearer: if we have `PerseusItemScore` should `PerseusScore` be renamed to `PerseusWidgetScore`?
- I think the comments above each type could be clearer
- Seems like some code still needs to be updated; for example `validateUserInput` still returns `PerseusScore` even though it's flattening the scores. `flattenScores` also seems semantically incorrect, I think it would be more correct to have it return `PerseusItemScore` vs `PerseusScore`. What about `custom-matchers.ts`?

handeyeco (approved):

packages/perseus-core/src/validation.types.ts
jeremywiebe: I think our consumers do still use this field. I don't think it's core to this PR so I'm going to leave it as is for now.

packages/perseus-core/src/validation.types.ts
jeremywiebe: This is why we do code reviews. You bring up a bunch of things I hadn't considered when I was putting this type in place. `widgetScores` won't make sense for the `validateUserInput`. 

If I remove the custom type and just adjust the return value for `scorePerseusItem` to be `PerseusScore & {widgetScores:...}`, then we can leave the naming as it is.

packages/perseus-core/src/validation.types.ts
jeremywiebe: I did some re-work of this PR.  I did end up adding `type: "invalid"` scores to `widgetScores` when the overall score is "invalid".

packages/perseus-core/src/validation.types.ts
jeremywiebe: Meh, I didn't feel I was justifying the type well either and you have deeper experience with SSS.  

I've taken another, different run at modelling this, but _just_ touching the return value of `scorePerseusItem`. I think that's a bit cleaner and less invasive. It also feels closer to what we're intending, which is to _only_ have widgetScores on the return value of `scorePerseusItem` (not on `validateUserInput`, not touching `custom-matchers.ts`, etc).

jeremywiebe:
@benchristel @handeyeco Sorry to tag you for one more review and for the major re-workings.  All the discussion about the `PerseusItemScore` type felt significant, especially Matthew's feedback), so I took a different approach: simply extending `PerseusScore` to include `widgetScores` strictly for `scorePerseusItem`. 

I think this feels a bit better as it avoids introducing a new type concept and guarantees usage to _just_ the scoring function.  I feel I've gone back and forth somewhat on this PR, but hopefully we're close!

benchristel (approved):
LGTM! Before landing, can you remove the EDIT and struck-out portion of the PR message?

benchristel (changes requested):
Actually, I missed one thing: can you please update `scorePerseusItemWithInputNumberAsNumericInput` to include widget scores as well? Otherwise, my side-by-side tests for the numeric input conversion will start failing when this is deployed to the perseus service.

packages/perseus-score/src/score.ts
jeremywiebe: I'll make a TODO for when I get back from vacation.  I was thinking that `flattenScores` _might_ be the place to start aggregating per-widget scores also.  

Maybe we could pair on some re-thinking of this when I'm back?

packages/perseus/src/util/test-utils.ts
jeremywiebe: I'll update the comment.  Matthew noted in another place that this function doesn't really add any value today. I'll make a note to remove it after I finish this widget-score work.

jeremywiebe:
Ok.  Updated to take `scorePerseusItemWithInputNumberAsNumericInput` into account.  I'm avoiding touching the custom matchers for now, but I did look. I think eventually we could add an `options` object to most of the scoring matchers which takes an optional `widgetScores` to assert on.  I didn't do that for now because it feels like something we wouldn't use in most places that we use these assertions (just in the actual scoring function tests)

packages/perseus-score/src/score.test.ts
```diff
@@ -318,7 +324,12 @@ describe("scorePerseusItem", () => {
 
         // Assert:
         expect(score).toHaveInvalidInput();
-        expect(score).toEqual({type: "invalid", message: null});
+        expect(score).toEqual(
+            expect.objectContaining({
+                type: "invalid",
+                message: null,
+            }),
+        );
```
benchristel: This assertion is the same as `expect(score).toHaveInvalidInput().` (well, I guess it additionally asserts that `message` is null). It could be deleted.


```suggestion
```

packages/perseus-score/src/score.test.ts
```diff
@@ -341,7 +352,41 @@ describe("scorePerseusItem", () => {
 
         // Assert
         expect(score).toHaveInvalidInput();
-        expect(score).toEqual({type: "invalid", message: null});
+        expect(score).toEqual(
+            expect.objectContaining({
+                type: "invalid",
+                message: null,
+            }),
+        );
```
benchristel: Same here.


```suggestion
```

packages/perseus-score/src/score.test.ts
```diff
@@ -667,3 +782,84 @@ describe("scorePerseusItemWithInputNumberAsNumericInput", () => {
...
+            total: 2,
+            earned: 1,
+            widgetScores: {
+                "radio 1": {type: "points", total: 1, earned: 0},
+                "radio 2": {type: "points", total: 1, earned: 1},
+            },
+        });
+    });
+
+    it("should omit points scores overall score is invalid", () => {
```
benchristel:
```suggestion
    it("should omit points scores when overall score is invalid", () => {
```

config/test/custom-matchers.ts
```diff
@@ -14,6 +14,7 @@ declare global {
             }): R;
             toHaveInvalidInput(message?: string | null): R;
             toHaveBeenAnsweredIncorrectly(): R;
+            // TODO: Delete this unused custom matcher and associated implementation
```
jeremywiebe: I'm going to remove this TODO from this PR as I just created: https://github.com/Khan/perseus/pull/3626

benchristel (approved):
LGTM! I left a few minor suggestions inline. I also updated the PR description.

PR #3611, opened 2026-05-08
Update item-splitting check to provide a clearer diff

jeremywiebe:
~~Stacked on top of #3610 (now landed).~~ Switches the item-splitting comparison check from diffing the production rollup bundle (`packages/perseus-core/dist/es/index.item-splitting.js`) to diffing a bundle built by esbuild from `packages/perseus-core/src/index.item-splitting.ts`. 

Two reasons:

  1. The production bundle is minified, so the rendered diff in the check summary is one giant line per logical chunk with mangled identifiers (`e`, `t`, `r`). Even running prettier over it doesn't help because `dist/` is in our `.prettierignore` and prettier can't unmangle minifier-renamed identifiers anyway. esbuild's default non-minified output is already line-broken and keeps real identifier names, so the diff is actually readable.

  2. It also insulates the diff baseline from production build pipeline drift. If someone changes Vite/rollup/swc config tomorrow in a way that alters bundle output (different minifier passes, different identifier mangling), the comparison check is unaffected — esbuild emits what it emits given pinned flags.

`esbuild` is added as an explicit devDep at `0.24.0`, the same version Vite is already pulling in transitively. Pinning it explicitly costs us no install bytes (already in the lockfile) and lets us control the diff baseline. It's deliberately decoupled from Vite's transitive copy so that if we ever upgrade Vite or change to a different tool, we don't break this check — see the comment in the workflow for rationale.

_**NOTE**: The item splitting check will fail on this PR because the base branch (`main`) does not have `esbuild` installed. Once we land this PR, future PRs will work because both the base and PR branches will have `esbuild` available._

Sample test run: https://github.com/Khan/perseus/pull/3611/checks?check_run_id=75105820630

Issue: LEMS-4226

## Test plan:

- After this lands, open a follow-up PR with a real item-splitting change and confirm the rendered diff in the `🛠️ Item Splitting` check summary is readable: line-broken JS, real identifier names, no single-line walls of code.
- First run of this check after this lands will produce a wholesale-replacement diff (because the comparison artifact format changes); subsequent runs only show real changes. Acceptable noise for a one-time transition.


package.json
```diff
@@ -61,6 +61,7 @@
         "cypress-jest-adapter": "^0.1.1",
         "cypress-real-events": "^1.12.0",
         "cypress-wait-until": "^3.0.1",
+        "esbuild": "0.24.0",
```
jeremywiebe, line 64: This is already in our dependency tree indirectly via Rollup.

anakaren-rojas (approved):
Sorry Jeremy I think you need an ADR for that new `esbuild` dependency

PR #3612, opened 2026-05-08
OUTDATED - Wire Sortable into AssetContext for stable Chromatic snapshots

Myranae:
**NEW APPROACH - A new approach is being reviewed in https://github.com/Khan/perseus/pull/3692. Not closing this one yet as I am not sure if the other approach is sound/better.**

## Summary:
Removes the 1000ms `chromatic: {delay}` workaround on the matcher `WithTeX` regression story by having Sortable participate in Perseus's existing AssetContext system. The renderer's `onRendered` callback now waits for Sortable's measurement cascade to settle alongside MathJax typeset, so Chromatic doesn't snapshot mid-layout.

See `sortable-asset-context-explainer.md` on the Jira ticket for the full explanation: the render cascade that caused the flakiness, why a quiescence timer (vs. other approaches), and the 50ms tuning.

Issue: LEMS-4117

## Test plan:
  - [x] All checks pass
  - [x] Open Storybook Publish → `Widgets/Matcher/Visual Regression Tests/Initial State/WithTeX` and  → renders cleanly without manual delay
  - [x]  Open Storybook Publish → `Widgets/Sorter/Visual Regression Tests/Initial State/WithTeX` → renders cleanly without manual delay
  - [x] Open Storybook Publish → `Widgets/Sorter/...` → renders cleanly (no regression from new asset wiring)
  - [x] Chromatic snapshot for `WithTeX` is stable across re-runs


packages/perseus/src/components/sortable.tsx
```diff
@@ -500,13 +511,17 @@ class Sortable extends React.Component<SortableProps, SortableState> {
...
     }
 
-    constructor(props: SortableProps) {
-        super(props);
+    constructor(
+        props: SortableProps,
+        context: React.ContextType<typeof AssetContext>,
+    ) {
+        super(props, context);
         // Don't call this.setState() here!
```
jeremywiebe: Can we delete this comment?  We aren't calling `setState()` so it's completely meaningless. lol

packages/perseus/src/components/sortable.tsx
```diff
@@ -600,6 +623,14 @@ class Sortable extends React.Component<SortableProps, SortableState> {
 
         this.setState({items}, () => {
             this.props.onMeasure?.({widths: widths, heights: heights});
+            this.context.setAssetStatus(this._assetKey, false);
+            if (this._settledTimer != null) {
+                clearTimeout(this._settledTimer);
+            }
+            this._settledTimer = setTimeout(() => {
```
jeremywiebe: We should probably be using the WB timing API here instead of raw `setTimeout`.  

https://khan.github.io/wonder-blocks/?path=/docs/packages-timing-withactionscheduler-migration-from-standard-api--docs#migration-example

packages/perseus/src/components/sortable.tsx
```diff
@@ -600,6 +623,14 @@ class Sortable extends React.Component<SortableProps, SortableState> {
 
         this.setState({items}, () => {
             this.props.onMeasure?.({widths: widths, heights: heights});
+            this.context.setAssetStatus(this._assetKey, false);
```
jeremywiebe: So if we `setState()`, we reset the asset context until the timer expires - signalling no re-measurements in the quiescence period. Is that right?

.claude/plans/sortable-asset-context-explainer.md
```diff
@@ -0,0 +1,239 @@
...
+
+1. Renderer mounts. `<TeX>` components register their formulas with
+   AssetContext. MathJax begins typesetting in the background.
+2. MathJax finishes a formula. `<TeX>` flips `setAssetStatus(true)`
+   for that formula.
+3. The DOM size of that card has now changed (TeX glyph rendered).
+4. The Sortable containing that card receives an `onRender` from its
+   Draggable child. It debounces (20ms) and re-measures.
+5. The new measurement might differ from the old one (taller now).
+   `onMeasure` fires up to Matcher.
```
jeremywiebe: Not asking for changes here, but I feel we might benefit from a [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) instead of manual timing and re-measuring.

.claude/plans/sortable-asset-context-explainer.md
```diff
@@ -0,0 +1,239 @@
...
+- Each Sortable instance gets a unique asset key (`sortable-N`,
+  generated from a module-level counter).
+- The constructor calls `setAssetStatus(key, false)`.
+- `componentWillUnmount` calls `setAssetStatus(key, true)` and clears
+  any pending timer (so the renderer doesn't stay blocked on a
+  Sortable that no longer exists).
+- The interesting part: in the `setState` callback inside
+  `measureItems`, after the new dimensions have been applied, we:
+  1. Re-assert `setAssetStatus(key, false)` (a measurement just
+     happened, so we are *not* settled).
```
jeremywiebe: Are we sure that we can flip resolved assets (marked as `true`) back to `false` safely?  

As long as the sorter doesn't set itself to `true` when its the last pending asset, because I think `ServerItemRenderer` would then mark itself as fully rendered and trigger OnRender: https://github.com/Khan/perseus/blob/651efcc50bbd7c03549ba4b62a53b49c0dbff12e/packages/perseus/src/server-item-renderer.tsx#L146

🤔 

I think this code is an improvement worth landing, but one niggle to think about.

.claude/plans/sortable-asset-context-explainer.md
```diff
@@ -0,0 +1,239 @@
...
+  setStates → constraints prop change → Sortable clears + re-
+  measures) is one React render cycle plus a `setTimeout(0)` —
+  let's call it ~5–15ms.
+
+50ms is roughly twice the largest of those intervals, giving some
+headroom but not so much that we noticeably slow the renderer's
+`onRendered` callback in the common case.
+
+If Chromatic is still flaky after this lands, **the first lever to
+reach for is bumping `SORTABLE_SETTLED_MS`**. 100ms or 150ms would
```
jeremywiebe: Agreed!

packages/perseus/src/components/__tests__/sortable.test.tsx
```diff
@@ -271,3 +273,171 @@ describe("Sortable.clearItemMeasurements", () => {
...
+
+            // Assert
+            expect(setAssetStatus).toHaveBeenLastCalledWith(key, true);
+        } finally {
+            jest.useRealTimers();
+        }
+    });
+
+    it("does not mark itself loaded before the quiescence window expires", () => {
+        jest.useFakeTimers();
```
jeremywiebe: All of our tests use fake timers by default: https://github.com/Khan/perseus/blob/086d4a43d3db50fe462603a6ab921348bf519ed1/config/test/test-setup.ts#L52


Is this necessary?

jeremywiebe: Yeah, looking through the tests, I don't think we need to be managing flipping to real timers and all the try...catch around these tests.

packages/perseus/src/components/__tests__/sortable.test.tsx
```diff
@@ -271,3 +273,171 @@ describe("Sortable.clearItemMeasurements", () => {
...
+            });
+
+            // Assert: now settled.
+            expect(setAssetStatus).toHaveBeenCalledWith(key, true);
+        } finally {
+            jest.useRealTimers();
+        }
+    });
+
+    it("clears the pending timer and marks itself loaded on unmount", () => {
```
jeremywiebe: Can we break this into two discrete tests?  The fact that we have two Act and Assert sections in this test signals we should split them into two tests.

packages/perseus/src/components/__tests__/sortable.test.tsx
```diff
@@ -271,3 +273,171 @@ describe("Sortable.clearItemMeasurements", () => {
...
+            });
+
+            // Assert
+            expect(setAssetStatus).not.toHaveBeenCalledWith(key, true);
+        } finally {
+            jest.useRealTimers();
+        }
+    });
+
+    it("resets the quiescence timer when a new measurement arrives", () => {
```
jeremywiebe: This feels like two intermixed tests also: 

1. `setAssetStatus` is not called when a measurement arrives during quiescence period
2. `setAssetStatus` settled (marked `true`) even if measurement arrives to reset the timer

jeremywiebe (approved):
I think I reviewed a slightly stale version of this PR, but I think this makes sense.  I don't _love_ still depending on a timer, but I don't know of a way to determine that everything has settled deterministically.  So timer is fine for me. :)

jeremywiebe:
Oops... I reviewed/approved a Draft PR.  Feel free to re-request if/when you make this a non-draft.

packages/perseus/src/components/sortable.tsx
Myranae: I think it is there to remind folks not to change `this.state` to `this.setState()` since that's used other times in the file, but it looks like the tests do catch it if it gets switched, so I'll delete it :)

packages/perseus/src/components/sortable.tsx
Myranae: Just updated :) Brief summary of the changes for this here. Not sure I understand all the implications of this, but needing the functional component to use a hook makes sense and having a bump to reset the timer after each measure makes sense:

> Migrated the quiescence timer from raw `setTimeout`/`clearTimeout` to the WB timing API (`useTimeout` with `SchedulePolicy.OnDemand`). Since `useTimeout` is a hook, the timer lives in a new functional component `SortableSettledSignal` that renders nothing but exposes a `bump()` ref handle — the class parent calls `bump()` from the `measureItems` `setState` callback, which restarts the 50ms countdown. WB's `useTimeout` also handles cleanup on unmount automatically, so the explicit `clearTimeout` in `componentWillUnmount` could be removed.

packages/perseus/src/components/sortable.tsx
Myranae: Yes. Just did some updates connected to this, but still, that's what is happening here. Every time `measureItems` runs and does a `setState`, the asset is marked as in progress and the timer gets restarted until at least 50ms of time has passed since the last bump.

.claude/plans/sortable-asset-context-explainer.md
Myranae: Claude and I dug into `server-item-renderer.tsx` to check this. The `maybeCallOnRendered` method ([lines 145-156](https://github.com/Khan/perseus/blob/main/packages/perseus/src/server-item-renderer.tsx#L145-L156)) has a `_fullyRendered` one-way latch:

```js
maybeCallOnRendered() {
    if (!this._fullyRendered) {
        const assetsLoaded = Object.values(this.state.assetStatuses).every(Boolean);
        if (assetsLoaded) {
            this._fullyRendered = true;
            this.props.onRendered(true);
        }
    }
}
```

So `onRendered(true)` fires exactly once on the first "all assets loaded" moment. If Sortable later flips back to `false` and then `true` again, the latch ignores it — no duplicate fires, no Chromatic confusion.

The subtler version of your question is whether the *first* settled moment is the right one. For Chromatic, yes: nothing interacts with the page after initial render, so once we hit quiescence we stay there. In production, by the time a student drags a card, `onRendered` has long since fired and downstream consumers don't care about subsequent flips.

Is this missing something though? Does this address your concern?

packages/perseus/src/components/sortable.tsx
jeremywiebe: You shouldn't need a functional component. There's an HoC that can be used for class components.  See the migration link I shared. 

Basically wrap the exported component in `withActionScheduler` and then spread `WithActionSchedulerProps` into the class's Props type. Then in the code where you were doing `setTimeout`, you can use `this.props.schedule.timeout(...)`.

.claude/plans/sortable-asset-context-explainer.md
jeremywiebe: Nope.  I think that addresses my concerns.  Thanks!

packages/perseus/src/components/sortable.tsx
Myranae: Okay, it looks like the main issue with this is related to refs. Matcher and Sorter rely on refs, which don't get passed correctly. Claude helped me summarize the situation below. 

> That was actually our first instinct, but we hit a ref-forwarding issue. The HOC ([source](https://github.com/Khan/wonder-blocks/blob/main/packages/wonder-blocks-timing/src/components/with-action-scheduler.tsx)) returns a plain function component, not a `React.forwardRef`:
> 
> ```js
> function withActionScheduler(WrappedComponent) {
>     const C = props => (
>         <ActionSchedulerProvider>
>             {schedule => <WrappedComponent {...props} schedule={schedule} />}
>         </ActionSchedulerProvider>
>     );
>     C.displayName = `withActionScheduler(${WrappedComponent.displayName})`;
>     return C;
> }
> ```
> 
> `Matcher` and `Sorter` both reach into Sortable via string refs to call its imperative API ([matcher.tsx#L115-L146](https://github.com/Khan/perseus/blob/main/packages/perseus/src/widgets/matcher/matcher.tsx#L115-L146)):
> 
> ```js
> this.refs.left.getOptions()
> this.refs.left.moveOptionToIndex(option, index)
> this.refs.right.getOptions()
> ```
> 
> If we wrap Sortable with `withActionScheduler`, `this.refs.left` resolves to the HOC's outer function component (refs to function components return `null` in strict React 18), so `getOptions()` would throw at runtime.

I think Matcher and Sorter are being looked at soon for drag n drop though, so we may not need to do a huge rework for this. I don't understand refs and HOCs as much as I would like, so don't have a good feel on what is the best way forward. We could add `React.forwardRef` to the WB HOC upstream, then come back to this. Or keep it as is and just let the WB team know about this? Totally could be missing something too.

packages/perseus/src/components/__tests__/sortable.test.tsx
Myranae: Done! :D

packages/perseus/src/components/__tests__/sortable.test.tsx
Myranae: Removed that part of the tests! Thanks! :D

packages/perseus/src/components/__tests__/sortable.test.tsx
Myranae: Also split this test :)

packages/perseus/src/components/sortable.tsx
jeremywiebe: Ok.  My sense is to just leave it as using `setTimeout` and we can adopt the WB timing APIs once these are functional components.  I don't think we need to let the WB know... this is a structural issue with our component being a class component, right?

Myranae: Update on the WB timing approach: We reverted the HOC attempt and stuck with raw `setTimeout` for now. We also extracted the quiescence timer into a small pure utility to allow for better testing:

- **`sortable-asset-tracker.ts`** — `markUnsettled()` / `markSettled()`, no React, no DOM.
- **`sortable-asset-tracker.test.ts`** — 7 isolated tests, no jsdom mocking, runs in ~1s.
- **`sortable.tsx`** — three timer-management sites now collapse to one-line method calls.
- **`sortable.test.tsx`** — replaced the brittle 7-test integration block with 2 mount/unmount smoke tests. Dropped the jQuery import entirely.
- **`sortable-tex-regression.stories.tsx`** — 3 Chromatic stories with TeX content (horizontal / vertical / mixed sizes) so we get visual regression coverage for the case the AssetContext wiring exists to stabilize.

.claude/plans/sortable-asset-context-explainer.md
Myranae: For this, do you think a ticket to explore replacing some of our size measurement code with using ResizeObserver instead would be useful? Or just something to keep in mind going forward?

Myranae:
@claude review once

packages/perseus/src/components/__docs__/sortable-tex-regression.stories.tsx
```diff
@@ -0,0 +1,60 @@
...
+                component:
+                    "Regression tests for Sortable rendering TeX content. " +
+                    "Verifies the AssetContext wiring produces stable layouts " +
+                    "after the quiescence window settles — guards against the " +
+                    "Chromatic flake where snapshots fired before MathJax " +
+                    "finished sizing the cards.",
+            },
+        },
+        chromatic: {disableSnapshot: false, modes: themeModes},
+    },
```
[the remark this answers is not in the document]
Myranae: This is huge. Thanks for this catch! We removed the sortable stories in favor of testing this at the widget level, which uses the renderer and allows for AssetContext.  Added a sorter story to match the matcher TeX story.

packages/perseus/src/components/sortable.tsx
```diff
@@ -500,13 +510,20 @@ class Sortable extends React.Component<SortableProps, SortableState> {
...
         this.state = {
             items: Sortable.itemsFromProps(this.props),
             texRendererLoaded: false,
         };
+        this._assetTracker = new SortableAssetTracker(
+            this._assetKey,
+            context.setAssetStatus,
+        );
+        context.setAssetStatus(this._assetKey, false);
     }
```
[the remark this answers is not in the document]
Myranae: Did the suggested fix

benchristel (changes requested):
This seems to be swapping one timeout for another (the 50ms timeout in the asset tracker). It's not always true that the component has re-rendered 50ms after `setState` is called, so I'm not convinced this approach will actually fix the problem.

Instead, what I think we should do is call `context.setAssetStatus(key, true)` in `componentDidUpdate` (so, after the component has rerendered) if dimensions are set on the items. Something like this:

```ts
componentDidUpdate() {
    if (this.state.items.length > 0) {
        // Measure items if their dimensions have been reset
        if (this.state.items[0].width == 0 && this.state.items[0].height == 0) {
            // Measure on the next frame to allow items size to settle.
            // TODO(jeff, CP-3128): Use Wonder Blocks Timing API.
            // eslint-disable-next-line no-restricted-syntax
            setTimeout(() => this.measureItems(), 0);
        } else {
            // Items have been measured and re-rendered.
            // Notify the context that the UI is ready.
            this.context.setAssetStatus(this.assetKey, true);
        }
    }
}
```

The explainer doc you linked to says that the current solution tries to handle "spurious re-renders" and measurement changes, but I don't think that's necessary. We only care about the *first* measurement.

Myranae:
@benchristel It looks like `remeasureItems` is what can cause content to be re-measured, and it sounds like it might be more common than "spurious" as originally described. Below is the summary of what Claude found after reviewing your suggestion. Mostly that it's actually the second measurement done via `remeasureItems` that is the final measurement. So it sounds like we might care about the second measurement too. What do you think?

> remeasureItems is called from each Draggable's inner <Renderer>'s onRender, which fires on mount, on update, and on every  load inside the Draggable's content (renderer.tsx:1234-1240). The 20ms debounce coalesces the burst of mount-time calls from N Draggables into a single clear+remeasure. So the second measurement isn't defensive — it's the authoritative pass that runs after all Draggables have committed to the DOM


<details>
<summary>More about remeasureItems</summary>

remeasureItems call sites:

- remeasureItems is the onRender callback on each Draggable's inner <Renderer> ([sortable.tsx:884](https://github.com/Khan/perseus/blob/tb/fix-sortable-TeX-timing/packages/perseus/src/components/sortable.tsx#L884)).
- It's debounced 20ms ([sortable.tsx:476-481](https://github.com/Khan/perseus/blob/tb/fix-sortable-TeX-timing/packages/perseus/src/components/sortable.tsx#L476-L481)) and clears item dimensions to 0×0.

When the Renderer fires onRender (per Renderer.handleRender, (renderer.tsx:1220-1241):
1. Synchronously on Renderer componentDidMount ([renderer.tsx:286-291](https://github.com/Khan/perseus/blob/tb/fix-sortable-TeX-timing/packages/perseus/src/renderer.tsx#L286-L291)).
2. Synchronously on Renderer componentDidUpdate ([renderer.tsx:352](https://github.com/Khan/perseus/blob/tb/fix-sortable-TeX-timing/packages/perseus/src/renderer.tsx#L352)).
3. Asynchronously on every  load event inside the rendered content ([renderer.tsx:1234-1236](https://github.com/Khan/perseus/blob/tb/fix-sortable-TeX-timing/packages/perseus/src/renderer.tsx#L1234-L1236)).

Resulting measurement cycle in Sortable:
1. Items render with 0×0 dimensions.
2. componentDidUpdate schedules measureItems via setTimeout(0) — runs before any Draggable's onRender has fired.
3. Each Draggable mounts and calls onRender → debounced remeasureItems fires once 20ms after the last call → clears dimensions to 0×0.
4. componentDidUpdate sees 0×0 → schedules another measureItems.
5. Subsequent  loads inside Draggable content fire onRender again → repeats steps 3-4.

Behavior under the proposed componentDidUpdate + immediate setAssetStatus(true) approach:

- setAssetStatus(true) would fire after step 2's measurement.
- ServerItemRenderer._fullyRendered is a one-way latch ([server-item-renderer.tsx:145-156](https://github.com/Khan/perseus/blob/main/packages/perseus/src/server-item-renderer.tsx#L145-L156)) — once latched, subsequent unsettled signals are ignored.
- Step 3's clear-and-remeasure and step 5's image-load remeasurements happen after the latch.

Behavior under the current quiescence-timer approach:

- Each measureItems call invokes markUnsettled, which resets the 50ms timer.
- Timer fires once 50ms passes with no new markUnsettled calls — i.e., after steps 2-5 have all stopped firing.

</details>

benchristel:
I looked into this a bit more — I don't think we can reliably solve this just at the Sortable level. I think Claude is not entirely correct about the second render being authoritative. Based on console logging in storybook, I think it's actually the *third* render that's authoritative when Sortable is used in the Matcher widget.

The sequence of events is:

- Initial render of the Draggable cards.
- Sortable measures the cards' natural sizes
- Sortable calls onMeasure with the sizes
- The Matcher widget has two columns, each of which is a Sortable. Once both Sortables call onMeasure, Matcher takes the maximum card height across both columns and sends that back down to the Sortables as a constraint. This ensures that all the cards are the same size across both columns.
- The Sortables then re-render with the constraint.

The problem is that the Sortable component itself cannot know whether it will be re-rendered in the future with a new constraint. Therefore, it cannot reliably say when it is "done" rendering. The parent component of the Sortable (the Matcher or Sorter widget) is the component that knows how many re-measurements are needed.

In order to solve this, the Matcher and Sorter widgets need to count the onMeasure events and call setAssetStatus once all expected events have happened.

Myranae:
Closing this for now. Check PR mentioned in summary for second approach. Not moving forward there either, so will just increase delay in original story.

PR #3615, opened 2026-05-08
Update sync-dependencies.ts to preserve comments

jeremywiebe:
`utils/sync-dependencies.ts` rewrites `pnpm-workspace.yaml` by parsing it to a plain JS object with `yaml.parse` and re-emitting it with `yaml.stringify`, which silently dropped every comment in the file. The script worked around this for the file-level header by hardcoding the comment block as a string constant in the script and prepending it on each write — but inline comments elsewhere (notably the explanation above the `jest-environment-jsdom>jsdom: 20.0.3` override) had no such protection and would disappear on the next sync.

This switches the parse path to `yaml.parseDocument`, which returns a node tree that retains comments, and mutates the catalog entries in place via the `YAMLMap` items rather than reassigning JS object keys. The header constant goes away — the header now lives only in the file and round-trips automatically alongside every other comment. Three small helpers (`getCatalogMap`, `packageNameOf`, `setVersion`) narrow the unknown-typed return values from `getIn` / `Pair` without using as casts.

Verified that running the script against an already-in-sync workspace produces a byte-identical file (zero diff), so this change is purely a fidelity improvement — no formatting churn for reviewers on future syncs.

Issue: none

Test plan:
  
  - Run `utils/sync-dependencies.ts ../frontend/pnpm-workspace.yaml` against a clean checkout and confirm `git diff pnpm-workspace.yaml` is empty
  - Inspected `pnpm-workspace.yaml` and confirmed both the file-level header and the inline `jsdom` override comment are still present
  - Run pnpm tsc
  - Run pnpm test


utils/sync-dependencies.ts
```diff
@@ -79,84 +99,44 @@ function main(argv: string[]) {
...
-    const ourPeerDeps: string[] = Object.keys(ourWorkspace.catalogs.peerDeps);
-    const ourDevDeps: string[] = Object.keys(ourWorkspace.catalogs.devDeps);
+    const peerDepsMap = getCatalogMap(ourWorkspace, "peerDeps");
+    const devDepsMap = getCatalogMap(ourWorkspace, "devDeps");
 
     // In our peer dependencies, declare that Perseus will work with any
     // package version compatible with the one we install in dev.
-    for (const pkgName of ourPeerDeps) {
+    for (const pair of peerDepsMap.items) {
+        const pkgName = packageNameOf(pair);
```
benchristel: Just a naming suggestion if we rename the function above to `getKey`.

```suggestion
    for (const peerDepsMapEntry of peerDepsMap.items) {
        const pkgName = getKey(peerDepsMapEntry);
```

utils/sync-dependencies.ts
```diff
@@ -30,12 +30,32 @@ function printHelp() {
...
+}
+
+function packageNameOf(pair: yaml.Pair<unknown, unknown>): string {
+    if (!yaml.isScalar(pair.key) || typeof pair.key.value !== "string") {
+        throw new Error(
+            `Expected catalog key to be a string scalar; got ${JSON.stringify(pair.key)}`,
+        );
+    }
+    return pair.key.value;
+}
```
benchristel, line 49: This function doesn't appear to know anything about package names. It just gets the string key name from a key-value pair. Maybe it should be named `getKey` or `getStringKey`

utils/sync-dependencies.ts
```diff
@@ -30,12 +30,32 @@ function printHelp() {
...
+
+function setVersion(pair: yaml.Pair<unknown, unknown>, version: string): void {
+    if (!yaml.isScalar(pair.value)) {
+        throw new Error(
+            `Expected catalog value to be a scalar; got ${JSON.stringify(pair.value)}`,
+        );
+    }
+    pair.value.value = version;
```
benchristel: Similarly, this function could be named `setValue`.

Also, is the `isScalar` check necessary?

benchristel (approved):
Thank you for doing this! It's so much better!

I left some suggestions about naming inline.

utils/sync-dependencies.ts
jeremywiebe: Good point.  I'll adjust that.

utils/sync-dependencies.ts
jeremywiebe: EDIT: Yes, it's needed to narrow the `yaml.Pair` `value`'s type to be a `yaml.Scalar<unknown>`.  By default, all pair keys and values are `unknown` and assigning `pair.value.value = version` results in a typescript error: `'pair.value' is of type 'unknown'`

~~It's probably a bit over-paranoid.  We already go through all of the catalog and peer deps and check their package names using `getKey()` which validates that they have a `string` value.~~

benchristel: I tried removing the `isScalar` check, and it's required by the types. Non-scalar YAML nodes don't have a `value` property.

Perhaps the check+throw could be abbreviated using `tiny-invariant`:

```ts
invariant(
    yaml.isScalar(pair.value),
    () => `Non-scalar YAML node: ${JSON.stringify(pair.value)}`,
);
```

jeremywiebe: I've changed the three assertions on YAML nodes to use `tiny-invariant`.  I'm not sure it's that much clearer. 🤷‍♂️  I'll leave it with the invariant calls.

PR #3616, opened 2026-05-08
docs(changeset): Resolve several Safari-based issues for the Interactive Graph's Circle graph

SonicScrewdriver:
## Summary:
While testing another ticket, our wonderful QA tester discovered two pre-existing issues on our Circle Graph. This PR aims to resolve both: 

- In Safari, dragging a circle and moving your mouse past the graph's boundary started selecting page text. This is not reproducible on any other graph type.
- In Safari, moving the circle and then shift-tabbing backwards would get you stuck in a “focus trap”. This is not reproducible on any other graph type.

While reviewing useControlPoint, I noticed it uses a different structure than `MovableCircle`: two separate `<g>` elements (one focusable with a11y support, one containing the visual elements that's the gesture/pointer-capture target). Mirroring that structure in `MovableCircle` resolves both bugs.

Issue: LEMS-4118

## Test plan:
- tests pass

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/circle.test.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/circle.test.tsx
@@ -76,9 +76,12 @@ describe("Circle graph", () => {
     test("Shows hairlines when dragging and 'markings' are NOT set to 'none'", () => {
         // Arrange
         useGraphConfigMock.mockReturnValue(baseGraphConfigContext);
-        // Only mock once so it applies just to the circle and not
-        // to the radius point's MovablePoint.
-        useDraggableMock.mockReturnValueOnce({dragging: true});
+        // MovableCircle uses two `useDraggable` calls (keyboard + mouse) —
+        // mock both so they apply to the circle but not the radius point's
+        // MovablePoint that follows.
+        useDraggableMock
+            .mockReturnValueOnce({dragging: true})
+            .mockReturnValueOnce({dragging: true});
         const {container} = render(
             <Mafs width={200} height={200}>
                 <CircleGraph graphState={baseCircleState} dispatch={() => {}} />
@@ -195,13 +198,13 @@ describe("Circle graph screen reader", () => {
         );
     });
 
-    test("should have aria label for circle graph", () => {
+    test("should have aria label for circle graph", async () => {
         // Arrange
         render(<MafsGraph {...baseMafsGraphProps} state={baseCircleState} />);
 
         // Act
-        // eslint-disable-next-line testing-library/no-node-access
-        const circleGraph = document.querySelector(".movable-circle");
+        const buttons = await screen.findAllByRole("button");
+        const circleGraph = buttons[0];
         const radiusPoint = screen.getByTestId(
             "movable-point__focusable-handle",
         );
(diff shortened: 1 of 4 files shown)


packages/perseus/src/widgets/interactive-graphs/graphs/circle.tsx
```diff
@@ -129,50 +129,80 @@ function MovableCircle(props: {
...
 
-    const draggableRef = useRef<SVGGElement>(null);
+    // Focus and gesture targets are kept on separate `<g>`s: in Safari,
+    // a focusable SVG group that's also the pointer-capture target leaks
+    // host-page text selection during mouse drag past the graph boundary.
+    const focusableHandleRef = useRef<SVGGElement>(null);
+    const visibleGroupRef = useRef<SVGGElement>(null);
 
+    // Keyboard support (on focusableHandleRef)
+    useDraggable({
```
SonicScrewdriver, line 140: This split and double `useDraggable` structure matches how `useControlPoint` is implemented! I wondered about finding a way to only use a single `useDraggable` call, but I'd rather match our existing standard.

packages/perseus/src/widgets/interactive-graphs/graphs/circle.test.tsx
```diff
@@ -195,13 +198,13 @@ describe("Circle graph screen reader", () => {
         );
     });
 
-    test("should have aria label for circle graph", () => {
+    test("should have aria label for circle graph", async () => {
         // Arrange
         render(<MafsGraph {...baseMafsGraphProps} state={baseCircleState} />);
 
         // Act
-        // eslint-disable-next-line testing-library/no-node-access
```
SonicScrewdriver, line 203: Might as well clean this up

packages/perseus/src/widgets/interactive-graphs/graphs/circle.test.tsx
```diff
@@ -195,13 +198,13 @@ describe("Circle graph screen reader", () => {
...
-        const circleGraph = document.querySelector(".movable-circle");
+        const buttons = await screen.findAllByRole("button");
```
handeyeco: 1. Do you need async/await if you use something like `getAllByRole`?
2. Is there not an accessible name for what you're looking for? ie `getAllByRole("button", {name: "My accessible name"})`? It's always a red flag when I see `ByRole` without a name.

packages/perseus/src/widgets/interactive-graphs/graphs/circle.tsx
handeyeco: I wonder if that indicates the shared logic that leverages two `useDraggable` should be abstracted into a shared hook?

handeyeco (approved):

packages/perseus/src/widgets/interactive-graphs/graphs/circle.test.tsx
SonicScrewdriver: Great call, updated! 

I can't reply on the other comment you left, but I think we could probably abstract a shared hook! We might want to tackle that in an additional PR though as that could possibly increase the risk of regressions due to the surface area of useControlPoint.

PR #3619, opened 2026-05-11
[Interactive Graph] Initial Announcer infrastructure.

catandthemachines:
## Summary:
Adding in the basic functionality for adding Announcer to our interactive graph state. Specifically we're going in the direction of adding a new `stateAnnouncement` field to identify when we need to announce to the user that the graph's state has changed.

Issue: LEMS-3943

## Test plan:
Nothing should change in our user experience at this point, so simply running `pnpm test` should be sufficient.


packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -506,17 +507,23 @@ function doMovePoint(
...
                 coords: setAtIndex({
                     array: state.coords,
                     index: action.index,
-                    newValue: boundToEdgeAndSnapToGrid(
-                        action.destination,
-                        state,
-                    ),
+                    newValue: snappedPoint,
+                }),
+                announcement: state.strings.srPointAtCoordinates({
```
catandthemachines: Note to self: delete this implementation once this goes to review. This PR should just be the initial implementation, moving aria-live commands to announcer will be in follow-up PRs.

packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.tsx
```diff
@@ -78,6 +84,15 @@ export const StatefulMafsGraph = React.forwardRef<
         prevState.current = state;
     }, [onChange, state, graph]);
 
+    useEffect(() => {
+        if (state.announcement) {
+            announceMessage({
+                message: state.announcement,
+                debounceThreshold: 150,
```
catandthemachines: Note: temporary debounce. Follow-up and check if this is the threshold we want.

packages/perseus/src/widgets/interactive-graphs/types.ts
```diff
@@ -54,6 +55,12 @@ export interface InteractiveGraphStateCommon {
     range: [xRange: Interval, yRange: Interval];
     // snapStep = [xStep, yStep] in Cartesian units
     snapStep: vec.Vector2;
+    // Screen reader announcement to be spoken after the current action.
+    // null means nothing to announce.
+    announcement: string | null;
+    // i18n strings and locale, passed in via initializeGraphState.
+    strings: PerseusStrings;
+    locale: string;
```
benchristel: `strings` never changes, so it kind of feels like it shouldn't be part of the state. It also contains functions, which the [Redux docs explicitly warn against putting in state](https://redux.js.org/style-guide/#do-not-put-non-serializable-values-in-state-or-actions) because it messes up the dev tools.

I think what we can do is replace the `announcement` state field with a `stateChange` field typed as:

```
type InteractiveGraphStateChange = {
  pointIndex: number,
  x: number,
  y: number,
}
```

and convert that to the announcement string in the component.

packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.tsx
benchristel: FYI, [looks like the default debounceThreshold is 250ms](https://khan.github.io/wonder-blocks/?path=/docs/packages-announcer--docs).

packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.tsx
```diff
@@ -78,6 +83,24 @@ export const StatefulMafsGraph = React.forwardRef<
         prevState.current = state;
     }, [onChange, state, graph]);
 
+    useEffect(() => {
+        if (!state.stateAnnouncement) {
+            return;
+        }
+
+        let message: string;
+        switch (state.stateAnnouncement.type) {
```
catandthemachines: Current implementation does not trigger this code so we're fine here. But wanted to leave this so showcase how this new state object will be used going forward.

packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.tsx
```diff
@@ -78,6 +83,24 @@ export const StatefulMafsGraph = React.forwardRef<
...
+            case "move-point":
+                message = strings.srPointAtCoordinates({
+                    num: state.stateAnnouncement.pointIndex + 1,
+                    x: srFormatNumber(state.stateAnnouncement.x, locale),
+                    y: srFormatNumber(state.stateAnnouncement.y, locale),
+                });
+                break;
```
benchristel: You can throw an `UnreachableCaseError` (from wonder-stuff-core) so TypeScript will check the exhaustiveness of this switch statement. Then any additions to the `InteractiveGraphStateAnnouncement` type will force this switch statement to be updated.

```suggestion
                break;
            default:
                throw new UnreachableCaseError(state.stateAnnouncement);
```

packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.tsx
```diff
@@ -78,6 +83,24 @@ export const StatefulMafsGraph = React.forwardRef<
...
+        }
```
benchristel: I think it would be good to pull the message construction out into a function — both so we can test it, and to keep it from overwhelming this component as it gets larger.

catandthemachines:
@claude review once

handeyeco (approved):

packages/perseus/src/widgets/interactive-graphs/graphs/screenreader-text.test.ts
```diff
@@ -1,4 +1,33 @@
...
+        const result = getAnnouncementText(
+            {type: "move-point", pointIndex: 0, x: 3, y: 5},
+            mockStrings,
+            "en",
+        );
+
+        expect(result).toBe("Point 1 at 3 comma 5.");
+    });
+
+    it("throws an UnreachableCaseError for an unhandled announcement type", () => {
```
benchristel, line 20: I'd question whether we have to test this, because the type system is supposed to prevent this case from ever happening.

packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.tsx
```diff
@@ -78,6 +83,20 @@ export const StatefulMafsGraph = React.forwardRef<
...
+
+        announceMessage({
+            message: getAnnouncementText(
+                state.stateAnnouncement,
+                strings,
+                locale,
+            ),
+        });
+    }, [state.stateAnnouncement, strings, locale]);
```
benchristel, line 98: We'll call `announceMessage` whenever the identity of the `stateAnnouncement` changes. I think this will probably do exactly what we want (including re-sending the announcement if two updates in a row produce the same message) but it's a little subtle. I'm not sure what we should do about it. Maybe we should add one of the following:

- a test asserting that the announcement is sent when the object identity changes and not sent when it hasn't changed.
- a comment explaining how this works.

benchristel (approved):
LGTM! I left a couple suggestions inline.

packages/perseus/src/widgets/interactive-graphs/graphs/screenreader-text.test.ts
catandthemachines: We probably don't, but I like that it captures the expected functionality even through the type system would prevent this. I'll keep this here, but if you feel strongly I can reconsider.

packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.tsx
catandthemachines: Added a test to capture this functionality.

packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.test.tsx
```diff
@@ -123,6 +135,45 @@ describe("StatefulMafsGraph", () => {
...
+                          ...baseState,
+                          stateAnnouncement: {
+                              type: "move-point" as const,
+                              pointIndex: 0,
+                              x: 3,
+                              y: 5,
+                          },
+                      }
+                    : baseState,
+        );
```
benchristel, line 159: This seems weird. Why do we have to mock the reducer? I'd rather test the real behavior.

I notice that this PR doesn't contain any changes to the reducer itself — seems like maybe the announcement message will never actually be sent.

PR #3621, opened 2026-05-12
Outline Locked Labels with SVG filter

handeyeco:
## Summary:
Replaces the transparent white background with an opaque white border for locked labels.

Issue: LEMS-4106

### Before:

Screenshot 2026-05-12 at 10 44 27 AM

### After:

Screenshot 2026-05-12 at 10 44 51 AM

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/locked-figures/locked-label.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/locked-figures/locked-label.tsx
@@ -2,7 +2,7 @@ import {
     lockedFigureColors,
     type LockedLabelType,
 } from "@khanacademy/perseus-core";
-import {font} from "@khanacademy/wonder-blocks-tokens";
+import {font, semanticColor} from "@khanacademy/wonder-blocks-tokens";
 import * as React from "react";
 
 import {getDependencies} from "../../../dependencies";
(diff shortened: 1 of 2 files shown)


packages/perseus/src/widgets/interactive-graphs/locked-figures/locked-label.tsx
```diff
@@ -19,19 +19,75 @@ export default function LockedLabel(props: LockedLabelType) {
...
+ * which isn't as obvious as it would seem because Mathjax
+ * renders HTML and uses borders for some lines.
+ *
+ * This creates a composite of the rendered output and
+ * thickens the output.
+ *
+ * See: LEMS-4106 for more information
+ */
+function LockedLabelStrokeFilter() {
+    const color = semanticColor.core.border.knockout.default;
```
benchristel, line 53: Are we sure this is the right color to use? Don't we want it to be the same as the graph background? I wonder if `core.background.base.default` would be a better choice.

I'm not sure what "knockout" means in the context of color tokens, so maybe I'm wrong.

packages/perseus/src/widgets/interactive-graphs/locked-figures/locked-label.tsx
```diff
@@ -19,19 +19,75 @@ export default function LockedLabel(props: LockedLabelType) {
...
+
+    return (
+        <svg width="0" height="0" style={{position: "absolute"}}>
+            <defs>
+                <filter
+                    id="math-stroke"
+                    x="-5%"
+                    y="-5%"
+                    width="110%"
+                    height="110%"
```
benchristel: What do these properties do?

packages/perseus/src/widgets/interactive-graphs/locked-figures/locked-label.tsx
```diff
@@ -19,19 +19,75 @@ export default function LockedLabel(props: LockedLabelType) {
...
+                    {/* expand edges outward */}
+                    <feMorphology
+                        operator="dilate"
+                        radius="2"
+                        in="SourceGraphic"
+                        result="expanded"
+                    />
+
+                    {/* flood with color */}
+                    <feFlood floodColor={color} result={color} />
```
benchristel: `result={color}` is weird. The `result` attribute specifies a "variable" name to put the result of the filter primitive into. We reference this name on line 78 (I think?). I would change this to `result="flood"` and line 78 to `in="flood"` if we can.

packages/perseus/src/widgets/interactive-graphs/locked-figures/locked-label.tsx
handeyeco: I just copied this from Caitlyns requirements:

> The stroke color should use the following WB token: Core Border Knockout Default (semanticColor.core.border.knockout.default)

packages/perseus/src/widgets/interactive-graphs/locked-figures/locked-label.tsx
handeyeco: Seemingly nothing.

packages/perseus/src/widgets/interactive-graphs/locked-figures/locked-label.tsx
handeyeco: Good catch.

benchristel (approved):
LGTM!

PR #3622, opened 2026-05-12
[Color] Apply color filter to images and math when in dark mode

mark-fitzgerald:
## Summary:
This PR applies a filter to images (PNG & GIF), Graphie elements, and MathJax which inverts the color of those items. This filter is applied only when Dark Mode is specified in the theme name (provided by Wonder Blocks), and only affects these items within a Perseus container.

Issue: LEMS-4126

## Test plan:
1. Launch Storybook
1. Navigate to the Dark Mode stories for the Renderer (Renderers => Dark Mode)
1. Examine the items in that section and see that they are viewable even with the dark background
1. Notice that the JPEG and GIF photos are NOT inverted


packages/perseus/src/styles/styles.css
```diff
@@ -361,3 +361,20 @@
...
+    img[src$=".png"],
+    img[src$=".svg"],
+    .graphie,
+    .mathjax-wrapper {
+        /* This filter is applied to all PNG/SVG images,
+           and to the content of Graphie images (text labels, etc.),
+           and to any MathJax.
+           It leaves JPEGs & GIFs untouched.
+        */
+        filter: invert(1) hue-rotate(180deg) brightness(1.5);
```
Myranae, line 378: This is so cool! 😍

packages/perseus/src/__docs__/dark-mode.stories.tsx
```diff
@@ -0,0 +1,106 @@
...
+};
+
+export const Photos: Story = {
+    render: RenderImages(
+        "$~\\grayH\\text{JPEG}$\n\n![Earth and Moon](https://cdn.kastatic.org/ka-content-images/61831c1329dbc32036d7dd0d03e06e7e2c622718.jpg)\n\n" +
+            "$~\\grayH\\text{GIF}$\n\n![Skateboarder](https://cdn.kastatic.org/ka-content-images/589905512e0a593ba12e4ac290c142cd54879ae4.gif)",
+    ),
+};
+
+const MathJaxColors = [
```
Myranae, line 86: This list might need an update when we do the math ColorSync work. Might be helpful to leave a comment with a TODO referencing the math ticket? Something like:

```suggestion
// TODO(LEMS-3505): Review this list when completing the math ColorSync conversion
const MathJaxColors = [
```

packages/perseus/src/__docs__/dark-mode.stories.tsx
```diff
@@ -0,0 +1,106 @@
...
+            },
+        },
+        chromatic: {disableSnapshot: false},
+    },
+};
+export default meta;
+
+const RenderImages = (content: string): (() => React.JSX.Element) => {
+    return function Render() {
+        // This is needed to apply a fictitious dark mode them to the body element.
```
Myranae, line 28:
```suggestion
        // This is needed to apply a fictitious dark mode theme to the body element.
```

packages/perseus/src/__docs__/dark-mode.stories.tsx
```diff
@@ -0,0 +1,106 @@
...
+        useEffect(() => {
+            setTimeout(() => {
+                document.body.setAttribute(
+                    THEME_DATA_ATTRIBUTE,
+                    "some-theme-in-dark",
+                );
+            }, 10);
+        }, []);
```
Myranae, line 36: Is this needed because the content needs to render before the theme can be applied? 

On another note, should we have regression tests set up for this functionality for renderer? We'd probably need to take a delay like this into account with Chromatic snapshots though.

packages/perseus/src/__docs__/dark-mode.stories.tsx
```diff
@@ -0,0 +1,106 @@
+import {generateTestPerseusRenderer} from "@khanacademy/perseus-core";
+import {THEME_DATA_ATTRIBUTE} from "@khanacademy/wonder-blocks-theming";
+import * as React from "react";
+import {useEffect} from "react";
+
+import QuestionRendererForStories from "../widgets/__testutils__/question-renderer-for-stories";
+
+import type {Meta, StoryObj} from "@storybook/react-vite";
+
+type Story = StoryObj;
```
Myranae, line 10: Nisha pointed out to me in another PR that this is usually listed after the meta export. I'm not sure it matters as much, but consistency is probably good. We could just switch to consistently have it up here, but we'd probably want to update the other places too 😂

Myranae (approved):
This looks great! 😍 So cool to have this functionality becoming available. Left some non-blocking comments, but it looks good to go!

My only other thought is just to make sure that this meshes with whatever other dark mode styling gets applied, but that might require some testing later.

packages/perseus/src/__docs__/dark-mode.stories.tsx
mark-fitzgerald: Yeah, this is something that I'm not so sure about. The current Storybook setup (preview.tsx) adds a toolbar that includes a dropdown for the two Wonder Blocks themes. When the page loads, it gets the Thunderblocks theme by default. I need  a way to override this for this page, and I couldn't figure out how to adjust that dropdown for just this story. I don't like the `setTimeout`, but I don't know any other way to add the theme name. I am open to any suggestions.

packages/perseus/src/__docs__/dark-mode.stories.tsx
mark-fitzgerald: Meaning, move it to line 9?

mark-fitzgerald:
> My only other thought is just to make sure that this meshes with whatever other dark mode styling gets applied, but that might require some testing later.

@Myranae - Would you expand on this some more? In what way are you thinking that it should mesh with other dark mode styling?

packages/perseus/src/__docs__/dark-mode.stories.tsx
jeremywiebe: We usually put it after `meta` because it should be based on the `meta` type. 

```typescript
type Story = StoryObj<typeof meta>;
```

jeremywiebe: Although, you don't have a `component` type here, so these stories don't really have much typing and won't have a control panel that has anything.

Myranae: No, move the `type Story` code to line 25. After the `export default meta` line

packages/perseus/src/__docs__/dark-mode.stories.tsx
jeremywiebe: Can you override it in the Story (and possibly even Meta) like this?

https://github.com/Khan/wonder-blocks/blob/dda49e53ef73058db7b28de87df3d10509d70bd7/__docs__/wonder-blocks-typography/typography.stories.tsx#L82

Myranae:
@mark-fitzgerald I'm not exactly sure how the other styling will be done, so my only thought was to make sure whatever you've done here aligns with how they do it. I don't know anything about where the other dark mode styling is at, but was mostly thinking about the CSS values for the filter property and making sure those align with whatever values the other folks use.

packages/perseus/src/__docs__/dark-mode.stories.tsx
mark-fitzgerald: That was my first thought. However, I'm applying a fictitious theme to the page, whereas the referenced example is pulling from one of the existing themes.

I was thinking that I would leave this in for now, but then update later when WB has dark mode themes fully in place. Then I could just switch to the WB Dark Mode theme.

mark-fitzgerald:
> I'm not exactly sure how the other styling will be done, so my only thought was to make sure whatever you've done here aligns with how they do it. I don't know anything about where the other dark mode styling is at, but was mostly thinking about the CSS values for the filter property and making sure those align with whatever values the other folks use.

Great point! I posted a message to Bea in Wonder Blocks team about that. There doesn't appear to be any other team needing to handle images, but Bea has a note to work on background images. I'll check in with her on that at that time.

PR #3628, opened 2026-05-13
Chromatic: Don't block action waiting for tests to complete

jeremywiebe:
Adds `exitOnceUploaded: true` to the chromaui action so the workflow finishes as soon as the Storybook upload completes, instead of waiting for Chromatic's visual diff to run. Chromatic reports results back to the PR via its GitHub integration, so blocking the workflow on build results is redundant. This is the optimization Chromatic itself suggested in our recent workflow logs ("Speed up Continuous Integration … add the option `with: exitOnceUploaded: true`").

Chromatic docs: https://www.chromatic.com/docs/configure/#exitonceuploaded

Issue: none

## Test plan:

- After merging, confirm the `Publish Storybook to Chromatic` workflow exits after upload and that Chromatic's check still shows up on subsequent PRs (ie. we don't block the action waiting for Chromatic to run its tests)


benchristel (approved):
Thanks!

PR #3629, opened 2026-05-13
Update `convertInputNumberOptionsToNumericInput` to account for the `inexact` widget option

benchristel:
## Summary:
This PR fixes scoring mismatches found in side-by-side testing.  When `inexact`
is undefined or false for an input-number widget, the scoring logic changes in
two ways:

- we force `maxError` to 0.
- decimal answers are rounded to 9 decimal places.

We can invoke the same logic for a numeric-input widget by setting maxError to
zero.

This PR also adds data-driven regression tests that do essentially what the
side-by-side testing in production does. The plan is: any new side-by-side
errors we see get copied to this test suite. The test should fail initially.
When the test passes, it means the side-by-side error is fixed.

Issue: LEMS-4110

## Test plan:

Deploy to the perseus service. Side-by-side testing errors should decrease by
at least 90%.

### DIFF
--- a/packages/perseus-score/src/input-number-to-numeric-input-regression.test.ts
+++ b/packages/perseus-score/src/input-number-to-numeric-input-regression.test.ts
@@ -18,6 +18,7 @@ interface TestCase {
 // - Search for "side-by-side" in the Perseus service
 // - select "Copy > Copy as JSON"
 // - `pbpaste | jq '.jsonPayload.metadata | {userInput: .inputNumberUserInputs, widgets: .inputNumberWidgets}'`
+// TODO(LEMS-4085): Delete this test file.
 const cases: TestCase[] = [
     {
         title: "when inexact = false",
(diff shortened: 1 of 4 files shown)


packages/perseus-score/src/input-number-to-numeric-input-regression.test.ts
```diff
@@ -0,0 +1,124 @@
...
+interface TestCase {
+    title: string;
+    userInput: UserInputMap;
+    widgets: PerseusWidgetsMap;
+}
+
+// NOTE: To add a test case to this list:
+// - Go to https://console.cloud.google.com/logs
+// - Search for "side-by-side" in the Perseus service
+// - select "Copy > Copy as JSON"
```
anakaren-rojas, line 19: What do you think about moving this to the scoring README, along with a note on when to add a new case? That would make it easier for others to find in the future.

anakaren-rojas (approved):
Overall LGTM. Left one non-blocking comment

packages/perseus-score/src/input-number-to-numeric-input-regression.test.ts
benchristel: Hmm, I'm not sure I agree that a readme is easier to find than right next to the tests. In any case, these tests are a short-term thing (they're just scaffolding the migration and we can remove them once the migration is complete).

benchristel: I added a TODO to delete these tests once the project wraps up.

PR #3630, opened 2026-05-13
Write changeset when we sync deps to avoid guesswork

jeremywiebe:
## Summary:

> **NOTE**: I'd be open to a different way of doing this if you (reviewer) disagree. I was prompted to do this because of the uncertainty in [Synchronize Perseus Dependencies](https://khanacademy.atlassian.net/wiki/spaces/LC/pages/4802216180/Releasing+Deploying+Perseus#Synchronize-Perseus-Dependencies) documentation.

This PR extends `utils/sync-dependencies.ts` so that a Changesets entry is generated as part of the dependency-sync flow. This removes the manual step (and ambiguity) we currently have during every release of figuring out which workspace packages need a version bump after a sync.

After the script does its original work, it remembers the list of packages that were affected by dependency changes. The script then resolves the frontend commit SHA via `git -C <client repo dir> rev-parse HEAD`, and writes `.changeset/sync-deps-<YYYY-MM-DD>.md` bumping each affected package as a patch.
  
A couple of design notes worth flagging:

  - The match is restricted to `peerDependencies`. `devDependencies` references to the same catalog entries don't change a package's published contract, so they don't warrant a release bump even though the catalogs always change in lockstep.
  - Bumps are always `patch`. Wonder Blocks syncs are essentially always patch-level for us (hopefully); if a sync ever pulls in a breaking range and we need a `minor`/`major`, the generated changeset is meant to be edited before merging.

Issue: none
  
## Test plan:
  
  - Run `pnpm tsc`
  - From a branch with no catalog changes, run `./utils/sync-dependencies.ts ../frontend/pnpm-workspace.yaml` and confirm it logs `No catalog version changes detected` and writes no changeset
  - From a branch with catalog changes, run the script and confirm `.changeset/sync-deps-<today>.md` appears with patch bumps for the affected packages and a body listing the synced packages


utils/sync-dependencies.ts
```diff
@@ -8,13 +8,20 @@
...
 import invariant from "tiny-invariant";
 import yaml from "yaml";
 
+import {createSyncChangeset} from "./internal/create-sync-changeset";
 import {updateCatalogHashes} from "./internal/update-catalog-hashes";
 
+/**
+ * The directory (relative to the repo root) where changeset files live.
+ */
+export const CHANGESET_DIR = ".changeset";
```
anakaren-rojas, line 23: NIT: this seems like its only used in this file, might not need to export

utils/internal/create-sync-changeset.ts
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * Produces the file name and contents of a changeset that bumps each affected
+ * package by a `patch`, with a summary that references the frontend commit the
+ * dependencies were synced from. The caller is responsible for writing the
+ * returned contents to disk.
+ *
+ * We generate the changeset filename ourselves rather than depending on the
+ * changeset CLI (because it does not provide any sort of command-line options
+ * or API that is usage in a script). Syncing twice on the same day reuses the
+ * same name, so the later changeset overwrites the earlier one.
```
anakaren-rojas, line 21: In the sync deps file, an existing changeset would cause a failure, but this comment says that the changeset is overwritten? 
```
if (fs.existsSync(changesetPath)) {
            console.error(
                `❌ Changeset file (${changesetPath}) already exists. Please delete if you want to re-run this script.`,
            );
            process.exit(1);
        }
```

utils/internal/__tests__/create-sync-changeset.test.ts
```diff
@@ -0,0 +1,46 @@
+/**
+ * @jest-environment node
+ */
+import {describe, expect, it, jest, beforeEach} from "@jest/globals";
+
+import {createSyncChangeset} from "../create-sync-changeset";
+
+describe("createSyncChangeset", () => {
+    beforeEach(() => {});
```
anakaren-rojas, line 9: did this need to have something in there?

PR #3632, opened 2026-05-13
[ColorSync] Update Radio widget regression stories to use best practices

nishasy:
## Summary:
Update the radio regression stories so that
- the meta is typed
- the args are the widget's direct args, with wrappers being decorators

Issue: https://khanacademy.atlassian.net/browse/LEMS-3497

## Test plan:
Check the UI Tests below - there should be no differences.

### DIFF
--- a/packages/perseus/src/widgets/radio/__docs__/multiple-choice-initial-state-regression.stories.tsx
+++ b/packages/perseus/src/widgets/radio/__docs__/multiple-choice-initial-state-regression.stories.tsx
@@ -1,26 +1,30 @@
 import {
-    generateTestPerseusItem,
     generateRadioChoice,
+    generateTestPerseusItem,
 } from "@khanacademy/perseus-core";
+import * as React from "react";
 
 import {themeModes} from "../../../../../../.storybook/modes";
+import {ServerItemRenderer} from "../../../server-item-renderer";
+import {testDependenciesV2} from "../../../testing/test-dependencies";
 import {radioRendererDecorator} from "../../__testutils__/radio-renderer-decorator";
 import {narrowViewportDecorator} from "../../__testutils__/story-decorators";
 import {
-    choicesWithGraphie,
-    choicesWithImages,
+    choicesWithGraphieArgs,
+    choicesWithGraphieContent,
+    choicesWithGraphieImages,
+    choicesWithImagesArgs,
     overflowContentInGradedGroupSet,
-    questionWithRationale,
 } from "../__tests__/radio.testdata";
 
-import type {RadioPublicWidgetOptions} from "@khanacademy/perseus-core";
+import type {PerseusRadioWidgetOptions} from "@khanacademy/perseus-core";
 import type {Meta, StoryObj} from "@storybook/react-vite";
 
 /**
  * These are visual regression stories for the radio widget.
  */
 
-const meta: Meta<RadioPublicWidgetOptions> = {
+const meta: Meta<PerseusRadioWidgetOptions> = {
     title: "Widgets/Radio/Visual Regression Tests/Initial State",
     tags: ["!autodocs", "!manifest"],
     decorators: [radioRendererDecorator],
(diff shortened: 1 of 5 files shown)


packages/perseus/src/widgets/radio/__docs__/multiple-choice-interactions-regression.stories.tsx
```diff
@@ -39,54 +51,18 @@ export default {
...
-        const choiceToClick = canvas.getByRole("button", {
-            name: "(Choice C) Correct",
-        });
-        await userEvent.click(choiceToClick);
-        const checkAnswerButton = canvas.getAllByRole("button", {
-            name: "Check",
-        })[0];
-        await userEvent.click(checkAnswerButton);
-        await checkAnswerButton.blur();
-    },
```
nishasy, line 82: Just moved this down since they're not directly rendering the radio widget.

packages/perseus/src/widgets/radio/__docs__/multiple-choice-interactions-regression.stories.tsx
```diff
@@ -4,31 +4,43 @@ import {
...
+            ),
+            generateRadioChoice(
+                "Only $8$ satisfies the equation $\\sqrt{64}=x$",
+            ),
+            generateRadioChoice(
+                "No value of $x$ satisfies the equation $\\sqrt{64}=x$",
+            ),
+        ],
+    };
+};
```
nishasy, line 40: Moved this over from the testdata file since it was only being used here.

packages/perseus/src/widgets/radio/__docs__/multiple-choice-interactions-regression.stories.tsx
```diff
@@ -116,51 +89,17 @@ export const ChoiceTextColorInMultipleSelect = {
...
-            dependencies={storybookDependenciesV2}
-        />
-    );
-};
-ChoiceTextColorInArticle.play = async ({canvas}) => {
-    const choiceToToggle = canvas.getByRole("button", {
-        name: /Warp drive$/,
-    });
-    choiceToToggle.click();
-};
```
nishasy, line 150: Moved this down since it's not directly rendering a Radio widget.

packages/perseus/src/widgets/radio/__docs__/multiple-choice-initial-state-regression.stories.tsx
```diff
@@ -44,369 +41,354 @@ export default {
...
+        ],
     },
 };
 
 export const SingleSelectWithRationale = {
+    decorators: [radioRendererDecorator],
     args: {
-        item: generateTestPerseusItem({
-            question: questionWithRationale,
-        }),
```
nishasy, line 120: `questionWithRationale` was only used here, so I put it directly in the story and deleted it from testdata.

packages/perseus/src/widgets/radio/__docs__/multiple-choice-interactions-regression.stories.tsx
```diff
@@ -39,54 +51,18 @@ export default {
...
-    args: {
-        item: generateTestPerseusItem({
-            question: choicesWithMathFont(),
-        }),
+export default meta;
+
+type Story = StoryObj<typeof meta>;
+
+export const ChoiceTextColorInSingleSelect: Story = {
+    decorators: [radioRendererDecoratorWithDebugUI],
```
mark-fitzgerald, line 61: I know this was already set up this way previous, but I wonder why we need "DebugUI" in the interactions stories, but not in the initial state stories?

mark-fitzgerald (approved):

packages/perseus/src/widgets/radio/__docs__/multiple-choice-interactions-regression.stories.tsx
nishasy: My guess is that it was an oversight. I don't think it makes sense to have the debug UI in regression stories, so I'd even advocate to switch everything over to the non-debug UI version in the future.

PR #3633, opened 2026-05-14
Remove unused `static` option from numeric-input widget

benchristel:
## Summary:
The static field in the widget options has no effect. The `static` field on the
wrapper object is the real one. This PR removes the unused option.

Issue: LEMS-4134

## Test plan:

CI checks should pass.

Deploying: Since the current numeric-input parser tolerates a missing `static`
field, and the new parser tolerates an extra `static` field, we can deploy the
various services and frontends of webapp in any order (if this is the only
schema change in the deploy).


anakaren-rojas (approved):
Do we need any FE or Webapp changes?

PR #3637, opened 2026-05-19
Deprecate EditorJsonify

handeyeco:
## Summary:
We should be explicit about what we serialize.

See:
- [Remove EditorJsonify and the denylist from Perseus](https://khanacademy.atlassian.net/browse/LEMS-4108)
- ["EditorJsonify needs to die"](https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/4765221980/graded+breaks+content+pipeline?focusedCommentId=4766990629)

### DIFF
--- a/.changeset/ten-rivers-grow.md
+++ b/.changeset/ten-rivers-grow.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": minor
+---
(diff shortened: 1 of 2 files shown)


benchristel (approved):
Thanks!

PR #3638, opened 2026-05-19
Improve WidgetLogic type

handeyeco:
## Summary:
When working on `interactiveGraphWidgetLogic.accessible` I noticed this:

``` TS
    accessible: (widgetOptions: PerseusWidgetOptions): boolean => {
        const interactiveGraphOptions =
            // eslint-disable-next-line no-restricted-syntax
            widgetOptions as PerseusInteractiveGraphWidgetOptions;
```

I figured maybe we could improve the types using generics for WidgetLogic.

### DIFF
--- a/packages/perseus-core/src/widgets/core-widget-registry.ts
+++ b/packages/perseus-core/src/widgets/core-widget-registry.ts
@@ -40,9 +40,9 @@ import type {
 import type {PerseusWidgetOptions, PerseusWidget} from "../data-schema";
 import type {Alignment} from "../types";
 
-const widgets = new Registry<WidgetLogic>("Core widget registry");
+const widgets = new Registry<WidgetLogic<any, any>>("Core widget registry");
 
-export function registerWidget(type: string, logic: WidgetLogic) {
+export function registerWidget(type: string, logic: WidgetLogic<any, any>) {
     widgets.set(type, logic);
 }
 
--- a/packages/perseus-core/src/widgets/core-widget-registry.ts
+++ b/packages/perseus-core/src/widgets/core-widget-registry.ts
@@ -221,6 +221,7 @@ export function registerCoreWidgets() {
     ];
 
     widgets.forEach((w) => {
-        registerWidget(w.name, w);
+        // eslint-disable-next-line no-restricted-syntax
+        registerWidget(w.name, w as WidgetLogic<any, any>);
     });
 }
--- a/packages/perseus-core/src/widgets/image/index.ts
+++ b/packages/perseus-core/src/widgets/image/index.ts
@@ -1,7 +1,4 @@
-import type {
-    PerseusImageWidgetOptions,
-    PerseusWidgetOptions,
-} from "../../data-schema";
+import type {PerseusImageWidgetOptions} from "../../data-schema";
 import type {WidgetLogic} from "../logic-export.types";
 
 export type ImageDefaultWidgetOptions = Pick<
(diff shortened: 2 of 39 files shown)


packages/perseus-core/src/widgets/core-widget-registry.ts
```diff
@@ -221,6 +221,7 @@ export function registerCoreWidgets() {
     ];
 
     widgets.forEach((w) => {
-        registerWidget(w.name, w);
+        // eslint-disable-next-line no-restricted-syntax
+        registerWidget(w.name, w as WidgetLogic<any, any>);
```
handeyeco, line 225: I feel like this is a safe bet, but I'm open to push back. It allows us to use `WidgetLogic` without type params for widgets that don't need the type params.

---

From Claude:

Yes. The error is about TypeScript's contravariant parameter checking.

  WidgetLogic (no args) now defaults to WidgetLogic<never, never>. That gives it:
  - accessible?: boolean | ((options: never) => boolean)
  - getPublicWidgetOptions?: (options: never) => never

  The registry is typed WidgetLogic<any, any>, which expects:
  - accessible?: boolean | ((options: any) => boolean)

  Assigning the first into the second requires the parameter any to be assignable to never
  (contravariance: callback param types flip direction). Under strictFunctionTypes, that check fails
  — hence "Type 'any' is not assignable to type 'never'". Intuitively: the registry promises to call
  the callback with anything (any), but a (options: never) => boolean claims to accept nothing.

  This means WidgetLogic<never, never> is more restrictive than WidgetLogic<any, any> on its inputs,
  not more general — so the widgets I just simplified (definition, etc.) can no longer go into the
  registry.

  A few ways forward:
  1. Cast at the registration site. Keep the defaults as never, but registerWidget(w.name, w as
  WidgetLogic<any, any>). Honest about the type erasure happening at the storage boundary.
  2. Default TWidgetOptions = any (not never). Loses some safety inside widget callbacks if widgets
  accept the default, but the 10 widgets without callbacks don't reference it anyway. Assignment to
  the registry's WidgetLogic<any, any> then trivially works.
  3. Revert defaulting TWidgetOptions. Keep TPublicWidgetOptions = never as the only default — that
  one's safe because never in covariant return position widens to any. Re-add the explicit
  <PerseusXxxOptions> to the 10 widgets. Less clean per-widget, no friction at the registry.

packages/perseus-core/src/widgets/interactive-graph/index.ts
```diff
@@ -56,30 +54,31 @@ const defaultWidgetOptions: InteractiveGraphDefaultWidgetOptions = {
...
+> = {
     name: "interactive-graph",
     defaultWidgetOptions,
     getPublicWidgetOptions: getInteractiveGraphPublicWidgetOptions,
     // Function determining if a interactive graph is accessible.
     // Interactive Graphs are accessible as long as:
     // 1. They do not contain a protractor
     // 2. They do not contain a graphie background image
-    accessible: (widgetOptions: PerseusWidgetOptions): boolean => {
-        const interactiveGraphOptions =
```
handeyeco, line 68: Original motivation

packages/perseus-core/src/widgets/image/index.ts
```diff
@@ -44,17 +41,15 @@ const imageWidgetLogic: WidgetLogic = {
     defaultAlignment: "block",
     // This widget's accessibility depends on its widget option: if the image
     // has a background but no alt text, it is not accessible
-    accessible: (widgetOptions: PerseusWidgetOptions): boolean => {
-        // eslint-disable-next-line no-restricted-syntax
-        const imageOptions = widgetOptions as PerseusImageWidgetOptions;
```
handeyeco, line 49: Also helped here.

benchristel (approved):
LGTM. Note that this doesn't give us type safety at the call sites of e.g. `getPublicWidgetOptions`, because it's not known at compile time whether the widget registered for a type name actually accepts a given options type. I think this PR is an improvement since it gets rid of some casts, though.

PR #3639, opened 2026-05-19
Disallow decimal input when input-number answer has > 10 decimal places

benchristel:
## Summary:
When scoring input-number widgets using the numeric-input logic, we now
disallow decimal and integer input under the following conditions: `answerType`
is `number` or not set, `inexact` is not `true`, and the correct answer `value`
has more than 10 decimal places. This ensures that learners are not
approximating a rational answer with a decimal. It preserves the existing
scoring behavior of input-number.

This fixes a side-by-side test mismatch. I added a regression test to reproduce
the mismatch observed in production.

See this doc for more info: https://khanacademy.atlassian.net/wiki/spaces/LC/pages/4767219957/Side-by-side+test+failure+analysis+2026-05-18

Issue: LEMS-4110

## Test plan:

Deploy and watch the perseus service logs. Side-by-side errors should decrease.

### DIFF
--- a/packages/perseus-core/src/widgets/input-number/to-numeric-input.test.ts
+++ b/packages/perseus-core/src/widgets/input-number/to-numeric-input.test.ts
@@ -208,4 +208,42 @@ describe("convertInputNumberOptionsToNumericInput", () => {
             "mixed",
         ]);
     });
+
+    it("does not exclude `decimal` and `integer` from `answerForms` when the answer has 10 decimal places", () => {
+        const options: PerseusInputNumberWidgetOptions = {
+            ...baseOptions,
+            answerType: undefined,
+            inexact: undefined,
+            value: "0.1231231234",
+        };
+
+        const result = convertInputNumberOptionsToNumericInput(options);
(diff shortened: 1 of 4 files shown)


packages/perseus-core/src/widgets/input-number/to-numeric-input.test.ts
```diff
@@ -138,4 +138,74 @@ describe("convertInputNumberOptionsToNumericInput", () => {
...
+        const result = convertInputNumberOptionsToNumericInput(options);
+        expect(result.answers).toHaveLength(1);
+        expect(result.answers[0].answerForms).toEqual([
+            "proper",
+            "improper",
+            "mixed",
+        ]);
+    });
+
+    it("excludes `decimal` and `integer` from `answerForms` when answerType is undefined, inexact is false, and `value` has more than 10 decimal places", () => {
```
SonicScrewdriver, line 159: I love the coverage!

SonicScrewdriver (approved):
Looks good to me, I really appreciated all the documentation to help me get context!

PR #3635, opened 2026-05-18
IG: linter rule when missing description and contains locked figures

handeyeco:
## Summary:

This PR will likely be closed in favor of: https://github.com/Khan/perseus/pull/3640

---

Just a simple rule for checking for a description when there are locked figures.

While doing this work, I ran into some tech debt that I think needs to be addressed. I didn't want to bloat the PR, so I followed established patterns and added new tickets to circle back to:

- [LEMS-4138](https://khanacademy.atlassian.net/browse/LEMS-4138): test should use IG generators
- [LEMS-4139](https://khanacademy.atlassian.net/browse/LEMS-4139): break up the `interactive-graph-widget-error` rule
- [LEMS-4140](https://khanacademy.atlassian.net/browse/LEMS-4140): `expectPass` and `expectWarning` should not contain `it`

Issue: LEMS-4125

[LEMS-4138]: https://khanacademy.atlassian.net/browse/LEMS-4138?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LEMS-4139]: https://khanacademy.atlassian.net/browse/LEMS-4139?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LEMS-4140]: https://khanacademy.atlassian.net/browse/LEMS-4140?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3640, opened 2026-05-19
LockedFigures/Description accessible check

handeyeco:
## Summary:
Add check in `interactiveGraphWidgetLogic.accessible` to make sure there's a description if an IG has locked figures.

Also adds a message to content creators.

Screenshot 2026-05-19 at 2 02 03 PM

Issue: LEMS-4125

Original PR that this one replaces: https://github.com/Khan/perseus/pull/3635


packages/perseus-core/src/widgets/interactive-graph/accessible.ts
catandthemachines, line 1: Yay! Thank you for moving it into it's own function for easier testing ❤️

packages/perseus-core/src/widgets/interactive-graph/accessible.ts
```diff
@@ -0,0 +1,34 @@
...
+    widgetOptions: PerseusInteractiveGraphWidgetOptions,
+): boolean {
+    // inaccessible if the interactive graph contains
+    // a protractor.
+    if (widgetOptions.showProtractor) {
+        return false;
+    }
+
+    // inaccessible if the interactive graph contains
+    // a graphie background image.
```
catandthemachines, line 16: Could you add a test for the graphie background case?

packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
```diff
@@ -415,6 +415,14 @@ class InteractiveGraphEditor extends React.Component<Props> {
             graph = <div className="perseus-error">{this.props.valid}</div>;
         }
 
+        // NOTE(matthewc): this duplicates some logic in `interactiveGraphWidgetLogic.accessible`.
+        // Ideally it wouldn't be duplicated, but the `isAccessible` API doesn't provide
+        // feedback as to why something is inaccessible. It could be in the linter which does
+        // provide specific feedback, but then we don't get the "is accessible" metrics we want
+        const showLockedFigureDescriptionWarning =
```
catandthemachines: My understanding from Jeanette's request is to have this in the description permanently not conditionally. Could you post this in our teams channel and confirm this functionality?

jeanettehead: I think you and I [agreed on only requiring this when there are locked figures present](https://khanacademy.slack.com/archives/C0AHH9H21H7/p1778263063994099?thread_ts=1778258120.422909&cid=C0AHH9H21H7) (since we have confidence the graph describes itself well if only interactive things are present).

catandthemachines: This is the additional description to the description field "_add a description to the description field that notes that interactive elements are automatically describe but locked figures are not._". So content folks can be aware of the limitation before getting the linter warning.

handeyeco: This seems like a wide-open two-way door given that it's easy to change and for an internal tool, so I'm just going to take the simplest path (always show the message) and we can iterate based on feedback.

handeyeco: Screenshot 2026-05-21 at 9 27 41 AM

catandthemachines (approved):
Thank you!

PR #3641, opened 2026-05-19
[ColorSync][Interactive Graph] Support exising orange locked figures after color migration

nishasy:
## Summary:
The new math learning tokens use `gold` instead of `orange`. I updated our locked figures colors as such,
but that means existing content with `orange` locked figures started erroring out.

We need to support `orange` for existing content, at least until some sort of backfill can happen.

Issue: none

## Test plan:
Storybook
- `/?path=/story/editors-editorpage--demo`
- Create a new interactive graph
- Add a locked point
- Use the JSON editor to change the locked point color to orange
- Go back to the editor view
- Confirm the selected value is "gold" and the locked figure is still colored appropriately

| setting | graph | json|
| --- | --- | --- |
 | image | image | image |

### DIFF
--- a/packages/perseus-core/src/data-schema.ts
+++ b/packages/perseus-core/src/data-schema.ts
@@ -990,6 +990,8 @@ export const lockedFigureColorNames = [
     "purple",
     "pink",
     "red",
+    // deprecated
+    "orange",
 ] as const;
 
 export type LockedFigureColor = (typeof lockedFigureColorNames)[number];
@@ -1002,6 +1004,9 @@ export const lockedFigureColors: Record<LockedFigureColor, string> = {
     pink: "var(--wb-semanticColor-learning-math-foreground-pink)",
     purple: "var(--wb-semanticColor-learning-math-foreground-purple)",
     red: "var(--wb-semanticColor-learning-math-foreground-red)",
+    // Deprecated: "orange" is the old name for the current "gold" color.
+    // We still have to support this in existing content.
+    orange: "var(--wb-semanticColor-learning-math-foreground-gold)",
 } as const;
 
 export type StrokeWeight = "thin" | "medium" | "thick";
--- a/packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/color-select.tsx
+++ b/packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/color-select.tsx
@@ -1,4 +1,4 @@
-import {lockedFigureColors} from "@khanacademy/perseus-core";
+import {lockedFigureColorNames} from "@khanacademy/perseus-core";
 import {View} from "@khanacademy/wonder-blocks-core";
 import {OptionItem, SingleSelect} from "@khanacademy/wonder-blocks-dropdown";
 import {Strut} from "@khanacademy/wonder-blocks-layout";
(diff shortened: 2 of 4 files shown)


packages/perseus-core/src/data-schema.ts
```diff
@@ -990,6 +990,8 @@ export const lockedFigureColorNames = [
     "purple",
     "pink",
     "red",
+    // deprecated
+    "orange",
 ] as const;
```
mark-fitzgerald, line 995: Just curious (I know it's not part of this PR) - Can't this array be derived from the keys of the `lockedFigureColors` object?

mark-fitzgerald (approved):

packages/perseus-core/src/data-schema.ts
nishasy: IIRC, it was for type safety? We wouldn't want to type `lockedFigureColors` as `Record<string, string>`.

(@jeremywiebe please correct me if I'm wrong. I think we worked on this code together.)

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/color-select.tsx
```diff
@@ -30,7 +32,9 @@ const ColorSelect = (props: Props) => {
                 color
                 <Strut size={spacing.xxSmall_6} />
                 <SingleSelect
-                    selectedValue={selectedValue}
+                    selectedValue={
+                        selectedValue === "orange" ? "gold" : selectedValue
```
handeyeco, line 36: Do we need to maintain this in code? I wonder if this would be a good use for the parser instead; that way all Perseus data gets converted before it enters the main flow of our code and we don't have to handle legacy values indefinitely.

nishasy: That would be great, although I think the parser refers directly to the list of colors from `data-schema.ts`, so I wasn't really sure if that would be the right way to go.

benchristel: I agree. It's not clear to me why we had to rename orange to gold in the first place, but now that both exist in our data, the parsers should be adjusted to convert both to one standard color name.

benchristel: I believe what Matthew is suggesting is that the parsers should accept both `gold` and `orange` but convert them to `gold`. Something like:

```ts
const parseLockedFigureColor = pipeParsers(enumeration("orange", "gold", ...))
  .then(convert((color) => color === "orange" ? "gold" : color))
  .parser,
```

Currently, the parsers reference `lockedFigureColorNames` from data-schema, but they could just as easily duplicate the list or add more elements. Duplicating that is not harmful because the type system keeps everything in sync.

packages/perseus-core/src/data-schema.ts
jeremywiebe: We want the data-schema.ts to be self-contained (ie. it shouldn't import anything) because the types here are the source of truth. If we derive them from other types (like ones for our widget React components), then the schema types can change implicitly as other types in the system change.

PR #3642, opened 2026-05-19
[Image] Remove scale flag and old sizing logic (make scale feature permanent)

nishasy:
## Summary:
Now that the scale feature has been on for everyone without any issues for a while, we can make it permanent.

- Remove the old logic
- Remove the scale flag

Note that the flag will also have to be removed from `frontend` after this.

Issue: https://khanacademy.atlassian.net/browse/LEMS-3912

## Test plan:
`pnpm jest`
`pnpm typecheck`

Make sure there are no regressions in the UI tests below.


catandthemachines:
Is this changing in spacing expected behavior?
image

catandthemachines (commented):
Had a few questions 😊

nishasy:
@catandthemachines The DOM differences in those snapshots are expected because I changed how the zoom is handled for images without sizes. But I'll look into getting rid of the 2px difference.

nishasy:
@catandthemachines I think I'm in favor of allowing the 2px caption shift for images that don't have size. I tried to make it consistent by adding `display: block` in the image loader to match the style for images that have sizes, but it causes markdown images within radio options to no longer support inline styles.

Do you think that's okay, or would you like to pair on it and brainstorm another idea?

catandthemachines (approved):
Looks good to me. The style differences look like a reasonable trade off for the improved functionality.

PR #3643, opened 2026-05-19
[Interactive Graph] Wire pointLabels through point graphs

ivyolamit:
## Summary:
- Adds the foundation that PRs 3–7 will reuse: state plumbing on `InteractiveGraphStateCommon`, a `buildPointAriaLabel` helper, an optional `pointLabel` prop bundle on `CoordInput`, and a per-type-aware `changePointLabels` editor handler that already covers all 15 graph types so follow-up PRs don't have to touch it.
- Wires it through the **point** graph end-to-end. After this PR, the LEMS-3995 reference question announces *"Point T at zero comma zero"* instead of *"Point 1 at zero comma zero"*.
- Existing content is unchanged: when `pointLabels` is unset, the announcement falls back to the original numeric `"Point 1"`, `"Point 2"`, … default.
- The author-facing tip that originally sat under the "Start coordinates" heading was dropped before merge — only `point` graphs expose the name field in PR 2 (polygon shares `start-coords-point.tsx` but is gated off in `start-coords-settings.tsx` until PR 3 wires its render side), so a global tip would mislead authors of every other graph type. It will return as a tooltip on the per-row name field once PRs 3–7 land (tracked in the plan doc's Follow-up work section).

## The bug

Interactive coordinate-plane questions ask the learner to plot a specific named point (e.g. *"Plot point T to complete the rectangle"*), but JAWS announces it as *"Point 1 at zero comma zero"*. The label is hard-coded to a numeric index in `useControlPoint`, so it never matches the prompt.

[PR 1](https://github.com/Khan/perseus/pull/3605) added the optional `pointLabels` field to every interactive graph-type schema. The field's shape mirrors `coords` / `startCoords` per type: `[string, string, string]` for `angle` / `quadratic`, `[string, string]` for `linear` / `ray` / `vector` / `absolute-value`, and `string[]` for the variable-arity rest. This PR makes that field actually do something — for `point` graphs first.

Issue: LEMS-3995

## Test plan

- [x] `pnpm typecheck` — passes
- [x] `pnpm lint packages/perseus packages/perseus-core packages/perseus-editor` — passes
- [x] `pnpm prettier . --check` — passes
- [x] `pnpm jest packages/perseus/src/widgets/interactive-graphs packages/perseus-editor/src/widgets/interactive-graph-editor packages/perseus-editor/src/widgets/__tests__/interactive-graph-editor.test.tsx` — 1584 tests pass (28 pre-existing skips)
- [ ] Reviewer manual: open the `PointWithCustomLabel` story, enable Storybook a11y addon (or VoiceOver/JAWS), and confirm the announcement matches *"Point T at 0 comma 0"*.
- [ ] Reviewer manual: open the `PointWithDefaultLabel` story, enable Storybook a11y addon (or VoiceOver/JAWS), and confirm the announcement matches *"Point 1 at 0 comma 0"*. Confirm default stories without `pointLabels` are unchanged.

## PR Series and Follow-ups (PRs 3–7)

Independent and parallelizable. None of them touch `data-schema.ts`, the parser, `strings.ts`, `InteractiveGraphStateCommon`, `interactive-graph-editor.tsx`'s handler/serialization, or `buildPointAriaLabel`. They do widen the `type === "point"` gate in `start-coords-settings.tsx` to include each new graph type as it lights up.

- 👉🏼 **PR 2** — [Foundation + Point graph](https://github.com/Khan/perseus/pull/3643)
- 👉🏼 **follow-up PR** — [Refactor linear, point, polygon, and ray graphs to use usePointAriaLabel instead of buildPointAriaLabel](https://github.com/Khan/perseus/pull/3694)


packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-settings.tsx
```diff
@@ -44,10 +44,18 @@ type Props = PerseusGraphType & {
...
 
 const StartCoordsSettingsInner = (props: Props) => {
-    const {type, range, step, allowReflexAngles, onChange} = props;
+    const {
+        type,
+        range,
+        step,
+        allowReflexAngles,
+        onChange,
+        onChangePointLabels,
```
catandthemachines, line 57: nit: Might as well expose the `pointLabels` since you also expose `onChangePointLabels`.

packages/perseus/src/strings.ts
```diff
@@ -135,7 +135,7 @@ export type PerseusStrings = {
         x,
         y,
     }: {
-        num: number;
+        num: number | string;
```
catandthemachines, line 142: small nit, it would be great to change this property to `label`, since that better identifies it's purpose.

packages/perseus/src/widgets/interactive-graphs/graphs/point.test.ts
```diff
@@ -56,4 +56,29 @@ describe("getPointGraphDescription", () => {
...
+    });
+
+    it("falls back to numeric defaults for indices without a custom label", () => {
+        const state: PointGraphState = {
+            ...baseState,
+            coords: [
+                [0, 0],
+                [1, 1],
+            ],
+            pointLabels: ["T"],
```
catandthemachines, line 78: Question, say a content creator added a label for the second coordinate, but not the first. Would `pointLabels: [undefined, "T"]`?

packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
```diff
@@ -195,6 +195,67 @@ class InteractiveGraphEditor extends React.Component<Props> {
...
+            case "circle":
+            case "exponential":
+            case "linear-system":
+            case "logarithm":
+            case "point":
+            case "polygon":
+            case "segment":
+            case "sinusoid":
+            case "tangent": {
+                const reshaped = [...pointLabels];
```
catandthemachines: Looking at this section of code, what it essentially says is "angle/quadratic has 3 point labels, absolute-value/linear/etc. has 2 point labels, none has no point labels, and everything else has varying number of point labels." Wouldn't it make more sense to just have two cases (graphs with point labels, and graphs without (none)) instead of having all these carve outs for 2 or 3 point labels? 🤔

catandthemachines:
Non-blocking issue, but could you post in the project channel and get a sanity check from Caitlyn if this is how we want to structure the label inputs for points?
image

ivyolamit:
> Non-blocking issue, but could you post in the project channel and get a sanity check from Caitlyn if this is how we want to structure the label inputs for points? image

Good call 👍🏼 Updated based on [discussion](https://khanacademy.slack.com/archives/C0AHH9H21H7/p1779474212715989) with Caitlyn, the label will be below the coordinates x and y fields.

image

packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
ivyolamit: That's a fair point. But this was intentional change so this will enforce the type check because the schema is per-type tuples, not a uniform array as defined in PR1 [here](https://github.com/Khan/perseus/pull/3605/changes#diff-17ad85d7f821dc080c7b004a2c0213fa3fad6c1e919a525d47d262202417b871).

In short,  the reshape is what makes the assignment nextGraph = {...graph, pointLabels: reshaped} type-check against PerseusGraphTypeAngle.pointLabels (which is [string, string, string], not string[]).

packages/perseus/src/strings.ts
ivyolamit: That's a good point.
I intentionally did not rename this for now for backward compatibility reason since this will also affect the existing translations The placeholder name in the source message (%(num)s at strings.ts:832) is what translators key their locale strings off of. Renaming source → %(pointLabel)s without updating every locale file in lockstep would silently break the substitution at runtime in non-English locales. That needs coordination — feels safer to do it as a focused refactor PR after PRs 3–7 land rather than expand this one.
For now I added a todo and logged this in the plan as follow-up work after all the graph code changes.

packages/perseus/src/widgets/interactive-graphs/graphs/point.test.ts
ivyolamit: Current logic will handle this as "" as implemented [here](https://github.com/Khan/perseus/pull/3643/changes#diff-70058fe5a65c240e99a83354493cdbc3a19fd852dd0f990a7b3a47df93293313R39)

packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-settings.tsx
ivyolamit: I initially applied as suggested but turns out this will not be an easy change without having to update the data schema. The `pointLabels` can't be destructured above because `PerseusGraphTypeNone` (one variant of the `PerseusGraphType` union that `Props` extends) does not declare the field. The `"pointLabels" in props` guard narrows the union to the variants that do.
I'll just leave this as is.

packages/perseus-editor/src/components/coordinate-pair-input.tsx
```diff
@@ -56,7 +60,11 @@ const CoordinatePairInput = (props: Props) => {
     return (
         <View style={[styles.row, style]}>
             <BodyText tag="label" style={styles.row}>
-                {labels ? labels[0] : "x coord"}
+                {labelStyle != null ? (
+                    <View style={labelStyle}>{xLabel}</View>
+                ) : (
+                    xLabel
+                )}
```
handeyeco, line 69: Couldn't `labelStyle` just default to an empty style object when destructuring from props so we don't need this conditional?

```suggestion
                <View style={labelStyle}>{xLabel}</View>
```

packages/perseus-editor/src/components/coordinate-pair-input.tsx
```diff
@@ -71,7 +79,11 @@ const CoordinatePairInput = (props: Props) => {
             <Strut size={spacing.medium_16} />
 
             <BodyText tag="label" style={styles.row}>
-                {labels ? labels[1] : "y coord"}
+                {labelStyle != null ? (
```
handeyeco, line 85: Same here

packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/coord-input.tsx
```diff
@@ -12,13 +13,47 @@ interface CoordInputProps {
     label: string;
     coord: Coord;
     onChange: (coord: Coord) => void;
+    // Custom label for each interactive point that will help with the screen reader.
+    pointLabel?: {
+        value: string | undefined;
+        placeholder: string;
```
handeyeco: Would it make sense to just derive the placeholder from the input label?

packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/coord-input.tsx
```diff
@@ -12,13 +13,47 @@ interface CoordInputProps {
...
+        onChange: (newLabel: string) => void;
```
handeyeco: Could we just reuse the `onChange` above?

packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-point.tsx
```diff
@@ -7,10 +7,12 @@ import type {Coord} from "@khanacademy/perseus";
 type Props = {
     startCoords: Coord[];
     onChange: (startCoords: Coord[]) => void;
+    pointLabels?: ReadonlyArray<string>;
+    onChangePointLabels?: (pointLabels: ReadonlyArray<string>) => void;
```
handeyeco, line 11: Could we just reuse the onChange above?

packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
```diff
@@ -195,6 +195,67 @@ class InteractiveGraphEditor extends React.Component<Props> {
...
+                    pointLabels[0] ?? "",
+                    pointLabels[1] ?? "",
+                    pointLabels[2] ?? "",
+                ];
+                nextGraph = {...graph, pointLabels: reshaped};
+                if (correct.type === graph.type) {
+                    nextCorrect = {...correct, pointLabels: reshaped};
+                }
+                break;
+            }
```
handeyeco: Could be nice to break these out into their own functions. It's manageable now, but I could this switch growing over time.

packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
```diff
@@ -231,6 +292,11 @@ class InteractiveGraphEditor extends React.Component<Props> {
                     type: correct.type,
                     startCoords:
                         this.props.graph && getStartCoords(this.props.graph),
+                    ...(this.props.graph &&
+                    "pointLabels" in this.props.graph &&
+                    this.props.graph.pointLabels
```
handeyeco, line 248: Would this essentially be the same?

```suggestion
                    this.props.graph?.pointLabels
```

packages/perseus/src/widgets/interactive-graphs/graphs/components/build-point-aria-label.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import type {PerseusStrings} from "../../../../strings";
+import type {vec} from "mafs";
+
+/**
+ * Build a screen-reader aria-label for an interactive point using a
+ * caller-supplied custom label (e.g. "T") in place of the numeric default.
+ * Returns `undefined` when no custom label is set for this index — callers
+ * should fall back to the existing numeric "Point N" announcement built
+ * inside `useControlPoint`.
```
handeyeco: Would it make sense to have this function provide the default "Point N" instead of the consumer providing it?

packages/perseus/src/widgets/interactive-graphs/graphs/components/build-point-aria-label.test.ts
```diff
@@ -0,0 +1,55 @@
...
+        ).toBe("Point B at -1 comma 2.");
+    });
+
+    it("returns undefined for non-string entries (null, undefined, number) — defensive against malformed hand-authored JSON bypassing the parser", () => {
+        // eslint-disable-next-line no-restricted-syntax -- cast simulates malformed JSON the parser would reject
+        const labels = [
+            null,
+            undefined,
+            42,
+        ] as unknown as ReadonlyArray<string>;
```
handeyeco, line 44: In what scenario would this happen?

packages/perseus-editor/src/widgets/__tests__/interactive-graph-editor.test.tsx
```diff
@@ -871,4 +871,155 @@ describe("InteractiveGraphEditor", () => {
...
+                    graph={{type: "angle"}}
+                    correct={{type: "angle"}}
+                    onChange={onChangeMock}
+                    ref={ref}
+                />,
+                {wrapper: RenderStateRoot},
+            );
+
+            // Act
+            ref.current?.changePointLabels(["A", "B", "C"]);
```
handeyeco: It would be my preference if we didn't call methods on class components, even in tests. It locks us even more into class-based component and adds more barriers to switching to functional components.

packages/perseus/src/widgets/interactive-graphs/graphs/point.test.ts
```diff
@@ -56,4 +56,58 @@ describe("getPointGraphDescription", () => {
...
     });
+
+    it("uses the custom point label when pointLabels is set", () => {
+        const state: PointGraphState = {
+            ...baseState,
+            coords: [[0, 0]],
+            pointLabels: ["T"],
+        };
+        expect(getPointGraphDescription(state, mockPerseusI18nContext)).toBe(
+            "Interactive elements: Point T at 0 comma 0.",
```
handeyeco, line 67: Probably not related to your PR, but should it be plural? "Interactive elements", not "Interactive element"?

packages/perseus/src/widgets/interactive-graphs/graphs/point.tsx
```diff
@@ -79,6 +81,8 @@ function PointGraph(props: Props) {
 
 function LimitedPointGraph(statefulProps: StatefulProps) {
     const {dispatch} = statefulProps;
+    const {pointLabels} = statefulProps.graphState;
+    const {strings, locale} = usePerseusI18n();
```
handeyeco, line 85: Random thought that's not a blocker: would it make sense to have `usePointAriaLabel`? Because if you had a custom hook, that hook could use `usePerseusI18n` directly and it might simplify the API of using `buildPointAriaLabel`.

packages/perseus/src/widgets/interactive-graphs/reducer/initialize-graph-state.ts
```diff
@@ -43,6 +43,7 @@ export function initializeGraphState(
         hasBeenInteractedWith: false,
         range,
         snapStep,
+        pointLabels: "pointLabels" in graph ? graph.pointLabels : undefined,
```
handeyeco, line 46: I'm guessing you have to do it the way you did for TS and that this won't work:

```suggestion
        pointLabels: graph?.pointLabels,
```

packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.test.tsx
```diff
@@ -175,6 +175,61 @@ describe("StatefulMafsGraph", () => {
...
+
+        // Act: rerender with a custom label on the only point.
+        rerender(
+            <StatefulMafsGraph
+                {...baseProps}
+                graph={{
+                    type: "point",
+                    numPoints: 1,
+                    coords: [[0, 0]],
+                    pointLabels: ["T"],
```
handeyeco, line 212: Would this ever happen outside of the editing experience?

handeyeco (approved):
My main concern is that I think some of this could be made less complicated.

- Do we need to add custom onChange handlers? Maybe the answer is "yes" but most everything else has gotten by reusing the existing onChange handler.
- When passing point labels, does it need to be an object with a custom placeholder and its own onChange?

Other than that, just random thoughts and questions.

packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-point.tsx
ivyolamit: The reason why I didn't reuse the `onChange` above is because the label change handle is for string compared to the one above it's meant for Coord which is expecting two numbers. I could reuse it but that would mean I would need to handle multiple type for it to cater the x and y coordinate inputs plus the label input. Separating it makes it more explicit handling.

packages/perseus-editor/src/components/coordinate-pair-input.tsx
ivyolamit: Good point! But i'll hold off the change for now, instead i'll add a TODO comment. I already have identified a work here after the custom label changes to update this code to change the old way of styling here. I'll include simplifying this in that PR. In that way all the applicable graphs that will have custom label will apply this styling below

Image

while the locked figures using the same component will retain its styling

Image

packages/perseus/src/strings.ts
catandthemachines: That sounds perfect. Thank you ❤️

packages/perseus-editor/src/widgets/__tests__/interactive-graph-editor.test.tsx
catandthemachines, line 1: Thank you for the update!

packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-shared.module.css
```diff
@@ -32,10 +32,27 @@
     gap: var(--wb-sizing-size_120) !important;
 }
 
+.label-field {
+    display: flex !important;
```
catandthemachines, line 36: (not related to your change) I look forward to a day with fewer `!important` 🥲

packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-settings.tsx
```diff
@@ -176,10 +184,20 @@ const StartCoordsSettingsInner = (props: Props) => {
                 type === "point"
                     ? getPointCoords(props, range, step)
                     : getPolygonCoords(props, range, step);
+            // Only `point` consumes `pointLabels` at runtime in PR 2 —
+            // polygon's render side wires up in PR 3. Gate the editor
+            // name fields to match so polygon authors don't fill in
```
catandthemachines, line 189: Thank you for this comment!

packages/perseus/src/widgets/interactive-graphs/graphs/components/build-point-aria-label.ts
catandthemachines: I would be for that, it either returns the label "Point [Custom]" or "Point N" depending on whether there is a custom label or not.

catandthemachines (approved):
Looks good to me! Added thumbs up to some of Matthews suggestions, but overall this is looking great.

packages/perseus-editor/src/widgets/interactive-graph-editor/interactive-graph-editor.tsx
ivyolamit: Good point, it's not the same. Since not all graphs has pointLabels. This check is doing TypeScript type narrowing so that accessing this.props.graph.pointLabels on the next line is type-safe. The optional chaining only guards against null/undefined TS would complain that pointLabels isn't a property on every variant of the union.

packages/perseus/src/widgets/interactive-graphs/graphs/components/build-point-aria-label.test.ts
ivyolamit: This is simulating an edge case scenario when the content author would use JSON object when adding an interactive graph and not via the UI from our editor.

packages/perseus/src/widgets/interactive-graphs/graphs/components/build-point-aria-label.ts
ivyolamit: let me rephrase and simplify this comment here. The fallback is "Point 1", "Point 2", and so on. The "N" is the incremental value we have that's the current behavior in prod.

packages/perseus/src/widgets/interactive-graphs/graphs/point.test.ts
ivyolamit: good catch, the current translation for this is
```
srInteractiveElements: ({elements}) => `Interactive elements: ${elements}`,
```

if we want to correct grammar here we can pass in the count so we can pluralize the value. I'll not include that change here since this PR is already big. But since I have a follow-up work related to the translation for `srPointAtCoordinates` i'll include it in that work. I'm adding this in my Follow-up plan.

packages/perseus/src/widgets/interactive-graphs/graphs/point.tsx
ivyolamit: That's a good point. Something I haven't considered. Upon checking this approach, this will cascade to multiple things. The `buildPointAriaLabel` is also being used in `getPointGraphDescription` which is used by `renderPointGraph`. For a simple localize forming aria label, for now with this 3 usage probably not the right time to do the refactor. I'll leave this as is for now.

packages/perseus/src/widgets/interactive-graphs/reducer/initialize-graph-state.ts
ivyolamit: yes unfortunately.

packages/perseus/src/widgets/interactive-graphs/stateful-mafs-graph.test.tsx
ivyolamit: This is specific for editing.

packages/perseus-editor/src/widgets/__tests__/interactive-graph-editor.test.tsx
ivyolamit: thanks for pointing this one out. I've updated this.

ivyolamit:
Chatted with @catandthemachines regrading the conflict in `point.tsx` added [here](https://github.com/Khan/perseus/pull/3674/changes#diff-e912784ffdfa53864f07de1621cac94b5cd5073c032d73dc860d60c213c9bed6R90) since the WB Announcer changes was just merged in first in main and this will cause the custom label not to be announce. Next steps:
1. Ivy will merge this PR 2
2. Will create a follow-up PR to ensure that the point graph will have the correct screen reader text in this line [here](https://github.com/Khan/perseus/pull/3619/changes#diff-3dd9ea20340a1e532705224ba35ca89fad5c4adbe4470ccb1178a2227ce0cf2fR13) can use the current `buildPointAriaLabel` function.
3. Will discuss with Cat how we want the rest of the PRs resolve since they're closely related.

PR #3649, opened 2026-05-21
You put the Numeric in the Input Number and you drink 'em both together

SonicScrewdriver:
## Summary:
This main purpose of this PR is to update Input Number so that it renders as a Numeric Input if the `input-number-to-numeric-input` flag is turned on. 

In order to do this, I needed to make too updates to the Numeric Input Widget:

1. I added a `blur()` method to Numeric input in order to satisfy the `Focusable` type so that we could update the old string ref to a more modern implementation. This seemed like the best approach to keep the logic as clean as possible. This method can be temporary until we remove the old render logic for Input Number, or we can keep it around so that Numeric Input follows the `Focusable` standard. However, it really doesn't seem like `blur() `ever gets called from anywhere. 
2. I had to add support to Numeric Input for rendering questions when the `answerForms` includes `percent`. This is a pre-existing bug with Numeric Input, despite the fact that we have logic to support `percent` during scoring. All that was required was to ensure that we consider `percent` when generating the example text. I've added a new story for Numeric Input for this answer form. 

I've also created a clone of the Input Number Widget story that hard-codes the `input-number-to-numeric-input` flag to `true`, so that we can easily compare with the original stories during the conversion process.

Issue: LEMS-4098

## Test plan:
- Tests Pass 
- Upcoming manual testing with blur / focus to ensure no regressions

### DIFF
--- a/packages/perseus/src/widgets/input-number/input-number-as-numeric-input.stories.tsx
+++ b/packages/perseus/src/widgets/input-number/input-number-as-numeric-input.stories.tsx
@@ -12,7 +12,7 @@ import * as React from "react";
 
 import {getFeatureFlags} from "../../testing/feature-flags-util";
 import {ServerItemRendererWithDebugUI} from "../../testing/server-item-renderer-with-debug-ui";
-import {getAnswerfulItem} from "../../util/test-utils";
+import {getAnswerfulItem, getAnswerlessItem} from "../../util/test-utils";
 
 import {question1, question2, question3} from "./input-number.testdata";
 
@@ -82,14 +82,11 @@ const meta: Meta = {
 };
 export default meta;
 
-type Question = PerseusRenderer;
-type InputNumberOptions = PerseusInputNumberWidgetOptions;
-
 const updateWidgetOptions = (
-    question: Question,
+    question: PerseusRenderer,
     widgetId: string,
-    options: InputNumberOptions,
-): Question => {
+    options: PerseusInputNumberWidgetOptions,
+): PerseusRenderer => {
     const widget = question.widgets[widgetId];
     return {
         ...question,
@@ -105,7 +102,9 @@ const updateWidgetOptions = (
     };
 };
 
-export const Rational = (args: InputNumberOptions): React.ReactElement => {
+export const Rational = (
+    args: PerseusInputNumberWidgetOptions,
+): React.ReactElement => {
     const question = updateWidgetOptions(question1, "input-number 1", args);
     return (
         <ServerItemRendererWithDebugUI
@@ -116,7 +115,9 @@ export const Rational = (args: InputNumberOptions): React.ReactElement => {
 };
 Rational.args = question1.widgets["input-number 1"].options;
 
-export const PiSimplify = (args: InputNumberOptions): React.ReactElement => {
+export const PiSimplify = (
+    args: PerseusInputNumberWidgetOptions,
+): React.ReactElement => {
     const question = updateWidgetOptions(question2, "input-number 1", args);
     return (
         <ServerItemRendererWithDebugUI
@@ -127,7 +128,9 @@ export const PiSimplify = (args: InputNumberOptions): React.ReactElement => {
 };
 PiSimplify.args = question2.widgets["input-number 1"].options;
 
-export const Percent = (args: InputNumberOptions): React.ReactElement => {
+export const Percent = (
+    args: PerseusInputNumberWidgetOptions,
+): React.ReactElement => {
     const question = updateWidgetOptions(question3, "input-number 1", args);
     return (
         <ServerItemRendererWithDebugUI
@@ -156,7 +159,7 @@ export const Answerful = (): React.ReactElement => {
 };
 
 export const Answerless = (): React.ReactElement => {
-    const item = getAnswerfulItem("input-number", {
+    const item = getAnswerlessItem("input-number", {
         simplify: "optional",
         size: "normal",
         value: 42,
--- a/packages/perseus/src/widgets/input-number/input-number-as-numeric-input.stories.tsx
+++ b/packages/perseus/src/widgets/input-number/input-number-as-numeric-input.stories.tsx
@@ -1,0 +1,173 @@
+// Mirrors `input-number.stories.tsx`, but renders every story with the
+// `input-number-to-numeric-input` feature flag enabled — so we can verify the
+// flag-on rendering side-by-side with the existing stories during the
+// migration. This file should be deleted once the flag is removed and
+// numeric-input is the only rendering path.
+import {
+    type PerseusRenderer,
+    type PerseusInputNumberWidgetOptions,
+    generateTestPerseusItem,
+} from "@khanacademy/perseus-core";
+import * as React from "react";
+
+import {getFeatureFlags} from "../../testing/feature-flags-util";
+import {ServerItemRendererWithDebugUI} from "../../testing/server-item-renderer-with-debug-ui";
+import {getAnswerfulItem} from "../../util/test-utils";
+
+import {question1, question2, question3} from "./input-number.testdata";
+
+import type {Meta} from "@storybook/react-vite";
+
+const apiOptionsWithFlag = {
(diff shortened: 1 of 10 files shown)


packages/perseus/src/widgets/input-number/input-number-as-numeric-input.stories.tsx
```diff
@@ -0,0 +1,173 @@
+// Mirrors `input-number.stories.tsx`, but renders every story with the
```
SonicScrewdriver, line 1: I thought about creating a new global "Flags" feature beside our "Theme" feature in Storybook, but opted to discuss that in another PR as it was getting messy. 

This is a temporary story so that we can test both Input Number (using the original story) and Numeric Input (using this new story) to ensure there's no regressions.

packages/perseus/src/widgets/numeric-input/numeric-input.testdata.ts
```diff
@@ -152,6 +152,25 @@ export const properProblem: PerseusRenderer = generateTestPerseusRenderer({
     },
 });
 
+// Demonstrates the renderer crash when "percent" appears in `answerForms`.
+// `NumericExampleStrings` (in utils.ts) is missing a `percent` entry even though
+// `MathFormat` lists it as valid.
+export const percentFormProblem: PerseusRenderer = generateTestPerseusRenderer({
```
SonicScrewdriver, line 158: Turns out we never provided rendering support for _rendering_ Numeric Input with the percent form, despite the fact that it's considered an acceptable `answerForm` in both types/scoring. This oversight was resulting in a rendering error for any Numeric Input with `answerForm` of "percent". 

I've added this new story after fixing support.

_Note: We currently don't have "percent" as a selectable `answerForm` in the Numeric Input **Editor**. This should be added in a new ticket, as we already have the full scoring logic to support it._

packages/perseus/src/widgets/numeric-input/utils.ts
```diff
@@ -25,6 +25,7 @@ const NumericExampleStrings: {
             : strings.simplifiedImproperExample,
     mixed: (form, strings) => strings.mixedExample,
     decimal: (form, strings) => strings.decimalExample,
+    percent: (form, strings) => strings.percentExample,
```
SonicScrewdriver, line 28: This is what was causing the rendering error. :) It was a simple enough fix.

packages/perseus/src/widgets/numeric-input/numeric-input.class.tsx
```diff
@@ -90,6 +90,13 @@ export class NumericInput
         return true;
     };
 
+    // [LEMS-4085] While we cannot find any callers of this method,
+    // adding it is the simplest way to resolve temporary type issues
+    // regarding rendering the Input Number widget as a Numeric Input
+    blur: () => void = () => {
+        this.inputRef.current?.blur();
+    };
+
```
SonicScrewdriver, line 99: I'm fully open to push back on this. We were previously using string refs for invoking our focus methods on input number. Unfortunately, the lack of a "blur" method on Numeric Input was resulting in type conflicts as it was not matching the `Focusable` interface. 

I suspect that we'll be able to clean out the old render paths in Input Number after receiving QE approval to remove the feature flag, at which point we may revisit whether `blur()` is still needed here, or if we want to keep it as a Focusable-compatibility method.

This seemed faster / cleaner than having to have separate refs, or diving too deeply into refactoring our focus handling as I felt that was getting outside the scope of this ticket.

packages/perseus/src/widgets/input-number/input-number.tsx
```diff
@@ -108,6 +118,10 @@ class InputNumber extends React.Component<Props> implements Widget {
         });
     }
 
+    private handleInputRef = (instance: Focusable | null): void => {
```
SonicScrewdriver, line 121: Using a callback ref instead of a normal one so we can share a single Focusable-typed ref across all three of our render targets (NumericInput, SimpleKeypadInput, InputWithExamples).

packages/perseus/src/widgets/input-number/input-number-as-numeric-input.stories.tsx
```diff
@@ -0,0 +1,173 @@
...
+    return (
+        <ServerItemRendererWithDebugUI
+            item={item}
+            apiOptions={apiOptionsWithFlag}
+        />
+    );
+};
+
+export const Answerless = (): React.ReactElement => {
+    const item = getAnswerfulItem("input-number", {
```
benchristel: Should this be `getAnswerlessItem`?


```suggestion
    const item = getAnswerlessItem("input-number", {
```

packages/perseus/src/widgets/input-number/input-number-as-numeric-input.stories.tsx
```diff
@@ -0,0 +1,173 @@
...
+        },
+        rightAlign: {
+            control: {type: "boolean"},
+        },
+    },
+};
+export default meta;
+
+type Question = PerseusRenderer;
+type InputNumberOptions = PerseusInputNumberWidgetOptions;
```
benchristel: Why do we alias these types?

packages/perseus/src/widgets/numeric-input/numeric-input.class.tsx
benchristel: +1, I think this is the best solution.

packages/perseus/src/widgets/numeric-input/numeric-input.class.tsx
```diff
@@ -90,6 +90,13 @@ export class NumericInput
         return true;
     };
 
+    // [LEMS-4085] While we cannot find any callers of this method,
```
benchristel: Can you change this to


```suggestion
    // TODO(LEMS-4085): While we cannot find any callers of this method,
```

please? Just for consistency and to be sure we find it later.

packages/perseus/src/widgets/input-number/input-number.tsx
```diff
@@ -86,6 +92,10 @@ class InputNumber extends React.Component<Props> implements Widget {
     static contextType = PerseusI18nContext;
     declare context: React.ContextType<typeof PerseusI18nContext>;
 
+    // [LEMS-4085] Single ref shared by both the legacy and NumericInput render paths.
```
benchristel:
```suggestion
    // TODO(LEMS-4085): Single ref shared by both the legacy and NumericInput render paths.
```

packages/perseus/src/widgets/input-number/input-number.tsx
```diff
@@ -191,11 +191,28 @@ class InputNumber extends React.Component<Props> implements Widget {
     }
 
     render(): React.ReactNode {
+        // [LEMS-4085] This logic renders the Numeric Input widget instead of the
```
benchristel:
```suggestion
        // TODO(LEMS-4085): This logic renders the Numeric Input widget instead of the
```

benchristel (approved):
This looks great! I left a few small suggestions inline.

packages/perseus/src/widgets/numeric-input/numeric-input.testdata.ts
benchristel: I'm actually not sure we want percent as a selectable `answerForm`. We should at least check with content creators first. They might have guidelines around this; e.g. they might prefer to hardcode the % sign in the content and let the learner enter an integer for the percentage.

packages/perseus/src/widgets/input-number/input-number-as-numeric-input.stories.tsx
SonicScrewdriver: Great question, I don't know! I basically just copied the story directly with the flag. This seems needless so I'll update both stories :)

packages/perseus/src/widgets/input-number/input-number-as-numeric-input.stories.tsx
SonicScrewdriver: It should! I copied this from the original story file. I'll fix it there too :)

PR #3652, opened 2026-05-21
GAAD: Remove a11y lint warning in perseus-api.tsx

handeyeco:
## Summary:

There was a warning that `<a>` elements needed content. I believe it's valid to do `<a children={children} />` which is essentially what `<a {...props} />` was doing, but the linter didn't think so. I think my change is a good alternative.


mark-fitzgerald (approved):
I think this makes it more clear, as well. Thanks!

PR #3654, opened 2026-05-21
GAAD: Remove some redundent roles

handeyeco:
## Summary:
Seems like an easy fix, but I'm wondering if they were intentional?


catandthemachines (approved):
I agree this looks unnecessary. Approved!

handeyeco:
Anakaren said these were intentional

PR #3655, opened 2026-05-21
GAAD: Convert clickable span into a button

handeyeco:
## Summary:

I think it's fine...it's just in a test...


catandthemachines (approved):
Looks good to me. Thanks!

PR #3656, opened 2026-05-21
Add a v1 version of input-number options; migrate from v0

benchristel:
## Summary:
We are in the process of replacing `input-number` with `numeric-input`.  See
[LEMS-4081] for context.

This PR changes the input-number options schema to match the options of
`numeric-input`.  The idea is that the migrated options can be passed directly
to a `NumericInput` component, paving the way for us to replace `InputNumber`
with `NumericInput` simply by registering `NumericInput` under the
`input-number` widget name.

[LEMS-4081]: https://khanacademy.atlassian.net/browse/LEMS-4081

Issue: LEMS-4111

## Test plan:

CI checks should pass.


packages/perseus-core/src/parse-perseus-json/perseus-parsers/widgets-map.test.ts
```diff
@@ -384,11 +384,27 @@ describe("parseWidgetsMap", () => {
...
+                                "improper",
+                                "mixed",
+                            ],
+                            message: "",
+                            maxError: 0,
+                        },
+                    ],
                     size: "normal",
-                    value: "",
+                    coefficient: false,
```
benchristel, line 407: We could use the generator functions here. It might be good to update all these tests to use them.

packages/perseus-core/src/data-schema.ts
```diff
@@ -2219,8 +2219,35 @@ export type PerseusInputNumberAnswerType =
...
+     */
+    answerForms: MathFormat[];
+    /**
+     * Whether we should check the answer strictly against the configured
+     * answerForms (strict = true) or include the set of default answerForms
+     * (strict = false).
+     */
+    strict: boolean;
+    maxError?: number | undefined;
+    /** Unsimplified answers are Ungraded, Accepted, or Wrong. */
```
benchristel, line 2245: This comment could be better.

packages/perseus-core/src/data-schema.ts
```diff
@@ -2230,6 +2257,39 @@ export type PerseusInputNumberWidgetOptions = {
...
+    /**
+     * A list of correct and incorrect answers. Each answer can have a
+     * message explaining why it is correct/incorrect. There may be
+     * multiple correct answers if multiple formats are accepted. For
+     * example, some questions might accept either a fraction or a
+     * decimal as correct.
+     *
+     * The first answer that matches (correct or incorrect) is the scoring
+     * result (so order of answers is very important).
+     */
```
benchristel, line 2271: This comment isn't really accurate since all input-number widgets will have exactly one answer. Maybe change this type to `[PerseusInputNumberAnswer]`. Then we could get rid of some optional chaining, and we wouldn't even need assertions!

PR #3660, opened 2026-05-22
[Image] Fix flakey graphie snapshot again

nishasy:
## Summary:
I did a deep dive with Mark and we figured out that the source of the flakiness
is coming from inconsistent span height, which was coming from fonts not having
been loaded yet.

To fix this, I'm adding a small timeout (5 milliseconds) to let the fonts load
before attempting to calculate the label margins.

Issue: https://khanacademy.atlassian.net/browse/LEMS-4053

## Test plan:
- check below for UI Test updates
- marginTop should be `-15.389285714285714` for all labels

### DIFF
--- a/.changeset/four-humans-pull.md
+++ b/.changeset/four-humans-pull.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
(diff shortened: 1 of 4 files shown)


SonicScrewdriver (approved):
Thank you for investigating this, it seems like a tricky one!

PR #3661, opened 2026-05-22
[Renderer] Add regression stories for text

mark-fitzgerald:
## Summary:
This PR adds Chromatic visual regression stories for the Perseus renderer,
covering common text content in both Article and Exercise contexts. The stories
test regular text (with headings at multiple levels) and list content (ordered,
unordered, and nested).

Issue: LEMS-4102

## Test plan:
1. Run `pnpm storybook` and navigate to **Renderers > Visual Regression Tests**
   to verify all four stories render correctly: ArticleRegularText,
   ArticleLists, ExerciseRegularText, ExerciseLists.


packages/perseus/src/__docs__/renderer-initial-state-regression.stories.tsx
```diff
@@ -0,0 +1,76 @@
...
+    tags: ["!manifest"],
+    parameters: {
+        docs: {
+            description: {
+                component: "Examples of graphics in dark mode.",
+            },
+        },
+        chromatic: {disableSnapshot: false, modes: themeModes},
+    },
+};
```
nishasy, line 29: Does this Meta purposefully not have a type or a `component` field?

packages/perseus/src/__docs__/renderer-initial-state-regression.stories.tsx
```diff
@@ -0,0 +1,76 @@
...
+export default meta;
+
+const RenderContent = (
+    content: string,
```
nishasy: Can you add a comment explaining that these stories are specifically for the paragraph styles (_not_ for widgets), which is why it's okay that this doesn't take any widget info?

packages/perseus/src/__testdata__/renderer.testdata.ts
```diff
@@ -198,3 +197,50 @@ export const mockedRandomItem: PerseusRenderer = {
...
+
+export const listsContent = `## Primary Federation ships involved in the tachyon blockade during the Klingon Civil War:
+
+1. USS Enterprise-D
+1. USS Excalibur
+1. USS Merrimack
+1. USS Sutherland
+1. USS Charleston
+1. USS Hood
+1. USS Tian An Men
```
nishasy, line 223: Are all the list numbers supposed to be 1?

nishasy (approved):
Left a couple comments, but makes sense to me.

Question: Is there anything else that we should be testing in the renderer paragraphs? Like bold or italics or other formatting? Or is this everything that's relevant to `packages/perseus/src/styles/paragraph-styles.css`?

mark-fitzgerald:
> Question: Is there anything else that we should be testing in the renderer paragraphs? Like bold or italics or other formatting? Or is this everything that's relevant to `packages/perseus/src/styles/paragraph-styles.css`?

I included bold and italics in the regular text story. But, to your main question, this is focused on items affected by the paragraph styles. We can do additional PRs for regular content, like tables, math, etc.

packages/perseus/src/__testdata__/renderer.testdata.ts
mark-fitzgerald: Markdown doesn't care what the numbers are in a list - it will render it with the proper order. I like to use just ones so that I can re-order the items and not have to renumber everything.

packages/perseus/src/__docs__/renderer-initial-state-regression.stories.tsx
mark-fitzgerald: What would the component be? Just Renderer?

packages/perseus/src/__docs__/renderer-initial-state-regression.stories.tsx
mark-fitzgerald: Done.

packages/perseus/src/__docs__/renderer-initial-state-regression.stories.tsx
nishasy: Yeah, I think it would just be `Renderer`.

mark-fitzgerald: Does this look correct, now:
```
const meta: Meta<PerseusRenderer> = {
```

packages/perseus/src/__docs__/renderer-initial-state-regression.stories.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+type Story = StoryObj;
+
+const meta: Meta<PerseusRenderer> = {
+    title: "Renderers/Visual Regression Tests",
+    tags: ["!manifest"],
+    parameters: {
+        docs: {
+            description: {
+                component: "Examples of graphics in dark mode.",
```
Myranae: Is this what you want? This PR is for text stories, right?

packages/perseus/src/__docs__/renderer-initial-state-regression.stories.tsx
```diff
@@ -0,0 +1,83 @@
...
+                    json={json}
+                    dependencies={storybookDependenciesV2}
+                />
+            );
+        } else {
+            return (
+                <div style={{padding: "24px"}}>
+                    <QuestionRendererForStories
+                        question={generateTestPerseusRenderer({
+                            content: content,
```
Myranae:
```suggestion
                            content,
```
Just a tiny nit Claude caught. Totally no worries if you just keep it as it is too. I like being able to shorten things like this though, so wanted to point out in case you do too :)

packages/perseus/src/__docs__/renderer-initial-state-regression.stories.tsx
```diff
@@ -0,0 +1,83 @@
...
+            },
+        },
+        chromatic: {disableSnapshot: false, modes: themeModes},
+    },
+};
+export default meta;
+
+const RenderContent = (
+    content: string,
+    asArticle: boolean,
```
Myranae: The  two separate component trees inside the if/else signaled a code smell to Claude. It looks like Boolean flag arguments can be an anti-pattern when it leads to two logically separate responsibilities. Claude gave the suggestion below, but it makes sense to me (and I wouldn't have recognized it XD). I think it's fine either way, but just a suggestion.

>  const RenderContent = (content: string, asArticle: boolean): ...
>   At call sites — RenderContent(regularTextContent, true) — the true is
>    opaque. Two separate factory functions would be clearer and
>   eliminate the if/else:
>   const renderAsArticle = (content: string) => function Render() { ...
>   };
>   const renderAsExercise = (content: string) => function Render() { ...
>    };

packages/perseus/src/__testdata__/renderer.testdata.ts
```diff
@@ -198,3 +197,50 @@ export const mockedRandomItem: PerseusRenderer = {
         },
     },
 } as PerseusRenderer;
+
+export const regularTextContent = `# Sensors and Scanners
```
Myranae, line 202: <3

Myranae (approved):
Looks good! Left a few comments :)

PR #3667, opened 2026-05-26
Use the `numeric-input` scoring code to score `input-number` widgets.

benchristel:
## Summary:
This changes behavior in only one case: when the input-number widget's answer
is between 0 and 1 and the answerType is `"number"`, user inputs formatted as
percentages are accepted as correct. Previously, percentage inputs were never
accepted when answerType was `"number"`.

More information: https://khanacademy.slack.com/archives/C0B0ZG7T4AF/p1779485472979299

Issue: LEMS-4112

## Test plan:

Deploy to the perseus service. There should be no more side-by-side scoring
mismatches in the logs, since the scoring logic is now identical for
input-number and numeric-input.

### DIFF
--- a/packages/perseus-core/src/validation.types.ts
+++ b/packages/perseus-core/src/validation.types.ts
@@ -44,7 +44,7 @@ import type {
     PerseusGraphCorrectType,
     MakeWidgetMap,
     PerseusFreeResponseWidgetScoringCriterion,
-    PerseusRenderer,
+    PerseusRenderer, PerseusInputNumberWidgetOptions,
 } from "./data-schema";
 import type {ErrorCode} from "./error-codes";
 import type {Relationship} from "./types";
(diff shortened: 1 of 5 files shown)


packages/perseus-core/src/validation.types.ts
```diff
@@ -628,7 +556,8 @@ export interface RubricRegistry {
     "graded-group": PerseusGradedGroupRubric;
     grapher: PerseusGrapherRubric;
     group: PerseusGroupRubric;
-    "input-number": PerseusInputNumberRubric;
+    // TODO(LEMS-4085): change to PerseusNumericInputRubric;
+    "input-number": PerseusInputNumberWidgetOptions;
```
benchristel, line 560: The `convertInputNumberOptionsToNumericInput` takes the `PerseusInputNumberWidgetOptions` type. We can't pass `PerseusInputNumberRubric` to it, since the rubric is missing the `size` property.

The rubric types are intended to be the parameters of the scoring functions. However, in production, the `PerseusInputNumberWidgetOptions` type is always passed to the scoring function. The only advantage of having the rubric types is that we don't have to specify unused properties like `size` in the tests for scoring.

Since the tests for `scoreInputNumber` will be removed as part of this project, I opted to add `size` there and change the types so `scoreInputNumber` takes `PerseusInputNumberWidgetOptions`.

packages/perseus-score/src/widgets/input-number/score-input-number.test.ts
```diff
@@ -201,12 +212,13 @@ describe("scoreInputNumber", () => {
...
+        const rubric: PerseusInputNumberWidgetOptions = {
             maxError: 0.1,
             inexact: false,
             value: 16.5,
             simplify: "optional",
             answerType: "decimal",
+            size: "normal",
         };
 
         const userInput: PerseusInputNumberUserInput = {
```
benchristel, line 224: Kind of interesting that I didn't have to meaningfully change any tests to swap out the scoring logic. I should add tests for the interesting cases we found during this project. I'll do that in a separate PR. ([LEMS-4185])

[LEMS-4185]: https://khanacademy.atlassian.net/browse/LEMS-4185

packages/perseus-core/src/validation.types.ts
SonicScrewdriver: Makes sense! :)

packages/perseus-score/src/widgets/input-number/score-input-number.ts
```diff
@@ -44,53 +44,21 @@ export const inputNumberAnswerTypes = {
...
-        earned: result.correct ? 1 : 0,
-        total: 1,
-        message: result.message,
-    };
+    return scoreNumericInput(
+        userInput,
+        convertInputNumberOptionsToNumericInput(rubric),
+        locale,
+    );
 }
```
SonicScrewdriver, line 62: Exciting! We're getting closer and closer

SonicScrewdriver (approved):
Looks great to me! Very excited to see this slowly get cleaned up.

PR #3668, opened 2026-05-27
 [Interactive Graph] PROTOTYPE: Opt-in visible point labels via showLabels

EmiliaPalaghita:
## Summary

Adds an opt-in `showLabels` field to every interactive-graph type that has movable points (angle, polygon, quadratic, sinusoid, tangent, exponential, logarithm, absolute-value, circle, linear, ray, segment, linear-system, point). When true, each interactive point gets a visible label drawn next to it on the canvas — author-supplied via the existing `pointLabels` field, or auto-generated as A/B/C/… when not provided. Labels track each point as the learner drags it.

The field defaults to off, no changes to existing graphs. Authors only see new behavior when they opt in.

Accessibility is preserved end-to-end: a single shared helper (`getEffectivePointLabels`) drives both the visible on-canvas text and the screen-reader announcements, so a sighted learner who sees "A" hears "Point A at …" — never "Point 1". The visible label itself is wrapped in `<g aria-hidden="true">` to avoid double-announcing alongside the focusable handle's aria-label.
  
`vector` and `none` are intentionally skipped (no labelable points by schema).

## JSON format
  
`showLabels` is an optional boolean, available on every graph type that supports `pointLabels`. Three usage patterns:

# Off (default — no change from today):
```
  {
    "graph": {
      "type": "polygon",
      "numSides": 3,
      "startCoords": [[-2, 0], [2, 0], [0, 3]]
    }
  }
```

# On, no pointLabels (auto A/B/C):
``` 
 {
    "graph": {
      "type": "polygon",
      "numSides": 4,
      "startCoords": [[-3, 2], [3, 2], [3, -2], [-3, -2]],
      "showLabels": true
    }
  }
```
  → Sighted sees A/B/C/D. Screen reader hears "Point A at …", "Point B at …", etc.

# On, with pointLabels (author-controlled):
```
  {
    "graph": {
      "type": "point",
      "numPoints": 3,
      "startCoords": [[-2, 0], [2, 0], [2, 3]],
      "pointLabels": ["P", "Q", "R"],
      "showLabels": true
    }
  }
```
  → Visible: P/Q/R. Screen reader: "Point P / Q / R at …". Author labels always win at their index; gaps fall back to A/B/C.

The `pointLabels` shape per graph type is unchanged from what's already in the schema. 

## Demo

https://github.com/user-attachments/assets/6329570f-0834-4a54-a98d-70d4cd15700f

## Testing

- Unit tests (`point-labels.test.ts`) cover the shared helpers in isolation: A/B/C fallback, author label preference, empty-string passthrough, alphabet wrap at index 26, and every combination of showLabels: true | false | undefined × pointLabels: present | absent | partial.

- Aria parity tests (`point.test.ts`, 3 new specs) verify that the interactive-elements screen-reader description uses the same per-point letter the canvas displays — including the "partial pointLabels" case which proves the fill-in rule reaches the SR announcement, not just the visible label.

- Component tests (`movable-point.test.tsx`, 4 new specs) confirm the visible `<TextLabel>` is rendered only when label is provided, that it appears with the exact string passed in, that the empty string suppresses it entirely, and that the wrapping group carries `aria-hidden="true"`.

- Storybook stories (3): PointWithAutoVisibleLabels and PointWithCustomVisibleLabels exercise the two point paths; PolygonWithVisibleLabels exists specifically to prove the cross-type wiring — every other graph type follows the same pattern and would render identically in a story, so we kept the addition minimal as agreed.
  
- Regression: the full interactive-graph jest suite passes unchanged
- `pnpm tsc` pass

### DIFF
--- a/packages/perseus-core/src/data-schema.ts
+++ b/packages/perseus-core/src/data-schema.ts
@@ -1157,6 +1157,12 @@ export type PerseusGraphTypeAngle = {
     startCoords?: [Coord, Coord, Coord];
     /** Custom label for each interactive point that will help with the screen reader. */
     pointLabels?: [string, string, string];
+    /**
+     * When true, render a visible label (A, B, C, …) next to each interactive
+     * point on the graph. If `pointLabels[i]` is provided, that string is used
+     * instead of the auto-generated letter. Opt-in; defaults to off.
+     */
+    showLabels?: boolean;
 };
 
 export type PerseusGraphTypeCircle = {
@@ -1170,6 +1176,12 @@ export type PerseusGraphTypeCircle = {
     };
     /** Custom label for each interactive point that will help with the screen reader. */
     pointLabels?: string[];
+    /**
+     * When true, render a visible label (A, B, C, …) next to each interactive
+     * point on the graph. If `pointLabels[i]` is provided, that string is used
+     * instead of the auto-generated letter. Opt-in; defaults to off.
+     */
+    showLabels?: boolean;
 };
 
 export type PerseusGraphTypeLinear = {
@@ -1180,6 +1192,12 @@ export type PerseusGraphTypeLinear = {
     startCoords?: CollinearTuple;
     /** Custom label for each interactive point that will help with the screen reader. */
     pointLabels?: [string, string];
+    /**
+     * When true, render a visible label (A, B, C, …) next to each interactive
+     * point on the graph. If `pointLabels[i]` is provided, that string is used
+     * instead of the auto-generated letter. Opt-in; defaults to off.
+     */
+    showLabels?: boolean;
 };
 
 export type PerseusGraphTypeLinearSystem = {
@@ -1190,6 +1208,12 @@ export type PerseusGraphTypeLinearSystem = {
     startCoords?: CollinearTuple[];
     /** Custom label for each interactive point that will help with the screen reader. */
     pointLabels?: string[];
+    /**
+     * When true, render a visible label (A, B, C, …) next to each interactive
+     * point on the graph. If `pointLabels[i]` is provided, that string is used
+     * instead of the auto-generated letter. Opt-in; defaults to off.
+     */
+    showLabels?: boolean;
 };
 
 export type PerseusGraphTypeNone = {
@@ -1210,6 +1234,12 @@ export type PerseusGraphTypePoint = {
     coord?: Coord;
     /** Custom label for each interactive point that will help with the screen reader. */
     pointLabels?: string[];
+    /**
+     * When true, render a visible label (A, B, C, …) next to each interactive
+     * point on the graph. If `pointLabels[i]` is provided, that string is used
+     * instead of the auto-generated letter. Opt-in; defaults to off.
+     */
+    showLabels?: boolean;
 };
 
 export type PerseusGraphTypePolygon = {
@@ -1229,6 +1259,12 @@ export type PerseusGraphTypePolygon = {
     startCoords?: Coord[];
     /** Custom label for each interactive point that will help with the screen reader. */
     pointLabels?: string[];
+    /**
+     * When true, render a visible label (A, B, C, …) next to each interactive
+     * point on the graph. If `pointLabels[i]` is provided, that string is used
+     * instead of the auto-generated letter. Opt-in; defaults to off.
+     */
+    showLabels?: boolean;
 };
 
 export type PerseusGraphTypeQuadratic = {
@@ -1239,6 +1275,12 @@ export type PerseusGraphTypeQuadratic = {
     startCoords?: [Coord, Coord, Coord];
     /** Custom label for each interactive point that will help with the screen reader. */
     pointLabels?: [string, string, string];
+    /**
+     * When true, render a visible label (A, B, C, …) next to each interactive
+     * point on the graph. If `pointLabels[i]` is provided, that string is used
+     * instead of the auto-generated letter. Opt-in; defaults to off.
+     */
+    showLabels?: boolean;
 };
 
 export type PerseusGraphTypeSegment = {
@@ -1254,6 +1296,12 @@ export type PerseusGraphTypeSegment = {
     startCoords?: CollinearTuple[];
     /** Custom label for each interactive point that will help with the screen reader. */
     pointLabels?: string[];
+    /**
+     * When true, render a visible label (A, B, C, …) next to each interactive
+     * point on the graph. If `pointLabels[i]` is provided, that string is used
+     * instead of the auto-generated letter. Opt-in; defaults to off.
+     */
+    showLabels?: boolean;
 };
 
 export type PerseusGraphTypeSinusoid = {
@@ -1264,6 +1312,12 @@ export type PerseusGraphTypeSinusoid = {
     startCoords?: Coord[];
     /** Custom label for each interactive point that will help with the screen reader. */
     pointLabels?: string[];
+    /**
+     * When true, render a visible label (A, B, C, …) next to each interactive
+     * point on the graph. If `pointLabels[i]` is provided, that string is used
+     * instead of the auto-generated letter. Opt-in; defaults to off.
+     */
+    showLabels?: boolean;
 };
 
 export type PerseusGraphTypeTangent = {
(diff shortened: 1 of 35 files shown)


packages/perseus-core/src/data-schema.ts
```diff
@@ -1162,6 +1162,12 @@ export type PerseusGraphTypeAngle = {
     startCoords?: [Coord, Coord, Coord];
     /** Custom label for each interactive point that will help with the screen reader. */
     pointLabels?: [string, string, string];
+    /**
+     * When true, render a visible label (A, B, C, …) next to each interactive
+     * point on the graph. If `pointLabels[i]` is provided, that string is used
+     * instead of the auto-generated letter. Opt-in; defaults to off.
```
benchristel, line 1163: I am not so sure we want to be auto-generating labels in Perseus.
As the comment in `point-labels.ts` mentions, generating labels in JS makes it impossible to translate them. If the labels are explicitly defined in the data, on the other hand, they will be picked up by our normal translation process for Perseus content.

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-line.tsx
```diff
@@ -30,6 +30,14 @@ type Props = {
         start: boolean;
         end: boolean;
     };
+    /**
+     * Visible on-canvas labels for the two endpoints, indexed [start, end].
+     * Each entry is forwarded to the corresponding endpoint's `useControlPoint`;
```
benchristel: This is an implementation detail and shouldn't appear in a doc comment.

```suggestion
```

packages/perseus/src/widgets/interactive-graphs/point-labels.ts
benchristel, line 1: It seems like this file is scar tissue from a change that the AI made and then reverted? `getEffectivePointLabels` doesn't do anything (the `?? ""` default has no effect because `pointLabels` cannot contain null or undefined elements). 

This file, and its tests, should be deleted.

packages/perseus-core/src/data-schema.ts
```diff
@@ -1157,6 +1157,12 @@ export type PerseusGraphTypeAngle = {
...
+     */
+    showLabels?: boolean;
```
benchristel, line 1165: All these `showLabels` fields should be renamed to `showVisibleLabels` or similar. The `pointLabels` are shown to screenreaders regardless of the value of this field.


```suggestion
    showVisibleLabels?: boolean;
```

packages/perseus-linter/src/rules/interactive-graph-widget-error.ts
```diff
@@ -84,7 +84,40 @@ export default Rule.makeRule({
...
+            "showLabels is true but pointLabels is missing. Provide a label for every point.",
+        );
+        return;
+    }
+    if (labels.some((l) => l == null || l === "")) {
+        issues.push(
+            "showLabels is true but pointLabels has empty entries. Provide a label for every point.",
+        );
+    }
+}
```
benchristel, line 123: I'm not sure this linter rule is what we want. If we want to enforce a rule that either all points must have labels or none have labels, we could do that. However, the current rule is that all points must have labels _if showLabels is true_. Since `pointLabels` is used for the accessible labels regardless of whether `showLabels` is true, I don't think that validation makes much sense.

I think the linter rule should be changed to this:

- `showLabels` may only be `true` if at least one element of `pointLabels` is not null/empty.

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-line.tsx
```diff
@@ -0,0 +1,371 @@
...
+    extend?: {
+        start: boolean;
+        end: boolean;
+    };
+    /**
+     * Accepted for compatibility with PR-4-series graphs still wired
+     * against the old prop. Visible labels now render via
+     * `MovablePointLabelsLayer`, which reads coords + label text from
+     * the reducer state. Not forwarded to either endpoint; has no
+     * rendering effect.
```
benchristel, line 43: What's a "PR-4-series" graph? What "old prop" is this talking about? I suspect this is the result of AI fixating on some intermediate state of the code.

This `pointLabels` prop should be removed.

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.test.tsx
```diff
@@ -371,6 +371,12 @@ describe("MovablePoint", () => {
         expect(focusSpy).toHaveBeenCalled();
     });
 
+    // Visible labels no longer render inside `MovablePoint`. The
+    // `MovablePointLabelsLayer` component now renders them in an HTML
+    // overlay, sourced from reducer state, so it can use TeX. See
+    // `movable-point-labels-layer.test.tsx` and
+    // `movable-point-labels.test.ts` for the replacement coverage.
+
```
benchristel, line 379: This comment refers to an intermediate state of the code that no one in the future is going to know about.


```suggestion
```

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-point.tsx
```diff
@@ -26,6 +26,14 @@ type Props = {
...
      */
     sequenceNumber?: number;
+    /**
+     * Accepted for compatibility with PR-4-series graphs still wired
+     * against the old prop. Visible labels now render via
+     * `MovablePointLabelsLayer`, which reads coords + label text from
+     * the reducer state. This prop is intentionally not forwarded to
+     * `useControlPoint` and has no rendering effect.
+     */
+    label?: string;
```
benchristel, line 36:
```suggestion
```

packages/perseus/src/widgets/interactive-graphs/graphs/components/use-control-point.tsx
```diff
@@ -137,6 +137,11 @@ export function useControlPoint(params: Params): Return {
             }}
         />
     );
+    // Visible labels are rendered separately by
+    // `movable-point-labels-layer.tsx` (HTML overlay with TeX), not
+    // inside the SVG — TeX cannot render inside an SVG. The layer reads
+    // the live coordinates straight from the reducer state, so a drag
+    // pushes the label along with the point glyph automatically.
```
benchristel, line 144:
```suggestion
```

packages/perseus/src/widgets/interactive-graphs/movable-point-labels.ts
```diff
@@ -0,0 +1,194 @@
...
+    // In graph units, larger y = further north. We want the label to
+    // sit further from the graph's center than the point, so:
+    //   point in upper half → attach north (label goes above point)
+    //   point in lower half → attach south (label goes below point)
+    //   point in right half → attach east (label goes right of point)
+    //   point in left half  → attach west (label goes left of point)
+    const vertical = coord[1] >= yMid ? "n" : "s";
+    const horizontal = coord[0] >= xMid ? "e" : "w";
+    return ATTACH_BY_QUADRANT[vertical][horizontal];
+}
```
benchristel, line 194: This logic seems kind of weird. We are positioning the label based on its location relative to the bounds of the graph, not based on where other visible objects (like lines) are positioned. Depending on where the user moves a graphed function or polygon, the labels might jump from one side of the point to the other.

I think that ideally, the positioning of the labels would be based on the function or figure being plotted. E.g. for a movable line, the labels would always be below their point, and for a polygon, the labels would be positioned outside the polygon.

If that's too complicated, I'd prefer to start with very simple logic — e.g. always put the label to the right of the point. We can always do something more elaborate later.

benchristel (changes requested):
It seems like there's a lot of code in this PR left over from previous iterations of this change. I'd maybe suggest starting from a clean slate on a new branch, to reduce confusion.

My thoughts on the overall strategy are:

- Adding a widget option to show visible labels seems like a fine approach.
- I would question whether we want to interpret the labels as TeX. There is no handling of TeX for the screenreader labels, so if labels are authored with TeX syntax it will mess up the screenreader experience. If you just want the labels to display in a serif math font, [use Symbola like we do for the numeric-input widget](https://github.com/Khan/perseus/blob/1062044db9c860425c4f30b2da6971fff6d64123/packages/perseus/src/widgets/numeric-input/numeric-input_legacy-styles.js#L8).
- If we don't interpret the labels as TeX, we don't need a separate layer for rendering the labels.
- The visible labels should be exactly the same as the accessible labels. Only labels that appear in the JSON should be shown. I don't want Perseus to be generating labels.
- The linter does not need to ensure that all points have labels. It might be worthwhile to ensure that `showPoints` is true only if there is _at least one_ label.

PR #3671, opened 2026-05-27
[ColorSync] Add more image regression stories

nishasy:
## Summary:
Adding some more image widget regression stories identified by QE.

Regression tests identified from QE:

(If not otherwise covered)
- Float left and float right image in same article
- Tall image with caption
- Landscape/wide image with caption
- SVG image
- PNG image
- Graphie image
- GIF image in mobile
- Non-animated GIF

Interactions:
- Long description (Clicked) tall GIF image in modal
- Zoom clicked right to left
- Long description Right to left
- Long description Clicked right to left

Added to the list later:
- full-width alignment

Issue: https://khanacademy.atlassian.net/browse/LEMS-4135

## Test plan:
Confirm in the UI tests section below


packages/perseus/src/widgets/__testutils__/story-decorators.tsx
```diff
@@ -20,10 +20,33 @@ export const mobileArticleDecorator: Decorator = (Story) => (
...
+        };
+    }, []);
+    return <div dir="rtl">{children}</div>;
+};
+
 export const rtlDecorator: Decorator = (Story) => (
-    <div style={{direction: "rtl"}}>
+    <RtlWrapper>
         <Story />
-    </div>
```
nishasy, line 26: As specified in the comment, I had to change this to get the explore image modal to actually show RTL styles. It wasn't working because WB modal uses react portals, which means it renders outside the main page body and doesn't inherit those styles.

packages/perseus/src/widgets/image/__docs__/image-initial-state-regression.stories.tsx
```diff
@@ -46,8 +50,15 @@ const frescsoLongDescription =
...
+const bioContent1 =
+    "Prophase (sometimes divided into prophase and prometaphase):";
+const bioContent2 =
+    "Chromosomes: In prophase, the chromosomes condense, forming the characteristic “X” shape that is often shown in diagrams. Each “X” is a duplicated chromosome. The two sides of the “X” are called sister chromatids, and they are attached at a point called the centromere. Even though the chromosome has been copied at this point of the cell cycle, as long as the two copies (sister chromatids) are attached, they are considered a single chromosome.";
+const bioContent3 =
+    "The nucleolus (a structure inside the nucleus where ribosomes are made) disappears during prophase. The mitotic spindle begins to form during prophase, starting at regions called centrosomes. These regions contain the material needed for building the spindle, and also function to regulate the spindle throughout mitosis.";
+
 const meta: Meta<PerseusImageWidgetOptions> = {
-    title: "Widgets/Image/Visual Regression Tests",
+    title: "Widgets/Image/Visual Regression Tests/Initial State",
```
nishasy, line 64: Note: All the stories are showing up as new because I put them in a folder

packages/perseus/src/widgets/image/__docs__/image-interactions-regression.stories.tsx
```diff
@@ -176,3 +180,58 @@ export const ZoomClickedLargePortraitImage: Story = {
...
+            ...ApiOptions.defaults,
+            flags: getFeatureFlags({
+                "image-widget-upgrade-gif-controls": true,
+            }),
+        },
+    },
+    args: {
+        backgroundImage: {
+            url: "https://cdn.kastatic.org/ka-content-images/1e6f6fd4de01058c3d548b7a942bd9e76d565fa3.gif",
+        },
```
Myranae: I don't think you have to hard code here. Can just use the `gifImage` value after importing it

```suggestion
        backgroundImage: gifImage,
```

packages/perseus/src/widgets/__testutils__/story-decorators.tsx
```diff
@@ -20,10 +20,33 @@ export const mobileArticleDecorator: Decorator = (Story) => (
     </div>
 );
 
+// Set `dir="rtl"` on both the wrapper and `<html>` so portaled content
+// (e.g. Wonder Blocks modal) inherits RTL.
+// CSS `direction` inherits through the DOM tree, not the React tree, so
+// portals not properly inherit the RTL direction without the HTML mutation.
```
Myranae:
```suggestion
// portals don't properly inherit the RTL direction without the HTML mutation.
```

packages/perseus/src/widgets/__testutils__/story-decorators.tsx
Myranae: Thanks for getting this working! I think in the past I just didn't have a RTL story if the setting didn't work. I wonder if other widgets could have successful RTL stories with a change like this. 🤔 Not related to this PR, but do you think that would be valuable for other widgets? I'd say we should add a ticket, but maybe that's something Sarah is checking for.

packages/perseus/src/widgets/image/__docs__/image-interactions-regression.stories.tsx
```diff
@@ -176,3 +180,58 @@ export const ZoomClickedLargePortraitImage: Story = {
...
     },
 };
+
+export const LongDescriptionClickedStateWithTallGif: Story = {
+    decorators: [imageRendererDecorator],
+    parameters: {
+        apiOptions: {
+            ...ApiOptions.defaults,
+            flags: getFeatureFlags({
+                "image-widget-upgrade-gif-controls": true,
```
Myranae, line 191: We just did some work with Radio to make it so that the story doesn't break when the flag is removed. Might need to think about that here. The [PR](https://github.com/Khan/frontend/pull/11138/changes) was in frontend though, so maybe it's not the same. Does this rely on the Growthbook flag?

packages/perseus/src/widgets/image/__docs__/image-interactions-regression.stories.tsx
Myranae:
```suggestion
        backgroundImage: gifImage,
```

I think I mentioned this before, but could do it here too I think

Myranae (approved):
Left some comments, but looks good :) Thanks for adding these!

packages/perseus/src/widgets/image/__docs__/image-interactions-regression.stories.tsx
nishasy: It is related to Growthbook, but the `getFeatureflags` util lets us manually set the flag for tests. I'm going to remove the GIF flag next week, which will also include removing this `parameters` block entirely as part of that task.

packages/perseus/src/widgets/image/__docs__/image-interactions-regression.stories.tsx
nishasy: This story is `LongDescriptionClickedStateWithTallGif` and `gifImage` is not a vertical gif. I'll add a new one in the util.

nishasy: sorry, I just realized that means the alt is wrong too

nishasy: nvm I merged main, and the tall gif has been updated

packages/perseus/src/widgets/__testutils__/story-decorators.tsx
nishasy: I think it would be valuable for any widgets that have side-by-side components, just to make sure they flip correctly.

packages/perseus/src/widgets/image/__docs__/image-interactions-regression.stories.tsx
nishasy: Is this a repeat of the previous comment? It looks like it's referring to the same lines?

packages/perseus/src/widgets/__testutils__/story-decorators.tsx
mark-fitzgerald: Not trying to expand scope, but do we want a Storybook toggle for switching LTR <=> RTL?

nishasy: @mark-fitzgerald That would be awesome! It could be worth a larger conversation, as I'm not sure how that would work for regression alerts.

packages/perseus/src/widgets/image/__docs__/image-initial-state-regression.stories.tsx
```diff
@@ -371,6 +383,85 @@ export const GraphieImage: Story = {
...
+        },
+    },
+    args: {
+        backgroundImage: gifImage,
+        alt: gifImageAlt,
+        caption: gifImageAlt,
+    },
+};
+
+export const NonAnimatedGifImage: Story = {
```
mark-fitzgerald: Nit: I think that GIF images should be assumed to be non-animated, unless the word "animated" is connected to "GIF". Or, we can preface all GIFs as either "animated" or "non-animated".

packages/perseus/src/widgets/image/__docs__/image-interactions-regression.stories.tsx
Myranae: So weird! Apologies, it looks like it is. I thought for sure I was looking at a different set. I might have been in the PR in a different tab or something 🤷‍♀️ In any case, looks like the instances of a url are gone from what I can tell. Thanks! :)

packages/perseus/src/widgets/image/__docs__/image-initial-state-regression.stories.tsx
Myranae: I feel like general understanding is that GIFs are always animated (unless otherwise described), but I could be wrong. Otherwise you would just have it be or call it an image 🤷‍♀️  But it sounds like this is being discussed elsewhere so I'll just slowly back away... 😂

packages/perseus/src/widgets/image/__docs__/image-initial-state-regression.stories.tsx
```diff
@@ -428,3 +519,106 @@ export const MarkdownTableWithMarkdownImages: Story = {
...
+                                    alt: "Fresco image - block",
+                                }),
+                            }),
+                            "image 4": generateImageWidget({
+                                alignment: "full-width",
+                                options: generateImageOptions({
+                                    backgroundImage: frescoImage,
+                                    alt: "Fresco image - full-width",
+                                }),
+                            }),
```
mark-fitzgerald, line 565: Do we need these last 2 image tests ("block" and "full-width") included in the Renderer regression tests, as well? Or, is having them here sufficient? I'm thinking the latter, but I want to check with both of you.

nishasy: I think it's sufficient to have them here.

packages/perseus/src/widgets/image/__docs__/image-initial-state-regression.stories.tsx
nishasy: I think in today's day and age, among the general public, GIFs are assumed to be animated unless stated otherwise. So I'm inclined to leave this one as is. But I also agree it make sense to rename the other ones to specify that they're animated, just for clarity.

mark-fitzgerald (approved):
Just a few minor comments.

PR #3672, opened 2026-05-27
[Interactive Graph] Wire pointLabels through line and ray graphs for custom points label

ivyolamit:
## Summary:
- Wires `pointLabels` through two of the three line-family interactive graphs — `linear` and `ray` — end-to-end on the render side and unlocks the editor label fields for those two types.
- Each interactive endpoint now announces its author-supplied label (e.g. *"Point A at -5 comma 5"*); when no custom label is set the existing default behavior is preserved per graph type:
  - **linear** falls back to the generic *"Point 1 at …"* / *"Point 2 at …"* defaults (same as before).
  - **ray** falls back to the semantic *"Endpoint at …"* / *"Through point at …"* defaults (same as before).
- **Vector is excluded** from this PR and from the LEMS-3995 rollout entirely. The `pointLabels` field on `PerseusGraphTypeVector` (originally shipped by PR 1) is **removed from the schema and parser in this PR** so the type reflects the runtime behaviour. The original plan listed `vector` alongside `linear` and `ray` because all three share the `[string, string]` schema shape and the `start-coords-line.tsx` editor component. The vector has tail and tip which is already mathematically correct for screen readers.

Issue: LEMS-3995

## Risk

Low. The new code paths are:
- Two render files (linear, ray) each gaining a few lines that read `pointLabels` from state and pass the helper output as `ariaLabel`.
- `start-coords-line.tsx` extended with `pointLabels` / `onChangePointLabels` props, the flat `pointLabel` / `onPointLabelChange` forwarded to each `<CoordInput>`, and a small `updatePointLabel` helper that pads the 2-tuple shape. Vector continues to call `<StartCoordsLine>` without those props, so its editor UX is unchanged.
- One gate widening per case branch in `start-coords-settings.tsx` (only linear / ray; vector deliberately not widened).
- The schema cleanup (removing `pointLabels` from `PerseusGraphTypeVector` and the parser) is two line deletions touching `data-schema.ts` and `interactive-graph-widget.ts`. Because the field was only ever optional and was never written to by the editor or read at runtime, no production content can break: any content that *does* carry `pointLabels` on a vector was hand-authored (the editor never produced it) and the field was already a no-op.
- One case relocation in `reshape-point-labels.ts` (`case "vector":` moves from `toPair` to the no-op group) — required for TypeScript to keep type-checking after the schema field disappears.

Reviewer focus:
- For each of linear / ray, (a) the focusable handle uses the custom label when set, (b) the right default is preserved when it's not (generic vs semantic).
- For vector, confirm that nothing changed at runtime: the SR announcements are still `"Tip point at X comma Y"` regardless of any `pointLabels` value (now impossible to set via the schema), and the editor's `<StartCoordsLine>` for vector still renders without the `Label [_]` field.
- **Schema cleanup verification.** Confirm no consumer of `PerseusGraphTypeVector` reads `pointLabels`. The `pnpm tsc` pass is the structural lock — any consumer that *was* reading it would fail compilation now. The full `perseus-core` jest suite is the runtime lock — any parser regression test asserting the field was accepted on vector would fail; none did, which means no such test existed (the field was never authored in fixtures).
- The rationale for excluding vector — see [Vector intentionally skipped](#vector-intentionally-skipped). Anyone uncomfortable with the schema-level removal can be directed at the decision log entry in the plan doc.

## Test plan

- [x] `pnpm tsc` — passes (the schema cleanup is structurally consistent across `data-schema.ts`, `interactive-graph-widget.ts`, and `reshape-point-labels.ts`)
- [x] `pnpm lint <changed files>` — passes
- [x] `pnpm prettier --check <changed files>` — passes
- [x] `pnpm jest packages/perseus-core packages/perseus-editor/src/widgets/interactive-graph-editor/utils/reshape-point-labels.test.ts packages/perseus/src/widgets/interactive-graphs/graphs/{linear,ray,vector}.test.tsx packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-settings.test.tsx` — 1762/1763 pass, 1 pre-existing skip. Includes the full `perseus-core` suite (parser regression tests, generator tests) as a non-regression check for the schema cleanup; includes `reshape-point-labels.test.ts` to confirm the relocated vector case behaves the same as `none`.
- [ ] Reviewer manual: open the `LinearWithCustomLabels` / `RayWithCustomLabels` stories. Enable the Storybook a11y addon (or VoiceOver / JAWS) and confirm each interactive endpoint announces *"Point A at …"* / *"Point B at …"* matching its prompt. Confirm the existing `Linear` / `Ray` stories still announce the legacy defaults.
- [ ] Reviewer manual: open the `InteractiveGraphLinear` / `InteractiveGraphRay` editor stories, expand "Start coordinates", and confirm each row reads as `Point N:`.

## PR Series and Follow-ups (PRs 5–7)

Independent and parallelizable. None of them touch the foundation files. Each wires its graph type's render side and adds the per-type `start-coords-<type>.tsx` plumbing using the same flat `CoordInput` `pointLabel` / `onPointLabelChange` props PR 2 take-2 introduced.

- 👉🏼 **PR 4** — [Lines: linear, ray](https://github.com/Khan/perseus/pull/3672)

### DIFF
--- a/packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
+++ b/packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
@@ -168,7 +168,6 @@ const parsePerseusGraphTypeVector = object({
     coords: optional(nullable(pair(pairOfNumbers, pairOfNumbers))),
     startCoords: optional(pair(pairOfNumbers, pairOfNumbers)),
     match: optional(enumeration("exact", "congruent")),
-    pointLabels: optional(pair(string, string)),
 });
 
 export const parsePerseusGraphType = discriminatedUnionOn("type")
--- a/packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-line.tsx
+++ b/packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-line.tsx
@@ -4,21 +4,21 @@ import CoordInput from "./coord-input";
 
 import type {CollinearTuple} from "@khanacademy/perseus-core";
 
-type Props = {
+interface StartCoordsLineProps {
     startCoords: CollinearTuple;
     onChange: (startCoords: CollinearTuple) => void;
-    pointLabels?: ReadonlyArray<string>;
-    onChangePointLabels?: (pointLabels: ReadonlyArray<string>) => void;
-};
+    pointLabels: ReadonlyArray<string>;
+    onChangePointLabels: (pointLabels: ReadonlyArray<string>) => void;
+}
 
-const StartCoordsLine = (props: Props) => {
+const StartCoordsLine = (props: StartCoordsLineProps) => {
     const {startCoords, onChange, pointLabels, onChangePointLabels} = props;
     const updatePointLabel = (index: number, newLabel: string) => {
         const next: [string, string] = [
-            index === 0 ? newLabel : pointLabels?.[0] ?? "",
-            index === 1 ? newLabel : pointLabels?.[1] ?? "",
+            index === 0 ? newLabel : pointLabels[0] ?? "",
+            index === 1 ? newLabel : pointLabels[1] ?? "",
         ];
-        onChangePointLabels?.(next);
+        onChangePointLabels(next);
     };
 
     return (
@@ -27,21 +27,15 @@ const StartCoordsLine = (props: Props) => {
                 label="Point 1"
                 coord={startCoords[0]}
                 onChange={(value) => onChange([value, startCoords[1]])}
-                pointLabel={pointLabels?.[0]}
-                onPointLabelChange={
-                    onChangePointLabels &&
-                    ((newLabel) => updatePointLabel(0, newLabel))
-                }
+                pointLabel={pointLabels[0]}
+                onPointLabelChange={(newLabel) => updatePointLabel(0, newLabel)}
             />
             <CoordInput
                 label="Point 2"
                 coord={startCoords[1]}
                 onChange={(value) => onChange([startCoords[0], value])}
-                pointLabel={pointLabels?.[1]}
-                onPointLabelChange={
-                    onChangePointLabels &&
-                    ((newLabel) => updatePointLabel(1, newLabel))
-                }
+                pointLabel={pointLabels[1]}
+                onPointLabelChange={(newLabel) => updatePointLabel(1, newLabel)}
             />
         </>
     );
--- a/packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-line.tsx
+++ b/packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-line.tsx
@@ -7,10 +7,19 @@ import type {CollinearTuple} from "@khanacademy/perseus-core";
 type Props = {
     startCoords: CollinearTuple;
     onChange: (startCoords: CollinearTuple) => void;
+    pointLabels?: ReadonlyArray<string>;
+    onChangePointLabels?: (pointLabels: ReadonlyArray<string>) => void;
 };
 
 const StartCoordsLine = (props: Props) => {
-    const {startCoords, onChange} = props;
+    const {startCoords, onChange, pointLabels, onChangePointLabels} = props;
+    const updatePointLabel = (index: number, newLabel: string) => {
+        const next: [string, string] = [
+            index === 0 ? newLabel : pointLabels?.[0] ?? "",
+            index === 1 ? newLabel : pointLabels?.[1] ?? "",
+        ];
+        onChangePointLabels?.(next);
+    };
 
     return (
         <>
@@ -18,11 +27,21 @@ const StartCoordsLine = (props: Props) => {
                 label="Point 1"
                 coord={startCoords[0]}
                 onChange={(value) => onChange([value, startCoords[1]])}
+                pointLabel={pointLabels?.[0]}
+                onPointLabelChange={
+                    onChangePointLabels &&
+                    ((newLabel) => updatePointLabel(0, newLabel))
+                }
             />
             <CoordInput
                 label="Point 2"
                 coord={startCoords[1]}
                 onChange={(value) => onChange([startCoords[0], value])}
+                pointLabel={pointLabels?.[1]}
+                onPointLabelChange={
+                    onChangePointLabels &&
+                    ((newLabel) => updatePointLabel(1, newLabel))
+                }
             />
         </>
     );
--- a/packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-settings.test.tsx
+++ b/packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-settings.test.tsx
@@ -18,6 +18,7 @@ const defaultProps = {
         [-10, 10],
     ] satisfies [Range, Range],
     step: [1, 1] satisfies [number, number],
+    onChangePointLabels: () => {},
 };
 describe("StartCoordSettings", () => {
     let userEvent: UserEvent;
@@ -176,24 +177,7 @@ describe("StartCoordSettings", () => {
             ).toBeInTheDocument();
         });
 
-        it(`does NOT render point name fields when onChangePointLabels is omitted for ${type}`, () => {
-            // Arrange, Act
-            render(
-                <StartCoordsSettings
-                    {...defaultProps}
-                    type={type}
-                    onChange={() => {}}
-                />,
-                {wrapper: RenderStateRoot},
-            );
-
-            // Assert
-            expect(
-                screen.queryByRole("textbox", {name: "Point 1 name"}),
-            ).not.toBeInTheDocument();
-        });
-
-        it(`calls onChangePointLabels with the schema's 2-tuple shape when a name is typed for ${type}`, async () => {
+        it(`${type}: calls onChangePointLabels with the schema's 2-tuple shape when a name is typed`, async () => {
             // Arrange
             const onChangePointLabelsMock = jest.fn();
             render(
@@ -215,6 +199,31 @@ describe("StartCoordSettings", () => {
             // Assert — the empty slot must be preserved so the schema
             // shape stays a 2-tuple even when only one point is named.
             expect(onChangePointLabelsMock).toHaveBeenLastCalledWith(["T", ""]);
+        });
+    });
+
+    describe("vector graph", () => {
+        it("shows the start coordinates UI without point name fields", () => {
+            // Arrange, Act
+            render(
+                <StartCoordsSettings
+                    {...defaultProps}
+                    type="vector"
+                    onChange={() => {}}
+                />,
+                {wrapper: RenderStateRoot},
+            );
+
+            // Assert — coord inputs render
+            expect(screen.getByText("Point 1:")).toBeInTheDocument();
+            expect(screen.getByText("Point 2:")).toBeInTheDocument();
+            // Assert — vector is intentionally excluded from custom labels
+            expect(
+                screen.queryByRole("textbox", {name: "Point 1 name"}),
+            ).not.toBeInTheDocument();
+            expect(
+                screen.queryByRole("textbox", {name: "Point 2 name"}),
+            ).not.toBeInTheDocument();
         });
     });
 
@@ -828,24 +837,6 @@ describe("StartCoordSettings", () => {
             ).toBeInTheDocument();
         });
 
-        it("does NOT render point name fields when onChangePointLabels is omitted", () => {
-            // Arrange, Act
-            render(
-                <StartCoordsSettings
-                    {...defaultProps}
-                    type="point"
-                    numPoints={2}
-                    onChange={() => {}}
-                />,
-                {wrapper: RenderStateRoot},
-            );
-
-            // Assert
-            expect(
-                screen.queryByRole("textbox", {name: "Point 1 name"}),
-            ).not.toBeInTheDocument();
-        });
-
         it("calls onChangePointLabels when a point name is typed", async () => {
             // Arrange
             const onChangePointLabelsMock = jest.fn();
@@ -945,23 +936,6 @@ describe("StartCoordSettings", () => {
             expect(
                 screen.getByRole("textbox", {name: "Point 3 name"}),
             ).toBeInTheDocument();
-        });
-
-        it("does NOT render point name fields when onChangePointLabels is omitted", () => {
-            // Arrange, Act
-            render(
-                <StartCoordsSettings
-                    {...defaultProps}
-                    type="polygon"
-                    onChange={() => {}}
-                />,
-                {wrapper: RenderStateRoot},
-            );
-
-            // Assert
-            expect(
-                screen.queryByRole("textbox", {name: "Point 1 name"}),
-            ).not.toBeInTheDocument();
         });
 
         it("calls onChangePointLabels when a point name is typed", async () => {
--- a/packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-settings.test.tsx
+++ b/packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-settings.test.tsx
@@ -154,6 +154,68 @@ describe("StartCoordSettings", () => {
                 expect(onChangeMock).toHaveBeenLastCalledWith(expectedCoords);
             },
         );
+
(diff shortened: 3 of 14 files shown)


packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
ivyolamit, line 1: We're cleaning up the data type related to vector pointLabels since we are excluding it that was introduced [here](https://github.com/Khan/perseus/pull/3605/changes#diff-d343348f71c73496134f4e1dc910ece414b89bc84da52748cfe59d98fe42a647R171). The vector screen reader is mathematically correct already having the tail and tip terms.

packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-line.tsx
```diff
@@ -7,22 +7,41 @@ import type {CollinearTuple} from "@khanacademy/perseus-core";
...
         <>
             <CoordInput
                 label="Point 1"
                 coord={startCoords[0]}
                 onChange={(value) => onChange([value, startCoords[1]])}
+                pointLabel={pointLabels?.[0]}
+                onPointLabelChange={
+                    onChangePointLabels &&
+                    ((newLabel) => updatePointLabel(0, newLabel))
+                }
```
handeyeco: I think it would be cleaner to simplify the JSX portion and just have `updatePointLabel` exit early if there's no `onChangePointLabels`.
```suggestion
                onPointLabelChange={
                    (newLabel) => updatePointLabel(0, newLabel)
                }
```

packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-line.tsx
```diff
@@ -7,22 +7,41 @@ import type {CollinearTuple} from "@khanacademy/perseus-core";
...
             />
             <CoordInput
                 label="Point 2"
                 coord={startCoords[1]}
                 onChange={(value) => onChange([startCoords[0], value])}
+                pointLabel={pointLabels?.[1]}
```
handeyeco: (Random thought, not a blocker) These two uses of `CoordInput` seem so similar, I wonder if they could just be called in a loop. The benefit would be that they're less likely to get out of sync. For example you had to add `onPointLabelChange` twice when making this change; if we consolidated them into one loop, you'd only have to add `onPointLabelChange` in one spot.

packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-settings.test.tsx
```diff
@@ -154,6 +154,68 @@ describe("StartCoordSettings", () => {
...
+        it(`renders point name fields when onChangePointLabels is provided for ${type}`, () => {
+            // Arrange, Act
+            render(
+                <StartCoordsSettings
+                    {...defaultProps}
+                    type={type}
+                    onChange={() => {}}
+                    onChangePointLabels={() => {}}
+                />,
+                {wrapper: RenderStateRoot},
```
handeyeco, line 168: Just a question: do you actually need `RenderStateRoot` here? If not, should we remove it? If so, should we make an abstraction around this?

packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-line.tsx
```diff
@@ -7,22 +7,41 @@ import type {CollinearTuple} from "@khanacademy/perseus-core";
 type Props = {
     startCoords: CollinearTuple;
     onChange: (startCoords: CollinearTuple) => void;
+    pointLabels?: ReadonlyArray<string>;
+    onChangePointLabels?: (pointLabels: ReadonlyArray<string>) => void;
```
handeyeco: `StartCoordsSettings` is only ever called once and it's given `onChangePointLabels`. Why make `onChangePointLabels` it optional? This PR seems like it could be simpler if it wasn't optional.

packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-settings.test.tsx
```diff
@@ -154,6 +154,68 @@ describe("StartCoordSettings", () => {
...
+            );
+
+            // Assert
+            expect(
+                screen.queryByRole("textbox", {name: "Point 1 name"}),
+            ).not.toBeInTheDocument();
+        });
+
+        it(`calls onChangePointLabels with the schema's 2-tuple shape when a name is typed for ${type}`, async () => {
```
handeyeco: nit, feel free to ignore:

```suggestion
        it(`${type}: calls onChangePointLabels with the schema's 2-tuple shape when a name is typed`, async () => {
```

I just think that would make it easier to differentiate in Jest reports and maybe make the test easier to find because it's a little more obvious what part of the name is interpolated.

packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-settings.test.tsx
```diff
@@ -154,6 +154,68 @@ describe("StartCoordSettings", () => {
...
+            // Assert
+            expect(
+                screen.getByRole("textbox", {name: "Point 1 name"}),
+            ).toBeInTheDocument();
+            expect(
+                screen.getByRole("textbox", {name: "Point 2 name"}),
+            ).toBeInTheDocument();
+        });
+
+        it(`does NOT render point name fields when onChangePointLabels is omitted for ${type}`, () => {
```
handeyeco: To my other point, these tests wouldn't be needed if `onChangePointLabels` was required instead of optional.

packages/perseus-editor/src/widgets/interactive-graph-editor/utils/reshape-point-labels.ts
```diff
@@ -64,6 +63,7 @@ export function reshapePointLabelsForGraphType(
             }
             break;
         }
+        case "vector":
```
handeyeco, line 66: This is safe to do? The old logic wasn't needed for `vector`?

packages/perseus/src/widgets/interactive-graphs/__docs__/interactive-graph.stories.tsx
```diff
@@ -150,6 +165,20 @@ export const Ray: Story = {
     },
 };
 
+/**
+ * A ray whose endpoint and through point are named "A" and "B" via
+ * `pointLabels`. The default semantic "Endpoint at …" / "Through point at …"
+ * labels are overridden so the SR announcement matches the prompt's
+ * naming convention.
+ */
+export const RayWithCustomLabels: Story = {
```
handeyeco, line 174: Is it useful to have a story for each feature for each graph type? I'm just concerned that it could make finding things in Storybook (which is already a challenge) more difficult. Feel free to disregard.

packages/perseus/src/widgets/interactive-graphs/graphs/linear.test.tsx
```diff
@@ -241,6 +241,80 @@ describe("Linear graph screen reader", () => {
...
+            testDependencies,
+        );
+    });
+
+    it("uses custom pointLabels in each endpoint's accessible name", () => {
+        // Arrange, Act
+        render(
+            <MafsGraph
+                {...baseMafsGraphProps}
+                state={{...baseLinearState, pointLabels: ["A", "B"]}}
```
handeyeco: I'd much rather see us using the IG generators and `renderQuestion` personally. Setting state directly like this isn't representative of how we'd be displaying the IG to learners.

packages/perseus/src/widgets/interactive-graphs/graphs/linear.test.tsx
```diff
@@ -241,6 +241,80 @@ describe("Linear graph screen reader", () => {
...
+            />,
+        );
+        const [point1, , point2] = screen.getAllByRole("button");
+
+        // Assert
+        expect(point1).toHaveAttribute("aria-label", "Point A at -5 comma 5.");
+        expect(point2).toHaveAttribute("aria-label", "Point B at 5 comma 5.");
```
handeyeco, line 269:
```suggestion
        const pointA = screen.getAllByRole(
            "button",
            name: "Point A at -5 comma 5."
        );
        const pointB = screen.getAllByRole(
            "button",
            name: "Point B at 5 comma 5."
        );

        // Assert
        expect(pointA).toBeInTheDocument();
        expect(pointB).toBeInTheDocument();
```
I think this would be more future proof; for example what would happen to the test if we were to add a third button?

packages/perseus/src/widgets/interactive-graphs/graphs/linear.test.tsx
```diff
@@ -241,6 +241,80 @@ describe("Linear graph screen reader", () => {
...
+
+    it("falls back to the numeric default for truthy non-string entries (defensive against malformed hand-authored JSON bypassing the parser)", () => {
+        // Arrange, Act
+        render(
+            <MafsGraph
+                {...baseMafsGraphProps}
+                state={{
+                    ...baseLinearState,
+                    // eslint-disable-next-line no-restricted-syntax -- cast simulates malformed JSON the parser would reject
+                    pointLabels: [42, "B"] as unknown as string[],
```
handeyeco: When would this happen?

packages/perseus/src/widgets/interactive-graphs/graphs/ray.test.tsx
```diff
@@ -30,7 +30,7 @@ const baseRayState: InteractiveGraphState = {
 
 const overallGraphLabel = "A ray on a coordinate plane.";
 
-describe("Linear graph screen reader", () => {
+describe("Ray graph screen reader", () => {
```
handeyeco, line 33: lol, makes sense

packages/perseus/src/widgets/interactive-graphs/graphs/linear.tsx
```diff
@@ -60,7 +61,23 @@ const LinearGraph = (props: LinearGraphProps, key: number) => {
...
-                ariaLabels={{grabHandleAriaLabel: srLinearGrabHandle}}
+                ariaLabels={{
+                    grabHandleAriaLabel: srLinearGrabHandle,
+                    point1AriaLabel: buildPointAriaLabel(
+                        pointLabels,
+                        0,
+                        line[0],
+                        strings,
+                        locale,
+                    ),
```
handeyeco, line 72: `buildPointAriaLabel`s API still feels a little awkward to me. I think I brought this up in another PR and I'm guessing you determined it was best as-is. No action needed, just wanted to point it out again.

packages/perseus/src/widgets/interactive-graphs/graphs/ray.tsx
```diff
@@ -61,8 +62,22 @@ const RayGraph = (props: Props) => {
...
+                            locale,
+                        ) ?? srRayEndpoint,
+                    point2AriaLabel:
+                        buildPointAriaLabel(
+                            pointLabels,
+                            1,
+                            line[1],
+                            strings,
+                            locale,
+                        ) ?? srRayTerminalPoint,
```
handeyeco: I wonder if it would be worth moving these out of JSX:

``` TS
const point1AriaLabel = buildPointAriaLabel(
    pointLabels,
    0,
    line[0],
    strings,
    locale,
) ?? srRayEndpoint
```

Then:


```suggestion
                    point1AriaLabel,
                    point2AriaLabel,
```

packages/perseus/src/widgets/interactive-graphs/interactive-graph.testdata.ts
```diff
@@ -1311,6 +1311,63 @@ export const pointWithDefaultLabelQuestion: PerseusRenderer =
...
+        content:
+            "**Drag points $A$ and $B$ so the line passes through both labeled points.**\n\n[[☃ interactive-graph 1]]",
+        markings: "graph",
+        gridStep: [1, 1],
+        snapStep: [1, 1],
+        step: [1, 1],
+        range: [
+            [-10, 10],
+            [-10, 10],
+        ],
```
handeyeco, line 1328: Doesn't `generateInteractiveGraphQuestion` default a lot of this stuff for you?

handeyeco (commented):
I'm concerned about [the change to vector in this PR](https://github.com/Khan/perseus/pull/3672#discussion_r3318813801), but if you tell me that's fine I can approve.

packages/perseus-editor/src/widgets/interactive-graph-editor/utils/reshape-point-labels.ts
catandthemachines: Same, should this be moved?

catandthemachines (approved):
Looks good to me, did have one concern on the vector case in utils.ts but non blocking as long as it's correct.

packages/perseus-editor/src/widgets/interactive-graph-editor/utils/reshape-point-labels.ts
ivyolamit: @handeyeco @catandthemachines yes this is fine, this was added in the [previous PR](https://github.com/Khan/perseus/pull/3643/changes#diff-d0c1d6716f3467b69b9515bc7f0ee3b6aaa2ea032af2fae36b2579983fcc8203R18) as part of the custom point label and is used being used [here](https://github.com/Khan/perseus/pull/3643/changes#diff-476b671baea6e731b5088781374ec6398400cb03b83afb2dc05a948861de15f8R200). So the decision now to not include the vector will not have any negative impact in prod.

packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-line.tsx
ivyolamit: This is because of vector [here](https://github.com/Khan/perseus/blob/main/packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-settings.tsx#L158), since we are not including vector with custom point label, it will not pass the `pointLabels` and `onChangePointLabels`. The other option I can think of is to not use `StartCoordsLine` for vector either use inline CoordInput (which doesn't follow the existing pattern in start coordinate file) or create a new StartCoordsVector which will follow the existing pattern and this file will no longer need to handle these two optional props.

packages/perseus/src/widgets/interactive-graphs/__docs__/interactive-graph.stories.tsx
ivyolamit: The only reason I added this is to make it easier when viewing what it looks like without having to think and manually do it in the editor. But yeah I also feel you're point.

packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-settings.test.tsx
ivyolamit: Yes, RenderStateRoot is needed for tests that render the label TextField.

packages/perseus/src/widgets/interactive-graphs/graphs/linear.test.tsx
ivyolamit: This is when content editor use the json object to create the content/widget/

packages/perseus/src/widgets/interactive-graphs/graphs/linear.tsx
ivyolamit: I just saw Cat's PR, i'll revisit this again in this ticket: see note in the ticket.
https://khanacademy.atlassian.net/browse/LEMS-4206

ivyolamit:
Note: The data-schema change [here](https://github.com/Khan/perseus/pull/3672/changes#diff-17ad85d7f821dc080c7b004a2c0213fa3fad6c1e919a525d47d262202417b871L1345) for vector is Ok since this is new type that was added recently and not used in prod right now.

ivyolamit:
Rebased from latest master to pickup the recent change introduced in https://github.com/Khan/perseus/pull/3682

ivyolamit:
> I'm concerned about [the change to vector in this PR](https://github.com/Khan/perseus/pull/3672#discussion_r3318813801), but if you tell me that's fine I can approve.

@handeyeco thanks for the review, regarding your concern about vector. This change is okay. See my response in the comment above here: https://github.com/Khan/perseus/pull/3672#discussion_r3321255534

handeyeco (approved):

PR #3673, opened 2026-05-28
[Interactive Graph] Wire pointLabels through linear-system and segment graphs for custom points label

ivyolamit:
## Summary:
- Wires `pointLabels` through the two multi-line interactive graphs — `linear-system` and `segment` — end-to-end on the render side and unlocks the editor label fields for both types.
- Each interactive endpoint now announces its author-supplied label (e.g. *"Point A at -5 comma 5"*). When no custom label is set the existing per-type defaults are preserved unchanged:
  - **linear-system** falls back to *"Point N on line M at …"* (the `srLinearSystemPoint` template).
  - **segment** falls back to *"Endpoint N at …"* / *"Endpoint N on segment M at …"* (the `srSingleSegmentGraphEndpointAriaLabel` / `srMultipleSegmentGraphEndpointAriaLabel` templates, depending on segment count).
- `pointLabels` for both types is a single **flat array** across all lines/segments (`string[]` schema): `[line0.p1, line0.p2, line1.p1, line1.p2, …]`. Renderer maps `(lineIndex i, endpointIndex j) → flat index i * 2 + j`. The editor's `updatePointLabel(flatIndex, newLabel)` helper rebuilds the full `startCoords.length * 2`-length array on every keystroke so the unedited slots stay as their existing value (or `""`).

Issue: LEMS-3995

## Risk

Low. No foundation changes. The new code paths are:
- Two render files (linear-system, segment) each gaining a per-component `buildLabel` closure and a `?? defaultLabel` chain on each `MovableLine.ariaLabels` slot.
- `start-coords-multiline.tsx` refactored to consume `CoordInput` (the shared two-row tile component) with `pointLabels` / `onChangePointLabels` plumbed through, plus a local `updatePointLabel` helper that pads the flat array.
- One gate widening per case branch in `start-coords-settings.tsx` (linear-system and segment, mirroring how PR 2/3/4 widened the `point` / `polygon` / `linear` / `ray` cases).

Reviewer focus:
- The flat-array indexing `i * 2 + j` is consistent across the renderer (`linear-system.tsx`, `segment.tsx`) and the editor (`start-coords-multiline.tsx`'s `updatePointLabel`). The render-side test cases assert both endpoints of both lines/segments to lock the indexing in.
- The fallback chain `buildLabel(...) ?? formatSegment(...)` correctly preserves segment's single-vs-multi default phrasing (`Endpoint N at …` vs `Endpoint N on segment M at …`) — locked in by separate tests for single-segment and multi-segment states.
- The `start-coords-multiline.tsx` refactor drops `Strut` / `spacing` / Aphrodite — these were the only remaining users of those imports in this file, so the removal is clean.

## Test plan:

- [x] `pnpm tsc` — passes
- [x] `pnpm lint <changed files>` — passes
- [x] `pnpm prettier --check <changed files>` — passes
- [x] `pnpm jest packages/perseus/src/widgets/interactive-graphs/graphs/linear-system.test.tsx packages/perseus/src/widgets/interactive-graphs/graphs/segment.test.tsx packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-settings.test.tsx` — 173 pass, 0 fail
- [ ] Reviewer manual: open the `LinearSystemWithCustomLabels` / `SegmentWithCustomLabels` stories. Enable the Storybook a11y addon (or VoiceOver / JAWS) and confirm each interactive endpoint announces *"Point A / B / C / D at …"* matching its prompt. Confirm the existing `LinearSystem` / `Segment` stories still announce the legacy per-line / per-segment defaults.
- [ ] Reviewer manual: open the `InteractiveGraphLinearSystem` / `InteractiveGraphSegment` editor stories, expand "Start coordinates", and confirm each line/segment accordion now shows the shared `CoordInput` stacked tile (`Point N:` heading on top, `x [_]  y [_]` coord row, `Label [_]` field below) instead of the previous bespoke single-row `Point 1:  x [_]  y [_]` layout. Type a letter into one of the label fields and confirm the rendered graph's endpoint `aria-label` *and* the **Screen reader tree** panel update on the same render.

## PR Series and Follow-ups (PRs 6–7)

Independent and parallelizable. None of them touch the foundation files. Each wires its graph type's render side and adds the per-type `start-coords-<type>.tsx` plumbing using the same flat `CoordInput` `pointLabel` / `onPointLabelChange` props PR 2 take-2 introduced.

- 👉🏼 **PR 5** — [Multi-line: linear-system, segment](https://github.com/Khan/perseus/pull/3673)

### DIFF
--- a/packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-multiline.tsx
+++ b/packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-multiline.tsx
@@ -1,12 +1,9 @@
-import {View} from "@khanacademy/wonder-blocks-core";
-import {Strut} from "@khanacademy/wonder-blocks-layout";
-import {spacing} from "@khanacademy/wonder-blocks-tokens";
 import {BodyText} from "@khanacademy/wonder-blocks-typography";
-import {StyleSheet} from "aphrodite";
 import * as React from "react";
 
-import CoordinatePairInput from "../../../components/coordinate-pair-input";
 import PerseusEditorAccordion from "../../../components/perseus-editor-accordion";
+
+import CoordInput from "./coord-input";
 
 import type {CollinearTuple} from "@khanacademy/perseus-core";
 
@@ -14,12 +11,22 @@ type Props = {
     type: "linear-system" | "segment";
     startCoords: CollinearTuple[];
     onChange: (startCoords: CollinearTuple[]) => void;
+    pointLabels?: ReadonlyArray<string>;
+    onChangePointLabels?: (pointLabels: ReadonlyArray<string>) => void;
 };
 
 const StartCoordsMultiline = (props: Props) => {
-    const {startCoords, type, onChange} = props;
+    const {startCoords, type, onChange, pointLabels, onChangePointLabels} =
+        props;
 
     const graphName = type === "segment" ? "Segment" : "Line";
+    const updatePointLabel = (flatIndex: number, newLabel: string) => {
+        const totalPoints = startCoords.length * 2;
+        const next = Array.from({length: totalPoints}, (_, i) =>
+            i === flatIndex ? newLabel : pointLabels?.[i] ?? "",
+        );
+        onChangePointLabels?.(next);
+    };
 
     return (
         <>
@@ -35,48 +42,39 @@ const StartCoordsMultiline = (props: Props) => {
                     }
                     expanded={true}
                 >
-                    <View style={styles.nestedTile}>
-                        <BodyText size="medium" weight="bold" tag="span">
-                            Point 1:
-                        </BodyText>
-                        <Strut size={spacing.small_12} />
-                        <CoordinatePairInput
-                            coord={coordPair[0]}
-                            labels={["x", "y"]}
-                            onChange={(value) => {
-                                const newCoords = [...startCoords];
-                                newCoords[i] = [value, coordPair[1]];
-                                onChange(newCoords);
-                            }}
-                        />
-                    </View>
-                    <View style={styles.nestedTile}>
-                        <BodyText size="medium" weight="bold" tag="span">
-                            Point 2:
-                        </BodyText>
-                        <Strut size={spacing.small_12} />
-                        <CoordinatePairInput
-                            coord={coordPair[1]}
-                            labels={["x", "y"]}
-                            onChange={(value) => {
-                                const newCoords = [...startCoords];
-                                newCoords[i] = [coordPair[0], value];
-                                onChange(newCoords);
-                            }}
-                        />
-                    </View>
+                    <CoordInput
+                        label="Point 1"
+                        coord={coordPair[0]}
+                        onChange={(value) => {
+                            const newCoords = [...startCoords];
+                            newCoords[i] = [value, coordPair[1]];
+                            onChange(newCoords);
+                        }}
+                        pointLabel={pointLabels?.[i * 2]}
+                        onPointLabelChange={
+                            onChangePointLabels &&
+                            ((newLabel) => updatePointLabel(i * 2, newLabel))
+                        }
+                    />
+                    <CoordInput
+                        label="Point 2"
+                        coord={coordPair[1]}
+                        onChange={(value) => {
+                            const newCoords = [...startCoords];
+                            newCoords[i] = [coordPair[0], value];
+                            onChange(newCoords);
+                        }}
+                        pointLabel={pointLabels?.[i * 2 + 1]}
+                        onPointLabelChange={
+                            onChangePointLabels &&
+                            ((newLabel) =>
+                                updatePointLabel(i * 2 + 1, newLabel))
+                        }
+                    />
                 </PerseusEditorAccordion>
             ))}
         </>
     );
 };
-
-const styles = StyleSheet.create({
-    nestedTile: {
-        paddingBottom: spacing.small_12,
-        flexDirection: "row",
-        alignItems: "center",
-    },
-});
 
 export default StartCoordsMultiline;
--- a/packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-multiline.tsx
+++ b/packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-multiline.tsx
@@ -7,15 +7,15 @@ import CoordInput from "./coord-input";
 
 import type {CollinearTuple} from "@khanacademy/perseus-core";
 
-type Props = {
+interface StartCoordsMultilineProps {
     type: "linear-system" | "segment";
     startCoords: CollinearTuple[];
     onChange: (startCoords: CollinearTuple[]) => void;
-    pointLabels?: ReadonlyArray<string>;
-    onChangePointLabels?: (pointLabels: ReadonlyArray<string>) => void;
-};
+    pointLabels: ReadonlyArray<string>;
+    onChangePointLabels: (pointLabels: ReadonlyArray<string>) => void;
+}
 
-const StartCoordsMultiline = (props: Props) => {
+const StartCoordsMultiline = (props: StartCoordsMultilineProps) => {
     const {startCoords, type, onChange, pointLabels, onChangePointLabels} =
         props;
 
@@ -23,9 +23,9 @@ const StartCoordsMultiline = (props: Props) => {
     const updatePointLabel = (flatIndex: number, newLabel: string) => {
         const totalPoints = startCoords.length * 2;
         const next = Array.from({length: totalPoints}, (_, i) =>
-            i === flatIndex ? newLabel : pointLabels?.[i] ?? "",
+            i === flatIndex ? newLabel : pointLabels[i] ?? "",
         );
-        onChangePointLabels?.(next);
+        onChangePointLabels(next);
     };
 
     return (
@@ -50,10 +50,9 @@ const StartCoordsMultiline = (props: Props) => {
                             newCoords[i] = [value, coordPair[1]];
                             onChange(newCoords);
                         }}
-                        pointLabel={pointLabels?.[i * 2]}
-                        onPointLabelChange={
-                            onChangePointLabels &&
-                            ((newLabel) => updatePointLabel(i * 2, newLabel))
+                        pointLabel={pointLabels[i * 2]}
+                        onPointLabelChange={(newLabel) =>
+                            updatePointLabel(i * 2, newLabel)
                         }
                     />
                     <CoordInput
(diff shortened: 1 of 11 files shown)


packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-multiline.tsx
```diff
@@ -35,48 +42,39 @@ const StartCoordsMultiline = (props: Props) => {
                     }
                     expanded={true}
                 >
```
ivyolamit, line 44: Swapped CoordinatePairInput → CoordInput here so multi-line picks up the same per-endpoint Label [_] field the other start-coords editors got in PRs 2–4. CoordInput is the shared wrapper that adds the heading + optional label TextField on top of CoordinatePairInput; the lower-level component is still used (just one layer up). Also drops the local Aphrodite / Strut / spacing imports since CoordInput handles layout.

packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-multiline.tsx
```diff
@@ -1,25 +1,32 @@
...
+import CoordInput from "./coord-input";
+
 import type {CollinearTuple} from "@khanacademy/perseus-core";
 
 type Props = {
     type: "linear-system" | "segment";
     startCoords: CollinearTuple[];
     onChange: (startCoords: CollinearTuple[]) => void;
+    pointLabels?: ReadonlyArray<string>;
+    onChangePointLabels?: (pointLabels: ReadonlyArray<string>) => void;
```
handeyeco: Same question as the last PR. `StartCoordsMultiline` is only called one place and you're passing these to it; why make these optional?

packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-multiline.tsx
```diff
@@ -35,48 +42,39 @@ const StartCoordsMultiline = (props: Props) => {
...
+                        onChange={(value) => {
+                            const newCoords = [...startCoords];
+                            newCoords[i] = [value, coordPair[1]];
+                            onChange(newCoords);
+                        }}
+                        pointLabel={pointLabels?.[i * 2]}
+                        onPointLabelChange={
+                            onChangePointLabels &&
+                            ((newLabel) => updatePointLabel(i * 2, newLabel))
+                        }
```
handeyeco: If `onChangePointLabels` was required, this would be simpler.

packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-settings.test.tsx
```diff
@@ -315,6 +315,75 @@ describe("StartCoordSettings", () => {
                 expect(onChangeMock).toHaveBeenLastCalledWith(expectedCoords);
             },
         );
+
+        it("renders point name fields when onChangePointLabels is provided", () => {
```
handeyeco, line 328: If `onChangePointLabels` was required, you wouldn't need these two tests.

packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-settings.test.tsx
```diff
@@ -315,6 +315,75 @@ describe("StartCoordSettings", () => {
...
+            const onChangePointLabelsMock = jest.fn();
+            render(
+                <StartCoordsSettings
+                    {...defaultProps}
+                    type="segment"
+                    numSegments={2}
+                    onChange={() => {}}
+                    onChangePointLabels={onChangePointLabelsMock}
+                />,
+                {wrapper: RenderStateRoot},
```
handeyeco, line 361: I wonder if there's a Jest/RTL setting we can set to auto-wrap things for us.

packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-settings.test.tsx
```diff
@@ -315,6 +315,75 @@ describe("StartCoordSettings", () => {
...
+            );
+
+            // Act
+            const seg2Point1NameInput = screen.getAllByRole("textbox", {
+                name: "Point 1 name",
+            })[1];
```
handeyeco: This to me is an a11y red flag (for the content creation UI): you have two textboxes in the same component with the same name.

It might be too big of a lift to fix right now, just calling it out. The reason I noticed it was because the test was confusing to me ("why is the third element in `toHaveBeenLastCalledWith` getting changed?"..."Oh, because it's the second segment."), so I would recommend adding some comments.

packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-settings.test.tsx
```diff
@@ -391,6 +460,72 @@ describe("StartCoordSettings", () => {
                 expect(onChangeMock).toHaveBeenLastCalledWith(expectedCoords);
             },
         );
+
+        it("renders point name fields when onChangePointLabels is provided", () => {
```
handeyeco, line 461: If `onChangePointLabels` was required, you wouldn't need these two tests.

packages/perseus-editor/src/widgets/interactive-graph-editor/start-coords/start-coords-settings.test.tsx
```diff
@@ -391,6 +460,72 @@ describe("StartCoordSettings", () => {
...
+            })[0];
+            await userEvent.type(line1Point1NameInput, "T");
+
+            // Assert
+            expect(onChangePointLabelsMock).toHaveBeenLastCalledWith([
+                "T",
+                "",
+                "",
+                "",
+            ]);
```
handeyeco: Some comments here would be useful too.

packages/perseus/src/widgets/interactive-graphs/graphs/linear-system.test.tsx
```diff
@@ -300,6 +300,146 @@ describe("Linear System graph screen reader", () => {
...
+        const line1Point1 = interactiveElements[0];
+        const line1Point2 = interactiveElements[2];
+        const line2Point1 = interactiveElements[3];
+        const line2Point2 = interactiveElements[5];
+
+        // Assert
+        expect(line1Point1).toHaveAttribute(
+            "aria-label",
+            "Point A at -5 comma 5.",
+        );
```
handeyeco: Same comment from the other PR:


```suggestion
        const line2Point2 = screen.getByRole("button", {
            name: "Point A at -5 comma 5.",
        });

        // Assert
        expect(line1Point1).toBeInTheDocument();
```

packages/perseus/src/widgets/interactive-graphs/graphs/linear-system.tsx
```diff
@@ -36,11 +37,13 @@ type LinearSystemGraphProps = MafsGraphProps<LinearSystemGraphState>;
...
 const LinearSystemGraph = (props: LinearSystemGraphProps) => {
     const {dispatch} = props;
-    const {coords: lines} = props.graphState;
+    const {coords: lines, pointLabels} = props.graphState;
 
     const {strings, locale} = usePerseusI18n();
     const id = React.useId();
     const intersectionId = `${id}-intersection`;
+    const buildLabel = (index: number, point: vec.Vector2) =>
+        buildPointAriaLabel(pointLabels, index, point, strings, locale);
```
handeyeco: Nice

packages/perseus/src/widgets/interactive-graphs/graphs/linear-system.tsx
```diff
@@ -91,18 +94,22 @@ const LinearSystemGraph = (props: LinearSystemGraphProps) => {
...
-                            y: srFormatNumber(line[1][1], locale),
-                        }),
+                        point1AriaLabel:
+                            buildLabel(i * 2, line[0]) ??
+                            strings.srLinearSystemPoint({
+                                lineNumber: i + 1,
+                                pointSequence: 1,
+                                x: srFormatNumber(line[0][0], locale),
+                                y: srFormatNumber(line[0][1], locale),
+                            }),
```
handeyeco, line 103: nit: Still would be nice if this was pulled out of JSX and into the logic section of the component.

packages/perseus/src/widgets/interactive-graphs/graphs/segment.test.tsx
```diff
@@ -337,6 +337,159 @@ describe("Segment graph screen reader", () => {
...
+                    pointLabels: ["A", "B"],
+                }}
+            />,
+        );
+        const interactiveElements = screen.getAllByRole("button");
+        const [point1, , point2] = interactiveElements;
+
+        // Assert
+        expect(point1).toHaveAttribute("aria-label", "Point A at -5 comma 5.");
+        expect(point2).toHaveAttribute("aria-label", "Point B at 5 comma 5.");
```
handeyeco: Same here

packages/perseus/src/widgets/interactive-graphs/graphs/segment.tsx
```diff
@@ -33,11 +34,13 @@ export function renderSegmentGraph(
...
 
 const SegmentGraph = ({dispatch, graphState}: SegmentProps) => {
-    const {coords: segments} = graphState;
+    const {coords: segments, pointLabels} = graphState;
     const {strings, locale} = usePerseusI18n();
     const segmentUniqueId = React.useId();
     const lengthDescriptionId = segmentUniqueId + "-length";
     const wholeGraphDescriptionId = segmentUniqueId + "-whole-graph";
+    const buildLabel = (index: number, point: vec.Vector2) =>
+        buildPointAriaLabel(pointLabels, index, point, strings, locale);
```
handeyeco: This is _so_ close to just having a custom hook, which could abstract away `buildPointAriaLabel`'s API, the use of `usePerseusI18n`, and could be shared between components.

packages/perseus/src/widgets/interactive-graphs/interactive-graph.testdata.ts
```diff
@@ -1368,6 +1368,87 @@ export const rayWithCustomLabelsQuestion: PerseusRenderer =
...
+        content:
+            "**Drag the endpoints so line 1 passes through $A$ and $B$, and line 2 passes through $C$ and $D$.**\n\n[[☃ interactive-graph 1]]",
+        markings: "graph",
+        gridStep: [1, 1],
+        snapStep: [1, 1],
+        step: [1, 1],
+        range: [
+            [-10, 10],
+            [-10, 10],
+        ],
```
handeyeco: Same here, I think `generateInteractiveGraphQuestion` will default a lot of this stuff for you.

handeyeco (approved):

packages/perseus/src/widgets/interactive-graphs/graphs/linear-system.test.tsx
```diff
@@ -300,6 +300,146 @@ describe("Linear System graph screen reader", () => {
...
+            "aria-label",
+            "Point C at -5 comma -5.",
+        );
+        expect(line2Point2).toHaveAttribute(
+            "aria-label",
+            "Point D at 5 comma -5.",
+        );
+    });
+
+    it("falls back to the per-line default for truthy non-string entries", () => {
```
catandthemachines, line 402: Interesting safe guard. Do you foresee something like this happening in the current implementation, or just safeguarding against the potential for this to occur?

catandthemachines (approved):
Looks good, had one question on this test. 😉

packages/perseus/src/widgets/interactive-graphs/graphs/linear-system.test.tsx
ivyolamit: This pattern of test was added in one of your comment in previous PR. this scenario is an edge case when the input is coming from the json object when creating a widget.

PR #3674, opened 2026-05-28
[Interactive Graph] Use WB Announcer in Point Graph.

catandthemachines:
## Summary:
Wires the Point graph's `movePoint` action into the existing WB Announcer
pipeline.

What this PR does:
- Reducer: the `case "point"` in `interactive-graph-reducer.ts` now emits
  `stateAnnouncement: {type: "move-point", pointIndex, x, y}` on every
  move. The existing `stateful-mafs-graph.tsx → getAnnouncementText →
  WB Announcer` plumbing then announces it.
- Point graph: both `<MovablePoint>` call sites in `point.tsx`
  (Limited + Unlimited) pass `ariaLive="off"` so the focusable handle's
  legacy `aria-live="polite"` no longer fires alongside the WB Announcer.
  This is the only graph type with the announcer wired up, so other graphs
  (polygon, angle, linear, etc.) are intentionally untouched.
  which tracks removing `aria-live` from `useControlPoint`, `MovablePoint`,
  and `MovableLine` once every graph's reducer is wired to the announcer.

Issue: LEMS-3943

### Manual (reviewer)
- [ ] Open the **Interactive Graph → Point** story in Storybook (`pnpm storybook`). Turn on VoiceOver / NVDA (or enable the Storybook a11y addon's live-region inspector).
- [ ] Tab to one of the point handles. Confirm the initial focus announces *"Point N at X comma Y."* (this comes from `aria-label`, unchanged).
- [ ] Press arrow keys to move the point one snap step at a time. Confirm:
  - You hear *"Point N at X comma Y."* after each move (this is the new WB Announcer payload).
  - You do **not** hear the announcement twice. Before this PR, both `aria-live="polite"` on the handle AND the WB Announcer would have fired.
- [ ] Repeat in the **Unlimited Point** story (multi-point graph) — confirm each point announces only its own move.
- [ ] Regression: open the **Polygon**, **Linear**, and **Circle** stories and confirm they still announce moves the same way they did before (their behavior is unchanged in this PR).

[LEMS-4189]: https://khanacademy.atlassian.net/browse/LEMS-4189?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/circle.tsx
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/circle.tsx
@@ -162,7 +162,6 @@ function MovableCircle(props: {
                 role="button"
                 aria-label={ariaLabel}
                 aria-describedby={ariaDescribedBy}
-                aria-live="polite"
                 aria-disabled={disableKeyboardInteraction}
                 onFocus={() => setFocused(true)}
                 onBlur={() => setFocused(false)}
--- a/packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
@@ -525,6 +525,10 @@ function doMovePoint(
                 coords: newCoords,
             };
         case "point": {
+            const newCoord = boundToEdgeAndSnapToGrid(
+                action.destination,
+                state,
+            );
             return {
                 ...state,
                 hasBeenInteractedWith: true,
@@ -532,11 +536,14 @@ function doMovePoint(
                 coords: setAtIndex({
                     array: state.coords,
                     index: action.index,
-                    newValue: boundToEdgeAndSnapToGrid(
-                        action.destination,
-                        state,
-                    ),
+                    newValue: newCoord,
                 }),
+                stateAnnouncement: {
+                    type: "move-point",
+                    pointIndex: action.index,
+                    x: newCoord[X],
+                    y: newCoord[Y],
+                },
             };
         }
         case "sinusoid": {
--- a/.changeset/tidy-doors-allow.md
+++ b/.changeset/tidy-doors-allow.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/perseus": patch
+---
+
+[Interactive Graph] Add WB Announcer to Point graph events.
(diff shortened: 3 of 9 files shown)


packages/perseus/src/widgets/interactive-graphs/graphs/circle.tsx
```diff
@@ -162,7 +162,6 @@ function MovableCircle(props: {
                 role="button"
                 aria-label={ariaLabel}
                 aria-describedby={ariaDescribedBy}
-                aria-live="polite"
```
handeyeco, line 165: For some you used `aria-live="off"` and for this one you just removed `aria-live`. Is that intentional? Not a loaded question, I genuinely don't know. Not a blocker, I'll defer to your expertise.

handeyeco: Also why is this `aria-live` and the others are `ariaLive`?

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -525,18 +525,25 @@ function doMovePoint(
...
                 hasBeenInteractedWith: true,
                 focusedPointIndex: action.index,
                 coords: setAtIndex({
                     array: state.coords,
                     index: action.index,
-                    newValue: boundToEdgeAndSnapToGrid(
-                        action.destination,
-                        state,
-                    ),
+                    newValue: newCoord,
```
handeyeco, line 539: Nice

handeyeco (approved):

packages/perseus/src/widgets/interactive-graphs/graphs/circle.tsx
catandthemachines: Great question! Essentially the `ariaLive` property is used on `MovablePoint` and `MovableLine`. I can't delete that property since that would break the aria-lives for all graphs that use it and require updating them all at once.

So for these instances I need to set `ariaLive="off"` and delete `aria-live="polite"` on graph specific elements. Once all the graphs no longer use this `ariaLive` property I'll go back through and delete it. https://khanacademy.atlassian.net/browse/LEMS-4189

PR #3678, opened 2026-05-28
[Interactive Graph] Add WB Announcer to Quadratic graph.

catandthemachines:
## Summary:
Wires the Quadratic graph's `movePoint` action into the WB Announcer
pipeline.

What this PR does:
- Reducer: `case "quadratic"` in `doMovePoint` now emits
  `stateAnnouncement: {type: "move-quadratic-point", pointIndex, x, y, vertex}`.
  The reducer pre-computes the vertex from the new quadratic
  coefficients — `vertex` is `null` when `a === 0` (the parabola
  degenerates to a line and has no vertex).
- Types: adds `MoveQuadraticPointAnnouncement` to `InteractiveGraphStateAnnouncement`.
- Screen reader text: `getAnnouncementText` handles
  `"move-quadratic-point"` by composing the quadrant-aware point label
  with the vertex string when present:
  - `getQuadraticPointString(pointIndex + 1, [x, y], strings, locale)` for the moved point
    (origin / on-axis / quadrant 1-4 variants)
  - When `vertex \!== null`, appends `getQuadraticVertexString(vertex, strings)`
    (origin / on-axis / quadrant 1-4 variants)
- Helper move: relocates `getCoordQuadrant`, `getQuadraticVertexString`,
  and `getQuadraticPointString` from `graphs/utils.ts` to
  `graphs/screenreader-text.ts`. `utils.ts` was already importing
  `srFormatNumber` from `screenreader-text.ts`, so importing the
  quadratic string helpers back would have created a cycle. Moving them
  keeps the dependency one-way (`utils.ts` → `screenreader-text.ts`).
  `quadratic.tsx` is updated to import the helpers from their new home.
- Quadratic component: passes `ariaLive="off"` to each `MovablePoint`
  so the WB Announcer is the only source of move announcements for
  quadratic. Includes a `TODO(LEMS-4189)` on the call site to remove
  the prop once `aria-live` is dropped from `useControlPoint`.
- Tests: adds two reducer tests for `movePoint` on quadratic (a normal
  move emits the announcement with a non-null `vertex`; a move that
  produces collinear points emits the announcement with `vertex: null`),
  and four `screenreader-text` tests covering each branch of the
  composed label (point in a quadrant + vertex in a quadrant, point on
  an axis + vertex at the origin, point at the origin + vertex on an
  axis, degenerate-line case where the vertex is omitted entirely).

Issue: LEMS-4197

### Automated
- [x] `pnpm tsc` — passes

### Manual (reviewer)
- [ ] Open the **Interactive Graph → Quadratic** story in Storybook (`pnpm storybook`). Turn on VoiceOver / NVDA (or use the Storybook a11y addon's live-region inspector).
- [ ] Tab to any of the three control-point handles. Confirm the initial focus announces the point's `aria-label` (unchanged — still includes the vertex context).
- [ ] Press arrow keys to move the focused point one snap step at a time. Confirm:
  - You hear the point label appropriate to the point's new location, e.g.:
    - *"Point 1 on parabola in quadrant 2 at -2 comma 4."* when the point lands in a quadrant.
    - *"Point 2 on parabola at 3 comma 0."* when the point lands on an axis.
    - *"Point 3 on parabola at the origin."* when the point lands at the origin.
  - When the parabola is a real curve, you also hear the vertex line, e.g. *"Vertex is in quadrant 4."* / *"Vertex is at the origin."* / *"Vertex is on the X-axis."* / *"Vertex is on the Y-axis."* based on the new vertex location.
  - You do **not** hear the announcement twice. Before this PR, both the per-point `aria-live` and the WB Announcer would have fired.
- [ ] Arrange the three control points so they are collinear (e.g. all on `y = x`). Move one of them and confirm the announcement is **only** the point label — no vertex sentence — since the parabola has degenerated to a line.


packages/perseus/src/widgets/interactive-graphs/graphs/utils.ts
catandthemachines, line 1: Moved these functions to `screenreader-text.ts` to avoid circular dependency.

packages/perseus/src/widgets/interactive-graphs/graphs/quadratic.tsx
```diff
@@ -222,7 +228,9 @@ export function describeQuadraticGraph(
     // Only describe vertex if the quadratic graph is not a line.
     // (Undefined means the quadratic graph is a line and has no vertex.)
     const srQuadraticVertex =
-        a !== 0 ? getQuadraticVertexString(vertex, strings) : undefined;
+        vertex === undefined
```
catandthemachines, line 236: Since the vertex is now calculated using the above function. It will return defined or undefined if there is a vertex to describe.

packages/perseus/src/widgets/interactive-graphs/graphs/utils.ts
```diff
@@ -242,6 +161,22 @@ export function getQuadraticXIntercepts(
...
+/**
+ * Calculate the vertex of a quadratic from its coefficients.
+ *
+ * When a === 0 the parabola degenerates to a line and has no vertex, so
+ * we return undefined; otherwise the vertex is (-b/2a, c - b²/4a).
+ */
+export function getQuadraticVertex(
+    coeffs: QuadraticCoefficient,
+): Coord | undefined {
+    const [a, b, c] = coeffs;
```
handeyeco, line 173: API is weird to me.

It's called like this:

``` TS
getQuadraticVertex([a, b, c])
```

And then it's immediately destructured.

```
const [a, b, c] = coeffs;
```

Could the function just take 3 numbers?

packages/perseus/src/widgets/interactive-graphs/graphs/utils.ts
```diff
@@ -242,6 +161,22 @@ export function getQuadraticXIntercepts(
     return [x1, x2];
 }
 
+/**
+ * Calculate the vertex of a quadratic from its coefficients.
+ *
+ * When a === 0 the parabola degenerates to a line and has no vertex, so
+ * we return undefined; otherwise the vertex is (-b/2a, c - b²/4a).
+ */
+export function getQuadraticVertex(
```
handeyeco, line 170: Are there tests for this?

packages/perseus/src/widgets/interactive-graphs/graphs/utils.ts
```diff
@@ -242,6 +161,22 @@ export function getQuadraticXIntercepts(
...
+    coeffs: QuadraticCoefficient,
+): Coord | undefined {
+    const [a, b, c] = coeffs;
+    if (a === 0) {
+        return undefined;
```
handeyeco, line 175: I think it's more common in JS to return `null` rather than `undefined`. From Gemini: 

> undefined means a variable has been declared but has not yet been assigned a value, while null is an assignment value that represents the intentional absence of any object value

packages/perseus/src/widgets/interactive-graphs/graphs/screenreader-text.ts
```diff
@@ -19,11 +20,37 @@ export function getAnnouncementText(
...
         case "move-center":
             return srCircleCenterLabel(state.x, state.y, strings, locale);
+        case "move-quadratic-point":
+            return srQuadraticPointLabel(state, strings, locale);
         default:
             throw new UnreachableCaseError(state);
     }
 }
 
+function srQuadraticPointLabel(
```
handeyeco, line 81: `srCircleRadiusPointLabel` has tests, but `srQuadraticPointLabel` doesn't. Is that intentional?

packages/perseus/src/widgets/interactive-graphs/graphs/screenreader-text.ts
```diff
@@ -19,11 +20,37 @@ export function getAnnouncementText(
...
+): string {
+    const pointString = getQuadraticPointString(
+        state.pointIndex + 1,
+        [state.x, state.y],
+        strings,
+        locale,
+    );
+    // When vertex is undefined the parabola is degenerate (a line) — no
+    // vertex string to append.
+    if (state.vertex === undefined) {
```
handeyeco, line 111: Related to my other my other comment about not defining things as `undefined`:


```suggestion
    if (state.vertex == null) {
```

packages/perseus/src/widgets/interactive-graphs/graphs/screenreader-text.ts
```diff
@@ -54,6 +81,87 @@ export function srCircleCenterLabel(
     });
 }
 
+type GraphLocations = "origin" | "x-axis" | "y-axis" | 1 | 2 | 3 | 4;
+
+function getCoordQuadrant(coord: Coord): GraphLocations {
```
handeyeco: `getCoordQuadrant`, `getQuadraticVertexString`, and `getQuadraticPointString` also don't have specific tests. Are they tested indirectly by other tests?

handeyeco (approved):
Ideally we'd prefer using `null` over `undefined` directly. Also some questions about test coverage.

packages/perseus/src/widgets/interactive-graphs/graphs/screenreader-text.ts
catandthemachines: Hmm, that was not, let me clean those up.

packages/perseus/src/widgets/interactive-graphs/graphs/utils.ts
catandthemachines: I will say this is just extrapolating existing functionality and moving it into a shared function. To re-align it to return null rather than undefined would require refactoring more logic than the scope of this PR. So I'm going to pass on this suggestion.

packages/perseus/src/widgets/interactive-graphs/graphs/utils.ts
catandthemachines: We could, I just really love that with this property it provides a type with the context of what these values symbolize. I'll keep it as is, but let me know if this is a dealbreaker.

packages/perseus/src/widgets/interactive-graphs/graphs/screenreader-text.ts
catandthemachines: Yeah the `srCircleRadiusPointLabel` tests are a bit redundant. I'll delete those 😉

PR #3680, opened 2026-05-28
Add a "flipbook" story for testing multiple Perseus items

benchristel:
## Summary:
We have an upcoming playtest for the input-number widget. Because input-number
widgets are deprecated and somewhat rare / hard to find on khanacademy.org, we
want an easier way to manually test them (including rendering, user
interaction, and scoring).

This flipbook story lets you paste in a bunch of newline-separated Perseus
items and page through them one at a time.

Issue: none

## Test plan:

```
pnpm storybook
open http://localhost:6006/?path=/story/perseus-item-flipbook--flipbook
```


packages/perseus/src/testing/item-flipbook/item-flipbook-model.ts
benchristel, line 1: The "model" pattern is somewhat unusual for us, but it's a lot less code than the equivalent reducer + actions would be.

The downside is that if you forget to call `this.observer()` in a mutating method, the view doesn't update.

packages/perseus/src/testing/item-flipbook/item-flipbook-model.ts
```diff
@@ -0,0 +1,123 @@
...
+        this.targetItemIndex = this.clampItemIndex(this.targetItemIndex - 1);
+        this.observer();
+    };
+
+    requestItemNumber = (requested: string) => {
+        const requestedIndex = Number.parseInt(requested, 10) - 1;
+        if (this.clampItemIndex(requestedIndex) === requestedIndex) {
+            this.targetItemIndex = requestedIndex;
+            this.observer();
+        }
```
benchristel: Perhaps if the requested index is too low or too high, we should update `targetItemIndex` to the clamped value.

The only time we really don't want to set `targetItemIndex` is if `parseInt` returns `NaN`.

packages/perseus/src/testing/item-flipbook/item-flipbook-model.test.ts
```diff
@@ -0,0 +1,273 @@
...
+        model.setTextareaValue(`
+            {"question":{"content":"1"}}
+            {"question":{"content":"2"}}
+            {"question":{"content":"3"}}
+        `);
+
+        model.requestItemNumber("3");
+
+        expect(model.present().selectedItemNumber.value).toBe("3");
+        expect(observer).toHaveBeenCalledTimes(2);
```
benchristel: I don't love that the number of observer calls is coupled to the test setup. Maybe I should rethink the design of these tests. If we could clear observer calls before the `act` section, this assertion could be changed to expect a single call.

SonicScrewdriver (approved):
Ooooo I love it, thank you so much! Apologies, I thought I had already approved this earlier.

PR #3682, opened 2026-05-29
[Interactive Graph] Fix screen reader text for point graph to include the pointLabel, custom point label

ivyolamit:
## Summary:
Fix screen reader text for point graph to include the pointLabel, custom point label

There was a conflicting work between the custom point label and the using the WB announcer tasks.
This PR will make sure that the custom point label is included in the new logic for WB Annoucer.

Related PRs:
1. https://github.com/Khan/perseus/pull/3619
2. https://github.com/Khan/perseus/pull/3674
3. https://github.com/Khan/perseus/pull/3643

This will also serve as the new pattern moving forward for the related work in implementing the feature for the rest of the graph types.

Issue: LEMS-3943

## Test plan:

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/components/build-point-aria-label.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/components/build-point-aria-label.ts
@@ -3,15 +3,8 @@ import {srFormatNumber} from "../screenreader-text";
 import type {PerseusStrings} from "../../../../strings";
 import type {vec} from "mafs";
 
-/**
- * Returns the screen-reader label for an interactive point at `index`.
- * Returns the custom label from `pointLabels` if present and well-formed;
- * otherwise the 1-indexed sequence number (used as the numeric default).
- *
- * Shared by `buildPointAriaLabel` (focusable-handle aria-label path) and
- * the reducer (live announcement payload) so both surfaces apply the same
- * rule for "what counts as a usable label".
- */
+// Returns the custom label from `pointLabels`, or the 1-indexed sequence
+// number if no custom label is set.
 export function resolvePointLabel(
     pointLabels: ReadonlyArray<string> | undefined,
     index: number,
@@ -19,6 +12,8 @@ export function resolvePointLabel(
     const customLabel = pointLabels?.[index];
     // Fall back to the numeric default if a non-string slips past the parser.
     if (typeof customLabel !== "string" || !customLabel) {
+        // Convert from a 0-indexed array position to the 1-indexed sequence
+        // number screen readers announce (index 0 → "Point 1", etc.).
         return index + 1;
     }
     return customLabel;
--- a/packages/perseus/src/widgets/interactive-graphs/graphs/components/build-point-aria-label.ts
+++ b/packages/perseus/src/widgets/interactive-graphs/graphs/components/build-point-aria-label.ts
@@ -4,6 +4,27 @@ import type {PerseusStrings} from "../../../../strings";
 import type {vec} from "mafs";
 
 /**
+ * Returns the screen-reader label for an interactive point at `index`.
+ * Returns the custom label from `pointLabels` if present and well-formed;
+ * otherwise the 1-indexed sequence number (used as the numeric default).
+ *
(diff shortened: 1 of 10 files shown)


packages/perseus/src/widgets/interactive-graphs/types.ts
```diff
@@ -50,7 +50,7 @@ export type UnlimitedGraphState = PointGraphState | PolygonGraphState;
 
 type MovePointAnnouncement = {
     type: "move-point";
-    pointIndex: number;
+    pointLabel: string;
```
ivyolamit, line 53: Our end goal is to use the pointLabel which will cater the new logic for custom point label. Since this is a new code, we can make this forward compatible for easier code clean up later for https://khanacademy.atlassian.net/browse/LEMS-4206 where we change the translation parameter `num: number | string` -> `pointLabel: string`.

ivyolamit: We do the conversion in https://github.com/Khan/perseus/pull/3682/changes#diff-82b20407edc8b1775676a593d9191c13c456a8ad5044306a3dc8cdd79d037e60R15

packages/perseus/src/widgets/interactive-graphs/graphs/screenreader-text.ts
```diff
@@ -11,7 +11,7 @@ export function getAnnouncementText(
     switch (state.type) {
         case "move-point":
             return strings.srPointAtCoordinates({
-                num: state.pointIndex + 1,
+                num: state.pointLabel, // TODO(LEMS-4206): fix num -> pointLabel
```
ivyolamit, line 14: base on new logic in the reducer this will now be string
🔗 https://github.com/Khan/perseus/pull/3682/changes#diff-ed3550c09301a3a9903bb56053e300ccd8339fac2822f32f26e69fc64ce56119R544

packages/perseus/src/widgets/interactive-graphs/graphs/components/build-point-aria-label.ts
```diff
@@ -3,6 +3,27 @@ import {srFormatNumber} from "../screenreader-text";
...
+ * rule for "what counts as a usable label".
+ */
+export function resolvePointLabel(
+    pointLabels: ReadonlyArray<string> | undefined,
+    index: number,
+): string | number {
+    const customLabel = pointLabels?.[index];
+    // Fall back to the numeric default if a non-string slips past the parser.
+    if (typeof customLabel !== "string" || !customLabel) {
+        return index + 1;
```
catandthemachines, line 17: Since this confused us both at different points. Let's add a comment here.

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -540,7 +541,9 @@ function doMovePoint(
                 }),
                 stateAnnouncement: {
                     type: "move-point",
-                    pointIndex: action.index,
+                    pointLabel: String(
```
catandthemachines, line 544: Idea: what if we just pass the state.pointLabels into the stateAnnouncement. This object is already designed to contain everything needed to announce a state change, and then we don't need to reference the `resolvePointLabel` in this file. Thoughts?

catandthemachines (approved):
LGTM, I agree there might be a better way to organize the shared functions. But it'll likely be easier once both our work is completed. Let's continue to discuss this and if anything it's easy to move functions to different files. But the logic here is sound.

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
ivyolamit: that's a good observation.
If we pass here the state.pointLabels that would mean the processing for getting the label either _Point 1_ (default) or _Point T_ (custom) would then happen in packages/perseus/src/widgets/interactive-graphs/graphs/screenreader-text.ts and it will be for each case in the switch.
By having it hear we would only do this once.

PR #3683, opened 2026-05-29
[Interactive Graph] Add WB Announcer to Linear graph

catandthemachines:
## Summary:
Wires the Linear graph's `movePoint` (endpoint) and `moveLine` (whole-line
grab handle) actions into the existing WB Announcer pipeline.

What this PR does:
- Reducer: `case "linear"` in `doMovePointInFigure` now emits
  `stateAnnouncement: {type: "move-linear-point", pointIndex, x, y}` when an
  endpoint moves. `case "linear"` in `doMoveLine` emits
  `stateAnnouncement: {type: "move-linear-line", coords}` when the whole
  line is dragged via the grab handle. Both `case "linear"` blocks are split
  out from the shared `case "ray"`/`case "vector"` blocks so ray and vector
  (not yet migrated) are intentionally untouched.
- Types: adds `MoveLinearPointAnnouncement` and `MoveLinearLineAnnouncement`
  to `InteractiveGraphStateAnnouncement`.
- Screen reader text: `getAnnouncementText` handles `"move-linear-point"`
  with the generic `srPointAtCoordinates` string (*"Point N at X comma Y."*,
  matching the endpoints' `aria-label`) and `"move-linear-line"` with
  `srLinearGrabHandle` (*"Line going through point X1 comma Y1 and point X2
  comma Y2."*).
- Component: adds an optional `ariaLive` override prop to `MovableLine`. It
  defaults to `undefined`, preserving the internal `aria-live` state machine
  for the other graphs that still share `MovableLine` (ray, segment,
  linear-system). `linear.tsx` passes `ariaLive="off"` so the WB Announcer
  is the only source of move announcements for the Linear graph.
  which tracks removing `aria-live` from `useControlPoint` and `MovableLine`
  once every line graph's reducer is wired to the announcer.
- Tests: adds reducer tests for `movePoint` and `moveLine` on linear, adds
  two `screenreader-text` tests (`move-linear-point` + `move-linear-line`),
  and removes the now-obsolete "Should update the aria-live when X is moved"
  test from `linear.test.tsx`.

Issue: LEMS-4198

### Automated
- [x] `pnpm tsc` — passes
- [x] `pnpm jest packages/perseus/src/widgets/interactive-graphs` — passes

### Manual (reviewer)
- [ ] Open the **Interactive Graph → Linear** story in Storybook (`pnpm storybook`). Turn on VoiceOver / NVDA (or use the Storybook a11y addon's live-region inspector).
- [ ] Tab to one of the endpoint handles. Confirm the initial focus announces *"Point N at X comma Y."* (from `aria-label`, unchanged).
- [ ] Press arrow keys to move the endpoint one snap step at a time. Confirm:
  - You hear *"Point N at X comma Y."* after each move (new WB Announcer payload).
  - You do **not** hear the announcement twice. Before this PR, both the endpoint's `aria-live` and the WB Announcer would have fired.
- [ ] Tab to the line itself (the whole-line grab handle). Press arrow keys to drag the line. Confirm you hear *"Line going through point X1 comma Y1 and point X2 comma Y2."* after each keyboard drag — this is the new `move-linear-line` announcement, and you should hear it only once.
- [ ] Regression: open the **Ray**, **Segment**, and **Linear System** stories and confirm they still announce moves the same way they did before (their behavior is unchanged in this PR).

[LEMS-4189]: https://khanacademy.atlassian.net/browse/LEMS-4189


packages/perseus/src/widgets/interactive-graphs/graphs/screenreader-text.ts
```diff
@@ -19,11 +19,35 @@ export function getAnnouncementText(
             return `${srCircleRadiusPointLabel(state.x, state.y, state.centerX, strings, locale)} ${strings.srCircleRadius({radius: state.radius})}`;
         case "move-center":
             return srCircleCenterLabel(state.x, state.y, strings, locale);
+        case "move-linear-point":
```
catandthemachines: Nit: Let me consolidate this with the generic `move-point` announcement. No need to duplicate if it's the same.

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -393,7 +421,23 @@ function doMoveLine(
                 coords: newCoords,
             };
         }
-        case "linear":
+        case "linear": {
+            const constrainedLine = constrainShapePreservingMove(
```
catandthemachines, line 430: nit: Add a similar comment like above so this logic is consoludated once ray and vector are moved over.

packages/perseus/src/widgets/interactive-graphs/graphs/components/movable-line.tsx
ivyolamit, line 1: Just a heads up, i'll be doing a change to this file for a bug fix. But i'll wait for your changes in this PR, Linear, and the Ray. So I can also refactor this component since this file is getting larger.

ivyolamit (approved):
👍🏼

PR #3685, opened 2026-05-29
[Interactive Graph] Add WB Announcer to Ray graph

catandthemachines:
## Summary:
Wires the Ray graph's `movePoint` (endpoint) and `moveRay` (whole-line grab
handle) actions into the existing WB Announcer pipeline.

What this PR does:
- Reducer: `case "ray"` in `doMovePointInFigure` now emits
  `stateAnnouncement: {type: "move-ray-point", pointIndex, x, y}` when an
  endpoint moves. `case "ray"` in `doMoveLine` emits
  `stateAnnouncement: {type: "move-ray-line", coords}` when the whole ray is
  dragged via the grab handle. Both `case "ray"` blocks are split out from
  the shared `case "linear"`/`case "vector"` blocks so linear and vector
  are intentionally untouched.
- Types: adds `MoveRayPointAnnouncement` and `MoveRayLineAnnouncement` to
  `InteractiveGraphStateAnnouncement`.
- Screen reader text: `getAnnouncementText` handles `"move-ray-point"` by
  index — the endpoint (index 0) uses `srRayEndpoint` (*"Endpoint at X comma
  Y."*) and the terminal point (index 1) uses `srRayTerminalPoint`
  (*"Through point at X comma Y."*) — and `"move-ray-line"` uses
  `srRayGrabHandle` (*"Ray with endpoint X1 comma Y1 going through point X2
  comma Y2."*).
- Component: adds an optional `ariaLive` override prop to `MovableLine`. It
  defaults to `undefined`, preserving the internal `aria-live` state machine
  for the other graphs that still share `MovableLine` (linear, segment,
  linear-system). `ray.tsx` passes `ariaLive="off"` so the WB Announcer is
  the only source of move announcements for the Ray graph.
  which tracks removing `aria-live` from `useControlPoint` and `MovableLine`
  once every line graph's reducer is wired to the announcer.
- Tests: adds reducer tests for `movePoint` and `moveRay` on ray, adds
  `screenreader-text` tests (`move-ray-point` at both indices +
  `move-ray-line`), and removes the now-obsolete "Should update the
  aria-live when X is moved" test from `ray.test.tsx`.

> Note: the `ariaLive` override prop on `MovableLine` is the same shared
> infrastructure introduced in the Linear PR (#3683). This PR re-introduces
> it identically so the two can merge in either order with a trivial /
> no-op conflict on `movable-line.tsx`.

Issue: LEMS-4199

### Automated
- [x] `pnpm tsc` — passes
- [x] `pnpm jest packages/perseus/src/widgets/interactive-graphs` — passes

### Manual (reviewer)
- [ ] Open the **Interactive Graph → Ray** story in Storybook (`pnpm storybook`). Turn on VoiceOver / NVDA (or use the Storybook a11y addon's live-region inspector).
- [ ] Tab to the ray's endpoint handle. Confirm the initial focus announces *"Endpoint at X comma Y."* (from `aria-label`, unchanged).
- [ ] Press arrow keys to move the endpoint one snap step at a time. Confirm:
  - You hear *"Endpoint at X comma Y."* after each move (new WB Announcer payload).
  - You do **not** hear the announcement twice. Before this PR, both the endpoint's `aria-live` and the WB Announcer would have fired.
- [ ] Tab to the terminal point (the point the ray passes through). Confirm moves announce *"Through point at X comma Y."*.
- [ ] Tab to the ray itself (the whole-line grab handle). Press arrow keys to drag the ray. Confirm you hear *"Ray with endpoint X1 comma Y1 going through point X2 comma Y2."* after each keyboard drag — this is the new `move-ray-line` announcement, and you should hear it only once.
- [ ] Regression: open the **Linear**, **Segment**, and **Linear System** stories and confirm they still announce moves the same way they did before (their behavior is unchanged in this PR).

[LEMS-4189]: https://khanacademy.atlassian.net/browse/LEMS-4189


ivyolamit (approved):
👍🏼

PR #3686, opened 2026-05-29
[Interactive Graph] Add WB Announcer to Vector graph.

catandthemachines:
## Summary:
Wires the Vector graph's keyboard move actions into the WB Announcer pipeline.

- **Reducer + types:** the `vector` cases now emit `move-vector-point` / `move-vector-line` announcements (added to `InteractiveGraphStateAnnouncement`); the point announcement carries the point index (tail = 0, tip = 1) and the line announcement carries both endpoints.
- **Screen reader text:** `getAnnouncementText` handles both via a new `srVectorPointLabel` helper — the tail uses the generic point label while the tip gets its dedicated label, and whole-vector drags describe the tail/tip coordinates.
- **Component:** the vector's `aria-live="polite"` is dropped and the tip arrowhead gets `ariaLive="off"` so the WB Announcer is the only source of move announcements (no double-announce), with a `TODO(LEMS-4189)` to remove the prop once `aria-live` is dropped from `useControlArrowhead`.

Issue: LEMS-4200

### Automated
- [x] `pnpm tsc` — passes

### Manual (reviewer)
- [ ] Open the **Interactive Graph → Vector** story in Storybook (`pnpm storybook`). Turn on VoiceOver / NVDA (or use the Storybook a11y addon's live-region inspector).
- [ ] Tab to the vector's tail handle and press arrow keys to move it one snap step at a time. Confirm you hear the point's new location, e.g. *"Point 1 at -3 comma 2."*
- [ ] Tab to the vector's tip handle and move it. Confirm you hear the tip summary, e.g. *"Tip of vector at 7 comma -4."*
- [ ] Tab to the vector's grab handle and move the whole vector. Confirm you hear the vector summary describing the tail and tip coordinates.
- [ ] Confirm you do **not** hear the announcement twice — before this PR, both the per-point `aria-live` and the WB Announcer would have fired.


packages/perseus/src/widgets/interactive-graphs/graphs/screenreader-text.test.ts
```diff
@@ -330,61 +330,117 @@ describe("getAnnouncementText", () => {
         });
     });
 
-    it("returns the endpoint label for a move-ray-point announcement at index 0", () => {
-        const result = getAnnouncementText(
-            {type: "move-ray-point", pointIndex: 0, pointLabel: 1, x: -3, y: 2},
-            mockStrings,
-            "en",
-        );
+    describe("move-ray-point", () => {
```
catandthemachines, line 333: Some tests were missing their describe blocks, went back through and added them back 😉

packages/perseus/src/widgets/interactive-graphs/reducer/interactive-graph-reducer.ts
```diff
@@ -489,38 +476,29 @@ function doMoveLine(
...
-                    ? {type: "move-ray-line", coords: constrainedLine}
-                    : {type: "move-linear-line", coords: constrainedLine};
-            return {
-                ...state,
-                type: state.type,
-                hasBeenInteractedWith: true,
+
+            const stateAnnouncement: InteractiveGraphStateAnnouncement = {
+                type:
+                    state.type === "ray"
```
ivyolamit: This is just my preference, using this kind of conditional is ideally for one liner a simple if else. Suggestion use a helper function to get the type.

ivyolamit (approved):
@catandthemachines changes look good to me, i just have one suggestion

PR #3687, opened 2026-05-29
[Interactive Graph] Add WB Announcer to Linear System graph

catandthemachines:
## Summary:
Wires the Linear System graph's keyboard move actions into the WB Announcer pipeline.

- **Reducer + types:** the `linear-system` cases in `doMovePointInFigure` and `doMoveLine` now emit `move-linear-system-point` / `move-linear-system-line` announcements (added to `InteractiveGraphStateAnnouncement`), each carrying the line index.
- **Screen reader text:** `getAnnouncementText` handles both via a new `srLinearSystemPointLabel` helper — a custom author label overrides the wording, otherwise it falls back to the `"Point N on line M at X, Y"` / grab-handle semantics.
- **Component:** each `MovableLine` gets `ariaLive="off"` so the WB Announcer is the only source of move announcements (no double-announce), with a `TODO(LEMS-4189)` to remove the prop once `aria-live` is dropped from `useControlPoint`.

Issue: LEMS-4201

### Automated
- [x] `pnpm tsc` — passes

### Manual (reviewer)
- [ ] Open the **Interactive Graph → Linear System** story in Storybook (`pnpm storybook`). Turn on VoiceOver / NVDA (or use the Storybook a11y addon's live-region inspector).
- [ ] Tab to one of the line's endpoint handles and press arrow keys to move it one snap step at a time. Confirm you hear the point's new location, e.g. *"Point 1 on line 2 at -3 comma 2."*
- [ ] Tab to a line's grab handle and move the whole line. Confirm you hear the line summary, e.g. *"Line 2 going through point -3 comma -4 and point 7 comma -4."*
- [ ] Set a custom label on a point and confirm the move announcement uses it (e.g. *"Point B at -3 comma 2."*) instead of the default line/point wording.
- [ ] Confirm you do **not** hear the announcement twice — before this PR, both the per-point `aria-live` and the WB Announcer would have fired.


packages/perseus/src/widgets/interactive-graphs/graphs/screenreader-text.ts
```diff
@@ -127,6 +137,34 @@ function srAnglePointLabel(
...
+    const y = srFormatNumber(state.y, locale);
+    // A custom author label overrides the line/point semantics, matching
+    // the static aria-label behavior in linear-system.tsx.
+    // TODO(LEMS-4206): Once we update the translation keys to allow custom labels
+    // we can remove this block in favor of using the logic below.
+    if (typeof state.pointLabel === "string") {
+        return strings.srPointAtCoordinates({num: state.pointLabel, x, y});
+    }
+    return strings.srLinearSystemPoint({
+        lineNumber: state.lineIndex + 1,
```
ivyolamit, line 167: @catandthemachines what's the difference with lineNumber and pointSequence?

to my knowledge the sequence number is what we currently have that will mention _Point 1_  like this [here](https://github.com/Khan/perseus/blob/3e96e9ec8c906625d58ad3e36cc85f9249e0296b/packages/perseus/src/widgets/interactive-graphs/graphs/point.tsx#L95)

catandthemachines: That is correct, pointSequence refers to the point number `Point 1 ...`. lineNumber refers to the index of the line in the sequence of lines that get's read out as `Endpoint 1 on segment 2`. Is there something I'm missing in the implementation?

ivyolamit: ah gotcha thanks for the info

ivyolamit (approved):

PR #3689, opened 2026-05-29
Remove deprecated/unused APIOptions and renderer props

benchristel:
## Summary:
All of the removed properties were either not used by Perseus or not used in
any clients.

This PR will be easiest to review one commit at a time.

Issue: none

## Test plan:

CI checks should pass.

### DIFF
--- a/packages/perseus/src/components/svg-image.tsx
+++ b/packages/perseus/src/components/svg-image.tsx
@@ -18,14 +18,14 @@ import {ZoomImageButton} from "./zoom-image-button";
 
 import type {ImageProps} from "./image-loader";
 import type {Coord} from "../interactive2/types";
-import type {APIOptions, Dimensions} from "../types";
+import type {APIOptions} from "../types";
 import type {Alignment, Size} from "@khanacademy/perseus-core";
 
 function isImageProbablyPhotograph(imageUrl) {
     return /\.(jpg|jpeg)$/i.test(imageUrl);
 }
 
-function defaultPreloader(dimensions: Dimensions) {
+function spinner() {
     return (
         <span
             style={{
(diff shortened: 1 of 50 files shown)


packages/perseus/src/components/svg-image.tsx
```diff
@@ -485,7 +472,7 @@ class SvgImage extends React.Component<Props, State> {
                                 <ImageLoader
                                     src={imageSrc}
                                     imgProps={imageProps}
-                                    preloader={preloader}
+                                    preloader={spinner}
```
nishasy, line 475: I'd even go as far as to remove the `preloader` prop from `ImageLoader` entirely, and just return the spinner in its default state branch rather than `this.props.preloader()`.

Ideally, we'd wrap it in a container that has the same width/height as the image that's being loaded to avoid layout shift.

nishasy (commented):
One small optional suggestion, but this looks right to me otherwise.

I'm holding off on approving, since I think someone with more knowledge about APIOptions should also take a look at this PR first.

packages/perseus/src/perseus-api.tsx
jeremywiebe, line 35: **note**: Eventually, I'd love to get rid of this old `prop-types` `ApiOptions` entirely. No action needed for this PR.

jeremywiebe (approved):
Love it! Thanks for cleaning up dead code/options!

packages/perseus/src/components/svg-image.tsx
benchristel: I looked into making this change, but some uses of `SvgImage` don't pass a `preloader` prop, which causes `SvgImage` to not display any loading indicator. To preserve the existing behavior, I think we need to keep the `preloader` prop.

nishasy: Cool, thanks for checking

PR #3690, opened 2026-05-29
Delete unused highlightLint field from PerseusCategorizerWidgetOptions

benchristel:
Issue: none

## Test plan:

CI checks should pass.

### DIFF
--- a/packages/perseus-core/src/data-schema.ts
+++ b/packages/perseus-core/src/data-schema.ts
@@ -537,9 +537,6 @@ export type PerseusCategorizerWidgetOptions = {
      * the category. e.g. [0, 1, 0, 1, 2]
      */
     values: number[];
-    /** Whether we should highlight i18n linter errors found on this widget */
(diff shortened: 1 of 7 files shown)


packages/perseus-core/src/data-schema.ts
```diff
@@ -537,9 +537,6 @@ export type PerseusCategorizerWidgetOptions = {
      * the category. e.g. [0, 1, 0, 1, 2]
      */
     values: number[];
-    /** Whether we should highlight i18n linter errors found on this widget */
-    // TODO(benchristel): highlightLint doesn't seem to be used. Delete it.
-    highlightLint?: boolean;
```
benchristel, line 542: @ivyolamit heads up since you mentioned you and Earl will be doing the next release. There's some test data in webapp that still includes `highlightLint: false` — you might need to update that if there are type errors.

SonicScrewdriver (approved):

PR #3691, opened 2026-05-29
Don't show ungraded text for 'none' graph

handeyeco:
## Summary:

Small change to hide "ungraded" text when a graph is "none" type (which is already not graded).

See: https://khanacademy.slack.com/archives/C05MUPEJUPP/p1780076351228929


ivyolamit (approved):
LGTM 👍🏼

PR #3692, opened 2026-05-29
Fix flaky TeX Chromatic snapshots for Matcher and Sorter widgets

Myranae:
## Summary:
- Fixes flaky Chromatic snapshots for Matcher and Sorter widgets containing TeX content. Previously, `ServerItemRenderer.onRendered` fired as soon as TeX finished typesetting, but before the cross-column measurement cascade settled — so snapshots could capture intermediate layouts.
- Moves readiness coordination from the Sortable component (the approach attempted in #3612) up to the widget level where the cascade structure is actually known. Matcher waits for both columns to measure at heights consistent with the shared constraint; Sorter settles after its first measurement (no constraint distribution loop). The Sortable component itself remains a dumb measurement primitive — it doesn't know how many constraint-propagation rounds its parent needs and shouldn't be making that decision.
- Introduces `withAssetContext`, a small HOC that mirrors `withDependencies` and injects `setAssetStatus` plus a `useId()`-generated `assetKey` as props. This lets class-component widgets consume AssetContext without changing their existing `contextType`.
- Supersedes #3612 — see that PR's discussion thread for the analysis that led to this architectural reframe (and the reasons the Sortable-level timer approach was abandoned).

Issue: LEMS-4117

## Test plan:
- [ ] Manually verify in Storybook that `Widgets/Matcher/Visual Regression Tests/Initial State/With TeX` renders stably with no layout flicker.
- [ ] Manually verify `Widgets/Sorter/Visual Regression Tests/Initial State/With TeX` renders stably.
- [ ] Confirm existing Matcher behavior is unaffected (drag/drop reordering, scoring, focus handling).
- [ ] Confirm existing Sorter behavior is unaffected.
- [ ] Run Chromatic and confirm the new `WithTeX` stories produce stable baselines across multiple runs.
- [ ] Confirm CI passes: `pnpm test`, `pnpm tsc`, `pnpm lint`.

### DIFF
--- a/packages/perseus/src/components/__tests__/with-asset-context.test.tsx
+++ b/packages/perseus/src/components/__tests__/with-asset-context.test.tsx
@@ -20,7 +20,6 @@ function MockComponent(props: MockComponentProps): React.ReactElement {
     return (
         <button
             onClick={() => props.setAssetStatus(props.assetKey, true)}
-            data-testid={props.label}
             data-asset-key={props.assetKey}
         >
             {props.label}
@@ -46,7 +45,7 @@ describe("withAssetContext", () => {
                 <Wrapped label="hello" />
             </AssetContext.Provider>,
         );
-        screen.getByTestId("hello").click();
+        screen.getByRole("button", {name: "hello"}).click();
 
         // Assert
         expect(setAssetStatusSpy).toHaveBeenCalled();
@@ -58,9 +57,9 @@ describe("withAssetContext", () => {
         render(<Wrapped label="hello" />);
 
         // Assert
-        expect(screen.getByTestId("hello").dataset.assetKey).toMatch(
-            /^widget-/,
-        );
+        expect(
+            screen.getByRole("button", {name: "hello"}).dataset.assetKey,
+        ).toMatch(/^widget-/);
     });
 
     it("generates a unique assetKey per rendered instance", () => {
@@ -74,8 +73,8 @@ describe("withAssetContext", () => {
         );
 
         // Assert
-        const first = screen.getByTestId("first");
-        const second = screen.getByTestId("second");
+        const first = screen.getByRole("button", {name: "first"});
+        const second = screen.getByRole("button", {name: "second"});
         expect(first.dataset.assetKey).not.toBe(second.dataset.assetKey);
     });
 
@@ -87,7 +86,7 @@ describe("withAssetContext", () => {
         // Assert — the HOC still generates an assetKey (useId works without
         // a Provider) and the default no-op setAssetStatus is callable (so
         // the click handler doesn't crash).
-        const button = screen.getByTestId("hello");
+        const button = screen.getByRole("button", {name: "hello"});
         expect(button.dataset.assetKey).toMatch(/^widget-/);
         expect(() => button.click()).not.toThrow();
     });
--- a/packages/perseus/src/components/__tests__/with-asset-context.test.tsx
+++ b/packages/perseus/src/components/__tests__/with-asset-context.test.tsx
@@ -1,0 +1,94 @@
+import {render, screen} from "@testing-library/react";
+import * as React from "react";
+
+import AssetContext from "../../asset-context";
+import {withAssetContext} from "../with-asset-context";
+
+// Minimal test-only component used to exercise withAssetContext. It accepts
+// the props the HOC injects (`setAssetStatus` and `assetKey`) and surfaces
+// them in two ways the tests can observe: as a `data-asset-key` attribute on
+// the rendered button (for inspecting which key the HOC generated), and via
+// a click handler that calls `setAssetStatus` (so tests can verify the
+// injected function reaches the wrapped component and is callable).
+type MockComponentProps = {
+    label: string;
+    setAssetStatus: (assetKey: string, loaded: boolean) => void;
+    assetKey: string;
+};
+
+function MockComponent(props: MockComponentProps): React.ReactElement {
+    return (
+        <button
+            onClick={() => props.setAssetStatus(props.assetKey, true)}
+            data-testid={props.label}
+            data-asset-key={props.assetKey}
+        >
+            {props.label}
+        </button>
+    );
+}
+
+describe("withAssetContext", () => {
+    it("injects setAssetStatus from AssetContext as a prop", () => {
+        // Arrange
+        const setAssetStatusSpy = jest.fn();
+        const Wrapped = withAssetContext(MockComponent, "widget");
+
+        // Act — clicking calls setAssetStatus via the injected prop, which
+        // proves the function from the Provider reached MockComponent.
+        render(
+            <AssetContext.Provider
+                value={{
+                    assetStatuses: {},
+                    setAssetStatus: setAssetStatusSpy,
+                }}
+            >
+                <Wrapped label="hello" />
+            </AssetContext.Provider>,
+        );
+        screen.getByTestId("hello").click();
+
+        // Assert
+        expect(setAssetStatusSpy).toHaveBeenCalled();
+    });
+
+    it("injects an assetKey with the configured prefix", () => {
+        // Arrange, Act
+        const Wrapped = withAssetContext(MockComponent, "widget");
+        render(<Wrapped label="hello" />);
+
+        // Assert
+        expect(screen.getByTestId("hello").dataset.assetKey).toMatch(
+            /^widget-/,
+        );
+    });
+
+    it("generates a unique assetKey per rendered instance", () => {
+        // Arrange, Act
+        const Wrapped = withAssetContext(MockComponent, "widget");
+        render(
+            <>
+                <Wrapped label="first" />
+                <Wrapped label="second" />
+            </>,
+        );
+
+        // Assert
+        const first = screen.getByTestId("first");
+        const second = screen.getByTestId("second");
+        expect(first.dataset.assetKey).not.toBe(second.dataset.assetKey);
+    });
+
+    it("uses the AssetContext default no-op when no Provider is above", () => {
+        // Arrange, Act — render without an AssetContext.Provider.
+        const Wrapped = withAssetContext(MockComponent, "widget");
+        render(<Wrapped label="hello" />);
+
+        // Assert — the HOC still generates an assetKey (useId works without
+        // a Provider) and the default no-op setAssetStatus is callable (so
+        // the click handler doesn't crash).
+        const button = screen.getByTestId("hello");
+        expect(button.dataset.assetKey).toMatch(/^widget-/);
+        expect(() => button.click()).not.toThrow();
+    });
+});
--- a/packages/perseus/src/components/with-asset-context.tsx
+++ b/packages/perseus/src/components/with-asset-context.tsx
@@ -1,0 +1,43 @@
+import * as React from "react";
+
+import AssetContext from "../asset-context";
+
+type SetAssetStatus = (assetKey: string, loaded: boolean) => void;
+
+type InjectedProps = {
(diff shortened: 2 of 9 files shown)


Myranae:
@claude review once

Myranae:
@benchristel After your review on the previous iteration of this work, I decided to start a new PR with the new direction. I went over this pretty closely with Claude, but still waiting for the Github Claude review to finish and will get to that on Monday. Does this new direction align more with how you would expect this to look? It no longer uses any sort of timer mechanic.

packages/perseus/src/components/__tests__/with-asset-context.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+    label: string;
+    setAssetStatus: (assetKey: string, loaded: boolean) => void;
+    assetKey: string;
+};
+
+function MockComponent(props: MockComponentProps): React.ReactElement {
+    return (
+        <button
+            onClick={() => props.setAssetStatus(props.assetKey, true)}
+            data-testid={props.label}
```
jeremywiebe: **nit**: If you omit this, can you use `screen.getByRole("button", {name: /label/});` instead?  That'd avoid using test ids, I think.

packages/perseus/src/components/with-asset-context.tsx
```diff
@@ -0,0 +1,43 @@
...
+                // eslint-disable-next-line no-restricted-syntax -- TypeScript can't verify that spreading `Omit<P, K>` + adding `K`-properties reconstitutes P (the "P could be a different subtype" limitation). The runtime behavior is sound: we provide every field P needs.
+                {...(props as P)}
+                setAssetStatus={setAssetStatus}
+                assetKey={assetKey}
+                ref={ref}
+            />
+        );
+    });
+
+    WithAssetContextComponent.displayName = `withAssetContext(${
```
jeremywiebe, line 38: Nice. This is helpful!

packages/perseus/src/widgets/matcher/matcher.tsx
```diff
@@ -81,22 +87,73 @@ export class Matcher extends React.Component<Props, State> implements Widget {
...
+    // the constraint to change. Settling requires both sides stable (see
+    // _maybeSettle). The check is intentionally not symmetric with the
+    // *opposite* side's state; it just verifies "this measurement won't
+    // drive future change."
     onMeasureLeft: (arg1: any) => void = (dimensions) => {
         const height = _.max(dimensions.heights);
+        const currentConstraint = _.max([
+            this.state.leftHeight,
+            this.state.rightHeight,
+        ]);
```
jeremywiebe: Can we use `Math.max()` instead? 


```suggestion
        const currentConstraint = Math.max(
            this.state.leftHeight,
            this.state.rightHeight,
        );
```

packages/perseus/src/widgets/matcher/matcher.tsx
```diff
@@ -81,22 +87,73 @@ export class Matcher extends React.Component<Props, State> implements Widget {
...
+            this._maybeSettle();
+        }
     };
 
     onMeasureRight: (arg1: any) => void = (dimensions) => {
         const height = _.max(dimensions.heights);
+        const currentConstraint = _.max([
+            this.state.leftHeight,
+            this.state.rightHeight,
+        ]);
```
jeremywiebe: Can we use `Math.max()` instead? 


```suggestion
        const currentConstraint = Math.max(
            this.state.leftHeight,
            this.state.rightHeight,
        );
```

jeremywiebe (commented):
I think this looks right. I'll hold off approving as Ben had the original thoughts on this re-working.

packages/perseus/src/widgets/matcher/__docs__/matcher-initial-state-regression.stories.tsx
```diff
@@ -57,30 +57,28 @@ export const WithoutLabels: Story = {
...
+            "$f(x) = \\dfrac{1}{x}$",
+            "$f(x) = \\dfrac{1}{x^2}$",
+            "$f(x) = \\dfrac{1}{x^3}$",
+        ],
+        right: [
+            "$f'(x) = -\\dfrac{1}{x^2}$",
+            "$f'(x) = -\\dfrac{2}{x^3}$",
+            "$f'(x) = -\\dfrac{3}{x^4}$",
+        ],
+    },
```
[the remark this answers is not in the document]
Myranae: Switched to using serverItemRendererWithDebugUI in TeX stories here

packages/perseus/src/widgets/matcher/matcher.tsx
```diff
@@ -81,22 +87,73 @@ export class Matcher extends React.Component<Props, State> implements Widget {
...
         this.setState({rightHeight: height});
+        if (height === currentConstraint && currentConstraint > 0) {
+            this._rightStable = true;
+            this._maybeSettle();
+        }
```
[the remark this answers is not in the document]
Myranae: Fixed so that the check checks <== instead of ===. This is to recognize that smaller heights won't ever equal the constraight so this allows it to get moved to true instead of staying false forever.

benchristel:
> Does this new direction align more with how you would expect this to look? It no longer uses any sort of timer mechanic.

Yes, thanks! Not using a timer is the crucial point. Going to do a thorough review this afternoon.

packages/perseus/src/components/with-asset-context.tsx
```diff
@@ -0,0 +1,43 @@
...
+        const {setAssetStatus} = React.useContext(AssetContext);
+        const uniqueId = React.useId();
+        const assetKey = `${keyPrefix}-${uniqueId}`;
+
+        return (
+            <WrappedComponent
+                // eslint-disable-next-line no-restricted-syntax -- TypeScript can't verify that spreading `Omit<P, K>` + adding `K`-properties reconstitutes P (the "P could be a different subtype" limitation). The runtime behavior is sound: we provide every field P needs.
+                {...(props as P)}
+                setAssetStatus={setAssetStatus}
+                assetKey={assetKey}
```
benchristel, line 32: The components that use this HOC don't do anything with `assetKey` except pass it to `setAssetStatus`. We can therefore omit the `assetKey` prop and parameter.

```suggestion
                setAssetStatus={(loaded) => setAssetStatus(assetKey, loaded)}
```

packages/perseus/src/widgets/matcher/matcher.test.tsx
benchristel, line 55: It seems like we might want our fake TeX component to call its `onRender` callback in other tests too. What do you think about moving this code to `test-dependencies.tsx` — replacing the fake TeX component there?

packages/perseus/src/widgets/matcher/matcher.tsx
benchristel: The constraint calculation is now duplicated in 3 places: `onMeasureLeft`, `onMeasureRight`, and `render`. It should be extracted to a method.

packages/perseus/src/widgets/matcher/matcher.test.tsx
```diff
@@ -216,4 +224,175 @@ describe("matcher widget", () => {
...
+                matcher.onMeasureLeft({heights: [74, 74, 74], widths: []});
+            });
+            act(() => {
+                matcher.onMeasureRight({heights: [80, 80, 80], widths: []});
+            });
+
+            // Assert — both sides stable and settled despite left < constraint.
+            expect(matcher._leftStable).toBe(true);
+            expect(matcher._rightStable).toBe(true);
+            expect(setAssetStatusSpy).toHaveBeenCalledWith(true);
```
benchristel, line 395: I don't think this test is correct. `_leftStable` should not be `true` at this point. We've just measured the left column to have shorter cards than the right column, so the left column needs to be rerendered with `constraints={{height: 80}}` to make its cards taller.

On a totally unrelated note, all of these tests are directly manipulating the internals of the Matcher widget, which makes them unrealistic, brittle in the face of changes to Matcher, and mistake-prone.

Instead, we should have:

- a realistic test that asserts that a Renderer's `onRendered` callback is eventually called when there is a Matcher widget containing TeX.
- unit tests for a state machine that keeps track of card measurements. Those tests would test the same logic that's being covered here, but without involving React.

packages/perseus/src/widgets/matcher/matcher.tsx
```diff
@@ -81,22 +87,71 @@ export class Matcher extends React.Component<Props, State> implements Widget {
...
+    // trigger another re-measurement cycle. Settling requires both sides
+    // stable (see _maybeSettle).
     onMeasureLeft: (arg1: any) => void = (dimensions) => {
         const height = _.max(dimensions.heights);
+        const currentConstraint = _.max([
+            this.state.leftHeight,
+            this.state.rightHeight,
+        ]);
         this.setState({leftHeight: height});
+        if (height <= currentConstraint && currentConstraint > 0) {
```
benchristel, line 135: This doesn't seem right. `currentConstraint` is effectively the minimum height that a card must have, so we're not stable until all the cards are at least that height. I think the `<=` should be flipped to `>=`.

The `currentConstraint > 0` check also means that we'll never "settle" if all the cards have zero height (which might be the case if, e.g. they are all empty). To fix this, I think we need to keep track of whether we have ever received a measurement (via `onMeasureLeft` or `onMeasureRight`) — if we have, then the constraint is valid.

packages/perseus/src/widgets/sorter/sorter.tsx
```diff
@@ -53,12 +56,30 @@ class Sorter extends React.Component<Props> implements Widget {
...
         this._isMounted = false;
+        // Settle on unmount so a tear-down mid-cascade doesn't block the
+        // renderer's onRendered indefinitely.
+        this._setAssetStatus(true);
     }
 
+    onMeasure: () => void = () => {
+        // Sorter has no constraint distribution, so Sortable's first
+        // measurement is authoritative. Settle on each call (idempotent);
+        // the renderer's _fullyRendered is a one-way latch.
```
benchristel: What is a "constraint distribution"? Claude really seems to love making up random technical-sounding terms.

This component cannot know that `setAssetStatus` ultimately sets the Renderer's `_fullyRendered` field. It might not even be rendered by a Renderer. The requirement that `setAssetStatus` be idempotent is part of the behavioral contract of this component, and should be documented in this component's Props type.


```suggestion
        // `setAssetStatus` is assumed to be idempotent.
```

benchristel (commented):
This is getting closer! I have some concerns about the logic and the tests, but it's great that you were able to get rid of the timers.

I'd be happy to pair on refactoring the Matcher measurement code to a state machine that can be tested without React.

packages/perseus/src/components/__tests__/with-asset-context.test.tsx
Myranae: Updated!

packages/perseus/src/widgets/sorter/sorter.tsx
Myranae: Updated the comments based on your comments here. Thanks for reading through these!

packages/perseus/src/widgets/matcher/matcher.tsx
Myranae: Pulled out to a method and uses Math.max now :)

packages/perseus/src/widgets/matcher/matcher.tsx
Myranae: Done!

Myranae:
Closing this for now as it may require too much refactoring for widgets that are already in process of being redone. Will increase the delay in the original story and see if that helps in the meantime.

PR #3695, opened 2026-05-30
[Explanation] Add more visual regression tests

mark-fitzgerald:
## Summary:
This PR expands the visual regression test coverage for the Explanation widget,
adding stories that cover embedded video, images, and data tables in explanation
content. It also adds interaction regression stories for hints renderer behavior,
and moves nested widget tests into a dedicated "Nested Widgets" folder to
centralize tests for widgets that contain other widgets.

Issue: LEMS-4130

## Test plan:
1. Open Storybook and navigate to the Explanation widget stories.
   - Verify the new initial-state regression stories render correctly: content
     with images, embedded video, and tables should display without visual
     artifacts.
2. Navigate to the new Nested Widgets folder and verify nested widget stories
   render correctly for both initial state and interactions.
3. Navigate to the Hints Renderer stories and verify the new interactions
   regression story works as expected.

### DIFF
--- a/packages/perseus-core/src/utils/generators/explanation-widget-generator.ts
+++ b/packages/perseus-core/src/utils/generators/explanation-widget-generator.ts
@@ -22,6 +22,9 @@ export function generateExplanationWidget(
         // Explanations are not graded
         graded: false,
         version: {major: 0, minor: 0},
+        // NOTE: The explanation widget doesn't consume this directly,
+        // instead, Perseus renders an overlay <div /> over top of the
+        // widget that intercepts interactions to it.
         static: false,
         alignment: "default",
         options: generateExplanationOptions(),
--- a/packages/perseus-core/src/utils/generators/explanation-widget-generator.ts
+++ b/packages/perseus-core/src/utils/generators/explanation-widget-generator.ts
@@ -22,9 +22,6 @@ export function generateExplanationWidget(
         // Explanations are not graded
         graded: false,
         version: {major: 0, minor: 0},
-        // NOTE: The explanation widget doesn't consume this directly,
-        // instead, Perseus renders an overlay <div /> over top of the
-        // widget that intercepts interactions to it.
         static: false,
         alignment: "default",
         options: generateExplanationOptions(),
--- a/packages/perseus/src/__docs__/hints-renderer-initial-state-regression.stories.tsx
+++ b/packages/perseus/src/__docs__/hints-renderer-initial-state-regression.stories.tsx
@@ -17,32 +17,42 @@ import {
 import {View} from "@khanacademy/wonder-blocks-core";
 import React from "react";
 
+import {themeModes} from "../../../../.storybook/modes";
 import HintsRenderer from "../hints-renderer";
 import {ApiOptions} from "../perseus-api";
 import {storybookDependenciesV2} from "../testing/test-dependencies";
+import {ipsumExample} from "../widgets/explanation/explanation.testdata";
 import {earthMoonImage} from "../widgets/image/utils";
+
+import {bibliotronExerciseDecorator} from "./hints-renderer-decorator";
 
 import type {Meta, StoryObj} from "@storybook/react-vite";
 
 const defaultApiOptions = ApiOptions.defaults;
 
 const meta: Meta<typeof HintsRenderer> = {
-    title: "Renderers/Hints Renderer",
+    title: "Renderers/Hints Renderer/Visual Regression Tests/Initial State",
     component: HintsRenderer,
+    tags: ["!autodocs", "!manifest"],
     decorators: [
         (Story) => {
             return (
-                <View style={{left: 80}}>
+                <View style={{paddingLeft: 80}}>
                     <Story />
                 </View>
             );
         },
     ],
-    argTypes: {
-        hintsVisible: {
-            control: {min: 0},
-            defaultValue: 3,
+    parameters: {
+        docs: {
+            description: {
+                component:
+                    "Regression tests for the Hints renderer that do NOT " +
+                    "need any interactions to test, which will be used with " +
+                    "Chromatic.",
+            },
         },
+        chromatic: {disableSnapshot: false, modes: themeModes},
     },
 };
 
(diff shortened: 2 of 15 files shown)


packages/perseus-core/src/utils/generators/explanation-widget-generator.ts
```diff
@@ -22,9 +22,6 @@ export function generateExplanationWidget(
         // Explanations are not graded
         graded: false,
         version: {major: 0, minor: 0},
-        // NOTE: The explanation widget doesn't consume this directly,
-        // instead, Perseus renders an overlay <div /> over top of the
-        // widget that intercepts interactions to it.
```
Myranae, line 27: So this comment is no longer true then?

packages/perseus/src/__docs__/hints-renderer-initial-state-regression.stories.tsx
Myranae, line 1: Claude flagged this and it caught my attention. This used to be a regular story, but now it's a regression story. Adding the `manifest` tag to a story keeps it out of the documentation for Perseus. We may want to have a separate file for the regression stories so we can have a regression one and one that adds to the documentation. Or, we may want to take off the manifest related one if we want this to be in the documentation. Not sure if there are other concerns for swapping a regular stories for a regression stories. Claude flagged the !dev to !autodocs, but I think we've already discussed that change and want it this way for regression stories.

packages/perseus-core/src/utils/generators/explanation-widget-generator.ts
mark-fitzgerald: Good point. I think I misunderstood what this property was and how it was/wasn't used. I'll put it back.

packages/perseus/src/__docs__/hints-renderer-initial-state-regression.stories.tsx
mark-fitzgerald: Good point. I forgot about the documentation aspect that these stories provide. I've added that back in, though with just one of the stories. I think the other stories that were in that file are better placed in the regression test stories, but let me know if you think otherwise.

nishasy, line 1: ~~How did you pick interactive graph, image, and explanation to be the widgets to test within the hints renderer?~~

I see you took these stories from the existing hints renderer stories. Are those the extent of what needs to be tested for Hints Renderer? I'm wondering if there are more widgets we should add later, or if it doesn't really matter...

packages/perseus/src/widgets/__docs__/nested-widgets-initial-state-regression.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+                component:
+                    "Regression tests for widgets nested inside other widgets that do NOT " +
+                    "need any interactions to test, which will be used with " +
+                    "Chromatic. Stories are all displayed on one page.",
+            },
+        },
+        chromatic: {disableSnapshot: false, modes: themeModes},
+        controls: {disable: true},
+    },
+};
```
nishasy, line 24: This one makes sense not to have a Meta type or component.

packages/perseus/src/widgets/explanation/__docs__/explanation-initial-state-regression.stories.tsx
```diff
@@ -1,19 +1,18 @@
...
+
+import type {PerseusExplanationWidgetOptions} from "@khanacademy/perseus-core";
 import type {Meta, StoryObj} from "@storybook/react-vite";
 
 /**
  * This is a visual regression story for the Explanation widget.
  */
 
-const meta: Meta = {
+const meta: Meta<PerseusExplanationWidgetOptions> = {
```
nishasy, line 13: nice

nishasy (approved):
Thanks for adding more regression stories!

packages/perseus/src/widgets/__docs__/nested-widgets-renderer-decorator.tsx
```diff
@@ -0,0 +1,13 @@
+import * as React from "react";
+
+import ArticleRenderer from "../../article-renderer";
+import {storybookDependenciesV2} from "../../testing/test-dependencies";
+
+export const articleRendererDecorator = (_, {parameters}) => {
```
Myranae: Apparently Claude noted that you can type the parameters value here with StoryContext. Not sure I've seen this before, but sharing in case you want to explore. Looks like it would also be relevant in `explanation-renderer-decorator.tsx`

<details>
<summary>Details</summary>
Files: nested-widgets-renderer-decorator.tsx, explanation-renderer-decorator.tsx

export const articleRendererDecorator = (_, {parameters}) => { ... };
export const explanationRendererDecorator = (_, {args, parameters}) => { ... };

Both destructure parameters (and args) from an untyped second argument. In Storybook 8 with Vite, these can be typed against StoryContext. Without types, errors in parameter names (e.g., parameters.quesion instead of parameters.question) fail silently at runtime. At minimum:

import type {StoryContext} from "@storybook/react-vite";

export const articleRendererDecorator = (_: React.ReactNode, {parameters}: StoryContext) => { ... };
</details>

mark-fitzgerald: Done.

packages/perseus/src/widgets/__docs__/nested-widgets-interactions-regression.stories.tsx
```diff
@@ -0,0 +1,90 @@
...
+    },
+    play: async ({canvas, userEvent}) => {
+        const explanationTrigger = canvas.getByRole("button", {
+            name: "Show explanation",
+        });
+        await userEvent.click(explanationTrigger);
+    },
+};
+
+const videoExample = videoInContent.widgets["explanation 1"]?.options;
```
Myranae: Claude pointed out a potential "null-unsafe access on optionally-chained widget options" here, though it's in a couple other files too. It sounds like, with this optional chain here, this value could be undefined, but there's no check for undefined where this is used.

> // No null check before property access
args: {
    hidePrompt: videoExample.hidePrompt, // TS error if dict index returns undefined

I think you can remove the optional chaining is the bottom line. In this file and also in these:
- explanation-interactions-regression.stories.tsx
- explanation-initial-state-regression.stories.tsx
- hints-renderer-interactions-regression.stories.tsx

mark-fitzgerald: Removed.

PR #3699, opened 2026-06-01
[AI] Reorganize AI documentation into AGENTS.md files

Myranae:
## Summary

Reorganized the AI assistant development documentation by moving content from `CLAUDE.md` into `AGENTS.md` at appropriate levels of the project hierarchy, with `CLAUDE.md` files now serving as pointers to the relevant `AGENTS.md` documentation.

## Changes

- **Root level (`AGENTS.md`)**: Consolidated general Perseus development guidance including quick start commands, key directories, package dependencies, testing guidelines, and deployment notes
- **`packages/perseus/AGENTS.md`**: Extracted Storybook MCP documentation with setup instructions, available tools, workflow guidance, and Perseus-specific rules for UI development
- **`packages/perseus/src/widgets/AGENTS.md`**: Created widget-specific development guide covering widget creation patterns, implementation details, common issues, and debugging tips
- **`CLAUDE.md` files**: Added at root, `packages/perseus/`, and `packages/perseus/src/widgets/` levels as simple pointers to their respective `AGENTS.md` files

## Implementation Details

This change improves documentation discoverability by:
- Placing documentation closer to the code it describes (widget guide in the widgets directory, Storybook guide in the perseus package)
- Maintaining a consistent naming convention (`AGENTS.md`) for AI assistant documentation
- Using `CLAUDE.md` as a lightweight redirect mechanism for backward compatibility
- Reducing duplication while keeping related guidance together at each level of the project hierarchy

https://claude.ai/code/session_012gTAqjsq2hiFkS3UYL2Pad

**Issue**: [LEMS-4211](https://khanacademy.atlassian.net/browse/LEMS-4211)

[LEMS-4211]: https://khanacademy.atlassian.net/browse/LEMS-4211?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Issue: LEMS-4327


Myranae:
@claude review once

Myranae:
@claude review once

packages/perseus/src/widgets/AGENTS.md
```diff
@@ -0,0 +1,80 @@
...
+## Focus Management
+
+All widgets must implement proper focus management for accessibility.
+
+## Common Issues & Solutions
+
+### Math Rendering (TeX)
+- Use `$...$` for inline math, `$$...$$` for display math
+- For complex expressions, use `\dfrac` instead of `\frac`
+- Test math rendering in different contexts (articles, exercises, hints)
```
benchristel: Widgets shouldn't implement their own math rendering. They should use Renderer to render a Markdown document with TeX support, if that's needed. Most widgets won't need to do this, since math notation should generally be part of the content surrounding the widget.

packages/perseus/src/widgets/AGENTS.md
```diff
@@ -0,0 +1,80 @@
...
+    name: "widget-name",
+    displayName: "Widget Display Name",
+    widget: WidgetComponent,
+    isLintable: true, // For use by the editor
+} as WidgetExports<typeof WidgetComponent>;
+```
+
+## Focus Management
+
+All widgets must implement proper focus management for accessibility.
```
benchristel: This probably needs more detail in order for an AI to do anything useful with it. Maybe a list of the methods that need to be implemented, with references to exemplary code.

packages/perseus/src/widgets/AGENTS.md
```diff
@@ -0,0 +1,80 @@
...
+   - `__docs__/[widget-name].stories.tsx` - Storybook story
+   - `__docs__/a11y.mdx` - Accessibility documentation
+3. **Register widget** in `packages/perseus/src/widgets.ts`
+4. **If scorable, add scoring functions** in `packages/perseus-score/src/widgets/[widget-name]/`:
+   - `score-[widget-name].ts` - Scoring logic
+   - `score-[widget-name].test.ts` - Scoring tests
+   - `validate-[widget-name].ts` - Input validation (optional)
+   - `validate-[widget-name].test.ts` - Validation tests (optional)
+5. **Register scoring** in `packages/perseus-score/src/widgets/widget-registry.ts`
+6. **Add types** to `packages/perseus-core/src/data-schema.ts`
```
benchristel, line 20: We should also add editor components and linters for new widgets.

packages/perseus/src/widgets/AGENTS.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Common Issues & Solutions
+
+### Math Rendering (TeX)
+- Use `$...$` for inline math, `$$...$$` for display math
+- For complex expressions, use `\dfrac` instead of `\frac`
+- Test math rendering in different contexts (articles, exercises, hints)
+
+### Widget State Management
+- Use `useState` for local component state
```
benchristel: `useState` should only be used for state that isn't meaningful user input (that should be scored or validated). User input state should not be stored locally; changes should be emitted via `onChange` and new widget state should be received via the `userInput` prop.

We (I) did this incorrectly in the new interactive-graph widget — it duplicates the state that's passed to and from the UserInputManager. :/

benchristel: Also, `useState` is not an absolute requirement — `useReducer` is also acceptable for local state (though probably overkill in most cases).

packages/perseus/src/widgets/AGENTS.md
```diff
@@ -0,0 +1,80 @@
...
+4. **If scorable, add scoring functions** in `packages/perseus-score/src/widgets/[widget-name]/`:
+   - `score-[widget-name].ts` - Scoring logic
+   - `score-[widget-name].test.ts` - Scoring tests
+   - `validate-[widget-name].ts` - Input validation (optional)
+   - `validate-[widget-name].test.ts` - Validation tests (optional)
+5. **Register scoring** in `packages/perseus-score/src/widgets/widget-registry.ts`
+6. **Add types** to `packages/perseus-core/src/data-schema.ts`
+
+## Widget Implementation Pattern
+
```
benchristel, line 28: Might be worth mentioning that new widgets should be implemented as functional components.

packages/perseus/src/widgets/AGENTS.md
```diff
@@ -0,0 +1,80 @@
...
+## Common Issues & Solutions
+
+### Math Rendering (TeX)
+- Use `$...$` for inline math, `$$...$$` for display math
+- For complex expressions, use `\dfrac` instead of `\frac`
+- Test math rendering in different contexts (articles, exercises, hints)
+
+### Widget State Management
+- Use `useState` for local component state
+- Props flow down from parent renderer
```
benchristel: This item seems to just describe how React works. More specifically,


```suggestion
- The props type of widgets should be derived from `WidgetProps<...>` and the widget options type in data-schema.ts.
```

packages/perseus/src/widgets/AGENTS.md
```diff
@@ -0,0 +1,80 @@
...
+- Call `onChange` to notify parent of state changes
+- Implement proper serialization for persistent state
```
benchristel: Is this referring to the `getSerializedState` method? That's deprecated, and new widgets should not implement it with anything but a trivial `return {}` or something like that.

packages/perseus/src/widgets/AGENTS.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Mobile Considerations
+- All widgets must work on mobile devices
+- Support touch interactions
+- Consider on-screen keypad for math inputs
+- Test with different screen sizes using Storybook
+
```
benchristel, line 54: Is this a good place to list accessibility requirements? Mark had a good doc on that: https://khanacademy.atlassian.net/wiki/spaces/LC/pages/4691919152/Accessibility+Instructions+for+AI

packages/perseus/src/widgets/AGENTS.md
```diff
@@ -0,0 +1,80 @@
...
+### Performance Optimization
+- Use `React.memo()` for expensive components
+- Implement `useMemo()` for complex calculations
+- Avoid unnecessary re-renders in widget hierarchies
+- Profile performance with React DevTools
```
benchristel: Can AI do this reliably? I.e. does it have good judgement about when to optimize and how to benchmark? Memoization can make components _slower_ in many cases.

I would lean towards not encouraging performance optimization.

packages/perseus/src/widgets/AGENTS.md
```diff
@@ -0,0 +1,80 @@
...
+- Test different props combinations
+- Verify accessibility with Storybook a11y addon
+- Check mobile layouts with device frame addon
+
+### Console Debugging
+```typescript
+// Temporary debugging (remove before commit)
+console.log("Widget state:", this.state);
+console.log("Props received:", this.props);
+```
```
benchristel: 😂 I don't think AI needs to be told how to use console.log.

packages/perseus/src/widgets/AGENTS.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Error Boundaries
+- Widgets are wrapped in error boundaries
```
benchristel:
```suggestion
- Widgets are wrapped in error boundaries by their parent component.
```

packages/perseus/src/widgets/AGENTS.md
```diff
@@ -0,0 +1,80 @@
...
+- Check browser console for widget-specific errors
+- Implement graceful fallbacks for failed widgets
```
benchristel: Widgets should not implement fallbacks. The parent component's error boundary catches the error and renders a placeholder icon.
```suggestion
```

packages/perseus/AGENTS.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Setup (first time)
+
+Make sure Storybook is running:
+```bash
+pnpm storybook
+```
+Then register the MCP server with Claude Code:
+```bash
+claude mcp add storybook-mcp --transport http http://localhost:6006/mcp --scope project
```
benchristel: Can claude add new MCP servers? I.e. does it have permission?

In any case, we probably don't want to include this line for non-Claude agents. I might just leave it out and let humans manage their own MCP servers.

packages/perseus/AGENTS.md
```diff
@@ -0,0 +1,48 @@
...
+| `get-storybook-story-instructions`  | Get story authoring guidelines for this project      |
+| `preview-stories`                   | Get live Storybook URLs to visually verify stories   |
+
+## Workflow for UI Work
+
+1. Call `get-storybook-story-instructions` once per session — the guidelines
+   don't change between calls, so there's no need to re-fetch them.
+2. Write or modify the component and its stories.
+3. Call `preview-stories` only after finishing changes, not speculatively
+   during exploration.
```
benchristel: Why don't we want the agent to check its work as it goes?

packages/perseus/AGENTS.md
```diff
@@ -0,0 +1,48 @@
...
+
+| Tool                                | Purpose                                              |
+| ----------------------------------- | ---------------------------------------------------- |
+| `get-storybook-story-instructions`  | Get story authoring guidelines for this project      |
+| `preview-stories`                   | Get live Storybook URLs to visually verify stories   |
+
+## Workflow for UI Work
+
+1. Call `get-storybook-story-instructions` once per session — the guidelines
+   don't change between calls, so there's no need to re-fetch them.
```
benchristel, line 24: Seems like an opportunity to inject these guidelines directly into the context (perhaps via this file?) so the agent _can't_ forget to read them or read them multiple times. I'm not sure why we need MCP for static instructions.

packages/perseus/AGENTS.md
```diff
@@ -0,0 +1,48 @@
...
+2. Write or modify the component and its stories.
+3. Call `preview-stories` only after finishing changes, not speculatively
+   during exploration.
+4. Prefer MCP tool calls over crawling raw source files — reading story
+   conventions via MCP is cheaper than navigating through source directories.
+
+## Perseus-specific Rules
+
+- **Never invent widget props.** Perseus widgets have a strict versioned data
+  schema. Always verify against documented stories before using any prop.
```
benchristel: Does this line fix a misbehavior we've observed? If agents don't tend to hallucinate props, this line might be wasted tokens or might even be confusing.

Addy Osmani suggests only including info in your context when you've observed the AI doing the wrong thing _without_ that info. Can't find the link at the moment unfortunately.

packages/perseus/AGENTS.md
```diff
@@ -0,0 +1,48 @@
...
+- **Use widget generators for test data.** All widget test data must come from
+  `packages/perseus-core/src/utils/generators` — never hand-roll it.
+- **Check for existing stories first.** New stories go in `__docs__/` directories.
+  Some older stories are colocated directly but haven't been migrated yet — avoid
+  duplicating what's already there.
+- **Every new component needs stories** covering each key visual state.
```
benchristel: Maybe:

```suggestion
- **Every new widget needs stories** covering each key visual state.
```

I don't know if we want stories for _every_ component, since that makes it harder to break up one component into several or inline components into a widget.

packages/perseus/AGENTS.md
```diff
@@ -0,0 +1,48 @@
...
+- **Add `!manifest` to non-component stories.** Documentation pages, visual
+  regression tests (`*-regression.stories.tsx`), a11y docs, and dev utilities
+  should include `"!manifest"` in their tags to keep them out of the component
+  manifest. Actual widget/component stories should NOT have this tag.
```
benchristel, line 39: This `!manifest` thing is new to me — can you double-check that this is right?

packages/perseus/AGENTS.md
```diff
@@ -0,0 +1,48 @@
...
+   during exploration.
+4. Prefer MCP tool calls over crawling raw source files — reading story
+   conventions via MCP is cheaper than navigating through source directories.
+
+## Perseus-specific Rules
+
+- **Never invent widget props.** Perseus widgets have a strict versioned data
+  schema. Always verify against documented stories before using any prop.
+- **Use widget generators for test data.** All widget test data must come from
+  `packages/perseus-core/src/utils/generators` — never hand-roll it.
```
benchristel: Might be worth mentioning that values that are **important to the test** (used in assertions, or in the specific logic being tested) should be explicitly passed as params to the widget generator functions. We don't want test assertions to be coupled to shared setup, since that makes it hard to understand what the test is doing.

Unfortunately, Claude is not great at following that guideline IME, but it probably doesn't hurt to put it in.

packages/perseus/AGENTS.md
```diff
@@ -0,0 +1,48 @@
...
+- **Check for existing stories first.** New stories go in `__docs__/` directories.
+  Some older stories are colocated directly but haven't been migrated yet — avoid
+  duplicating what's already there.
```
benchristel, line 34: Is this true? I thought we were moving away from `__docs__` and `__tests__` directories and putting everything next to the component files, but maybe I'm misremembering.

AGENTS.md
```diff
@@ -0,0 +1,136 @@
+# Perseus Development Guide
+
+Perseus is Khan Academy's educational content rendering system that powers all exercises and articles. It's a TypeScript monorepo that extends Markdown with interactive widgets and beautiful math rendering.
+
+**Core Architecture:**
+- **Renderers**: Display content (ServerItemRenderer for exercises, ArticleRenderer for articles)
+- **Widgets**: Interactive components (radio, numeric-input, interactive-graph, etc.)
+- **Editors**: Authoring interfaces for content creators
+- **Math**: TeX expressions rendered via MathJax
```
benchristel: Probably don't need to mention math here, since it's handled by the renderers automatically.

benchristel: We probably **do** want to mention scoring, validation, the data schema, and the parsers, since those are all significant architectural components. Maybe the editor linters too.

AGENTS.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Quick Start Commands
+
+### Development
+```bash
+pnpm storybook             # Launch Storybook documentation
+pnpm test                  # Run tests
```
benchristel: Agents should run `pnpm tesc` and `pnpm fixc` to speed up testing and autofixing of lint errors.
```suggestion
pnpm tesc                  # Run tests for changed files
pnpm fixc                  # Run linter and autofix formatting
```

AGENTS.md
```diff
@@ -0,0 +1,136 @@
...
+pnpm tsc                   # Type-check all packages
+pnpm knip                  # Find unused files, exports, and dependencies
+```
+
+### Testing Specific Packages
+```bash
+pnpm --filter perseus test                   # Test main perseus package
+pnpm --filter perseus-editor test            # Test editor package
+pnpm test packages/perseus/src/widgets/radio # Test specific widget
+```
```
benchristel: This info is generic pnpm/jest stuff that AI almost certainly knows already. I'd probably take it out.

AGENTS.md
```diff
@@ -0,0 +1,136 @@
...
+- Import order: builtin > external > internal > relative > types
+
+### Example Correct Imports
+```typescript
+import React from "react";                           // external
+import {ApiOptions} from "@khanacademy/perseus";     // internal package
+import {WidgetContainer} from "../widget-container"; // relative
+
+import type {WidgetProps} from "@khanacademy/perseus-core";
+```
```
benchristel: We lint for all this stuff so I don't think the AI needs to be told.

packages/perseus/AGENTS.md
benchristel: Info about generators is also in the root AGENTS.md file, so maybe this section should be removed and my suggested changes should go in the root file.

AGENTS.md
```diff
@@ -0,0 +1,136 @@
...
+        render(<WidgetComponent {...question1} onChange={onChange} />);
+        await user.click(screen.getByRole("button"));
+
+        expect(onChange).toHaveBeenCalled();
+    });
+});
+```
+
+### Writing Tests
+- use `it` for individual test cases and not `test`
```
benchristel: I'm not sure this is a rule? The reason Jest has both is to make tests read like an English sentence:

```js
test("input validation")
it("escapes backslashes")
```

AGENTS.md
```diff
@@ -0,0 +1,136 @@
...
+- use `describe` to group related tests
```
benchristel: I'm also not sure we want to encourage this. If the test has many `describe`s, it's usually a sign that the code being tested is doing too much and should be split up into smaller, individually tested pieces.

AGENTS.md
```diff
@@ -0,0 +1,136 @@
...
+- Test titles should describe the requirement or observable outcome, not the implementation.
+  Prefer verbs like `returns`, `renders`, `disables`, `throws` over vague phrases like "should handle".
```
benchristel, line 85: ❤️

AGENTS.md
```diff
@@ -0,0 +1,136 @@
...
+3. Lint and format: `pnpm lint --fix && pnpm prettier . --write`
+4. Test in Storybook: `pnpm storybook`
+5. Verify accessibility compliance
+
+### Common Pre-commit Failures
+- ESLint errors (unused imports, console statements)
+- Prettier formatting (spacing, quotes, semicolons)
+- TypeScript type errors
+- Missing accessibility documentation
+- Test failures
```
benchristel: Probably don't need to include this?
```suggestion
```

AGENTS.md
```diff
@@ -0,0 +1,136 @@
...
+- Test titles should describe the requirement or observable outcome, not the implementation.
+  Prefer verbs like `returns`, `renders`, `disables`, `throws` over vague phrases like "should handle".
+  A failing test title should tell you which requirement broke without reading the test body.
+  ❌ `"should handle empty input"` → ✅ `"returns null when input is empty"`
+
+## Deployment Notes
+
+### Before Submitting PR
+1. Run full test suite: `pnpm test`
+2. Check types: `pnpm build:types`
```
benchristel: I don't think building the types is necessary. `pnpm tsc` is enough.

AGENTS.md
```diff
@@ -0,0 +1,136 @@
...
+3. Lint and format: `pnpm lint --fix && pnpm prettier . --write`
+4. Test in Storybook: `pnpm storybook`
+5. Verify accessibility compliance
```
benchristel, line 110: I think the AI needs more information to do a good job with a11y... see Mark's doc: https://khanacademy.atlassian.net/wiki/spaces/LC/pages/4691919152/Accessibility+Instructions+for+AI

AGENTS.md
```diff
@@ -0,0 +1,136 @@
...
+
+- **Storybook**: Interactive component documentation and testing
+- **Perseus Architecture**: See `__docs__/introduction.mdx` for detailed overview
+- **Widget Gallery**: Browse existing widgets in Storybook for patterns
+- **Accessibility Guidelines**: Each widget should have `a11y.mdx` documentation
+- **Khan Academy Design System**: Wonder Blocks components for consistent UI
+
+---
+
+*This document is maintained for AI assistants. For human developers, see the main README.md files in each package.*
```
benchristel: Once thing I don't see here is instructions about how to format commit messages. By default, Claude's messages are very verbose, full of self-praise, and duplicate information that's otherwise available via Git or in the PR UI.

A commit message should contain only what's likely to be useful to someone 5 years in the future:

- What behavior changed
- Why it was changed (link to Jira and background info — though note, it's unlikely that the AI has all the context!)
- Links to any slack convos or relevant docs (again, unlikely that the AI has this)
- Notes for anything surprising (e.g. "We aren't using the feature flag yet; it will be wired up in a future PR.")
- A test plan

benchristel (approved):
Nice! I think it's fine if this lands as-is, but I also left a lot of suggestions inline for future improvements.

Myranae:
@benchristel Made the easy fixes. Here's a summary of those. Going to review the ones that need a bit more review and might do a couple or might create a follow-up ticket for them so your suggestions don't get lost :) Once that's done, I'll go ahead and land this.

> Root AGENTS.md — Commands now use pnpm tesc/pnpm fixc; removed the per-package test examples, pre-commit failures list, pnpm build:types, and Additional Resources section.
> 
> [packages/perseus/AGENTS.md](https://github.com/Khan/perseus/blob/claude/compassionate-dirac-3d8Xc/packages/perseus/AGENTS.md) — Removed the claude mcp add line (tool-specific); "every new component" → "every new widget".
> 
> [packages/perseus/src/widgets/AGENTS.md](https://github.com/Khan/perseus/blob/claude/compassionate-dirac-3d8Xc/packages/perseus/src/widgets/AGENTS.md) — Added functional components requirement; removed math rendering section; rewrote widget state guidance with the onChange/userInput pattern and WidgetProps<...> props type note; removed performance optimization and console.log sections; simplified error boundaries.

packages/perseus/src/widgets/AGENTS.md
mark-fitzgerald: I agree that this needs more detail. I'm wary of the statement as is because past code showed a heavy-hand in regards to focus management that caused various UI problems (and actually interfered with accessibility). I would prefer not to bring this up at all unless there is something universal about focus management that we want to call out.

packages/perseus/src/widgets/AGENTS.md
```diff
@@ -0,0 +1,62 @@
...
+- Consider on-screen keypad for math inputs
+- Test with different screen sizes using Storybook
+
+## Debugging Tips
+
+### Storybook Development
+- Use Storybook for isolated component development
+- Test different props combinations
+- Verify accessibility with Storybook a11y addon
+- Check mobile layouts with device frame addon
```
mark-fitzgerald, line 61: Should we also call out the need for visual regression stories?

AGENTS.md
mark-fitzgerald: I have an instructions document that is specifically focused on accessibility that we could add as a reference document.

packages/perseus/src/widgets/AGENTS.md
```diff
@@ -0,0 +1,62 @@
...
+Widgets must be implemented as **functional components**.
+
+```typescript
+// Export interface following WidgetExports<T> pattern
+export default {
+    name: "widget-name",
+    displayName: "Widget Display Name",
+    widget: WidgetComponent,
+    isLintable: true, // For use by the editor
+} as WidgetExports<typeof WidgetComponent>;
```
jeremywiebe: Can we make this `satisfies WidgetExports..` instead?

packages/perseus/src/widgets/AGENTS.md
```diff
@@ -0,0 +1,62 @@
...
+```
```
jeremywiebe, line 39: This feels confusing.  We mention that widgets must be functional components, but the example here is basically the widget export definition.  Maybe we should say:

> Widgets are defined by creating an object the that conforms to the `WidgetExports<T>` interface. The widget component (`widget` key in the export) must be implemented as **functional components**.

packages/perseus/src/widgets/AGENTS.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Focus Management
+
+All widgets must implement proper focus management for accessibility.
+
+## Common Issues & Solutions
+
+### Widget State Management
+- Use `useState` or `useReducer` for local UI state that doesn't represent user input
+- User input state must not be stored locally — emit changes via `onChange` and receive state via the `userInput` prop
```
jeremywiebe: I thought we propagated user input up through the `handleUserImput` prop.  Eg: https://github.com/Khan/perseus/blob/d41078308789fa63a3b815a2e0cc59eece986ede/packages/perseus/src/widgets/sorter/sorter.tsx#L67

packages/perseus/src/widgets/AGENTS.md
```diff
@@ -0,0 +1,62 @@
...
+- The props type should be derived from `WidgetProps<...>` and the widget options type in `data-schema.ts`
+
+### Mobile Considerations
+- All widgets must work on mobile devices
+- Support touch interactions
+- Consider on-screen keypad for math inputs
+- Test with different screen sizes using Storybook
+
+## Debugging Tips
```
jeremywiebe: Should we delete this if we don't actually provide any tips?

packages/perseus/src/widgets/AGENTS.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Storybook Development
+- Use Storybook for isolated component development
+- Test different props combinations
+- Verify accessibility with Storybook a11y addon
+- Check mobile layouts with device frame addon
+
+### Error Boundaries
+- Widgets are wrapped in error boundaries by their parent component
```
jeremywiebe:
```suggestion
- Widgets are wrapped in error boundaries by their parent component (`WidgetContainer`)
```

AGENTS.md
```diff
@@ -0,0 +1,107 @@
+# Perseus Development Guide
```
jeremywiebe, line 1: Alot of this feels duplicated from the root README.md.  If this is good for the AI, it's likely good for humans.  Is it possible to just point this file at the root README to avoid duplicating the content so much?

AGENTS.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Development
+```bash
+pnpm storybook             # Launch Storybook documentation
+pnpm tesc                  # Run tests for changed files
+```
+
+### Code Quality
+```bash
+pnpm fixc                  # Run linter and autofix formatting
```
jeremywiebe, line 25: Is it fixing formatting or fixable linting issues? I thought the latter.

jeremywiebe: That said, we could mention `pnpm format` here which re-formats all files. That feels useful for AI generated code.

jeremywiebe: Oh boy, I should read the code.  `lint.sh` does invoke `prettier`.  So just ignore all this dithering.

jeremywiebe (commented):
Some drive-by comments, nothing blocking!

packages/perseus/src/widgets/AGENTS.md
Myranae: Removed this section!

packages/perseus/src/widgets/AGENTS.md
Myranae: Removed!

packages/perseus/src/widgets/AGENTS.md
Myranae: Added!

packages/perseus/src/widgets/AGENTS.md
Myranae: Updated!

packages/perseus/src/widgets/AGENTS.md
Myranae: Added!

packages/perseus/src/widgets/AGENTS.md
Myranae: Updated!

packages/perseus/src/widgets/AGENTS.md
Myranae: Added a comment about this

AGENTS.md
Myranae: Removed~

AGENTS.md
Myranae: @mark-fitzgerald , is that separate from what Ben linked above?

packages/perseus/src/widgets/AGENTS.md
Myranae: Removed~

packages/perseus/src/widgets/AGENTS.md
Myranae: Removed~

packages/perseus/AGENTS.md
Myranae: Removed~

packages/perseus/AGENTS.md
Myranae: These were all pre-existing instructions, so not sure 😅 Claude says to avoid spamming storybook maybe? All agree it's badly worded. Changed to: "Call `preview-stories` after making changes to verify your work visually."

packages/perseus/AGENTS.md
Myranae: @benchristel The only thing I worry about is missing out on updates to the instructions. Not sure we would get any indicator as I'm not quite sure how MCPs get updated over time. Claude checked the info from the call and found it was already trained on most of it except two import related changes, but if those change or anything else changes, then we'd be out-of-date. I think I'd lean towards keeping it or just adding the import related changes that Claude found were specific to Perseus. What do you think?

packages/perseus/AGENTS.md
Myranae: I didn't write these, so I'm not sure where this came from or why it was added. I'm cool with removing it for now and just updating these instructions as issues are discovered. Removed~

packages/perseus/AGENTS.md
Myranae: We've started adding this to all of our visual regression stories during color sync. We use `!autodocs` and `!manifest` so that the stories don't get included as documentation, either for humans (autodocs) or for AI (manifests). I think there have been some discussions in Slack about these tags too. Here's the [Storybook Manifests documentation](https://storybook.js.org/docs/ai/manifests). I think the thought is that these types of stories are not meant for documentation but for Chromatic snapshots, so we don't include them.

packages/perseus/AGENTS.md
Myranae: We've been using `__docs__` folders within the widget folders during Color Sync. So, for example, we have the numeric-input folder with a child `__docs__` folder containing all the stories. Not all widgets are like that, but that's what we've been doing.

packages/perseus/AGENTS.md
Myranae: Updated!

AGENTS.md
Myranae: Updated!

AGENTS.md
Myranae: Removed~

AGENTS.md
Myranae: Removed~

AGENTS.md
Myranae: Removed~

AGENTS.md
Myranae: Wrote up some [commit message guidance](https://github.com/Khan/perseus/pull/3699/changes#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9R88).

packages/perseus/src/widgets/AGENTS.md
Myranae: Called them out here and also in the new widget checklist :)

packages/perseus/src/widgets/AGENTS.md
Myranae: Updated!

packages/perseus/src/widgets/AGENTS.md
Myranae: Updated!

packages/perseus/src/widgets/AGENTS.md
Myranae: Updated to reference handleuserInput!

packages/perseus/src/widgets/AGENTS.md
Myranae: Changed to Development tips. Added some content during this PR review phase :)

packages/perseus/src/widgets/AGENTS.md
Myranae: Added!

AGENTS.md
Myranae: I think that's a possibility at some point, but I have a preference for keeping them separate as I feel like instructions cut down to what the AI might not know compared to its training might be shorter/more efficient than what we would want to share with human users, although I guess we assume they have some level of training too. Also this point from Claude makes me still want to keep them separate:

> **Automatic loading**: Claude Code injects AGENTS.md into the AI's context before it starts working. If it just said "see README.md," the AI would have to read that file separately — and might not prioritize the right parts.

I think symlinking might work better than saying "see README", so it might not be an issue, but I like being able to streamline the AI instructions as much as possible to cut down on context usage. Totally cool if this direction changes in the future or if I get overruled here though.

packages/perseus/src/widgets/AGENTS.md
Myranae: Added it!

AGENTS.md
Myranae: Added his doc!

jeremywiebe (approved):
Thanks for re-working all these docs

AGENTS.md
```diff
@@ -100,37 +80,30 @@ describe("WidgetComponent", () => {
...
-5. Verify accessibility compliance
+When doing any UI work, follow the guidelines in `accessibility-instructions.md` (at the repo root).
 
-### Common Pre-commit Failures
-- ESLint errors (unused imports, console statements)
-- Prettier formatting (spacing, quotes, semicolons)
-- TypeScript type errors
-- Missing accessibility documentation
-- Test failures
+@accessibility-instructions.md
```
benchristel, line 93: Question: I haven't seen this `@` syntax before. Does it have a special meaning to AI agents?

EDIT: Claude says

> The @ prefix is an import directive: it tells Claude Code (or other agentic CLIs that read AGENTS.md/CLAUDE.md) to inline the contents of the referenced file into context when the parent file is loaded.

Interesting that Claude thinks it reads AGENTS.md...

in any case, I don't know if we want to necessarily have the instructions inlined into this file, since they're kind of long. But I don't know if it matters. At this point, I feel like I don't have the information to make good decisions about what should go in this file. I'd like to just try something and see how it works out. So I'm happy to accept this PR as-is.

packages/perseus/AGENTS.md
benchristel: I don't know for sure. I get nervous whenever there are multiple sources of the same information that can get out of sync. It's even worse when we can't even tell which one the AI is using.

My first choice would be to add the import-related information to AGENTS.md and remove the MCP instructions. MCP is opaque to me, but if the info is in this file I know how to review and edit it.

My second choice would be to remove all Storybook instructions from AGENTS.md that are duplicated on the MCP server, and rely only on MCP for Storybook info.

I would also be fine to leave this as-is for now and revisit it if we run into problems.

benchristel (approved):
Thanks for the updates! It all looks good. I left some nonblocking comments inline.

PR #3701, opened 2026-06-01
[ColorSync] Migrate Sortable component colors and font-size to semantic tokens

Myranae:
## Summary:
Part of the ongoing ColorSync migration to align Perseus components with the Wonder Blocks semantic token system. Sortable is used by both the Sorter and Matcher widgets (which were migrated in earlier PRs), so this closes the remaining gap in that widget family. Adds Chromatic regression coverage for the initial, dragging, and placeholder visual states, which had no prior snapshot coverage.

The snapshot files for Sorter and Matcher are also updated in this PR. Aphrodite generates CSS class names by hashing the style object content — replacing hardcoded hex values with token references changes the hash, so any test that snapshots rendered Sortable markup (including Sorter and Matcher tests, which render Sortable internally) needed regeneration. No behavioral change.

[Figma Designs for Sorter](https://www.figma.com/design/HlLQJqNeMTLenuDfkyzYzE/Perseus-Widgets?node-id=2469-4152&m=dev) 
(which uses Sortable)

Issue: LEMS-4099

## Test plan:
- [x] Chromatic diff approved — visual token changes look correct in both SYL Light and ThunderBlocks themes
- [x] `PlaceholderVisible` story in Storybook shows the gray placeholder slot exposed alongside the warm dragging card
- [x] `DraggingCard` story shows the dragging background without card movement (baseline for the held state)
- [x] `pnpm test`, `pnpm tsc`, `pnpm lint` all pass


.claude/plans/sortable-font-color-progress-report.md
Myranae, line 1: A check in place will prevent this from landing, so it will get removed before finalizing this work. The failing check is because of this file.

Myranae:
@claude review once

packages/perseus/src/components/__docs__/sortable-initial-state-regression.stories.tsx
```diff
@@ -0,0 +1,49 @@
...
+    },
+};
+export default meta;
+
+type Story = StoryObj<typeof Sortable>;
+
+const defaultOptions = ["Apple", "Banana", "Cherry"];
+
+// Verifies the default card appearance in horizontal layout: white background
+// and light gray border (card style).
```
Myranae: Not sure if these "Verifies..." comments are helpful. I think I'm leaning towards removing them. Does anyone think they are good to keep?

Myranae:
Before landing this, I need to reach out to stakeholders to let them know the dragging color is changing to blue for Sorter and Matcher!

packages/perseus/src/components/__docs__/sortable-initial-state-regression.stories.tsx
Myranae: Removed them :D

packages/perseus/src/components/__docs__/sortable-interactions-regression.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+    play: async ({canvasElement, userEvent}) => {
+        const canvas = within(canvasElement);
+        const cards = canvas.getAllByRole("listitem");
+        // Press and hold to trigger the DRAGGING state without releasing
+        await userEvent.pointer({target: cards[0], keys: "[MouseLeft>]"});
+        // requestAnimationFrame inside onMouseDown fires asynchronously;
+        // wait for the placeholder to appear before Chromatic snapshots.
+        await waitFor(() =>
+            expect(canvas.getAllByRole("listitem")).toHaveLength(4),
+        );
```
mark-fitzgerald, line 40: Are these steps a consistent method of producing a visual snapshot with a mouse-down or `:active` pseudo-state in place? If so, I'm wondering if there is a way to make a util function for use by (all?) widgets?

packages/perseus/src/components/__docs__/sortable-interactions-regression.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+        const cards = canvas.getAllByRole("listitem");
+        const cardRect = cards[0].getBoundingClientRect();
+        // Press and hold to trigger the DRAGGING state
+        await userEvent.pointer({target: cards[0], keys: "[MouseLeft>]"});
+        // Wait for the placeholder to appear in the DOM
+        await waitFor(() =>
+            expect(canvas.getAllByRole("listitem")).toHaveLength(4),
+        );
+        // Sortable's Draggable binds mousemove on document directly; userEvent
+        // does not dispatch events that reach document-level native listeners.
```
mark-fitzgerald: Nit: This is totally your call, but I had trouble understanding this comment. Perhaps something like:
```
// To move the sortable card, we need to "fire" a "mouseMove" event on the "document" element
// because the sortable's event listener is registered on the "document" instead of the card.
// Also, events dispatched via "userEvent" don't bubble up to the "document".
```

packages/perseus/src/components/sortable.tsx
```diff
@@ -917,8 +918,8 @@ const styles = StyleSheet.create({
 
     card: {
         boxSizing: "border-box",
-        background: "#fff",
-        border: "1px solid #ddd",
+        background: semanticColor.core.background.base.default,
+        border: `1px solid ${semanticColor.core.border.neutral.subtle}`,
```
mark-fitzgerald: I know it isn't our focus (color & fonts), but perhaps we could also update these borders:
```
border: `${border.width.thin} solid ${semanticColor.core.border.neutral.subtle}`,
```

mark-fitzgerald (approved):

packages/perseus/src/components/__docs__/sortable-interactions-regression.stories.tsx
Myranae: Thank you so much! Just used your suggested wording :)

packages/perseus/src/components/sortable.tsx
Myranae: Thanks for pointing this out! Made me notice the transparent color I'd left behind one one of the borders below this 😅 Updated!

packages/perseus/src/components/__docs__/sortable-interactions-regression.stories.tsx
Myranae: Added a [function called "startDrag"](https://github.com/Khan/perseus/pull/3701/changes#diff-1b824d90ab3712e00c29708d45bfb98008ae8b4bccbd92332a7a5b6120610910) to be available to all packages (that's why it's in .Storybook instead of lower). Would love your thoughts!

.storybook/play-utils.ts
```diff
@@ -0,0 +1,11 @@
+import type {UserEvent} from "@testing-library/user-event";
+
+// "[MouseLeft>]" holds the button down without releasing — triggers drag-start
+// handlers and the CSS :active pseudo-state. The ">" is userEvent pointer syntax
+// for "press without release".
+export async function startDrag(
```
mark-fitzgerald: I realize this is currently intended to initiate a drag, but technically it is a click-and-hold action. Would you be open to calling this function something like `clickAndHold` or `mouseDown`?

Myranae: Updated! Did `mouseDown` :)

nishasy (approved):
LGTM! I'm glad we're doing the drag and drop project, because these are so inaccessible and don't have clear styling for different interaction states 😵‍💫

packages/perseus/src/components/__docs__/sortable-interactions-regression.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+export const DraggingCard = {
+    args: {
+        layout: "horizontal",
+        options: ["Apple", "Banana", "Cherry"],
+        waitForTexRendererToLoad: false,
+        onMeasure: fn(),
+    },
+    play: async ({canvasElement, userEvent}) => {
+        const canvas = within(canvasElement);
```
nishasy: I don't think you need the `within` anymore. You can just use `canvas` in the parameters and at the call site.

Myranae: Updated!

PR #3702, opened 2026-06-01
[Definitions] Added a label to define with the definition widget button does for screenreaders

Evelas78:
## Summary:
[LEMS-4104/the-purpose-of-clickable-definition-buttons-in-question-text-is-unclear] Added a label to define what the definition widget button does for screenreaders

Issue: LEMS-4104

## Test plan:
- Go to definition widget and use the voiceover tool to see if the button is read out as "Definition of: {Word}"

### DIFF
--- a/packages/perseus/src/widgets/definition/definition.tsx
+++ b/packages/perseus/src/widgets/definition/definition.tsx
@@ -91,6 +91,7 @@ class Definition extends React.Component<DefinitionProps> implements Widget {
                                 this.props.trackInteraction();
                                 setActiveDefinitionId(this.props.widgetId);
                             }}
+                            aria-label={`Definition of: ${this.props.togglePrompt}`}
                         >
                             {() => (
                                 <span className={styles.definition}>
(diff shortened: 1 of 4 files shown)


ivyolamit (approved):
LGTM 👍🏼

nishasy:
Nice!!

packages/perseus/src/widgets/definition/definition.tsx
```diff
@@ -91,6 +91,7 @@ class Definition extends React.Component<DefinitionProps> implements Widget {
                                 this.props.trackInteraction();
                                 setActiveDefinitionId(this.props.widgetId);
                             }}
+                            aria-label={`Definition of: ${this.props.togglePrompt}`}
```
nishasy, line 94: Hey! I know you already merged this, but I think you might have to revisit this issue. We put our strings in `strings.js`. I believe the best practice for this would be to add `"Definition of: "` as a new string and then do a concatenation here. That way "Definition of" and the toggle prompt both get translated.

nishasy: After looking into this some more, it seems like string interpolation should be fine, but you still need to add it as a function within `strings.ts`.

You can use this as an example: https://github.com/Khan/perseus/blob/main/packages/perseus/src/strings.ts#L40

PR #3703, opened 2026-06-01
[ColorSync][Expression] Convert expression fonts and colors to semantic tokens

Myranae:
## Summary:
Converts the hardcoded `backgroundColor: "white"` in the mobile `KeypadInput` component (`math-input` package) to `semanticColor.core.background.base.default`. Adds visual regression stories for the expression widget covering desktop and mobile input states, focus states, keypad tabs, hover, and text-in-field. Also adds a `TODO(LEMS-4238)` on the MathQuill imperative `fontSize` DOM style, flagged for re-evaluation when Aphrodite is removed

[Figma Designs for Expression](https://www.figma.com/design/HlLQJqNeMTLenuDfkyzYzE/Perseus-Widgets?node-id=2396-416&m=dev)

Issue: LEMS-3496

## Test plan:
- [x] Chromatic diffs show only intentional color/font changes against the baseline
- [x] All checks pass


Myranae:
@claude review once

packages/perseus/src/widgets/expression/__docs__/expression-initial-state-regression.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+        times: false,
+        extraKeys: [],
+        visibleLabel: "Enter your answer",
+        ariaLabel: "Enter your answer",
+    } satisfies Partial<PerseusExpressionWidgetOptions>,
+};
+
+// Verifies the mobile input (KeypadInputWithInterface / KeypadInput) renders
+// correctly — captures the hardcoded white background and neutral border that
+// need color token migration
```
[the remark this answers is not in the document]
Myranae: Fixed!

packages/perseus/src/widgets/input-number/__snapshots__/input-number.test.ts.snap
```diff
@@ -176,7 +176,7 @@ exports[`rendering supports mobile rendering: mobile render 1`] = `
             >
               <div
                 class="mq-editable-field mq-math-mode"
-                style="background-color: white; min-height: 44px; min-width: 64px; max-width: 128px; box-sizing: border-box; position: relative; border-style: solid; border-radius: 4px; border-width: 1px;"
```
nishasy, line 179: Why would using the token make the background image go away completely from the snapshot?

packages/perseus/src/widgets/expression/__docs__/expression-initial-state-regression.stories.tsx
```diff
@@ -0,0 +1,71 @@
+import {themeModes} from "../../../../../../.storybook/modes";
+import {getWidget} from "../../../widgets";
+
+import {expressionRendererDecorator} from "./expression-renderer-decorator";
+
+import type {PerseusExpressionWidgetOptions} from "@khanacademy/perseus-core";
+import type {Meta, StoryObj} from "@storybook/react-vite";
+
+const ExpressionWidget = getWidget("expression")!;
```
nishasy: I think we decided to move away from the `getWidget` approach, right? I don't think we've updated this in the Claude files yet.

packages/perseus/src/widgets/expression/__docs__/expression-interactions-regression.stories.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+import {themeModes} from "../../../../../../.storybook/modes";
+import {getWidget} from "../../../widgets";
+
+import {expressionRendererDecorator} from "./expression-renderer-decorator";
+
+import type {PerseusExpressionWidgetOptions} from "@khanacademy/perseus-core";
+import type {Meta, StoryObj} from "@storybook/react-vite";
+
+const ExpressionWidget = getWidget("expression")!;
```
nishasy: same here

.claude/plans/expression-font-color-progress-report.md
nishasy, line 1: Do we want to keep this file here? I was under the impression the progress reports were supposed to be ignored by git.

nishasy (commented):
Left a couple questions, but looking good overall

.claude/plans/expression-font-color-progress-report.md
Myranae: I usually remove them before landing. I don't remember how we set up the ignore, but I might have mine in a different file structure. Will remove before landing though. :) Wanted it available for reviewers. I should probably move it to the plans folder since we have checks to make sure we don't land with files there.

Update: Moved to the plans folder which prevents landing if there are any documents inside. Should see a failing check for it too. Once I get approvals, I can remove it. :)

packages/perseus/src/widgets/input-number/__snapshots__/input-number.test.ts.snap
Myranae: When I saw it, I assumed it was removed from inline but carried out somewhere else, and it looks like that's the case. Thanks for calling this out though as I didn't fully understand why either and didn't realize it 😅 Per Claude:

> The property is still set as an inline style on the element in the browser — it's now `background-color:  var(--wb-semanticColor-core-background-base-default)` instead of `background-color: white`. The browser resolves the CSS variable to the actual color using Wonder Blocks' theme stylesheet. jsdom is what causes it to disappear from the snapshot: it doesn't support CSS custom properties in inline style assignment, so when React sets the value via CSSOM, jsdom silently drops it. The snapshot reflects what jsdom sees, not what a browser renders.

I tested it as well. Switched the color to red and was able to see it in a story, so it is affecting the page still :)

packages/perseus/src/widgets/expression/__docs__/expression-initial-state-regression.stories.tsx
Myranae: Fixed!

packages/perseus/src/widgets/expression/__docs__/expression-interactions-regression.stories.tsx
Myranae: Fixed! Thanks so much for pointing these out! <3

packages/perseus/src/widgets/expression/__docs__/expression-initial-state-regression.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+    component: Expression,
+    tags: ["!autodocs", "!manifest"],
+    parameters: {
+        docs: {
+            description: {
+                component:
+                    "Regression tests for the Expression widget that do NOT " +
+                    "need any interactions to test.",
+            },
+        },
```
Myranae, line 19: Might remove this since it's only useful in the code and doesn't seem to appear in Storybook. Discussion in Slack about this.

packages/perseus/src/widgets/expression/__docs__/expression-initial-state-regression.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+        functions: [],
+        times: false,
+        extraKeys: [],
+        visibleLabel: "Enter your answer",
+        ariaLabel: "Enter your answer",
+    },
+};
+
+// Verifies the mobile input (KeypadInputWithInterface / KeypadInput) renders
+// correctly in the mobile keypad context — background and border use semantic tokens
```
Myranae: Also not sure if the "Verifies" comments here are useful. Might remove?

packages/perseus/src/widgets/expression/__docs__/expression-interactions-regression.stories.tsx
```diff
@@ -0,0 +1,172 @@
...
+    component: Expression,
+    tags: ["!autodocs", "!manifest"],
+    parameters: {
+        docs: {
+            description: {
+                component:
+                    "Regression tests for the Expression widget that DO need " +
+                    "some sort of interaction to test.",
+            },
+        },
```
Myranae, line 21: Might remove

packages/perseus/src/widgets/expression/__docs__/expression-initial-state-regression.stories.tsx
Myranae: We want to keep it just for devs. Sounds good!

packages/perseus/src/widgets/expression/__docs__/expression-initial-state-regression.stories.tsx
Myranae: Removed the comments in this file.

packages/perseus/src/widgets/expression/__docs__/expression-interactions-regression.stories.tsx
Myranae: Keeping!

.claude/artifacts/implementation/widget-progress-reports/sorter-font-color-progress-report.md
Myranae, line 1: This seems to have somehow slipped into main. Whoops! Removing

packages/perseus/src/widgets/expression/__docs__/expression-interactions-regression.stories.tsx
```diff
@@ -0,0 +1,164 @@
...
+        extraKeys: [],
+    },
+    play: async ({canvas, userEvent}) => {
+        const mathInput = canvas.getByRole("textbox");
+        await userEvent.click(mathInput);
+    },
+};
+
+// Verifies the keypad popout renders with the Numbers tab selected by default —
+// captures the number buttons (0–9, operators, backspace)
```
mark-fitzgerald: Nit: Maybe simplify these comments to just the basics?
```
// Number buttons (0-9, operators, backspace)
```

mark-fitzgerald (approved):

packages/perseus/src/widgets/expression/__docs__/expression-interactions-regression.stories.tsx
Myranae: Updated all the comments on this file :) Thanks so much!

nishasy (approved):
Looks good to me!

PR #3704, opened 2026-06-01
Fix GIF related visual regression stories for image

Myranae:
## Summary:

- Fixes a flaky Chromatic snapshot in the existing `TallGifImage` story caused by the GIF rendering inconsistently across snapshots -- discovered when the new stories triggered a Chromatic run that surfaced the pre-existing issue.

- The story omitted `width` and `height` on `backgroundImage`. `SvgImage` requires both to enter the responsive rendering path, which is the only path that uses the canvas-based `GifImage` component. Without dimensions, the GIF rendered as a plain ``, giving the browser full control over animation frame timing — a race condition Chromatic can't resolve. With dimensions present, `GifImage` renders, starts paused at frame 0 for a11y, and the play function waits for the first frame to be drawn before Chromatic snapshots.

Issue: LEMS-4214

## Test plan:
- [x] Open **Widgets > Image > Visual Regression Tests > Initial State > TallGifImage**
  - [x] The tall portrait GIF renders as a canvas (not a plain ``) — visible as an animation frozen at frame 0
  - [x] The Play Animation button and Explore image buttons appear in the info area below
  - [x] The story does not error in the browser console
  - [x] Open the explore modal → verify the description reads the surreal illustration text, not the skateboard description


Myranae:
@claude review once

catandthemachines (approved):
Yeah this looks to have done the trick! Thank you @Myranae for doing this quick fix. 😊

packages/perseus/src/widgets/image/__docs__/image-initial-state-regression.stories.tsx
```diff
@@ -340,12 +342,23 @@ export const TallGifImage: Story = {
         },
     },
     args: {
-        backgroundImage: {
-            url: "https://cdn.kastatic.org/ka-content-images/1e6f6fd4de01058c3d548b7a942bd9e76d565fa3.gif",
-        },
-        alt: gifImageAlt,
-        caption: gifImageAlt,
-        longDescription: gifImageAlt,
+        backgroundImage: tallGifImage,
```
nishasy, line 345: I think it's possible that the background image actually having a size would fix this even without the `play`.

Myranae: Looks like there could still be a race between the blank canvas and frame 0 even with the sizes added. I included Claude's response below. Is there any downside to using the play function?

<details>
<summary>Claude's Response</summary>
  With dimensions added, SvgImage routes to GifImage (canvas-based) instead of ImageLoader (). GifImage starts paused at frame 0 for a11y — so unlike the native  path where the browser could be anywhere in the animation, you're now down to two possible states:   
  

  1. Blank canvas — Chromatic snapshots before decodeGifFrames resolves (the async CDN fetch + parse)
  2. Frame 0 drawn — Chromatic snapshots after

  GifImage uses fetch() internally to load the GIF data. Chromatic only auto-waits for resources the browser loads natively (``, CSS, JS) — not arbitrary fetch() calls from JavaScript. So without the play function, there's still a race between blank and frame 0.

  The play function is what collapses those two states into one by telling Chromatic "don't snapshot until canvas.width > 0", which only becomes true after drawFrame(0) runs.
  </details>

nishasy: I was actually thinking about this after work yesterday and realized I'm wrong 😅 This looks good to me.

nishasy (approved):

PR #3705, opened 2026-06-01
refactor: locked polygon points shape

mojadem:
## Summary

This is the first PR in a stack to enable per-vertex angle labeling in locked polygons.

- Add `LockedPolygonPointType` with shape `{coord: Coord}`
- Change `LockedPolygonType.points` from type `Coord[]` to `LockedPolygonPointType[]`
- Update rendering, editor, linter, defaults, fixtures, stories, and tests to use `points.coord`
- Keep parser backwards-compatible by accepting legacy raw `Coord[]` polygon points and normalizing to the new shape
- Add parser coverage for both new and legacy locked polygon point formats

## Next steps

- Extend `LockedPolygonPointType` with a `showAngle` field that determines whether the angle should be shown for that point (#3730)
- Extend `LockedPolygonPointType` with a `computedAngle` field that is populated during parsing with the computed angle measurement

## Test plan
- `pnpm jest packages/perseus-core/src/parse-perseus-json/regression-tests/parse-perseus-json-regression.test.ts -u --runInBand --watchman=false`
- `pnpm jest packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.test.ts packages/perseus-core/src/utils/get-default-figure-for-type.test.ts
packages/perseus-core/src/utils/generators/interactive-graph-widget-generator.test.ts packages/perseus-linter/src/rules/interactive-graph-widget-error.test.ts --runInBand --watchman=false`

### DIFF
--- a/packages/perseus-core/src/data-schema.ts
+++ b/packages/perseus-core/src/data-schema.ts
@@ -1076,12 +1076,12 @@ export type LockedEllipseType = {
 };
 
 export type LockedPolygonPointType = {
-    coord: Coord[];
+    coord: Coord;
 };
 
 export type LockedPolygonType = {
     type: "polygon";
-    points: LockedPolygonPointType;
+    points: LockedPolygonPointType[];
     color: LockedFigureColor;
     showVertices: boolean;
     fillStyle: LockedFigureFillType;
--- a/packages/perseus-core/src/data-schema.ts
+++ b/packages/perseus-core/src/data-schema.ts
@@ -1075,9 +1075,13 @@ export type LockedEllipseType = {
     ariaLabel?: string;
 };
 
+export type LockedPolygonPointType = {
+    coord: Coord[];
+};
+
 export type LockedPolygonType = {
     type: "polygon";
-    points: Coord[];
+    points: LockedPolygonPointType;
     color: LockedFigureColor;
     showVertices: boolean;
     fillStyle: LockedFigureFillType;
--- a/packages/perseus-core/src/parse-perseus-json/regression-tests/item-data/interactive-graph-locked-figures-missing-weights.ts
+++ b/packages/perseus-core/src/parse-perseus-json/regression-tests/item-data/interactive-graph-locked-figures-missing-weights.ts
@@ -140,9 +140,9 @@ export default {
                         {
                             type: "polygon",
                             points: [
-                                {coord: [-8, -7]},
-                                {coord: [-8, -3]},
-                                {coord: [-4, -3]},
+                                [-8, -7],
+                                [-8, -3],
+                                [-4, -3],
                             ],
                             color: "pink",
                             showVertices: false,
--- a/packages/perseus-core/src/parse-perseus-json/regression-tests/item-data/interactive-graph-locked-figures-missing-weights.ts
+++ b/packages/perseus-core/src/parse-perseus-json/regression-tests/item-data/interactive-graph-locked-figures-missing-weights.ts
@@ -139,13 +139,11 @@ export default {
                         },
                         {
                             type: "polygon",
-                            points: {
-                                coord: [
-                                    [-8, -7],
-                                    [-8, -3],
-                                    [-4, -3],
-                                ],
-                            },
+                            points: [
+                                {coord: [-8, -7]},
+                                {coord: [-8, -3]},
+                                {coord: [-4, -3]},
+                            ],
                             color: "pink",
                             showVertices: false,
                             fillStyle: "none",
--- a/packages/perseus-core/src/parse-perseus-json/regression-tests/item-data/interactive-graph-locked-figures-missing-weights.ts
+++ b/packages/perseus-core/src/parse-perseus-json/regression-tests/item-data/interactive-graph-locked-figures-missing-weights.ts
@@ -139,11 +139,13 @@ export default {
                         },
                         {
                             type: "polygon",
-                            points: [
-                                [-8, -7],
-                                [-8, -3],
-                                [-4, -3],
-                            ],
+                            points: {
+                                coord: [
+                                    [-8, -7],
+                                    [-8, -3],
+                                    [-4, -3],
+                                ],
+                            },
                             color: "pink",
                             showVertices: false,
                             fillStyle: "none",
--- a/packages/perseus-core/src/utils/generators/interactive-graph-widget-generator.test.ts
+++ b/packages/perseus-core/src/utils/generators/interactive-graph-widget-generator.test.ts
@@ -1179,7 +1179,17 @@ describe("generateIGLockedPolygon", () => {
         // Assert
         expect(lockedPolygon).toEqual({
             type: "polygon",
-            points: [{coord: [0, 2]}, {coord: [-1, 0]}, {coord: [1, 0]}],
+            points: [
+                {
+                    coord: [0, 2],
+                },
+                {
+                    coord: [-1, 0],
+                },
+                {
+                    coord: [1, 0],
+                },
+            ],
             color: "grayH",
             showVertices: false,
             fillStyle: "none",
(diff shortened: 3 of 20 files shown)


packages/perseus-core/src/data-schema.ts
```diff
@@ -1075,9 +1075,13 @@ export type LockedEllipseType = {
...
 };
 
+export type LockedPolygonPointType = {
+    coord: Coord;
+};
+
 export type LockedPolygonType = {
     type: "polygon";
-    points: Coord[];
+    points: LockedPolygonPointType[];
```
SonicScrewdriver, line 1084: It might be nice to add a comment here explaining the intention to add more data in the LockedPolygonPointType, but it's not strictly neccessary

SonicScrewdriver (approved):
I think this looks good! Thanks for working on this. I'll also tag @jeremywiebe in case he has any further thoughts on the approach. 

We'll need to be very careful deploying this, as we need to give the Content Platform a heads up about the schema change before we do. Otherwise we can break publishing. :) Happy to provide support with deploying if you need any!

packages/perseus-core/src/parse-perseus-json/regression-tests/item-data/interactive-graph-locked-figures-missing-weights.ts
```diff
@@ -140,9 +140,9 @@ export default {
                         {
                             type: "polygon",
                             points: [
-                                [-8, -7],
-                                [-8, -3],
-                                [-4, -3],
+                                {coord: [-8, -7]},
+                                {coord: [-8, -3]},
+                                {coord: [-4, -3]},
```
jeremywiebe: These `item-data/` files should never be modified as they represent data imported from production. Instead, create a second file with the new structure(s). 

See my comment at the top of this file.

packages/perseus-core/src/parse-perseus-json/regression-tests/item-data/interactive-graph-locked-figures-missing-weights.ts
jeremywiebe, line 3: @mojadem just want to make sure you saw this.

packages/perseus-core/src/utils/generators/interactive-graph-widget-generator.test.ts
```diff
@@ -1179,11 +1179,7 @@ describe("generateIGLockedPolygon", () => {
         // Assert
         expect(lockedPolygon).toEqual({
             type: "polygon",
-            points: [
-                [0, 2],
-                [-1, 0],
-                [1, 0],
-            ],
+            points: [{coord: [0, 2]}, {coord: [-1, 0]}, {coord: [1, 0]}],
```
jeremywiebe: **nit**: I'm not sure if prettier chose this, but the original, multi-line layout reads more easily.  No worries if this new inline layout if `prettier`'s choice.

packages/perseus-editor/src/widgets/__tests__/interactive-graph-editor-locked-figures.test.tsx
```diff
@@ -1079,7 +1079,7 @@ describe("InteractiveGraphEditor locked figures", () => {
                     lockedFigures: [
                         expect.objectContaining({
                             type: "polygon",
-                            points: [...defaultPolygon.points, [0, 0]],
+                            points: [...defaultPolygon.points, {coord: [0, 0]}],
```
jeremywiebe, line 1082: I'm assuming this spread works because `defaultPolygon.points` has been changed to be an array of these new `{coord: [number, number]}` format?

packages/perseus-editor/src/widgets/__tests__/interactive-graph-editor-locked-figures.test.tsx
```diff
@@ -1119,7 +1119,9 @@ describe("InteractiveGraphEditor locked figures", () => {
                     lockedFigures: [
                         expect.objectContaining({
                             type: "polygon",
-                            points: squarePolygonPoints.slice(1),
+                            points: squarePolygonPoints
+                                .slice(1)
+                                .map((coord) => ({coord})),
```
jeremywiebe: **suggestion**: This predates your change, but using the original test data in the assertion is a smell We should inline the values like we do later in this file around lines 1152-1156 (next chunk in this PR)

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-polygon-settings.tsx
```diff
@@ -66,6 +66,9 @@ const LockedPolygonSettings = (props: Props) => {
         onMove,
         onRemove,
     } = props;
+    const coords = points.map((point) => point.coord);
+    const coordsToPoints = (coords: Coord[]) =>
+        coords.map((coord) => ({coord}));
```
jeremywiebe: I haven't looked through all the usages of `coords` in this area, but I wonder if it'd be cleaner to use the new structure everywhere rather than have to keep dealing with both `[number, number]` and `{x: number, y: number}`.

jeremywiebe (approved):
A few thoughts, but nothing blocking!  Thanks Matt!

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-polygon-settings.tsx
mojadem: Agreed! I updated the logic to always derive from the points prop

packages/perseus-core/src/parse-perseus-json/regression-tests/item-data/interactive-graph-locked-figures-missing-weights.ts
mojadem: Ah, thanks for the heads up! I missed the note at the top of the file. Do we need a regression test in this context, or is unit testing the parser enough?

packages/perseus-core/src/utils/generators/interactive-graph-widget-generator.test.ts
mojadem: Prettier doesn't allow:
```typescript
points: [
    { coord: [1, 1] },
    { coord: [2, 2] },
    { coord: [3, 3] },
],
```
but does allow:
```typescript
points: [
    {
        coord: [1, 1],
    },
    {
        coord: [2, 2],
    },
    {
        coord: [3, 3],
    },
],
```
which mirrors the labels below, so I'll update it to that.

packages/perseus-editor/src/widgets/__tests__/interactive-graph-editor-locked-figures.test.tsx
mojadem: Right! We do that in `packages/perseus-core/src/utils/get-default-figure-for-type.ts`

packages/perseus-editor/src/widgets/__tests__/interactive-graph-editor-locked-figures.test.tsx
mojadem: Agreed! The assertion has been updated.

packages/perseus-core/src/data-schema.ts
mojadem: To clarify, I intend on merging the stacked PRs together as one schema change. I will be sure to get the Go schema change deployed first once the target schema is finalized.

packages/perseus-core/src/parse-perseus-json/regression-tests/item-data/interactive-graph-locked-figures-missing-weights.ts
jeremywiebe: You should add a new file with this new data format. That way we ensure that our parser remains compatible with all previous known schema versions _and_ supports the new version.

PR #3706, opened 2026-06-01
Fix package.json key order for 'exports'

jeremywiebe:
## Summary:

While working on improving the item-splitting check, I noticed a warning in the console output: 

image

I did [some research](https://nodejs.org/api/packages.html#conditional-exports) and it turns out that the order of keys in our `exports` objects is important and they are processed in the order they appear. 

This PR re-orders all `package.json` files that use `exports` and updates the re-formatting script to format them in this way also.

Issue: LEMS-4229 

## Test plan:

Run `./utils/format-package.json.sh; git diff` (should be no changes)

Note the order of each object in our `exports` objects: they are always `types`, `source`, `import`, `require`.


utils/format-package-json.sh
jeremywiebe, line 1: This script is a helper I've been maintaining to keep our `package.json` files somewhat consistent. It's not tied into any process and I just use it manually. However, I like the idea of keeping it up to date as then we can run it over the repo whenever we want.

nishasy (approved):
Good catch

PR #3708, opened 2026-06-02
Added definition widget identifier string for screenreader into strings.ts

Evelas78:
## Summary:
Added the identifier string for the definitions widget that'll be used as a screenreader into strings.ts

Issue: https://khanacademy.atlassian.net/browse/LEMS-4104

## Test plan:
- Used voiceover to verify that Defintion of: {Word} showed up in screenreader


nishasy (approved):
Looks good to me! I was also able to confirm that it works manually in Storybook.

PR #3711, opened 2026-06-02
Change PR comparison checks to have separate concurrency groups

jeremywiebe:
## Background 

The `PR Comparison Checks` workflow defines two independent checks — `schema-change` and `item-splitting` — that each fail a PR until the author acknowledges the change by adding an `-ack` label. Once the `-ack` label is applied the check re-runs and passes. 

## The Problem 

The problem is that both checks shared a single workflow-level `concurrency` group keyed only on the workflow name and PR number, with `cancel-in-progress: true`. Because GitHub fires a separate `labeled` event per label, adding both `-ack` labels at once (or in quick succession) produces two events landing in the same group, and the second one cancels the first's in-flight run. Whichever job the cancelled run owned then gets skipped in the surviving run (its `if` guard only matches its own ack label), so it never re-posts a passing check — leaving a stale failing check that the author already acknowledged. 

## The Fix 

This PR moves the `concurrency` configuration out of the workflow level and into each job, giving them distinct group keys (`…-schema-change-…` and `…-item-splitting-…`). The two ack flows can now run concurrently without cancelling each other, while each job still cancels its own superseded runs.

Issue: LEMS-4236

## Test plan:

- On a test PR that trips both checks, add both `-ack` labels at the same time and confirm both checks re-run and end green (neither cancels the other).
- Push a new commit to a PR and confirm an in-flight run is still superseded by the newer push within each check.

## Follow-up (after this lands):

This PR is a prerequisite for making the comparison checks *required* in branch protection. Once these checks are required, the concurrency bug fixed here would otherwise turn into a hard merge-block (a cancelled ack re-run leaves a stuck failing check), so that enforcement work is intentionally gated on this fix.

Remaining steps, to be done after this merges:

- Flip the [`main branch protection`](https://github.com/khan/perseus/settings/rules/17139544) ruleset (currently in Evaluate) to **Active**. It already lists the three required checks: `Lint, Typecheck, Format, and Test`, `PR Comparison: 🗄️ Schema Change`, and `PR Comparison: 🛠️ Item Splitting`.
- Verify on a live PR that an unacknowledged schema / item-splitting change now blocks merge until the `-ack` label flips the check green.
- Delete the legacy classic [branch-protection rule](https://github.com/khan/perseus/settings/branch_protection_rules/25361949) for `main`, which the ruleset replaces.


.github/workflows/pr-comparison-checks.yml
```diff
@@ -16,13 +16,14 @@ permissions:
     issues: write
     checks: write
 
-concurrency:
-    group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
-    cancel-in-progress: true
-
```
jeremywiebe, line 22: Having the `concurrency` defined at the workflow level, means that all jobs are treated as a unit and new events for one job can cancel in-progress events for the other. Bad. 

We don't want to turn off `cancel-in-progress` to solve this because then we'd have to wait for jobs to run that we know have been superseded, wasting time and Github $$$.

nishasy (approved):
Great sleuthing! Thanks for fixing this.

PR #3712, opened 2026-06-03
Fix how we're converting the 'number' answerType for Input Number

SonicScrewdriver:
## Summary:
[less-is-more] Update how we convert Input Number's "number" answerType

still need to test this. 

Issue: LEMS-4216

## Test plan:

### DIFF
--- a/packages/perseus-core/src/widgets/input-number/to-numeric-input.test.ts
+++ b/packages/perseus-core/src/widgets/input-number/to-numeric-input.test.ts
@@ -178,7 +178,7 @@ describe("convertInputNumberOptionsToNumericInput", () => {
         ]);
     });
 
-    it("does not exclude `decimal` and `integer` from `answerForms` when inexact is true", () => {
+    it("returns an empty array for answerForms when inexact is true and value has more than 10 decimal places", () => {
         const options: PerseusInputNumberWidgetOptions = {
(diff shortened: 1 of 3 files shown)


packages/perseus-core/src/widgets/input-number/to-numeric-input.test.ts
```diff
@@ -190,7 +178,7 @@ describe("convertInputNumberOptionsToNumericInput", () => {
         ]);
     });
 
-    it("does not exclude `decimal` and `integer` from `answerForms` when inexact is true", () => {
+    it("returns an empty array for answerForms when inexact is true and value has more than 10 decimal places", () => {
```
handeyeco, line 181: The original test talks about `inexact` and this test talks about 10 decimal places.

1. Is this intentional?
2. What's the relevance of 10 decimal places?

handeyeco (approved):

benchristel (approved):
Thanks for the fix!

packages/perseus-core/src/widgets/input-number/to-numeric-input.test.ts
benchristel: See [this PR](https://github.com/Khan/perseus/pull/3639) for the context on the "10 decimal places" case.

PR #3720, opened 2026-06-03
[Radio] Update Radio image editor so it autosaves

nishasy:
## Summary:
Previously, content authors using the "Add image" feature within the Radio editor would
have to fill out all the image info first and press the "save" button for the image to
be saved and show up in the content.

This led to images sometimes falling through the cracks if the save button wasn't clicked
for any reason.

To prevent images from getting missed, we can update the Radio image editor to always save
whenever the URL and/or alt text are updated. This means no longer having the initial
editor tile, but rather adding the image accordions, just not populated, from the getgo.

Changes in PR:
- Remove the editor tile with the save button.
- Move the image preview from `radio-option-content-and-image-editor.tsx` to `radio-image-editor.tsx`.
- Move the useEffect from `radio-option-content-and-image-editor.tsx` to `radio-image-editor.tsx` and only have it run with the URL is changed (stop the preview from flickering while the alt is being updated).
- Update tests.

Issue: https://khanacademy.atlassian.net/browse/LEMS-4240

## Test plan:
`pnpm jest packages/perseus-editor/src/widgets/__tests__/radio-editor.test.tsx`

Storybook
- Go to `/?path=/story/widgets-radio-editor-demo--multi-choice`
- Click "Add image" button in the first Radio option
- Paste in a sample image URL (`https://cdn.kastatic.org/ka-content-images/61831c1329dbc32036d7dd0d03e06e7e2c622718.jpg`)
- Confirm that the image shows up in the content box and both previews
- Update the alt text - make sure the image doesn't flicker as you type

## Before

https://github.com/user-attachments/assets/b96bdea0-a40c-421c-a5ae-8674db8d9b93

## After

https://github.com/user-attachments/assets/3eb37c10-f094-4e9a-8cc1-7382f54377dc


packages/perseus-editor/src/widgets/__tests__/radio-editor.test.tsx
```diff
@@ -889,154 +889,6 @@ describe("radio-editor", () => {
             );
         });
 
-        /***** Tests for image adder tile *****/
-
-        it("should show the 'Add image' button by default", () => {
```
anakaren-rojas, line 894: should there be new "add image" tests for the new behavior?

packages/perseus-editor/src/widgets/radio/radio-image-editor.tsx
```diff
@@ -131,7 +122,7 @@ export default function RadioImageEditor({
                 value={imageUrl}
                 placeholder="cdn.kastatic.org/..."
                 onChange={(value) => {
-                    setImageUrl(value);
+                    handleUrlChange(value);
```
anakaren-rojas: this would save on each keystroke, right? would that cause any issues for content if they decide to type the url instead of pasting it?

packages/perseus-editor/src/widgets/__tests__/radio-editor.test.tsx
```diff
@@ -1117,147 +962,165 @@ describe("radio-editor", () => {
...
-            const saveImageButton = screen.getByRole("button", {
-                name: "Save image",
-            });
-
-            // Assert
-            expect(saveImageButton).toHaveAttribute("aria-disabled", "true");
-        });
-
-        it("should enable the 'Save image' button when the image URL and alt text are changed", async () => {
+        it("should call onChange when the image URL is changed", async () => {
```
anakaren-rojas: can there also be a test for when a user is typing the url?

packages/perseus-editor/src/widgets/radio/radio-image-editor.tsx
```diff
@@ -1,119 +1,110 @@
...
-        // Only fetch if URL changed and we don't already have dimensions
-        if (imageUrl !== initialImageUrl) {
-            void fetchDimensions();
-        }
-    }, [imageUrl, initialImageUrl]);
+        void fetchDimensions();
+
+        // Cleanup function to cancel the fetch if the component unmounts.
+        return () => {
+            cancelled = true;
```
mark-fitzgerald, line 92: Probably a "me" thing - I'm having trouble understanding how setting `cancelled = true` here overrides the assignment on line 48 that sets it to `false`.

packages/perseus-editor/src/widgets/radio/radio-image-editor.tsx
```diff
@@ -1,119 +1,110 @@
...
-            {/* Close button */}
-            {onClose && (
-                <IconButton
-                    aria-label="Close"
-                    icon={xIcon}
-                    size="small"
-                    kind="tertiary"
-                    onClick={handleClose}
-                    style={{position: "absolute", top: 4, right: 4}}
+        <div>
```
mark-fitzgerald: If no class is being applied to this `<div>`, is it still needed, or could it be replaced with a React Fragment?

packages/perseus-editor/src/widgets/__tests__/radio-editor.test.tsx
nishasy: Yes, thank you! I'll add a test to confirm that an empty image markdown is added to the content when the button is pressed.

packages/perseus-editor/src/widgets/__tests__/radio-editor.test.tsx
nishasy: Great point. I'm trying to test typing stuff into the URL field right now, and it keeps taking the focus off the field. I doubt a lot of manual typing is happening, but I think I'll update it to save on blur instead of with each keystroke.

packages/perseus-editor/src/widgets/radio/radio-image-editor.tsx
nishasy: My understanding is that the cleanup function would run when the component is unmounted, so if the URL changes or something, `cancelled` is set to true and the effect stops trying to fetch the dimensions.

I think Claude was able to come up with a more in-depth explanation:
> Line 67 doesn't override line 48 - they run at different times within the same effect execution, and `cancelled` is a fresh variable each time the effect runs.
> 
> When the effect runs, line 48 declares a brand-new local `canceled = false`. Both `fetchDimensions` (line 49) and the cleanup function (line 66) close over that same variable. Line 48 sets the initial value; line 67 only runs later, when React invokes the returned cleanup function (on unmount, or before the effect re-runs because `imageUrl` changed). At that point it flips that closure's `cancelled` to `true`, which is what the in-flight `await` checks on lines 52/57 when it finally resolves.
> 
> So the sequence for a single effect run is: `cancelled = false` (now) → fetch starts → [later] cleanup sets `cancelled = true` → [even later] the awaited promise resolves and sees `true`, so it skips `setDimensions`. They never "compete"; line 67 is simply a later event in that variable's lifetime.
>
> Why this pattern exists
> The whole point is to avoid calling setDimensions after the relevant image is no longer the current one. Without it, this race could happen:
>  1. imageUrl is image A → effect runs, fetch A starts (slow).
> 2. User changes URL to image B → React runs cleanup for A's effect, then runs a new effect for B, fetch B starts (fast).
> 3. Fetch B resolves → sets dimensions for B. ✅
> 4. Fetch A finally resolves → without the guard, it would overwrite with A's stale dimensions. ❌
> 
> Because each effect run has its own cancelled, step 2's cleanup flips A's flag to true, so when A resolves in step 4, the if (!cancelled) check on line 52 is false and it bails out. B's flag is still false, so B's update went through normally.

packages/perseus-editor/src/widgets/radio/radio-image-editor.tsx
nishasy: Good catch

packages/perseus-editor/src/widgets/radio/radio-image-editor.tsx
nishasy: Yup. I replied to your other comment about this, but I think I'm going to do this on blur to avoid issues while typing.

packages/perseus-editor/src/widgets/radio/radio-option-content-and-image-editor.tsx
```diff
@@ -263,37 +206,20 @@ export const RadioOptionContentAndImageEditor = React.forwardRef<
...
-                        field={
-                            <ImagePreview
-                                src={image.url}
-                                alt={`Preview: ${image.altText ?? "No alt text"}`}
-                                width={image.width}
-                                height={image.height}
-                            />
-                        }
+                    <RadioImageEditor
+                        imageUrl={image.url}
```
anakaren-rojas, line 210: In the scenario that a content author decides to use the same image in multiple places, this would cause a collision, right?

packages/perseus-editor/src/widgets/radio/radio-image-editor.tsx
```diff
@@ -1,119 +1,104 @@
...
-                    style={{position: "absolute", top: 4, right: 4}}
+        <>
+            {/* Preview */}
+            {imageUrl ? (
+                <LabeledField
+                    label="Preview"
+                    field={
+                        <ImagePreview
+                            src={imageUrl}
+                            alt={`Preview: ${imageAltText ?? "No alt text"}`}
```
anakaren-rojas: NIT: Can probably remove the `?? "No alt text"` since it would never trigger

packages/perseus-editor/src/widgets/radio/radio-image-editor.tsx
```diff
@@ -1,119 +1,104 @@
...
-        initialImageHeight,
-    ]);
-
     const uniqueId = React.useId();
     const imageUrlTextAreaId = `${uniqueId}-image-url-textarea`;
     const imageAltTextTextAreaId = `${uniqueId}-image-alt-text-textarea`;
 
-    // Fetch image dimensions when URL changes, this will be used for the
-    // proper display of the preview on the left side of the editor
+    const [stateImageUrl, setStateImageUrl] = React.useState<string>(imageUrl);
```
anakaren-rojas, line 33: Does `stateImageUrl` need a useEffect in case the `imageUrl` changes?

packages/perseus-editor/src/widgets/radio/radio-image-editor.tsx
mark-fitzgerald: Oh, I see now. The function `fetchDimensions` is an internal async function that references the `cancelled` variable.

I'm still concerned that this isn't obvious - again, maybe it's a "me" thing. Would you please add a comment that explains the workflow of the `fetchDimensions` function within the `useEffect`, and how the `cancelled` variable is involved? When I read the code, I can't see how the function would be running while a "cancelled" event occurs. Is there a confluence of events that can illustrate how all of this works?

packages/perseus-editor/src/widgets/radio/radio-option-content-and-image-editor.tsx
nishasy: Was this comment supposed to be on line 194, with the `key={image.url}`?

Good catch. I'll update that.

packages/perseus-editor/src/widgets/radio/radio-image-editor.tsx
nishasy: I'm a bit confused about how it would be possible to get into that position.

I think that would only happen on the initial load, and any subsequent image URL updates would happen from the input field, so it would always happen via state. I _think_ other widgets need it because some other setting could also update the field in question, so the useEffect was needed to keep them in sync?

packages/perseus-editor/src/widgets/radio/radio-option-content-and-image-editor.tsx
anakaren-rojas: yes! how did this shift 😕

anakaren-rojas (approved):
Overall LGTM! 
I did find a random bug while testing that I think would annoy someone if they're continuously adding and deleting images on the same choice, but I don't think its too likely of a scenario

https://github.com/user-attachments/assets/d9eafa0b-5a26-4f29-9cb4-7e819fd03f86

nishasy:
@anakaren-rojas oh haha yeah, that does happen. I think trying to fix it causes a lack of space when adding a new one, so I didn't bother to stop it. I'm inclined to leave it as is until someone from content finds it to be a problem 😅

PR #3722, opened 2026-06-04
[Editor] Restore cursor positioning after paste/widget-insert

jeremywiebe:
## Summary:

While working on the Preview NG work, I was going to clean up the two extra parameters in the `onChange` callback prop (`silent` and `cb` (aka "callback")). While cleaning up the code I saw that there was some code in the paste path that actually provides a function for the `cb`. 

This led me to discover that we'd broken cursor handling when we handle paste in the editor. A while back, we did [some cleanup](https://github.com/Khan/perseus/pull/3318) that broke this cursor handling when we stopped passing `silent` and `cb` parameters in higher-level editor components, but `Editor` was still sending them - effectively they were dropped on the floor. This results in the cursor always jumping to the end of the content when you paste anywhere.

Thsi PR restores this behaviour, but does so entirely within the `Editor` component instead of depending on the deprecated `cb` `onChange` parameter. This allows us to still fully remove the `silent` and `cb` parameters, but restores cursor handling (all within the `Editor`).

paste-cursor-demo

Issue: LEMS-4241 

## Test plan:

Run storybook
Navigate to the Editor Demo page
Add a widget and configure it
Select the widget in the Markdown editor, copy it
Move the cursor to the beginning of the field and type `STARTEND`
Place the cursor between `START` and `END` (eg. `START|END`)
Paste
The copied widget should be pasted between `START` and `END` and the cursor should be sitting at the end fo the pasted widget (just before `END`).


packages/perseus-editor/src/editor.tsx
```diff
@@ -233,32 +238,32 @@ class Editor extends React.Component<Props, State> {
...
-            if (
-                document.execCommand(
-                    "insertText",
-                    false,
-                    this.props.content,
-                ) === false
-            ) {
-                // This command is not implemented. Fall back to setting `value`
-                // directly.
-                textarea.value = this.props.content;
```
jeremywiebe, line 257: All of our supported browsers support `execCommand` now, so there's no point in keeping this extra fallback code!

packages/perseus-editor/src/editor.tsx
```diff
@@ -564,20 +569,16 @@ class Editor extends React.Component<Props, State> {
...
+            // After React commits the new content, place the cursor at the end
+            // of what we just pasted in.
+            this._pendingCursorPos = selectionStart + safeText.length;
+            this.props.onChange({
+                content: newContent,
+                widgets: {
+                    ...safeWidgetData,
+                    ...this.getWidgetsReferencedIn(newContent),
                 },
-            );
```
jeremywiebe, line 580: We are moving to remove the callback prop from `onChange`. This is currently the only known usage, and we can easily manage it without this legacy prop - plus it doesn't require plumbing the callback value up and down the editor component tree (which we did... and which was broken for a while now!)

packages/perseus-editor/src/__tests__/editor.test.tsx
```diff
@@ -307,6 +308,127 @@ describe("Editor", () => {
...
+        };
+
+        beforeEach(() => {
+            // jsdom doesn't implement document.execCommand, which the Editor
+            // uses to keep programmatic content changes in the browser's undo
+            // stack. We stub it with a minimal impl that does what
+            // execCommand("insertText") would do in a real browser: replace
+            // the focused textarea's current selection with the supplied text.
+            document.execCommand = jest
+                .fn()
```
benchristel: Since none of the tests assert that execCommand was called, this could just be a normal function instead of a Jest mock.

packages/perseus-editor/src/__tests__/editor.test.tsx
```diff
@@ -307,6 +308,127 @@ describe("Editor", () => {
...
+            const pasteAtIndex = 5;
+            textarea.focus();
+            textarea.setSelectionRange(pasteAtIndex, pasteAtIndex);
+            // NOTE: we use `fireEvent` because userEvent.paste doesn't go
+            // through the jQuery-bound paste listener, but fireEvent does.
+            // eslint-disable-next-line testing-library/prefer-user-event
+            fireEvent.paste(textarea);
+
+            await waitFor(() => {
+                expect(textarea.value).toBe(`hello${pastedText} world`);
```
benchristel: I have always maintained that expected values in tests should be hardcoded, not created dynamically.

```suggestion
                expect(textarea.value).toBe(`helloPASTED world`);
```

If a test fails, the expected value is printed in the failure message. Having that value also appear in the test establishes a visual correspondence that makes the tests easy to skim and easy to debug.

Additionally, the more dynamic stuff a test does, the more it ends up duplicating the production logic it is trying to test. This pushes the test toward becoming tautological.

There are no tests for the tests, so they have to be kept extremely simple.

benchristel: Also, should this line be below the `// Assert` comment?

packages/perseus-editor/src/__tests__/editor.test.tsx
```diff
@@ -307,6 +308,127 @@ describe("Editor", () => {
...
+            });
+
+            // Assert
+            const expectedCursorPos = pasteAtIndex + pastedText.length;
+            expect(textarea.selectionStart).toBe(expectedCursorPos);
+            expect(textarea.selectionEnd).toBe(expectedCursorPos);
```
benchristel:
```suggestion
            expect(textarea.selectionStart).toBe(11);
            expect(textarea.selectionEnd).toBe(11);
```

If this makes it too hard to see why `11` is the correct value, that's a sign that the input values should be simplified — e.g. start with the text `"AZ"` and insert `"B"` at position 1. Then the final cursor position would be 2.

packages/perseus-editor/src/editor.tsx
```diff
@@ -233,32 +238,32 @@ class Editor extends React.Component<Props, State> {
...
+        // as one step, the same as a user-typed edit would.
         // See https://stackoverflow.com/questions/41273569/how-to-enable-ctrl-z-when-you-change-an-input-text-dynamically-with-react
-        // Note: On unsupported browsers (*cough* Firefox *cough*) `execCommand`
-        // will return false. However at least in Firefox setting `value` on a
-        // textbox clears the undo stack so we don't get unexpected undo
-        // behavior.
+        // `execCommand` is deprecated per MDN but remains the only widely
+        // supported way to keep programmatic edits in the undo stack; all of
+        // Khan Academy's supported browsers implement it (per caniuse, global
+        // support is ~96%).
```
benchristel: I feel like comments like this should always have a date.


```suggestion
        // Khan Academy's supported browsers implement it (per caniuse, global
        // support is ~96% as of June 2026).
```

packages/perseus-editor/src/editor.tsx
```diff
@@ -705,24 +706,13 @@ class Editor extends React.Component<Props, State> {
...
-                );
-            },
-        );
+        // After React commits the new content, place the cursor after the
+        // newly-inserted widget syntax (and any added newlines).
+        this._pendingCursorPos = newContent.length - postlude.length;
+        this.props.onChange({
+            content: newContent,
+            widgets: newWidgets,
+        });
```
benchristel, line 715: Since we called onChange, the next render should have the same `value` for the textarea that we just set. I think we could set the cursor position here, and `_pendingCursorPos` could go away.

jeremywiebe: I am pretty sure we have to wait for the next render so that our cursor position refers to the new content string (also, in my testing, even if I set the cursor position no the textarea ref after calling `this.props.onChange` it jumps to the end of the string after rendering.

Perhaps I'm misunderstanding you.

packages/perseus-editor/src/__tests__/editor.test.tsx
jeremywiebe: Those are good points.  I still find myself cheating a little bit and doing calculations in test assertions. I've removed them so they should be cleaner/less dynamic now.

packages/perseus-editor/src/editor.tsx
benchristel: Ohh, sorry, I thought we were inserting text with execCommand in this function. Since we're not doing that, we do need to move the cursor after the next render.

benchristel (approved):
LGTM! Thanks for the fix, and sorry I broke this originally!

PR #3723, opened 2026-06-04
Parser import lint rule, part 1: create the rule

handeyeco:
## Summary:

This does a couple of things:
- Adds a lint rule to prevent importing things into `packages/perseus-core/src/parse-perseus-json/perseus-parsers`
- Moves `typetest` files from `packages/perseus-core/src/parse-perseus-json/perseus-parsers` to `packages/perseus-core/src/parse-perseus-json/perseus-parsers-typetests` (so they won't be affected by the rule)
- Ignores the rule for now (will be addressed in future PRs)

This kind of feels wrong, which is why I'm submitting this before looking into the fixes; but if people don't mind moving `*.typetest.ts` and ignoring the rule in `*.test.ts`, then I can move forward with this.


.eslintrc.js
```diff
@@ -297,6 +297,13 @@ module.exports = {
                         target: "./packages/(?!.*test.js).*",
                         from: "utils",
                     },
+                    {
+                        // Files in parse-perseus-json may not import from the
+                        // rest of the codebase — it must stay self-contained.
+                        target: "./packages/perseus-core/src/parse-perseus-json/perseus-parsers",
+                        from: "./packages/perseus-core/src",
+                        except: ["./parse-perseus-json"],
+                    },
```
handeyeco, line 306: This is the added rule down here

jeremywiebe: One way you could avoid the moving of typetest files is to exclude them in the `target` with a file glob:

```
    target: "./packages/perseus-core/src/parse-perseus-json/perseus-parsers/**/!(*.typetest).ts"
```

jeremywiebe (approved):
Non-blocking, but I'd vote not to move the typetest files and use the glob approach I commented about in `.eslintrs.js`

.eslintrc.js
handeyeco: 🤦 I originally tried something like that and couldn't get it to work. Claude failed me; long live JeremyAI.

PR #3724, opened 2026-06-04
Comparison Checks: Use the job's own result as the required check

jeremywiebe:
## Summary:

The `PR Comparison Checks` workflow gates schema and item-splitting changes: it builds an artifact on the PR branch and on the base branch, diffs them, and requires the author to acknowledge an intentional change via an acknowledgement label. Once these were made *required* in branch protection, they started getting stuck on "Expected — Waiting for status to be reported" — the check had already reported green/red, yet the PR sat blocked.

The root cause was the verdict mechanism. The action posted its pass/fail as a **custom Checks-API check run**, and GitHub's Checks API gives no way to associate a check created with the API to the workflow run that produced it — GitHub files it into a check suite of its own choosing, so the check is effectively detached from its run. A single PR commit accumulates many check suites (the open event, the bot's own label-add events, concurrency-cancelled re-runs), and under that churn GitHub's required-status evaluation flakily failed to reconcile its "Expected" placeholder with the detached check. It bound on some PRs and not others.

This replaces the detached check with the **job's own pass/fail** as the verdict — a native check, which GitHub binds to required-status rules reliably (the existing required `Lint, Typecheck, Format, and Test` check never has this problem, for exactly this reason). The job now exits non-zero on an unacknowledged change. To keep it always-reporting without paying a full rebuild on every event, the expensive build+diff runs only on code events; on label events the verdict is derived from the labels (the bot keeps the change label in sync with the diff), so toggling the ack label is a ~15s job rather than a rebuild. (We do this because otherwise a *skipped* required check would be treated as passing, which would let an unacked change slip through!)

The original reason for the Checks API was to keep the diff out of the buried run-Summary page. To preserve that, the failing verdict step echoes the diff and a run-Summary link straight into its **log**, so clicking the failing check lands you right on it; the rendered Markdown version is also written to the run Summary. (On a label-triggered re-check there's no fresh diff, so the log explains that and points to the latest code-push run.)

Folded in along the way: the action now asserts a `pull_request` context (it's PR-only by design), and the label-management and verdict steps no longer act on cancelled runs — previously a run cancelled mid-build read an empty diff result and spuriously added the change label, causing add/remove label churn on PRs that never touched these files.

**Rollout note:** the required-check *names* change (`PR Comparison: Schema Change` → `PR Comparison: Schema Change (ubuntu-latest, 24.x)`), so the `main branch protection` ruleset must be re-pointed. The two comparison checks have already been temporarily removed from the required list; after this merges, re-add the new `… (ubuntu-latest, 24.x)` names as required.

Issue: LEMS-4236

## Test plan:

- After merge (and re-pointing the ruleset): open a PR that trips a schema or item-splitting change and confirm `PR Comparison: Schema Change` goes red, with the diff visible in the failing step's log (and rendered on the run Summary).
- Add the corresponding `-ack` label and confirm the check re-runs in ~15s (no rebuild) and goes green; remove it and confirm it fails again.
- Confirm a PR with no schema/item-splitting change passes ("No Changes").
- Push several commits in quick succession and confirm cancelled runs don't toggle the change labels or post spurious verdicts.


.github/actions/branch-compare/action.yml
```diff
@@ -1,213 +1,272 @@
...
+
+    To stay cheap, the expensive build+diff runs only on code events
+    (opened/synchronize/reopened/edited). On label events the verdict is derived
+    from the current labels without rebuilding, so toggling the ack label is a
+    ~15s job rather than a full rebuild.
+
+    Must be invoked from a workflow triggered by `pull_request` events: it reads
+    the PR's base ref, head SHA, number, and labels from the event payload and
+    has no meaning outside a pull request. Assumes the repository has already
+    been checked out (it's loaded as a local action).
```
jeremywiebe, line 21: This is alot of words, but I spent alot of time flip-flopping and testing the two approaches and the one we're moving to in this PR is definitely the best. Hopefully this comment captures the why - and prevents future "us" from attempting the failure path again.

.github/actions/branch-compare/action.yml
```diff
@@ -1,213 +1,272 @@
...
+              echo "::error::branch-compare must run from a pull_request-triggered workflow (got event '${{ github.event_name }}')."
+              exit 1
+
+        # ---------------------------------------------------------------------
+        # Build + diff: code events only. On label events every step here is
+        # skipped and the verdict is derived from the labels below, so a label
+        # toggle (e.g. adding the ack label) is a ~15s job, not a rebuild.
+        # ---------------------------------------------------------------------
+        - name: Install & cache node_modules
+          if: ${{ github.event.action != 'labeled' && github.event.action != 'unlabeled' }}
```
jeremywiebe, line 76: This is done on every subsequent step. We want to prevent the label/unlabel events from triggering a bunch of work when nothing's changed. 🤷‍♂️

anakaren-rojas: Would it be worth it to compute this once and have each step check that, so the definition lives in one place?

.github/actions/branch-compare/action.yml
```diff
@@ -1,213 +1,272 @@
...
-                             "${{ steps.capture-pr-artifacts.outputs.artifact-path }}" >> "$GITHUB_OUTPUT"
-              DIFF_EXIT_CODE=$?
-              echo "$DELIM" >> "$GITHUB_OUTPUT"
-              echo "diff-exit-code=$DIFF_EXIT_CODE" >> "$GITHUB_OUTPUT"
-
-        - name: Add or Remove Label Based on Diff
-          if: always()
+              DIFF_FILE="${{ runner.temp }}/branch-compare-diff.txt"
+              diff --unified \
+                  "${{ runner.temp }}/branch-compare/base" \
```
anakaren-rojas, line 133: Is there a risk these artifact paths ever get out of sync with the ones in the capture step? Can they be passed as step outputs so each path is defined in one place?

.github/actions/branch-compare/action.yml
```diff
@@ -1,213 +1,272 @@
...
+                  } >> "$GITHUB_STEP_SUMMARY"
+                  echo "::error::Comparison could not be completed (a prior step failed)."
+                  exit 1
+              fi
+
+              # Read current labels from the API (not the event payload) so we
+              # see any label edits this run just made.
+              change_present=false
+              acked=false
+              labels="$(gh pr view "$PR_NUMBER" --json labels --jq '.labels[].name')"
```
anakaren-rojas: should we give it the same friendly message the "comparison could not be completed" case gets just above, so a flaky API call reads as "re-run me" rather than a real failure?

.github/actions/branch-compare/action.yml
jeremywiebe: Two reasons I decided against it:
1. If we centralize it and it is ever "broken" (calculated incorrectly to `false` - ie. it thinks its a labelled event) the failure mode is that we skip all the checks and consider the check passing. With the duplication, we'd need to update every one intentionally.
2. Centralizing the logic adds more indirection especially as we'd have to pass in the flag from the callers.

.github/actions/branch-compare/action.yml
jeremywiebe: Yes, that's a small risk. It's been like this for a while.  I think I'll queue this up for a future PR as this PR is already quite full of churn.

.github/actions/branch-compare/action.yml
jeremywiebe: We have `set -euo pipefail` at the top of this script block, which will cause the entire step to blow up if `gh` returns a non-zero exit (errors). This will turn the check red. `gh` should be writing to `stderr` in error cases and so that should come out in the step logs.

I'll add an 'echo' above this so that if `gh` errors, we'll at least know what it was trying to do...

jeremywiebe: I changed my mind.  I asked Claude and it had a suggestion to change it to be more explicit. I liked that so I implemented it.

anakaren-rojas (approved):

PR #3726, opened 2026-06-04
Parser import lint rule, part 2: copy data arrays

handeyeco:
## Summary:

My first thought was to copy the array data that the parser was using, but I kind of wonder if it wouldn't be better to convert the enum types to `string` or `string[]`.

Issue: LEMS-4224


packages/perseus-core/src/parse-perseus-json/perseus-parsers/expression-keys.ts
```diff
@@ -0,0 +1,110 @@
+export const KeypadKeys = [
```
handeyeco: This kind of makes me question this linter strategy and `deriveExtraKeys` _really_ makes me question this strategy.

This is kind of looking like: any time we make a change to KeypadKeys, we need to have a new type - one for the original version and one for the new version.

I kind of wonder if `extraKeys: optional(array(enumeration(...KeypadKeys)))` should just be `extraKeys: optional(array(string))` and we let the widget figure it out.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/plotter-widget.ts
```diff
@@ -15,6 +12,14 @@ import {defaulted} from "../general-purpose-parsers/defaulted";
 
 import {parseWidget} from "./widget";
 
+const plotterPlotTypes = [
```
handeyeco: Same here, I wonder if `type: enumeration(...plotterPlotTypes)` should be `type: string`

packages/perseus-core/src/parse-perseus-json/perseus-parsers/expression-keys.ts
benchristel: > This is kind of looking like: any time we make a change to KeypadKeys, we need to have a new type - one for the original version and one for the new version.

Can you explain this a bit more? I think we only ever need one KeypadKeys type. Additions of new enum values to the KeypadKeys type are backward-compatible with existing data.

I still prefer having an explicit list of the keys we support in the data schema.

benchristel: @handeyeco Could you make the export from this file an actual parser?

```
export const KeypadKeys = enumeration(...);
```

if we export an array here, it might accidentally get used for non-parser stuff.

benchristel (approved):
I'd prefer not to have arrays exported from one file and used to construct a parser in another file. I left a suggestion inline. LGTM overall, though!

packages/perseus-core/src/parse-perseus-json/perseus-parsers/expression-keys.ts
handeyeco: > Can you explain this a bit more?

My concern is that different versions of the widget might support different sets of keys. We could probably get away with a list of keys that is additive (so that it supports all keys that ever existed) but that wouldn't be _technically_ correct if a widget was no longer using the key.

For example if v1 used: `["plus", "minus", "multiply", "divide"]` and v2 used `["plus", "minus", "multiply", "fraction"]` - we could support this with `["plus", "minus", "multiply", "divide", "fraction"]` but that would still be supporting `"divide"` in the types even if we transform `"divide" -> "fraction"` and no longer used "divide" in the widget.

I don't think this really matters though: the parser will handle `"divide"` in v2 widgets but the schema won't (theoretically).

PR #3730, opened 2026-06-05
feat: add `showAngle` field to `LockedPolygonPointType`

mojadem:
## Summary

This is a follow-up to #3705.

- Added `showAngle` to locked polygon points, defaulting to false for parsed legacy data and newly created
points.
- Reused existing polygon angle-label rendering so locked polygons can selectively show interior angle labels
per vertex.
- Added editor controls for toggling angle labels on individual locked polygon points.
- Updated defaults, fixtures, and tests for the new field.

## Test plan
- `pnpm test`


packages/perseus/src/widgets/interactive-graphs/locked-figures/locked-polygon.tsx
```diff
@@ -42,6 +46,29 @@ const LockedPolygon = (props: LockedPolygonType) => {
                     },
                 }}
             />
+            {points.map((point, i) => {
+                const previous = points.at(i - 1);
+                const next = points.at((i + 1) % points.length);
+                // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
```
mojadem, line 52: This is mirroring the pattern in `packages/perseus/src/widgets/interactive-graphs/graphs/polygon.tsx`.

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-polygon-settings.tsx
```diff
@@ -336,7 +350,10 @@ const LockedPolygonSettings = (props: Props) => {
...
                         onClick={() => {
                             props.onChangeProps({
-                                points: [...points, {coord: [0, 0]}],
+                                points: [
+                                    ...points,
+                                    {coord: [0, 0], showAngle: false},
+                                ],
                             });
                         }}
                     >
```
jeremywiebe, line 359: I know this button predates your change, but would you mind adding an `aria-label` to this button (it took me a bit to decipher that it was the "add point" button... right? 😬 ).

packages/perseus/src/widgets/interactive-graphs/locked-figures/locked-polygon.tsx
```diff
@@ -42,6 +46,29 @@ const LockedPolygon = (props: LockedPolygonType) => {
...
+                        key={`locked-polygon-angle-${i}`}
+                        centerPoint={point.coord}
+                        endPoints={[previous.coord, next.coord]}
+                        areEndPointsClockwise={clockwise(
+                            points.map(({coord}) => coord),
+                        )}
+                        showAngles={point.showAngle}
+                        // Locked polygons don't snap; "grid" is chosen as a
+                        // default to preserve decimal angle labels.
+                        snapTo="grid"
```
jeremywiebe, line 68: **thought**: This is an unfortunate instance where we propagate a value from higher up and its meaning is completely different inside this component.  Ideally, this prop would be named something like `roundLabelAngleToInteger` or something.

No change required.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/interactive-graph-widget.ts
```diff
@@ -258,6 +258,7 @@ const parseLockedEllipseType = object({
 
 const parseLockedPolygonPointType = object({
     coord: pairOfNumbers,
+    showAngle: defaulted(boolean, () => false),
```
jeremywiebe, line 261: I think we want to add a regression test item now that we have a new value here.  

See: https://github.com/Khan/perseus/blob/75d06c5a69d70d9783df34f26d5d5575f0068602/packages/perseus-core/src/parse-perseus-json/README.md#regression-testing-against-old-data

jeremywiebe (changes requested):
In the [Chromatic version of Storybook](https://650db21c3f5d1b2f13c02952-uziqcwynpi.chromatic.com/?path=/story/editors-editorpage--demo), the editor layout looks pretty janky. 

Image

Could you double-check that before landing, please?  If you need design help, please ask Caitlyn about the layout.

packages/perseus-editor/src/widgets/interactive-graph-editor/locked-figures/locked-polygon-settings.tsx
```diff
@@ -305,6 +305,20 @@ const LockedPolygonSettings = (props: Props) => {
...
+                                checked={point.showAngle}
+                                size="small"
+                                onChange={(newValue) => {
+                                    const newPoints = [...points];
+                                    newPoints[index] = {
+                                        ...point,
+                                        showAngle: newValue,
+                                    };
+                                    props.onChangeProps({points: newPoints});
+                                }}
```
SonicScrewdriver, line 320: I do think this needs some tweaking in the editor UI, as the option is clipping and doesn't quite match the rest of the styles. Unfortunately we're quite constrained in the width of the editor settings area right now. I'm happy to pair if you want to discuss how we could approach the layout! 

I did notice that the start coordinates handle the extra input for a point by adding a new surrounding box:

Image

packages/perseus/src/widgets/interactive-graphs/locked-figures/locked-polygon.tsx
SonicScrewdriver: We could probably improve the undefined check to avoid the strict boolean expressions :) I think the old logic predated the inclusion of the linter rule

packages/perseus/src/widgets/interactive-graphs/locked-figures/locked-polygon.tsx
SonicScrewdriver: Yeah this does seem like a potential ticket I could throw in our LEMS backlog, since it's a little outside of this PR

packages/perseus/src/widgets/interactive-graphs/__docs__/interactive-graph-initial-state-regression.stories.tsx
```diff
@@ -539,10 +539,10 @@ function lockedFiguresWithWeight(
...
-                {coord: [-6.5, -2.5]},
-                {coord: [-5.5, -3.5]},
-                {coord: [-6.5, -4.5]},
+                {coord: [-7.5, -3.5], showAngle: false},
+                {coord: [-6.5, -2.5], showAngle: false},
+                {coord: [-5.5, -3.5], showAngle: false},
+                {coord: [-6.5, -4.5], showAngle: false},
             ],
             weight,
             color: "pink",
```
SonicScrewdriver, line 548: It'd be great if you could also add a story to this storybook regression test where `showAngle` is on for a locked polygon. These stories get snapshot and checked by Chromatic/Storybook so that we're notified on a PR if anything changes visually, which has been quite helpful for our Interactive Graph Widget.

SonicScrewdriver (changes requested):
Great work! Overall this is nearly there I think we just need to polish up the design a little!

PR #3731, opened 2026-06-05
Add support for centered NumericInputs

handeyeco:
## Summary:

**Go schema change:** https://github.com/Khan/webapp/pull/39970

This PR adds support for centered NumericInputs. This required a change to the schema for NumericInput's widgetOptions:

```diff
- rightAlign?: Boolean;
+ textAlign: "left" | "right" | "center";
```

This meant that I needed to update the widget, editor, parser, stories, and tests. Ultimately it's a small change to the UI and a lot of changes to our safety nets.

~~I decided to go with `start`/`end` vs `left`/`right` to set us up for i18n in the future. However maybe we want to keep `left`/`right` and we can _add_ `start`/`end` later? That would let content creators pick: "this needs to be on the left" or "this can be at the start of whatever the locale wants".~~

Ben asked around and found out that we probably don't want `start`/`end` and in fact do want `left`/`right`.

Issue: LEMS-4144

## Test plan:

Screenshot 2026-06-05 at 1 14 24 PM

Screenshot 2026-06-05 at 2 28 53 PM

Most of this PR is adding/updating tests including:

- Adding a story here: `?path=/story/widgets-numeric-input-visual-regression-tests-initial-state--center-text-align`
- Making the control interactive here: `?path=/docs/widgets-numeric-input--docs`


packages/perseus-core/src/parse-perseus-json/perseus-parsers/numeric-input-widget.test.ts
```diff
@@ -45,3 +45,124 @@ describe("parseSimplify", () => {
         expect(parse("foobar", parseSimplify)).toEqual(anyFailure);
     });
 });
+
+describe("textAlign", () => {
```
handeyeco, line 49: Tests to make sure:
1. We can migrate from `rightAlign` to `textAlign`
2. We can parse `textAlign` correctly

packages/perseus-core/src/parse-perseus-json/perseus-parsers/numeric-input-widget.ts
```diff
@@ -64,32 +67,71 @@ function deprecatedSimplifyValuesToRequired(
...
+    object({
+        answers: parseNumericInputAnswers,
         labelText: optional(string),
         size: string,
         coefficient: defaulted(boolean, () => false),
         rightAlign: optional(boolean),
     }),
 );
+
+function migrateV0ToV1(
```
handeyeco, line 120: Replacing `rightAlign` with `textAlign` is a major change and requires a migration.

packages/perseus-core/src/parse-perseus-json/regression-tests/item-data/numeric-input-center-aligned.ts
```diff
@@ -0,0 +1,51 @@
...
+                            simplify: true,
+                            status: "correct",
+                            strict: false,
+                            value: 1.125,
+                        },
+                    ],
+                    coefficient: false,
+                    labelText: "",
+                    size: "normal",
+                    textAlign: "center",
```
handeyeco, line 41: Test to make sure we can parse the new `textAlign`

packages/perseus-core/src/widgets/input-number/to-numeric-input.test.ts
```diff
@@ -246,4 +246,37 @@ describe("convertInputNumberOptionsToNumericInput", () => {
             "mixed",
         ]);
     });
+
+    it("converts alignment when rightAlign is undefined", () => {
```
handeyeco, line 207: Make sure IN-to-NI still works.

packages/perseus-editor/src/widgets/__tests__/numeric-input-editor.test.tsx
```diff
@@ -67,19 +67,24 @@ describe("numeric-input-editor", () => {
         );
     });
 
-    it("should be possible to select right alignment", async () => {
+    it("should be possible to change text alignment", async () => {
```
handeyeco, line 70: Test to make sure the editor can select the new options

packages/perseus/src/widgets/numeric-input/__docs__/numeric-input-initial-state-regression.stories.tsx
```diff
@@ -46,14 +46,26 @@ export const SizeSmall: Story = {
...
     },
     args: {
         size: "normal",
-        rightAlign: true,
+        textAlign: "end",
+    },
+};
+
+// Verifies the center-aligned text input variant with a pre-filled value — the value should appear in the center
+export const CenterTextAlign: Story = {
```
handeyeco, line 62: Visual regression test for centered NIs.

packages/perseus-core/src/parse-perseus-json/regression-tests/__snapshots__/parse-perseus-json-regression.test.ts.snap
```diff
@@ -14311,13 +14316,131 @@ exports[`parseAndMigratePerseusItem given numeric-input-answer-without-value.ts
...
+              "simplify": "required",
+              "status": "correct",
+              "strict": false,
+              "value": 1.125,
+            },
+          ],
+          "coefficient": false,
+          "labelText": "",
+          "size": "normal",
+          "textAlign": "center",
```
handeyeco, line 15104: This is the important bit here.

packages/perseus-core/src/parse-perseus-json/regression-tests/__snapshots__/parse-perseus-json-regression.test.ts.snap
```diff
@@ -14844,6 +14969,124 @@ Heart rate is described as the number of heart beats per minute.  The normal hum
...
+              "simplify": "required",
+              "status": "correct",
+              "strict": false,
+              "value": 1.125,
+            },
+          ],
+          "coefficient": false,
+          "labelText": "",
+          "size": "normal",
+          "textAlign": "end",
```
handeyeco: This is the expected transformation of the legacy schema: `rightAlign: true` -> `textAlign: end`.

packages/perseus-core/src/parse-perseus-json/regression-tests/item-data/numeric-input-right-aligned.ts
```diff
@@ -0,0 +1,51 @@
...
+                            simplify: true,
+                            status: "correct",
+                            strict: false,
+                            value: 1.125,
+                        },
+                    ],
+                    coefficient: false,
+                    labelText: "",
+                    size: "normal",
+                    rightAlign: true,
```
handeyeco, line 41: Test to make sure we can still parse `rightAlign`

packages/perseus-core/src/data-schema.ts
```diff
@@ -1617,8 +1617,11 @@ export type PerseusNumericInputWidgetOptions = {
      * empty string to mean 1.
      */
     coefficient: boolean;
-    /** Whether to right-align the text or not */
-    rightAlign?: boolean;
+    /**
+     * How to align the text in the input
+     * it's "start"/"end" vs "left"/"right" to support i18n in the future
+     */
+    textAlign: "start" | "end" | "center";
```
handeyeco: I could have done `textAlign?` but I feel like it's easier to just let the parser provide a default so we always have a value.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/numeric-input-widget.test.ts
```diff
@@ -45,3 +45,124 @@ describe("parseSimplify", () => {
...
+        // Act
+        const result = parse(widget, parseNumericInputWidget);
+
+        // Assert
+        assertSuccess(result);
+        expect(result.value.version).toEqual({major: 1, minor: 0});
+        expect(result.value.options.textAlign).toBe("start");
+    });
+
+    it("migrates from v0 to v1 when textAlign is true", () => {
```
jeremywiebe:
```suggestion
    it("migrates from v0 to v1 when rightAlign is true", () => {
```

packages/perseus-core/src/data-schema.ts
jeremywiebe: I like the idea of working to reduce optionality in the schema and have the parser fill in details. I think we found that depending on the React widget components do the defaulting is tricky now that we have server-side scoring, etc.

packages/perseus-editor/src/widgets/numeric-input-editor.tsx
```diff
@@ -456,28 +457,21 @@ class NumericInputEditor extends React.Component<Props, State> {
...
+                    placeholder="Select text alignment"
                 >
-                    Right
-                </Pill>
-            </fieldset>
+                    <OptionItem value="start" label="Left" />
+                    <OptionItem value="center" label="Center" />
+                    <OptionItem value="end" label="Right" />
+                </SingleSelect>
+            </label>
```
jeremywiebe: Not directly related, but I'd love to move away from creating variables with snippets of JSX which we then render at some place in favour of bespoke little components. 

Although we don't use the React Component view much, modelling these as react components helps them show up much better in that component debugger. 

I also think it's a bit more idiomatic as React can manage re-rendering at that level. 

🤷‍♂️   


Something like this: 

```typescript
function AlignmentSelector(props: {
    value: "left" | "right" | "centre",
    onChange: (alignment: string) => void,
}) {
    return (<label>
                Text alignment
                <SingleSelect
                    selectedValue={this.props.textAlign}
                    onChange={(value) => {
                        this.props.onChange({textAlign: value});
                    }}
                    placeholder="Select text alignment"
                >
                    <OptionItem value="start" label="Left" />
                    <OptionItem value="center" label="Center" />
                    <OptionItem value="end" label="Right" />
                </SingleSelect>
            </label>)
}
```

And then ... 

```typescript
<AlignmentSelector 
    value={this.props.textAlign} 
    onChange={(alignment) => {   
        this.props.onChange({textAlign: alignment}); 
    } 
/>
```

packages/perseus/src/styles/perseus-renderer-part-1.css
```diff
@@ -56,6 +56,15 @@ div.graphie {
 .framework-perseus.perseus-mobile .perseus-input-right-align .keypad-input {
     text-align: right;
 }
```
jeremywiebe, line 58: Do these need to be updated? Like `.perseus-input-end-align` or something?

packages/perseus/src/widgets/numeric-input/numeric-input.class.tsx
```diff
@@ -74,7 +74,7 @@ export class NumericInput
 
     static defaultProps: DefaultProps = {
         size: "normal",
-        rightAlign: false,
+        textAlign: "start",
```
jeremywiebe: If the widget options don't make this optional, can this prop be made required here and avoid needing a default?

packages/perseus/src/widgets/numeric-input/numeric-input.module.css
jeremywiebe, line 25: Is this `.end-align` now?

jeremywiebe (commented):
Drive-by comments. I'll leave it to project folks to approve. Thanks for this Matthew!

packages/perseus-core/src/widgets/input-number/to-numeric-input.test.ts
SonicScrewdriver: Greatly appreciate you adding these tests for us!

packages/perseus/src/widgets/numeric-input/__docs__/numeric-input.stories.tsx
```diff
@@ -58,7 +58,7 @@ const meta: Meta<typeof NumericInput> = {
     args: {
         coefficient: false,
         userInput: {currentValue: ""},
-        rightAlign: false,
+        alignment: "start",
```
SonicScrewdriver: Should this be `textAlign`?

SonicScrewdriver (commented):
This generally looks great to me! I just had a comment for numeric-input.stories.tsx, and there's a couple things Jeremy pointed out that we should fix up with the styles. (I'm not sure if IN/NI widgets share any styles with the Expression widget, but hopefully not.)

packages/perseus-core/src/data-schema.ts
benchristel: +1 for making the field required. This way there's no opportunity for different parts of the code to assign different defaults.

benchristel: I'm not so sure about using `"start" | "end"` "for i18n". In Arabic, numbers are left-aligned. https://khanacademy.slack.com/archives/C3SLM30AW/p1780933046797839?thread_ts=1780932840.393159&cid=C3SLM30AW

We probably *don't* want the input's text alignment to follow that of the surrounding text. I think `"right" | "left"` would be my choice for now. It's not hard to change if we find we need `"start" | "end"` later.

handeyeco: I switched back to `left`/`right` vs `start`/`end`.

packages/perseus-editor/src/widgets/numeric-input-editor.tsx
handeyeco: I just decided to inline the JSX because I didn't think it warranted its own component and also I was arguing with TS.

packages/perseus/src/styles/perseus-renderer-part-1.css
handeyeco: No longer relevant since we're not using `start`/`end`.

packages/perseus/src/widgets/numeric-input/numeric-input.module.css
handeyeco: No longer relevant since we're not using `start`/`end`.

benchristel (approved):
LGTM!

packages/perseus/src/widgets/numeric-input/numeric-input.class.tsx
benchristel, line 224: The new version number needs to be added ~~here~~ to WidgetLogic; otherwise, the editor won't produce the correct version.

benchristel (changes requested):
Requesting changes because the widget version number needs to be updated before this lands (see my inline comment). If you rebase this branch on the latest `main`, you'll see test failures related to this.

packages/perseus/src/widgets/numeric-input/numeric-input.class.tsx
benchristel: After [my change to widget versions yesterday](https://github.com/Khan/perseus/pull/3784), you should add the version to the WidgetLogic object.

benchristel (approved):
LGTM!

PR #3733, opened 2026-06-05
[Interactive Graph] Add WB Announcer to Absolute Value Graph

catandthemachines:
## Summary

- Wires the Absolute Value graph's vertex (`movePoint` index 0) and arm-point (`movePoint` index 1) keyboard moves into the WB Announcer — the reducer's `case "absolute-value"` now emits a `move-absolute-value-point` `stateAnnouncement`, and `absolute-value.tsx` passes `ariaLive="off"` to `MovablePoint` so moves aren't double-announced.
- `getAnnouncementText` handles `move-absolute-value-point` via a new `srAbsoluteValuePointLabel` helper: index 0 reads as the vertex (*"Vertex point at X comma Y."*) and index 1 as the arm point (*"Point on arm at X comma Y."*), matching the static aria-labels. An author-supplied custom label (*"Point A at X comma Y."*) overrides the vertex/arm phrasing when one is set — mirroring sinusoid's `pointLabels` handling.
- The same-x rejection (which keeps the slope defined) happens before the announcement, so a rejected move emits no `stateAnnouncement`.
- TODOs reference [LEMS-4189](https://khanacademy.atlassian.net/browse/LEMS-4189), which tracks removing `aria-live` from `useControlPoint` once every graph is wired to the announcer.

Issue: LEMS-4195

### Automated
- [x] `pnpm tsc` — passes
- [x] `pnpm lint` (changed files) — passes
- [x] `pnpm jest packages/perseus/src/widgets/interactive-graphs` — passes (reducer, screenreader-text, and absolute-value suites)

### Manual (reviewer)
- [ ] Open the **Interactive Graph → Absolute Value** story in Storybook with VoiceOver/NVDA (or the a11y addon's live-region inspector).
- [ ] Arrow-move the vertex: confirm you hear *"Vertex point at X comma Y."* **once** — not doubled.
- [ ] Arrow-move the arm point: confirm you hear *"Point on arm at X comma Y."* once.
- [ ] With custom point labels set, confirm a labeled point announces *"Point A at …"* on move.
- [ ] Regression: open **Sinusoid**, **Linear**, and **Ray** and confirm their announcements are unchanged.

[LEMS-4189]: https://khanacademy.atlassian.net/browse/LEMS-4189?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ivyolamit (approved):

packages/perseus/src/widgets/interactive-graphs/graphs/screenreader-text.test.ts
```diff
@@ -215,6 +215,60 @@ describe("getAnnouncementText", () => {
...
 
+    describe("move-absolute-value-point", () => {
+        // Coord layout: [vertex(0), arm point(1)]. The vertex uses the
+        // vertex label; the arm point uses the second-point label.
+        it("uses the vertex label for index 0", () => {
+            const result = getAnnouncementText(
+                {
+                    type: "move-absolute-value-point",
+                    pointIndex: 0,
+                    pointLabel: 1,
```
benchristel, line 225: What does it mean when `pointLabel` is a number? This seems confusing.

packages/perseus/src/widgets/interactive-graphs/types.ts
```diff
@@ -160,6 +160,17 @@ type MoveSinusoidPointAnnouncement = {
     otherY: number;
 };
 
+// Absolute-value graph: the vertex (index 0) and the point on the arm
+// (index 1) use different labels, chosen by index — mirroring the static
+// aria-labels in absolute-value.tsx.
+type MoveAbsoluteValuePointAnnouncement = {
+    type: "move-absolute-value-point";
+    pointIndex: number;
+    pointLabel: string | number;
```
benchristel, line 183: It looks like pointLabel has no effect unless it is a string. I think the type should be:


```suggestion
    pointLabel?: string;
```

packages/perseus/src/widgets/interactive-graphs/graphs/screenreader-text.test.ts
catandthemachines: `pointIndex` is used to determine the "type" of point is being used, while the pointLabel can be a number (index +1 by default) or a custom label.

ivyolamit: We also have a follow-up ticket here:
https://khanacademy.atlassian.net/browse/LEMS-4206

PR #3737, opened 2026-06-06
[Color Sync] Add more regression stories for the Definition widget

nishasy:
## Summary:
Add stories for
- Definition in Hint (initial + interaction)
- Definition in Explanation (initial + interaction)
- Definition in Graded Group (initial + interaction)
- Definition in both main article and in explanation widget (interaction)
- Mobile Definition (initial + interaction)

+ Added a comment about the hover state existing but not being testable.

Issue: https://khanacademy.atlassian.net/browse/LEMS-4129

## Test plan:
Storybook
- `/?path=/story/renderers-hints-renderer-visual-regression-tests-initial-state--definition-widget-in-hint`
- `/?path=/story/renderers-hints-renderer-visual-regression-tests-interactions--definition-widget-in-hint`
- `/?path=/story/widgets-nested-widgets-visual-regression-tests-initial-state--explanation-with-definition`
- `/?path=/story/widgets-nested-widgets-visual-regression-tests-initial-state--graded-group-with-radio-and-definition`
- `/?path=/story/widgets-nested-widgets-visual-regression-tests-interactions--definition-in-content-and-explanation`
- `/?path=/story/widgets-nested-widgets-visual-regression-tests-interactions--explanation-with-definition`
- `/?path=/story/widgets-nested-widgets-visual-regression-tests-interactions--graded-group-definition-clicked`
- `/?path=/story/widgets-definition-visual-regression-tests-initial-state--mobile`
- `/?path=/story/widgets-definition-visual-regression-tests-interactions--mobile-clicked-state`


packages/perseus/src/widgets/definition/__docs__/definition-interactions-regression.stories.tsx
```diff
@@ -19,33 +22,48 @@ const meta: Meta = {
...
+    play: async ({canvas, userEvent}) => {
+        const definitionTrigger = canvas.getByRole("button", {
+            name: "Definition of: Allies",
+        });
+        await userEvent.click(definitionTrigger);
+    },
+};
+
+export const MobileClickedState: Story = {
+    decorators: [definitionRendererDecorator, mobileDecorator],
```
Myranae: Claude pointed out how the definitionRendererDecorator ends up being applied twice in this story since you have the decorator at the meta level already. It suggests removing it from here, but if you need to have the two listed together here, then you may want to move it out of the meta and just have each story list its own decorator.

packages/perseus/src/widgets/definition/__docs__/definition-renderer-decorator.tsx
```diff
@@ -0,0 +1,40 @@
...
+    generateDefinitionWidget,
+    generateTestPerseusRenderer,
+} from "@khanacademy/perseus-core";
+import * as React from "react";
+
+import QuestionRendererForStories from "../../__testutils__/question-renderer-for-stories";
+
+import type {APIOptions} from "../../../types";
+
+export const definitionRendererDecorator = (
```
Myranae:
```suggestion
import type {APIOptions} from "../../../types";
import type {Decorator} from "@storybook/react-vite";

export const definitionRendererDecorator: Decorator = (
```

I don't think this is the right place for that import statement, but the lint fix option should put it in the right place :)

packages/perseus/src/widgets/definition/__docs__/definition-renderer-decorator.tsx
```diff
@@ -0,0 +1,40 @@
...
+    _,
+    {
+        args,
+        parameters,
+    }: {
+        args: Record<string, unknown>;
```
Myranae:
```suggestion
        args: Partial<DefinitionDefaultWidgetOptions>;
```

Would still need an import of `DefinitionDefaultWidgetOptions` from "@khanacademy/perseus-core" also

Myranae (approved):
Looks good! Left some small fix comments. The mobile input focused story still has small font for the definition mobile initial state, but the mobile clicked state interactions story font looks good. Was that

mobile-definition-input-focused

nishasy:
@Myranae The small font in the mobile storybook is consistent with the experience on the mobile app. I think this is something that will get fixed with the color sync project when we update the mobile font size.

PR #3740, opened 2026-06-08
Add Claude Code preview config for Storybook

Myranae:
## Summary:
Adds `.claude/launch.json` so Claude Code web sessions can launch and preview Storybook (port 6006) with a single click, mirroring what wonder-blocks did in Khan/wonder-blocks#3097. Also makes the preview panel available in the Claude Code Desktop app.

This allows you to click on specific elements in the preview panel, annotate specific elements, refresh, go back and forth, all from within the Claude app. You can share the elements you select and your annotations with Claude in the chat!

image

## Test plan:
- Checkout the branch
- Point a Claude Code Desktop conversation at it
- Check the Preview pane (shift+cmd+P). It should load Storybook.


.claude/launch.json
```diff
@@ -0,0 +1,12 @@
+{
+    "version": "0.0.1",
+    "configurations": [
+        {
+            "name": "Storybook",
+            "runtimeExecutable": "pnpm",
+            "runtimeArgs": ["storybook"],
+            "port": 6006,
+            "autoPort": true
```
jeremywiebe, line 9: We probably want this `false` (https://code.claude.com/docs/en/desktop#port-conflicts).  Storybook prompts to use a different port if 6006 is in use, but that's almost always an error or means that Storybook is already running.

jeremywiebe (approved):

.claude/launch.json
Myranae: Hmm, Kevin mentioned this might be good if you use worktrees, and I use worktrees all the time. I was assuming it would allow us to use the Preview pane for each worktree and have it look at the stories from each worktree

Myranae: Discussed in Slack. Keeping it for now to assist with worktrees. Will remove if doesn't work well or causes issues for folks not using worktrees.

PR #3743, opened 2026-06-08
Convert Grapher widgets to Interactive Graph

benchristel:
This PR makes it so when the `grapher-to-interactive-graph` feature flag
is on, we render Grapher widgets with the interactive graph components.
From the learner's perspective, the Grapher widget appears to have been
replaced by Interactive Graph.

There are two cases where we can't convert Grapher to Interactive Graph:

- Quadratic graphs. The Grapher data and UI use two movable points (the
  vertex and a second point) to determine the quadratic function.
  IG uses three arbitrary points to determine the quadratic function. This
  makes it impossible to losslessly convert IG user input to Grapher user
  input.
- Choose-your-own-function graphs. IG has no equivalent for this case.

Issue: LEMS-4258

## Test plan:

Review Grapher widgets in Test Everything and elsewhere. When the feature
flag is on, they should display as Interactive Graphs. Validation and
scoring should work.

### DIFF
--- a/packages/perseus-core/src/widgets/grapher/to-interactive-graph.ts
+++ b/packages/perseus-core/src/widgets/grapher/to-interactive-graph.ts
@@ -3,6 +3,7 @@ import invariant from "tiny-invariant";
 
 import type {
     GrapherAnswerTypes,
+    GrapherFunctionType,
     PerseusGrapherWidgetOptions,
     PerseusGraphType,
     PerseusInteractiveGraphWidgetOptions,
@@ -29,6 +30,10 @@ export function convertGrapherOptionsToInteractiveGraph(
         return null;
     }
 
+    const graph: PerseusGraphType = {
+        type: grapherFunctionTypeToInteractiveGraphType(type),
+    };
+
     return {
         step: grapherOptions.graph.step,
         gridStep: grapherOptions.graph.gridStep,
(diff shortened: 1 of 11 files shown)


packages/perseus-core/src/widgets/grapher/to-interactive-graph.ts
```diff
@@ -0,0 +1,184 @@
...
+        case "quadratic":
+            return {
+                type: "quadratic",
+                coords:
+                    grapherAnswerTypes.coords == null
+                        ? null
+                        : grapherQuadraticCoordsToInteractiveGraphCoords(
+                              grapherAnswerTypes.coords,
+                          ),
+            };
```
benchristel: We probably don't want this case, since there's no way to map user input for an iG quadratic graph back to the Grapher format.

packages/perseus/src/widgets/grapher/grapher.tsx
```diff
@@ -606,6 +624,36 @@ class Grapher extends React.Component<Props> implements Widget {
             </div>
         );
     }
+
+    render(): React.ReactNode {
+        if (!this.props.apiOptions.flags?.["grapher-to-interactive-graph"]) {
+            return this.renderLegacyGrapher();
+        }
+
+        const interactiveGraphOptions = convertGrapherOptionsToInteractiveGraph(
```
SonicScrewdriver, line 633: I _really_ love how simple our logic works out for these conversions

SonicScrewdriver (approved):
Honestly this is gorgeous, I don't see anything to even suggest! Everything seems to score correctly in Storybook, I think this is pretty much ready for QE

PR #3744, opened 2026-06-08
[ColorSync][math-input] Convert math-input font and colors to semantic tokens

Myranae:
## Summary:
Part of the ColorSync migration. Converts hardcoded color and font values in the math-input package to WonderBlocks semantic tokens. See images below for color comparisons and affected widget areas.

math-input-color-comparison

### Main Changes Visible
**Keypad Button Pressed (outline and background)**
Screenshot 2026-06-10 at 4 01 13 PM
image

**Cursor Handle**
Screenshot 2026-06-10 at 10 13 44 AM

**Navigation Button Pressed**
Screenshot 2026-06-10 at 10 13 34 AM

Issue: LEMS-4244

## Test plan:
- [x] Confirm all checks pass
- [x] Chromatic diffs reviewed and approved — 3 expected stories:
  - `CursorHandleVisible` (new regression story, cursor handle fill/stroke)
  - `KeypadButtonPressed` (pressed state background + border)
  - `NavigationButtonPressed` (pressed state border)


Myranae:
@claude review once

packages/math-input/src/__docs__/math-input-initial-state-regression.stories.tsx
```diff
@@ -0,0 +1,47 @@
+import * as React from "react";
+
+import {themeModes} from "../../../../.storybook/modes";
+import CursorHandle from "../components/input/cursor-handle";
+
+import type {Meta, StoryObj} from "@storybook/react-vite";
+
+const meta: Meta<typeof CursorHandle> = {
+    title: "Math Input/Components/Visual Regression Tests/Initial State",
+    component: CursorHandle,
```
nishasy: Would this still be considered general math input initial state regression story if the main component being tested is CursorHandle? I almost feel like this should be `cursor-handle-regression.stories.tsx` instead?

packages/math-input/src/__docs__/math-input-interactions-regression.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+import {themeModes} from "../../../../.storybook/modes";
+import NavigationPad from "../components/keypad/navigation-pad";
+
+import TintedBackgroundDecorator from "./tinted-background-decorator";
+
+import type {Meta, StoryObj} from "@storybook/react-vite";
+
+const meta: Meta<typeof NavigationPad> = {
+    title: "Math Input/Components/Visual Regression Tests/Interactions",
+    component: NavigationPad,
```
nishasy: Same comment here but for `NavigationPad`.

packages/math-input/src/components/keypad/keypad-button.tsx
```diff
@@ -108,9 +114,9 @@ const styles = StyleSheet.create({
         boxShadow: "none",
     },
     pressed: {
-        border: "2px solid #1B50B3",
+        border: `2px solid ${semanticColor.core.border.instructive.strong}`,
         padding: 0,
-        background: `linear-gradient(0deg, rgba(24, 101, 242, 0.32), rgba(24, 101, 242, 0.32)), ${semanticColor.core.background.base.default}`,
```
nishasy, line 113: oh whoa, what was this linear gradient doing before?

packages/perseus/src/widgets/expression/__docs__/expression-interactions-regression.stories.tsx
```diff
@@ -11,6 +12,28 @@ const meta: Meta<typeof Expression> = {
     title: "Widgets/Expression/Visual Regression Tests/Interactions",
     component: Expression,
     tags: ["!autodocs", "!manifest"],
+    decorators: [
+        (Story) => {
+            // MathQuill toggles `mq-blink` via setInterval, making snapshots
+            // non-deterministic. The blink rules in main.css use `!important`
```
nishasy: I love a good use of `non-deterministic`, but can we say `flakey` here so it's more immediately obvious what the problem is?

Or maybe just add a sentence above this that says "text field cursor blinking can make snapshots flakey"

packages/math-input/src/__docs__/math-input-initial-state-regression.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+        chromatic: {disableSnapshot: false, modes: themeModes},
+    },
+};
+export default meta;
+
+type Story = StoryObj<typeof CursorHandle>;
+
+// This component is touch-only in production and never reached via mouse
+// interactions, so it is tested here in isolation.
+export const CursorHandleVisible: Story = {
```
nishasy: I don't think I see this in the UI Tests check on your PR?

packages/perseus/src/widgets/expression/__docs__/expression-interactions-regression.stories.tsx
```diff
@@ -139,6 +162,17 @@ export const WithTextInField: Story = {
     },
 };
 
+export const KeypadButtonPressed: Story = {
```
nishasy, line 164: I don't think I see this in the UI Tests check on your PR either?

packages/math-input/src/__docs__/math-input-initial-state-regression.stories.tsx
Myranae: I'd be okay with renaming the file, but I kind of still want it to be in the math input components visual regressions folder. Would you want these to not be considered initial state / interactions, or would it stay that and we would just change the file name?

packages/math-input/src/components/keypad/keypad-button.tsx
Myranae: It was a semi-transparent blue tint composited over white for the pressed state, replaced with `background.instructive.subtle`.  Checked with Caitlyn about this and she wants rgba converted to tokens. I wonder if this gradient would still work if we just replaced the rgba instead of the whole gradient. I'll check!

Myranae: > The original was a "non-gradient gradient" — both stops were identical (rgba(24, 101, 242, 0.32) repeated twice), so there was no actual color transition. It was a flat 32%-opacity blue overlay on top of the base white background. The linear-gradient syntax was used here as a CSS layering trick, not for an actual gradient effect.

So it sounds like even though a linear gradient was in the code, it wasn't actually a gradient. It was just a blue overlay on top of white, and then the color that's used now is an approximate blue of the same kind. It sounds like the background semantic token now is just a little bit lighter, but it's the semantic version of that similar color. We have a Slack thread with Caitlyn about this where I show the difference here if you're interested :) 

Also sounds like the gradient won't work with tokens, and it's not super semantically clear keeping the gradient, so this is the most semantically correct and designer approved option for now.

packages/perseus/src/widgets/expression/__docs__/expression-interactions-regression.stories.tsx
Myranae: Added a sentence!

packages/perseus/src/widgets/expression/__docs__/expression-interactions-regression.stories.tsx
Myranae: It's an ancestor of the current build. Had a couple updates along the way, but checked the snapshots at each step. There's a [list of the Chromatic builds for the branch](https://www.chromatic.com/builds?appId=650db21c3f5d1b2f13c02952&branch=tb%2FLEMS-4244%2Fmath-input-color-sync)! Just found this now, but it seems pretty handy! :D Thanks for asking!

packages/math-input/src/__docs__/math-input-initial-state-regression.stories.tsx
Myranae: See comment below.

packages/math-input/src/__docs__/math-input-initial-state-regression.stories.tsx
nishasy: I think this is fine to stay in the math input folder, but I think we should either
- not declare the component/type here as `CursorHandle`. Maybe we can follow the non-typed meta format that's in the nested widget regression stories?

OR
- put this under math input, but break it out into CursorHandle-specific regression stories.

Maybe Mark will have thoughts on this when he's back tomorrow?

mark-fitzgerald: I'm open to what you all feel makes sense. If it helps, I am leaning towards using the non-typed meta format and continue to include all regression stories in here. I also think it would make sense to move the regression stories out of "Components" up one level, so that the regression stories are for everything related to the math input.

packages/math-input/src/components/keypad/navigation-button.tsx
```diff
@@ -120,8 +120,8 @@ const styles = StyleSheet.create({
         boxShadow: "none",
     },
     pressed: {
-        border: "2px solid #1B50B3",
-        background: `linear-gradient(0deg, rgba(24, 101, 242, 0.32), rgba(24, 101, 242, 0.32)), ${semanticColor.core.background.base.default}`,
+        border: `2px solid ${semanticColor.core.border.instructive.strong}`,
```
mark-fitzgerald: Since we are here, I recommend using the WB border thickness token, as well.

packages/perseus/src/widgets/expression/__docs__/expression-interactions-regression.stories.tsx
```diff
@@ -11,6 +12,29 @@ const meta: Meta<typeof Expression> = {
...
+            // inside `@layer shared`; CSS layers reverse !important priority
+            // (layered beats unlayered), so our overrides must be inside the
+            // same layer.
+            React.useLayoutEffect(() => {
+                const style = document.createElement("style");
+                style.textContent = `
+                    @layer shared {
+                        :root .mq-cursor.mq-blink { visibility: visible !important; }
+                        :root .keypad-input .mq-editable-field .mq-cursor { transition: none !important; }
+                        :root .keypad-input .mq-editable-field .mq-cursor.mq-blink { opacity: 1 !important; }
```
mark-fitzgerald: Is `:root` needed here?

mark-fitzgerald (approved):

packages/math-input/src/__docs__/math-input-initial-state-regression.stories.tsx
Myranae: Okay, changed it to the non-typed meta format and moved it out of the components folder up a level :)

packages/math-input/src/__docs__/math-input-interactions-regression.stories.tsx
Myranae: Same fix as above :)

packages/math-input/src/components/keypad/navigation-button.tsx
Myranae: Updated~

packages/perseus/src/widgets/expression/__docs__/expression-interactions-regression.stories.tsx
Myranae: It looks like it is needed on the first one to win specificity-wise, but the others are just to be consistent across all three. Do you have a preference towards removing the two unnecessary ones over having it consistent across all three?
image

mark-fitzgerald: I don't think it's needed on the first one, either. The `!important` flag should cause it to win out over specificity.

packages/math-input/src/__docs__/math-input-initial-state-regression.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+                    "Regression tests for math-input components that do NOT " +
+                    "need any interactions to test.",
+            },
+        },
+        chromatic: {disableSnapshot: false, modes: themeModes},
+    },
+};
+export default meta;
+
+type Story = StoryObj<typeof CursorHandle>;
```
nishasy: Since there's potential for other types being tested here later, could you rename this to `CursorHandleStory`?

Similar to these in the nested widget stories:
- https://github.com/Khan/perseus/blob/dd8d25c7df584b86f96d1011125a5eba9d423b79/packages/perseus/src/widgets/__docs__/nested-widgets-initial-state-regression.stories.tsx#L37
- https://github.com/Khan/perseus/blob/dd8d25c7df584b86f96d1011125a5eba9d423b79/packages/perseus/src/widgets/__docs__/nested-widgets-initial-state-regression.stories.tsx#L53

packages/math-input/src/__docs__/math-input-interactions-regression.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+        ),
+        TintedBackgroundDecorator,
+    ],
+    args: {
+        onClickKey: action("onClickKey"),
+    },
+};
+export default meta;
+
+type Story = StoryObj<typeof NavigationPad>;
```
nishasy: same here

nishasy (approved):
Yay, thanks for the updates! I left one more small suggestion, but this looks good to me.

packages/perseus/src/widgets/expression/__docs__/expression-interactions-regression.stories.tsx
Myranae: Thanks so much! Updated :)

packages/math-input/src/__docs__/math-input-initial-state-regression.stories.tsx
Myranae: Great point! Updated :)

PR #3745, opened 2026-06-08
[Interactive Graph] Improve graph instruction and add repeat shortcut

ivyolamit:
## Summary:
Reworks the screen reader-only interactive-graph instructions to lead with a focus-mode warning and end with an on-demand repeat hint, and adds the Insert + I shortcut that repeats the instructions while the graph is focused.

Screenshot 2026-06-12 at 2 31 51 PM

Issue: LEMS-4120

## Test plan:

### DIFF
--- a/packages/perseus/src/widgets/interactive-graphs/__docs__/notes/screen-reader.md
+++ b/packages/perseus/src/widgets/interactive-graphs/__docs__/notes/screen-reader.md
@@ -285,7 +285,7 @@ Since this implementation is an improvement on existing functionality no QE is n
 - `role="figure"` on the outer focusable `mafs-graph` View (LEMS-4119; OQ6a primary, OQ6b decision A).
 - DOM reorder so the instructions `<View>` is the first sr-only child inside the graph wrapper (LEMS-4121; OQ4).
 - Bounded vs. unlimited instruction-string split (LEMS-4120; OQ2a decision B), including the leading mode-toggle warning (OQ2b decision A) and the on-demand-shortcut sentence (OQ2c decision A).
-- Insert + I (Windows) / Fn + Enter + I (macOS) keyboard shortcut to repeat the instruction string on demand (LEMS-4120; OQ2c decision A). Platform-detected dual binding.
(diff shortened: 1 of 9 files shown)


catandthemachines:
Just adding a note that this short cut isn't even working in Windows + NVDA. We'll need to investigate and troubleshoot this a bit more to ensure we have a version that works.

packages/perseus/src/widgets/interactive-graphs/__docs__/notes/screen-reader.md
```diff
@@ -285,7 +285,7 @@ Since this implementation is an improvement on existing functionality no QE is n
 - `role="figure"` on the outer focusable `mafs-graph` View (LEMS-4119; OQ6a primary, OQ6b decision A).
 - DOM reorder so the instructions `<View>` is the first sr-only child inside the graph wrapper (LEMS-4121; OQ4).
 - Bounded vs. unlimited instruction-string split (LEMS-4120; OQ2a decision B), including the leading mode-toggle warning (OQ2b decision A) and the on-demand-shortcut sentence (OQ2c decision A).
-- Insert + I (Windows) / Fn + Enter + I (macOS) keyboard shortcut to repeat the instruction string on demand (LEMS-4120; OQ2c decision A). Platform-detected dual binding.
+- Insert + I keyboard shortcut to repeat the instruction string on demand (LEMS-4120; OQ2c). **Windows-only** — the originally-planned Fn + Enter + I macOS binding was dropped during implementation: Macs have no Insert key (Fn + Return reports as `key: "Enter"`, never "Insert") and macOS VoiceOver doesn't use Insert as a modifier, so the chord can't fire on a Mac. Mac SR users re-focus the graph to re-hear the instructions. No platform detection needed.
```
handeyeco, line 288: Should we really be doing something that's only supported on Windows and not Mac?

Are we identifying if the user is on Windows/Mac to conditionally change the instructions? (I didn't see that when doing a rough pass over the PR)

How does this affect development and QE? Adding a feature that many people won't be able to test because we don't have Windows?

I would recommend not adding a feature that's only available to one OS, but if we still decide to, this should probably be a bigger conversation in LEMS.

ivyolamit:
Based on team [discussion](https://khanacademy.slack.com/archives/C0AHH9H21H7/p1781715047436629?thread_ts=1780940625.179129&cid=C0AHH9H21H7) we will hold off implementation for future work.

PR #3746, opened 2026-06-08
Require the `sinusoid` and `tangent` subtypes of interactive graphs to have exactly two points.

benchristel:
## Summary:
This will make it slightly easier to adapt the Grapher widget to the
Interactive Graph UI. Grapher uses tuple types for its point coordinates.

Issue: LEMS-4252

## Test plan:

Search the content for interactive graphs, and verify that all sinusoid
and tangent graphs have exactly two points.

```
rg -l '"type":"interactive-graph"' \
  | xargs jq --indent 0 '
      ..
      | objects
      | select(.type == "sinusoid" or .type == "tangent")
      | {
          coords: (if .coords == null then null else .coords | length end),
          startCoords: (if .startCoords == null then .startCoords else .startCoords | length end)
        }' \
  | sort \
  | uniq -c

8455 {"coords":2,"startCoords":null}
 102 {"coords":null,"startCoords":2}
3443 {"coords":null,"startCoords":null}
```

### DIFF
--- a/packages/perseus-core/src/utils/generators/interactive-graph-widget-generator.test.ts
+++ b/packages/perseus-core/src/utils/generators/interactive-graph-widget-generator.test.ts
@@ -784,12 +784,10 @@ describe("generateIGSinusoidGraph", () => {
             coords: [
                 [0, 0],
                 [1, 1],
-                [2, 2],
             ],
             startCoords: [
                 [0, 0],
                 [1, 1],
-                [2, 2],
             ],
         });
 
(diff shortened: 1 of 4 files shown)


packages/perseus-core/src/utils/generators/interactive-graph-widget-generator.test.ts
```diff
@@ -784,12 +784,10 @@ describe("generateIGSinusoidGraph", () => {
             coords: [
                 [0, 0],
                 [1, 1],
-                [2, 2],
             ],
             startCoords: [
                 [0, 0],
                 [1, 1],
-                [2, 2],
```
benchristel, line 792: These test inputs (accidentally?) had an extra point.

SonicScrewdriver (approved):
Love it! Thanks for checking the content as well.

PR #3752, opened 2026-06-09
[Interactive Graph] Remove aria-live from interactive-graph in favor of using the WB Annoucer recently added logic

ivyolamit:
## Summary:
Interactive graphs historically announced point/line moves to screen readers via an `aria-live` attribute on the focusable handle: when a handle moved, its `aria-label` ("Point 1 at x, y") updated and the live region read it out. That mechanism has been superseded by the WB Announcer: each graph's reducer now sets a `stateAnnouncement`, which `stateful-mafs-graph.tsx` consumes via `announceMessage(...)`. This is centralized and gives us per-graph control over the spoken copy.

The changes include:
- removed aria-live plumbing
- deleted the now-unused `AriaLive`
- updated existing docs for latest information
- For 13 of the 15 affected graphs this is a pure no-op — they already passed `ariaLive="off"`, and `aria-live="off"` is equivalent to the attribute being absent.
- **Circle** (radius point) and **vector** are the exceptions: they never opted out, so they rode the `"polite"` default *and* emitted a `stateAnnouncement` — i.e. they **double-announced** before this PR. After it, they announce once, through the WB Announcer only:
- circle → `move-center` / `move-radius-point`
- vector → `move-vector-point` / `move-vector-line`

Issue: LEMS-4189

## Test plan:


catandthemachines (approved):
This implementation looks correct and I can confirm that I don't see any other usage or implementation of `aria-live` on our graphs. 🥳

PR #3754, opened 2026-06-09
[Interactive Graph] Update the order of the graph type drop down

Evelas78:
## Summary:
Changed order of graph type dropdown items, renamed sinusoid and logarithm according to ticket, set default to None.

Issue: https://khanacademy.atlassian.net/browse/LEMS-4208

## Test plan:
- Verify order of dropdown starts with one, followed by Absolute Value. Exponential Function should be right below Tangent Function. Linear system should be followed by line segments then points
- Verify Sinusoidal function is found (name exact)
- Verify Logarithmic function is found (name exact)
- Default option should be None
- All words after the first word in each option of the dropdown should be lower case


packages/perseus-editor/src/widgets/interactive-graph-editor/components/graph-type-selector.tsx
```diff
@@ -16,19 +16,22 @@ const GraphTypeSelector = (props: GraphTypeSelectorProps) => {
             placeholder="Select an answer type"
             style={styles.singleSelectShort}
         >
-            <OptionItem value="absolute-value" label="Absolute value" />
             <OptionItem value="none" label="None" />
+            <OptionItem
+                value="absolute-value"
+                label="Absolute value function"
+            />
             <OptionItem value="linear" label="Linear function" />
```
ivyolamit, line 25: Since we're already grouping the functions after the None. Let's alphabetize them. Then below the functions will be the rest of the graph types still in alphabetic order. You can also show a screenshot and respond to the slack message in the ticket that Emily started in this way we can confirm if it's better order for them.

ivyolamit (commented):
@Evelas78 overall PR looks good, see my comment below and let's just confirm with Emily if she's happy with the order.

packages/perseus-editor/src/widgets/interactive-graph-editor/components/graph-type-selector.tsx
Evelas78: Part of the ticket was grouping functions logically (ie, reversing the order of exponential and tangent so that tangent + sinusoidal were next to each other and exponentional + logarthmic were next to each other). I think alphabetizing would be against that request, but I can still bring it up

ivyolamit (approved):

packages/perseus-editor/src/widgets/interactive-graph-editor/components/graph-type-selector.tsx
ivyolamit: based on content author's response:
> the team prefers alphabetical. So None, then the functions in alphabetical order, then everything else in alphabetical order.

PR #3759, opened 2026-06-10
Disallow changes to parser test data files

nishasy:
## Summary:
All the parser test data files have a warning telling people not to modify them,
but that doesn't fully protect them from modifications.

Adding a Github check here to fail in PRs where anyone modifies any of these files.
New test data files can be added.

Issue: https://khanacademy.atlassian.net/browse/LEMS-4223

## Test plan:
- Make a change to an `item-data` file and push it to this PR.
- Confirm that there is a failure in the Github checks.
- Remove the change to the `item-data` file.
- Confirm the PR has no failures.


.github/workflows/protect-regression-data.yml
```diff
@@ -0,0 +1,44 @@
+name: Protect Regression Data
```
jeremywiebe: important distinction :) 

```suggestion
name: Protect Parser Regression Data
```

.github/workflows/protect-regression-data.yml
```diff
@@ -0,0 +1,44 @@
...
+
+# The Perseus JSON regression-test fixtures under
```
jeremywiebe:
```suggestion
# The Perseus parser regression-test fixture data under
```

utils/check-frozen-regression-data.sh
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bash
+
+# For use in Github Actions.
```
jeremywiebe: tbh, I'd probably just inline this into the action. I'm not sure there's value in having this external. 🤔

utils/check-frozen-regression-data.sh
```diff
@@ -0,0 +1,53 @@
...
+#   D = deleted, M = modified, R = renamed, C = copied, T = type changed.
+# Added (A) is intentionally excluded — adding new fixtures is allowed.
+#
+# The `A...B` (three-dot) range diffs B against the merge-base of A and B, i.e.
+# exactly the changes this PR introduces on top of the base branch.
+OFFENDING=$(git diff --name-status --diff-filter=DMRCT \
+    "$BASE_SHA...$HEAD_SHA" -- "${PROTECTED_DIRS[@]}")
+
+if [ -n "$OFFENDING" ]; then
+    echo "::error title=Frozen regression fixture changed::Existing regression-test fixtures may not be modified, renamed, or deleted."
```
jeremywiebe: I'd really like to make sure we keep talking about these as "parsing regression fixtures" - parsing is the only place we care about this. :) 

A I think linking to this part of the parser README would help:
 https://github.com/Khan/perseus/tree/main/packages/perseus-core/src/parse-perseus-json#regression-testing-against-old-data


```suggestion
    echo "::error title=Frozen parser regression fixture changed::Existing regression-test fixtures may not be modified, renamed, or deleted."
```

utils/check-frozen-regression-data.sh
```diff
@@ -0,0 +1,53 @@
...
+    echo ""
+    echo "The following protected files were changed:"
+    echo "$OFFENDING" | sed 's/^/    /'
+    echo ""
+    echo "These fixtures capture historical Perseus data formats and must stay"
+    echo "stable so the parser is always tested against real, published data."
```
jeremywiebe:
```suggestion
    echo "These fixtures capture historical Perseus data formats and must never"
    echo "change so the parser is always tested against real, published data."
```

jeremywiebe (approved):

packages/perseus-core/src/parse-perseus-json/regression-tests/item-data/interactive-graph-locked-figures-missing-weights.ts
```diff
@@ -32,7 +32,7 @@ export default {
                         {
                             type: "point",
                             coord: [-9, 9],
-                            color: "blue",
+                            color: "green",
```
benchristel: I assume you're going to revert this before landing :)

benchristel (approved):
Very cool! Thanks for setting this up.

packages/perseus-core/src/parse-perseus-json/regression-tests/item-data/interactive-graph-locked-figures-missing-weights.ts
nishasy: Yes!

PR #3760, opened 2026-06-10
Delete most InputNumber code, replace with NumericInput

benchristel:
## Summary:
The input-number widget has been deprecated for... a decade?  In this PR, we
replace the input-number editor and component with the numeric-input ones.  To
make this a drop-in replacement, we migrate the input number data to a new v1
schema in the parser. The v1 schema is identical to the numeric-input schema.

Issue: LEMS-4112

## Test plan:

Deploy the dev build to a ZND. Test that input-number widgets can be used and
scored.  Test editing: existing input-numbers should display the numeric-input
editor (but the placeholder in the content will still say input-number).

- [x] incorporate Matthew's textAlign changes
- [x] deploy Go schema changes
- [x] change major version number output by editor to 1.
- [x] make input-number data-schema types alias numeric-input types
- [x] address all TODO and FIXME comments
- [x] ensure the perseus service no longer calls `scorePerseusItemWithInputNumberAsNumericInput`


packages/perseus-core/src/data-schema.ts
```diff
@@ -2248,6 +2249,45 @@ export type PerseusInputNumberWidgetOptions = {
...
+     * The maximum difference between the answer key `value` and a correct
+     * response.
+     */
+    maxError?: number | undefined;
+    /** Determines how unsimplified responses are handled */
+    simplify: PerseusNumericInputSimplify;
+};
+
+export type PerseusInputNumberWidgetOptions = {
+    answers: [PerseusInputNumberAnswer];
```
benchristel: I'm trying to be a bit too clever here by restricting `answers` to a single element, and by forcing the answer's `status` to "correct," `strict` to `true`, etc. I did this because the fields will always have these values when we migrate them from v0 input-numbers. The problem is, the numeric-input editor can produce configurations different from this. I'll have to change these types to be exactly the same as the numeric-input types... or we could just reuse the numeric-input types.

benchristel:
Before this lands, we need to update the Go schema: https://github.com/Khan/webapp/pull/39496

packages/perseus-score/src/widgets/input-number/score-input-number.test.ts
```diff
@@ -25,75 +25,68 @@ describe("scoreInputNumber", () => {
...
+                {
+                    status: "correct",
+                    value: 241.90263432641407,
+                    maxError: 0,
+                    simplify: "required",
+                    answerForms: [],
+                    message: "",
+                    strict: true,
+                },
+            ],
```
benchristel, line 84: This test and the ones below could use the generator functions.

packages/perseus/src/widget-ai-utils/input-number/input-number-ai-utils.test.ts
```diff
@@ -21,18 +21,25 @@ const question: PerseusRenderer = {
...
+                        maxError: 0,
+                        simplify: "required",
+                        answerForms: [],
+                        message: "",
+                        strict: true,
+                    },
+                ],
             },
             alignment: "default",
         } as InputNumberWidget,
```
benchristel: We should not need to cast here.

packages/perseus/src/widget-ai-utils/input-number/input-number-ai-utils.test.ts
```diff
@@ -21,18 +21,25 @@ const question: PerseusRenderer = {
...
+                    {
+                        status: "correct",
+                        value: 0.5,
+                        maxError: 0,
+                        simplify: "required",
+                        answerForms: [],
+                        message: "",
+                        strict: true,
+                    },
+                ],
```
benchristel, line 41: Could use the generator functions.

packages/perseus/src/widgets/input-number/input-number.test.ts
```diff
@@ -137,21 +140,32 @@ describe("input-number", function () {
...
+                                    answerForms: [
+                                        "integer",
+                                        "proper",
+                                        "improper",
+                                        "mixed",
+                                    ],
+                                    message: "",
+                                    strict: true,
+                                },
+                            ],
```
benchristel, line 161: Could use the generator functions.

packages/perseus/src/widgets/input-number/input-number.testdata.ts
benchristel, line 1: This file doesn't have many usages. The data in it could be inlined into the tests that need it, and maybe replaced with generation function calls.

packages/perseus/src/widgets/input-number/input-number.tsx
```diff
@@ -1,300 +1,7 @@
-import {
```
SonicScrewdriver, line 1: ❤️

packages/perseus/src/widget-ai-utils/input-number/input-number-ai-utils.ts
```diff
@@ -41,10 +35,7 @@ export const getPromptJSON = (
 ): InputNumberPromptJSON => {
     return {
         type: "input-number",
-        options: {
-            simplify: widgetData.simplify,
-            answerType: widgetData.answerType,
-        },
+        label: widgetData.labelText,
```
SonicScrewdriver, line 38: I do wonder about removing `simplify` and `answerType` from Khanmigo, as that seems like helpful information for it to have in order to support student questions. I suspect VERY few of our Numeric Inputs have the `labelText` set, and obviously none of our our Input Numbers will have it either.

I can see that you've just brought this inline to match Numeric Input, but I wonder if we should check in with the Khanmigo team to see if we should be providing these values for both Input Number and Numeric Input.

packages/perseus-score/src/widgets/input-number/score-input-number.test.ts
```diff
@@ -212,13 +243,22 @@ describe("scoreInputNumber", () => {
     });
 
     it("should consider decimals as the thousands separator in FR locale", () => {
+        // TODO(benchristel): This test seems wrong. The correct answer is 16.5, but
+        // 16.500,00 is accepted.
```
SonicScrewdriver, line 241: Ooooo that's a good find. I recall that the decimal separator logic is a little rough between languages, but I have a feeling we can fix this! :)

packages/perseus-core/src/parse-perseus-json/perseus-parsers/input-number-widget.ts
```diff
@@ -46,8 +99,100 @@ export const parseInputNumberWidget = parseWidget(
...
+        rightAlign: inputNumberOptions.rightAlign,
+        size: inputNumberOptions.size,
+        answers: [
+            {
+                status: "correct",
+                value: Number(inputNumberOptions.value),
+                simplify: inputNumberOptions.simplify,
+                message: "",
+                maxError: getMaxError(inputNumberOptions),
+                strict: true,
```
SonicScrewdriver, line 120: I thought we had strict set to `false` before, did you find something that indicated that it would be better to change it to `true`?

SonicScrewdriver (commented):
SO STOKED for this! It looks great :) 

I just have one blocking comment to check in on the shift of the default of "strict" from `false` to `true`, as I want to make sure this was intentional. I also left a non-blocking comment about supporting Khanmigo.

packages/perseus-core/src/parse-perseus-json/perseus-parsers/input-number-widget.ts
benchristel: It's been true from the beginning of the project. Here it is on main: https://github.com/Khan/perseus/blob/9c8873aab0b04775ff4aa45cabff8e1368254933/packages/perseus-core/src/widgets/input-number/to-numeric-input.ts#L24

`strict: false` causes the "default" answer forms to be accepted in addition to the ones explicitly listed. [Here's where that logic lives](https://github.com/Khan/perseus/blob/7fcef1a49d15bc9ba524a39d00b51f88e5e49bec/packages/perseus-score/src/widgets/numeric-input/score-numeric-input.ts#L92-L98) in the numeric-input scoring code:

```ts
        // When an answer is set to strict, we validate using ONLY
        // the provided answerForms. If strict is false, or if there
        // were no provided answer forms, we will include all
        // of the default answer forms in our validator.
        if (!answer.strict || validatorForms.length === 0) {
            validatorForms.push(...defaultAnswerForms);
        }
```

InputNumber never had any equivalent of `strict` — that is, it always used an explicit allowlist of answer forms. There was no way to also allow the "default" set of forms. Here's the original code for input-number answer forms: https://github.com/Khan/perseus/blob/4e74d94fc0356d6769e7750ad301e3c9c6b08bc8/packages/perseus-score/src/widgets/input-number/score-input-number.ts#L70

SonicScrewdriver: Oh weird! I was sure it was false. Okay wonderful! My apologies for the confusion

SonicScrewdriver (approved):

PR #3763, opened 2026-06-11
[Image] | [GIF] |  Turn off pause/play experience for single frame gif images.

Evelas78:
## Summary:
Added checks to see if a gif is a single frame gif, then turn off the gif pause and play if so

Issue: https://khanacademy.atlassian.net/browse/LEMS-4234

## Test plan:
- Verified that single frame gifs have no pause/play button in storybook


packages/perseus/src/widgets/image/components/explore-image-modal-content.tsx
```diff
@@ -108,14 +113,17 @@ export default function ExploreImageModalContent({
...
+                            }
                         />
                     )}
                 </AssetContext.Consumer>
             </div>
             <div
                 className={`perseus-image-modal-description ${styles.modalDescriptionContainer}`}
             >
-                {imageIsGif && (
+                {isAnimatedGif && imageIsGif && (
```
catandthemachines: Could we remove the `imageIsGif` in favor of just `isAnimatedGif`, because based on this logic it already includes `imageIsGif` in the boolean 😉

packages/perseus/src/widgets/image/components/explore-image-modal.test.tsx
```diff
@@ -68,6 +79,21 @@ describe("ExploreImageModal", () => {
 
         unmockImageLoading = mockImageLoading();
 
+        // Make gifuct-js return two fake frames (100ms total loop).
+        // eslint-disable-next-line no-restricted-syntax
+        (parseGIF as jest.Mock).mockReturnValue({});
```
catandthemachines: Can we move this into a `beforeEach` for the gif controls describe section? Or do you now need it for all tests?

packages/perseus/src/widgets/image/components/image-info-area.tsx
catandthemachines: If we have `isAnimatedGif` I would say we no longer need to check whether this image is a gif, because that flag will only be true if it's a gif and longer than one fram.

packages/perseus/src/components/gif-image.tsx
catandthemachines: Idea: what if we instead create a context provider, or call this function from image.tsx (if the image is a `.gif` file) and then pass this value down? I ask because there's a lot of checking whether the image is a gif and whether it has multiple frames in several areas, and it's being passed up from this file and distributed among others. I would prefer this starting at the parent component and disseminating it down.

packages/perseus/src/widgets/image/image.test.ts
```diff
@@ -40,6 +56,21 @@ describe.each([[true], [false]])("image widget - isMobile(%j)", (isMobile) => {
 
         unmockImageLoading = mockImageLoading();
 
+        // Make gifuct-js return two fake frames (100ms total loop).
```
catandthemachines: Same question, is this required for the gif controls tests or all image tests?

nishasy (commented):
~~Not done reviewing yet, but I figured I'd leave these comments before leaving for the day.~~

Oops, they were repeats of Cat's comments.

packages/perseus/src/widgets/image/components/explore-image-modal-content.tsx
Evelas78: Got it

packages/perseus/src/widgets/image/components/explore-image-modal.test.tsx
Evelas78: Hey Cat, so specifically in this file, it works fine, but for the image test file, a lot of other tests render a gif (one example would be in the zoom feature tests, "does not render a zoom image button for gif images") so moving the mock lines to a beforeEach for gif controls doesn't work without adding multiple beforeEach's

Should we aim for a generalized system where we just keep it at the top in case for gif tests or should we just add a beforeEach to each section that may use a gif?

packages/perseus/src/components/gif-image.tsx
Evelas78: I'll definitely look into that, would make everything a little more clear and layers more separated

packages/perseus/src/widgets/image/utils.ts
```diff
@@ -7,6 +10,22 @@ export function isGif(url: string): boolean {
     return url.endsWith(".gif");
 }
 
+/**
+ * Fetches a GIF and decodes it into individual frames using gifuct-js.
+ * Returns the parsed frames array, which includes per-frame pixel data,
+ * delay, dimensions, and disposal type.
+ */
+export async function decodeGifFrames(src: string): Promise<ParsedFrame[]> {
```
catandthemachines, line 18: nit: could we add 1-2 tests on this decodeGifFrames function?

packages/perseus/src/widgets/image/image.tsx
```diff
@@ -47,16 +53,34 @@ export const ImageComponent = (props: ImageWidgetProps) => {
...
+        }
+        setImageGif(true);
+        let mounted = true;
+        decodeGifFrames(url).then((frames) => {
+            if (mounted) {
+                setGifFrames(frames);
+            }
+        });
+        return () => {
+            mounted = false;
```
catandthemachines, line 74: Clarification, what edge case is this solving for? Is it for instances where the backgroundImage.url has changed during the async call do decode frames?

catandthemachines (approved):
Approved, have a few comments for clarification and additional tests, but nothing blocking.

packages/perseus/src/widgets/image/image.tsx
Evelas78: Yeah, thats the exact reason, it had the same thing gif-image.tsx, just had to migrate this to image.tsx if we wanted to grab the data for the frames before passing it downwards

packages/perseus/src/widgets/image/utils.ts
Evelas78: Shouldn't be a problem, depends ofc. What kind of things to do we want to test for for this function?

